PR #5850, opened 2016-06-03
Clear entityInsertions for specific entityName

ricohumme:
See https://github.com/doctrine/doctrine2/issues/5849

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2391,7 +2391,7 @@ class UnitOfWork implements PropertyChangedListener
             $this->orphanRemovals = array();
         } else {
             $this->clearIdentityMap($entityName);
-            $this->clearIdentityInsertions($entityName);
+            $this->clearEntityInsertions($entityName);
         }
 
         if ($this->evm->hasListeners(Events::onClear)) {
@@ -3476,7 +3476,7 @@ class UnitOfWork implements PropertyChangedListener
     /**
      * @param $entityName
      */
-    private function clearIdentityInsertions($entityName)
+    private function clearEntityInsertions($entityName)
     {
         foreach ($this->entityInsertions as $hash => $entity) {
             if (get_class($entity) === $entityName) {
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2390,24 +2390,8 @@ class UnitOfWork implements PropertyChangedListener
             $this->visitedCollections =
             $this->orphanRemovals = array();
         } else {
-            $visited = array();
-
-            foreach ($this->identityMap as $className => $entities) {
-                if ($className !== $entityName) {
-                    continue;
-                }
-
-                foreach ($entities as $entity) {
-                    $this->doDetach($entity, $visited, false);
-                }
-            }
-
-            foreach ($this->entityInsertions as $hash => $entity) {
-                if (get_class($entity) != $entityName) {
-                    continue;
-                }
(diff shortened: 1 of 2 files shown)


Ocramius:
Requires test cases

ricohumme:
Test case provided

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2401,6 +2401,13 @@ public function clear($entityName = null)
                     $this->doDetach($entity, $visited, false);
                 }
             }
+
+            foreach ($this->entityInsertions as $hash => $entity) {
+                if (get_class($entity) != $entityName) {
```
Ocramius: Strict comparison needed here

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2401,6 +2401,13 @@ public function clear($entityName = null)
...
+                    continue;
```
Ocramius: In this scenario, using `continue` is more confusing, IMO. `unset()` can live inside the `if()` if you don't negate the comparison

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2401,6 +2401,13 @@ public function clear($entityName = null)
                     $this->doDetach($entity, $visited, false);
                 }
             }
+
+            foreach ($this->entityInsertions as $hash => $entity) {
```
Ocramius: This block of code (starting at line [2392](https://github.com/doctrine/doctrine2/pull/5850/files#diff-6e8c1c1e78b054ba05e20ea09d877865R2392)) seems to become too complex. Can you move it all into its own private method?

Ocramius:
> Test case provided

Thanks! \o/

lib/Doctrine/ORM/UnitOfWork.php
ricohumme: I could ofcourse, but those are not my contributions. 
Would you like me to alter that ?

lib/Doctrine/ORM/UnitOfWork.php
ricohumme: What do you mean by this?

Ocramius: `===` or `!==`. Never rely on `==` and `!=` unless strictly required to do so

lib/Doctrine/ORM/UnitOfWork.php
Ocramius: Yes, please :-)

ricohumme: Fixed in latest commit

lib/Doctrine/ORM/UnitOfWork.php
ricohumme: Fixed in latest commit

lib/Doctrine/ORM/UnitOfWork.php
ricohumme: Fixed in latest commit

Ocramius: @ricohumme no need to comment on those: they disappear from the diff as soon as you push :-)

Ocramius:
Merged, thanks!

`2.5` - doctrine/doctrine2@0af9ee0140e5ac758e4a1011721810768ceb11cd
`master` - doctrine/doctrine2@3bc61d5f5e4aa75479dfe141e52326243972f530

ricohumme:
Hi,

When will 2.5.5 be released? I'm waiting for this one for one of my projects,

Thnx in advance

PR #5853, opened 2016-06-05
Fix typo

lcobucci:


lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
```diff
@@ -217,7 +217,7 @@ public function getRegion(array $cache)
 
             if ( ! $this->fileLockRegionDirectory) {
                 throw new \LogicException(
-                    'If you what to use a "READ_WRITE" cache an implementation of "Doctrine\ORM\Cache\ConcurrentRegion" is required, ' .
+                    'If you want to use a "READ_WRITE" cache an implementation of "Doctrine\ORM\Cache\ConcurrentRegion" is required, ' .
                     'The default implementation provided by doctrine is "Doctrine\ORM\Cache\Region\FileLockRegion" if you what to use it please provide a valid directory, DefaultCacheFactory#setFileLockRegionDirectory(). '
```
fsevestre: Same typo next line (starting with `The default implementation provided by doctrine is...`)

fsevestre:
It seems that you have to fix some tests

lcobucci:
Already fixed 😉

lcobucci:
But we have failing tests that are unrelated with this change (only for PHP 7) will also fix it here...

lcobucci:
Actually it just happens on PHP 7.0.7 (cool!)

lcobucci:
The problem here is that `DefaultQueryCache` is trying to fetch the `country` property from `City` instead of the `State`.

As you can see [here](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Cache/DefaultQueryCache.php#L266) the `$entity` is always the same (`City` in this case) while `$metadata` points to an association of the `$entity` (`State`).

@Ocramius how should we tackle this problem? Maybe merging this PR and opening a new issue?

Ocramius:
Unrelated failures are not your problem ;-)

lcobucci:
> lated failures are not your problem ;-)

Nice, I'll create an issue about that bug.

Ocramius:
Merged, will go into 2.6.0 only though, as it is not really critical :-)

Thanks, @lcobucci!

lcobucci:
Nice, is indeed way far from being critical HAHAHA
Thanks @Ocramius 😉

Issue #5854, opened 2016-06-05
Default query cache using wrong object


lcobucci:
On the fix of PHP 7 [bug](https://bugs.php.net/bug.php?id=72174), `ReflectionProperty#getValue()` is now raising a `Notice` when accessing an undefined property. That change revealed a false positive on ORM tests!

As you can see [here](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Cache/DefaultQueryCache.php#L266) the `DefaultQueryCache` is always trying to fetch data from `$entity` even when the association is not with that object (the test is failing because `DefaultQueryCache` is trying to get the `country` from the `City` instead of the `State`).

**How to reproduce:** execute the test suite while running PHP 7.0.7

Ocramius:
@lcobucci can you take a look at #5856? It fixes the issue, but also disables some of the SLC caching happening during hydration of multi-level DQL fetch-joins.

lcobucci:
@Ocramius I totally forgot to do a follow up on this issue here.

I'd say it's fixed, let's close it?

PR #5856, opened 2016-06-05
#5854 workaround to avoid populating Second Level Cache from DQL queries with multiple nested DQL aliases

Ocramius:
Provides workaround for #5854

Doesn't really fix the issue though.

The problem is following:

``` sql
SELECT
    a, b, c
FROM
    A a
JOIN
    a.b b
JOIN
    b.c c
```

Caching a query like this one via SLC will have the `DefaultQueryCache` loop over associations of `A`, but using metadata of all DQL aliases/associations.

This workaround prevents that, but also disables caching of fetched `c` instances from this example.

Note: no tests provided, as existing tests already show the regression.


.travis.yml
```diff
@@ -5,6 +5,7 @@ php:
   - 5.5
   - 5.6
   - 7.0
+  - 7.1
```
Ocramius: Hmm, not yet available. I guess `nightly` would be an alternative

lcobucci: Indeed

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -261,6 +261,14 @@ public function put(QueryCacheKey $key, ResultSetMapping $rsm, $result, array $h
             // @TODO - move to cache hydration components
             foreach ($rsm->relationMap as $alias => $name) {
                 $metadata = $this->em->getClassMetadata($rsm->aliasMap[$rsm->parentAliasMap[$alias]]);
+                $className = $metadata->getName();
```
lcobucci: Shouldn't this change impact on https://github.com/doctrine/doctrine2/blob/master/tests/Doctrine/Tests/ORM/Cache/DefaultQueryCacheTest.php#L138?

Ocramius: I think the design of the test is broken. The current assertions pass, but further details are indeed skipped

lcobucci:
I think that we also have a small thing on https://github.com/doctrine/doctrine2/blob/master/tests/Doctrine/Tests/ORM/Cache/DefaultQueryCacheTest.php#L134. Shouldn't it be `$state->getName()` instead of `$city->getName()`?

Ocramius:
Yep, that needs fixing too.

lcobucci:
I believe that we can fix those small problems in order to prevent that `Notice` and then think on a permanent solution.

FabioBatSilva:
@Ocramius I made a couple changes here, hopefully will fix the issue.

Ocramius:
Ok, seems like I'll have to do some `git bisect` on `master`: loads of issues popping up lately...

This patch is good to go meanwhile, thanks @FabioBatSilva!

PR #5866, opened 2016-06-09
Added afterFlush event

simPod:


Ocramius:
Use-case scenario?

simPod:
For example, you cannot persist another entity in postPersist. So you schedule these updates after the lifecycle has finished and dispatch it all from the queue

Issue #4097, opened 2014-09-10
DDC-3303: @ORM\Embedded does not work with extending classes


doctrinebot:
Jira issue originally created by user TheBelgarion:

I update my symfony 2.5 with ORM 2.5. to try out the Embedded function.
I currently does not work correctly with extending class
```
/****
 * @ORM\Entity
 * @ORM\Table(name="A", uniqueConstraints={@UniqueConstraint(name="id", columns={"id"})})
 */
class A extends B {}
```

```
abstract class Item {

    /****
     * @var $bonus
     * @ORM\Embedded(class="Bonus")
     *
     */
    private $bonus;
}
```

it only works if i put the definition of the Embedded properties directly in class A ( but it works with all normal properties defined in the Abstract )

martinopic:
Setting the parent class properties to "protected" (instead of private") works and could be an acceptable workaround for some.

Issue #4277, opened 2014-12-30
DDC-3467: Embeddable in inheritance currently must be protected


doctrinebot:
Jira issue originally created by user nemo64:

I have a setup similar to this one :
```
/****
 * @ORM\Table()
 * @ORM\Entity(readOnly=true)
 * @ORM\InheritanceType("SINGLE_TABLE")
 ****/
abstract class Planet
{
    /** ... **/
    /****
     * @var Position
     * @ORM\Embedded(class="Position")
     */
    private $position;
}
```
```
/****
 * @ORM\Entity(readOnly=true)
 ****/
class Sun extends Planet
{
}
```
This causes the following error:
```
[ReflectionException]
Property Sun::$position does not exist
```

This can easily be fixed by making the embeddable protected, but if I didn't miss anything, a private embed should work too.

acanimal:
Any news on this issue?

lcobucci:
This duplicates #4097 right?
I'll work on a test for it now.

acanimal:
@lcobucci It seems so. Thanks !!!

PR #5867, opened 2016-06-09
Allow the usage of embedded objects on parent classes.

lcobucci:
The `ClassMetadataInfo` was always using the "current class" to fetch the reflection of a property even when a field is declared on the parent class (which causes `ReflectionProperty` to throw an exception).

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -941,8 +941,13 @@ class ClassMetadataInfo implements ClassMetadata
                 continue;
             }
 
-            $parentReflFields[$property] = $reflService->getAccessibleProperty($this->name, $property);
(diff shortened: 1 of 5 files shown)


lcobucci:
Fixes #4097 and #4277.

lcobucci:
@Ocramius is this good enough?

Ocramius:
@lcobucci looks good, but can't merge atm (in a hurry)

lcobucci:
Don't worry was just checking if needed some more work ;)

Ocramius:
@lcobucci merged, thanks!

`master`: doctrine/doctrine2@3dfc1807209e131e691cbe782024d84387273d86
`2.5`: doctrine/doctrine2@aa5820309e80e746411100f073b8bf0a0a684082

Note: I inlined all the tests into the TestCase that required them, and removed all unused API

lcobucci:
@Ocramius Nice! Will inline the mapped classes on the next functional tests I may write. Thanks!

PR #5872, opened 2016-06-13
Improved Output in Validation with no Entities

rogerwalt:
Instead of "OK - The mapping files are correct" it now says "FAIL - No
entities defined!".

Maybe someone can help me write a correct test for the
ValidateSchemaCommand, I wasn't able to provide that, mostly because
there is no test for this command yet.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
@@ -76,7 +76,9 @@ EOT
 
         if ($input->getOption('skip-mapping')) {
             $output->writeln('<comment>[Mapping]  Skipped mapping check.</comment>');
-        } elseif ($errors = $validator->validateMapping()) {
+        } elseif (($errors = $validator->validateMapping()) === false) {
+            $output->writeln('<error>[Mapping]  FAIL - No entities defined!</error>');
+        } elseif (count($errors) > 0) {
             foreach ($errors as $className => $errorMessages) {
                 $output->writeln("<error>[Mapping]  FAIL - The entity-class '" . $className . "' mapping is invalid:</error>");
 
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -67,6 +67,10 @@ class SchemaValidator
         $cmf = $this->em->getMetadataFactory();
         $classes = $cmf->getAllMetadata();
 
+        if (count($classes) === 0) {
+            return false;
+        }
+
         foreach ($classes as $class) {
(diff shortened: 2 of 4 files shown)


lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
```diff
@@ -76,7 +76,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         if ($input->getOption('skip-mapping')) {
             $output->writeln('<comment>[Mapping]  Skipped mapping check.</comment>');
-        } elseif ($errors = $validator->validateMapping()) {
+        } elseif (($errors = $validator->validateMapping()) === false) {
+            $output->writeln('<error>[Mapping]  FAIL - No entities defined!</error>');
+        } elseif (count($errors) > 0) {
```
rogerwalt: Now by looking at this I think $error won't be defined here. I should really write a test for the `ValidateSchemaCommand`.

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -67,6 +67,10 @@ public function validateMapping()
         $cmf = $this->em->getMetadataFactory();
         $classes = $cmf->getAllMetadata();
 
+        if (count($classes) === 0) {
+            return false;
```
Ocramius: 1. return type is inconsistent with signature
2. an empty set of mappings is perfectly valid here

lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
Ocramius: I suggest counting `$em->getMetadataFactory()->getAllMetadata()` and dealing with the case where 0 mappings are found directly in the CLI command. The validator context is much more specific, and an empty set of metadata is indeed valid.

tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
```diff
@@ -24,6 +27,22 @@ public function setUp()
...
+        $config->setMetadataDriverImpl(new ZeroEntitiesMappingDriver());
+        $config->setProxyDir(__DIR__ . '/Proxies');
+        $config->setProxyNamespace('Doctrine\Tests\Proxies');
+        $em = $this->_getTestEntityManager(null, $config, null, true);
+
+        // setup new validator for $em
+        $validator = new SchemaValidator($em);
+
+        $result = $validator->validateMapping();
+        $this->assertFalse($result);
```
Ocramius: Should simply be `self::assertEmpty($result)`

Ocramius (changes requested):

PR #5884, opened 2016-06-19
Drop PHP 5.5 support

Ocramius:


Majkl578:
Cool, now is the time for generators! :100:

PR #5860, opened 2016-06-08
Removed hacky switch/case, migrated to if/else and early return statements

peter-gribanov:

PR #5885, opened 2016-06-19
Partially reverting #5860 due to type juggling horrors

Ocramius:
See #5860

See this example on why the revert is needed: https://3v4l.org/8T34v

Code copied for reference:

``` php
<?php

$a = 1;

switch ($a) {
    case "1";
        echo "STUPID LANGUAGE!";
        break;
    case 1;
        echo __LINE__;
        break;
}
```

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -421,14 +421,14 @@ use Doctrine\Common\Util\ClassUtils;
                 return null;
             }
 
-            switch ($lockMode) {
-                case LockMode::OPTIMISTIC:
+            switch (true) {
+                case LockMode::OPTIMISTIC === $lockMode:
                     $this->lock($entity, $lockMode, $lockVersion);
                     break;
 
-                case LockMode::NONE:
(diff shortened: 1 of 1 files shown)


peter-gribanov:
@Ocramius you can force a check on `null`.
```php
$lockMode = is_int($lockMode) ? $lockMode : -1; // -1 is out of range LockModes

switch ($lockMode) {
    case LockMode::OPTIMISTIC:
        $this->lock($entity, $lockMode, $lockVersion);
        break;
    case LockMode::NONE:
    case LockMode::PESSIMISTIC_READ:
    case LockMode::PESSIMISTIC_WRITE:
        $persister = $unitOfWork->getEntityPersister($class->name);
        $persister->refresh($sortedId, $entity, $lockMode);
        break;
}
```

Ocramius:
Not useful then: just increases complexity

peter-gribanov:
@Ocramius then you should at least add a comment to make it clear why it was done so. And then a year later again will be the PR which will break all.

Ocramius:
Tests will break if this is changed anyway :-\

PR #5890, opened 2016-06-20
Allow to clear up UOW only with given entity

ImmRanneft:
When flushing only given entities you are loosing already computed changesSets. This is little bit weird, because you may want to wirte something in very far distant place (like some listener or something like this).

With this changes nothing become borken, and you need no changes in your code, but you are allowed now to keep already computed changesets and save them after you already flushed single (or several) entities.

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -343,7 +343,11 @@ use Doctrine\Common\Util\ClassUtils;
      * If an entity is explicitly passed to this method only this entity and
      * the cascade-persist semantics + scheduled inserts/removals are synchronized.
      *
+     * Also if you want to keep your changeSets for another entities you should explicitly provide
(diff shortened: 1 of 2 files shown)


Ocramius:
This is a "won't fix" directly. Sorry, I'm expressing my veto on it because:

 * this mixes listeners on changesets with an already very messy feature (clearing subsets of the UoW)
 * clearing subsets of the UoW is likely going to die as a feature altogether, as it is fundamentally broken

I want to get rid of `flush($something)` and `clear($something)`, moving to `flush()` and `clear()`. I absolutely don't want to allow further complexity to land into this already bug-ridden functionality (that can't be fixed).

ImmRanneft:
@Ocramius so your advice is just **stop** using this rotten stuff? :smile: Thanks for quick answer, btw.

Ocramius:
Pretty much, yes :-\

Issue #5882, opened 2016-06-17
PHP 5.4 compatibility issues


ic1011:
[Getting Started](http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/tutorials/getting-started.html)

The above page says that Doctrine 2 is PHP 5.4+ compatible but in reality it is not.

The class Doctrine\Common\Proxy\ProxyGenerator contains the following code which requires PHP 5.5 and above:

```
    protected $placeholders = [
        'baseProxyInterface'   => Proxy::class, 
        'additionalProperties' => '',
    ];
```

[PHP 5.5 - New features - class name](http://php.net/manual/en/migration55.new-features.php#migration55.new-features.class-name)

Majkl578:
You can explicitly require older doctrine/common. Version 2.6 (current) requires 5:5+, older version PHP 2.5 requires PHP 5.4+.

ic1011:
Okay. Thank you. Are they interchangeable? Should I expect any major drawbacks in the functionality?
I already have the Domain structure and I am just trying to use it on different machine where the PHP currently cannot be upgraded.

Ocramius:
Docs page indeed needs updating: specifically, removal of explicit requirements, since composer tells you those.
We could simplify it to something like "always use the latest stable version of PHP".

Anyway, this is typically caused by a `composer update` issued on a machine with newer PHP versions, and then deploying via `composer install` on a machine with older PHP versions (tip: don't do that, or specify a platform version in your `composer.json`).

If you want to help out with the wording, check [`getting-started.rst`](https://github.com/doctrine/doctrine2/blob/659f6a386462b9dd1b171c64d9b5bca8ab8cf1a6/docs/en/tutorials/getting-started.rst) and edit it to reflect the wording that you'd prefer.

Ocramius:
Was handled by @BreiteSeite in #5892

PR #5892, opened 2016-06-20
#5882 recommend using latest stable PHP version in the getting started guide

BreiteSeite:
As @Ocramius suggested in #5882: dropping the specific PHP version from the getting started guide.


Ocramius:
@BreiteSeite thanks!

PR #5896, opened 2016-06-21
Pull default mapping values from concrete types

DASPRiD:
This PR requires the https://github.com/doctrine/dbal/pull/2421 PR to be merged first.

It solves the problem where the comparator sees columns as changed when a custom type overrides parameters like length internally when returning the SQL statement to create/update the column.

This feature allows to supply those override values via a specific function, so the values are known to the schema tool and the comparator can compare the actual values. As an example, the following custom type always triggered the comparator to see a column as changed, even though it wasn't:

``` php
class LocaleType extends StringType
{
    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
    {
        return $platform->getVarcharTypeDeclarationSQL([
            'length' => '5',
            'fixed' => true,
        ]);
    }
}
```

In comparison, with the new method for types introduced, one can write an equal type which will work with the comparator:

``` php
class LocaleType extends StringType
{
    public function getDefaultMappingValues()
    {
        return [
            'length' => '5',
            'fixed' => true,
        ];
    }
}
```

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -26,6 +26,7 @@ use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\Visitor\DropSchemaSqlCollector;
 use Doctrine\DBAL\Schema\Visitor\RemoveNamespacedAssets;
+use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Tools\Event\GenerateSchemaTableEventArgs;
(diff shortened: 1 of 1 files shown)


Ocramius:
As per off-github discussion:

This addition assumes that `Type` (and its subclasses) convey default type information of DDL columns, such as length, encoding, nullability, etc.

While the feature is interesting, type information is mostly bound to how platform deal with it, while our type system is designed to produce runtime type conversion.

In order to get DDL type information transfered correctly, the type system always has to strictly communicate with platforms to define what the actual DDL string will look like (and what it should look like, when using the comparator), as well as which modifiers are even available (stuff like `AUTO_INCREMENT`, or `DEFAULT NULL` for LOB are really weird, for example).
As I've said on IRC, the fact that our `Type` works to define column types is quite similar to "luck", sadly.

Therefore, we should probably come up with a way to disentangle the DDL definitions for types:

 * they should be platform-specific
 * they cannot exist only in a `Type` object, unless we make `Type` instances platform-specific (3.x)
 * we may separate type conversion and type DDL specification, eventually (unsure about this one).

@deeky666 I assigned this to you to get your opinion on it, since this is strictly related to doctrine/dbal#2421, which I closed as `incomplete`

deeky666:
I'm -1 for this because of the things @Ocramius pointed out. Introducing something like this will potentially break current custom type implementations and introduce more issues than it solves. We may keep this in mind for a new type system in `3.0` and take this for consideration but for `2.x` I am -1.

DASPRiD:
How exactly may this break the current custom type system though? It's fully backward compatible (as long as you don't override the method, everything acts as before).

Ocramius:
> How exactly may this break the current custom type system though? 

It's mostly a maintenance concern. The entire type system as it is right now is pretty much a jenga tower of platform and static type information. Ref: 
jenga-tower.

If we keep enhancing the type system based on its current (broken) architecture, we'll end up having even more BC concerns here, especially when platform-specific rules are involved.

The patch doesn't break anything, but it adds a drop to an already overflowing glass.

Ocramius:
Closing as `Can't fix`. This becomes relevant once we went through a DBAL 3.x iteration (not happening yet).

PR #5897, opened 2016-06-22
Check for Xcache availability from INI setting

mikaelz:
On a shared webhosting the Xcache module is loaded, but it can be disabled with .htaccess `php_flag xcache.cacher Off`
So module is loaded, but Xcache is disabled via php.ini


Ocramius:
Looked into how to test this one, but there ain't a proper way unless we mock `ini_get()`, and that's just fragile.

Thanks @mikaelz, merging, although only to `2.6.0`.

Issue #5889, opened 2016-06-20
Second Cache Level is not working correctly with "indexBy" on association


orfin:
Hi,

when using indexBy on OneToMany association, returned collection is not indexed by given field.

* getTranslations with 2ndlevel:  ```[0 => Translation, 1 => Translation]```
* getTranslations without 2ndlevel:   ```['en' => Translation, 'pl' => Translation]```

Sample mapping:

```
    /**
     * @ORM\Cache(usage="NONSTRICT_READ_WRITE", region="cms")
     * 
     * @ORM\OneToMany(targetEntity="Translation", mappedBy="page", cascade={"persist", "remove"}, indexBy="language", orphanRemoval=true)
     */
    protected $translations;


    /**
     * Get translations
     *
     * @return \Doctrine\Common\Collections\Collection
     */
    public function getTranslations()
    {
        return $this->translations;
    }
```

lcobucci:
@Ocramius as per @FabioBatSilva's review on https://github.com/doctrine/doctrine2/pull/5899#issuecomment-228787124 I'd say we close this and improve the documentation, what do you think?

orfin:
> I don't think that adding indexes to collections is something we want to support.

We have many projects with a vast number of associations with indexBy attribute. That is cool feature of doctrine, because it improves performance on PHP-level - we don't have to iterate over a whole array to get an element.

Removing ability to use indexBy with 2nd level cache makes this mechanizm totally unusable in our applications. We had to return to old-style ResultCache due to this limitation but would be more than happy to return to 2nd level cache :-)

lcobucci:
@orfin the problem here is a different thing...

The ORM assumes that you have valid state on your objects even without the persistence layer. Let's consider that everything is running in memory and use your models from #5889.

If you do:

```php
$page = new DDC5889PageModel('phpunit');
$translation1 = new DDC5889TranslationModel($page, 'phpunit-1', 'en');
$translation2 = new DDC5889TranslationModel($page, 'phpunit-2', 'pl');

$page->getTranslations()->add($translation1);
$page->getTranslations()->add($translation2);

assert($page->getTranslations()->get('en') === $translation1);
assert($page->getTranslations()->get('pl') === $translation2);
```

The script will fail since the collection doesn't know how do you want to index things in it (you didn't tell it).

On the other hand, everything works with:

```php
$page = new DDC5889PageModel('phpunit');
[...]

PR #5899, opened 2016-06-23
respect indexBy when using second level cache

BreiteSeite:
Hello,

this fixes issue #5889.

As this is my first contribution to doctrine, please review thoroughly. Any feedback is very welcome.

Things to consider:
- Should i rename the functional test from DDC5889 to GH5889?
- I would prefer to already write the correct index to the cache (because written to the cache may occur less often then reading from it). However, there is [code which expects the index from the cache keys to be the same as the index from the initial collection](https://github.com/BreiteSeite/doctrine2/blob/ff4e02f202a6977492ed530b435fb28e828d3d8b/lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php#L193). Thats why "reindex" occurs on reading from the cache now. As a small positive side effect: this means the caching adapter does not have to support indexes. Also no need for cache purging for metadata changes when indexing occurs on the read-side.
- If you agree that the implementation is okay so far, i might add a unit test for the DefaultCollectionHydrator class if you want.

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultCollectionHydrator.php
+++ b/lib/Doctrine/ORM/Cache/DefaultCollectionHydrator.php
@@ -87,9 +87,20 @@ class DefaultCollectionHydrator implements CollectionHydrator
             return null;
         }
 
-        /* @var $entityEntries \Doctrine\ORM\Cache\EntityCacheEntry[] */
-        foreach ($entityEntries as $index => $entityEntry) {
-            $list[$index] = $this->uow->createEntity($entityEntry->class, $entityEntry->resolveAssociationEntries($this->em), self::$hints);
+        if (isset($assoc['indexBy'])) {
+            /* @var $entityEntries \Doctrine\ORM\Cache\EntityCacheEntry[] */
+            foreach ($entityEntries as $entityEntry) {
+                $entity = $this->uow->createEntity($entityEntry->class, $entityEntry->resolveAssociationEntries($this->em), self::$hints);
+
+                $entityClassMetadata = $this->em->getClassMetadata($entityEntry->class);
+                $index = $entityClassMetadata->getFieldValue($entity, $assoc['indexBy']);
+                $list[$index] = $entity;
+            }
+        } else {
+            /* @var $entityEntries \Doctrine\ORM\Cache\EntityCacheEntry[] */
+            foreach ($entityEntries as $index => $entityEntry) {
+                $list[$index] = $this->uow->createEntity($entityEntry->class, $entityEntry->resolveAssociationEntries($this->em), self::$hints);
+            }
         }
 
         array_walk($list, function($entity, $index) use ($collection) {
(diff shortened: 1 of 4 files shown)


FabioBatSilva:
Hi @BreiteSeite,

This seems invalid to me, (probably something we should add to known limitations)
Your are creating a collection with numeric indexes and expects a collection indexed by ``language``

You should be using ``Collection#set`` to create a indexed collection;

```
$page->getTranslations()->set('en', new DDC5889TranslationModel($page, 'phpunit-1', 'en'));
$page->getTranslations()->set('pl', new DDC5889TranslationModel($page, 'phpunit-2', 'pl'));
```

BreiteSeite:
Hi @FabioBatSilva,

i'm not sure what do you mean exactly. You seem to refer to the functional test. In this test, i append two entities to an collection and persist them. After that, i retrieve them again (with second level cache enabled) and expect them to be indexed by language, cause this is [specified in the association mapping](https://github.com/BreiteSeite/doctrine2/blob/ff4e02f202a6977492ed530b435fb28e828d3d8b/tests/Doctrine/Tests/Models/DDC5889/DDC5889PageModel.php#L23).

I don't explicitly set them indexed because adding them to the collection will probably also won't set them to a specific index.

FabioBatSilva:
Your test its OK, I don`t think the issue is valid.

SLC aims to be transparent but that is not aways possible.

Collections are not able to index on the fly.
SLC relies on the PHP object that represents the collection to store its cache.

Using ``add`` instead of ``set`` its inconsistent,
If you create a new collection using ``add`` and then try to use it without reloading from the DB you won`t have the index.
In this case SLC accentuates a issue that is already there.

BreiteSeite:
> If you create a new collection using add and then try to use it without reloading from the DB you won`t have the index.

I can see your point, but how is this related to SLC/this PR? This issue exists already if you use the indexBy annotation and add entries to the collection via `Collection#add` instead of `Collection#set`, isn't it?

So what would be needed is something like a collection which indexes by a property from the entities that are added so behavior is not coupled to the persistence logic.

Would that be an acceptable solution in your opinion?

FabioBatSilva:
SLC will take whatever collection you give it and store as the association cache.
As your collection is created without indexes you end up  with a cache that reflects that.

The only reason it works without SLC is that the collection index gets rebuild when you reload it.

I don't think that adding indexes to collections is something we want to support.

IMO the solution here is to add this [known limitations](https://github.com/doctrine/doctrine2/blob/master/docs/en/reference/second-level-cache.rst#limitations) and simple use ```Collection#set``` instead of ```Collection#add``` when you create your indexed collections.

BreiteSeite:
Hey @FabioBatSilva, so we gonna close the PR as invalid?

lcobucci:
@BreiteSeite yes indeed, this PR is invalid since this works fine:

```php
$page = new DDC5889PageModel('phpunit');

$translation1 = new DDC5889TranslationModel($page, 'phpunit-1', 'en');
$translation2 = new DDC5889TranslationModel($page, 'phpunit-2', 'pl');

$page->getTranslations()->add($translation1);
$page->getTranslations()->add($translation2);

$this->_em->persist($page);
$this->_em->flush();
$this->_em->clear();

$fetchedPage = $this->_em->find(DDC5889PageModel::class, $page->getId());
$translations = $fetchedPage->getTranslations();

self::assertCount(2, $translations, 'Page does contain translations');
self::assertEquals($translation1, $translations->get(0));
self::assertEquals($translation2, $translations->get(1));
```

PR #5903, opened 2016-06-27
Add keyword aware quote strategy

Deltachaos:
Related to the discussion in doctrine/dbal#1592 it is probably a good idea to add this quote strategy. It can be enabled by the user optionally and therefore he should be aware of the upercase/lowercase issue in some DBMS.

For the integration of this into doctrine-bundle see this PR: doctrine/DoctrineBundle#553


Ocramius:
@guilhermeblanco can you take a look at this?

Ocramius:
@Deltachaos tests are missing tho

Deltachaos:
@Ocramius will add tests as soon as is see that there is a change that this gets merged

lib/Doctrine/ORM/Mapping/KeywordQuoteStrategy.php
```diff
@@ -0,0 +1,197 @@
...
+     *
+     * @return string
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    public function getQuotedName(AbstractPlatform $platform, $name, $force = false)
+    {
+        $keywords = $platform->getReservedKeywordsList();
+        $parts = explode(".", $name);
+        foreach ($parts as $k => $v) {
+            $parts[$k] = ($force || $keywords->isKeyword($v)) ? $platform->quoteIdentifier($v) : $v;
```
Majkl578: Unnecessary write of the unmodified value here, rather use normal if.

lib/Doctrine/ORM/Mapping/KeywordQuoteStrategy.php
```diff
@@ -0,0 +1,197 @@
...
+     * @param string $name
+     * @param bool $force
+     *
+     * @return string
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    public function getQuotedName(AbstractPlatform $platform, $name, $force = false)
+    {
+        $keywords = $platform->getReservedKeywordsList();
+        $parts = explode(".", $name);
```
Majkl578: `"` -> "'"

lib/Doctrine/ORM/Mapping/KeywordQuoteStrategy.php
```diff
@@ -0,0 +1,197 @@
...
+{
+   /**
+     * @param AbstractPlatform $platform
+     * @param string $name
+     * @param bool $force
+     *
+     * @return string
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    public function getQuotedName(AbstractPlatform $platform, $name, $force = false)
```
Majkl578: Is there a reason to make this public?

Majkl578:
It'd be really good to have a native way to quote reserved words. 👍

lib/Doctrine/ORM/Mapping/KeywordQuoteStrategy.php
```diff
@@ -0,0 +1,197 @@
...
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+/**
+ * A set of rules for determining the physical column, alias and table quotes
+ *
+ * @since   2.6
+ * @author  Maximilian Ruta <mr@xtain.net>
+ */
+class KeywordQuoteStrategy implements QuoteStrategy
```
Majkl578: I think _keyword_ is may be a bit misleading, what about e.g. _ReservedWordsQuotingStrategy_?

lib/Doctrine/ORM/Mapping/KeywordQuoteStrategy.php
Deltachaos: not realy

Deltachaos:
@Majkl578 tests are added and class was renamed

mikeSimonson:
@Deltachaos Do you still plan on fixing the broken tests?

Deltachaos:
@mikeSimonson yes i have this still on my list. But i can not guarantee to fix this within the next month. Probably start of next year.

mikeSimonson:
@Deltachaos Thanks

billschaller:
Quoting is deeply wtf in DBAL - 3.0 needs to address more than just reserved word quoting...

Ocramius:
@Deltachaos I think we're going for quoting-by-default in ORM 3.x, so I'm not sure if this will be required by then... /cc @guilhermeblanco just to confirm.

guilhermeblanco:
@Ocramius everything is quoted (and cannot be turned off) in `master`.
QuoteStrategy support is also removed from ORM 3.X.

Ocramius:
Closing as `duplicate` then.

PR #5904, opened 2016-06-28
Fixed overwriting explicit cache namespace

JanJakes:
When trying to use explicit namespace revision (i.e. revision as cache namespace: `$cache->setNamespace($revision)`), Doctrine overwrites the revision by some default namespace derived from proxy dir (which in environments like Docker does not differ from revision to revision).

### DIFF
--- a/lib/Doctrine/ORM/Tools/Setup.php
+++ b/lib/Doctrine/ORM/Tools/Setup.php
@@ -145,7 +145,7 @@ class Setup
             $cache = new ArrayCache();
         }
 
-        if ($cache instanceof CacheProvider) {
+        if ($cache instanceof CacheProvider && $cache->getNamespace() === '') {
             $cache->setNamespace("dc2_" . md5($proxyDir) . "_"); // to avoid collisions
         }
 
--- a/tests/Doctrine/Tests/ORM/Tools/SetupTest.php
+++ b/tests/Doctrine/Tests/ORM/Tools/SetupTest.php
@@ -71,6 +71,22 @@ class SetupTest extends OrmTestCase
         $this->assertInstanceOf('Doctrine\ORM\Mapping\Driver\YamlDriver', $config->getMetadataDriverImpl());
     }
 
+    public function testConfigureCacheNamespace()
+    {
+        // automatically set namespace
+        $config = Setup::createConfiguration(false, '/foo');
+
+        $this->assertSame('dc2_1effb2475fcfba4f9e8b8a1dbc8f3caf_', $config->getMetadataCacheImpl()->getNamespace());
+
+
+        // manually set namespace
+        $cache = new ArrayCache();
+        $cache->setNamespace('foo');
+        $config = Setup::createConfiguration(false, '/foo', $cache);
+
+        $this->assertSame('foo', $config->getMetadataCacheImpl()->getNamespace());
+    }
+
     /**
      * @group DDC-1350
      */


lib/Doctrine/ORM/Tools/Setup.php
```diff
@@ -145,7 +145,7 @@ public static function createConfiguration($isDevMode = false, $proxyDir = null,
             $cache = new ArrayCache();
         }
 
-        if ($cache instanceof CacheProvider) {
+        if ($cache instanceof CacheProvider && $cache->getNamespace() === '') {
```
Ocramius: `=== ''` is not needed. Tests required tho

JanJakes:
Added tests.

r3ntg3n:
Hi there.

Is there a way to upvote this Pull Request to be accepted? We're facing the same issue, but it has more dramatic impact for us.

We use Doctrine for two separate application. These applications have shared Memcache cluster and same `proxyDir` configuration value. Applications have several entities with same names, but totally different relations.

So, as you may already guess, well you clear metadata cache from one application, actually another application's metadata cache is cleared as well.

And, as soon as any application regenerates it's metadata cache, the other application becomes broken, because cached metadata does not match actual entities and source metadata.

thekia19:
Overriding the namespace is very confusing. And breaks the initial purpose of that - keeping more than one application separated from each other.
Please take a look on pull request.

Thanks

tests/Doctrine/Tests/ORM/Tools/SetupTest.php
```diff
@@ -71,6 +71,22 @@ public function testYAMLConfiguration()
         $this->assertInstanceOf('Doctrine\ORM\Mapping\Driver\YamlDriver', $config->getMetadataDriverImpl());
     }
 
+    public function testConfigureCacheNamespace()
```
lcobucci: ` : void` missing

tests/Doctrine/Tests/ORM/Tools/SetupTest.php
```diff
@@ -71,6 +71,22 @@ public function testYAMLConfiguration()
...
+    {
+        // automatically set namespace
+        $config = Setup::createConfiguration(false, '/foo');
+
+        $this->assertSame('dc2_1effb2475fcfba4f9e8b8a1dbc8f3caf_', $config->getMetadataCacheImpl()->getNamespace());
```
lcobucci: `self::assert*` is preferable since the method is static

tests/Doctrine/Tests/ORM/Tools/SetupTest.php
```diff
@@ -71,6 +71,22 @@ public function testYAMLConfiguration()
...
+
+
+        // manually set namespace
+        $cache = new ArrayCache();
+        $cache->setNamespace('foo');
+        $config = Setup::createConfiguration(false, '/foo', $cache);
+
+        $this->assertSame('foo', $config->getMetadataCacheImpl()->getNamespace());
```
lcobucci: `self::assert*` is preferable since the method is static

tests/Doctrine/Tests/ORM/Tools/SetupTest.php
```diff
@@ -71,6 +71,22 @@ public function testYAMLConfiguration()
...
+    public function testConfigureCacheNamespace()
+    {
+        // automatically set namespace
+        $config = Setup::createConfiguration(false, '/foo');
+
+        $this->assertSame('dc2_1effb2475fcfba4f9e8b8a1dbc8f3caf_', $config->getMetadataCacheImpl()->getNamespace());
+
+
+        // manually set namespace
+        $cache = new ArrayCache();
```
lcobucci: It would be better if we move this to another test case, to make things more explicit

lib/Doctrine/ORM/Tools/Setup.php
lcobucci: I think that we should use the same approach of L2C where we always change change the namespace but using the current namespace as prefix, like:

https://github.com/doctrine/doctrine2/blob/b7cace86a078feb9d6da4f69fa4a49b99b353258/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php#L234-L251

lcobucci (changes requested):
The general idea LGTM to me, I'll be applying some changes to get this merged ASAP!

lib/Doctrine/ORM/Tools/Setup.php
lcobucci: not necessarily generating a new instance, though

lcobucci:
> Overriding the namespace is very confusing. And breaks the initial purpose of that - keeping more than one application separated from each other.

@thekia19 this is a generic way to setup the ORM, you can (should IMO) create the configuration object by yourself and fine tune things the way you need. I'd say that we should override the namespace to avoid conflicts but the original namespace MUST be considered.

lcobucci:
@JanJakes 🚢 and backported to 2.5! Thanks for your contribution

JanJakes:
@lcobucci Thanks!

PR #5911, opened 2016-07-02
Scrutinizer clear code

ReenExe:
only UnitOfWork

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1566,11 +1566,9 @@ class UnitOfWork implements PropertyChangedListener
     {
         $stringIdHash = (string) $idHash;
 
-        if (isset($this->identityMap[$rootClassName][$stringIdHash])) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1155,9 +1155,9 @@ private function getCommitOrder(array $entityChangeSet = null)
                 }
 
                 $joinColumns = reset($assoc['joinColumns']);
-                $isNullable  = isset($joinColumns['nullable']) ? $joinColumns['nullable'] : false;
+                $weight  = (int)empty($joinColumns['nullable']);
```
Ocramius: Consider inlining this into `$calc->addDependency(`, since the variable isn't used anywhere else.

ReenExe: @Ocramius, ok

Ocramius:
@ReenExe thanks!

Note: please be more specific with commit messages in further PRs - the current commit messages don't really explain what/why something was done :-)

PR #5918, opened 2016-07-05
Make setSQL and setDQL function calls case-sensitive.

tonynelson19:


Ocramius:
@tonynelson19 thanks!

PR #5924, opened 2016-07-07
Allow empty string identifiers

Ocramius:
This PR fixes a subtle bug that disallowed the UnitOfWork from containing entities with a `''` identifier.

Note: this fix targets ORM 2.6, because empty identifiers are still subject to bug https://github.com/doctrine/doctrine2/issues/5923


deeky666:
*hiding*

Ocramius:
@deeky666 thanks!

PR #5926, opened 2016-07-07
Fix #5923 - disambiguate identifier hashing

### LINKED ISSUE #5923: UnitOfWork does not distinguish between ['', ' '] and [' ', ''] when hashing identifiers
I haz zeh perfect database table that haz 1 composite identifier and contains zeh values `['', ' ']` and `[' ', '']` and now Doctrine explodes cause it cannot generate different identifier hashes for dat :((((

Simply another edge case for #5922.

Ocramius:
Fixes #5923

Requires #5924 to be merged first.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1402,7 +1402,7 @@ class UnitOfWork implements PropertyChangedListener
             throw ORMInvalidArgumentException::entityWithoutIdentity($classMetadata->name, $entity);
         }
 
-        $idHash    = implode(' ', $identifier);
+        $idHash    = $this->hashIdentifier($identifier);
         $className = $classMetadata->rootEntityName;
 
         if (isset($this->identityMap[$className][$idHash])) {
@@ -1512,7 +1512,7 @@ class UnitOfWork implements PropertyChangedListener
     {
         $oid           = spl_object_hash($entity);
         $classMetadata = $this->em->getClassMetadata(get_class($entity));
-        $idHash        = implode(' ', $this->entityIdentifiers[$oid]);
+        $idHash        = $this->hashIdentifier($this->entityIdentifiers[$oid]);
 
         if ($idHash === '') {
             throw ORMInvalidArgumentException::entityHasNoIdentity($entity, "remove from identity map");
@@ -1585,7 +1585,7 @@ class UnitOfWork implements PropertyChangedListener
         }
 
         $classMetadata = $this->em->getClassMetadata(get_class($entity));
-        $idHash        = implode(' ', $this->entityIdentifiers[$oid]);
+        $idHash        = $this->hashIdentifier($this->entityIdentifiers[$oid]);
 
         return isset($this->identityMap[$classMetadata->rootEntityName][$idHash]);
     }
@@ -2503,7 +2503,7 @@ class UnitOfWork implements PropertyChangedListener
         //$isReadOnly = isset($hints[Query::HINT_READ_ONLY]);
 
         $id = $this->identifierFlattener->flattenIdentifier($class, $data);
-        $idHash = implode(' ', $id);
+        $idHash = $this->hashIdentifier($id);
 
         if (isset($this->identityMap[$class->rootEntityName][$idHash])) {
             $entity = $this->identityMap[$class->rootEntityName][$idHash];
@@ -2665,7 +2665,7 @@ class UnitOfWork implements PropertyChangedListener
                     // Check identity map first
                     // FIXME: Can break easily with composite keys if join column values are in
                     //        wrong order. The correct order is the one in ClassMetadata#identifier.
-                    $relatedIdHash = implode(' ', $associatedId);
+                    $relatedIdHash = $this->hashIdentifier($associatedId);
 
                     switch (true) {
                         case (isset($this->identityMap[$targetClass->rootEntityName][$relatedIdHash])):
@@ -2943,7 +2943,7 @@ class UnitOfWork implements PropertyChangedListener
      */
     public function tryGetById($id, $rootClassName)
     {
-        $idHash = implode(' ', (array) $id);
+        $idHash = $this->hashIdentifier((array) $id);
 
         return isset($this->identityMap[$rootClassName][$idHash])
             ? $this->identityMap[$rootClassName][$idHash]
@@ -3338,7 +3338,17 @@ class UnitOfWork implements PropertyChangedListener
             ? $this->entityIdentifiers[$oid2]
             : $this->identifierFlattener->flattenIdentifier($class, $class->getIdentifierValues($entity2));
 
-        return $id1 === $id2 || implode(' ', $id1) === implode(' ', $id2);
+        return $id1 === $id2 || $this->hashIdentifier($id1) === $this->hashIdentifier($id2);
+    }
+
+    /**
+     * @param array $identifier
+     *
+     * @return string
+     */
+    private function hashIdentifier(array $identifier)
+    {
+        return json_encode(array_values(array_map('strval', $identifier)));
     }
 
     /**
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -3348,7 +3348,7 @@ class UnitOfWork implements PropertyChangedListener
      */
     private function hashIdentifier(array $identifier)
     {
-        return json_encode(array_values(array_map('strval', $identifier)));
+        return serialize(array_values(array_map('strval', $identifier)));
     }
 
     /**
--- a/tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
+++ b/tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
@@ -422,12 +422,15 @@ class UnitOfWorkTest extends OrmTestCase
         $booleanFalse->id = false;
 
         return [
-            'empty string, single field'     => [$emptyString, ''],
-            'non-empty string, single field' => [$nonEmptyString, $nonEmptyString->id],
-            'empty strings, two fields'      => [$emptyStrings, ' '],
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3329,7 +3329,17 @@ private function isIdentifierEquals($entity1, $entity2)
...
+        return $id1 === $id2 || $this->hashIdentifier($id1) === $this->hashIdentifier($id2);
+    }
+
+    /**
+     * @param array $identifier
+     *
+     * @return string
+     */
+    private function hashIdentifier(array $identifier)
+    {
```
beberlei:
```
if (!isset($identifier[1])) {
     return (string)$identifier[0];
}
```

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3329,7 +3329,17 @@ private function isIdentifierEquals($entity1, $entity2)
...
+        return serialize(array_values(array_map('strval', $identifier)));
```
Ocramius: As discussed on IRC, the performance impact of this functionality may be huge.

lib/Doctrine/ORM/UnitOfWork.php
Ocramius: Sadly, `$identifier` is not guaranteed to be `0-indexed`.

lib/Doctrine/ORM/UnitOfWork.php
Ocramius: Performance impact is ~2% of hydration time. Seems acceptable. Thoughts @beberlei @deeky666 @guilhermeblanco

deeky666: Well I don't have a strong opinion on this. I tend to prefer functionality also considering that the whole communication process between Doctrine and the DB takes a lot of more time than having that serialization here. Sometimes I find that optimization Doctrine wise over exaggerated because if the retrieval from DB will always be much slower than PHP code evaluation (especially if you do not have a blazing fast DB server). Also considering that PHP7 will most probably be a lot faster, I don't see any problems with this. Furthermore I think that people having those edge cases will surely be willing to take the minimal performance difference. But that is just my 2 cents. In the end if you want to query a lot of entities you might nevertheless be better off doing pure DBAL queries. Hope that helps...

deeky666: I mean lets take a real world example. Lets say we have a website that takes 100 entities (which is quite a lot already) and that needs to render. What impact (absolute) does that have? Like 20ms? Who will recognize that? Nobody. Everything that goes beyond 100 entries at a time should not be done with ORM IMO. Of course if my stupidly guessed 20ms is in fact 2 seconds, it's a different story but otherwise who cares?

deeky666: Btw I still have a lot of projects using Doctrine 1 and it's a lot worse than Doctrine 2 performance and memory wise and still acceptable (at least performance wise) as long as the database server is strong enough to deliver the stuff. Even with 100+ entities. So I really don't see a problem with this.

Ocramius: I agree that it looks like pennies on this side, but we should still care very much for anything that is `O(n)` or in the same league.

We can't do miracles, and 2% here is indeed not bad, but we should strive for better performance also to get rid of the silly/misplaced "ORMs are slow" reputation.

We did a lot of research on this for 3.x, so we will see some impact there (better performance), but we should never just go full "oh, nobody cares".

guilhermeblanco: Keep in mind that Asmir worked on a similar issue a few years ago and we didn't accept because of the 2% performance impact... We might take a look at his approach too to verify all issues got solved. =)

lcobucci:
@Ocramius @guilhermeblanco what should we do with this one?

Ocramius:
@lcobucci I'd say that we need to add benchmarks for these code paths

lcobucci:
@Ocramius alright should we move it to `v2.7` or maybe `v2.6.1` then?

Ocramius:
@lcobucci 3.x

lcobucci:
@Ocramius any reason for moving it back to `2.6.x`?

Ocramius:
I want to give this another shot before abandoning all hope: it's a bad one
and it should die

lcobucci:
Benchmark results (using #6575)
---

Test | Time (μs) - before | Time (μs) - after | Diff (μs) |
--------------|-------------------|-----------------|----|
UnitOfWorkComputeChangesBench#benchChangeSetComputation                          |12,055    |12,338    |+283     |
MixedQueryFetchJoinArrayHydrationPerformanceBench#benchHydration                 |437,631   |441,650   |+4,019   |
MixedQueryFetchJoinFullObjectHydrationPerformanceBench#benchHydration            |455,700   |494,812   |+39,112  |
MixedQueryFetchJoinPartialObjectHydrationPerformanceBench#benchHydration         |242,108   |262,340   |+20,232  |
SimpleHydrationBench#benchHydration                                              |1,338,824 |1,395,454 |+56,630  |
SimpleInsertPerformanceBench#benchHydration                                      |3,647,768 |3,666,269 |+18,501  |
SimpleQueryArrayHydrationPerformanceBench#benchHydration                         |243,605   |245,064   |+1,459   |
SimpleQueryFullObjectHydrationPerformanceBench#benchHydration                    |1,608,842 |1,742,565 |+133,723 |
SimpleQueryPartialObjectHydrationPerformanceBench#benchHydration                 |489,523   |578,972   |+89,449  |
SimpleQueryScalarHydrationPerformanceBench#benchHydration                        |183,123   |211,087   |+27,964  |
SingleTableInheritanceHydrationPerformanceBench#benchHydrateFixContracts         |5,822     |6,949     |+1,127   |
SingleTableInheritanceHydrationPerformanceBench#benchHydrateFlexContracts        |9,851     |10,972    |+1,121   |
SingleTableInheritanceHydrationPerformanceBench#benchHydrateUltraContracts       |6,610     |7,475     |+865     |
SingleTableInheritanceHydrationPerformanceBench#benchHydrateAllContracts         |12,951    |14,048    |+1,097   |
SingleTableInheritanceInsertPerformanceBench#benchInsertFixContracts             |14,536    |15,768    |+1,232   |
SingleTableInheritanceInsertPerformanceBench#benchInsertFlexContracts            |15,428    |18,049    |+2,621   |
SingleTableInheritanceInsertPerformanceBench#benchInsertUltraContracts           |16,413    |17,102    |+689     |
SingleTableInheritanceInsertPerformanceBench#benchInsertAllContracts             |33,686    |38,128    |+4,442   |
ProxyInitializationTimeBench#benchCmsUserInitialization                          |645,743   |680,637   |+34,894  |
ProxyInitializationTimeBench#benchCmsEmployeeInitialization                      |142,335   |152,550   |+10,215  |
ProxyInitializationTimeBench#benchInitializationOfAlreadyInitializedCmsUsers     |9,178     |9,249     |+71      |
ProxyInitializationTimeBench#benchInitializationOfAlreadyInitializedCmsEmployees |9,065     |8,666     |-399     |
ProxyInstantiationTimeBench#benchCmsUserInstantiation                            |993,616   |1,057,435 |+63,819  |
ProxyInstantiationTimeBench#benchCmsEmployeeInstantiation                        |448,051   |532,088   |+84,037  |


**Summary**

Metric    | Before                  | After                   |
----------|-------------------------|-------------------------|
best      | 5,822.000               | 6,949.000               |
mean mode | 459,269.333 459,269.333 | 484,152.792 484,152.792 |
worst     | 5,822.000               | 6,949.000               |
⅀T        | 11,022,464.000μs        | 11,619,667.000μs        |

Ocramius:
That is *a lot*. Assuming that identifiers don't change, except for generated identifiers, the hashing should probably be cached...

lcobucci:
@Ocramius I've rebased this branch to have phpbench, so the diff is:
[`master`](https://travis-ci.org/doctrine/doctrine2/jobs/256545517#L578): ⅀T: 7,077,983.000μs
[`this PR`](https://travis-ci.org/doctrine/doctrine2/jobs/256547475#L579): ⅀T: 5,950,296.000μs

But Travis' containers might not be 100% reliable regarding performance, idk.

guilhermeblanco:
This PR is wrong. It introduces a behavior worse than we have today.
Let me quickly try to explain: If you have an entity that contains an association as identifier and the target entity implements `__toString()`, everything breaks.

I'm at the point where I have to properly address this issue for once and for all.
The solution is to NOT store flattened identifiers in `$uow->entityIdentifiers`. It's also not to hash using any kind of `implode()`, `array_map()` or any transformation. 
It should purely get the identifier values (in master, `$metadata->getIdentifierValues()`; in develop, `$persister->getIdentifier()`) and call `serialize()`. You can see why this is required since the `UnitOfWork::isIdentifierEquals()` method is currently broken, since flattened identifiers and entityIdentifiers *can* be different today.
Using `serialize()` will allow developers to also control the hash if they want (by implementing `Serializable`).

As soon as you make these type of changes, all EntityPersisters break. The reason is because association identifiers can only go 1 level deep (associated id whose target entity also has associated id are not supported). This would require to completely rewrite both `prepareInsertData` and `prepareUpdateData`. I've done 50% of the work already in `develop`, but I consider it worthless to try in `master`.

IMHO, we should just close this PR.

lcobucci:
@Ocramius let's close this then?

Ocramius:
Closing is fine by me, but we still need to solve the ID => object problem,
and hashing is a semi-decent solution.

My last attempt used the SplObjectMap (or whatever it was called), and
performance was dreadful. Unless we enforce usage of the php-dd extension,
I don't see this happening 😥

Ocramius:
Delaying to 3.0 - while this is a critical bug, fixing it now is more trouble, as we'd only introduce more dirty hacks.

derrabus:

PR #5935, opened 2016-07-14
Fix #5684 - Fix mapping on insertion for autoincrement columns with custom types

### LINKED ISSUE #5684: Custom type ignored by identity generator
I have a custom type which extends the `IntegerType`. It encodes the integer on the way out of the database, and decodes it on the way in with the use of `convertToPHPValue`/`convertToDatabaseValue`. I tried using it like so:

``` php
<?php

namespace App\Entities;

use Doctrine\ORM\Mapping AS ORM;

class Location
{
    /**
     * @var string
     * @ORM\Id
     * @ORM\Column(type="hash", nullable=false)
     */
    protected $id;

    // ...
}
```

The problem with this implementation is that the generated value returned from the database after a `persist`/`flush` operation is the integer stored in the database; not the value as it would be had it passed through the `convertToPHPValue` method of my custom type.

mantiz:
Fixes #5684

The problem when using a custom type for autoincrement columns is that the value of autoincrement column returned by the database gets assigned directly without respecting mapping information. This only happens for insertion. When you fetch the records, the value is already mapped correctly.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -20,6 +20,7 @@
 namespace Doctrine\ORM;
 
 use Doctrine\Common\Persistence\Mapping\RuntimeReflectionService;
+use Doctrine\DBAL\Types;
 use Doctrine\DBAL\LockMode;
 use Doctrine\ORM\Internal\HydrationCompleteHandler;
 use Doctrine\ORM\Mapping\Reflection\ReflectionPropertiesGetter;
@@ -987,6 +988,7 @@ class UnitOfWork implements PropertyChangedListener
         $className  = $class->name;
         $persister  = $this->getEntityPersister($className);
         $invoke     = $this->listenersInvoker->getSubscribedSystems($class, Events::postPersist);
+        $platform   = $this->em->getConnection()->getDatabasePlatform();
 
         foreach ($this->entityInsertions as $oid => $entity) {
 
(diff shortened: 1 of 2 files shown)


mantiz:
Ok, it turned out to be a bigger problem than I initially thought :wink: 

I'm not sure if everything is covered but it looks good to me. Let me know if I forgot something or in the case a test case is missing.

I don't know what I think about the requirement to add a `__toString()` method to the custom type which is required for e.g. https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/UnitOfWork.php#L1590. At least, it seems to work and it's easy to implement. :smile:

Ocramius:
@lcobucci can you check over this? Is it similar to #6020?

lcobucci:
@Ocramius sure, I think the changes made in #6020 are a bit simpler (can still be improved). This evening I'll check out the other branch and see if these tests are passing and provide feedback.

mantiz:
I visually compared this PR to #6020 and I think they both are effectively the same but the implementation in #6020 looks much cleaner than mine. So, I vote for closing this PR in favour of #6020.

lcobucci:
@mantiz I've combined your test case with @renan's code and rebased it with `master` on https://github.com/lcobucci/doctrine2/commits/fix/id-generator-custom-type. I think that the combination of the two PRs is the best thing to the project. Just waiting for @Ocramius opinion

Ocramius:
Handled in #6152

Ocramius:
@mantiz your tests were merged into #6152, thanks!

PR #5954, opened 2016-07-27
Add OR to PersistentCollection matching()

michsk:
Hi guys, i'm wondering if this could be a option.

Now when i have a `OneToMany`, and want to add a option to also include many entities which have a null. We have entities which are assigned to a other entity, but we also have them assigned to null, which means they are default and are assiged to all entities.

The method to get the collection would look like:

``` php
    public function getElements(): Collection
    {
        $cr = Criteria::create();
        $cr->orWhere(Criteria::expr()->isNull('house'));

        return $this->elements->matching($cr, true);
    }
```

But because the `PersistentCollection` only uses andX this was not possible. Therefore this pull request.

### DIFF
--- a/lib/Doctrine/ORM/PersistentCollection.php
+++ b/lib/Doctrine/ORM/PersistentCollection.php
@@ -646,7 +646,7 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
      *
      * @throws \RuntimeException
      */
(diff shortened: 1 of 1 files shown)


mcfedr:
Take a look at the more extensive solution to this, https://github.com/doctrine/doctrine2/pull/5669

PR #5958, opened 2016-07-31
Avoid adding a column for the discriminator if the column is already defined

thomasbeaujean:
Avoid adding a column for the discriminator if the column is already defined.

Use case:
The User entity has a ManyToOne relationship to the UserType entity.
The ManyToOne relationship is done using the column user_type_id.

I want to use this column user_type_id for the discriminator map.

Everything works except the doctrine:shema:update.

An exception is raised: The column is already defined.
The solution: For the discriminator column, do not try to add the column if this one has already been set.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -358,23 +358,26 @@ class SchemaTool
     {
         $discrColumn = $class->discriminatorColumn;
 
-        if ( ! isset($discrColumn['type']) ||
-            (strtolower($discrColumn['type']) == 'string' && $discrColumn['length'] === null)
-        ) {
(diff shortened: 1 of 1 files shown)


Ocramius:
This is invalid for a few reasons:

 * the mapped column/association value can be changed, allowing data changes to mutate the actual type on an `EntityManager#refresh()` call (for example). This is a huge headache.
 * the discriminator column and the non-transient column have to be selected twice by the ORM - this will go wrong in the hydration, and isn't currently supported
 * this assumes that the discriminator is part of the fetched entity data: it's not

Closing as `invalid`. The correct fix is to throw an exception due to column duplication (if that's not already happening).

thomasbeaujean:
@Ocramius 

I agree with you.
Thank you for your review.

PR #5809, opened 2016-05-04
Verify embeddables do not define associations

nicoschoenmaker:
If I define an association on an embeddable, the following happens:
- The schema validates without errors;
- Yet my association is ignored!

There are two ways to solve this problem: invalidate the schema, or add associations of the embeddable.

Personally I would find it intuitive if the associations work. Currently we've placed a workaround, but we'd prefer adding the association mapping in the embeddable.

According to #4291, Doctrine consciously does not support associations in embeddables. That is why I opted for stricter validation. That way the error would have at least been found sooner.

PR #5962, opened 2016-08-03
Allow manyToOne associations to be embedded

### LINKED ISSUE #4291: DDC-3480: ORM\Embeddable does not create ManyToOne column in the database
Jira issue originally created by user tvoslar:

Following Embeddable

```
/****
 * @ORM\Embeddable
 */
class Address
{
    /****
     * @ORM\ManyToOne(targetEntity="Country")
     */
    protected $country;
```

won't save country attribute into the database table, other simple attributes (like type="text") saved normally.

```
    /****
     * @ORM\Embedded(class="LuciniLucini\Adsender\CommonBundle\Entity\Address")
     */
    private $address;
```

scaytrase:
We've recently faced, that embeddables are not allowed to have an oneToMany relation. My investigations basically finished that it could be solved by several lince of code, allowing inlining the embeddable's relations into the parent entity.

I'm not sure that I've patched all the placed and did it properly, so I'm ready to any improvements.

Currently we have to override the metadata factory to use patched class metadata class

Fix #4291

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -965,10 +965,19 @@ class ClassMetadataInfo implements ClassMetadata
                 : $reflService->getAccessibleProperty($this->name, $field);
         }
 
-        foreach ($this->associationMappings as $field => $mapping) {
-            $this->reflFields[$field] = isset($mapping['declared'])
-                ? $reflService->getAccessibleProperty($mapping['declared'], $field)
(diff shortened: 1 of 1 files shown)


Ocramius:
Ref: https://github.com/doctrine/doctrine2/pull/5809

We won't implement associations in embeddable fields, but we will instead disallow them forcefully.

scaytrase:
Thanks for fast decline, @Ocramius.
As far as i see doctrine starts to be less "OO-first" as stated before.

I hope it would be easy to override #5809 implemented restriction

PR #5964, opened 2016-08-04
Fist steps to close DDC-2134

marcobuschini:
This pull request creates a new command `deadentities` that seeks for child entities that have no parent. This is for helping people getting to implement referential integrity on databases that didn't have it. It does it all by analyzing the database schema via Doctrine API, and then generating queries to seek for the "missing" data.
It still does not work with composite foreign keys, possibly due to my lack of knowledge of the Doctrine schema API, or query language.
Testing data should be inserted by hand in the database, as I do not know how I can violate foreign keys with Doctrine, but everything else should be database independent.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/DeadEntitiesCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/DeadEntitiesCommand.php
@@ -87,46 +87,44 @@ EOT
             try {
                 $tableName = array_search($className, $tableNames);
                 $keys = $schema->listTableForeignKeys($tableName);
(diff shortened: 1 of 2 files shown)


Ocramius:
@marcobuschini this patch is not fit for core, in my opinion.

1. consider moving it to a separate component
2. split the logic that creates the SQL queries from the logic that runs the command
3. make sure that you *iterate* over the results (do not dump an array, it will just kill your memory)
4. use the symfony output helpers to make the output nicer

Closing here. I agree that this is a much needed tool, but it doesn't fit ORM core.

Also: this tool can be implemented without ORM (just use DBAL, look for foreign keys, repeat the same logic there ;-) )

PR #5974, opened 2016-08-12
documenation -- fix getting started instructions

jackdpeterson:
Resolve issue with 'em' not found
Add annotation driver so files can actually be read

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -141,6 +141,14 @@ step:
     //$config = Setup::createXMLMetadataConfiguration(array(__DIR__."/config/xml"), $isDevMode);
     //$config = Setup::createYAMLMetadataConfiguration(array(__DIR__."/config/yaml"), $isDevMode);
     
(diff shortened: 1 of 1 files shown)


docs/en/tutorials/getting-started.rst
```diff
@@ -186,7 +194,10 @@ doctrine command. Its a fairly simple file:
     // cli-config.php
     require_once "bootstrap.php";
     
-    return \Doctrine\ORM\Tools\Console\ConsoleRunner::createHelperSet($entityManager);
+    $helperSet = ConsoleRunner::createHelperSet($entityManager->getConnection());
+    // register the entity manager helper
+    $helperSet->set(new EntityManagerHelper($entityManager), 'em');
```
Ocramius: Wait, doesn't `createHelperSet` already create the `EntityManagerHelper`? Is it missing an alias?

Ocramius: If we are miss the `'em'` helper, then we should get rid of it from the docs instead (if the name doesn't match)

PR #5996, opened 2016-08-26
[QueryBuilder] setParameter : trim the input key when comparing it to…

fancyweb:
… existing parameters keys

Basically when you first use setParameter with a parameter name starting with ":" it trims it in the Doctrine\ORM\Query\Parameter constructor.
When you later want to change the value of this parameter, still using the name starting with ":", the parameter is not found because it compares it with the trimmed name stored in the Parameter object so it duplicates it.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -325,8 +325,10 @@ abstract class AbstractQuery
         $filteredParameters = $this->parameters->filter(
             function ($parameter) use ($key)
             {
+                $trimmedKey = trim($key, ':');
+
                 // Must not be identical because of string to integer conversion
-                return ($key == $parameter->getName());
+                return ($trimmedKey == $parameter->getName());
             }
         );
 
@@ -374,8 +376,10 @@ abstract class AbstractQuery
         $filteredParameters = $this->parameters->filter(
             function ($parameter) use ($key)
             {
+                $trimmedKey = trim($key, ':');
+
                 // Must not be identical because of string to integer conversion
-                return ($key == $parameter->getName());
+                return ($trimmedKey == $parameter->getName());
             }
         );
 
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -536,6 +536,7 @@ class QueryBuilder
             function ($parameter) use ($key)
             {
                 $trimmedKey = trim($key, ':');
+
                 /* @var Query\Parameter $parameter */
                 // Must not be identical because of string to integer conversion
                 return ($trimmedKey == $parameter->getName());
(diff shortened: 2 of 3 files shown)


Ocramius:
`setParameter` should be called without the `:` in the parameter name. Could you show a test case where this is allowed, and fails with what you described?

I'd also suggest that we deprecate support for `:` in bound parameter names (`:` is just a delimiter, not part of the name).

Also, I'd suggest checking the docs to see if we ever suggest that `:` is allowed (shouldn't be)

fancyweb:
@Ocramius 

Actually I have found that there is the same problem with the Query class and with the `getParameter` method so I fixed those cases and I also added tests.

I agree that it's not that good to use a parameter name starting with `:` but my point is that it is currently allowed and working when you use the `setParameter` method so I expect it to work again when I reuse this method or the `getParameter` method with the same input key.

phansys:
IMO a better approach could be validate the parameter name and throw an exception when it doesn't accomplish the expectations, instead of silent the issue with a workaround.

fancyweb:
@phansys Wouldn't this be a BC break ?

phansys:
@fancyweb, I didn't noticed what you've mentioned about [`Parameter::__construct()`](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Query/Parameter.php#L61), sorry.
I really don't know which is the reason for that `trim()` call, even in the worst case it should be `ltrim()` I think. But again, IMO, that call is hiding the issue.
Given the current behavior is already what you're proposing here, I think yes, throwing an exception in that case could be considered as a BC break. In order to face this situation, it would be nice to add a deprecation path for the next major release removing this behavior, but let's wait for @Ocramius' concerns.

phansys:
I think this could work:
```php
// lib/Doctrine/ORM/Query/Parameter.php

public function __construct($name, $value, $type = null)
{
    if (!preg_match('#^[\w\-]+$#', $name)) {
        @trigger_error('Using names with chars other than alphanumeric, "_" and "-" is deprecated since 2.6, its support will be removed in 3.0', E_USER_DEPRECATED);
    }
    $this->name = trim($name, ':');

    $this->setValue($value, $type);
}
```
If this proposal is accepted, you should add the same deprecation check in any place that trims the parameter name, but ensuring that the deprecation error is triggered just one time in each case.
Note that the regex I've used is very simple, so maybe it could needs to be updated regarding the concerns of doctrine members.

Ocramius:
Throwing a deprecation notice is acceptable for 2.6, IMO. Removal/exception in 3.0

Majkl578:
@fancyweb Can you please prepare new PRs for deprecating and removing the behavior?
One PR against master with deprecation warning using trigger_error (see doctrine/dbal#2846 and @phansys' comment for examples), second one against develop where exception (or no trimming - to be discussed) will be thrown. Thanks.

PR #6001, opened 2016-09-02
Make query cache rely on entity timestamp region

lcobucci:
As far as I saw that's the default Hibernate behavior (https://vladmihalcea.com/2015/06/08/how-does-hibernate-query-cache-work/) to keep the query cache consistent.

I'm just not sure if that is the way to go, basically because of the change on the `CachedEntityPersister` interface (I didn't want to duplicate the logic that generates the timestamp key).

### DIFF
--- a/lib/Doctrine/ORM/Cache/QueryCacheEntry.php
+++ b/lib/Doctrine/ORM/Cache/QueryCacheEntry.php
@@ -38,18 +38,18 @@ class QueryCacheEntry implements CacheEntry
     /**
      * READ-ONLY: Public only for performance reasons, it should be considered immutable.
      *
-     * @var integer Time creation of this cache entry
+     * @var float Time creation of this cache entry
      */
     public $time;
 
     /**
-     * @param array   $result
-     * @param integer $time
+     * @param array $result
+     * @param float $time
      */
     public function __construct($result, $time = null)
     {
         $this->result = $result;
-        $this->time   = $time ?: time();
+        $this->time   = $time ?: microtime(true);
     }
 
     /**
--- a/lib/Doctrine/ORM/Cache/QueryCacheKey.php
+++ b/lib/Doctrine/ORM/Cache/QueryCacheKey.php
@@ -45,14 +45,27 @@ class QueryCacheKey extends CacheKey
     public $cacheMode;
 
     /**
-     * @param string  $hash      Result cache id
-     * @param integer $lifetime  Query lifetime
-     * @param integer $cacheMode Query cache mode
+     * READ-ONLY: Public only for performance reasons, it should be considered immutable.
+     *
+     * @var TimestampCacheKey|null
      */
-    public function __construct($hash, $lifetime = 0, $cacheMode = Cache::MODE_NORMAL)
-    {
-        $this->hash      = $hash;
-        $this->lifetime  = $lifetime;
-        $this->cacheMode = $cacheMode;
+    public $timestampKey;
+
+    /**
+     * @param string $hash Result cache id
+     * @param integer $lifetime Query lifetime
+     * @param int $cacheMode Query cache mode
+     * @param TimestampCacheKey|null $timestampKey
+     */
+    public function __construct(
+        $hash,
+        $lifetime = 0,
+        $cacheMode = Cache::MODE_NORMAL,
+        TimestampCacheKey $timestampKey = null
+    ) {
+        $this->hash         = $hash;
+        $this->lifetime     = $lifetime;
+        $this->cacheMode    = $cacheMode;
+        $this->timestampKey = $timestampKey;
     }
 }
--- a/lib/Doctrine/ORM/Cache/TimestampQueryCacheValidator.php
+++ b/lib/Doctrine/ORM/Cache/TimestampQueryCacheValidator.php
@@ -27,14 +27,48 @@ namespace Doctrine\ORM\Cache;
 class TimestampQueryCacheValidator implements QueryCacheValidator
 {
     /**
+     * @var TimestampRegion
+     */
+    private $timestampRegion;
+
+    /**
+     * @param TimestampRegion $timestampRegion
+     */
+    public function __construct(TimestampRegion $timestampRegion)
+    {
+        $this->timestampRegion = $timestampRegion;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function isValid(QueryCacheKey $key, QueryCacheEntry $entry)
     {
+        if ($this->regionUpdated($key, $entry)) {
+            return false;
+        }
+
         if ($key->lifetime == 0) {
             return true;
         }
 
         return ($entry->time + $key->lifetime) > microtime(true);
+    }
+
+    /**
+     * @param QueryCacheKey   $key
+     * @param QueryCacheEntry $entry
+     *
+     * @return bool
+     */
+    private function regionUpdated(QueryCacheKey $key, QueryCacheEntry $entry)
+    {
+        if ($key->timestampKey === null) {
+            return false;
+        }
+
+        $timestamp = $this->timestampRegion->get($key->timestampKey);
+
+        return $timestamp && $timestamp->time > $entry->time;
     }
 }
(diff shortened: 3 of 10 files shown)


lcobucci:
I had an idea to improve this PR a bit by moving the timestamp region check to the query cache validator but this also means moving the `timestampKey` to the `QueryCacheKey`.

I think it's a way better solution what's your opinion?

guilhermeblanco:
@lcobucci indeed that's a MUCH BETTER solution. If I understood your idea correctly, you'd eliminate the BC break by adding a new interface method.
This means it could potentially be part of 2.X series instead of 3.X only. =)

If you're willing to work on that, I'd appreciate.

lcobucci:
@guilhermeblanco I really need this to be part of 2.X so I'm glad to help 😉 

> If I understood your idea correctly, you'd eliminate the BC break by adding a new interface method.

My idea basically is:

1. Add `timestampRegion` to the `TimestampQueryCacheValidator` constructor
1. Add `timestampKey` to the `QueryCacheKey` constructor
1. Make `TimestampQueryCacheValidator#isValid()` also validate the timestamp using the region and the key

I didn't see the BC break you said, can you please elaborate a bit more?

The biggest question here is: what'd be the best place to retrieve the `timestampKey`?
Passing it to the `QueryCacheKey` constructor means that we need to have it available also on the `AbstractQuery#executeUsingQueryCache()`.

Moving those things to que validator also simplifies the `AbstractEntityPersister#getHash()` logic, since we don't need concatenate the `$timestamp` anymore (and we could also remove the unused `timestampRegion` from the persister).

lib/Doctrine/ORM/Cache/Persister/Entity/CachedEntityPersister.php
```diff
@@ -44,4 +44,9 @@ public function getEntityHydrator();
      * @return boolean
      */
     public function storeEntityCache($entity, EntityCacheKey $key);
+
+    /**
+     * @return \Doctrine\ORM\Cache\TimestampCacheKey
+     */
+    public function getTimestampKey();
```
Ocramius: bc break, sadly

Ocramius: Also, I don't know how to regenerate a timestamp key, if not retrieved from here.

lcobucci:
Ow ofc. Do you think that the solution I explained is applicable?

Ocramius:
@lcobucci moving more info to the `QueryCacheKey` seems good to me, since it's information that sticks together anyway. I don't have the full picture though, sorry.

lcobucci:
@guilhermeblanco @Ocramius introduced a tiny duplication in order to remove the BC break. I think it should be ok to be merged on v2.X now.

Anything else?

lib/Doctrine/ORM/Cache/TimestampQueryCacheValidator.php
```diff
@@ -27,14 +27,39 @@
...
         if ($key->lifetime == 0) {
             return true;
         }
 
         return ($entry->time + $key->lifetime) > time();
     }
+
+    private function regionUpdated(QueryCacheKey $key, QueryCacheEntry $entry)
+    {
+        if ($key->timestampKey === null) {
```
Ocramius: The docblock states that it is never `null`: can you check that?

lib/Doctrine/ORM/Cache/TimestampQueryCacheValidator.php
```diff
@@ -27,14 +27,39 @@
...
+        }
+
         if ($key->lifetime == 0) {
             return true;
         }
 
         return ($entry->time + $key->lifetime) > time();
     }
+
+    private function regionUpdated(QueryCacheKey $key, QueryCacheEntry $entry)
```
Ocramius: Docblock plz, even if it just says that a boolean is returned

lib/Doctrine/ORM/Cache/TimestampQueryCacheValidator.php
lcobucci: The docblock is a liar! Will fix that, thanks

lib/Doctrine/ORM/Cache/TimestampQueryCacheValidator.php
lcobucci: 😉

lcobucci:
@guilhermeblanco @Ocramius comments processed let me know if there's anything else preventing this to be merged and let's :shipit:

lib/Doctrine/ORM/Cache/QueryCacheEntry.php
```diff
@@ -38,18 +38,18 @@ class QueryCacheEntry implements CacheEntry
     /**
      * READ-ONLY: Public only for performance reasons, it should be considered immutable.
      *
-     * @var integer Time creation of this cache entry
+     * @var float Time creation of this cache entry
```
Ocramius: Since there was a change here, did you check all usages of this property?

lib/Doctrine/ORM/Cache/QueryCacheKey.php
```diff
@@ -45,14 +45,27 @@ class QueryCacheKey extends CacheKey
...
+     *
+     * @var TimestampCacheKey|null
+     */
+    public $timestampKey;
+
+    /**
+     * @param string $hash Result cache id
+     * @param integer $lifetime Query lifetime
+     * @param int $cacheMode Query cache mode
+     * @param TimestampCacheKey|null $timestampKey
```
Ocramius: When is `null` happening here, exactly?

tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheQueryCacheTest.php
```diff
@@ -1120,4 +1120,37 @@ public function testNonCacheableQueryUpdateStatementException()
...
+        $this->_em->clear();
+
+        $queryCount = $this->getCurrentQueryCount();
+
+        $result2 = $this->_em->createQuery($dql)
+            ->setCacheable(true)
+            ->getResult();
+
+        $this->assertCount(3, $result2);
+        $this->assertEquals($queryCount + 1, $this->getCurrentQueryCount());
```
Ocramius: Wait... Shouldn't this be `$queryCount + 2`?

Ocramius: I would expect that because of the region being updated (and therefore cache evicted)

lib/Doctrine/ORM/Cache/QueryCacheEntry.php
lcobucci: I did but I definitely forgot to change `($entry->time + $key->lifetime) > time()` to `((int) $entry->time + $key->lifetime) > time()` on the validator

lib/Doctrine/ORM/Cache/QueryCacheKey.php
lcobucci: Basically because of https://github.com/doctrine/doctrine2/blob/8a87fa2d01a01f4ae939559b10cb5733132930cf/tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheQueryCacheTest.php#L1106 and https://github.com/doctrine/doctrine2/blob/8a87fa2d01a01f4ae939559b10cb5733132930cf/tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheQueryCacheTest.php#L1117

Since the ResultSetMapping don't give the entity name.

tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheQueryCacheTest.php
Ocramius: Argh, totally misread, sorry. Missed that you are resetting the `$queryCount` after the first query.

Ocramius:
This goes into `2.5.5`! Thanks @lcobucci 

`master`: doctrine/doctrine2@009e94720b5dcf25e4399fb16072ac9cbc0dcae6
`2.5`: doctrine/doctrine2@e16de704a18a29dd71beb0f4d44afff98a890893

Issue #5989, opened 2016-08-24
SimpleObjectHydrator produces wrong values with inheritance table and simple array type 


w3sami:
I am getting empty array instead of array with some values, since SimpleObjectHydrator converts the value with convertToPHPValue

line 129 $value = $type->convertToPHPValue($value, $this->_platform);

this doesent play well with the null check below

line 134:
// Prevent overwrite in case of inherit classes using same property name (See AbstractHydrator)
            if ( ! isset($data[$fieldName]) || $value !== null) {
                $data[$fieldName] = $value;
            }

since arrayType->convertToPHPValue  returns [] when null, resulting in a leak from the wrong left joined table.

solution is easy, dont use the converted value in the if, but instead a raw value.

eg line 125 + $rawValue = $value;
and
136: if ( ! isset($data[$fieldName]) || $rawValue !== null) {

sorry I don't have time to write a pr with tests

Ocramius:
> since arrayType->convertToPHPValue returns [] when null, resulting in a leak from the wrong left joined table.

Seems like an issue with the array type, rather than with the ORM itself?

w3sami:
SimpleObjectHydrator is the file with the problem, type is ok imo. Same
issue is with eg json array type. Hydrator compares with type converted
version not raw value.
For me it makes sense that null is empty array for array type.

w3sami:
bumb! this is actually a quite serious bug, as it renders all our entities with joined inheritance + (json/)array type useless!!

Ocramius:
The fix in #6004 seems good, besides minor cleanup needed :-)

Ocramius:
Fixed in #6004, thanks @w3sami and @cvuorinen!

PR #6004, opened 2016-09-03
Fix hydration in a joined inheritance with simple array or json array

cvuorinen:
Both SimpleArrayType and JsonArrayType convert the value `null` to an empty array, which fails the null check that is used to prevent overwrite (in https://github.com/doctrine/doctrine2/blob/v2.5.4/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php#L134).
Fixes issue #5989 

It could be debated that the issue is actually with the types that convert null to something else (especially since ArrayType actually returns null when the value is null). But changing them would be a BC break so I think it is a separate issue. Also, there is no way to know how custom types will convert null values, so it seems like a reasonable approach to check for null before the conversion since it's quite a serious bug to overwrite data in the hydrator.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -122,7 +122,7 @@ class SimpleObjectHydrator extends AbstractHydrator
             }
 
             // Check if value is null before conversion (because some types convert null to something else)
-            $valueIsNull = $value === null;
+            $valueIsNull = null === $value;
 
             // Convert field to a valid PHP value
             if (isset($cacheKeyInfo['type'])) {
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -121,6 +121,9 @@ class SimpleObjectHydrator extends AbstractHydrator
                 continue;
             }
 
+            // Check if value is null before conversion (because some types convert null to something else)
+            $valueIsNull = $value === null;
+
             // Convert field to a valid PHP value
             if (isset($cacheKeyInfo['type'])) {
                 $type  = $cacheKeyInfo['type'];
@@ -130,7 +133,7 @@ class SimpleObjectHydrator extends AbstractHydrator
             $fieldName = $cacheKeyInfo['fieldName'];
 
             // Prevent overwrite in case of inherit classes using same property name (See AbstractHydrator)
-            if ( ! isset($data[$fieldName]) || $value !== null) {
+            if ( ! isset($data[$fieldName]) || ! $valueIsNull) {
                 $data[$fieldName] = $value;
             }
         }
--- a/tests/Doctrine/Tests/Models/Issue5989/Issue5989Employee.php
+++ b/tests/Doctrine/Tests/Models/Issue5989/Issue5989Employee.php
@@ -9,19 +9,9 @@ namespace Doctrine\Tests\Models\Issue5989;
 class Issue5989Employee extends Issue5989Person
 {
     /**
-     * @column(type="simple_array", nullable=true)
+     * @Column(type="simple_array", nullable=true)
      *
      * @var array
      */
-    private $tags;
-
-    public function getTags()
-    {
-        return $this->tags;
-    }
-
-    public function setTags(array $tags)
-    {
-        $this->tags = $tags;
-    }
+    public $tags;
 }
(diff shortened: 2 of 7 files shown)


lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
```diff
@@ -121,6 +121,9 @@ protected function hydrateRowData(array $sqlResult, array &$result)
                 continue;
             }
 
+            // Check if value is null before conversion (because some types convert null to something else)
+            $valueIsNull = $value === null;
```
phansys: Please, use Yoda condition instead.

tests/Doctrine/Tests/Models/Issue5989/Issue5989Employee.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Doctrine\Tests\Models\Issue5989;
+
+/**
+ * @Entity
+ * @Table(name="issue5989_employees")
+ */
+class Issue5989Employee extends Issue5989Person
+{
+    /**
+     * @column(type="simple_array", nullable=true)
```
Ocramius: case sensitivity, careful :-)

tests/Doctrine/Tests/Models/Issue5989/Issue5989Employee.php
```diff
@@ -0,0 +1,27 @@
...
+     *
+     * @var array
+     */
+    private $tags;
```
Ocramius: can just work with `public`

tests/Doctrine/Tests/Models/Issue5989/Issue5989Manager.php
```diff
@@ -0,0 +1,27 @@
...
+ * @Table(name="issue5989_managers")
+ */
+class Issue5989Manager extends Issue5989Person
+{
+    /**
+     * @column(type="simple_array", nullable=true)
+     *
+     * @var array
+     */
+    private $tags;
```
Ocramius: `public`

tests/Doctrine/Tests/Models/Issue5989/Issue5989Manager.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Doctrine\Tests\Models\Issue5989;
+
+/**
+ * @Entity
+ * @Table(name="issue5989_managers")
+ */
+class Issue5989Manager extends Issue5989Person
+{
+    /**
+     * @column(type="simple_array", nullable=true)
```
Ocramius: case sensitivity

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue5989Test.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+class Issue5989Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    public function setUp()
+    {
+        $this->useModelSet('issue5989');
+        parent::setUp();
+    }
+
+    public function testSimpleArrayTypeHydratedCorrectlyInJoinedInheritance()
```
Ocramius: Besides the functional test, is there a unit test that we can write for the `SimpleObjectHydrator`?

Please don't attempt to do that if it's too complex, just please check it and write one, if it is trivial.

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue5989Test.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $manager = new Issue5989Manager();
+
+        $managerTags = array('tag1', 'tag2');
```
Ocramius: You can use short array syntax overall

Ocramius:
Looks good! Feedback is only technical/minor stuff to be fixed, but this is going to be merged once that's fixed :-)

tests/Doctrine/Tests/Models/Issue5989/Issue5989Employee.php
cvuorinen: Yea, I think it was copy paste from some existing models you have there. There are quite a few with lowercase. But I'll fix these.

Ocramius:
@cvuorinen this looks good! Should I take over for a merge?

cvuorinen:
@Ocramius I found a way to add a unit test that also reproduces this issue and passes with the fix.
Does that look OK to you?

Ocramius:
@cvuorinen yes, that looks exactly like what I was asking for.

cvuorinen:
OK, then it's ready for merge.

Ocramius:
Merged, thanks a lot @cvuorinen!

`master`: doctrine/doctrine2@5eebdcf6308277347da88e1478fc4bc1374ed3ce
`2.5`: doctrine/doctrine2@d592c14a6c4fd09d7f1ea05f77f211a092509fb2

PR #6011, opened 2016-09-09
Use `ClassMedatada#isIdGeneratorIdentity()` instead of constant to add the ID on query

lcobucci:
It's quite handy when creating CUSTOM id generators that should also rely on AUTO_INCREMENT stuff (one can just extend the `ClassMedatadaFactory` and create a different instance of `ClassMetadata` that overrides that method).

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -1442,7 +1442,7 @@ class BasicEntityPersister implements EntityPersister
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


lcobucci:
I'm definitely not a big fan of having a `CUSTOM` id generator but we needed to create one for a project but regardless that this piece of code is duplicated and can be simplified by using the method 😉

guilhermeblanco:
I wouldn't bother tbh...

lcobucci:
@guilhermeblanco you're right, thanks 😉

Ocramius:
If it's not a bugfix, then no :-)

PR #6014, opened 2016-09-09
As of HHVM 3.15 pgsql is supported, add php 7.1 to tests

photodude:
- Remove the HHVM exclusion now that pgsql is in core HHVM
- add PHP 7.1
- fix the xdebug line that was erroring the php 7.1+ and HHVM builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,7 @@ matrix:
       env: DB=mariadb
       addons:
         mariadb: 10.1
-    - php: hhvm-3.15
+    - php: hhvm
       sudo: true
       dist: trusty
       group: edge # until the next Trusty update
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ language: php
 php:
   - 5.6
   - 7.0
+  - 7.1
   - nightly
-  - hhvm
 
 env:
   - DB=mysql
@@ -27,6 +27,32 @@ after_script:
 
 matrix:
   include:
+    - php: hhvm-3.15
+      sudo: true
+      dist: trusty
+      group: edge # until the next update
+      addons:
+        apt:
+          packages:
+            - mysql-server-5.6
+            - mysql-client-core-5.6
+            - mysql-client-5.6
+      services:
+        - mysql
+      env: DB=mysql
+    - php: hhvm-3.15
+      sudo: true
+      dist: trusty
+      group: edge # until the next update
+      services:
+        - postgresql
+      env: DB=pgsql
+    - php: hhvm-3.15
+      sudo: true
+      dist: trusty
+      group: edge # until the next update
+      env: DB=sqlite
+
     - php: 5.6
       env: DB=mariadb
       addons:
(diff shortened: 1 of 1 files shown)


photodude:
@Ocramius this is ready for consideration now.

.travis.yml
```diff
@@ -26,7 +26,34 @@ after_script:
   - if [[ $TRAVIS_PHP_VERSION = '7.0' && $DB = 'sqlite' ]]; then php ocular.phar code-coverage:upload --format=php-clover build/logs/clover.xml; fi
 
 matrix:
+  fast_finish: true
   include:
+    - php: hhvm-3.15
```
Ocramius: This is painful to maintain, and will require changing every time there is a minor version bump. Can we just rely on `hhvm` at the base config, and have travis-ci update to latest stable?

.travis.yml
```diff
@@ -35,10 +62,17 @@ matrix:
       env: DB=mariadb
       addons:
         mariadb: 5.5
-    - php: hhvm
+    - php: 7.1
       env: DB=mariadb
       addons:
         mariadb: 5.5
+    - php: hhvm-3.15
```
Ocramius: Same as above

.travis.yml
```diff
@@ -48,15 +82,21 @@ matrix:
...
         mariadb: 10.1
-    - php: hhvm
+    - php: 7.1
       env: DB=mariadb
       addons:
         mariadb: 10.1
-  exclude:
-    - php: hhvm
-      env: DB=pgsql  # driver for PostgreSQL currently unsupported by HHVM, requires 3rd party dependency
+    - php: hhvm-3.15
```
Ocramius: Same

.travis.yml
photodude: Yes and no...

I chose [3.15 since it's LTS](http://hhvm.com/blog/6083/hhvm-long-term-support), I believe the next LTS will be 3.18 or 3.20, or maybe 4.0. Considering 3.15 like an LTS php version 5.6 or 7.0

To be on the actual latest of HHVM we can skip specifying the version, but we can't put it in the base configuration since the Trusty container builds on Travis CI have not been released, and the precise containers only support HHVM 3.9. [See HHVM options](https://docs.travis-ci.com/user/languages/php#HHVM-versions)

So it's just a matter of do you only want to test latest, LTS versions, both, etc. 

We can also add builds for [HHVM in php 7 mode](https://docs.hhvm.com/hhvm/configuration/INI-settings#php-7-settings) rather than the default php 5 mode. (I skipped that option since [HHVM has an issue with the PHP 7 mode](https://github.com/facebook/hhvm/issues/7198) that is blocking composer from working)

photodude:
@Ocramius can the marina 5.5 builds be removed?

Ocramius:
@photodude yes, we can safely move everything to 5.6+

photodude:
Done, ready for review again.

photodude:
@Ocramius bump for merge consideration

Ocramius:
@photodude 👍 thanks for the poke!

photodude:
You are welcome

PR #6017, opened 2016-09-11
Avoid error when entityName isn't a string

j0k3r:
Previously I wasn't used the `->clear()` method in the correct way. I always put the full entity object in parameter instead of the entity name. It _worked_ before the 2.5.5 even if in fact it cleared nothing ...

With 2.5.5 a check has been added in `clearIdentityMapForEntityName` to verify if the given entity name exist in the entityMap.

Using `isset` with the entity object generated a PHP error `Illegal offset type in isset or empty`.

This PR aim to remove that error.

I'm not sure if this is the right way to fix that problem. I'm maybe the only person who gave the entity object as parameter to the `clear` function.
Maybe we should better detect if it's an object in the `clear` function and use the `:class` attribute to get the class name?

Like 

``` php
public function clear($entityName = null)
{
    if (is_object($entityName)) {
        $entityName = $entityName::class;
    }
```

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2370,7 +2370,7 @@ class UnitOfWork implements PropertyChangedListener
     public function clear($entityName = null)
     {
         if ($entityName !== null && !is_string($entityName)) {
-            throw new \InvalidArgumentException(sprintf('Argument 1 passed to %s() must be a string, %s given', __METHOD__, gettype($entityName)));
+            throw ORMException::invalidEntityName($entityName);
         }
 
         if ($entityName === null) {
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2369,6 +2369,10 @@ class UnitOfWork implements PropertyChangedListener
      */
     public function clear($entityName = null)
     {
+        if ($entityName !== null && !is_string($entityName)) {
+            throw new \InvalidArgumentException(sprintf('Argument 1 passed to %s() must be a string, %s given', __METHOD__, gettype($entityName)));
+        }
+
         if ($entityName === null) {
             $this->identityMap =
             $this->entityIdentifiers =
@@ -3448,10 +3452,6 @@ class UnitOfWork implements PropertyChangedListener
      */
     private function clearIdentityMapForEntityName($entityName)
     {
-        if (is_object($entityName)) {
-            return;
-        }
-
         if (! isset($this->identityMap[$entityName])) {
             return;
         }
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -3448,6 +3448,10 @@ class UnitOfWork implements PropertyChangedListener
      */
     private function clearIdentityMapForEntityName($entityName)
     {
+        if (is_object($entityName)) {
+            return;
+        }
+
         if (! isset($this->identityMap[$entityName])) {
             return;
         }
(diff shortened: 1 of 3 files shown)


phansys:
Since the docblock states that the argument must be an string, I think your approach is more a workaround than a fix. Think by instance what will happen when the PHP version is bumped to `^7.0` and the `string` type declaration is used. For this reason, if you think it's a common situation to provide a non supported type for this method, IMO a better solution could be emulate the type declaration:
```php
if (!is_string($entityName)) {
    throw new \InvalidArgumentException(sprintf('Argument 1 passed to %s() must be a string, %s given', __METHOD__, gettype($entityName)));
}
```

j0k3r:
Might be a better solution yeah.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2369,6 +2369,10 @@ public function getCommitOrderCalculator()
      */
     public function clear($entityName = null)
     {
+        if ($entityName !== null && !is_string($entityName)) {
```
phansys: Please, use Yoda condition.

j0k3r: I did it but then checked the file and see that it doesn't use yoda condition. Like on line below `if ($entityName === null) {` that's why I changed it :confused:

j0k3r:
Should I fix the coding style so this PR can have a chance to be merged?

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2369,6 +2369,10 @@ public function getCommitOrderCalculator()
...
+            throw new \InvalidArgumentException(sprintf('Argument 1 passed to %s() must be a string, %s given', __METHOD__, gettype($entityName)));
```
Ocramius: This should be something like `throw ORMException::fromInvalidEntityName($entityName)`

Ocramius (changes requested):

Ocramius:
@j0k3r sorry for missing the review of this before - likely ended in my mail backlog and was never checked later.

The patch is indeed legit, just need to throw an `ORMException` and code the message construction in a named constructor (static method) on the `ORMException` itself.

j0k3r:
@Ocramius no problem. I've added `ORMException::invalidEntityName`

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
```diff
@@ -349,6 +350,32 @@ public function testPersistedEntityAndClearManager()
...
     }
 
+    public function testClearManagerWithObject()
+    {
+        $this->expectException(ORMException::class);
+        $this->expectExceptionMessage('must be a string');
+
+        $entity = new Country(456, 'United Kingdom');
+
+        $this->_unitOfWork->persist($entity);
```
Ocramius: Not needed

j0k3r: Both removed

Ocramius:
@j0k3r I'm taking ownership of the branch in order to merge it and apply a few adjustments that I just thought of. Thanks for fixing that unused method call!

Commits will be displayed here, stand by.

j0k3r:
👍

Ocramius:
@j0k3r rebased and merged, thanks! doctrine/doctrine2@555e8ae641ef269c5b4793245d13d05170e69d2f

Here's what I added:

 * Checks were moved from `UnitOfWork` to `EntityManager`. While this reduces safety, I still consider `clear($entityName)` to be a deprecated feature, so it will be removed in future. Cluttering `UnitOfWork` seemed excessive, while `EntityManager` is doing the job just fine.
 * `$entityName` is now also checked for entity name validity (just a string is not sufficient!). If `$entityName` is not a valid class name, then a `MappingException` is thrown
 * `$entityName` can now also be a proxy name, so `$entityManager->clear(get_class($aProxy))` will work out of the box

👍

PR #6018, opened 2016-09-11
Convert generated id value to its PHP representation.

renan:
According to the conversion rules of a specific DBAL mapping type.


renan:
Spotted some failures on tests, will fix them first.

renan:
Re-created at #6020

PR #6020, opened 2016-09-11
Convert generated id value to its PHP representation.

renan:
This makes it possible for post-insert ids to be converted into a custom object as it happens for any other identity strategy.

According to the conversion rules of a specific DBAL mapping type.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -890,7 +890,11 @@ class UnitOfWork implements PropertyChangedListener
             $idValue = $idGen->generate($this->em, $entity);
 
             if ( ! $idGen instanceof \Doctrine\ORM\Id\AssignedGenerator) {
-                $idValue = array($class->identifier[0] => $idValue);
+                $idField = $class->identifier[0];
+                $idType  = $class->getTypeOfField($idField);
+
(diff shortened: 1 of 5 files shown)


lcobucci:
@renan I'm not sure what's @Ocramius opinion here but I'd say that I don't think that the ORM should be responsible for that.

Auto generated ids are DB-dependent and wouldn't make sense to have them converted to custom objects. IMHO it's an edge case and if you **really** need/want to do that you can override `ClassMetadata#setIdentifierValues()` and do the conversion there.

renan:
The ORM/DBAL offers a feature that can convert any given column to a different, including custom, object type. Said feature is documented [here](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html).

Currently **loading/hydrating** an entity that have a custom object type mapped for an auto-generated id will **work as expected**: converting any value to the mapped type.
What is missing is converting such value to the mapped type after a given entity is **newly persisted**. Again, saving an existing entity, or loading the entity, will also work as expected.

Hence I respectfully disagree, and argue this might even be a bug.

Ocramius:
This does indeed look like a bug. Can't check the code right now, but labeling it.

Ocramius:
@renan this was handled in #6152. Tests from here were ported there as well, thanks!

PR #6048, opened 2016-09-22
Add an optional method on entity to check if entities from cache are valid

BourotBenjamin:
To prevent error, you could add checks on entities with a method isCacheValid on your entities.
If you create this method, and it returns false, the EntityPersister will load it frtom th database. In other case, it will keep it original working.

### DIFF
--- a/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
+++ b/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
@@ -567,12 +567,14 @@ abstract class AbstractEntityPersister implements CachedEntityPersister
             $key     = $this->buildCollectionCacheKey($assoc, $ownerId);
             $list    = $persister->loadCollectionCache($coll, $key);
 
-            foreach ($list as $entity) {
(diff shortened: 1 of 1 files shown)


lcobucci:
> To prevent error, you could add checks on entities with a method isCacheValid on your entities.

What error you're preventing here? And why would you like to put cache control (infrastructure layer) on your entities (domain layer)?

BourotBenjamin:
I'm preventing errors like that : 
https://github.com/snc/SncRedisBundle/issues/272 
Those errors appears both in memcache and redis.

And we though it could be a good idea to let developers checks on others field than just the timestamp.

Your cache shouldn't know what is inside your entity, so it could not manually check fields on your entities.
But you're loading your entity here and looking if you send it back to the controller or loading a new one from the database, so I thought it was the place to check if the entity is valid.

lcobucci:
> And we though it could be a good idea to let developers checks on others field than just the timestamp.

You must keep in mind that the entities **shouldn't** know anything about persistence and caching.

About the L2C, the tablespace timestamp should be enough to validate if any change happened, if that is not happening it means we have a bug.

Can you try to reproduce this error on a functional test instead of changing the ORM?

PR #6055, opened 2016-09-27
[WIP] Fixed wrong operator use in condition

juniwalk:
Hello,
this fix will allow **Unit of Work** to stop early when there is nothing to do. Currently Doctrine starts _transaction_ and _commits_ it each time flush is called as this condition is invalid.


juniwalk:
Looks like the condition is okay, I just worked with entities in a bad way.

PR #6059, opened 2016-09-28
Remove final class comment

techibert:
The comment in this final should be removed. Git will take care of it.


Ocramius:
It's still supposed to be final-ish. We can't do it in 2.x due to BC boundaries (we didn't have an interface until 2.3).

PR #6074, opened 2016-10-06
Added afterFlush event

simPod:
You cannot persist another entity in postPersist. So you schedule these updates after the lifecycle has finished and dispatch it all from the queue. It is for better control over the whole operation

### DIFF
--- a/lib/Doctrine/ORM/Events.php
+++ b/lib/Doctrine/ORM/Events.php
@@ -158,6 +158,13 @@ final class Events
     const postFlush = 'postFlush';
 
     /**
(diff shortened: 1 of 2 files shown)


lcobucci:
@simPod the ORM event system was not made with the intention of binding domain things to it (like new entities).

Lifecycle callbacks should be avoided as @Ocramius explains in his [talk](https://www.youtube.com/watch?v=rzGeNYC3oz0) (you can also read the [slides too](http://ocramius.github.io/doctrine-best-practices/#/58)).

Having this said, I don't think we should have this new event.

simPod:
Thank you for reference, will check it as soon as I have a few minutes :)

PR #533, opened 2012-12-11
added DiscriminatorMapEntry mapping 

Padam87:
In Symfony2, if the main entity is in a separate bundle, from a vendor, changing the DiscriminatorMap is not a good idea.
Thats why we should be able to add entries to the map.

Mapping:

yaml: 

discriminatorMapEntry: keyHere

annotation:

@DiscriminatorMapEntry("keyHere")

xml:

<discriminator-map-entry name="keyHere" />

PR #6082, opened 2016-10-12
Added annotation DiscriminatorValue 

KorsaR-ZN:
| Q | A |
| --- | --- |
| Branch? | `master` |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | #2889, #533 |
| License | MIT |
| Doc PR | n/a |

New feature the `DiscriminatorValue` specifies the discriminator value for each class in discriminator map when it is automatically generated.

It also supports XML (`discriminator-value`) and Yaml (`discriminatorValue`) driver
## Example

``` php
/**
 * @ORM\Entity()
 * @ORM\DiscriminatorColumn(name="type", type="string")
 */
abstract class Vehicle
{
   ...
}

/**
 * @ORM\Entity()
 * @ORM\DiscriminatorValue("wheel4.car")
 */
class Car extends Vehicle
{
   ...
}

/**
 * @ORM\Entity()
 * @ORM\DiscriminatorValue("wheel4.quadricycle")
 */
class Quadricycle extends Vehicle
{
   ...
}

/**
 * @ORM\Entity()
 * @ORM\DiscriminatorValue("wheel2.motorbike")
 */
class Motorbike extends Vehicle
{
   ...
}
```
## Result

Automatically generated discriminator map as defaults:

``` php
{
   "wheel4.car": "Car",
   "wheel4.quadricycle": "Quadricycle",
   "wheel2.motorbike": "Motorbike"
}
```


KorsaR-ZN:
@Ocramius 
It will not be included in 2.6.0?

Ocramius:
@KorsaR-ZN unsure - didn't look at it yet, nor did evaluate if it's a valid/useful scenario

KorsaR-ZN:
@Ocramius 
This is useful in cases where automatically generated `DiscriminatorMap` and as the default key takes the class name in lowercase. 
This PR allows you to set key `DiscriminatorMap`  explicitly, that solves the problem:
- Collides for entities of same name in different namespace,
- Breaks in terrible way when class is renamed,
- Is ugly (generates unnecesarily too long discriminator columns).

Ocramius:
Sounds like an edge case, whereas defining the discriminator map in one location is relatively OK in my opinion.

tkleinhakisa:
If think that is a very nice feature to have.

The problem when defining the discriminator map in one location is that you introduce coupling between the main class and the sub classes.

Ex: i have an InvitationBundle with an Invitation entity which is the root of the inheritance. In another bundle i have extended the Invitation entity eg: ContactInvitation. I don't want to couple my invitation bundle to my contact bundle, so my only choice is to let doctrine generate the discriminator map, with the problems mentioned by @KorsaR-ZN (collides, breaks, ugly)

KorsaR-ZN:
I have same case as in @tkleinhakisa, so static definition `DiscriminatorMap` is not suitable...

Ocramius:
> Ex: i have an InvitationBundle with an Invitation entity which is the root of the inheritance. In another bundle i have extended the Invitation entity eg: ContactInvitation. I don't want to couple my invitation bundle to my contact bundle, so my only choice is to let doctrine generate the discriminator map, with the problems mentioned by @KorsaR-ZN (collides, breaks, ugly)

That's easily fixable with a metadata postLoad event listener, as far as I can see.

This feature seems to be well outside the Pareto rule.

I know you put some effort in it, but I will close this once I wrote a snippet that demonstrates how to handle these edge cases in userland.

KorsaR-ZN:
@Ocramius This feature also is supported in [Hibernate](http://docs.oracle.com/javaee/5/api/javax/persistence/DiscriminatorValue.html), in Doctrine, this function would be useful when creating a complex inheritance entities...

Ocramius:
Creating complex inheritances *IS* an edge case, and also discouraged
(regardless of doctrine)

Seb33300:
I am working on projects with different bundles too and I would like to see this feature in doctrine.
I think this will make code easier to read and understand.

lcobucci:
As discussed in #6612 we won't make it part of `v2.x` and it might happen for `v3`.

PR #6085, opened 2016-10-13
More specific named native query throw exceptions test

arjank:
- Makes sure the correct method is called in the test
- Verifies that the correct method is called by checking the exception message


Ocramius:
@arjank thanks!

PR #6088, opened 2016-10-17
Fix CS for OrderBy PHP annotation

teohhanhui:


teohhanhui:
Never mind. This seems to be the CS of the documentation. It is different from the tests.

PR #6089, opened 2016-10-18
Remove duplicated enum type comment from declaration.

damour:
Diff command generate duplicated comment declaration:
ALTER TABLE test CHANGE type type ENUM(\'enabled\', \'disabled\', \'bonus\', \'jackpot\') COMMENT \'(DC2Type:enum_status)\' NOT NULL COMMENT \'(DC2Type:enum_status)\''

### DIFF
--- a/docs/en/cookbook/mysql-enums.rst
+++ b/docs/en/cookbook/mysql-enums.rst
@@ -98,7 +98,7 @@ For example for the previous enum type:
 
         public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
         {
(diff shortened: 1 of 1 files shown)


Ocramius:
@damour can you please write a test that fails with the original version, and passes with your modified one?

Ocramius:
@damour seems good, but can you please also add a [`requiresSQLCommentHint`](https://github.com/doctrine/dbal/blob/ec8fc16604915110eea8046e3ea594c133aee880/lib/Doctrine/DBAL/Types/Type.php#L331-L344) that always returns `true` please?

damour:
@Ocramius done, thanks.

Ocramius:
@damour thanks!

PR #6094, opened 2016-10-20
[PMakowski] fix #5622-Mysql-5.7-requires-order-by-clause-in-select-DI…

### LINKED ISSUE #5622: Mysql 5.7 requires order by clause in select DISTINCT statement (Paginator)
Hi,

There seems to be an issue with the Paginator. In mysql 5.7 it is required to have the order by fields to be present in the select statement when using DISTINCT. I have situated where the problem at: 
https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryOutputWalker.php#L276

By preserving the order by clause as described in issue #2630 it now fails in mysql 5.7. I tested by simply adding the ordered fields to the select statement and everything worked again.

Astinus-Eberhard:
Fix for #5622 - seems to work for me


Ocramius:
Is this supposed to be closed? What is the resolution?

Astinus-Eberhard:
sorry, accidentally deleted branch

PR #6101, opened 2016-10-26
Add license checker to the build

malukenho:
- [x] ~~Add license checker to build~~ 
- [x] Fix license by file inconsistencies

I did separated commits in case of the license checker be considered overkill, them we can just revert it.


composer.json
```diff
@@ -25,7 +25,8 @@
     },
     "require-dev": {
         "symfony/yaml": "~2.3|~3.0",
-        "phpunit/phpunit": "^5.4"
+        "phpunit/phpunit": "^5.4",
+        "malukenho/docheader": "^0.1.4"
```
Ocramius: @malukenho one added dependency to just check some docblocks? I'm quite against this...

Ocramius:
@malukenho the fixes on the docblocks are OK, but:

 * no added dependency, please
 * it is hardly a vital bit, so we should NOT add it to the build process

malukenho:
@Ocramius there's some file under `tests` that have the license on header. It should add license to files under `tests` or remove it to preserve consistency?

I agree with you about the dependency, but license is very important to OSS projects, and it's a vital part of any OSS project as it says who can use or not this code.

Ocramius:
@malukenho I'd rather strip the licence from all files than having to deal with this sort of complexity due to vogonism. It's MIT, after all, and we have signed releases, should anybody ever complain about authorship.

Ocramius:
@malukenho LGTM 👍 Thanks!

PR #6109, opened 2016-10-28
Support for ClassMetadataInfo::isSingleAssociationJoinColumnNullable()

holtkamp:
Some context here: https://github.com/doctrine/doctrine2/issues/5696, when generating getters / setters etc based on metadata, it is good to know whether an association allows null values:

```php
class Image
{
    /**
     * Multiple Images "can" belong to one Article, it is not mandatory, so NULL is allowed, for example to unset.
     */
    public function setArticle(Article $article = null){
        //..code here
    }

}

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -2893,6 +2893,24 @@ class ClassMetadataInfo implements ClassMetadata
     }
 
     /**
+     * Checks whether the single association join column (if any) is nullable.
+     *
+     * @param string $fieldName
+     *
+     * @return bool
+     *
+     * @throws MappingException
+     */
+    public function isSingleAssociationJoinColumnNullable($fieldName)
+    {
+        if ( ! $this->isAssociationWithSingleJoinColumn($fieldName)) {
+            throw MappingException::noSingleAssociationJoinColumnFound($this->name, $fieldName);
+        }
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Ma27** [comment]
ok, I understand the usecase so far, but I'm unsure whether the doctrine maintainers really want to expose their public API as it's currently planned to reduce the ORM's featureset and yet another public method causes more public API (see #6211)

**Comment by holtkamp** [comment]
Well... from my perspective it is just a helper function to ease up access to publicly available ClassMetadataInfo(rmation), which is useful for other utilities like code generation... 

Yes it does add a public method. But that public method is just informative / returning a boolean... about information that is stored in an (already) [public array](https://github.com/doctrine/doctrine2/blob/06691e01505a0bccca7f63c0efc2e6c4b8989245/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L532)

So it does not expose more than currently is already exposed...?

**Comment by guilhermeblanco** [comment]
This is already done as part of the 3.0 branch through its public API.
Access to this information will be through:

```php
$metadata->getProperty('article')->getJoinColumns()[0]->isNullable()
```

holtkamp:
@guilhermeblanco, ok, I think I get your point, but that is the ```3.0``` branch, due somewhere in 201X. This was just well intended, small, isolated and properly tested contribution to the ```2.x``` branch which would be useful directly. Just don't understand this, now people will have such helper functions in their own code. 

Will lie down and [try not to cry](http://i0.kym-cdn.com/entries/icons/facebook/000/012/073/7686178464_fdc8ea66c7.jpg) now 😄

Ocramius:
@holtkamp sorry, that's the destiny of most API that isn't going to be used internally :-\

PR #6110, opened 2016-10-31
Clear $this->collection even when empty, to reset keys

steevanb:
To have same behavior as when $this->empty() === false, that is to say reset $this->collection keys


Ocramius:
I've manually cherry-picked this into `2.5` and merged into it and also into `master`. Thanks @steevanb: good catch! 👍 

`master`: doctrine/doctrine2@c1480595930166a77a597b65dc8850509d57a934
`2.5`: doctrine/doctrine2@4d699789a25990f5514461f6b17b50f887d65f31

steevanb:
Thanks !

Do you know when 2.5.6 will be available, with this fix ?

Ocramius:
At earliest occasion.

PR #6120, opened 2016-11-06
Fixed typos regarding the strategy for composite keys (xml mapping)

mschroeder:
The strategy for composite primary keys must be "NONE" although the id generator class is \Doctrine\ORM\Id\AssignedGenerator. Calling the strategy "ASSIGNED" is misleading.


Ocramius:
👍 thanks!

PR #6121, opened 2016-11-06
Static Code Analysis with Php Inspections (EA Extended)

kalessil:
Couple of findings

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1629,7 +1629,6 @@ class ClassMetadataInfo implements ClassMetadata
                     ? $joinColumn['fieldName']
                     : $joinColumn['name'];
             }
-            unset($joinColumn);
 
             if ($uniqueConstraintColumns) {
                 if ( ! $this->table) {
@@ -1755,7 +1754,6 @@ class ClassMetadataInfo implements ClassMetadata
                 $mapping['relationToSourceKeyColumns'][$joinColumn['name']] = $joinColumn['referencedColumnName'];
                 $mapping['joinTableColumns'][] = $joinColumn['name'];
             }
-            unset($joinColumn);
 
             foreach ($mapping['joinTable']['inverseJoinColumns'] as &$inverseJoinColumn) {
                 if (empty($inverseJoinColumn['name'])) {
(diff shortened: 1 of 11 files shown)


lib/Doctrine/ORM/Tools/Console/Command/ConvertDoctrine1SchemaCommand.php
```diff
@@ -180,6 +180,7 @@ public function convertDoctrine1Schema(array $fromPaths, $destPath, $toType, $nu
                 );
             }
         }
+        unset($dirName);
```
Ocramius: Not sure if I like this inspection. `$dirName` is not used after this location: let the stack garbage collection do its job.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1629,6 +1629,7 @@ protected function _validateAndCompleteOneToOneMapping(array $mapping)
                     ? $joinColumn['fieldName']
                     : $joinColumn['name'];
             }
+            unset($joinColumn);
```
Ocramius: Same as below: this inspection is (IMO) not a good idea.

Ocramius (changes requested):

lib/Doctrine/ORM/Tools/Console/Command/ConvertDoctrine1SchemaCommand.php
kalessil: It's more tricky: https://3v4l.org/QKOKH - here unset should clean the scope after foreach.

Ocramius: I know why this is applied, it just doesn't make sense if the variables are unused. I'd rather prefer extracting the loop into a private method instead.

kalessil: I see, good approach btw. Reverting the change.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
kalessil: Reverting. I also curious why it's not a good idea, why?

Ocramius: It's just an additional instruction (also to read - not worried about performance)

kalessil:
@Ocramius : I reverted usset constructs here, but checks still saying that you requested changes.

Ocramius:
@kalessil thanks for this, merging!

Issue #6123, opened 2016-11-09
Remove on a specific Entity and a subsequent find will retrieve the same entity with a Managed state


bertuz:
I don't know if it's the expected behavior, beh when removing a specific Entity, it is removed from the Identity Map. This implies that the subsequent finds within a single UoW will fetch the Entity with a 'MANAGED' state. Is this the expected behavior?

```

/**
     * INTERNAL:
     * Schedules an entity for deletion.
     *
     * @param object $entity
     *
     * @return void
     */
    public function scheduleForDelete($entity)
    {
        $oid = spl_object_hash($entity);

        if (isset($this->entityInsertions[$oid])) {
            if ($this->isInIdentityMap($entity)) {
                $this->removeFromIdentityMap($entity);
            }
```

Ocramius:
@bertuz looks like a bug. Can you try reproducing it against https://github.com/doctrine/doctrine2/tree/d3f6c5ec70aac4b029a4b61ecf1e2ba61a1a4a6d/tests/Doctrine/Tests/ORM/Functional/Ticket please? (see that dir for examples)

bertuz:
of cource, by tomorrow

bertuz:
@Ocramius I was about to add a test into `Tests/ORM/Funcitonal/Ticket` but I don't know where to get the ticket number in order to name the file + open a pull req. Can you please show me the right way to that?

Ocramius:
You can call it like this issue (6123)

bertuz:
here you go https://github.com/doctrine/doctrine2/pull/6126

xificurk:
Fixed by #11428

PR #6126, opened 2016-11-10
#6123 removing and retrieving an entity will still retrieve the removed entity

bertuz:


lcobucci:
@bertuz @Ocramius moving the removal from identity map to `UnitOfWork#removeFromIdentityMap()` solves this issue but breaks `UnitOfWorkTest#testRemovedAndRePersistedEntitiesAreInTheIdentityMapAndAreNotGarbageCollected()`.

Not sure what's the expected behaviour though.

bertuz:
IMHO I think that's the point: a subsequent find does not retrieve the same entity instance. Hence, the identity map is not keeping it until flushing. As a developer that's the functional behavior I would expect when removing. I guess is the developer's responsibility checking if an entity is in a remove state within a UoW.

PR #6128, opened 2016-11-14
added IdGeneratorInterface and generateId for custom entity-managers

addiks:
Hello doctrine maintainers,

while trying to write a replacement for doctrine's own entity-manager (to experiment with different behavior regarding transactions) i have noticed that currently there are hard-coded dependencies to doctrine's own entity-manager in the id-generator classes. In order to make the id-generation process work with third-party entity-managers while still maintaining full backwards-compatibility, i have added an interface called IdGeneratorInterface which specifies a "generateId" method that depends only on the EntityManagerInterface instead of EntityManager, allowing third-party entity-managers to be used with it.

The AbstractIdGenerator class still exists as an adapter translating the new method to the old (if possible). This allows full backward compatibility because other (old) third party id-generators (like these defined in the unit-tests) extend the AbstractIdGenerator which in turn translates all calls from "generateId" to "generate" if doctrine's own entity-manager is used. The only situation where this fails is when an old id-generator (extending AbstractIdGenerator) is used with a third-party entity-manager, which is a situation that is currently also not supported anyway. Old third-party id-generators can (in almost all cases) simply be upgraded to use the IdGeneratorInterface and the "generateId" method to be able to be used with third-party entity-managers.

I have made sure that all tests pass on my computer with the inmemory sqlite driver.

### DIFF
--- a/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
+++ b/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
@@ -20,11 +20,14 @@
 namespace Doctrine\ORM\Id;
 
 use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\Id\IdGeneratorInterface;
+use Doctrine\ORM\ORMException;
+use Doctrine\ORM\EntityManagerInterface;
 
-abstract class AbstractIdGenerator
+abstract class AbstractIdGenerator implements IdGeneratorInterface
 {
     /**
-     * Generates an identifier for an entity.
+     * {@inheritDoc}
      *
      * @param EntityManager $em
      * @param \Doctrine\ORM\Mapping\Entity $entity
@@ -32,10 +35,29 @@ abstract class AbstractIdGenerator
      */
     abstract public function generate(EntityManager $em, $entity);
 
+    public final function generateId(EntityManagerInterface $em, $entity)
+    {
+        /* @var $id mixed */
+        $id = null;
+
+        if ($em instanceof EntityManager) {
+            $id = $this->generate($em, $entity);
+
+        } else {
+            throw new ORMException(sprintf(
+                "Tried to use non-doctrine entity-manager %s with old id-generator %s which is not supported! ".
+                "This id-generator must be upgraded to use the %s to make it work with different entity-managers!",
+                get_class($em),
+                get_class($this),
+                IdGeneratorInterface::class
+            ));
+        }
+
+        return $id;
+    }
+
     /**
-     * Gets whether this generator is a post-insert generator which means that
-     * {@link generate()} must be called after the entity has been inserted
-     * into the database.
+     * {@inheritDoc}
      *
      * By default, this method returns FALSE. Generators that have this requirement
      * must override this method and return TRUE.
--- a/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
+++ b/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
@@ -37,13 +37,7 @@ abstract class AbstractIdGenerator implements IdGeneratorInterface
 
     public final function generateId(EntityManagerInterface $em, $entity)
     {
-        /* @var $id mixed */
-        $id = null;
-
-        if ($em instanceof EntityManager) {
-            $id = $this->generate($em, $entity);
-
-        } else {
+        if (!$em instanceof EntityManager) {
             throw new ORMException(sprintf(
                 "Tried to use non-doctrine entity-manager %s with old id-generator %s which is not supported! ".
                 "This id-generator must be upgraded to use the %s to make it work with different entity-managers!",
@@ -53,7 +47,7 @@ abstract class AbstractIdGenerator implements IdGeneratorInterface
             ));
         }
 
-        return $id;
+        return $this->generate($em, $entity);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


lcobucci:
@addiks I have mixed feelings regarding your changes... I understand that you tried to not insert a BC break but at the same time it's weird to have a method `generateId` on an `IdGenerator` (I just don't expect to have an `IdGenerator` that generates other things than an ID).
And I'm not sure if it's worth to have this "temporary complexity"...

addiks:
Example:
`
class GenerateMyThings implements IdGenerator, NameGenerator
{
    public function generateId()
    {
        // Generates an id
    }

    public function generateName()
    {
        // Generates a name
    }
}
`
The method-name "generate" makes much more sense if the class extends the AbstractIdGenerator because the class is practically guaranteed to be just an id-generator. The name "generateId" makes more sense in the context of an IdGeneratorInterface because here all we know is that the class can also generate id's besides maybe other things. If the method was just named "generate" it would be unclear *what* would be generated here and it could possibly cause name-conflicts. I don't want to forbid a class to implement multiple (generator-) interfaces. Something like that might be considered bad design by some, but that is a decision the user has to make and i don't want to impose my idea of good design on anybody by causing name-conflicts that otherwise could easily be avoided. Also the method-name "generate" would conflict with the generate-method defined in AbstractIdGenerator, which would kill of the backwards compatibility.

As for if it's worth the extra complexity, to me it does not look like that much more complexity because everything still works just the same as before, It's just a different name to be called (while the old one still works). The only additional "weight" is the AbstractIdGenerator for backwards compatibility. I think having this is worth it because it allows users to use their own implementation of an entity-manager.

In my case i created an [alternative entity-manager](https://github.com/addiks/doctrine2_tweaks) to experiment with a different behavior regarding transactions for better use with batch-processing and recovering from a faulty state but which should probably not be the default because for beginners it is the much more safe option to just stop if any error happens.

If you (or anybody else here) is at SymfonyCon in Berlin next week, we could talk in person. I'm looking forward to interesting discussions.

Ocramius:
I like how you mitigated the changes, but I don't know yet if I like working around everything like that. Possibly a good idea to allow dropping the abstract class for 3.x though.

Just one thing: please don't merge `master` into your branch: rebase instead.

Ocramius:
One thing missing: a test that passes an `EntityManagerInterface` to the id generator (regression checking, mostly)

addiks:
Thanks for pointing me to git rebase, i didn't knew about that. I have now rebased my commit on doctrine master, so the previous commits vanished and were replaced with a new one. I hope that does not confuse anybody.

I will look into writing a test for using the id-generators with another entity-manager, but first i need to make myself familiar with the current set of tests in doctrine to find out where and how to best put that test.

Ocramius:
> first i need to make myself familiar with the current set of tests in doctrine

Try looking for existing tests around the identifier generators. Unsure if they are segregated (one per SUT)

lib/Doctrine/ORM/Id/AbstractIdGenerator.php
```diff
@@ -20,22 +20,44 @@
...
      * @return mixed
      */
     abstract public function generate(EntityManager $em, $entity);
 
+    public final function generateId(EntityManagerInterface $em, $entity)
+    {
+        /* @var $id mixed */
+        $id = null;
+
+        if ($em instanceof EntityManager) {
```
malukenho:
```php
if ($em instanceof EntityManager) {
    return $this->generate($em, $entity);
}
 
throw new ORMException(sprintf(
     "Tried to use non-doctrine entity-manager %s with old id-generator %s which is not supported! ".
     "This id-generator must be upgraded to use the %s to make it work with different entity-managers!",
     get_class($em),
     get_class($this),
     IdGeneratorInterface::class
));
```

addiks: Right, i forgot about exit early when writing this. My bad. I would switch the two cases around so that when more code get's added to the valid branch it would not be unnecessarily indented. That way the exception would act more like a guard clause. It would be easier to add more such clauses or any other code beneath it. Would you agree?
```
if (!$em instanceof EntityManager) {
    throw new ORMException(sprintf(
         "Tried to use non-doctrine entity-manager %s with old id-generator %s which is not supported! ".
         "This id-generator must be upgraded to use the %s to make it work with different entity-managers!",
         get_class($em),
         get_class($this),
         IdGeneratorInterface::class
    ));
}

return $this->generate($em, $entity);
```

Ocramius: Yes, the exceptional case should be before the successful one (last)

addiks:
I've now modified the assigned-id-generator-test to also execute with a test-double of the entity-manager-interface. I don't know if this is to your standards so just tell me if anything is wrong here. I also tried to modify the sequence-generator-test, but that one seems very tightly coupled to the actual db-connection, which i think would be hard to write test-doubles for.

Majkl578:
Hi @addiks, just wanted to let you know that we've made some changes to id generators in `develop`:
- they were moved to Sequencing namespace;
- we're now typehinting EntityManagerInterface, not EntityManager;
- generators have been generalized as more generic value generators so they could be used for generating values even for non-ids (this is not yet implemented);
- we now have a value generation plan to encapsulate value generation on multiple places.

With these changes, I think this PR is obsoleted for 3.x and adding such code in 2.x wouldn't be worth it. Let me know what you think and whether this PR could be closed.

Ocramius:
Closing as per @Majkl578's comment. Indeed, the scope changed here, so the patch would need to be thought through (new discussion/pr if relevant), while this thread should end here.

Issue #6129, opened 2016-11-16
XmlDriver incorrectly parses boolean option values


kalimatas:
I have a mapping for a column like that:

```
<field name="alpha2Code" column="country_code" type="string" length="2" column-definition="CHAR(2) NOT NULL">
    <options>
        <option name="fixed">false</option>
    </options>
</field>
```

Despite that I specified `false` as a value of `fixed` option, it will be `true`, because in `XmlDriver::_parseOptions` method the string `false` will be casted to `true` with `(bool)`:

```
if ($option->count()) {
    $value = $this->_parseOptions($option->children());
} else {
    // here
    $value = (string) $option;
}
```

It probably should be:

```
$value = $this->evaluateBoolean((string) $option);
```

Ocramius:
@kalimatas can this be converted into a test case?

kalimatas:
I'm on it right now.

Ocramius:
Tests in #6130

kalimatas:
I would like also to fix it, though I'm not sure, if maintaining an array of option names, that are supposed to be boolean in the `_parseOptions()` method , is a nice solution.

Ocramius:
@kalimatas give it a shot - it can be a quick fix with a later refactoring, but if there's a possible BC break, we need to expose that as well.

kalimatas:
I created a generic test for all drivers.

Ocramius:
Closing as per work done in #6130

Ocramius:
Note: this is only merged into `2.6.0` due to risk of regressions in the mapping driver change.

PR #6130, opened 2016-11-16
#6129 Added unit test for boolean option values.

kalimatas:
It fails now. In `XmlDriver::_parseOptions` we need somehow to maintain a list of options, that are supposed to be boolean, and then call `$this->evaluateBoolean()` on them.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -653,8 +653,6 @@ class XmlDriver extends FileDriver
     {
         $array = array();
 
-        $booleanOptions = ['unsigned', 'fixed'];
-
         /* @var $option SimpleXMLElement */
         foreach ($options as $option) {
             if ($option->count()) {
@@ -663,11 +661,11 @@ class XmlDriver extends FileDriver
                 $value = (string) $option;
             }
 
-            $attr = $option->attributes();
+            $attributes = $option->attributes();
 
-            if (isset($attr->name)) {
-                $attrName = (string) $attr->name;
-                $array[$attrName] = in_array($attrName, $booleanOptions)
+            if (isset($attributes->name)) {
+                $nameAttribute = (string) $attributes->name;
+                $array[$nameAttribute] = in_array($nameAttribute, ['unsigned', 'fixed'])
                     ? $this->evaluateBoolean($value)
                     : $value;
             } else {
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -653,6 +653,8 @@ class XmlDriver extends FileDriver
     {
         $array = array();
 
+        $booleanOptions = ['unsigned', 'fixed'];
+
         /* @var $option SimpleXMLElement */
         foreach ($options as $option) {
             if ($option->count()) {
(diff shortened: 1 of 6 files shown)


Ocramius:
Ref: #6129

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -132,14 +132,21 @@ public function testEmbeddedMapping()
...
      * @expectedExceptionMessage Invalid mapping file 'Doctrine.Tests.Models.Generic.SerializationModel.dcm.xml' for class 'Doctrine\Tests\Models\Generic\SerializationModel'.
      */
     public function testInvalidMappingFileException()
     {
         $this->createClassMetadata('Doctrine\Tests\Models\Generic\SerializationModel');
     }
 
+    public function testBooleanValuesForOptionIsSetCorrectly()
+    {
+        $class = $this->createClassMetadata('Doctrine\Tests\ORM\Mapping\User');
```
Ocramius: Can use `::class` for new code

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -132,14 +132,21 @@ public function testEmbeddedMapping()
...
-     * @expectedException Doctrine\Common\Persistence\Mapping\MappingException
+     * @expectedException \Doctrine\Common\Persistence\Mapping\MappingException
      * @expectedExceptionMessage Invalid mapping file 'Doctrine.Tests.Models.Generic.SerializationModel.dcm.xml' for class 'Doctrine\Tests\Models\Generic\SerializationModel'.
      */
     public function testInvalidMappingFileException()
     {
         $this->createClassMetadata('Doctrine\Tests\Models\Generic\SerializationModel');
     }
 
+    public function testBooleanValuesForOptionIsSetCorrectly()
```
Ocramius: Can you please add an `/** @group #6129 */` or similar?

Ocramius:
@kalimatas thanks for converting it into an actual failing scenario!

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -653,6 +653,8 @@ private function _parseOptions(SimpleXMLElement $options)
     {
         $array = array();
 
+        $booleanOptions = ['unsigned', 'fixed'];
```
Ocramius: Since this constant is not used anywhere else, it can be inlined into the `in_array()` call

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -664,7 +666,10 @@ private function _parseOptions(SimpleXMLElement $options)
             $attr = $option->attributes();
 
             if (isset($attr->name)) {
-                $array[(string) $attr->name] = $value;
+                $attrName = (string) $attr->name;
```
Ocramius: Please don't abbreviate. I know the previous code did, but we should really call it `$attributeName` or such

tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
```diff
@@ -245,6 +248,26 @@ public function testIdentifier($class)
     }
 
     /**
+     * @group #6129
+     *
+     * @depends testIdentifier
```
Ocramius: Shouldn't need `@depends`

Ocramius (changes requested):

kalimatas:
@Ocramius fixed

Ocramius:
@kalimatas thanks! This looks good, merging!

👍

PR #6142, opened 2016-11-22
Fix incompatibility w/ PHP7.2+

mbeccati:
Mock_ParserResult_*::getParameterMappings() was returning null, which was then passed to count() on Query.php:308, causing a "Parameter must be an array or an object that implements Countable" error.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/QueryCacheTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/QueryCacheTest.php
@@ -135,7 +135,9 @@ class QueryCacheTest extends OrmFunctionalTestCase
                     ->method('execute')
                     ->will($this->returnValue( 10 ));
 
-        $parserResultMock = $this->createMock(ParserResult::class);
+        $parserResultMock = $this->getMockBuilder(ParserResult::class)
+                                 ->setMethods(array('getSqlExecutor'))
+                                 ->getMock();
         $parserResultMock->expects($this->once())
                          ->method('getSqlExecutor')
                          ->will($this->returnValue($sqlExecMock));


lcobucci:
@mbeccati that sounds like a `phpunit-mock` issue, since when you call `createMock()` with a concrete class it should stub the public interface of the object...

mbeccati:
> By default, all methods of the original class are replaced with a dummy implementation that just returns null 

Which is not what you'd want there, @lcobucci

Ocramius:
Got a link to the failure? I don't understand the reason for this change.

lcobucci:
> Which is not what you'd want there, @lcobucci

We just don't care here if we stub everything or not, you might right when saying that we're just doing that for the method `getSqlExecutor()` but there's no reason to change doctrine's test suite if it's passing on all builds.

> Got a link to the failure? I don't understand the reason for this change.

It's failing on all `nightly` builds (like https://travis-ci.org/doctrine/doctrine2/jobs/177333526)

mbeccati:
@Ocramius https://revive.beccati.com/bamboo/browse/PHP-DOCTR-MAS-919/test/case/11810654 w/ php-src master, since count() has been restricted to work only on array|Countable.

Ocramius:
Merging, since I feel like digging further into this is just going to be more pain. 

The fix is simple/easy: let's roll with it, since it's a test-suite change only.

mbeccati:
Thanks @Ocramius. Since I've did that digging already, I'll try to write it down for posterity:

1. `\Doctrine\ORM\Query\ParserResult` has a private property `$_parameterMappings`, which is initialised as an empty array;
1. such property is never overwritten with non-array values;
1. `\Doctrine\ORM\Query\ParserResult::getParameterMappings()` returns its value, which means it is always returning an array;
1. `Doctrine/ORM/Query.php` on line 306 calls `getParameterMappings()` and on line 308 calls `count()` on its return value, which is guaranteed to be an array;
1. `Doctrine/Tests/ORM/Functional/QueryCacheTest.php` was injecting a stub of `ParserResult` that returned `null` for all methods but `getSqlExecutor()`;
1. that means that `getParameterMappings()` was unexpectedly returning `null` and causing the above `count()` to fail on PHP 7.2 since the introduction of [this RFC](https://wiki.php.net/rfc/counting_non_countables);
1. the test was fixed by using a mock that only overrides `getSqlExecutor()`.

PR #5973, opened 2016-08-12
This should fix MySQL 5.7 issues caused by ONLY_FULL_GROUP_BY

wucdbm:
First of all, I have to say that this is my first pull request ever created

This should fix MySQL 5.7 issues caused by ONLY_FULL_GROUP_BY

The proposed code works for me, maybe others could share opinions on a fix for this, it is very annoying when working with new versions of MySQL

PR #6143, opened 2016-11-22
fixed #5622 - mysql 5.7 ONLY_FULL_GROUP_BY

### LINKED ISSUE #5622: Mysql 5.7 requires order by clause in select DISTINCT statement (Paginator)

chihiro-adachi:
#5622

I fixed it with reference to #5973 
please check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ before_script:
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
   - if [[ $PHPSTAN = 1 ]]; then composer require --dev --prefer-stable phpstan/phpstan:^0.7 symfony/console:^3.0; fi
+  - if [ "$MYSQL_VERSION" == "5.7" ]; then bash ./tests/travis/install-mysql-5.7.sh; fi;
   - if [[ $DB == "mysql" || $DB == "mariadb" ]]; then mysql -e "CREATE SCHEMA doctrine_tests; GRANT ALL PRIVILEGES ON doctrine_tests.* to travis@'%'"; fi;
 
 script:
@@ -46,6 +47,13 @@ matrix:
       env:
         - DB=pgsql
         - PHPSTAN=1
+
+    - php: 7.1
+      env: DB=mysql MYSQL_VERSION=5.7
+      sudo: required
+    - php: nightly
+      env: DB=mysql MYSQL_VERSION=5.7
+      sudo: required
 
   allow_failures:
     - php: nightly
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "ext-pdo": "*",
         "doctrine/collections": "^1.4",
         "doctrine/dbal": ">=2.5-dev,<2.7-dev",
-        "doctrine/instantiator": "~1.0.1",
+        "doctrine/instantiator": "~1.1",
         "doctrine/common": "^2.7.1",
         "doctrine/cache": "~1.6",
         "doctrine/annotations": "~1.4",
--- a/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryOutputWalker.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryOutputWalker.php
@@ -43,6 +43,8 @@ use Doctrine\ORM\Query\AST\SelectStatement;
  */
 class LimitSubqueryOutputWalker extends SqlWalker
 {
+    private const ORDER_BY_PATH_EXPRESSION = '/(?<![a-z0-9_])%s\.%s(?![a-z0-9_])/i';
+
     /**
      * @var \Doctrine\DBAL\Platforms\AbstractPlatform
      */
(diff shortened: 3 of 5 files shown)


ruudk:
I just tested this in our project and it seems to solve the issues. Really awesome! :+1:

Would be great if this could get merged.

kissifrot:
Confirmed here too, +1 for a merge :)

netiul:
Looks good. Let's merge this!

chihiro-adachi:
@Ocramius 
this pr Is there something wrong?
anything else needed?

thank you.

Ocramius:
Simply didn't get to it, sorry.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

mauriciogtaborda:
lgtm :+1:

nateevans:
is this on deck to be pulled in anytime soon?

fefas:
+1

paaacman:
+1

Einenlum:
+1

lcobucci:
@Ocramius I've synced this PR with `master` and added MySQL 5.7 to Travis. I'm  just not fully convinced that this is the way to go (regex stuff and also the impact of wrapping the subquery on the other platforms).

teohhanhui:
> the impact of wrapping the subquery on the other platforms

Any SQL-compliant platform should require this fix.

Arkemlar:
why still not merge?

Ocramius:
@Arkemlar because I'd rather not have support for MySQL 5.7 than having security issues due to regex-based SQL manipulation.

The patch may be good, but it will need further checking for edge-cases in the newly added conditionals.

Also, tests were only modified so far, but none was added.

Ocramius:
Looking good according to @lcobucci's changes: will likely merge today.

Ocramius (approved):
LGTM :+1: :ship:

teohhanhui:
Will this be merged into stable versions as a bugfix?

Ocramius:
@chihiro-adachi this is already in `2.6.0`

teohhanhui:
Is `2.6.0` released yet? I can't find it anywhere. But I mean this should be treated as a bug, right? If so, might it be fixed in the currently supported versions?

lcobucci:
No, mysql 5.7 is just officially supported in 2.6 (unstable version).

Ocramius:
@teohhanhui no, newer mysql versions broke BC and we simply support them in a newer version only.

teohhanhui:
I thought this bug is also present when using other SQL-compliant platforms such as PostgreSQL... Because this bug only manifested itself when MySQL started becoming more SQL-compliant.

Ocramius:
@teohhanhui yes, but only MySQL crashes hard. So far, the only issue with such queries was non-deterministic result due to ordering.

ureimers:
@teohhanhui If you can't wait for Doctrine 2.6 to be released you still have the hackish way of overriding the Autoloader in your composer.json:

```
"autoload": {
        [...]
        "files": ["src/LimitSubqueryOutputWalkerPatch.php"]
    },
```

Just put the fixed source from this PR into the `LimitSubqueryOutputWalkerPatch.php`. When composer dumps your autoloader the class will take precedence over the Doctrine 2.5 one. Until Doctrine 2.6 is available you use this to temporarily fix the MySQL 5.7 incompatibility of the Paginator.

But keep in mind that this is pretty hackish. On the other hand if a project is blocked because of this you sometimes just don't have a choice.

InfopactMLoos:
Is this not backported to 2.5 ? Since my company uses Ubuntu 16.04 LTS and the 2.6 branch does not support PHP 7.0 I cannot upgrade to 2.6. @ureimers solution works but I would prefer a cleaner solution if possible. I tried using the patch generated by this pull request but it is not compatible with 2.5. Now trying to strip down this patch so the changes can be applied to 2.5 but was wondering why this bugfix is 2.6 only.

Ocramius:
@InfopactMLoos no, you will need to upgrade to `2.6`. Upgrading PHP version is easy peasy via http://ppa.launchpad.net/ondrej/php/ubuntu/pool/main/p/php7.1/

 > why this bugfix is 2.6 only.

MySQL 5.7 support is an addition.

InfopactMLoos:
Yeah I know how to do that through ppa it's easy indeed but company policy is to keep everything LTS compatible. We try to minimize customizations to base server images. Ok no worries I just completed rolling the patch and it seems to work fine. Was just wondering why, anyway thanks for the quick response!

shkkmo:
It seems like there should be a version of Doctrine that supports both PHP 7.0 and MySQL given that those are the available versions in Ubuntu 16.04

Ocramius:
Won't happen, as that is just pushing back the ecosystem with the excuse that the ecosystem is already being pushed back.

You can simply rely on the PPAs by ondrej, which reflect PHP core development, on which you can rely for security maintenance as well.

PR #6146, opened 2016-11-24
Allow EntityManager::transational() to return non-null values

enumag:


lcobucci:
@enumag could you please provide the reasons you need this change and also test cases for it?

Ocramius:
The correct fix would be to return anything from the executed callable, but that's not fixable, as it is an interface BC break. See https://github.com/doctrine/doctrine2/blob/bfa9a31ad77ce769f8e78ab70c2083e75f623c7a/lib/Doctrine/ORM/EntityManagerInterface.php#L73-L84

enumag:
@Ocramius Could it be changed in Doctrine 3.0?

Ocramius:
Yes, can be done. You can send a patch against `develop`, changing the interface too.

enumag:
Alright, thank you!

enumag:
Note for myself: Create a second PR for Doctrine/DBAL with the same thing for Connection::transactional().

enumag:
The transactional function is mentioned here in the docs: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/transactions-and-concurrency.html#approach-2-explicitly

PR #1392, opened 2015-04-23
transactional() wrapper corrupts return values

DHager:
The EntityManager::transactional() method has undocumented behavior, which unexpectedly rewrites all `false`-ish return values into `true` before passing them on:

```
$result = $this->_em->transactional(function (){
    return array();
}
assert(is_array($result)); // Fails
```

I think there's a very strong case to be made that this is undesirable, **however** it's been in existence for a few years now (since DDC-1125) and it's very likely at least a few users have code that _relies_ on the undocumented behavior.

This PR represents the most direct fix, but as a practical matter we may want to improve the documentation instead. If that is the decision, I'd be happy to create a separate PR for documentation changes.

PR #6147, opened 2016-11-24
EntityManager::transactional() returns value returned by the closure

enumag:
As requested this PR replaces #6146 and is meant for Doctrine 3.0.

`Doctrine\DBAL\Connection::transactional()` already works this way in the master branch of `doctrine/dbal`.

### DIFF
--- a/docs/en/reference/transactions-and-concurrency.rst
+++ b/docs/en/reference/transactions-and-concurrency.rst
@@ -104,18 +104,11 @@ functionally equivalent to the previously shown code looks as follows:
         $em->persist($user);
     });
 
-.. warning::
-
-    For historical reasons, ``EntityManager#transactional($func)`` will return
-    ``true`` whenever the return value of ``$func`` is loosely false.
-    Some examples of this include ``array()``, ``"0"``, ``""``, ``0``, and
-    ``null``.
 
 The difference between ``Connection#transactional($func)`` and
 ``EntityManager#transactional($func)`` is that the latter
 abstraction flushes the ``EntityManager`` prior to transaction
-commit and rolls back the transaction when an
-exception occurs.
+commit.
 
 .. _transactions-and-concurrency_exception-handling:
 
@@ -241,15 +234,15 @@ either when calling ``EntityManager#find()``:
     <?php
     use Doctrine\DBAL\LockMode;
     use Doctrine\ORM\OptimisticLockException;
-    
+
     $theEntityId = 1;
     $expectedVersion = 184;
-    
+
     try {
         $entity = $em->find('User', $theEntityId, LockMode::OPTIMISTIC, $expectedVersion);
-    
+
         // do the work
-    
+
         $em->flush();
     } catch(OptimisticLockException $e) {
         echo "Sorry, but someone else has already changed this entity. Please apply the changes again!";
@@ -262,16 +255,16 @@ Or you can use ``EntityManager#lock()`` to find out:
     <?php
     use Doctrine\DBAL\LockMode;
     use Doctrine\ORM\OptimisticLockException;
-    
+
     $theEntityId = 1;
     $expectedVersion = 184;
-    
+
     $entity = $em->find('User', $theEntityId);
-    
+
     try {
         // assert version
         $em->lock($entity, LockMode::OPTIMISTIC, $expectedVersion);
-    
+
     } catch(OptimisticLockException $e) {
         echo "Sorry, but someone else has already changed this entity. Please apply the changes again!";
     }
@@ -310,7 +303,7 @@ See the example code, The form (GET Request):
 
     <?php
     $post = $em->find('BlogPost', 123456);
-    
+
     echo '<input type="hidden" name="id" value="' . $post->getId() . '" />';
     echo '<input type="hidden" name="version" value="' . $post->getCurrentVersion() . '" />';
 
@@ -321,7 +314,7 @@ And the change headline action (POST Request):
     <?php
     $postId = (int)$_GET['id'];
     $postVersion = (int)$_GET['version'];
-    
+
     $post = $em->find('BlogPost', $postId, \Doctrine\DBAL\LockMode::OPTIMISTIC, $postVersion);
 
 .. _transactions-and-concurrency_pessimistic-locking:
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -222,12 +222,8 @@ use Doctrine\Common\Util\ClassUtils;
     /**
      * {@inheritDoc}
      */
-    public function transactional($func)
+    public function transactional(callable $func)
     {
-        if (!is_callable($func)) {
-            throw new \InvalidArgumentException('Expected argument of type "callable", got "' . gettype($func) . '"');
-        }
-
         $this->conn->beginTransaction();
 
         try {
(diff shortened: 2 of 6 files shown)


lib/Doctrine/ORM/EntityManagerInterface.php
```diff
@@ -81,7 +81,7 @@ public function beginTransaction();
      *
      * @param callable $func The function to execute transactionally.
      *
-     * @return mixed The non-empty value returned from the closure or true instead.
+     * @return mixed The value returned from the closure.
      */
     public function transactional($func);
```
lcobucci: Since it's a major release you can add `callable` as type-hint, right?

enumag:
@Ocramius The tests failed quite miserably but I don't think it's related to my changes.

lib/Doctrine/ORM/EntityManagerInterface.php
enumag: Yeah I think so. Thanks for the suggestion.

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -104,18 +104,11 @@ functionally equivalent to the previously shown code looks as follows:
...
-
-    For historical reasons, ``EntityManager#transactional($func)`` will return
-    ``true`` whenever the return value of ``$func`` is loosely false.
-    Some examples of this include ``array()``, ``"0"``, ``""``, ``0``, and
-    ``null``.
 
 The difference between ``Connection#transactional($func)`` and
 ``EntityManager#transactional($func)`` is that the latter
 abstraction flushes the ``EntityManager`` prior to transaction
-commit and rolls back the transaction when an
```
Ocramius: The fact that a rollback happens on an exception is still true

lib/Doctrine/ORM/EntityManagerInterface.php
```diff
@@ -81,9 +81,9 @@ public function beginTransaction();
      *
      * @param callable $func The function to execute transactionally.
      *
-     * @return mixed The non-empty value returned from the closure or true instead.
+     * @return mixed The value returned from the closure.
```
Ocramius: Should document `@throws \Throwable`

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -236,7 +232,7 @@ public function transactional($func)
             $this->flush();
             $this->conn->commit();
 
-            return $return ?: true;
+            return $return;
         } catch (Exception $e) {
```
Ocramius: This is to be replaced with `} catch (\Throwable $e) {`

tests/Doctrine/Tests/ORM/EntityManagerTest.php
```diff
@@ -191,12 +191,13 @@ public function testTransactionalAcceptsVariousCallables()
         self::assertSame('callback', $this->_em->transactional(array($this, 'transactionalCallback')));
     }
 
-    public function testTransactionalThrowsInvalidArgumentExceptionIfNonCallablePassed()
+    public function testTransactionalAcceptsReturnFalse()
     {
-        $this->expectException(\InvalidArgumentException::class);
-        $this->expectExceptionMessage('Expected argument of type "callable", got "object"');
+        $return = $this->_em->transactional(function ($em) {
```
Ocramius: Can assert directly: no need for assigning to `$return`

tests/Doctrine/Tests/ORM/EntityManagerTest.php
```diff
@@ -191,12 +191,13 @@ public function testTransactionalAcceptsVariousCallables()
...
+            return false;
+        });
 
-        $this->_em->transactional($this);
+        self::assertEquals(false, $return);
```
Ocramius: I suggest adding a scenario that covers also returning `null`

Ocramius (changes requested):

docs/en/reference/transactions-and-concurrency.rst
enumag: Of course it is. I removed it because the original text said it was a difference from Connection::transactional() - implying that Connection::transactional() does not call the rollback - which is of course not the case.

Ocramius: 👍

lib/Doctrine/ORM/EntityManager.php
enumag: Oh develop branch requires php 7.0. Ok then. 👍

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -222,12 +222,8 @@ public function getCache()
     /**
      * {@inheritDoc}
      */
-    public function transactional($func)
+    public function transactional(callable $func)
```
Majkl578: Now (for 7.0+) `call_user_func($func, $this)` may also be replaced by `$func($this)` I think.

enumag: @Majkl578 It seems this was possible even with older versions. https://3v4l.org/I7d3P

tests/Doctrine/Tests/ORM/EntityManagerTest.php
enumag: @Ocramius As this would add third test for the same thing I refactored them to a dataProvider.

lib/Doctrine/ORM/EntityManagerInterface.php
enumag: Done.

tests/Doctrine/Tests/ORM/EntityManagerTest.php
```diff
@@ -174,31 +174,33 @@ public function testAffectedByErrorIfClosedException($methodName)
         $this->_em->$methodName(new \stdClass());
     }
 
+    public function dataToBeReturnedByTransactional()
+    {
+        return array(
+            array(null),
+            array(false),
+            array('foo'),
+        );
```
enumag: I'm using the long-array-syntax here because it's used in other data providers. I'm guessing it will be changed to the short-syntax later in a batch.

enumag:
@Ocramius @Majkl578 All done, thank you for code review! 👍

lib/Doctrine/ORM/EntityManager.php
Majkl578: Yes, but not for some alternative/weird syntaxes. :) Don't exactly remember, but one example: https://3v4l.org/a9Ae4

tests/Doctrine/Tests/ORM/EntityManagerTest.php
Ocramius: Yeah, can be done via CS tools

Ocramius:
@enumag given the travis failures, can you check the [`EntityManagerDecorator`](https://github.com/doctrine/doctrine2/blob/bfa9a31ad77ce769f8e78ab70c2083e75f623c7a/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php) as well? Also check any classes implementing the `EntityManagerInterface` in the test assets/mocks.

enumag:
@Ocramius Done. I only found one other class in the tests.

Ocramius:
LGTM, thanks!

DHager:
It probably doesn't matter, but some additional unit-test stuff can be salvaged from PR #1392, which was an attempt to make some of the same changes.

PR #6149, opened 2016-11-25
VersionedOneToOneTest: added failing test with update

janedbal:
Updating versioned entity when using one-to-one association is broken as shown in failing test. It results in `Undefined index: second_entity_id` in IdentifierFlattener. Similar (but maybe not 100% same) error will occur when I use own type (e.g. Uuid object) for the id column of `SecondRelatedEntity`.

-----

*I was not able to find and fix the bug itself, so I hope this failing test will help you resolve this issue easily.*

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/VersionedOneToOneTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/VersionedOneToOneTest.php
@@ -60,4 +60,35 @@ class VersionedOneToOneTest extends OrmFunctionalTestCase
         $this->assertSame($secondRelatedEntity, $secondEntity);
         $this->assertSame($firstEntity->secondEntity, $secondEntity);
     }
+
+    public function testUpdateVersion()
+	{
+		$secondRelatedEntity = new SecondRelatedEntity();
+		$secondRelatedEntity->name = 'Bob';
+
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/VersionedOneToOneTest.php
```diff
@@ -50,8 +50,11 @@ public function testSetVersionOnCreate()
         $this->_em->persist($firstRelatedEntity);
         $this->_em->flush();
 
+        $firstRelatedEntity->name = 'Fred2';
+        $this->_em->flush();
+
         $firstEntity = $this->_em->getRepository('Doctrine\Tests\Models\VersionedOneToOne\FirstRelatedEntity')
-            ->findOneBy(array('name' => 'Fred'));
+            ->findOneBy(array('name' => 'Fred2'));
```
Ocramius: Can you please add a test method, instead of reusing an existing one for displaying the issue?

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/VersionedOneToOneTest.php
janedbal: Sure, added.

PR #6151, opened 2016-11-27
Embeddable attribute override support

andy-shea:
This allows for overriding of embeddable mapping definitions from the parent entity/embeddable. 

Nested embeddables are supported via dot notation along with the ability to drop column(s) from the embedded object if necessary.

The syntax is equivalent to attribute overrides on a subclass with the addition of passing a `null` value for a overridden column to drop it from the parent's field mappings.

Via annotations:
```php
/** @Embeddable */
class ValueObject {

    /** @Column(type="string") */
    public $value;

    /** @Column(type="integer") */
    public $count;

}

/** @Embeddable */
class NestedValueObject {

    /** @Embedded(class="ValueObject", columnPrefix=false) */
    public $nested;

}

/** @Entity */
class FooEntity {

    /** @Id @Column(type="integer") */
    private $id;

    /**
     * @Embedded(class="ValueObject", columnPrefix=false)
     * @AttributeOverride(name="value", column=@Column(type="string", name="single_value_override"))
     */
    public $valueObjectWithSingleOverride;

    /**
     * @Embedded(class="ValueObject", columnPrefix=false)
     * @AttributeOverrides({
     *     @AttributeOverride(name="value", column=@Column(type="string", name="multiple_value_override")),
     *     @AttributeOverride(name="count", column=null)
     * })
     */
    public $valueObjectWithMultipleOverrides;

    /**
     * @Embedded(class="NestedValueObject", columnPrefix=false)
     * @AttributeOverrides({
     *     @AttributeOverride(name="nested.value", column=@Column(type="string", name="nested_value_override")),
     *     @AttributeOverride(name="nested.count", column=null)
     * })
     */
    public $valueObjectWithNestedOverrides;
}
```

Via XML:
```xml
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                            https://raw.github.com/doctrine/doctrine2/master/doctrine-mapping.xsd">
    <embeddable name="ValueObject">
        <field name="value" type="string" />
        <field name="count" type="integer" />
    </embeddable>
</doctrine-mapping>

<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                            https://raw.github.com/doctrine/doctrine2/master/doctrine-mapping.xsd">
    <embeddable name="NestedValueObject">
        <embedded name="nested"
                class="ValueObject"
                use-column-prefix="false"/>
    </embeddable>
</doctrine-mapping>

<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                            https://raw.github.com/doctrine/doctrine2/master/doctrine-mapping.xsd">
    <entity name="FooEntity" table="entity">
        <id name="id" column="id"/>
        <embedded name="valueObjectWithSingleOverride"
                class="ValueObject"
                use-column-prefix="false">
            <attribute-override name="value">
                <field column="single_value_override" type="string"/>
            </attribute-override>
        </embedded>
        <embedded name="valueObjectWithMultipleOverrides"
                class="ValueObject"
                use-column-prefix="false">
            <attribute-overrides>
                <attribute-override name="value">
                    <field column="multiple_value_override" type="string"/>
                </attribute-override>
                <attribute-override name="count"/>
            </attribute-overrides>
        </embedded>
        <embedded name="valueObjectWithNestedOverrides"
                class="NestedValueObject"
                use-column-prefix="false">
            <attribute-overrides>
                <attribute-override name="nested.value">
                    <field column="nested_value_override" type="string"/>
                </attribute-override>
                <attribute-override name="nested.count"/>
            </attribute-overrides>
        </embedded>
    </entity>
</doctrine-mapping>
```

Via YML:
```yaml
ValueObject:
  type: embeddable
  fields:
    value:
      type: string
    count:
      type: integer

NestedValueObject:
  type: embeddable
  embedded:
    nested:
      class: ValueObject
      columnPrefix: false

FooEntity:
  type: entity
  id:
    id:
      type: integer
  embedded:
    valueObjectWithSingleOverride:
      class: ValueObject
      columnPrefix: false
      attributeOverrides:
        value:
          column: single_value_override
          type: string
    valueObjectWithMultipleOverrides:
      class: ValueObject
      columnPrefix: false
      attributeOverrides:
        value:
          column: multiple_value_override
          type: string
        count:
          column: null
    valueObjectWithNestedOverrides:
      class: NestedValueObject
      columnPrefix: false
      attributeOverrides:
        nested.value:
          column: nested_value_override
          type: string
        nested.count:
          column: null
```

A further improvement could be to merge the column override over the top of the original column instead of replacing it.  Doing so would remove the need to redefine original column attributes even if they have not been changed.  This change is outside the scope of this pull request.

References #6047

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -185,6 +185,12 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
                     $this->addNestedEmbeddedClasses($embeddableMetadata, $class, $property);
                 }
 
+                if ($embeddableClass['attributeOverrides']) {
+                    foreach ($embeddableClass['attributeOverrides'] as $name => $attributeOverride) {
+                        $embeddableMetadata->setAttributeOverride($name, $attributeOverride);
+                    }
+                }
+
                 $identifier = $embeddableMetadata->getIdentifier();
 
                 if (! empty($identifier)) {
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -187,7 +187,11 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
 
                 if ($embeddableClass['attributeOverrides']) {
                     foreach ($embeddableClass['attributeOverrides'] as $name => $attributeOverride) {
-                        $embeddableMetadata->setAttributeOverride($name, $attributeOverride);
+                        if ($attributeOverride) {
+                            $embeddableMetadata->setAttributeOverride($name, $attributeOverride);
+                        } else {
+                            $embeddableMetadata->removeFieldMapping($name);
+                        }
                     }
                 }
 
--- a/doctrine-mapping.xsd
+++ b/doctrine-mapping.xsd
@@ -301,6 +301,11 @@
   </xs:complexType>
 
   <xs:complexType name="embedded">
+    <xs:sequence>
+      <xs:element name="attribute-overrides" type="orm:attribute-overrides" minOccurs="0" maxOccurs="unbounded" />
+      <xs:element name="attribute-override" type="orm:attribute-override" minOccurs="0" maxOccurs="1" />
+      <xs:any minOccurs="0" maxOccurs="unbounded" namespace="##other"/>
+    </xs:sequence>
     <xs:attribute name="name" type="xs:string" use="required" />
     <xs:attribute name="class" type="xs:string" use="required" />
     <xs:attribute name="column-prefix" type="xs:string" use="optional" />
@@ -582,7 +587,7 @@
 
   <xs:complexType name="attribute-override">
     <xs:sequence>
-      <xs:element name="field" type="orm:attribute-override-field" minOccurs="1" />
+      <xs:element name="field" type="orm:attribute-override-field" minOccurs="0" />
       <xs:any minOccurs="0" maxOccurs="unbounded" namespace="##other"/>
     </xs:sequence>
     <xs:attribute name="name" type="xs:NMTOKEN" use="required" />
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -2203,6 +2203,18 @@ class ClassMetadataInfo implements ClassMetadata
             $overrideMapping['id'] = $mapping['id'];
         }
 
+        if (isset($mapping['originalClass'])) {
+            $overrideMapping['originalClass'] = $mapping['originalClass'];
+        }
+
+        if (isset($mapping['declaredField'])) {
+            $overrideMapping['declaredField'] = $mapping['declaredField'];
+        }
+
+        if (isset($mapping['originalField'])) {
+            $overrideMapping['originalField'] = $mapping['originalField'];
+        }
+
         if ( ! isset($overrideMapping['type'])) {
             $overrideMapping['type'] = $mapping['type'];
         }
@@ -3286,6 +3298,7 @@ class ClassMetadataInfo implements ClassMetadata
             'columnPrefix' => $mapping['columnPrefix'],
             'declaredField' => isset($mapping['declaredField']) ? $mapping['declaredField'] : null,
             'originalField' => isset($mapping['originalField']) ? $mapping['originalField'] : null,
+            'attributeOverrides' => isset($mapping['attributeOverrides']) ? $mapping['attributeOverrides'] : null,
         );
     }
 
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -2227,13 +2227,32 @@ class ClassMetadataInfo implements ClassMetadata
             throw MappingException::invalidOverrideFieldType($this->name, $fieldName);
         }
 
-        unset($this->fieldMappings[$fieldName]);
-        unset($this->fieldNames[$mapping['columnName']]);
-        unset($this->columnNames[$mapping['fieldName']]);
-
+        $this->removeFieldMapping($fieldName);
         $this->_validateAndCompleteFieldMapping($overrideMapping);
 
         $this->fieldMappings[$fieldName] = $overrideMapping;
+    }
+
+    /**
+     * Remove a mapped field.
+     *
+     * @param string $fieldName
+     *
+     * @return void
+     *
+     * @throws MappingException
+     */
+    public function removeFieldMapping($fieldName)
+    {
+        if ( ! isset($this->fieldMappings[$fieldName])) {
+            throw MappingException::mappingNotFound($this->name, $fieldName);
+        }
+
+        $mapping = $this->fieldMappings[$fieldName];
+
+        unset($this->fieldMappings[$mapping['fieldName']]);
+        unset($this->fieldNames[$mapping['columnName']]);
+        unset($this->columnNames[$mapping['fieldName']]);
     }
 
     /**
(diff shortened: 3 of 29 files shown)


Arkemlar:
It would be awesome to have this functionality. Can you tell me how to applly these changes using composer? (to not wait for merge)

andy-shea:
@Arkemlar yes, instructions are on the composer website: https://getcomposer.org/doc/05-repositories.md#vcs

Arkemlar:
@andy-shea Thank you! Installation successful. Working with Embeddables much more easier now.
Btw is it really needs php ^5.6 version? My dev system is ok with it but production runs on 5.6 so it needs upgrade.

Arkemlar:
I found bug.
Setting up overridings in one entity automaticly reflects on other entities:
I have embeddable with two fields:
```
<MyEmbeddable>
  type: embeddable
  fields:
    valueDefault:
      column: default
      type: string
      length: 255
      nullable: false   # This should be NOT NULL by default
    valueTranslated:
      column: translated
      type: string
      length: 255
      nullable: true
```
When I set up overridings in Product mapping, mapping in other tables changes:
```
<MyProductEntity>
  embedded:
    name:
      class: <MyEmbeddable>
      attributeOverrides:
        valueDefault:
          column: default
          nullable: true
    manufacturer:
      class: <MyEmbeddable>
      attributeOverrides:
        valueDefault:
          column: default
          nullable: true
```
When overriding is on:
```
$ bin/console doctrine:schema:update --dump-sql
ALTER TABLE product CHANGE name_default name_default VARCHAR(255) DEFAULT NULL, CHANGE name_translated name_translated VARCHAR(255) DEFAULT NULL, CHANGE manufacturer_default manufacturer_default VARCHAR(255) DEFAULT NULL, CHANGE manufacturer_translated manufacturer_translated VARCHAR(255) DEFAULT NULL;
(Down below is unexpected)
ALTER TABLE product_category CHANGE name_default name_default VARCHAR(255) DEFAULT NULL;
ALTER TABLE product_mark CHANGE name_default name_default VARCHAR(255) DEFAULT NULL, CHANGE description_default description_default VARCHAR(255) DEFAULT NULL;
```
When overriding is off (commented whole 'embedded' section in Product mapping):
```
$ bin/console doctrine:schema:update --dump-sql
ALTER TABLE product CHANGE name_default name_default VARCHAR(255) NOT NULL, CHANGE name_translated name_translated VARCHAR(255) DEFAULT NULL, CHANGE manufacturer_default manufacturer_default VARCHAR(255) NOT NULL, CHANGE manufacturer_translated manufacturer_translated VARCHAR(255) DEFAULT NULL;
```

danitome24:
Hi, 
that will be useful for me. When will this be accepted?

abdelior:
I think this feature is mandatory when working with value objects. Until the merge is done, is there any alternative to map orm value objects with multiple fields columns names?

Ocramius:
Looking at this, and I don't see the feature as necessary.

Overrides on embeddables are indeed useful, but for the few edge cases where they apply, a metadata listener would suffice.

That's my take on it at least, so I won't merge and will pass the token to the next maintainer looking at this.

ovidals:
@Ocramius 

Then what's the way to map the orm on a table that has "phone1" and "phone2", and you want the same value object "phone" to be used?

Ocramius:
@ovidals you just change the columns to be `phone1_phone` and `phone2_phone`. Then you can use your `Phone` embeddable for both the `$phone1` and the `$phone2` properties of your object.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -185,6 +185,16 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
                     $this->addNestedEmbeddedClasses($embeddableMetadata, $class, $property);
                 }
 
+                if ($embeddableClass['attributeOverrides']) {
```
phansys: Which types can store `$embeddableClass['attributeOverrides']`? If it can be only `array`, the `if` isn't required.

phansys (changes requested):
Status: Pending response.

danitome24:
> @ovidals you just change the columns to be phone1_phone and phone2_phone. Then you can use your Phone embeddable for both the $phone1 and the $phone2 properties of your object.

@Ocramius Can you show us an example, please? Do you mean to change de db column names? it's a legacy code so that's not possible.

Reference: http://stackoverflow.com/questions/41246632/doctrine-orm-value-object-for-mutiple-database-fields

Ocramius:
> Do you mean to change de db column names? it's a legacy code so that's not possible.

Don't expect doctrine to always 1:1 map on legacy stuff. A bit of flexibility here is required, like a r/w view for the legacy app, and moving the table to a newer structure. Expecting Doctrine to fix your legacy problems is out of question.

Arkemlar:
What about my bug report I provided above? It needs to be fixed I suppose...

Arkemlar:
@Ocramius my use case for this feauture is overriding some column settings:
- 'Product' entity has embeddable 'Name' with $name and $nameTranslated fields
- 'ProductCategory' has embeddable 'Name' too
- for 'Product' names I want allow nullable values for both variables (nullable=true) due to business needs
- for 'ProductCategory' I want the $name to be set as nullable=false and $nameTranslated to be set as nullable=true
___
Generally: 
- I use this embeddable in other entities too and may need to change varchar length or nullable flag for some entities so it will differ from default implementation
- I want to use one embeddable class everywhere (to avoid a lot of descendants with one/two rewritten property)

I don't see the way to make this without this feature.

lcobucci:
This is a no go for me too...

Defining the attribute overrides in the class level should be sufficient IMO (and according to [@schmittjoh](https://gist.github.com/lcobucci/9e68501cfad6931119d7#gistcomment-1365387) embeddables were designed to be overridden in that way).

I've just debugged the `ClassMetadataFactory` and the reason that prevents it is because we're calling `ClassMetadataInfo#setAttributeOverride()` [while loading the class metadata](https://github.com/doctrine/doctrine2/blob/ce4abdea55111ec265cb379704aa9f4a2e77b627/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L151) which happens before of [inlining the class' embeddables](https://github.com/doctrine/doctrine2/blob/ce4abdea55111ec265cb379704aa9f4a2e77b627/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L194).

With that said, I think that we don't need a new way to override attributes but rather fix the one we have now, the only thing is that it can be a bit tricky to fix it properly (without introducing a lot of duplication but is doable IMO).

Ocramius:
Closing as per @lcobucci's feedback. Sorry @andy-shea, marking as `won't fix` :-\

danitome24:
I couldn't make work embeddable overriding and I try to find another way. The way I found is using [doctrine events](http://doctrine-orm.readthedocs.io/en/latest/reference/events.html#entity-listeners-class). I build an EntityListener which is subscribed to postLoad event and where I build my ValueObject and inject it to my Entity.

What do you think about it?

lcobucci:
@danitome24 as said on https://github.com/doctrine/doctrine2/pull/6151#issuecomment-271432401 the proper solution would be fixing the order of the methods' calls (to make sure that the metadata state is right).

creocoder:
@lcobucci Are there any plans to fix this?

This is a fairly basic and popular feature in Hibernate: https://docs.jboss.org/hibernate/orm/6.4/userguide/html_single/Hibernate_User_Guide.html#embeddable-override

lcobucci:
@creocoder I'm not an active maintainer atm but I believe @beberlei is the best person to answer you 👍

PR #6152, opened 2016-11-27
Use the correct types for generated ids

lcobucci:
This merges #5935 and #6020 in order to solve #5684.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -19,36 +19,33 @@
 
 namespace Doctrine\ORM;
 
-use Doctrine\Common\Persistence\Mapping\RuntimeReflectionService;
-use Doctrine\DBAL\LockMode;
-use Doctrine\ORM\Internal\HydrationCompleteHandler;
-use Doctrine\ORM\Mapping\Reflection\ReflectionPropertiesGetter;
-use Exception;
-use InvalidArgumentException;
-use UnexpectedValueException;
-
 use Doctrine\Common\Collections\ArrayCollection;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -890,7 +887,14 @@ private function persistNew($class, $entity)
             $idValue = $idGen->generate($this->em, $entity);
 
             if ( ! $idGen instanceof \Doctrine\ORM\Id\AssignedGenerator) {
-                $idValue = array($class->identifier[0] => $idValue);
+                $idField = $class->identifier[0];
```
Ocramius: I just digged into this conversion in depth, and it seems wrong. The `AssignedGenerator` already uses a PHP value internally, and never produces an SQL value. This code can therefore be dropped, and test code needs to be re-validated

Ocramius (changes requested):

lib/Doctrine/ORM/UnitOfWork.php
Ocramius: Nvm, applies to non-assigned generators -.- Misread the `!` in front of it.

lcobucci: But then it's a bit confusing... should this conversion be generators' responsibility?

Ocramius: Could be, but can't refactor it now, since generators have very little access to external scope.

Ocramius:
@lcobucci merged and closed all related issues, thanks a lot!

Note: I added a few additional commits to just extract the common code.

PR #6155, opened 2016-12-01
[WIP][Bugfix] Eager loaded associations and orderBy mappings

guiwoda:
This PR aims to solve #4256 

I'm starting with the regression test. I've been digging and I believe this has mostly to do with the fact that the orderBy clause is read when `UnitOfWork` calls the `EntityPersister::loadOneToManyCollection`, while eager loading is done at `EntityPersister` level through the join analysis in `getSelectColumnsSQL()`.

I could hack into the `getSelectColumnsSQL()` method and leave some sort of `orderedJoinedQuery` mark somewhere, but it _will be hacky_.

I'm opening this here to ask for opinions on how this could be tackled. I'm available to get it done, just point my fingers in the right direction!

Cheers

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3448Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3448Test.php
@@ -1,0 +1,105 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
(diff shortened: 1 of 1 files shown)


Ocramius:
> I could hack into the `getSelectColumnsSQL()` method and leave some sort of `orderedJoinedQuery` mark somewhere, but it _will be hacky_.

Give it a spin first, I'd say. The fact that you think that it will be hacky also means that you may come up with a better idea while going through it.

isaackearl:
Is a fix for this still planned?

PR #6156, opened 2016-12-02
Allow doctrine/common 2.7

Slamdunk:
To ease the process I just edited the current constraint schema, but I would rather use caret version ^2.4 if you agree.


Ocramius:
@Slamdunk 👍

PR #6168, opened 2016-12-09
Fixed #6167 - Forced SELECT NEXTVAL to be run on master

### LINKED ISSUE #6167: Fails to insert to a PostgreSQL master-slave config due to SequenceGenerator(NEXTVAL) issues
Background: I have a master-slave connection and I'm using auto-generated id for my entity;

Doctrine ORM fails to insert data because it tries to get the sequence NEXTVAL value from a slave.

```php
    {
        if ($this->_maxValue === null || $this->_nextValue == $this->_maxValue) {
            // Allocate new values
            $conn = $em->getConnection();

            $this->_nextValue = (int)$conn->fetchColumn($sql); //this line fails
            $this->_maxValue  = $this->_nextValue + $this->_allocationSize;
        }

        return $this->_nextValue++;
    }
```

PostgreSQL allows to fetch nextval only from a master server, and using fetchColumn runs the query using a slave connection.

I could start a transaction before, but I assume this case should be handled by ORM.

In case someone tries to reproduce this issue, just start a PostgreSQL cluster (docker-compose and https://hub.docker.com/r/bitnami/postgresql/ might be helpful) and try to persist any entity with an auto-generated ID.

mkurzeja:
A possible solution for #6167 - Doctrine ORM failing to insert to a master-slave PostgreSQL setup because of running NEXTVAL on a slave.


Ocramius:
Requires a test case

mkurzeja:
And it seems it broke support for other databases.

mkurzeja:
Ok, so this does not break any DB support, it just does not fit the tests written. I will provide a fix and a new test for this scenario.

mkurzeja:
@Ocramius I have fixed the existing test, but I cannot think about a good test that could be run in CI. Travis does not support master-slave setup, and the whole logic is inside MasterSlaveConnection that I did not touch at all. I would have to write a custom test that wouldn't be run on Travis, or I would have to mock a lot of things just to check that MasterSlaveConnection::query() is run on the master server. Please let me know if I'm missing something and there is a better way to test this case.

Ocramius:
@mkurzeja what you could do is writing a test that asserts that `query` is used, whereas `fetchColumn` isn't, is sufficient.

Can be done via a mock/spy, and indeed needs some documentation around it.

tests/Doctrine/Tests/ORM/Id/SequenceGeneratorTest.php
```diff
@@ -23,17 +24,20 @@ protected function setUp()
...
     public function testGeneration()
     {
+        $this->_em->getConnection()->setFetchOneException(
+            new \Exception('Fetch* method used. Query method should be used instead, as NEXTVAL should be run on a master server in master-slave setup.')
+        );
+
         for ($i=0; $i < 42; ++$i) {
             if ($i % 10 == 0) {
-                $this->_em->getConnection()->setFetchOneResult((int)($i / 10) * 10);
+                $nextId = array(array((int)($i / 10) * 10));
```
lcobucci: @mkurzeja please use short-array syntax here ;)

tests/Doctrine/Tests/Mocks/ConnectionMock.php
```diff
@@ -86,12 +97,24 @@ public function lastInsertId($seqName = null)
      */
     public function fetchColumn($statement, array $params = array(), $colnum = 0, array $types = array())
     {
+        if ($this->_fetchOneException != null) {
```
phansys: Please, use Yoda condition and strict comparison.

tests/Doctrine/Tests/Mocks/ConnectionMock.php
```diff
@@ -113,6 +136,16 @@ public function setFetchOneResult($fetchOneResult)
     }
 
     /**
+     * @param \Exception $exception
```
phansys: Please, update this docblock in order to reflect that this param also accepts `null`.

tests/Doctrine/Tests/Mocks/ConnectionMock.php
```diff
@@ -133,6 +166,14 @@ public function setLastInsertId($id)
     }
 
     /**
+     * @param Statement $result
+     */
+    public function setQueryResult($result)
```
phansys: Can `$result` argument be declared as `Statement`?

tests/Doctrine/Tests/ORM/Id/SequenceGeneratorTest.php
```diff
@@ -23,17 +24,20 @@ protected function setUp()
 
     public function testGeneration()
     {
+        $this->_em->getConnection()->setFetchOneException(
+            new \Exception('Fetch* method used. Query method should be used instead, as NEXTVAL should be run on a master server in master-slave setup.')
```
phansys: SPL `\LogicException` or `\RuntimeException` could be less generic.

phansys (changes requested):
Status: Needs work.

mkurzeja:
Thanks @phansys and @lcobucci for review and @Ocramius for help. Hope this is acceptable now ;)

tests/Doctrine/Tests/Mocks/ConnectionMock.php
```diff
@@ -86,12 +97,24 @@ public function lastInsertId($seqName = null)
      */
     public function fetchColumn($statement, array $params = [], $colnum = 0, array $types = [])
     {
+        if (null !== $this->_fetchOneException) {
```
Ocramius: This makes sense, but I'll probably have to change it to use the mocking framework to do this (on merge)

Ocramius:
This has been manually rebased and merged into `master` via doctrine/doctrine2@07a9b10f3678664a15049be0fc3ff47374435152

Thanks @mkurzeja \o/

wlzch:
Hi,

Any reason why is this not merge into 2.5 branch? Also is there a workaround if I want to use it in v2.5? Thanks.

gigi:
Hi!
We need this fix in 2.5 too. Any solutions? Thanx

lcobucci:
@wlzch @gigi I've just tried to backport this and unfortunately it cannot be done easily, since it has dependencies on things that only exists on `v2.6` (and backporting these things might lead us down the rabbit hole).

As stated on every [single `v2.5.x` milestone in repo since `2.5.8`](https://github.com/doctrine/doctrine2/milestones?state=closed), we no longer actively support 2.5.x since we must release `v2.6` ASAP. So, sorry but this won't be part of any `v2.5.x` release.

gigi:
@lcobucci 
Thanks for answer! 
But is it not enough to replace `$this->_nextValue = (int)$conn->fetchColumn($sql);` with `$this->_nextValue = (int) $conn->query($sql)->fetchColumn();` in `\Doctrine\ORM\Id\SequenceGenerator::generate`?

lcobucci:
@gigi not really, we need to get the tests as well to make sure things work and that's the tricky part

lcobucci:
@gigi if you have the time and need this really badly (and can't update your stack to use 2.6) you can try to cherry-pick the commits related to the merge commit that @Ocramius mentioned with all the dependencies (ensuring that things are compatible with PHP 5.4) please do it and send us a PR.

gigi:
@lcobucci https://github.com/doctrine/doctrine2/pull/6852

PR #6171, opened 2016-12-13
Include example of a cascaded "persist" operation

ThomasLandauer:
Following up on https://github.com/doctrine/doctrine2/issues/2943 I started to clarify how it's supposed to be done.

Please check if this would be necessary (at line 511):
`$myFirstComment->setUser($user);`
...and add it (in case).

### DIFF
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -500,7 +500,8 @@ and the "remove" operation.
         //...
     }
 
-Now you can persist the `User` entity like that:
+Since ``cascade: persist`` is configured for the ``User#commentsAuthored``
+association, you can now create a user and persist his comment as follows:
 
 .. code-block:: php
 
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -453,7 +453,7 @@ following code:
     $user->addComment($myFirstComment);
     
     $em->persist($user);
-    $em->persist($myFirstComment); // mandatory if `persist` isn't set
+    $em->persist($myFirstComment); // mandatory if `cascade: persist` isn't set
     $em->flush();
 
 Even if you *persist* a new User that contains our new Comment this
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -453,11 +453,11 @@ following code:
     $user->addComment($myFirstComment);
     
     $em->persist($user);
-    $em->persist($myFirstComment);
+    $em->persist($myFirstComment); // mandatory if `persist` isn't set
     $em->flush();
 
 Even if you *persist* a new User that contains our new Comment this
-code would fail if you removed the call to
+code requires the explicit call to
 ``EntityManager#persist($myFirstComment)``. Doctrine 2 does not
 cascade the persist operation to all nested entities that are new
 as well.
@@ -499,6 +499,18 @@ and the "remove" operation.
         private $commentsAuthored;
         //...
     }
+
+Now you can persist the `User` entity like that:
+
+.. code-block:: php
+
+    <?php
+    $user = new User();
+    $myFirstComment = new Comment();
+    $user->addComment($myFirstComment);
+    
+    $em->persist($user);
+    $em->flush();
 
 Even though automatic cascading is convenient it should be used
 with care. Do not blindly apply cascade=all to all associations as
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -405,25 +405,11 @@ There are two approaches to handle this problem in your code:
 Transitive persistence / Cascade Operations
 -------------------------------------------
 
-Persisting, removing, detaching, refreshing and merging individual entities can
-become pretty cumbersome, especially when a highly interweaved object graph
-is involved. Therefore Doctrine 2 provides a
-mechanism for transitive persistence through cascading of these
-operations. Each association to another entity or a collection of
-entities can be configured to automatically cascade certain
-operations. By default, no operations are cascaded.
-
-The following cascade options exist:
-
-
--  persist : Cascades persist operations to the associated
-   entities.
--  remove : Cascades remove operations to the associated entities.
--  merge : Cascades merge operations to the associated entities.
--  detach : Cascades detach operations to the associated entities.
--  refresh : Cascades refresh operations to the associated entities.
--  all : Cascades persist, remove, merge, refresh and detach operations to
-   associated entities.
+When working with many associated entities, you sometimes don't want to "expose" all entities to your PHP application.
+Therefore Doctrine 2 provides a mechanism for transitive persistence through cascading of certain operations.
+Each association to another entity or a collection of
+entities can be configured to automatically cascade the following operations to the associated entities:
+``persist``, ``remove``, ``merge``, ``detach``, ``refresh`` or ``all``.
 
 .. note::
 
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -453,7 +453,7 @@ following code:
     $user->addComment($myFirstComment);
     
     $em->persist($user);
-    $em->persist($myFirstComment); // mandatory if `cascade: persist` isn't set
+    $em->persist($myFirstComment); // required, if `cascade: persist` isn't set
     $em->flush();
 
 Even if you *persist* a new User that contains our new Comment this
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -513,8 +513,14 @@ association, you can now create a user and persist his comment as follows:
     $em->persist($user);
     $em->flush();
 
-Even though automatic cascading is convenient it should be used
-with care. Do not blindly apply cascade=all to all associations as
+.. note::
+
+    This code does not associate the newly created comment with the user.
+    To achieve this, you *always* have to call ``$myFirstComment->setUser($user);`` –
+    no matter if ``cascade: persist`` is set or not.
+
+Even though automatic cascading is convenient, it should be used
+with care. Do not blindly apply ``cascade=all`` to all associations as
 it will unnecessarily degrade the performance of your application.
 For each cascade operation that gets activated Doctrine also
 applies that operation to the association, be it single or
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -442,7 +442,7 @@ The following cascade options exist:
 
 The following example is an extension to the User-Comment example
 of this chapter. Suppose in our application a user is created
-whenever he writes his first comment. In this case we would use the
+whenever they write their first comment. In this case we would use the
 following code:
 
 .. code-block:: php
@@ -462,7 +462,7 @@ code requires an explicit call to
 cascade the persist operation to all nested entities that are new
 as well.
 
-More complicated is the deletion of all of a user's comments when he is
+More complicated is the deletion of all user's comments when the user is
 removed from the system:
 
 .. code-block:: php
@@ -501,7 +501,7 @@ and the "remove" operation.
     }
 
 Since ``cascade: persist`` is configured for the ``User#commentsAuthored``
-association, you can now create a user and persist his comment as follows:
+association, you can now create a user and persist their comments as follows:
 
 .. code-block:: php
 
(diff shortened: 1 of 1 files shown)


docs/en/reference/working-with-associations.rst
```diff
@@ -453,11 +453,11 @@ following code:
     $user->addComment($myFirstComment);
     
     $em->persist($user);
-    $em->persist($myFirstComment);
+    $em->persist($myFirstComment); // mandatory if `persist` isn't set
```
Ocramius: "persist isn't set"? What did you mean with this comment?

docs/en/reference/working-with-associations.rst
```diff
@@ -453,11 +453,11 @@ following code:
...
     $em->flush();
 
 Even if you *persist* a new User that contains our new Comment this
-code would fail if you removed the call to
+code requires the explicit call to
```
Ocramius: `requires an explicit call`

docs/en/reference/working-with-associations.rst
```diff
@@ -500,6 +500,18 @@ and the "remove" operation.
         //...
     }
 
+Now you can persist the `User` entity like that:
```
Ocramius: Instead of `Now`, describe the context again. Something like `Since cascade persist is configured for the User#commentsAuthored association, you can now create a user and persist its comments with it as following:`

Ocramius (changes requested):

ThomasLandauer:
@Ocramius: I edited the PR but I'm not sure if I commited it the right way?

Actually, the entire PR is also a question (to which I don't know the answer): If I do it like in lines 508-514, will the persisted Comment "automatically" get its `UserID` field set to the User's id?? Or is `$myFirstComment->setUser($user);` still required?
Is it this what you mean with `You are responsible for keeping your object graph consistent.` there? https://github.com/doctrine/doctrine2/issues/2943

=> Please clarify! :-)

docs/en/reference/working-with-associations.rst
```diff
@@ -500,6 +500,19 @@ and the "remove" operation.
         //...
     }
 
+Since ``cascade: persist`` is configured for the ``User#commentsAuthored``
+association, you can now create a user and persist his comment as follows:
```
Ocramius: s/`his`/`its`

Ocramius: s/`comment`/`comments`

docs/en/reference/working-with-associations.rst
```diff
@@ -453,11 +453,11 @@ following code:
     $user->addComment($myFirstComment);
     
     $em->persist($user);
-    $em->persist($myFirstComment);
+    $em->persist($myFirstComment); // mandatory if `cascade: persist` isn't set
```
Ocramius: s/`mandatory if`/`mandatory, if`

Ocramius (changes requested):
`$myFirstComment->setUser($user);` is required anyway, because the ORM will not change any state of your objects (except for identifiers). You are responsible for assigning the values on both the owning and the inverse side of the association. The ORM will do that only when loading fresh instances.

docs/en/reference/working-with-associations.rst
ThomasLandauer: I'm not 100% sure what you mean. However:
* You can't call a person "it"! I followed the convention on this page, referring to a user as "he".
* The example is about just one comment, not many.

If you mean something else, let me know :-)

ThomasLandauer:
To me, the question if `$myFirstComment->setUser($user);` is required or not is the most important issue. So I put it into a new "note" block.

Open question: What's the *practical* advantage of `cascade:persist` at all?
Without it, I need 5 lines:
````php
$user = new User();
$myFirstComment = new Comment();
$myFirstComment->setUser($user);
$em->persist($user);
$em->persist($myFirstComment);
````
And with it, I need 5 lines too:
````php
$user = new User();
$myFirstComment = new Comment();
$user->addComment($myFirstComment);
$myFirstComment->setUser($user);
$em->persist($user);
````
So to me it looks like `cascade:persist` just "trades" the `$em->persist($myFirstComment);` line for the `$user->addComment($myFirstComment);` line - what is gained here?

File: docs/en/reference/working-with-associations.rst [hunk 6]
```diff
@@ -500,8 +500,27 @@ and the "remove" operation.
...
+    $user->addComment($myFirstComment);
+    
+    $em->persist($user);
+    $em->flush();
+
+.. note::
+
+    This code does not associate the newly created comment with the user.
+    To achieve this, you *always* have to call ``$myFirstComment->setUser($user);`` –
+    no matter if ``cascade: persist`` is set or not.
```
Ocramius: We shouldn't show a broken example: let's correct the example and then explain why we do it, rather than explaining why the example is broken

Ocramius (changes requested):

docs/en/reference/working-with-associations.rst
Ocramius: > You can't call a person "it"! I followed the convention on this page, referring to a user as "he".

Although I agree with you, there's an entire social movement (and a quite aggressive one) about making stuff gender neutral, and I just want to avoid them bother the project due to a misplaced word here and there.

 > The example is about just one comment, not many.

Yes, but a user always has multiple comments, regardless if it's 0, 1 or 10000.

docs/en/reference/working-with-associations.rst
ThomasLandauer: I changed several (hopefully all) occurances of "he" to gender neutral.

Broken example: Well, I wouldn't call the exmple broken. That's just how `cascade:persist` apparently works ;-)

ThomasLandauer: If you help me understand the very purpose of `cascade:persist` (see above question with the 2x5 lines), maybe I can come up with a better code example + explanation.

Ocramius: @ThomasLandauer Let me clarify:

Let's say that you have an association tree `A -> B -> C`, where `A` "owns" `B`, and `B` owns `C`. You don't really want to expose `B` and `C` to the outside world, so they need to stay hidden in the implementation.

The various parts of the app should therefore only touch `A`, so you'd have `$em->persist($a)` and `$em->remove($a)`, but never touch instances of `B`, `C`.

This is the perfect use-case scenario for cascade operations, since the ORM will therefore interact with child elements only via mappings, and not via explicit API usage in your application layer (which would cause references of `B` and `C` to leak around).

There are other use-cases, but this is the main one.

As for "why does this not just work out of the box?", the reason is that otherwise accidents may happen when instances leak into the graph, and they get either persisted or removed automagically. You want to retain control over this, so an explicit cascade operation has to happen here.

Also, cascades are expensive (performance-wise) and should be used sparingly.

Makes sense?

ThomasLandauer: > Makes sense?

Yeah, **thanks**, it's starting to make sense :-)

IMO what you're saying should be explained in the docs! I think I'll somewhat rewrite the entire 8.6 chapter - right now the very first sentence is already leading to the wrong direction ("can become pretty cumbersome"), namely promoting the cascading as a *convenience* feature.

I would explain the question "who should use it" at first, deferring all other topics to later. Reason: For a feature which is only useful for a small minority, it doesn't make sense to start with details.

Is that OK?

However, one more question I have:  
What exactly do you mean by "expose to the outside world"? I mean, the Comment entity has to be instantiated anyway: `$myFirstComment = new Comment();` - so it is always "exposed" (in the sense of "visible", or "modifyable"), and never "hidden in the implementation".  
Or do you mean that it's not directly "persistable" by the PHP app - i.e. the only way to get a Comment into the database is "through" the User entity?

Ocramius: > IMO what you're saying should be explained in the docs! 

Add it! :D

 > cascading as a *convenience feature.

Cascading **is** a convenience feature. It is not to be used as a default, but it is very commonly used when dealing with DDD aggregate roots.

 >  I think I'll somewhat rewrite the entire 8.6 chapter

Be my guest! Any improvements are hugely welcomed! 

 > I would explain the question "who should use it" at first, deferring all other topics to later. Reason: For a feature which is only useful for a small minority, it doesn't make sense to start with details.
 > 
 > Is that OK?

Absolutely!

 > However, one more question I have:
 > What exactly do you mean by "expose to the outside world"?

Take a `BankAccount` object (typical sad example): you can most probably deposit and withdraw from it, but you can't really modify its internals from the outside, as that would be an obvious violation of encapsulation. Inside it, you may have `Transaction` instances that aren't really visible to the outside world: all you want is for them to be saved/loaded correctly, but you don't want to give access to others.

Ocramius: Please note that these are DDD concepts, not really ORM concepts. This is just a scenario where this fits well.

ThomasLandauer:
OK, for now I started with the first paragraph.

However, I still don't understand the "exposure" aspect well enough to explain it ;-)  
The "internals" of a bank account are probably a `Deposit` and a `Withdrawal` entity. How does `cascade` help you in **hiding** those? Public methods `Account::addDeposit()`, `Deposit::setAmount()` and `Deposit::setAccount()` need to exist anyway!

Or is it all about the fact that Deposit entities can't be persisted directly? But `Deposit::setAccount()` has to be called manually, so the `cascade` doesn't even save you from having orphaned Deposits in your database. And I still can't see any (security) advantage of `$account->addDeposit()` over `$em->persist($deposit)`

Ocramius:
@ThomasLandauer you'd have `Account#deposit($amount, $reason, $dateTime)`, but the actual `Deposit` instance is created inside the method, or a more hidden private method. Same goes for `Withdrawal`.

The actual `Deposit` and `Withdrawal` objects are never instantiated nor type-hinted outside an `Account`'s implementation.

```php
class Account
{
    private $id;
    private $deposits;
    private $withdrawals;

    public function __construct()
    {
        $this->id = AccountId::new();
        $this->deposits = new ArrayCollection();
        $this->withdrawals = new ArrayCollection();
    }

    public function deposit(AccountId $origin, Amount $amount, DateTimeInterface $time, string $reason) : void
    {
        $this->deposits->add(Deposit::from($origin, $amount, $time, $reason));
    }

    public function withdraw(Amount $amount, DateTimeInterface $time) : void
    {
        if (! $this->availableFunds()->fullfills($amount)) {
            throw InsufficientFunds::forWithdrawal($this->id, $amount);
        }

        $this->deposits->add(Withdrawal::from($origin, $amount, $time, $reason));
    }

    private function availableFunds() : Amount
    {
        // ...
    }
}
```

ThomasLandauer:
Sorry for the delay!

Yeah, it's finally starting to make sense :-)  
But I'm losing track as to what the current status of the page is now. So could you please merge this PR for now, and I'll start a new one with further changes?

docs/en/reference/working-with-associations.rst
```diff
@@ -500,8 +486,27 @@ and the "remove" operation.
...
-with care. Do not blindly apply cascade=all to all associations as
+Since ``cascade: persist`` is configured for the ``User#commentsAuthored``
+association, you can now create a user and persist their comments as follows:
+
+.. code-block:: php
+
+    <?php
+    $user = new User();
+    $myFirstComment = new Comment();
+    $user->addComment($myFirstComment);
```
Ocramius: Before merging, `$comment->user` must be set in this example.

Ocramius (changes requested):

Ocramius:
@ThomasLandauer awesome, thanks!

ThomasLandauer:
@Ocramius after reading it all again, I think:  
Your point with the "exposure" aspect only makes sense if the Deposit entity is created **inside the Account entity**, right?  
But in the docs line 496-503 it's code from the controller, not from the entity. So this isn't suitable for illustrating what you mean, right?!  
So I'd like to change the file to convey these messages:  
1. `cascade:persist` is mainly for avoiding "exposure"
1. If you use it for this reason, it only makes sense to instantiate the Depsit entity inside the Account entity (and not in the controller).  
Plus:
1. It's not meant to save you some lines of code.
1. It's bad for performance.  

So:
- It's no mainstream feature.

OK?

Ocramius:
> `cascade:persist` is mainly for avoiding "exposure"

That's just the main use-case

 > If you use it for this reason, it only makes sense to instantiate the Depsit entity inside the Account entity (and not in the controller).

Aye, creating such an example is likely more complex

 > It's not meant to save you some lines of code.

Correct, it's mostly about keeping code safe/clean and decoupled

 > It's bad for performance.

Full ack.

ThomasLandauer:
@Ocramius See https://github.com/doctrine/doctrine2/pull/6219

PR #5570, opened 2015-12-22
Fix PrePersist EventListener when using merge instead of persist

moroine:
The merge operation trigger PRE_PERSIST & POST_PERSIST events when we merge a new entity.

The bug that I encounter, is that the PRE_PERSIST event is trigger with the managed copy empty ! I that case my event listener can't access original data.

In order to fix this, I firstly merge entity state into managed copy. I think this is not the best move as it will be copied a second time at line 1888.

I tried to persist after this is done, but it break some unit tests.

PR #6174, opened 2016-12-15
Merging a new entity with PrePersist event make changes in callback not be considered

moroine:
When merging a new entity, the must act like a basic persist. But it seems that changes made into the PrePersist event callback are not persisted.

To fix this I firstly refactor two features in the  `doMerge`:
   - Added `ensureVersionMatch`: that check the requested merge entity version is correct (in case of versionned one) 
   - Move the `isLoaded` test used before  ̀mergeEntityStateIntoManagedCopy` into `mergeEntityStateIntoManagedCopy`

Then I move `mergeEntityStateIntoManagedCopy` calls to ensure that it will be called always before `persistNew`.

This will also fixed one of my previously PR (#5570) that is schedule for the next patch release.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1816,6 +1816,7 @@ class UnitOfWork implements PropertyChangedListener
             if ( ! $id) {
                 $managedCopy = $this->newInstance($class);
 
+                $this->mergeEntityStateIntoManagedCopy($entity, $managedCopy);
                 $this->persistNew($class, $managedCopy);
             } else {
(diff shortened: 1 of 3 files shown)


moroine:
@Ocramius it seems tests fail due to Travis

Ocramius:
@moroine everything seems fine in the code changes, but the test is reusing some code that seems unrelated (entity resolver listeners).

I'm currently rewriting the test to get rid of that.

Ocramius:
Moved to #6177, applied fixes locally, rewrote the tests to reduce cross-test dependencies. 👍

moroine:
Amazing ! Thanks a lot ! 

The next patch version will make me abandon my one year fork !!!

PR #6178, opened 2016-12-18
Backport #6177 - fix #6174 #5570: merging new entities should also trigger prepersist lifecycle callbacks with the merged data

Ocramius:
Backport of #6177


Ocramius:
Travis looking good!

PR #6179, opened 2016-12-19
Minor fixes to the 'getting started' page

biggianteye:
- Use consistent visibility for some fields.
- Fixing path of yaml file.
- Minor grammar changes.

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -717,8 +717,8 @@ the bi-directional reference:
     {
         // ... (previous code)
 
-        private $reportedBugs = null;
-        private $assignedBugs = null;
+        protected $reportedBugs = null;
+        protected $assignedBugs = null;
 
         public function addReportedBug($bug)
         {
@@ -969,7 +969,7 @@ The last missing definition is that of the User entity:
 
     .. code-block:: yaml
 
-        # config/xml/User.dcm.yml
+        # config/yaml/User.dcm.yml
         User:
           type: entity
           table: users
@@ -1008,7 +1008,7 @@ Update your database running:
 Implementing more Requirements
 ------------------------------
 
-For starters we need a create user entities:
+For starters we need to create user entities:
 
 .. code-block:: php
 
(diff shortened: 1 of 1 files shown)


docs/en/tutorials/getting-started.rst
```diff
@@ -717,8 +717,8 @@ the bi-directional reference:
     {
         // ... (previous code)
 
-        private $reportedBugs = null;
-        private $assignedBugs = null;
+        protected $reportedBugs = null;
```
Ocramius: Should stay `private`

docs/en/tutorials/getting-started.rst
```diff
@@ -717,8 +717,8 @@ the bi-directional reference:
...
+        protected $assignedBugs = null;
```
Ocramius: Should stay `private`

Ocramius (changes requested):

docs/en/tutorials/getting-started.rst
biggianteye: Could you explain why? All other instance variables are marked as `protected`. Why should these two be `private`?

More importantly, they were declared as `protected` when first added to the `User` class. If there is a legitimate need for the change to the visibility, let me know and I'll incorporate it into the section where the visibility changes.

Ocramius: I'd rather have everything as `private` around the code, but if you think that `protected` is better for consistency (for now), I can merge the PR as-is.

Should generally strive for `private` in future.

biggianteye: I'm not trying to advocate one over the other (though if I were, I'd go for private). I'm just trying to remove any confusion caused by modifiers changing half way through introductory documentation without any explanation.

Ocramius: Good, merging as-is then :-)

docs/en/tutorials/getting-started.rst
biggianteye: 👍

docs/en/tutorials/getting-started.rst
biggianteye: 👍

PR #6180, opened 2016-12-19
Remove final restriction from Doctrine\ORM\Query to allow mocking

guiwoda:
The `Doctrine\ORM\Query` class is marked as final but does not have a full interface, so mocking it is basically next to impossible.

I propose dropping the final restriction, leaving it commented as it was done in the `EntityManager`. Documentation could also recommend **not to** extend it.

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -35,7 +35,7 @@ use Doctrine\Common\Collections\ArrayCollection;
  * @author  Konsta Vesterinen <kvesteri@cc.hut.fi>
  * @author  Roman Borschel <roman@code-factory.org>
  */
-final class Query extends AbstractQuery
+/* final */class Query extends AbstractQuery
 {
     /**
      * A query object is in CLEAN state when it has NO unparsed/unprocessed DQL parts.


Ocramius:
There's no need to extend the `Query` object. The semantics are absolutely terrifying and specific to the ORM: just don't do it.

guiwoda:
@Ocramius I'm not extending, I'm trying to mock it.

Ocramius:
@guiwoda a query object is code, and cannot really be mocked. It's like mocking `eval()`: doesn't really work.

Keep `Query` and `QueryBuilder` as part of your private API only. Also, there's no way to really unit test it: needs integration tests, as asserting on a DQL string or structure is not sufficient to assert its correctness.

guiwoda:
@Ocramius `Query` is final, but there is no interface that covers it. Other parts of Doctrine (right now `Paginator` is the one that bugs me) depend on a final, non-mockable class.

How do you propose I test a code that instantiates a `Paginator` without the ability to mock a `Query`?

Ocramius:
> How do you propose I test a code that instantiates a `Paginator` without the ability to mock a `Query`?

Code that instantiates a paginator is statically coupled to both the paginator (which doesn't have an interface, besides being an `Iterator`) and the query/query builder: it's in fact query logic that is not mockable. Consumers of the paginator should expect a `Countable` or an `Iterator` or `IteratorAggregate` instead.

guiwoda:
@Ocramius the code instantiates the `Paginator` but passes through a `Query` from parameters. That's why I would have mocked the `Query` and let the `Paginator` be built with it.

This is Laravel Doctrine bridge code. You can check the linked PR if you want to.

PR #6181, opened 2016-12-19
Added a QueryInterface to decouple code from the final Query class

guiwoda:
The final `Query` class is impossible to mock, so to decouple its dependants from it, I refactor->extracted an interface out.

https://github.com/doctrine/doctrine2/pull/6180

### DIFF
--- a/lib/Doctrine/ORM/EntityManagerInterface.php
+++ b/lib/Doctrine/ORM/EntityManagerInterface.php
@@ -104,7 +104,7 @@ interface EntityManagerInterface extends ObjectManager
      *
      * @param string $dql The DQL string.
      *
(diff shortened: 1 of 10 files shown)


Ocramius:
As mentioned before, queries don't need mocking, as they are actually DB-specific `eval` (which is also why we get so many SQL injections).

The fact that you expose `Query` as either parameter or return type is the problem here, and it is intentional to have that part as hard/problematic coupling. Simply don't type-hint (in any way) against `QueryBuilder` or `Query` and you will be fine: they are meant to be used only in private APIs.

Please don't pursue this further: `Query` is staying final and un-mockable.

PR #6185, opened 2016-12-21
[BUG] Incorrect handling of ManyToMany relation with unique constraint

aivus:
See https://github.com/doctrine/doctrine2/issues/6186

### DIFF
--- a/tests/Doctrine/Tests/Models/AivusTest/Category.php
+++ b/tests/Doctrine/Tests/Models/AivusTest/Category.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Doctrine\Tests\Models\AivusTest;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+
+/**
+ * @Entity
+ * @Table(name="categories")
+ */
+class Category
+{
+    /**
+     * @Column(type="integer")
+     * @Id
+     * @GeneratedValue
+     *
+     * @var int
+     */
+    protected $id;
+
+    /**
+     * @ManyToMany(targetEntity="Product")
+     * @JoinTable(name="category_to_products",
+     *      joinColumns={@JoinColumn(name="category_id", referencedColumnName="id")},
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/Models/AivusTest/Category.php
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @var int
+     */
+    protected $id;
+
+    /**
+     * @ManyToMany(targetEntity="Product")
+     * @JoinTable(name="category_to_products",
+     *      joinColumns={@JoinColumn(name="category_id", referencedColumnName="id")},
+     *      inverseJoinColumns={@JoinColumn(name="product_id", referencedColumnName="id", unique=true)}
```
Ocramius: This is not valid, as by definition a many-to-many association allows many of both sides. This is a `one-to-many` association that may be implemented incorrectly.

Ocramius:
Closing as `invalid`: the test uses a `ManyToMany` association as a `OneToMany` by adding a unique constraint.

That's not a valid scenario, as the ORM simply assumes a primary key on the join table.

If this is still a problem, redesign the many-to-many association with a join entity, then use surrogate identifiers for joins.

aivus:
> That's not a valid scenario,

It so strange, because:

1) I see tests with `@ManyToMany` with unique=true
At least, here:
https://github.com/doctrine/doctrine2/blob/22b5fb1ad4aaed91f77d4b334d2cbd287bc16bec/tests/Doctrine/Tests/Models/Global/GlobalNamespaceModel.php

2) Here is example of this usage at docs.doctrine-project.org
http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/annotations-reference.html#jointable

3) Also, it's possible case when you need to have `OneToMany` without modifying `*-many` side.

Ocramius:
Yes, uniqueness is indeed an available mapping option, along with all other
join column flags: doesn't mean that it makes sense to the ORM, especially
since many-to-many does in fact allow for multiple instances on each side
by definition.

One possible approach is to add this to mapping validation, and then throw
an exception for unsupported column uniqueness setups.

Issue #6190, opened 2016-12-22
Performing a count on a PersistentCollection is clearing its collection property


andrewjt71:
**Release version which introduced this issue**
v.2.5.6

**Issue Summary**
Many to many relationship between entities retrieved from the DB are being cleared from memory when a count is performed on a PersistentCollection. This issue was introduced by the above release.

**Replication Steps**
Please find in the link below an example scenario affected by this issue.
https://gist.github.com/andrewjt71/711d737e03b23b9ad7cb638e0b88a4f2

NB. This issue only occurs when creating relations between entities which are fetched from the DB.

Ocramius:
Here's the inlined code:

```php
/**
 * User
 *
 * @ORM\Table(name="user")
 * @ORM\Entity
 */
class User
{
    /**
     * @var integer
     *
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * Many Users have Many Groups.
     * @ORM\ManyToMany(targetEntity="Group", inversedBy="users")
     * @ORM\JoinTable(name="users_groups")
     */
    private $groups;

    /**
     * Constructor
     */
    public function __construct()
    {
        $this->groups = new ArrayCollection();
    }
}

/**
 * Group
 *
 * @ORM\Table(name="group")
 * @ORM\Entity
 */
class Group
{
    /**
     * @var integer
     *
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * Many Groups have Many Users.
     * @ORM\ManyToMany(targetEntity="User", mappedBy="groups")
     */
    private $users;

    /**
     * Constructor
     */
    public function __construct()
    {
        $this->users = new ArrayCollection();
    }

    /**
     * @param User $user
     */
    public function addUser(User $user)
    {
        $this->users[] = $user;
    }

    /**
     * @return ArrayCollection
     */
    public function getUsers()
    {
        return $this->users;
    }
}

// Create user.
$user = new User();

// Create Group.
$group = new Group();

// Persist user and group to the DB.
$this->getEntityManager()->persist($user);
$this->getEntityManager()->persist($group);
$this->getEntityManager()->flush();

// Fetch user from DB.
$user = $this->getEntityManager()->getRepository('User')->find(1);

// Fetch Group from DB.
[...]

PR #6191, opened 2016-12-23
Bug - #6190 many to many collection cleared on inverse side association is cleared when unbalanced

Ocramius:
This is a test case imported from #6190

Created because the issue seemed very bad, but seems invalid after a first glance.


Ocramius:
See https://travis-ci.org/doctrine/doctrine2/builds/186232782 for reference: cannot reproduce.

Also: test case is invalid, as it only updates the inverse side of a many-to-many association

Issue #6189, opened 2016-12-22
Pagination: countQuery overrides existing ResultSetMapping


nikolajlovenhardt:
Hello,

```php
public function foo()
{
    $queryBuilder->setParameter('entityClass', $this->getEntityManager()->getClassMetadata(EntityClass::class));

    $queryBuilder->andWhere(sprintf('%s instance of :entityClass', $this->getName()));

    $query = $queryBuilder->getQuery();
}
```

**The issue**
The `getCountQuery` method in `Doctrine\ORM\Tools\Pagination\Paginator` overrides the result mapping in line 251. It is not appropriate because it overwrites the existing result mapping, and therefore in my case returns a count of 0.

So the Paginator has results, but the count is zero.

I'm not sure how to fix this, because the `getResultMapping` method in the Query is private.

Thank you in advance

Ocramius:
Linking [logic that overwrites the mappings](https://github.com/doctrine/doctrine2/blob/cd1a5fcadce6cd5fe3ee17360503d73e82852339/lib/Doctrine/ORM/Tools/Pagination/Paginator.php#L252-L256)

@nikolajlovenhardt I don't see any custom mappings in your case? Can you make this an isolated test case?

nikolajlovenhardt:
@Ocramius 

I don't add any custom mappings, but I tried to change the getResultMapping method to public, and change `$rsm = new ResultSetMapping();` into `$rsm = $countQuery->getResultMapping()`, and that fixed it.

I've made an example below
https://gist.github.com/nikolajlovenhardt/dcaee906027a57e8bb0ee4f8f29c8816

Ocramius:
I understand the issue even less now, or the issue description doesn't
reflect the test that you linked.

I'll try running the test locally though.

Ocramius:
Converted into a failing test case in https://github.com/doctrine/doctrine2/pull/6194

Ocramius:
I applied the patch to #6194, but exposing the resultset mapping is a no-go from here.

Will need an alternate approach instead, or checking why the new mapping doesn't fit.

nikolajlovenhardt:
@Ocramius oh! I didn't test the code. I just wrote a quick example similar to my case. Would you like me to write (and test) some code?

Simply explained: The issue (count() eq 0) appears every time I paginate a query with a where clause using instance of.

PR #6194, opened 2016-12-23
#6189 paginator counting and iterating are inconsistent between each other with inheritances

Ocramius:
This PR provides tests for #6189


lib/Doctrine/ORM/Query.php
```diff
@@ -222,7 +222,7 @@ public function getAST()
     /**
      * {@inheritdoc}
      */
-    protected function getResultSetMapping()
+    public function getResultSetMapping()
```
billschaller: Ugh

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -249,7 +249,8 @@ private function getCountQuery()
         if ($this->useOutputWalker($countQuery)) {
             $platform = $countQuery->getEntityManager()->getConnection()->getDatabasePlatform(); // law of demeter win
 
-            $rsm = new ResultSetMapping();
+            $rsm = $countQuery->getResultSetMapping();
```
billschaller: (┛◉Д◉)┛彡┻━┻

billschaller (approved):

lib/Doctrine/ORM/Query.php
Ocramius: Yes, this is not going to happen.

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6189Test.php
```diff
@@ -0,0 +1,87 @@
...
+ * @group issue-6189
+ */
+class Issue6189Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(Issue6189Product::CLASSNAME),
```
lcobucci: This PR is for `master` lets just use `::class`

Ocramius:
Added `Can't Fix` and `Invalid` labels - don't merge this, if you are reading this in a while. The fix is a prototype, and not a correct one.

PR #6197, opened 2016-12-26
getSingleScalarResult() returns NULL when the query yields no results

BenMorel:
`NoResultException` is not thrown in this case.


Ocramius:
LGTM 👍

Issue #6170, opened 2016-12-13
oneToMany association not flushed


poolerMF:
hello

- class A has oneToMany class B .. cascade=persist
- entities already exists in DB

1. load A .. with B
2. make change to B
3. persist(A)
4. flush(A)

result is, that B is not changed
I think flush should commit A and B (if there is cascade) not only A
when I make flush() everything works OK

is this bug ? or how to flush every ToMany assotiations of entity A  ?

Ocramius:
Needs a test case. Check https://github.com/doctrine/doctrine2/tree/d3f6c5ec70aac4b029a4b61ecf1e2ba61a1a4a6d/tests/Doctrine/Tests/ORM/Functional/Ticket for examples on how to write one.

poolerMF:
added test:
https://github.com/poolerMF/doctrine2/blob/master/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6170Test.php

pull request failed:
https://github.com/doctrine/doctrine2/pull/6198

maybe same problem here:
[[3.0] Single entity flush](https://github.com/doctrine/doctrine2/pull/1139)
[DDC-3344: Flush on a specific entity is not correctly cascaded to associated entities](https://github.com/doctrine/doctrine2/issues/4142)
[Drop support for `EntityManager#flush($something)`](https://github.com/doctrine/doctrine2/issues/6118)
[UnitOfWork::commit with specific entity flashes all removed and persisted entity but not change dependent entity](https://github.com/doctrine/doctrine2/issues/6116)

Ocramius:
Closing as per https://github.com/doctrine/doctrine2/pull/6198#discussion_r93874559

PR #6198, opened 2016-12-26
Added test for issue #6170. (oneToMany association not flushed)

poolerMF:


tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6170Test.php
```diff
@@ -0,0 +1,47 @@
...
+        $address->city = 'Snina';
+
+        $user->setAddress($address);
+
+        $this->_em->persist($user);
+        $this->_em->flush();
+
+        $address->zip = '12345';
+
+        $this->_em->flush($user);
```
Ocramius: Is flushing `$user` intentional? Should just `flush()`.

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6170Test.php
poolerMF: yes, intentional

Ocramius: Then the non-flushed one-to-many is expected behavior, since one-to-many is always inverse side (many-to-one is owning side, and therefore checked for changes)

Ocramius:
Linking #6170

PR #6207, opened 2017-01-02
Documentation: Fixes missing namespaces on class call

gabidavila:
On http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/tools.html#reverse-engineering there are calls like the following:

```php
$cmf = new DisconnectedClassMetadataFactory();
```

Which would cause fatal error if not in the namespace call list, however as documentation for this feature already starts with 

```php
$em->getConfiguration()->setMetadataDriverImpl(
    new \Doctrine\ORM\Mapping\Driver\DatabaseDriver(
        $em->getConnection()->getSchemaManager()
    )
);
```

I added the namespace on the calls itself to maintain consistency.

### DIFF
--- a/docs/en/reference/tools.rst
+++ b/docs/en/reference/tools.rst
@@ -385,7 +385,7 @@ First you need to retrieve the metadata instances with the
         )
     );
     
-    $cmf = new DisconnectedClassMetadataFactory();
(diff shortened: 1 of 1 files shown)


Ocramius:
@gabidavila thanks! 👍

gabidavila:
;)

PR #6208, opened 2017-01-03
[WIP] Generate unique index name across tables

real34:
SQLite does not support multiple unique indexes with same name across tables

... this fixes the issue with name generation

This is an early PR to get some feedbacks in case I am missing somthing. I'll come up with a test case for it by the end of the week.


Ocramius:
Requires test cases that run against both a DB like mysql and sqlite, as well as unit tests.

real34:
🧹 _I'm cleaning up my PRs list. Sorry for not being able to followup on this with test cases. I don't use it Doctrine anymore today, and forgot about this PR. I won't do any further work on this._

PR #6216, opened 2017-01-05
Remove unused "$className" argument

malukenho:
/cc @Ocramius @lcobucci

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 2.5
 
+## Minor BC BREAK: remove $className parameter on AbstractEntityInheritancePersister#getSelectJoinColumnSQL
+
+As `$className` parameter is not used on its method body, it was safely removed.
+
 ## Minor BC BREAK: query cache key time is now a float
 
 As of 2.5.5, the `QueryCacheEntry#time` property will contain a float value
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,8 +1,8 @@
 # Upgrade to 2.5
 
-## Minor BC BREAK: remove $className parameter on AbstractEntityInheritancePersister#getSelectJoinColumnSQL
+## Minor BC BREAK: removed $className parameter on `AbstractEntityInheritancePersister#getSelectJoinColumnSQL`
 
-As `$className` parameter is not used on its method body, it was safely removed.
+As `$className` parameter was not used in the method, it was safely removed.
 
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Persisters/Entity/AbstractEntityInheritancePersister.php
```diff
@@ -82,12 +82,11 @@ protected function getSelectColumnSQL($field, ClassMetadata $class, $alias = 'r'
     /**
      * @param string $tableAlias
      * @param string $joinColumnName
-     * @param string $className
      * @param string $type
      *
      * @return string
      */
-    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $className, $type)
+    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $type)
```
Ocramius: This is a minor BC break, and needs to be documented in [`UPGRADE.md`](https://github.com/doctrine/doctrine2/blob/master/UPGRADE.md)

Ocramius (changes requested):

malukenho:
@Ocramius now that I looked at it more deeply, this change can break code on users' side as `AbstractEntityInheritancePersister#getSelectJoinColumnSQL` is protected and can be overwrited.

Still valid or can be closed?

UPGRADE.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade to 2.5
 
+## Minor BC BREAK: remove $className parameter on AbstractEntityInheritancePersister#getSelectJoinColumnSQL
```
Ocramius: `removed`

Ocramius: Put backticks around `AbstractEntityInheritancePersister#getSelectJoinColumnSQL`

UPGRADE.md
```diff
@@ -1,5 +1,9 @@
...
+
+As `$className` parameter is not used on its method body, it was safely removed.
```
Ocramius: s/`is not used on its method body`/`was not used in the method`

Ocramius (changes requested):

Ocramius:
> @Ocramius now that I looked at it more deeply, this change can break code on users' side as `AbstractEntityInheritancePersister#getSelectJoinColumnSQL` is protected and can be overwrited.

Persisters are internal API, and we don't support custom userland persisters anyway. This is still valid.

PR #6217, opened 2017-01-05
Handle AssociationCacheEntry identifier

gadelkareem:
Fixes `Notice: Undefined property: Doctrine\ORM\Cache\AssociationCacheEntry::$id` when retrieving an associated entity from a Composite primary key


Ocramius:
A test is needed

PR #6222, opened 2017-01-10
Fix invalid joinColumn in ClassMetadata::associationMappings

gadelkareem:
Fix getting `Notice: Undefined index: user_id` for a user field association

### DIFF
--- a/lib/Doctrine/ORM/Utility/IdentifierFlattener.php
+++ b/lib/Doctrine/ORM/Utility/IdentifierFlattener.php
@@ -85,6 +85,9 @@ final class IdentifierFlattener
                 }
 
                 $flatId[$field] = implode(' ', $associatedId);
+            } elseif (array_key_exists($field, $id)) {
+                $flatId[$field] = $id[$field];
+
             } elseif (isset($class->associationMappings[$field])) {
                 $associatedId = array();
 
@@ -93,8 +96,7 @@ final class IdentifierFlattener
                 }
 
                 $flatId[$field] = implode(' ', $associatedId);
-            } else {
-                $flatId[$field] = $id[$field];
+
             }
         }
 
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Utility/IdentifierFlattener.php
```diff
@@ -85,6 +85,9 @@ public function flattenIdentifier(ClassMetadata $class, array $id)
                 }
 
                 $flatId[$field] = implode(' ', $associatedId);
+            } elseif (array_key_exists($field, $id)) {
```
phansys: Can `$id[$field]` be `null`? If that isn't the case, `isset()` is able to deliver better performance.

lib/Doctrine/ORM/Utility/IdentifierFlattener.php
```diff
@@ -93,8 +96,7 @@ public function flattenIdentifier(ClassMetadata $class, array $id)
                 }
 
                 $flatId[$field] = implode(' ', $associatedId);
-            } else {
-                $flatId[$field] = $id[$field];
+
```
phansys: Please, remove surplus line break.

phansys (changes requested):
Status: Needs work.

Ocramius:
Please re-send this against master. Also, tests are absolutely needed, as well as an example of where this is happening.

lib/Doctrine/ORM/Utility/IdentifierFlattener.php
gadelkareem: Yes $id[$field] is null if Eager loading is not enabled.

mpdude:
@gadelkareem is this still an issue?

PR #6229, opened 2017-01-13
Fix bug when source class has parents

kovalevmixail:
An error occurs when the sourceEntity has @Id propperty in parent class.


Ocramius:
Needs test case, can't evaluate it otherwise.

lcobucci:
@Ocramius should we close this one? I mean, it doesn't have test, is from 2013 and the destination branch is `2.4`...

PR #6237, opened 2017-01-16
Allow arithmetic expression as DATE_ADD()/DATE_SUB() interval parameter

ghost:
This PR makes possible to use an arithmetic expression as the interval parameter of DATE_ADD and DATE_SUB functions.
My use case is a complex query with the following HAVING clause:
`HAVING f.releaseDate <= DATE_ADD(DATE_ADD(p.registerTimestamp, SUM(l.validityDays), 'day'), 365, 'day')`
While cascading two DATE_ADD calls works fine in this case, the same clause is much better expressed as
`f.releaseDate <= DATE_ADD(p.registerTimestamp, SUM(l.validityDays) + 365, 'day')`
which is correctly compiled (PostgreSQL) as:
`HAVING f0_.release_date <= (p3_.register_timestamp + SUM(p5_.validity_days) + 365 || ' DAY')::interval)`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 7.0
   - 7.1
   - nightly
+  - hhvm
 
 env:
   - DB=mysql
@@ -70,7 +71,6 @@ matrix:
       env: DB=sqlite
 
   allow_failures:
-    - php: 7.1
     - php: nightly
     - php: hhvm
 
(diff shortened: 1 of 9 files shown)


Ocramius:
Requires test case additions and documentation changes (if there is any EBNF definition for this particular function)

gigib82:
Will do it in a couple of days

Ocramius:
@gigib82 please do rebases, not merges

PR #6242, opened 2017-01-19
Remove primary key requirement from DatabaseDriver.php.

danyPatient:


Ocramius:
Any mapping must come with a non-empty set of identifier fields. This patch leads to the database driver producing invalid mapping definitions, and is therefore invalid.

The correct approach here is to simply skip tables with no PK when reverse-engineering the DB, and that is usually done by passing an exclusion pattern to the CLI tools.

danyPatient:
But what can I do when I have a legacy base which has not any id and needs to be mapped?

danyPatient:
@Ocramius Is updating the database the only way?

Ocramius:
You'll have to reverse engineer it manually.

On 19 Jan 2017 12:00, "danyPatient" <notifications@github.com> wrote:

But what can I do when I have a legacy base which has not any id and needs
to be mapped?

—
You are receiving this because you modified the open/close state.

<https://github.com/doctrine/doctrine2/pull/6242#issuecomment-273745262>,
or mute the thread

danyPatient:
And one I reversed engineer manually, how do I resolve the Exception `No identifier/primary key specified for Entity "AppBundle\Entity\SomeEntity". Every Entity must have an identifier/primary key`?

Ocramius:
You will have to add an identifier to your entities: the ORM can't work without it.

PR #5831, opened 2016-05-20
Multi get implementation in QueryCache

polyanin:
Hi

There is significant performance degradation when doctrine tries to check the cache (memcache) after fetching multiple thousands of rows. It occurs because it doesn't use memcache's multiple get function.

This modification implements multiple get cache checking as it has been done here: https://github.com/doctrine/doctrine2/pull/954

PR #6245, opened 2017-01-20
Use `getMultiple()` to fetch associations as well in L2 cache fetch operations

lcobucci:
Just as it was done on #5831

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -122,11 +122,9 @@ class DefaultQueryCache implements QueryCache
 
         // @TODO - move to cache hydration component
         foreach ($entry->result as $index => $entry) {
-
             $entityEntry = is_array($entries) && array_key_exists($index, $entries) ? $entries[$index] : null;
 
             if ($entityEntry === null) {
-
                 if ($this->cacheLogger !== null) {
                     $this->cacheLogger->entityCacheMiss($regionName, $cacheKeys->identifiers[$index]);
                 }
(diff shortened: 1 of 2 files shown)


lcobucci:
btw I know this is a bit messy, so please suggest a nice way to improve this (I'm a bit sleepy right now)

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -178,14 +176,20 @@ public function get(QueryCacheKey $key, ResultSetMapping $rsm, array $hints = []
                     continue;
                 }
 
-                $collection  = new PersistentCollection($this->em, $assocMetadata, new ArrayCollection());
+                $generateKeys = function ($id) use ($assocMetadata): EntityCacheKey {
```
guilhermeblanco: PHP7 only feature!

guilhermeblanco (dismissed):

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
lcobucci: `master` already dropped PHP 5.6

Ocramius: @guilhermeblanco we dropped PHP 5 support in `master` as well

LuffyLyon:
Hello,

This PR will be merged to the 2.6 version that's right?
Because I did the same work in my project to improve performance, I didn't see this PR before, so II've overriden the DefaultQueryCache class.

The PR is still in improvement?

lcobucci:
@LuffyLyon this is only for 2.6 indeed and @Ocramius wanted to have some tests but we bumped into mocking hell. We still need to decide what to do with it (there're some duplications that I'm not really happy with)

LuffyLyon:
Ok thank you, which duplications are you talking about? Maybe I can contribute to improving this feature

Ocramius (approved):
LGTM :+1:

PR #6250, opened 2017-01-23
Note on Entity listeners

johndodev:
As requested here https://github.com/doctrine/doctrine2/issues/6247#issuecomment-274123780 I thought it should be interesting to note that.
The note could be longer, like "You should avoid to have multiple methods for the same event (e.g. multiple @PrePersist) since the execution order is not guaranteed".


Ocramius (approved):

Ocramius:
@johndodev thanks!

PR #6251, opened 2017-01-24
Add missing group to DDC2660Test.php

mbeccati:
With the current PHPUnit version and PHP 7.2/master, the test fails with:

1) Doctrine\Tests\ORM\Functional\Ticket\DDC2660Test::testIssueWithExtraColumn
array_flip(): Can only flip STRING and INTEGER values!

2) Doctrine\Tests\ORM\Functional\Ticket\DDC2660Test::testIssueWithoutExtraColumn
array_flip(): Can only flip STRING and INTEGER values!

Due to the group being NULL vs "" on older PHP versions.

I will also file a bug report or fix to PHPUnit and/or PHP itself, but it sounds
like the missing group was just a typo.


Ocramius (approved):

Ocramius:
@mbeccati nice find!

PR #6253, opened 2017-01-25
Fix xmlDriver mapping when cannot find using simplexml_load_file

mistraal:
Please refer to [this issue](https://github.com/wallabag/wallabag/issues/1828) and [this issue](https://github.com/FriendsOfSymfony/FOSUserBundle/issues/1062). 

While this is a very specific case, it is not exclusive to Nginx server as it also happens on my Apache setup.


mistraal:
Sorry this PR got screwed up. Just wanted to make a small change in one file and it added a load more.

PR #930, opened 2014-02-04
Transactional callback isn't returning false

joni2back:
Trying to fix any situation that if we've something callback with explicit FALSE return, like:

``` php
$rtn = $mgr->transactional(function() {...; return false;});
$rtn //true 
```

Should be the return value of the callback method, 
If the method is not returning any value (null), then.. we consider the callback was done, then return true.

Regards.

PR #6254, opened 2017-01-25
Fix issue where xml mapping cannot find file on certain configurations

mistraal:
Please refer to [this issue](https://github.com/wallabag/wallabag/issues/1828) and [this issue](https://github.com/FriendsOfSymfony/FOSUserBundle/issues/1062). 

While this is a very specific case, it is not exclusive to Nginx server as it also happens on my Apache setup.


Ocramius:
This needs a test case.

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -911,7 +911,8 @@ private function getCascadeMappings(SimpleXMLElement $cascadeElement)
     protected function loadMappingFile($file)
     {
         $result = [];
-        $xmlElement = simplexml_load_file($file);
+        $xmlData = file_get_contents($file);
```
Ocramius: Can be inlined

Ocramius (changes requested):

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -911,7 +911,7 @@ private function getCascadeMappings(SimpleXMLElement $cascadeElement)
     protected function loadMappingFile($file)
     {
         $result = [];
-        $xmlElement = simplexml_load_file($file);
+        $xmlElement = simplexml_load_string(file_get_contents($file));
```
phansys: Is there any situation where [`simplexml_load_file()`](http://php.net/manual/en/function.simplexml-load-file.php) is not returning `false` with a not found file?

phansys (changes requested):
Status: Answer pending.

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
Ocramius: > Is there any situation where [`simplexml_load_file()`](http://php.net/manual/en/function.simplexml-load-file.php) is not returning `false` with a not found file?

Indeed, that's what a test would verify, but I see no tests.

phansys:
This is related to #930.

lcobucci:
@mistraal can you please rebase with `develop` dropping all commits that aren't yours please?

PR #6257, opened 2017-01-25
Adding a note about the `Doctrine\ORM\QueryBuilder#addCriteria()` method

weaverryan:
Hi guys!

Just adding a note to connect things better.

Thanks!

### DIFF
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -616,10 +616,10 @@ address reference. When flush is called not only are the references removed
 but both the old standing data and the one address entity are also deleted 
 from the database.
 
+.. _filtering-collections:
+
 Filtering Collections
 ---------------------
-
-.. filtering-collections:
 
 Collections have a filtering API that allows to slice parts of data from
 a collection. If the collection has not been loaded from the database yet,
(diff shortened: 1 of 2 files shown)


Ocramius (approved):

Ocramius:
👍 thanks @weaverryan

docs/en/reference/working-with-associations.rst
```diff
@@ -616,11 +616,11 @@ address reference. When flush is called not only are the references removed
 but both the old standing data and the one address entity are also deleted 
 from the database.
 
+.. _filtering-collections:
+
 Filtering Collections
 ---------------------
 
-.. filtering-collections:
```
weaverryan: I believe the `_` was left off accidentally before - I'm not familiar with this syntax without an underscore, and I searched the codebase and found NO other references like this (`git grep '\.\. ' | grep -v 'code-block' | grep -v 'note' | grep -v 'section' | grep -v '\.\. _' | grep -v 'configuration-block' | grep -v versionadded | grep -v 'warning' | grep -v '//' | grep -v '\-\->' | grep -v 'include' | grep -v 'caution' | less` gets you a short list, where you can see this). Also, this section header already gave us a `#filtering-collections` anchor, even without the added reference.

PR #6265, opened 2017-01-29
Removing detach and merge from the entity manager

mikeSimonson:
and the related tests


mikeSimonson:
@Ocramius Not sure what to do with the last test that fails Doctrine\Tests\ORM\Functional\Ticket\DDC3303Test::testEmbeddedObjectsAreAlsoInherited ?

Travis does weird stuff for some reasons.

Is this sort of what you expected ?

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -669,51 +669,6 @@ public function refresh($entity)
...
-     *
-     * @throws ORMInvalidArgumentException
-     */
-    public function detach($entity)
-    {
-        if ( ! is_object($entity)) {
-            throw ORMInvalidArgumentException::invalidObject('EntityManager#detach()', $entity);
-        }
-
-        $this->unitOfWork->detach($entity);
```
Ocramius: These methods are to be removed as well

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -669,51 +669,6 @@ public function refresh($entity)
...
-     */
-    public function merge($entity)
-    {
-        if ( ! is_object($entity)) {
-            throw ORMInvalidArgumentException::invalidObject('EntityManager#merge()', $entity);
-        }
-
-        $this->errorIfClosed();
-
-        return $this->unitOfWork->merge($entity);
```
Ocramius: These methods are to be removed as well

Ocramius (changes requested):

Ocramius:
@mikeSimonson can't see failures, as the `EntityManagerInterface` is not fully covered by an implementation now.

lcobucci:
@mikeSimonson embeddables were temporarily disabled by @guilhermeblanco on `develop` 😉

mikeSimonson:
@lcobucci Thanks for the info, but still it's weird that the test passed before I removed the merge and detach functionality no ?

guilhermeblanco:
@lcobucci there's much more to be removed and decided than a blank removal of the EM methods.
UoW also contains a bunch of related methods to merge/detach, but the key decisions to be made are how do we handle the entity states now that there's no more detached state. If you look at UoW methods, there're lots of state detached handling there.

IMHO, whenever we have state detached we should either reintegrate the state as managed or consider it as new.

Once you have that sorted out, I can properly review the PR. =)

Ocramius:
I would honestly just don't deal with that, and instead let users decide
how to:

 * Hydrate from array data (custom hydrator)
 * Manage dirty state (if integrated entities need to be added to the ORM)
 * Load existing state on their own (they can assume existing records or
not, since they decided how to detach on their own)

Ocramius:
Handled in https://github.com/doctrine/doctrine2/pull/1577

Issue #5887, opened 2016-06-20
Failing inverse One-to-One relationship using a custom data type


alexb-uk:
## Issue

The BasicEntityPersister, possibly loadOneToOneEntity() method, does not handle One-to-One relationships on the inverse side when using a custom data type.

I've stepped through the code and can see that getTypes() method is failing to find a mapping due to the SQL field name being passed in rather than the class field. E.g. from the example below its "t0.shipping_id" but I think should be "shipping". Unfortunately my understanding of the internals of Doctrine is not enough to spot the exact cause / place to fix.
## Steps to reproduce

Create two classes with a one-to-one relationship using a custom data type. I used the Ramsey UUID library and Doctrine integration.

Create DQL that filters based on the inverse side and not that hydrated object returned has NULL for the relationship.

Filtering on owning side or using entityManager-find('Shipping', $id) both work correctly.

If I can give any more information or perform any tests please don't hesistate to ask.
## Versions

doctrine/common                     v2.5.3
doctrine/dbal                       v2.5.4
doctrine/doctrine-module            1.0.0
doctrine/doctrine-orm-module        0.10.0
doctrine/orm                        v2.5.4
ramsey/uuid-doctrine                1.2.0
## YAML mappings

```
Product:
  type: entity
  table: products
  id:
    id:
      type: uuid_binary
      generator:
        strategy: CUSTOM
      customIdGenerator:
        class: Ramsey\Uuid\Doctrine\UuidGenerator

  fields:
    name:
      type: string

  oneToOne:
      shipping:
        targetEntity: Voicesafe\Test\Shipping
        joinColumn:
          name: shipping_id
          referencedColumnName: id

Shipping:
  type: entity
  table: shipping
  id:
    id:
      type: uuid_binary
      generator:
        strategy: CUSTOM
      customIdGenerator:
        class: Ramsey\Uuid\Doctrine\UuidGenerator

  fields:
      transport:
        type: string

  oneToOne:
      product:
        targetEntity: Voicesafe\Test\Product
        mappedBy: shipping
```
## Working Example - Owning Side

```
$qb = $entityManager->createQueryBuilder();

$qb->select('p')
    ->from('Product', 'p')
    ->where('p.shipping = ?1')
    ->setParameter(1, $id, 'uuid_binary');

$query  = $qb->getQuery();
$result = $query->getResult();

Debug::dump($result);
```
## SQL Generated

```
[...]

PR #6274, opened 2017-02-03
Fixing #5887 - lazy loading of one-to-one relationship with custom id object

necsord:
Fixing case when lazy loading of entity in one-to-one relationship on the side without foreign key
did not return the entity because of custom type not being detected and properly formatted before
applying to database query.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -785,7 +785,7 @@ class BasicEntityPersister implements EntityPersister
 
             // unset the old value and set the new sql aliased value here. By definition
             // unset($identifier[$targetKeyColumn] works here with how UnitOfWork::createEntity() calls this method.
-            $identifier[$this->getSQLTableAlias($targetClass->name) . "." . $targetKeyColumn] =
+            $identifier[$targetClass->getFieldForColumn($targetKeyColumn)] =
                 $sourceClass->reflFields[$sourceClass->fieldNames[$sourceKeyColumn]]->getValue($sourceEntity);
 
             unset($identifier[$targetKeyColumn]);
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5887Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5887Test.php
@@ -1,0 +1,234 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\DBAL\Types\StringType;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group 5887
+ */
+class GH5887Test extends OrmFunctionalTestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+
+        Type::addType(GH5887CustomIdObjectType::NAME, GH5887CustomIdObjectType::class);
+
+        $this->_schemaTool->createSchema(
+            [
+                $this->_em->getClassMetadata(GH5887Cart::class),
+                $this->_em->getClassMetadata(GH5887Customer::class),
+            ]
+        );
+    }
+
+    public function testLazyLoadsForeignEntitiesInOneToOneRelationWhileHavingCustomIdObject()
+    {
+        $customerId = new GH5887CustomIdObject(1);
+        $customer = new GH5887Customer();
+        $customer->setId($customerId);
+
+        $cartId = 2;
+        $cart = new GH5887Cart();
+        $cart->setId($cartId);
+        $cart->setCustomer($customer);
+
+        $this->_em->persist($customer);
+        $this->_em->persist($cart);
+        $this->_em->flush();
+
+        // Clearing cached entities
+        $this->_em->clear();
+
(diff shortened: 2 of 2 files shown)


tests/Doctrine/Tests/Models/Issue5887/Issue5887Cart.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Doctrine\Tests\Models\Issue5887;
+
+/**
+ * @Entity
+ * @Table(name="issue5887_cart")
+ */
+class Issue5887Cart
```
lcobucci: Please move this class to the test file (like we have [here](https://github.com/doctrine/doctrine2/blob/e4704beaf9fad5a521fe7592dec382ae209b3cc1/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6141Test.php)).

tests/Doctrine/Tests/Models/Issue5887/Issue5887CustomIdObject.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Doctrine\Tests\Models\Issue5887;
+
+class Issue5887CustomIdObject
```
lcobucci: Please move this class to the test file (like we have [here](https://github.com/doctrine/doctrine2/blob/e4704beaf9fad5a521fe7592dec382ae209b3cc1/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6141Test.php)).

tests/Doctrine/Tests/Models/Issue5887/Issue5887CustomIdObjectType.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Doctrine\Tests\Models\Issue5887;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\Type;
+
+class Issue5887CustomIdObjectType extends Type
```
lcobucci: Please move this class to the test file (like we have [here](https://github.com/doctrine/doctrine2/blob/e4704beaf9fad5a521fe7592dec382ae209b3cc1/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6141Test.php)).

lcobucci: It can also inherit `StringType` instead so you don't need to implement the method `getSQLDeclaration()`.

tests/Doctrine/Tests/Models/Issue5887/Issue5887Customer.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+namespace Doctrine\Tests\Models\Issue5887;
+
+/**
+ * @Entity
+ * @Table(name="issue5887_customer")
+ */
+class Issue5887Customer
```
lcobucci: Please move this class to the test file (like we have [here](https://github.com/doctrine/doctrine2/blob/e4704beaf9fad5a521fe7592dec382ae209b3cc1/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6141Test.php)).

tests/Doctrine/Tests/ORM/Functional/SchemaValidatorTest.php
```diff
@@ -16,7 +16,7 @@ static public function dataValidateModelSets()
     {
         $modelSets = [];
         foreach (self::$_modelSets as $modelSet => $classes) {
-            if ($modelSet == "customtype") {
+            if (in_array($modelSet, ["customtype", "issue5887"])) {
```
lcobucci: Not needed for inlined classes

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -302,6 +302,10 @@
             Models\Issue5989\Issue5989Employee::class,
             Models\Issue5989\Issue5989Manager::class,
         ],
+        'issue5887' => [
```
lcobucci: Not needed for inlined classes

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -576,6 +580,11 @@ protected function tearDown()
             $conn->executeUpdate('DELETE FROM issue5989_managers');
         }
 
+        if (isset($this->_usedModelSets['issue5887'])) {
```
lcobucci: Not needed for inlined classes

lcobucci (changes requested):
@Necsord thanks for the PR but unfortunately it doesn't solve the issue.

Your test is only passing because you're setting the parameter type on the query (`->setParameter('id', $customerId, Issue5887CustomIdObjectType::NAME)`).

Necsord:
@lcobucci Thanks for looking into this. I've applied changes regarding your comments and simplified test case to test only the most basic case I could come up with.

The problem is not with the first query for `Customer` which as you pointed out uses type, it's with the second one for `Cart` initiated in `\Doctrine\ORM\UnitOfWork::createEntity`:

```
// Inverse side of x-to-one can never be lazy
$class->reflFields[$field]->setValue($entity, $this->getEntityPersister($assoc['targetEntity'])->loadOneToOneEntity($assoc, $entity));
```

If you remove the applied change in `BasicEntityPersister` the test will fail.

lcobucci (approved):

lcobucci:
@Necsord I've just rebased your branch to sync with `master` and squash the commits. Patch LGTM just waiting for tests to pass and I'll merge it.

Thanks for your contribution 😉

PR #6284, opened 2017-02-10
#6217 when hydrating an entity with a composite primary key that is both an `EAGER` and a `LAZY` association and also cached, the `DefaultQueryCache` tries to pass L2 cache implementation detail objects to the `UnitOfWork`

gadelkareem:
I recreated the branch for https://github.com/doctrine/doctrine2/pull/6217 which forced closing to PR so I am submitting this as a new PR. 

DESC: Handle AssociationCacheEntry identifier

Fixes Notice: Undefined property: Doctrine\ORM\Cache\AssociationCacheEntry::$id when retrieving an associated entity from a Composite primary key

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -29,6 +29,7 @@ use ReflectionClass;
 use Doctrine\Common\Persistence\Mapping\ClassMetadata;
 use Doctrine\Common\ClassLoader;
 use Doctrine\ORM\Cache\CacheException;
+use Doctrine\ORM\Cache\AssociationCacheEntry;
 
 /**
  * A <tt>ClassMetadata</tt> instance holds all the object-relational mapping metadata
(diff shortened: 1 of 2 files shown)


Ocramius:
@lcobucci if you merge this one, can you please check if the test can be moved to a smaller SUT test?

lcobucci (approved):
@gadelkareem thanks for the patch!

@Ocramius I've managed to reduce the SUT and saw that the problem ONLY happens when the fetch mode is `EAGER`.

I've rebased it, removed some unrelated CS fixes and applied the same logic [here](https://github.com/doctrine/doctrine2/pull/6284/files#diff-093aaad4da679d0374260757bf7bafcaR722).

@Ocramius although it LGTM would be nice to have your eyes too

Ocramius (changes requested):
Ouch, forgot to send in this review :-\

Ocramius:
Re-iterating this: I don't think that classmetadata should ever interact with an `AssociationCacheEntry` object. This patch fixes a symptom, rather than the cause of the issue, which is the fact that an `AssociationCacheEntry` object is given to the metadata in first place.

Ocramius:
I checked this patch locally, and the bug is fixed on the wrong end here: the `DefaultQueryCache` is passing incorrect data to `UnitOfWork#createEntity()`. I'm currently fixing that :+1:

Ocramius:
@gadelkareem please check #6640 and verify it against your codebase.

gadelkareem:
@Ocramius thanks for the fix and sorry for having no time to help
All tests are passing including GH6217Test so I think this is fixed.

PR #6286, opened 2017-02-12
Fix HHVM build according to Travis documentation

lcobucci:
Reference: https://docs.travis-ci.com/user/languages/php#HHVM-versions-on-Trusty

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,15 @@
 sudo: false
 language: php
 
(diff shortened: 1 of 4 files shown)


lcobucci:
@Ocramius now the test suite at least runs but something is failing miserably when trying to create a connection using ubuntu trusty 😞

lcobucci:
Alright, I give up too 😞

PR #6287, opened 2017-02-12
Fix EntityManagerDecorator tests

lcobucci:
We were expecting a return value for all methods but on https://github.com/doctrine/common/commit/6a56df9a2445db166dbdae8079b9608f72a8496e
we changed some methods to be void and of course that affected the
ORM.


Ocramius:
LGTM 👍

Issue #6297, opened 2017-02-22
ManyToOne association entity identifier


podorozhny:
Hi everyone.

I ran into a problem using many-to-one association between entities while using custom property (not @ORM\Id() one) to join mapped entity.

Simplified example:
``` php
class User {
    /**
     * @ORM\ManyToOne(targetEntity="...\Group", inversedBy="users")
     * @ORM\JoinColumn(referencedColumnName="second_id", nullable=false)
     */
    private $group;

    ...
}

class Group {
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @ORM\Column(type="string")
     */
    private $secondId;

    /**
     * @ORM\OneToMany(targetEntity="...\User", mappedBy="users")
     */
    private $users;
}
```

`bin/console doctrine:schema:validate` said all mappings are okay, but when I tried to insert new user I got this error: 
```
CRITICAL - Uncaught PHP Exception Symfony\Component\Debug\Exception\ContextErrorException: "Notice: Undefined index: secondId" at /application/vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php line 671
```

https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L671

I got `['id' => 100500]` [here](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L658) instead of `['secondId' => 100500]`.

Thanks.

lcobucci:
@podorozhny it seems odd, can you please send a PR with a test that reproduce that issue (we can use https://github.com/doctrine/doctrine2/blob/e4704beaf9fad5a521fe7592dec382ae209b3cc1/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5762Test.php as example).

podorozhny:
@lcobucci

https://github.com/doctrine/doctrine2/pull/6301

```
➜  doctrine2 git:(master) ✗ ./vendor/bin/phpunit --filter=GH6297Test
PHPUnit 5.7.14-1-g7d5e096 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 555 ms, Memory: 36.00MB

There was 1 error:

1) Doctrine\Tests\ORM\Functional\Ticket\GH6297Test::testIssue
[...]

PR #6301, opened 2017-02-23
Issue #6297 functional test

podorozhny:
https://github.com/doctrine/doctrine2/issues/6297

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
@@ -36,9 +36,8 @@ class GH6297Test extends OrmFunctionalTestCase
      */
     public function testIssue()
     {
-        $group = new GH6297Group(1, 'test');
-
-        $user = new GH6297User(1, $group);
+        $group = new GH6297Group(1, 1);
+        $user  = new GH6297User(1, $group);
 
         $this->_em->persist($group);
         $this->_em->persist($user);
@@ -57,7 +56,7 @@ class GH6297Test extends OrmFunctionalTestCase
 class GH6297User
 {
     /**
-     * @Id()
+     * @Id
      * @Column(type="integer")
      */
     public $id;
@@ -72,9 +71,9 @@ class GH6297User
      * @param int         $id
      * @param GH6297Group $group
      */
-    public function __construct($id, GH6297Group $group)
+    public function __construct(int $id, GH6297Group $group)
     {
-        $this->id    = (int) $id;
+        $this->id    = $id;
         $this->group = $group;
     }
 }
@@ -86,7 +85,7 @@ class GH6297User
 class GH6297Group
 {
     /**
-     * @Id()
+     * @Id
      * @Column(name="primary_id", type="integer")
      */
     public $primaryId;
@@ -105,10 +104,10 @@ class GH6297Group
      * @param int $primaryId
      * @param int $secondaryId
      */
-    public function __construct($primaryId, $secondaryId)
+    public function __construct(int $primaryId, int $secondaryId)
     {
-        $this->primaryId   = (int) $primaryId;
-        $this->secondaryId = (int) $secondaryId;
+        $this->primaryId   = $primaryId;
+        $this->secondaryId = $secondaryId;
         $this->users       = new ArrayCollection();
     }
 }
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
```diff
@@ -0,0 +1,114 @@
...
+     * @ManyToOne(targetEntity="GH6297Group", inversedBy="users")
+     * @JoinColumn(referencedColumnName="secondary_id", nullable=false)
+     */
+    public $group;
+
+    /**
+     * @param int         $id
+     * @param GH6297Group $group
+     */
+    public function __construct($id, GH6297Group $group)
```
lcobucci: `master` requires PHP7+ so you can use `int $id`

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * @OneToMany(targetEntity="GH6297User", mappedBy="users")
+     */
+    public $users;
+
+    /**
+     * @param int $primaryId
+     * @param int $secondaryId
+     */
+    public function __construct($primaryId, $secondaryId)
```
lcobucci: Add typehints here instead of casting the values

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * @Entity
+ * @Table(name="users")
+ */
+class GH6297User
+{
+    /**
+     * @Id()
```
lcobucci: Parentheses are not needed here

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * @Entity
+ * @Table(name="groups")
+ */
+class GH6297Group
+{
+    /**
+     * @Id()
```
lcobucci: Parentheses are not needed here

lcobucci (changes requested):
Now I could see better what's happening.

According to the [documentation](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/annotations-reference.html#joincolumn)

> referencedColumnName: Name of the primary key identifier that is used for joining of this relation.

So that's indeed an invalid mapping and I'd say that the `SchemaValidator` should warn about that. What do you say @Ocramius?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
```diff
@@ -0,0 +1,114 @@
...
+        $this->_em->flush();
+        $this->_em->clear();
+
+        self::assertTrue(true);
+    }
+}
+
+/**
+ * @Entity
+ * @Table(name="gh6297_users")
```
lcobucci: You can just drop `@Table` here

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
```diff
@@ -0,0 +1,114 @@
...
+    public function __construct($id, GH6297Group $group)
+    {
+        $this->id    = (int) $id;
+        $this->group = $group;
+    }
+}
+
+/**
+ * @Entity
+ * @Table(name="gh6297_groups")
```
lcobucci: You can just drop `@Table` here

podorozhny:
> referencedColumnName: Name of the primary key identifier that is used for joining of this relation.

So, I can't use "secondary" properties to join entities? Seems like must-have feature.
It's a pity!

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
podorozhny: Already renamed table name with issue number prefix.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6297Test.php
podorozhny: Already renamed table name with issue number prefix.

podorozhny:
@Ocramius

Ocramius:
An identifier is an identifier, and for the ORM it means that:
 * it doesn't change (immutable)
 * it is unique
 
The ORM will always require identifiers to be referenced for any kind of association mapping, as it would otherwise also be impossible to identify those associations in the ORM's in-memory state.

This is both a reasonable and an intentional limitation, and also is good DB design (if you remove the ORM from the discussion).

Closing as `invalid`

PR #6316, opened 2017-03-03
Issue-6313: orphanRemoval doesn't start cascade delete on child entit…

yurybykov:
Add ticket that show bug with orphanRemoval child entity deletion in MySQL (MariaDB).


Ocramius:
Duplicate of https://github.com/doctrine/doctrine2/pull/6318#pullrequestreview-25186354

Issue #6317, opened 2017-03-04
Doctrine build wrong deletion graph on One-To-Many Self-referencing relation with JOINED table inheritance.


yurybykov:
Doctrine build wrong deletion graph on One-To-Many Self-referencing relation with JOINED table inheritance.

PR #6318, opened 2017-03-04
Issue-6313: fix orphanRemoval deletion if orphan has own relations

yurybykov:


yurybykov:
Why
`
if ($enableSecondLevelCache) {
  $evm->addEventListener('loadClassMetadata', new CacheMetadataListener());
}
`
in OrmFunctionalTestCase and not
`if ($this->isSecondLevelCacheEnabled || $enableSecondLevelCache) {
            $evm->addEventListener('loadClassMetadata', new CacheMetadataListener());
        }`

Travis enable second level cache like ENABLE_SECOND_LEVEL_CACHE=1

lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
```diff
@@ -182,10 +182,15 @@ public function removeElement(PersistentCollection $collection, $element)
             return false;
         }
 
+        $this->uow->scheduleOrphanRemoval($element);
+
+        return true;
+        /*
         return $this
```
Ocramius: In the context of an `EXTRA_LAZY` association, an immediate delete seems more correct here, if I read it correctly.

tests/Doctrine/Tests/ORM/Functional/ExtraLazyCollectionTest.php
```diff
@@ -1146,6 +1146,7 @@ public function testRemoveOrphanedManagedElementFromOneToManyExtraLazyCollection
 
         $user->userLists->removeElement($this->_em->find(UserList::class, $userListId));
 
+        $this->_em->flush();
```
Ocramius: Test changes in here indicate a BC break in the current logical changes.

tests/Doctrine/Tests/ORM/Functional/ExtraLazyCollectionTest.php
```diff
@@ -1199,6 +1200,7 @@ public function testRemoveOrphanedManagedLazyProxyFromExtraLazyOneToMany()
 
         $user->userLists->removeElement($this->_em->getReference(UserList::class, $userListId));
 
+        $this->_em->flush();
```
Ocramius: Test changes in here indicate a BC break in the current logical changes.

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6313Test.php
```diff
@@ -0,0 +1,245 @@
...
+    }
+}
+
+/**
+ * @Entity
+ * @InheritanceType("SINGLE_TABLE")
+ * @DiscriminatorColumn(name="type_id", type="integer")
+ * @DiscriminatorMap({1 = "Issue6313_Unit1", 2 = "Issue6313_Unit2"})
+ */
+abstract class Issue6313_Unit
```
Ocramius: Is an inheritance really needed?

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6313Test.php
```diff
@@ -0,0 +1,245 @@
...
+
+final class Issue6313Test extends OrmFunctionalTestCase
+{
+    protected function setUp()
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema(
+            [
+                $this->_em->getClassMetadata(Issue6313_Unit::class),
```
Ocramius: Please reduce the test case to the minimum possible amount of fields/entities needed to replicate the issue without your patch applied, then please rename entities so they make sense in the context of this test, rather than their origin context (your codebase)

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6313Test.php
```diff
@@ -0,0 +1,245 @@
...
+
+        $unit1 = $this->_em->find(Issue6313_Unit::class, $unit1->id);
+        $port1 = $this->_em->find(Issue6313_Port::class, $port1->id);
+        $unit1->removePort($port1);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $unit1 = $this->_em->find(Issue6313_Unit::class, $unit1->id);
+
+        self::assertEquals(1, $unit1->ports->count());
```
Ocramius: `assertCount`

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6313Test.php
yurybykov: It's part of my app and I need this logic. You are right, test need to be reduced for orphanRemoval, but when I test this code I found one more bug, so I open new issue 6318. #6318 different from this because its show wrong CommitOrderCalculator for One-To-Many self-ref with JOINED table.

lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
yurybykov: OrphanRemoval its hook for remove Entity from collection and it doesn't matter which it type of.
If no OrphanRemoval we do 2 operation: remove from collection and remove from EM, if yes - just one remove from collection and Doctrine automatically collect entity for deletion.
And I think this operation should be execute in transaction on flush. Early, I saw issue about this.

tests/Doctrine/Tests/ORM/Functional/ExtraLazyCollectionTest.php
yurybykov: Yes, it's BC break, but it will become in logic of Doctrine. All changes must be complete with flush.

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6313Test.php
Ocramius: I missed this in the #6317 reference explaining that this applies specifically to inheritances, sorry.

tests/Doctrine/Tests/ORM/Functional/ExtraLazyCollectionTest.php
Ocramius: Yes, but it's still a BC break ;-)

I'm fine with documenting it after we looked at what scenarios may be affected by it.

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6313Test.php
yurybykov: Yes, of course. Thank you!

yurybykov:
@Ocramius test updated

PR #6323, opened 2017-03-06
The Entity Generator can't generate annotation correctly

LinboLen:
…`options` should mapping into Entity annotation.

By the way, I don't like to use
```
'options={' . $this->exportTableOptions((array) $metadata->table['options']) . '}';
```
just use json_encode instead

 Changes to be committed:
	modified:   lib/Doctrine/ORM/Tools/EntityGenerator.php

### DIFF
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -1070,7 +1070,7 @@ public function __construct(<params>)
         }
 
         if (isset($metadata->table['options']) && $metadata->table['options']) {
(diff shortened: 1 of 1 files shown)


xsilen:
same as https://github.com/doctrine/doctrine2/pull/6322

Ocramius:
Looks good, although:

 * we need tests
 * we need tests for unicode characters
 * Full JSON support in annotations is not complete yet, see https://github.com/doctrine/annotations/pull/122

Majkl578:
Hello, I'm going to close this incomplete PR as we will be removing EntityGenerator in Doctrine 3.0 completely and 2.x is closed feature-wise. Thank you for the PR.

PR #6325, opened 2017-03-06
Improve exception message for DQL single-valued association path expression to an inverse side (unsupported)

Isinlor:
Hello,
I have added path expression to the exception about single-valued association to an inverse side being not supported. The path expression makes it easier to identify the issue among many other expression in potentially complex query. 

[Seems like many people had an issue with this exception.](https://www.google.com/search?q=A+single-valued+association+path+expression+to+an+inverse+side+is+not+supported+in+DQL+queries.)

For a query:
```sql
SELECT u FROM Doctrine\Tests\Models\CMS\CmsUser u WHERE u.address IN (?1, ?2)
```

The exception message before:

> A single-valued association path expression to an inverse side is not supported in DQL queries. Use an explicit join instead.

The exception message after:

> A single-valued association path expression to an inverse side is not supported in DQL queries. Instead of 'u.address' use an explicit join.

### DIFF
--- a/lib/Doctrine/ORM/Query/QueryException.php
+++ b/lib/Doctrine/ORM/Query/QueryException.php
@@ -204,13 +204,15 @@ class QueryException extends \Doctrine\ORM\ORMException
     }
 
     /**
+     * @param object $pathExpr
+     *
      * @return QueryException
      */
-    public static function associationPathInverseSideNotSupported()
+    public static function associationPathInverseSideNotSupported($pathExpr)
     {
         return new self(
-            "A single-valued association path expression to an inverse side is not supported".
-            " in DQL queries. Use an explicit join instead."
+            "A single-valued association path expression to an inverse side is not supported in DQL queries. " .
+            "Instead of '" . $pathExpr->identificationVariable . "." . $pathExpr->field . "' use an explicit join."
         );
     }
 
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -671,7 +671,7 @@ class SqlWalker implements TreeWalker
                 $assoc = $class->associationMappings[$fieldName];
 
                 if ( ! $assoc['isOwningSide']) {
-                    throw QueryException::associationPathInverseSideNotSupported();
+                    throw QueryException::associationPathInverseSideNotSupported($pathExpr);
                 }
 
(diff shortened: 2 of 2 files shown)


lcobucci (approved):
Even though it's a very tiny BC break it LGTM. @Ocramius should we document this in the v2.6 changelog?

lib/Doctrine/ORM/Query/QueryException.php
```diff
@@ -204,13 +204,15 @@ public static function overwritingJoinConditionsNotYetSupported($assoc)
     }
 
     /**
+     * @param object $pathExpr
+     *
      * @return QueryException
      */
-    public static function associationPathInverseSideNotSupported()
+    public static function associationPathInverseSideNotSupported($pathExpr)
```
Ocramius: I know the caller side uses abbreviations, but please don't abbreviate here.

lib/Doctrine/ORM/Query/QueryException.php
```diff
@@ -204,13 +204,15 @@ public static function overwritingJoinConditionsNotYetSupported($assoc)
     }
 
     /**
+     * @param object $pathExpr
```
Ocramius: Is there really nothing more detailed than `object`? :| Sounds like the `SqlWalker` internal API could need a DTO

Ocramius (changes requested):
LGTM besides nitpicking. As @lcobucci mentioned, we need an entry in `UPGRADE.md` for this, since it still counts as BC Break.

Ocramius:
And yep, a test for the static method is needed :-\

lib/Doctrine/ORM/Query/QueryException.php
Isinlor: [I was following this example.](https://github.com/Isinlor/doctrine2/blob/4f854ae7cca3f94df363321b5e9ab878c0580249/lib/Doctrine/ORM/Query/QueryException.php#L150) Are you fine with the inconsistency?

Ocramius: If you can add it to the patch, could also rename the variable there please?  👍

Ocramius:
@Isinlor I manually applied the patch after a rebase and some minor fixes: see doctrine/doctrine2@9b9128ae774ed20d8ccfd03d2fd2a540c140acdd

Thanks!

PR #6326, opened 2017-03-06
[Documentation] Add column check-option to documentation

SenseException:
Looking into the documentation, I saw that the check constraint in the `Column` mapping isn't in the documentation yet and added it.


Ocramius (approved):

Ocramius:
@SenseException thanks!

PR #5680, opened 2016-02-26
DDC-2780 - Fixed issue with IS NULL on join aliases

byonchev:
#3528 Added fallback to SingleValuedPathExpression

PR #6328, opened 2017-03-07
DDC-2780 - Fixed issue with IS NULL on join aliases

### LINKED ISSUE #3528: DDC-2780: IS [NOT] NULL conditions with JOINs
Jira issue originally created by user maryo:

<?php
When trying to apply a [NOT] NULL condition on LEFT JOINed association. It ends with an error.

"Cannot add having condition on a non result variable."

$queryBuilder
    ->leftJoin('r.players', 'players', 'WITH', 'players = :player')
    ->where('players IS NULL');

It was working 3 month ago. But then, this commit broke it.

https://github.com/doctrine/doctrine2/commit/d9c1782a4f6d46f66e9deb2c375830f9192d4482

Or in other words if this is not a bug. How to test if the collection does not contain let's say a user with ID 5 (AND XX OR YY) with DQL/queryBuilder API? IN SQL I would just simply LEFT JOIN it with ON condition and then i would test it for existence using IS NULL in WHERE condition. How to do this using DQL when the pasted example ends with an error?

lcobucci:
Apply fix from #5680 (with some minor changes)

Fixes #3528

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -3264,6 +3264,12 @@ class Parser
                     $this->semanticalError('Cannot add having condition on undefined result variable.');
                 }
 
+                // Validate SingleValuedPathExpression (ie.: "product")
+                if (isset($this->queryComponents[$lookaheadValue]['metadata'])) {
+                    $expr = $this->SingleValuedPathExpression();
+                    break;
+                }
+
                 // Validating ResultVariable
                 if ( ! isset($this->queryComponents[$lookaheadValue]['resultVariable'])) {
                     $this->semanticalError('Cannot add having condition on a non result variable.');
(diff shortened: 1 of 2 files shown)


guilhermeblanco (approved):

githoober:
Can this fix be backported to 2.5, we are really suffering from this bug, and yet we cannot upgrade to 2.6 due to the current PHP limitation in our organization.

githoober:
I don't understand why a bug that is introduced in one version of doctrine/orm is never fixed in the same version of the library. It's impossible to use it. For example, L2 was introduced in 2.5, but all bug fixes were applied in 2.6.

githoober:
https://github.com/doctrine/doctrine2/issues/7293

Ocramius:
> I don't understand why a bug that is introduced in one version of doctrine/orm is never fixed in the same version of the library. It's impossible to use it. 

We have very limited development resources here, based on free time and volunteering.

We chose to not involve ourselves in long term support plans, as we can't sustain them. Anything that isn't trivial or very important to backport won't be backported to a stable version, but only merged into next minor/major releases depending on whether BC breaks are required or involved.

PR #6354, opened 2017-03-22
Keep track of full qualified type name when dealing with collections

sandvige:
This pull request should be considered with this one :
https://github.com/doctrine/dbal/pull/2685


Ocramius (changes requested):
Requires test case.

sandvige:
@Ocramius How could I make this code to run against a specific DBAL version? This fix requires this feature: https://github.com/doctrine/dbal/pull/2685

Ocramius:
As per https://github.com/doctrine/dbal/pull/2685#discussion_r126886505 and related comments below it, this patch seems like the wrong approach. We need a better way of representing types that isn't string-based.

PR #6359, opened 2017-03-26
Allow to configure hints for the QueryBuilder

sstok:
Background: In the RollerworksSearch (Doctrine ORM ConditionGenerator) I use a Query-Hint to pass conversion information to DQL functions. But now to integrate RollerworksSearch with the Api-platform I need to work with the QueryBuilder which doesn't allow to set hints.

I copied the description from the AbstractQuery class. If possible can this be back ported to 2.5 also? I know you prefer to merge everything to master but without this I'm kinda stuck, and I don't know when 2.6 will be released 😅

### DIFF
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -118,6 +118,13 @@ class QueryBuilder
      */
     private $joinRootAliases = [];
 
+    /**
+     * The map of query hints.
+     *
+     * @var array
+     */
+    private $_hints = [];
+
      /**
      * Whether to use second level cache, if available.
      *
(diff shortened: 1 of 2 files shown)


sstok:
Travis failures seem unrelated 😃

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -145,6 +145,13 @@ class QueryBuilder
     protected $lifetime = 0;
 
     /**
+     * The map of query hints.
+     *
+     * @var array
+     */
+    private $_hints = [];
```
issei-m: It would be better to be at after `$joinRootAliases` (the last one of the bundle of private fields)

sstok: Done.

lcobucci:
But what's wrong with `$queryBuilder->getQuery()->setHint();`? I don't think it's really necessary to support this...

sstok:
The Query is generated after the building process, so I don't have any control over it.

https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/CollectionDataProvider.php#L65-L70

In the case an extension is able to provide the data, `return $queryBuilder->getQuery()->getResult();` is never executed and the Extension (which can be any) will generate the Query and use the result in a wrapped object (like a paginator).

I set the Query hint as follow:

```php
    public function applyToCollection(QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null)
    {
        $request = $this->requestStack->getCurrentRequest();

        /** @var SearchCondition $condition */
        if (!$request || null === $condition = $request->attributes->get('_api_search_condition')) {
            return;
        }

        if (!method_exists($queryBuilder, 'setHint')) {
            return;
        }

        $fieldSetName = $condition->getFieldSet();
        $resourceMetadata = $this->resourceMetadataFactory->create($resourceClass);

        if (null !== $attributes = $resourceMetadata->getAttribute('rollerworks_search')) {
            return;
        }

        if (empty($attributes['doctrine_orm']['*']) && empty($attributes['doctrine_orm'][$fieldSetName])) {
            return;
        }

        if (isset($resourceMetadata['accepted_fieldsets'])) {
            $condition->assertFieldSetName(...array_push($resourceMetadata['accepted_fieldsets'], $resourceClass));
        }

        $fieldSetName = $condition->getFieldSet();
        $setName = isset($attributes['doctrine_orm'][$fieldSetName]) ? $fieldSetName : '*';
        $configuration = $attributes['doctrine_orm'][$setName];

        $this->configureRelations($resourceClass, $configuration, $setName, $queryBuilder);

        $query = $queryBuilder->getEntityManager()->createQuery(); // Dummy to get the EntityManager
        $conditionGenerator = $this->ormFactory->createCachedConditionGenerator($this->ormFactory->createConditionGenerator($query, $condition));

        $this->configureMappings($resourceClass, $configuration, $setName, $conditionGenerator);

        $queryBuilder->andWhere($conditionGenerator->getWhereClause());
        $queryBuilder->setHint($conditionGenerator->getQueryHintName(), $conditionGenerator->getQueryHintValue()); // This cannot be done afterwards, 
    }
```

sstok:
Notice. I no longer need this feature as it's solved different in https://github.com/rollerworks/search/pull/286 

Feel free to close this 👍

Issue #6362, opened 2017-03-27
Undefined Property error in ObjectHydrator when using JOIN WITH


SirWaddles:
I have a pretty strange issue that I can't quite work out, I have a fix but it's affecting code from 8yrs ago which seems pretty stable.

I have a DQL query resembling
```
    SELECT base, fruits, apples, data
    FROM AppBundle\Base base 
    LEFT JOIN base.fruits fruits 
    LEFT JOIN AppBundle\Apple apples WITH fruits.id = apples.id 
    LEFT JOIN apples.data data
```

where a Base can have many Fruits of multiple types. The error I get when trying to run this query is just
`Undefined property on AppBundle\Base::$data`

For some reason, the ObjectHydrator is trying to add the values retrieved from the database to the wrong entity. After doing some debugging, I found something a little odd in [ObjectHydrator.php](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php#L360)
```php
    // Get a reference to the parent object to which the joined element belongs.
    if ($this->_rsm->isMixed && isset($this->rootAliases[$parentAlias])) {
        $first = reset($this->resultPointers);
        $parentObject = $first[key($first)];
```
It appears that it can only ever get a reference to the first object that there is a root alias to. After changing the second line to `$first = $this->resultPointers[$parentAlias];` in a debug build, my problem is solved and although I didn't run through the unit tests (will do if it's a PR, but it seems more like I'm doing something wrong here) my application elsewhere works fine.

I realise there is a limitation when it comes to downcasting in Doctrine (which is what I am trying to do) but it seems like this issue would have broader implications.

lcobucci:
This is a known limitation of the ORM when working with class table inheritance and might not be properly documented.

@SirWaddles would be nice if you could reproduce and implement a good solution for this on a PR.

lcobucci:
Fixed by #6367

PR #6367, opened 2017-03-28
Correct hydration of resultsets coming from DQL queries where the root of the selection is aliased and also the root of an inheritance

SirWaddles:
Resolves Issue #6362 

Added a test case for reproducability.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6362Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6362Test.php
@@ -1,0 +1,144 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+use Doctrine\ORM\Query;
+use Doctrine\ORM\Query\ResultSetMapping;
+use Doctrine\Tests\Mocks\HydratorMockStatement;
+
+final class GH6362Test extends OrmFunctionalTestCase
+{
+    protected function setUp()
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema(
+            [
+                $this->_em->getClassMetadata(GH6362Start::class),
+                $this->_em->getClassMetadata(GH6362Base::class),
+                $this->_em->getClassMetadata(GH6362Child::class),
+                $this->_em->getClassMetadata(GH6362Join::class),
+            ]
+        );
+    }
+
+    /**
+     * @group 6362
+     *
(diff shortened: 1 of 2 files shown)


SirWaddles:
Travis failures are unrelated, something to do with the Symfony Yaml component.

SirWaddles:
The main reason for test is to check that the Undefined Property error does not occur, and if you run the test without the fix, it will error out. It does however also make some assertions about the format of the result which may be questionable.

The resultset only has one row, yet the hydrated result has two. My assumption was that this is intentional as this change did not break any tests, but it does seem odd to me.

SirWaddles:
Might be worth running the tests again, only one job failed and it had just stalled before even getting to my code. I have no idea why my change would cause only PHP7.1/MariaDB to halt.

Ocramius:
@SirWaddles restarted

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6362Test.php
```diff
@@ -0,0 +1,142 @@
...
+                $this->_em->getClassMetadata(GH6362Start::class),
+                $this->_em->getClassMetadata(GH6362Base::class),
+                $this->_em->getClassMetadata(GH6362Child::class),
+                $this->_em->getClassMetadata(GH6362Join::class),
+            ]
+        );
+    }
+
+    /**
+     * SELECT a as base, b, c, d
```
lcobucci: `@group GH6362` missing

lcobucci (commented):
The rest LGTM, what do you think @Ocramius?

lcobucci:
@SirWaddles I've rebased your branch to sync with the latest updates and group the commits. Patch LGTM but I'd love to have @guilhermeblanco's and/or @Ocramius' opinion just to double check.

Thanks for the contribution 😉

lcobucci:
Quite easy to backport for `v2.5.7` too...

Ocramius (approved):

Ocramius:
Good to go! Will also backport :+1: 

Thanks @SirWaddles, pristine work.

Ocramius:
Hah, I backported to `2.5` via doctrine/doctrine2@04fc7a9a1ccd377c33ec753151b425fc3db22c27, but it will obviously fail due to lack of PHP 5.4 compliance =_=

Ocramius:
Should be corrected in doctrine/doctrine2@d2c805b07120e0d31cd4c27dbe39124ccec65092 (see https://travis-ci.org/doctrine/doctrine2/builds/227851219)

PR #6373, opened 2017-03-30
Typehint @return at get method for OneToMany target

fliespl:
Currently when using doctrine:generate:entityies for @ORM\OneToMany relation get method will generate as follows

```
    /**
     * @ORM\OneToMany(targetEntity="AppBundle\Entity\Action", mappedBy="player")
     */
    protected $actions;

    /**
     * Get actions
     *
     * @return \Doctrine\Common\Collections\Collection
     */
    public function getActions()
    {
        return $this->actions;
    }
```
Proposed change will add entity array hint as alternative for return as follows:

```

    /**
     * Get actions
     *
     * @return \Doctrine\Common\Collections\Collection|\AppBundle\Entity\Action[]
     */
    public function getActions()
    {
        return $this->actions;
    }
```

### DIFF
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -1380,6 +1380,14 @@ public function __construct(<params>)
             $variableType   =  '\\' . ltrim($variableType, '\\');
             $methodTypeHint =  '\\' . $typeHint . ' ';
         }
+        
+        if($metadata->hasAssociation($fieldName)) {
+            $assoc = $metadata->getAssociationMapping($fieldName);
+            if($assoc['type'] === ClassMetadataInfo::ONE_TO_MANY) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -1380,6 +1380,14 @@ protected function generateEntityStubMethod(ClassMetadataInfo $metadata, $type,
             $variableType   =  '\\' . ltrim($variableType, '\\');
             $methodTypeHint =  '\\' . $typeHint . ' ';
         }
+        
+        if($metadata->hasAssociation($fieldName)) {
```
Ocramius: To be moved to private method

Ocramius (changes requested):
Change looks good, but tests are needed

Ocramius:
Closing as `incomplete` due to missing tests

PR #6374, opened 2017-03-31
Add failing test case DDC726 #5238

soyuka:
In order to fix this, and as I'm not really familiar with the Query AST/Parser code, I have some questions:

- Is there already an ASTExpression we can use to match `(foo, bar)`?
- The path for Where being ConditionalExpression => ConditionalTerm => ConditionalFactor => ConditionalPrimary => ConditionalExpression, I think that the fix should be [here](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Query/Parser.php#L2497). Or should it be upper in the chain?

@guilhermeblanco (you were assigned on #5238), did you already look into this? Could you share your thoughts on the subject?

Thanks.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC726Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC726Test.php
@@ -1,0 +1,105 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+class DDC726Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->_em->getConnection()->getConfiguration()->setSQLLogger(new \Doctrine\DBAL\Logging\EchoSQLLogger);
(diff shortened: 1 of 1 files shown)


Ocramius:
@soyuka is this supposed to be closed? What's the resolution?

soyuka:
Whoops, my bad! I'll let this open if you want to re-use my failing test but I don't think that I'll be able to fix this on my own, codebase is hard to assimilate. Also I didn't wanted to pollute the repository!

A temporary resolution for this is to have multiple subqueries, which is working really fine. Anyway this could be a nice feature!

Ocramius:
Possibly provided by https://github.com/doctrine/doctrine2/pull/6476 or https://github.com/doctrine/doctrine2/issues/6520

lcobucci:
@soyuka @Ocramius I've rebased https://github.com/doctrine/doctrine2/pull/6476 and ran this test and it still fails.

Reasons:

- `WHERE (b.id, c.id) EXPRESSION` syntax is not supported
- IDENTITY(d) is not returning all the identifiers

PR #6376, opened 2017-03-31
Pinning dev dependency symfony/yaml to working version

Sam-Burns:
You've been having some [weird build breaks](https://travis-ci.org/doctrine/doctrine2/jobs/215427291) in [pull requests](https://github.com/doctrine/doctrine2/pull/6359) recently.  This is because the latest version of `symfony/yaml` that you're getting with `"minimum-stability": "dev"` is `3.3.x-dev`, which isn't working for you.

This PR makes your build run with a stable version of `symfony/yaml`, which we know works for your test suite.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "symfony/console": "~2.5|~3.0"
     },
     "require-dev": {
-        "symfony/yaml": "~2.3|~3.0",
+        "symfony/yaml": "^3.2@stable",
         "phpunit/phpunit": "^5.4"
     },
     "suggest": {


Ocramius:
Removing Yaml v2 support seems unrelated to the patch though.

Sam-Burns:
@Ocramius as this is the `require-dev` section of the `composer.json` file, downstream projects aren't affected in terms of what versions of `symfony/yaml` they can use.  This simply affects which version is used in development and builds, when running Doctrine's own tests.

lcobucci:
I've already notified symfony devs (symfony/symfony#22193) and they're working to fix it (symfony/symfony#22216), with that said I'd say that we should just leave the ORM the way it is.

Ocramius:
@Sam-Burns brain-fart on my side, thanks.

lcobucci:
@Sam-Burns you do have a point regarding stability of the tests

Sam-Burns:
@lcobucci just thought it might help other contributors, what with all the PRs being red for no reason.  Very glad to hear someone has spoken to the good people at Symfony already though - good job.

Ocramius:
@Sam-Burns if we didn't keep dev dependencies, this sort of issue would never be discovered in first place. Restricting to stable versions will only cause more issues downstream (consumers)

Ocramius:
That said, closing as `won't fix` - it will stay red until fixed at the root of the problem.

lcobucci:
@Sam-Burns @Ocramius just to let you know that the issue was fixed on the component and I queued the tests for the last PRs we had 😉

PR #6379, opened 2017-04-01
EntityManager: Implement singleton pattern, use global injection (minor BC break)

Majkl578:
As per [the article](http://www.doctrine-project.org/2017/04/01/announcing-doctrine-4.html), since Doctrine is moving to modern static interface, let's start immediately with EntityManager.

This PR introduces:
* Singleton pattern - Why should anyone track the instance? Let EntityManager do it itself!
* Global injection - Parameters are boring, this is easier.

Targets 2.5 as this should be introduced ASAP - preferably in the next bugfix version.

TODO:
* [ ] Fix tests.


Ocramius:
Closing, since we missed the timeframe for merging this stuff. These features are now to be considered shit.
Feel free to reopen on April 1st next year 😂

PR #6384, opened 2017-04-06
Invalid discrimination column value when discriminator column has custom type.

daniel1302:
I have got discriminator config like below:
```xml
<discriminator-column name="type" type="dc_job_type_enum"></discriminator-column>
    <discriminator-map>
      <discriminator-mapping value="EMAIL" class="AppBundle\Entity\JobObjectEmail"/>
    </discriminator-map>
```

And EMAIL is PHP value and its mapping to 1 in database. 
Script throwing below error before changes:
```
HydrationException: The discriminator value "1" is invalid. It must be one of "EMAIL".
```

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -99,13 +99,21 @@ class SimpleObjectHydrator extends AbstractHydrator
             }
 
             $discrMap = $this->class->discriminatorMap;
+            $discrType = $discriminatorValue = $this->class->discriminatorColumn['type'];
 
-            if ( ! isset($discrMap[$sqlResult[$discrColumnName]])) {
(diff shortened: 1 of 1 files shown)


Ocramius:
Requires a test case

lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
```diff
@@ -99,13 +99,21 @@ protected function hydrateRowData(array $sqlResult, array &$result)
             }
 
             $discrMap = $this->class->discriminatorMap;
+            $discrType = $discriminatorValue = $this->class->discriminatorColumn['type'];
 
-            if ( ! isset($discrMap[$sqlResult[$discrColumnName]])) {
+            if (Type::hasType($discrType)) {
```
Ocramius: This should always be `true`: we actually want an exception to be thrown if it is not.

Ocramius (changes requested):

PR #6389, opened 2017-04-09
XSD definition improvements

mikeSimonson:


doctrine-mapping.xsd
```diff
@@ -200,6 +200,13 @@
     <xs:anyAttribute namespace="##other"/>
   </xs:complexType>
 
+  <xs:simpleType name="TABLENAME" id="TABLENAME">
+    <xs:restriction base="xs:token">
+      <xs:pattern value="[-._:A-Za-z0-9\`]" id="TABLENAME.pattern">
```
Ocramius: This is way too restrictive

doctrine-mapping.xsd
```diff
@@ -412,9 +419,16 @@
     <xs:sequence>
       <xs:any minOccurs="0" maxOccurs="unbounded" namespace="##other"/>
     </xs:sequence>
-    <xs:attribute name="class" type="xs:string" use="required" />
+    <xs:attribute name="class" type="orm:CLASSPATH" use="required" />
   </xs:complexType>
 
+  <xs:simpleType name="CLASSPATH" id="CLASSPATH">
+    <xs:restriction base="xs:token">
+      <xs:pattern value="([a-zA-Z_$][a-zA-Z\d_$\\]*\.)*[a-zA-Z_$][a-zA-Z\d_$\\]*" id="CLASSPATH.pattern">
```
Ocramius: The regex for class names is something like `/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+$/` - not just ascii

Ocramius (changes requested):

doctrine-mapping.xsd
```diff
@@ -200,6 +200,13 @@
     <xs:anyAttribute namespace="##other"/>
   </xs:complexType>
 
+  <xs:simpleType name="TABLENAME" id="TABLENAME">
```
Ocramius: What's up with the uppercase naming?

doctrine-mapping.xsd
```diff
@@ -412,9 +419,16 @@
     <xs:sequence>
       <xs:any minOccurs="0" maxOccurs="unbounded" namespace="##other"/>
     </xs:sequence>
-    <xs:attribute name="class" type="xs:string" use="required" />
+    <xs:attribute name="class" type="orm:CLASSPATH" use="required" />
   </xs:complexType>
 
+  <xs:simpleType name="CLASSPATH" id="CLASSPATH">
```
Ocramius: What's up with the uppercase naming?

Ocramius (changes requested):

doctrine-mapping.xsd
mikeSimonson: more something like that ```/[a-zA-Z_\x01-\xff\`]+/``` ?
Based on https://dev.mysql.com/doc/refman/5.7/en/identifiers.html

mikeSimonson:
@Ocramius updated. Is there an existing place to add tests or should I create it ?

Ocramius:
> Is there an existing place to add tests or should I create it ?

Good question. Wondering if there's an XSD validation tool anywhere (not an online tool)

mikeSimonson:
@Ocramius I was thinking using something along the line of http://php.net/manual/en/domdocument.schemavalidate.php

mikeSimonson:
and it seems that this mechanism is already used.

Ocramius:
If you want to create some valid/invalid documents, you can give it a shot.
I was more wondering about the XSD itself :-)

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

doctrine-mapping.xsd
```diff
@@ -412,9 +419,16 @@
     <xs:sequence>
       <xs:any minOccurs="0" maxOccurs="unbounded" namespace="##other"/>
     </xs:sequence>
-    <xs:attribute name="class" type="xs:string" use="required" />
+    <xs:attribute name="class" type="orm:classpath" use="required" />
```
mikeSimonson: fqcn is probably better here no ?

lcobucci:
> If you want to create some valid/invalid documents, you can give it a shot.
I was more wondering about the XSD itself :-)

@Ocramius AFAIK we would have issues when trying to validate a document with an invalid XSD, so I'd be fine once the test suite is passing

Ocramius (approved):
LGTM! :+1:

Ocramius:
@mikeSimonson this started failing in `master`, but I really have no clue why: all XML files in that specific test seem to be fine :-\

mikeSimonson:
I also have no clue why and can't reproduce it locally.

Ocramius:
Same here - can't reproduce, neither on PHP 7.0 nor on 7.1 :-\

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

guilliamxavier:
_(Sorry if this seems not very constructive, but if it may help a little:)_

Is `"[a-zA-Z_u01-uff.]+"` (and `"[a-zA-Z_u01-uff][a-zA-Z0-9_u01-uff]+"`) really the intended pattern?
- About the `u01-uff` part, isn't it missing backslashes like `\u01-\uff`, or probably more like `\u0001-\u00ff` (or `\x01-\xff`)? (As it is, I interpret it as "`u` or `0` or a char in the range `1-u` (which feels like [a pretty weird range](http://www.torsten-horn.de/techdocs/ascii.htm)) or `f` or `f` (bis)"...)
    - But wait, `[a-zA-Z_\x01-\xff.]` would be in fact the same as `[\x01-\xff]` (which already includes `a-z`, `A-Z`, `_` and `.`), which is probably not the intent?
    - Moreover, according to http://www.regular-expressions.info/xml.html, "XML regular expressions don't have any tokens like \xFF or \uFFFF to match particular (non-printable) characters", and e.g. http://www.utilities-online.info/xsdvalidation/ indeed flags it as invalid XSD schema. But it seems you could use [Unicode blocks](http://www.regular-expressions.info/unicode.html) like `\p{IsBasicLatin}` or `\p{IsLatin-1Supplement}`.

lcobucci:
@guilliamxavier @mikeSimonson mistery solved on #6515, ancient libxml versions (but now we're facing some other issues with MySQL/MariaDB on Trusty env 😜 )

guilliamxavier:
@lcobucci After looking deeper into the test failures, I have indeed realized that they were actually not related to pattern validation...
But, independently of the tests, I still wonder: is "`[u01-uff]`" actually a desired character class?

lcobucci:
@guilliamxavier tbh I'm not sure, I'll leave that question to @mikeSimonson

mikeSimonson:
@guilliamxavier The charachter class is based on what mysql accept as identifier (https://dev.mysql.com/doc/refman/5.7/en/identifiers.html). If you have a better proposition that also take into account other dbms fire away or better pull request away.

guilliamxavier:
@mikeSimonson Thanks for the feedback. My first and main concern is actually about the _syntax_ of the pattern:

`[a-zA-Z_u01-uff.]+`

which, as shown below, doesn't accept e.g. "`` `my-foo` ``".

### XML pattern syntax for Unicode characters

As I said in my first (i.e. second-to-last) comment, the bare (non-escaped) `u01` and `uff` just look _wrong_ to me, and some testing (script further below) actually confirmed that the previous pattern means this:

`([a-z]|[A-Z]|_|u|0|[1-u]|f|f|\.)+`

(where, notably, `f` is duplicated and `[1-u]` is an unusual ASCII range corresponding to the following characters: ``123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstu``) which is very probably _not_ what was intended. For example, it **doesn't accept** the legit quoted table name `` `my-foo` `` in the XML line

```xml
    <entity name="Foo" table="`my-foo`">
```

and gives errors:

    [ERROR 1839] Element '{http://doctrine-project.org/schemas/orm/doctrine-mapping}entity', attribute 'table': [facet 'pattern'] The value '`my-foo`' is not accepted by the pattern '[a-zA-Z_u01-uff.]+'.
    [ERROR 1824] Element '{http://doctrine-project.org/schemas/orm/doctrine-mapping}entity', attribute 'table': '`my-foo`' is not a valid value of the atomic type '{http://doctrine-project.org/schemas/orm/doctrine-mapping}tablename'.

(because `-` isn't included in the character class) while it accepts the disputable unquoted table name `my;foo` in

```xml
    <entity name="Foo" table="my;foo">
```

(because `;` is).

So I tried different _backslash-escaped_ forms, e.g. for `u01` which I assume to intend Unicode character U+0001, I tested: `\x01`, `\x{1}`, `\u0001`, `\u{1}`, both lower- and upper-case, with and without braces, varying the amount of leading zeros (and same for `uff` for U+00FF), but each time got the following kind of errors:

    Warning: DOMDocument::schemaValidate(): Invalid Schema in .../test.php on line 10

    [FATAL 1450] failed to compile: Wrong escape sequence, misuse of character '\'
    [FATAL 1450] failed to compile: xmlFAParseCharClass: ']' expected
    [FATAL 1450] failed to compile: xmlFAParseRegExp: extra characters
    [ERROR 1756] Element '{http://www.w3.org/2001/XMLSchema}pattern': The value '[a-zA-Z_\x01-\xff.]+' of the facet 'pattern' is not a valid regular expression.

Note that those errors (and PHP warning) occur when libxml is parsing the XSD Schema, before even trying to validate any XML document.

The correct syntax in "XML Regex" is actually to use _XML entities_ e.g. `&#x01;` (or `&#x0001;` or `&#x1;`). But now I get:

    Warning: DOMDocument::schemaValidate(): Invalid Schema in .../test.php on line 10

    [FATAL 9] xmlParseCharRef: invalid xmlChar value 1
    [ERROR 3067] Failed to parse the XML resource '.../doctrine-mapping.xsd'.

because U+0001 is not a valid character in XML 1.0 (note: but it is in XML 1.1). It seems that the only valid chars before U+0020 (space) are U+0009 (`\t`), U+000A (`\n`) and U+000D (`\r`). So if I change the pattern to

`[a-zA-Z_&#x09;-&#xff;.]+`

then it "works", i.e. no more XSD errors, and the quoted table name `` `my-foo` `` is accepted (which was the purpose of your commit f537eb2915edda7820b53ff2f94bf395ed489ff7 after all).

### Actual meaning

But you must realize that `[&#x09;-&#xff;]` already includes all the chars in `[a-zA-Z_.]`, so the pattern boils down to

`[&#x09;-&#xff;]+`

### [My]SQL identifier and FQCN

**[My]SQL identifier**

Now, according to your link, the correct pattern for a MySQL identifier would be more like

``[0-9a-zA-Z$_&#x0080;-&#xFFFF;]+|`[&#x0001;-&#x007F;&#x0080;-&#xFFFF;]+` ``

(NB: `FFFF`, not `00FF` which was meant by `ff` (ignoring the case)). But again XML doesn't like 0001 and also not FFFF neither FFFE, plus we can merge the ranges ...-007F and 0080-..., so

``[0-9a-zA-Z$_&#x0080;-&#xFFFD;]+|`[&#x0009;-&#xFFFD;]+` ``

but it is still not complete (emphasis mine):

> - Identifiers may begin with a digit but _unless quoted_ may not consist solely of digits.
> - Database, table, and column names cannot end with space characters.
> - Identifier quote characters can be included within an identifier if you quote the identifier. (...) then you need to double the character. The following statement creates a table named ``a`b`` (...): ```CREATE TABLE `a``b` (...);```

(and not even accounting that with `ANSI_QUOTES` you can also use `"` as the quote character).

And as you said, there are other DBMS, with possibly different accepted chars and restrictions... At this point, maybe one could as well consider going with `<xs:attribute name="table" type="xs:string" />` (as was done in commit fafb8166f40585e0c186fdb06f228d4a83e4b1ce)?

**FQCN**

The same concerns apply to your second commit b43325760ddddf4983768c7eae802a3644837c5f, and the correct character class for a FQCN will of course be different than for a MySQL identifier (e.g. it must accept `\`, and should not accept `` ` `` or `"`...)

---

Finally here is my `test.php` script (inspired from https://github.com/doctrine/doctrine2/blob/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php#L151 + PHP manual):

```php
<?php

$xmlMappingFile = '.../test.xml';
$xsdSchemaFile  = '.../doctrine-mapping.xsd';
$dom            = new \DOMDocument('UTF-8');

$dom->load($xmlMappingFile);

libxml_use_internal_errors(true); // to use libxml_get_errors() instead of getting warnings
if ($dom->schemaValidate($xsdSchemaFile)) {
    echo "OK.\n";
} else {
    echo "FAILURE!\n";
    /** @var \LibXMLError[] $errors */
    $errors     = libxml_get_errors();
    $errorCount = count($errors);
    echo "$errorCount error(s):\n";
    $errorLevelNames = [
        LIBXML_ERR_NONE     => 'NONE',
        LIBXML_ERR_WARNING  => 'WARNING',
        LIBXML_ERR_ERROR    => 'ERROR',
        LIBXML_ERR_FATAL    => 'FATAL',
    ];
    $i = 0;
    foreach ($errors as $error) {
        $i++;
        $errorLevelName = $errorLevelNames[$error->level] ?? 'UNKNOWN';
        $errorMessage   = trim($error->message);
        echo "($i/$errorCount) at $error->file:$error->line:$error->column:\n";
        echo "[$errorLevelName $error->code] $errorMessage\n";
        echo "---\n";
    }
    libxml_clear_errors();
}
```

And some minimalist XML test file examples (change `xxx` to the values to test):

* for `table` (`tablename`):

```xml

<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                          http://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd">

    <entity name="Foo" table="xxx">
        <id name="id" />
    </entity>

</doctrine-mapping>
```

* for `class` (`fqcn`):

```xml

<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                          http://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd">

    <entity name="Foo">
        <id name="id">
            <generator strategy="CUSTOM" />
            <custom-id-generator class="xxx" />
        </id>
    </entity>

</doctrine-mapping>
```

PR #1441, opened 2015-06-29
`INSTANCE OF` now behaves correctly by including all possible child classes in `IN()` filtering conditions

taueres:
There was a bug in the "INSTANCE OF" operator as described in
https://groups.google.com/forum/#!topic/doctrine-user/B8raq8CNMgg

"INSTANCE OF" was not taking into account subclasses.
It was merely translating the class to its discriminator.
This is not correct since the class can have subtypes and those
are, indeed, still instance of the superclass.
Also, classes may not have a discriminator (e.g. abstract classes).

This commit also provides useful tests to avoid regression.

PR #6392, opened 2017-04-12
Correct DQL `INSTANCE OF` to filter all possible child classes

### LINKED ISSUE #4646: DDC-3800: [GH-1441] [QUERY] "INSTANCE OF" now behaves correctly with subclasses
Jira issue originally created by user @doctrinebot:

This issue is created automatically through a Github pull request on behalf of taueres:

Url: https://github.com/doctrine/doctrine2/pull/1441

Message:

There was a bug in the "INSTANCE OF" operator as described in
https://groups.google.com/forum/#!topic/doctrine-user/B8raq8CNMgg

"INSTANCE OF" was not taking into account subclasses.
It was merely translating the class to its discriminator.
This is not correct since the class can have subtypes and those
are, indeed, still instance of the superclass.
Also, classes may not have a discriminator (e.g. abstract classes).

This commit also provide useful tests to avoid regression.

Jean85:
This is the rebased version of #1441, which fixes #4646.
Here is the original text:

> There was a bug in the "INSTANCE OF" operator as described in
> https://groups.google.com/forum/#!topic/doctrine-user/B8raq8CNMgg
> 
> "INSTANCE OF" was not taking into account subclasses.
> It was merely translating the class to its discriminator.
> This is not correct since the class can have subtypes and those
> are, indeed, still instance of the superclass.
> Also, classes may not have a discriminator (e.g. abstract classes).
> 
> This commit also provides useful tests to avoid regression.

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -2038,8 +2038,10 @@ class SqlWalker implements TreeWalker
 
         $sql .= $class->discriminatorColumn['name'] . ($instanceOfExpr->not ? ' NOT IN ' : ' IN ');
 
-        $sqlParameterList = [];
+        $knownSubclasses = array_flip($discrClass->subClasses);
 
+        $sqlParameterList = [];
+        $discriminators = [];
         foreach ($instanceOfExpr->value as $parameter) {
             if ($parameter instanceof AST\InputParameter) {
                 $this->rsm->addMetadataParameterMapping($parameter->name, 'discriminatorValue');
@@ -2049,21 +2051,31 @@ class SqlWalker implements TreeWalker
                 continue;
             }
 
-            // Get name from ClassMetadata to resolve aliases.
-            $entityClassName    = $this->em->getClassMetadata($parameter)->name;
-            $discriminatorValue = $class->discriminatorValue;
+            // Trim first backslash
+            $parameter = ltrim($parameter, '\\');
 
-            if ($entityClassName !== $class->name) {
-                $discrMap = array_flip($class->discriminatorMap);
-
-                if ( ! isset($discrMap[$entityClassName])) {
-                    throw QueryException::instanceOfUnrelatedClass($entityClassName, $class->rootEntityName);
-                }
-
-                $discriminatorValue = $discrMap[$entityClassName];
+            // Check parameter is really in the hierarchy
+            if ($parameter !== $discrClass->name && ! array_key_exists($parameter, $knownSubclasses)) {
+                throw QueryException::instanceOfUnrelatedClass($parameter, $discrClass->name);
             }
 
-            $sqlParameterList[] = $this->conn->quote($discriminatorValue);
+            // Include discriminators for parameter class and its subclass
+            $metadata = $this->em->getClassMetadata($parameter);
+            $hierarchyClasses = $metadata->subClasses;
+            $hierarchyClasses[] = $metadata->name;
+
+            foreach ($hierarchyClasses as $class) {
+                $currentMetadata = $this->em->getClassMetadata($class);
+                $currentDiscriminator = $currentMetadata->discriminatorValue;
+
+                if (is_string($currentDiscriminator) && ! array_key_exists($currentDiscriminator, $discriminators)) {
+                    $discriminators[$currentDiscriminator] = true;
+                }
+            }
+        }
+
+        foreach (array_keys($discriminators) as $dis) {
+            $sqlParameterList[] = $this->conn->quote($dis);
         }
 
         $sql .= '(' . implode(', ', $sqlParameterList) . ')';
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -2280,7 +2280,6 @@ class SqlWalker implements TreeWalker
      */
     private function getChildDiscriminatorsFromClassMetadata(ClassMetadataInfo $discrClass, AST\InstanceOfExpression $instanceOfExpr)
     {
-        $knownSubclasses = array_flip($discrClass->subClasses);
         $sqlParameterList = [];
         $discriminators = [];
         foreach ($instanceOfExpr->value as $parameter) {
@@ -2292,15 +2291,13 @@ class SqlWalker implements TreeWalker
                 continue;
             }
 
-            // Trim first backslash
-            $parameter = ltrim($parameter, '\\');
+            $metadata = $this->em->getClassMetadata($parameter);
 
-            if ($parameter !== $discrClass->name && ! array_key_exists($parameter, $knownSubclasses)) {
+            if ($metadata->getReflectionClass()->isSubclassOf($discrClass->name)) {
                 throw QueryException::instanceOfUnrelatedClass($parameter, $discrClass->name);
             }
 
             // Include discriminators for parameter class and its subclass
-            $metadata = $this->em->getClassMetadata($parameter);
             $hierarchyClasses = $metadata->subClasses;
             $hierarchyClasses[] = $metadata->name;
 
@@ -2308,15 +2305,13 @@ class SqlWalker implements TreeWalker
                 $currentMetadata = $this->em->getClassMetadata($class);
                 $currentDiscriminator = $currentMetadata->discriminatorValue;
 
-                if (is_string($currentDiscriminator) && ! array_key_exists($currentDiscriminator, $discriminators)) {
-                    $discriminators[$currentDiscriminator] = true;
+                if (null !== $currentDiscriminator && ! array_key_exists($currentDiscriminator, $discriminators)) {
+                    $discriminators[$currentDiscriminator] = null;
                 }
             }
         }
 
-        foreach (array_keys($discriminators) as $dis) {
-            $sqlParameterList[] = $this->conn->quote($dis);
-        }
+        $sqlParameterList = array_map([$this->conn, 'quote'], array_keys($discriminators));
 
         return '(' . implode(', ', $sqlParameterList) . ')';
     }
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -25,6 +25,7 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\OptimisticLockException;
 use Doctrine\ORM\Query;
+use Doctrine\ORM\Utility\HierarchyDiscriminatorResolver;
 use Doctrine\ORM\Utility\PersisterHelper;
 
 /**
@@ -2292,7 +2293,7 @@ class SqlWalker implements TreeWalker
                 throw QueryException::instanceOfUnrelatedClass($parameter, $rootClass->name);
             }
 
-            $discriminators = $discriminators + $this->getAllDiscriminators($metadata);
+            $discriminators += HierarchyDiscriminatorResolver::resolveDiscriminatorsForClass($metadata, $this->em);
         }
 
         foreach (array_keys($discriminators) as $dis) {
@@ -2300,24 +2301,5 @@ class SqlWalker implements TreeWalker
         }
 
         return '(' . implode(', ', $sqlParameterList) . ')';
-    }
-
-    private function getAllDiscriminators(ClassMetadata $classMetadata)
-    {
-        // FIXME: this code is identical to Query->getAllDiscriminators()
-        $hierarchyClasses = $classMetadata->subClasses;
-        $hierarchyClasses[] = $classMetadata->name;
-
-        $discriminators = [];
-        foreach ($hierarchyClasses as $class) {
-            $currentMetadata = $this->em->getClassMetadata($class);
-            $currentDiscriminator = $currentMetadata->discriminatorValue;
-
-            if (null !== $currentDiscriminator) {
-                $discriminators[$currentDiscriminator] = null;
-            }
-        }
-
-        return $discriminators;
     }
 }
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -2290,7 +2290,7 @@ class SqlWalker implements TreeWalker
 
             $metadata = $this->em->getClassMetadata($parameter);
 
-            if (! $metadata->getReflectionClass()->isSubclassOf($discrClass->name)) {
+            if ($metadata->getName() !== $discrClass->name && ! $metadata->getReflectionClass()->isSubclassOf($discrClass->name)) {
                 throw QueryException::instanceOfUnrelatedClass($parameter, $discrClass->name);
             }
 
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -2276,7 +2276,7 @@ class SqlWalker implements TreeWalker
      * @return string The list in parentheses of valid child discriminators from the given class
      * @throws QueryException
      */
-    private function getChildDiscriminatorsFromClassMetadata(ClassMetadataInfo $rootClass, AST\InstanceOfExpression $instanceOfExpr)
+    private function getChildDiscriminatorsFromClassMetadata(ClassMetadataInfo $rootClass, AST\InstanceOfExpression $instanceOfExpr): string
     {
         $sqlParameterList = [];
         $discriminators = [];
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -2277,7 +2277,6 @@ class SqlWalker implements TreeWalker
      */
     private function getChildDiscriminatorsFromClassMetadata(ClassMetadataInfo $discrClass, AST\InstanceOfExpression $instanceOfExpr)
     {
-        $knownSubclasses = array_flip($discrClass->subClasses);
         $sqlParameterList = [];
         $discriminators = [];
         foreach ($instanceOfExpr->value as $parameter) {
@@ -2289,15 +2288,13 @@ class SqlWalker implements TreeWalker
                 continue;
             }
 
-            // Trim first backslash
-            $parameter = ltrim($parameter, '\\');
+            $metadata = $this->em->getClassMetadata($parameter);
 
-            if ($parameter !== $discrClass->name && ! array_key_exists($parameter, $knownSubclasses)) {
+            if ($metadata->getReflectionClass()->isSubclassOf($discrClass->name)) {
                 throw QueryException::instanceOfUnrelatedClass($parameter, $discrClass->name);
             }
 
             // Include discriminators for parameter class and its subclass
-            $metadata = $this->em->getClassMetadata($parameter);
             $hierarchyClasses = $metadata->subClasses;
             $hierarchyClasses[] = $metadata->name;
 
@@ -2305,15 +2302,13 @@ class SqlWalker implements TreeWalker
                 $currentMetadata = $this->em->getClassMetadata($class);
                 $currentDiscriminator = $currentMetadata->discriminatorValue;
 
-                if (is_string($currentDiscriminator) && ! array_key_exists($currentDiscriminator, $discriminators)) {
-                    $discriminators[$currentDiscriminator] = true;
+                if (null !== $currentDiscriminator && ! array_key_exists($currentDiscriminator, $discriminators)) {
+                    $discriminators[$currentDiscriminator] = null;
                 }
             }
         }
 
-        foreach (array_keys($discriminators) as $dis) {
-            $sqlParameterList[] = $this->conn->quote($dis);
-        }
+        $sqlParameterList = array_map([$this->conn, 'quote'], array_keys($discriminators));
 
         return '(' . implode(', ', $sqlParameterList) . ')';
     }
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -37,7 +37,6 @@ use Doctrine\ORM\Utility\PersisterHelper;
  * @author Alexander <iam.asm89@gmail.com>
  * @author Fabio B. Silva <fabio.bat.silva@gmail.com>
  * @since  2.0
- * @todo Rename: SQLWalker
  */
 class SqlWalker implements TreeWalker
 {
@@ -2017,6 +2016,7 @@ class SqlWalker implements TreeWalker
 
     /**
      * {@inheritdoc}
+     * @throws \Doctrine\ORM\Query\QueryException
      */
     public function walkInstanceOfExpression($instanceOfExpr)
     {
@@ -2034,48 +2034,7 @@ class SqlWalker implements TreeWalker
         }
 
         $sql .= $class->discriminatorColumn['name'] . ($instanceOfExpr->not ? ' NOT IN ' : ' IN ');
-
-        $knownSubclasses = array_flip($discrClass->subClasses);
-
-        $sqlParameterList = [];
-        $discriminators = [];
-        foreach ($instanceOfExpr->value as $parameter) {
-            if ($parameter instanceof AST\InputParameter) {
-                $this->rsm->addMetadataParameterMapping($parameter->name, 'discriminatorValue');
-
-                $sqlParameterList[] = $this->walkInputParameter($parameter);
-
-                continue;
-            }
-
-            // Trim first backslash
-            $parameter = ltrim($parameter, '\\');
-
-            // Check parameter is really in the hierarchy
-            if ($parameter !== $discrClass->name && ! array_key_exists($parameter, $knownSubclasses)) {
-                throw QueryException::instanceOfUnrelatedClass($parameter, $discrClass->name);
-            }
-
-            // Include discriminators for parameter class and its subclass
-            $metadata = $this->em->getClassMetadata($parameter);
-            $hierarchyClasses = $metadata->subClasses;
-            $hierarchyClasses[] = $metadata->name;
-
-            foreach ($hierarchyClasses as $class) {
-                $currentMetadata = $this->em->getClassMetadata($class);
-                $currentDiscriminator = $currentMetadata->discriminatorValue;
-
-                if (is_string($currentDiscriminator) && ! array_key_exists($currentDiscriminator, $discriminators)) {
-                    $discriminators[$currentDiscriminator] = true;
-                }
-            }
-        }
-
-        foreach (array_keys($discriminators) as $dis) {
-            $sqlParameterList[] = $this->conn->quote($dis);
-        }
-
-        $sql .= '(' . implode(', ', $sqlParameterList) . ')';
+        $sql .= $this->getChildDiscriminatorsFromClassMetadata($discrClass, $instanceOfExpr);
 
         return $sql;
     }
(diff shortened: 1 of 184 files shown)


tests/Doctrine/Tests/ORM/Functional/InstanceOfTest.php
```diff
@@ -0,0 +1,119 @@
...
+
+    use Doctrine\Tests\OrmFunctionalTestCase;
+
+    class InstanceOfTest extends OrmFunctionalTestCase
+    {
+        protected function setUp()
+        {
+            parent::setUp();
+
+            $this->_schemaTool->createSchema(array(
```
phansys: Please, use short array syntax instead.

tests/Doctrine/Tests/ORM/Functional/InstanceOfAbstractTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+            $dql = 'SELECT p FROM Doctrine\Tests\ORM\Functional\InstanceOfAbstractTest\Person p
+                    WHERE p INSTANCE OF Doctrine\Tests\ORM\Functional\InstanceOfAbstractTest\Person';
+            $query = $this->_em->createQuery($dql);
+            $result = $query->getResult();
+
+            $this->assertCount(1, $result);
+
+            foreach ($result as $r) {
+                $this->assertInstanceOf('Doctrine\Tests\ORM\Functional\InstanceOfAbstractTest\Person', $r);
```
phansys: Please, use `::class` constant instead of hardcoding the QCN here.

tests/Doctrine/Tests/ORM/Functional/InstanceOfAbstractTest.php
```diff
@@ -0,0 +1,113 @@
...
+                $this->assertInstanceOf('Doctrine\Tests\ORM\Functional\InstanceOfAbstractTest\Employee', $r);
+                $this->assertEquals('bar', $r->getName());
```
phansys: Please, use `assertSame()` instead.

phansys (changes requested):
Status: Needs work.

Jean85:
All done, thanks for the review @phansys

lcobucci (changes requested):
Can you please extract some methods from `SqlWalker#walkInstanceOfExpression()` it was already messy and now it's a bit harder to understand things properly.

Also could you try to relate the tests, and entities that each test use, to the ticket (as we have on `Functional\Ticket` classes) just to keep the "standard" we're having...

Jean85:
Thanks for the review @lcobucci!
I hope that I addressed al your concerns with my latest commits.

Jean85:
A lot of tests are failing, probably for an unrelated issue... should I rebase? Or merge master?

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -2300,4 +2271,53 @@ public function walkResultVariable($resultVariable)
...
+            // Include discriminators for parameter class and its subclass
+            $metadata = $this->em->getClassMetadata($parameter);
+            $hierarchyClasses = $metadata->subClasses;
+            $hierarchyClasses[] = $metadata->name;
+
+            foreach ($hierarchyClasses as $class) {
+                $currentMetadata = $this->em->getClassMetadata($class);
+                $currentDiscriminator = $currentMetadata->discriminatorValue;
+
+                if (is_string($currentDiscriminator) && ! array_key_exists($currentDiscriminator, $discriminators)) {
```
Ocramius: A discriminator may be also an integer or similar - why do you want this to only work with `string`?

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -2300,4 +2271,53 @@ public function walkResultVariable($resultVariable)
...
+            if ($parameter instanceof AST\InputParameter) {
+                $this->rsm->addMetadataParameterMapping($parameter->name, 'discriminatorValue');
+
+                $sqlParameterList[] = $this->walkInputParameter($parameter);
+
+                continue;
+            }
+
+            // Trim first backslash
+            $parameter = ltrim($parameter, '\\');
```
Ocramius: You should do `$this->em->getClassMetadata($parameter);` here - this already provides the `ltrim()` logic for you (internally)

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -2300,4 +2271,53 @@ public function walkResultVariable($resultVariable)
...
+
+            if ($parameter !== $discrClass->name && ! array_key_exists($parameter, $knownSubclasses)) {
```
Ocramius: `$metadata->getClass()->isSubclassOf(...)` maybe?

Ocramius: This also has the effect of getting rid of `INSTANCE OF` checks where `INSTANCE OF` is pointing at the topmost class (which is most likely a bug in userland code)

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -2300,4 +2271,53 @@ public function walkResultVariable($resultVariable)
...
+            $metadata = $this->em->getClassMetadata($parameter);
+            $hierarchyClasses = $metadata->subClasses;
+            $hierarchyClasses[] = $metadata->name;
+
+            foreach ($hierarchyClasses as $class) {
+                $currentMetadata = $this->em->getClassMetadata($class);
+                $currentDiscriminator = $currentMetadata->discriminatorValue;
+
+                if (is_string($currentDiscriminator) && ! array_key_exists($currentDiscriminator, $discriminators)) {
+                    $discriminators[$currentDiscriminator] = true;
```
Ocramius: Consider using `null` if that `true` has no meaning

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -2300,4 +2271,53 @@ public function walkResultVariable($resultVariable)
...
+                }
+            }
+        }
+
+        foreach (array_keys($discriminators) as $dis) {
+            $sqlParameterList[] = $this->conn->quote($dis);
+        }
+
+        return '(' . implode(', ', $sqlParameterList) . ')';
```
Ocramius: According to this logic, the list may be empty, leading to a crash

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -2300,4 +2271,53 @@ public function walkResultVariable($resultVariable)
...
+                $currentDiscriminator = $currentMetadata->discriminatorValue;
+
+                if (is_string($currentDiscriminator) && ! array_key_exists($currentDiscriminator, $discriminators)) {
+                    $discriminators[$currentDiscriminator] = true;
+                }
+            }
+        }
+
+        foreach (array_keys($discriminators) as $dis) {
+            $sqlParameterList[] = $this->conn->quote($dis);
```
Ocramius: `array_map([$this->conn, 'quote'], array_keys($discriminators))`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfAbstractTest.php
```diff
@@ -0,0 +1,114 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket {
```
Ocramius: Please put the entire file under a single namespace

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfMultiLevelTest.php
```diff
@@ -0,0 +1,156 @@
...
+     * @Table(name="instance_of_multi_level_engineer")
+     */
+    class Engineer extends Employee
+    {
+        /**
+         * @Column(type="string")
+         */
+        private $specialization;
+
+        public function getSpecialization()
```
Ocramius: Remove all accessors that aren't strictly needed: prefer public properties if they are needed (although note that these tests don't really need the properties to work, so you should drop everything but the identifier)

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
```diff
@@ -502,7 +502,7 @@ public function testSupportsInstanceOfExpressionsInWherePart()
     {
         $this->assertSqlGeneration(
             "SELECT u FROM Doctrine\Tests\Models\Company\CompanyPerson u WHERE u INSTANCE OF Doctrine\Tests\Models\Company\CompanyEmployee",
-            "SELECT c0_.id AS id_0, c0_.name AS name_1, c0_.discr AS discr_2 FROM company_persons c0_ WHERE c0_.discr IN ('employee')"
+            "SELECT c0_.id AS id_0, c0_.name AS name_1, c0_.discr AS discr_2 FROM company_persons c0_ WHERE c0_.discr IN ('manager', 'employee')"
```
Ocramius: Ouch, this is a bad bug. Good catch!

Ocramius (changes requested):

lib/Doctrine/ORM/Query/SqlWalker.php
Jean85: Right! I was also already doing the `getClassMetadata` a few lines after, I'll just move that up!

lib/Doctrine/ORM/Query/SqlWalker.php
Jean85: I didn't know! I'll do a `!== null` then.

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
Jean85: All credits to @taueres !

lib/Doctrine/ORM/Query/SqlWalker.php
Jean85: We may already have values inside `$sqlParameterList`, see [line 2289](https://github.com/doctrine/doctrine2/pull/6392/files#diff-0c8825d1265a66bb8b20f3d99c276961R2289), so I think it's better to do it like that.

lib/Doctrine/ORM/Query/SqlWalker.php
Jean85: How that could happen? Since [I check if the class is unrelated at line 2297](https://github.com/doctrine/doctrine2/pull/6392/files#diff-0c8825d1265a66bb8b20f3d99c276961R2297), at least one discriminator should pop up, or an exception be thrown...

Jean85:
Ok seems that I've addressed all your concerns, @Ocramius @lcobucci 
Build is green with low deps, other failures seems unrelated, maybe they come from master?

taueres:
Hi! Thank you very much for resuming this PR, I really appreciate that!
I checked this fix once more but unfortunately the issue still occurs when using `INSTANCE OF` with parameters.

Here is an example:
```php
$dql = 'SELECT p FROM Person p WHERE p INSTANCE OF :parameter';
$query = $em->createQuery($dql);
$query->setParameter('parameter', $em->getClassMetadata(Person::class));

// $result contains only direct instances of `Person`, no subclasses
$result = $query->getResult();
```

taueres:
I wrote a failing test case for the issue I mentioned before: https://gist.github.com/taueres/05ff035bb2f72cddd7711b7f93cd9d6a

lib/Doctrine/ORM/Query/SqlWalker.php
Ocramius: Then just `array_merge()` with it

lib/Doctrine/ORM/Query/SqlWalker.php
Ocramius: Yeah, but that flow makes it hard to follow: ideally we should `array_map()` over `$instanceOfExpr->value`, but let's see where the code goes after the additional tests provided by @taueres

Jean85:
Thanks @taueres!
I've found that your test case fails also when doing this:

```php
$dql = 'SELECT p FROM Doctrine\Tests\ORM\Functional\Ticket\PersonTicket4646Parametric p
        WHERE p INSTANCE OF :parameter';
$query = $this->_em->createQuery($dql);
$query->setParameter(
    'parameter',
    PersonTicket4646Parametric::class
);
```
Do we want to check for this additional uses, @Ocramius ?

Ocramius:
```php
$query->setParameter(
    'parameter',
    PersonTicket4646Parametric::class
);
```

This currently requires binding a `ClassMetadata` instance:

```php
$query->setParameter(
    'parameter',
    $em->getClassMetadata(PersonTicket4646Parametric::class)
);
```

We can fix it, yes

Jean85:
Ok I've added the failing test case. I'll try to work on that soon; @taueres if you have any tips, I'm all ears! :smile:

taueres:
Unfortunately solving the parameter issue is not as easy as static DQL queries.
The main problem is that **Doctrine is always translating each DQL parameter to one single SQL parameter**.
For `INSTANCE OF`, this 1-to-1 relationship is not always true: one class might be mapped to multiple discriminators.

The issue is not easy to address because the parameter values could still be undefined while Doctrine is parsing the query.

Think about the following scenario:
```php
$dql = 'SELECT p FROM Person p WHERE p INSTANCE OF :parameter';
$sql = $query->getSQL();
// $sql is something like "SELECT * FROM person WHERE person.kind IN (?)"
// which is WRONG because we don't know yet how many SQL parameters are needed.
```

Do you have any idea of how we could tackle this problem?
Is DBAL supporting `IN` operator with array parameters?

Ocramius:
> Is DBAL supporting `IN operator with array parameters?

Yep, but it only supports [these two types of bindings](https://github.com/doctrine/dbal/blob/03e17189c48a92bff9c5fd7d9242d13a15780d15/lib/Doctrine/DBAL/Connection.php#L73-L85), so custom discriminators may be a problem

taueres:
Okay, I have a solution! `INSTANCE OF` with parameters now works correctly on my branch https://github.com/taueres/doctrine2/tree/fix-instance-of-subclasses

It still needs some minor improvements, e.g. duplicated methods.

How should I proceed with this PR? Should I create a new one from my branch?

Ocramius:
@taueres I suggest waiting for @Jean85 to merge the changes into his branch

Jean85:
I've merged your fixes @taueres, thanks a lot!
Now the only issue that I see is the duplication of the `getAllDiscriminators(ClassMetadata $classMetadata)` method in both the `Query` and `SqlWalker` classes. Any suggestion on how to address that? It uses `$this->getEntityManager()->getClassMetadata($class)`, which I think is the biggest hurdle to overcome, otherwise it could be moved inside the `ClassMetadata` class.

taueres:
To solve the duplication issue we could create something like:
```php
class HierarchyDiscriminatorResolver
{
    public static function resolveDiscriminatorsForClass(ClassMetadata $rootClassMetadata, EntityManagerInterface $entityManager)
    {
        /* Duplicate code goes here */
        /* return list of discriminator */
    }
}
```

In the proposed solution, the `EntityManager` dependency is needed in order to retrieve `ClassMetadata` for subClasses.

I would not put such logic into `ClassMetadata` because atm `ClassMetadata` is not coupled to `EntityManager`.

Jean85:
This seems a feasible solution for me... @Ocramius ? What do you think?

Ocramius:
@Jean85 as long as the class is marked as `@internal`, sure :+1: 

Totally missed these comments before btw.

Jean85:
Don't worry @Ocramius, I was on vacation anyway :smile: 
Deduplication of code done, build is green, this is ready for a final review!

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
```diff
@@ -0,0 +1,34 @@
...
+
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
+use Doctrine\ORM\EntityManagerInterface;
+
+/**
+ * Class HierarchyDiscriminatorResolver
+ * @package Doctrine\ORM\Utility
+ * @internal This class exists only to avoid code duplication, do not reuse it externally
+ */
+class HierarchyDiscriminatorResolver
```
Ocramius: This class needs to be tested in isolation

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Doctrine\ORM\Utility;
+
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
+use Doctrine\ORM\EntityManagerInterface;
+
+/**
+ * Class HierarchyDiscriminatorResolver
```
Ocramius: Remove this line

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
```diff
@@ -0,0 +1,34 @@
...
+ * @package Doctrine\ORM\Utility
```
Ocramius: Remove this line

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
Ocramius: `final`

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
```diff
@@ -0,0 +1,34 @@
...
+ * @internal This class exists only to avoid code duplication, do not reuse it externally
+ */
+class HierarchyDiscriminatorResolver
+{
+    public static function resolveDiscriminatorsForClass(
```
Ocramius: Add a `private` constructor

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
```diff
@@ -0,0 +1,34 @@
...
+        ClassMetadata $rootClassMetadata,
+        EntityManagerInterface $entityManager
+    ) {
```
Ocramius: Return type

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
```diff
@@ -0,0 +1,34 @@
...
+        $hierarchyClasses = $rootClassMetadata->subClasses;
+        $hierarchyClasses[] = $rootClassMetadata->name;
+
+        $discriminators = [];
+        foreach ($hierarchyClasses as $class) {
```
Ocramius: Empty line before logical blocks

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
Ocramius: An explanation of the method is needed

Ocramius (changes requested):

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -2297,4 +2269,37 @@ public function walkResultVariable($resultVariable)
...
         return $resultAlias;
     }
+
+    /**
+     * @param ClassMetadataInfo $rootClass
+     * @param AST\InstanceOfExpression $instanceOfExpr
+     * @return string The list in parentheses of valid child discriminators from the given class
+     * @throws QueryException
+     */
+    private function getChildDiscriminatorsFromClassMetadata(ClassMetadataInfo $rootClass, AST\InstanceOfExpression $instanceOfExpr)
```
lcobucci: Please add the  return type declaration

taueres:
According to the DQL spec. Users can provide multiple classes for an INSTANCE OF statement.
```
InstanceOfExpression     ::= IdentificationVariable ["NOT"] "INSTANCE" ["OF"] (InstanceOfParameter | "(" InstanceOfParameter {"," InstanceOfParameter}* ")")
```
Do we have some tests covering this specific use case?

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfAbstractTest.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class Ticket4646InstanceOfAbstractTest extends OrmFunctionalTestCase
+{
+    protected function setUp()
```
lcobucci: `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfAbstractTest.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(PersonTicket4646Abstract::class),
+            $this->_em->getClassMetadata(EmployeeTicket4646Abstract::class),
+        ]);
+    }
+
+    public function testInstanceOf()
```
lcobucci: `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfAbstractTest.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $this->_em->persist(new EmployeeTicket4646Abstract());
+        $this->_em->flush();
+
+        $dql = 'SELECT p FROM Doctrine\Tests\ORM\Functional\Ticket\PersonTicket4646Abstract p
+                WHERE p INSTANCE OF Doctrine\Tests\ORM\Functional\Ticket\PersonTicket4646Abstract';
+        $query = $this->_em->createQuery($dql);
+        $result = $query->getResult();
+
+        $this->assertCount(1, $result);
```
lcobucci: Please use `self::assert*` for new tests

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfMultiLevelTest.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class Ticket4646InstanceOfMultiLevelTest extends OrmFunctionalTestCase
+{
+    protected function setUp()
```
lcobucci: `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfMultiLevelTest.php
```diff
@@ -0,0 +1,77 @@
...
+        parent::setUp();
+
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(PersonTicket4646MultiLevel::class),
+            $this->_em->getClassMetadata(EmployeeTicket4646MultiLevel::class),
+            $this->_em->getClassMetadata(EngineerTicket4646MultiLevel::class),
+        ]);
+    }
+
+    public function testInstanceOf()
```
lcobucci: `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfMultiLevelTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $this->_em->persist(new EmployeeTicket4646MultiLevel());
+        $this->_em->persist(new EngineerTicket4646MultiLevel());
+        $this->_em->flush();
+
+        $dql = 'SELECT p FROM Doctrine\Tests\ORM\Functional\Ticket\PersonTicket4646MultiLevel p
+                WHERE p INSTANCE OF Doctrine\Tests\ORM\Functional\Ticket\PersonTicket4646MultiLevel';
+        $query = $this->_em->createQuery($dql);
+        $result = $query->getResult();
+
+        $this->assertCount(3, $result);
```
lcobucci: Here too (`self::assert*`)

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfParametricTest.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class Ticket4646InstanceOfParametricTest extends OrmFunctionalTestCase
+{
+    protected function setUp()
```
lcobucci: `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfParametricTest.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        parent::setUp();
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(PersonTicket4646Parametric::class),
+            $this->_em->getClassMetadata(EmployeeTicket4646Parametric::class),
+        ]);
+    }
+
+    public function testInstanceOf()
```
lcobucci: `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfParametricTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->_em->flush();
+        $dql = 'SELECT p FROM Doctrine\Tests\ORM\Functional\Ticket\PersonTicket4646Parametric p
+                WHERE p INSTANCE OF :parameter';
+        $query = $this->_em->createQuery($dql);
+        $query->setParameter(
+            'parameter',
+            $this->_em->getClassMetadata(PersonTicket4646Parametric::class)
+        );
+        $result = $query->getResult();
+        $this->assertCount(2, $result);
```
lcobucci: `self::assert*` here too

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfTest.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class Ticket4646InstanceOfTest extends OrmFunctionalTestCase
+{
+    protected function setUp()
```
lcobucci: `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfTest.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(PersonTicket4646::class),
+            $this->_em->getClassMetadata(EmployeeTicket4646::class),
+        ]);
+    }
+
+    public function testInstanceOf()
```
lcobucci: `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfMultiLevelTest.php
```diff
@@ -0,0 +1,77 @@
...
+class PersonTicket4646MultiLevel
+{
+    /**
+     * @Id()
+     * @GeneratedValue()
+     * @Column(type="integer")
+     */
+    private $id;
+
+    public function getId()
```
lcobucci: `: ?int`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfParametricTest.php
```diff
@@ -0,0 +1,67 @@
...
+class PersonTicket4646Parametric
+{
+    /**
+     * @Id()
+     * @GeneratedValue()
+     * @Column(type="integer")
+     */
+    private $id;
+
+    public function getId()
```
lcobucci: `: ?int`

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfTest.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->_em->persist(new PersonTicket4646());
+        $this->_em->persist(new EmployeeTicket4646());
+        $this->_em->flush();
+
+        $dql = 'SELECT p FROM Doctrine\Tests\ORM\Functional\Ticket\PersonTicket4646 p
+                WHERE p INSTANCE OF Doctrine\Tests\ORM\Functional\Ticket\PersonTicket4646';
+        $query = $this->_em->createQuery($dql);
+        $result = $query->getResult();
+
+        $this->assertCount(2, $result);
```
lcobucci: Here too

tests/Doctrine/Tests/ORM/Functional/Ticket/Ticket4646InstanceOfTest.php
```diff
@@ -0,0 +1,66 @@
...
+class PersonTicket4646
+{
+    /**
+     * @Id()
+     * @GeneratedValue()
+     * @Column(type="integer")
+     */
+    private $id;
+
+    public function getId()
```
lcobucci: `: ?int`

lcobucci (dismissed):
Some nitpicking 😄

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
Jean85: Sorry, I forgot that the target branch is 7+! Good!

lib/Doctrine/ORM/Utility/HierarchyDiscriminatorResolver.php
Jean85: You should enforce this in the Lint stage of the CI!

lcobucci: We're going to :wink

Jean85:
@lcobucci @Ocramius done as per review!

Jean85:
I've also added a new functional test to cover the case of multiple parameters passed to the `INSTANCEOF`, as suggested by @taueres

Ocramius (approved):

Ocramius (approved):
LGTM :+1: :ship:

Ocramius:
Thanks @Jean85 @taueres and @phansys!

PR #6394, opened 2017-04-12
Fix for #6393 - use field names when extracting identifier fields

### LINKED ISSUE #6393: Exception in `IdentifierFlattener::flattenIdentifier` - column name used instead of field name in identifier extraction
Consider the following two entities:
```
class A
{
    /**
     * @Id 
     **/
    private $id;

    private $something;

    /**
     * @Version
     */
    private $version;
};

class B
{
	/**
	 * @Id
	 * @ManyToOne(targetEntity="A")
	 * @JoinColumn(name="aid", referencedColumnName="id)
	 */
	private $A
	
	private $something;
	
	/**
	 * @Version
	 */
	private $version
}
````

I get an Exception in IdentifierFlattener::flattenIdentifier called from BasicEntityPersister::fetchVersionValue because. BasicEntityPersister::fetchVersionValue calls IdentifierFlattener::flattenIdentifier with the identifier field name, but IdentifierFlattener::flattenIdentifier  expects the identifier column names.

### PR DESCRIPTION
**Pull request by tobiasstadler**
Fixes #6393

### CODE REVIEW

File: lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php [hunk 1]
```diff
@@ -334,17 +334,26 @@ protected function fetchVersionValue($versionedClass, array $id)
         $versionField = $versionedClass->versionField;
         $fieldMapping = $versionedClass->fieldMappings[$versionField];
         $tableName    = $this->quoteStrategy->getTableName($versionedClass, $this->platform);
-        $identifier   = $this->quoteStrategy->getIdentifierColumnNames($versionedClass, $this->platform);
         $columnName   = $this->quoteStrategy->getColumnName($versionField, $versionedClass, $this->platform);
+		
+		$where = array();
```
lcobucci: Please use short array syntax

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -334,17 +334,26 @@ protected function fetchVersionValue($versionedClass, array $id)
...
+        $params = array();
```
lcobucci: Please use short array syntax

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -334,17 +334,26 @@ protected function fetchVersionValue($versionedClass, array $id)
...
+        foreach ($versionedClass->identifier as $idField) {
```
lcobucci: Can you extract a method from this code?

lcobucci (changes requested):
Also we need tests 😄

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -2071,4 +2070,30 @@ protected function switchPersisterContext($offset, $limit)
 
         $this->currentPersisterContext = $this->limitsHandlingContext;
     }
+
+    /**
+     * Determains the identifier columns and values for an entity
+     *
+     * @param \Doctrine\ORM\Mapping\ClassMetadata $class
```
phansys: This class is already imported, so please use `ClassMetadata` instead FQCN.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -2071,4 +2070,30 @@ protected function switchPersisterContext($offset, $limit)
...
+     * @param array                               $id
+     * @return array
+     */
+    private function determineIdentifierColumnsAndValues($class, array $id)
```
phansys: Could you please declare the type for `$class` argument?

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,131 @@
...
+    public function testFetchVersionValueForDifferentIdFieldAndColumn()
+    {
+        $a = new A(1);
+        $this->_em->persist($a);
+
+        $b = new B($a, 'foo');
+        $this->_em->persist($b);        
+        $this->_em->flush();
+        
+        $this->assertEquals(1, $b->getVersion());
```
phansys: Please, use `assertSame()` instead.

phansys (changes requested):
Status: Needs work.

tobiasstadler:
Any chance for getting this in?

Ocramius:
I think this looks fine, but I don't have my laptop atm. Will try merging later today :+1:

Ocramius (changes requested):
I went through the patch again and tried applying it locally. There are some major issues that need fixing:

 * [ ] only the first association column is considered, whereas the `IdentifierFlattener` considered all columns
 * [ ] the association column is not being quoted
 * [ ] overall, the logic in `determineIdentifierColumnsAndValues` is an identifier lookup, and it should produce a `[ $values, $columns, $types]` triplet
 * [ ] The code in `determineIdentifierColumnsAndValues` should be abstracted away, or an existing abstraction should be found. The `IdentifierFlattener` is close to it, but it was designed for the identity map internals, not for column operations.

tobiasstadler:
Hi I've tried a different (and hopefully better) approach for solving #6394

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -386,7 +386,16 @@ public function update($entity)
         $this->updateTable($entity, $quotedTableName, $data, $isVersioned);
 
         if ($isVersioned) {
-            $id = $this->em->getUnitOfWork()->getEntityIdentifier($entity);
+            $identifier = $this->em->getUnitOfWork()->getEntityIdentifier($entity);
+
+            $id = [];
+            foreach ($this->class->identifier as $idField) {
```
lcobucci: This seems like a workaround for me... [as @Ocramius said](https://github.com/doctrine/doctrine2/issues/6393#issuecomment-293533668) we should rather use field names everywhere.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+class DDC6393Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+
+    protected function setUp()
```
lcobucci: ` : void` missing

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+                $this->_em->getClassMetadata(B::class)
+            ]
+        );
+    }
+
+    /**
+     * Test the the version of an entity can be fetched, when the id field and
+     * the id column are different.
+     */
+    public function testFetchVersionValueForDifferentIdFieldAndColumn()
```
lcobucci: ` : void` missing

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+                $this->_em->getClassMetadata(A::class),
+                $this->_em->getClassMetadata(B::class)
+            ]
+        );
+    }
+
+    /**
+     * Test the the version of an entity can be fetched, when the id field and
+     * the id column are different.
+     */
```
lcobucci: Please add `@group 6393`

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+    public function testFetchVersionValueForDifferentIdFieldAndColumn()
+    {
+        $a = new A(1);
+        $this->_em->persist($a);
+
+        $b = new B($a, 'foo');
+        $this->_em->persist($b);        
+        $this->_em->flush();
+        
+        $this->assertSame(1, $b->getVersion());
```
lcobucci: `self::assert*` is more appropriated since the method is static

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+
+        $b->setSomething('bar');
+        $this->_em->flush();
+        
+        $this->assertSame(2, $b->getVersion());
```
lcobucci: `self::assert*` is more appropriated since the method is static

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+     * @Column(type="integer")
+     */
+    private $version;
+
+    public function __construct($id)
+    {
+        $this->id = $id;
+    }
+
+    public function getId()
```
lcobucci: Please this method and make the property `public`, so that we simplify the entity a bit

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+     */
+    private $id;
+
+    /**
+     * @Version
+     * @Column(type="integer")
+     */
+    private $version;
+
+    public function __construct($id)
```
lcobucci: `int $id`

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        $this->id = $id;
+    }
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function getVersion()
```
lcobucci: Please this method and make the property `public`, so that we simplify the entity a bit

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+     */
+    private $something;
+
+    /**
+     * @Version
+     * @Column(type="integer")
+     */
+    private $version;
+
+    public function __construct($a, $something)
```
lcobucci: `public function __construct(A $a, string $something)`

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        $this->a = $a;
+        $this->something = $something;
+    }
+
+    public function getA()
```
lcobucci: Please this method and make the property `public`, so that we simplify the entity a bit

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return $this->a;
+    }
+
+    public function getSomething()
```
lcobucci: Please this method and make the property `public`, so that we simplify the entity a bit

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return $this->something;
+    }
+
+    public function setSomething($something)
+    {
+        $this->something = $something;
+    }
+
+    public function getVersion()
```
lcobucci: Please this method and make the property `public`, so that we simplify the entity a bit

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6393Test.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return $this->a;
+    }
+
+    public function getSomething()
+    {
+        return $this->something;
+    }
+
+    public function setSomething($something)
```
lcobucci: Please this method and simply change the property on the test

lcobucci (changes requested):

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
beberlei: This doesn't seem possible, because in `UnitOfWork::createEntity` it passes the whole $row ($data) into the identifier flattener and the hydration works by passing the association data as columns not as fields, because it must support composite primary keys as assocations where using field name as a single name would not be able to represent two ore more columns.

beberlei: ACtually the other way around, BasicEntityPersister::update used the wrong identifier representation. See #8384

Issue #5575, opened 2015-12-27
Getting started documentation links are broken?


Ocramius:
Latest documentation (compiled) seems to have broken links in http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/getting-started.html ([getting-started.rst](https://github.com/doctrine/doctrine2/blob/3ca68285449cbf43a20c2b8383bd82745d68bdb6/docs/en/tutorials/getting-started.rst#obtaining-the-entitymanager)):
![e9fd57](https://cloud.githubusercontent.com/assets/154256/12010723/88555be4-acb2-11e5-83c8-a71848576c54.png)

harikt:
Accidentally related #5583 :) .

PR #6395, opened 2017-04-12
Updated broken links

yannickl88:
Updated all links pointing to
http://www.doctrine-project.org/documentation/* to their
http://docs.doctrine-project.org counterparts

This should fix the broken link of issue #5575


Ocramius (approved):
LGTM! Thanks!

PR #6399, opened 2017-04-14
Fix incorrect YAML example for versioning a field.

neobyteUK:
The example failed for me in practice, until I made the fix here. A casual comparison of the YAML and XML example in the doc file also supports "version" and "type" being configured at the same indentation level.


Ocramius (approved):
LGTM, thanks! :+1:

PR #6400, opened 2017-04-15
Fix incorrect YAML datetime version field.

neobyteUK:
Apologies @Ocramius , as I wasn't personally using a datetime version field I didn't notice the documentation was incorrect for datetime as well. This corrects the other example.


Ocramius:
@neobyteUK fair enough, thanks for noticing :+1:

PR #6412, opened 2017-04-25
"Notice: Undefined offset: 0" when there is no ID Column defined

JarJak:
When you forget about defining the ID/PK Column, then this ugly Notice appear.

This PR is not ready for merge, but shows where the problem exist.

Could you fix this or tell me the way this Pull Request would be acceptable? Thanks.

How to reproduce?
Remove ORM\Id annotation, then try to regenerate entities.


Ocramius:
@JarJak metadata without identifier isn't really valid metadata - is this for an embeddable? Can a test be added?

JarJak:
@Ocramius I agree that metadata without identifier is invalid, however if you throw nice Exception than this Notice, then this mistake would be much easier to debug and fix.

I will add test later.

JarJak:
@Ocramius I've updated PR with two nearly identical tests (I don't know where exactly I should put my test). However this change might be a BC Break if someone has error reporting turned off (previously it would silently fail instead of throwing exception).

lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -426,6 +426,16 @@ public static function singleIdNotAllowedOnCompositePrimaryKey($entity)
 
     /**
      * @param string $entity
+     *
+     * @return MappingException
+     */
+    public static function noIdDefined($entity)
+    {
+        return new self('No ID defined for entity '.$entity);
```
lcobucci: Please add spaces before and after the concatenation operator (`'No ID defined for entity ' . $entity`)

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6412Test.php
```diff
@@ -0,0 +1,23 @@
...
+
+use Doctrine\Common\Persistence\Mapping\RuntimeReflectionService;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\Tests\Models\DDC6412\DDC6412File;
+use Doctrine\Tests\OrmTestCase;
+
+/**
+ * @author JarJak
+ */
+class DDC6412Test extends OrmTestCase
```
lcobucci: I'd say it's better to leave this check in `ClassMetadataTest.php` only, then we keep it together with the other mapping assertions

lcobucci (changes requested):

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1803,6 +1803,10 @@ public function getSingleIdentifierFieldName()
         if ($this->isIdentifierComposite) {
             throw MappingException::singleIdNotAllowedOnCompositePrimaryKey($this->name);
         }
+        
+        if ( ! isset($this->identifier[0])) {
```
lcobucci: Can you also mention this change in the docblock? We currently have `@throws MappingException If the class has a composite primary key.` but would be good to say something like `@throws MappingException If the class doesn't have an identifier or it has a composite primary key.`

lcobucci:
> This change might be a BC Break if someone has error reporting turned off

@JarJak It's not really considered a BC Break since we already expected that type of exception. It was actually something that we missed. Thanks for addressing it 😉

JarJak:
@lcobucci I polished the code according to your tips, thanks :)

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
JarJak: done

lcobucci (approved):
LGTM I'll just rebase the branch, remove the methods in the sample entity (since they're not needed) and merge it. Basically nitpicking 😜

Thanks @JarJak

lcobucci:
Merged manually in 38bfcc6a7ad264b880b847b69cafa10e3df5cf5d. @Ocramius should we backport this to 2.5.x?

Ocramius:
@lcobucci no, I think this is fine with `2.6.0`-only

PR #6404, opened 2017-04-18
Fix quoting issues related to ticket #6402

lemartin:
Apply quoting strategy to foreign key columns of one-to-one relation so
that correct select statements are generated. And use unquoted column
names in result mapping instead of quoted ones, as consumers (namely
IdentifierFlattener) expect unquoted column names.

Issue #6402, opened 2017-04-17
Quote strategy not applied to foreign key columns


lemartin:
The quote strategy is not applied to foreign key columns used in one-to-one associations. When fetching the [Address entity](https://github.com/doctrine/doctrine2/blob/2.5/tests/Doctrine/Tests/Models/Quote/Address.php) the column `user-id` is not quoted. 

```sql
SELECT 
  q0_."address-id" AS addressid_0,
  q0_."address-zip" AS addresszip_1,
  q0_.user-id AS userid_2
FROM "quote-address" q0_
```

The tests do not cover this scenario, as they force partially loaded entities. Applying the patch below should yield a failing test case. The columns come from a collection called `targetToSourceKeyColumns` where, if I understood correctly, all columns are stored that are used as 

As far as I can tell, the problem is in `SqlWalker::walkSelectClause`, where the foreign key columns are added to the list of select expressions.

Patch for creating a test case, based on branch 2.5 (48e8c02cb8c20d10240248b0232e31bf785b0209):
```diff
diff --git a/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php b/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
index a3a34f8..21b6961 100644
--- a/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
@@ -1972,0 +1973,6 @@ class SelectSqlGenerationTest extends \Doctrine\Tests\OrmTestCase
+
+        $this->assertSqlGeneration(
+            'SELECT a FROM Doctrine\Tests\Models\Quote\Address a',
+            'SELECT q0_."address-id" AS addressid_0, q0_."address-zip" AS addresszip_1, q0_."user-id" AS userid_2 FROM "quote-address" q0_',
+            [Query::HINT_FORCE_PARTIAL_LOAD => false]
+        );
```

theofidry:
AFAIKS it's a duplicate of https://github.com/doctrine/doctrine2/issues/5874 :/

Ocramius:
@theofidry it's similar, but not a duplicate. @lemartin is the join column mapped with backticks?

lemartin:
Yes, the column is mapped with back-ticks.

```php
    /**
     * @OneToOne(targetEntity="User", inversedBy="address")
[...]

PR #6416, opened 2017-04-30
Fix quoting when using inheritance as well

lcobucci:
Related to #6402 and #6404

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/AbstractEntityInheritancePersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/AbstractEntityInheritancePersister.php
@@ -82,16 +82,18 @@ abstract class AbstractEntityInheritancePersister extends BasicEntityPersister
     /**
      * @param string $tableAlias
      * @param string $joinColumnName
+     * @param string $quotedColumnName
+     *
      * @param string $type
      *
      * @return string
      */
-    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $type)
+    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $quotedColumnName, $type)
     {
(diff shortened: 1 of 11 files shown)


lib/Doctrine/ORM/Persisters/Entity/AbstractEntityInheritancePersister.php
```diff
@@ -82,16 +82,18 @@ protected function getSelectColumnSQL($field, ClassMetadata $class, $alias = 'r'
...
      * @param string $tableAlias
      * @param string $joinColumnName
+     * @param string $quotedColumnName
+     *
      * @param string $type
      *
      * @return string
      */
-    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $type)
+    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $quotedColumnName, $type)
```
lcobucci: Theoretically this is as minor BC break, should we add to `UPGRADING.md`?

guilhermeblanco: Not a valid BC break since we do not allow custom persisters.

guilhermeblanco:
`develop` is way ahead of the game and honestly any change around quoting is ok since it'll all changed already.

Ocramius:
@guilhermeblanco just a note: rebasing on this will likely be annoying. Still, the fix is valid IMO, and it shall be merged in `master`

lib/Doctrine/ORM/Persisters/Entity/AbstractEntityInheritancePersister.php
Ocramius: Indeed, change is fine here

Ocramius (approved):
LGTM :ship:

PR #6390, opened 2017-04-11
DDC-2248: Expire result cache functionality not implemented #2947

bruno-ds:
PR for this issue : https://github.com/doctrine/doctrine2/issues/2947#issuecomment-293181787

I don't know how to run the test... sorry

PR #6417, opened 2017-04-30
Fix expire result cache

lcobucci:
Simplifying test provided in #6390 and process the result set cache eviction when flag was passed.

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -20,6 +20,7 @@
 namespace Doctrine\ORM;
 
 use Doctrine\DBAL\LockMode;
+use Doctrine\ORM\Query\Exec\AbstractSqlExecutor;
 use Doctrine\ORM\Query\Parser;
 use Doctrine\ORM\Query\ParserResult;
 use Doctrine\ORM\Query\QueryException;
@@ -322,7 +323,25 @@ final class Query extends AbstractQuery
 
         list($sqlParams, $types) = $this->processParameterMappings($paramMappings);
 
+        $this->evictResultSetCache($executor, $sqlParams, $types);
+
         return $executor->execute($this->_em->getConnection(), $sqlParams, $types);
+    }
+
+    private function evictResultSetCache(AbstractSqlExecutor $executor, array $sqlParams, array $types)
+    {
+        if (null === $this->_queryCacheProfile || ! $this->getExpireResultCache()) {
+            return;
+        }
+
+        $cacheDriver = $this->_queryCacheProfile->getResultCacheDriver();
+        $statements  = (array) $executor->getSqlStatements(); // Type casted since it can either be a string or an array
+
+        foreach ($statements as $statement) {
+            $cacheKeys = $this->_queryCacheProfile->generateCacheKeys($statement, $sqlParams, $types);
+
+            $cacheDriver->delete(reset($cacheKeys));
+        }
     }
 
     /**
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH2947Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH2947Test.php
@@ -1,0 +1,95 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Cache\ArrayCache;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group 2947
+ */
+class GH2947Test extends OrmFunctionalTestCase
+{
+    protected function setUp()
+    {
+        $this->resultCacheImpl = new ArrayCache();
+
+        parent::setUp();
+
+        $this->_schemaTool->createSchema([$this->_em->getClassMetadata(GH2947Car::class)]);
+    }
+
+    public function testIssue()
+    {
+        $this->createData();
+        $initialQueryCount = $this->getCurrentQueryCount();
+
+        $query = $this->createQuery();
+        self::assertEquals('BMW', (string) $query->getSingleResult());
+        self::assertEquals($initialQueryCount + 1, $this->getCurrentQueryCount());
+
+        $this->updateData();
+        self::assertEquals('BMW', (string) $query->getSingleResult());
+        self::assertEquals($initialQueryCount + 2, $this->getCurrentQueryCount());
+
+        $query->expireResultCache(true);
+        self::assertEquals('Dacia', (string) $query->getSingleResult());
+        self::assertEquals($initialQueryCount + 3, $this->getCurrentQueryCount());
+
+        $query->expireResultCache(false);
+        self::assertEquals('Dacia', (string) $query->getSingleResult());
+        self::assertEquals($initialQueryCount + 3, $this->getCurrentQueryCount());
+    }
+
+    private function createQuery()
+    {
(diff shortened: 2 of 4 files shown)


lib/Doctrine/ORM/Query.php
```diff
@@ -322,6 +322,12 @@ protected function _doExecute()
 
         list($sqlParams, $types) = $this->processParameterMappings($paramMappings);
 
+        if ($this->_queryCacheProfile && $this->getExpireResultCache()) {
+            $this->_queryCacheProfile->getResultCacheDriver()->delete(
+                $this->_queryCacheProfile->generateCacheKeys($executor->getSqlStatements(), $sqlParams, $types)[0]
```
guilhermeblanco: If this is a list, iterate over it (even though now it only returns one item) and delete each record.

lcobucci: I thought that but although the docblock says it returns an `array` it's actually a string 😱

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -682,6 +689,11 @@ protected function _getEntityManager($config = null, $eventManager = null) {
             self::$_queryCacheImpl = new ArrayCache();
         }
 
+        if (is_null(self::$_resultCacheImpl)) {
```
Ocramius: `null ===`

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -40,6 +40,13 @@
     private static $_queryCacheImpl = null;
 
     /**
+     * The result cache shared between all functional tests.
+     *
+     * @var \Doctrine\Common\Cache\Cache|null
+     */
+    private static $_resultCacheImpl = null;
```
Ocramius: If we introduce new properties, we shouldn't use the prefix... Also, the fact that this is a static smells a lot: what if it is accidentally not unset between test cycles? Is it supposed to always be set?

lib/Doctrine/ORM/Query.php
```diff
@@ -322,9 +323,24 @@ protected function _doExecute()
...
 
+    private function evictResultSetCache(AbstractSqlExecutor $executor, array $sqlParams, array $types)
+    {
+        if (null === $this->_queryCacheProfile || ! $this->getExpireResultCache()) {
+            return;
+        }
+
+        $cacheDriver = $this->_queryCacheProfile->getResultCacheDriver();
+
+        foreach ((array) $executor->getSqlStatements() as $statement) {
```
Ocramius: Why the `(array)` cast? Smelly :|

lib/Doctrine/ORM/Query.php
```diff
@@ -322,9 +323,24 @@ protected function _doExecute()
...
+            $cacheDriver->delete($this->_queryCacheProfile->generateCacheKeys($statement, $sqlParams, $types)[0]);
```
Ocramius: That `[0]` is also a bit smelly

tests/Doctrine/Tests/ORM/Functional/Ticket/GH2947Test.php
```diff
@@ -0,0 +1,92 @@
...
+     * @GeneratedValue(strategy="NONE")
+     */
+    public $brand;
+
+    public function __construct($brand)
+    {
+        $this->brand = $brand;
+    }
+
+    public function __toString()
```
Ocramius: We can start violently adding type hints everywhere

tests/Doctrine/Tests/ORM/Functional/Ticket/GH2947Test.php
```diff
@@ -0,0 +1,92 @@
...
+class GH2947Car
+{
+    /**
+     * @Id
+     * @Column(type="string", length=25)
+     * @GeneratedValue(strategy="NONE")
+     */
+    public $brand;
+
+    public function __construct($brand)
```
Ocramius: We can start violently adding type hints everywhere

tests/Doctrine/Tests/ORM/Functional/Ticket/GH2947Test.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group 2947
+ */
+class GH2947Test extends OrmFunctionalTestCase
```
Ocramius: Do we have a `Query`-specific test that we can use here?

Ocramius (changes requested):

lib/Doctrine/ORM/Query.php
lcobucci: It's indeed... cause: https://github.com/doctrine/doctrine2/pull/6417#discussion_r114082839

tests/Doctrine/Tests/ORM/Functional/Ticket/GH2947Test.php
lcobucci: except for `: void` 😭

lib/Doctrine/ORM/Query.php
Ocramius: @lcobucci can you add that as a comment in the code?

lcobucci: Sure 😉

tests/Doctrine/Tests/ORM/Functional/Ticket/GH2947Test.php
Ocramius: @lcobucci bump it

tests/Doctrine/Tests/ORM/Functional/Ticket/GH2947Test.php
lcobucci: As in a unit test?

lib/Doctrine/ORM/Query.php
lcobucci: What do you suggest? `first()`? `list()`?

Ocramius: `reset()` or similar works fine

tests/Doctrine/Tests/OrmFunctionalTestCase.php
lcobucci: True thing, I'll keep it non-static and leave the instantiation to the test that actually uses it. Thanks

tests/Doctrine/Tests/ORM/Functional/Ticket/GH2947Test.php
lcobucci: [`Doctrine\Tests\ORM\Query\QueryTest#testResultCacheEviction()`](https://github.com/doctrine/doctrine2/pull/6417/files#diff-8839a19e4228e39f37c3000d51e43beeR250) not strictly unit test though but should be enough 😉

Ocramius: Thanks for adding these :-)

Ocramius (approved):

Ocramius:
@lcobucci :ship: \o/

lcobucci:
@Ocramius awesome thanks!

PR #6439, opened 2017-05-13
align equals signs

sensorario:
According to the project coding standard I've aligned the equals signs.


lcobucci (approved):
@sensorario thanks!

PR #6440, opened 2017-05-14
Initial PHPStan integration

ondrejmirtes:
This is a starting point to begin using static analysis in Doctrine ORM codebase. [PHPStan](https://github.com/phpstan/phpstan) is already used by doctrine/common and doctrine/annotations so I think it's logical that the ORM repo starts using it too :)

The build will now fail, I wasn't sure what to do about this one error:

```
 ------ ------------------------------------------------------------------------------------------
  Line   lib/Doctrine/ORM/Query/ResultSetMapping.php
 ------ ------------------------------------------------------------------------------------------
  456    Return typehint of method Doctrine\ORM\Query\ResultSetMapping::getRelation() has invalid
         type Doctrine\ORM\Query\AssociationMapping.
 ------ ------------------------------------------------------------------------------------------
```

The method has no tests and no usage in the project, so I'm really not sure what it's supposed to return.

After you merge this, I can spread the analysis to tests and also increase the level of the analysis to find more errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ before_script:
 script:
   - ENABLE_SECOND_LEVEL_CACHE=0 ./vendor/bin/phpunit -v -c tests/travis/$DB.travis.xml $PHPUNIT_FLAGS
   - ENABLE_SECOND_LEVEL_CACHE=1 ./vendor/bin/phpunit -v -c tests/travis/$DB.travis.xml --exclude-group performance,non-cacheable,locking_functional
+  - if [[ $PHPSTAN = 1 ]]; then vendor/bin/phpstan analyse -l 0 lib; fi
 
 after_script:
   - if [[ $TRAVIS_PHP_VERSION = '7.0' && $DB = 'sqlite' ]]; then wget https://scrutinizer-ci.com/ocular.phar; fi
@@ -75,6 +76,10 @@ matrix:
       dist: trusty
       group: edge # until the next update
       env: DB=sqlite
+    - php: 7
+      env: PHPSTAN=1
+    - php: 7.1
+      env: PHPSTAN=1
 
   allow_failures:
     - php: nightly
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
     },
     "require-dev": {
         "symfony/yaml": "~2.3|~3.0",
-        "phpunit/phpunit": "^6.0"
+        "phpunit/phpunit": "^6.0",
+        "phpstan/phpstan": "^0.7"
     },
     "suggest": {
         "symfony/yaml": "If you want to use YAML Metadata Mapping Driver"
--- a/lib/Doctrine/ORM/PersistentCollection.php
+++ b/lib/Doctrine/ORM/PersistentCollection.php
@@ -514,10 +514,10 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
     public function offsetSet($offset, $value)
     {
         if ( ! isset($offset)) {
-            return $this->add($value);
+            $this->add($value);
         }
 
-        return $this->set($offset, $value);
+        $this->set($offset, $value);
     }
 
     /**
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -738,7 +738,7 @@ final class Query extends AbstractQuery
             ->getName();
 
         return md5(
-            $this->getDql() . serialize($this->_hints) .
+            $this->getDQL() . serialize($this->_hints) .
             '&platform=' . $platform .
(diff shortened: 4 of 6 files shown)


composer.json
```diff
@@ -26,7 +26,8 @@
     },
     "require-dev": {
         "symfony/yaml": "~2.3|~3.0",
-        "phpunit/phpunit": "^6.0"
+        "phpunit/phpunit": "^6.0",
+        "phpstan/phpstan": "^0.7"
```
theofidry: I would require it in a different `composer.json` via https://github.com/bamarni/composer-bin-plugin for example to avoid a too grater diff between the deps installed by the consumers and the deps installed for testing

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -514,10 +514,10 @@ public function offsetGet($offset)
     public function offsetSet($offset, $value)
     {
         if ( ! isset($offset)) {
-            return $this->add($value);
+            $this->add($value);
```
lcobucci: I'm not sure about this one since now both `add()` and `set()` will be called if `isset($offset) === false`, so we should at least have a `return` after `$this->add($value);`

composer.json
lcobucci: @ondrejmirtes the installation on both `doctrine/annotations` and `doctrine/common` is different from what you're proposing here, is there any reason for that?

lcobucci (changes requested):

lib/Doctrine/ORM/PersistentCollection.php
ondrejmirtes: Are you calling `offsetSet` directly somewhere? Because from `$collection['foo'] = $value`, you cannot get the return value of `offsetSet` method. I find out about this issue because `offsetSet` has `@return void`, so it should not return anything.

lcobucci: What I meant was this:

```php
if ( ! isset($offset)) {
    $this->add($value);
    return;
}
```

I totally agree that it should return any value since it's a `void` but we shouldn't change the behaviour ;)

ondrejmirtes: Oh yeah, sure, I missed that :)

lcobucci:
@Ocramius do you agree on applying the very same approach we have in the other projects?

Ocramius:
> very same approach

Unsure what this is about

lcobucci:
@Ocramius https://github.com/doctrine/annotations/blob/7c830a576f9471f71096d200d633b6400513ccfb/.travis.yml#L24

Ocramius:
Ah, yep, we shouldn't add it to `require-dev` due to the amount of additional dependencies...

theofidry:
You can use something like https://github.com/bamarni/composer-bin-plugin to install it whilst keeping references of it in the project

ondrejmirtes:
You could use `phpstan/phpstan-shim` dependency that only installs a PHAR and does not have any other dependencies. It's still a little bit experimental but should work.

Ocramius:
I would rather rely on "the real thing".

The problem is only for the dependency restrictions being added, so it's
fine as long as we manually run `composer require phpstan/phpstan`

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

.travis.yml
```diff
@@ -23,6 +23,7 @@ before_script:
 script:
   - ENABLE_SECOND_LEVEL_CACHE=0 ./vendor/bin/phpunit -v -c tests/travis/$DB.travis.xml $PHPUNIT_FLAGS
   - ENABLE_SECOND_LEVEL_CACHE=1 ./vendor/bin/phpunit -v -c tests/travis/$DB.travis.xml --exclude-group performance,non-cacheable,locking_functional
+  - if [[ $PHPSTAN = 1 ]]; then vendor/bin/phpstan analyse -l 0 lib; fi
```
Ocramius: The fact that we are at `-l 0` is quite... frightening ^_^''

Majkl578:
Rebased and tested PHPStan 0.7 against master.

This is for `lib/` only:

level | # of errors
----- | ------
level 0 | 2 errors
level 1 | 21 errors
level 2 | 557 errors
level 3 | 638 errors
level 4 | 682 errors
level 5 | 871 errors
level 6 | 886 errors
level 7 | 1053 errors

So what would we want to do here? Getting this into master at level 0 sounds almost useless, so maybe level 3 (like in annotations) - is it actually worth it for 2.x? And then, gradually "upgade" develop to level 5?

Ocramius:
I would say level 1 for the first patch: rebasing `develop` on it will be painful enough

Majkl578:
-> #6535

lcobucci:
Handled by #6535

PR #6441, opened 2017-05-15
Verify that the `fileLockRegionDirectory` passed to the `DefaultCacheFactory` cannot be empty

sensorario:
The expression $this->fileLockRegionDirectory of type string|null is loosely compared to false; this is ambiguous if the string can be empty. You might want to explicitly use === null instead.

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
+++ b/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
@@ -210,7 +210,10 @@ class DefaultCacheFactory implements CacheFactory
 
         if ($cache['usage'] === ClassMetadata::CACHE_USAGE_READ_WRITE) {
 
-            if ( ! $this->fileLockRegionDirectory) {
+            if (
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
```diff
@@ -210,7 +210,7 @@ public function getRegion(array $cache)
 
         if ($cache['usage'] === ClassMetadata::CACHE_USAGE_READ_WRITE) {
 
-            if ( ! $this->fileLockRegionDirectory) {
+            if ( null === $this->fileLockRegionDirectory) {
```
Ocramius: Space before should be removed

Ocramius (changes requested):

phansys:
@sensorario, could you please add a test in order to avoid regressions related to this change?

Ocramius (approved):
Looking good! :ship:

PR #6442, opened 2017-05-15
Compute size of loaders in the test suite beforehand, and not on each iteration

sensorario:
It's just a best practice, not really a performance improvement.


Ocramius:
LGTM :+1:

Issue #6449, opened 2017-05-17
[Bug] Reference to an entity is lost after selecting a referenced entity separately


ifif14:
Hello,

I've discovered a following bug today.

I select an entity like so, which has manyToOne relationship to the next entity.

```
$suggest_item =
  Doctrine::em()
  ->getRepository('\DPO\Model\Core\PurchaseOrderItemsSuggest')
  ->findOneBy(
    array(
      'projectid' => $this->projectID,
      'qbpurchaseorderitemsuggestid' => $item->QBItemID
    )
  );

echo ($suggest_item->qbaccount->accountnumber); // echo#1

$accountnumber =
  Doctrine::em()
  ->getRepository('\DPO\Model\Core\AccountNumber')
  ->findOneBy(
    array(
      'projectid' => $this->projectID,
      'qbaccountnumberid' => $item->QBAccountNumberID
    )
  );
echo ($suggest_item->qbaccount->accountnumber); // echo#2
```

When I echo#1 and cause the referenced object to hydrate, everything is okay. 
If I skip the echo#1, then the echo#2 will deliver null.

suggest_item has following relation to account
```
manyToOne:
    qbaccount:
        targetEntity: DPO\Model\Core\AccountNumber
        cascade: {}
        fetch: LAZY
        joinColumns:
            -
                name: ProjectID
                referencedColumnName: ProjectID
            -
                name: QBAccountNumberID
                referencedColumnName: QBAccountNumberID
```
Key of accountnumber looks like follows
```
id:
    qbaccountnumberid:
        type: bigint
        nullable: false
        options:
            unsigned: true
        id: true
        column: QBAccountNumberID
        generator:
            strategy: NONE
    projectid:
        type: bigint
        nullable: false
        options:
            unsigned: true
        id: true
        column: ProjectID
        generator:
            strategy: NONE
```

lcobucci:
@ifif14 could you please try to reproduce that on a test case and send us a PR? You can use the files in  https://github.com/doctrine/doctrine2/tree/334b7e68a7d14443962eff5452f1c7dfe92eff17/tests/Doctrine/Tests/ORM/Functional/Ticket as example.

PR #6453, opened 2017-05-19
Ticket6449Test.php

igorfialko:
So, there is a first test. Here I'm not able to create an DDC6449PurchaseOrderItemsSuggest, due to not recognized changes on object.

Tested with SQLite.


Ocramius:
Linking #6449

PR #6462, opened 2017-05-22
Fix #6460 - \Doctrine\ORM\Mapping\ClassMetadataInfo::hasField should return true for embedded fields

### LINKED ISSUE #6460: Getting embeddable should trigger entity initalization
Getting public embeddables on lazy-loaded entity which is not initialized does not trigger entity initialization.

Common ProxyGenerator in `\Doctrine\Common\Proxy\ProxyGenerator::getLazyLoadedPublicProperties` uses code:
```
if (($class->hasField($name) || $class->hasAssociation($name)) && ! $class->isIdentifier($name)) {
    $properties[$name] = $defaultProperties[$name];
}
```
The interface `\Doctrine\Common\Persistence\Mapping\ClassMetadata` cannot be extended without breaking backward compatibility so ORM has to return `true` for embeddables either in `hasField` or in `hasAssociation`. In my opinion this should be `hasField`

mkosiedowski:
This fixes https://github.com/doctrine/doctrine2/issues/6460

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Embeddable;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\ManyToOne;
+
+class DDC6460Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+
+        try {
+            $this->_schemaTool->createSchema(
+                [
+                    $this->_em->getClassMetadata(DDC6460Entity::class),
+                    $this->_em->getClassMetadata(DDC6460ParentEntity::class),
+                ]
+            );
+        } catch (\Exception $e) {
+        }
+    }
+
+    public function testInlineEmbeddable()
+    {
+        $isFieldMapped = $this->_em
+            ->getClassMetadata(DDC6460Entity::class)
+            ->hasField('embedded');
+
+        $this->assertTrue($isFieldMapped);
+    }
+
+    public function testInlineEmbeddableProxyInitialization()
+    {
+        $entity = new DDC6460Entity();
+        $entity->id = 1;
+        $entity->embedded = new DDC6460Embeddable();
+        $entity->embedded->field = 'test';
+        $this->_em->persist($entity);
+        $this->_em->flush();
+
+        $second = new DDC6460ParentEntity();
+        $second->id = 1;
+        $second->lazyLoaded = $entity;
+        $this->_em->persist($second);
+        $this->_em->flush();
+
+        $this->_em->clear();
+
+        $proxy = $this->_em->getRepository(DDC6460ParentEntity::class)->findOneById(1);
+
+        $this->assertNotNull($proxy->lazyLoaded->embedded);
+    }
+}
+
+/**
+ * @Embeddable()
+ */
+class DDC6460Embeddable
+{
+    /** @Column(type="string") */
+    public $field;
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
```diff
@@ -0,0 +1,100 @@
...
+        parent::setUp();
+
+        try {
+            $this->_schemaTool->createSchema(
+                [
+                    $this->_em->getClassMetadata(DDC6460Entity::class),
+                    $this->_em->getClassMetadata(DDC6460ParentEntity::class),
+                ]
+            );
+        } catch (\Exception $e) {
```
Ocramius: Please only catch schemaexception

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    public function testInlineEmbeddable()
+    {
+        $isFieldMapped = $this->_em
+            ->getClassMetadata(DDC6460Entity::class)
+            ->hasField('embedded');
+
+        $this->assertTrue($isFieldMapped);
```
Ocramius: This test should probably be on the class metadata tests

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
```diff
@@ -0,0 +1,100 @@
...
+        $second->id = 1;
+        $second->lazyLoaded = $entity;
+        $this->_em->persist($second);
+        $this->_em->flush();
+
+        $this->_em->clear();
+
+        $proxy = $this->_em->getRepository(DDC6460ParentEntity::class)->findOneById(1);
+
+        $this->assertNotNull($proxy->lazyLoaded->embedded);
```
Ocramius: Please use `self::assertInstanceOf()`

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
```diff
@@ -0,0 +1,100 @@
...
+        $entity->embedded = new DDC6460Embeddable();
+        $entity->embedded->field = 'test';
+        $this->_em->persist($entity);
+        $this->_em->flush();
+
+        $second = new DDC6460ParentEntity();
+        $second->id = 1;
+        $second->lazyLoaded = $entity;
+        $this->_em->persist($second);
+        $this->_em->flush();
```
Ocramius: Why two flush operations?

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
```diff
@@ -0,0 +1,100 @@
...
+
+        $this->_em->clear();
+
+        $proxy = $this->_em->getRepository(DDC6460ParentEntity::class)->findOneById(1);
```
Ocramius: This is not a proxy here: please add assertions to verify if it is a proxy, and if it is loaded

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
Ocramius: I would also check the value of `field` after this call

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
mkosiedowski: I've added similar test there

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6460Test.php
mkosiedowski: I've added cascade persist in order to remove one flush

Ocramius: Thanks!

Ocramius (approved):

Ocramius:
LGTM! :+1: 

Note that I'm not changing this for `2.5.x`, because there's a subtle behavioral change when this API is used when looping over fields. It will land in `2.6.0`

PR #6466, opened 2017-05-23
Update QueryBuilder docblock to reflect correct type of bound parameter type

fmagnan:
Hello,

thank you very much for your work, it's awesome!
I found a minor error in PHPDOC for method setParameter. Argument "type" can be a string or null but can be also an int if it comes from \PDO::PARAM_*. So I've changed typehint "string" into "mixed".

### DIFF
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -526,7 +526,7 @@ class QueryBuilder
      *
      * @param string|integer $key   The parameter position or name.
      * @param mixed          $value The parameter value.
-     * @param string|null    $type  PDO::PARAM_* or \Doctrine\DBAL\Types\Type::* constant
+     * @param string|integer|null    $type  PDO::PARAM_* or \Doctrine\DBAL\Types\Type::* constant
      *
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -526,7 +526,7 @@ public function getRootEntities()
      *
      * @param string|integer $key   The parameter position or name.
      * @param mixed          $value The parameter value.
-     * @param string|null    $type  PDO::PARAM_* or \Doctrine\DBAL\Types\Type::* constant
+     * @param mixed|null    $type  PDO::PARAM_* or \Doctrine\DBAL\Types\Type::* constant
```
lcobucci: Can you use `string|integer|null` instead? `mixed` is a bit too broad

lcobucci (changes requested):
@fmagnan thanks for your contribution! Just a small change 😄

fmagnan (commented):
done

lcobucci (approved):

lcobucci:
@fmagnan awesome! 🚢

Issue #6470, opened 2017-05-25
OneToOne second level cache load


shustrik:
```php
/**
 * @ORM\Entity()
  */
class Merchant
{
    /**
     * @var integer
     *
     * @ORM\Id
     * @ORM\Column(name="id", type="integer")
     * @ORM\GeneratedValue(strategy="IDENTITY")
     */
    protected $id;

    /**
     * @var Manager
     *
     * @ORM\OneToOne(targetEntity="AppBundle\Entity\Manager", mappedBy="merchant")
     * @ORM\Cache(usage="NONSTRICT_READ_WRITE", region="my_region")
     */
    protected $manager;

    /**
     * @var string
     *
     * @ORM\Column(name="name", type="string", length=255, nullable=false)
     */
    protected $name;
````
and

```php
/**
 * @ORM\Entity()
 * @ORM\Cache(usage="NONSTRICT_READ_WRITE", region="my_region")
 */
class Manager extends User
{

    /**
     * @var string
     *
     * @ORM\Column()
     */
    protected $username;

    /**
     * @var Merchant
     *
     * @ORM\OneToOne(targetEntity="AppBundle\Entity\Merchant", inversedBy="manager")
     */
    protected $merchant;
```

Manager from merchant is always loaded from database. I guess that the problem is in Doctrine\ORM\Cache\Persister\Entity\AbstractEntityPersister:loadOneToOneEntity:  only persister which gets objects from database is called in this method, but  loading from cache  is missing.

lcobucci:
@shustrik can you please reproduce this in a functional test case (based on `master`) and send us a PR with the failing test? You can find examples on https://github.com/doctrine/doctrine2/tree/2a239be45ec4e64c736e4277cde8bff376ef2402/tests/Doctrine/Tests/ORM/Functional/Ticket

shustrik:
@lcobucci test added

martixy:
I think I have the same problem. I've got `User (non-owning) <-> A bunch of extensions(e.g. customer, designer, etc)`.

User will not cache. Even if I mark all of the associations and owning entities.
Amusingly, it also creates a situation where caching makes performance _worse_ - when you have a cache hit for the owning entities, even if you make the query with eager fetch, it will retrieve the cached super entities, but then have to go and fetch the 1-1s individually for each cache hit.
Or rather, it will fetch them individually, until the first cache miss, then it will make the eager fetch query, loading all those associations eagerly.

lcobucci:
[...]

PR #6473, opened 2017-05-26
issue #6470

shustrik:
Tests for issue #6470

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6470Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6470Test.php
@@ -1,0 +1,181 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Tools\ToolsException;
+
+/**
+ */
+class DDC6470Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    public function setUp()
+    {
+        $this->enableSecondLevelCache();
+        parent::setUp();
+
+        try {
+            $this->_schemaTool->createSchema(
+                [
+                    $this->_em->getClassMetadata(DDC6470User::class),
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6470Test.php
```diff
@@ -0,0 +1,181 @@
...
+        $driver1 = new DDC6470Driver('Bar');
+
+        $this->_em->persist($driver1);
+        $this->_em->flush();
+        $this->_em->clear();
+
+
+        $this->assertTrue($this->_em->getCache()->containsEntity(DDC6470User::class, ['id' => $driver1->getUserProfile()->getId()]));
+        $queryCount = $this->getCurrentQueryCount();
+        $driver3 = $this->_em->createQueryBuilder()
```
lcobucci: @shustrik the problem with this test is that your query is not cacheable, therefore it won't load data from L2C.

Your test pass if you call `QueryBuilder#setCacheable(true)` or if you just use `$driver3 = $this->_em->find(DDC6470Driver::class, $driver1->getId());` here.

lcobucci:
@shustrik sorry about my delay to get back to you. As mentioned on the comment this is an invalid scenario and I'll be closing the PR as `invalid`.

wtorsi:
I've created pull request https://github.com/doctrine/doctrine2/pull/7149 , but actually it's not my problem, it's just a small bug in the 'warmup' state.

In my case, some of the OneToOne relations are not put into cache properly in 'warmup' state. 
After executing query some of the OneToOne fields are not stored in cache (according to my test, `target` is never stored). That's why all other requests to the same resource make one request to cache, and N requests to db for missed relations. 

According to my test, all my requests act like first QueryBuilder, but must act like the second.
Really hard to reproduce this error in tests. But i'll try

Issue #6464, opened 2017-05-22
Query builder yields invalid SQL with JOINED inheritance type on PostgreSQL


stesie:
Given a simple set of entities (adapted to a simpler domain here):

* a `Post` belongs to an `Author` which derives from `User`
* the inheritance is mapped by JOINED type
* yet the relation between `Post` and `Author` is not mapped (for reasons, as the real case is less simple)

```php
<?php
namespace Entity;

/**
 * @Entity @Table(name="posts")
 */
class Post
{
    /** @Id @GeneratedValue @Column(type="integer") */
    public $id;

    /** @Column(type="integer") */
    public $authorId;

    /** @Column(length=100) */
    public $title;

    /** @Column(type="text") */
    public $text;
}
```

```php
<?php
namespace Entity;

/**
 * @Entity @Table(name="authors")
 */
class Author extends User
{
    /** @Column(length=50) */
    public $displayName;
}
```

```php
<?php
namespace Entity;

/**
 * @Entity @Table(name="users")
 * @InheritanceType("JOINED")
 * @DiscriminatorColumn(name="discr", type="string")
 * @DiscriminatorMap({"author" = "Author"})
 */
abstract class User
{
    /** @Id @GeneratedValue @Column(type="integer") */
    public $id;

    /** @Column(length=50) */
    public $username;

    /** @Column(length=100) */
    public $password;
}
```

If I then instanciate and use a QueryBuilder like this:

```php
$qb = $entityManager->createQueryBuilder();

$qb
    ->select('p', 'a')
    ->from('Entity\\Post', 'p')
    ->innerJoin('Entity\\Author', 'a', 'WITH', 'p.authorId = a.id')
    ;
```

it generates the following SQL

```sql
SELECT ... FROM posts p0_ INNER JOIN authors a2_ INNER JOIN users u1_ ON a2_.id = u1_.id AND (p0_.authorId = u1_.id)
```

Which is valid for MySQL, however is *not* valid for PostgreSQL (tried with 9.6)

PostgreSQL seems to require an `ON` clause with each and every `INNER JOIN` (interestingly as opposed to `LEFT JOIN`). This is, the following query works with it:

```sql
SELECT ... FROM posts p0_ INNER JOIN users u1_ ON (p0_.authorId = u1_.id) INNER JOIN authors a2_ ON (a2_.id = u1_.id) ;
```

Pushing all conditions to `WHERE` also works on PostgreSQL:

```sql
[...]

PR #6475, opened 2017-05-27
Fix invalid sql generated for CTI JOINs when INNER JOIN'ing with a `WITH` clause

stesie:
See #6464 for a detailed example of the problem.

For a query like 

```php
$qb
            ->select('p', 'a')
            ->from(GH6464Post::class, 'p')
            ->innerJoin(GH6464Author::class, 'a', 'WITH', 'p.authorId = a.id')
            ->getQuery();
```

... where `GH6464Author` is a CTI entity, the sql generation currently generates two JOINs (correct) and attaches both join conditions to the latter JOIN. This works with MySQL and SQLite, however fails on PostgreSQL.

This patch

* adds a functional test case that actually tries to run such a query
* applies a fix to `SqlWalker` that makes it output a nested JOIN
* adapts test cases in `SelectSqlGenerationTest` to reflect the change (and make those SQL statements valid -- on PostgreSQL)

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -872,11 +872,23 @@ class SqlWalker implements TreeWalker
      * Walks down a RangeVariableDeclaration AST node, thereby generating the appropriate SQL.
      *
      * @param AST\RangeVariableDeclaration $rangeVariableDeclaration
+     *
+     * @return string
+     */
+    public function walkRangeVariableDeclaration($rangeVariableDeclaration)
+    {
+        return $this->generateRangeVariableDeclarationSQL($rangeVariableDeclaration, false);
+    }
+
+    /**
+     * Generate appropriate SQL for RangeVariableDeclaration AST node
+     *
+     * @param AST\RangeVariableDeclaration $rangeVariableDeclaration
      * @param bool $buildNestedJoins
      *
      * @return string
      */
-    public function walkRangeVariableDeclaration($rangeVariableDeclaration, $buildNestedJoins = false)
+    private function generateRangeVariableDeclarationSQL($rangeVariableDeclaration, $buildNestedJoins)
     {
         $class    = $this->em->getClassMetadata($rangeVariableDeclaration->abstractSchemaName);
         $dqlAlias = $rangeVariableDeclaration->aliasIdentificationVariable;
@@ -1132,7 +1144,7 @@ class SqlWalker implements TreeWalker
             : ' INNER JOIN ';
 
         switch (true) {
-            case ($joinDeclaration instanceof \Doctrine\ORM\Query\AST\RangeVariableDeclaration):
+            case ($joinDeclaration instanceof AST\RangeVariableDeclaration):
                 $class      = $this->em->getClassMetadata($joinDeclaration->abstractSchemaName);
                 $dqlAlias   = $joinDeclaration->aliasIdentificationVariable;
                 $tableAlias = $this->getSQLTableAlias($class->table['name'], $dqlAlias);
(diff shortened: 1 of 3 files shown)


stesie:
Test fails seem to be related to changes to `symfony/yaml` and unrelated to this change (as tests are green on PHP 7.0 and on 7.1 with low dependencies)

Majkl578:
Can you please rebase this PR onto current master? It should fix build errors on Travis. 👍

stesie:
@Majkl578 rebased, tests are green now

... but scrutinizer fails now. Yet the only issue now caused by my additions (https://scrutinizer-ci.com/g/doctrine/doctrine2/inspections/de206627-1966-4ae5-9710-e883754a9c58/issues/files/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6464Test.php) is a false positive as `getSql` returns a `string` (as opposed to reported `array`) ...

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -872,10 +872,11 @@ public function walkIndexBy($indexBy)
      * Walks down a RangeVariableDeclaration AST node, thereby generating the appropriate SQL.
      *
      * @param AST\RangeVariableDeclaration $rangeVariableDeclaration
+     * @param bool $buildNestedJoins
      *
      * @return string
      */
-    public function walkRangeVariableDeclaration($rangeVariableDeclaration)
+    public function walkRangeVariableDeclaration($rangeVariableDeclaration, $buildNestedJoins = false)
```
Ocramius: The additional parameter is problematic. The decision is one in the `SqlWalker` anyway, so the correct solution (IMO) would be to move the method body to a private method, then keep the public API as-is and delegate the call on the public API to the private API, where the new parameter is defined.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6464Test.php
```diff
@@ -0,0 +1,89 @@
...
+/** @Entity */
+class GH6464Post
+{
+    /** @Id @Column(type="integer") @GeneratedValue */
+    public $id;
+
+    /** @Column(type="integer") */
+    public $authorId;
+
+    /** @Column(length=100) */
```
Ocramius: Drop any details that aren't strictly needed. If columns aren't affected, they shouldn't be in here

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6464Test.php
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Verifies that SqlWalker generates valid SQL for an INNER JOIN to CTI table
+     *
+     * SqlWalker needs to generate nested INNER JOIN statements, otherwise there would be INNER JOIN
+     * statements without an ON clause, which are valid on e.g. MySQL but rejected by PostgreSQL.
+     */
+    public function testIssue()
+    {
+
+        $qb = $this->_em->createQueryBuilder();
```
Ocramius: Please either inline with the code below or use a DQL string instead of the builder

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -891,7 +892,11 @@ public function walkRangeVariableDeclaration($rangeVariableDeclaration)
         );
 
         if ($class->isInheritanceTypeJoined()) {
```
Ocramius: Can this block be moved to a private method too?

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -1137,11 +1142,11 @@ public function walkJoin($join)
                     $conditions[] = '(' . $this->walkConditionalExpression($join->conditionalExpression) . ')';
                 }
 
-                $condExprConjunction = ($class->isInheritanceTypeJoined() && $joinType != AST\Join::JOIN_TYPE_LEFT && $joinType != AST\Join::JOIN_TYPE_LEFTOUTER)
+                $condExprConjunction = ($class->isInheritanceTypeJoined() && $joinType != AST\Join::JOIN_TYPE_LEFT && $joinType != AST\Join::JOIN_TYPE_LEFTOUTER && empty($conditions))
```
Ocramius: The `empty($conditions)` is used twice - can you give it a name and assign it to a meaningful variable?

Ocramius (changes requested):

lib/Doctrine/ORM/Query/SqlWalker.php
stesie: I doubt that aids readability since the decision is made by the caller (actually even that one's caller) and not the extracted function itself ... and it just modifies local state (i.e. `$sql`).

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6464Test.php
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Verifies that SqlWalker generates valid SQL for an INNER JOIN to CTI table
+     *
+     * SqlWalker needs to generate nested INNER JOIN statements, otherwise there would be INNER JOIN
+     * statements without an ON clause, which are valid on e.g. MySQL but rejected by PostgreSQL.
+     */
+    public function testIssue()
+    {
+        $query = $this->_em->createQueryBuilder()
+            ->select('p', 'a')
```
Ocramius: This sort of `select` is invalid - removing `a` from the selection, as you should *NEVER* fetch-join results over an association that was filtered.

Ocramius:
Manually merged, thanks @stesie!

`master`: doctrine/doctrine2@6d428c90e24c914c1e5d817d74646f3b94757c43
`2.5`: doctrine/doctrine2@d89d238594621220c8a18e8c1abf224e7e967ac3

ivok:
Why this issue is not fixed yet?

For our new project we use Doctrine 2 and MSSQL and with first join query we encountered the same problem.

When I dump the sql build from Doctrine it wraps the joined table name with parenthesis "**JOIN (table tb) ON**...." which gives error for MSSQL

PR #6476, opened 2017-05-28
Allow arithmetic expressions within IN operator

astepin:
In some cases it is necessary to use not only scalar values or subselects within IN-conditions, but also own functions. 

With Oracle databases, the following is then possible:

DQL:
`SELECT u FROM User WHERE u.name IN ( CUSTOM_DQL_FUNCTION() );`

SQL:
`SELECT * FROM user WHERE name IN ( SELECT * FROM TABLE(ORACLE_PLSQL_FUNCTION()) );`

Another example which is also possible now:

DQL:
`[...] WHERE field IN ( CUSTOM_DQL_GET_ID('Foo'), CUSTOM_ESCAPE('Bar') )`

SQL:
`[...] WHERE field IN ( (SELECT id FROM settings WHERE name = 'Foo'), [...] )`

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -2759,7 +2759,7 @@ class Parser
     }
 
     /**
-     * InParameter ::= Literal | InputParameter
+     * InParameter ::= ArithmeticExpression | InputParameter
      *
      * @return AST\InputParameter|AST\Literal
      */
@@ -2769,7 +2769,7 @@ class Parser
             return $this->InputParameter();
         }
 
-        return $this->Literal();
+        return $this->ArithmeticExpression();
     }
 
     /**
--- a/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
@@ -643,6 +643,17 @@ class SelectSqlGenerationTest extends OrmTestCase
         );
     }
 
+    public function testCustomFunctionWithinInExpression() : void
+    {
+        $this->entityManager->getConfiguration()->addCustomStringFunction('FOO', MyAbsFunction::class);
+        $this->assertSqlGeneration(
+            "SELECT u FROM Doctrine\Tests\Models\Forum\ForumUser u WHERE u.username IN (FOO('Lo'), 'Lo', :name)",
+            <<<'SQL'
+SELECT f0_.id AS id_0, f0_.username AS username_1 FROM forum_users f0_ WHERE f0_.username IN (ABS('Lo'), 'Lo', ?)
+SQL
+        );
+    }
+
     public function testSupportsConcatFunctionForMysqlAndPostgresql(): void
     {
         $connMock    = $this->entityManager->getConnection();
--- a/docs/en/reference/dql-doctrine-query-language.rst
+++ b/docs/en/reference/dql-doctrine-query-language.rst
@@ -1712,7 +1712,7 @@ Literal Values
 .. code-block:: php
 
     Literal     ::= string | char | integer | float | boolean
-    InParameter ::= Literal | InputParameter
+    InParameter ::= ArithmeticExpression | InputParameter
 
 Input Parameter
 ~~~~~~~~~~~~~~~
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -2106,7 +2106,7 @@ class SqlWalker implements TreeWalker
     {
         return $inParam instanceof AST\InputParameter
             ? $this->walkInputParameter($inParam)
-            : $this->walkLiteral($inParam);
+            : $this->walkArithmeticExpression($inParam);
     }
 
     /**


Ocramius:
The idea sounds sane, but how does it behave with the "modify limit" logic in the paginator?

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -2694,7 +2694,7 @@ public function InParameter()
             return $this->InputParameter();
         }
 
-        return $this->Literal();
+        return $this->ArithmeticExpression();
```
Ocramius: Documentation changes in https://github.com/doctrine/doctrine2/blob/22ecc2d58cdf1e8a9b932d99b952c94afae7e873/docs/en/reference/dql-doctrine-query-language.rst#ebnf are also needed

Ocramius (changes requested):

lib/Doctrine/ORM/Query/Parser.php
astepin: I've updated the affected docs.

astepin:
As far as i know, does the limit clause not affect any where conditions at all. I don't see any conflicts for the pagination.

lib/Doctrine/ORM/Query/Parser.php
Ocramius: Thanks!

tests/Doctrine/Tests/ORM/Functional/CustomFunctionsTest.php
```diff
@@ -64,6 +64,31 @@ public function testCustomFunctionOverride()
...
+
+        $this->_em->getConfiguration()->addCustomStringFunction(
+            'FOO',
+            function (string $funcName) : NoOp {
+                return new NoOp($funcName);
+            }
+        );
+
+        $query = 'SELECT u FROM Doctrine\Tests\Models\CMS\CmsUser u WHERE u.name IN (FOO(\'Lo\'), \'Lo\', :name)';
+        $users = $this->_em->createQuery($query)
```
Ocramius: This test should be changed to an SQL generation test, in my opinion. Persistence logic is not needed here

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/CustomFunctionsTest.php
astepin: @Ocramius Can you provide me an example where I can look at? I'm not quite sure what you mean here.

Ocramius: See https://github.com/doctrine/doctrine2/blob/6a861756176a6bfcfe5c4ecb699e3b08c7582cfb/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php

ostrolucky:
@astepin are you interested in finishing this?

astepin:
@ostrolucky I've just fixed the last comments from the review.

Ocramius:
@astepin please rebase rather than merging `master`

astepin:
@Ocramius Slow and steady wins the race ... PR is rebased now

ostrolucky (dismissed):

SenseException (dismissed):

webda2l:
PR welcome :+1:

astepin:
I would like to continue working on this feature. What do I need to do so that it can continue?

greg0ire:
Hi! You should rebase on 2.11.x :)

derrabus:
Feature development happens on the 2.11.x branch currently. Can you try to rebase against that branch?

greg0ire:
`git rebase --onto origin/2.11.x HEAD^^ arithmetic-in-expression` should do the trick.

greg0ire:
It's quite complicated, so I'm going to do it for you.

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -2759,7 +2759,7 @@ public function Literal()
     }
 
     /**
-     * InParameter ::= Literal | InputParameter
+     * InParameter ::= ArithmeticExpression | InputParameter
      *
      * @return AST\InputParameter|AST\Literal
```
greg0ire, line 2764: As PHPStan pointed out, This line needs to be updated

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
```diff
@@ -643,6 +643,17 @@ public function testInvalidInExpressionWithSingleValuedAssociationPathExpression
         );
     }
 
+    public function testCustomFunctionWithinInExpression() : void
+    {
+        $this->entityManager->getConfiguration()->addCustomStringFunction('FOO', MyAbsFunction::class);
+        $this->assertSqlGeneration(
+            "SELECT u FROM Doctrine\Tests\Models\Forum\ForumUser u WHERE u.username IN (FOO('Lo'), 'Lo', :name)",
```
greg0ire, line 650: Why are you using strings in this example? It does not make a lot of sense to use an arithmetic function on a string, does it? Can you maybe write something closer to the real-world use-case you have?

greg0ire: You can disregard my comment if that's too hard: I just read the EBNF for https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/dql-doctrine-query-language.html#arithmetic-expressions and now I understand it's not just about `2+2` :slightly_smiling_face:

astepin: In my use case, the scenario looks like this: I have a database platform specific function that I want to use like `NVL2` from Oracle. For this I have created a user defined function for DQL in Doctrine and now I want to use this also for IN where conditions.

greg0ire: OK. To me it's kind of weird that `ArithmeticPrimary` contains that many things, for instance `FunctionsReturningStrings`. I don't think this has much to do with arithmetic, but anyway, that's not really your issue.

lib/Doctrine/ORM/Query/Parser.php
webda2l: @astepin You asked how continue this PR https://github.com/doctrine/orm/pull/6476#issuecomment-934618022, there is an awaited commit fix here if you want.
See https://github.com/doctrine/orm/pull/6476/checks?check_run_id=3806327129#step:6:30.

`Error: Method Doctrine\ORM\Query\Parser::InParameter() should return Doctrine\ORM\Query\AST\InputParameter|Doctrine\ORM\Query\AST\Literal but returns Doctrine\ORM\Query\AST\ArithmeticExpression.
`

so, `@return AST\ArithmeticExpression`

webda2l:
@astepin so you have time to update this PR (see https://github.com/doctrine/orm/pull/6476#discussion_r729626160) or I do another PR ? Thanks

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
beberlei: You could also add an arithmetic expression, it should not work before to do `IN (1+1)`

astepin:
> @astepin so you have time to update this PR (see [#6476 (comment)](https://github.com/doctrine/orm/pull/6476#discussion_r729626160)) or I do another PR ? Thanks

Hey @webda2l, I will not be able to finish it this year. If you want to take it over, it's okay with me.

derrabus:
Closing in favor of #9242.

PR #6485, opened 2017-06-05
#5804 Versioning doesn't work with a custom type primary key

### LINKED ISSUE #5804: Versioning doesn't work with a custom type primary key
I've been wondering for a while why does Doctrine resets my `@Version` entity field to `0` every time I flush to the database even though the correct version can be found in the database, so I decided to dig down into the entrails of Doctrine and found some peculiar issue.

When the execution flow reaches the `Doctrine\ORM\Persisters\Entity\BasicEntityPersister::fetchVersionValue` method (it's the point after the flush where Doctrine tries to retrieve the new version of the entity from the database to re-hydrate the entity) apparently it assumes that the primary key for this table is a known primitive type and doesn't bother trying to map it to anything. More specifically this line:

``` php
// Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php:346
$value = $this->conn->fetchColumn($sql, array_values($flatId));
```

The third and fourth parameters of `fetchColumn` should be `0` and the type(s) of the `$flatId` field(s). But nothing is sent. In my situation, my primary key is a custom type `uuid_binary`. Then down the line when you get to `Doctrine\DBAL\Connection::executeQuery` and the statement is being prepared, since there is no types involved in the `$types` parameter, the binding actually never happens!

``` php
// Doctrine/DBAL/Driver/PDOStatement.php:825
$stmt = $this->_conn->prepare($query);
if ($types) {
    $this->_bindTypedValues($stmt, $params, $types);
    $stmt->execute();
} else {
    $stmt->execute($params);
}
```

Which means that the versioning query...

``` sql
SELECT version FROM tablename WHERE primarykey = ?
```

... doesn't ever receive any value for the `primarykey` placeholder and the request ultimately fails and returns `false`. Doctrine then takes this `false` and assigns it to the `@Version` field of the entity in `Doctrine\ORM\Persisters\Entity\BasicEntityPersister::assignDefaultVersionValue`, which then translates to `0`.

``` php
// Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php:317
protected function assignDefaultVersionValue($entity, array $id)
{
    $value = $this->fetchVersionValue($this->class, $id);
    $this->class->setFieldValue($entity, $this->class->versionField, $value);
}
```

So it seems like this scenario is entirely missing.

alle:
This PR fixes #5804.
It's not a my own solution, but the one provided by @ngrevet-lightspeed.
Since I need this fixed, I made this PR on behalf of @ngrevet-lightspeed.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -346,9 +346,7 @@ class BasicEntityPersister implements EntityPersister
 
         $types = [];
         foreach ($id as $field => $value) {
-            foreach ($this->getTypes($field, $value, $versionedClass) as $type) {
-                $types[] = $type;
-            }
+            $types = array_merge($types, $this->getTypes($field, $value, $versionedClass));
         }
 
         $flatId = $this->identifierFlattener->flattenIdentifier($versionedClass, $id);
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -344,9 +344,13 @@ class BasicEntityPersister implements EntityPersister
              . ' FROM '  . $tableName
              . ' WHERE ' . implode(' = ? AND ', $identifier) . ' = ?';
 
+        $types = [];
+        foreach ($identifier as $fieldName) {
+            $types[] = $versionedClass->fieldMappings[$fieldName]['type'];
+        }
         $flatId = $this->identifierFlattener->flattenIdentifier($versionedClass, $id);
 
-        $value = $this->conn->fetchColumn($sql, array_values($flatId));
+        $value = $this->conn->fetchColumn($sql, array_values($flatId), 0, $types);
 
         return Type::getType($fieldMapping['type'])->convertToPHPValue($value, $this->platform);
     }
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -345,9 +345,12 @@ class BasicEntityPersister implements EntityPersister
              . ' WHERE ' . implode(' = ? AND ', $identifier) . ' = ?';
 
         $types = [];
-        foreach ($identifier as $fieldName) {
-            $types[] = $versionedClass->fieldMappings[$fieldName]['type'];
+        foreach ($id as $field => $value) {
+            foreach ($this->getTypes($field, $value, $versionedClass) as $type) {
+                $types[] = $type;
+            }
         }
+
         $flatId = $this->identifierFlattener->flattenIdentifier($versionedClass, $id);
 
         $value = $this->conn->fetchColumn($sql, array_values($flatId), 0, $types);
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -344,16 +344,28 @@ class BasicEntityPersister implements EntityPersister
              . ' FROM '  . $tableName
              . ' WHERE ' . implode(' = ? AND ', $identifier) . ' = ?';
 
+
+        $flatId = $this->identifierFlattener->flattenIdentifier($versionedClass, $id);
+
+        $value = $this->conn->fetchColumn(
+            $sql,
+            array_values($flatId),
+            0,
+            $this->extractIdentifierTypes($id, $versionedClass)
+        );
+
+        return Type::getType($fieldMapping['type'])->convertToPHPValue($value, $this->platform);
+    }
+
+    private function extractIdentifierTypes(array $id, ClassMetadata $versionedClass) : array
+    {
         $types = [];
+
         foreach ($id as $field => $value) {
             $types = array_merge($types, $this->getTypes($field, $value, $versionedClass));
         }
 
-        $flatId = $this->identifierFlattener->flattenIdentifier($versionedClass, $id);
-
-        $value = $this->conn->fetchColumn($sql, array_values($flatId), 0, $types);
-
-        return Type::getType($fieldMapping['type'])->convertToPHPValue($value, $this->platform);
+        return $types;
     }
 
     /**
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5804Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5804Test.php
@@ -84,7 +84,6 @@ final class GH5804Type extends Type
 
 /**
  * @Entity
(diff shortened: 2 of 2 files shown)


Ocramius (changes requested):
@alle the patch needs tests to be validated

alle:
@Ocramius @ngrevet-lightspeed Everything should be fine, now. I was too hurried, last time. Sorry.

Ocramius:
@alle I still don't see test changes though :-\

alle:
@Ocramius Well, I thought tests shouldn't change, should they?
I thought I fixed this bug, working on the `BasicEntityPersister::fetchVersionValue()` method, by passing the involved tests, without changing them. Humbly, am I missing something?

Ocramius:
@alle a test demonstrating the original bug (failing with the original repository state, passing with this patch) is needed ;-)

alle:
@ocramius: @astepin did the job (in #6496) that I should have done from myself.
@astepin The only change I made is in the custom type `getSQLDeclaration()` method: your version was failing against the Postgres platform, since Doctrine maps to a specific datatype for that platform.

alle:
I changed the involved classes in tests, complying to other tests' structure, in order to avoid a merge conflict on the `SchemaValidatorTest`, due to recent updates

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -344,9 +344,16 @@ protected function fetchVersionValue($versionedClass, array $id)
              . ' FROM '  . $tableName
              . ' WHERE ' . implode(' = ? AND ', $identifier) . ' = ?';
 
+        $types = [];
+        foreach ($id as $field => $value) {
+            foreach ($this->getTypes($field, $value, $versionedClass) as $type) {
```
lcobucci: Do we need to iterate this too? I mean in theory we could use `array_merge()` right, like:

```php
$types = [];

foreach ($id as $field => $value) {
    $types = array_merge($types, $this->getTypes($field, $value, $versionedClass));
}
```

alle: @lcobucci You're simply right.

alle:
@Ocramius @lcobucci Is there any possibility to have this PR merged? I didn't see any news since the last changes.

tests/Doctrine/Tests/Models/GH5804/GH5804Article.php
```diff
@@ -0,0 +1,29 @@
...
+ * @Table(name="gh5804_articles")
+ */
+class GH5804Article
+{
+    /**
+     * @Id
+     * @Column(type="GH5804Type")
+     * @GeneratedValue(strategy="CUSTOM")
+     * @CustomIdGenerator(class="\Doctrine\Tests\Models\GH5804\GH5804Generator")
+    ")
```
alcaeus: Syntax error ;)

tests/Doctrine/Tests/Models/GH5804/GH5804Article.php
```diff
@@ -0,0 +1,29 @@
...
+ * @Entity
+ * @Table(name="gh5804_articles")
+ */
+class GH5804Article
+{
+    /**
+     * @Id
+     * @Column(type="GH5804Type")
+     * @GeneratedValue(strategy="CUSTOM")
+     * @CustomIdGenerator(class="\Doctrine\Tests\Models\GH5804\GH5804Generator")
```
alcaeus: Could/Should use `::class` constant

tests/Doctrine/Tests/Models/GH5804/GH5804Article.php
alle: Aaargh! Didn't see that!

tests/Doctrine/Tests/Models/GH5804/GH5804Article.php
alle: In which form? Any example?

alcaeus: Paired with a `use` statement: `@CustomIdGenerator(class=GH5804Generator::class)`. Works with Annotations >= 1.2 IIRC.

alle: I guessed it was that way, but my local tests fail. I even update merged with the latest `master` release, updated the dependencies (my Annotations version was right, anyway), but your suggestion (which seems to be logical) doesn't work to me. Did you try by yourself?
Anyway, even without that `::class` notation, everything is fine.

alcaeus: That's strange. I've been using that for a couple of years now.

lcobucci:
I'll take a look at this in a bit

tests/Doctrine/Tests/Models/GH5804/GH5804Article.php
lcobucci: It requires the FQCN, just that 😉

lcobucci:
@alle thanks for your contribution, I've simplified things a bit (kept all test related classes together).

@Ocramius it LGTM, can you check this one to see if we're good to go?

lcobucci (approved):

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -344,9 +344,14 @@ protected function fetchVersionValue($versionedClass, array $id)
              . ' FROM '  . $tableName
              . ' WHERE ' . implode(' = ? AND ', $identifier) . ' = ?';
 
+        $types = [];
```
Ocramius: Can this block be moved to a private method?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH5804Test.php
```diff
@@ -0,0 +1,108 @@
...
+class GH5804Article
+{
+    /**
+     * @Id
+     * @Column(type="GH5804Type")
+     * @GeneratedValue(strategy="CUSTOM")
+     * @CustomIdGenerator(class=\Doctrine\Tests\ORM\Functional\Ticket\GH5804Generator::class)
+     */
+    public $id;
+    /**
```
Ocramius: Spacing

tests/Doctrine/Tests/ORM/Functional/Ticket/GH5804Test.php
```diff
@@ -0,0 +1,108 @@
...
+            return null;
+        }
+
+        return 'testGh5804DbValue';
+    }
+}
+
+/**
+ * @Entity
+ * @Table(name="gh5804_articles")
```
Ocramius: Not really needed

Ocramius (approved):
Some nitpicks apply

lcobucci:
@alle 🚢 thanks!

alle:
Thank you so much, guys!
This was my first opportunity to contribute to this project, even to "calibrate" myself to future contributions. Or at least I hope so!

PR #6487, opened 2017-06-06
Add failing test for DDC-3460

flack:
as discussed in #4270 here's my first stab at a failing test.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\SchemaTool;
+
+use Doctrine\ORM\Tools;
+
+class DDC3460Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    /**
+     * @group DDC-3460
+     */
+    public function testDetectMyISAM()
+    {
+        $conn = $this->_em->getConnection();
+
+        if (strpos($conn->getDriver()->getName(), "mysql") === false) {
+            $this->markTestSkipped('this test is only relevant for MySQL');
+        }
+
+        $conn->exec('CREATE TABLE `tweet_user` (id INT NOT NULL AUTO_INCREMENT, name VARCHAR(255) NOT NULL, PRIMARY KEY(id)) ENGINE = MyISAM');
+        $conn->exec('CREATE TABLE `tweet_tweet` (id INT NOT NULL AUTO_INCREMENT, content VARCHAR(255) NOT NULL, author_id INT, PRIMARY KEY(id)) ENGINE = MyISAM');
+
+        $fromSchema = $conn->getSchemaManager()->createSchema();
+
+        $classMetadata = [
+            $this->_em->getClassMetadata('Doctrine\Tests\Models\Tweet\Tweet'),
+            $this->_em->getClassMetadata('Doctrine\Tests\Models\Tweet\User')
+        ];
+        $schemaTool = new Tools\SchemaTool($this->_em);
+        $toSchema = $schemaTool->getSchemaFromMetadata($classMetadata);
+
+        $comparator = new \Doctrine\DBAL\Schema\Comparator();
+        $schemaDiff = $comparator->compare($fromSchema, $toSchema);
+        $this->assertEquals([], $schemaDiff->toSql($conn->getDatabasePlatform()));
+    }
+}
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function testDetectMyISAM()
+    {
+        $conn = $this->_em->getConnection();
+
+        if (strpos($conn->getDriver()->getName(), "mysql") === false) {
+            $this->markTestSkipped('this test is only relevant for MySQL');
+        }
+
+        $conn->exec('CREATE TABLE `tweet_user` (id INT NOT NULL AUTO_INCREMENT, name VARCHAR(255) NOT NULL, PRIMARY KEY(id)) ENGINE = MyISAM');
```
Ocramius: Add two `DROP TABLE IF EXISTS` before these statements

tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
```diff
@@ -0,0 +1,36 @@
...
+        $conn->exec('CREATE TABLE `tweet_tweet` (id INT NOT NULL AUTO_INCREMENT, content VARCHAR(255) NOT NULL, author_id INT, PRIMARY KEY(id)) ENGINE = MyISAM');
+
+        $fromSchema = $conn->getSchemaManager()->createSchema();
+
+        $classMetadata = [
+            $this->_em->getClassMetadata('Doctrine\Tests\Models\Tweet\Tweet'),
```
Ocramius: `::class`

tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
Ocramius: These models are already in use somewhere else - this makes the test a bit fragile...

tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
```diff
@@ -0,0 +1,36 @@
...
+            $this->_em->getClassMetadata('Doctrine\Tests\Models\Tweet\User')
```
Ocramius: `::class`

tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
```diff
@@ -0,0 +1,36 @@
...
+        ];
+        $schemaTool = new Tools\SchemaTool($this->_em);
+        $toSchema = $schemaTool->getSchemaFromMetadata($classMetadata);
+
+        $comparator = new \Doctrine\DBAL\Schema\Comparator();
+        $schemaDiff = $comparator->compare($fromSchema, $toSchema);
+        $this->assertEquals([], $schemaDiff->toSql($conn->getDatabasePlatform()));
```
Ocramius: Add a message to the assertion. Also, try checking the existing schema's `engine` attribute on the tables

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
flack: I was kind of hoping that this might get backported to 2.5 eventually, so I refrained from using PHP5.5-only stuff. But I can change it ofc

tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
flack: Yes, I was lazy :-) I can make new dedicated models for this or I can add a `tearDown` method which drops the tables, if you prefer

tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
Ocramius: Do make a `master`-compatible one - backporting is up to us :-)

tests/Doctrine/Tests/ORM/Functional/SchemaTool/DDC3460Test.php
flack: would this be the correct way to do it? 

```php
$fromSchema->getTable('tweet_user')->getOption('engine')
```

I just tried and it currentyl throws a notice "Undefined index: engine". But that's probably what this test is about :-)

Ocramius: Yes, I think that would be correct.

Ocramius: Well, `self::assertSame('MyISAM', $fromSchema->getTable('tweet_user')->getOption('engine'))`

flack:
Puhsed another commit which I think addresses all your points. I've done the `engine` check a bit differently so that it doesn't throw a notice, but still fails. Let me know if I should change something else

PR #6491, opened 2017-06-08
Fix Mock Implementations.

ThePixelDeveloper:
The implementations were not using the correct function signatures. Caused some failures, but easily resolved.


Ocramius (approved):
@ThePixelDeveloper LGTM, thanks!

PR #6496, opened 2017-06-09
#5804 #6485 Versioning doesn't work with a custom type primary key

### LINKED ISSUE #5804: Versioning doesn't work with a custom type primary key

astepin:
This PR is based on #6485. I've just added a test.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -344,9 +344,13 @@ class BasicEntityPersister implements EntityPersister
              . ' FROM '  . $tableName
              . ' WHERE ' . implode(' = ? AND ', $identifier) . ' = ?';
 
+        $types = [];
+        foreach ($identifier as $fieldName) {
+            $types[] = $versionedClass->fieldMappings[$fieldName]['type'];
+        }
         $flatId = $this->identifierFlattener->flattenIdentifier($versionedClass, $id);
 
-        $value = $this->conn->fetchColumn($sql, array_values($flatId));
+        $value = $this->conn->fetchColumn($sql, array_values($flatId), 0, $types);
 
         return Type::getType($fieldMapping['type'])->convertToPHPValue($value, $this->platform);
     }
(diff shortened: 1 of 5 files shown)


Ocramius:
Argh, this is a bit of a "lots of people dancing around, stepping on each other's toes". @lcobucci has already added the tests in #5804, so this is no longer needed, as @astepin's work is already integrated in `master`. 

Thanks nonetheless! \o/

PR #6498, opened 2017-06-12
Throw exception when APCu cache is used by cli commands

tommyseus:
It isn't possible to access the APCu cache (same behaviour as APC and XCache) from cli.
The commands (MetadataCommand, QueryCommand and ResultCommand) don't throw an exception when APCu cache is used.

**Changes**
- Created a AbstractCommand with one method to check the $cacheDriver and throws a \LogicException for APC, APCu and XCache.
- Created unittests for MetadataCommand, QueryCommand and ResultCommand.
- Moved RegionCache unittests to the corresponding namespace.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/AbstractCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/AbstractCommand.php
@@ -1,0 +1,56 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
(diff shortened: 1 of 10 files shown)


lib/Doctrine/ORM/Tools/Console/Command/ClearCache/AbstractCommand.php
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * @param object $cacheDriver
+     *
+     * @throws \LogicException
+     */
+    protected function checkCacheDriverType($cacheDriver)
+    {
+        if ($cacheDriver instanceof Cache\ApcCache) {
+            $name = 'APC';
+        } elseif ($cacheDriver instanceof Cache\ApcuCache) {
```
Ocramius: There's no need for the `else` keyword anywhere here

lib/Doctrine/ORM/Tools/Console/Command/ClearCache/AbstractCommand.php
```diff
@@ -0,0 +1,56 @@
...
+            $name = 'APCu';
+        } elseif ($cacheDriver instanceof Cache\XcacheCache) {
+            $name = 'XCache';
+        }
+
+        if (isset($name)) {
```
Ocramius: Instead of checking for a variable existence, make a private method that throws the exception

lib/Doctrine/ORM/Tools/Console/Command/ClearCache/AbstractCommand.php
```diff
@@ -0,0 +1,56 @@
...
+ * @since   2.5
+ */
+abstract class AbstractCommand extends Command
+{
+    /**
+     * @param object $cacheDriver
+     *
+     * @throws \LogicException
+     */
+    protected function checkCacheDriverType($cacheDriver)
```
Ocramius: This can be a public static utility: inheritance is not needed

lib/Doctrine/ORM/Tools/Console/Command/ClearCache/MetadataCommand.php
```diff
@@ -87,14 +84,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
-        if ($cacheDriver instanceof ApcCache) {
-            throw new \LogicException("Cannot clear APC Cache from Console, its shared in the Webserver memory and not accessible from the CLI.");
-        }
-
-        if ($cacheDriver instanceof XcacheCache) {
-            throw new \LogicException("Cannot clear XCache Cache from Console, its shared in the Webserver memory and not accessible from the CLI.");
-        }
-
+        $this->checkCacheDriverType($cacheDriver);
```
Ocramius: `assertCacheDriverAllowsClearingInConsole()` maybe?

Ocramius (changes requested):

lcobucci:
@tommyseus @Ocramius does it make sense to move the validation to the method that ensures production settings? I think it would be better to keep these things together.

Ocramius:
> does it make sense to move the validation to the method that ensures production settings?

These commands are unrelated to that though...

lcobucci:
>These commands are unrelated to that though... 

Yeah I know, I'm just saying that I think it's more logical to check the cache driver as part of the production settings checking and thus making users responsible for following our suggestions or not (and dealing with the consequences of that decision).

That reduces the amount of changes we need to do and brings the responsibility to the correct hands. I mean, we don't have control on how people decide to use the ORM (and that's why we can't solve all N+1 errors).

I'm fine on going on with this path but my intention is more to "educate" people to use the correct stuff. If we decide to proceed with this implementation we should also add it to L2C commands and also add `ArrayCache` to the list (since it simply doesn't need to be cleaned).

PR #6501, opened 2017-06-14
Clarify the impact of changing the fetch mode

mpdude:
Seems to work for one-to-many as well, but with a negative performance impact.


Ocramius (approved):
LGTM

Issue #6502, opened 2017-06-15
Wrong column order in inherited table metadata


mpajunen:
We ran into an issue where the column and parameter order of an insert statement for a new entity don't match. Something like this:

```
INSERT INTO Person (name, group_id, email) 
VALUES ("John Smith", "john@example.org", 3)
```

(I might be able to add a failing test case later, but thought I'd report this now in case there's a straightforward way to fix the problem.)

The issue occurs when using Class Table Inheritance and metadata cache. As far as I can tell the problem is that field order of the `ClassMetadataInfo` `reflFields` property is unstable.

Without cache:

1. `AbstractClassMetadataFactory` loads the metadata directly if [metadata is not found in the cache](https://github.com/doctrine/common/blob/7181fafc832e90e866000b7cbe86aa0939cf2de0/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php#L217) or if [cache is not enabled](https://github.com/doctrine/common/blob/7181fafc832e90e866000b7cbe86aa0939cf2de0/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php#L226).
2. `AbstractClassMetadataFactory` [calls the load implementation](https://github.com/doctrine/common/blob/7181fafc832e90e866000b7cbe86aa0939cf2de0/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php#L333) for each class in the inheritance hierarchy.
3. `ClassMetadataFactory` handles inherited [fields](https://github.com/doctrine/doctrine2/blob/fc67b398a1efd3d61837778e2cda2d05a395778c/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L130), [relations](https://github.com/doctrine/doctrine2/blob/fc67b398a1efd3d61837778e2cda2d05a395778c/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L131) and [embeds](https://github.com/doctrine/doctrine2/blob/fc67b398a1efd3d61837778e2cda2d05a395778c/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L132) and [adds them to `reflProperties`](https://github.com/doctrine/doctrine2/blob/fc67b398a1efd3d61837778e2cda2d05a395778c/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L413).
[...]

PR #6503, opened 2017-06-16
Fix metadata field order stability (failing test)

mpajunen:
A simplified, currently failing test case for #6502 .

Didn't find an easy way to fix this yet. The two paths to load the metadata are quite different and more drastic changes might break something else.

The test is more implementation specific than I'd like, but in practice the `reflFields` property is used directly.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Tests\ORM\Mapping;
 
+use Doctrine\Common\Cache\ArrayCache;
 use Doctrine\Common\EventManager;
 use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Common\Persistence\Mapping\RuntimeReflectionService;
@@ -14,6 +15,7 @@ use Doctrine\ORM\Events;
 use Doctrine\ORM\Id\AbstractIdGenerator;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataFactory;
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\ORM\ORMException;
 use Doctrine\Tests\Mocks\ConnectionMock;
@@ -26,6 +28,7 @@ use Doctrine\Tests\Models\DDC4006\DDC4006User;
 use Doctrine\Tests\Models\JoinedInheritanceType\AnotherChildClass;
 use Doctrine\Tests\Models\JoinedInheritanceType\ChildClass;
 use Doctrine\Tests\Models\JoinedInheritanceType\RootClass;
+use Doctrine\Tests\Models\PropertyOrder;
 use Doctrine\Tests\Models\Quote;
 use Doctrine\Tests\OrmTestCase;
 use DoctrineGlobal_Article;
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -451,6 +454,70 @@ public function testInheritsIdGeneratorMappingFromEmbeddable()
...
+        $this->assertSame(
+            ['id', 'name', 'group'],
+            array_keys($userMetadata->reflFields)
+        );
+    }
+
+    public function testPropertyOrderWithCache()
+    {
+        $noCacheCmf = $this->createPropertyOrderFactory();
+        $cache      = new ArrayCache();
```
Ocramius: Hmm, an `ArrayCache` doesn't really serialize anything though...

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
mpajunen: That's why the `reflFields` property is reset manually. I meant to add a comment about that.

Alternatively could (de)serialize manually, or use another cache implementation.

Ocramius: @mpajunen as long as the example suffices to demonstrate a failure, that's enough :)

tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -451,6 +454,70 @@ public function testInheritsIdGeneratorMappingFromEmbeddable()
...
+        $addCacheMetadata(PropertyOrder\Admin::class);
+        $addCacheMetadata(PropertyOrder\User::class);
+
+        $cmf = $this->createPropertyOrderFactory();
+        $cmf->setCacheDriver($cache);
+
+        /** @var ClassMetadataInfo $adminMetadata */
+        $adminMetadata = $cmf->getMetadataFor(PropertyOrder\Admin::class);
+        $this->assertSame(
+            ['id', 'name', 'group', 'email'],
```
Ocramius, line 500: So, if I get it correctly, the problem is that the sorting of the keys changes? How much does the ORM API rely on fetching fields and reflection fields in the correct order? I think that since this is a hashmap, the sorting should not really cause trouble (fragile consumer, not metadata)

mpajunen: Yes, order of the fields is the problem.

At least persisting new entities relies on this order. The field order is potentially relevant at least in the following places:

- [`BasicEntityPersister:1428`](https://github.com/doctrine/doctrine2/blob/388afb46d0cb3ed0c51332e8df0de9e942c2690b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1428)
- [`JoinedSubclassPersister:534`](https://github.com/doctrine/doctrine2/blob/388afb46d0cb3ed0c51332e8df0de9e942c2690b/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php#L534)
- [`UnitOfWork:580`](https://github.com/doctrine/doctrine2/blob/388afb46d0cb3ed0c51332e8df0de9e942c2690b/lib/Doctrine/ORM/UnitOfWork.php#L580)
- [`UnitOfWork:940`](https://github.com/doctrine/doctrine2/blob/388afb46d0cb3ed0c51332e8df0de9e942c2690b/lib/Doctrine/ORM/UnitOfWork.php#L940)

The field order of the derived collections like `$columns` and `$actualData` is relevant to the generated queries.

Ocramius: I think we'll need to change those code locations too then... Relying on the order is too fragile, and it's also too easy to introduce regressions there...

PR #6506, opened 2017-06-17
Checks if the Change set exists and is  not null before using it.

### LINKED ISSUE #6505: On update, doctrine doesn't generate a changeset when the property changed is an entity
## Scenario

I have a `Product` entity and an `Advert` entity.
Advert 1 was linked to Product 1, I change it to be linked to Product 2.

## Facts

The process calls `Doctrine\ORM\UnitOfWork::executeUpdates` and on line `1060` an Error is thrown:

Catchable Fatal Error: Argument 3 passed to Doctrine\ORM\Event\PreUpdateEventArgs::__construct() must be of the type array, null given, called in /var/www/project/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php on line 1060 and defined

because `$this->entityChangeSets[$oid]` is nul

fabricekabongo:
Fixes #6505

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1046,8 +1046,13 @@ class UnitOfWork implements PropertyChangedListener
             }
 
             if ($preUpdateInvoke != ListenersInvoker::INVOKE_NONE) {
-                $this->listenersInvoker->invoke($class, Events::preUpdate, $entity, new PreUpdateEventArgs($entity, $this->em, $this->getEntityChangeSet($entity)), $preUpdateInvoke);
-
+                if (!array_key_exists($oid, $this->entityChangeSets) || is_null($this->entityChangeSets[$oid])) {
+                    $emptyChangeSet = array();
+                    $this->listenersInvoker->invoke($class, Events::preUpdate, $entity, new PreUpdateEventArgs($entity, $this->em, $emptyChangeSet), $preUpdateInvoke);
+                } else {
+                    $this->listenersInvoker->invoke($class, Events::preUpdate, $entity, new PreUpdateEventArgs($entity, $this->em, $this->entityChangeSets[$oid]), $preUpdateInvoke);
+                }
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1046,8 +1046,13 @@ private function executeUpdates($class)
             }
 
             if ($preUpdateInvoke != ListenersInvoker::INVOKE_NONE) {
-                $this->listenersInvoker->invoke($class, Events::preUpdate, $entity, new PreUpdateEventArgs($entity, $this->em, $this->getEntityChangeSet($entity)), $preUpdateInvoke);
-
+                if (!array_key_exists($oid, $this->entityChangeSets) || is_null($this->entityChangeSets[$oid])) {
```
stloyd:
```php
if (isset($this->entityChangeSets[$oid])) {
    $this->listenersInvoker->invoke($class, Events::preUpdate, $entity, new PreUpdateEventArgs($entity, $this->em, $this->entityChangeSets[$oid]), $preUpdateInvoke);
} else {
    $this->listenersInvoker->invoke($class, Events::preUpdate, $entity, new PreUpdateEventArgs($entity, $this->em, array()), $preUpdateInvoke);
}

stloyd (changes requested):

lib/Doctrine/ORM/UnitOfWork.php
fabricekabongo: 1. I agree with the `isset`
2. I need a variable because it is passed by reference, it fails when I pass directly a value (with `array()` or with `[]`)

Ocramius:
Test case missing

Ocramius:
Looking at this again, something is fishy: how did you manage to get a scheduled entity update without an actually entity changeset? That's pretty much what I'd like to see in a test case.

fabricekabongo:
I'll create the test case,  Sorry for the delay.

PR #6508, opened 2017-06-19
[Doc] Fix links to api doc in the second level cache documentation page

magnetik:


lcobucci (approved):
@magnetik nice catch, thanks!

PR #6511, opened 2017-06-20
Define build failure conditions on scrutinizer

lcobucci:
So that @Ocramius can spend more time drinking and playing 😜

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -9,3 +9,16 @@ before_commands:
 tools:
     external_code_coverage:
         timeout: 3600
+
+filter:
+    excluded_paths:
+        - docs
+        - tools
+
+build_failure_conditions:
+    - 'elements.rating(<= C).new.exists'                        # No new classes/methods with a rating of C or worse allowed
+    - 'issues.label("coding-style").new.exists'                 # No new coding style issues allowed
+    - 'issues.severity(>= MAJOR).new.exists'                    # New issues of major or higher severity
+    - 'project.metric_change("scrutinizer.test_coverage", < 0)' # Code Coverage decreased from previous inspection
+    - 'patches.label("Doc Comments").new.exists'                # No new doc comments patches allowed
+    - 'patches.label("Unused Use Statements").new.exists'       # No new unused imports patches allowed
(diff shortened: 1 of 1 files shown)


.scrutinizer.yml
```diff
@@ -9,27 +9,117 @@ before_commands:
...
-    sensiolabs_security_checker: true
+
+filter:
+    excluded_paths:
+        - docs
+        - tools
+
+checks:
+    php:
+        code_rating: true
```
Ocramius: Urgh, do we really want to maintain this infinite list of flags? :-(

Ocramius (changes requested):

.scrutinizer.yml
lcobucci: Some are set by default, some are not... and those keep changing 😢

Ocramius: Then let's please not go in this direction :-\

Ocramius: The fact that it keeps changing is precisely the problem: if there was an XSD I would be OK with this, but not with some YAML cruft.

lcobucci: So you want to remove the default ones from the config, right?

lcobucci: I mean, since we don't have control over the format...

Ocramius: The point is having *LESS* flags to maintain. Let's keep all the defaults we can keep, and only override the ones we need to override.

This project will never reach "object calisthenics" quality within our lifetime, so it's useless to use a code quality analysis tool to drive improvements: we are good at reviewing stuff.

lcobucci: The point is to also keep https://scrutinizer-ci.com/g/doctrine/doctrine2/issues/master updated with things that we consider to be improved (and which people could just pick up).

I mean, the defaults are fine but they're too generic... maybe we could find a middle ground?

Ocramius:
I'd just keep the defaults for now: the tool is interesting for finding low
hanging fruit issues.

lcobucci:
@Ocramius  ok, but do you agree with adding the failing conditions?

Ocramius:
I'm fine with additional failure conditions, yes. Just overwhelmed with the
size of the diff

lcobucci:
@Ocramius have fun 😉

Ocramius (approved):
LGTM :+1:

PR #6515, opened 2017-06-22
Use the Travis' Trusty build environment

lcobucci:
To see if that solves the XML validation issues (since the issues are
possibly happening due to old libxml versions).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 sudo: false
 language: php
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_script:
   - composer install --prefer-source
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -18,6 +18,7 @@ before_script:
   - composer install --prefer-source
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
+  - if [[ $DB == "mysql" || $DB == "mariadb" ]]; then SCHEMA=`xmllint --xpath 'string(//var[@name="db_name"]/@value)' tests/travis/$DB.travis.xml` && mysql -e "CREATE SCHEMA $SCHEMA; GRANT ALL PRIVILEGES ON $SCHEMA.* to travis@'%'"; fi;
```
lcobucci: @Ocramius `xmllint` is not installed, I was trying to avoid hardcoded stuff but I'm not sure that we should install `libxml2-utils` just because of that...

Ocramius: Surely not just for this line. We can actually hardcode it in `.travis.yml` though...

lcobucci: alright will do it 😉

lcobucci:
@Ocramius done

Ocramius:
selection_001

Ocramius:
:ship:

PR #6507, opened 2017-06-19
Upgrade dependencies

lcobucci:
Related to #6424

PR #6519, opened 2017-06-22
Bump versions of some dependencies

cengizhancaliskan:
Added symfony/console 2.8 for support to symfony 2.8.x


lcobucci:
@cengizhancaliskan we just dropped that version on #6507, I'm not sure if we should put it back. The idea was to bump the dependencies just like we're doing with the PHP version. @Ocramius your call though...

Ocramius:
> I'm not sure if we should put it back.

We shouldn't: afaik we dropped it to remove some CLI tool deprecations that were fixed in symfony 3.0

PR #6522, opened 2017-06-24
[WIP] PHP7.1 types added

TomasVotruba:
Since PHP 7.1 is minimum, this can be added now.

There are 195 errors caused by invalid types used. Let me know this PR is way to go. I can fix them.


Ocramius:
@TomasVotruba this is a BC break, as the failures show: we can only apply it to develop

Majkl578:
Serious BC break indeed. I wouldn't even apply this to develop, I'd let @guilhermeblanco finish most of the refactorings first to avoid tons of conflicts etc. This has no added value at this point.

Ocramius:
Closing as `won't fix`, sorry @TomasVotruba :-\

TomasVotruba:
What's the final statement for this? Should be applied to develop or not at all or wait for something?

Ocramius:
@TomasVotruba we should *try* applying this to `develop` and see how many tests change. For each changed test, we need to check what the possible userland BC breaks are.

TomasVotruba:
@Ocramius: Thanks, that's clear to me.

Replaced by:
https://github.com/doctrine/doctrine2/pull/6524

Actually using scalar typehints is on the menu after that PR gets merged, right?

Ocramius:
@TomasVotruba yeah, we'll add scalar types gradually, but we need to document ALL API endpoints that change, which is a huge pain

PR #6523, opened 2017-06-24
Enable strict types

Majkl578:
Lets just see how many things this breaks.


lib/Doctrine/ORM/Cache/CacheConfiguration.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+declare(strict_types = 1);
```
Ocramius: Just a note on style: the `declare(strict_types=1)` (no spaces) should be after the licence

Ocramius (changes requested):

lib/Doctrine/ORM/Cache/CacheConfiguration.php
Majkl578: Just did a quick replace in PhpStorm. 😂 `^<?php` -> `<?php\n\ndeclare(strict_types = 1);\n`. Will adjust accordingly if this PR should land eventually.

Ocramius: @Majkl578 we can land it if it passes with green and not too many test changes. We'll need to look at the changed tests and see whether there are serious BC breaks for consumers.

Ocramius:
@Majkl578 urgh, I need to close this, as it targets `master`.

We should work on `develop` for this :-\

Ocramius:
Linking https://github.com/doctrine/doctrine2/pull/6522

Majkl578:
@Ocramius Master was intended target for this, to see how (im)possible would it be for 2.6.

Technically, in ideal world, this shouldn't be a BC break as strict types are enabled just internally. But there are already some strange errors due incorrect types/assumptions. :/

Majkl578:
Looks like this is gonna be a donkey work to properly support strict types. PHPStan (#6440) should catch & fix most of these issues (not on Level 0 though 😆 ). Could eventually retarget this PR to develop.

Ocramius:
> Could eventually retarget this PR to develop.

That would be great :-)

And yes, phpstan could help loads here.

Majkl578:
Retargeted to develop & force-pushed, but GitHub didn't update the PR, probably because it's closed. :/

Ocramius:
New PR needed, heh

PR #6530, opened 2017-06-27
Update Generation way of the options

PyRowMan:
When the options array was created, only the unsigned value were setted.

### DIFF
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -1660,7 +1660,7 @@ public function __construct(<params>)
             }
 
             if (isset($fieldMapping['options']) && !empty($fieldMapping['options'])) {
-                $column[] = $this->generateFormattedOptionsString($fieldMapping['options'])
+                $column[] = $this->generateFormattedOptionsString($fieldMapping['options']);
             }
 
             if (isset($fieldMapping['columnDefinition'])) {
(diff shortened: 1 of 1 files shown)


Ocramius:
Misses test cases, plus the additional conditionals should be stashed in a separate private method.

PyRowMan:
I updated the code, and put it into a function, but i'm not the best at doing tests and i currently cannot test them on my machine.

If someone could do them for me, it would be much appreciated :)

PyRowMan:
There still is a bug with the comments with double quotes, even escaping them doesn't work. for myself i replaced them all with simple quotes, but it's a solution only for my projet, not for long therm.

Ocramius:
> I updated the code, and put it into a function, but i'm not the best at doing tests and i currently cannot test them on my machine.

Travis can test things for you. Also, read https://github.com/doctrine/doctrine2/blob/e3ecec36ad77525553bfd4e6a2ec86e80d39ac27/CONTRIBUTING.md#unit-tests please :-)

Majkl578 (changes requested):
Just found this months old unsent review. :)
Please also add tests.

Majkl578:
I'm going to close this PR as we will be removing EntityGenerator in Doctrine 3.0 completely and 2.x is closed feature-wise. Thank you for the PR.

PyRowMan:
Yeah, I completely forgot to update theses :/


Sorry for that.

Issue #6499, opened 2017-06-12
Persisting entities with OneToOne and OneToMany/ManyToOne relationship between them.


Frikkle:
Last week I encountered an issue in one of our projects where we had two entities, call them A and B having a unidirectional OneToOne A->B (with the join column not nullable) and a OneToMany(A->B)/ManyToOne(B->A) bidirectional relationship between them.

The problem lies in the order of persisting them; when first persisting A, then B, a constraint violation is thrown: SQLSTATE[23000]: Integrity constraint violation: 19 NOT NULL constraint failed: a.b_id, indicating that the join column isn't allowed to be NULL, even though it should have been set.

I reworked my code to a minimal example:

**Entity A**
```php
/**
 * Class A
 * @ORM\Entity()
 */
class A
{
    /**
     * @ORM\Id()
     * @ORM\GeneratedValue(strategy="AUTO")
     * @ORM\Column(name="id", type="integer")
     */
    private $id;

    /**
     * @ORM\OneToMany(targetEntity="AppBundle\Entity\B", mappedBy="a", cascade={"persist", "remove"}, orphanRemoval=true)
     */
    private $bs;

    /**
     * @ORM\OneToOne(targetEntity="AppBundle\Entity\B", cascade={"persist"})
     * @ORM\JoinColumn(nullable=false)
     */
    private $b;

    /**
     * @return int
     */
    public function getId()
    {
        return $this->id;
    }

    /**
     * @return B[]|ArrayCollection
     */
    public function getBs()
    {
        return $this->bs;
    }

    /**
     * @param B $b
     */
    public function addB(B $b)
    {
        if ($this->bs->contains($b)) return;

        $this->bs->add($b);

        // Update owning side
        $b->setA($this);
    }

    /**
     * @param B $b
     */
    public function removeB(B $b)
    {
        if (!$this->bs->contains($b)) return;

        $this->bs->removeElement($b);

        // Not updating owning side due to orphan removal
    }

    /**
     * @return B
     */
    public function getB()
    {
        return $this->b;
    }

    /**
     * @param B $b
     */
    public function setB(B $b)
    {
        $this->b = $b;
    }
}
```

**Entity B**
```php
/**
 * Class B
 * @ORM\Entity()
 */
class B
{
    /**
     * @ORM\Id()
     * @ORM\GeneratedValue(strategy="AUTO")
     * @ORM\Column(name="id", type="integer")
     */
    private $id;

    /**
[...]

PR #6533, opened 2017-06-29
#6499 commit order must consider non-nullability of join columns as prioritised over nullable join columns

Frikkle:
As suggested by @lcobucci, I created a PR with a minimal test example.

While working on the test example I looked into the code and found the (probable) cause for this issue and proposed a fix for it in a second commit.

The issue seems to originate from the part of the code in UnitOfWork which calculates the weight for new dependencies in the graph based on whether or not the `nullable` clause on join columns is set and to which value.

In particular the part `(int)empty($joinColumns['nullable'])` which to me seems incorrect. With the current implementation in `master` the case where the `nullable` property on the join column is set to false (so, the relation is mandatory) yields the same weight as when the `nullable` property is not defined. But when the `nullable` property is not defined, the documention states that the default would be `nullable === true`.

Below, you find a simple truth table breakdown of the specific part of the code (for the possible values for `$joinColumns['nullable']`.

| `(int)` 	| `empty`	| `( $joinColumns['nullable'] )` 	|
|-------	|-------	|------------------------------	|
| 0     	| false 	| true                         	|
| 1     	| true  	| false                        	|
| 1     	| true  	| unset/null/etc.              	|

I tried to keep the (in-line) fix as compact as possible. All test keep running and the added test for this issue also succeeds after applying the fix. Before applying the fix, it throws a ConstraintViolationException and tries to insert null for a mandatory foreign key.

Since I'm not completely familiar with the codebase, please look into whether this does not influence other parts of the code. If there's anything I can do to assist, let me know.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1154,7 +1154,7 @@ class UnitOfWork implements PropertyChangedListener
 
                 $joinColumns = reset($assoc['joinColumns']);
 
-                $calc->addDependency($targetClass->name, $class->name, (int)empty($joinColumns['nullable']));
+                $calc->addDependency($targetClass->name, $class->name, (int) ! ( ! isset($joinColumns['nullable']) || $joinColumns['nullable'] === true));
 
                 // If the target class has mapped subclasses, these share the same dependency.
                 if ( ! $targetClass->subClasses) {
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6499Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6499Test.php
@@ -13,7 +13,7 @@ class DDC6499Test extends OrmFunctionalTestCase
     /**
      * {@inheritDoc}
      */
-    protected function setUp()
+    protected function setUp() : void
     {
         parent::setUp();
 
@@ -26,21 +26,55 @@ class DDC6499Test extends OrmFunctionalTestCase
     }
 
     /**
+     * {@inheritDoc}
+     */
+    protected function tearDown() : void
+    {
+        parent::tearDown();
+
+        $this->_schemaTool->dropSchema(
+            [
+                $this->_em->getClassMetadata(DDC6499A::class),
+                $this->_em->getClassMetadata(DDC6499B::class),
+            ]
+        );
+    }
+
+    /**
      * Test for the bug described in issue #6499.
      */
-    public function testIssue()
+    public function testIssue() : void
     {
         $a = new DDC6499A();
         $this->_em->persist($a);
 
         $b = new DDC6499B();
-        $a->setB($b);
+        $a->b = $b;
         $this->_em->persist($b);
 
         $this->_em->flush();
+        $this->_em->clear();
 
-        // Issue #6499 will result in a Integrity constraint violation before reaching this point
-        $this->assertEquals(true, true);
+        self::assertEquals($this->_em->find(DDC6499A::class, $a->id)->b->id, $b->id, "Issue #6499 will result in a Integrity constraint violation before reaching this point.");
+    }
+
+    /**
+     * Test for the bug described in issue #6499 (reversed order).
+     */
+    public function testIssueReversed() : void
+    {
+        $a = new DDC6499A();
+
+        $b = new DDC6499B();
+        $a->b = $b;
+
+        $this->_em->persist($b);
+        $this->_em->persist($a);
+
+        $this->_em->flush();
+        $this->_em->clear();
+
+        self::assertEquals($this->_em->find(DDC6499A::class, $a->id)->b->id, $b->id, "Issue #6499 will result in a Integrity constraint violation before reaching this point.");
     }
 }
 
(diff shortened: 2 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6499Test.php
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * @OneToOne(targetEntity="DDC6499B", cascade={"persist"})
+     * @JoinColumn(nullable=false)
+     */
+    private $b;
+
+    /**
+     * DDC6499A constructor.
+     */
+    public function __construct()
```
Ocramius: Can you drop all API that isn't strictly needed? Getters, setters, etc

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6499Test.php
```diff
@@ -0,0 +1,172 @@
...
+    {
+        $this->bs = new ArrayCollection();
```
Ocramius: Use `= []` as default property value (test is simple enough) ;-)

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6499Test.php
```diff
@@ -0,0 +1,172 @@
...
+            $this->_em->getClassMetadata(DDC6499A::class),
+            $this->_em->getClassMetadata(DDC6499B::class),
+            ]
+        );
+    }
+
+    /**
+     * Test for the bug described in issue #6499.
+     */
+    public function testIssue()
```
Ocramius: Mark as `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6499Test.php
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * @group DDC-6499
+ */
+class DDC6499Test extends OrmFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function setUp()
```
Ocramius: Mark as `: void`

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6499Test.php
```diff
@@ -0,0 +1,172 @@
...
+        $this->_em->persist($a);
+
+        $b = new DDC6499B();
+        $a->setB($b);
+        $this->_em->persist($b);
+
+        $this->_em->flush();
+
+        // Issue #6499 will result in a Integrity constraint violation before reaching this point
+        $this->assertEquals(true, true);
```
Ocramius: This should be dropped - instead, try a `find` after a `clear`, and verify that the collection and the association are correctly populated. Put the comment in the assertion message

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6499Test.php
```diff
@@ -0,0 +1,172 @@
...
+{
+    /**
+     * @Id()
+     * @GeneratedValue(strategy="AUTO")
+     * @Column(name="id", type="integer")
+     */
+    private $id;
+
+    /**
+     * @OneToMany(targetEntity="DDC6499B", mappedBy="a", cascade={"persist", "remove"}, orphanRemoval=true)
```
Ocramius: Drop all cascade operations, orphan removals and properties/mappings that aren't strictly related to the tests

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1154,7 +1154,7 @@ private function getCommitOrder(array $entityChangeSet = null)
 
                 $joinColumns = reset($assoc['joinColumns']);
 
-                $calc->addDependency($targetClass->name, $class->name, (int)empty($joinColumns['nullable']));
+                $calc->addDependency($targetClass->name, $class->name, (int) ! ( ! isset($joinColumns['nullable']) || $joinColumns['nullable'] === true));
```
Ocramius: Lots of negations. Also, this is only checking the first join column... Possibly requiring to add tests around this area. :-\

Ocramius (changes requested):

triasinformatica-gabe:
I'll adjust the above suggestions, fix Scrutinizer improvements and run some more tests locally.

lib/Doctrine/ORM/UnitOfWork.php
lcobucci: I think it makes sense to also cover this on the unit level (`UnitOfWorkTest`)

Ocramius: Let's keep it on integration tests for now - unit testing the UoW is a bit more messy, and we can add that just before merging, rather than early during the fixing process

triasinformatica-gabe: I just fixed the other requested changes, causing the example to be a lot smaller. This proposed change fixes our broken test, but consequently breaks two other tests. I will look a bit further into this when I have some spare time.

@Ocramius Do you agree that there is a logic inconsistency in the the original line in `UnitOfWork.php`?

guilliamxavier:
Just for the record: [the file's history](https://github.com/doctrine/doctrine2/commits/a22f1650268051d763bc847f6ad0f4b328085839/lib/Doctrine/ORM/UnitOfWork.php) shows that the current logic

`(int)empty($joinColumns['nullable'])`

originates from https://github.com/doctrine/doctrine2/commit/27e9b492152090d41c47bc30dc24b5ed316b9f2b#diff-6e8c1c1e78b054ba05e20ea09d877865L1158 as a simplification of

`(isset($joinColumns['nullable']) ? $joinColumns['nullable'] : false) ? 0 : 1`

<sup>(actually with an intermediate variable `$isNullable` for the parenthesized ternary expression)</sup>
which was originally introduced in https://github.com/doctrine/doctrine2/commit/8ea62b95b8e4a8b3de000a59f018ec250b41d714#diff-6e8c1c1e78b054ba05e20ea09d877865L1165

so the present PR is like saying that the `false` should actually have been a `true`, or (in PHP 7):

`($joinColumns['nullable'] ?? true) ? 0 : 1`

or alternatively

`(int) ! ($joinColumns['nullable'] ?? true)`
`(int) ! ( ! isset($joinColumns['nullable']) || $joinColumns['nullable'])`
`(int) (isset($joinColumns['nullable']) && ! $joinColumns['nullable'])`

all being equivalent to [the currently proposed fix](https://github.com/doctrine/doctrine2/pull/6533#discussion-diff-124767030R1157) just without the `=== true` strict equality.

Now, as you already noted, this is unfortunately causing [exceptions](https://travis-ci.org/doctrine/doctrine2/jobs/248315732) in an existing [test](https://github.com/doctrine/doctrine2/blob/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/CascadeRemoveOrderTest.php), and the original logic was precisely a [fix](https://github.com/doctrine/doctrine2/pull/1570) for the affected test.

triasinformatica-gabe:
@guilliamxavier Thanks for clarifying the file's history. As said, I'm not that familiar with the codebase and when digging into it I also could not find out what causes the other two tests to fail.

It might be that this proposed fix isn't correct for this issue, but then the question remains: what causes issue #6499 to arise? Or is #6499 intended behavior?

triasinformatica-gabe:
Any news on this issue?

Ocramius:
@triasinformatica-gabe adding it to my bucket of stuff to check

triasinformatica-gabe:
@Ocramius thanks.

If there's anything we (me or my colleagues) can do to help, please let us know. But currently we're unfamiliar with the history of this part of the doctrine codebase.

Ocramius:
@triasinformatica-gabe I manually merged after some cleanups and moving stuff around.

@lcobucci I couldn't find a proper way to test this in isolation, as it's deep in UnitOfWork internals, sadly.

This patch can only land in `2.6.0`, since the commit order calculator doesn't even know about weights of dependencies before that.

Ocramius:
@triasinformatica-gabe sorry, I had to revert the merge - this patch introduces errors when deleting foreign key referenced data.

I think this is much more complex than what it looks. To be clear, I know as much as you do about this issue. The problem area is exactly the one you worked on, except that the commit order isn't simply "reversible" for deletes (current approach).

bendavies:
@triasinformatica-gabe @Ocramius is tihs waiting on anything now? This fixes https://github.com/doctrine/doctrine2/issues/7006 for me

Ocramius:
@bendavies see https://github.com/doctrine/doctrine2/pull/6533#issuecomment-321911821 - this was reverted

bendavies:
@Ocramius yeah thanks. I got confused as this PR was still open. was `CascadeRemoveOrderTest` the issue? It seems to be passing on 2.6 now, with this patch.

Ocramius:
@bendavies no, we are lacking tests about cascade-deleted data, from what I remember (sorry, mind is also a bit clouded over here)

triasinformatica-gabe:
@bendavies It's passing the tests, but only in SQLite, which is ignoring the foreign keys. If you run the test set in MySQL or another database that does enforce this, some other tests will fail. There seems to be a non-trivial issue in building up the relationship dependency graph for cascades. I didn't have time so for to look into it, but it's still somewhere on my list.

The fix this PR originates from was a bit to rigid and therefore not solving the issue.

bendavies:
I ran against mysql and postres. They passed!

triasinformatica-gabe:
Then maybe some other fix has fixed this issue; I’ll look into it tomorrow morning if I have time and post back my findings. Should save Ocramius some time.

bendavies:
any luck @triasinformatica-gabe?

bendavies:
hey @triasinformatica-gabe. Any chance to look at this yet?

triasinformatica-gabe:
@bendavies Sorry for the late response, haven't had time to feedback on this before since I was abroad for a while. Any way; I did run all the tests on top of the latest changes and for me all the same tests keep failing the same way as they did last summer, so I'm not sure how you managed to get it working running Postgress and MySQL.

Could you elaborate?

bendavies:
Hmm, I'll try again.

bendavies:
@triasinformatica-gabe which test is failing for you please?

Ocramius:
> @Ocramius can you give more context about that?

The problem is:

 > (sorry, mind is also a bit clouded over here)

I also don't remember the details.

lcobucci:
> I also don't remember the details.

@Ocramius I BELIEVE IN YOU, you can do it!!!! Didn't help, did it? HAHAHA alright, thanks!

bendavies:
@triasinformatica-gabe trying to pick this up again....which test fails for you?

PR #6545, opened 2017-07-11
Fix newline in annotation example in documentation

Landerstraeten:


Ocramius:
@Landerstraeten thanks!

PR #6546, opened 2017-07-11
Fix OrderBy annotation with association

enumag:
According to SchemaValidator, the `@OrderBy` annotation can have either [a field name or an association name](https://github.com/doctrine/doctrine2/blob/43009682a49c9fa7fe1bacabf22a7230aa524e1f/lib/Doctrine/ORM/Tools/SchemaValidator.php#L231).

However if association is used it doesn't work correctly and the result is an `Undefined index` error in `DefaultQuoteStrategy::getColumnName()` - obviously since the association is not in `$classMetadata->fieldMappings`.

Of course some test is needed. But first I'd like to know how (if at all) to handle associations with more than one join column.

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -410,7 +410,16 @@ class SqlWalker implements TreeWalker
             $persister = $this->em->getUnitOfWork()->getEntityPersister($qComp['metadata']->name);
 
             foreach ($qComp['relation']['orderBy'] as $fieldName => $orientation) {
-                $columnName = $this->quoteStrategy->getColumnName($fieldName, $qComp['metadata'], $this->platform);
+                if ($qComp['metadata']->hasField($fieldName)) {
+                    $columnName = $this->quoteStrategy->getColumnName($fieldName, $qComp['metadata'], $this->platform);
+                } else {
+                    $columnName = $this->quoteStrategy->getJoinColumnName(
+                        $qComp['metadata']->associationMappings[$fieldName]['joinColumns'][0],
+                        $qComp['metadata'],
+                        $this->platform
+                    );
+                }
+
                 $tableName  = ($qComp['metadata']->isInheritanceTypeJoined())
                     ? $persister->getOwningTable($fieldName)
                     : $qComp['metadata']->getTableName();


Ocramius:
@enumag I suggest checking the `develop` branch. As far as I know, @guilhermeblanco did some work to simplify getting column information for associations or fields

enumag:
@Ocramius In develop it is clearly much different but it looks unfinished/broken - or maybe I missed something. [SqlWalker](https://github.com/doctrine/doctrine2/blob/develop/lib/Doctrine/ORM/Query/SqlWalker.php#L414-L416) calls `ClassMetadata::getProperty()` which returns an instance of [Property](https://github.com/doctrine/doctrine2/blob/develop/lib/Doctrine/ORM/Mapping/Property.php) interface. But then it calls `$property->getTableName()` and `$property->getColumnName()` - neither of which exists? 

Anyway I'm not sure how is develop relevant here. It is a bug and should be fixed sooner than in 3.0.0 in my opinion.

Ocramius:
> Anyway I'm not sure how is develop relevant here. It is a bug and should be fixed sooner than in 3.0.0 in my opinion.

It's just about the "correct" approach: we keep splitting logic wherever join-column or field-column is being used :-\

enumag:
Yes, with the refactoring made in develop fixing this would be much simpler - actually I suspect that it is already fixed there as a sideeffect of those changes. But I still can't see how is that relevant for this PR.

Majkl578:
> SqlWalker calls ClassMetadata::getProperty() which returns an instance of Property interface. But then it calls $property->getTableName() and $property->getColumnName() - neither of which exists?

That piece of code apparently works with assumption that property is a Doctrine\ORM\Mapping\ColumnMetadata which is the one responsible for holding such information.

If it's feasible to fix this for 2.6, I'd say go for it even though develop may have resolved it in different way.

enumag:
@Majkl578 In develop `ClassMetadata::getProperty()` is supposed to return an instance of `Property` interface. `ColumnMetadata` does not implement that interface. That's why I'm confused.

Ocramius:
> That's why I'm confused.

I wasn't asking to emulate `develop`, I was checking whether it was solved in a better way :-)

Anyway, as this won't be fixed here, I'd suggest adding a `// @TODO` comment on the code, then add the tests here and we will see if `develop` also reacts well once rebased :-)

Majkl578:
@enumag What is the resolution here, are you still targeting 2.6 or do we wait for 3.0 with a cleaner approach?

enumag:
@Majkl578 3.0

guilhermeblanco:
This looks incomplete. We can only consider ordering like that for fields or owning *ToOne associations.

Majkl578:
Master is now 3.0.x-dev, feel free to continue @enumag. 🥇

derrabus:

PR #6547, opened 2017-07-11
Case insensitive like

okoehler:
This pull request adds Comparison::ICONTAINS to create a case insensitive like. 

Existing case sensitive like
`fieldname LIKE %value%`

New case insensitive like
`LOWER(fieldname) LIKE (%value%)`
with value converted to lowercase by php, and fieldname surrounded by platforms lower function. 

UnitTests will run if you use [this version of doctrine collections](https://github.com/okoehler/collections/tree/6de7e7ce5e3b0f5d50b5152366efbb9e721db212). See doctrine collections [PR#123](https://github.com/doctrine/collections/pull/123)

### DIFF
--- a/lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
+++ b/lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
@@ -186,8 +186,8 @@ class QueryExpressionVisitor extends ExpressionVisitor
                 $this->parameters[] = $parameter;
 
                 return $this->expr->like($field, $placeholder);
-            case Comparison::CONTAINS_CI:
-                $parameter->setValue('%' . strtolower($parameter->getValue()) . '%', $parameter->getType());
+            case Comparison::ICONTAINS:
+                $parameter->setValue('%' . mb_strtolower($parameter->getValue()) . '%', $parameter->getType());
                 $this->parameters[] = $parameter;
 
                 return $this->expr->like($field, $placeholder);
--- a/lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
+++ b/lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
@@ -182,6 +182,7 @@ class QueryExpressionVisitor extends ExpressionVisitor
 
                 return $this->expr->neq($field, $placeholder);
             case Comparison::CONTAINS:
+            case Comparison::CONTAINS_CI:
                 $parameter->setValue('%' . $parameter->getValue() . '%', $parameter->getType());
                 $this->parameters[] = $parameter;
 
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
```diff
@@ -186,6 +186,11 @@ public function walkComparison(Comparison $comparison)
                 $this->parameters[] = $parameter;
 
                 return $this->expr->like($field, $placeholder);
+            case Comparison::CONTAINS_CI:
+                $parameter->setValue('%' . strtolower($parameter->getValue()) . '%', $parameter->getType());
```
Majkl578: This should use multibyte safe variant, mb_strtolower.

Majkl578 (changes requested):
Needs functional tests.

Majkl578:
Maybe `ICONTAINS` would be better naming for this?

mvrhov:
This will work only on Mysql, postgresql uses ILIKE keyword for that.

okoehler:
> This will work only on Mysql, postgresql uses ILIKE keyword for that.

I use **LIKE** with left and right value converted to lower for multi platform support:

`LOWER(fieldname) LIKE (%value%)`

tobiasoberrauch:
This is exactly what I need. I hope this feature will merged as soon a possible. Thank you for contributing @okoehler

Ocramius:
> `LOWER(fieldname) LIKE (%value%)`

This will disable any kind of index usage, unless the column has an index on `LOWER(fieldname)` (only possible in PostgreSQL and advanced RDBMSs)

tobiasoberrauch:
@Ocramius What solution what you suggest?
It would be sad if we don't use this PR

Ocramius:
@tobiasoberrauch I don't have a solution: I just exposed another problem.

okoehler:
@Ocramius I created this PR to provide a feature which some people are missing. I created tests, it works for all supported databases. Of course there are some possible improvements, which could be created if there is a need for. 
But for now, I think, the benefit is high enough to see this as a first releasable version. So, what do you think, is there any chance to that this PR will be merge?

Ocramius:
> So, what do you think, is there any chance to that this PR will be merge?

Not until a solution is found here, at least in my opinion. Other @doctrine/doctrinecore folks may disagree.

sschleue:
In my opinion the usage of indexes is nice to have not mandatory. This (some kind of slow) solution is better than none. Case insensitive like would be a very nice feature. Else you always have to build it on your own which bloats your code and makes it ugly. I think a notice in documentation should be enough to be aware of that performance issue. At least for now.
I highly appreciate this PR.

PR #6549, opened 2017-07-11
OrmTestCase: Always use full AnnotationReader, drop legacy versions

Majkl578:
Develop was failing locally due to class imports not being recognized. That's because SimpleAnnotationReader was in use rather than full AnnotationReader.
Also not sure why this was testing version of Common - annotations are separate package for a while now. :/
Dropped legacy setup & unused `$alias` parameter for `Doctrine\Tests\OrmTestCase#createAnnotationDriver()`.

cc @guilhermeblanco

### DIFF
--- a/tests/Doctrine/Tests/OrmTestCase.php
+++ b/tests/Doctrine/Tests/OrmTestCase.php
@@ -4,7 +4,6 @@ namespace Doctrine\Tests;
 
 use Doctrine\Common\Annotations;
 use Doctrine\Common\Cache\ArrayCache;
-use Doctrine\Common\Version;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\ORM\Cache\CacheConfiguration;
 use Doctrine\ORM\Cache\DefaultCacheFactory;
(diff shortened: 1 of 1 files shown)


Ocramius:
LGTM :+1:

carnage:
This doesn't apply cleanly to the code in 2.10.x and doesn't look like it makes sense to apply it:

```
    /**
     * @param mixed $alias
     */
    protected function createAnnotationDriver(array $paths = [], $alias = null): AnnotationDriver
    {
        // Register the ORM Annotations in the AnnotationRegistry
        $reader = new Annotations\SimpleAnnotationReader();

        $reader->addNamespace('Doctrine\ORM\Mapping');

        if (class_exists(Annotations\PsrCachedReader::class)) {
            $reader = new Annotations\PsrCachedReader($reader, new ArrayAdapter());
        } else {
            $reader = new Annotations\CachedReader($reader, DoctrineProvider::wrap(new ArrayAdapter()));
        }

        Annotations\AnnotationRegistry::registerFile(__DIR__ . '/../../../lib/Doctrine/ORM/Mapping/Driver/DoctrineAn
notations.php');

        return new AnnotationDriver($reader, (array) $paths);
    }
```

Possible to replace the top two lines with a creation of the full reader instead of the simple one; but not sure what the motivation behind that change was.

greg0ire:
I've discussed this with @beberlei the other day, the goal is to be able to drop `SimpleAnnotationReader` entirely, which was useful when Doctrine was the only library using annotations, but given the popularity of this lib, it's not useful anymore.

More details about this in https://github.com/doctrine/orm/search?q=SimpleAnnotationReader&type=issues

I remember @beberlei told me that would imply importing `use` statements in for every annotation in test files, which should be automatable.

greg0ire:
FYI: I'm on it.

PR #6550, opened 2017-07-13
Fix return value of `PersistentCollection#removeElement()` for extra-lazy collections: should be `bool`

### LINKED ISSUE #5745: PersistentCollection::removeElement() returns object or NULL instead of boolean
When using strict typing in PHP 7.0.4: `declare (strict_types = 1);` , I noticed that an error is returned for userland code like

``` php
public function removeUser(User $user) : bool
{
    return $this->users->removeElement($user);
}

```

This results in an error message like:

```
Return value of Project\Entity::removeUser() must be of the type boolean, object returned
```

I had a look at all `removeElement()` functions used in the project and I suspect [this section](https://github.com/doctrine/doctrine2/blob/b055d78ea19835fab563dfc234d4804a9d04966a/lib/Doctrine/ORM/PersistentCollection.php#L378-L380)

Will try to come up with a test but already wanted to register this issue so I will not forget it. 

**UPDATE**
It appears that this can be tested in [ExtraLazyCollectionTest::testRemoveElementManyToMany()](https://github.com/doctrine/doctrine2/blob/b055d78ea19835fab563dfc234d4804a9d04966a/tests/Doctrine/Tests/ORM/Functional/ExtraLazyCollectionTest.php#L639):

``` php
        $result = $user->groups->removeElement($group);
        $this->assertTrue($result);
```

However, this made me wonder, why does [this part](https://github.com/doctrine/doctrine2/blob/b055d78ea19835fab563dfc234d4804a9d04966a/lib/Doctrine/ORM/PersistentCollection.php#L382) returns `NULL` while the [Collection interface indicates](https://github.com/doctrine/collections/blob/392b8dd9e04c073e832fcda99270ec5a8f5e30cd/lib/Doctrine/Common/Collections/Collection.php#L98) it should return a `boolean`? Is returning the element/null on purpose?

alcaeus:
Fixes #5745.

As indicated in the ticket, calling `removeElement` on an uninitialized collection that uses the extra_lazy fetch mode does not return a bool, breaking the contract specified in the collection interface. This PR changes `removeElement` to return a bool value in those cases as well.

### DIFF
--- a/lib/Doctrine/ORM/PersistentCollection.php
+++ b/lib/Doctrine/ORM/PersistentCollection.php
@@ -373,11 +373,7 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
 
             $persister = $this->em->getUnitOfWork()->getCollectionPersister($this->association);
 
-            if ($persister->removeElement($this, $element)) {
-                return $element;
-            }
-
-            return null;
+            return $persister->removeElement($this, $element);
         }
 
         $removed = parent::removeElement($element);
--- a/tests/Doctrine/Tests/ORM/Functional/ExtraLazyCollectionTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/ExtraLazyCollectionTest.php
@@ -640,7 +640,7 @@ class ExtraLazyCollectionTest extends OrmFunctionalTestCase
         $group      = $this->_em->find(CmsGroup::class, $this->groupId);
         $queryCount = $this->getCurrentQueryCount();
 
-        $user->groups->removeElement($group);
+        $this->assertTrue($user->groups->removeElement($group));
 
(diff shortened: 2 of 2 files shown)


Ocramius:
@alcaeus an additional test for the `false` scenario (that was now removed, it was the `null` scenario) is needed.

The test would basically attempt to call `removeElement()` with a non-existing or unrelated entity.

Ocramius (changes requested):
@alcaeus an additional test for the `false` scenario (that was now removed, it was the `null` scenario) is needed.

The test would basically attempt to call `removeElement()` with a non-existing or unrelated entity.

alcaeus:
@Ocramius done!

Ocramius:
:ship:

Ocramius:
Backported to `2.5` via doctrine/doctrine2@b3ceef0fb6f822c876571311a5431b771efaab98, thanks!

PR #6551, opened 2017-07-13
[3.0] Moved IdentifierGenerator to Property.

guilhermeblanco:

PR #6554, opened 2017-07-14
Move identifier generation to FieldMetadata

Majkl578:
Removed ClassMetadata::$generatorType, ClassMetadata::$generatorDefinition and ClassMetadata::$idGenerator and related methods.

Some pieces of code (UoW in particular) are not really nice because of original assumptions about single vs. composite keys and generator types, but it'll eventually undergo further reactorings and generalization soon with introduction of ValueGenerationPlan...

Removed two tests:
* Functional\Ticket\DDC2415Test - tested id generator inheritance which no longer exists
* Mapping\ClassMetadataTest::testSetSequenceGeneratorThrowsExceptionWhenSequenceNameIsMissing - tested setGeneratorDefinition() method that was removed

Default test suite passes, test suite w/ postgres has same number of failures compared to develop.

_Working version of #6551._


guilhermeblanco (approved):
I'm assuming as going through all these changes you realize why we discussed about having `ValueGenerationPlan` to help... =)

PR #6555, opened 2017-07-14
Generalize identifier generators as value generators aware of target property

Majkl578:
Initial step towards generalizing identifier generators as more generic value generators (to allow i.e. non-id sequences).

BC break though - adds a required parameter to Generator::generate().

Built on top of #6554.


guilhermeblanco:
My thoughts:

1- If we manage to have only 1 instance across the board and have references on property generators, that is ideal, since the cost of having N instances is higher (we save memory resources)

2- Not all the time we can have a generator into a Property. Sometimes it can be in a LocalColumn (which is a higher up element). For example, partition tables usually suggest a non-mapped column to hold their values… this approach would not work then.
The Property, Column, LocalColumn, Field, Association etc in Metadata was a design decision I made that we can change if you really want.
In both EclipseLink and Hibernate, they split into 2 sets of components: Property (holds name, nullability, type, default value, etc) and the type, where it normally hosts a column or association.
I don’t see the value personally of making such change since it increases the allocation size for the sake of 2 properties (name and nullability)
So ideally, we make a design decision of supporting on Property and we force people to map partitioned by column as a field in their mapping. Not the end of the world, it was not supported before, so now there’d be a way… much better, not ideal

3- ValueGenerationPlan would have to be a per-class support (one instance per class), since we need to determine if it needs to map which properties needs to be executed on.

guilhermeblanco:
Feel free to remove the "WIP" label, so I can track what is mergeable once the label goes out. =)
You'll still have to notify me that you did it.

Majkl578:
Closing this - after some discussion, replaced by a different approach in #6585.

PR #6556, opened 2017-07-16
Update ConsoleRunner's command list to include L2 cache and DBAL reserved keywords commands

lcobucci:
These commands were missing and they're quite important...

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 2.6
 
+## Minor BC BREAK: `Doctrine\ORM\Tools\Console\ConsoleRunner` is now final
+
+Since it's just an utilitarian class and should not be inherited.
+
 ## Minor BC BREAK: removed `Doctrine\ORM\Query\Parser#isInternalFunction`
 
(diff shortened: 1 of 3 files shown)


Ocramius:
@lcobucci can we add a test checking the list of registered commands? Or is the console runner untestable?

lcobucci:
@Ocramius we can add an assertion like (for each command):
```php
self::assertTrue($app->has('dbal:run-sql'));
```
or:

```php
self::assertTrue($app->has((new \Doctrine\DBAL\Tools\Console\Command\RunSqlCommand())->getName()));
```

Maybe we can move the list of classes to a constant and reuse that constant to instantiate them and test things, like:

```php
foreach (ConsoleRunner::COMMAND_LIST as $command) {
    self::assertTrue($app->has((new $command())->getName()));
}
```

What do you prefer?

Ocramius:
The first one: ensures BC and we see the diff immediately if something
changes

Ocramius (approved):

Ocramius:
LGTM :ship:

PR #6557, opened 2017-07-17
Fix DDC2632.

enriquesomolinos:
This fix DDC2632 problem. I generate de nullable field for the foreingkey in the case you are using the tool to generate an xml or yml form a database.

To run the test you need to use a database that supports foreing keys like mysql.

When this PR gets closed I fix this problem for master branch.


Ocramius:
@enriquesomolinos please target `master` instead of `2.5`. Backporting is up to us.

PR #6569, opened 2017-07-21
Added a test case for postLoad on fetch-joined entities - fetch joined entities should have lifecycle events fired as well

stoccc:
see https://github.com/doctrine/doctrine2/issues/6568

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/LifecycleCallbackTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/LifecycleCallbackTest.php
@@ -296,12 +296,12 @@ FROM
 LEFT JOIN
     entA.entities AS entB
 WHERE
-    entA.id = %s
+    entA.id = :entA_id
 DQL;
 
         $fetchedA = $this
             ->_em
-            ->createQuery(sprintf($dql, $entA->getId()))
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/LifecycleCallbackTest.php
```diff
@@ -268,6 +268,47 @@ public function testCascadedEntitiesNotLoadedInPostLoadDuringIterationWithSimple
...
+
+        $dql = <<<'DQL'
+SELECT
+    entA, entB
+FROM
+    Doctrine\Tests\ORM\Functional\LifecycleCallbackCascader AS entA
+LEFT JOIN
+    entA.entities AS entB
+WHERE
+    entA.id = %s
```
Ocramius: Please use a named parameter instead of `sprintf()`

Ocramius:
Test looks good - let's see if it fails on CI. If that's the case, we got some hydrator hell to fix :-\

stoccc:
@Ocramius 
think you should remove Bug and Failing Test labels.
I closed the bug, you gained a test case if it can help.

Ocramius:
Marketing as `Improvement` then

Issue #6573, opened 2017-07-22
NewObjectExpression - multiple NEW syntax


lzakrzewski:
I would like to hydrate with DQL query an object with another object argument in the constructor. For example Money:
https://github.com/moneyphp/money/blob/master/src/Money.php#L57

So it should look sth like `NEW Money\Money(object.amount NEW Money\Currency(object.currency))`. It does not work. Are any chances to do it with doctrine?

Ocramius:
@lzakrzewski I think this was simply never attempted. Write a few test cases and then we can see what can be done :)

lzakrzewski:
Thank you @Ocramius PR: https://github.com/doctrine/doctrine2/pull/6574

PR #6574, opened 2017-07-23
#6573 multiple "NEW" operator syntax

lzakrzewski:
In some cases, it will be good to use multiple `NewObjectExpression`.

For example, I used [money](https://github.com/moneyphp/money) library for money calculations and I would like to return the price of each item in my bucket as a `Money` value object.

```
            SELECT
                new Doctrine\Tests\Models\DDC6573\DDC6573Money(
                    i.priceAmount,
                    new Doctrine\Tests\Models\DDC6573\DDC6573Currency(i.priceCurrency)
                )
            FROM
                Doctrine\Tests\Models\DDC6573\DDC6573Item i
```

See: #6573 

Right now I committed the only test.

### DIFF
--- a/tests/Doctrine/Tests/Models/DDC6573/DDC6573Currency.php
+++ b/tests/Doctrine/Tests/Models/DDC6573/DDC6573Currency.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Doctrine\Tests\Models\DDC6573;
+
+class DDC6573Currency
+{
+    /**
+     * Currency code.
+     *
+     * @var string
+     */
+    private $code;
+
+    public function __construct($code)
+    {
+        if (!is_string($code)) {
+            throw new \InvalidArgumentException('Currency code should be string');
+        }
+
+        $this->code = $code;
+    }
+
+    /**
+     * @return string
+     */
+    public function getCode()
+    {
+        return $this->code;
+    }
+}
--- a/tests/Doctrine/Tests/Models/DDC6573/DDC6573Item.php
+++ b/tests/Doctrine/Tests/Models/DDC6573/DDC6573Item.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Doctrine\Tests\Models\DDC6573;
+
+/**
+ * @Entity
+ * @Table(name="ddc6573_items")
+ */
+class DDC6573Item
+{
+    /**
+     * @Id
+     * @Column(type="integer")
+     * @GeneratedValue(strategy="AUTO")
+     */
+    public $id;
+
+    /**
+     * @Column(type="string", length=255)
+     */
+    public $name;
+
+    /**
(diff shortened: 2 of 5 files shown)


Majkl578:
I'm wondering how the tests could be passing when this behavior is not supported now... 😕

Ocramius:
> I'm wondering how the tests could be passing when this behavior is not supported now... :confused:

Big WTF indeed. @lzakrzewski did you expect this? :O

Bilge:
Why is it so hard to understand? He wrote `$this->markTestSkipped('Feature does not exist.');` at the head of test cases expected to fail.

Ocramius:
Meh, totally missed it =_=

@lzakrzewski can you remove the "test skipped" blocks?

lzakrzewski:
@Ocramius @Majkl578 sorry for misleading you. "Test skipped" blocks removed.

Majkl578:
Yep, I must be blind... 😶 
Anyway, I'd like to see this implemented - it may be a powerful tool for read-only layer separation to completely bypass full entity hydratation & management.
After a quick look: It's easy to support on parser level, but it will be more tricky to implement in hydrators since the code there assumes no DTOs could exist inside another DTOs. May even end up being rewritten/reimplented in order to have this. :/

TNAJanssen:
The parser is really easy to implement, is made it check for token type T_NEW.

The hydrator is a different story, I want to help but the issue also comes with the fact we don't know the nesting inside of the hydrator of somebody can help me with gathering this data I can continue to make an PR for this feature. 

`
        if ($token['type'] === Lexer::T_NEW) {
            $expression = $this->NewObjectExpression();

            return $expression;
        }
`

TNAJanssen:
https://github.com/doctrine/doctrine2/issues/4687

fesor:
@TNAJanssen I implemented this but I not sure is this implementation is ok:

https://github.com/fesor/doctrine2/commit/b48e6a4e642aaef0fcf8d0d17237368240fec688

 - In order to have this  instantiation should be moved from Array/Object hydrator to AbstractHydrator.
 - also we need additional information to store hierarchy of arguments (in ResultsSetMapping)

TNAJanssen:
@fesor indeed like that, do you already have it working?

fesor:
Yep, this commit contains all the implementation.

TNAJanssen:
@fesor tests are also successful? I will test it tomorrow!

fesor:
@TNAJanssen tests successful but I didn't used test cases in this PR. I also think that I need add few more test cases.

Majkl578:
Please create new PR for that and move discussion over there, it's not appropriate to discuss unrelated commits in this PR.

fesor:
Implementation for this PR: #6709

greg0ire:
Implemented in https://github.com/doctrine/orm/pull/11576

PR #6579, opened 2017-07-24
Document `string` as possible parameter type for `Expr#between()`

Nek-:
People may want to enter a date, or simply the parameter name (`:param`).

Also, linters may report an error while it's a good usage.


Ocramius:
LGTM thanks @Nek-!

PR #6580, opened 2017-07-24
Allow DBAL 2.6 and common 2.8 to be installed

Tobion:
DBAL 2.6 is released but currently dependencies can't be resolved as ORM 2.5 does not allow DBAL 2.6 and ORM 2.6 is not relased yet.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "php": ">=5.4",
         "ext-pdo": "*",
         "doctrine/collections": "~1.2",
-        "doctrine/dbal": ">=2.5-dev,<2.6-dev",
+        "doctrine/dbal": ">=2.5-dev,<2.7-dev",
         "doctrine/instantiator": "~1.0.1",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
         "php": ">=5.4",
         "ext-pdo": "*",
         "doctrine/collections": "~1.2",
-        "doctrine/dbal": ">=2.5-dev,<2.6-dev",
+        "doctrine/dbal": ">=2.5-dev,<2.7-dev",
         "doctrine/instantiator": "~1.0.1",
         "doctrine/common": ">=2.5-dev,<2.8-dev",
```
alcaeus: Could also bump this to `>=2.5-dev,<2.9-dev`

Tobion: done

Ocramius:
Test looks good - failures unrelated. Thanks @Tobion!

PR #6581, opened 2017-07-25
Travis enhancements: Use Build Stages, added PHP 7.2

Majkl578:
- migrated to [Build Stages](https://docs.travis-ci.com/user/build-stages)
- added PHP 7.2 to the chain
- added travis_retry to composer calls for cases when installation unexpectedly fails
- added --prefer-dist to composer for faster installation
- disable xdebug and only explicitly enable it for coverage job
- fixed executable flag on tests/travis/install-mysql-5.7.sh

It's also much more readable than before with tons of bash ifs (but it may be just my subjective opinion).


Ocramius:
SLICK! :+1:

lcobucci:
@Majkl578 that's awesome, thanks!

PR #6585, opened 2017-07-28
Added ValueGeneratorMetadata, implemented ValueGeneratorPlan

guilhermeblanco:


guilhermeblanco:
Meh... let's see what this breaks and fix. =)

Majkl578:
@guilhermeblanco Good plan indeed! 😃
Forgot to drop (now useless) AssignedGenerator, will push that directly separately...

PR #6586, opened 2017-07-28
Correctly process entities bound to query parameters, conver them to their identifier values when their metadata is not yet available in-memory

fancyweb:
Fixed issue : https://github.com/doctrine/doctrine2/issues/6073

Better late than never :-)

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -19,10 +19,11 @@
 
 namespace Doctrine\ORM;
 
-use Doctrine\Common\Util\ClassUtils;
+use Doctrine\Common\Persistence\Mapping\MappingException;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Collections\ArrayCollection;
 
+use Doctrine\ORM\Mapping\MappingException as ORMMappingException;
 use Doctrine\ORM\Query\Parameter;
 use Doctrine\ORM\Cache\QueryCacheKey;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
@@ -419,17 +420,16 @@ abstract class AbstractQuery
 
         if (is_object($value)) {
             try {
-                $this->_em->getMetadataFactory()->getMetadataFor(ClassUtils::getClass($value));
-            } catch (\Exception $e) {
-
-            }
-
-            if (!isset($e)) {
                 $value = $this->_em->getUnitOfWork()->getSingleIdentifierValue($value);
 
                 if ($value === null) {
                     throw ORMInvalidArgumentException::invalidIdentifierBindingEntity();
                 }
+            } catch (MappingException $e) {
+                // The object is not a mapped entity but we still want its processed value so we have to continue code
+                // execution
+            } catch (ORMMappingException $e) {
+
             }
         }
 
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -417,11 +417,19 @@ abstract class AbstractQuery
             return $value;
         }
 
-        if (is_object($value) && $this->_em->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
-            $value = $this->_em->getUnitOfWork()->getSingleIdentifierValue($value);
+        if (is_object($value)) {
+            try {
+                $this->_em->getMetadataFactory()->getMetadataFor(ClassUtils::getClass($value));
+            } catch (\Exception $e) {
 
-            if ($value === null) {
-                throw ORMInvalidArgumentException::invalidIdentifierBindingEntity();
+            }
+
+            if (!isset($e)) {
+                $value = $this->_em->getUnitOfWork()->getSingleIdentifierValue($value);
+
+                if ($value === null) {
+                    throw ORMInvalidArgumentException::invalidIdentifierBindingEntity();
+                }
             }
         }
 
--- a/tests/Doctrine/Tests/ORM/Functional/QueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/QueryTest.php
@@ -870,8 +870,9 @@ class QueryTest extends OrmFunctionalTestCase
 
     public function testProcessParameterValueWithObjects()
     {
-        $query  = $this->_em->createQuery();
+        $query = $this->_em->createQuery();
 
+        // non-mapped class
         $dt = new \DateTime();
         $this->assertSame(
             $dt,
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -417,11 +417,19 @@ public function processParameterValue($value)
             return $value;
         }
 
-        if (is_object($value) && $this->_em->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
-            $value = $this->_em->getUnitOfWork()->getSingleIdentifierValue($value);
+        if (is_object($value)) {
+            try {
+                $this->_em->getMetadataFactory()->getMetadataFor(ClassUtils::getClass($value));
```
Majkl578: I think you can safely replace this call by the contents of the `if (!isset($e))` here. `UnitOfWork::getSingleIdentifierValue()` will call `ClassMetadataFactory::getMetadataFor()` right away (and throw if class is not a valid entity).

Majkl578 (changes requested):

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -417,11 +417,19 @@ public function processParameterValue($value)
...
+            } catch (\Exception $e) {
```
fancyweb: Then I also need to catch `Doctrine\ORM\Mapping\MappingException` exceptions because when I run the tests, all the exceptions caught are from this class.

lib/Doctrine/ORM/AbstractQuery.php
fancyweb: Good code improvement indeed. I think the case is already tested under the "non-mapped class" comment with a \DateTime object.

tests/Doctrine/Tests/ORM/Functional/QueryTest.php
```diff
@@ -867,4 +867,40 @@ public function testMultipleJoinComponentsUsingLeftJoin()
...
+        );
+
+        $user = new CmsUser();
+        $user->name = 'Thomas';
+        $user->username = 'fancyweb';
+
+        $this->_em->persist($user);
+        $this->_em->flush();
+
+        $this->assertEquals(true, $this->_em->getMetadataFactory()->hasMetadataFor(CmsUser::class));
```
Ocramius: `self::assertTrue()`

tests/Doctrine/Tests/ORM/Functional/QueryTest.php
```diff
@@ -867,4 +867,40 @@ public function testMultipleJoinComponentsUsingLeftJoin()
...
+
+        // non-loaded metadata
+        $this->assertEquals(
+            $user->getId(),
+            $query->processParameterValue($user)
+        );
+
+        $this->_em->getMetadataFactory()->setMetadataFor(CmsUser::class, null);
```
Ocramius: This is a bit of a hack... I can imagine that the test will break if we change something, unless this is a mock metadata factory?

tests/Doctrine/Tests/ORM/Functional/QueryTest.php
```diff
@@ -867,4 +867,40 @@ public function testMultipleJoinComponentsUsingLeftJoin()
...
+        $this->assertEquals(false, $this->_em->getMetadataFactory()->hasMetadataFor(CmsUser::class));
```
Ocramius: `self::assertFalse()`

tests/Doctrine/Tests/ORM/Functional/QueryTest.php
```diff
@@ -867,4 +867,40 @@ public function testMultipleJoinComponentsUsingLeftJoin()
         $this->assertInstanceOf(CmsUser::class, $users[2]);
         $this->assertNull($users[3]);
     }
+
+    public function testProcessParameterValueWithObjects()
+    {
+        $query = $this->_em->createQuery();
+
+        // non-mapped class
```
Ocramius: Split the test in two methods, instead of testing first against a value type, then against an entity type.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -417,11 +418,18 @@ public function processParameterValue($value)
...
-        if (is_object($value) && $this->_em->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
-            $value = $this->_em->getUnitOfWork()->getSingleIdentifierValue($value);
+        if (is_object($value)) {
+            try {
+                $value = $this->_em->getUnitOfWork()->getSingleIdentifierValue($value);
+
+                if ($value === null) {
+                    throw ORMInvalidArgumentException::invalidIdentifierBindingEntity();
+                }
+            } catch (MappingException $e) {
```
Ocramius: Is this location hit when using `DateTime`, for example?

Ocramius: If that's the case, there is a performance impact

tests/Doctrine/Tests/ORM/Functional/QueryTest.php
```diff
@@ -867,4 +867,40 @@ public function testMultipleJoinComponentsUsingLeftJoin()
...
+        $this->assertEquals(
+            $user->getId(),
+            $query->processParameterValue($user)
+        );
+
+        $this->_em->getMetadataFactory()->setMetadataFor(CmsUser::class, null);
+        $this->assertEquals(false, $this->_em->getMetadataFactory()->hasMetadataFor(CmsUser::class));
+
+        // loaded metadata
+        $this->assertEquals(
```
Ocramius: `self::assertSame()` - also add an assertion message

Ocramius (changes requested):

lib/Doctrine/ORM/AbstractQuery.php
fancyweb: Yes there is a performance impact because we are now trying to get the identifier value of every object processed by this method. And it tries to load the metadata everytime. However, doing this might be the only solution to actually check if an object is an entity or not. Unless someone has a better way of doing it, we have to choose between performance loss or not fixing this bug I guess.

tests/Doctrine/Tests/ORM/Functional/QueryTest.php
fancyweb: Yes this is not that great but this is the only solution I have found to "clean" the metadata of a loaded entity. I added an assertion just after to ensure that the metadata are not loaded after this method call, so if anything changes the test will fail. Note that I also mistakenly exchanged the comments an that it tests first the loaded metadata then the non-loaded metadata.

lib/Doctrine/ORM/AbstractQuery.php
Ocramius: I think the performance impact is not worth the bugfix. This would mean automatic file I/O for anyone having a file-based metadata driver configured.

That said, I have no other/better solutions right now :-\

This will probably have to stay in a limbo unless we fix `hasMetadataFor()` for good.

PR #6588, opened 2017-07-30
Updating docs to support for APCu as a cache driver

reypm:


docs/en/reference/caching.rst
```diff
@@ -63,6 +63,24 @@ by itself.
...
+
+Below is a simple example of how you could use the APCu cache driver
+by itself.
+
+.. code-block:: php
+
+    <?php
+    $cacheDriver = new \Doctrine\Common\Cache\ApcuCache();
+    $cacheDriver->save('cache_id', 'my_data');
+
```
greg0ire: I'd put it above the doc for APC, which is deprecated (maybe add a note about that, or even consider removing it?)

docs/en/reference/caching.rst
```diff
@@ -63,6 +63,24 @@ by itself.
     $cacheDriver = new \Doctrine\Common\Cache\ApcCache();
     $cacheDriver->save('cache_id', 'my_data');
 
+APCu
+~~~
```
greg0ire: Please add one tilde here

greg0ire:
APC is used in examples for the query cache or the result cache, should be replaced with APCu IMO: https://github.com/doctrine/doctrine2/blob/cd11723e6310b5973f50e7d9819c58916721d290/docs/en/reference/caching.rst

greg0ire:
Same here: https://github.com/doctrine/doctrine2/blob/cd11723e6310b5973f50e7d9819c58916721d290/docs/en/reference/metadata-drivers.rst

docs/en/reference/caching.rst
```diff
@@ -60,7 +60,25 @@ by itself.
 .. code-block:: php
 
     <?php
-    $cacheDriver = new \Doctrine\Common\Cache\ApcCache();
+    $cacheDriver = new \Doctrine\Common\Cache\ApcuCache();
```
greg0ire: You don't want to change that one.

Ocramius (approved):
LGTM :+1:

PR #6590, opened 2017-08-01
The collection is deleted twice at the post flush event

vtsykun:
Some collection maybe deleted twice at the post flush event, because UOW not clear calculated change set before invoke postFlush. So if there is event listener which calls flush in postFlush event, the collection will be deleted a twice. For example I added failing tests for this bug

Thanks,
Vladimir

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -412,6 +412,8 @@ class UnitOfWork implements PropertyChangedListener
             $coll->takeSnapshot();
         }
 
+        $this->collectionDeletions = [];
+
         $this->dispatchPostFlushEvent();
 
         // Clear up
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6590Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6590Test.php
@@ -1,0 +1,110 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\EventSubscriber;
+use Doctrine\ORM\Event\PostFlushEventArgs;
+
+class DDC6590Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+        $this->_schemaTool->createSchema(
+            [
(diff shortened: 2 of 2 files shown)


lcobucci:
I'll be closing this PR as `invalid` due to that reason.

vtsykun:
Hi @lcobucci,
I found more than 10 places in vendor where it used. I think it makes sense to test this cases, because "why not" if we can enesure that it does not cause to an infinite loop of recursion.
You can explain what other problems are caused by called `EntityManager#flush()` inside its listeners.

Ocramius:
@vtsykun generally, inconsistent `UnitOfWork` states.

Unless you are flushing a separate `UnitOfWork`, flushing inside a `flush()` call stack is not going to be supported.

vtsykun:
The problem of inconsistency of `UOW` states is fixed here https://github.com/doctrine/doctrine2/pull/1503, but of course this is BC. In order to avoid the BC that caused by these changes, it is possible to keep the states UOW (`entityInsertions`, `entityUpdates`, `entityDeletions` etc.) in the stack in the order of calling the flush. So when calling `flush` inside its listeners we will have such behavior as during normal calling `flush` and  the state of `UOW` will available in `postFlush` event.

But in any case, the current `Doctrine` behavior is dangerous. As a developer, I can not be sure that in the vendor code there will not be call `flush` inside its listeners, but is the cause of the appearance of [heisenbug](https://en.wikipedia.org/wiki/Heisenbug)

So exist two ways:
1) Doctrine should not ignore that it does not support. It is **MAJOR** BC and very reduces its capabilities.
2) Clear up state before call postFlush. Here it is possible to avoid the BC and add support the new feature

vtsykun:
Currently, only one way to fix the problem as in the example this PR tests :)
```php
class CleanUpUnitOfWorkListener
{
    /**
     * @param PostFlushEventArgs $args
     */
    public function postFlush(PostFlushEventArgs $args)
    {
        $uow = $args->getEntityManager()->getUnitOfWork();
        if ($uow->getScheduledCollectionDeletions()) {
            $reflect = new \ReflectionObject($uow);
            $prop = $reflect->getProperty('collectionDeletions');
            $prop->setAccessible(true);
            $prop->setValue($uow, []);
        }
    }
}
```

Ocramius:
Let me put it more clearly: flush in flush is not supported, and if you or your vendors do it in userland, the problem is on your side.

No BC, simply no safety around the UoW, as it is already doctrine internals.

Issue #6589, opened 2017-07-31
Synchronization of bidirectional association cannot be performed in any order


janedbal:
Let's have following entities:
```php
/** @ORM\Entity */
class RoomType {
	/** @ORM\OneToMany(targetEntity=Room::class, mappedBy="roomType") */
	private $rooms;

	public function addRoom(Room $room) {
		$this->rooms->add($room);
	}

	public function removeRoom(Room $room)	{
		$this->rooms->removeElement($room);
	}
}
/** @ORM\Entity */
class Room {
	/** @ORM\ManyToOne(targetEntity=RoomType::class, inversedBy="rooms") */
	private $roomType;

	public function moveTo(RoomType $roomType) {
		$oldRoomType = $this->roomType;
		$this->roomType = $roomType;
		$roomType->addRoom($this);
		$oldRoomType->removeRoom($this); // changes $this->roomType back to $oldRoomType !
	}
}
```
([full example](https://gist.github.com/janedbal/f910792334a51a561623d3d6fc5b5430))

Calling method **moveTo** will never update association from Room to new RoomType, because removeElement on PersistentCollection magically changes reference on `Room::$roomType` back to the old value. To make this work, you first have to remove old association and then create new like this:

```php
$this->roomType->removeRoom($this);
$this->roomType = $roomType;
$roomType->addRoom($this);
```

Is there any reason, why Doctrine behaves like this? From my point of view, this looks like a bug.

Ocramius:
> PersistentCollection magically changes reference on Room::$roomType back to the old value

A test is needed to verify this. The code as-is seems sane, and a unit test around the `Room` class probably shows that the values are correct when not involving Doctrine ORM, but with Doctrine ORM plus your code involved, we need some sort of proof that there is a problem.

See https://github.com/doctrine/doctrine2/tree/master/tests/Doctrine/Tests/ORM/Functional/Ticket for examples on such tests.

janedbal:
Ok, I'll send failing test to prove described behaviour.

janedbal:
@Ocramius Failing test added.

FireLizard:
I've got the same issue although `fetch="EXTRA_LAZY"` is set on the OneToMany association.
Are there any news on this?

PR #6591, opened 2017-08-01
#6589 Synchronization of bidirectional association cannot be performed in any order 

janedbal:
Failing test for #6589

### DIFF
--- a/lib/Doctrine/ORM/PersistentCollection.php
+++ b/lib/Doctrine/ORM/PersistentCollection.php
@@ -160,18 +160,31 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
     {
         $this->collection->add($element);
 
-        // If _backRefFieldName is set and its a one-to-many association,
-        // we need to set the back reference.
         if ($this->backRefFieldName && $this->association['type'] === ClassMetadata::ONE_TO_MANY) {
-            // Set back reference to owner
-            $this->typeClass->reflFields[$this->backRefFieldName]->setValue(
-                $element, $this->owner
-            );
-
-            $this->em->getUnitOfWork()->setOriginalEntityProperty(
-                spl_object_hash($element), $this->backRefFieldName, $this->owner
-            );
+            $this->completeBidirectionalAssociationLink($element);
         }
+    }
+
+    /**
+     * Sets the owner on the inverse side of the association, when the property was not set yet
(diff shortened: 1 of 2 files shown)


lcobucci:
@janedbal thanks for your test! I've managed to find the cause of the issue and came up with a solution, however I'd like to have an opinion from the other maintainers just to make sure we won't have any side effect with this change (it touches on some stuff that @guilhermeblanco optimised).

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -187,13 +200,8 @@ public function hydrateSet($key, $element)
...
         $this->collection->set($key, $element);
 
-        // If _backRefFieldName is set, then the association is bidirectional
-        // and we need to set the back reference.
         if ($this->backRefFieldName && $this->association['type'] === ClassMetadata::ONE_TO_MANY) {
-            // Set back reference to owner
-            $this->typeClass->reflFields[$this->backRefFieldName]->setValue(
-                $element, $this->owner
-            );
+            $this->completeBidirectionalAssociationLink($element);
```
lcobucci, line 204: No tests failed when I added this here, now it's also executing this (but I'm not sure if it should):

```php
$this->em->getUnitOfWork()->setOriginalEntityProperty(
    spl_object_hash($element),
    $this->backRefFieldName,
    $this->owner
);
```

guilhermeblanco: The looks weird to me. Conceptually, if you're `hydrateSet`ing, you should have the back reference defined.
However, if you're doing that for a new collection owner entity, the you should not have the original entity property value defined.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6589Test.php
```diff
@@ -0,0 +1,114 @@
...
+        parent::setUp();
+
+        $this->_schemaTool->createSchema(
+            [
+                $this->_em->getClassMetadata(GH6589Room::class),
+                $this->_em->getClassMetadata(GH6589RoomType::class),
+            ]
+        );
+
+        $roomType1 = new GH6589RoomType(1);
```
guilhermeblanco: This needs to be moved to the `testMoveTo` method.

PR #6592, opened 2017-08-01
Resolved target entity and repositories

wesleylancel:
This isn't an actual fix but a test to show some weirdness going on with repositories of resolved target entities.

The added test will fail on the first `assertSame` because the class name return will actually be `AbstractMappedSuperclass::class`. However, when line 120 is uncommented, the test will pass.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Tools/ResolveTargetEntityListenerTest.php
+++ b/tests/Doctrine/Tests/ORM/Tools/ResolveTargetEntityListenerTest.php
@@ -109,6 +109,20 @@ class ResolveTargetEntityListenerTest extends OrmTestCase
                 ->getSQL()
         );
     }
+
+    public function testResolveTargetEntityListenerCanRetrieveTargetEntityByAbstractClass()
+    {
+        $this->listener->addResolveTargetEntity(AbstractMappedSuperclass::class, ConcreteClass::class, []);
+        $this->listener->addResolveTargetEntity(TheInterface::class, ConcreteClass::class, []);
+
+        $this->em->getEventManager()->addEventSubscriber($this->listener);
+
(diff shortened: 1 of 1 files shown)


wesleylancel:
@Ocramius: I looked into this some further and think I know what's happening.

The resolve target entities listener works in two ways:
- when no metadata was found for a class it replaces it with the metadata of the target class (this works for the interface, but not for the abstract mapped superclass, because it has its own metadata)
- when the metadata of a target class is loaded its sets its for all classes that use it as a target (that's why it'll return the correct repository after uncommenting the line that loads metadata separately).

Not sure how to fix this as the metadata of the mapped superclass also needs to be loaded in order for the class that extends it to work.

tests/Doctrine/Tests/ORM/Tools/ResolveTargetEntityListenerTest.php
```diff
@@ -109,6 +109,20 @@ public function testDoesResolveTargetEntitiesInDQLAlsoWithInterfaces()
...
+    public function testResolveTargetEntityListenerCanRetrieveTargetEntityByAbstractClass()
+    {
+        $this->listener->addResolveTargetEntity(AbstractMappedSuperclass::class, ConcreteClass::class, []);
+        $this->listener->addResolveTargetEntity(TheInterface::class, ConcreteClass::class, []);
+
+        $this->em->getEventManager()->addEventSubscriber($this->listener);
+
+//        $this->factory->getMetadataFor(TheInterface::class);
+
+        $this->assertSame($this->em->getRepository(AbstractMappedSuperclass::class)->getClassName(), ConcreteClass::class);
```
lcobucci: The order of the arguments are wrong, the expectations should come first 😉

PR #6595, opened 2017-08-02
Removed the final comment on EntityManager.php

usmanzafar:
Removed the comment for "final class keyword" on EntityManager to avoid ambiguity. The commit to put this into comment was made in early 2013 but does not talk about why the final was commented. Though i really wanted to add a final back but i am assuming there must be a reason for it.

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -60,7 +60,7 @@ use Doctrine\Common\Util\ClassUtils;
  * @author  Jonathan Wage <jonwage@gmail.com>
  * @author  Roman Borschel <roman@code-factory.org>
  */
-/* final */class EntityManager implements EntityManagerInterface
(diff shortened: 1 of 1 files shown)


Majkl578:
`final` comment was added back in 2012 as part of #524 to indicate that extending EntityManager is *not* a valid extension point, but could not be enforced by adding the final modifier in 2.x as it would be a BC break. Valid extension point would be EntityManagerInterface, but due to BC constraints some places still rely on EntityManager, not the interface. Hopefully to be sorted out in 3.0/develop.

usmanzafar:
Fair enough @Majkl578 - however what are your thoughts on this PR, shall i close this or do you reckon we should remove this comment for ambiguity sake . Thanks

Ocramius:
This should be replaced with a patch making the `EntityManager` `final` in `develop`

Ocramius:
Closing here - please re-open a separate patch against `develop`, if applicable.

PR #6601, opened 2017-08-03
Fix double spacing in "cascade persist missing" exception message

martin-georgiev:
This is an absolute minor style improvement that concerns wrong double spacing on string concatenation.


Ocramius:
@martin-georgiev thanks, LGTM! :+1:

PR #6607, opened 2017-08-06
Be strict with risky tests

lcobucci:
We were missing a risky test for a while, so this PR makes PHPUnit fails if we have a risky test.


Ocramius (approved):
LGTM 👍

PR #6614, opened 2017-08-11
`UNIQUE constraint failed` errors when manipulating a `LAZY` collection and calling `toArray()` on it

Sysaninster:
See https://github.com/doctrine/doctrine2/issues/6613

PR #6616, opened 2017-08-11
#6613 #6614 correct handling of new elements added to an initialized (yet dirty) collection

### LINKED ISSUE #6613: There is the fail "UNIQUE constraint failed" with a LAZY collection
I have the User entity with the LAZY collection field. If I do this code:

```php
        $user = new User();
        $user->id = 1;
        $this->_em->persist($user);
        $this->_em->flush();

        $this->_em->clear();

        /** @var User $user */
        $user = $this->_em->find(User::class, 1);
        $phone1 = new Phone();
        $phone1->id = 1;
        $user->phones->add($phone1);
        $this->_em->persist($phone1);
        $this->_em->flush();

        $phone2 = new Phone();
        $phone2->id = 2;
        $user->phones->add($phone2);
        $this->_em->persist($phone2);
        $user->phones->toArray();
        $this->_em->flush();
```

I will get the error:

```
Doctrine\DBAL\Exception\UniqueConstraintViolationException: An exception occurred while executing 'INSERT INTO user_phone (user_id, phone_id) VALUES (?, ?)' with params [1, 1]:

SQLSTATE[23000]: Integrity constraint violation: 19 UNIQUE constraint failed: user_phone.user_id, user_phone.phone_id
```

If I remove the *$user->phones->toArray();* line then it will work without a problem.

Ocramius:
Fixes #6613
Fixes #6614

This is a *very* bad bug. The fix introduces some ovehead, but the scenario is usually not that common, so it should be fine.

Ideally, we should keep `new` and `dirty` states completely separate, but that's not possible without completely changing also the base class (`Doctrine\Common\Collections\AbstractLazyCollection`)


lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -692,23 +692,41 @@ public function unwrap()
...
+     *       is because we want to prevent using `array_udiff()`, which is likely to cause very
+     *       high overhead (complexity of O(n^2)). `array_diff_key()` performs the operation in
+     *       core, which is faster than using a callback for comparisons
+     */
+    private function restoreNewObjectsInDirtyCollection(array $newObjects)
+    {
+        $loadedObjects               = $this->collection->toArray();
+        $newObjectsByOid             = array_combine(array_map('spl_object_hash', $newObjects), $newObjects);
+        $loadedObjectsByOid          = array_combine(array_map('spl_object_hash', $loadedObjects), $loadedObjects);
+        $newObjectsThatWereNotLoaded = array_diff_key($newObjectsByOid, $loadedObjectsByOid);
```
alcaeus: Clever...I'm totally stealing this one.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6613Test.php
```diff
@@ -0,0 +1,107 @@
...
+class DDC6613InverseSide
+{
+    /** @Id @Column(type="string") */
+    private $id;
+
+    public function __construct()
+    {
+        $this->id = uniqid('phone', true);
+    }
+}
```
alcaeus: Needs trailing newline 😛

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -692,23 +692,41 @@ public function unwrap()
...
         // Reattach NEW objects added through add(), if any.
-        if ($newObjects) {
-            foreach ($newObjects as $obj) {
-                $this->collection->add($obj);
-            }
+        array_walk($newObjectsThatWereNotLoaded, [$this->collection, 'add']);
 
-            $this->isDirty = true;
-        }
+        $this->isDirty = true;
```
alcaeus: The collection shouldn't be marked as dirty if no items were added after initializing.

alcaeus (changes requested):

lib/Doctrine/ORM/PersistentCollection.php
Ocramius: It actually should: it was loaded, but it was never flushed. I also had a previous assertion for that, see https://github.com/doctrine/doctrine2/pull/6616/commits/1fc7f81741be84bbce6f92c61d6f6c2b4a3e573f

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6613Test.php
```diff
@@ -0,0 +1,107 @@
+<?php
```
Ocramius: This entire test should probably be ported to `PersistentCollectionTest` and stripped of any `UnitOfWork` interactions

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6613Test.php
Ocramius: I'll probably get rid of the test once I convert it to a unit test

lib/Doctrine/ORM/PersistentCollection.php
alcaeus: Assume you add the following code after https://github.com/doctrine/doctrine2/commit/1fc7f81741be84bbce6f92c61d6f6c2b4a3e573f#diff-c32ade803f1349ec4da7da8644879ef6R65:
```
        $this->_em->flush();

        self::assertFalse($phones->isInitialized());
        self::assertFalse($phones->isDirty());

        $phones->add($item2);

        self::assertFalse($phones->isInitialized());
        self::assertTrue($phones->isDirty());

        $phones->initialize();
 
        self::assertTrue($phones->isInitialized());
        self::assertFalse($phones->isDirty());
```

After the initialize, the only item that was added to the uninitialized collection was already present after the collection was initialized. Thus, after initializing, the collection is not dirty.

Ocramius: That element does not exist in the DB though, so the collection does indeed need to go through persistence at the next `flush()` call.

If the collection wasn't dirty after this initialization, the object association would be completely lost (other tests break too if I do that, btw).

alcaeus: If you were to flush after adding `$item2`, then re-add `$item2` to the (now uninitialized) collection. Once the collection is initialized, there will be no element to be added in `restoreNewObjectsInDirtyCollection`, but the collection will be marked as dirty, even though it wasn't modified.

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -692,23 +692,41 @@ public function unwrap()
...
         // Has NEW objects added through add(). Remember them.
-        $newObjects = [];
+        $newlyAddedDirtyObjects = [];
 
         if ($this->isDirty) {
-            $newObjects = $this->collection->toArray();
+            $newlyAddedDirtyObjects = $this->collection->toArray();
         }
 
         $this->collection->clear();
```
guilhermeblanco: I think the issue could be solved by only not calling `clear()` here. 
IIRC, `UnitOfWork` already handles dirty elements in the collection and does not replace them when hydrating in `BasicEntityPersister::loadCollectionFromStatement` (which is where all calls go at the end).

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -692,23 +692,41 @@ public function unwrap()
...
+     *
+     * Note: the only reason why this entire looping/complexity is performed via `spl_object_hash`
+     *       is because we want to prevent using `array_udiff()`, which is likely to cause very
+     *       high overhead (complexity of O(n^2)). `array_diff_key()` performs the operation in
+     *       core, which is faster than using a callback for comparisons
+     */
+    private function restoreNewObjectsInDirtyCollection(array $newObjects)
+    {
+        $loadedObjects               = $this->collection->toArray();
+        $newObjectsByOid             = array_combine(array_map('spl_object_hash', $newObjects), $newObjects);
```
Ocramius: These variables are redundant

lib/Doctrine/ORM/PersistentCollection.php
Ocramius: That's not the case: removing the `clear()` causes this test to fail, as well as another 3:

```
There were 4 failures:

1) Doctrine\Tests\ORM\Functional\ManyToManyBasicAssociationTest::testRetrieveManyToManyAndAddMore
Failed asserting that 4 matches expected 3.

/home/ocramius/Documents/doctrine/doctrine2/tests/Doctrine/Tests/ORM/Functional/ManyToManyBasicAssociationTest.php:201

2) Doctrine\Tests\ORM\Functional\SecondLevelCacheOneToManyTest::testCacheInitializeCollectionWithNewObjects
Failed asserting that actual size 3 matches expected size 4.

/home/ocramius/Documents/doctrine/doctrine2/tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheOneToManyTest.php:377

3) Doctrine\Tests\ORM\Functional\Ticket\DDC1643Test::testNotCloneAndPassAroundFlush
Failed asserting that 2 matches expected 3.

/home/ocramius/Documents/doctrine/doctrine2/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC1643Test.php:118

4) Doctrine\Tests\ORM\Functional\Ticket\DDC6613Test::testFail
Failed asserting that actual size 3 matches expected size 2.

/home/ocramius/Documents/doctrine/doctrine2/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6613Test.php:71
```

Ocramius: I think the current solution is a decent one, as it keeps the responsibility of re-attaching the `NEW` records within the `PersistentCollection` implementation. Distributing it around multiple objects may be better for performance, but is actually a mess to track on the long term

lib/Doctrine/ORM/PersistentCollection.php
Ocramius: @alcaeus re-read the test carefully:

 1. add
 2. flush (collection clean, *no initialization happening*)
 3. add
 4. initialize (collection dirty, since we added and we didn't flush)
 5. flush

Ocramius: @alcaeus I converted these tests into more isolated `PersistentCollectionTest` scenarios: please check them.

Also, should we drop this integration test?

alcaeus: @Ocramius new tests look good. I believe we can drop the integration test.

I've also pushed `testWillNotMarkCollectionAsDirtyAfterInitializationIfNoElementsWereAdded` to show the issue I was mentioning above: after initializing, there are no new elements in the collection, but it is still marked as dirty. Feel free updating either the test or the collection logic to make it pass.

Ocramius:
Merged:

`master`: doctrine/doctrine2@633c8461c8cb7ee176249a8f55cbeabc9789e726
`2.5`: doctrine/doctrine2@095611c4b655db70aaaf41f63d7e22990bb13814

PR #6618, opened 2017-08-11
Quick wins by removal of `count`-related operations

Ocramius:


lib/Doctrine/ORM/Mapping/ClassMetadata.php
```diff
@@ -128,7 +128,7 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
      * READ-ONLY: The field names of all fields that are part of the identifier/primary key
      * of the mapped entity class.
      *
-     * @var array
+     * @var string[]
```
guilhermeblanco: I'd change it to be `array<int, string>` instead... =)

PR #6619, opened 2017-08-11
Remove expensive array iteration operations

Ocramius:


Ocramius:
YOLO :ship:

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -240,11 +240,12 @@ public function getSnapshot()
...
     {
-        return array_udiff_assoc(
-            $this->snapshot,
-            $this->collection->toArray(),
-            function($a, $b) { return $a === $b ? 0 : 1; }
-        );
+        $collectionItems = $this->collection->toArray();
+
+        return \array_values(\array_diff_key(
+            \array_combine(\array_map('spl_object_hash', $this->snapshot), $this->snapshot),
```
drupol, line 246: Note for later, maybe we should use `spl_object_id` as mentionned here: https://github.com/doctrine/orm/pull/6878

greg0ire: It's in the backport todo list, yes :)

carnage:
Tried pulling this one into latest 2.10.x but it seems that the commits no longer exist in a fresh clone of the repo. Perhaps @Ocramius still has them locally, otherwise I'm stuck to recreating via copy/paste from github as fresh commits.

greg0ire:
I think you won't find them by hash, but you can find them by commit message. At least it was doable last time I tried :)

greg0ire:
Lloks like they were applied several times, in fact :sweat_smile: 
```
git log --oneline --all|grep 'Replacing possible'
c6116d4e5 Replacing possible O(n^2) operation from `PersistentCollection` diffing operations
4fc48a66b Replacing possible O(n^2) operation from `PersistentCollection` diffing operations
200e53a27 Replacing possible O(n^2) operation from `PersistentCollection` diffing operations
```

:thinking: that must be because my clone isn't fresh. Anyway old-prototype-3.x is still up and must have those commits, but has been rebased.

simonberger:
This PR has like 30 changed lines. 8)
What I did in a similar, bigger case is to grab a patch from the PR and apply it.

https://github.com/doctrine/orm/pull/6619.patch

carnage:
I was trying to find an option to grab the patch from the PR yesterday and gave up, thanks for the link :D

PR #6620, opened 2017-08-12
fixed undefined variable

Powerhamster:
renamed $condition to $conditions so $isConditionalJoin is working


Powerhamster:
replaced with https://github.com/doctrine/doctrine2/pull/6621

PR #6621, opened 2017-08-12
Fixed usage of undefined variable in `empty()` expression in `SqlWalker`

Powerhamster:
changed $conditions to $condition so $isConditionalJoin is working

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -1143,7 +1143,7 @@ class SqlWalker implements TreeWalker
                 $dqlAlias   = $joinDeclaration->aliasIdentificationVariable;
                 $tableAlias = $this->getSQLTableAlias($class->table['name'], $dqlAlias);
                 $condition = '(' . $this->walkConditionalExpression($join->conditionalExpression) . ')';
-                $isUnconditionalJoin = empty($conditions);
+                $isUnconditionalJoin = empty($condition);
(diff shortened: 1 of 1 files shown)


Ocramius:
I must say that I have issues wrapping my head around the fact that builds fail on `2.5` and passed in this patch: at this point, it must be a caching issue.

Still, this is why you don't use `empty()`, kids!

Thanks @Powerhamster, I'm doing a fixup release with this patch in it ASAP

Ocramius:
Related:

#6464
#6475

This bug was introduced by an incorrect backport performed by me.

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -1143,7 +1143,7 @@ public function walkJoin($join)
                 $dqlAlias   = $joinDeclaration->aliasIdentificationVariable;
                 $tableAlias = $this->getSQLTableAlias($class->table['name'], $dqlAlias);
                 $condition = '(' . $this->walkConditionalExpression($join->conditionalExpression) . ')';
-                $isUnconditionalJoin = empty($conditions);
+                $isUnconditionalJoin = empty($condition);
```
jvasseur: Wouldn't this be always true given the previous line always add brackets to the `$condition` variable meaning it will never be empty ?

Ocramius: @jvasseur yes, that's not an issue though, as the code was cleaned up on `master`

Powerhamster: Should I try to fix it?

Ocramius: @Powerhamster no - ORM 2.5.8 is green and doesn't need improvement here.

PR #6623, opened 2017-08-13
Ensure that `AbstractHydrator#hydrateAll()` detaches the hydrator from the event manager

echo511:
I think someone overlooked hydrateAll() method and edited only iterate(). https://github.com/doctrine/doctrine2/pull/1515

As a result an event listener is not added which can cause problems if EventManager implementation does not check for existence of a listener upon its removal.

In my case it causes E_NOTICE (undefined index) because of https://github.com/Kdyby/Events/blob/master/src/Events/EventManager.php#L198

If my PR should be altered (commit message, etc.) please let me know.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -143,6 +143,9 @@ abstract class AbstractHydrator
         $this->_rsm   = $resultSetMapping;
         $this->_hints = $hints;
 
+        $evm = $this->_em->getEventManager();
+        $evm->addEventListener(array(Events::onClear), $this);
+
         $this->prepare();
 
         $result = $this->hydrateAllData();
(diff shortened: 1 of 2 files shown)


Ocramius:
@echo511 this still requires a test case

echo511:
@Ocramius OK, I'll do it as a fix of DDC-3146 and its test case? Where should it go? On top of 2.5.8?

Ocramius:
@echo511 all patches should go against `master` (this one included) -  backporting is up to us

Ocramius (approved):
LGTM! Preparing for merge, thanks @echo511!

Ocramius:
@echo511 I manually merged and backported to `2.5`, thanks!

`master`: doctrine/doctrine2@57a9509dec83fec705dde1792510b6420dfbcb02
`2.5`: doctrine/doctrine2@1554af0c07db130b8fd03d40c4a2a67ac1bc3110

PR #6630, opened 2017-08-16
Remove reliance on Symfony\Component\Yaml deprecated constants

Ocramius:
See https://travis-ci.org/doctrine/doctrine2/jobs/265260226#L546-L554 for reasons - constant is gone in 4.0, and that makes static analysis (obviously) crash.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,8 +24,11 @@
         "doctrine/annotations": "~1.4",
         "symfony/console": "~3.0|~4.0"
     },
+    "conflict": {
+        "symfony/yaml": "<3.4"
+    },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -24,8 +24,11 @@
         "doctrine/annotations": "~1.4",
         "symfony/console": "~3.0|~4.0"
     },
+    "conflict": {
+        "symfony/yaml": "<3.4"
```
Ocramius: People can really hurt themselves if they install ORM 2.6 + Yaml <3.4, so this is now here.

alcaeus (approved):
👍

Ocramius:
:ship:

Issue #3788, opened 2014-03-11
DDC-3020: simplexml_load_file(): I/O warning: failed to load external in XmlDriver.php


doctrinebot:
Jira issue originally created by user rmatrono:

PHP Warning:  simplexml_load_file(): I/O warning : failed to load external entity &quot;/path-to/doctrine/entities/mappings/my_entity.dcm.xml&quot; in /path-to/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php on line 711

PHP bug:
https://bugs.php.net/bug.php?id=62577

Possible solution:
https://github.com/owncloud/core/pull/7498/files

layanto:
This is an issue even with no external entities in orm mapping. 
See https://groups.google.com/forum/m/#!topic/doctrine-user/vAb_3zcz6vU

Ocramius:
@layanto as explained above, this has nothing to do with the ORM itself.

layanto:
@ocramius agree that the root of the problem is not Doctrine ORM, but rather it is a bug in PHP/libxml2. However, this bug has not been resolved for over 3 years and still exists in the latest versions of php5.6 and php7. Doctrine xml mapping will fail regardless of whether external entities are used or not (FosUserBundle xml mapping has no external entities) when libxml_disable_entity_loader is set to true.

If you do a Google search on "failed to load external entity symfony" or "failed to load external FosUserBundle" you will find quite a few people having issues due to Doctrine mapping failed to load xml mapping (mostly of FosUserBundle). Most people describe the problem as randomly happening, unexplained and resolved by restarting PHP (which will reset libxml_disable_entity_loader back to false).

As you described above, xml mapping is not a valid attack vector as such should be safe for Doctrine xml driver to set libxml_disable_entity_loader to false prior to loading xml mapping and then reset libxml_disable_entity_loader setting back (exactly what OwnCloud did to work around this bug https://github.com/owncloud/core/pull/7498/files).

Just wanting to say that despite being nothing to do with the ORM itself, it is a PHP/libxml bug that affects Doctrine xml mapping loader which, in my opinion, Doctrine should work around this bug. Will save people a lot of head scratching why they code randomly stopped working (as this bug is not threadsafe, another application on the same server can trigger this issue - in my case RSS app tt-rss.org).

Thanks for responding to a closed issue. At least I now know what to do to work around this issue. Just trying to save others the trouble.

[...]

PR #6633, opened 2017-08-18
XmlDriver: Avoid PHP bug #62577 by avoiding simplexml_load_file.

apapsch:
Doctrine is affected by PHP bug #62577. simplexml_load_file is not
able to load files if libxml_disable_entity_loader(true) has been
called. simplexml_load_file fails with the message:

I/O warning : failed to load external entity "/my/mappings/my_entity.dcm.xml"
in /path-to/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php on line 711

This error occurs even if there are no external entities in the XML file.

Waiting for the PHP bug to be resolved is infeasible, because it is
unresolved since years. Therefore Doctrine needs to circumvent the bug
by replacing simplexml_load_file with simplexml_load_string while getting
the file contents itself. simplexml_load_string is not affected by the
PHP bug.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -838,7 +838,7 @@ class XmlDriver extends FileDriver
     protected function loadMappingFile($file)
     {
         $result = [];
-        $xmlElement = simplexml_load_file($file);
+        $xmlElement = simplexml_load_string(file_get_contents($file));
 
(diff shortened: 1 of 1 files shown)


apapsch:
This would solve #3788 .

Ocramius:
Merged, thanks @apapsch!

`master`: doctrine/doctrine2@a426808a16dc7cfacea0e83587a4e71e6a5d1af0
`2.5`: doctrine/doctrine2@fb7b78c0048e570b3b671ef2342ff9aeb27e6ae7

PR #6640, opened 2017-08-22
#6217 #6284 when hydrating an entity with a composite primary key that is both an `EAGER` and a `LAZY` association and also cached, the `DefaultQueryCache` tries to pass L2 cache implementation detail objects to the `UnitOfWork` 

Ocramius:
Fixes #6217
Fixes #6284

This is a proper version of #6284, fixing the deeper problem.

Basically, what is happening is that you have a L2 cached entity class like following:

```php
class A
{
    /** @Id @ManyToOne(targetEntity=b) */
    private $b;
    /** @Id @ManyToOne(targetEntity=c)*/
    private $c;
}
```

And a DQL query like following:

```sql
SELECT a, b
FROM A a
JOIN a.b b
```

The L2 cache for `A.c` won't have any information to cache, but it will still contain the identifier for `A.c`. This means that for that association, a `AssociationCacheEntry` exists anyway, but no definition in the `QueryCacheEntry` on how to process it.

I also tried fixing the problem on the other side (where the `QueryCacheEntry` is produced, rather than consumed), but overfitting it when no association data is actually available causes empty `AssociationCacheEntry` objects to be hydrated into empty entities (really bad!).

This is just a shotgun patch to fix the problem by iterating over all possible association fields, so if you have any better ideas, please do let me know.

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -92,13 +92,13 @@ class DefaultQueryCache implements QueryCache
             return null;
         }
 
-        $entry = $this->region->get($key);
+        $cacheEntry = $this->region->get($key);
 
-        if ( ! $entry instanceof QueryCacheEntry) {
+        if ( ! $cacheEntry instanceof QueryCacheEntry) {
             return null;
         }
 
-        if ( ! $this->validator->isValid($key, $entry)) {
+        if ( ! $this->validator->isValid($key, $cacheEntry)) {
             $this->region->evict($key);
 
             return null;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -210,6 +210,25 @@ public function get(QueryCacheKey $key, ResultSetMapping $rsm, array $hints = []
...
+                // cache key information in `$cacheEntry` will not contain details
+                // for fields that are associations.
+                //
+                // This means that `$data` keys for some associations that may
+                // actually not be cached will not be converted to actual association
+                // data, yet they contain L2 cache AssociationCacheEntry objects.
+                //
+                // We need to unwrap those associations into proxy references,
+                // since we don't have actual data for them except for identifiers.
+                if ($unCachedAssociationData instanceof AssociationCacheEntry) {
```
lcobucci: It makes sense IMO, but I feel like we could do that conversion while iterating `$entry['associations']` in https://github.com/doctrine/doctrine2/pull/6640/files#diff-c5343cb0b202102a59de56dfdcb8ee47R147 to avoid iterating the entire object data.

lcobucci (changes requested):

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
Ocramius: `$entry['associations']` does not contain the keys we're iterating upon here.

I considered using an accumulator of already processed keys, but since these are usually about associations, and the number of fields and associations is (typically) < 10 on any entity, looping is actually faster than aggregating the keys and excluding them here.

Ocramius:
:ship:

PR #6646, opened 2017-08-24
add test for corrupted entry in the cache

jocel1:


lcobucci:
@jocel1 I've moved your test to doctrine/cache#231 and applied the fix, so I'll close this PR - since it's not related to the ORM. Thanks for your contribution!

Issue #6638, opened 2017-08-21
Changes to one-to-one relations get reversed when the inverse side is fetched again using findBy


johanderuijter:
I found some unexpected behaviour with one to one relations. I couldn't find any information on it, so i hope it's ok i opened an issue about it here.

My problem is the following:
I have 2 entities, with a bidirectional one to one relation. When i remove the relation between them (e.g. set the properties to null) the relation is removed. However, when the inverse side is fetched again (after setting the relation to null, but before a flush) the relation reappears. This is not the case for 'regular' properties which remain changed.

In my particular use case, this is a problem as the symfony unique constraint uses the repositories findBy to check if an entity already exists. In doing so, it restores the relation.

Here is a code example based on the [bidirectional one to one mapping from the docs](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/association-mapping.html#one-to-one-bidirectional):

Given the following two entities:
``` php
/**
 * @Entity
 */
class Customer
{
    /**
     * @Id
     * @Column(type="integer")
     * @GeneratedValue
     */
    private $id;

    /**
     * @Column(type="string")
     */
    private $name;

    /**
     * @OneToOne(targetEntity="Cart", mappedBy="customer", cascade={"all"})
     */
    private $cart;

    public function setName(string $name): void
    {
        $this->name = $name;
    }

    public function setCart(?Cart $cart): void
    {
        $prevCart = $this->cart;

        if ($prevCart) {
            $prevCart->setCustomer(null);
        }

        if ($cart) {
            $cart->setCustomer($this);
        }

        $this->cart = $cart;
    }

    public function getCart(): ?Cart
    {
        return $this->cart;
    }

    // ...
}

/**
 * @Entity
 */
class Cart
{
    /**
     * @Id
     * @Column(type="integer")
     * @GeneratedValue
     */
    private $id;

    /**
     * @OneToOne(targetEntity="Customer", inversedBy="cart")
     * @JoinColumn(name="customer_id", referencedColumnName="id")
     */
    private $customer;

    // ...
}
```
with the following schema:
```sql
CREATE TABLE Cart (id INTEGER NOT NULL, customer_id INTEGER DEFAULT NULL, PRIMARY KEY(id));
CREATE UNIQUE INDEX UNIQ_AB9127899395C3F3 ON Cart (customer_id);
[...]

PR #6648, opened 2017-08-25
#6638 one-to-one association is unexpectedly refreshed when using `findBy` or `findOneBy` repository methods 

johanderuijter:
Failing test case for #6638

Both `findBy` and `findOneBy` have the same effect where the relation suddenly returns.

Using either `find` or a DQL query work as expected (e.g. the relation remains null after)

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6638Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6638Test.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace Doctrine\Tests\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH6638Test extends OrmFunctionalTestCase
+{
+    public function setUp() : void
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(GH6638Customer::class),
+            $this->_em->getClassMetadata(GH6638Cart::class),
+        ]);
+    }
+
+    public function testFetchingOfOneToOneRelations() : void
+    {
+        $initialCustomer = new GH6638Customer();
+
+        $initialCart = new GH6638Cart();
+        $initialCustomer->cart = $initialCart;
+        $initialCart->customer = $initialCustomer;
+
+        $this->_em->persist($initialCustomer);
+        $this->_em->persist($initialCart);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $repository = $this->_em->getRepository(GH6638Customer::class);
+
+        $customer = $repository->find($initialCustomer->id);
+
+        $this->assertInstanceOf(GH6638Cart::class, $customer->cart);
+
+        $customer->cart = null;
+
+        $this->assertNull($customer->cart);
+
+        $repository->findBy(['id' => $initialCustomer->id]);
+
+        $this->assertNull($customer->cart);
+    }
+}
+
+/**
+ * @Entity
+ */
+class GH6638Customer
+{
+    /**
+     * @Id
+     * @Column(type="integer")
+     * @GeneratedValue
+     */
+    public $id;
+
+    /**
+     * @OneToOne(targetEntity="GH6638Cart", mappedBy="customer")
+     */
+    public $cart;
+}
+
+/**
+ * @Entity
+ */
+class GH6638Cart
+{
+    /**
+     * @Id
+     * @Column(type="integer")
+     * @GeneratedValue
+     */
+    public $id;
+
+    /**
+     * @OneToOne(targetEntity="GH6638Customer", inversedBy="cart")
+     * @JoinColumn()
+     */
+    public $customer;
+}


Ocramius:
Notes to self:

Did some initial digging: the issue is only reproducible with a bidirectional one-to-one association, as the reset to `null` is applied due to it being the inverse side of the association. The owning side of the association here is the source of truth.

So far, I [traced it down to this location in the code](https://github.com/doctrine/doctrine2/blob/61404e2d6d71a6d46331f2fa1771dad121e49058/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php#L434)

The ORM sees `null` on a field, so it decides to hydrate the association regardless of the correctness of the state.

johanderuijter:
Not sure of this matters for this issue, but in my initial example i set both sides of the relation to null (which i forgot to do here).

When i apply to following diff, in the end, both sides of the relation are refreshed.

```diff
diff --git a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6638Test.php b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6638Test.php
index 50f5d09df..afb672f8c 100644
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6638Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6638Test.php
@@ -32,16 +32,20 @@ final class GH6638Test extends OrmFunctionalTestCase
         $repository = $this->_em->getRepository(GH6638Customer::class);
 
         $customer = $repository->find($initialCustomer->id);
+        $cart = $customer->cart;
 
         $this->assertInstanceOf(GH6638Cart::class, $customer->cart);
 
         $customer->cart = null;
+        $cart->customer = null;
 
         $this->assertNull($customer->cart);
+        $this->assertNull($cart->customer);
 
         $repository->findBy(['id' => $initialCustomer->id]);
 
         $this->assertNull($customer->cart);
+        $this->assertNull($cart->customer);
     }
 }
```

Does this change anything? Do you want me to update the PR with this diff?

Ocramius:
> both sides of the relation are refreshed.

That's no good either.

What seems to happen is that the hydrator ignores the fact that the entity already is hydrated, and therefore keeps hydrating even if the `Query::HINT_REFRESH` is not given.

johanderuijter:
I wish i could be of more help, but most of the doctrine internals seems like black magic to me. If you told me it works by [sacrificing an intern every monday](http://theunderfold.com/wp-content/uploads/2017/06/2017-06-19-Server-Tips.jpg) i'd believe you.

Regardless, i've been trying to understand the issue a little better...

So far i've found out that you CAN recreate this using dql. In my earlier tests i did `SELECT c FROM Doctrine\Tests\Functional\Ticket\GH6638Customer`, but because of reasons beyond my understanding a `findBy` will do a fetch join on the one-to-one relationship. 
When i do a dql query with a fetch join like this `SELECT c, cc FROM Doctrine\Tests\Functional\Ticket\GH6638Customer c LEFT JOIN c.cart cc` it has the same effect as `findBy` and will refresh the relationship.

Unfortunately, i'm unable to figure out what the exact effect of a fetch join is on the hydration process.

Ocramius:
@johanderuijter the problem is quite simple: the `ObjectHydrator` (among its complexity, which is mind-boggling) has a design flaw.

Given a query like following:

```sql
SELECT a, b FROM A a JOIN a.b b
```

And a resultset like following:

```
a_id, a_field1, b_id
1, "foo", 2
```

And an in-memory entity (registered as managed) `new A(1, "foo", null /* this is where B should be */)`

It will loop over the resultset and fetch the current fields of the `A` instance.

The problem is specifically with `null`: if the hydrator detects that the field `b` is `null` in the following check:

https://github.com/doctrine/doctrine2/blob/61404e2d6d71a6d46331f2fa1771dad121e49058/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php#L434

It will overwrite the field, as if it was never loaded.

This check is too naive, as the correct check would be "is this entity already loaded or built by someone else, or did I build it?".

That ` ! $reflFieldValue` expression needs to become something like `if ($this->didIReallybuildThis($parentObject)) {`

Ocramius:
This is really just an edge-case with `null`: this behavior would be perfectly OK with `$this->_hints[Query::HINT_REFRESH]` being set, but it is not when the field is simply `null`

Ocramius:
By the way: sacrificing interns works too, I just don't have one available right now.

Ocramius:
@johanderuijter I found an intern to sacrifice and I started work on https://github.com/doctrine/doctrine2/pull/6649: can you please try it out?

It is not an acceptable patch as-is, but it's a start, as it correctly identifies the issue.

johanderuijter:
I wont be able to test it in my original use case until monday, but i tried it out in a different setting and as far as i can tell this fixes the problem. For all use cases i could find (`findBy`, `findOneBy` and a DQL query with a fetch join) the one-to-one relation is no longer refreshed.

Thank you for looking into this so swiftly and taking the time to explain the problem to me in more detail!

PR #6651, opened 2017-08-26
Replacing direct `quote` model-set schema creation with actually existing model-set setup

Ocramius:
This was causing random test failures due to the schema tool raising exceptions while running completely unrelated tests or just a subset of the test suite.


Ocramius:
:ship: (yes, tests pass)

PR #6653, opened 2017-08-28
Drop `Doctrine\ORM\Version`, using `PackageVersions\Version::getVersion("doctrine/orm")` instead, which is more reliable

Ocramius:
This removes the need to:

 * bump the release version each time, manually (or via `build.xml`)
 * figure out which version the users are running (just ask them to run a CLI command)


tests/Doctrine/Tests/ORM/Tools/SetupTest.php
```diff
@@ -37,10 +31,9 @@ public function tearDown()
 
         set_include_path($this->originalIncludePath);
         $loaders = spl_autoload_functions();
```
alcaeus: Remove: `$loaders` is no longer used below.

alcaeus (approved):

Majkl578:
I'm afraid this is a **HUGE** BC break even for 3.0. There is A LOT of code checking for versions this way - either for specific version or whether ORM is installed or not...
https://github.com/search?l=PHP&q=%22Doctrine%5CORM%5CVersion%22&type=Code&utf8=%E2%9C%93

kimhemsoe:
Would it make sense to keep the class Version but replace the constant with a static function which proxied PackageVersions\Version::getVersion() ?

Seems alot of packages are looking for Version to check if the ORM is installed. 
Or do we have recommended way to check for ORM?

Ocramius:
@Majkl578 since we are bumping to ORM v3, checking for the existence of `Doctrine\ORM\Version` existence is to be considered unsafe anyway. This is an intentional BC break, as nobody (in no library) should ever need to design such a `Version` class anymore, now that `ocramius/package-versions` exists.

The only reason why I didn't PR this earlier is because the lib only works on 7.0+, which wasn't the case for us (yet).

The `version_compare()` stuff was our fault btw: we changed annotation setup in ORM 2.1 to 2.2, so that was an actual BC break that was mitigated via the introduction of a series of checks. It then got copy-pasted, which is why you see so many hits on the Github search.

SenseException:
`Version` itself has a redundant functionality between the Doctrine repositories and only existed because of the constant, so thumbs up. This should be considered in the other repos too.

I assume ORM will depend on major release versions of common, dbal and the other dependencies?

Ocramius:
> I assume ORM will depend on major release versions of common, dbal and the other dependencies?

Yep, that will be required once we start working on the new `ObjectManager` and `ObjectRepository` API

Ocramius:
@alcaeus merge whenever you want :+1:

Ocramius:
Handled in #2967

Issue #6661, opened 2017-09-01
Handling DateTimes-Cookbook is … error-prone regarding Timezones


heiglandreas:
The DateTime-Cookbook favours to store all DateTimes by converting them to UTC and store them in the Database in UTC. That is a possible approach, but it can (and will) lead to problems as soon as the olson-DB changes. That will lead to differing offsets resulting in wrong DateTimes when hydrating them from the database. 

I have not yet found a good solution using the ORM (that's why there isn't a PR by now) but I think it is unwise to give advice that can lead to errors without stating that.

For more information on the topic feel free to have a look at [a blogpost I wrote](https://andreas.heigl.org/2016/12/22/why-not-to-convert-a-datetime-to-timestamp/)

Ocramius:
Can you suggest a storage format to be used?

heiglandreas:
Best storage-format IMHO would be *local* datetime and timezone in 2 separate fields. Important IMHO is that the datetime is not converted to UTC but left in local time. But that would require 2 DB-fields for one PHP-Object. dehydration would be like this:

``` 
function dehydrate(DateTimeInterface $datetime) 
{
    $dtStore = $datetime->format($platform->getDateTimeFormatString());
    $tzStore = $datetime->getTimezone()->getName();
}
```

and hydration on the other side would be like this:

```
function hydrate(string $dtStorage, string $tzStorage) : DateTime
{
    return new DateTime($dtStorage, new DateTimeZone($tzStorage));
}
```

That would also allow users to use in-DB timezone handling as I explained in these articles for [Postgres](https://andreas.heigl.org/2016/05/29/timezones-and-postgresql/) and [MySQL](https://andreas.heigl.org/2016/04/18/timezones-and-mysql/)

Currently I have no clue as to how one could bring that into Doctrine. I'm happy for all pointers on where to start or how to achieve that…

PR #6662, opened 2017-09-01
6661: Improve timezone handling documentation

heiglandreas:
This commit alters the page to better handle datetimes with timezones. The result will store the datetime with the local date as datetime-value and the timezone as string-value. That way the database-specific ways of doing datetime arithmentics with timezones (see examples for [MySQL](https://andreas.heigl.org/2016/04/18/timezones-and-mysql/) and [Postgres](https://andreas.heigl.org/2016/05/29/timezones-and-postgresql/)) are still possible, the datetime will not be prone to errors due to timezone-changes between creation and the actual datetime and still the application will be able to use a generic datetime-object that can then be converted to a local timezone of the current user.

### DIFF
--- a/docs/en/cookbook/working-with-datetime.rst
+++ b/docs/en/cookbook/working-with-datetime.rst
@@ -67,8 +67,8 @@ The problem is simple. Not a single database vendor saves the timezone, only the
 However with frequent daylight saving and political timezone changes you can have a UTC offset that moves
 in different offset directions depending on the real location.
 
-The solution for this dilemma seems simple. Don't use timezones with DateTime and Doctrine 2. However there are 2 workarounds
-that even allows correct date-time handling with timezones depending on your use case:
+The solution for this dilemma seems simple: don't use timezones with DateTime(Immutable) and Doctrine 2. However there are some
+workarounds that allow correct date-time handling with timezones depending on your use case:
 
 1a. Don't convert DateTimes to UTC
 1b. Always convert any DateTime instance to UTC
--- a/docs/en/cookbook/working-with-datetime.rst
+++ b/docs/en/cookbook/working-with-datetime.rst
@@ -84,9 +84,13 @@ But as soon as you are handling datetimes that are more than a few days in advan
 Handling DateTimes with timezone-informations
 _____________________________________________
 
-Say we have an international calendaring application where users can add events that occur at different places worldwide and therefore at different timezones. To determine the exact time an event will happen means to save both the local time of the event and the timezone the event will happen in.
+Say we have an international calendaring application where users can add events that occur at different places
+worldwide and therefore at different timezones. To determine the exact date and time an event will happen means to save
+both the local time of the event and the timezone the event will happen in.
 
-Using the defalt DateTimeType will store the DateTime in the local time but without the timezone information. Therefore we need to store the timezone information as well and also need to provide a way to get the datetime back from the database with the correct timezone-information.
+Using the default datetime-type will store the datetime in the local time but without the timezone information.
+Therefore we need to store the timezone information as well and also need to provide a way to get the datetime
+back from the database with the correct timezone-information.
 
 To be able to transform these values back into their real timezone we have to save the timezone in a separate field of the entity requiring timezoned datetimes:
 
--- a/docs/en/cookbook/working-with-datetime.rst
+++ b/docs/en/cookbook/working-with-datetime.rst
@@ -185,7 +185,7 @@ An alternative would be to use a post-processing lifecycle event to recreate the
             $this->timezone = $eventDateTime->getTimeZone()->getName();
         }
 
-        /** @PostLoad */
+        /** @ORM\PostLoad */
         public function correctTimezone(): void
         {
             $this->eventDateTime = new DateTimeImmutable(
--- a/docs/en/cookbook/working-with-datetime.rst
+++ b/docs/en/cookbook/working-with-datetime.rst
@@ -68,7 +68,7 @@ The problem is simple. Not a single database vendor saves the timezone, only the
 However with frequent daylight saving and political timezone changes you can have a UTC offset that moves
 in different offset directions depending on the real location.
 
-The solution for this dilemma seems simple: don't use timezones with DateTime(Immutable) and Doctrine 2. However there are some
+The solution for this dilemma seems simple: don't use timezones with DateTime(Immutable) and Doctrine ORM. However there are some
 workarounds that allow correct date-time handling with timezones depending on your use case:
 
 1a. Don't convert DateTimes to UTC
@@ -126,43 +126,82 @@ field of the entity requiring timezoned datetimes:
         /** @ORM\Column(type="string") */
         private string $timezone;
 
-        /**
-         * @var bool
-         */
-        private $localized = false;
-
         public function __construct(DateTimeInterface $eventDateTime)
         {
-            $this->localized = true;
             $this->eventDateTime = $eventDateTime;
             $this->timezone = $eventDateTime->getTimeZone()->getName();
         }
 
-        /**
-         * During hydration $this->eventDateTime will set with a new DateTimeImmutable instance with the servers
-         * default timezone. So when getting the eventDateTime for the first time after hydration we need to
-         * make sure that the correct timezone is set. Therefore we reset the property with a correctly created
-         * DateTimeImmutable object.
-         *
-         * DBALTypes sadly currently can not take care of this during hydration due to the fact that they are not
-         * allowing to access multiple fields.
-         */
         public function getEventDateTime()
         {
-            if ($this->localized === false) {
+            return new DateTimeImmutable(
+                $this->eventDateTime->format('Y-m-d H:i:s'),
+                new DateTimeZone($this->timezone)
+            );
+        }
+    }
 
-                $this->eventDateTime = new DateTimeImmutable(
-                    $this->eventDateTime->format('Y-m-d H:i:s'),
-                    new DateTimeZone($this->timezone)
-                );
-                $this->localized = true;
-            }
+During hydration $this->eventDateTime will set with a new DateTimeImmutable instance with the servers
+default timezone. So when getting the eventDateTime we need to make sure that the correct timezone is set.
+Therefore we return a correctly created DateTimeImmutable object in the `getEventDateTime()`-method
+
+DBALTypes sadly currently can not take care of this during hydration due to the fact that they are not
+allowing to access multiple fields.
+
+An alternative would be to use a post-processing lifecycle event to recreate the object in the right way like this:
+
+.. code-block:: php
+
+    <?php
+    namespace Calendaring;
+
+    use DateTimeImmutable;
+    use DateTimeInterface;
+    use DateTimezone;
+    use Doctrine\ORM\Mapping as ORM;
+
+    /**
+     * @ORM\Entity
+     * @ORM\HasLifecycleCallbacks
+     */
+    class Event
+    {
+        /**
+         * @ORM\Id
+         * @ORM\GeneratedValue
+         * @ORM\Column(type="integer")
+         */
+        private ?int $id;
+
+        /** @ORM\Column(type="datetime") */
+        private DateTimeInterface $eventDateTime;
+
+        /** @ORM\Column(type="string") */
+        private string $timezone;
+
+        public function __construct(DateTimeInterface $eventDateTime)
+        {
+            $this->eventDateTime = $eventDateTime;
+            $this->timezone = $eventDateTime->getTimeZone()->getName();
+        }
+
+        /** @PostLoad */
+        public function correctTimezone(): void
+        {
+            $this->eventDateTime = new DateTimeImmutable(
+                $this->eventDateTime->format('Y-m-d H:i:s'),
+                new DateTimeZone($this->timezone)
+            );
+        }
+
+        public function getEventDateTime(): DateTimeImmutable
+        {
             return $this->eventDateTime;
         }
     }
 
-Using this way of handling timezones allows you also to use the database-specific ways of
-doing DateTime-arithmetics with the appropriate timezones. Make sure though that the database
+Using these ways of handling timezones allow you also to use the database-specific ways of
+doing DateTime-arithmetics in SQL with the appropriate timezones. Make sure though that the database
 always has the latest version of the timezone-database when you use these features.
 
 Handling log-like DateTimes that shall be converted to UTC
--- a/docs/en/cookbook/working-with-datetime.rst
+++ b/docs/en/cookbook/working-with-datetime.rst
@@ -80,7 +80,9 @@ Wait: **What? Shall we convert to UTC or not?**
 
 That depends on what kind of DateTimes you are handling! When you are handling current dates like log-entries or
 datetimes that are within one or two days of the current date, you can - and even should - convert them to UTC.
-But as soon as you are handling datetimes that are more than a few days in advance (or back) you should **not** convert them to UTC but instead keep them in their local timezone. Why? You might want to read up on `why not to convert a datetime to a timestamp<https://andreas.heigl.org/2016/12/22/why-not-to-convert-a-datetime-to-timestamp/>`_
+But as soon as you are handling datetimes that are more than a few days in advance (or back) you should **not**
+convert them to UTC but instead keep them in their local timezone. Why? You might want to read up on
+`why not to convert a datetime to a timestamp<https://andreas.heigl.org/2016/12/22/why-not-to-convert-a-datetime-to-timestamp/>`_
 
 Handling DateTimes with timezone-informations
 _____________________________________________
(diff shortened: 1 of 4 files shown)


heiglandreas:
This PR would fix #6661

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -177,7 +113,12 @@ requiring timezoned datetimes:
         public function getCreated()
         {
             if (!$this->localized) {
-                $this->created->setTimeZone(new \DateTimeZone($this->timezone));
+                
+                $class = $this->created::class;
```
Ocramius: `::class` is not an actual constant, and will not resolve at runtime.

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -177,7 +113,12 @@ requiring timezoned datetimes:
...
+                $this->created = new $class(
```
Ocramius: I suggest using one of either `DateTimeImmutable` or `DateTime`. Since DBAL 2.6, `DateTimeImmutable` is viable, so I suggest just going with that rather than with a generic example

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -67,84 +67,20 @@ The problem is simple. Not a single database vendor saves the timezone, only the
 However with frequent daylight saving and political timezone changes you can have a UTC offset that moves
 in different offset directions depending on the real location.
 
-The solution for this dilemma is simple. Don't use timezones with DateTime and Doctrine 2. However there is a workaround
+The solution for this dilemma seems simple. Don't use timezones with DateTime and Doctrine 2. However there is a workaround
 that even allows correct date-time handling with timezones:
 
-1. Always convert any DateTime instance to UTC.
```
Ocramius: I suggest adding a section that links to your article about the original problem. Most people will try looking for a simple "convert to UTC" solution, and they will mess it up.

Ocramius (changes requested):
Overall good improvement, but IMO the `UTC` approach should be perfectly viable for logging events (specifically in the past, while the issue with zone time changes is for future data).

A typical requirement is to be able to sort data based on a `DATETIME` field: could you think about how to integrate that into the article?

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -58,25 +58,88 @@ Handling different Timezones with the DateTime Type
...
-in Databases <http://derickrethans.nl/storing-date-time-in-database.html>`_.
+however let me crush your expectations fast. Each database (that is supported by Doctrine 2) handles timezones in a different way - if they handle it at all! If you don't believe me, you can find more details on that in different blogposts about `the general aproach <http://derickrethans.nl/storing-date-time-in-database.html>`_ and - more up-to-date - about timezone-handling in `Postgres <https://andreas.heigl.org/2016/05/29/timezones-and-postgresql/>`_ and `MySQL <https://andreas.heigl.org/2016/04/18/timezones-and-mysql/>`_. 
 
 The problem is simple. Not a single database vendor saves the timezone, only the differences to UTC.
 However with frequent daylight saving and political timezone changes you can have a UTC offset that moves
 in different offset directions depending on the real location.
 
-The solution for this dilemma is simple. Don't use timezones with DateTime and Doctrine 2. However there is a workaround
-that even allows correct date-time handling with timezones:
+The solution for this dilemma seems simple. Don't use timezones with DateTime and Doctrine 2. However there are 2 workarounds
```
lcobucci: > The solution for this dilemma seems simple. Don't use timezones with DateTime and Doctrine 2.

Maybe is better?:

The solution for this dilemma seems simple: don't use timezones with DateTime and Doctrine 2.

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -58,25 +58,88 @@ Handling different Timezones with the DateTime Type
...
+that even allows correct date-time handling with timezones depending on your use case:
 
-1. Always convert any DateTime instance to UTC.
-2. Only set Timezones for displaying purposes
+1a. Don't convert DateTimes to UTC
```
lcobucci: Not sure if it's me, but this is a bit confusing... we mention "2 workarounds" and then present a list with 3 (4) items but these workarounds are actually explained as sub-sections of that area. 😕

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -58,25 +58,88 @@ Handling different Timezones with the DateTime Type
...
+datetimes that are within one or two days of the current date, you can - and even should - convert them to UTC. 
+But as soon as you are handling datetimes that are more than a few days in advance (or back) you should **not** convert them to UTC but instead keep them in their local timezone. Why? You might want to read up on `why not to convert a datetime to a timestamp<https://andreas.heigl.org/2016/12/22/why-not-to-convert-a-datetime-to-timestamp/>`_
+
+Handling DateTimes with timezone-informations
+_____________________________________________
+
+Say we have an international calendaring application where users can add events that occur at different places worldwide and therefore at different timezones. To determine the exact time an event will happen means to save both the local time of the event and the timezone the event will happen in.
+
+Using the defalt DateTimeType will store the DateTime in the local time but without the timezone information. Therefore we need to store the timezone information as well and also need to provide a way to get the datetime back from the database with the correct timezone-information.
```
lcobucci: Typo: `defalt` -> `default`

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -58,25 +58,88 @@ Handling different Timezones with the DateTime Type
...
+        /**
+         * @var bool
+         */
+        private $localized = false;
+
+        public function __construct(\DateTimeInterface $eventDateTime)
+        {
+            $this->localized = true;
+            $this->eventDateTime = $eventDateTime;
+            $this->timezone = $$eventDateTime->getTimeZone()->getName();
```
lcobucci: Typo `$$eventDateTime` -> `$eventDateTime`

lcobucci (changes requested):
It's going on good direction IMO, just some fixes/comments 😄

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -58,25 +58,88 @@ Handling different Timezones with the DateTime Type
...
 
 The problem is simple. Not a single database vendor saves the timezone, only the differences to UTC.
 However with frequent daylight saving and political timezone changes you can have a UTC offset that moves
 in different offset directions depending on the real location.
 
-The solution for this dilemma is simple. Don't use timezones with DateTime and Doctrine 2. However there is a workaround
-that even allows correct date-time handling with timezones:
+The solution for this dilemma seems simple. Don't use timezones with DateTime and Doctrine 2. However there are 2 workarounds
+that even allows correct date-time handling with timezones depending on your use case:
```
Majkl578: _that allow_

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -58,25 +58,88 @@ Handling different Timezones with the DateTime Type
...
+
+.. code-block:: php
+
+    <?php
+    namespace Calendaring;
+
+    /**
+     * @Entity
+     */
+    class Event
```
Majkl578: Strictly speaking, this is not a valid entity since it has no identifier, can we add one (or make one field one)?

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -58,25 +58,88 @@ Handling different Timezones with the DateTime Type
...
+            $this->localized = true;
+            $this->eventDateTime = $eventDateTime;
+            $this->timezone = $$eventDateTime->getTimeZone()->getName();
+        }
+
+        public function getEventDateTime()
+        {
+            if (!$this->localized) {
+                
+                $this->eventDateTime = new DateTimeImmutable(
```
Majkl578: This will trigger entity update as a side effect, that is not desired.

Majkl578 (changes requested):
I don't particulary like the naming inconsistencies in the text - once it's _DateTime_, then _datetime_, then _date-time_. Also we have DateTimeImmutable for a while which is actually preferred nowadays. Can we sync it to one variant?

lcobucci:
I've removed this issue from `2.6` since we can update the docs anytime

landsman:
This would be great!

docs/en/cookbook/working-with-datetime.rst
heiglandreas: Thanks for noting. I've changed that.

docs/en/cookbook/working-with-datetime.rst
heiglandreas: Thanks for noting. I've modified that part accordingly

docs/en/cookbook/working-with-datetime.rst
heiglandreas: How can that be avoided? I'm not too deep into that, so any help would be appreciated!

docs/en/cookbook/working-with-datetime.rst
heiglandreas: I was copy/pasting the example beginning at line 16 here which also does not have an identifier. Shall I add one everywhere? Or shall we skip that here as it is not relevant to the topic illustrated here?

docs/en/cookbook/working-with-datetime.rst
heiglandreas: I've modified that part

heiglandreas:
Hey folks: Can we get on with this somehow? There are some questions unanswered for some time now. So are these still a blocker? Or do we just close the PR and go on?

I'd rather not see the last option as the currently available information on the docs are actually harming users as the proposed solution will sooner or later harm users by introducing situations where wrong data is stored.

SenseException:
@heiglandreas The branches changed between the last changes and the new default branch approach. Can you please change the base branch of `heiglandreas:patch-1` to `2.9.x`? Then let's take a fresh look into this PR.

docs/en/cookbook/working-with-datetime.rst
SenseException: I suggest to add them in this file. In some past PRs we also added some missing lines and the identifier doesn't take much space that could distract from the core information. Since there are namespaces in use, some of the annotations need to be included with `use` too

docs/en/cookbook/working-with-datetime.rst
SenseException: What is the method supposed to do? Its name suggests just to return a datetime value but it changes `$eventDateTime` property.

docs/en/cookbook/working-with-datetime.rst
heiglandreas: This should be addressed now

docs/en/cookbook/working-with-datetime.rst
heiglandreas: Added the reasoning in a code-comment.

I'd rather remove that but that would mean that there needs to be a way to modify that value correctly during hydration which currently doesn't work as I can only access the current value while converting from a DB value to a PHP value. But I would also need access to all the other values to be able to get the timezone string :frowning:

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -67,53 +68,151 @@ The problem is simple. Not a single database vendor saves the timezone, only the
 However with frequent daylight saving and political timezone changes you can have a UTC offset that moves
 in different offset directions depending on the real location.
 
-The solution for this dilemma is simple. Don't use timezones with DateTime and Doctrine ORM. However there is a workaround
-that even allows correct date-time handling with timezones:
+The solution for this dilemma seems simple: don't use timezones with DateTime(Immutable) and Doctrine 2. However there are some
```
SenseException:
```suggestion
The solution for this dilemma seems simple: don't use timezones with DateTime(Immutable) and Doctrine ORM. However there are some
```

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -67,53 +68,151 @@ The problem is simple. Not a single database vendor saves the timezone, only the
...
+                    new DateTimeZone($this->timezone)
+                );
+                $this->localized = true;
+            }
+            return $this->eventDateTime;
+        }
+    }
+
+Using this way of handling timezones allows you also to use the database-specific ways of
+doing DateTime-arithmetics with the appropriate timezones. Make sure though that the database
```
SenseException: People might confuse that the timezone-database is related to ORM.

```suggestion
doing DateTime-arithmetics in SQL with the appropriate timezones. Make sure though that the database
```

docs/en/cookbook/working-with-datetime.rst
SenseException: Instead of writing the datetime back into the `$eventDateTime`, it could be returned directly. It doesn't have to return the same instance, does it? This way the update trigger can be prevented.

Or instead of having a boolean `$localized` property maybe you can store the created DateTimeImmutable in that property (With a different name) if a same instance matters.

What do you think about putting the comment below this class example? It seems like it belongs to the documentation itself than into a comment.

SenseException (changes requested):
Thank you for coming back to this PR. :-)

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -67,53 +68,190 @@ The problem is simple. Not a single database vendor saves the timezone, only the
...
+        /** @ORM\Column(type="string") */
+        private string $timezone;
+
+        public function __construct(DateTimeInterface $eventDateTime)
+        {
+            $this->eventDateTime = $eventDateTime;
+            $this->timezone = $eventDateTime->getTimeZone()->getName();
+        }
+
+        /** @PostLoad */
```
SenseException:
```suggestion
        /** @ORM\PostLoad */
```

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -67,53 +68,190 @@ The problem is simple. Not a single database vendor saves the timezone, only the
...
+        public function correctTimezone(): void
+        {
+            $this->eventDateTime = new DateTimeImmutable(
```
SenseException: Writing a new DateTimeImmutable into `$this->eventDateTime` will trigger an update for the entities when `$entityManager->flush()` is called. Do you need this alternative `Event` entity example?

SenseException (changes requested):

docs/en/cookbook/working-with-datetime.rst
heiglandreas: As this function is done via a `PostLoad`-Action I assumed that it does **not** trigger the update during the `flush`. But I then might have missunderstood [the docs regarding lifecycle-callbacks]( https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/events.html#lifecycle-callbacks). Espechially as the hydrated content is indeed not the one  the class was initially created with as there seems to be no way to actually get the `eventDateTime` property into the initial state without the `PostLoad` action.

Any ideas how to solve that dilemma?

heiglandreas: I just did a quick check. As far as I understand [the code](https://github.com/doctrine/orm/blob/2.9.x/lib/Doctrine/ORM/UnitOfWork.php#L2941), the `PostLoad` event listeners are run as the last part of the hydration process. Therefore I expect that the entity is not marked as changed as the `originalEntityData` array is populated with the data **after** the hydration finished completely. And therefore contained the already changed Data.

So that process should indeed be the preffered one. Or I completely misunderstood the code...

SenseException: Yes, having the following code in an Entity

``` php
    /** @ORM\PostLoad */
    public function fooLoad(): void
    {
        $this->bar = 'lol';
    }
```

and running 


``` php
$foo = $this->entityManager->find(\App\Entity\Foo::class, 1);
```

Will write "lol" into the mapped table column of `$this->bar`.

---

There's also a note about combined usage of `PostLoad` and `Doctrine\ORM\AbstractQuery#toIterable()` that is not considered safe.
https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/events.html#lifecycle-events

tests/Doctrine/Tests/Examples/DateTimeExampleTest.php
```diff
@@ -0,0 +1,54 @@
...
+        $schemaTool = new SchemaTool($this->entityManager);
+
+        $cmf = $this->entityManager->getMetadataFactory();
+        $classes = $cmf->getMetadataFor(DateTimeInstance::class);
+
+        $schemaTool->dropDatabase();
+        $schemaTool->createSchema([$classes]);
+    }
+
+    public function testDateTimeWithTimezoneExample(): void
```
SenseException, line 30: Is the test currently a work in progress?

SenseException:
@heiglandreas I'm not sure if you got my last question or if you just didn't find some time to work further on this PR.

heiglandreas:
No worries @SenseException: Life just got in the way. And I'm currently trying to wrap my head around the Embeddable idea. The test is indeed a "work in progress" and was a way to figure out whether the "PostLoad" actually triggers before or after setting the value in the UoW-cache. Which it does afterwards. Which IMO is breaking the idea of hydration but that is a completely separate discussion that we sxhould not have here in this PR :wink:

SenseException:
Thanks for the update. Feel free to catch on when you're ready. :-)

PR #6663, opened 2017-09-01
Use Generators instead of AppendIterator for properties iteration

Majkl578:
As discussed on Slack, AppendIterator doesn't work with inheritance depth >2, would have to be flattened somehow or FlattenIterator been created. Generators solve this issue in a cleaner way.

Tests with assertArrayHasKey had to be adjusted (some of them should probably use `hasProperty()` instead), strange it worked with AppendIterator. :)

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ComponentMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/ComponentMetadata.php
@@ -108,11 +108,13 @@ abstract class ComponentMetadata
     }
 
     /**
-     * @return \ArrayIterator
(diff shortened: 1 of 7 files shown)


Ocramius (changes requested):
> AppendIterator doesn't work with inheritance depth >2

This is missing an additional test case

Majkl578:
Yes, pretty much like whole ComponentMetadata or the new metadata system as whole...

Ocramius:
I didn't see pushes on the tests here 😋

PR #6683, opened 2017-09-03
Supply default values for optional values in sequence mapping if omitted by driver.

### LINKED ISSUE #6682: SequenceGenerator crashes with non-numeric value encountered when using XML Driver
Using XML Driver causes crash at https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Id/SequenceGenerator.php#L81 during addition operation, ~because parser casts sequence parameters as strings, not integers.~

Edit: because optional default values were missing.

alextech:
Fix #6682
~~Cannot figure out how to get XML to run properly from Ticket directory as in contributor guide. If this is a strict requirement will try again.~~ Figured out.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -3003,13 +3003,21 @@ class ClassMetadataInfo implements ClassMetadata
      */
     public function setSequenceGeneratorDefinition(array $definition)
     {
-        if ( ! isset($definition['sequenceName'])) {
+        if ( ! isset($definition['sequenceName']) || trim($definition['sequenceName']) === '') {
             throw MappingException::missingSequenceName($this->name);
         }
 
         if ($definition['sequenceName'][0] == '`') {
             $definition['sequenceName']   = trim($definition['sequenceName'], '`');
             $definition['quoted'] = true;
+        }
+
+        if ( ! isset($definition['allocationSize']) || trim($definition['allocationSize']) === '') {
+            $definition['allocationSize'] = '1';
+        }
+
+        if ( ! isset($definition['initialValue']) || trim($definition['initialValue']) === '') {
+            $definition['initialValue'] = '1';
         }
 
         $this->sequenceGeneratorDefinition = $definition;
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6682Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6682Test.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Doctrine\Test\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH6682Test extends OrmFunctionalTestCase
+{
+    /**
+     * @group 6682
+     */
+    public function testIssue() : void
+    {
+        $parsedDefinition = [
+            'sequenceName'   => 'test_sequence',
+            'allocationSize' => '',
+            'initialValue'   => '',
+        ];
+
+        $classMetadataInfo = new ClassMetadataInfo('test_entity');
+        $classMetadataInfo->setSequenceGeneratorDefinition($parsedDefinition);
+
+        self::assertSame(
+            ['sequenceName' => 'test_sequence', 'allocationSize' => '1', 'initialValue' => '1'],
+            $classMetadataInfo->sequenceGeneratorDefinition
+        );
+    }
+}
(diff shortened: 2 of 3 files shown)


lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -337,8 +337,8 @@ public function loadMetadataForClass($className, ClassMetadata $metadata)
                 $metadata->setSequenceGeneratorDefinition(
                     [
                         'sequenceName' => (string) $seqGenerator['sequence-name'],
-                        'allocationSize' => (string) $seqGenerator['allocation-size'],
-                        'initialValue' => (string) $seqGenerator['initial-value']
+                        'allocationSize' => (int) $seqGenerator['allocation-size'],
+                        'initialValue' => (int) $seqGenerator['initial-value']
```
Ocramius: Tbh, sequences are not necessarily int-based

alextech: They can be floats? I do not think I can do `CREATE SEQUENCE INCREMENT ` with anything but integer. Neither PostgreSQL, nor oracle https://docs.oracle.com/database/121/SQLRF/statements_6017.htm#SQLRF01314

Majkl578: Postgres only supports bigint:
> Sequences are based on bigint arithmetic, so the range cannot exceed the range of an eight-byte integer (-9223372036854775808 to 9223372036854775807).
> https://www.postgresql.org/docs/9.5/static/sql-createsequence.html

But on 32-bit platform, bigint would be string/float...

alextech: Interesting. That would be still on PostgreSQL side though. The bug at highlighted line in referenced issue is in PHP side, where SequenceGenerator trying to predict maximum value. It is expecting PHP integer, as enforced by the cast, to do the addition for the prediction. But if value comes from XML parser, prediction calculation fails having second addend being string.

alextech: Or are you saying because of this 32-bit corner case, the schema mapping should not be touched, but the cast should be fixed instead wherever the parameters are used? In this case, are you saying xml parser is not root cause, that `SequenceGenerator` should be fixed?

alextech: Also turns out migrations wants the value to be string. If driver casts to integer itself, migration diff generates INCREMENT BY 0. Unit tests here do not notice that potential regression. so I guess the crash is better fixed at the SequenceGenerator level.

alextech: sorry, discovered actual root cause. Thanks @Majkl578 for making me think further about it. Turns out xml driver is not supplying default values for optional fields when they are omitted in mapping file. This results in empty $definition keys, which causes "non numeric value" error during addition, *not* because it is a casting problem as I originally thought. 

Will redo the fix by either validating input in `ClassMetadataInfo::setSequenceGeneratorDefinition`  where it already does a little bit of parsing validation for sequence_name, but not the other parameters, or maybe I can find why XSD schema is not applied by XML driver.

alextech:
Kept default as string too as @Ocramius suggested, however I will still point out that

1. If plan to later convert to PHP 7 type syntax, `SequenceGenerator` constructor https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Id/SequenceGenerator.php#L63 wants an integer and will not be happy with string.
2. xsd schema claims it to be integer so IMO it should throw exception if non_numeric value is given by user.

lcobucci (approved):
I've rebased the branch and applied some minor changes. It LGTM now!

lcobucci:
> If plan to later convert to PHP 7 type syntax, SequenceGenerator constructor https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Id/SequenceGenerator.php#L63 wants an integer and will not be happy with string.

@alextech we only plan to add type declaration for `v3` so that's fine for now, I'd say that's important to mimic the current behaviour in order to avoid BC breaks.

alextech:
Thank you! I can now get off my custom branch.
@lcobucci sounds good. I am also seeing a very useful #6728 which is a great step, but I do not see it managing default values which is how these bugs occur. I do not know if it is validator's job to inject default values into driver, or driver should have defaults hardcoded in. ATM there is a disconnect between configuration passing schema validation, but the driver still expecting something else. So for `v3` should the suggestion be made to have validator or parser pass in missing defaults, or do an "audit" of all driver components to ensure they do not crash without missing values?

lcobucci:
> So for `v3` should the suggestion be made to have validator or parser pass in missing defaults, or do an "audit" of all driver components to ensure they do not crash without missing values?

@alextech the best person to answer that would be @guilhermeblanco 😄

PR #6684, opened 2017-09-03
Support composite & non-identifier generated values

### LINKED ISSUE #2400: DDC-1750: [GH-319] [WIP] Added support to Multiple ID Generators
Jira issue originally created by user @beberlei:

This issue is created automatically through a Github pull request on behalf of guilhermeblanco:

  Url: https://github.com/doctrine/doctrine2/pull/319

Message:

### LINKED ISSUE #4704: DDC-3853: [GH-1480] Allow custom id generators to handle composite keys
Jira issue originally created by user @doctrinebot:

This issue is created automatically through a Github pull request on behalf of kaiwa:

Url: https://github.com/doctrine/doctrine2/pull/1480

Message:

Currently when using composite keys, any other id generation strategy than
NONE (assigned identifiers) is triggering an exception in ClassMetadataInfo.

This commit allows to use the CUSTOM strategy with composite keys, therefore
to allow a custom id generator to handle the composite key identifier
generation.

### LINKED ISSUE #6422: Using two sequences in the same table
I need to handle two sequences for the same table like this:
```php
class Foo
{
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="SEQUENCE")
     * @SequenceGenerator(sequenceName="a_seq")
     */
    private $a;

    /**
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="SEQUENCE")
     * @SequenceGenerator(sequenceName="b_seq")
     */
    private $b;
}
```
But insert has failed with `not null violation` on "b".
A query looks like `INSERT INTO foo (a, b) VALUES (42, null)`.
My DDL allow using`DEFAULT nextval('b_seq'::regclass);`, but Doctrine pass `null` explicit and breaks query.
How can I avoid them?

Majkl578:
* enable generated values for composite identifiers
* enable generated values for non-identifier fields
* forbid IDENTITY in composite identifiers and on non-identifiers (not supported by platforms)

This is now possible thanks to f482d64d50ffec5f7918206f56beb849e48350e2.

Possible enhancements:
* More platform-specific tests (i.e. sequences)?
* Unsure about post-insert generators, should they be allowed? Or maybe removed in 3.0 as it really makes sense only for IDENTITY generator?

Closes #6422.
Closes #4704.
Closes #5416.
Closes #2400.


lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -486,7 +486,7 @@ private function gatherColumn($classMetadata, FieldMetadata $fieldMetadata, Tabl
             $options['customSchemaOptions'] = $fieldOptions;
         }
 
-        if ($fieldMetadata->hasValueGenerator() && $fieldMetadata->getValueGenerator()->getType() === GeneratorType::IDENTITY && $classMetadata->getIdentifierFieldNames() == [$fieldName]) {
+        if ($fieldMetadata->hasValueGenerator() && $fieldMetadata->getValueGenerator()->getType() === GeneratorType::IDENTITY && in_array($fieldName, $classMetadata->getIdentifierFieldNames(), true)) {
```
greg0ire: This line could use one or more linebreaks :P

PR #6690, opened 2017-09-05
Clarify association mapping language

bitwombat:
- 5.3 was wrong - should really be 'inversedBy'
- 11.1 - change 'use' to 'have' for consistency.  A class has an attribute.

- See https://github.com/doctrine/doctrine2/issues/6686


Ocramius (approved):

Ocramius:
Thanks @bitwombat!

PR #6695, opened 2017-09-07
Update table prefix docs

wshafer:
Updated docs to remove deprecated setTableName() and replace it with setPrimaryTable()


Ocramius:
What's the reason for this change? Broken API? What is not working as expected?

wshafer:
@Ocramius - setTableName() is marked as Deprecated and the deprecated message says to use setPrimaryTable() instead.

wshafer:
See: https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L2264

Ocramius (approved):
@wshafer thanks for clarifying :+1:

Issue #6531, opened 2017-06-27
Cannot persist parent entity if OneToOne shares the same id


davidbarratt:
If you have an Entity like this:
```php
/**
 * @ORM\Entity
 * @ORM\Table(name="takedown")
 */
class Takedown {

	/**
	 * @var int
	 *
	 * @ORM\Column(name="takedown_id", type="integer")
	 * @ORM\Id
	 * @ORM\GeneratedValue
	 */
	private $id;

	/**
	 * @var DigitalMillenniumCopyrightAct
	 *
	 * @ORM\OneToOne(targetEntity="DigitalMillenniumCopyrightAct", mappedBy="takedown", orphanRemoval=true, cascade={"persist"})
	 */
	private $dmca;

	public function setDmca( $dmca = null ) : self {
		if ( $dmca === null ) {
			$this->dmca = null;

			return $this;
		}

		$this->dmca = $dmca;
		$this->dmca->setTakedown( $this );

		return $this;
	}

	public function getDmca() :? DigitalMillenniumCopyrightAct {
		return $this->dmca;
	}
}
```

and the related entity:
```php
/**
 * @ORM\Entity
 * @ORM\Table(name="takedown_dmca")
 */
class DigitalMillenniumCopyrightAct {

	/**
	 * @ORM\Id
	 * @ORM\OneToOne(
	 *	targetEntity="Takedown",
	 *	inversedBy="dmca"
	 *)
         * @ORM\JoinColumn(name="takedown_id", referencedColumnName="takedown_id")
	 */
	private $takedown;

	public function setTakedown( Takedown $takedown ) : self {
		$this->takedown = $takedown;

		return $this;
	}

	public function getTakedown( ) :? Takedown {
		return $this->takedown;
	}
}
```

Doing something like this:
```
$takedown = new Takedown();
$takedown->setDmca(new DigitalMillenniumCopyrightAct());

$em->persist($takedown);
$em->flush();
```
will always fail with an error:

> The given entity of type 'DigitalMillenniumCopyrightAct' (DigitalMillenniumCopyrightAct@000000000c1a46f70000000005683920) has no identity/no id values set. It cannot be added to the identity map.

I've also tried explicit persisting:
```
$takedown = new Takedown();
$takedown->setDmca(new DigitalMillenniumCopyrightAct());

$em->persist($takedown);
$em->persist($takedown->getDmca());
$em->flush();
```

But also fails with the same error.

[...]

Issue #6043, opened 2016-09-19
Identity through foreign Entities - persist


Gorjunov:
I try to implement schema from http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/composite-primary-keys.html#use-case-2-simple-derived-identity (OneToOne relation between models and second model's primary key is also foreign key to first model)

When I try to create both entities at same time and persist them (`$entityManager->persist($address)`) I get `Doctrine\ORM\ORMInvalidArgumentException: The given entity of type ‘Address' (Address@0000000055b6ca6a000000001effbba7) has no identity/no id values set. It cannot be added to the identity map`

Is there any solution (except saving User, adding it to address and then saving address) to realize it.

P.S.
I've tried to change `Doctrine\ORM\Id\AssignedGenerator` by adding:

```
    public function isPostInsertGenerator()
    {
        return true;
    }
```

This change has fixed my problem, but I'm afraid, that AssignedGenerator change is bad idea.

PR #1176, opened 2014-11-07
[DDC-3378] Support merging entities with composite identities defined through to-one associations

adrienbrault:
I hope that I'm doing something wrong and that this is not a bug ...

PR #6701, opened 2017-09-13
Fix identity through foreign entities

vhenzl:
Failing tests for #6531 (and also #6043).

Tests are based on examples from [Composite and Foreign Keys as Primary Key](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/composite-primary-keys.html) tutorial. None of all three examples of **Identity through foreign Entities** works at the moment. All of them fail with a similar error:

```
[Doctrine\ORM\ORMInvalidArgumentException] The given entity of type 'GH6531Address' (GH6531Address@0000000013e28b9b000000003fade3e6) has no identity/no id values set. It cannot be added to the identity map.
```

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
@@ -1,0 +1,190 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+
+final class GH6531Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    protected function setUp() : void
+    {
+        parent::setup();
+
+        $this->setUpEntitySchema(
+            [
+                GH6531User::class,
+                GH6531Address::class,
+                GH6531Article::class,
+                GH6531ArticleAttribute::class,
+                GH6531Order::class,
+                GH6531OrderItem::class,
+                GH6531Product::class,
+            ]
+        );
+    }
+
+    /**
+     * @group 6531
+     */
+    public function testSimpleDerivedIdentity() : void
+    {
+        $user          = new GH6531User();
+        $address       = new GH6531Address();
+        $address->user = $user;
+
+        $this->_em->persist($user);
+        $this->_em->persist($address);
+        $this->_em->flush();
+
+        self::assertSame($user, $this->_em->find(GH6531User::class, $user->id));
+        self::assertSame($address, $this->_em->find(GH6531Address::class, $user));
+    }
+
+    /**
+     * @group 6531
+     */
+    public function testDynamicAttributes() : void
+    {
+        $article = new GH6531Article();
+        $article->addAttribute('name', 'value');
+
+        $this->_em->persist($article);
+        $this->_em->flush();
+
+        self::assertSame(
+            $article->attributes['name'],
+            $this->_em->find(GH6531ArticleAttribute::class, ['article' => $article, 'attribute' => 'name'])
+        );
+    }
+
+    /**
+     * @group 6531
+     */
+    public function testJoinTableWithMetadata() : void
+    {
+        $product = new GH6531Product();
+        $this->_em->persist($product);
+        $this->_em->flush();
+
+        $order = new GH6531Order();
+        $order->addItem($product, 2);
+
+        $this->_em->persist($order);
+        $this->_em->flush();
+
+        self::assertSame(
+            $order->items->first(),
+            $this->_em->find(GH6531OrderItem::class, ['product' => $product, 'order' => $order])
+        );
+    }
+}
+
+/**
+ * @Entity
+ */
+class GH6531User
+{
+    /** @Id @Column(type="integer") @GeneratedValue */
+    public $id;
+}
+
+/**
+ * @Entity
+ */
+class GH6531Address
+{
+    /** @Id @OneToOne(targetEntity=GH6531User::class) */
+    public $user;
+}
+
+/**
+ * @Entity
+ */
+class GH6531Article
+{
+    /** @Id @Column(type="integer") @GeneratedValue */
+    public $id;
+
+    /** @OneToMany(targetEntity=GH6531ArticleAttribute::class, mappedBy="article", cascade={"ALL"}, indexBy="attribute") */
+    public $attributes;
+
+    public function addAttribute(string $name, string $value)
+    {
+        $this->attributes[$name] = new GH6531ArticleAttribute($name, $value, $this);
+    }
+}
+
+/**
+ * @Entity
+ */
+class GH6531ArticleAttribute
+{
+    /** @Id @ManyToOne(targetEntity=GH6531Article::class, inversedBy="attributes") */
+    public $article;
+
+    /** @Id @Column(type="string") */
+    public $attribute;
+
+    /** @Column(type="string") */
+    public $value;
+
+    public function __construct(string $name, string $value, GH6531Article $article)
+    {
+        $this->attribute = $name;
+        $this->value     = $value;
+        $this->article   = $article;
+    }
+}
+
+/**
+ * @Entity
+ */
+class GH6531Order
+{
+    /** @Id @Column(type="integer") @GeneratedValue */
+    public $id;
+
+    /** @OneToMany(targetEntity=GH6531OrderItem::class, mappedBy="order", cascade={"ALL"}) */
+    public $items;
+
+    public function __construct()
+    {
+        $this->items = new ArrayCollection();
+    }
+
+    public function addItem(GH6531Product $product, int $amount) : void
+    {
+        $this->items->add(new GH6531OrderItem($this, $product, $amount));
+    }
+}
+
+/**
+ * @Entity
+ */
+class GH6531Product
+{
+    /** @Id @Column(type="integer") @GeneratedValue */
+    public $id;
+}
+
+/**
+ * @Entity
+ */
+class GH6531OrderItem
+{
+    /** @Id @ManyToOne(targetEntity=GH6531Order::class) */
+    public $order;
+
+    /** @Id @ManyToOne(targetEntity=GH6531Product::class) */
+    public $product;
+
+    /** @Column(type="integer") */
+    public $amount = 1;
+
+    public function __construct(GH6531Order $order, GH6531Product $product, int $amount = 1)
+    {
+        $this->order   = $order;
+        $this->product = $product;
+        $this->amount  = $amount;
+    }
+}
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -940,12 +940,30 @@ class UnitOfWork implements PropertyChangedListener
                 $class->setIdentifierValues($entity, $idValue);
             }
 
-            $this->entityIdentifiers[$oid] = $idValue;
+            // Some identifiers may be foreign keys to new entities.
+            // In this case, we don't have the value yet and should treat it as if we have a post-insert generator
+            if (! $this->hasMissingIdsWhichAreForeignKeys($class, $idValue)) {
+                $this->entityIdentifiers[$oid] = $idValue;
+            }
         }
 
         $this->entityStates[$oid] = self::STATE_MANAGED;
 
         $this->scheduleForInsert($entity);
+    }
+
+    /**
+     * @param mixed[] $idValue
+     */
+    private function hasMissingIdsWhichAreForeignKeys(ClassMetadata $class, array $idValue) : bool
+    {
+        foreach ($idValue as $idField => $idFieldValue) {
+            if ($idFieldValue === null && isset($class->associationMappings[$idField])) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     /**
@@ -1033,11 +1051,15 @@ class UnitOfWork implements PropertyChangedListener
         $persister  = $this->getEntityPersister($className);
         $invoke     = $this->listenersInvoker->getSubscribedSystems($class, Events::postPersist);
 
+        $insertionsForClass = [];
+
         foreach ($this->entityInsertions as $oid => $entity) {
 
             if ($this->em->getClassMetadata(get_class($entity))->name !== $className) {
                 continue;
             }
+
+            $insertionsForClass[$oid] = $entity;
 
             $persister->addInsert($entity);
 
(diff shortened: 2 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
```diff
@@ -0,0 +1,242 @@
...
+ */
+class GH6531Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setup();
+
+        try {
+            $this->_schemaTool->createSchema(array(
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531User'),
```
Ocramius: You should probably use the `::class` syntax here

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
```diff
@@ -0,0 +1,242 @@
...
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531Address'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531Article'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531ArticleAttribute'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531Customer'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531Order'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531OrderItem'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531Product'),
+            ));
+        } catch (\Exception $e) {
```
Ocramius: Catch only tools exceptions, please

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
```diff
@@ -0,0 +1,242 @@
...
+    private $id;
+}
+
+/**
+ * @Entity
+ */
+class GH6531Address
+{
+    /** @Id @OneToOne(targetEntity="GH6531User") */
+    private $user;
```
Ocramius: Use a `public` property instead

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
```diff
@@ -0,0 +1,242 @@
...
+/**
+ * @Entity
+ */
+class GH6531Article
+{
+    /** @Id @Column(type="integer") @GeneratedValue */
+    private $id;
+
+    /** @Column(type="string") */
+    private $title;
```
Ocramius: Remove any fields and methods that aren't strictly required for the test scenario

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
```diff
@@ -0,0 +1,242 @@
...
+
+    /** @OneToMany(targetEntity="GH6531ArticleAttribute", mappedBy="article", cascade={"ALL"}, indexBy="attribute") */
+    private $attributes;
+
+    public function __construct($title)
```
Ocramius: If the title is not required, drop this

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
```diff
@@ -0,0 +1,242 @@
...
+{
+    /** @Id @ManyToOne(targetEntity="GH6531Article", inversedBy="attributes") */
+    private $article;
+
+    /** @Id @Column(type="string") */
+    private $attribute;
+
+    /** @Column(type="string") */
+    private $value;
+
```
Ocramius: Same as above: drop any non-strictly-required fields

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
```diff
@@ -0,0 +1,242 @@
...
+    private $id;
+
+    /** @ManyToOne(targetEntity="GH6531Customer") */
+    private $customer;
+
+    /** @OneToMany(targetEntity="GH6531OrderItem", mappedBy="order", cascade={"ALL"}) */
+    private $items;
+
+    /** @Column(type="boolean") */
+    private $payed = false;
```
Ocramius: Same: drop anything that isn't strictly needed

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
```diff
@@ -0,0 +1,242 @@
...
+    /** @Id @Column(type="integer") @GeneratedValue */
+    private $id;
+}
+
+/**
+ * @Entity
+ */
+class GH6531Address
+{
+    /** @Id @OneToOne(targetEntity="GH6531User") */
```
Ocramius: Use `::class` also when referencing classes in annotations

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
```diff
@@ -0,0 +1,242 @@
...
+        $this->_em->persist($user);
+        $this->_em->persist($address);
+        $this->_em->flush();
+    }
+
+
+    /**
+     * @group GH-6531
+     */
+    public function testDynamicAttributes(): void
```
Ocramius: Drop passing tests fron the testcase

Ocramius: Ah, never mind, not this one 🤔

Ocramius (changes requested):

Ocramius:
@vhenzl does this set of tests fail also on 2.5.x?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
vhenzl: I've copy-pasted the try/catch from some older test. Is it actually good for something? It's never used in more recent tests as I can see now. Shouldn't I rather throw it away?

Ocramius:
Most likely, but if you throw it away you also need a tearDown to handle
dropping tables at the end of a test cycle

On 13 Sep 2017 08:26, "Vašek Henzl" <notifications@github.com> wrote:

*@vhenzl* commented on this pull request.
------------------------------

In tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php
<https://github.com/doctrine/doctrine2/pull/6701#discussion_r138533239>:

> +    protected function setUp(): void
+    {
+        parent::setup();
+
+        try {
+            $this->_schemaTool->createSchema(array(
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531User'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531Address'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531Article'),
+                $this->_em->getClassMetadata(__NAMESPACE__ .
'\GH6531ArticleAttribute'),
+                $this->_em->getClassMetadata(__NAMESPACE__ .
'\GH6531Customer'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531Order'),
+                $this->_em->getClassMetadata(__NAMESPACE__ .
'\GH6531OrderItem'),
+                $this->_em->getClassMetadata(__NAMESPACE__ . '\GH6531Product'),
+            ));
+        } catch (\Exception $e) {

I've copy-pasted the try/catch from some older test. Is it actually good
for something? It's never used in more recent tests as I can see now.
Shouldn't I rather throw it away?

—
You are receiving this because you commented.

<https://github.com/doctrine/doctrine2/pull/6701#discussion_r138533239>,
or mute
the thread

vhenzl:
> does this set of tests fail also on 2.5.x?

@Ocramius  On 2.5 branch fails only first test (`testSimpleDerivedIdentity`) where entities have `OneToOne` relation,  other two tests with `ManyToOne` pass.

vhenzl:
@Ocramius, I did some digging in the code and commit history regarding this issue and I have a question: Was this use case actually ever supported? Ie. identity through foreign entities with id strategy set to "AUTO" (ie. to "IDENTITY" or "SEQUENCE")?

Current error

> Doctrine\ORM\ORMInvalidArgumentException: The given entity of type 'GH6531Address' (GH6531Address@0000000013e28b9b000000003fade3e6) has no identity/no id values set. It cannot be added to the identity map.

appears since b889e18a. Before that commit, error message would be

> Doctrine\ORM\ORMException: Entity of type GH6531Address has identity through a foreign entity GH6531User, however this entity has no identity itself. You have to call EntityManager#persist() on the related entity and make sure that an identifier was generated before trying to persist 'GH6531Address'. In case of Post Insert ID Generation (such as MySQL Auto-Increment) this means you have to call EntityManager#flush() between both persist operations.

There was a discussion ([see here](https://github.com/doctrine/doctrine2/pull/1176/files#r20011861)) in PR #1176 if the check throwing this exception should be removed or not. It was removed eventually, but apparently shouldn't be.
Also note, that `ORMException::entityMissingForeignAssignedId($entity, $relatedEntity)` throwing this exception isn't call anywhere since then.

Anyway, according to this error message, I assume that identity through foreign entities with database generated id never was supported and this PR doesn't make sense then.

If this is true, then the documentation is wrong ([from the beginning](https://github.com/doctrine/doctrine2/blob/91b2c82c58bae2051df4e8c8b227c033df525143/en/tutorials/composite-primary-keys.rst)) and needs to be updated to use `@GeneratedValue(strategy="NONE")` instead of `@GeneratedValue`.

With `@GeneratedValue(strategy="NONE")` and manually assigned IDs everything works as expected.

-----
Side note on different behaviour for 2.5 and 2.6 when in master branch fail all three tests but on 2.5 only first one. 
This happens because of behaviour before #5924 (commit 1cb8d790) when entity identifier could contain `null`.  And because both `GH6531ArticleAttribute` and `GH6531OrderItem` have a composite key, `$idHash` wasn't an empty string and therefore the exception wasn't thrown and it worked somehow.

lcobucci:
@vhenzl @Ocramius I've rebased the branch to sync with `master` and updated the tests to add the proper assertions but honestly I don't see an easy solution for this, since it's related to how we read/write to/from the identity map. Maybe @guilhermeblanco could give us some ideas on how to fix this, but regardless of that I think that this should be fixed for `v2.6.1` or even `v3.0`.

guilhermeblanco:
@vhenzl you are indeed correct.
*ToOne Identifiers pointing to a postInsert generated values were never supported since 2.0.0 Beta 2, when we supported association identifiers.

We're now discussing internally the procedure to take (if we'll solve this in 2.X or 3.0 only). In the meantime, switch the User to assigned and it should work.

PS: And the documentation should also be addressed... =(

soyuka:
Just adding some use case for this particular issue. Assuming you have an association (can be seen as a Many To Many relation) that need additional information, you're defining a new Entity.

For example, here is a relational entity that has a "name":

```
/**
 * @ORM\Entity
 */
class RelationThatHasAName
{
    /**
     * @ORM\Column
     */
    public $name;

    /**
     * @ORM\Id
     * @ORM\ManyToOne(targetEntity="DummyFriend")
     * @ORM\JoinColumn(name="dummyfriend_id", referencedColumnName="id", nullable=false)
     */
    public $dummyFriend;

    /**
     * @ORM\Id
     * @ORM\ManyToOne(targetEntity="RelatedDummy", inversedBy="relatedToDummyFriend")
     * @ORM\JoinColumn(name="relateddummy_id", referencedColumnName="id", nullable=false, onDelete="CASCADE")
     */
    public $relatedDummy;
}
```

This is a common use case to me. Using `@ORM\Id` just helps identifying that the columns are our identifiers. Though, we don't need to add a generated value or any other identifier, the association ones are sufficient.

This used to work pre `2.6` and IMO should still work. My error is now:

```
identity/no id values set. It cannot be added to the identity map
```

lcobucci:
> Just adding some use case for this particular issue. Assuming you have an association (can be seen as a Many To Many relation) that need additional information, you're defining a new Entity.

Why not using what the [lib offers for `many-to-many`](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/association-mapping.html#many-to-many-unidirectional)? I mean, creating an entity in the middle is not same as a `many-to-many` association.

soyuka:
Because you can (could?) not add custom columns to the ManyToMany entity (the association table). I think that I discussed that with @Ocramius somewhere but can't find the issue anymore. Anyway the conclusion was that if you need to add fields to a ManyToMany relation, you need to create a new entity.

Obviously, in the case above you could easily add a column with an `@Id`. Though IMHO it's cleaner to have `RelationA-RelationB` as index (same constraint as you would have in a ManyToMany relation).

lcobucci:
> Because you can (could?) not add custom columns to the ManyToMany entity (the association table). I think that I discussed that with @Ocramius somewhere but can't find the issue anymore. Anyway the conclusion was that if you need to add fields to a ManyToMany relation, you need to create a new entity.
> 
> Obviously, in the case above you could easily add a column with an @Id. Though IMHO it's cleaner to have RelationA-RelationB as index (same constraint as you would have in a ManyToMany relation).

@soyuka I do agree that **if (and when) needed you can have an entity to connect stuff** but let's be clear that these are not `many-to-many` associations. Regardless of that, you still can have this kind of mapping, the point is that it cannot be done using a post generated ID, only that 😄

soyuka:
> let's be clear that these are not many-to-many

indeed those are not many-to-many, as said above *can be seen as*

It does work if my the two associations have been `flushed` before we save that entity yes :).

lcobucci:
> It does work if my the two associations have been flushed before we save that entity yes :).

Yeap, because then you do have identifiers for your relations 😉

guilhermeblanco:
@soyuka the easiest solution is to add a reference identifier to this table and remove the @Id from associations.

soyuka:
What do you mean by "reference identifier"? I like having associations as identifier. Anyway, this works just fine as long as you save the relation before saving the relation entity.

lcobucci:
@Ocramius I've moved the fix from #7003 (with some minor fixes) to this PR and things are working fine. We did have some impact on insertion performance (`SimpleInsertPerformanceBench`):

```diff
- benchHydration                I0 P0 	[μ Mo]/r: 480,363.000 480,363.000 (μs) 	[μSD μRSD]/r: 0.000μs 0.00%
+ benchHydration                I0 P0 	[μ Mo]/r: 482,352.000 482,352.000 (μs) 	[μSD μRSD]/r: 0.000μs 0.00%
```

Could you please check if this is acceptable or if we can optimise things a bit more?

Ocramius:
@lcobucci that impact is OK :+1:

lcobucci (approved):

lcobucci:
@Ocramius alright then, I'll merge this in and make it part of `v2.6.1`.

lcobucci:
@Ocramius I'm still working on porting this to `master` (which is being a little tricky), will continue on this tomorrow 👍

tonivdv:
@lcobucci @Ocramius I'm currently having this issue in doctrine 2.5 ... But even after manually applying this PR's patch I still have the issue. So I wanted to know 1) if this was, according to you, a 2.6 issue only, or 2) it's an issue in 2.5 too and fixing it requires more than the current patch ?

lcobucci:
@tonivdv since the release of v2.6.0 we no longer maintain v2.5.x and recommend every one to migrate to the newer version. We just backport security fixes to v2.5.x, which didn't happen so far 👍

tonivdv:
@lcobucci Yeah I know and trust me I'm crying (:p) here but one of the projects can't be migrated to 2.6 because it can't be migrated easily at this stage to php 7 . Anyway, I changed the entity to prevent this issue, so as soon as we migrate the php 7 we'll update doctrine accordingly.

But according to you the issue is still present in 2.5 right?

Thanks for the quick response.

Cheers

lcobucci:
> But according to you the issue is still present in 2.5 right?

It's highly possible that it still there.

PR #6705, opened 2017-09-16
Fix parameter name comparison in AbstractQuery regarding different types (fixes #6699)

### LINKED ISSUE #6699: setParameter issue when mixing :name and ?0 syntax
Example:
```php
$queryBuilder

    ->andWhere("alias.field = :paramname")
    ->setParameter('paramname', $paramval)

    ->andWhere("alias.field = ?0")
    ->setParameter(0, $paramval)

;
```

If I call `getParameters()`, only 1 param is returned.

When executing the query, an exception is thrown:

> Too few parameters: the query defines 2 parameters but you only bound 1

This seems to happen only if the param is named `?0`.
With `?1`, it seems to be ok.

Majkl578:
AbstractQuery::setParameter() uses weak comparison to compare parameter names and decide whether to update existing or add new parameter. AbstractQuery::getParameter() does so as well.
This is problematic for parameter named `0` since strings are downcasted to `0` as well, making it incorrectly assume that i.e. the `0` and `"foo"` are the same parameter.
In most cases, cast is not needed and we would ideally want to prevent it, thus added identity check first and only if it fails, string cast comparison is attempted.

Fixes #6699.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -321,14 +321,14 @@ abstract class AbstractQuery
     public function getParameter($key)
     {
         $filteredParameters = $this->parameters->filter(
-            function ($parameter) use ($key)
-            {
-                // Must not be identical because of string to integer conversion
-                return ($key == $parameter->getName());
+            function (Query\Parameter $parameter) use ($key) : bool {
+                $parameterName = $parameter->getName();
+
+                return $key === $parameterName || (string) $key === (string) $parameterName;
             }
         );
 
-        return count($filteredParameters) ? $filteredParameters->first() : null;
+        return ! $filteredParameters->isEmpty() ? $filteredParameters->first() : null;
     }
 
     /**
@@ -369,17 +369,10 @@ abstract class AbstractQuery
      */
     public function setParameter($key, $value, $type = null)
     {
-        $filteredParameters = $this->parameters->filter(
-            function ($parameter) use ($key)
-            {
-                // Must not be identical because of string to integer conversion
-                return ($key == $parameter->getName());
-            }
-        );
+        $existingParameter = $this->getParameter($key);
 
-        if (count($filteredParameters)) {
-            $parameter = $filteredParameters->first();
-            $parameter->setValue($value, $type);
+        if ($existingParameter !== null) {
+            $existingParameter->setValue($value, $type);
 
             return $this;
         }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6699Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6699Test.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\Models\CMS\CmsUser;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group 6699
+ */
+final class GH6699Test extends OrmFunctionalTestCase
+{
+    protected function setUp() : void
+    {
+        $this->useModelSet('cms');
+
+        parent::setUp();
+    }
+
+    public function testMixedParametersWithZeroNumber() : void
+    {
+        $query = $this->_em->createQueryBuilder()
+                           ->select('u')
+                           ->from(CmsUser::class, 'u')
+                           ->andWhere('u.username = :username')
+                           ->andWhere('u.id = ?0')
+                           ->getQuery();
+
+        $query->setParameter('username', 'bar');
+        $query->setParameter(0, 0);
+
+        $query->execute();
+
+        self::assertCount(2, $query->getParameters());
+        self::assertSame(0, $query->getParameter(0)->getValue());
+        self::assertSame('bar', $query->getParameter('username')->getValue());
+    }
+
+    public function testMixedParametersWithZeroNumberOnQueryBuilder() : void
+    {
+        $query = $this->_em->createQueryBuilder()
+                           ->select('u')
+                           ->from(CmsUser::class, 'u')
+                           ->andWhere('u.username = :username')
+                           ->andWhere('u.id = ?0')
+                           ->setParameter('username', 'bar')
+                           ->setParameter(0, 0)
+                           ->getQuery();
+
+        $query->execute();
+
+        self::assertCount(2, $query->getParameters());
+        self::assertSame(0, $query->getParameter(0)->getValue());
+        self::assertSame('bar', $query->getParameter('username')->getValue());
+    }
+}
--- a/tests/Doctrine/Tests/ORM/Query/QueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/QueryTest.php
@@ -12,6 +12,7 @@ use Doctrine\ORM\Query\QueryException;
 use Doctrine\Tests\Mocks\DriverConnectionMock;
 use Doctrine\Tests\Mocks\StatementArrayMock;
 use Doctrine\Tests\Models\CMS\CmsAddress;
+use Doctrine\Tests\Models\CMS\CmsUser;
 use Doctrine\Tests\OrmTestCase;
 
 class QueryTest extends OrmTestCase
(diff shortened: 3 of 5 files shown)


Majkl578:
Reworded & fix + more tests added.

Seb33300:
You have only updated the `AbstractQuery::setParameter()` function but I can find the exact same function in `QueryBuilder`: https://github.com/doctrine/doctrine2/blob/80f7824b3d8a2b5e3810c597582ee3524750b682/lib/Doctrine/ORM/QueryBuilder.php#L533 
Not sure if it should be modified too.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6699Test.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\Models\CMS\CmsUser;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group #6699
```
lcobucci: We're using the ticket id without the `#` everywhere else, we should keep it consistent here too

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -303,4 +304,65 @@ public function testSelectFromSubquery()
         $this->expectExceptionMessage('Subquery');
         $query->getSQL();
     }
+
+    /**
+     * @group #6699
```
lcobucci: We're using the ticket id without the `#` everywhere else, we should keep it consistent here too

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -303,4 +304,65 @@ public function testSelectFromSubquery()
...
+        $query = $this->_em->createQuery('select u from ' . CmsUser::class . ' u where u.id = ?0');
+
+        $query->setParameter(0, 0);
+
+        $this->assertSame(0, $query->getParameter(0)->getValue());
+        $this->assertSame(0, $query->getParameter('0')->getValue());
+    }
+
+    /**
+     * @group #6699
```
lcobucci: We're using the ticket id without the `#` everywhere else, we should keep it consistent here too

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -303,4 +304,65 @@ public function testSelectFromSubquery()
...
+        $query->setParameter(0, 0);
+        $query->setParameter('name', 'Doctrine');
+
+        $this->assertSame(2, $query->getParameters()->count());
+        $this->assertSame(0, $query->getParameter('0')->getValue());
+        $this->assertSame('Doctrine', $query->getParameter('name')->getValue());
+    }
+
+    /**
+     * @group #6699
```
lcobucci: We're using the ticket id without the `#` everywhere else, we should keep it consistent here too

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -303,4 +304,65 @@ public function testSelectFromSubquery()
...
+        $query->setParameter('name', 'Doctrine');
+        $query->setParameter(0, 0);
+
+        $this->assertSame(2, $query->getParameters()->count());
+        $this->assertSame(0, $query->getParameter(0)->getValue());
+        $this->assertSame('Doctrine', $query->getParameter('name')->getValue());
+    }
+
+    /**
+     * @group #6699
```
lcobucci: We're using the ticket id without the `#` everywhere else, we should keep it consistent here too

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -303,4 +304,65 @@ public function testSelectFromSubquery()
...
+    public function testSetParameterWithTypeJugglingWorks()
+    {
+        $query = $this->_em->createQuery("select u from Doctrine\Tests\Models\CMS\CmsUser u where u.id != ?0 and u.username = :name");
+
+        $query->setParameter('0', 1);
+        $query->setParameter('name', 'Doctrine');
+        $query->setParameter(0, 2);
+        $query->setParameter('0', 3);
+
+        $this->assertSame(2, $query->getParameters()->count());
```
lcobucci: `self::assertCount()` would be more appropriated here

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6699Test.php
```diff
@@ -0,0 +1,38 @@
...
+            ->andWhere('u.username = :username')
+            ->andWhere('u.id = ?0');
+        $query = $qb->getQuery();
+
+        $query->setParameter('username', 'bar');
+        $query->setParameter(0, 0);
+
+        $query->execute();
+
+        $this->assertSame(2, $query->getParameters()->count());
```
lcobucci: `self::assertCount()` would be more appropriated here

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -303,4 +304,65 @@ public function testSelectFromSubquery()
...
+     * @group #6699
+     */
+    public function testSetParameterWithNameZeroIsNotOverridden() : void
+    {
+        $query = $this->_em->createQuery('select u from ' . CmsUser::class . ' u where u.id != ?0 and u.username = :name');
+
+        $query->setParameter(0, 0);
+        $query->setParameter('name', 'Doctrine');
+
+        $this->assertSame(2, $query->getParameters()->count());
```
lcobucci: `self::assertCount()` would be more appropriated here

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -303,4 +304,65 @@ public function testSelectFromSubquery()
...
+     * @group #6699
+     */
+    public function testSetParameterWithNameZeroDoesNotOverrideAnotherParameter() : void
+    {
+        $query = $this->_em->createQuery('select u from ' . CmsUser::class . ' u where u.id != ?0 and u.username = :name');
+
+        $query->setParameter('name', 'Doctrine');
+        $query->setParameter(0, 0);
+
+        $this->assertSame(2, $query->getParameters()->count());
```
lcobucci: `self::assertCount()` would be more appropriated here

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6699Test.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    protected function setUp()
+    {
+        $this->useModelSet('cms');
+        parent::setUp();
+    }
+
+    public function testMixedParametersWithZeroNumber()
+    {
+        $qb = $this->_em->createQueryBuilder()
```
lcobucci: We can simplify this bit, since the query builder reference is not really needed, like:

```php
$query = $this->_em->createQueryBuilder()
                   ->select('u')
                   ->from(CmsUser::class, 'u')
                   ->andWhere('u.username = :username')
                   ->andWhere('u.id = ?0')
                   ->getQuery();
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6699Test.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+class GH6699Test extends OrmFunctionalTestCase
+{
+    protected function setUp()
+    {
+        $this->useModelSet('cms');
+        parent::setUp();
+    }
+
+    public function testMixedParametersWithZeroNumber()
```
lcobucci: ` : void`

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6699Test.php
```diff
@@ -0,0 +1,38 @@
...
+
+use Doctrine\Tests\Models\CMS\CmsUser;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group #6699
+ */
+class GH6699Test extends OrmFunctionalTestCase
+{
+    protected function setUp()
```
lcobucci: ` : void`

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -303,4 +304,65 @@ public function testSelectFromSubquery()
...
+        $this->assertSame(0, $query->getParameter(0)->getValue());
+        $this->assertSame('Doctrine', $query->getParameter('name')->getValue());
+    }
+
+    /**
+     * @group #6699
+     */
+    public function testSetParameterWithTypeJugglingWorks()
```
lcobucci: ` : void`

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -303,4 +304,65 @@ public function testSelectFromSubquery()
...
+    {
+        $query = $this->_em->createQuery("select u from Doctrine\Tests\Models\CMS\CmsUser u where u.id != ?0 and u.username = :name");
```
lcobucci: Could use `' . CmsUser::class . '` to keep consistency

lcobucci (changes requested):
Some nitpicking, as usual 😂 
We should also use `self::assert*()` everywhere.

I got some time now so I can apply the changes and investigate any performance impact, if you don't mind.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -369,17 +369,10 @@ public function setParameters($parameters)
      */
     public function setParameter($key, $value, $type = null)
     {
-        $filteredParameters = $this->parameters->filter(
```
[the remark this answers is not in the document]
lcobucci: Just tested, the answer is: we have no performance impact

lcobucci (approved):

lcobucci:
@Majkl578 🚢 and backported to `v2.5.x`

Thanks!

Majkl578:
We should probably copy this behavior to AbstractQuery::setParameter() as well (thanks for spotting @Seb33300).
For 3.0 we should consider different approach - I was already thinking about having ParameterCollection or ParameterBag (Symfony-style naming).

lcobucci:
> We should probably copy this behavior to AbstractQuery::setParameter() as well (thanks for spotting @Seb33300).

I've applied it there as well

Majkl578:
@lcobucci Cool, thanks!

PR #6706, opened 2017-09-18
Add "fixed" & "default" options to entity generator

MrSpoocy:
Add missing "fixed" option

### DIFF
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -1669,6 +1669,10 @@ public function __construct(<params>)
                 $options[] = '"unsigned"=true';
             }
 
+            if (isset($fieldMapping['options']['fixed']) && $fieldMapping['options']['fixed']) {
+				$options[] = '"fixed"=true';
+			}
+
             if ($options) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -1669,6 +1669,14 @@ protected function generateFieldMappingPropertyDocBlock(array $fieldMapping, Cla
                 $options[] = '"unsigned"=true';
             }
 
+            if (isset($fieldMapping['options']['fixed']) && $fieldMapping['options']['fixed']) {
+                $options[] = '"fixed"=true';
+            }
+
+            if (isset($fieldMapping['options']['default'])) {
+                $options[] = '"default"="'. $fieldMapping['options']['default'].'"';
```
lcobucci: Missing spaces before and after the concatenation operator (it should be `'"default"="' . $fieldMapping['options']['default'] . '"';`). Alternatively you can also use `sprintf()` if you prefer...

lcobucci (changes requested):
Tests are mandatory here too 😉

Majkl578:
Hi, 2.x is in a feature freeze now and we are removing EntityGenerator in 3.0 entirely, so this PR is no longer needed. Thank you.

PR #6709, opened 2017-09-19
NewObjectExpression: Allows use of nested new operators

fesor:
Covers #4687, #6573

This PR adds ability to use aggregation of DTOs as query result:

```php
$dql = '
    SELECT
        new UserDTO(
           u.id,
           u.name,
           new AddressDTO(
                address.country,
                address.city,
                address.street,
                address.postcode
           )
        )
    FROM User u
    JOIN u.location address
';
```

Use cases:

 - reporting and aggregations results
 - read-only queries

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -293,6 +293,29 @@ abstract class AbstractHydrator
             }
         }
 
+        foreach ($this->rsm->nestedNewObjectArguments as $objIndex => ['ownerIndex' => $ownerIndex, 'argIndex' => $argIndex]) {
+            $newObject = $rowData['newObjects'][$objIndex];
+            unset($rowData['newObjects'][$objIndex]);
+
+            $class  = $newObject['class'];
+            $args   = $newObject['args'];
+            $obj    = $class->newInstanceArgs($args);
+
+            $rowData['newObjects'][$ownerIndex]['args'][$argIndex] = $obj;
+        }
+
+
+        if (isset($rowData['newObjects'])) {
+            foreach ($rowData['newObjects'] as $objIndex => $newObject) {
+                $class  = $newObject['class'];
+                $args   = $newObject['args'];
+                $obj    = $class->newInstanceArgs($args);
+
+                $rowData['newObjects'][$objIndex]['obj'] = $obj;
+            }
+        }
+
+
         return $rowData;
     }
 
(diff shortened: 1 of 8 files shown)


lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -324,6 +325,30 @@ protected function gatherRowData(array $data, array &$id, array &$nonemptyCompon
...
+
+            $rowData['newObjects'][$ownerIndex]['args'][$argIndex] = $obj;
+        }
+
+
+        if (isset($rowData['newObjects'])) {
+            foreach ($rowData['newObjects'] as $objIndex => $newObject) {
+                $class  = $newObject['class'];
+                $args   = $newObject['args'];
+                ksort($args);
```
fesor: Needs test case

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -324,6 +325,30 @@ protected function gatherRowData(array $data, array &$id, array &$nonemptyCompon
             }
         }
 
+        foreach ($this->_rsm->nestedNewObjectArguments as $objIndex => ['ownerIndex' => $ownerIndex, 'argIndex' => $argIndex]) {
```
fesor: Looks like scrutinizer doesn't know how to deal with destructuring.

lcobucci:
@fesor looks like this patch is going in a nice way, however I wonder that we should rather do this on v3 instead. Not sure about what @guilhermeblanco and @Ocramius think about this, but I believe it's about time to do a feature freeze on 2.x.

Ocramius:
This will unlikely land in 2.x due to the current pressure to get 2.6 out
and start with 3.x overall.

I'd say we leave it as-is and ask for a rebase once `develop` is merged.

fesor:
@Ocramius rebased on master.

tkleinhakisa:
Very nice! Does it handle collections ? eg if User has multiple addresses and constructor param is an array in UserDTO

fesor:
@tkleinhakisa you mean handling *-to-many relations? Currently no... need to check how many changes in hydrator is required to make it possible.

I think that whole hydration system should be refactored so this and many other use cases could be possible without huge overhead, like the one in #2553 for example. But in order to make things both maintainable and performant, this would require some kind of inlining functionality and code generation...

Majkl578:
I'd probably suggest to postpone further work here for now if it requires further redesign. 2.x branch is in feature-freeze and for 3.0 we've been discussing a full DQL parser rewrite. No roadmap for that yet, but I've already been prepping a switch to [Hoa\Compiler](https://github.com/hoaproject/Compiler).

joshhornby:
What are the chances of this making it into v3?

Majkl578:
It's definitely something I'd like to pursue since I believe DTOs have high potential, but likely not before rewriting the parser itself, for that see #7017.

derrabus:

norkunas:
Hoa packages are abandoned. Could this be reconsidered?

greg0ire:
> Please open a new one if you want to pick up the work one day.

PR #6714, opened 2017-09-21
Remove unnecessary else

guiajlopes:
Just a small improvement in the code since the else is unnecessary on this case

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -752,6 +752,7 @@ class SqlWalker implements TreeWalker
                 if ( ! ($assoc['isOwningSide'] && $assoc['type'] & ClassMetadata::TO_ONE)) {
                     continue;
                 }
+                
                 if ( !$addMetaColumns && !isset($assoc['id'])) {
                     continue;
                 }
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -749,11 +749,8 @@ class SqlWalker implements TreeWalker
 
             // Add foreign key columns of class and also parent classes
             foreach ($class->associationMappings as $assoc) {
-                if ( ! ($assoc['isOwningSide'] && $assoc['type'] & ClassMetadata::TO_ONE)) {
-                    continue;
-                }
-                
-                if ( !$addMetaColumns && !isset($assoc['id'])) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -751,7 +751,8 @@ public function walkSelectClause($selectClause)
             foreach ($class->associationMappings as $assoc) {
                 if ( ! ($assoc['isOwningSide'] && $assoc['type'] & ClassMetadata::TO_ONE)) {
                     continue;
-                } else if ( !$addMetaColumns && !isset($assoc['id'])) {
+                }
+                if ( !$addMetaColumns && !isset($assoc['id'])) {
```
Ocramius: Add an empty line between logical blocks

guiajlopes: @Ocramius thanks, done!

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -751,7 +751,9 @@ public function walkSelectClause($selectClause)
             foreach ($class->associationMappings as $assoc) {
                 if ( ! ($assoc['isOwningSide'] && $assoc['type'] & ClassMetadata::TO_ONE)) {
                     continue;
-                } else if ( !$addMetaColumns && !isset($assoc['id'])) {
+                }
+                
+                if ( !$addMetaColumns && !isset($assoc['id'])) {
```
lcobucci: @guiajlopes according to our CS we need spaces before AND after the negation operator, so this should be:

```php
if ( ! $addMetaColumns && ! isset($assoc['id'])) {
```

lcobucci (changes requested):

lib/Doctrine/ORM/Query/SqlWalker.php
lcobucci: Does it make sense to turn this into a single `IF` with an `OR` instead? Like:

```php
if ( ! ($assoc['isOwningSide'] && $assoc['type'] & ClassMetadata::TO_ONE)
    || ( ! $addMetaColumns && ! isset($assoc['id']))) {
```

I mean, both are fine... just wondering 😜

guiajlopes: @lcobucci thanks for your review, make sense to turn in the same if statement

Ocramius (approved):
LGTM, thanks @guiajlopes

PR #6715, opened 2017-09-21
Promote nullable types from PHP 7.1

PowerKiKi:
Also don't show type hinting in a bad light with sentences such as "If you insist on type-hinting"

### DIFF
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -238,14 +238,19 @@ the database permanently.
 
 Notice how both sides of the bidirectional association are always
 updated. Unidirectional associations are consequently simpler to
-handle. Also note that if you use type-hinting in your methods, i.e.
-``setAddress(Address $address)``, PHP will only allow null
-values if ``null`` is set as default value. Otherwise
-setAddress(null) will fail for removing the association. If you
-insist on type-hinting a typical way to deal with this is to
-provide a special method, like ``removeAddress()``. This can also
-provide better encapsulation as it hides the internal meaning of
-not having an address.
+handle.
+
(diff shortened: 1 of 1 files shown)


docs/en/reference/working-with-associations.rst
```diff
@@ -238,14 +238,19 @@ the database permanently.
...
-insist on type-hinting a typical way to deal with this is to
-provide a special method, like ``removeAddress()``. This can also
-provide better encapsulation as it hides the internal meaning of
-not having an address.
+handle.
+
+Also note that if you use type-hinting in your methods, i.e.
+``setAddress(Address $address)``, you will have to specifically
+allow null values, otherwise ``setAddress(null)`` will fail to 
+remove the association. Starting from PHP 7.1 you should use
```
Ocramius: I wouldn't add the detail of PHP 7.1: just the fact that a nullable hint should be used

docs/en/reference/working-with-associations.rst
```diff
@@ -238,14 +238,19 @@ the database permanently.
...
-provide better encapsulation as it hides the internal meaning of
-not having an address.
+handle.
+
+Also note that if you use type-hinting in your methods, i.e.
+``setAddress(Address $address)``, you will have to specifically
+allow null values, otherwise ``setAddress(null)`` will fail to 
+remove the association. Starting from PHP 7.1 you should use 
+nullable types by prefixing the type with a ``?``, 
+``setAddress(?Address $address)``. Older PHP versions will only
```
Ocramius: We really don't care about older PHP versions here

Ocramius (changes requested):

Ocramius (approved):
LGTM :+1: 

Thanks @PowerKiKi!

PR #6717, opened 2017-09-21
Code cleanup in UnitOfWork class

guiajlopes:
Some code cleanup in the Doctrine\ORM\UnitOfWork class

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -339,9 +339,13 @@ class UnitOfWork implements PropertyChangedListener
         // Compute changes done since last commit.
         if (null === $entity) {
             $this->computeChangeSets();
-        } elseif (is_object($entity)) {
+        }
+
+        if (is_object($entity)) {
             $this->computeSingleEntityChangeSet($entity);
-        } elseif (is_array($entity)) {
+        }
+
+        if (is_array($entity)) {
             foreach ($entity as $object) {
                 $this->computeSingleEntityChangeSet($object);
             }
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2541,7 +2541,10 @@ private function newInstance($class)
     public function createEntity($className, array $data, &$hints = [])
     {
         $class = $this->em->getClassMetadata($className);
-        //$isReadOnly = isset($hints[Query::HINT_READ_ONLY]);
+
+        if ( ! $class instanceof ClassMetadata) {
+          return $entity;
```
Ocramius: We use 4 spaces indentation. Also, `ClassMetadata` is always an instance of `Doctrine\ORM\Mapping\ClassMetadata`: the check should be removed (and yes, I know the metadata factory interface returns a `Doctrine\Common\Persistence\Mapping\ClassMetadata` as per contract)

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2828,7 +2830,7 @@ public function createEntity($className, array $data, &$hints = [])
      */
     public function triggerEagerLoads()
     {
-        if ( ! $this->eagerLoadingEntities) {
+        if (empty($this->eagerLoadingEntities)) {
```
Ocramius: This should be reverted, since a simple check on the array is sufficient, while `empty()` introduces further risks instead. Overall, avoid using `empty()`, as it is one of the many weird PHP operator-non-operator-will-work-with-too-many-things.

Ocramius (changes requested):
Overall, the only diff remaining in this patch after my review are two line deletions, so I'm closing this as `invalid`, sorry @guiajlopes :-\

lib/Doctrine/ORM/UnitOfWork.php
guiajlopes: I've just added that because of Scrutinizer complaining about it

lib/Doctrine/ORM/UnitOfWork.php
guiajlopes: I've also just added that because of scrutinizer complaining about it https://scrutinizer-ci.com/g/doctrine/doctrine2/inspections/2976232d-f9dc-4542-8680-7b6cb411aab7/issues/files/lib/Doctrine/ORM/UnitOfWork.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

PR #1241, opened 2015-01-09
Lazy-load on a per-property base (removes doctrine proxies, replaced by ProxyManager)

Ocramius:
This is just a proof of concept that I hacked together real quick.

This changes how proxies are generated and lazy-loaded in a very radical way:
- method calls do not cause lazy-loading
- lazy-loading is triggered by access to non-transient properties (things that need to be lazy-loaded)
- access to identifiers or non-mapped properties won't cause lazy-loading
- you can now use `$class = get_class($proxy); $object = new $class();` without any side-effects, as the constructor logic is preserved

Following BC breaks need to be fixed or discussed before going forward on this idea:
- [x] proxies don't implement `Doctrine\Common\Proxy\Proxy` anymore, but `ProxyManager\Proxy\GhostObjectInterface` ~~(BC break, needs fixing, can be easily done with some effort)~~ BC break is voluntary
- [x] `serialize($proxy)` now causes proxy initialization ~~(probably needs fixing, as this is a major BC break)~~ - BC break is voluntary

Pending TODOs:
- [ ] cloning a proxy is still not fully supported (requires dedicated logic in `__clone`)
- [ ] this is just a PoC, so the code that writes proxies to disk is not yet in place
- [ ] hhvm compatibility (Ocramius/ProxyManager#210, facebook/hhvm#4560)
- [x] ProxyManager 2.0.0 is required first (release highly depends on this issue)
- [ ] Lots of dead code to remove

To give you an example, it is impossible (without reflection) to cause lazy-loading for the following class:

``` php
class Counter
{
    private $id;
    private $otherUnrelatedThing;
    public function doThings() { /* empty body (on purpose) */ }
}
```

[Review on Reviewable](https://reviewable.io/reviews/doctrine/doctrine2/1241)

PR #6649, opened 2017-08-25
#6638 #6648 one-to-one association being refreshed even if not built by the hydrator or having `HINT_REFRESH` set

Ocramius:
Fixes #6638
Fixes #6648 

This patch is just a first stab at the issue: the `ObjectHydrator` is massively broken because of this bug, and the idea is that we need to internally cache whether an object is BUILT in the current `ObjectHydrator`, or whether it already exists as a loaded and managed entity in the `UnitOfWork`.

This has to happen before calling `createEntity()` on the `UnitOfWork`, or else information is lost.

PR #6719, opened 2017-09-22
Replace proxy layer with ProxyManager's layer

Ocramius:
Replaces #1241

TODOs:

 * [x] replace proxy instantiation with ProxyManager proxies
 * [x] ensure that `ProxyManager` class names are properly recognized
 * [x] remove `ClassUtils` shim
 * [x] ~~~move proxy instantiation and definition to the metadata~~~
 * [x] remove `Proxy*` related code in ORM
 * [x] update `Doctrine\ORM\Configuration` to work with `ProxyManager` utilities
 * [x] update `ClassUtils` to work with `ProxyManager` proxy class names
 * [x] remove public properties hack
 * [x] extract identifier normalization to its own object that is the opposite of [`IdentifierFlattener`](https://github.com/doctrine/doctrine2/blob/1d5c87ee4bcf22d699530b3792739f33b91cf6cd/lib/Doctrine/ORM/Utility/IdentifierFlattener.php)
 * [x] test "friend object" behavior with proxies (shared private scope access)
 * [x] test that `EntityManager#getReference()` calls `ProxyFactory#getProxy()` with un-touched data
 * [x] document BC break extent:
    * [x] `__sleep` behavior changes
    * [x] `__wakeup` behavior changes
    * [x] `serialize()` no longer breaks the graph -> potentially dangerous if serializing over large association graphs
    * [x] `Proxy` interface gone
    * [x] `private` property access now triggers lazy-loading
    * [x] `ReflectionProperty` property access now triggers lazy-loading
 * [x] document improvements
    * [x] "friend object" behavior with proxies (shared private scope access)
    * [x] `private` property access now triggers lazy-loading
    * [x] `ReflectionProperty` property access now triggers lazy-loading
    * [x] proxies no longer lazy-load when no internal state is accessed
    * [x] no more "identifier getter" hacks
    * [x] no more type conversion hacks
    * [x] identifier fields are always set, and always have the expected type
 * [x] benchmark regression testing


lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
```diff
@@ -124,11 +124,11 @@ protected function hydrateRowData(array $sqlResult, array &$result)
             }
         }
 
-        if (isset($this->hints[Query::HINT_REFRESH_ENTITY])) {
-            $id = $this->em->getIdentifierFlattener()->flattenIdentifier($this->class, $data);
-
-            $this->em->getUnitOfWork()->registerManaged($this->hints[Query::HINT_REFRESH_ENTITY], $id, $data);
-        }
+//        if (isset($this->hints[Query::HINT_REFRESH_ENTITY])) {
```
Ocramius: This code was extremely broken, but I still need to see if any other related code may have been broken by it.

lib/Doctrine/ORM/Proxy/Factory/DefaultProxyResolver.php
```diff
@@ -5,9 +5,6 @@
 
 namespace Doctrine\ORM\Proxy\Factory;
 
-use Doctrine\Common\Util\ClassUtils;
-use Doctrine\ORM\Proxy\Proxy;
-
 class DefaultProxyResolver implements ProxyResolver
```
Ocramius: To be removed

lib/Doctrine/ORM/Proxy/Factory/ProxyDefinitionFactory.php
```diff
@@ -58,11 +58,12 @@ public function build(ClassMetadata $classMetadata) : ProxyDefinition
     {
         $definition = $this->createDefinition($classMetadata);
 
-        if (! class_exists($definition->proxyClassName, false)) {
-            $proxyClassPath = $this->resolver->resolveProxyClassPath($classMetadata->getClassName());
-
-            $this->generatorStrategy->generate($proxyClassPath, $definition);
-        }
+        // @TODO kill this. We no longer generate proxies the "old way"
```
Ocramius: To be removed

lib/Doctrine/ORM/Proxy/Factory/ProxyFactory.php
```diff
@@ -56,11 +57,6 @@ public function generateProxyClasses(array $classMetadataList) : int;
...
      * Gets a reference proxy instance for the entity of the given type and identified by
      * the given identifier.
-     *
-     * @param string $className
-     * @param array  $identifier
-     *
-     * @return \Doctrine\ORM\Proxy\Proxy
      */
-    public function getProxy(string $className, array $identifier) : Proxy;
+    public function getProxy(string $className, array $identifier) : GhostObjectInterface;
```
Ocramius: Should probably become `ClassMetadata $class` to avoid repeating metadata lookup operations

lib/Doctrine/ORM/Proxy/Factory/ProxyGenerator.php
```diff
@@ -6,7 +6,7 @@
 namespace Doctrine\ORM\Proxy\Factory;
```
Ocramius: To be removed

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -7,8 +7,13 @@
 
 use Doctrine\ORM\Configuration\ProxyConfiguration;
```
Ocramius: (probably) to be removed

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -82,18 +87,110 @@ public function generateProxyClasses(array $classMetadataList) : int
...
      */
-    public function getProxy(string $className, array $identifier) : Proxy
+    public function getProxy(string $className, array $identifier) : GhostObjectInterface
     {
+        $metadata = $this->entityManager->getClassMetadata($className);
+
+        $proxyInstance = (new LazyLoadingGhostFactory(new Configuration()))
+            ->createProxy(
+                $metadata->getClassName(),
+                function (
```
Ocramius: This loader should be cached

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -82,18 +87,110 @@ public function generateProxyClasses(array $classMetadataList) : int
...
     /**
      * {@inheritdoc}
+     * @throws \Doctrine\ORM\EntityNotFoundException
      */
-    public function getProxy(string $className, array $identifier) : Proxy
+    public function getProxy(string $className, array $identifier) : GhostObjectInterface
     {
+        $metadata = $this->entityManager->getClassMetadata($className);
+
+        $proxyInstance = (new LazyLoadingGhostFactory(new Configuration()))
```
Ocramius: This factory should be injected

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -82,18 +87,110 @@ public function generateProxyClasses(array $classMetadataList) : int
...
+                            $metadata->getClassName(),
+                            $identifier
+                        );
+                    }
+
+                    return true;
+                },
+                [
+                    // @TODO this should be a constant reference, not a magic constant
+                    'skippedProperties' => \array_merge(
```
Ocramius: For now, we'll have to use a `private const`

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -965,11 +969,25 @@ public function AbstractSchemaName()
      *
      * @throws QueryException if the name does not exist.
      */
-    private function validateAbstractSchemaName($schemaName)
+    private function validateAbstractSchemaName($schemaName) : void
     {
-        if (! (class_exists($schemaName, true) || interface_exists($schemaName, true))) {
-            $this->semanticalError("Class '$schemaName' is not defined.", $this->lexer->token);
+        // @TODO are these checks even needed?
```
Ocramius: The check should probably be inversed: `getClassMetadata()` first, then `(class|interface)_exists()`. The reason being that `class_exists()` can trigger I/O, while `getClassMetadata()` usually doesn't.

This makes `SELECT new Foo(b.c) FROM Bar b` slower, as `getClassMetadata()` fails more often, but it makes other DQL queries faster.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2108,29 +2112,31 @@ public function createEntity($className, array $data, &$hints = [])
...
+                // @TODO ideally, all this scenario should be completely gone
+                $entity = $unmanagedProxy;
                 // DDC-1238 - we have a managed instance, but it isn't the provided one.
                 // Therefore we clear its identifier. Also, we must re-fetch metadata since the
                 // refreshed object may be anything
-                foreach ($class->identifier as $fieldName) {
-                    $property = $class->getProperty($fieldName);
+                // @TODO DDC-1238 seems to be invalid.
+                // @TODO We shouldn't touch the object state, but instead just consider it "unmanaged" and skip further hydration
+//                foreach ($class->identifier as $fieldName) {
```
Ocramius: This block was super-broken: we just want an un-managed instance, but we really don't want to change its identifier. Replacing the identifier = bad.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2108,29 +2112,31 @@ public function createEntity($className, array $data, &$hints = [])
                 isset($hints[Query::HINT_REFRESH])
                 && isset($hints[Query::HINT_REFRESH_ENTITY])
                 && ($unmanagedProxy = $hints[Query::HINT_REFRESH_ENTITY]) !== $entity
-                && $unmanagedProxy instanceof Proxy
+                && $unmanagedProxy instanceof GhostObjectInterface
                 && $this->isIdentifierEquals($unmanagedProxy, $entity)
             ) {
+                // @TODO messy, but for now we try continuing operation as normal
```
Ocramius: No longer valid once removing this block

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2108,29 +2112,31 @@ public function createEntity($className, array $data, &$hints = [])
...
+                // @TODO ideally, all this scenario should be completely gone
```
Ocramius: No longer valid once removing this block

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2108,29 +2112,31 @@ public function createEntity($className, array $data, &$hints = [])
...
+                $entity = $unmanagedProxy;
                 // DDC-1238 - we have a managed instance, but it isn't the provided one.
                 // Therefore we clear its identifier. Also, we must re-fetch metadata since the
                 // refreshed object may be anything
-                foreach ($class->identifier as $fieldName) {
-                    $property = $class->getProperty($fieldName);
+                // @TODO DDC-1238 seems to be invalid.
```
Ocramius: This refers to the [`DDC1238Test`](https://github.com/doctrine/doctrine2/blob/1d5c87ee4bcf22d699530b3792739f33b91cf6cd/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC1238Test.php)

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2108,29 +2112,31 @@ public function createEntity($className, array $data, &$hints = [])
...
+//                    $property = $class->getProperty($fieldName);
+//
+//                    $property->setValue($unmanagedProxy, null);
+//                }
 
-                    $property->setValue($unmanagedProxy, null);
-                }
-
-                return $unmanagedProxy;
+                // @TODO overall, this logic seems wrong and harmful. Even if the proxy is un-managed, we want to load its data here
```
Ocramius: No longer valid

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2108,29 +2112,31 @@ public function createEntity($className, array $data, &$hints = [])
...
+//                return $unmanagedProxy;
```
Ocramius: No longer valid

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2158,17 +2164,17 @@ public function createEntity($className, array $data, &$hints = [])
...
-            }
-
             $overrideLocalValues = true;
         }
 
         if ( ! $overrideLocalValues) {
             return $entity;
         }
 
+        if ($entity instanceof NotifyPropertyChanged) {
```
Ocramius: This was moved, because the `addPropertyChangedListener()` call must happen both for initialized and un-initialized proxies once the `UnitOfWork` starts touching them

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2339,28 +2345,50 @@ public function createEntity($className, array $data, &$hints = [])
                     // Proxies do not carry any kind of original entity data until they're fully loaded/initialized
                     $managedData = [];
 
+                    // @TODO refactor: `getProxy`: no scalars if the identifier is an association, use real value
```
Ocramius: Comment block to be removed: replicated below

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2339,28 +2345,50 @@ public function createEntity($className, array $data, &$hints = [])
...
                         default:
                             // TODO: This is very imperformant, ignore it?
-                            $newValue    = $this->em->find($targetEntity, $associatedId);
+                            $newValue    = $this->em->find($targetEntity, $normalizedAssociatedId);
                             // Needed to re-assign original entity data for freshly loaded entity
                             $managedData = $this->originalEntityData[spl_object_hash($newValue)];
                             break;
                     }
 
+                    // @TODO using `$associatedId` here seems to be risky.
```
Ocramius: Very risky indeed: this should be a `flattenId(normalizeId(flatId))` instead, in order to guarantee consistency

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2369,7 +2397,16 @@ public function createEntity($className, array $data, &$hints = [])
             $this->originalEntityData[$oid][$field] = $newValue;
             $association->setValue($entity, $newValue);
 
-            if ($association->getInversedBy() && $association instanceof OneToOneAssociationMetadata) {
+            if (
```
Ocramius: I think this entire block seems wrong and should be removed. The code in this block overwrites the association field of the related entity, which is wrong.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2531,6 +2568,86 @@ public function getSingleIdentifierValue($entity)
     }
 
     /**
+     * @param array  $id
+     * @param string $rootClassName
+     *
+     * @return GhostObjectInterface|object
+     */
+    private function tryGetByIdOrLoadProxy(array $id, string $rootClassName)
```
Ocramius: This method is extracted from the `EntityManager` API

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2531,6 +2568,86 @@ public function getSingleIdentifierValue($entity)
...
+     * @TODO we need to therefore have a generic utility that, given a scalar identifier, gives us
+     *       a "normalized" identifier. This is pretty much the opposite of what the
+     *       `IdentifierFlattener` does
+     * @TODO note that we also need to sort the identifier fields here
+     *
+     * @TODO Start of the block to be refactored into "make a deep identifier from a flat one"
+     *
+     * @return mixed[]
+     */
+    private function convertFlatIdentifierIntoRealIdentifierFieldValues(
```
Ocramius: This method is to be extracted to the `ClassMetadata` API, which should (given a `UnitOfWork`) normalize/denormalize identifiers for us.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2531,6 +2568,86 @@ public function getSingleIdentifierValue($entity)
...
+        foreach ($targetClass->getDeclaredPropertiesIterator() as $name => $declaredProperty) {
+            if (! \array_key_exists($name, $flatIdentifier)) {
+                continue;
+            }
+
+            if ($declaredProperty instanceof ToOneAssociationMetadata) {
+                $targetIdMetadata = $this->em->getClassMetadata($declaredProperty->getTargetEntity());
+
+                $normalizedAssociatedId[$name] = $this->tryGetByIdOrLoadProxy(
+                    [reset($targetIdMetadata->identifier) => $flatIdentifier[$name]], // @TODO this is where stuff breaks with composite IDs :-(
```
Ocramius: This assumes single-field identifiers on associated proxies: extremely dangerous

tests/Doctrine/Tests/ORM/Functional/ProxiesLikeEntitiesTest.php
```diff
@@ -52,6 +59,8 @@ protected function setUp()
         $this->em->persist($this->user);
         $this->em->flush();
         $this->em->clear();
+
+        $this->proxyClassName = (new Configuration())->getClassNameInflector()->getProxyClassName(CmsUser::class);
```
Ocramius: This is probably not the correct approach

tests/Doctrine/Tests/ORM/Functional/ReferenceProxyTest.php
```diff
@@ -165,46 +167,30 @@ public function testInitializeChangeAndFlushProxy()
...
+        /* @var $entity ECommerceProduct|GhostObjectInterface */
         $entity = $this->em->getReference(ECommerceProduct::class , $id);
 
         self::assertEquals('Doctrine 2 Cookbook', $entity->getName());
     }
 
-    /**
-     * @group DDC-1022
-     */
-    public function testWakeupCalledOnProxy()
```
Ocramius: This no longer holds true: we don't call `__wakeup` on proxies anymore.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC1238Test.php
```diff
@@ -58,14 +59,29 @@ public function testIssueProxyClear()
...
-        self::assertNull($user->getId(), "Now this is null, we already have a user instance of that type");
+        $user->initializeProxy();
+
+        self::assertInternalType(
+            'integer',
+            $user->getId(),
+            'Even if a proxy is detached, it should still have an identifier'
+        );
+
+        $user2->initializeProxy();
```
Ocramius: Just added a block checking that we didn't ruin the fresh instance.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3223Test.php
```diff
@@ -67,7 +81,11 @@ class Participant
     /** @ORM\Id @ORM\Column(type="integer") @ORM\GeneratedValue */
     public $id;
 
-    /** @ORM\ManyToOne(targetEntity="ProfileStatus") */
+    /**
+     * @ORM\ManyToOne(targetEntity="ProfileStatus")
+     * @ORM\JoinColumn(name="status_id", nullable=false)
```
Ocramius: Added the `nullable=false` since I added the `@var` block with no nullability

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
```diff
@@ -185,40 +191,42 @@ public function testFailedProxyCloningDoesNotMarkTheProxyAsInitialized()
         } catch (EntityNotFoundException $exception) {
         }
 
-        self::assertFalse($proxy->__isInitialized());
+        self::assertFalse($proxy->isProxyInitialized());
     }
 
     public function testProxyClonesParentFields()
     {
-        $companyEmployee = new CompanyEmployee();
```
Ocramius: This was completely broken: the persister was returning a new instance instead of just hydrating the given object.

shims/ClassUtils.php
```diff
@@ -0,0 +1,131 @@
+<?php
```
Ocramius: This file is to be removed before merge

Ocramius:
I must say that I have NO IDEA why travis fails and local tests pass...

Slamdunk:
>  Replace proxy layer with ProxyManager's layer

To me sounds a lot like a big _Conflict of interest_

> I must say that I have NO IDEA why travis fails and local tests pass...

Works on my machine

Ocramius:
Tbh, I was just sick of fixing the same bugs on both projects, plus the
proxy implementation advantages have yet to be documented 😁

lib/Doctrine/ORM/Configuration.php
```diff
@@ -46,12 +57,16 @@ class Configuration extends \Doctrine\DBAL\Configuration
...
+
+        $this->getProxyManagerConfiguration()->setProxiesTargetDir($dir);
     }
 
     /**
      * Gets the directory where Doctrine generates any necessary proxy class files.
      *
      * @return string|null
+     *
+     * @deprecated please do not use this anymore
```
Ocramius: Should we drop this?

lib/Doctrine/ORM/Configuration.php
```diff
@@ -64,6 +79,8 @@ public function getProxyDir()
      * Gets the strategy for automatically generating proxy classes.
      *
      * @return int Possible values are constants of Doctrine\ORM\Proxy\Factory\StaticProxyFactory.
+     *
+     * @deprecated please do not use this anymore
```
Ocramius: Should we drop this?

lib/Doctrine/ORM/Configuration.php
```diff
@@ -83,12 +100,32 @@ public function getAutoGenerateProxyClasses()
...
+                return;
+        }
     }
 
     /**
      * Gets the namespace where proxy classes reside.
      *
      * @return string|null
+     *
+     * @deprecated please do not use this anymore
```
Ocramius: Should we drop this?

Ocramius:
This is done so far: next up is just the documentation

Ocramius:
Benchmark results:

```
benchmark: UnitOfWorkComputeChangesBench, subject: benchChangeSetComputation
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 1,305.988μs             | 1,113.150μs                                   |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: ProxyInitializationTimeBench, subject: benchCmsEmployeeInitialization
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 4,492.494μs             | 4,945.614μs                                   |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: ProxyInstantiationTimeBench, subject: benchCmsEmployeeInstantiation
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 72,182.124μs            | 194,942.108μs                                 |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: ProxyInitializationTimeBench, subject: benchCmsUserInitialization
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 8,947.280μs             | 5,902.738μs                                   |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: ProxyInstantiationTimeBench, subject: benchCmsUserInstantiation
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 119,298.386μs           | 137,659.690μs                                 |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SingleTableInheritanceHydrationPerformanceBench, subject: benchHydrateAllContracts
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 1,219.090μs             | 1,245.426μs                                   |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SingleTableInheritanceHydrationPerformanceBench, subject: benchHydrateFixContracts
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 604.638μs               | 617.326μs                                     |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SingleTableInheritanceHydrationPerformanceBench, subject: benchHydrateFlexContracts
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 912.946μs               | 939.842μs                                     |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SingleTableInheritanceHydrationPerformanceBench, subject: benchHydrateUltraContracts
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 670.386μs               | 689.746μs                                     |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: MixedQueryFetchJoinArrayHydrationPerformanceBench, subject: benchHydration
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 10,006.246μs            | 10,198.008μs                                  |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: MixedQueryFetchJoinFullObjectHydrationPerformanceBench, subject: benchHydration
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 15,528.624μs            | 15,298.930μs                                  |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: MixedQueryFetchJoinPartialObjectHydrationPerformanceBench, subject: benchHydration
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 7,344.548μs             | 7,238.552μs                                   |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SimpleHydrationBench, subject: benchHydration
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 114,677.968μs           | 112,961.742μs                                 |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SimpleInsertPerformanceBench, subject: benchHydration
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 548,459.208μs           | 518,215.356μs                                 |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SimpleQueryArrayHydrationPerformanceBench, subject: benchHydration
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 5,696.064μs             | 5,807.282μs                                   |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SimpleQueryFullObjectHydrationPerformanceBench, subject: benchHydration
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 51,989.224μs            | 50,381.912μs                                  |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SimpleQueryPartialObjectHydrationPerformanceBench, subject: benchHydration
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 12,397.502μs            | 12,096.792μs                                  |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SimpleQueryScalarHydrationPerformanceBench, subject: benchHydration
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 4,043.822μs             | 4,151.792μs                                   |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: ProxyInitializationTimeBench, subject: benchInitializationOfAlreadyInitializedCmsEmployees
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 2,865.794μs             | 480.390μs                                     |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: ProxyInitializationTimeBench, subject: benchInitializationOfAlreadyInitializedCmsUsers
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 3,205.172μs             | 571.826μs                                     |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SingleTableInheritanceInsertPerformanceBench, subject: benchInsertAllContracts
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 2,088.532μs             | 2,101.618μs                                   |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SingleTableInheritanceInsertPerformanceBench, subject: benchInsertFixContracts
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 832.212μs               | 855.418μs                                     |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SingleTableInheritanceInsertPerformanceBench, subject: benchInsertFlexContracts
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 914.024μs               | 922.926μs                                     |
+--------+--------+------+-------------------------+-----------------------------------------------+

benchmark: SingleTableInheritanceInsertPerformanceBench, subject: benchInsertUltraContracts
+--------+--------+------+-------------------------+-----------------------------------------------+
| groups | params | revs | vcs_branch:develop:mean | vcs_branch:cleanup/replace-proxy-factory:mean |
+--------+--------+------+-------------------------+-----------------------------------------------+
|        | []     | 5    | 946.498μs               | 942.446μs                                     |
+--------+--------+------+-------------------------+-----------------------------------------------+
```

Overall quite satisfied by these, even if initialization times times went up, you have to remember that we removed one layer of indirection for method calls, and also removed some lazy-loading for any operations not accessing internal object state.

lib/Doctrine/ORM/Configuration.php
Ocramius: Not removed: too commonly used, therefore keeping it and documented as deprecated

lib/Doctrine/ORM/Configuration.php
Ocramius: Not removed: too commonly used, therefore keeping it and documented as deprecated

lib/Doctrine/ORM/Configuration.php
Ocramius: Not removed: too commonly used, therefore keeping it and documented as deprecated

lib/Doctrine/ORM/UnitOfWork.php
Ocramius: This code cannot be fixed in here

lib/Doctrine/ORM/UnitOfWork.php
Ocramius: This code cannot be fixed in this patch: possibly related to #6638, #6648, #6649

lib/Doctrine/ORM/Proxy/Factory/ProxyFactory.php
Ocramius: Not changing this for now - can be done later, eventually

docs/en/reference/advanced-configuration.rst
```diff
@@ -221,45 +223,76 @@ option that controls this behavior is:
...
-value. This is still possible, ``FALSE`` being equivalent to
-AUTOGENERATE_NEVER and ``TRUE`` to AUTOGENERATE_ALWAYS.
+Here is an example of such setup:
+
+.. code-block:: json
+
+    {
+        "autoload": {
+            "psr-4": {
+                "MyProject\\": "path/to/project/sources/"
```
jaapio: A , (comma) is missing at the end of this line

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
```diff
@@ -401,7 +400,7 @@ protected function hydrateRowData(array $row, array &$result)
                     $reflFieldValue = $association->getValue($parentObject);
 
                     if (! $reflFieldValue || isset($this->hints[Query::HINT_REFRESH]) ||
-                        ($reflFieldValue instanceof Proxy && ! $reflFieldValue->__isInitialized())) {
+                        ($reflFieldValue instanceof GhostObjectInterface && ! $reflFieldValue->isProxyInitialized())) {
```
jaapio: It doesn't really matter, but this check is everywhere in the code. I would personally extract it to a helper function. Although an extra method call could have impact on the execution time, depending on the location of these checks.

jaapio (commented):
I'm not a doctrine insider. So I didn't understand every change. Especially the usage of final methods should be documented for me as end-user. What does it mean of my entities when I use final methods?

docs/en/tutorials/getting-started.rst
```diff
@@ -49,14 +49,14 @@ What are Entities?
...
-abstract base class or interface. An entity class must not be final
-or contain final methods. Additionally it must not implement
-**clone** nor **wakeup**, unless it :doc:`does so safely <../cookbook/implementing-wakeup-or-clone>`.
+abstract base class or interface.
 
 An entity contains persistable properties. A persistable property
 is an instance variable of the entity that is saved into and retrieved from the database
 by Doctrine's data mapping capabilities.
 
+An entity class must not be ``final``, although it can contain ``final`` methods.
```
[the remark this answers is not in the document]
guilhermeblanco: I'm assuming some sort of Exception should be triggered in this case. If this correct, @Ocramius ?

lib/Doctrine/ORM/Configuration.php
guilhermeblanco: Drop it! Unless you can make it fully functional to keep BC.
I didn't for Metadata, so I'm assuming if people will already change for Metadata, they could do for Proxy also.

lib/Doctrine/ORM/Proxy/Factory/ProxyFactory.php
```diff
@@ -47,7 +47,7 @@
     const AUTOGENERATE_EVAL = 3;
 
     /**
-     * @param array<Doctrine\ORM\Mapping\ClassMetadata> $classMetadataList
```
guilhermeblanco: https://i.redditmedia.com/0UfebEHWhsAa6CyTcxXnHzSnVdaw3x2TRFJjdT_-CGw.jpg?w=538&s=724ae8ed0e8391ab9f3f36c4458cb43c

lib/Doctrine/ORM/Proxy/Factory/ProxyFactory.php
guilhermeblanco: Do it now and make it fully finished! Unless you plan to make subsequent PRs on the funcionality.

guilhermeblanco:
@Ocramius overall PR looks solid. Feel free to merge once you've looked at the comments.
The `NormalizeIdentifier` is yet for discussion, but I don't want to disrupt the main purpose of this PR for no reason.

docs/en/tutorials/getting-started.rst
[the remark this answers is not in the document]
Ocramius: Yes, an exception is raised by ProxyManager when trying to generate proxies for classes that aren't really proxy-able

lib/Doctrine/ORM/Proxy/Factory/ProxyFactory.php
Ocramius: I'll dig in today 👍

lib/Doctrine/ORM/Configuration.php
Ocramius: I kept it because this is a high visibility endpoint: most integration layers with the ORM touch the configuration object. I will do a github search before considering a droo

lib/Doctrine/ORM/Configuration.php
```diff
@@ -848,4 +887,15 @@ public function setDefaultQueryHint($name, $value)
     {
         $this->attributes['defaultQueryHints'][$name] = $value;
     }
+
+    public function buildGhostObjectFactory() : LazyLoadingGhostFactory
+    {
+        return new LazyLoadingGhostFactory(clone $this->getProxyManagerConfiguration());
+    }
+
+    public function getProxyManagerConfiguration() : ProxyManagerConfiguration
```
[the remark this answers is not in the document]
Ocramius: It is mutable on purpose to allow users to customise details later on, but internally, a new copy is made every time the proxy factory is used. 

This behaviour is also tested though (to avoid the mutability bugs you mentioned)

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
Ocramius: Sadly, this is a very common problems with the ORM: we cannot extract logic to methods without massive performance impact.

It is indeed very true that it is horrible, but php-src does not yet optimise cross-file calls.

lib/Doctrine/ORM/Proxy/Factory/ProxyFactory.php
Ocramius: First requires to be fixed in known code analysis tools, then we can use it 😁

Ocramius:
@jaapio 

 > What does it mean of my entities when I use final methods?

Before this change, `final` methods in entities were NOT allowed. This limitation simply goes away :-)

Ocramius:
:ship:

carnage:
Done a quick scan of the diff and this seems to be a massive BC break (literally spelled out in upgrading.md); should it not be targetting 3.0.x not 2.10.x ?

greg0ire:
I'm AFK for a week and I don't remember what the plan was here. @beberlei, can you please advise?

Ocramius:
This should really land in 3.x, as the type of the proxies changes completely.

beberlei:
Its not really a madsive BC break as the Proxy interface is mostly internal.

The idea is, when possible and i think that might be, a user should be able to decide which proxy generator to use via config, which would allow downstreams to gradually prepare for 3.0.

That would require proxy manager to be flexible enough to also implement our Proxy interface if possible. Otherwise we can provide APIs that would enable forward compatibility

Issue #6722, opened 2017-09-23
Initialize proxies with all their collections already being `PersistentCollection`


Ocramius:
Starting with #6719, proxies can be partially loaded by just deciding which fields are to be lazy-loaded and which are not: this allows us to initialize them with empty `PersistentCollection` instances upfront.

Take this example:

```php
class Comment
{
    public $id;
}

class BlogPost
{
    public $id;
    public $comments = [];
}

$blogPost = $em->getReference(BlogPost::class, ['id' => 123]);

var_dump(count($blogPost->comments));
```

The logic above triggers two `SELECT` queries in ORM 2.x. In ORM 3.x, we could make it perform just the `SELECT` query that initializes the `BlogPost#$comments` collection if the property is always replaced with a `Doctrine\ORM\PersistentCollection`

PR #6724, opened 2017-09-23
#6722 initialize proxies with their to-many fields already initialized to a `PersistentCollection` instance

Ocramius:
Implements #6722


lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -127,6 +138,15 @@ public function getProxy(string $className, array $identifier) : GhostObjectInte
 
         $persister->setIdentifier($proxyInstance, $identifier);
 
+        foreach ($toManyFields as $toMany) {
+            $collection = $toMany->wrap($proxyInstance, null, $this->entityManager);
+
+            $collection->setInitialized(false); // @TODO not the right approach, IMO
```
Ocramius, line 133: Seems correct, although I'd prefer the `$toMany` `Property` to do this for me :-\

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -186,6 +207,44 @@ private function transientFieldsFqns(ClassMetadata $metadata) : array
...
+    }
+
+    /**
+     * @return ToManyAssociationMetadata[]
+     */
+    private function toManyFields(ClassMetadata $metadata) : array
+    {
+        $toMany = [];
+
+        foreach ($metadata->getDeclaredPropertiesIterator() as $property) {
```
Ocramius: Note: not using `array_map()` because this is an iterator

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -112,6 +121,8 @@ public function getProxy(string $className, array $identifier) : GhostObjectInte
                 ->entityManager
                 ->getUnitOfWork()
                 ->getEntityPersister($metadata->getClassName());
+        $toManyFields = $this->cachedToManyFields[$className]
```
guilhermeblanco, line 113: Blank line yo

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -186,6 +207,44 @@ private function transientFieldsFqns(ClassMetadata $metadata) : array
...
+        foreach ($metadata->getDeclaredPropertiesIterator() as $name => $property) {
+            if (! $property instanceof ToManyAssociationMetadata) {
+                continue;
+            }
+
+            $transientFieldsFqns[] = $this->propertyFqcn(
+                $property
+                    ->getDeclaringClass()
+                    ->getReflectionClass()
+                    ->getProperty($name) // @TODO possible NPR. This should never be null, why is it allowed to be?
```
guilhermeblanco: and it's not. unless you got a bug. =\

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
guilhermeblanco: There're several pro/cons on Iterator for properties. Keep in mind that `getDeclaredProperties` is different from `getProperties` as the latter also covers hierarchy.

I have yet to decide if we'll make it an iterator forever or no.

Ocramius: Ideally, I'd like to have pre-filtered iterators for things like identifiers, association types, etc 😁

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
Ocramius: Same group of assignments, hence no added newline

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
Ocramius: Yes, the problem is that `getProperty()` can lead to a NPE here. Would probably need an API that gives me a reflection property or throws

Ocramius:
This is to be checked: not sure if collection updates are picked up for un-initialized proxies.

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
stof: shouldn't this be done in `->wrap()` instead ? It looks weird to me that wrapping `null` gives you an initialized PersistentCollection.

Majkl578:
Needs a rebase.

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
Ocramius: Yeah, TODO in fact. `wrap` was not working as expected either.

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
```diff
@@ -175,6 +196,44 @@ private function transientFieldsFqns(ClassMetadata $metadata) : array
         return $transientFieldsFqns;
     }
 
+    private function collectionFieldsFqns(ClassMetadata $metadata) : array
+    {
+        $transientFieldsFqns = [];
+
+        foreach ($metadata->getDeclaredPropertiesIterator() as $name => $property) {
+            if (! $property instanceof ToManyAssociationMetadata) {
```
Ocramius: I don't like this kind of filtering of an iterator - do want a dedicated metadata API endpoint

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
Ocramius: Indeed, accessor that throws and has a non-null return type is needed.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -745,6 +774,73 @@ public function computeChangeSets()
         }
     }
 
+    private function computeToManyAssociationChanges(ToManyAssociationMetadata $association, $value)
```
Ocramius: This code is mostly duplicated/extracted from `computeAssociationChanges`. The code duplication can be massively reduced if we use dedicated iterators over property types.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -673,11 +687,31 @@ public function computeChangeSet(ClassMetadata $class, $entity)
...
-            }
+            if ($property instanceof ToManyAssociationMetadata) {
+                $value = $property->getValue($entity);
+
+                if (null === $value) {
+                    continue;
+                }
+
+                $this->computeToManyAssociationChanges($property, $value);
+            } else {
```
Ocramius: Didn't find a cleaner way for now. Also here, iterating over dedicated accessors would be better.

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
Majkl578: Get method should throw imho, unlike find.

lib/Doctrine/ORM/UnitOfWork.php
ostrolucky: `elseif ($property instanceof AssociationMetadata && !$isUnInitializedProxy)` :upside_down_face:

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
Majkl578: Thinking about how to avoid ton of getDeclared*() methods, quick idea: what would you say to PropertyFilterIterator + binary flags to pick your own set + factory methods (forAssociations())?

lib/Doctrine/ORM/UnitOfWork.php
Ocramius: Aye, but the `if` blocks still need de-duplication anyway

lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php
Ocramius: Bad idea: lots of untyped methods then. Having multiple dedicated accessors for metadata isn't a problem IMO.

Ocramius:
Argh, I'll need to rebase this

Pierstoval:
Yep, you do 😛

derrabus:

PR #6725, opened 2017-09-24
Bump versions of some dependencies

cengizhancaliskan:
Enabled to work in Symfony 2.8


Ocramius:
This is not a bump: it is a downgrade.

Also, we no longer work with symfony 2.x: please use `doctrine/orm` `2.5.*`

PR #6728, opened 2017-09-25
Validate XML mapping against XSD file

greg0ire:
# TODO

- [x] fix all the tests I broke
- [x] make it optional and disabled by default, and trigger deprecation if disabled?
- [x] establish what is the right kind of exception to throw (there is a test that asserts a mapping exception from doctrine/common should be thrown, I don't know why yet)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "doctrine/annotations": "<1.13 || >= 2.0"
     },
     "suggest": {
+        "ext-dom": "Provides support for XSD validation for XML mapping files",
         "symfony/cache": "Provides cache support for Setup Tool with doctrine/cache 2.0",
         "symfony/yaml": "If you want to use YAML Metadata Mapping Driver"
     },
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -5,11 +5,13 @@ declare(strict_types=1);
 namespace Doctrine\ORM\Mapping\Driver;
 
 use Doctrine\Common\Collections\Criteria;
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Mapping\Builder\EntityListenerBuilder;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\Persistence\Mapping\ClassMetadata as PersistenceClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\FileDriver;
+use DOMDocument;
 use InvalidArgumentException;
 use LogicException;
 use SimpleXMLElement;
@@ -22,6 +24,9 @@ use function explode;
 use function extension_loaded;
 use function file_get_contents;
 use function in_array;
+use function libxml_clear_errors;
+use function libxml_get_errors;
+use function libxml_use_internal_errors;
 use function simplexml_load_string;
 use function sprintf;
 use function str_replace;
@@ -36,10 +41,13 @@ class XmlDriver extends FileDriver
 {
     public const DEFAULT_FILE_EXTENSION = '.dcm.xml';
 
+    /** @var bool */
+    private $isXsdValidationEnabled;
+
     /**
      * {@inheritDoc}
      */
-    public function __construct($locator, $fileExtension = self::DEFAULT_FILE_EXTENSION)
+    public function __construct($locator, $fileExtension = self::DEFAULT_FILE_EXTENSION, bool $isXsdValidationEnabled = false)
     {
         if (! extension_loaded('simplexml')) {
             throw new LogicException(sprintf(
@@ -47,6 +55,25 @@ class XmlDriver extends FileDriver
                 . ' Please configure PHP with SimpleXML or choose a different metadata driver.'
             ));
         }
+
+        if (! $isXsdValidationEnabled) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/pull/6728',
+                sprintf(
+                    'Using XML mapping driver with XSD validation disabled is deprecated'
+                    . ' and will not be supported in Doctrine ORM 3.0.'
+                )
+            );
+        }
+
+        if ($isXsdValidationEnabled && ! extension_loaded('dom')) {
+            throw new LogicException(sprintf(
+                'XSD validation cannot be enabled because the DOM extension is missing.'
+            ));
+        }
+
+        $this->isXsdValidationEnabled = $isXsdValidationEnabled;
 
         parent::__construct($locator, $fileExtension);
     }
@@ -940,6 +967,7 @@ class XmlDriver extends FileDriver
      */
     protected function loadMappingFile($file)
     {
+        $this->validateMapping($file);
         $result = [];
         // Note: we do not use `simplexml_load_file()` because of https://bugs.php.net/bug.php?id=62577
         $xmlElement = simplexml_load_string(file_get_contents($file));
@@ -965,6 +993,27 @@ class XmlDriver extends FileDriver
         }
 
         return $result;
+    }
+
+    private function validateMapping(string $file): void
+    {
+        if (! $this->isXsdValidationEnabled) {
+            return;
+        }
+
+        $backedUpErrorSetting = libxml_use_internal_errors(true);
+
+        try {
+            $document = new DOMDocument();
+            $document->load($file);
+
+            if (! $document->schemaValidate(__DIR__ . '/../../../../../doctrine-mapping.xsd')) {
+                throw MappingException::fromLibXmlErrors(libxml_get_errors());
+            }
+        } finally {
+            libxml_clear_errors();
+            libxml_use_internal_errors($backedUpErrorSetting);
+        }
     }
 
     /**
--- a/lib/Doctrine/ORM/Mapping/MappingException.php
+++ b/lib/Doctrine/ORM/Mapping/MappingException.php
@@ -6,6 +6,7 @@ namespace Doctrine\ORM\Mapping;
 
 use BackedEnum;
 use Doctrine\ORM\Exception\ORMException;
+use LibXMLError;
 use ReflectionException;
 use ValueError;
 
@@ -16,6 +17,8 @@ use function get_debug_type;
 use function get_parent_class;
 use function implode;
 use function sprintf;
+
+use const PHP_EOL;
 
 /**
  * A MappingException indicates that something is wrong with the mapping setup.
@@ -999,5 +1002,22 @@ EXCEPTION
             $value,
             $enumType
         ), 0, $previous);
+    }
+
+    /**
+     * @param LibXMLError[] $errors
+     */
+    public static function fromLibXmlErrors(array $errors): self
+    {
+        $formatter = static function (LibXMLError $error): string {
+            return sprintf(
+                'libxml error: %s in %s at line %d',
+                $error->message,
+                $error->file,
+                $error->line
+            );
+        };
+
+        return new self(implode(PHP_EOL, array_map($formatter, $errors)));
     }
 }
--- a/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
@@ -9,35 +9,34 @@ use Doctrine\ORM\Cache\Exception\CacheException;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataFactory;
 use Doctrine\ORM\Mapping\Driver\XmlDriver;
+use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Persistence\Mapping\RuntimeReflectionService;
 use Doctrine\Tests\Models\DDC117\DDC117Translation;
 use Doctrine\Tests\Models\DDC3293\DDC3293User;
 use Doctrine\Tests\Models\DDC3293\DDC3293UserPrefixed;
 use Doctrine\Tests\Models\DDC889\DDC889Class;
+use Doctrine\Tests\Models\DDC889\DDC889Entity;
+use Doctrine\Tests\Models\DDC889\DDC889SuperClass;
 use Doctrine\Tests\Models\Generic\SerializationModel;
 use Doctrine\Tests\Models\GH7141\GH7141Article;
 use Doctrine\Tests\Models\GH7316\GH7316Article;
 use Doctrine\Tests\Models\ValueObjects\Name;
 use Doctrine\Tests\Models\ValueObjects\Person;
-use DOMDocument;
 
-use function array_filter;
-use function array_map;
-use function assert;
-use function glob;
-use function in_array;
-use function is_array;
-use function pathinfo;
+use function substr_count;
 
 use const DIRECTORY_SEPARATOR;
-use const PATHINFO_FILENAME;
 
 class XmlMappingDriverTest extends AbstractMappingDriverTest
 {
     protected function loadDriver(): MappingDriver
     {
-        return new XmlDriver(__DIR__ . DIRECTORY_SEPARATOR . 'xml');
+        return new XmlDriver(
+            __DIR__ . DIRECTORY_SEPARATOR . 'xml',
+            XmlDriver::DEFAULT_FILE_EXTENSION,
+            true
+        );
     }
 
     public function testClassTableInheritanceDiscriminatorMap(): void
@@ -163,32 +162,89 @@ class XmlMappingDriverTest extends AbstractMappingDriverTest
      * @dataProvider dataValidSchema
      * @group DDC-2429
      */
-    public function testValidateXmlSchema(string $xmlMappingFile): void
-    {
-        $xsdSchemaFile = __DIR__ . '/../../../../../doctrine-mapping.xsd';
-        $dom           = new DOMDocument();
+    public function testValidateXmlSchema(
+        string $class,
+        string $tableName,
+        array $fieldNames,
+        array $associationNames
+    ): void {
+        $metadata = $this->createClassMetadata($class);
 
-        $dom->load($xmlMappingFile);
-
-        self::assertTrue($dom->schemaValidate($xsdSchemaFile));
+        $this->assertInstanceOf(ClassMetadata::class, $metadata);
+        $this->assertEquals($metadata->getTableName(), $tableName);
+        $this->assertEquals($metadata->getFieldNames(), $fieldNames);
+        $this->assertEquals($metadata->getAssociationNames(), $associationNames);
     }
 
     /**
-     * @psalm-return list<array{string}>
+     * @psalm-return []array{0: class-string, 1: string, 2: list<string>, 3: list<string>}
      */
     public static function dataValidSchema(): array
     {
-        $list    = glob(__DIR__ . '/xml/*.xml');
-        $invalid = ['Doctrine.Tests.Models.DDC889.DDC889Class.dcm'];
-        assert(is_array($list));
+        return [
+            [
+                User::class,
+                'cms_users',
+                ['name', 'email', 'version', 'id'],
+                ['address', 'phonenumbers', 'groups'],
+            ],
+            [
+                DDC889Entity::class,
+                'DDC889Entity',
+                [],
+                [],
+            ],
+            [
+                DDC889SuperClass::class,
+                'DDC889SuperClass',
+                ['name'],
+                [],
+            ],
+        ];
+    }
 
-        $list = array_filter($list, static function (string $item) use ($invalid): bool {
-            return ! in_array(pathinfo($item, PATHINFO_FILENAME), $invalid, true);
-        });
+    /**
+     * @param class-string                 $class
+     * @param non-empty-array<string, int> $expectedExceptionOccurrences
+     *
+     * @dataProvider dataInvalidSchema
+     */
+    public function testValidateIncorrectXmlSchema(string $class, array $expectedExceptionOccurrences): void
+    {
+        try {
+            $this->createClassMetadata($class);
 
-        return array_map(static function (string $item) {
-            return [$item];
-        }, $list);
+            $this->fail('XML schema validation should throw a MappingException');
+        } catch (MappingException $exception) {
+            foreach ($expectedExceptionOccurrences as $exceptionContent => $occurrencesCount) {
+                $this->assertEquals($occurrencesCount, substr_count($exception->getMessage(), $exceptionContent));
+            }
+        }
+    }
+
+    /**
+     * @return non-empty-list<array{0: class-string, 1: non-empty-array<string, int>}>
+     */
+    public static function dataInvalidSchema(): array
+    {
+        return [
+            [
+                DDC889Class::class,
+                ['This element is not expected' => 1],
+            ],
+            [
+                UserIncorrectAttributes::class,
+                [
+                    'attribute \'field\': The attribute \'field\' is not allowed' => 2,
+                    'The attribute \'name\' is required but missing' => 2,
+                    'attribute \'fieldName\': The attribute \'fieldName\' is not allowed' => 1,
+                ],
+            ],
+            [
+                UserMissingAttributes::class,
+                ['The attribute \'name\' is required but missing' => 1],
+            ],
+        ];
     }
 
     /**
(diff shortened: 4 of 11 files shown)


lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -820,4 +820,21 @@ public static function infiniteEmbeddableNesting($className, $propertyName)
             )
         );
     }
+
+    /**
+     * @param LibXMLError[] $errors
+     */
+    public static function fromLibXmlErrors(array $errors) : self
+    {
+        return new self(array_reduce(array_map(function (\LibXMLError $error) : string {
```
greg0ire: First time ever playing with `array_reduce`, this is fun! But I should probably use implode instead, right?

alcaeus: Yeah, that looks like it could perfectly be implemented with `implode(PHP_EOL, array_map(...);`.

greg0ire: Will change

greg0ire:
Regarding the exception issue, I found that the xml driver extends a class from another package (doctrine/common). [One of the parent methods throws an exception *from that packge* if the result from `loadMappingFile` is missing a `$className` key](https://github.com/doctrine/common/blob/a3e240fa07ec9748387c13534949d543c0e177fa/lib/Doctrine/Common/Persistence/Mapping/Driver/FileDriver.php#L115-L118).

Would you consider throwing an exception from this package a BC-break or not? Also, if we forget about this PR for a moment, wouldn't it be better to do this:

```
<?php
try {
   // method call from another package
} catch (ExceptionFromAnotherPackage $e) {
    throw new ExceptionFromThisPackage('message', 0, $e);
}
```

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -861,6 +862,21 @@ protected function loadMappingFile($file)
         return $result;
     }
 
+    private function assertValid(string $file): void
```
Majkl578: How about call it `validateMapping`?

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -861,6 +862,21 @@ protected function loadMappingFile($file)
...
+    {
+        $backedUpErrorSetting = libxml_use_internal_errors(true);
+        $document = new \DOMDocument();
+        $document->load($file);
```
Majkl578: This will cause duplicated IO, better pass already fetched XML string to this function and use `DOMDocument::loadXML()`.

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -861,6 +862,21 @@ protected function loadMappingFile($file)
...
+        if (!$document->schemaValidate(__DIR__ . '/../../../../../doctrine-mapping.xsd')) {
+            $errors = libxml_get_errors();
+            libxml_clear_errors();
+            libxml_use_internal_errors(false);
```
Majkl578: These two directives could be put into a more DRY `finally` block, and use `$backedUpErrorSetting`.

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -861,6 +862,21 @@ protected function loadMappingFile($file)
         return $result;
     }
 
+    private function assertValid(string $file): void
+    {
+        $backedUpErrorSetting = libxml_use_internal_errors(true);
+        $document = new \DOMDocument();
+        $document->load($file);
+        if (!$document->schemaValidate(__DIR__ . '/../../../../../doctrine-mapping.xsd')) {
```
Majkl578: Space around `!` is needed.

Majkl578 (changes requested):
I think this would be a nice addition to 3.0 and have it enabled by default.

_Note that this introduces a dependency on ext-dom._

Majkl578:
> Would you consider throwing an exception from this package a BC-break or not?

In 3.0: No.
In 2.x: No, because this behavior would not be enabled by default (if accepted for 2.x - not sure about it).

---

Please add some tests to illustrate how the error propagation from libxml looks like.
It'd also be interesting to know how many (if any) tests fail with this enabled.

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
greg0ire: Could be less misleading since I don't use actual assertions here

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
greg0ire: Yeah but it gives worse error messages

greg0ire:
Right now it's not optional so... congrats, no failures!

greg0ire:
> Right now it's not optional so... congrats, no failures!

Uuuuh what am I saying, there is one failure that was not intended (I think):

```
1) Doctrine\Tests\ORM\Tools\Export\XmlClassMetadataExporterTest::testExportDirectoryAndFilesAreCreated
Doctrine\ORM\Mapping\MappingException: libxml error: Element '{http://doctrine-project.org/schemas/orm/doctrine-mapping}join-column', attribute 'on-update': The attribute 'on-update' is not allowed.
/home/greg/dev/doctrine2/tests/Doctrine/Tests/ORM/Tools/Export/xml/Doctrine.Tests.ORM.Tools.Export.User.dcm.xml:35
libxml error: Element '{http://doctrine-project.org/schemas/orm/doctrine-mapping}cascade-persist': This element is not expected. Expected is one of ( {http://doctrine-project.org/schemas/orm/doctrine-mapping}cascade-detach, ##other{http://doctrine-project.org/schemas/orm/doctrine-mapping}* ).
/home/greg/dev/doctrine2/tests/Doctrine/Tests/ORM/Tools/Export/xml/Doctrine.Tests.ORM.Tools.Export.User.dcm.xml:59

/home/greg/dev/doctrine2/lib/Doctrine/ORM/Mapping/MappingException.php:829
/home/greg/dev/doctrine2/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php:873
/home/greg/dev/doctrine2/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php:840
/home/greg/dev/doctrine2/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/Driver/FileDriver.php:115
/home/greg/dev/doctrine2/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php:59
/home/greg/dev/doctrine2/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:151
/home/greg/dev/doctrine2/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:333
/home/greg/dev/doctrine2/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:78
/home/greg/dev/doctrine2/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:226
/home/greg/dev/doctrine2/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:115
/home/greg/dev/doctrine2/tests/Doctrine/Tests/ORM/Tools/Export/AbstractClassMetadataExporterTest.php:91
```

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
greg0ire: See https://github.com/doctrine/doctrine2/pull/6728#issuecomment-333406605

greg0ire:
> Note that this introduces a dependency on ext-dom.

Made that explicit

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
greg0ire: I lost my train of thought it seems

composer.json
```diff
@@ -15,6 +15,7 @@
     "minimum-stability": "dev",
     "require": {
         "php": "^7.1",
+        "ext-dom": "*",
```
SenseException: I have a question: Isn't this extension optional? The usage of XML in Doctrine is optional too for a user, isn't it?

alcaeus: Technically, it's an optional dependency. However, making it an explicit dependency is the defensive way here: we prevent the code from blowing up in the users face with a fatal error if they use XML mappings.

Remember that other bundles (e.g. the FOSUserBundle) specify XML mappings which are automatically imported, so the user won't know that he's using XML mappings and thus should add the `ext-dom` dependency.

One more thought: moving this to `suggest` would require all bundles that specify XML mappings to add a `require` entry for `ext-dom`, which would be a BC break towards the bundles. IMO it's safer to add it as an explicit dependency.

SenseException: Talking about "Bundles" sounds more like Symfony instead of Doctrine. This repository is independent from frameworks. Should in this case the `require` be added in DoctrineBundle? What are "all bundles"?

alcaeus: @SenseException replace with "bundles" with "anything that provides XML mappings". Suggesting `ext-dom` instead of requiring it will break third party software if they already use XML mappings. That's a BC break that would move this feature back to 3.0. The way it is now it's possible to add this to the 2.x line.

greg0ire: @alcaeus I rebased on 3.x yesterday :sweat_smile:

alcaeus: Oh, didn't see that, my bad. IMO it's better to have the explicit dependency anyways. 👍

Majkl578:
Needs a rebase.

composer.json
Majkl578: I have mixed feelings about this. One one hand, it's true that user would need explicit ext-dom dependency. On the other one, we should check at runtime for DOM presence, it just feels wrong to force people with annotations install DOM. (Well, we can say the same about doctrine/annotations, but that's slightly different story).
Since this PR is now 3.0 only, it should be fine fine to do either way, but question is which one is better.

Ocramius: > On the other one, we should check at runtime for DOM presence, it just feels wrong to force people with annotations install DOM.

https://webmozart.io/blog/2014/10/24/defining-php-annotations-in-xml/

greg0ire:
>  establish what is the right kind of exception to throw (there is a test that asserts a mapping exception from doctrine/common should be thrown, I don't know why yet)

#7199 should solve this

lcobucci (changes requested):
@greg0ire @Majkl578 I've rebased this branch and adjusted the files that were actually invalid according to our XSD. The only broken test we have now are about the validation of an invalid XML document (which was being tested in a different way).

About the design of the solution, this implementation describes EXACTLY the point I was trying to make regarding deprecations. IMO we **should not** modify the existing class but rather create a new one which validates the XML files and simply add the `@deprecated` annotation on the existing one. This new class can be a simple decorator for now.

We should also consider that the current approach breaks [OCP](https://web.archive.org/web/20060822033314/http://www.objectmentor.com/resources/articles/ocp.pdf)...

lcobucci:
BTW I do believe that this would fit nicely for v2.7 and I don't really see a reason for blocking this until we decide about #7199

greg0ire:
> simply add the @deprecated annotation on the existing one. 
> This new class can be a simple decorator for now.

This would create a situation where you have an unfixable deprecation.

> BTW I do believe that this would fit nicely for v2.7 and I don't really see a reason for blocking this until we decide about #7199

I use `throw MappingException` from ORM, but it does not extend `MappingException` from Common, and should.

lcobucci:
> This would create a situation where you have an unfixable deprecation.

Nope, the migration path is to use the new class. Users would only be notified about the deprecation when using directly the old class. Nobody should be bothered if we're using internally our own deprecated items (this already happens with `ClassMetadata`		and `ClassMetadataInfo` btw).

greg0ire:
> Nope, the migration path is to use the new class. Users would only be notified about the deprecation when using directly the old class. Nobody should be bothered if we're using internally our own deprecated items (this already happens with ClassMetadata	and ClassMetadataInfo btw).

Somewhat agree, but are there any tools that are capable of detecting calls like you describe and only act on direct calls? phpstan-deprecation-rules, maybe?

greg0ire:
> (this already happens with ClassMetadata	and ClassMetadataInfo btw

Is this why https://github.com/doctrine/doctrine2/pull/6886 never got merged?

lcobucci:
> Somewhat agree, but are there any tools that are capable of detecting calls like you describe and only act on direct calls? phpstan-deprecation-rules, maybe?

That's correct, that PHPStan extension behaves like that indeed. But please keep in mind that my main concern is about our design principles and values... IMO we can always adapt tools but sticking to what we believe is the most appropriated is usually harder - specially on these days we live where we're always eager to push anything to production (definitely not talking about the code here, just a rant of an old grumpy developer HAHAHA).

Anyway, sorry if this was way philosophical :smile:

Kern046:
Hello there, I volunteered in #9637 to help in this PR which seems to have mobilized some energy for some time now ! As a first contribution it also seems quite tough viewing the discussions, so I will rely on your advice to make it progress properly !

My first question for now is about the target branch, I saw that it was targeting 2.10.x a few months ago, should it target 2.11.x now ? I also noticed [here](https://github.com/doctrine/orm/pull/6728#discussion_r142091148) that the branch was rebased with 3.x, but if I rebase it now shouldn't I use 2.11.x instead ?

Thanks a lot :)

greg0ire:
Hi! Thanks for volunteering! I wasn't sure if you wanted to do it so I did some cleanup. I retargeted to 2.12.x, which is the next minor branch. You can start from there :)

greg0ire:
Apparently I broke the tests in `tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php`, it seems that the mapping files (one of which is now invalid) were used for more than one test. And maybe what I put in the message of `markTestSkipped` is wrong.

Kern046:
I did a PR to your branch with the following elements :

* Restoration of the removed DDC889 XML mapping files which made a test fail
* Implementation of new tests for incorrect cases (invalid and missing attributes)

Though I have several concerns about this PR :

* Currently we have to retrieve the XSD file through `www.doctrine-project.org`, so I wonder : what if someday this website is unavailable for some reason or encounters heavy latencies ? Won't this have a strong impact on everyone implementing XML mapping ?
* In the PR tests we seem unable to reach `XmlDriver::validateMapping` to ensure the whole workflow works as intended. The tests use directly `DOMDocument` and doesn't produce the same error output as the driver which uses `libxml_errors`. Due to that, I think that the current tests do not succeed in their objective of protecting `XmlDriver::validateMapping` against regressions or misbehavior.

WDYT ?

greg0ire:
> Currently we have to retrieve the XSD file through www.doctrine-project.org, so I wonder : what if someday this website is unavailable for some reason or encounters heavy latencies ? Won't this have a strong impact on everyone implementing XML mapping ?

Where/when does that happen exactly? Are you sure it happens? Aren't we avoiding that with this line: https://github.com/doctrine/orm/pull/6728/files#diff-3613bafacbade59e06613442d37c5f0debfed82f00f548143197ed7caf98a635R978

> In the PR tests we seem unable to reach XmlDriver::validateMapping to ensure the whole workflow works as intended. The tests use directly DOMDocument and doesn't produce the same error output as the driver which uses libxml_errors. Due to that, I think that the current tests do not succeed in their objective of protecting XmlDriver::validateMapping against regressions or misbehavior.

:thinking: don't we reach it through `testInvalidEntityOrMappedSuperClassShouldMentionParentClasses`? If yes, let's do something similar to that?

Kern046:
> Where/when does that happen exactly? Are you sure it happens? Aren't we avoiding that with this line: https://github.com/doctrine/orm/pull/6728/files#diff-3613bafacbade59e06613442d37c5f0debfed82f00f548143197ed7caf98a635R978

Oh yeah absolutely, I assumed that we had to read the XML mapping file XSD location and then access it remotely. Problem solved :smile: !

> thinking don't we reach it through `testInvalidEntityOrMappedSuperClassShouldMentionParentClasses`? If yes, let's do something similar to that?

Very nice ! I adapted the tests, will push it tomorrow once it's all green !

Kern046:
I submitted the tests improvement to your repo branch. Let me know what you think about it :)

greg0ire:
You can open your own PR to `doctrine/orm` if you want to iterate faster. Anyway, this is starting to look good, and we should think how we could make this optional and disabled by default. I guess we should probably add a new argument to `ORMSetup::createXMLMetadataConfiguration`, set to `false` by default,  deprecate not setting it to `true`, and remove it in 3.0?

Kern046:
> You can open your own PR to `doctrine/orm` if you want to iterate faster. Anyway, this is starting to look good, and we should think how we could make this optional and disabled by default. I guess we should probably add a new argument to `ORMSetup::createXMLMetadataConfiguration`, set to `false` by default, deprecate not setting it to `true`, and remove it in 3.0?

If you don't mind I think it's better to keep the current PR to keep its history ? I may be wrong on this, this is my first contribution to Doctrine so if you think it's better to open my own PR I will :).

Okay for the optional design, I'll give it a try asap.

greg0ire:
> If you don't mind I think it's better to keep the current PR to keep its history ? I may be wrong on this, this is my first contribution to Doctrine so if you think it's better to open my own PR I will :).

If you don't mind making PRs to my repo, I'm fine with it, this was a suggestion for your convenience. You somehow manage to get green tests everytime so it should be fine :)

composer.json
```diff
@@ -21,6 +21,7 @@
     },
     "require": {
         "php": "^7.1 || ^8.0",
+        "ext-dom": "*",
```
greg0ire: @Kern046 this needs to be moved to `suggest`, and then if it's not loaded, validation should be skipped.

Kern046: Should it also trigger a deprecation or not ? If not, that means that people could have XSD validation enabled with the boolean, but ineffective due to the absence of the DOM extension.

greg0ire: Gave if some thought, and I think it should, and then we throw in 3.0, otherwise people are going to report issues that could have been prevented by installing that extension. In fact, I think you can even throw an exception rightaway: it's not like people will get the exception by default.

SenseException (dismissed):

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -962,6 +990,27 @@ protected function loadMappingFile($file)
...
+    private function validateMapping(string $file): void
+    {
+        if (! $this->isXsdValidationEnabled) {
+            return;
+        }
+
+        $backedUpErrorSetting = libxml_use_internal_errors(true);
+
+        $document = new DOMDocument();
+        $document->load($file);
```
derrabus: These lines should be moved into the `try` block, just in case.

lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -1000,4 +1003,21 @@ public static function invalidEnumValue(
...
     }
+
+    /**
+     * @param LibXMLError[] $errors
+     */
+    public static function fromLibXmlErrors(array $errors): self
+    {
+        $formatter = static function (LibXMLError $error): string {
+            return sprintf(
+                'libxml error: %s%s:%d',
```
derrabus: Don't we need some kind of separator between message and file?

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -163,32 +163,88 @@ public function testInvalidMappingFileException(): void
...
+     *
+     * @dataProvider dataInvalidSchema
+     */
+    public function testValidateIncorrectXmlSchema(string $class, array $expectedExceptionOccurrences): void
+    {
+        try {
+            $this->createClassMetadata($class);
+        } catch (Throwable $throwable) {
+            $this->assertInstanceOf(MappingException::class, $throwable);
+
```
derrabus:
```suggestion
        } catch (MappingException $throwable) {
```

Additionally, we need to make sure the test fails if no exception is raised at all.

derrabus (changes requested):

beberlei (dismissed):
An unrelated follow up could be to get rid of SimpleXML in the driver and use only ext-dom

derrabus (dismissed):

greg0ire:
@lcobucci regarding the OCP, I read the link you gave and it says (rightly IMO) that the OCP is something that should be taken into account when designing a class. `XMLDriver` does not seem closed against adding validation functionality, as there is no way to inject a validator in it. Also, I agree that decoration is a great way to add functionality, and use it a lot in my projects, but here the method we would want to decorate is `loadMappingFile()`, which is protected. That means we cannot use decoration, but would be forced to use inheritance instead to "decorate" the method. Even if it was public and we could do proper decoration, the interface of `XMLDriver` is not exactly small, which means we would have to proxify a lot of methods to achieve that.

I think a sensible way to violate the OCP as little as possible could be to do the following:

```php
interface XMLValidator
{
    public function __invoke(string $file);
}

class DomXMLValidator implements XMLValidator
{
    // … implementation goes here
}

class XMLDriver
{
    // inject validator in constructor and call it
}
```

But given the size of the added piece of code, and given how unlikely I think it is people would want to inject their own implementation of the XMLValidator, I'm not sure it's worth it.

Do you still have concerns about runtime deprecations, now that we have switched to https://github.com/doctrine/deprecations/ ? That lib wasn't discussed at all in https://github.com/doctrine/doctrine-website/issues/200

lcobucci:
> @lcobucci regarding the OCP, I read the link you gave and it says (rightly IMO) that the OCP is something that should be taken into account when designing a class. `XMLDriver` does not seem closed against adding validation functionality, as there is no way to inject a validator in it. (...)

The great thing about "principles" that version of me from 4 years ago hadn't learned well is that they aren't absolute rules :laughing:. I (now) know that understanding the context and allowing some flexibility is important - even though I tend to stick to them as much as possible.

If I was making this feature, I'd still aim to add a new component instead of modifying the existing one to conditionally change its behaviour based on a boolean flag. I believe that adding a new (`final`) class (`SchemaEnforcingXmlDriver`, whateva) that - initially - inherits the `XmlDriver` makes things more explicit in the long run and gives more flexibility on refactoring than the boolean flag approach :smile:

With that said, I understand your arguments and agree that a completely different design would make things much better buuuuut :rofl:


> Do you still have concerns about runtime deprecations, now that we have switched to https://github.com/doctrine/deprecations/ ? That lib wasn't discussed at all in [doctrine/doctrine-website#200](https://github.com/doctrine/doctrine-website/issues/200)

Doctrine has fully embraced that deprecations library so I believe that's a moot point now :+1:

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -163,32 +162,88 @@ public function testInvalidMappingFileException(): void
...
+                [],
+            ],
+        ];
+    }
 
-        $list = array_filter($list, static function (string $item) use ($invalid): bool {
-            return ! in_array(pathinfo($item, PATHINFO_FILENAME), $invalid, true);
-        });
+    /**
+     * @param array{0: class-string, 1: array<string, int>} $expectedExceptionOccurrences
```
lcobucci: This doesn't seem correct =)

https://phpstan.org/r/7db17906-8a40-4f4e-a45d-16ec59ef3f8d on the `non-empty-array` :+1: 

```suggestion
     * @param class-string $class
     * @param non-empty-array<string, int> $expectedExceptionOccurrences
```

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -163,32 +162,88 @@ public function testInvalidMappingFileException(): void
...
+                $this->assertEquals($occurrencesCount, substr_count($exception->getMessage(), $exceptionContent));
+            }
+        }
+    }
 
-        return array_map(static function (string $item) {
-            return [$item];
-        }, $list);
+    /**
+     * @psalm-return []array{0: class-string, 1: array<string, int>}
```
lcobucci: This doesn't look correct (https://phpstan.org/r/3986d513-0d9f-4591-9d33-6a9fae6ec4b8 and https://psalm.dev/r/1d25f7f2f6).

I'd also just use `@return`, IDEs are much better nowadays...

```suggestion
     * @return non-empty-list<array{0: class-string, 1: non-empty-array<string, int>}>
```

lcobucci (dismissed):
Just some minor improvements on the SA side of things =)

greg0ire:
Thanks for reviving this @Kern046 ! And thanks for your help @lcobucci ! 5 years after the start, let's finally merge this :)

PR #6732, opened 2017-09-27
Update BasicEntityPersister::count()

afurculita:
`$this->entityPersister->count($this->criteria)` will return a string, because it returns the result of a count query without casting.


Ocramius:
@afurculita please provide a test case. See https://github.com/doctrine/doctrine2/pull/6550 for an example.

Also, please target `master`: backporting is up to us.

afurculita:
@Ocramius actually the problem is from `BasicEntityPersister::count()` and it is only in the versions <=2.5. It has been fixed in master. No tests have been added in master for this change. Should I still add them in 2.5?

Ocramius:
> Should I still add them in 2.5?

Yes please: I do not merge stuff without tests. If a change made it through to `master` without a test, I will forward port the one you write there.

Majkl578:
This was fixed by  #6818 in 2.5.

PR #6734, opened 2017-09-28
Add DateInterval type parameter

ossinkine:
This PR adds the ability to pass a DateInterval object as a query parameter

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "php": "^7.1",
         "ext-pdo": "*",
         "doctrine/collections": "^1.4",
-        "doctrine/dbal": ">=2.5-dev,<2.7-dev",
+        "doctrine/dbal": "^2.6",
         "doctrine/instantiator": "~1.1",
         "doctrine/common": "^2.7.1",
         "doctrine/cache": "~1.6",
--- a/tests/Doctrine/Tests/ORM/Query/ParameterTypeInfererTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/ParameterTypeInfererTest.php
@@ -13,17 +13,18 @@ class ParameterTypeInfererTest extends OrmTestCase
 
     public function providerParameterTypeInferer()
     {
-         $data = [
+        $data = [
             [1,                 Type::INTEGER],
             ["bar",             PDO::PARAM_STR],
             ["1",               PDO::PARAM_STR],
             [new \DateTime,     Type::DATETIME],
+            [new \DateInterval('P1D'), Type::DATEINTERVAL],
             [[2],          Connection::PARAM_INT_ARRAY],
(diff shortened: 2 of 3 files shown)


Ocramius:
Looks good to me.

We need to plan something smarter for ORM v3 (maybe a replaceable infer mechanism?

ossinkine:
Tests with the lowest dependencies didn't pass: https://travis-ci.org/doctrine/doctrine2/jobs/280750301
Should I increase the version in composer.json or add the version check in the class?

Ocramius:
Bumping dependency version for 2.6.x is quite OK

lcobucci (approved):
LGTM 👍 

Thanks for the CS fixes as well 😄

tests/Doctrine/Tests/ORM/Query/ParameterTypeInfererTest.php
```diff
@@ -13,17 +13,18 @@ class ParameterTypeInfererTest extends OrmTestCase
...
+            [new \DateInterval('P1D'), Type::DATEINTERVAL],
             [[2],          Connection::PARAM_INT_ARRAY],
             [["foo"],      Connection::PARAM_STR_ARRAY],
             [["1","2"],    Connection::PARAM_STR_ARRAY],
             [[],           Connection::PARAM_STR_ARRAY],
             [true,              Type::BOOLEAN],
-         ];
+        ];
 
         if (PHP_VERSION_ID >= 50500) {
```
lcobucci: Unrelated, but we can simplify this since we require PHP 7.1+

lcobucci:
@ossinkine 🚢

composer.json
```diff
@@ -17,7 +17,7 @@
         "php": "^7.1",
         "ext-pdo": "*",
         "doctrine/collections": "^1.4",
-        "doctrine/dbal": ">=2.5-dev,<2.7-dev",
+        "doctrine/dbal": "^2.6",
```
Majkl578: Did we allow >=2.7-dev intentionally here or is it just accidental?

ossinkine: I set it up intentionally for look at tests and they were passed, so I think it's ok.

lcobucci: I think it was accidental =) It should indeed be `>=2.6-dev,<2.7-dev` or even `>=2.6-dev,<2.8-dev`

alcaeus: With how carefully we're checking BC, does it even matter? Might as well use the semver operator here.

lcobucci: That's true... @Majkl578 @Ocramius anything against that?

Ocramius:
Nah, fine as-is

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

PR #6735, opened 2017-09-28
parameter type guessing fails for identifier with custom type

maglnet:
Hi,

regarding #6443 this PR currenlty only adds a failing test case to discuss how one could improve type inferring.

The main part where things go wrong is here:
https://github.com/doctrine/doctrine2/blob/53245e8a73cb16d5e9ba35ffa5bbb15cc417bb9f/lib/Doctrine/ORM/Query.php#L406-L409

The reason is that `ParameterTypeInferer::inferType($value)` does not handle metadata and other information to detect the type.

I suggest to extend `ParameterTypeInferer` to take the class metadata into account, if it is available for the given parameter.

Are there any hints on this?
I'd try to create a first draft within the next days.

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -386,7 +386,7 @@ final class Query extends AbstractQuery
             if (is_object($value) && $this->em->getMetadataFactory()->hasMetadataFor(StaticClassNameConverter::getClass($value))) {
                 $metadata = $this->em->getClassMetadata(StaticClassNameConverter::getClass($value));
                 if ($metadata->isIdentifierComposite()) {
-                    ORMInvalidArgumentException::invalidCompositeIdentifier();
+                    throw ORMInvalidArgumentException::invalidCompositeIdentifier();
                 }
                 $type = $metadata->getColumn($metadata->getIdentifier()[0])->getTypeName();
             }
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -385,7 +385,7 @@ final class Query extends AbstractQuery
 
             if (is_object($value) && $this->em->getMetadataFactory()->hasMetadataFor(StaticClassNameConverter::getClass($value))) {
                 $metadata = $this->em->getClassMetadata(StaticClassNameConverter::getClass($value));
-                $type = $metadata->getColumn($metadata->getIdentifier()[0])->getTypeName();
+                $type     = $metadata->getColumn($metadata->getIdentifier()[0])->getTypeName();
             }
 
             $value = $this->processParameterValue($value);
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -19,6 +19,7 @@ use Doctrine\ORM\Query\Parser;
 use Doctrine\ORM\Query\ParserResult;
 use Doctrine\ORM\Query\QueryException;
 use Doctrine\ORM\Utility\HierarchyDiscriminatorResolver;
+use ProxyManager\Proxy\GhostObjectInterface;
 use function array_keys;
 use function array_values;
 use function count;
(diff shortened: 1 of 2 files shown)


Majkl578:
Auto detection / inference is very problematic issue. There are multiple edge cases that have to be taken into consideration, otherwise the queries could be corrupted.
Some of these cases are:
* parameter reuse (i.e. `WHERE e.id = :foo OR e.name = :foo`)
* parameter usage unbound to a field (i.e. `WHERE :foo < 42`)

Maybe exception for ambiguous types should be thrown instead, since it can cause very nasty and hard-to-debug bugs, for example DATETIME vs. DATETIMETZ.

maglnet:
@Majkl578 well, I probably referenced an issue that isn't excatly facing the problem we had.

Our main problem is, that an entity with a custom type (UUID in our case) is passed in as a parameter (as in the test). 
After processing the vaule through `$this->processParameterValue($value);` it is becoming the entity's identifier value but without correctly setting the type of the identifier.

It seems that most database-drivers (e.g. mysql, sqlite) automagically string-cast this value before passing it to the database, so one is not directly facing the problem on common oss db systems. But mssql raises an exception if there is an object passed to the driver.
The first try to fix this was casting the value, but was rejected in doctrine/dbal#2788 and @Ocramius pointed out that the type needs to be correctly set or infereed to handle this.

The current implementation only handles exactly this case and does not handle type-guessing depending on the given objects. So the mentioned converting of e.g. `DATETIME` / `DATETIMETZ` will not happen.

lib/Doctrine/ORM/Query.php
```diff
@@ -403,10 +406,17 @@ private function processParameterMappings($paramMappings)
                 $value = array_keys(HierarchyDiscriminatorResolver::resolveDiscriminatorsForClass($value, $this->em));
             }
 
+            if (is_object($value) && $this->em->getMetadataFactory()->hasMetadataFor(get_class($value))) {
```
Ocramius: Can the block deciding `$type` and `$value` be isolated into a `private` method?

Ocramius (changes requested):

lib/Doctrine/ORM/Query.php
Majkl578: `$value` may also be a proxy, then this code will not work

lib/Doctrine/ORM/Query.php
```diff
@@ -403,10 +406,17 @@ private function processParameterMappings($paramMappings)
...
+                $metadata = $this->em->getClassMetadata(get_class($value));
+                $type     = $metadata->getColumn($metadata->getIdentifier()[0])->getTypeName();
```
Majkl578: What happens when the `$value` is a composite-identified entity? Should we throw?

Majkl578 (changes requested):

lib/Doctrine/ORM/Query.php
maglnet: > 
> 
> Can the block deciding `$type` and `$value` be isolated into a `private` method?

I'll work on this later, for now "it works"... :-)

lcobucci:
@maglnet could you please check if this got fixed in the last patch release?

ChrisSchwerdt:
We recently began work on a project that is using `ramsey/uuid-doctrine` to support a `UuidBinaryOrderedTimeType` for entity identifiers on Doctrine 2.7, and we're running into this same issue. Are the maintainers open to accepting an updated PR to fix this issue?

chapterjason:
@maglnet It is still an issue if you want to use ulid as identifier and just set the parameter value to an entity.

https://github.com/doctrine/orm/blob/81d472f6f96b8b571cafefe8d2fef89ed9446a62/lib/Doctrine/ORM/Query.php#L455-L462

In the case of using ulid, the `$processedValue` will be an ulid instance.

```
object(Symfony\Component\Uid\Ulid)#1380 (1) {
  ["uid":protected]=>
  string(26) "01FJN4099H28M1HYC0YRH3CYEP"
}
```

`$parameter->getType()` will be null and the `ParameterTypeInferer::inferType` will return `ParameterType::STRING` which is wrong.

Would be great to have something which can inter custom types.

This version based on you code, works for me:

https://github.com/chapterjason/orm/blob/f5f7c7dc9f9ad932e62616ce491fe0812096bf4a/lib/Doctrine/ORM/Query.php#L456-L475

Workaround for Ulid:

```php
class Article
{
    #[ORM\Id]
    #[ORM\Column(type: 'uuid')]
    private Uuid $id;

    public function __construct(Ulid $id)
    {
        $this->id = Uuid::fromRfc4122($id->toRfc4122());
    }

    public function getId(): Ulid
    {
        return Ulid::fromRfc4122($this->id->toRfc4122());
    }
}
```

PR #6742, opened 2017-10-02
Add support second, minute, week, year on DATE_ADD and DATE_SUB

SebLevDev:
The DATE_ADD and DATE_SUB built-in function of Doctrine supports only the "hour", "month" and "day" intervals. I don't see the point of not having all the intervals available. These should be "minute", "second" and "year". I didn't check compatibility with other database vendors rather than MySQL but I don't think it should be a problem

Cfr Also Ticket https://github.com/doctrine/doctrine2/issues/3698

### DIFF
--- a/lib/Doctrine/ORM/Query/AST/Functions/DateAddFunction.php
+++ b/lib/Doctrine/ORM/Query/AST/Functions/DateAddFunction.php
@@ -51,7 +51,11 @@ class DateAddFunction extends FunctionNode
                     $this->firstDateExpression->dispatch($sqlWalker),
                     $this->intervalExpression->dispatch($sqlWalker)
                 );
-
+            case 'minute':
+                return $sqlWalker->getConnection()->getDatabasePlatform()->getDateAddMinutesExpression(
+                    $this->firstDateExpression->dispatch($sqlWalker),
+                    $this->intervalExpression->dispatch($sqlWalker)
+                );
             case 'hour':
                 return $sqlWalker->getConnection()->getDatabasePlatform()->getDateAddHourExpression(
                     $this->firstDateExpression->dispatch($sqlWalker),
@@ -62,16 +66,25 @@ class DateAddFunction extends FunctionNode
                     $this->firstDateExpression->dispatch($sqlWalker),
                     $this->intervalExpression->dispatch($sqlWalker)
                 );
-
+            case 'week':
+                return $sqlWalker->getConnection()->getDatabasePlatform()->getDateAddWeeksExpression(
+                    $this->firstDateExpression->dispatch($sqlWalker),
+                    $this->intervalExpression->dispatch($sqlWalker)
+                );
             case 'month':
                 return $sqlWalker->getConnection()->getDatabasePlatform()->getDateAddMonthExpression(
+                    $this->firstDateExpression->dispatch($sqlWalker),
+                    $this->intervalExpression->dispatch($sqlWalker)
+                );
+            case 'year':
+                return $sqlWalker->getConnection()->getDatabasePlatform()->getDateAddYearsExpression(
                     $this->firstDateExpression->dispatch($sqlWalker),
                     $this->intervalExpression->dispatch($sqlWalker)
                 );
 
             default:
                 throw QueryException::semanticalError(
-                    'DATE_ADD() only supports units of type second, hour, day and month.'
+                    'DATE_ADD() only supports units of type second, minute, hour, day, week, month and year.'
                 );
         }
     }
--- a/tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Tests\ORM\Functional;
 
+use Doctrine\ORM\AbstractQuery;
 use Doctrine\Tests\Models\Company\CompanyManager;
 use Doctrine\Tests\OrmFunctionalTestCase;
 
@@ -176,7 +177,7 @@ class QueryDqlFunctionTest extends OrmFunctionalTestCase
         $this->assertEquals('Gui', $result[1]['str1']);
         $this->assertEquals('Jon', $result[2]['str1']);
         $this->assertEquals('Rom', $result[3]['str1']);
-        
+
         $this->assertEquals('amin E.', $result[0]['str2']);
         $this->assertEquals('herme B.', $result[1]['str2']);
         $this->assertEquals('than W.', $result[2]['str2']);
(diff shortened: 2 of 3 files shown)


lib/Doctrine/ORM/Query/AST/Functions/DateAddFunction.php
```diff
@@ -51,7 +51,11 @@ public function getSql(SqlWalker $sqlWalker)
                     $this->firstDateExpression->dispatch($sqlWalker),
                     $this->intervalExpression->dispatch($sqlWalker)
                 );
-
+			case 'minute':
```
Ocramius: I spotted some `\t` here ;-)

Ocramius (changes requested):
Overall, patch looks good, but misses testing (especially integration with other DBs

Legenyes:
Can i do something again ?

Ocramius:
Looks good to me: so far.

Assigning to @lcobucci for another pair of eyes.

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
```diff
@@ -301,6 +301,24 @@ public function testDateAdd()
         $this->assertTrue(strtotime($arg[0]['add']) > 0);
     }
 
+    /**
+     * @group DDC-2938
+     */
+    public function testDateAddTime()
```
lcobucci: Please add ` : void`

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
```diff
@@ -301,6 +301,24 @@ public function testDateAdd()
...
+    {
+        $arg = $this->_em->createQuery("SELECT DATE_ADD(CURRENT_TIMESTAMP(), 10, 'second') AS add FROM Doctrine\Tests\Models\Company\CompanyManager m")
+                ->getArrayResult();
+        $this->assertTrue(strtotime($arg[0]['add']) > 0);
```
lcobucci: Use `self::assert*()` instead

lcobucci: Wouldn't be better to use a more specific assertion?

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
```diff
@@ -301,6 +301,24 @@ public function testDateAdd()
...
+
+        $arg = $this->_em->createQuery("SELECT DATE_ADD(CURRENT_TIMESTAMP(), 10, 'minute') AS add FROM Doctrine\Tests\Models\Company\CompanyManager m")
+                ->getArrayResult();
+        $this->assertTrue(strtotime($arg[0]['add']) > 0);
+
+        $arg = $this->_em->createQuery("SELECT DATE_ADD(CURRENT_TIMESTAMP(), 10, 'hour') AS add FROM Doctrine\Tests\Models\Company\CompanyManager m")
+                ->getArrayResult();
+        $this->assertTrue(strtotime($arg[0]['add']) > 0);
```
lcobucci: Use `self::assert*()` instead

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
```diff
@@ -301,6 +301,24 @@ public function testDateAdd()
...
+     */
+    public function testDateAddTime()
+    {
+        $arg = $this->_em->createQuery("SELECT DATE_ADD(CURRENT_TIMESTAMP(), 10, 'second') AS add FROM Doctrine\Tests\Models\Company\CompanyManager m")
+                ->getArrayResult();
+        $this->assertTrue(strtotime($arg[0]['add']) > 0);
+
+        $arg = $this->_em->createQuery("SELECT DATE_ADD(CURRENT_TIMESTAMP(), 10, 'minute') AS add FROM Doctrine\Tests\Models\Company\CompanyManager m")
+                ->getArrayResult();
+        $this->assertTrue(strtotime($arg[0]['add']) > 0);
```
lcobucci: Use `self::assert*()` instead

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
```diff
@@ -335,6 +353,24 @@ public function testDateSub()
     }
 
     /**
+     * @group DDC-2938
+     */
+    public function testDateSubTime()
+    {
+        $arg = $this->_em->createQuery("SELECT DATE_SUB(CURRENT_TIMESTAMP(), 10, 'second') AS add FROM Doctrine\Tests\Models\Company\CompanyManager m")
+                ->getArrayResult();
+        $this->assertTrue(strtotime($arg[0]['add']) > 0);
```
lcobucci: Use `self::assert*()` instead

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
```diff
@@ -335,6 +353,24 @@ public function testDateSub()
...
+
+        $arg = $this->_em->createQuery("SELECT DATE_SUB(CURRENT_TIMESTAMP(), 10, 'minute') AS add FROM Doctrine\Tests\Models\Company\CompanyManager m")
+                ->getArrayResult();
+        $this->assertTrue(strtotime($arg[0]['add']) > 0);
```
lcobucci: Use `self::assert*()` instead

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
```diff
@@ -335,6 +353,24 @@ public function testDateSub()
...
+
+        $arg = $this->_em->createQuery("SELECT DATE_SUB(CURRENT_TIMESTAMP(), 10, 'hour') AS add FROM Doctrine\Tests\Models\Company\CompanyManager m")
+                ->getArrayResult();
+        $this->assertTrue(strtotime($arg[0]['add']) > 0);
```
lcobucci: Use `self::assert*()` instead

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
lcobucci: Wouldn't be better to use a more specific assertion?

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
lcobucci: Wouldn't be better to use a more specific assertion?

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
lcobucci: Wouldn't be better to use a more specific assertion?

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
lcobucci: Wouldn't be better to use a more specific assertion?

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
lcobucci: Wouldn't be better to use a more specific assertion?

tests/Doctrine/Tests/ORM/Functional/QueryDqlFunctionTest.php
```diff
@@ -335,6 +353,24 @@ public function testDateSub()
     }
 
     /**
+     * @group DDC-2938
+     */
+    public function testDateSubTime()
```
lcobucci: Please add ` : void`

lcobucci (changes requested):
It LGTM, just some nitpicking. Also, aren't we missing tests for year and week add/sub?

Legenyes:
@lcobucci I cannot change for` self::assert*()` and `: void `. 
These changes must be the subject of another PR.

lcobucci:
@Legenyes you can change for the code you're inserting.

Legenyes:
It's worse. 
Mixing two syntax in the same file is not a good practice.
I will not make this change in this PR

What about BC for `: void` ?

Ocramius:
This patch would land in 2.6 only, which is 7.1 only anyway

lcobucci:
@Legenyes `self::assert*()` has already been fixed for v3 on the entire codebase, this means that new code should adhere to that, otherwise we need to reapply things over and over again. In the other hand, if we apply the change on v2 as well we'll be creating conflicts.

I agree that it's ideal to have mixed syntaxes but the least worse solution given this context.

Legenyes:
What' news ?

lcobucci:
@Legenyes we still don't have tests for `week` and `year` and we still have the same `self::assertTrue()` checks that IMO are not really helpful since it doesn't provide you enough info about the operation that has been done.

greg0ire:
[The docs](http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html) currently say:

> DATE_ADD(date, days, unit) - Add the number of days to a given date. (Supported units are DAY, MONTH)

please change that too (add your units, and change "number of days")

greg0ire:
Closes #5835 ?

lcobucci (approved):

JoppeDC:
This wasn't in the docs yet, leading me to a search for how to do it. Luckily i found this old PR. I've created a PR to add this to the docs. 
https://github.com/doctrine/orm/pull/7729

PR #6743, opened 2017-10-02
Split `ORMException` into more fine-grained, better documented and typed exception classes

greg0ire:
This class is too big, and some exceptions could inherit from `doctrine/common` exceptions instead of this one.

# TODO

- [x] remove unused exception methods
    - [x] `entityMissingAssignedIdForField`    
    - [x] `entityMissingForeignAssignedId`     
    - [x] `invalidFindByCall`                  
    - [x] `invalidFlushMode`                   
- [x] `notSupported`                       
- [x] `namedNativeQueryNotFound`           
- [x] `namedQueryNotFound`                 
- [x] `proxyClassesAlwaysRegenerating`     
- [x] `invalidEntityRepository`            
- [x] `unknownEntityNamespace`             
- [x] `entityManagerClosed`                
- [x] `invalidHydrationMode`               
- [x] `mismatchedEventManager`             
- [x] `missingIdentifierField`             
- [x] `missingMappingDriverImpl`           
- [x] `unrecognizedIdentifierFields`       
- [x] `cantUseInOperatorOnCompositeKeys`   
- [x] `invalidOrientation`                 
- [x] `unexpectedAssociationValue`         
- [x] `unrecognizedField`                  
- [x] `findByRequiresParameter`            
- [x] `invalidMagicCall`                   
- [x] `invalidFindByInverseAssociation`    
- [x] `invalidResultCacheDriver`           
- [x] `queryCacheNotConfigured`            
- [x] `metadataCacheNotConfigured`         
- [x] `queryCacheUsesNonPersistentCache`   
- [x] `metadataCacheUsesNonPersistentCache`

# Proposal

| method                                | used in class                                         | should implement   |
|---------------------------------------|-------------------------------------------------------|--------------------------|
| `notSupported`                        | `Doctrine\ORM\Tools\SchemaTool`                       | `SchemaToolException`    |
| `namedNativeQueryNotFound`            | `Doctrine\ORM\Configuration`                          | `ConfigurationException` |
| `namedQueryNotFound`                  | `Doctrine\ORM\Configuration`                          | `ConfigurationException` |
| `proxyClassesAlwaysRegenerating`      | `Doctrine\ORM\Configuration`                          | `ConfigurationException` |
| `invalidEntityRepository`             | `Doctrine\ORM\Configuration`                          | `ConfigurationException` |
| `unknownEntityNamespace`              | `Doctrine\ORM\Configuration`                          | `ConfigurationException` |
| `entityManagerClosed`                 | `Doctrine\ORM\EntityManager`                          | `ManagerException`       |
| `entityMissingAssignedIdForField`     | no usages                                             | /dev/null                |
| `entityMissingForeignAssignedId`      | no usages                                             | /dev/null                |
| `invalidHydrationMode`                | `Doctrine\ORM\EntityManager`                          | `ManagerException`       |
| `mismatchedEventManager`              | `Doctrine\ORM\EntityManager`                          | `ManagerException`       |
| `missingIdentifierField`              | `Doctrine\ORM\EntityManager`                          | `ManagerException`       |
| `missingMappingDriverImpl`            | `Doctrine\ORM\EntityManager`                          | `ManagerException`       |
| `unrecognizedIdentifierFields`        | `Doctrine\ORM\EntityManager`                          | `ManagerException`       |
| `cantUseInOperatorOnCompositeKeys`    | `Doctrine\ORM\Persisters\Entity\BasicEntityPersister` | `PersisterException`     |
| `invalidFlushMode`                    | no usages                                             | /dev/null                |
| `invalidOrientation`                  | `Doctrine\ORM\Persisters\Entity\BasicEntityPersister` | `PersisterException`     |
| `unexpectedAssociationValue`          | `Doctrine\ORM\UnitOfWork`                             | ???                      |
| `unrecognizedField`                   | `Doctrine\ORM\Persisters\Entity\BasicEntityPersister` | `PersisterException`     |
| `findByRequiresParameter`             | `Doctrine\ORM\EntityRepository`                       | `RepositoryException`    |
| `invalidFindByCall`                   | no usages                                             | /dev/null                |
| `invalidMagicCall`                    | `Doctrine\ORM\EntityRepository`                       | `RepositoryException`    |
| `invalidFindByInverseAssociation`     | `Doctrine\ORM\Persisters\Entity\BasicEntityPersister` | `RepositoryException`    |
| `invalidResultCacheDriver`            | `Doctrine\ORM\AbstractQuery`                          | `CacheException`         |
| `queryCacheNotConfigured`             | `Doctrine\ORM\Configuration`                          | `CacheException`         |
| `metadataCacheNotConfigured`          | `Doctrine\ORM\Configuration`                          | `CacheException`         |
| `queryCacheUsesNonPersistentCache`    | `Doctrine\ORM\Configuration`                          | `CacheException`         |
| `metadataCacheUsesNonPersistentCache` | `Doctrine\ORM\Configuration`                          | `CacheException`         |


Ocramius:
@greg0ire the split is indeed needed (see https://github.com/doctrine/common/issues/681), but splitting just per component is a bit silly.

These are (almost) all separate exceptions, implementing a component-specific exception interface (for example, `CacheException`), and extend `ORMException` (need to do that for BC compliance).

For example:

```php
class InvalidRepositoryMagicMethodCall extends ORMException implements RepositoryException
{
    public static function fromMethodCallAndArguments(string $method, array $arguments) : self {
        ...
    }
}
```

In an ideal world, this should be:

```php
class InvalidRepositoryMagicMethodCall extends \InvalidArgumentException implements RepositoryException
{
    public static function fromMethodCallAndArguments(string $method, array $arguments) : self {
        ...
    }
}
```

Moving away from `ORMException` would be quite a BC break, so I don't think we can totally fix this.

greg0ire:
> Moving away from ORMException would be quite a BC break, so I don't think we can totally fix this.

@Ocramius My plan was to make `ORMException` an interface, and have all the split exceptions implement that interface. Is that a big BC-break? It's not for `instanceof`.

Ocramius:
@greg0ire given that the constructors would move around anyway, having an `ORMException` as an interface would be OK

greg0ire:
Ok so if I understand correctly, I should move every method to its separate exception class, and use the split I proposed to create interfaces instead of classes. I'll try to push an example soon based on the "ideal world" snippet you gave + `implements ORMException` for BC

Ocramius:
@greg0ire the `implements ORMException` can be done on the namespace-specific exception marker interface:

```php
namespace Doctrine\ORM\Repository;

interface RepositoryException extends \Throwable, \Doctrine\ORM\ORMException
{
}
```

greg0ire:
Even better!

Majkl578:
ORMException being an interface was exactly my idea.

```
namespace Doctrine\ORM;

interface ORMException extends \Throwable {}
```
Then specific exceptions would implement it (or have its own interface, as shown above):
```
use  Doctrine\ORM\ORMException;

interface MappingException extends ORMException {}

class BlahException extends \RuntimeException implements MappingException {}
```

lib/Doctrine/ORM/NewORMException.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM;
+
+/**
+ * This interface should be implemented by all exceptions in this package.
+ */
+interface NewORMException extends \Throwable
```
greg0ire: Will of course be refactored to `ORMException`

lib/Doctrine/ORM/Repository/InvalidMagicMethodCall.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Repository;
+
+use Doctrine\ORM\RepositoryException;
+
+final class InvalidMagicMethodCall extends \InvalidArgumentException implements RepositoryException
```
greg0ire: I made the exception final

lib/Doctrine/ORM/Repository/InvalidMagicMethodCall.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Repository;
```
greg0ire: I put the exception in the namespace of the relevant component.

lib/Doctrine/ORM/RepositoryException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM;
```
greg0ire: I put that at the top level namespace to avoid the repetition. An alternative would be to name the class `Doctrine\ORM\Repository\Exception`, tell me what you prefer.

lib/Doctrine/ORM/Repository/InvalidMagicMethodCall.php
greg0ire: I removed the `Repository` bit from the name since it already is in the namespace.

lib/Doctrine/ORM/Repository/InvalidMagicMethodCall.php
```diff
@@ -0,0 +1,21 @@
...
+
+use Doctrine\ORM\RepositoryException;
+
+final class InvalidMagicMethodCall extends \InvalidArgumentException implements RepositoryException
+{
+    public static function fromEntityNameFieldNameAndMethod(
```
greg0ire: This method name is a bit long, I'm open to proposals

lib/Doctrine/ORM/Repository/InvalidMagicMethodCall.php
SenseException: Was `InvalidArgumentException` chosen instead of `BadMethodCallException` because of BC or because magic methods are getting the method name as an argument?

lib/Doctrine/ORM/Repository/InvalidMagicMethodCall.php
SenseException: `fieldNameNotFound` or `fieldNotFound` because it's the reason for not be able to use a magic method in an entity.

lib/Doctrine/ORM/Repository/InvalidMagicMethodCall.php
greg0ire: I don't know, I copy pasted a snippet from @Ocramius but `BadMethodCallException` might be better indeed: it's not about the arguments being invalid, it's about the field not existing.

lib/Doctrine/ORM/Repository/InvalidMagicMethodCall.php
greg0ire: So no `from`?

SenseException: The static name may lack fluent reading because of a missing method prefix like "from". Sounds funny with "because" `InvalidMagicMethodCall::becauseFieldNotFoundIn`, but I like "In" as suffix. Maybe the suffix can make it more readable without having needed arguments in the method name.

lib/Doctrine/ORM/Repository/InvalidMagicMethodCall.php
```diff
@@ -0,0 +1,21 @@
...
+        string $entityName,
+        string $fieldName,
+        $method
```
greg0ire: Woops I forgot the typehint here

lib/Doctrine/ORM/Repository/InvalidFindByInverseAssociation.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Repository;
+
+use Doctrine\ORM\RepositoryException;
+
+final class InvalidFindByInverseAssociation extends \BadMethodCallException implements RepositoryException
```
greg0ire: I'm not a huge fan of this class name. I know I created it myself, but it's derived from the former method name. What I don't like is that it kind of implies there could be valid find by inverse association method calls. Should I drop the `Invalid` part?

lib/Doctrine/ORM/Repository/InvalidFindByInverseAssociation.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    public static function becauseIsInverseAssociation(
```
greg0ire: @Ocramius @SenseException can you please agree with a "because" vs "from" naming recommendation before I migrate all the exceptions?

lib/Doctrine/ORM/Repository/InvalidFindByInverseAssociation.php
SenseException: I guess the name can be discussed in combination with the exception methods. I suggest `InvalidAssociation` or `InvalidAssociationSide`. I've read the exception message in the tests and used the mentioned names for it.

lib/Doctrine/ORM/Repository/InvalidFindByInverseAssociation.php
SenseException: Like I previously wrote, "because" sounds too funny, so I can agree on a different prefix. If we go with my exception name suggestions of a previous comment about this class name, we can create a method like `InvalidAssociationSide::fromInverseSide()` or `InvalidAssociation::fromInverseSideUsage()`. We can recombine method name and class name between those two method-suggestions.

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
```diff
@@ -683,7 +684,7 @@ public function testCanRetrieveRepositoryFromClassNameWithLeadingBackslash()
     /**
      * @group DDC-1376
      *
-     * @expectedException Doctrine\ORM\ORMException
+     * @expectedException Doctrine\ORM\Repository\InvalidFindByInverseAssociation
      * @expectedExceptionMessage You cannot search for the association field 'Doctrine\Tests\Models\CMS\CmsUser#address', because it is the inverse side of an association.
```
SenseException: I just realized it now, that one test is using `$this->expectException()` and ` $this->expectExceptionMessage()` while  we have annotations here.

lib/Doctrine/ORM/Repository/InvalidFindByInverseAssociation.php
greg0ire: `InvalidAssociationSide` sounds great

greg0ire: But the other one leaves more room for other named constructors.

lib/Doctrine/ORM/Repository/InvalidFindByInverseAssociation.php
greg0ire: I don't think there is a point to using named constructor if we don't have several of them. I'll therefore go with your second proposition.

greg0ire: Or rather, `InvalidFindByCall::fromInverseSideUsage`

greg0ire: I think I'm gonna cram `findByRequiresParameter` in there, Ocramius said those should be almost all separate exceptions, I think using the "almost" card here is right.

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
greg0ire: I think annotations should be completely dropped in favor of those method calls :P

lib/Doctrine/ORM/Repository/InvalidFindByInverseAssociation.php
greg0ire: `fromMissingParameter` sounds weird though, maybe `because` is not that bad 🤔

greg0ire: How about `onMissingParameter`?

greg0ire: > I think I'm gonna cram findByRequiresParameter in there, Ocramius said those should be almost all separate exceptions, I think using the "almost" card here is right.

as @ostrolucky pointed out, this one is not about `findBy`, but about a magic call, so I guess I'm gonna put it in `InvalidMagicCall` instead

greg0ire:
Wait why is `invalidFindByInverseAssociation` only called in `BasicEntityPersister`? Does it really have to do with `findBy`?

greg0ire:
> Wait why is invalidFindByInverseAssociation only called in BasicEntityPersister? Does it really have to do with findBy?

Looking at the unit tests, it seems this exception is indeed supposed to be thrown when using `findBy` on a repo, so I guess it is ok.

lib/Doctrine/ORM/Repository/InvalidFindByInverseAssociation.php
SenseException: `EvilMagicCall`  😆
It sounds right that it belongs into an exception for magic calls.
About just one needed named constructor: They become or already are a standard in Doctrine, so it would be weird if we have one case where an instance was created with `new`. Especially if that exception class might need a second exception case one day, we have to look that the one with `new` will get an own named constructor too.

lib/Doctrine/ORM/RepositoryException.php
SenseException: We have a `PersisterException` in `Doctrine\ORM\Persisters` namespace. Also a `CacheException` in `Doctrine\ORM\Cache`. If we keep that structure, it suggests that `RepositoryException` will be in `Doctrine\ORM\Repository`.

I would like to see a structure as you suggested with `Doctrine\ORM\Repository\Exception`, even with the already existing exception classes, but I don't think that this would be an anticipated change.

Majkl578:
Needs a rebase.
Also I think all exceptions should be suffixed by `Exception`, at least for consistency.

greg0ire:
>  Also I think all exceptions should be suffixed by Exception, at least for consistency.

That would make the already long names a bit longer, and IMO the exception suffix should be deprecated (copy and extend w/o suffix)

lib/Doctrine/ORM/Persisters/UnrecognizedField.php
```diff
@@ -0,0 +1,15 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Persisters;
+
+use Doctrine\ORM\PersisterException;
+
+class UnrecognizedField extends \Exception implements PersisterException
+{
+    public static function fromField(string $field): self
```
SenseException: What about a method name like `UnrecognizedField::byName()` or `UnrecognizedField::ofName()`?

Also: missing space in return value at `:`.

Majkl578:
> That would make the already long names a bit longer, and IMO the exception suffix should be deprecated (copy and extend w/o suffix)

The issue there is that some exceptions become ambiguous. Especially names like  `InvalidEntityRepository` invoke feeling that it's real repository that is invalid. It's even more obvious when they're not put into Exception sub-namespace.

greg0ire:
I'd rather put them in a sub-namespace, *if that's ok with everyone*. So please decide together.

greg0ire:
It think some exception might be regrouped too, like `MetadataCacheUsesNonPersistentCache` and `QueryCacheUsesNonPersistentCache`, please tell me what you think.

Majkl578:
> I'd rather put them in a sub-namespace, if that's ok with everyone.

This is rather for a question on doctrine/coding-standard, it's not a simple yes/no here.

So far we:
* don't use sub-namespaces anywhere (ORM, DBAL, Common, ODM) (although I'd be +1 on this);
* do suffix all exceptions with `Exception`.

Majkl578:
@Ocramius @lcobucci @guilhermeblanco What's the concensus about naming? I still think exceptions must be suffixed with `Exception` for consistency and clarity, and ideally be in Exception sub-namespace.

greg0ire:
That's the same information twice though

SenseException:
The Exception class name should of course indicate, that it represents something that went wrong. This is not only a thing for Exception suffixes, but also for design patterns (Factory, Decorator, etc.) or abstractions, where some languages prefer to have an "I" class name prefix for interfaces while others prefer the suffix "Interface". Starting with PHP 7, errors are `\Throwable` like an Exception, but doesn't have the suffix like the default or SPL ones, but this might be because they want to differentiate between "regular" Exceptions and errors.

I kind of started to like not having suffixes, but it depends on the understanding of the devs. Is there anything we can improve the exception class names? Or do we still depend on suffixes?

Ocramius:
I'm fine with not having suffixes, but the naming must be extremely clear, as in it shouldn't be possible to confuse a service with an exception by just looking at the name.

lcobucci:
> What's the concensus about naming? I still think exceptions must be suffixed with Exception for consistency and clarity.

You know that I'm not a big fan of prefixes and suffixes, so I'm happy with removing it (considering the the class name is consistent and very clear about it's intent).

> and ideally be in Exception sub-namespace.

I disagree on grouping all exceptions in the Exception sub-namespace. I believe that things that change together should be grouped together, so I'm quite happy with the suggestion given in https://github.com/doctrine/doctrine2/pull/6743#issuecomment-333908743

greg0ire:
>  but the naming must be extremely clear

That's a feature of not using suffixes: instead of lazy, vague naming, you get precise naming.

I think the suffix should stay for interfaces such as `MappingException` or `ORMException`, because it groups several exceptions so you cannot be precise.

greg0ire:
The February bug has been fixed, let's trigger the build again

greg0ire:
Hmmm not for master it seems…

lcobucci:
@greg0ire it should be fixed by #7055

lib/Doctrine/ORM/Cache/Exception/FeatureNotImplemented.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Cache\Exception;
+
+class FeatureNotImplemented extends \Exception implements CacheException
+{
+    public static function scalarResults() : self
+    {
+        return new self('Second level cache does not support scalar results.');
```
Majkl578: Should return, not throw.

lib/Doctrine/ORM/Cache/Exception/FeatureNotImplemented.php
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public static function multipleRootEntities() : self
+    {
+        throw new self('Second level cache does not support multiple root entities.');
```
Majkl578: Should return, not throw.

lib/Doctrine/ORM/Cache/Exception/FeatureNotImplemented.php
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public static function nonSelectStatements() : self
+    {
+        throw new self('Second-level cache query supports only select statements.');
```
Majkl578: Should return, not throw.

lib/Doctrine/ORM/Cache/Exception/FeatureNotImplemented.php
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public static function partialEntities() : self
+    {
+        throw new self('Second level cache does not support partial entities.');
```
Majkl578: Should return, not throw.

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -677,12 +677,11 @@ private function gatherRelationJoinColumns(
             );
 
             if (! $definingClass) {
-                throw new ORMException(sprintf(
-                    'Column name "%s" referenced for relation from %s towards %s does not exist.',
+                throw new MissingColumnException(
```
Majkl578: Why is there no factory for this exception?

lib/Doctrine/ORM/ORMException.php
```diff
@@ -4,221 +4,9 @@
...
-use Exception;
-use function get_class;
-use function implode;
-use function sprintf;
-
 /**
  * Base exception class for all ORM exceptions.
  */
-class ORMException extends Exception
+interface ORMException extends \Throwable
```
Majkl578: Since we target master, can we maybe get rid of this polyfill and move it to its proper place (`Doctrine\ORM\Exception`)?

tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterCompositeTypeSqlTest.php
```diff
@@ -51,11 +52,9 @@ public function testSelectConditionStatementNeqNull()
         self::assertEquals('t0."admin1" IS NOT NULL AND t0."country" IS NOT NULL', $statement);
     }
 
-    /**
-     * @expectedException Doctrine\ORM\ORMException
-     */
     public function testSelectConditionStatementIn()
     {
+        $this->expectException(CantUseInOperatorOnCompositeKeys::class);
```
Majkl578: Why converting it into a method here while keeping it in annotations on other places?

lib/Doctrine/ORM/Cache/Exception/InvalidResultCacheDriver.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Cache\Exception;
+
+final class InvalidResultCacheDriver extends \Exception implements CacheException
+{
+    public static function create() : self
```
Majkl578: Generic `create()` method doesn't say much, could we have them express what they really do? I.e. `notImplementingCacheInterface()`?

Majkl578 (dismissed):
Generally I would advise to never extend Exception directly. Code should either extend RuntimeException or LogicException, whichever fits to the use case.
Also for cases like `throw NotSupported::create()`, I'd probably be in favor of partial uses, i.e. `throw Exception\NotSupported::create()`, for clarity.

lib/Doctrine/ORM/ORMException.php
greg0ire: Sure, why not?

tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterCompositeTypeSqlTest.php
greg0ire: Laziness/rabbit hole avoiding techniques I think

lib/Doctrine/ORM/Cache/Exception/InvalidResultCacheDriver.php
greg0ire: Won't that be a bit long?

lib/Doctrine/ORM/Tools/SchemaTool.php
greg0ire: There is, but I don't seem to use it, my mistake.

tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterCompositeTypeSqlTest.php
greg0ire: Or was it maybe line length?

greg0ire:
> Also for cases like throw NotSupported::create(), I'd probably be in favor of partial uses, i.e. throw Exception\NotSupported::create(), for clarity.

IMO If it's unclear, then the name is bad. Maybe rename to `FeatureNotSupported` in that case?

tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterCompositeTypeSqlTest.php
SenseException: The recommendation is to use the method instead of the annotation. It keeps expected results like exceptions and assertions in the test's code.

greg0ire: Yeah and also, exception triggered *before* the call that expected to trigger them will not make the tests pass. But that's out of scope, sorry, that PR is big enough as is.

greg0ire:
> Generally I would advise to never extend Exception directly. Code should either extend RuntimeException or LogicException, whichever fits to the use case.

@Majkl578 I addressed this, please review again

> Also for cases like throw NotSupported::create(), I'd probably be in favor of partial uses, i.e. throw Exception\NotSupported::create(), for clarity.

Doctrine CS?

Ocramius (approved):
This goes in as-is :+1: 

The BC breaks to be documented are to be added to the `UPGRADE.md`, but I'm raising a separate issue for that.

greg0ire:
Thanks a lot for merging! Will list the BC-break this week-end.

Majkl578:
@greg0ire Any update on documentating these changes?

Majkl578:
Also what do we do with all the exceptions left in root namespace?
```
lib/Doctrine/ORM/EntityNotFoundException.php
lib/Doctrine/ORM/NonUniqueResultException.php
lib/Doctrine/ORM/NoResultException.php
lib/Doctrine/ORM/OptimisticLockException.php
lib/Doctrine/ORM/ORMInvalidArgumentException.php
lib/Doctrine/ORM/PessimisticLockException.php
lib/Doctrine/ORM/TransactionRequiredException.php
lib/Doctrine/ORM/UnexpectedResultException.php
```

Current state is highly inconsistent.

PR #6746, opened 2017-10-03
[DOCS] Emphasize that Embeddable can only contain simple fields

malarzm:
I personally overlooked this sentence and now I'm cursing :P


Ocramius:
:ship:

PR #6747, opened 2017-10-03
Add Doctrine\ORM\Query\Expr::mod()

holtkamp:
As suggested in https://github.com/doctrine/doctrine2/issues/6739

### DIFF
--- a/lib/Doctrine/ORM/Query/Expr.php
+++ b/lib/Doctrine/ORM/Query/Expr.php
@@ -329,6 +329,17 @@ class Expr
     }
 
     /**
+     * Creates a MOD($x, $y) function expression to return the remainder of $x divided by $y.
+     *
+     * @param mixed $x
+     * @param mixed $y
+     */
+    public function mod($x, $y): Expr\Func
+    {
+        return new Expr\Func('MOD', array($x, $y));
+    }
+
+    /**
      * Creates a product mathematical expression with the given arguments.
      *
      * First argument is considered the left expression and the second is the right expression.
--- a/tests/Doctrine/Tests/ORM/Query/ExprTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/ExprTest.php
@@ -119,6 +119,11 @@ class ExprTest extends OrmTestCase
         self::assertEquals('ABS(1)', (string) $this->expr->abs(1));
     }
 
+    public function testModExpr() : void
+    {
+        self::assertEquals('MOD(10, 1)', (string) $this->expr->mod(10, 1));
+    }
+
     public function testProdExpr() : void
     {
         self::assertEquals('1 * 2', (string) $this->expr->prod(1, 2));


Ocramius:
This looks OK-ish to me, but where do we draw the line between adding new expressions to the fluent interface?

If this keeps going, we'll soon have a lot of added API... Maybe stop at the `Expr`?

holtkamp:
mmm, I guess I understand this doubt. I would expect all [DQL functions](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html#dql-functions) to be supported. 

What would be the alternative? Create dedicated classes like 
- Doctrine\ORM\Query\Expr\Math\Mod 

?

ostrolucky:
We could get rid of a lot of methods in this class via generic magic call for Expr\Func. Personally I would get rid of this class completely though.

holtkamp:
> via generic magic call for Expr\Func

Should we not prevent the use of magic `__call()` approaches?

ostrolucky:
Nothing is white or black. I don't like magic calls in general but they are good fit for removing boilerplate methods like these.

But it was just an idea thrown out there, it probably won't pass anyway.

holtkamp:
Encountered this again, any update on the stance on this?

holtkamp:
@Ocramius and @ostrolucky maybe it is time to take a decision and accept or close this PR?

ostrolucky:
Please do not merge, rebase. Please remove merge commit. I'm fine with change otherwise.

holtkamp:
@ostrolucky I would be happy though, although I am afraid I have reached the limits of my GIT skills here.. What I did/tried:


```
git reset --hard HEAD^ 
git push origin -f
git merge-base patch-add-modulo-expression upstream/master
git rebase -i 2ade863bca0ffbe08b9f2999b1dfe726c10c3e95
```

ostrolucky:
It's
```
git reset --hard HEAD^ 
git rebase master
// solve conflicts
git rebase --continue
```

ostrolucky (dismissed):

lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -328,6 +328,19 @@ public function abs($x)
         return new Expr\Func('ABS', [$x]);
     }
 
+    /**
+     * Creates a MOD($x, $y) function expression to return the remainder of $x divided by $y.
+     *
+     * @param mixed $x
+     * @param mixed $y
```
greg0ire: What types exactly can be passed here? Not that many I guess?

lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -328,6 +328,19 @@ public function abs($x)
...
+     *
+     * @return Expr\Func
```
greg0ire: I would remove this and use a return type declaration here, regardless of the context, not sure what other contributors think. For me it's just more code to migrate.

tests/Doctrine/Tests/ORM/Query/ExprTest.php
```diff
@@ -119,6 +119,11 @@ public function testAbsExpr() : void
         self::assertEquals('ABS(1)', (string) $this->expr->abs(1));
     }
 
+    public function testModExpr()
```
greg0ire:
```suggestion
    public function testModExpr() : void
```

beberlei:
@ostrolucky to go on a tangent here for your generic argument, it doesn't have to be magic call, what about `->fn('mod', ...$args)`. Maybe something we can investigate for an improvement later?

ostrolucky:
@beberlei Unlike `->fn('mod', ...$args)` magic functions can be documented with `@method`, so you still get IDE auto complete and SA tools inference.

beberlei:
@ostrolucky Technically yes, but then you need to change the code every time we add a new function *again*, and for custom functions registers by users we can't add them there either.

lib/Doctrine/ORM/Query/Expr.php
ostrolucky: I imagine it can be int|string|float|object (with _toString). It also follows current standard in file, I am not sure we should increase the scope of this PR and possibly document wrong types. Somebody should instead investigate types for all functions in file and document types in separate PR.

ostrolucky:
Well yes we do need to do that, IDE and SA support is important. How else would we get it? My suggestion was only to reduce boilerplate. Otherwise let's just remove the class, because then I don't see what it brings on table compared to just doing `new Expr\Func('mod', [$x, $y]);` in userspace.

holtkamp:
Is there anything I can do to get this PR merged? We are almost reaching its 3 year anniversary... 😄

ostrolucky (approved):

ostrolucky:
Indeed. Thanks for ping and patience, let's merge.

PR #6748, opened 2017-10-04
Fix missing null check on `AbstractQuery#setResultCacheProfile()`

KonstantinKuklin:
subject

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -521,7 +521,7 @@ abstract class AbstractQuery
      */
     public function setResultCacheProfile(QueryCacheProfile $profile = null)
     {
-        if ( ! $profile->getResultCacheDriver()) {
+        if ($profile !== null && ! $profile->getResultCacheDriver()) {
             $resultCacheDriver = $this->_em->getConfiguration()->getResultCacheImpl();
             $profile = $profile->setResultCacheDriver($resultCacheDriver);
         }
--- a/tests/Doctrine/Tests/ORM/Query/QueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/QueryTest.php
@@ -5,6 +5,7 @@ namespace Doctrine\Tests\ORM\Query;
 use Doctrine\Common\Cache\ArrayCache;
 use Doctrine\Common\Collections\ArrayCollection;
 
+use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Internal\Hydration\IterableResult;
 use Doctrine\ORM\Query\Parameter;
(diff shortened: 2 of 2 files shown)


Ocramius:
While the failure is trivial, we still need a test case for this

KonstantinKuklin:
@Ocramius done

tests/Doctrine/Tests/ORM/QueryTest.php
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    /**
+     * setResultCacheProfile should works correct with null
+     */
+    public function testSetResultCacheProfile()
+    {
+        $query = $this->_em->createQuery('SELECT u FROM ' . CmsUser::class . ' u');
+        $result = $query->setResultCacheProfile(null);
+        self::assertEquals($result, $query);
```
jwage: I think we should be using `$this` instead of `self::` like the rest of the code base right?

alcaeus: Strictly speaking, the `assert*` methods in PHPUnit are static, so they should be called using `self::assert*`, not `$this->assert*`. IIRC @Ocramius and @lcobucci have started changing these in new tests and for v3.

KonstantinKuklin:
so, any updates here guys?
@Ocramius , @jwage

tests/Doctrine/Tests/ORM/QueryTest.php
```diff
@@ -0,0 +1,34 @@
...
+
+    protected function setUp()
+    {
+        $this->_em = $this->_getTestEntityManager();
+    }
+
+    /**
+     * setResultCacheProfile should works correct with null
+     */
+    public function testSetResultCacheProfile()
```
lcobucci: Shouldn't this be part of https://github.com/doctrine/doctrine2/blob/f2dc9a8f92b7eb9a566b3ba7af16559cc2bc0eb7/tests/Doctrine/Tests/ORM/Query/QueryTest.php?

tests/Doctrine/Tests/ORM/QueryTest.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        $query = $this->_em->createQuery('SELECT u FROM ' . CmsUser::class . ' u');
+        $result = $query->setResultCacheProfile(null);
```
lcobucci: This is a bit misleading, `$result` is not the query result but rather the query

tests/Doctrine/Tests/ORM/QueryTest.php
lcobucci: Shouldn't we assert that the cache profile was NOT configured?

lcobucci (changes requested):

Majkl578:
Linking #6861 fixing the same bug, but missing tests.

lcobucci (approved):

lcobucci:
@KonstantinKuklin 🚢 and backported via 56ebf4bea91d4c0bbcf411c43f2f322a1a65fafa, thanks for your contribution!

PR #6759, opened 2017-10-06
Inverse side of OneToOne is loaded without identifier condition

janlanger:
When testing of 2.6.x-dev, I found out that under certain conditions is inverse side of OneToOne relation loaded without **any SQL condition** - whith results in loading first row returned by database and possibly corrupted data.

Condition when this occurs are:
- Have bidirectional 1-1 relation
- Have relation column named same as entity field (eg. `@JoinColumn(name="field")`) 
- Load entity on inverse side via DQL query (issue does not occur with em->find)

=> 💥  Mapping side of the realtion will be loaded with second SQL query, but WHERE condition is missing

**Attached is failing test case illustration the issue.**

Bug is probably caused by https://github.com/doctrine/doctrine2/commit/a360da5a7e5463f6c12b0922f6063e1cf32dde6d

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/OneToOneInverseSideLoadAfterDqlQueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/OneToOneInverseSideLoadAfterDqlQueryTest.php
@@ -1,0 +1,109 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use Doctrine\ORM\Tools\SchemaTool;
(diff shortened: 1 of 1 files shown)


Ocramius:
I fixed this locally and am trying to reduce the test to an extreme minimum.

Just an interesting remark: this only applies to `joinColumn === fieldName`. Weird coincidence caused by a variable reuse.

Ocramius:
Moved to https://github.com/doctrine/doctrine2/pull/6760

PR #6762, opened 2017-10-10
Use of static assert methods in tests

SenseException:
Based on comment https://github.com/doctrine/doctrine2/pull/6748#discussion_r142661060, I changed the asserts of the existing tests to a static method call. Just a PR for getting more consistency.


Ocramius:
@SenseException I think this was already handled in `develop`

SenseException:
Do'h. Seems like this wasn't a "everything to master first" issue like the contribution guide explains. I'll close this PR then. Thank you for pointing that out. I haven't thought about looking into develop, expecting such a state in master too.

SenseException:
@Ocramius Is the "everything to master first" part of the contribution guide still correct?

Ocramius:
@SenseException please don't change that - we're merging `develop` into `master` as soon as we're done with 2.6.0

PR #6767, opened 2017-10-12
Allow association mappings as IDs for joined-table inherited entity

Parent5446:
SchemaTool has custom logic for creating the primary key of a
joined-table inherited entity. This logic overlooked association maps
as a possible source for identity columns, resulting in a fatal error
when fetching the primary key list for child entities.

Removed any custom logic for generating primary keys for root entities
in joined-table inheritance, deferring to the common logic used for
other entities.

Also adjusted the child entity logic, scanning association maps for
identity columns, and including the column as appropriate. It also
ensures that the primary key columns are in the correct order.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -183,19 +183,9 @@ class SchemaTool
                 }
             } elseif ($class->isInheritanceTypeJoined()) {
                 // Add all non-inherited fields as columns
-                $pkColumns = [];
                 foreach ($class->fieldMappings as $fieldName => $mapping) {
                     if ( ! isset($mapping['inherited'])) {
-                        $columnName = $this->quoteStrategy->getColumnName(
-                            $mapping['fieldName'],
-                            $class,
-                            $this->platform
-                        );
                         $this->gatherColumn($class, $mapping, $table);
-
-                        if ($class->isIdentifier($fieldName)) {
-                            $pkColumns[] = $columnName;
-                        }
                     }
                 }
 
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -221,9 +213,39 @@ public function getSchemaFromMetadata(array $classes)
...
+                                    $classes,
+                                    function (ClassMetadata $class) use ($idMapping) : bool {
+                                        return $class->name === $idMapping['targetEntity'];
+                                    }
+                                )
+                            );
+
+                            foreach ($idMapping['joinColumns'] as $joinColumn) {
+                                if (isset($targetEntity->fieldMappings[$joinColumn['referencedColumnName']])) {
+                                    $idMapping = $targetEntity->fieldMappings[$joinColumn['referencedColumnName']];
```
lcobucci: should be removed since`$idMapping` is not being used

lcobucci (changes requested):
I've applied some minor changes on your patch, reducing the complexity and improving the tests. Apart from the unused variable it LGTM

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -496,8 +518,8 @@ private function gatherColumn($class, array $mapping, Table $table)
      */
     private function gatherRelationsSql($class, $table, $schema, &$addedFks, &$blacklistedFks)
     {
-        foreach ($class->associationMappings as $mapping) {
-            if (isset($mapping['inherited'])) {
+        foreach ($class->associationMappings as $id => $mapping) {
+            if (isset($mapping['inherited']) && array_search($id, $class->identifier) === false) {
```
Majkl578: Should be using strict mode - 3rd parameter `true`.

guilhermeblanco (approved):
Aside from the 2 comments, that looks correct to me.

lib/Doctrine/ORM/Tools/SchemaTool.php
lcobucci: Isn't `in_array()` more appropriated here?

Majkl578: Good point, although both should do. :)

Ocramius:
Manually merged into `master` for `2.6` in doctrine/doctrine2@4c894983597d2e615069106a3bac52b7f0b6a6f2

PR #6777, opened 2017-10-16
Add validation on return values for Memcache and Redis connect() meth…

ywarnier:
…ods and obfuscate uncatcheable warning in Tools/Setup::createConfiguration()

The current switch throws a warning and generates false expectations on the cache. Instead, this modified approach makes sure the memcache and redis connections work before considering they should be used as cache mechanisms.

If left as is, the createConfiguration() method will *always* throw a warning if the php-redis or php-memcache extensions are installed and enabled although there is no memcache/memcached/redis server on localhost. Typically, a cheap hosting solution might have the memcache or redis extensions available and enabled without possibility to disable them, although the redis/memcache servers are not part of the package.

Also, the if/elseif structure did not allow for a finely tuned check on the results of connections.


lib/Doctrine/ORM/Tools/Setup.php
```diff
@@ -126,19 +126,27 @@ public static function createConfiguration($isDevMode = false, $proxyDir = null,
...
-            } elseif (extension_loaded('memcache')) {
+            }
+            if ($cache === null && extension_loaded('memcache')) {
                 $memcache = new \Memcache();
-                $memcache->connect('127.0.0.1');
-                $cache = new \Doctrine\Common\Cache\MemcacheCache();
-                $cache->setMemcache($memcache);
-            } elseif (extension_loaded('redis')) {
+                $r = @$memcache->connect('127.0.0.1');
+                if ($r !== false) {
```
Ocramius: This should throw if failed.

lib/Doctrine/ORM/Tools/Setup.php
```diff
@@ -126,19 +126,27 @@ public static function createConfiguration($isDevMode = false, $proxyDir = null,
...
+                }
+            }
+            if ($cache === null && extension_loaded('redis')) {
                 $redis = new \Redis();
-                $redis->connect('127.0.0.1');
-                $cache = new \Doctrine\Common\Cache\RedisCache();
-                $cache->setRedis($redis);
-            } else {
+                $r = @$redis->connect('127.0.0.1');
+                if ($r !== false) {
```
Ocramius: This should throw if failed.

Ocramius (changes requested):

lib/Doctrine/ORM/Tools/Setup.php
Majkl578: It should not, it's a _dumb_ attempt to discover running memcache simply based on ext existence without configuration - if not found, try another storage.
@lcobucci suggested to drop Setup entirely in 3.0, sounds good to me.

Ocramius: Yeah, probably a better solution. This is way too magic and simplistic.

ywarnier: @Majkl578 could you be more specific about "drop Setup entirely"? You mean dropping the "connect" call, right?

lcobucci: @ywarnier no, dropping the class entirely. It is completely unnecessary.

PR #6780, opened 2017-10-19
getRepository actually returns ObjectRepository

cryptiklemur:
Because the `repositoryFactory` in `EntityManager` must implement the `RepositoryFactory` interface. `RepositoryFactory`'s `getRepository` method advertises that it returns an instance of `\Doctrine\Common\Persistence\ObjectRepository`.

I've kept the `EntityRepository` return as an or, because in practice, people still expect the remaining functions from `EntityManager` (like `createQueryBuilder`) on the response.

cc/ @weaverryan

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -709,7 +709,7 @@ use Throwable;
      *
      * @param string $entityName The name of the entity.
      *
-     * @return \Doctrine\ORM\EntityRepository The repository class.
+     * @return \Doctrine\Common\Persistence\ObjectRepository|\Doctrine\ORM\EntityRepository The repository class.
      */
     public function getRepository($entityName)
     {


lib/Doctrine/ORM/EntityManager.php
```diff
@@ -709,7 +709,7 @@ public function lock($entity, $lockMode, $lockVersion = null)
      *
      * @param string $entityName The name of the entity.
      *
-     * @return \Doctrine\ORM\EntityRepository The repository class.
+     * @return \Doctrine\Common\Persistence\ObjectRepository|\Doctrine\ORM\EntityRepository The repository class.
```
SenseException: Isn't `EntityRepository` the wrong return type for `getRepository()`? The user can change the default repository class in `Doctrine\ORM\Configuration::setDefaultRepositoryClassName()` and it has to be a class that implements `ObjectRepository`. So we had the wrong return type all the time for repository classes.

Is it a BC break to remove `EntityRepository` from the phpdoc return annotation when the type was wrong in the past?

alcaeus: It's not a BC break to remove `EntityRepository`, but then I think it's also wrong: without adding a specific repository, the `EntityRepository` type hint is quite valid. From my personal experience I've also found most people to extend `EntityRepository` when creating their own repository classes, so the return type hint is still correct in most cases.

SenseException: `EntityRepository` implements `ObjectRepository` and `Selectable`, while a custom Repository class only needs to have `ObjectRepository`. If devs are using the `Selectable` implementations in `EntityRepository`, a custom Repo allows break compatibility. That's why I wasn't sure about BC.

aequasi: Like alcaeus says, technically it's not a BC break because it's not a contract in the code to return EntityRepository, but users are expecting it in autocompletion.

Majkl578: EntityRepository should probably be an interface anyway (maybe in 3.0) and be made a parent of user repositories.

This return type looks strange but if people can register _anything_ as their base repository class, well...

Majkl578 (approved):
I think we can live with this in 2.x for convenience, we may eventually reconsider how base repository class/interface is treated for 3.0 anyway.

lib/Doctrine/ORM/EntityManager.php
lcobucci: Honestly `EntityManager#getRepository()` should (eventually) die a terrible death.

Majkl578:
Going to merge this into 2.x since it's a design failure in the contract that cannot be fixed in 2.x without BC break. For 3.x I've proposed an interface: #6871.

PR #6788, opened 2017-10-23
[Docs] Type::convertToDatabaseValue is always called

mateuszsip:
Custom types documentation includes a false statement that `Type::convertToDatabaseValue` is not called for NULL values.

According to [DBAL issue](https://github.com/doctrine/dbal/issues/2896) this looks like expected behavior so it would be great to remove misleading line from ORM docs.


Ocramius (approved):
LGTM 👍

Thanks @kejwmen!

Issue #4543, opened 2015-04-22
DDC-3708: Doctrine SQL filters and lazy loading causing EntityNotFoundException


doctrinebot:
Jira issue originally created by user pavle.predic:

When using a global SQL filter, lazy loading ManyToOne relations may lead to an EntityNotFoundException. You will always get a proxy object for the relation, but as soon as you try to access any property, doctrine will attempt to load the entity from DB, get zero results because the global filter has been applied (assuming that the filter affects the entity in question) and throw an EntityNotFoundException. 

What is the recommended way of dealing with this issue globally? The use case is a CMS with a soft-delete feature. A global sql filter filters out entities with a deleted flag. One solution would be to write a custom ProxyFactory that doesn't throw the EntityNotFoundException or to write a custom ProxyGenerator that puts the initializer call in a try/catch block and returns null if an EntityNotFoundException is thrown. 

This issue can be resolved by using "fetch EAGER" on ManyToOne relations, but this can lead to performance issues due to the large number of DB queries. How can this problem be solved while still using the lazy loading mechanism?

althaus:
>One possible solution is to always do joins on filtered associations, and then fetch either NULL or the identifier for those associations.

@Ocramius Where should we do that? I started to use `JOIN`s every now and than, but that's leading to other issues. It's simply making the filter more or less a waste of code. :(

Ocramius:
SQL filters (and soft deletes) are most probably a waste of code in most scenarios anyway.
Extending the proxy factory is a no-go, but the problem is much wider than what it looks like from in this tiny issue.

Basically, the ORM would have to completely change how hydration works, at the risk of breaking persistence later on.

Here's an example:

```sql
SELECT f FROM Foo f
```

Where `f.bar` is a `Bar` `*-to-one` association. Say following record is fetched (raw):

```php
[
    'id' => 123,
    'name' => 'yadda',
    'bar_id' => 456,
]
```

Assuming that `Bar` with `id` `456` is soft-deleted, we'd obviously have a failure during lazy-loading.

If we instead fix DQL so that the results are fetched with filtered values, we get:

```php
[
    'id' => 123,
    'name' => 'yadda',
    'bar_id' => null,
]
```

While this works for read APIs, persisting an object with this data hydrated into it will indeed set `bar_id` to `null` (breaks the reason why you used soft-deletes in first place) during the next flush operation.

[...]

PR #6791, opened 2017-10-24
Add option for QueryBuilder to use deferred Criteria

thomasklein94:
In one of my projects using doctrine, I needed to achieve something like adding a criteria to a QueryBuilder, and only evaluates this Criteria when getting the real Query from the QueryBuilder. I have not found anything to fulfill my needs, so I created this small patch to contribute to the community, maybe someone needs this feature too.

An example when this approach can be usefull:
You want some business logic like selecting only some entities based on the environment or something like that. You implemented the EntityRepository for these entities to aware this constraints, but an awful lot of external modules wants to override the where constraints by not using the andWhere() method instead of where() on the QueryBuilder instance, throwing out your carefull filtering, causing confusions with the users. With deferred Critetia, you can define filtering which can not be getting around by external modules of your application.

### DIFF
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -145,6 +145,13 @@ class QueryBuilder
     protected $lifetime = 0;
 
     /**
+     * Array of deferred criteria
+     *
+     * @var array<Criteria>
+     */
+    protected $deferredCriteria = [];
+
+    /**
      * Initializes a new <tt>QueryBuilder</tt> that uses the given <tt>EntityManager</tt>.
      *
      * @param EntityManagerInterface $em The EntityManager to use.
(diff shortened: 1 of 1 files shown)


lcobucci:
@thomasklein94 thanks for sharing your ideas. I have mixed feelings here, though.

I'm not sure if this should be part of the ORM... I mean, the fact that you have modules overriding filters that were previously defined sounds more like design issues on your application.

Let's wait to hear from other maintainers (@Ocramius @Majkl578 @alcaeus) and then we can do a more in-depth review of the code.

Ocramius:
> but an awful lot of external modules wants to override the where constraints by not using the andWhere() method instead of where() on the QueryBuilder instance, throwing out your carefull filtering, causing confusions with the users. 

This is exactly why custom repositories should be used in different modules (or even simple query methods, segregated each in their own class).

I don't think that adding complexity on the query builder aids with customisation problems. In fact, it will probably even worsen things.

As a rule of thumb, do not share a query builder across two classes: this applies both to inheritance and composition.

thomasklein94:
Hi,
I can see why you think this is a flaw, originated from the application i'm currently developing, but as I mentioned, not just our Bundles using the QueryBuilder - lets say - in a sily way, but lot of 3rd party Bundles as well. 

But let put this aside for a moment. Let's say you want Soft

thomasklein94:
Hi,
I can see why you think this is a flaw in my application, but i still think this is a small changset helping a simplifying application logic. Let me give you an other, more specific example.

Lets filter our entities based on its 'deletedAt' field. As stated in DDC-3708 (Issue #4543 ), it can't be done (easily) with SqlFilter, as it is more like a business logic task, and SqlFilter operates on the data layer. 
So let's move the filtering to the application layer. The first thing that came in the mind of almost everyone is that we can put this filtering in the Entity's Repository class. It may sounds good, but when using with other Bundles, lots of them ask the repository to create a QueryBuilder for the given Entity, and later using that like it have not contain any filtering, Basicly throwing out your Criteria, and often result in an Exception, because the parameters was not reseted and there is a mismatch in used and given parameters. Yes, this is the other module's fault, but its really hard to fix every third party module, but more easy to adopt this deferred criteria feature. 
I'm aware that there is a donctrine extension for soft deletes, i only used this as an easy example (in my inherited application, there is an even bigger and complicated problem, and i don't want to share that problems with you, sorry).

Issue #6786, opened 2017-10-22
How to avoid updating of the referenced entity, when the owning side gets changed?


automatix:
Hello!

I have two entities: `Foo` and `Bar` with a unidirectional `ManyToOne` relationship between them (`Bar` has `Foo`).

![foo-bar-buz_model](https://user-images.githubusercontent.com/440005/31862924-bd25caf2-b746-11e7-84a4-8ce521ff3c59.png)

When I try to re-/unset `Bar#foo`, Doctrine tries not only change the reference (change/reomve the value in the column `bar.foo_name`), but also to `UPDATE` the value of `foo.name`.

In my case it leads to an `Integrity constraint violation` error. But even if not -- it's not the expected behavior.

Is it a bug or a layer-8-issue? :)

Ocramius:
Requires test case, see https://github.com/doctrine/doctrine2/tree/master/tests/Doctrine/Tests/ORM/Functional/Ticket for examples

automatix:
Thank you for your reply. I'm trying to write a test case, but have a problem with the schema creation from meta data. Just posted it [here](https://stackoverflow.com/questions/46936753/class-is-not-a-valid-entity-error-for-entitymanagergetclassmetadata). Do you / Does anyone here have an idea, what's going wrong there?

automatix:
I wrote a test and have just opened a [pull request](https://github.com/doctrine/doctrine2/pull/6796). The test is working, so it won't be a Doctrine issue. I'm closing this issue. But, anyway, if you have an idea, what it might be caused by, it would be nice, if you share your thoughts about it. Thanks.

PR #6796, opened 2017-10-26
Test for the issue #6786

automatix:
The test is working, so the issue seems not to be a Doctrine issue.


lcobucci:
@automatix closing as `invalid` since the test is passing.

PR #6801, opened 2017-10-30
Improve misleading ORMInvalidArgumentException message

jonny-no1:
See https://github.com/doctrine/doctrine2/issues/6029#issuecomment-340304346.

### DIFF
--- a/lib/Doctrine/ORM/ORMInvalidArgumentException.php
+++ b/lib/Doctrine/ORM/ORMInvalidArgumentException.php
@@ -216,11 +216,7 @@ class ORMInvalidArgumentException extends \InvalidArgumentException
      */
     public static function invalidAssociation(ClassMetadata $targetClass, $assoc, $actualValue)
     {
-        $expectedType = 'Doctrine\Common\Collections\Collection|array';
-
-        if (($assoc['type'] & ClassMetadata::TO_ONE) > 0) {
-            $expectedType = $targetClass->getName();
-        }
+        $expectedType = $targetClass->getName();
 
         return new self(sprintf(
             'Expected value of type "%s" for association field "%s#$%s", got "%s" instead.',
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6029Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6029Test.php
@@ -1,0 +1,137 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ORM\ORMInvalidArgumentException;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH6029Test extends OrmFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function setUp()
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema(
+            [
+                GH6029User::class,
+                GH6029Group::class,
+                GH6029Group2::class,
+                GH6029Product::class,
+                GH6029Feature::class,
+            ]
+        );
+    }
+
+    /**
+     * Verifies that when wrong entity is persisted via relationship field, the error message does not correctly state
+     * the expected class name.
+     *
+     * @group 6029
+     */
+    public function testManyToManyAssociation() : void
+    {
+        $user = new GH6029User();
+        $user->groups->add(new GH6029Group2());
+
+        $this->expectException(ORMInvalidArgumentException::class);
+        $this->expectExceptionMessage(
+            sprintf(
+                'Expected value of type "%s" for association field "%s#$groups", got "%s" instead.',
+                GH6029Group::class,
+                GH6029User::class,
+                GH6029Group2::class
+            )
+        );
+
+        $this->_em->persist($user);
+        $this->_em->flush();
+    }
+
+    /**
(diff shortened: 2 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6029Test.php
```diff
@@ -0,0 +1,83 @@
...
+/** @Entity */
+class DDC6029Group implements DDC6029GroupInterface
+{
+    /** @Id @Column(type="integer") @GeneratedValue */
+    public $id;
+
+    /**
+     * @ManyToMany(targetEntity="DDC6029User", mappedBy="groups")
+     */
+    private $users;
```
lcobucci: Is this needed? Let's keep only the things required by the tests

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6029Test.php
```diff
@@ -0,0 +1,83 @@
...
+        ]);
+    }
+
+    /**
+     * Verifies that when wrong entity is persisted via relationship field, the error message does not correctly state
+     * the expected class name.
+     */
+    public function testIssue()
+    {
+        $user = new DDC6029User();
```
lcobucci: Shouldn't we add here the expected exception and the expected exception message? Like:

```php
$this->expectException(ORMInvalidArgumentException::class);
$this->expectExceptionMessage('Expected value of type "Doctrine\Tests\ORM\Functional\Ticket\DDC6029Group" for association field "Doctrine\Tests\ORM\Functional\Ticket\DDC6029User#$groups", got "Doctrine\Tests\ORM\Functional\Ticket\DDC6029Group2" instead.');
```

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6029Test.php
```diff
@@ -0,0 +1,83 @@
...
+            $this->_em->getClassMetadata(DDC6029Group::class),
+            $this->_em->getClassMetadata(DDC6029Group2::class),
+        ]);
+    }
+
+    /**
+     * Verifies that when wrong entity is persisted via relationship field, the error message does not correctly state
+     * the expected class name.
+     */
+    public function testIssue()
```
lcobucci: ` : void`

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6029Test.php
```diff
@@ -0,0 +1,83 @@
...
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(DDC6029User::class),
+            $this->_em->getClassMetadata(DDC6029Group::class),
+            $this->_em->getClassMetadata(DDC6029Group2::class),
+        ]);
+    }
+
+    /**
+     * Verifies that when wrong entity is persisted via relationship field, the error message does not correctly state
+     * the expected class name.
```
lcobucci: Please add `@group` with the ticket number so that we can trace things back

lcobucci (changes requested):

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6029Test.php
```diff
@@ -0,0 +1,79 @@
...
+     */
+    public function testIssue(): void
+    {
+        $user = new DDC6029User();
+        $user->addGroup(new DDC6029Group2());
+
+        self::expectException(ORMInvalidArgumentException::class);
```
lcobucci: This method is not static, so we should use `$this` instead

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6029Test.php
```diff
@@ -0,0 +1,79 @@
...
+        self::expectExceptionMessage('Expected value of type "Doctrine\Tests\ORM\Functional\Ticket\DDC6029Group" for association field "Doctrine\Tests\ORM\Functional\Ticket\DDC6029User#$groups", got "Doctrine\Tests\ORM\Functional\Ticket\DDC6029Group2" instead.');
```
lcobucci: This method is not static, so we should use `$this` instead

lib/Doctrine/ORM/ORMInvalidArgumentException.php
```diff
@@ -195,11 +195,7 @@ public static function invalidIdentifierBindingEntity()
      */
     public static function invalidAssociation(ClassMetadata $targetClass, $assoc, $actualValue)
     {
-        $expectedType = 'Doctrine\Common\Collections\Collection|array';
-
-        if (($assoc['type'] & ClassMetadata::TO_ONE) > 0) {
```
lcobucci: Is there any explanation on why we had this code there (commit msg)? Maybe @Ocramius can give us some context

lcobucci (changes requested):

lib/Doctrine/ORM/ORMInvalidArgumentException.php
foaly-nr1: No details in the commit message of https://github.com/doctrine/doctrine2/commit/71a6a88de82d23232cdf5377a44804a5f55d29b8 and not sure where I could find details on DDC-3490.

lcobucci (approved):
I've rebased and added a test for one-to-many association as well.

lcobucci:
@foaly-nr1 🚢 thanks for your contribution!

PR #6802, opened 2017-10-31
Is detached collection orphan removal a valid scenario?

pt24:
Hi,

* let's have an entity with collection of subentities, that are owned by entity
* detach it
* remove subentities (orpan removal)
* merge entity

is this a valid scenario? If so, the scenario is failing
If not, how shall I achieve similar behavior?

### DIFF
--- a/tests/Doctrine/Tests/Models/Detach/Member.php
+++ b/tests/Doctrine/Tests/Models/Detach/Member.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Doctrine\Tests\Models\Detach;
+
+/**
+ * @Entity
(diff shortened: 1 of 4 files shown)


lcobucci:
> detach it
> merge entity

@pt24 just don't rely on these features, they were already removed from the [next major release - `v3`](https://github.com/doctrine/doctrine2/pull/1577) - because they do more harm than good.

Majkl578:
Closing as per @lcobucci's comment - detach & merge are being removed from ORM 3.0.

Ocramius:
Errata: they are already removed in `develop` \o/

PR #6805, opened 2017-11-01
[Docs] Corrected NamingStrategy demo code to match example

yesdevnull:
From the [NamingStrategy docs](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/namingstrategy.html#implementing-a-naming-strategy):
> ...all column names should be upper case...

Example paragraph mentions changing column titles to upper case, yet `strtolower` was being used.  Corrected demo code to match the instructions.

### DIFF
--- a/docs/en/reference/namingstrategy.rst
+++ b/docs/en/reference/namingstrategy.rst
@@ -97,7 +97,7 @@ a naming strategy for database tables and columns.
 Implementing a naming strategy
 -------------------------------
 If you have database naming standards, like all table names should be prefixed
-by the application prefix, all column names should be upper case, you can easily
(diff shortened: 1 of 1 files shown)


carlosbuenosvinos:
Thanks for the contribution, @yesdevnull! In this case, I think it probably makes more sense to change the description for than the code. I would say is more typical to have lowercase table names than uppercase names, what do you think?

yesdevnull:
@carlosbuenosvinos likewise :) I will adjust accordingly.

(apologies for very late reply, didn't see the notification for this)

Majkl578 (approved):

Majkl578:
@yesdevnull Thanks!

PR #6811, opened 2017-11-06
Fix syntax error when join unrelated entity with discriminator entity

Deltachaos:
This is the backporting PR for #6812


Majkl578:
Going to close this one, backporting to lower branches is up to the Doctrine maintainers. Let's stay with #6811 only.

Issue #6815, opened 2017-11-08
Discriminator "fieldName" is not set in ClassMetadata


matthias-z:
When a discriminator column is defined with a field name (property "fieldName") the corresponding value in ClassMetadata is set to the column name (property "name") instead.
It is my understanding that this should only occur as a default when no field name is configured.

I observed this behavior using the `AnnotationDriver`:
```php
/**
 * @ORM\Entity
 * @ORm\DiscriminatorColumn(fieldName="discriminatorField")
 */
```
But from what I've seen in the code `XmlDriver` or `YamlDriver` should behave the same in that regard.

Ocramius:
Closing as `invalid` as per https://github.com/doctrine/doctrine2/pull/6817#issuecomment-354667371

PR #6817, opened 2017-11-08
#6815 Pass the dicriminator column's "fieldName" config to ClassMetadata

matthias-z:
Fix for #6815 

I fixed this for Annotation, Xml and Yaml drivers, however the test works for the Annotation driver only, as I failed to get the `OrmFunctionalTestCase` working with the other drivers.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
@@ -250,6 +250,7 @@ class AnnotationDriver extends AbstractAnnotationDriver
                             'name'             => $discrColumnAnnot->name,
                             'type'             => $discrColumnAnnot->type ?: 'string',
                             'length'           => $discrColumnAnnot->length ?: 255,
+                            'fieldName'        => $discrColumnAnnot->fieldName,
                             'columnDefinition' => $discrColumnAnnot->columnDefinition,
                         ]
                     );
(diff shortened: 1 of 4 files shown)


lcobucci:
@matthias-z how did you end up using `fieldName`? It's not even documented: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/annotations-reference.html#discriminatorcolumn

AFAIK that's an internal property used only in non-object hydration and shouldn't be set by users... Maybe @Ocramius or @guilhermeblanco can give us more context here..

matthias-z:
@lcobucci Interesting, I did not notice that it's not in the documentation. I guess I ended up using it because it was suggested by autocompletion in my IDE and the corresponding property in the annotation class actually is somehow documented:
```php
    /**
     * Field name used in non-object hydration (array/scalar).
     *
     * @var mixed
     */
    public $fieldName;
```

It is in fact used for array/scalar hydration to set a custom key for the field in the result array (if it is not set, the column name is used instead).
This is also what I'm using it for primarily.

So as the underlying feature actually works, I was surprised I could not read it from `ClassMetadata`. This is especially strange, since it is taken care of in the corresponding setter method `ClassMetadataInfo::setDiscriminatorColumn`. It simply is not passed to the method by the mapping driver implementations.

Ocramius:
Closing as `invalid`. A discriminator is not a field, so `fieldName` shouldn't ever land in the metadata.

PR #6818, opened 2017-11-08
BasicEntityPersister::count() return type fix

### LINKED ISSUE #6810: BasicEntityPersister::count() return type contract violation in 2.5.x branch
There is a return type contract violation in stable 2.5.x branch for `count()` method in `BasicEntityPersister` class: string returned instead of integer
https://github.com/doctrine/doctrine2/blob/e1b851f2e93a4afd82fb7e7f32c6f114d0ea7b71/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L811-L824

causing incorrect behavior in case like:
```php
/** @var $coll Doctrine\ORM\LazyCriteriaCollection */
if ($coll->count() === 0) {
  // unreachable code
}
```
it seems to be fixed as a collateral change in `master` and needs to be backported somehow:
https://github.com/doctrine/doctrine2/commit/a90035e81af2e9a424c0268a36101c3483e06571#diff-25f5292f65d5536bb733559f15e0d37fR825

vkurdin:
Resolve #6810 by casting return value to integer (`2.5.x` branch)

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -820,7 +820,7 @@ class BasicEntityPersister implements EntityPersister
             ? $this->expandCriteriaParameters($criteria)
             : $this->expandParameters($criteria);
 
-        return $this->conn->executeQuery($sql, $params, $types)->fetchColumn();
+        return (int) $this->conn->executeQuery($sql, $params, $types)->fetchColumn();
     }
 
     /**


lcobucci:
@vkurdin thanks for helping us to backport this fix but we do need to have the tests as well. Maybe cherry-picking the original commit(s) and addressing the the conflicts would be a better solution?

vkurdin:
@lcobucci no conflicts introduced after cherry-picking the original commit to 2.5 branch. On the other side new functionality was added: `EntityRepository::count` method. Is that ok to introduce new feature to 2.5 branch?

lcobucci:
@vkurdin we shouldn't introduce new features on `2.5` but don't we have any test that ensures that we're returning the correct type on that method (not even an `assertSame()`)?

lcobucci:
I've meant tests for `BasicEntityPersister#count()`. If we don't we should probably do something about that on `2.6` as well

lcobucci:
@vkurdin I've found (on `v2.6`) the test that covers this method and, unfortunately, we are using `assertEquals()` and not `assertSame()`. So it's fine to merge this on `2.5` but we should fix `BasicEntityPersisterTypeValueSqlTest#testCountEntities()` on `master` (eventually).

lcobucci (approved):

lcobucci:
@vkurdin 🚢 thanks for your contribution!

Issue #6819, opened 2017-11-09
Pagination LimitSubqueryWalker optimization


plfort:
LimitSubqueryWalker create useless select expressions when "resultVariable" are involved : https://github.com/doctrine/doctrine2/blob/a82f6c5725cb31348eae0b4414e8992b1400ba30/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php#L76-L82

We need to "Preserve mixed data in query for ordering." only if these fields are used in the "order by" clause.

plfort:
The current behavior produces horrible performances when a lot of subselect are performed in the select clause.

mvrhov:
AFAIK the double sort was kept because the PHP would die with core dump. At least this was the case with both 5.3 and 5.4 versions at that time.

plfort:
@mvrhov Sorry but I don't get the point, can you be more specific ?

mvrhov:
I just wanted to tell you, that there is the reason that this was kept.

plfort:
Ok but what do you mean with "postgresql requires this ...".

mvrhov:
That you should run the same query on whatever database you are using and the postgresql.
Most complaints here are from ppl. using MySQL which is not SQL standard conformant.

Ocramius:
Handled in #6820

PR #6820, opened 2017-11-09
#6819 Optimize LimitSubqueryWalker

plfort:
#6819  Don't add useless select expressions when fields are not involved in "order by" clause


guilhermeblanco (approved):
LGTM

PR #6826, opened 2017-11-14
Add Type::GUID to $typeAlias list in EntityGenerator

JKapitein:
Right now, when generating entities, a column of type 'guid' will generate the following PHPDoc:

```
/**
 * Get id
 *
 * @return guid
 */
public function getId() {...}
```

Since guid is not a valid PHP type, this throws a warning in PHPStorm, and I assume most IDEs.

Adding the type to the type alias list fixes the problem.

### DIFF
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -164,6 +164,7 @@ class EntityGenerator
         Type::TEXT          => 'string',
         Type::BLOB          => 'string',
         Type::DECIMAL       => 'string',
+        Type::GUID          => 'string',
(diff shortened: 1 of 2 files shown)


carlosbuenosvinos:
@JKapitein, thanks for your contribution! Would you mine to go a bit deeper and provide a unit test for this change? It would be awesome! Thanks!

JKapitein:
I've added the test. I ran it, and it passed with the type alias set and failed without it.

lcobucci (approved):
I've rebased the branch to sync it with the latest version of `master`. LGTM 👍

lcobucci:
@JKapitein 🚢 it will be available on our next minor release. Thanks for your contribution!

PR #6829, opened 2017-11-17
Fix collection not deleted after two flush

### LINKED ISSUE #6831: The old collection not deleted after a sequence of two flushes
When you in full rewrite a collection for an entity, sometimes an old collection is not removed from the database. 

Example
```
class Contact
{
    /**
     * @Id
     * @Column(type="integer")
     */
    public $id;

    /**
     * @Column(type="string")
     */
    public $name;

    /**
     * @ManyToMany(targetEntity="Source")
     */
    public $sources;
}
```
```
class Source
{
    /**
     * @Id
     * @Column(type="string")
     */
    public $id;
}
```
Origin entity

| id | name |  sources |
|:--:|:--:|:--:|
|1|name 1| cck |

```php
$contact->name = 'name 2';
$em->flush();

$sources = $contact->sources->toArray();
$contact->sources = new ArrayCollection($sources);
$em->flush();
```
After the first flush, the changes for the collection fields are no longer tracked and an [error](https://travis-ci.org/doctrine/doctrine2/jobs/303710005#L612) occurred during the second flush. 
```
SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "gh6829contact_gh6829source_pkey"
DETAIL:  Key (gh6829contact_id, gh6829source_id)=(1, cck) already exists.
```
If you remove the first flush then it works fine. 

```php
$contact->name = 'name 2';
$sources = $contact->sources->toArray();
$contact->sources = new ArrayCollection($sources);
$em->flush();
```

vtsykun:
Fixes #6831

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -622,6 +622,7 @@ class UnitOfWork implements PropertyChangedListener
             if ($class->isCollectionValuedAssociation($name) && $value !== null) {
                 if ($value instanceof PersistentCollection) {
                     if ($value->getOwner() === $entity) {
+                        $actualData[$name] = $value;
                         continue;
                     }
 
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6829Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6829Test.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH6829Test extends OrmFunctionalTestCase
+{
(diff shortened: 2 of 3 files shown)


lcobucci:
@vtsykun I don't really get the point of this test... I mean, why would you want to do that with your object? You simply shouldn't need to overwrite the entire collection, IMO if you want to remove all items from the collection you should rely on `Collection#clear()`.

vtsykun:
Hi @lcobucci 
No. I don't want do. I provided only simple example of failed test. More common use case - usage `PropertyAccessor` in inner api methods.

lcobucci:
@vtsykun what I'm trying to say is that that's not the way the objects were designed to be used, as @Majkl578 explained in https://github.com/doctrine/doctrine2/issues/6831#issuecomment-352160796 the collection should not be overridden.

@Ocramius and @guilhermeblanco were discussing some changes for v3.0 but IMO this shouldn't change on `2.x`.

vtsykun:
Ok, I understood
Thanks

PR #6830, opened 2017-11-19
fix applying column options on foreign key columns

Tobion:
Fixes https://github.com/doctrine/dbal/issues/2811

Since the collation is not a standard option, see https://github.com/doctrine/dbal/pull/2919, it must explicitly be set as platform option on a join column when the referenced column also has a collation set.

PR #6833, opened 2017-11-20
Fix applying target column options on foreign key columns

mrxotey:
Fixes doctrine/dbal#2811
Similar PR: #6830 

Tobion's PR fixes only applying of collation, which is not enough. This fix applies all options from target column to make type of source and target columns equal.

P.S.
Sorry for my terrible english.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -439,19 +439,7 @@ class SchemaTool
             $options['columnDefinition'] = $mapping['columnDefinition'];
         }
 
-        if (isset($mapping['options'])) {
-            $knownOptions = array('comment', 'unsigned', 'fixed', 'default');
-
-            foreach ($knownOptions as $knownOption) {
-                if (array_key_exists($knownOption, $mapping['options'])) {
-                    $options[$knownOption] = $mapping['options'][$knownOption];
-
-                    unset($mapping['options'][$knownOption]);
-                }
-            }
-
-            $options['customSchemaOptions'] = $mapping['options'];
-        }
+        $this->gatherColumnOptions($options, $mapping);
 
         if ($class->isIdGeneratorIdentity() && $class->getIdentifierFieldNames() == array($mapping['fieldName'])) {
             $options['autoincrement'] = true;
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -871,4 +857,21 @@ public function getUpdateSchemaSql(array $classes, $saveMode = false)
 
         return $schemaDiff->toSql($this->platform);
     }
+
+    private function gatherColumnOptions(array &$options, array $mapping)
+    {
+        if (isset($mapping['options'])) {
+            $knownOptions = array('comment', 'unsigned', 'fixed', 'default');
```
Tobion: I don't think it makes sense to copy the column comment from the PK to the foreign key. So that should probably be excluded, but makes it kinda more ugly.

tests/Doctrine/Tests/Mocks/HydratorMockStatement.php
```diff
@@ -34,7 +34,7 @@ public function __construct(array $resultSet)
      *
      * @return array
      */
-    public function fetchAll($fetchStyle = null, $columnIndex = null, array $ctorArgs = null)
+    public function fetchAll($fetchStyle = null, $columnIndex = null, $ctorArgs = null)
```
Tobion: why was all those unrelated stuff changed?

tests/Doctrine/Tests/Models/Forum/ForumCategory.php
```diff
@@ -9,7 +9,7 @@
 class ForumCategory
 {
     /**
-     * @Column(type="integer")
+     * @Column(type="guid", options={"fixed":true, "collation":"latin1_bin", "foo":"bar"})
```
Tobion: guid might not be a string so applying fixed or collation might not make sense in this case. maybe just use a string simply.

Tobion:
Fixing it generically makes sense 👍

Tobion:
@mrxotey ping

lcobucci:
@mrxotey @Tobion as mentioned on our [contributions guidelines](https://github.com/doctrine/doctrine2/blob/master/CONTRIBUTING.md#we-only-accept-prs--to-master) we accept PRs only to `master` and we handle the backporting if needed (and when possible). With that said, please let's apply the fixes on #6830.

lcobucci:
I'll close this PR as `invalid` as per my comment.

mrxotey:
Sorry. Was busy. I'll create new PR with review fixes. Thanks.

PR #6844, opened 2017-11-23
Configure PHPCS

lcobucci:


phpcs.xml.dist
```diff
@@ -0,0 +1,24 @@
...
+
+    <exclude-pattern>*/tests/Doctrine/Tests/Proxies/__CG__/*</exclude-pattern>
+
+    <rule ref="Doctrine" />
+
+    <rule ref="PSR1.Classes.ClassDeclaration.MultipleClasses">
+        <exclude-pattern>*/tests/*</exclude-pattern>
+    </rule>
+</ruleset>
+
```
phansys: Please, remove surplus LF.

phansys (commented):
Status: Needs work.

.travis.yml
```diff
@@ -78,6 +78,11 @@ jobs:
       before_script: wget https://phpbench.github.io/phpbench/phpbench.phar https://phpbench.github.io/phpbench/phpbench.phar.pubkey
       script: php phpbench.phar run -l dots --report=default
 
+    - stage: Coding standard
+      php: nightly
```
stof: I suggest using a stable PHP version instead. It is less likely to break PHPCS itself.

lcobucci: That will be done on `v3` after we fix all CS violations

Issue #6712, opened 2017-09-20
ClosureExpressionVisitor::sortByField() fails when comparing Objects/Dates


t11n:
In method `ClosureExpressionVisitor::sortByField()` we have the following sort function...

```
    return function ($a, $b) use ($name, $next, $orientation) {
        $aValue = ClosureExpressionVisitor::getObjectFieldValue($a, $name);
        $bValue = ClosureExpressionVisitor::getObjectFieldValue($b, $name);

        if ($aValue === $bValue) {
            return $next($a, $b);
        }

        return (($aValue > $bValue) ? 1 : -1) * $orientation;
    };
```

... which does not work with Objects, especially DateTime objects. I do not know if it makes sense to compare objects in general if they are greater or lesser, but it definitely makes sense for DateTime objects.

The problem is, that equal DateTime objects will not be recognised by `$aValue === $bValue`. Instead `$aValue > $bValue` will evaluate to false in these cases. Not a big deal you may think, as the a element is put below the b element, which is fine, as they are equal. That's right, but $next will neither be executed.

In my use case, sorting by date is the first order criteria, sorting by ID is the second. That code will never evaluate the second criteria, though.

Possible fix:
```
        if ($aValue instanceof \DateTime && $aValue == $bValue || $aValue === $bValue) {
            return $next($a, $b);
        }
```

You have to judge whether it makes sense to compare other object types. For DateTimes however, this is necessary, I suppose. Sorting by date with a second order by criteria is not possible otherwise.

Please have a look into the matter.

lcobucci:
@t11n comparing `DateTime` objects is actually fine, but maybe we should be using the space ship operator instead of `($aValue > $bValue) ? 1 : -1`. But regardless of that, could you send a PR with a failing functional test? That would be really helpful!

You can find examples on https://github.com/doctrine/doctrine2/tree/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket

t11n:
@lcobucci Thanks for looking into the issue.

I prepared three tests, that demonstrate the problem: https://github.com/doctrine/doctrine2/pull/6851
(branched from v2.5, since I do not have a php7.1 setup running, yet)

First test shows sorting by scalar values, that works as expected. The other two tests sort by fields, where first ordered by field is a DateTime object and a general stdClass.

[...]

PR #6851, opened 2017-11-27
#6712 Added tests to demonstrate, that sorting by two criterias is un…

t11n:
…reliable, when first field holds objects.

Demonstrates issue https://github.com/doctrine/doctrine2/issues/6712

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6712Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6712Test.php
@@ -1,0 +1,130 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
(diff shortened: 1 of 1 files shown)


lcobucci:
@t11n we REALLY need this to be based on `master`, `v2.5.x` is not being actively maintained.

lcobucci:
I understand that you don't have the setup, but you could use a docker container or a VM to get PHP 7.1 running and ensure that this behaviour also happens on `v2.6`

t11n:
@lcobucci The code is the same in 2.6, so the same happens with this version, unless PHP changed behaviour of the identical operator.

I do not have time to set things up in PHP7. But trust me, it happens there, too..

mpdude:
Is this an ORM or doctrine/collections thing?

PR #6853, opened 2017-11-27
Refactoring tests

carusogabriel:
I've refactored some tests with [`assertFileExists`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFileExists), [`assertNull`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertNull) and [`assertEmpty`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertEmpty) methods.

PR #6854, opened 2017-11-28
Refactoring tests

carusogabriel:
Simple version of #6853, just using `assertNull`, `assertFileExists`, `assertTrue` and `assertFalse`. Also, in `develop` branch.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Cache/FileLockRegionTest.php
+++ b/tests/Doctrine/Tests/ORM/Cache/FileLockRegionTest.php
@@ -80,7 +80,7 @@ class FileLockRegionTest extends AbstractRegionTest
 
         self::assertFileExists($file);
         self::assertInstanceOf(Lock::class, $lock);
-        self::assertEquals($lock->value, file_get_contents($file));
+        self::assertStringEqualsFile($file, $lock->value);
 
         // should be not available after lock
         self::assertFalse($this->region->contains($key));
@@ -102,10 +102,10 @@ class FileLockRegionTest extends AbstractRegionTest
 
         file_put_contents($file, 'foo');
         self::assertFileExists($file);
-        self::assertEquals('foo' , file_get_contents($file));
+        self::assertStringEqualsFile($file, 'foo');
 
         self::assertNull($this->region->lock($key));
-        self::assertEquals('foo' , file_get_contents($file));
+        self::assertStringEqualsFile($file, 'foo');
         self::assertFileExists($file);
 
         // should be not available
@@ -124,17 +124,17 @@ class FileLockRegionTest extends AbstractRegionTest
         self::assertTrue($this->region->contains($key));
 
         self::assertInstanceOf(Lock::class, $lock = $this->region->lock($key));
-        self::assertEquals($lock->value, file_get_contents($file));
+        self::assertStringEqualsFile($file, $lock->value);
         self::assertFileExists($file);
 
         // change the lock
         file_put_contents($file, 'foo');
         self::assertFileExists($file);
-        self::assertEquals('foo' , file_get_contents($file));
+        self::assertStringEqualsFile($file, 'foo');
 
         //try to unlock
         self::assertFalse($this->region->unlock($key, $lock));
-        self::assertEquals('foo' , file_get_contents($file));
+        self::assertStringEqualsFile($file, 'foo');
         self::assertFileExists($file);
 
         // should be not available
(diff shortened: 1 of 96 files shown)


lcobucci (commented):
Awesome, LGTM 👍

lcobucci (approved):

lcobucci:
Since @guilhermeblanco is rebasing `develop` I'll leve this to him.

carusogabriel:
I've improved more tests:
- Instead of `count` PHP method, use [`assertCount`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertCount);
- Avoid use PHP functions `is_*`. Instead, use [`assertInternalType`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertInternalType);
- Use [`assertEquals`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertEquals) instead of comparisions `==` and [`assertSame`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertSame) instead of strict comparasions `===`;
- Use [`assertStringEqualsFile`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertStringEqualsFile) when comparing strings and files;
- Use [`assertTrue`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertTrue) or [`assertFalse`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFalse), instead of comparing `true` or `false` keyworkds with [`assertEquals`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertEquals)/[`assertSame`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertSame);

docs/en/reference/batch-processing.rst
```diff
@@ -174,7 +174,7 @@ problems using the following approach:
         // do stuff with the data in the row, $row[0] is always the object
     
         // detach all entities from Doctrine, so that Garbage-Collection can kick in immediately
-        $this->em->clear();
+        $this->_em->clear();
```
Majkl578: This change is incorrect.

carusogabriel:
I’m trying to rebase, but I guess I lost my work 😭

Majkl578:
You've probably just incorrectly rebased, but your commits still exist: 6999d39 + 60cd9cf.
Hint: Sometimes it's easier to just hard-reset local copy and re-apply the commits again by cherry-picking them.

carusogabriel:
I’m gonna try this, thanks for the tip

carusogabriel:
Should we refactory `assertEquals('', $foo)/assertSame('', $foo)` with `assertEmpty($foo)`, and `assertEquals([], $bar)/assertSame([], $bar)` with `assertEmpty($bar)`?

Ocramius:
Don't think so: `assertEmpty()` is too lax

Majkl578 (approved):
Develop has already been rebased so this can be merged now.

Majkl578:
@carusogabriel Thanks!

carusogabriel:
Any time @Majkl578 :man_technologist:

carnage:
Took a quick look at back porting this one and well, lets just say it's potentially going to be a big job. 

94 out of 96 files have conflicts, a 95th is a deleted/modified conflict.

One of the biggest things I've noticed is there are conflicts on a lot of lines where $this->assert* has been changed to self::assert* (technically, the correct way to call the asserts) however this wasn't done in this PR. My guess is that we're missing a PR somewhere that did all the changes from $this-> to self:: and that finding and merging that one first might reduce the number of conflicts in this PR or at the very least make it possible to see what's going on in the diffs. 

It is of cause possible that that PR will also have a huge number of conflicts, so it might be prudent to redo the work to change $this-> to self:: as it can be automated by php cs fixer.

Alternatively, if it's the project's choice to use $this-> to call the asserts php cs fixer could be run it in the opposite direction against the changes in this PR to make it closer to the current codebase before a merge is attempted.

greg0ire:
I think redoing the change would be best, because you would not catch all occurrences anyway. We use phpcs in this project, maybe we can find a rule to add to our coding standard?

carnage:
This is the rule class for php cs fixer https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php but I can't find an equivalent in php codesniffer. We'd also need to decide which style to support (options being Assert::assertX, static::assertX, self::assertX and $this->assertX with the latter two being the usual choices)

greg0ire:
🤔 I wonder why it is a phpunit-specific rule. My personal preference would be to run it once and make a PR, and pick `self`, because it won't conflict but because if we already merged that in the old-prototype-3.x, then there must be some consensus on using that.

carnage:
In case anyone else comes across this ticket as a TODO, I have started work on it; however there are still ~60 conflicted files to go through so it's going to take a while :)

PR #6858, opened 2017-11-29
Fix preFlush event documentation stating incorrectly that flush can be called safely

egonolieux:


docs/en/reference/events.rst
```diff
@@ -548,8 +548,8 @@ preFlush
 ~~~~~~~~
 
 ``preFlush`` is called at ``EntityManager#flush()`` before
-anything else. ``EntityManager#flush()`` can be called safely
-inside its listeners.
+anything else. ``EntityManager#flush()`` should never be called inside its listeners,
+because it will cause an infinite recursive call.
```
Majkl578: Can we replace _cause an infinite recursive call_ with _result in an undefined behavior_? That doesn't expose implementation and covers other potential issues as well.

egonolieux: Yes, a more general explanation might be better indeed.

lcobucci: Also "should never be called inside its listeners" is misleading because it really depends on the type of listener you're using and how `flush()` is being called.

egonolieux: What do you mean by "the type of the listener"? This is specifically about `preFlush`.

lcobucci: That comment was just wrong, sorry... too sleepy

egonolieux: No problem ;)

Issue #5699, opened 2016-03-07
Iterate undefined offset 0


iskyd:
From my function i return by a native query and iterator : 
`return $q->iterate();`
Now in accordance with documentation : 

```
if ($iterator) {
            while (($row = $iterator->next()) !== false) {
                $var = $row[0];
                var_dump($var);
                $em->clear();
            }
}
```

But after the first execution i get : 

> Undefined offset 0

If i dump $row variable i notice that the first execution i get from $row[0] the second $row[1] and so on...

whatda:
I think you're right, because I noticed the same. For others having this issue: You could try `$iterator->key()` for the currently valid index of `$row`.

But it would be more intuitive if `$row` would already contain the correct value, instead of an array, imo (maybe there are cases idk about yet which make it necessary).

flack:
I just stumbled across this, too: It happens if your hydration mode is not `HYDRATE_OBJECT`, i.e. if you iterate over objects, it works as documented, but for e.g. array results, you'll get this error

flack:
Seems to be the same as #3238. 

@Ocramius is this intended behavior?

lcobucci:
@iskyd @whatda @flack could you please send us a failing test case that reproduces that behaviour? It would help us a lot to identify and fix the issue you're describing.

You can find examples on https://github.com/doctrine/doctrine2/tree/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket

flack:
@lcobucci I can make a unittest if you want, but it's really very simple. The code to reproduce is right there in the ticket description

lcobucci:
@flack so, please, send us PR 👍

Steveb-p:
I've stumbled across this as well with exactly the same situation. `iterate` works alright if you're using object hydration, and at the point when you start using arrays (for me I just needed the ID) `iterate` starts using 0,1,2,3... as keys.

Question is it this intended behavior. For people already going around this issue it will be kind of a BC break.

peter-gribanov:
I solve this problem using the `current()` function. But i agree. This is a very strange behavior.

```php
public function iterate(): iterable
{
    foreach ($this->rep->createQueryBuilder('e')->getQuery()->iterate() as $row) {
        yield current($row);
    }
}
```

beberlei:
This is fixed by moving to the new method `Query::toIterable` in 2.8

PR #6860, opened 2017-11-30
Add failing testcase for #5699

flack:
As per https://github.com/doctrine/doctrine2/issues/5699#issuecomment-347038010


lcobucci:
@flack thanks, I'll be looking at this ASAP.

PR #6867, opened 2017-12-06
Ensure `ObjectManagerAware` and `EntityManagerAware` are correctly handled by UoW load/refresh

Ocramius:
Couple things here:

 * [x] tests fail because `Doctrine\Common\Persistence\ClassMetadata` is no longer an ancestor of the ORM metadata
 * [x] since lazy-loading is now performed on a per-property fashion, the assertion that the `om` and `em` fields are `null` before lazy-loading seems wrong. `getReference` should inject the `EntityManager` instance

@guilhermeblanco feedback needed on the above, then I can continue on this :-)

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Doctrine\ORM;
 
 use Doctrine\Common\EventManager;
-use Doctrine\Common\Persistence\ObjectManagerAware;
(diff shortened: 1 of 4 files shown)


guilhermeblanco:
@Ocramius that reminded me that I brought several `Doctrine\Common\Persistence` classes to ORM because they would be incompatible with new APIs.

We can safely remove the `ObjectManagerAware` and use the `EntityManagerAware` only... =)
That would allow us to remove the ClassMetadata question you asked.

Ocramius:
:ship:

PR #6869, opened 2017-12-08
Added deprecation warnings for 2.x

Majkl578:
Based on UPGRADE document for 3.0.

Deprecations added for:
* EntityManager::copy() method
* EntityManager::merge() method
* EntityManager::detach() method
* EntityManager::flush() calls with arguments
* YAML mapping driver
* Version class
* Proxy interface
* Configuration::getProxy*()
* ProxyFactory
* proxy Autoloader

I have not added any notes about mapping/metadata changes here because it's still pretty unstable.

Tests marked as `@legacy` use any of the deprecated methods.
Added explicit deprecation tests for deprecated methods.

~I'd like to get these into 2.6, there is no reason to delay it.~

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,10 @@
+# Upgrade to 2.7
+
+## Deprecated: `Doctrine\ORM\EntityManagerInterface#copy()`
+
+Method `Doctrine\ORM\EntityManagerInterface#copy()` never got its implementation and is deprecated.
+It will be removed in 3.0.
+
 # Upgrade to 2.6
 
 ## Minor BC BREAK: `Doctrine\ORM\Tools\Console\ConsoleRunner` is now final
@@ -29,11 +36,6 @@ As a consequence, automatic cache setup in Doctrine\ORM\Tools\Setup::create*Conf
 - APCu extension (ext-apcu) will now be used instead of abandoned APC (ext-apc).
 - Memcached extension (ext-memcached) will be used instead of obsolete Memcache (ext-memcache).
 - XCache support was dropped as it doesn't work with PHP 7.
-
-## Deprecated: `Doctrine\ORM\EntityManagerInterface#copy()`
-
-Method `Doctrine\ORM\EntityManagerInterface#copy()` never got its implementation and is deprecated.
-It will be removed in 3.0.
 
 # Upgrade to 2.5
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,103 @@
 # Upgrade to 2.7
 
+## Deprecated code generators and related console commands
+ 
+These console commands have been deprecated:
+
+ * `orm:convert-mapping`
+ * `orm:generate:entities`
+ * `orm:generate-repositories`
+
+These classes have been deprecated:
+
+ * `Doctrine\ORM\Tools\EntityGenerator`
+ * `Doctrine\ORM\Tools\EntityRepositoryGenerator`
+
+Whole Doctrine\ORM\Tools\Export namespace with all its members have been deprecated as well.
+
+## Deprecated `Doctrine\ORM\Proxy\Proxy` marker interface
+
+Proxy objects in Doctrine 3.0 will no longer implement `Doctrine\ORM\Proxy\Proxy` nor
+`Doctrine\Common\Persistence\Proxy`: instead, they implement
+`ProxyManager\Proxy\GhostObjectInterface`.
+
+These related classes have been deprecated:
+
+ * `Doctrine\ORM\Proxy\ProxyFactory`
+ * `Doctrine\ORM\Proxy\Autoloader` - we suggest using the composer autoloader instead
+ 
+These methods have been deprecated:
+
+ * `Doctrine\ORM\Configuration#getAutoGenerateProxyClasses()`
+ * `Doctrine\ORM\Configuration#getProxyDir()`
+ * `Doctrine\ORM\Configuration#getProxyNamespace()`
+
+## Deprecated `Doctrine\ORM\Version`
+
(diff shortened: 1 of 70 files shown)


composer.json
```diff
@@ -31,7 +31,8 @@
         "doctrine/coding-standard": "^1.0",
         "phpunit/phpunit": "^6.0",
         "squizlabs/php_codesniffer": "^3.1",
-        "symfony/yaml": "~3.4|~4.0"
+        "symfony/phpunit-bridge": "^3.3|^4.0",
+        "symfony/yaml": "4.0.1"
```
Majkl578: This is just temporary hack to work around symfony/symfony#25405 on CI.

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -353,6 +353,13 @@ public function createQueryBuilder()
      */
     public function flush($entity = null)
     {
+        if (func_num_args() === 1) {
```
greg0ire: Consider using > 0

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -652,9 +659,13 @@ public function refresh($entity)
      * @return void
      *
      * @throws ORMInvalidArgumentException
+     *
+     * @deprecated This method is deprecated and will be removed in Doctrine 3.0.
      */
     public function detach($entity)
     {
+        @trigger_error('Method ' . __METHOD__ . '() is deprecated and will be removed in Doctrine 3.0, please migrate to ' . self::class . '::clear().', E_USER_DEPRECATED);
```
greg0ire: Linebreak here

lib/Doctrine/ORM/Mapping/Driver/YamlDriver.php
```diff
@@ -44,6 +44,8 @@ class YamlDriver extends FileDriver
      */
     public function __construct($locator, $fileExtension = self::DEFAULT_FILE_EXTENSION)
     {
+        @trigger_error('YAML mapping driver is deprecated and will be removed in Doctrine 3.0, please migrate to annotation or XML driver.', E_USER_DEPRECATED);
```
greg0ire: Linebreak?

lib/Doctrine/ORM/EntityManager.php
lcobucci: Isn't `$entity !== null` simpler? I mean, only the first argument is used anyway

lib/Doctrine/ORM/Proxy/Proxy.php
```diff
@@ -26,6 +26,8 @@
  *
  * @author Roman Borschel <roman@code-factory.org>
  * @since 2.0
+ *
+ * @deprecated This class is deprecated and will be removed in Doctrine 3.0, proxies will no longer implement it.
```
lcobucci: Does it make sense to mention which interface should be used? So that people can already start preparing their stuff

lib/Doctrine/ORM/EntityManager.php
lcobucci: Also, `EntityManager#clear()` has a different behaviour... and I'm not even sure if we should point people to do partial cleaning of the UoW, it might have the very same issues we have with `EntityManager#detach()`.

lcobucci (changes requested):
LGTM, just some minor comments.

We just need to keep in mind that our plan is to release 2.6 WITHOUT the deprecation notices. 2.7 would be the one adding them.

lib/Doctrine/ORM/EntityManager.php
greg0ire: Yeah even better

Majkl578: That would still allow `flush(null)` though. I'll change it to `!== 0`.

lib/Doctrine/ORM/Proxy/Proxy.php
Majkl578: Not sure about this one, we would be referencing 3rd party code (ProxyManager) which is not required in Composer.

Majkl578:
> We just need to keep in mind that our plan is to release 2.6 WITHOUT the deprecation notices.

Well, the sooner we ship them, the better for users, but if there are other reasons, 2.7 is fine by me.

Majkl578:
Updated, added deprecations for codegen (6870).

Majkl578:
I've also added deprecatons for:
* Configuration::getProxy*()
* ProxyFactory
* proxy Autoloader

Also partially imported UPGRADE notes from develop and reworded them (no mapping-related changes since it's unfinished).

UPGRADE.md
```diff
@@ -1,3 +1,103 @@
+# Upgrade to 2.7
+
+## BC Break: Deprecated code generators and related console commands
```
greg0ire: Deprecations are not BC-breaks, are they?

lib/Doctrine/ORM/Configuration.php
```diff
@@ -70,6 +70,8 @@ public function setProxyDir($dir)
      * Gets the directory where Doctrine generates any necessary proxy class files.
      *
      * @return string|null
+     *
+     * @deprecated
```
greg0ire: Why not also trigger a deprecation error?

Majkl578: There is no replacement, it would be an error that cannot be muted/fixed, it would be just annoying without benefit.

Majkl578:
Rebased.

hkdobrev:
I think deprecations for named queries should be added as well.

https://github.com/doctrine/doctrine2/blob/master/UPGRADE.md#bc-break-removed-support-for-named-queries-and-named-native-queries

PR #6870, opened 2017-12-08
Remove code generation tools & commands

Majkl578:
Removes entity & repository codegen, drops related console commands. See UPGRADE.
We've been hearing about this change for a while so let's do it for 3.0.

This also removes support for converting between formats which relies on entity generator.

If anyone feels interested, they can adopt this code and maintain it in their own repository under different namespace but let's not keep it in the ORM itself.


guilhermeblanco:
Is this done already? If not, once finished, feel free to merge! =)

Ocramius (approved):
LGTM :+1:

PR #6871, opened 2017-12-09
Introduce EntityRepositoryInterface, ancestor for custom repositories

Majkl578:
Introducing a common ancestor for custom repositories.
Currently there is a slight mess, some pieces of code only specify ObjectRepository which:
a) is insufficient since it's missing Selectable;
b) is incorrect since following code expects EntityRepository.

I've taken the opportunity to also add return types to the interface and default repository. Also  updated some phpDocs - specialized `array` and removed duplicate information.

This change should not affect 99.9% of user code since almost everyone extends EntityRepository, but allows 1% to use custom repositories in fancier way.

TODO:
* [ ] constructor should be part of the interface or made parameterless
* [ ] make Scrutinizer happy

### DIFF
--- a/docs/en/reference/advanced-configuration.rst
+++ b/docs/en/reference/advanced-configuration.rst
@@ -411,7 +411,7 @@ implementations.
 Default Repository (***OPTIONAL***)
 -----------------------------------
 
-Specifies the FQCN of a subclass of the EntityRepository.
+Specifies the FQCN of a class implementing EntityRepositoryInterface.
 That will be available for all entities without a custom repository class.
 
 .. code-block:: php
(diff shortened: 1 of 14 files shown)


guilhermeblanco:
This interface should be part of `Doctrine\Persistence` module, and also we do not suffix with `Interface`

Majkl578:
We already have ObjectRepository as a base interface. This interface extends it with ORM-specific methods like createQueryBuilder or by adding Criteria support. I don't think those are methods that'd belong to base persistence layer, i.e. ODM doesn't support Criteria.
Suffix is there because EntityRepository is already taken by default imlementation.

PR #6880, opened 2017-12-12
[Parameter] Deprecate names starting or ending with a colon

fancyweb:
As requested by @Majkl578 in https://github.com/doctrine/doctrine2/pull/5996

I wasn't able to make a test as I don't know how to test a deprecation message without the @expectedDeprecation annotation of the Symfony PHPUnit Bridge ^^

### DIFF
--- a/lib/Doctrine/ORM/Query/Parameter.php
+++ b/lib/Doctrine/ORM/Query/Parameter.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\ORM\Query;
 
+use function preg_match;
+use function trigger_error;
 use function trim;
 
 /**
@@ -67,6 +69,10 @@ class Parameter
      */
     public function __construct($name, $value, $type = null)
     {
+        if (preg_match('/^:|:$/', $name)) {
+            @trigger_error('Starting or ending a parameter name with ":" is deprecated since 2.7 and will cause an error in 3.0', E_USER_DEPRECATED);
+        }
+
         $this->name          = trim($name, ':');
         $this->typeSpecified = $type !== null;
 
--- a/tests/Doctrine/Tests/ORM/Query/ParameterTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/ParameterTest.php
@@ -1,0 +1,29 @@
+<?php
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Query;
+
+use Doctrine\ORM\Query\Parameter;
+use Doctrine\Tests\DoctrineTestCase;
+use Doctrine\Tests\VerifyDeprecations;
+
+final class ParameterTest extends DoctrineTestCase
+{
+    use VerifyDeprecations;
+
+    /**
+     * @test
+     * @group GH6880
+     */
+    public function deprecationMustBeTriggeredWhenUsingColonInParameterNames() : void
+    {
+        $this->expectDeprecationMessage('Starting or ending a parameter name with ":" is deprecated since 2.7 and will cause an error in 3.0');
+        new Parameter(':user_name', 'Testing');
+
+        $this->expectDeprecationMessage('Starting or ending a parameter name with ":" is deprecated since 2.7 and will cause an error in 3.0');
+        new Parameter('user_name:', 'Testing');
+
+        $this->expectDeprecationMessage('Starting or ending a parameter name with ":" is deprecated since 2.7 and will cause an error in 3.0');
+        new Parameter(':user_name:', 'Testing');
+    }
+}


Ocramius:
Requires a test case

fancyweb:
@Ocramius As I said in my comment I need help for this test.

coudenysj:
@fancyweb Just construct the class with a faulty string, and watch what happens.

> By default, PHPUnit will install an error handler that converts the following errors to exceptions: https://phpunit.de/manual/current/en/appendixes.configuration.html

fancyweb:
I tried that and it was just marked as a risky test since there was no assertion. No exception was thrown.

Majkl578:
Please wait until #6869 is merged (after 2.6 is released), it will add the infrastructure for testing deprecations.

lcobucci:
Just missing the test so I'll keep on 2.7.0 and we can easily test it with things added in #7901

lcobucci:
@fancyweb thanks, I've updated things to add the test. I'm just struggling on explaining the value of this deprecation in `UPGRADING.md`. I'll move this 2.8.0 so we have enough time to discuss this :+1:

SenseException:
Thanks for adding the tests.

> I'm just struggling on explaining the value of this deprecation in `UPGRADING.md`

Maybe describe it as a clear information about how a parameter has to look like and being less ambiguous?

lcobucci:
@SenseException focusing on the ambiguity is a good idea. I guess what made me hesitate was also doing the check using regex. We could do the trim and then compare with the original value, sending the deprecation notice when they are different.

lib/Doctrine/ORM/Query/Parameter.php
```diff
@@ -67,6 +69,10 @@ class Parameter
      */
     public function __construct($name, $value, $type = null)
     {
+        if (preg_match('/^:|:$/', $name)) {
+            @trigger_error('Starting or ending a parameter name with ":" is deprecated since 2.7 and will cause an error in 3.0', E_USER_DEPRECATED);
```
greg0ire: Let's add the package name and use some linebreaks:
```suggestion
            @trigger_error(
                'Starting or ending a parameter name with ":" is deprecated since Doctrine ORM 2.7 and will cause an error in 3.0',
                E_USER_DEPRECATED
            );
```

tests/Doctrine/Tests/ORM/Query/ParameterTest.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    use VerifyDeprecations;
+
+    /**
+     * @test
+     * @group GH6880
+     */
+    public function deprecationMustBeTriggeredWhenUsingColonInParameterNames() : void
+    {
+        $this->expectDeprecationMessage('Starting or ending a parameter name with ":" is deprecated since 2.7 and will cause an error in 3.0');
```
greg0ire:
```suggestion
        $this->expectDeprecationMessage('Starting or ending a parameter name with ":" is deprecated since Doctrine ORM 2.7 and will cause an error in 3.0');
```

tests/Doctrine/Tests/ORM/Query/ParameterTest.php
```diff
@@ -0,0 +1,29 @@
...
+        new Parameter(':user_name', 'Testing');
+
+        $this->expectDeprecationMessage('Starting or ending a parameter name with ":" is deprecated since 2.7 and will cause an error in 3.0');
```
greg0ire:
```suggestion
        $this->expectDeprecationMessage('Starting or ending a parameter name with ":" is deprecated since Doctrine ORM 2.7 and will cause an error in 3.0');
```

tests/Doctrine/Tests/ORM/Query/ParameterTest.php
```diff
@@ -0,0 +1,29 @@
...
+        new Parameter('user_name:', 'Testing');
+
+        $this->expectDeprecationMessage('Starting or ending a parameter name with ":" is deprecated since 2.7 and will cause an error in 3.0');
```
greg0ire:
```suggestion
        $this->expectDeprecationMessage('Starting or ending a parameter name with ":" is deprecated since Doctrine ORM 2.7 and will cause an error in 3.0');
```

tests/Doctrine/Tests/ORM/Query/ParameterTest.php
lcobucci: It should be 2.8 too :)

SenseException (changes requested):

derrabus:
Can we pick this up for 2.13?

PR #6885, opened 2017-12-13
Add throw Annotation

LauLaman:
The AbstractHydrator:: hydrateAllData() is missing the `throws` annotation.

the SingleScalarHydrator can throw  `NoResultException` and `NonUniqueResultException`


Majkl578:
Hmm, technically, adding this to the abstract method could be considered as a BC break...

Ocramius:
Not a BC break as this is only for documentation. Still, patch is outdated due to https://github.com/doctrine/doctrine2/pull/6743 being merged now, so I'm going to close this for now.

PR #6886, opened 2017-12-13
Deprecate ClassMetadataInfo

greg0ire:
The deprecation happened 7 years ago: https://github.com/doctrine/doctrine2/pull/249

Here is how I proceeded:

1. I moved all the code back from `ClassMetadataInfo` to `ClassMetadata`
2. I aliased `ClassMetadataInfo` to `ClassMetadata`.
3. I kept a `ClassMetadataInfo` class definition in an `if (\false)` statement to provide IDE autocompletion.


tests/Doctrine/Tests/ORM/Mapping/ClassMetadataInfoTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+class ClassMetadataInfoTest extends TestCase
+{
+    /**
+     * @group legacy
+     * @expectedDeprecation Doctrine\ORM\Mapping\ClassMetadataInfo is deprecated since 2.x and will be removed in 3.0. Use Doctrine\ORM\Mapping\ClassMetadata instead.
+     */
+    public function testTheClassIsDeprecated()
+    {
+        class_exists(ClassMetadataInfo::class);
```
Majkl578: Should at least has assertTrue(). Maybe instantiating ReflectionClass would be better as it would throw i.e. once CMI BC is dropped.

tests/Doctrine/Tests/ORM/Mapping/ClassMetadataInfoTest.php
```diff
@@ -0,0 +1,32 @@
...
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use PHPUnit\Framework\TestCase;
+
+class ClassMetadataInfoTest extends TestCase
+{
+    /**
+     * @group legacy
+     * @expectedDeprecation Doctrine\ORM\Mapping\ClassMetadataInfo is deprecated since 2.x and will be removed in 3.0. Use Doctrine\ORM\Mapping\ClassMetadata instead.
+     */
+    public function testTheClassIsDeprecated()
```
Majkl578: return types here and below please

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -19,3381 +19,16 @@
...
-use InvalidArgumentException;
-use RuntimeException;
-use Doctrine\DBAL\Types\Type;
-use Doctrine\DBAL\Platforms\AbstractPlatform;
-use ReflectionClass;
-use Doctrine\Common\Persistence\Mapping\ClassMetadata;
-use Doctrine\Common\ClassLoader;
-use Doctrine\ORM\Cache\CacheException;
+@trigger_error(sprintf(
+    '%s is deprecated since 2.x and will be removed in 3.0. Use %s instead.',
```
Majkl578: Note: Revisit message later, it may be ComponentMetadata instead of ClassMetadata.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -19,3381 +19,16 @@
...
-     * Gets the type of a field.
-     *
-     * @param string $fieldName
-     *
-     * @return \Doctrine\DBAL\Types\Type|string|null
-     *
-     * @todo 3.0 Remove this. PersisterHelper should fix it somehow
-     */
-    public function getTypeOfField($fieldName)
-
```
Majkl578: Why do we need this? Should also use ::class.

greg0ire: We don't: it says "remove this"

tests/Doctrine/Tests/ORM/Mapping/ClassMetadataInfoTest.php
greg0ire: Once CMI BC is dropped, the class will no longer exist, and this will throw too. Even without the `assertTrue`, right? I mean there is the `@expectedDeprecation`, and it will no longer be satisfied, I don't know why you want me to add `assertTrue`. I'll add it but I don't understand what the point is.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Majkl578: This comment was for:
```php
class_exists('Doctrine\ORM\Mapping\ClassMetadata');
```
GitHub somehow put it on wrong line, maybe too large diff...

tests/Doctrine/Tests/ORM/Mapping/ClassMetadataInfoTest.php
Majkl578: Doesn't matter really, just thought it may be easier to see failed assertion first than _no deprecation thrown_ or whatever that message looks like.

Majkl578:
Also please add note to UPGRADE document, thanks.

greg0ire:
Triggering the build again

greg0ire:
Rebased

lcobucci:
@greg0ire the destination branch should be `2.7`

greg0ire:
@lcobucci fixed :)

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
```diff
@@ -253,12 +253,12 @@ private function getColumnAliasMap($className, $mode, array $customRenameColumns
     /**
      * Adds the mappings of the results of native SQL queries to the result set.
      *
-     * @param ClassMetadataInfo $class
+     * @param ClassMetadata $class
```
Majkl578: The comments should be re-aligned everywhere.

Majkl578 (commented):
LGTM, we just need to revisit the message later and consider adding forward-compatible alias to ComponentMetadata.

greg0ire:
@Majkl578 I realigned the comments… with a tool… which I won't mention :trollface:

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -19,3381 +19,19 @@
...
```
alcaeus: PhpStorm interprets `class_alias`, and I hope other IDEs do as well. There's no need to declare the class like this.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -19,3381 +19,10 @@
...
```
SenseException: With `class_alias()`, `ClassMetadataInfo` is shown as unknown class in my PHPStorm IDE. I guess we can live with that since we don't code for IDEs and an error will be shown in the IDE telling us, that something important changed.

SenseException (commented):
You missed one `ClassMetadataInfo` in `lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php`

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
SenseException: Hm, it took the deleted side of the diff instead of the one with the class_alias. Sorry.

Majkl578:
Just realized - we can't use `class_alias` in the file and rely on autoloading - will not work with Composer's authoritative classmap. :(
This would require an explicit autoload in composer.json then:
```json
"autoload": {
    "files": [
        "lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php"
    ]
}

**Comment by greg0ire** [comment]
I just fixed it by restoring the dead code block (I added an extra comment to explain).
How to test I'm not bullshitting?

```
cd /tmp
mkdir whatever
cd whatever
cat > composer.json <<'EOT'
{
    "repositories": [{
        "type": "vcs",
        "url": "https://github.com/greg0ire/doctrine2"

    }],
    "require": {
        "doctrine/orm": "dev-deprecate_classmetadatainfo"
    }
}
EOT
cat > test.php <<'EOT'
<?php

require 'vendor/autoload.php';

var_dump(class_exists(ClassMetadataInfo::class));
EOT
composer install -a
php test.php
```

**Comment by greg0ire** [comment]
Switched to a centralized class deprecation system (@Majkl578 's idea, I like it, do you?)

**Comment by SenseException** [comment]
@greg0ire Will it stay in the project or was this created just for `ClassMetadataInfo`? It's currently one affected class, but could be used for coming deprecations, if this was your intention.

**Comment by greg0ire** [comment]
I think it will stay in doctrine2 because it's hard to externalize without adding overhead (see [twitter thread](https://twitter.com/greg0ire/status/952188793792139265)), and yeah, the goal is to use it for more classes.

File: composer.json [hunk 8]
```diff
@@ -30,13 +30,15 @@
...
+        "symfony/phpunit-bridge": "^4.0",
         "symfony/yaml": "~3.4|~4.0"
     },
     "suggest": {
         "symfony/yaml": "If you want to use YAML Metadata Mapping Driver"
     },
     "autoload": {
-        "psr-4": { "Doctrine\\ORM\\": "lib/Doctrine/ORM" }
+        "psr-4": { "Doctrine\\ORM\\": "lib/Doctrine/ORM" },
+        "files": ["deprecating_autoloader.php"]
```
nicolas-grekas: the commit introducing this line and the corresponding logic should be removed: it couples Doctrine to Composer, making it not-PSR4 compliant anymore. Furthermore, adding a dedicated autoloaded for Doctrine's deprecation just doens't scale. Imagine if all libs were to do so? No way on my side.

greg0ire:
Moved the controversial autoloading to another PR, so that this is easy to merge.

greg0ire:
Don't merge, this crashes:

```
<?php

require 'vendor/autoload.php';

class A
{
    public function __construct(ClassMetadataInfo $a)
    {
    }
}

new A(new ClassMetadata('whatever'));
```

greg0ire:
Forgot to say that it is now "fixed", but the fix is a bit ugly

lcobucci:
@greg0ire moving this to 2.8.0 as we have conflicts to address and want to release 2.7.0 ASAP.

PR #6887, opened 2017-12-13
Use stable packages

greg0ire:
This will make it easier to contribute. It will also make reviews safer
since you will be able to tell if a PR is reponsible for a broken build
or not. Catching errors early is still possible by having a look at
allowed failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,8 +56,10 @@ jobs:
       install: travis_retry composer update --prefer-dist --prefer-lowest
 
     - stage: Test
-      env: DB=sqlite DEPENDENCIES=stable
-      install: travis_retry composer update --prefer-dist --prefer-stable
+      env: DB=sqlite STABILITY=dev
+      install:
+        - composer config minimum-stability $STABILITY
+        - travis_retry composer update --prefer-dist
 
     - stage: Test
       env: DB=sqlite COVERAGE
@@ -91,6 +93,7 @@ jobs:
 
   allow_failures:
     - php: nightly
+    - env: DB=sqlite STABILITY=dev
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,9 +56,10 @@ jobs:
       install: travis_retry composer update --prefer-dist --prefer-lowest
 
     - stage: Test
-      env: DB=sqlite STABILITY=dev
+      if: type = cron
+      env: DB=sqlite
       install:
-        - composer config minimum-stability $STABILITY
+        - composer config minimum-stability dev
         - travis_retry composer update --prefer-dist
 
     - stage: Test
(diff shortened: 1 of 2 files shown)


Majkl578:
👍 See latest 8 PRs for reference - Symfony introduced a bug in YAML and broke our test suite (already fixed for next stable: symfony/symfony#25405).
Testing against dev dependencies is highly unreliable and non-deterministic.

greg0ire:
Why does Scrutinizer fail exactly?

Majkl578:
According to documentation:
>```
> # Code Coverage decreased from previous inspection
> - 'project.metric_change("scrutinizer.test_coverage", < 0)'
> ```

Looks like some sort of a rounding bug to me.

lcobucci:
@greg0ire can you please rebase this branch to get the latest changes? TBH I like to run everything on with dev dependencies but I understand the frustration when you're trying to collaborate to a project and stuff doesn't work because weird stuff on dependencies. @Ocramius @guilhermeblanco any take on this?

BTW I've configured a cron job on travis to run the build on `master` daily, that should also help us to catch issues before of contributors.

lcobucci:
> Why does Scrutinizer fail exactly?

I've reported that, let's see what they say =)

Majkl578:
@lcobucci Can we maybe add [conditional stage](https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions) that will allow failure (or wouldn't execute) if the build is PR? So it would fail for master (and cron build), not for PR.

greg0ire:
Something is fishy, the build is entirely passing now. But unstable packages are still installed, so maybe the bug was fixed upstream?

greg0ire:
Hmmm... it's supposed to have been fixed 6 days ago, so prior to this bug. I don't understand this mystery. Maybe there was another bug in the meantime, fixed too?

greg0ire:
Or maybe is it the latest changes @lcobucci ?

greg0ire:
Implemented @Majkl578 's proposal: the unstable build will "only" run for the `cron` build, which should be a bit more noticeable.

.travis.yml
```diff
@@ -56,8 +56,18 @@ jobs:
...
     - stage: Test
-      env: DB=sqlite DEPENDENCIES=stable
-      install: travis_retry composer update --prefer-dist --prefer-stable
+      if: type = cron
+      env: DB=sqlite
+      install:
+        - composer config minimum-stability dev
+        - travis_retry composer update --prefer-dist
+
+    - stage: Lint
```
Majkl578: Looks like incorrect rebase, this shouldn't be here.

greg0ire: Indeed, sorry

lcobucci (approved):
LGTM 👍

@Majkl578 do you have any comment?

Majkl578 (approved):
LGTM after recent changes. Maybe just add env DEV_DEPENDENCIES to distinguish the stage on overview in Travis.

greg0ire:
@Majkl578 done

lcobucci:
@greg0ire thanks!

shipit

PR #6888, opened 2017-12-14
Yoda conditions abolished should be

Majkl578:
On this special day of Star Wars VIII release, to remove Yoda conditions I propose, in honor to Mr. Yoda. Hmmmm.

Nowadays no real advantage there is, most of the arguments obsoleted have been. Namely:
* _Myth `#​1_: Assignments in conditions dangerous are.
  * Not using asignments in conditions fixes the problem.
  * See doctrine/coding-standard#9 - addresses both assignment and Yoda.
  * Almost all the time `===` or `!==` used is, thus confusion with `=` problem is not.
* _Myth #​2_: Code is more readable.
  * No, it simply is not.
  * i.e. readability of `2 < func_get_args()` (found in Symfony) is extremely bad.
* In languages like Java Yoda used to avoid nulllptr exceptions is, this case of PHP is not.

I know a proponent of Yoda conditions @Ocramius is, but to see why anymore I fail, maybe explain he can.

Thoughts?


Ocramius:
Tbh, I always keep suggesting and actively asking for yoda conditions: they
read pretty much OK  and there is always the certainty that a constant is
on the left. FWIW, I wouldn't even allow a variable on the left if a
constant is on the right (at language level).

👎 from me.

goetas:
I always hated them. (they come less natural to me when writing code...)
I must admit that there were maybe 3 cases in 15 years of development when a yoda condition saved me from a bug.

greg0ire:
>  I must admit that there were maybe 3 cases in 15 years of development when a yoda condition saved me from a bug.

:+1:  To save 3x 30 mn, you have to spend time running the cs-fixer when there is a rule for that, you have to develop an automatism and stick with it, and you have to make every single if-with-comparison in your codebase harder to read. Forbidding assignments in ifs is a simple rule that you can apply to avoid that... in far, far less cases. Order 66 FTW

greg0ire (approved):

lcobucci:
Shall we discuss and decide this on https://github.com/doctrine/coding-standard/pull/9?

Majkl578:
Yes, that would better place for further discussion, just wanted to trigger some feedback with this here, and eventually get it merged. 😉

Majkl578:
Part of #6964.

PR #6892, opened 2017-12-15
Re-order attributes of relation complex types

thexpand:
This will provide the same look for all of the relations: one-to-many, many-to-one, one-to-one, many-to-many.
It helps during auto-completion of XML code when creating XML schema for an entity.
The order is as follows: field, target-entity, mapped-by, inversed-by, indexed-by, fetch, orphan-removal.


lcobucci (approved):
LGTM 👍

lcobucci:
@thexpand 🚢 thanks!

PR #6897, opened 2017-12-16
Upgrade PHPStan to 0.9.x

lcobucci:
To avoid having weird bugs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,9 +72,10 @@ jobs:
 
     - stage: Code Quality
       env: DB=none STATIC_ANALYSIS
+      install: travis_retry composer update --prefer-dist --prefer-stable
       before_script:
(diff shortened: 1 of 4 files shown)


Majkl578:
I'm 👎 on this, it will break in long run and cause unexpected test failures. :/

lcobucci:
@Majkl578 I think this is not a problem since I've configured a cron job in Travis to run the build in master daily, then we ensure that PRs won't be influenced by phpstan fixes. But I can modify the patch to simply bump to `0.9`

lcobucci:
@Majkl578 updated ;)

Majkl578:
> then we ensure that PRs won't be influenced by phpstan fixe

Only if someone finds time to fix them in meantime. 😄 

I prefer just bumping the version to 0.9. I've been using PHPStan since version 0.6, originally on level 5, now level 7, and every single version introduced new sniffs or features that found new errors. Since PHPStan is still 0.x version I think it's safer to stay locked at specific version for now. :)

PR #6902, opened 2017-12-18
Documentation about getters/setters to DTOs and Value Objects

Pierstoval:
I recently [tweeted](https://twitter.com/Pierstoval/status/942768021038186497) about this, but the original questioning comes from @javiereguiluz:

The "Getting started" guide refers to getters and setters as the base of an entity's behavior, but the Doctrine core team and other experts certainly prefer mapping rich accessors and mutators. @Ocramius also gives this advice in [his "Doctrine beset practices" conference](https://www.youtube.com/watch?v=rzGeNYC3oz0) ([slides here](https://ocramius.github.io/doctrine-best-practices/)).

This PR is not a _final_ proposal, but more a starter about a change to the documentation to suggest using a rich model for our domain instead of relying only on getters and setters.

I thought it would still be nice to bring the getters and setters and then talk about DTOs and rich model, to keep consistency with the 95% (arbitrary number) entities created by developers these days, especially the ones using Symfony (and mostly because the Symfony Form component uses getters and setters **a lot**, and most docs propose to inject them in our forms rather than relying on a DTO).

WDYT about this?

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -234,30 +234,97 @@ When creating entity classes, all of the fields should be protected or private (
 Adding the Entity behaviors
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You have two options to define the accessors mutators that will
-update the state of your objects.
+You have two options to define to define methods in your entity:
+**getters/setters**, or **mutators and DTOs**,
+respectively for **anemic models** or **rich models**.
 
-**Getters and setters**
+**Anemic models: Getters and setters**
 
-The most popular method is to create getter and setter methods for each property (except ``$id``).
-The use of mutators allows Doctrine to hook into calls which
-manipulate the entities in ways that it could not if you just
-directly set the values with ``entity#field = foo;``.
+The most popular method is to create two kinds of methods to
+**read** (getter) and **update** (setter) the object's properties.
 
-The id field has no setter since, generally speaking, your code
-should not set this value since it represents a database id value.
-(Note that Doctrine itself can still set the value using the
-Reflection API instead of a defined setter function)
+Some fields such as ``$id`` are unlikely to be changed, so it's ok to
+not provide any setter for thme.
 
 .. note::
 
-    This method, although very common, is **not** the best approach if
-    you want a perfect domain design. It can lead to unexpected behavior
-    when dealing with different steps of logic in between different
-    calls to ``EntityManager::flush()``, for example when you change
-    the state of an object via event listeners or lifecycle callbacks.
+    Setters are unrelated to Doctrine itself, because the ORM does not
+    use them to change the values of the fields. It instead relies on
+    the Reflection API and do not run neither the constructor nor other
+    methods.
 
-**Mutators and DTOs**
+This method is mostly used when you want to focus on behavior-less
+entities, and when you need to move all your business logic in your
+services rather than in the objects themselves.
+
+It is a common convention which makes it possible to expose each field
+of your entity to external services, while allowing you to keep type
+safety in place.
+
+Such approach is good choice for RAD (rapid application development),
+but may bring problems later down the road, because providing such
+easy way to modify any field in your entity means the entity itself cannot
+ensure it's in valid state. Having entity in invalid state is dangerous,
+because it's one step away from being implicitly saved in database, thereby
+leading to corrupted or inconsistent data in your storage.
+
+.. note::
+
+    This method, although very common, is inappropriate for Domain Driven
+    Design (DDD), because in such design, methods should be named according
+    to actual business operation and entity should be valid anytime.
+
+Here is an example of a simple **anemic model**:
+
+.. configuration-block::
+
+    .. code-block:: php
+
+        <?php
+        class User
+        {
+            private $username;
+            private $passwordHash;
+            private $bans;
+
+            public function getUsername(): string
+            {
+                return $this->username;
+            }
+
+            public function setUsername(string $username): void
+            {
+                $this->username = $username;
+            }
+
+            public function getPasswordHash(): string
+            {
+                return $this->passwordHash;
+            }
+
+            public function setPasswordHash(string $passwordHash): void
+            {
+                $this->passwordHash = $passwordHash;
+            }
+
+            public function getBans(): array
+            {
+                return $this->bans;
+            }
+
+            public function addBan(Ban $ban): void
+            {
+                $this->bans[] = $ban;
+            }
+        }
+
+Here, we avoid all possible logic from the model and only care about injecting
+data into it without validation nor consideration about the object's state.
+
+And as Doctrine ORM is a persistence tool for your domain, the state of an
+object is really important. This is why we recommend using rich models.
+
+**Rich models: Mutators and DTOs**
 
 We recommend using a rich model design and rely on more complex
 mutators, ideally based on DTOs.
@@ -267,40 +334,7 @@ implement methods that represent the **behavior** of your domain.
 For example, when having a ``User`` entity, we could foresee
 the following kind of optimization.
 
-Before, an anemic model with getters and setters:
-
-.. configuration-block::
-
-    .. code-block:: php
-
-        <?php
-        class User {
-            private $username;
-            private $passwordHash;
-            private $bans;
-
-            public function getUsername() :string {
-                return $this->username;
-            }
-
-            public function setUsername(string $username) {
-                $this->username = $username;
-            }
-
-            public function getPasswordHash() : string {
-                return $this->passwordHash;
-            }
-
-            public function setPasswordHash(string $passwordHash) {
-                $this->passwordHash = $passwordHash;
-            }
-
-            public function addBan(Ban $ban) {
-                $this->bans[] = $ban;
-            }
-        }
-
-After, a rich model with proper accessors and mutators:
+Example of a rich model with proper accessors and mutators:
 
 .. configuration-block::
 
@@ -313,38 +347,39 @@ After, a rich model with proper accessors and mutators:
             private $passwordHash;
             private $bans;
 
-            public function toNickname() : string
+            public function toNickname(): string
             {
                 return $this->username;
             }
 
-            public function authenticate(string $password, callable $checkHash) : bool
+            public function authenticate(string $password, callable $checkHash): bool
             {
                 return $checkHash($password, $this->passwordHash) && ! $this->hasActiveBans();
             }
 
-            public function changePass(string $password, callable $hash)
+            public function changePass(string $password, callable $hash): void
             {
                 $this->passwordHash = $hash($password);
             }
 
-            public function ban()
+            public function ban(\DateInterval $duration): void
             {
+                assert($duration->invert !== 1);
+
                 $this->bans[] = new Ban($this);
             }
         }
 
-Additionnally, our entities should never see their state change
+Additionally, our entities should never see their state change
 partially without validation. For example, creating a ``new Product()``
 object without any data makes it an **invalid object**.
-Entities should represent **behavior**, not **data**, therefore
+Rich models should represent **behavior**, not **data**, therefore
 they should be valid even after a ``__construct()``.
 
 To help creating such objects, we can rely on ``DTO``s, and/or make
-our entities Value Objects.
-
-You can know more about Value Objects by reading the `following article
-<https://webmozart.io/blog/2015/09/09/value-objects-in-symfony-forms/>`_.
+our entities always up-to-date. This can be performed with static constructors,
+or rich mutators that accept the ``DTO`` as argument. And DTOs can be
+validated before being sent to the object for update.
 
 By using DTOs, if we take our previous ``User`` example, we could create
 a ``ProfileFormUser`` DTO object that will be a plain model, totally
@@ -357,7 +392,12 @@ Then we can add a new mutator to our ``User``:
         <?php
         class User
         {
-            public function updateFromProfile(ProfileFormUser $profileForm) : void
+            public function updateFromProfile(ProfileFormUser $profileForm): void
+            {
+                // ...
+            }
+
+            public static function createFromRegistration(RegistrationFormUser $registrationForm): self
             {
                 // ...
             }
@@ -372,9 +412,11 @@ update portions of the entity that should already be valid.
 logic in the domain.
 * DTOs can be reused in other components, like when deserializing
 mixed content, using forms...
-* Anemic model tends to isolate the entity from the logic, and a
-rich model fixes that because entity's logic can be put in the
-entity itself.
+* Classic and static constructors can be used to manage different
+ways to create our objects, and they can also use DTOs.
+* Anemic models tend to isolate the entity from logic, whereas
+rich models fix that because logic related to the entity can be
+put in the entity itself, including data validation.
 
 The next step for persistence with Doctrine is to describe the
 structure of the ``Product`` entity to Doctrine using a metadata
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -135,7 +135,6 @@ step:
     $config = Setup::createAnnotationMetadataConfiguration(array(__DIR__."/src"), $isDevMode);
     // or if you prefer XML
     //$config = Setup::createXMLMetadataConfiguration(array(__DIR__."/config"), $isDevMode);
-
     // database configuration parameters
     $conn = array(
         'driver' => 'pdo_sqlite',
@@ -228,33 +227,154 @@ entity definition:
          * @var string
          */
         protected $name;
-
-        public function getId()
-        {
-            return $this->id;
-        }
-
-        public function getName()
-        {
-            return $this->name;
-        }
-
-        public function setName($name)
-        {
-            $this->name = $name;
-        }
     }
 
-When creating entity classes, all of the fields should be protected or private
-(not public), with getter and setter methods for each one (except $id).
+When creating entity classes, all of the fields should be protected or private (not public).
+
+Adding the Entity behaviors
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You have two options to define the accessors mutators that will
+update the state of your objects.
+
+**Getters and setters**
+
+The most popular method is to create getter and setter methods for each property (except ``$id``).
 The use of mutators allows Doctrine to hook into calls which
 manipulate the entities in ways that it could not if you just
-directly set the values with ``entity#field = foo;``
+directly set the values with ``entity#field = foo;``.
 
 The id field has no setter since, generally speaking, your code
 should not set this value since it represents a database id value.
 (Note that Doctrine itself can still set the value using the
 Reflection API instead of a defined setter function)
+
+.. note::
+
+    This method, although very common, is **not** the best approach if
+    you want a perfect domain design. It can lead to unexpected behavior
+    when dealing with different steps of logic in between different
+    calls to ``EntityManager::flush()``, for example when you change
+    the state of an object via event listeners or lifecycle callbacks.
+
+**Mutators and DTOs**
+
+We recommend using a rich model design and rely on more complex
+mutators, ideally based on DTOs.
+This way, you should **not** use getters nor setters, and instead,
+implement methods that represent the **behavior** of your domain.
+
+For example, when having a ``User`` entity, we could foresee
+the following kind of optimization.
+
+Before, an anemic model with getters and setters:
+
+.. configuration-block::
+
+    .. code-block:: php
+
+        <?php
+        class User {
+            private $username;
+            private $passwordHash;
+            private $bans;
+
+            public function getUsername() :string {
+                return $this->username;
+            }
+
+            public function setUsername(string $username) {
+                $this->username = $username;
+            }
+
+            public function getPasswordHash() : string {
+                return $this->passwordHash;
+            }
+
+            public function setPasswordHash(string $passwordHash) {
+                $this->passwordHash = $passwordHash;
+            }
+
+            public function addBan(Ban $ban) {
+                $this->bans[] = $ban;
+            }
+        }
+
+After, a rich model with proper accessors and mutators:
+
+.. configuration-block::
+
+    .. code-block:: php
+        <?php
+        class User
+        {
+            private $banned;
+            private $username;
+            private $passwordHash;
+            private $bans;
+
+            public function toNickname() : string
+            {
+                return $this->username;
+            }
+
+            public function authenticate(string $password, callable $checkHash) : bool
+            {
+                return $checkHash($password, $this->passwordHash) && ! $this->hasActiveBans();
+            }
+
+            public function changePass(string $password, callable $hash)
+            {
+                $this->passwordHash = $hash($password);
+            }
+
+            public function ban()
+            {
+                $this->bans[] = new Ban($this);
+            }
+        }
+
+Additionnally, our entities should never see their state change
+partially without validation. For example, creating a ``new Product()``
+object without any data makes it an **invalid object**.
+Entities should represent **behavior**, not **data**, therefore
+they should be valid even after a ``__construct()``.
+
+To help creating such objects, we can rely on ``DTO``s, and/or make
+our entities Value Objects.
+
+You can know more about Value Objects by reading the `following article
+<https://webmozart.io/blog/2015/09/09/value-objects-in-symfony-forms/>`_.
+
+By using DTOs, if we take our previous ``User`` example, we could create
+a ``ProfileFormUser`` DTO object that will be a plain model, totally
+unrelated to our database, that will be populated via a form and validated.
+Then we can add a new mutator to our ``User``:
+
+.. configuration-block::
+
+    .. code-block:: php
+        <?php
+        class User
+        {
+            public function updateFromProfile(ProfileFormUser $profileForm) : void
+            {
+                // ...
+            }
+        }
+
+There are several advantages to using such model:
+
+* Entity **state is always valid**. No setters means that we only
+update portions of the entity that should already be valid.
+* Instead of having plain getters and setters, our entity now has
+**real behaviors**, meaning it is much easier to determine the
+logic in the domain.
+* DTOs can be reused in other components, like when deserializing
+mixed content, using forms...
+* Anemic model tends to isolate the entity from the logic, and a
+rich model fixes that because entity's logic can be put in the
+entity itself.
 
 The next step for persistence with Doctrine is to describe the
 structure of the ``Product`` entity to Doctrine using a metadata
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -250,7 +250,7 @@ not provide any setter for thme.
 
     Setters are unrelated to Doctrine itself, because the ORM does not
     use them to change the values of the fields. It instead relies on
-    the Reflection API and do not run neither the constructor nor other
+    the Reflection API and does not run either the constructor or other
     methods.
 
 This method is mostly used when you want to focus on behavior-less
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -244,7 +244,7 @@ The most popular method is to create two kinds of methods to
 **read** (getter) and **update** (setter) the object's properties.
 
 Some fields such as ``$id`` are unlikely to be changed, so it's ok to
-not provide any setter for thme.
+omit them.
 
 .. note::
 
@@ -261,8 +261,8 @@ It is a common convention which makes it possible to expose each field
 of your entity to external services, while allowing you to keep type
 safety in place.
 
-Such approach is good choice for RAD (rapid application development),
-but may bring problems later down the road, because providing such
+Such approach is a good choice for RAD (rapid application development),
+but may bring problems later down the road, because providing such an
 easy way to modify any field in your entity means the entity itself cannot
 ensure it's in valid state. Having entity in invalid state is dangerous,
 because it's one step away from being implicitly saved in database, thereby
@@ -272,7 +272,7 @@ leading to corrupted or inconsistent data in your storage.
 
     This method, although very common, is inappropriate for Domain Driven
     Design (DDD), because in such design, methods should be named according
-    to actual business operation and entity should be valid anytime.
+    to actual business operations, and entities should be valid anytime.
 
 Here is an example of a simple **anemic model**:
 
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -229,52 +229,67 @@ entity definition:
         protected $name;
     }
 
-When creating entity classes, all of the fields should be protected or private (not public).
+When creating entity classes, all of the fields should be ``private``.
 
-Adding the Entity behaviors
+Use ``protected`` when strictly needed and very rarely if not ever ``public``.
+
+Adding behavior to Entities
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You have two options to define to define methods in your entity:
+There are two options to define methods in entities:
 **getters/setters**, or **mutators and DTOs**,
-respectively for **anemic models** or **rich models**.
+respectively for **anemic entities** or **rich entities**.
 
-**Anemic models: Getters and setters**
+**Anemic entities: Getters and setters**
 
 The most popular method is to create two kinds of methods to
 **read** (getter) and **update** (setter) the object's properties.
 
-Some fields such as ``$id`` are unlikely to be changed, so it's ok to
+Some fields such as ``$id`` are unlikely to be changed, so it is ok to
 omit them.
 
 .. note::
 
-    Setters are unrelated to Doctrine itself, because the ORM does not
-    use them to change the values of the fields. It instead relies on
-    the Reflection API and does not run either the constructor or other
-    methods.
+    Doctrine ORM does not use any of the methods you defined: it uses
+    reflection to read and write values to your objects, and will never
+    call methods, not even ``__construct``.
 
-This method is mostly used when you want to focus on behavior-less
-entities, and when you need to move all your business logic in your
+This approach is mostly used when you want to focus on behavior-less
+entities, and when you want to have all your business logic in your
 services rather than in the objects themselves.
 
-It is a common convention which makes it possible to expose each field
-of your entity to external services, while allowing you to keep type
-safety in place.
+Getters and setters are a common convention which makes it possible to
+expose each field of your entity to the external world, while allowing
+you to keep some type safety in place.
 
-Such approach is a good choice for RAD (rapid application development),
-but may bring problems later down the road, because providing such an
-easy way to modify any field in your entity means the entity itself cannot
-ensure it's in valid state. Having entity in invalid state is dangerous,
-because it's one step away from being implicitly saved in database, thereby
-leading to corrupted or inconsistent data in your storage.
+Such an approach is a good choice for RAD (rapid application development),
+but may lead to problems later down the road, because providing such an
+easy way to modify any field in your entity means that the entity itself
+cannot guarantee validity of its internal state. Having any object in
+invalid state is dangerous:
+
+- An invalid state can bring bugs in your business logic.
+- The state can be implicitly saved in the database: any forgotten ``flush``
+  can persist the broken state.
+- If persisted, the corrupted data will be retrieved later in your application
+  when the data is loaded again, thereby leading to bugs in your business logic.
+- When bugs occur after corrupted data is persisted, troubleshooting will
+  become much harder, and you might be aware of the bug too late to fix it in a
+  proper manner.
+
+implicitly saved in database, thereby leading to corrupted or inconsistent
+data in your storage, and later in your application when the data is loaded again.
 
 .. note::
 
     This method, although very common, is inappropriate for Domain Driven
-    Design (DDD), because in such design, methods should be named according
-    to actual business operations, and entities should be valid anytime.
+    Design (`DDD <https://en.wikipedia.org/wiki/Domain-driven_design>`)
+    where methods should represent real business operations and not simple
+    property change, And business invariants should be maintained both in the
+    application state (entities in this case) and in the database, with no
+    space for data corruption.
 
-Here is an example of a simple **anemic model**:
+Here is an example of a simple **anemic entity**:
 
 .. configuration-block::
 
@@ -318,23 +333,24 @@ Here is an example of a simple **anemic model**:
             }
         }
 
-Here, we avoid all possible logic from the model and only care about injecting
-data into it without validation nor consideration about the object's state.
+In the example above, we avoid all possible logic in the entity and only care
+about putting and retrieving data into it without validation (except the one
+provided by type-hints) nor consideration about the object's state.
 
-And as Doctrine ORM is a persistence tool for your domain, the state of an
-object is really important. This is why we recommend using rich models.
+As Doctrine ORM is a persistence tool for your domain, the state of an object is
+really important. This is why we strongly recommend using rich entities.
 
-**Rich models: Mutators and DTOs**
+**Rich entities: Mutators and DTOs**
 
-We recommend using a rich model design and rely on more complex
-mutators, ideally based on DTOs.
-This way, you should **not** use getters nor setters, and instead,
+We recommend using a rich entity design and rely on more complex mutators,
+and if needed based on DTOs.
+In this design, you should **not** use getters nor setters, and instead,
 implement methods that represent the **behavior** of your domain.
 
 For example, when having a ``User`` entity, we could foresee
 the following kind of optimization.
 
-Example of a rich model with proper accessors and mutators:
+Example of a rich entity with proper accessors and mutators:
 
 .. configuration-block::
 
@@ -357,7 +373,7 @@ Example of a rich model with proper accessors and mutators:
                 return $checkHash($password, $this->passwordHash) && ! $this->hasActiveBans();
             }
 
-            public function changePass(string $password, callable $hash): void
+            public function changePassword(string $password, callable $hash): void
             {
                 $this->passwordHash = $hash($password);
             }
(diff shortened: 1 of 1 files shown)


fesor:
I don't think that "getting started" should provide insides of "whole value" concepts, but it will be useful to describe reasons why anemic model is most common way of doctrine usage.  Also there should be no phrases like, which enforces users to write setters.

> all of the fields should be protected or private (not public), with getter and setter methods for each one (except $id).

You could give user a hint that anemic model is not preferred way by slightly different example:

```
<?php
// src/Product.php
class Product
{
    /**
     * @var int
     */
    private $id;
    /**
     * @var string
     */
    private $name;

    public function __construct($name)
    {
        $this->name = $name;
    }

    public function getId()
    {
        return $this->id;
    }

    public function getName()
    {
        return $this->name;
    }
}
```

More detailed example of how to design your entities and fill it with behaviour could be placed in separate cookbook (series). Also it will be useful to cover typical problems of first steps in rich model. Like "how to put logic in your entities but not let them became too big (`User` entity with 5000 LoC for example) and so on.

Anyway, I think this would be great!

Majkl578:
Agreed on constructor + immutable state being easier to understand in Getting Started. But behavioral methods should be explained too at some later point. For example, `changePassword($newPassword)` would be good example.

Referencing symfony/symfony-docs#8893 - similar issue for Symfony docs.

fesor:
I don't think that information about proper modelling fits to "getting started" section... This is far more complicated topic. You could change examples, like remove setters:

```
<?php
// src/Product.php
class Product
{
    /**
     * @var int
     */
    private $id;
    /**
     * @var string
     */
    private $name;

    public function __construct($name)
    {
        $this->name = $name;
    }

    public function getId()
    {
        return $this->id;
    }

    public function getName()
    {
        return $this->name;
    }
}
```

and

```
$product->rename($name);
// instead of
$product->setName($name);
```

Doctrine doesn't enforce users to use setters. But with getters there are different story. Since #6719 will not be available to users until 3.0 will be released, it will be just to complicated to describe why you still need `getId()` getter so you could fetch ID of related entity for example without actually preform db query.

Anyway, right now documentation contains something like:

> When creating entity classes, all of the fields should be protected or private (not public), with getter and setter methods for each one (except $id). 

this literally enforce users to use getters in setters in their models. And the only reason for that is how doctrine use accessors for lazy load. 

Also I think that it would be useful to describe concepts of persistence ignorance instead of digging into DTO/UoW stuff in the getting started section. You could provide links on separate sections which will provide more detailed information. Like "how to use entities with forms" and so on. Something for cookbook rather than "getting started".

Ocramius:
The requirement for accessors is no longer needed since Doctrine ORM 2.3,
where we added full support for public properties. In 3.x we will even
support friend objects.

docs/en/tutorials/getting-started.rst
```diff
@@ -233,33 +233,154 @@ entity definition:
...
+
+We recommend using a rich model design and rely on more complex
+mutators, ideally based on DTOs.
+This way, you should **not** use getters nor setters, and instead,
+implement methods that represent the **behavior** of your domain.
+
+For example, when having a ``User`` entity, we could foresee
+the following kind of optimization.
+
+Before, an anemic model with getters and setters:
```
fesor: there are sections "Getting Started: Database First" and "Getting Started: Model First". I think this sections could cover "anemic vs rich model" cases.

docs/en/tutorials/getting-started.rst
```diff
@@ -233,33 +233,154 @@ entity definition:
...
+            private $bans;
+
+            public function toNickname() : string
+            {
+                return $this->username;
+            }
+
+            public function authenticate(string $password, callable $checkHash) : bool
+            {
+                return $checkHash($password, $this->passwordHash) && ! $this->hasActiveBans();
```
fesor: what about cases with "password_needs_rehash"?

docs/en/tutorials/getting-started.rst
Pierstoval: I think it's a.... rich idea :sunglasses:

docs/en/tutorials/getting-started.rst
Pierstoval: It's an example retrieved from Marco's talk about doctrine best practices, but of course it can be changed :)

Ocramius: @fesor I think this is a good example of how a rich domain model aids in better maintainability later on. Not sure if it would fit a simplified example though. Maybe just skip it for now?

fesor: What I thought is... `password_hash` API is already a good abstraction... maybe we don't need to pass closure for that specific case? 

@Ocramius what is your thoughts about `password_needs_rehash` case?

fesor: Yes, probably this something that could be covered in cookbook (like how to deal with dependencies).

docs/en/tutorials/getting-started.rst
```diff
@@ -233,33 +233,154 @@ entity definition:
...
+        class User {
+            private $username;
+            private $passwordHash;
+            private $bans;
+
+            public function getUsername() :string {
+                return $this->username;
+            }
+
+            public function setUsername(string $username) {
```
Majkl578: 👍 for types, also add void

docs/en/tutorials/getting-started.rst
```diff
@@ -233,33 +233,154 @@ entity definition:
...
+            {
+                return $checkHash($password, $this->passwordHash) && ! $this->hasActiveBans();
+            }
+
+            public function changePass(string $password, callable $hash)
+            {
+                $this->passwordHash = $hash($password);
+            }
+
+            public function ban()
```
Majkl578: It may be interesting to see some parameters here. I.e. `\DateInterval $duration` + assert() that it's non-negative. WDYT?

ostrolucky:
Some pretty big words unnecessarily used here for getting started guide. I suggest following changes (I focused on getter/setters option, not sure how to change the rest)

```diff
 Adding the Entity behaviors
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You have two options to define the accessors mutators that will
-update the state of your objects.
+You have two options how to define methods in your entity.
 
 **Getters and setters**
 
-The most popular method is to create getter and setter methods
-for each property (except ``$id``). The use of mutators allows
-Doctrine to hook into calls which manipulate the entities in ways
-that it could not if you just directly set the values with ``entity#field = foo;``.
+The most popular option is creating 2 separate methods for reading
+(getter) and updating (setter) for each field in your entity.
+Some fields, such as ``$id`` are undesirable to be modified and as
+such it's ok to not provide setters for them.
 
-The id field has no setter since, generally speaking, your code
-should not set this value since it represents a database id value.
-(Note that Doctrine itself can still set the value using the
-Reflection API instead of a defined setter function)
+This is a convention which makes it possible to expose each field
+in your entity to external objects in automated way, while allowing
+you to keep type safety in place.
+
+Such approach is good choice for RAD (rapid application development),
+but may prove problematic later down the road, because providing such
+easy way to modify any field in your entity means entity itself cannot
+ensure it's in valid state. Having entity in invalid state is dangerous,
+because it's one step away from being implicitly saved in database.
 
 .. note::
 
-    This method, although very common, is **not** the best approach if
-    you want a perfect domain design. It can lead to unexpected behavior
-    when dealing with different steps of logic in between different
-    calls to ``EntityManager::flush()``, for example when you change
-    the state of an object via event listeners or lifecycle callbacks.
+    This method, although very common, is inappropriate for domain driven
+    design, because in such design, methods should be named according to
+    actual business operation and entity should be valid at all times.
 
 **Mutators and DTOs**
 ```

**Comment by fesor** [comment]
I thought about this a little bit (and asked some of the developers in local community about pit-fails in learning doctrine).

The goal of "getting started" section of every tool is to have it running it as soon as possible (within 10-15 minutes). This is very important to not scare new developers with complexity of the tool. So there should be bare minimal information with only hints about preferred way of designing entities (like remove setters from samples). Al other things should be covered in other sections.

From what I know, getters/setters are mostly used since getting started guides of other tools (like forms, CRUD generators and so on) are provided this examples as most straightforward way to start. later in documentation most of this tools covers how to handle DTO, value objects and other stuff, but in format of cook books. However most of the tools doesn't cover this information (api-platform for example, or sonata admin bundle) which make it hard to developers to start without at-least setters.

So if Doctrine's documentation will push to one direction, and other tools like forms components in to another, and there will be no handy tutorial on how to mix concepts, developers will fall into cognitive dissonance. 

Another thought. I faced with fact that most of the developers just don't know how to handle updates in CRUD based application. Basically they need to replace most of the entity data in update operation, with some kind of control. I teach my team to rely in immutable data since in this way we could replace "update" to another "create" operation but in different context. This requires understanding of concepts like value objects, which represented in Doctrine as `Embeddable`. Unfortunately "Separating Concerns using Embeddables" section is just not contain enough information and many developer event don't know about embeddable. 

I think that embeddable should be covered right after entities in basics of the ORM. Also I found structure of documentation not user friendly. Use of embeddables are simplifies understanding of how to better structure business logic into separate objects.

Another "pattern" in behaviour which I found, is how data modeling is happening. For example in some application we have users. And we also have some kind of referral program. And also users could be both merchants and customers. And each has profiles. Most of the devs which I talked with putted all this logic in same `User` entity eventually will break SRP and will be too big to maintain. And having multiple entities which are referencing same ID and used in different contexts is counterintuitive to people which doesn't want to know anything about DDD and other fancy stuff.

This probably something that should be at lest slightly covered in "Getting Started: Model First". Or maybe instead of two sections there should be one - "Getting Started: Entity Design Approaches" which will have high-level description of different approaches with list of pros/cons of each.  

Also I think that restructuring welcome page will have much more benefits to simplify learning projects. Documentation structure should provide much more cleaners way of learning Doctrine.

**Comment by Pierstoval** [comment]
I'm not sure we should document Embeddables right after the "Getting started" guide. It is related only to specific needs which are not very common, as most of the times, having entities and relations is a better option, could it be for the DBMS or for your domain.

For the rest, I take all your comments in account and will make updates on this subject in the next days/weeks.

In the meantime, thanks everyone for participating in this very interesting discussion, happy new year, and may 2018 give you more free time for OSS 😄 🎉

**Comment by fesor** [comment]
> right after the "Getting started" guide

not right after "getting started". But at least in "Mapping Objects onto a Database"

File: docs/en/tutorials/getting-started.rst [hunk 5]
```diff
@@ -103,7 +103,7 @@ Install Doctrine using the Composer Dependency Management tool, by calling:
     $ composer install
 
 This will install the packages Doctrine Common, Doctrine DBAL, Doctrine ORM,
-Symfony YAML and Symfony Console into the `vendor` directory. The Symfony 
+Symfony YAML and Symfony Console into the `vendor` directory. The Symfony
```
lcobucci: Maybe we should remove `Symfony YAML` from this page? I mean v3 already removed support for it, v2.7 deprecates it, and I think we should start phasing it out ASAP.

Can you please also update the version in L90?

Pierstoval: Yes, I will, I'm just waiting for a moment when I'll have more time for this PR, it's hard work 😉

docs/en/tutorials/getting-started.rst
```diff
@@ -233,33 +233,154 @@ entity definition:
...
+            }
+
+            public function ban()
+            {
+                $this->bans[] = new Ban($this);
+            }
         }
 
-        public function setName($name)
+Additionnally, our entities should never see their state change
```
mheki: Additionnally -> Additionally

docs/en/tutorials/getting-started.rst
```diff
@@ -233,33 +233,154 @@ entity definition:
...
+partially without validation. For example, creating a ``new Product()``
+object without any data makes it an **invalid object**.
+Entities should represent **behavior**, not **data**, therefore
+they should be valid even after a ``__construct()``.
+
+To help creating such objects, we can rely on ``DTO``s, and/or make
```
mheki: >  make our entities Value Objects.
Can we rephrase it? From DDD point of view VOs and entities are different concepts.

docs/en/tutorials/getting-started.rst
```diff
@@ -233,33 +233,154 @@ entity definition:
...
+the following kind of optimization.
+
+Before, an anemic model with getters and setters:
+
+.. configuration-block::
+
+    .. code-block:: php
+
+        <?php
+        class User {
```
greg0ire: This is not PSR compliant

docs/en/tutorials/getting-started.rst
```diff
@@ -233,33 +233,154 @@ entity definition:
...
-Reflection API instead of a defined setter function)
+* Entity **state is always valid**. No setters means that we only
+update portions of the entity that should already be valid.
+* Instead of having plain getters and setters, our entity now has
+**real behaviors**, meaning it is much easier to determine the
+logic in the domain.
+* DTOs can be reused in other components, like when deserializing
+mixed content, using forms...
+* Anemic model tends to isolate the entity from the logic, and a
+rich model fixes that because entity's logic can be put in the
```
greg0ire: "the entity's" logic I think

docs/en/tutorials/getting-started.rst
pylebecq: Hi guys.
I appreciate the effort you put in updating the documentation and promote better programming practices.
I agree with @fesor that the "dealing with dependencies" topic must be covered. Yes, it can be in some cookbook, but it must be covered because it will be an issue people will run into very quickly and not finding anything related to this will probably make them to give up and go to the well known "don't put behavior in your entities, just put data and make some services that will operate on your entities through some setters".

Pierstoval:
Fixed all your review @greg0ire, but next time, could you add them in a Github review so we can hide them when fixed? 😅

Pierstoval:
By the way, for reviewers, I just made some changes to the text, if you could review again 😉

greg0ire:
@Pierstoval that's what you get for mentioning me in a commit message :P. You can still squash if you want them to disappear though

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
          * @var string
          */
         protected $name;
+    }
 
-        public function getId()
+When creating entity classes, all of the fields should be protected or private (not public).
```
Ocramius: We should probably rephrase to something like:

 > should be `private`, `protected` when strictly needed and very rarely if not ever `public`

I can also add that `protected` and `public` have some performance overhead (minimal, but it is there). Unsure if we can pack it into the article.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+
+Adding the Entity behaviors
```
Ocramius: `Adding behavior to Entities`?

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You have two options to define to define methods in your entity:
```
Ocramius: It is a good idea to keep things very impersonal.

`There are two options` and `in entities`

Also, `to define` is used twice here

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+**getters/setters**, or **mutators and DTOs**,
+respectively for **anemic models** or **rich models**.
+
+**Anemic models: Getters and setters**
+
+The most popular method is to create two kinds of methods to
+**read** (getter) and **update** (setter) the object's properties.
+
+Some fields such as ``$id`` are unlikely to be changed, so it's ok to
```
Ocramius: overall, try not abbreviating `it's` - keep `it is`

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+omit them.
+
+.. note::
+
+    Setters are unrelated to Doctrine itself, because the ORM does not
```
Ocramius: It is true that they are `unrelated`, but that's not how it should be explained to the reader.

I think a better approach is to say something like:

 > Doctrine ORM does not use any of the methods you defined: it uses reflection to read and write values to your objects, and will never call methods, not even `__construct`

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+    use them to change the values of the fields. It instead relies on
+    the Reflection API and does not run either the constructor or other
+    methods.
+
+This method is mostly used when you want to focus on behavior-less
```
Ocramius: `method` is confusing, since we are talking about methods. `approach` maybe?

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+entities, and when you need to move all your business logic in your
```
Ocramius: s/move/have

Ocramius: s/need/want to

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+services rather than in the objects themselves.
+
+It is a common convention which makes it possible to expose each field
```
Ocramius: Instead of `It`, prefer repeating the subject. `Getters and setters are a common convention which`

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+of your entity to external services, while allowing you to keep type
```
Ocramius: s/external services/the external world

Ocramius: s/keep type safety/keep some type safety

You said it yourself a bit further: invalid state ;-)

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+safety in place.
+
+Such approach is a good choice for RAD (rapid application development),
+but may bring problems later down the road, because providing such an
```
Ocramius: s/bring/lead to

Ocramius: s/down the road/when a software project grows

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+easy way to modify any field in your entity means the entity itself cannot
```
Ocramius: means that the

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+ensure it's in valid state. Having entity in invalid state is dangerous,
```
Ocramius: s/ensure/guarantee validity of its internal state

(yes, `its` without `'`)

Ocramius: s/Having entity in invalid state/Having any object in invalid state

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+because it's one step away from being implicitly saved in database, thereby
+leading to corrupted or inconsistent data in your storage.
```
Ocramius: in your storage, and later in your application when the data is loaded again.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+
+It is a common convention which makes it possible to expose each field
+of your entity to external services, while allowing you to keep type
+safety in place.
+
+Such approach is a good choice for RAD (rapid application development),
+but may bring problems later down the road, because providing such an
+easy way to modify any field in your entity means the entity itself cannot
+ensure it's in valid state. Having entity in invalid state is dangerous,
+because it's one step away from being implicitly saved in database, thereby
```
Ocramius: need something less figurative than `it's one step away`

Ocramius: The concept that needs to be conveyed here is that:

 * broken state -> bugs
 * broken state -> db -> data corruption
 * corrupt db -> broken state -> bugs

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+leading to corrupted or inconsistent data in your storage.
+
+.. note::
+
+    This method, although very common, is inappropriate for Domain Driven
```
Ocramius: I think most readers don't know about DDD here: can we add a wikipedia link or such?

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+    Design (DDD), because in such design, methods should be named according
```
Ocramius: `because in such design` is repeating `design` again. Needs rephrasing.

Ocramius: `methods should be named` - not just named, but also perform operations close to business operations. To clarify: the naive reader will pick this up as "I have to name my setter so it looks like biz shit"

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+    to actual business operations, and entities should be valid anytime.
```
Ocramius: > and entities should be valid anytime.

And business invariants should be maintained both in the application state (entities in this case) and in the database, with no space for data corruption.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+
+Here is an example of a simple **anemic model**:
```
Ocramius: `anemic entity`. Let's avoid the word "model" overall please.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+            }
+
+            public function addBan(Ban $ban): void
+            {
+                $this->bans[] = $ban;
+            }
         }
 
-        public function getName()
+Here, we avoid all possible logic from the model and only care about injecting
```
Ocramius: s/Here/in the example above

Ocramius: > all possible logic from the model 

 * `model` needs to disappear
 * that `from` is super-weird. It's `in` the model, not `from`

Ocramius: > injecting

More like "putting and retrieving" - injecting seems to lead to dependency injection, which we don't discuss in the ORM docs

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+data into it without validation nor consideration about the object's state.
+
+And as Doctrine ORM is a persistence tool for your domain, the state of an
+object is really important. This is why we recommend using rich models.
```
Ocramius: `why we strongly recommend`

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+            {
+                $this->bans[] = $ban;
+            }
         }
 
-        public function getName()
+Here, we avoid all possible logic from the model and only care about injecting
+data into it without validation nor consideration about the object's state.
+
+And as Doctrine ORM is a persistence tool for your domain, the state of an
```
Ocramius: Don't start a sentence with `And`: simply drop it.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+object is really important. This is why we recommend using rich models.
+
+**Rich models: Mutators and DTOs**
+
+We recommend using a rich model design and rely on more complex
+mutators, ideally based on DTOs.
```
Ocramius: s/ideally/and if needed

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+This way, you should **not** use getters nor setters, and instead,
```
Ocramius: s/this way/In this design

Ocramius: s/use/implement

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+implement methods that represent the **behavior** of your domain.
+
+For example, when having a ``User`` entity, we could foresee
+the following kind of optimization.
+
+Example of a rich model with proper accessors and mutators:
```
Ocramius: s/model/entity

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+            }
+
+            public function changePass(string $password, callable $hash): void
+            {
+                $this->passwordHash = $hash($password);
+            }
+
+            public function ban(\DateInterval $duration): void
+            {
+                assert($duration->invert !== 1);
```
Ocramius: `invert`? O_o That's something I didn't know, thanks! :D

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+            {
+                return $this->username;
+            }
+
+            public function authenticate(string $password, callable $checkHash): bool
+            {
+                return $checkHash($password, $this->passwordHash) && ! $this->hasActiveBans();
+            }
+
+            public function changePass(string $password, callable $hash): void
```
Ocramius: s/changePass/changePassword - no abbreviations please

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+For example, when having a ``User`` entity, we could foresee
+the following kind of optimization.
+
+Example of a rich model with proper accessors and mutators:
+
+.. configuration-block::
+
+    .. code-block:: php
+        <?php
+        class User
```
Ocramius: This class has no constructor: we should at least have a `::register()` or such

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+            {
+                assert($duration->invert !== 1);
+
+                $this->bans[] = new Ban($this);
+            }
         }
 
-        public function setName($name)
+Additionally, our entities should never see their state change
+partially without validation. For example, creating a ``new Product()``
```
Ocramius: s/partially//

I suggest replacing the entire sentence with something like `The entities should only mutate state after checking that all business logic invariants are being respected.`. Avoid negative language to ease readability :)

docs/en/tutorials/getting-started.rst
Ocramius: Disregard: I saw the description below (you add the ctor later on)

docs/en/tutorials/getting-started.rst
Ocramius: We should clarify that this is a stub, since you keep editing it later in the documentation

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+object without any data makes it an **invalid object**.
+Rich models should represent **behavior**, not **data**, therefore
+they should be valid even after a ``__construct()``.
```
Ocramius: > after a `__construct` call

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+        <?php
+        class User
         {
-            $this->name = $name;
+            public function updateFromProfile(ProfileFormUser $profileForm): void
+            {
+                // ...
+            }
+
+            public static function createFromRegistration(RegistrationFormUser $registrationForm): self
```
Ocramius: s/RegistrationFormUser/UserRegistrationForm

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+Then we can add a new mutator to our ``User``:
+
+.. configuration-block::
+
+    .. code-block:: php
+        <?php
+        class User
         {
-            $this->name = $name;
+            public function updateFromProfile(ProfileFormUser $profileForm): void
```
Ocramius: s/ProfileFormUser/ProfileEditingForm

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
         }
 
-        public function setName($name)
+Additionally, our entities should never see their state change
+partially without validation. For example, creating a ``new Product()``
+object without any data makes it an **invalid object**.
+Rich models should represent **behavior**, not **data**, therefore
+they should be valid even after a ``__construct()``.
+
+To help creating such objects, we can rely on ``DTO``s, and/or make
```
Ocramius: I don't think `DTO` is linked/described anywhere in here...

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+our entities always up-to-date. This can be performed with static constructors,
+or rich mutators that accept the ``DTO`` as argument. And DTOs can be
```
Ocramius: s/the DTO as argument/DTOs as arguments

Ocramius: s/arguments/parameters

Ocramius: Also here: sentence started with `And`: don't do that. Split over a new paragraph and explain that a DTO maintains some internal state consistency that we can rely upon.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+validated before being sent to the object for update.
+
+By using DTOs, if we take our previous ``User`` example, we could create
+a ``ProfileFormUser`` DTO object that will be a plain model, totally
```
Ocramius: Naming suggestion below - the `Form` name in the middle of the DTO class name is super-weird

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
-(not public), with getter and setter methods for each one (except $id).
-The use of mutators allows Doctrine to hook into calls which
-manipulate the entities in ways that it could not if you just
-directly set the values with ``entity#field = foo;``
 
-The id field has no setter since, generally speaking, your code
-should not set this value since it represents a database id value.
-(Note that Doctrine itself can still set the value using the
-Reflection API instead of a defined setter function)
+There are several advantages to using such model:
```
Ocramius: `model` needs to go away :P

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+
+* Entity **state is always valid**. No setters means that we only
```
Ocramius: highlight `Entity` as part of the bold section

Ocramius: s/No setters means that/Since no setters exist,

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+update portions of the entity that should already be valid.
+* Instead of having plain getters and setters, our entity now has
+**real behaviors**, meaning it is much easier to determine the
```
Ocramius: s/behaviors/behavior

Ocramius: Also: you really like the word "meaning" :D

Ocramius: Can be simplified: `**real behaviors**: it `

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+logic in the domain.
+* DTOs can be reused in other components, like when deserializing
```
Ocramius: s/like when/for example

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+mixed content, using forms...
+* Classic and static constructors can be used to manage different
+ways to create our objects, and they can also use DTOs.
```
Ocramius: s/use/rely on

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+* Anemic models tend to isolate the entity from logic, whereas
+rich models fix that because logic related to the entity can be
```
Ocramius: >  fix that because

"fix that" is strange to read in this context, because it is not immediately obvious that this is a problem. We only briefly explained that anemic models lead to some issues, but we didn't explain that logic outside entities versus inside is worse.

Ocramius (changes requested):

docs/en/tutorials/getting-started.rst
fesor: I don't think that we need to introduce any limitations in getting started at all. Only one that are really limitations of Doctrine. Proper entity modeling could be covered elsewhere.

greg0ire:
@Pierstoval I have access to a computer again and deleted my comments to keep this PR relatively clean

docs/en/tutorials/getting-started.rst
Pierstoval: I'm not sure the performance overhead might be interesting for end-users to know.

greg0ire: @fesor it's not a limitation (of Doctrine) is it? Just a good recommendation. There's no explanation though, and the performance concerns are additional to something else.

When creating entity classes, fields should be `private`, `protected` when strictly needed and very rarely if not ever `public`, to make sure encapsulation is not broken by needless exposition.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You have two options to define to define methods in your entity:
+**getters/setters**, or **mutators and DTOs**,
+respectively for **anemic models** or **rich models**.
+
+**Anemic models: Getters and setters**
+
+The most popular method is to create two kinds of methods to
+**read** (getter) and **update** (setter) the object's properties.
```
greg0ire: "read" and "update" sounds weird, how about:

- "read" and "write"?
- "share" and "mutate"?

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+
+This method is mostly used when you want to focus on behavior-less
+entities, and when you need to move all your business logic in your
+services rather than in the objects themselves.
+
+It is a common convention which makes it possible to expose each field
+of your entity to external services, while allowing you to keep type
+safety in place.
+
+Such approach is a good choice for RAD (rapid application development),
```
greg0ire: "Such an approach"

docs/en/tutorials/getting-started.rst
greg0ire: How about:

> Having any object in an invalid state can lead to bugs that will be even harder to troubleshoot and fix if that broken state is persisted to a database and detected later on when the corrupted data is loaded.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,33 +227,196 @@ entity definition:
...
+
+            public function addBan(Ban $ban): void
+            {
+                $this->bans[] = $ban;
+            }
         }
 
-        public function getName()
+Here, we avoid all possible logic from the model and only care about injecting
+data into it without validation nor consideration about the object's state.
```
greg0ire: "without any validation but the one provided by the type hints"

docs/en/tutorials/getting-started.rst
greg0ire: TIL :bulb:

docs/en/tutorials/getting-started.rst
greg0ire: DTO object expands to "data transfer object object"

docs/en/tutorials/getting-started.rst
greg0ire: should be "such a model", BTW

greg0ire (changes requested):

docs/en/tutorials/getting-started.rst
Pierstoval: Yep, we can't create "negative" `DateInterval` objects, so we can use this flag to specify it is 😄

docs/en/tutorials/getting-started.rst
Pierstoval: Yeah, I need to improve my english language a bit 🤣

docs/en/tutorials/getting-started.rst
Pierstoval: I thought read & update were nicer. By adding "write", I'm confused with the fact that we _read something_ but we _write in something_, and I didn't want to bother with my imperfect english 😆 

"share" and "mutate" are closer to the DDD & rich entities approach, and these are words that are possibly unknown by many developers, so I didn't want to add them in the common anemic design

Pierstoval:
I made another big update, thanks a lot for all reviewers, it makes me both improve my english and my understanding of how we explain such a complex design as DTOs and rich entities 😄

Ocramius (approved):
:ship: this is good as-is IMO (besides more nitpicking that I will avoid here).

If anybody has additions/suggestions, please send more PRs :-)

docs/en/tutorials/getting-started.rst
```diff
@@ -228,39 +227,235 @@ entity definition:
...
+- An invalid state can bring bugs in your business logic.
+- The state can be implicitly saved in the database: any forgotten ``flush``
+  can persist the broken state.
+- If persisted, the corrupted data will be retrieved later in your application
+  when the data is loaded again, thereby leading to bugs in your business logic.
+- When bugs occur after corrupted data is persisted, troubleshooting will
+  become much harder, and you might be aware of the bug too late to fix it in a
+  proper manner.
+
+implicitly saved in database, thereby leading to corrupted or inconsistent
```
javiereguiluz: @Pierstoval this phrase looks a missed leftover. Is it intended?

docs/en/tutorials/getting-started.rst
```diff
@@ -228,39 +227,235 @@ entity definition:
...
-should not set this value since it represents a database id value.
-(Note that Doctrine itself can still set the value using the
-Reflection API instead of a defined setter function)
-
-The next step for persistence with Doctrine is to describe the
-structure of the ``Product`` entity to Doctrine using a metadata
-language. The metadata language describes how entities, their
-properties and references should be persisted and what constraints
-should be applied to them.
+There are several advantages to using such a model:
```
javiereguiluz: For the previous "anemic entities" you mentioned 1 advantage (RAD) and 4 drawbacks ... here you mention 4 advantages and 0 drawbacks. Do these "rich entities" really have no drawbacks? Thanks!

docs/en/tutorials/getting-started.rst
Ocramius: Dropped in https://github.com/doctrine/doctrine2/pull/7137

docs/en/tutorials/getting-started.rst
```diff
@@ -228,39 +227,235 @@ entity definition:
          * @var string
          */
         protected $name;
```
tomasfejfar: The vairable is protected. 

> When creating entity classes, all of the fields should be ``private``.

docs/en/tutorials/getting-started.rst
Ocramius: Not really: being able to have the database as an implementation detail is primary the reason for using a data mapper over other types of persistence layers, so this approach goes much much closer to the business requirements. It still has the flaws of mutable state management, but that's inherent from RDBMS-backed applications in general.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,39 +227,235 @@ entity definition:
...
+and if needed based on DTOs.
+In this design, you should **not** use getters nor setters, and instead,
+implement methods that represent the **behavior** of your domain.
+
+For example, when having a ``User`` entity, we could foresee
+the following kind of optimization.
+
+Example of a rich entity with proper accessors and mutators:
+
+.. configuration-block::
```
tomasfejfar: This does not render!

image

Ocramius:
@tomasfejfar this patch is already merged: send further PRs to fix pending issues :+1:

docs/en/tutorials/getting-started.rst
```diff
@@ -228,39 +227,235 @@ entity definition:
...
-Reflection API instead of a defined setter function)
-
-The next step for persistence with Doctrine is to describe the
-structure of the ``Product`` entity to Doctrine using a metadata
-language. The metadata language describes how entities, their
-properties and references should be persisted and what constraints
-should be applied to them.
+There are several advantages to using such a model:
+
+* **Entity state is always valid**. Since no setters exist, this means that we
```
tomasfejfar: image

docs/en/tutorials/getting-started.rst
```diff
@@ -228,39 +227,235 @@ entity definition:
...
+* Classic and static constructors can be used to manage different ways to
+create our objects, and they can also use DTOs.
+* Anemic entities tend to isolate the entity from logic, whereas rich
+entities allow putting the logic in the object itself, including data
+validation.
+
+The next step for persistence with Doctrine is to describe the structure of
+the ``Product`` entity to Doctrine using a metadata language. The metadata
+language describes how entities, their properties and references should be
+persisted and what constraints should be applied to them.
```
tomasfejfar: The section following this edit mentions again setters and getters. That's IMO confusing.

docs/en/tutorials/getting-started.rst
```diff
@@ -228,39 +227,235 @@ entity definition:
...
+Then we can add a new mutator to our ``User``:
+
+.. configuration-block::
+
+    .. code-block:: php
+        <?php
+        class User
         {
-            $this->name = $name;
+            public function updateFromProfile(ProfileEditingForm $profileForm): void
```
morozov: While the `ProfileEditingForm` is defined above as a DTO, the code itself reads as the domain object (user) depending on the application logic (form). This way, it violates the Open/closed principle: every time a new form is added, the `User` class needs to be updated to enable its mutation by the form. Does the user really have to accept multiple types of DTO or at least, could all forms be represented by one DTO type?

BTW, this code is not rendered on the website:
screenshot from 2018-03-21 12-21-36

docs/en/tutorials/getting-started.rst
sstok: There needs to a an empty line after the `.. code-block:: php`

PR #6903, opened 2017-12-18
Travis: Allow failures for tests on non-SQLite RDBMS, disable SLC tests

Majkl578:
Disable everything except SQLite on SLC=0. Because these simply don't work in develop at this stage. Also silent b0rked embedded.
Trying to make CI at least a bit useful for now.

Also bumbed PHPUnit version, older were was confused by Generator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,8 @@ install: travis_retry composer update --prefer-dist
 script:
   - if [[ "$DB" == "mysql" || "$DB" == "mariadb" ]]; then mysql -e "CREATE SCHEMA doctrine_tests; GRANT ALL PRIVILEGES ON doctrine_tests.* to travis@'%'"; fi
   - ENABLE_SECOND_LEVEL_CACHE=0 ./vendor/bin/phpunit -v -c tests/travis/$DB.travis.xml
-  - ENABLE_SECOND_LEVEL_CACHE=1 ./vendor/bin/phpunit -v -c tests/travis/$DB.travis.xml --exclude-group performance,non-cacheable,locking_functional
+  # temporarily disabled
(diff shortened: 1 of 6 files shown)


Majkl578:
This now fails on PHPStan check due to errors in code - void return returns, LSP violations.

.travis.yml
```diff
@@ -19,37 +19,37 @@ before_install:
...
-        - ./tests/travis/install-mysql-$MYSQL_VERSION.sh
-      sudo: required
+#    - stage: Test
+#      env: DB=mariadb
+#      addons:
+#        mariadb: 10.1
+#
+#    - stage: Test
+#      env: DB=mysql MYSQL_VERSION=5.7
+#      php: 7.1
```
lcobucci: Didn't we remove PHP 7.1 from `develop`?

composer.json
```diff
@@ -27,7 +27,7 @@
         "symfony/console": "~3.0|~4.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^6.0"
+        "phpunit/phpunit": "^6.5"
```
lcobucci: Let's bump this on `master`? Pleaaaaaase!!!

lcobucci (changes requested):

.travis.yml
Majkl578: Yes, but @guilhermeblanco lost that commit during recent rebase. We should re-apply it.

lcobucci: 😭

Majkl578:
Although I'm not a fan of this approach in particular, I'll merge it as is so we can at least get relevant CI output for basic tests + Code Quality for the time being. We already got some bugs slipped through...
I'll start fixing PHPStan and CS builds later today - once done, we can require green Travis for PRs again.

Fixing SLC and non-SQLite tests will require additional effort at later stage. Hopefully we'll be able to re-enable those soon.

PR #6905, opened 2017-12-19
Setup: Switch Apc -> Apcu and Memcache -> Memcached

Majkl578:
Quick fix for 2.6 release.
Without this change it would be impossible to autoconfigure APC/Memcache since previous extensions do not exist anymore in 7.1 world [1][2]. Removed XCache, never upgraded to PHP 7 [3].

Marked as BC break, but it's really not one since those extensions wouldn't just work - this is actually _broken_ by requiring PHP 7.1.

[1] https://pecl.php.net/package-changelog.php?package=APC&release=3.1.13
[2] https://pecl.php.net/package-changelog.php?package=memcache&release=3.0.8
[3] https://xcache.lighttpd.net/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,7 @@ jobs:
       env: DB=none STATIC_ANALYSIS
       install: travis_retry composer update --prefer-dist --prefer-stable
       before_script:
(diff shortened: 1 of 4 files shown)


lcobucci (approved):
LGTM although I'd love to have this class removed from the ORM for good in v3

PR #6896, opened 2017-12-16
Run all code quality tools in the same stage

lcobucci:
In order to execute things in parallel and speed up the build.

PR #6906, opened 2017-12-19
Travis: Moved coverage stage back to Code Quality group

Majkl578:
Was moved as part of  #6896, probably accidentally. It slows Test stage from 2-3min to 7-8min. If that was intentional, feel free to close right away. 😊 

(Getting tired of these false positives on Scrutinizer... 😢 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ jobs:
         - composer config minimum-stability dev
         - travis_retry composer update --prefer-dist
 
-    - stage: Test
+    - stage: Code Quality
       env: DB=sqlite COVERAGE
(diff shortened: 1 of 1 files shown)


lcobucci:
@Majkl578 sorry if I didn't make it clear on the previous change but this was done consciously. I wanted to run coverage in parallel in the test stage, not only because it's related to tests but also to not block scrutinizer builds for too long

Majkl578:
@lcobucci Right, it looks strange without context, makes somewhat sense now. Can we do anything about these random Scrutinizer failures? Those can get pretty annoying... :(

PR #6910, opened 2017-12-19
Combine consecutives issets

carusogabriel:
`isset` allow us to pass multiple values to check.

[PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) helped me on that.


Ocramius (approved):
LGTM :+1:

PR #6912, opened 2017-12-20
[CS] Heredoc to Nowdoc

Slamdunk:
Nowdocs aren't parsed, so the developer immediately knows that they don't cointain interpreted variable; hence are faster to read and maintain.

The porting was done with [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) rules `escape_implicit_backslashes` and `heredoc_to_nowdoc`, so only heredocs - even with dollar signs and backslashes - eligible to the change were fixed.


Ocramius (approved):
LGTM :+1:

PR #6913, opened 2017-12-20
[CS] Prefer `elseif` over `else if`

Slamdunk:
As of yet there is a mixed usage of the two conditions `elseif` and `else if`.

Symfony prefers the former: the fix was applied with [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) rule `elseif`.


Ocramius (approved):
LGTM :+1:

PR #6916, opened 2017-12-20
[CS] One black line after opening tag

Slamdunk:
The fix was applied with [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) rule `blank_line_after_opening_tag`.


Ocramius (approved):
LGTM :+1:

PR #6918, opened 2017-12-20
[CS] No alias functions

Slamdunk:
The fix was applied with [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) rule `no_alias_functions`.


Ocramius:
:ship:

PR #6919, opened 2017-12-20
[CS] Remove useless semicolon statements

Slamdunk:
The fix was applied with [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) rule `no_empty_statement`.


Ocramius (approved):
LGTM :+1:

PR #6921, opened 2017-12-20
[CS] Properties must not be explicitly initialized with null

Slamdunk:
The fix was applied with [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) rule `no_null_property_initialization`.

Ref: https://github.com/doctrine/doctrine2/pull/6917#issuecomment-353032273


Ocramius (approved):
LGTM :+1:

PR #6935, opened 2017-12-22
Remove support for entity namespaces (Foo:Bar)

Majkl578:
Another possibly controversial PR removing existing feature. Entity namespaces are basically obsolete since PHP 5.5 released 4.5 years ago and are superseded by much safer `::class` pseudo-constant, which is refactoring- and autocomplete-friendly in IDEs.


Ocramius (approved):
:+1:

Xymanek:
I'm using this notation in DQL rather than FQCN as they can be quite long. Take a look at following:

* `Acme\Cms\Modules\BlogModule\Entity\Article` (yes, this is a bit on extreme side but it shows the idea)
* `Blog:Article` or `BlogModule:Article`

Removing this feature will make using DQL without builder quite cumbersome I think

Majkl578:
> Removing this feature will make using DQL without builder quite cumbersome I think

The issue with entity namespaces is that they're solving a problem that doesn't exist anymore (it did back then when Doctrine 2.0 was released for PHP 5.3, `::class` is PHP 5.5+). Furthermore, it makes code harder to maintain, harder to read (subjectively) and harder to refactor (rename class, move to different namespace).


```
SELECT a FROM BlogModule:Article a WHERE a.author = :author
```
would become:
```
SELECT a FROM ' . Article::class . ' a WHERE a.author = :author
```
And with query builder, you won't even need string concatenation:
```
$qb->select('a')
    -> from(Article::class, 'a')
    -> where('a.author = :author')
```

Majkl578:
Finally ready for shipping. :shipit:

PR #6936, opened 2017-12-23
Remove support for same-namespace class resolution

### LINKED ISSUE #4845: DDC-3985: Drop support for same-namespace names resolution
Jira issue originally created by user @ocramius:

When resolving target classes, the ORM currently allows to reference classes in the same namespace via their short name.

Given a class `Foo\Bar\Baz`, when referencing a class `Foo\Bar\Tab`, it is possible to simply reference it as `Tab` inside the mappings of `Foo\Bar\Baz`.

This is just a useless complication that causes confusion and makes introspection harder. Additionally, since annotation mappings support the `::class` meta-constant, it now is a useless feature, as the PHP namespace system already solves the problem.

Therefore, this sort of "magic resolving" should be dropped.

Majkl578:
Looks huge, but most of the changes are (regex) replacement of same-NS references by `::class` reference in tests.

BC breaks:
- annotations should be migrated to `::class` references
- XML needs to be migrated to FQCN references
- Doctrine\ORM\Mapping\ClassMetadata::fullyQualifiedClassName() removed

Closes #4845.


simPod:
Will it work for `::class`-referencing entities in different namespaces?

Ocramius:
Already does

Ocramius (approved):
LGTM :+1:

PR #6947, opened 2018-01-02
[CS] Require explicit visibility

Slamdunk:
The fix was applied with [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) rule `visibility` on `methods`.


Ocramius (approved):
LGTM :+1:

PR #6948, opened 2018-01-02
[CS] is_null($var) -> null === $var

Slamdunk:
The fix was applied with [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) rule `is_null`.


Ocramius:
:ship: thanks @Slamdunk!

Majkl578:
FYI: This shouldn't be using Yoda as per vote in doctrine/coding-standard#9.

Slamdunk:
If I got it right, when you'll finish the work on https://github.com/doctrine/coding-standard/ you'll apply it to all the doctrine codebase, and so the code will be fixed, right?

Majkl578:
Yes, just better to not introduce code that needs modifications a week later. :) Doesn't matter now when merged.

Slamdunk:
> Yes, just better to not introduce code that needs modifications a week later.

I agree, I asked because in opposite case I would have opened another PR.

PR #6952, opened 2018-01-02
[CS] Add curly braces to indirect variables

Slamdunk:
The fix was applied with [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) rule `explicit_indirect_variable`.

Add curly braces to indirect variables to make them clear to understand.

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -302,6 +302,6 @@ class EntityRepository implements ObjectRepository, Selectable
             throw ORMException::invalidMagicCall($this->entityName, $fieldName, $method . $by);
         }
 
-        return $this->$method([$fieldName => $arguments[0]], ...array_slice($arguments, 1));
+        return $this->{$method}([$fieldName => $arguments[0]], ...array_slice($arguments, 1));
     }
 }
--- a/lib/Doctrine/ORM/Event/ListenersInvoker.php
+++ b/lib/Doctrine/ORM/Event/ListenersInvoker.php
@@ -84,7 +84,7 @@ class ListenersInvoker
     {
         if ($invoke & self::INVOKE_CALLBACKS) {
             foreach ($metadata->lifecycleCallbacks[$eventName] as $callback) {
-                $entity->$callback($event);
+                $entity->{$callback}($event);
             }
         }
 
@@ -94,7 +94,7 @@ class ListenersInvoker
                 $method     = $listener['method'];
                 $instance   = $this->resolver->resolve($class);
 
-                $instance->$method($entity, $event);
+                $instance->{$method}($entity, $event);
             }
         }
 
--- a/lib/Doctrine/ORM/PersistentObject.php
+++ b/lib/Doctrine/ORM/PersistentObject.php
@@ -120,7 +120,7 @@ abstract class PersistentObject implements EntityManagerAware
 
         switch (true) {
             case ($property instanceof FieldMetadata && ! $property->isPrimaryKey()):
-                $this->$field = $args[0];
+                $this->{$field} = $args[0];
                 break;
 
             case ($property instanceof ToOneAssociationMetadata):
@@ -130,7 +130,7 @@ abstract class PersistentObject implements EntityManagerAware
                     throw new \InvalidArgumentException("Expected persistent object of type '".$targetClassName."'");
                 }
 
-                $this->$field = $args[0];
+                $this->{$field} = $args[0];
                 $this->completeOwningSide($property, $args[0]);
                 break;
         }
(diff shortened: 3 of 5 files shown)


Majkl578:
I'm inclined to decline this, it doesn't really improve readibility at all... Leaving to someone else to decide.

Slamdunk:
> it doesn't really improve readibility at all

Consider this change:
```diff
-    $entity->$callback($event);
+    $entity->{$callback}($event);
```
Without the fix, a non-expert user may doubt if the callback is executed before or after the object property reference.

Another improvement should be renaming the variable: in that case the correct name would be `$callbackName`.

Majkl578:
> Without the fix, a non-expert user may doubt if the callback is executed before or after the object property reference.

In that case, you would probably use even more explicit syntax:
```php
[$foo, $bar]();
```
https://3v4l.org/O0OSL

Ocramius:
@Majkl578 unless opcache does some magic, that is also an array declaration happening (see opcodes)

lcobucci:
> I'm inclined to decline this, it doesn't really improve readibility at all... Leaving to someone else to decide.

I agree that this doesn't improve readability at all.

> Without the fix, a non-expert user may doubt if the callback is executed before or after the object property reference.

Didn't PHP 7 standardise the order of execution? I'd say there's no room for doubt...

Slamdunk:
> Didn't PHP 7 standardise the order of execution? I'd say there's no room for doubt...

Reference: https://secure.php.net/manual/en/migration70.incompatible.php#migration70.incompatible.variable-handling

The mess comes when using array referencing like ` $foo->$bar['baz']()`: without explicitness and perfect memory of the manual, it's impossible to guess the right precedence.

If a developer would ever be tempted to write something like
```php
$callbacks = [$mainCallback];
$entity->$mainCallback($event);
$entity->$callbacks[0]($event);
```
Results would be far from expected behaviour (@Ocramius scourging aside).

Hence, because it is possible to fall in bad code starting from easy readable code, I prefer to always be explicit about indirect variables.

Majkl578:
> unless opcache does some magic, that is also an array declaration happening (see opcodes)

Yes, that is true, considering the two pieces of code:
```php
$foo = new DateTimeImmutable();
$bar = 'format';
[$foo, $bar]('U');
```
vs.
```php
$foo = new DateTimeImmutable();
$bar = 'format';
$foo->$bar('U');
```
and their VLD output:
```
$ php7.2 -dopcache.enable=1 -dopcache.enable_cli=1 -dvld.active=1 -dvld.execute=0 /tmp/call-dynamic.php 
Finding entry points
Branch analysis from position: 0
1 jumps found. (Code = 62) Position 1 = -2
filename:       /tmp/call-dynamic.php
function name:  (null)
number of ops:  8
compiled vars:  !0 = $foo, !1 = $bar
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   3     0  E >   NEW                                              $2      :-5
         1        DO_FCALL                                      0          
         2        ASSIGN                                                   !0, $2
   4     3        ASSIGN                                                   !1, 'format'
   6     4        INIT_METHOD_CALL                                         !0, !1
         5        SEND_VAL_EX                                              'U'
         6        DO_FCALL                                      0          
   7     7      > RETURN                                                   1

branch: #  0; line:     3-    7; sop:     0; eop:     7; out0:  -2
path #1: 0,
```
vs.
```$ php7.2 -dopcache.enable=1 -dopcache.enable_cli=1 -dvld.active=1 -dvld.execute=0 /tmp/call-array.php 
Finding entry points
Branch analysis from position: 0
1 jumps found. (Code = 62) Position 1 = -2
filename:       /tmp/call-array.php
function name:  (null)
number of ops:  10
compiled vars:  !0 = $foo, !1 = $bar
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   3     0  E >   NEW                                              $2      :-5
         1        DO_FCALL                                      0          
         2        ASSIGN                                                   !0, $2
   4     3        ASSIGN                                                   !1, 'format'
   6     4        INIT_ARRAY                                       ~2      !0
         5        ADD_ARRAY_ELEMENT                                ~2      !1
         6        INIT_DYNAMIC_CALL                                        ~2
         7        SEND_VAL_EX                                              'U'
         8        DO_FCALL                                      0          
   7     9      > RETURN                                                   1
```

So yes, the array variant has two more opcodes + difference in INIT_METHOD_CALL vs. INIT_DYNAMIC_CALL, but is still more readable and I'd use it if the code in question is not performance sensitive.

> ```php
> $callbacks = [$mainCallback];
> $entity->$mainCallback($event);
> $entity->$callbacks[0]($event);
> ```

These are tricky, sure. The last one especially is pretty confusing and doesn't do what you'd expect (it won't call $mainCallback, but throw array to string conversion).

Ocramius:
@Majkl578 yes, I was just being clear on its (minimal) disadvantages.

Anyway, taking a stance here and merging this. If we ever want to step back and change course on this we should do so in the `doctrine/coding-standard` repo.

Ocramius (approved):
:ship:

Issue #6942, opened 2017-12-28
SQL Function ROUND is not parsed in Query\Parser


cgardiennet:
The SQL function ROUND is not parsed in Query\Parser.
Need to be added.

cgardiennet:
Need this in a future 2.5 version

Ocramius:
@cgardiennet this is available in [DoctrineExtensions](https://github.com/beberlei/DoctrineExtensions/blob/f592c495f1c43851f5e3418cf9c2514e61d67e64/src/Query/Mysql/Round.php)

cgardiennet:
The solution to use DoctrineExtension is not good one for me :

Ocramius:
> Round function class is not created for every database platform (ie : doesn't exist for PostgreSql ???!!!)

Yep, you should send a patch to `DoctrineExtensions` then :-)

Patches here also welcome, as long as behavior is consistent on all platforms.

cgardiennet:
The most important annoying fact for me is that i have to declare a custom DQL function (ROUND) when this is a standard SQL function...

Ocramius:
That holds true for most SQL functions. DQL is kinda like a subset of SQL, and we can only expand it where functionality applies to all DB vendors.

cgardiennet:
ROUND exist in all DB !

cgardiennet:
And most important, it is implemented in Doctrine DBAL for all platforms !

Ocramius:
Send a patch then :+1:

cgardiennet:
This is what i tried to do : https://github.com/doctrine/doctrine2/pull/6954

PR #6954, opened 2018-01-02
#6942 SQL Function ROUND is not parsed in Query\Parser

cgardiennet:
Branch created from doctrine/doctrine2/2.5
Need to have this feature in next 2.5 version

### DIFF
--- a/lib/Doctrine/ORM/Query/AST/Functions/RoundFunction.php
+++ b/lib/Doctrine/ORM/Query/AST/Functions/RoundFunction.php
@@ -1,0 +1,72 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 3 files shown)


Ocramius:
@cgardiennet ORM 2.5 and 2.6 are closed for modification/improvement - you will need to send a PR to `develop` (3.x)

beberlei:
Functions can be added via `Doctrine\ORM\Configuration` registration, this is not needed as a core feature.

cgardiennet:
@ocramius I didn't take time to make a new PR to be merged on branch develop 2 years ago. I should have... 
@beberlei This is unbelievable that an SQL Core function won't be in Doctrine Core features. I understand more why some people are not using Doctrine at all. 
I'll do a fork myself and I'll will patch with all the missing SQL standard functions that I need. Doctrine is not at the level we can expect just because of answers like yours and a lack of quality in functionalities.

Ocramius:
> Doctrine is not at the level we can expect

You can't "expect" things from software developed by volunteers in their free time, sorry.

Besides that, @beberlei is correct, and the DQL functions are already extensible via configuration, so this doesn't need to live in core, especially if `ROUND()` semantics change across platforms.

cgardiennet:
Anyway I'm a volunteer too (even though I'm not a big contributor to the code) and also a consumer.

> especially if ROUND() semantics change across platforms.

Explain to me the purpose of a Database Abstraction Layer then, if not to able to work with different database platforms.

PR #6956, opened 2018-01-02
README: Update versions

Majkl578:
Coverage badge only works for master, Scrutinizer apparently doesn't know about other branches - should they be added manually?


Ocramius (approved):
LGTM :+1:

PR #6957, opened 2018-01-03
Fix getHash param

muglug:
Because it's incorrect, I think?


muglug:
Closing in favour a grouped PR

PR #6958, opened 2018-01-03
Remove impossible-to-hit condition

muglug:
Because we have a hard type check for `$coll`


muglug:
Closing in favour of a grouped PR

PR #6959, opened 2018-01-03
Fix bugs that Psalm found, and add/improve docblocks

muglug:
It catches a bunch of things that I think are fatal errors (too few arguments, attempting array access on an object etc.) and a few other things that make its type inference easier.

[Here's the config](https://github.com/muglug/doctrine2/blob/master/psalm.xml) that was used to find these issues.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/IterableResult.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/IterableResult.php
@@ -26,7 +26,7 @@ class IterableResult implements \Iterator
     private $key = -1;
 
     /**
-     * @var object|null
+     * @var mixed[]|false|null
      */
     private $current;
 
@@ -65,7 +65,7 @@ class IterableResult implements \Iterator
     }
 
     /**
-     * @return mixed
+     * @return mixed[]|false|null
      */
     public function current()
     {
--- a/lib/Doctrine/ORM/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadata.php
@@ -421,20 +421,17 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
      */
     public function validateAssociations() : void
     {
-        array_map(
-            function (Property $property) {
-                if (! ($property instanceof AssociationMetadata)) {
-                    return;
-                }
+        foreach ($this->declaredProperties as $property) {
+            if (! ($property instanceof AssociationMetadata)) {
+                continue;
+            }
 
-                $targetEntity = $property->getTargetEntity();
+            $targetEntity = $property->getTargetEntity();
 
-                if (! class_exists($targetEntity)) {
-                    throw MappingException::invalidTargetEntityClass($targetEntity, $this->className, $property->getName());
-                }
-            },
-            $this->declaredProperties
-        );
+            if (! class_exists($targetEntity)) {
+                throw MappingException::invalidTargetEntityClass($targetEntity, $this->className, $property->getName());
+            }
+        }
     }
 
     /**
@@ -1307,7 +1304,10 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
     public function addInheritedProperty(Property $property)
     {
         $inheritedProperty = clone $property;
-        $declaringClass    = $property->getDeclaringClass();
+        /** @var ClassMetadata $declaringClass */
+        $declaringClass = $property->getDeclaringClass();
+
+        assert($declaringClass instanceof ClassMetadata);
 
         if ($inheritedProperty instanceof FieldMetadata) {
             if (! $declaringClass->isMappedSuperclass) {
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -135,8 +135,10 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
         $this->completeIdentifierGeneratorMappings($classMetadata);
 
         if ($parent) {
-            if ($parent->getCache()) {
-                $classMetadata->setCache(clone $parent->getCache());
+            $parentCache = $parent->getCache();
+
+            if ($parentCache) {
+                $classMetadata->setCache(clone $parentCache);
             }
 
             if (! empty($parent->namedNativeQueries)) {
@@ -589,16 +591,15 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
                     $platform->quoteIdentifier($definition['sequenceName']),
                     $definition['allocationSize']
                 );
-                break;
 
             case GeneratorType::UUID:
                 return new Sequencing\UuidGenerator();
-                break;
 
             case GeneratorType::CUSTOM:
                 $class = $property->getValueGenerator()->getDefinition()['class'];
                 return new $class();
-                break;
         }
+
+        throw new \UnexpectedValueException('Platform ' . $property->getValueGenerator()->getType() . ' is not supported');
     }
 }
--- a/lib/Doctrine/ORM/OptimisticLockException.php
+++ b/lib/Doctrine/ORM/OptimisticLockException.php
@@ -16,8 +16,8 @@ class OptimisticLockException extends ORMException
     private $entity;
 
     /**
-     * @param string $msg
-     * @param object $entity
+     * @param string      $msg
+     * @param object|null $entity
      */
     public function __construct($msg, $entity)
     {
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -280,6 +280,7 @@ class BasicEntityPersister implements EntityPersister
      */
     protected function fetchVersionValue(VersionFieldMetadata $versionProperty, array $id)
     {
+        /** @var ClassMetadata $versionedClass */
         $versionedClass = $versionProperty->getDeclaringClass();
         $tableName      = $versionedClass->table->getQuotedQualifiedName($this->platform);
         $columnName     = $this->platform->quoteIdentifier($versionProperty->getColumnName());
@@ -817,6 +818,8 @@ class BasicEntityPersister implements EntityPersister
         $sourceClass       = $association->getDeclaringClass();
         $owningAssociation = $targetClass->getProperty($association->getMappedBy());
         $targetTableAlias  = $this->getSQLTableAlias($targetClass->getTableName());
+
+        assert($sourceClass instanceof ClassMetadata);
 
         foreach ($owningAssociation->getJoinColumns() as $joinColumn) {
             $sourceKeyColumn = $joinColumn->getReferencedColumnName();
--- a/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
@@ -8,6 +8,7 @@ use Doctrine\Common\Collections\Criteria;
 use Doctrine\DBAL\LockMode;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Mapping\AssociationMetadata;
+use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\FieldMetadata;
 use Doctrine\ORM\Mapping\GeneratorType;
 use Doctrine\ORM\Mapping\JoinColumnMetadata;
@@ -143,6 +144,7 @@ class JoinedSubclassPersister extends AbstractEntityInheritancePersister
         // Make sure the table with the version column is updated even if no columns on that
         // table were affected.
         if ($isVersioned) {
+            /** @var ClassMetadata $versionedClass */
             $versionedClass = $this->class->versionProperty->getDeclaringClass();
             $versionedTable = $versionedClass->getTableName();
 
@@ -183,6 +185,7 @@ class JoinedSubclassPersister extends AbstractEntityInheritancePersister
         $parentClass  = $this->class;
 
         while (($parentClass = $parentClass->getParent()) !== null) {
+            /** @var ClassMetadata $parentClass */
             $parentTable = $parentClass->table->getQuotedQualifiedName($this->platform);
 
             $this->conn->delete($parentTable, $id);
@@ -302,6 +305,7 @@ class JoinedSubclassPersister extends AbstractEntityInheritancePersister
 
         while (($parentClass = $parentClass->getParent()) !== null) {
             $conditions = [];
+            /** @var ClassMetadata $parentClass */
             $tableName  = $parentClass->table->getQuotedQualifiedName($this->platform);
             $tableAlias = $this->getSQLTableAlias($parentClass->getTableName());
             $joinSql   .= ' INNER JOIN ' . $tableName . ' ' . $tableAlias . ' ON ';
@@ -499,6 +503,7 @@ class JoinedSubclassPersister extends AbstractEntityInheritancePersister
 
         while (($parentClass = $parentClass->getParent()) !== null) {
             $conditions = [];
+            /** @var ClassMetadata $parentClass */
             $tableName  = $parentClass->table->getQuotedQualifiedName($this->platform);
             $tableAlias = $this->getSQLTableAlias($parentClass->getTableName());
             $joinSql   .= ' INNER JOIN ' . $tableName . ' ' . $tableAlias . ' ON ';
--- a/lib/Doctrine/ORM/Query/Exec/MultiTableDeleteExecutor.php
+++ b/lib/Doctrine/ORM/Query/Exec/MultiTableDeleteExecutor.php
@@ -35,10 +35,10 @@ class MultiTableDeleteExecutor extends AbstractSqlExecutor
      * {@internal Any SQL construction and preparation takes place in the constructor for
      *            best performance. With a query cache the executor will be cached. }}
      *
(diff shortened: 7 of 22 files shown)


Majkl578:
Hi, thanks for your PR, great timing. 🎖 We've merged develop (3.0) into master just recently and the work is still in progress.
Additionally I've also started upgrading code base to match the updated [Doctrine Coding Standard](https://github.com/doctrine/coding-standard/) (WIP [here](https://github.com/doctrine/doctrine2/compare/cs)). Once done, we will start porting code to 7.2-based scalar typehinting and more. After these tasks are finished, we can focus on more similar refactorings and improving SA compliance - and since we use [PHPStan](https://github.com/phpstan/phpstan), increase its analysis level.

muglug:
Thanks!

A bunch of these changes are bugfixes (as opposed to docblock changes). Would you prefer they be separated out for speedier consideration?

Majkl578:
Of course, bugfixes are always welcomed. 👍 But at this moment, I'd suggest to wait until we handle CS changes (to avoid further conflicts/rebases on both ends), I will ping you here once that is done (soon™).

muglug:
Ok, and thanks for forcing me to take the 7.2 leap. So bleeding edge that `spl_object_id` isn’t documented yet...

lib/Doctrine/ORM/OptimisticLockException.php
```diff
@@ -17,7 +17,7 @@ class OptimisticLockException extends ORMException
 
     /**
      * @param string $msg
-     * @param object $entity
+     * @param ?object $entity
```
Majkl578: Please use `object|null` here, `?` is invalid in phpDoc. (It will become a native type later.)

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -380,9 +380,11 @@ public function parse()
      * declarations (from ... x join ... y join ... z ...) appear in the query
      * as the hydration process relies on that order for proper operation.
      *
-     * @param AST\SelectStatement|AST\DeleteStatement|AST\UpdateStatement $AST
+     * @param AST\SelectStatement $AST
+     *
+     * @return void
      */
-    private function fixIdentificationVariableOrder($AST)
+    private function fixIdentificationVariableOrder(AST\SelectStatement $AST)
```
Majkl578: This one introduced some test failures, it must accept all three AST statement types as previously specified.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -978,8 +978,9 @@ private function executeUpdates($class)
                 continue;
             }
 
-            if ($preUpdateInvoke !== ListenersInvoker::INVOKE_NONE) {
-                $this->listenersInvoker->invoke($class, Events::preUpdate, $entity, new PreUpdateEventArgs($entity, $this->em, $this->getEntityChangeSet($entity)), $preUpdateInvoke);
+            if ($preUpdateInvoke != ListenersInvoker::INVOKE_NONE) {
```
Majkl578: Likely should stay strict.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -137,8 +137,10 @@ protected function doLoadMetadata(
         if ($parent) {
             $this->addInheritedNamedQueries($classMetadata, $parent);
 
-            if ($parent->getCache()) {
-                $classMetadata->setCache(clone $parent->getCache());
+            $parent_cache = $parent->getCache();
```
Majkl578: Please use camelCase here.

Majkl578 (changes requested):
Thanks for rebasing it before I even got here to ping you as promised. 💯 There are some new test failures though, can you please look into them? Thanks.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
muglug: sorry, house style creeping in

lib/Doctrine/ORM/OptimisticLockException.php
muglug: Absolutely, good to know

lib/Doctrine/ORM/UnitOfWork.php
muglug: Sorry, I think this was a merge error

muglug:
I've fixed those issues. And I think I haven't introduced any new issues?

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -979,7 +979,8 @@ private function executeUpdates($class)
             }
 
             if ($preUpdateInvoke !== ListenersInvoker::INVOKE_NONE) {
-                $this->listenersInvoker->invoke($class, Events::preUpdate, $entity, new PreUpdateEventArgs($entity, $this->em, $this->getEntityChangeSet($entity)), $preUpdateInvoke);
+                $changeSet = $this->getEntityChangeSet($entity);
+                $this->listenersInvoker->invoke($class, Events::preUpdate, $entity, new PreUpdateEventArgs($entity, $this->em, $changeSet), $preUpdateInvoke);
```
Majkl578: Still one new error there:
> 1) Doctrine\Tests\ORM\Functional\LifecycleCallbackTest::testLifecycleListener_ChangeUpdateChangeSet
> Exception: [Doctrine\ORM\NoResultException] No result was found for query although at least one row was expected.

`getEntityChangeSet()` returns a reference which gets lost during assignment here.

muglug: Oh sorry! That’s a bug in Psalm. Will fix.

muglug: Psalm bugfix: https://github.com/vimeo/psalm/commit/7fa541e39bce2423e517bd60b861cf73de522db7
PHPStan ticket: https://github.com/phpstan/phpstan/issues/757

lib/Doctrine/ORM/Internal/Hydration/IterableResult.php
```diff
@@ -65,7 +65,7 @@ public function next()
     }
 
     /**
-     * @return mixed
+     * @return mixed[]|false|null
```
Majkl578: `false` is not a type, it should be `bool` instead

lib/Doctrine/ORM/Internal/Hydration/IterableResult.php
```diff
@@ -26,7 +26,7 @@ class IterableResult implements \Iterator
     private $key = -1;
 
     /**
-     * @var object|null
+     * @var mixed[]|false|null
```
Majkl578: `false` is not a type, it should be `bool` instead

lib/Doctrine/ORM/Mapping/ComponentMetadata.php
```diff
@@ -154,7 +154,8 @@ public function hasProperty(string $propertyName) : bool
     }
 
     /**
-     * @return \ArrayIterator|ColumnMetadata[]
+     * @return \ArrayIterator
+     * @psalm-return \ArrayIterator<ColumnMetadata>
```
Majkl578: What was the problem here? `\ArrayIterator|ColumnMetadata[]` says _\ArrayIterator that contains ColumnMetadatas_. Wouldn't add any vendor-specific hints with non-standard syntax if possible. Unfortunately generic-like syntax (`type<K, V>`) is not well supported by SA tools and IDEs so it's not used even on places where it makes more sense (hashes vs. lists). :(

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -279,6 +280,7 @@ protected function assignDefaultVersionValue($entity, array $id)
      */
     protected function fetchVersionValue(VersionFieldMetadata $versionProperty, array $id)
     {
+        /** @var ClassMetadata */
```
Majkl578: `@var` needs the variable name.

lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
```diff
@@ -143,6 +144,7 @@ public function update($entity)
         // Make sure the table with the version column is updated even if no columns on that
         // table were affected.
         if ($isVersioned) {
+            /** @var ClassMetadata */
```
Majkl578: `@var` needs the variable name.

lib/Doctrine/ORM/Mapping/ClassMetadata.php
```diff
@@ -1346,6 +1346,7 @@ public function addProperty(Property $property)
     public function addInheritedProperty(Property $property)
     {
         $inheritedProperty = clone $property;
+        /** @var ClassMetadata */
```
Majkl578: `@var` needs the variable name.

Majkl578 (commented):
Just some details.
I suppose Psalm is more benevolent in interpreting some things like anonymous inline `@var`, but we should keep it consistent.
(I should probably experiment with Psalm sometime to know the background of some of these fixes.)

lib/Doctrine/ORM/Internal/Hydration/IterableResult.php
muglug: Counterpoint: `false` is absolutely a type, at least as far as PHP uses it. This one's backed up by phpDocumentor: https://docs.phpdoc.org/guides/types.html

Moreover, an understanding of `false` is necessary to properly check code like
```php
function takesInt(int $i) : void {}

function foo(string $s) : void {
  $a_pos = strpos($s, "a");
  
  if ($a_pos === false) {
    return;
  }
  
  takesInt($a_pos);
}
```

lib/Doctrine/ORM/Mapping/ComponentMetadata.php
muglug: > `\ArrayIterator|ColumnMetadata[]` says *\ArrayIterator that contains ColumnMetadatas*

As I understand union types, `A|B` means either `A` or `B` are acceptable values. So that implies that an `array` is an acceptable value, which is not the case here.

The type given is just what PHPStorm needs to have an idea about what's being iterated over, and that's fairly vendor-specific, just not explicitly so.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
muglug: Sure, but also I realise that `@var` should [only be used for properties](https://docs.phpdoc.org/references/phpdoc/tags/var.html), according to phpDoc (Psalm and PHPStorm support it, but that support is outside the spec).

Maybe I should replace with an `assert`?

lib/Doctrine/ORM/Internal/Hydration/IterableResult.php
Majkl578: > Counterpoint: false is absolutely a type

It's not a real _type_, type is _bool_; phpDocumentor describes it as _keyword_ that is allowed in type specification as a specialization for bool.
That code should really use array|null, not false - that is php4-ism just like most builtin functions (and that may be the reason why phpDocumentor allows it).

```php
function foo() : false {} // fatal error - https://3v4l.org/b9imQ
```

lib/Doctrine/ORM/Mapping/ComponentMetadata.php
Majkl578: Yes, you're right. But unfortunately, due to limitation of current IDEs and SA/CS tools, it's not currently possible to express these properly. So the least troublesome variant is _abusing_ union syntax for now. I really wish this gets improved in future, we already see increasing adoption of Psalm and PhpStorm which leverage better type systems, let's hope PhpStorm will not be ignorant in this manner and improve too. 👍 
So far, generics syntax is not well supported and intersection syntax is completely unsupported (i.e. PhpStorm fails to interpret both).

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
Majkl578: Wouldn't be pedantic about this one, it's generally well supported.

lib/Doctrine/ORM/Internal/Hydration/IterableResult.php
muglug: > It's not a real type, type is bool;

You're right, I'm wrong.

> That code should really use `array|null`

Truly. `false` is an abomination. But it's one we're saddled with?

lib/Doctrine/ORM/Mapping/ComponentMetadata.php
muglug: > due to limitation of current IDEs and SA/CS tools

*some* SA/CS tools...

> So the least troublesome variant is abusing union syntax for now

I'm not sure that follows. Surely the least troubling variant is just something like
```
 @return \ArrayIterator of ColumnMetadata
```

Nothing is violated there.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
muglug: ok!

lib/Doctrine/ORM/Internal/Hydration/IterableResult.php
muglug: If we have `bool`, then analysis of the `foreach` is confused by a possibly `true` value for `$row`:

```php
function foo(IterableResult $i) {
  $row = $i->current();
  if ($row) {
    foreach ($row as $column) {}
  }
}

**Comment by muglug** [comment]
> I should probably experiment with Psalm

https://github.com/muglug/doctrine2 has the commits from all my PRs and a config that suppresses all other issues Psalm finds. Its config ignores function docblocks, using only PHP typehints and property docblocks to understand the code. That narrowing removes 200+ errors where the violated contracts are just defined in function docblocks.

**Comment by Majkl578** [comment]
Just fixed the build in master so after rebasing you should get relevant build status here. Then we can finish the rest.

File: lib/Doctrine/ORM/Query/Exec/MultiTableDeleteExecutor.php [hunk 12]
```diff
@@ -35,10 +35,10 @@ class MultiTableDeleteExecutor extends AbstractSqlExecutor
      * {@internal Any SQL construction and preparation takes place in the constructor for
      *            best performance. With a query cache the executor will be cached. }}
      *
-     * @param \Doctrine\ORM\Query\AST\Node  $AST       The root AST node of the DQL query.
+     * @param \Doctrine\ORM\Query\AST\DeleteStatement  $AST       The root AST node of the DQL query.
      * @param \Doctrine\ORM\Query\SqlWalker $sqlWalker The walker used for SQL generation from the AST.
```
Majkl578: This will need realignment.

lib/Doctrine/ORM/Query/Exec/MultiTableUpdateExecutor.php
```diff
@@ -46,10 +46,10 @@ class MultiTableUpdateExecutor extends AbstractSqlExecutor
      * {@internal Any SQL construction and preparation takes place in the constructor for
      *            best performance. With a query cache the executor will be cached. }}
      *
-     * @param \Doctrine\ORM\Query\AST\Node  $AST       The root AST node of the DQL query.
+     * @param \Doctrine\ORM\Query\AST\UpdateStatement  $AST       The root AST node of the DQL query.
      * @param \Doctrine\ORM\Query\SqlWalker $sqlWalker The walker used for SQL generation from the AST.
```
Majkl578: This will need realignment.

lib/Doctrine/ORM/Mapping/ComponentMetadata.php
Majkl578: I think the original return type should be kept, at least until we decide to start using intersections - which won't likely happen before being supported in IDEs.

lib/Doctrine/ORM/Internal/Hydration/IterableResult.php
Majkl578: Ok, keep false, we can revisit in future.

lib/Doctrine/ORM/Mapping/ComponentMetadata.php
muglug: `ArrayIterator&ColumnMetadata[]` isn't a valid intersection type either (as an array can never be an `ArrayIterator` object.

But anyway, I'll revert this change and [add support for this annotation to Psalm](https://github.com/vimeo/psalm/issues/479) – if nothing else, this back and forth tells me that PHPStorm users really like this convention.

muglug:
> Then we can finish the rest.

I've fixed the style issues - want me to squash commits?

Majkl578:
> want me to squash commits?

Not sure what is general policy here, but I'd prefer that much more than 10 fix/add commits. 👍 

Also please let's not include Psalm requirement & psalm.xml in this PR. May I suggest to move it i.e. to Gist or separate repo, or at least separate PR? Thanks.

muglug:
> Also please let's not include Psalm requirement & psalm.xml in this PR. May I suggest to move it i.e. to Gist or separate repo, or at least separate PR? Thanks.

God sorry, not my intention – I rebased off my fork's master a few hours ago (which is why the backdated commits have only just appeared)

muglug:
And any Psalm PR is a while away. This is the current state of the build: https://travis-ci.org/muglug/doctrine2/jobs/335491979

muglug:
Closing as this isn't particularly useful now

jwage:
@muglug is this not something you want to get merged? Happy to help push it through if it is still relevant.

muglug:
It's bound to be outdated - I think a better option is to use Psalm's baseline functionality to encourage small positive changes: https://github.com/doctrine/orm/pull/7691

PR #6960, opened 2018-01-03
fix UPGRADE instruction to correct version - `AbstractEntityInheritancePersister#getSelectJoinColumnSQL()` was changed in 2.6.0, not 2.5.0

hultberg:
For v2.6.0, $className argument was removed from `AbstractEntityInheritancePersister#getSelectJoinColumnSQL()` via #6216. However, the documented minor BC break for upgrading to v2.6 was put under Upgrade to 2.5. This commit fixes this by moving it to v2.6


Ocramius:
@hultberg thanks!

PR #6962, opened 2018-01-03
Use Null Coalesce Operator

carusogabriel:
Following #6874

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
@@ -494,11 +494,10 @@ class DatabaseDriver implements MappingDriver
      */
     private function getClassNameForTable($tableName)
     {
-        if (isset($this->classNamesForTables[$tableName])) {
-            return $this->namespace . $this->classNamesForTables[$tableName];
-        }
-
-        return $this->namespace . Inflector::classify(strtolower($tableName));
+        return $this->namespace . (
+            $this->classNamesForTables[$tableName]
+                ?? Inflector::classify(strtolower($tableName))
+        );
     }
 
     /**
--- a/lib/Doctrine/ORM/Repository/DefaultRepositoryFactory.php
+++ b/lib/Doctrine/ORM/Repository/DefaultRepositoryFactory.php
@@ -25,11 +25,8 @@ final class DefaultRepositoryFactory implements RepositoryFactory
     {
         $repositoryHash = $entityManager->getClassMetadata($entityName)->getClassName() . spl_object_id($entityManager);
 
-        if (isset($this->repositoryList[$repositoryHash])) {
-            return $this->repositoryList[$repositoryHash];
-        }
-
-        return $this->repositoryList[$repositoryHash] = $this->createRepository($entityManager, $entityName);
+        return $this->repositoryList[$repositoryHash]
+                ?? $this->repositoryList[$repositoryHash] = $this->createRepository($entityManager, $entityName);
     }
 
     /**
--- a/lib/Doctrine/ORM/Query/Filter/SQLFilter.php
+++ b/lib/Doctrine/ORM/Query/Filter/SQLFilter.php
@@ -97,11 +97,7 @@ abstract class SQLFilter
      */
     final public function hasParameter($name)
     {
-        if (! isset($this->parameters[$name])) {
-            return false;
-        }
-
-        return true;
+        return isset($this->parameters[$name]);
     }
 
     /**


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -475,11 +475,7 @@ public function & getEntityChangeSet($entity)
         $oid  = spl_object_id($entity);
         $data = [];
 
-        if (!isset($this->entityChangeSets[$oid])) {
-            return $data;
-        }
-
-        return $this->entityChangeSets[$oid];
+        return $this->entityChangeSets[$oid] ?? $data;
```
Ocramius: This change is incorrect: the method returns a reference, so an expression is not valid when combined with `return`

Ocramius (changes requested):

lib/Doctrine/ORM/UnitOfWork.php
carusogabriel: Should I rollback so?

Ocramius: Yeap

lib/Doctrine/ORM/Repository/DefaultRepositoryFactory.php
```diff
@@ -28,11 +28,8 @@ public function getRepository(EntityManagerInterface $entityManager, $entityName
     {
         $repositoryHash = $entityManager->getClassMetadata($entityName)->getClassName() . spl_object_id($entityManager);
 
-        if (isset($this->repositoryList[$repositoryHash])) {
-            return $this->repositoryList[$repositoryHash];
-        }
-
-        return $this->repositoryList[$repositoryHash] = $this->createRepository($entityManager, $entityName);
+        return $this->repositoryList[$repositoryHash] ??
+            $this->repositoryList[$repositoryHash] = $this->createRepository($entityManager, $entityName);
```
Slamdunk: I don't like this change: it's easy to mess with operator precedence this way.

Ocramius: This is quite standard

Slamdunk: Btw in this context I would prefer:
```php
if (!isset($this->repositoryList[$repositoryHash])) {
    $this->repositoryList[$repositoryHash] = $this->createRepository($entityManager, $entityName);
}

return $this->repositoryList[$repositoryHash];
```
So Setter and Getter parts are strictly detached

Ocramius: I moved all my lazy getters to following:

```php
public function getFoo()
{
    return $this->foo
       ?? $this->foo = $this->makeFoo();
}
```

The intent is clearer and the expression is quite simple

Ocramius: See for example https://github.com/Ocramius/ProxyManager/blob/81d53b2878f1d1c40ad27270e64b51798485dfc5/src/ProxyManager/Configuration.php :-)

Slamdunk: Ok then.

@carusogabriel may you move the `??` operator to the new line so conjunctions operators are easy to spot?

carusogabriel: Sure. I’m on city now. I’ll update my PR at Doctrine’s projects this afternoon 😄

carusogabriel:
@Ocramius @Slamdunk Every requested change done :+1:

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -524,11 +524,10 @@ private function getTablePrimaryKeys(Table $table)
      */
     private function getClassNameForTable($tableName)
     {
-        if (isset($this->classNamesForTables[$tableName])) {
-            return $this->namespace . $this->classNamesForTables[$tableName];
-        }
-
-        return $this->namespace . Inflector::classify(strtolower($tableName));
+        return $this->namespace . (
+            $this->classNamesForTables[$tableName] ??
```
Majkl578: This ?? should also be at the beginning of next line.

carusogabriel: Should I indent?
```php
$this->classNamesForTables[$tableName]
    ?? Inflector::classify(strtolower($tableName))
```

Majkl578: OF course, 4 spaces as usual. 👍

Majkl578 (approved):

Majkl578:
Waiting for resolution of #6964 as this will likely conflict.

Majkl578:
This will need a rebase now, sorry. :/

carusogabriel:
No problem, I’ll rebase and solve the conflicts

Majkl578 (approved):

PR #6964, opened 2018-01-05
[CS] Ported lib/ to updated CS 2.0

Majkl578:
Beware, tons of changes. Most of them automatically fixed by phpcbf.

Most of the changes should be harmless except some weak->strict comparison changes (one was kept because juggling is part of expected behavior (really), hopefully we can eliminate it later by explicit casts).

Disabled:
* native parameter and return type declaration checks - BC breaking, will do in phase two (needs to be propagated to common libs as well)
* Throwable check - Exception is still used on multiple places

Tried to narrow down as many iterables and mixeds as possible (manually), but it's still a mixed land more than I'd like. (Whoever invented wide `array` type in phpDoc standard killed some kittens btw.)

Made lib/ check mandatory on CI (although now irrelevant until we fix CI).
https://travis-ci.org/doctrine/doctrine2/jobs/325277840

∑ 3,599 additions, 7,202 deletions.

Future scope:
* rebase?
* port tests/
* native parameters and native types, moving from phpDoc to native declaration


Ocramius:
If I may add: "argh".

Majkl578:
@Ocramius I'd like to merge this today so it doesn't block @guilhermeblanco's work and other PRs. Will happily address any post-merge review anytime later (it's master-only after all). Are you ok with that?

Ocramius (approved):
:shipit:

goetas:
huge!

PR #6967, opened 2018-01-07
replace 'internal note:' with {@internal .. }}

egircys:
This PR replaces the `Internal note:` phpdoc part with the `{@internal .. }}` syntax. Like specified in 
https://docs.phpdoc.org/references/phpdoc/tags/internal.html

Also, some methods have `INTERNAL:` https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/PersistentCollection.php#L106

I'm not really sure if these methods are meant to be `@internal` or the following text is a internal note? This would require another PR to fix imho.


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2105,7 +2105,7 @@ public function newInstance(ClassMetadata $class)
      * INTERNAL:
      * Creates an entity. Used for reconstitution of persistent entities.
      *
-     * Internal note: Highly performance-sensitive method.
+     * {@internal Highly performance-sensitive method. }}
```
Ocramius: Did something go wrong with closing brackets?

lib/Doctrine/ORM/Configuration.php
```diff
@@ -33,7 +33,7 @@
  * Configuration container for all configuration options of Doctrine.
  * It combines all configuration options from DBAL & ORM.
  *
- * Internal note: When adding a new configuration option just write a getter/setter pair.
+ * {@internal When adding a new configuration option just write a getter/setter pair. }}
```
Ocramius: We reverted these because IDEs started marking entire classes and methods as internal (warnings everywhere). How does this appear now in something like PHPStorm?

Ocramius (changes requested):

lib/Doctrine/ORM/UnitOfWork.php
egircys: From the phpdoc reference `To increase readability and ease parsing should the tag be terminated with a double closing brace, instead of a single one.`

lib/Doctrine/ORM/Configuration.php
egircys: I believe it's related to this https://github.com/Elgg/Elgg/issues/7714
PHPStorm show it strike through only when `@internal` (first version in phpdoc reference) is given :)

egircys: image

Majkl578: Configuration is definitely not an internal class (regardless how it looks in PhpStorm), the note is only meant to be a hint for anyone adding new options.

egircys: Right, the class itself is not marked internal, just the text for adding the new options.

lib/Doctrine/ORM/UnitOfWork.php
Ocramius: 👍

lib/Doctrine/ORM/Configuration.php
Ocramius: The problem is just when the class is used - does the IDE still strike-through the name?

egircys: With PhpStorm 2017.3.1 It's only strike-through in this case
```
/**
 * Configuration container for all configuration options of Doctrine.
 * It combines all configuration options from DBAL & ORM.
 *
 * {@internal When adding a new configuration option just write a getter/setter pair. }}
 *
 * @internal
 */
```

If anything else comes after the second `@internal` in the example, it's not striked-through anymore

Ocramius: Verified locally 👍

Ocramius:
:ship: thanks @egircys!

lib/Doctrine/ORM/UnitOfWork.php
Majkl578: This really looks like a typo/bug, I would prefer not doubling it if it is not required (I see "should" above)...

egircys: True, it does not look obvious from the first look. Here's a discussion about it https://github.com/phpDocumentor/fig-standards/pull/80. Until it's not changed, I would stick to the current syntax, tbh.

PR #6974, opened 2018-01-09
Use $strict param in functions that have it

carusogabriel:
Some functions, e.g. `in_array`, have a third param to enable strict comparison :shield:

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3192Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3192Test.php
@@ -155,7 +155,7 @@ class DDC3192CurrencyCode extends Type
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
-        return array_search($value, self::$map);
+        return array_search($value, self::$map, true);
     }
 
     /**
--- a/lib/Doctrine/ORM/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadata.php
@@ -622,7 +622,7 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
                 throw MappingException::sqlConversionNotAllowedForPrimaryKeyProperties($property);
             }
 
-            if (! in_array($fieldName, $this->identifier)) {
+            if (! in_array($fieldName, $this->identifier, true)) {
                 $this->identifier[] = $fieldName;
             }
         }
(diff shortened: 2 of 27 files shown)


lcobucci (changes requested):
IMO we should also use https://github.com/phpstan/phpstan-strict-rules to enforce this

carusogabriel:
I used https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.9/src/Fixer/Strict/StrictParamFixer.php, but if you want, I can change the PR :sweat_smile:

lcobucci:
@carusogabriel of course we can solve this using various tools, but the point is that IMO doing the strict checks from phpstan is preferable to prevent bugs in the future

carusogabriel:
@lcobucci I'm just installing PHP 7.2 locally, I'll modify this PR to use `phpstan-strict-rules` :smile:

Majkl578:
>  I'll modify this PR to use phpstan-strict-rules

I'm not sure whether it's wise at this moment. It will cause a lot of errors imho, especially implicit bool casts in conditions. :/ Lets postpone it for now and just switch on these `$strict` params, there could also be a PHPCS rule for that (requiring explicit value for `$strict` params).

Majkl578:
This will need a rebase now, sorry. :/

carusogabriel:
No problem, I’ll rebase and solve the conflicts

Majkl578 (approved):
Don't see any apparent coercion issue, so LGTM.

Majkl578:
@lcobucci @carusogabriel Let's address `phpstan-strict-rules` later, PHPStan build is currently red anyway and needs fixes first.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3192Test.php
```diff
@@ -155,7 +155,7 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
-        return array_search($value, self::$map);
+        return array_search($value, self::$map, true);
```
Majkl578: This change has actually broken the test:
```
1) Doctrine\Tests\ORM\Functional\Ticket\DDC3192Test::testIssue
Failed asserting that false matches expected 'BYR'.

/www/doctrine/doctrine2/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3192Test.php:55
```

There is likely string to integer coercion that doesn't work after enabling strict mode.

carusogabriel: Have you reverted the change or should I? Sorry about that!

Majkl578: Fixed it in d791f7f6ca77ab9acb282317664d1ea7316bae8d.

PR #6977, opened 2018-01-10
Use precise workers

greg0ire:
This should make a warning banner disappear. Maybe builds will be more
stable too, who knows?


greg0ire:
Oh wow

PR #6981, opened 2018-01-11
Fix exporter signatures

muglug:
All these classes extend another that has a wider type for the first argument of `exportInstantiation()`, and using these classes causes a fatal error in PHP.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Exporter/FieldMetadataExporter.php
+++ b/lib/Doctrine/ORM/Mapping/Exporter/FieldMetadataExporter.php
@@ -4,13 +4,14 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Mapping\Exporter;
 
+use Doctrine\ORM\Mapping\ColumnMetadata;
 use Doctrine\ORM\Mapping\FieldMetadata;
 
 class FieldMetadataExporter extends LocalColumnMetadataExporter
 {
     public const VARIABLE = '$property';
 
-    protected function exportInstantiation(FieldMetadata $metadata) : string
+    protected function exportInstantiation(ColumnMetadata $metadata) : string
     {
         $lines   = [];
         $lines[] = sprintf(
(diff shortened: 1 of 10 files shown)


Majkl578:
AFAIK this code is not yet in use (infrastructure for upcoming changes) so it doesn't trigger errors in tests. We were actually talking about these with @guilhermeblanco as well.

Manually triggered static analysis stage and it looks like there are some other signature problems, but we can address those separately: https://travis-ci.org/doctrine/doctrine2/jobs/327872765#L593

lib/Doctrine/ORM/Mapping/Exporter/FieldMetadataExporter.php
```diff
@@ -4,13 +4,14 @@
...
 
+use Doctrine\ORM\Mapping\ColumnMetadata;
 use Doctrine\ORM\Mapping\FieldMetadata;
 
 class FieldMetadataExporter extends LocalColumnMetadataExporter
 {
     public const VARIABLE = '$property';
 
-    protected function exportInstantiation(FieldMetadata $metadata) : string
+    protected function exportInstantiation(ColumnMetadata $metadata) : string
```
Majkl578: We should keep contravariance information either in `@param` or in `assert(... instanceof ...)` to help static analysis in future - i.e. `$metadata->getName()` would be undefined on ColumnMetadata.

muglug: I've added asserts, as I think that makes the most sense.

muglug:
This PR now also fixes that bad `void` return type.

Psalm found one other [potentially problematic return type](https://github.com/doctrine/doctrine2/blob/a0071b17deeb71d5542e82f74577258fc766d1fb/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L594) in a typehint – you may want to throw an exception at the end of the method.

Majkl578 (approved):
LGTM. I like the `assert()`s to maintain safety.

Ocramius (approved):
LGTM 🚢

Ocramius:
Thanks @muglug!

Majkl578:
Thanks @muglug, managed to fix some other issues and got PHPStan working again in #7013.

PR #6982, opened 2018-01-12
Make ClassMetadata (class) implement ClassMetadata (interface)

muglug:
I don't know if this is good or not. All I know is a ton of Psalm's errors disappear when I do this, because you're passing the class into a whole bunch of places that expect the interface.

And 4a427b8b7b8b15fd3eba2613ca388b047fc3b5a9 looks to be a bit of a smoking gun...


muglug:
This was dumb. A much larger change is needed.

Majkl578:
These are intentionally _disconnected_ in master to allow API changes. Once API is stabilized and updated in Persistence as well, it will be _connected_ again. Consider master as WIP for now, especially in the Mapping part.

PR #6983, opened 2018-01-13
Fix property docblocks where type is easy to infer

muglug:
Again, picked up by Psalm


Ocramius (approved):
LGTM, thanks @muglug!

PR #6989, opened 2018-01-16
fix Catchable Fatal Error: Argument 3 passed to Doctrine\ORM\Event\Pr…

ibogdan94:
Hi, guys
When I want to remove entity, that has few relations, I get this error:

Catchable Fatal Error: Argument 3 passed to Doctrine\ORM\Event\PreUpdateEventArgs::__construct() must be of the type array, null given, called in /var/www/akela/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php on line 1065 and defined

After debugging and understanding what's going on, I found a solution

Hope, you can help me to fix this issue


Ocramius:
Please do check this against `2.6` - should already be fixed there.

ibogdan94:
@Ocramius thanks for reply 
Hope, you are right

PR #6990, opened 2018-01-17
Centralize class deprecations

greg0ire:
Don't review until #6886 is merged


lcobucci:
@greg0ire moving to 2.8.0 for the same reason as #6886

greg0ire:
I'll just close that one, it's not going to happen I'm afraid.

PR #6993, opened 2018-01-18
Updated docblocks

caciobanu:
As it can be seen from https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Internal/Hydration/SingleScalarHydrator.php#L24 **NoResultException** is thrown even when HYDRATE_SINGLE_SCALAR mode is used.


Ocramius (approved):
LGTM :+1:

PR #6995, opened 2018-01-19
Add information about PHP constants for default options

maks-rafalko:
Document feature about using PHP contanst in `options->default` annotation

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -118,7 +118,7 @@ Optional attributes:
 -  **options**: Array of additional options:
 
    -  ``default``: The default value to set for the column if no value
-      is supplied.
+      is supplied. It's possible to use PHP contstants here: ``options={"default": Domain::CONSTANT_VALUE}``
 
    -  ``unsigned``: Boolean value to determine if the column should
       be capable of representing only non-negative integers
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -118,7 +118,7 @@ Optional attributes:
 -  **options**: Array of additional options:
 
    -  ``default``: The default value to set for the column if no value
-      is supplied. It's possible to use PHP contstants here: ``options={"default": Domain::CONSTANT_VALUE}``
+      is supplied. It's possible to use PHP constants here: ``options={"default": Domain::CONSTANT_VALUE}``
 
    -  ``unsigned``: Boolean value to determine if the column should
       be capable of representing only non-negative integers


borNfreee:
Travis failures unrelated. Should I change the target branch from master to something from 2.*?

docs/en/reference/annotations-reference.rst
```diff
@@ -118,7 +118,7 @@ Optional attributes:
 -  **options**: Array of additional options:
 
    -  ``default``: The default value to set for the column if no value
-      is supplied.
+      is supplied. It's possible to use PHP contstants here: ``options={"default": Domain::CONSTANT_VALUE}``
```
lcobucci: Typo: `contstants`

lcobucci (changes requested):
I'm not sure if this should be here, it's related to how annotation parsing works (which is part of `doctrine/annotations` instead)

lcobucci:
> Travis failures unrelated. Should I change the target branch from master to something from 2.*?

Don't bother with that, your changes contains documentation only.

docs/en/reference/annotations-reference.rst
borNfreee: fixed

borNfreee:
> I'm not sure if this should be here, it's related to how annotation parsing works (which is part of `doctrine/annotations` instead)

I see your point (after reading the whole documentation about Annotations [here](http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html#constants 😄  )

But for average user, it won't be obvious (as it was for me). Maybe it would be nice to have this addition.


I like using cross references in documentation in my projects. From my point of view - they help users to find new features

We can even add something like ".. for more information see [Annotations Documentation](http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html#constants)"

docs/en/reference/annotations-reference.rst
```diff
@@ -118,7 +118,7 @@ Optional attributes:
 -  **options**: Array of additional options:
 
    -  ``default``: The default value to set for the column if no value
-      is supplied.
+      is supplied. It's possible to use PHP constants here: ``options={"default": Domain::CONSTANT_VALUE}``
```
Ocramius: It is possible to use php constants in any annotation value, so I don't really understand the point of this change...

Ocramius (changes requested):

borNfreee:
No problem, then I'm closing it.

The point was to let developers know that this is possible without reading the annotations documentation, which is I guess 90% of the cases when people working with doctrine - they just google how to describe default value and know nothing about constant feature

docs/en/reference/annotations-reference.rst
borNfreee: the key word is not "here", key word is a possibility of using constants which is not obvious from doctrine's documentation. this is just to let users know and redirect them to the full documentation

PR #6997, opened 2018-01-19
ManyToManyPersister fails to remove join table entry if there is multiple join columns

NicolaF:
This is due to a typo in `\Doctrine\ORM\Persisters\Collection\ManyToManyPersister::getDeleteSQLParameters`:

```php
        foreach ($mapping['relationToSourceKeyColumns'] as $columnName => $refColumnName) {
            $params[] = isset($sourceClass->fieldNames[$refColumnName])
                ? $identifier[$sourceClass->fieldNames[$refColumnName]]
                : $identifier[$sourceClass->getFieldForColumn($columnName)];
        }

```

It should be `$sourceClass->getFieldForColumn($refColumnName)`, as `$columnName` is the join table column, not the entity table column.

### DIFF
--- a/tests/Doctrine/Tests/Models/ManyToManyPersister/ChildClass.php
+++ b/tests/Doctrine/Tests/Models/ManyToManyPersister/ChildClass.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Doctrine\Tests\Models\ManyToManyPersister;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+
+/**
+ * @Entity
+ * @Table(name="manytomanypersister_child")
+ */
+class ChildClass
+{
+    /**
+     * @Id
+     * @Column(name="id1", type="integer")
+     *
+     * @var integer
+     */
+    public $id1;
+
+    /**
+     * @Id
+     * @ManyToOne(targetEntity=OtherParentClass::class, cascade={"persist"})
+     * @JoinColumn(name="other_parent_id", referencedColumnName="id")
+     *
+     * @var OtherParentClass
+     */
+    public $otherParent;
+
+    /**
+     * @ManyToMany(targetEntity=ParentClass::class, inversedBy="children")
+     * @JoinTable(
+     *     name="parent_child",
+     *     joinColumns={
+     *         @JoinColumn(name="child_id1", referencedColumnName="id1"),
+     *         @JoinColumn(name="child_id2", referencedColumnName="other_parent_id")
+     *     },
+     *     inverseJoinColumns={@JoinColumn(name="parent_id", referencedColumnName="id")}
+     * )
+     *
+     * @var Collection|ParentClass[]
+     */
+    public $parents;
+
+    public function __construct(int $id1, OtherParentClass $otherParent)
+    {
+        $this->id1         = $id1;
+        $this->otherParent = $otherParent;
+        $this->parents     = new ArrayCollection();
+    }
+}
--- a/tests/Doctrine/Tests/Models/ManyToManyPersister/ParentClass.php
+++ b/tests/Doctrine/Tests/Models/ManyToManyPersister/ParentClass.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Doctrine\Tests\Models\ManyToManyPersister;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Table;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\ManyToMany;
+
+/**
+ * @Entity
+ * @Table(name="manytomanypersister_parent")
+ */
+class ParentClass
+{
+    /**
+     * @Id
+     * @Column(name="id", type="integer")
+     *
+     * @var integer
+     */
+    public $id;
+
+    /**
+     * @ManyToMany(targetEntity=ChildClass::class, mappedBy="parents", orphanRemoval=true, cascade={"persist"})
+     *
+     * @var Collection|ChildClass[]
+     */
+    public $children;
+
+    public function __construct(int $id)
+    {
+        $this->id       = $id;
+        $this->children = new ArrayCollection();
+    }
+}
--- a/tests/Doctrine/Tests/ORM/Persisters/ManyToManyPersisterTest.php
+++ b/tests/Doctrine/Tests/ORM/Persisters/ManyToManyPersisterTest.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Persisters;
+
+use Doctrine\ORM\Persisters\Collection\ManyToManyPersister;
+use Doctrine\Tests\Mocks\ConnectionMock;
+use Doctrine\Tests\Models\ManyToManyPersister\ChildClass;
+use Doctrine\Tests\Models\ManyToManyPersister\OtherParentClass;
+use Doctrine\Tests\Models\ManyToManyPersister\ParentClass;
+use Doctrine\Tests\OrmTestCase;
+
+/**
+ * @covers \Doctrine\ORM\Persisters\Collection\ManyToManyPersister
+ */
+final class ManyToManyPersisterTest extends OrmTestCase
+{
+    /**
+     * @group 6991
+     * @group ManyToManyPersister
+     *
+     * @throws \Doctrine\ORM\ORMException
+     */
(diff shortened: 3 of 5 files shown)


tests/Doctrine/Tests/Models/ManyToManyPersister/ChildClass.php
```diff
@@ -0,0 +1,103 @@
+<?php
+/**
```
Ocramius: To be dropped

tests/Doctrine/Tests/Models/ManyToManyPersister/ChildClass.php
```diff
@@ -0,0 +1,103 @@
...
+    private $parents;
+
+    public function __construct() {
+        $this->parents = new ArrayCollection();
+    }
+
+    /**
+     * @return int
+     */
+    public function getId1(): int {
```
Ocramius: Remove all public API that isn't strictly needed. Use public properties instead

tests/Doctrine/Tests/Models/ManyToManyPersister/ParentClass.php
```diff
@@ -0,0 +1,72 @@
...
+    private $children;
+
+    public function __construct() {
+        $this->children = new ArrayCollection();
+    }
+
+    /**
+     * @return int
+     */
+    public function getId(): int {
```
Ocramius: Remove all public API that isn't strictly needed

tests/Doctrine/Tests/Models/ManyToManyPersister/ChildClass.php
```diff
@@ -0,0 +1,103 @@
...
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\ManyToMany;
+use Doctrine\ORM\Mapping\JoinColumn;
+use Doctrine\ORM\Mapping\JoinTable;
+/**
+ * Class ChildClass
+ * @package Doctrine\Tests\Models\ManyToManyPersister
+ * @Entity()
+ * @Table(name="child")
+ */
```
Ocramius: Prefix table names to avoid collisions with other test cases.

tests/Doctrine/Tests/Models/ManyToManyPersister/ChildClass.php
```diff
@@ -0,0 +1,103 @@
...
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Table;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\ManyToMany;
+use Doctrine\ORM\Mapping\JoinColumn;
+use Doctrine\ORM\Mapping\JoinTable;
+/**
+ * Class ChildClass
```
Ocramius: Redundant

tests/Doctrine/Tests/Models/ManyToManyPersister/ChildClass.php
```diff
@@ -0,0 +1,103 @@
...
+ * @package Doctrine\Tests\Models\ManyToManyPersister
```
Ocramius: Redundant

tests/Doctrine/Tests/Models/ManyToManyPersister/ParentClass.php
```diff
@@ -0,0 +1,72 @@
+<?php
+/**
```
Ocramius: To be removed

tests/Doctrine/Tests/ORM/Persisters/ManyToManyPersisterTest.php
```diff
@@ -0,0 +1,65 @@
+<?php
+/**
```
Ocramius: To be removed

tests/Doctrine/Tests/ORM/Persisters/ManyToManyPersisterTest.php
```diff
@@ -0,0 +1,65 @@
...
+
+        /** @var ConnectionMock $conn */
+        $conn = $this->em->getConnection();
+
+        $updates = $conn->getExecuteUpdates();
+
+        $lastUpdate = array_pop($updates);
+
+        $this->assertEquals('DELETE FROM parent_child WHERE child_id1 = ? AND child_id2 = ?', $lastUpdate['query']);
+        $this->assertEquals([1,2], $lastUpdate['params']);
```
Ocramius: Consider using fixed assigned identifiers instead of `AUTO_INCREMENT` to prevent issues with subsequent test runs.

tests/Doctrine/Tests/ORM/Persisters/ManyToManyPersisterTest.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->em->flush();
+
+        $childReloaded = $this->em->find(ChildClass::class, ['id1' => 1, 'id2' => 2]);
+
+        $this->persister->delete($childReloaded->getParents());
+
+        /** @var ConnectionMock $conn */
+        $conn = $this->em->getConnection();
+
+        $updates = $conn->getExecuteUpdates();
```
Ocramius: Wondering if a full mock connection would be better here, assuming that we completely get rid of that `EntityManager#flush()` call

tests/Doctrine/Tests/ORM/Persisters/ManyToManyPersisterTest.php
Ocramius: Errata: I see you are already doing that 👍

tests/Doctrine/Tests/ORM/Persisters/ManyToManyPersisterTest.php
```diff
@@ -0,0 +1,65 @@
...
+    /** @var ManyToManyPersister */
+    private $persister;
+
+    protected function setUp() {
+        parent::setUp();
+        $this->em = $this->_getTestEntityManager();
+        $this->persister = new ManyToManyPersister($this->em);
+    }
+
+    /**
```
Ocramius: Docblock redundant

Ocramius (changes requested):
Overall good patch. Comments are really just nits, but we may need to check whether removal of the `EntityManager#flush()` call is possible. This is done elsewhere via `UnitOfWork#registerManaged()` (which may not even be needed here)

NicolaF:
Hi @Ocramius,

I cleaned the test case as you requested. Howerver, I'm not sure to understand what you want to do with the `EntityManager#flush()` call. `ManyToManyPersister#delete` requires a `PersistentCollection` (because it needs the mapping to construct the delete statement), so `UnitOfWork#registerManaged()` won't do the job IMO.

Best Regards,
NicolaF

lcobucci (changes requested):
I've just executed your tests without the fix and it passed, this doesn't look like an issue IMO...

NicolaF:
@lcobucci Yes it looks like the test case is not correct. The original code is :
```php
            $params[] = isset($sourceClass->fieldNames[$refColumnName])
                ? $identifier[$sourceClass->fieldNames[$refColumnName]]
                : $identifier[$sourceClass->getFieldForColumn($columnName)];
```

When I stumble upon this issue, `isset($sourceClass->fieldNames[$refColumnName])` evaluated to `false` which caused the subsequent call to `$sourceClass->getFieldForColumn($columnName)`

which cannot work as `$columnName` is the column in the join table, not in the entity table.

I will investigate further and update the unit test to make it useful.

NicolaF:
Test case update. To actually reproduce the issue, we need :
* multiple join columns
* the join column and the referenced column have different names
* the referenced column is a foreign key.

In this case the `$referencedColumn` is not in $sourceCles

lcobucci:
@NicolaF I will check it in a bit, can you please execute `phpcbf` on the files you created to ensure they follow our coding standard?

NicolaF:
@lcobucci done

Ocramius:
@NicolaF I think this is good to go. @lcobucci can you give it another check? This would need merging to `2.6.1`, `2.7` and `master`

lcobucci (approved):

lcobucci:
> @NicolaF I think this is good to go. @lcobucci can you give it another check? This would need merging to 2.6.1, 2.7 and master

@Ocramius I've made some minor changes to simplify the SUT and organise the annotations. It LGTM now. The build is failing for nightly PHP and mysql/pgsql but I think this is fine.

lcobucci:
@NicolaF 🚢 thanks!

lcobucci:
@NicolaF the good thing is that your test already passes in the current state of `master`, but I've added to the suite to ensure we won't have any regression (via 79282317d1006f48e6016ba6876f298605eb207d).

NicolaF:
You're welcome!

PR #6998, opened 2018-01-19
FlintCI integration

soullivaneuh:
As suggested to @Majkl578, here is a proposition to integrate FlintCI for coding style review.

This tool will use your current CS configuration files and will publish PR status, diff comment and fix your code.

Another tools are available: https://flintci.io/docs

To finish this integration, just enable the repositories and re-open the PR to check if it works! :+1: 

I recommend to deactivate the "Comment on failure" this one is more useful for platform that does not support PR status, like GitLab.

As a side note, phpstan is also planned: https://gitlab.com/flintci/flintci/issues/133

Thanks!

### DIFF
--- a/.flintci.yml
+++ b/.flintci.yml
@@ -1,0 +1,5 @@
+services:
+  phpcs:
+    config_path: phpcs.xml.dist
+  phpcbf:
+    config_path: phpcs.xml.dist
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,13 +79,6 @@ jobs:
       before_script: wget https://phpbench.github.io/phpbench/phpbench.phar https://phpbench.github.io/phpbench/phpbench.phar.pubkey
       script: php phpbench.phar run -l dots --report=default
 
-    - stage: Code Quality
-      env: DB=none CODING_STANDARDS
-      php: 7.2
-      script:
-        - ./vendor/bin/phpcs lib
-        - ./vendor/bin/phpcs tests tools || true # disabled on CI until ported to new CS
-
   allow_failures:
     - php: nightly
 


Ocramius:
@Soullivaneuh do you have an example of how this works already on other projects? I'm being cautiously conservative on adding more CI tools here, especially since our travis pipeline already does all this stuff (except sending PRs to fix issues)

Soullivaneuh:
@Ocramius Hi and sorry for the delay.

You can see this PR as example: https://github.com/sonata-project/SonataAdminBundle/pull/4909

From here you can see the report download the patch directly if needed. Useful for people not familiar with cli commands.

On Pull Request, the report is also directly set as diff comment by FlintCI.

> except sending PRs to fix issues

Here is an auto-fix PR: https://github.com/sonata-project/SonataAdminBundle/pull/4900

It can also be a direct-push or simply nothing.

You may also change the repository access token if you want to setup a bot for the commit authoring instead of your account.

Of course, any suggestion is welcomed! :+1:

Ocramius:
@Soullivaneuh I'm going through backlog and found thia one. What I'd ideally want is to send PRs against contributor forks: is that something the platform can do?

Soullivaneuh:
> What I'd ideally want is to send PRs against contributor forks: is that something the platform can do?

You mean having FlintCI working on the forks?

Well, only if the fork owner subscribed it to FlintCI. Otherwise, the platform won't have access to it.

Did I well understand the question? :-)

Ocramius:
@Soullivaneuh clarifying on wished workflow:

 1. new contributor opens a patch against `doctrine/doctrine2`
 2. CS issues are detected in the patch
 3. auto-fixer PR is issued to the original PR
 4. contributor merges patch into their own branch

**NOT** wanted workflow:

 1. new contributor opens a patch against `doctrine/doctrine2`
 2. patch is merged
 3. CS issues are detected in the target branch
 3. auto-fixer PR is issued to the `doctrine/doctrine2` PR
 4. CS fix PR merged into the `doctrine/doctrine2` repository

Soullivaneuh:
@Ocramius The auto-fixer is for the subscribed repository only. In this case it will be `doctrine/doctrine2`.

But this auto-fix is not mandatory and can be disabled. In this case, the tool will just be useful for branch and PR report.

If you clean the CS at first case to make FlintCI green and refuse PR without FlintCI validation, you won't have to get any auto-fix, activated or not. The only case you might have auto-fix (if enabled) is when the chosen CS tools bring new fixer possibility, but this is not the responsibility of FlintCI.

If the PR author want an auto-fix on his fork, he has to subscribe it himself on the FlintCI platform.

Is that enough understandable? :-)

Ocramius:
> If the PR author want an auto-fix on his fork, he has to subscribe it himself on the FlintCI platform.

Makes sense.

We discussed this internally and this kind of behavior is really not what would simplify the workflow here, as we have a simple CI step that does the CS checks for us already, and we don't let our target branches ever deviate from the standard, so fixup PRs are not really wished.

Closing here as `won't fix` then, sorry!

PR #7003, opened 2018-01-23
Cannot persist new entity if primary key contains a foreign key and the referenced object is in state new and its id is not assigned

NicolaF:
As the id of the referenced entity is generated, when `UnitOfWork#persistNew` computes the primary key, it may contain some `null` values.

Because of https://github.com/doctrine/doctrine2/commit/1cb8d790b64f542406a576b878d67abe284e7275 , `UnitOfWork#addToIdentityMap` does not allow this anymore, which is a good thing.

We need to delay the insertion into `identityMap` and `entityIdentifiers` after the actual insert to ensure that we have a complete primary key, as we do with post-insert ID generators.

This kinda worked in 2.5, but in the end `identityMap` and `entityIdentifiers` contained wrong values anyway.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -939,8 +939,11 @@ class UnitOfWork implements PropertyChangedListener
 
                 $class->setIdentifierValues($entity, $idValue);
             }
-
-            $this->entityIdentifiers[$oid] = $idValue;
+            // Some identifiers may be foreign keys to new entities.
+            // In this case, we don't have the value yet and should treat it as if we have a post-insert generator
+            if(!in_array(null, $idValue, true)) {
+                $this->entityIdentifiers[$oid] = $idValue;
+            }
         }
 
         $this->entityStates[$oid] = self::STATE_MANAGED;
@@ -1033,11 +1036,15 @@ class UnitOfWork implements PropertyChangedListener
         $persister  = $this->getEntityPersister($className);
         $invoke     = $this->listenersInvoker->getSubscribedSystems($class, Events::postPersist);
 
+        $theseInsertions = [];
+
         foreach ($this->entityInsertions as $oid => $entity) {
 
             if ($this->em->getClassMetadata(get_class($entity))->name !== $className) {
                 continue;
             }
+
+            $theseInsertions[$oid] = $entity;
 
             $persister->addInsert($entity);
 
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1033,12 +1044,16 @@ private function executeInserts($class)
         $persister  = $this->getEntityPersister($className);
         $invoke     = $this->listenersInvoker->getSubscribedSystems($class, Events::postPersist);
 
+        $theseInsertions = [];
```
Ocramius: `$theseInsertions` is a bad name here

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1067,6 +1082,34 @@ private function executeInserts($class)
...
                 $this->addToIdentityMap($entity);
             }
+        } else {
+            foreach ($theseInsertions as $oid => $entity) {
+                if(!isset($this->entityIdentifiers[$oid])) {
+                    //entity was not added to identity map because some identifiers are foreign keys to new entities.
+                    //add it now
+                    $identifier = [];
+                    $idFields   = $class->getIdentifierFieldNames();
+                    foreach ($idFields as $idField) {
```
Ocramius: This block seems to detect whether an identifier is incomplete. Can it be moved to a `private` method?

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1067,6 +1082,34 @@ private function executeInserts($class)
 
                 $this->addToIdentityMap($entity);
             }
+        } else {
+            foreach ($theseInsertions as $oid => $entity) {
```
Ocramius: If the commit order calculator is working correctly, then this should simply be pushed down in the execution path, no?

Ocramius (changes requested):

Ocramius:
Related: https://github.com/doctrine/doctrine2/issues/7002
Related: https://github.com/doctrine/doctrine2/issues/7006

Please check if #7006 fixes this for you.

NicolaF:
Hi @Ocramius,
I'll make the requested changes.

This is actually not related to #7006. Here, the code path starts with `UnitOfWork#persist`
and what happens is:
* set identifier if possible
* schedule for insert
* add to identity map if identifier has been set
* cascade persist for all associations

The commit order calculator is not involved here, it is just basic tree walking.

Actually the commit order calculator does the job, otherwise the second part of the patch (in `executeInserts`) would not work : the referenced entity is correctly inserted first, and we can finally have a complete identifier for the referencing entity.

Ocramius:
Thanks for clarifying that!

tmbdrogba:
Is this PR ready to be merged ?

NicolaF:
It is IMHO, but I'm not the maintainer. @Ocramius, is it OK for you ?

ste93cry:
Actually I think this is breaking a lot of projects as it's not possible anymore after upgrading to 2.6 to persist any entity related to a new one in the same batch. Can we get this merged or at least a feedback to get it merged asap?

Ocramius:
I'm simply busy with other work stuff ATM - don't expect any deep involved OSS work from me for a while, sorry.

ste93cry:
Can we get at least another maintainer have a look at this? It's a really high priority bug to fix

lcobucci:
The code provided here doesn't follow our coding standards and we already had a test that reproduces this issue in #6701 (with a simpler set of entities too). I'll try to combine everything and adjust stuff to make this part of `v2.6.1` (also checking for any kind of performance impact).

lcobucci:
Fix was ported to #6701 and everything works as expected (with a minor, acceptable, performance impact) and it will be released on `v2.6.1` ASAP.

@NicolaF thank you for you help on the fix.

ste93cry:
This is a huge news, thank you for your time!

Issue #6976, opened 2018-01-09
Doctrine deletes all records when changing items in relation


Sharoff45:
I have 2 entities:
1. Project
2. Issue

Each Project can have many Issues related to it.

I use this code to filter the Issues:
~~~php

$filtered = []
foreach($project->getIssues() as $issue) {
   if(somethingIsTrue()) {
      $filtered[] = $issue;
      $issue->setAuthor('some-author');
   } else {
      $this->em->remove($issue);
   } 
}

$project->setIssues($filtered);
...
$this->em->flush();
~~~

In the end here is what I see in the log:

~~~sql
BEGIN TRANSACTION;
DELETE FROM issues WHERE project_id = ? [777];
UPDATE issues SET author = 'some-author' WHERE id = ? [123];
DELETE FROM issues WHERE id = ? [321];
COMMIT;
~~~

This output is not what I expected to see. I wouldn't expect it to delete all Issues within the Project. Could you please explain this behavior?

I also need to mention that this started happening only after we updated from Doctrine/ORM 2.5.14 to Doctrine/ORM 2.6.0

solexsylvain:
I didn't expect such a regression when upgrading. My database in production was flushing order items when saving the order. I lost a whole day on fixing the mess it created. my fault, I should have had more behavioral tests in place.

Ocramius:
This looks like an orphan removal misconfiguration issue. Can you isolate this behaviour in a test case?

solexsylvain:
I can probably set a test case but it will look the same as the code already posted up.  
Do you need something in particular ? Yml file, small code block ?
When I go back to version 2.5.14 as mentioned by Sharoff45 who reported the bug, everything is awesome.

Ocramius:
> I can probably set a test case but it will look the same as the code already posted up.

It should be something runnable, like the tests in https://github.com/doctrine/doctrine2/tree/ddb3cd17bdc771e0d8ef1eb1b441f140e3b80cd9/tests/Doctrine/Tests/ORM/Functional/Ticket

Sharoff45:
I'm created pull request with test: https://github.com/doctrine/doctrine2/pull/7011

Ocramius:
Thanks!

Ocramius:
@Sharoff45 @solexsylvain see #7011 discussion - this is indeed behaving as per design :-\

solexsylvain:
Hi Ocramius and Sharoff45,

[...]

PR #7011, opened 2018-01-26
#6976 Doctrine deletes all records when changing items in relation

Sharoff45:
- added tests

### DIFF
--- a/tests/Doctrine/Tests/Models/Issue6976/Issue6976Project.php
+++ b/tests/Doctrine/Tests/Models/Issue6976/Issue6976Project.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace Doctrine\Tests\Models\Issue6976;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+
+/**
+ * @Entity
+ * @Table(name="issue6976_project")
+ */
+class Issue6976Project
+{
+    /**
+     * @Id()
+     * @Column(type="integer")
+     * @GeneratedValue()
+     *
+     * @var int
+     */
+    private $id;
+
+    /**
+     * @var string
+     * @Column(name="title")
+     */
+    private $title;
+
+    /**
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/Models/Issue6976/Issue6976Project.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        return $this->issues;
+    }
+
+    /**
+     * @param Collection|Issue6976Issue[] $issues
+     *
+     * @return Issue6976Project
+     */
+    public function setIssues(iterable $issues)
```
Ocramius: This replaces the collection, essentially marking the existing `PersistentCollection` as removed

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue6976Test.php
```diff
@@ -0,0 +1,83 @@
...
+        foreach ($project->getIssues() as $issue) {
+            if (self::AUTHOR_1 === $issue->getAuthor()) {
+                $issue->setAuthor('new author');
+                $filtered[] = $issue;
+            } else {
+                $this->_em->remove($issue);
+            }
+        }
+
+        $project->setIssues($filtered);
```
Ocramius: This call removes all the existing `$project->issues` by replacing the existing `PersistentCollection` with an `array|Issue6976Issue[]`. This will mark the `PersistentCollection` entries as orphan-removed.

Ocramius:
Tests correctly reflect the issue report in #6976, but the ORM is behaving as per design here, since a collection with `orphanRemoval` on it is being replaced.

Ocramius:
Also: no, I don't think the design of the ORM around `orphanRemoval` is good here. It has always been a pain. For example, removing and re-adding an entity to a collection that has `orphanRemoval` will require it to be re-persisted to avoid it being removed. This should probably be changed in `3.x`...

lcobucci:
@Ocramius should we close this PR or leave it open?

Ocramius:
Closing here and linking #6976

PR #7012, opened 2018-01-26
Fixed updating of entity with quoted identifier join column

janlanger:
When entity has `JoinColumn` as its identifier, and this `JoinColumn` is quoted, the identifier is not quoted when building `UPDATE` query.

Quoting works fine when entity is inserted or selected...

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/QuotedIdentifierAssociationTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/QuotedIdentifierAssociationTest.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use Doctrine\Tests\Models\Quote\User;
+use Doctrine\Tests\Models\Quote\UserData;
+
+/**
+ * Tests that association with quoted JoinColumn is updated
+ */
+class QuotedIdentifierAssociationTest extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+
+	protected function setUp()
+	{
+		$this->useModelSet('quote');
+		parent::setUp();
+	}
+
+	public function testUpdateEntityWithIdentifierAssociationWithQuotedJoinColumn()
+	{
+		$user = new User();
+		$user->name = 'John Doe';
+		$this->_em->persist($user);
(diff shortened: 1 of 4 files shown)


janlanger:
Not really sure how to fix the failing tests... :(

tests/Doctrine/Tests/ORM/Functional/QuotedIdentifierAssociationTest.php
```diff
@@ -0,0 +1,37 @@
...
+		$userData->name = '123456789';
+		$userData->user = $user;
+		$this->_em->persist($userData);
+		$this->_em->flush();
+
+		$userData->name = '4321';
+		$this->_em->flush();
+
+		$queries = $this->_sqlLoggerStack->queries;
+		$this->assertSQLEquals('UPDATE `quote-user-data` SET `name` = ? WHERE `user-id` = ?', $queries[$this->_sqlLoggerStack->currentQuery - 1]['sql']);
```
Ocramius: Quoting depends on the platform, so the backticks are not valid with all the various phpunit setups we have for different platforms. In postgresql (and other standard RDBMSs), identifiers are quoted via `"`, for example.

The test will need to be adapted as following, in my opinion:

1. quote all these 3 identifiers (`quote-user-data`, `name` and `user-id`)
2. expect them to NOT be equals to the non-quoted version
3. assert on the actual performed query with the quoted identifiers passed in via `sprintf()`

Ocramius (changes requested):

janlanger:
@Ocramius I hope I understand you correctly :-) Updated.

Ocramius:
@janlanger looks good 👍 

Will just need few cleanups (reducing code size, mostly). Shall do on merge.

Ocramius:
@lcobucci can you take over the merging and merging upstream here?

lcobucci:
@Ocramius will do

lcobucci:
@janlanger I've synced the branch with the latest state of `v2.6` and simplified things to have a single test to port to `master`. I was going to override your branch but didn't have access so had to open a new PR (#7093) to solve that, which will be merged when we have everything green.

Thanks for your contribution!

PR #7013, opened 2018-01-27
Fix PHPStan stage

Majkl578:
Back on level 1. Six more to go (currently 1079 errors). 💯

### DIFF
--- a/lib/Doctrine/ORM/Mapping/EntityClassMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/EntityClassMetadata.php
@@ -115,7 +115,7 @@ abstract class EntityClassMetadata extends ComponentMetadata
      */
     public function __construct(string $className, ClassMetadataBuildingContext $metadataBuildingContext)
     {
-        parent::__construct($className);
+        parent::__construct($className, $metadataBuildingContext);
 
         $this->entityName = $className;
     }
@@ -184,7 +184,8 @@ abstract class EntityClassMetadata extends ComponentMetadata
      */
     public function addSubClass(SubClassMetadata $subClassMetadata) : void
     {
-        $superClassMetadata = $this->getSuperClass();
+        /** @var EntityClassMetadata|null $superClassMetadata */
+        $superClassMetadata = $this->getParent();
 
         while ($superClassMetadata !== null) {
             if ($superClassMetadata->entityName === $subClassMetadata->entityName) {
--- a/lib/Doctrine/ORM/Mapping/Exporter/VariableExporter.php
+++ b/lib/Doctrine/ORM/Mapping/Exporter/VariableExporter.php
@@ -6,8 +6,6 @@ namespace Doctrine\ORM\Mapping\Exporter;
 
 class VariableExporter implements Exporter
 {
-    public const INDENTATION =  '    ';
-
     /**
      * {@inheritdoc}
      */
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -1831,7 +1831,7 @@ class SqlWalker implements TreeWalker
     public function walkWhereClause($whereClause)
     {
         $condSql  = $whereClause !== null ? $this->walkConditionalExpression($whereClause->conditionalExpression) : '';
-        $discrSql = $this->generateDiscriminatorColumnConditionSql($this->rootAliases);
+        $discrSql = $this->generateDiscriminatorColumnConditionSQL($this->rootAliases);
 
         if ($this->em->hasFilters()) {
             $filterClauses = [];
--- a/lib/Doctrine/ORM/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadata.php
@@ -619,7 +619,7 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
             }
 
             if ($property->getType()->canRequireSQLConversion()) {
-                throw MappingException::sqlConversionNotAllowedForPrimaryKeyProperties($property);
+                throw MappingException::sqlConversionNotAllowedForPrimaryKeyProperties($this->className, $property);
             }
 
             if (! in_array($fieldName, $this->identifier, true)) {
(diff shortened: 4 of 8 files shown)


lib/Doctrine/ORM/Mapping/EntityClassMetadata.php
```diff
@@ -184,7 +184,8 @@ public function isReadOnly() : bool
      */
     public function addSubClass(SubClassMetadata $subClassMetadata) : void
     {
-        $superClassMetadata = $this->getSuperClass();
+        /** @var EntityClassMetadata|null $superClassMetadata */
+        $superClassMetadata = $this->getParent();
```
Majkl578: getSuperClass() doesn't exist so I assumed it's equivalent to getParent().

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -1831,7 +1831,7 @@ public function walkUpdateItem($updateItem)
     public function walkWhereClause($whereClause)
     {
         $condSql  = $whereClause !== null ? $this->walkConditionalExpression($whereClause->conditionalExpression) : '';
-        $discrSql = $this->generateDiscriminatorColumnConditionSql($this->rootAliases);
+        $discrSql = $this->generateDiscriminatorColumnConditionSQL($this->rootAliases);
```
muglug: Psalm doesn't bug you about this at any level, because *sigh* PHP. Should it be something to warn a dev about?

Majkl578: Well, PHP is case-insensitive when it comes to function/method/class names, but mixing them may have dire consequences (especially for classes and PSR-0/4).
I think check for case sensitive invocation is essential and should exist. But PHPStan does that on level 0 although it's not a bug and works fine for functions. Probably because this check is a side-effect of CallToNonExistentFunctionRule which checks for fatal error on level 0. I'd say it should be split and case-sensitivity check moved to different level.

muglug: > Well, PHP is case-insensitive when it comes to function/method/class names, but mixing them may have dire consequences (especially for classes and PSR-0/4)

Definitely has dire consequences for class names (with autoloaders) but I've yet to see it have any consequences for function/method names. Still, obviously bad code.

Ocramius: Yeah, this is more CS than static analysis stuff. Still good idea to keep the casing consistent 👍

lib/Doctrine/ORM/Mapping/Exporter/VariableExporter.php
```diff
@@ -6,8 +6,6 @@
 
 class VariableExporter implements Exporter
 {
-    public const INDENTATION =  '    ';
```
Ocramius: This is a BC break that needs documenting

Ocramius (changes requested):

lib/Doctrine/ORM/Mapping/Exporter/VariableExporter.php
Ocramius: I'm doing this.

Ocramius (approved):

Ocramius:
Manually merged via doctrine/doctrine2@c14dbc240935f229cf08be0677f784744e46f21d

lib/Doctrine/ORM/Mapping/Exporter/VariableExporter.php
Majkl578: Nope? It was causing fatal error and Exporter is 3.x only code. :blush:

Ocramius: Ah... I documented this. May need removal.

Majkl578:
Damn, PHPStan 0.9.2 was released in meantime and introduced one new failure: https://travis-ci.org/doctrine/doctrine2/jobs/334591697
I'll fix it in master directly.

lib/Doctrine/ORM/Query/SqlWalker.php
ondrejmirtes: Can you CC me on any suggestions/ideas for PHPStan in the future? :) I might move the case-sensitivity check on functions/methods to `strict-rules`...

ondrejmirtes:
It's hard and even questionable how to keep semver for static analysis tools - any bugfix is potentially a BC break.

For `0.x.y`, I do bugfixes in `y`, but bugfixes usually lead to better understanding of the code which can uncover new bugs and thus breaking the build, like in this case.

In `x`, I usually add new rules and features which can find new types of errors.

ondrejmirtes:
But at least the fact that new bugs are found is welcome and keeps developers on my side :) I still think that using `^0.9` version constraint is better than locking to a specific version and depriving yourselves of improvemnets.

Majkl578:
Fix prepared in #7020 since it's a regular bug.

muglug:
@ondrejmirtes 

> It's hard and even questionable how to keep semver for static analysis tools - any bugfix is potentially a BC break.

I don't think it's a violation of semver to fix a bug in the tool which improves its analysis. A bc-incompatible change for a static analysis tool, IMO, is one that changes the API – e.g. it invalidates entries in that tool's config, or that tool's plugin architecture or public API.

Majkl578:
I agree, but on the other hand, if your CI starts failing without you changing anything, it's sub-optimal as well.

Ocramius:
New inspections with the same config is a BC break when the exit code of
the program changes (that's the actual API change here).

We had a similar discussion on
https://github.com/doctrine/coding-standard/pull/18

muglug:
> New inspections with the same config

By new inspections do you mean when [new types of inspection](https://github.com/vimeo/psalm/blob/master/docs/issues.md) are introduced, or when a bug is fixed so that an existing inspection can be applied to more code?

For example, fixing [this bug](https://github.com/phpstan/phpstan/issues/747) doesn't count as a new inspection, but it will likely cause many new issues to be found.

Ocramius:
Yeah, it is a bit of a gray area, but if new results pop up it should go to
the next MAJOR IMO.

We're also not always able to do that either though.

The idea is simple: if you install a certain version of the tool, all minor
and patch versions should behave pretty much the same way since they got
introduced in that minor.

muglug:
Realistically, though, that would mean that every Psalm release would be a major release. And that could get quite silly.

Ocramius:
It could, but it is also not going to be a big issue for consumers, as the
tool isn't supposed to be a runtime dependency. The problem with releasing
many major versions is ecosystem stability, and these tools do not affect
the stability of their consumers.

PR #7015, opened 2018-01-27
Update EntityGenerator.php

bieleski:
When running the EntityGenerator, the comment column was not created


Ocramius:
Please send patches for bug-fixes against `2.6`. `2.5` will only receive security issue fixes.

Also, please make sure patches are covered by test additions.

PR #7020, opened 2018-01-29
SchemaValidator: Fixed bug in inverse JoinTable column name

Majkl578:
Fixes regular bug found by PHPStan 0.9.2 (slipped through #7013 which was built on top of 0.9.1 at that time).

@Ocramius Worth backporting to 2.6 as well (if relevant)?

This should make Travis green for master. 🎉

### DIFF
--- a/tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
+++ b/tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
@@ -59,7 +59,7 @@ class SchemaValidatorTest extends OrmTestCase
         $class1 = $this->em->getClassMetadata(InvalidEntity1::class);
         $class2 = $this->em->getClassMetadata(InvalidEntity2::class);
 
-        $ce = $this->validator->validateClass($class1);
+        $errors = $this->validator->validateClass($class1);
 
         $message1 = "The inverse join columns of the many-to-many table '%s' have to contain to ALL identifier columns of the target entity '%s', however '%s' are missing.";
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
```diff
@@ -200,6 +200,21 @@ public function testInvalidOrderByAssociationInverseSide()
             $ce
         );
     }
+
+    public function testInvalidReferencedJoinTableColumnIsNotPrimary() : void
+    {
+        $class = $this->em->getClassMetadata(InvalidEntity3::class);
+        $ce = $this->validator->validateClass($class);
```
lcobucci: We could have a better name for this, maybe `$errors`? I know that it's the same that we have in the implementation, but that doesn't mean it's a good choice 😂

lcobucci (changes requested):
Just a tiny thing

lcobucci (approved):

tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
Majkl578: Just a lazy copy&paste. Renamed all occurrences.

Majkl578:
:shipit: 

@lcobucci Let me know if we want the test backported to 2.6.

lcobucci:
It sounds like a good idea, I was a bit lazy to do it myself yesterday
(shame)

PR #7023, opened 2018-01-30
Avoid `returning` on void methods

malukenho:


Majkl578 (approved):
LGTM
We'll add native void return types later.

PR #7026, opened 2018-01-31
adds optional $message parameter

ibronit:
It makes possible to provide custom exception message to the exception.

### DIFF
--- a/lib/Doctrine/ORM/NoResultException.php
+++ b/lib/Doctrine/ORM/NoResultException.php
@@ -9,11 +9,13 @@ namespace Doctrine\ORM;
  */
 class NoResultException extends UnexpectedResultException
 {
+    public const DEFAULT_MESSAGE = 'No result was found for query although at least one row was expected.';
+    
     /**
      * Constructor.
      */
-    public function __construct()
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/NoResultException.php
```diff
@@ -9,11 +9,13 @@
  */
 class NoResultException extends UnexpectedResultException
 {
+    public const DEFAULT_MESSAGE = 'No result was found for query although at least one row was expected.';
```
Ocramius: This exception is not supposed to be thrown by userland code.

Ocramius (changes requested):

Ocramius:
Closing as `invalid`: please define your own exception types.

smatyas:
@Ocramius so why does this file have the same feature?
https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/NonUniqueResultException.php

Ocramius:
@smatyas probably shouldn't, tbh. /cc @greg0ire

Majkl578:
Also see #6743.

PR #7038, opened 2018-02-05
add option to bypass LimitSubqueryWalker Validation

smaignan:
Add option to bypass LimitSubqueryWalker Validation.
Useful when a filter is present on the relationship TO_MANY, turning it into ONE_TO_ONE


Ocramius:
If you need to go into this level of detail, it is much better to implement a custom paginator yourself rather than adding hacks to the original one ;-)

lcobucci:
Closing as per @Ocramius comment

PR #7040, opened 2018-02-05
Fix the result cache documentation

coudenysj:
Since #172 (#2861) the result cache is raw data, not hydrated entities.


Ocramius:
@coudenysj thanks! 👍

Issue #7043, opened 2018-02-06
HTTP 404 in a tutorial page


Cladis:
Link to [http://www.doctrine-project.org/jira/browse/DDC-213](DDC-213) in /blob/master/docs/en/tutorials/working-with-indexed-associations.rst results in 404

Cladis:
@Ocramius , I have no idea what the correct location is, so no. (Unless you want me to file a dummy pull request)

Cladis:
Ah. Sorry being slow.

Ocramius:
Handled in #7044

PR #7044, opened 2018-02-06
Fixing link to DDC-213 from Jira to Github

Cladis:
As per the desc; I previously reported this in #7043


Ocramius (approved):
LGTM 👍 

Thanks @Cladis!

PR #7046, opened 2018-02-06
[CS] Fix CS in tests

carusogabriel:

PR #7048, opened 2018-02-07
Remove redundant php-docs

malukenho:
I also took the liberty to add some php7 hints to it.

### DIFF
--- a/lib/Doctrine/ORM/Cache/Lock.php
+++ b/lib/Doctrine/ORM/Cache/Lock.php
@@ -7,9 +7,6 @@ namespace Doctrine\ORM\Cache;
 use function time;
 use function uniqid;
 
-/**
- * Cache Lock
- */
 class Lock
 {
     /**
@@ -22,11 +19,7 @@ class Lock
      */
     public $time;
 
-    /**
-     * @param string $value
-     * @param int    $time
-     */
-    public function __construct($value, $time = null)
+    public function __construct(string $value, int $time = null)
     {
         $this->value = $value;
         $this->time  = $time ?: time();
--- a/lib/Doctrine/ORM/Cache/QueryCacheKey.php
+++ b/lib/Doctrine/ORM/Cache/QueryCacheKey.php
@@ -32,18 +32,13 @@ class QueryCacheKey extends CacheKey
      */
     public $timestampKey;
 
-    /**
-     * @param string $hash      Result cache id
-     * @param int    $lifetime  Query lifetime
-     * @param int    $cacheMode Query cache mode
-     */
     public function __construct(
-        $hash,
-        $lifetime = 0,
-        $cacheMode = Cache::MODE_NORMAL,
+        string $cacheId,
+        int $lifetime = 0,
+        int $cacheMode = Cache::MODE_NORMAL,
         ?TimestampCacheKey $timestampKey = null
     ) {
-        $this->hash         = $hash;
+        $this->hash         = $cacheId;
         $this->lifetime     = $lifetime;
         $this->cacheMode    = $cacheMode;
         $this->timestampKey = $timestampKey;
--- a/lib/Doctrine/ORM/Query/AST/Functions/ConcatFunction.php
+++ b/lib/Doctrine/ORM/Query/AST/Functions/ConcatFunction.php
@@ -38,7 +38,7 @@ class ConcatFunction extends FunctionNode
             $args[] = $sqlWalker->walkStringPrimary($expression);
         }
 
-        return call_user_func_array([$platform, 'getConcatExpression'], $args);
+        return $platform->getConcatExpression(...$args);
     }
 
     /**
--- a/lib/Doctrine/ORM/Cache/Logging/CacheLoggerChain.php
+++ b/lib/Doctrine/ORM/Cache/Logging/CacheLoggerChain.php
@@ -8,9 +8,6 @@ use Doctrine\ORM\Cache\CollectionCacheKey;
 use Doctrine\ORM\Cache\EntityCacheKey;
 use Doctrine\ORM\Cache\QueryCacheKey;
 
-/**
- * Cache logger chain
- */
 class CacheLoggerChain implements CacheLogger
 {
     /**
--- a/lib/Doctrine/ORM/Cache/Logging/StatisticsCacheLogger.php
+++ b/lib/Doctrine/ORM/Cache/Logging/StatisticsCacheLogger.php
@@ -11,7 +11,6 @@ use function array_sum;
 
 /**
  * Provide basic second level cache statistics.
- *
  */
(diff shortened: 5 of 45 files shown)


Majkl578:
Please see https://github.com/doctrine/coding-standard/pull/24 and https://github.com/Majkl578/doctrine2/commit/c2b786d6d3e826b81bb862f5d7a94bf3ba1aa0de (WIP).

lib/Doctrine/ORM/Cache/QueryCacheKey.php
```diff
@@ -32,15 +32,10 @@ class QueryCacheKey extends CacheKey
      */
     public $timestampKey;
 
-    /**
-     * @param string $hash      Result cache id
```
Majkl578: It's ok to remove these, but the variable should be self-explaining. `$hash` is not such case.

lib/Doctrine/ORM/Event/OnFlushEventArgs.php
```diff
@@ -17,19 +17,11 @@ class OnFlushEventArgs extends EventArgs
...
-     */
     public function __construct(EntityManagerInterface $em)
     {
         $this->em = $em;
     }
 
-    /**
-     * Retrieve associated EntityManager.
-     *
-     * @return \Doctrine\ORM\EntityManagerInterface
```
Majkl578: This is incorrect - EntityManagerInterface becomes mixed.

lib/Doctrine/ORM/Event/PostFlushEventArgs.php
```diff
@@ -17,19 +17,11 @@ class PostFlushEventArgs extends EventArgs
...
-     */
     public function __construct(EntityManagerInterface $em)
     {
         $this->em = $em;
     }
 
-    /**
-     * Retrieves associated EntityManager.
-     *
-     * @return \Doctrine\ORM\EntityManagerInterface
```
Majkl578: This is incorrect - EntityManagerInterface becomes mixed.

lib/Doctrine/ORM/Event/PreFlushEventArgs.php
```diff
@@ -17,17 +17,11 @@ class PreFlushEventArgs extends EventArgs
...
-    /**
-     * Constructor.
-     */
     public function __construct(EntityManagerInterface $em)
     {
         $this->em = $em;
     }
 
-    /**
-     * @return \Doctrine\ORM\EntityManagerInterface
```
Majkl578: This is incorrect - EntityManagerInterface becomes mixed.

lib/Doctrine/ORM/Mapping/AssociationMetadata.php
```diff
@@ -51,10 +51,7 @@ class AssociationMetadata implements Property
     /** @var CacheMetadata|null */
     private $cache;
 
-    /**
-     * AssociationMetadata constructor.
-     */
-    public function __construct(string $name)
+     public function __construct(string $name)
```
Majkl578: Wrong indentation.

lib/Doctrine/ORM/Mapping/Driver/Annotation/AnnotationDriver.php
```diff
@@ -194,7 +194,7 @@ public function getAllClassNames()
                     new \RecursiveDirectoryIterator($path, \FilesystemIterator::SKIP_DOTS),
                     \RecursiveIteratorIterator::LEAVES_ONLY
                 ),
-                '/^.+' . preg_quote($this->fileExtension) . '$/i',
+                '/^.+' . preg_quote($this->fileExtension, '/') . '$/i',
```
Majkl578: This looks like an unrelated bugfix.

lib/Doctrine/ORM/Query/Exec/AbstractSqlExecutor.php
```diff
@@ -10,6 +10,7 @@
 /**
  * Base class for SQL statement executors.
  * @todo Rename: AbstractSQLExecutor
+ * ???
```
Majkl578: What's wrong here?

lib/Doctrine/ORM/Cache/Lock.php
```diff
@@ -19,20 +16,13 @@ class Lock
...
     {
         $this->value = $value;
         $this->time  = $time ?: time();
     }
 
-    /**
-     * @return \Doctrine\ORM\Cache\Lock
-     */
-    public static function createLockRead()
+    public static function createLockRead(): Lock
```
Majkl578: Let's not change public API just yet, will be done later everywhere + needs to be documented in UPGRADING.

Majkl578 (changes requested):
Looks ok, but some changes are redundant (either here or in CS 3.0 changes - see comment above).
Actually some of the removed comments here would make sense as part of the CS itself (ForbiddenComments rule), do you mind to send PR? 😉

lib/Doctrine/ORM/Query/Exec/AbstractSqlExecutor.php
malukenho: Seems like the class was already renamed. May we can drop the comment?

Majkl578: I don't think it was, the name is same, but casing is different `Sql` vs. `SQL`, hasn't been touched since 2009.

Majkl578:
@malukenho Hi, please rebase onto current master where CS 3.0 is applied (#7057). Then once comments are resolved, it's ready to go. :shipit:

malukenho:
@Majkl578 please, take another look at it.

lcobucci:
@malukenho we have some CS violations on your patch: https://travis-ci.org/doctrine/doctrine2/jobs/340480444#L554

Can you please check those?

carusogabriel:
@malukenho This is being doing in #7046 :)

lib/Doctrine/ORM/Query/AST/Functions/ConcatFunction.php
```diff
@@ -38,7 +38,7 @@ public function getSql(SqlWalker $sqlWalker)
             $args[] = $sqlWalker->walkStringPrimary($expression);
         }
 
-        return call_user_func_array([$platform, 'getConcatExpression'], $args);
+        return $platform->getConcatExpression(...$args);
```
carusogabriel: @Majkl578 @Ocramius Maybe something for our CS as well?

Ocramius: @carusogabriel not really that important, plus `call_user_func_array()` is quite magic, so I'd avoid fiddling with it at CS level.

Ocramius (approved):
LGTM :ship:

PR #7052, opened 2018-02-09
Null proxy data after clearing the UnitOfWork and loading same object via another proxy

BitOne:
Given 3 objects, childA, childB and parent. Parent is accessible from both child objects by the parent attribute.
When I load childA, and I clear the UnitOfwork, and I load childB, and I access its parent.
When I access childA parent,
Then childA parent have its data null.

I've created the following test that reproduces this issue.
As per your contributing guidelines, I know you prefer PR on master. But this test passes properly on master.
And it fails in 2.5, 2.6 and 2.7.

So if you prefer, I can do the PR on master, and then backport it to the 2.5, 2.6 and 2.7 version. But I'm not sure it would make sense.

Is there another process when the bug is only in a branch, but master has changed so much I can really do a cherry pick of a potential solution from master to that branch?

Thanks a lot!

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Tests\Models\CMS\CmsEmployee;
+
+/**
+ * @group issue-7052
+ */
+class Issue7052Test extends \Doctrine\Tests\OrmFunctionalTestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema(array(
+            $this->_em->getClassMetadata(__NAMESPACE__ . '\\Issue7052Child'),
+            $this->_em->getClassMetadata(__NAMESPACE__ . '\\Issue7052Parent'),
+        ));
+    }
+
+    public function testIssue()
+    {
+        $parent = new Issue7052Parent();
+        $parent->name = "parent test";
+
+        $childA = new Issue7052Child();
+        $childA->name = "child A";
+        $childA->parent = $parent;
+
+        $childB = new Issue7052Child();
+        $childB->name = "child B";
+        $childB->parent = $parent;
+
+        $this->_em->persist($parent);
+        $this->_em->persist($childA);
+        $this->_em->persist($childB);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $childA = $this->_em->find('Doctrine\Tests\ORM\Functional\Ticket\Issue7052Child', $childA->id);
+
+        $this->_em->clear();
+
+        $childB = $this->_em->find('Doctrine\Tests\ORM\Functional\Ticket\Issue7052Child', $childB->id);
+
+        $parentFromChildB = $childB->parent;
+
+        $this->assertNotNull($childA->parent->name, "Unable to get parent name after EM cleared.");
+    }
+}
+
+/**
+ * @Entity
+ * @Table(name="issuebitonexxx_child")
+ */
+class Issue7052Child
+{
+    /** @Id @GeneratedValue @Column(type="integer") */
+    public $id;
+
+    /**
+     * @Column
+     * @var string
+     */
+    public $name;
+
+    /**
+     * @ManyToOne(targetEntity="Issue7052Parent", fetch="LAZY")
+     * @JoinTable(
+     *   name="issuebtonexxx_parent",
+     *   inverseJoinColumns={@JoinColumn(name="parent_id", referencedColumnName="id")}
+     * )
+     */
+    public $parent;
+}
+
+/**
+ * @Entity
+ * @Table(name="issuebitonexxx_parent")
+ */
+class Issue7052Parent
+{
+    /** @Id @GeneratedValue @Column(type="integer") */
+    public $id;
+
+    /**
+     * @Column
+     * @var string
+     */
+    public $name;
+}
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
@@ -6,7 +6,7 @@ use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Tests\Models\CMS\CmsEmployee;
 
 /**
- * @group issue-7052
+ * @group 7052
  */
 class Issue7052Test extends \Doctrine\Tests\OrmFunctionalTestCase
 {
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * @Column
+     * @var string
+     */
+    public $name;
+
+    /**
+     * @ManyToOne(targetEntity="Issue7052Parent", fetch="LAZY")
+     * @JoinTable(
```
Ocramius: Please remove all mapping, properties and details that aren't strictly related to reproducing the bug

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
```diff
@@ -0,0 +1,93 @@
...
+        $this->_em->persist($childA);
+        $this->_em->persist($childB);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $childA = $this->_em->find('Doctrine\Tests\ORM\Functional\Ticket\Issue7052Child', $childA->id);
+
+        $this->_em->clear();
+
+        $childB = $this->_em->find('Doctrine\Tests\ORM\Functional\Ticket\Issue7052Child', $childB->id);
```
Ocramius: Can you add an assertion that verifies that this entity is different from the one created before `clear()`? A new variable name is needed.

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
```diff
@@ -0,0 +1,93 @@
...
+
+        $parentFromChildB = $childB->parent;
+
+        $this->assertNotNull($childA->parent->name, "Unable to get parent name after EM cleared.");
```
Ocramius: Are the identifiers set? Is the issue only around the state of `$name`?

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
BitOne: Yes, the identifier is set (which is pretty strange as we go through this if: https://github.com/doctrine/doctrine2/blob/2.5/lib/Doctrine/ORM/UnitOfWork.php#L2531 ), but all others properties are at null. Here only the name, because it's the only property apart from the id.
But I can add other properties and test them if that makes sense. That would maybe make the code a little bit less straight to the point.

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
BitOne: Done!

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
BitOne: Done: new variable name and assertion to check that they are indeed different objects.
By the way, on our project where we have this issue, the creation and data persistence is done in a different request than the one reading the data. So totally different memory for the PHP process.

Ocramius:
@BitOne `master` has fixes that we obviously cannot backport (mostly architectural and BC breaks), so this may indeed not be fixable without huge workarounds. Still, test looks much simpler and to the point now, thanks!

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Tests\Models\CMS\CmsEmployee;
+
+/**
+ * @group issue-7052
```
lcobucci: Please just use `7052` as group

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
```diff
@@ -0,0 +1,87 @@
...
+{
+    /**
+     * @Id
+     * @GeneratedValue
+     * @Column(type="integer")
+    */
+    public $id;
+
+    /**
+     * @ManyToOne(targetEntity="Issue7052Parent", fetch="LAZY")
```
lcobucci: Use `Issue7052Parent::class` (unquoted)

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
```diff
@@ -0,0 +1,87 @@
...
+
+        $childAFromDb = $this->_em->find('Doctrine\Tests\ORM\Functional\Ticket\Issue7052Child', $childA->id);
+
+        $this->_em->clear();
+
+        $childBFromDb = $this->_em->find('Doctrine\Tests\ORM\Functional\Ticket\Issue7052Child', $childB->id);
+
+        $parentFromChildB = $childBFromDb->parent;
+
+        $this->assertNotSame($childA, $childAFromDb);
```
lcobucci: Please use `self::assert*()` since the method is static

lcobucci (changes requested):
We don't support `2.5` anymore, so please use `2.6` as base and fix coding standard issues using `phpcbf`.

BitOne:
Hey @lcobucci, thanks for the remarks!

I applied the beautifier and rebase the test on 2.6.

Thank you!

lcobucci:
@BitOne awesome, I'd take a look ASAP. Thanks for your help!

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue7052Test.php
```diff
@@ -0,0 +1,87 @@
...
+        $childAFromDb = $this->_em->find('Doctrine\Tests\ORM\Functional\Ticket\Issue7052Child', $childA->id);
+
+        $this->_em->clear();
+
+        $childBFromDb = $this->_em->find('Doctrine\Tests\ORM\Functional\Ticket\Issue7052Child', $childB->id);
+
+        $parentFromChildB = $childBFromDb->parent;
+
+        self::assertNotSame($childA, $childAFromDb);
+        self::assertNotNull($childAFromDb->parent->name, "Unable to get parent name on second loaded child after EM cleared.");
```
Majkl578: I don't think this assertion is valid.
1. You loaded `$childAFromDb` from DB.
2. You called `EntityManager::clear()`, making the obtained entity unmanaged.
3. You assert that lazy association works on an unmanaged entity.

This test is also failing in ORM 2.5.x.

Majkl578: Also, if this works in master, it's rather a bug there, since once unmanaged, entities/proxies should be ignored by ORM.

Majkl578: @Ocramius @lcobucci Please verify the assumptions above. Removing from 2.6 milestone for now.

BitOne:
Hey guys,

Closing this PR as it's a on pretty old version of Doctrine anyway. Let's do some cleanup ;)

Thanks!

PR #7053, opened 2018-02-09
Fixed parser rejecting valid DQL

carnage:
Doctrine 2.5 allows the following as valid DQL:

"DELETE FROM Namespace\Entity"

Doctrine 2.6 rejects this as a syntax error.

The bug appears to be on the changed line, as the code didn't match the logic specified in the comment. I've not got a massive amount of knowledge on how the parser/lexer is built but changing this to a >= means that the above DQL is accepted as valid and appears to fix the problem.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -315,7 +315,7 @@ class Parser
             }
 
             // If parameter is T_IDENTIFIER, then matches T_IDENTIFIER (100) and keywords (200+)
-            if ($token === Lexer::T_IDENTIFIER && $lookaheadType < Lexer::T_IDENTIFIER) {
+            if ($token === Lexer::T_IDENTIFIER && $lookaheadType >= Lexer::T_IDENTIFIER) {
                 $this->syntaxError($this->lexer->getLiteral($token));
             }
         }
--- a/tests/Doctrine/Tests/ORM/Query/DeleteSqlGenerationTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/DeleteSqlGenerationTest.php
@@ -36,6 +36,14 @@ class DeleteSqlGenerationTest extends OrmTestCase
         }
     }
 
+    public function testSupportsDeleteWithoutWhereAndAlias()
+    {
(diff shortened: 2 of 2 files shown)


Ocramius:
@carnage can you still add a DQL SQL generation test? There's an already existing test class for that.

Ocramius:
Seems to fail tho: https://travis-ci.org/doctrine/doctrine2/jobs/339380152#L652

carnage:
new test passes, fix breaks other tests. Will have to look into it further.

Is there a BNF grammar defined for DQL?

Ocramius:
@carnage yeah, it is at http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html#ebnf

carnage:
Dug into this some more.

Seems the syntax "DELETE FROM \namespace\Entity" isn't legal according to the grammar. However, it was accepted as a valid statement in >2.6 and performed as expected. 

There are two possible options.
a) Close this as will not fix as it's against the defined grammar
b) Update the grammar and code to accept this statement and preserve BC

lcobucci:
@carnage thanks for your help on digging up things 👍 

@Ocramius @guilhermeblanco it looks like [this commit](https://github.com/doctrine/doctrine2/commit/b7bd42638dcf8aa049698f01466688ff7762953b) is the cause of this issue, which was actually a fix to ensure that the grammar is properly followed. That seems to be the most relevant change from `v2.5.x` to `v2.6.x`, since it affects how we parse the abstract schema name...

I've managed to make the tests pass with a nasty _gambiarra™_ - which essentially consist in creating an alias when it doesn't exist.

Although it's quite dirty, I'd suggest that we should remove the BC-break in `v2.6.x`, add a deprecation notice there on `v2.7.x`, and don't support this anymore on `v3.0`. I'll open a PR for that.

---

For those wondering what's a _gambiarra_: it's a Brazilian word for a very creative and terrible solution.

If that description is not enough, here's an [image](http://3.bp.blogspot.com/_KAhZfQh3zUE/TLNgb7plNBI/AAAAAAAAAOM/iq0yCBgNzoE/s640/gambiarra-lifestyle-5.jpg).

PR #7054, opened 2018-02-09
Fix ID generation of foreign keys

### LINKED ISSUE #7002: Flush new entity with relation when use relation in composite key throws ORMInvalidArgumentException
```php
/**
 * Class Entity
 *
 * @Entity(repositoryClass="TestRepo")
 * @Table(name="test_entity")
 */
class Entity
{
    /**
     * @var int
     * @Id
     * @Column(type="integer")
     * @GeneratedValue
     */
    protected $id;

    /**
     * @var string
     * @Column(type="string")
     */
    protected $label = '';

    /**
     * @var Collection|Friend[]
     * @OneToMany(targetEntity="Friend", mappedBy="entity", cascade={"all"}, orphanRemoval=true)
     */
    protected $friends;

    public function __construct()
    {
        $this->friends = new ArrayCollection();
    }

    public function addFriend(Friend $friend)
    {
        $friend->setEntity($this);
        $this->friends->add($friend);

        return $this;
    }

    public function setLabel(string $label)
    {
        $this->label = $label;

        return $this;
    }
}

/**
 * Class Entity
 * @Entity
 * @Table(name="test_entity_friend")
 */
class Friend
{
    /**
     * @var Entity
     * @id
     * @ManyToOne(targetEntity="Entity", inversedBy="friends")
     */
    protected $entity;

    /**
     * @var string
     * @id
     * @Column(type="string", length=400)
     */
    protected $name = '';

    /**
     * @return string
     */
    public function getName(): string
    {
        return $this->name;
    }

    /**
     * @param string $name
     * @return $this
     */
    public function setName(string $name)
    {
        $this->name = $name;

        return $this;
    }

    public function setEntity(Entity $entity)
    {
        $this->entity = $entity;

        return $this;
    }
}
```

Usage:

```php
$entity = new Entity();
$entity->setLabel('labelll');

$friend = new Friend();
$friend->setName('name');
$entity->addFriend($friend);

$entityManager->persist($entity);
$entityManager->flush();
```

Throws ORM/ORMInvalidArgumentException.php:68 "The given entity of type '\TestDoctrine\Entity\Friend' (\TestDoctrine\Entity\Friend@00000000309854b5000000005dfe4c27) has no identity/no id values set. It cannot be added to the identity map".

Because $entity has null id.

Flush after creating new Entity() and before adding $friend to $entity resolve this problem, but it makes many duct tapes.

Doctrine 2.6. Prev version works fine.

cb8:
this patch is an alternate approach to #7003 which tries to fix #7002. however, it applies to the `master` branch. so far, i didn't clean up my test cases. in case this patch will get merged despite/instead of #7003, i will push them for review.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -962,9 +962,9 @@ class UnitOfWork implements PropertyChangedListener
             if ($generationPlan->containsDeferred()) {
                 // Entity has post-insert IDs
                 $oid = spl_object_id($entity);
-                $id  = $this->em->getIdentifierFlattener()->flattenIdentifier($class, $persister->getIdentifier($entity));
+                $id  = $persister->getIdentifier($entity);
 
-                $this->entityIdentifiers[$oid]  = $id;
+                $this->entityIdentifiers[$oid]  = $this->em->getIdentifierFlattener()->flattenIdentifier($class, $id);
                 $this->entityStates[$oid]       = self::STATE_MANAGED;
                 $this->originalEntityData[$oid] = $id + $this->originalEntityData[$oid];
 
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -13,10 +13,12 @@ use Doctrine\ORM\Event\OnClassMetadataNotFoundEventArgs;
 use Doctrine\ORM\Events;
 use Doctrine\ORM\ORMException;
 use Doctrine\ORM\Sequencing;
+use Doctrine\ORM\Sequencing\Planning\AssociationValueGeneratorExecutor;
 use Doctrine\ORM\Sequencing\Planning\ColumnValueGeneratorExecutor;
 use Doctrine\ORM\Sequencing\Planning\CompositeValueGenerationPlan;
 use Doctrine\ORM\Sequencing\Planning\NoopValueGenerationPlan;
 use Doctrine\ORM\Sequencing\Planning\SingleValueGenerationPlan;
+use Doctrine\ORM\Sequencing\Planning\ValueGenerationExecutor;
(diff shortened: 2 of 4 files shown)


lcobucci:
@cb8 thanks for your patch, we just merged in the fix for 2.6 and were porting it to master. I'll check your changes to see if it works better for the new version of the code.

lcobucci:
@cb8 your changeset makes the tests of #6701 pass but breaks `SecondLevelCacheCompositePrimaryKeyWithAssociationsTest` because it tries to update a read only entity, I'll investigate further this week but it would be nice if you could check it as well.

cb8:
thank you for the review. the deferred IDs were inadvertently flattened when they were added to the original data of the entity. as far as i could see, the failing tests on Travis are not caused by the patch (but don't hold me to that ;)).

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -962,9 +962,9 @@ private function executeInserts(ClassMetadata $class) : void
             if ($generationPlan->containsDeferred()) {
                 // Entity has post-insert IDs
                 $oid = spl_object_id($entity);
-                $id  = $this->em->getIdentifierFlattener()->flattenIdentifier($class, $persister->getIdentifier($entity));
+                $id  = $persister->getIdentifier($entity);
 
-                $this->entityIdentifiers[$oid]  = $id;
+                $this->entityIdentifiers[$oid]  = $this->em->getIdentifierFlattener()->flattenIdentifier($class, $id);
```
lcobucci: @guilhermeblanco this looks ok, but I'd like your eyes here

lcobucci (changes requested):
Some small changes, I'll take care of these in a bit. @cb8 thanks a lot for your help!

lcobucci (approved):

guilhermeblanco (approved):

lcobucci:
@cb8 🚢 thanks for your contribution!

PR #7056, opened 2018-02-09
[CS] Apply Doctrine CS 3.0

Majkl578:
Changes from 2.0 -> 3.0:
* Import all functions and constants in `use`
* Remove useless comments
* Use new with parenthesis
* Unqualified class references in annotations

Excluded:
* EarlyExit sniff: To be handled separately as is not just cosmetics.
* ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName: Excluded for Annotation namespace - annotations do not work with `use`.

~I'll drop WIP label and dev-master dependency once Slevomat CS 4.4.5 with bugfixes is out, but feel free to review already.~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "symfony/console": "~3.2|~4.0"
     },
     "require-dev": {
-        "doctrine/coding-standard": "~2.0.0",
+        "doctrine/coding-standard": "~3.0.0",
         "phpunit/phpunit": "^7.0"
     },
     "autoload": {
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -4,10 +4,10 @@
     <arg name="extensions" value="php"/>
     <arg name="parallel" value="80"/>
     <arg name="cache" value=".phpcs-cache"/>
-    <arg name="colors" />
+    <arg name="colors"/>
 
-    <!-- Ignore warnings and show progress of the run -->
-    <arg value="np"/>
+    <!-- Ignore warnings, show progress of the run and show sniff names -->
+    <arg value="nps"/>
 
     <file>lib</file>
     <file>tests</file>
@@ -19,6 +19,7 @@
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint"/>
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingReturnTypeHint"/>
         <exclude name="SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly.ReferencedGeneralException"/>
+        <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
     </rule>
 
     <rule ref="PSR1.Classes.ClassDeclaration.MultipleClasses">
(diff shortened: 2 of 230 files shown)


carusogabriel (approved):
Couldn't find anything wrong :clap:

phpcs.xml.dist
```diff
@@ -19,6 +19,7 @@
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint"/>
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingReturnTypeHint"/>
         <exclude name="SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly.ReferencedGeneralException"/>
+        <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
```
Ocramius: I guess this still breaks too much stuff

phpcs.xml.dist
```diff
@@ -57,4 +58,8 @@
     <rule ref="SlevomatCodingStandard.Classes.UnusedPrivateElements.UnusedMethod">
         <exclude-pattern>lib/Doctrine/ORM/Mapping/*</exclude-pattern>
     </rule>
+
+    <rule ref="SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName">
+        <exclude-pattern>lib/Doctrine/ORM/Annotation/*</exclude-pattern>
```
Ocramius: This looks fishy. Worth inspecting in a second PR just doing this one /cc @carusogabriel

composer.json
```diff
@@ -29,7 +29,7 @@
         "symfony/console": "~3.2|~4.0"
     },
     "require-dev": {
-        "doctrine/coding-standard": "~2.0.0",
+        "doctrine/coding-standard": "~3.0.0",
```
Ocramius: This should be `^3.0.0` as per https://github.com/doctrine/coding-standard/pull/18 /cc @carusogabriel

Ocramius (approved):
🚢 

The comments need to be fixed, but due to the huge diff, tracking these changes would be a pain, so this is going in as-is for now. @carusogabriel can you maybe add these changes to #7046 if we figure out what failed in the test suite there?

phpcs.xml.dist
Majkl578: This is because imho doctrine/annotations doesn't properly support unqualified names in annotations, at least I repeatedly ran into test failures so I explicitly disabled it until further investigation.

phpcs.xml.dist
Majkl578: Most of the issues were fixed, but I didn't want to do this here as it would make the diff a lot harder to review, rather separately, some code could even be refactored using private methods.

composer.json
Majkl578: Damn, I think I changed it but maybe forgot to amend. 🤔 Will fix.

carusogabriel:
@Ocramius @Majkl578 Thanks for ping me. Gonna rebase and reconstruct #7046 with new CS version and excludes Sniffs.

PR #7065, opened 2018-02-14
Fix compatibility with DBAL develop

Majkl578:
* ResultStatement signature BC break
* PDO::FETCH_* -> FetchMode::*
* PDO::PARAM_* -> ParameterType::*
* AbstractPlatform::DATE_INTERVAL_UNIT_* -> DateIntervalUnit::*
* AbstractPlatform::TRIM_* -> TrimMode::*

Changes from doctrine/dbal#2958 & doctrine/dbal#2998.

/cc @morozov

PR #7076, opened 2018-02-17
Make mocks compatible with latest version of interface

lcobucci:
Since we changed the signature in DBAL v3.x-dev we must apply the changes here too.


lcobucci:
I'll just go YOLO here, things are green so it's all good 😂

Majkl578:
Can you maybe check existing PRs first? #7065

PR #7077, opened 2018-02-18
Fix BC-break on delete without alias DQL

lcobucci:
As mentioned in #7053, this fixes the issue without breaking other tests.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -304,21 +304,24 @@ class Parser
         $lookaheadType = $this->lexer->lookahead['type'];
 
         // Short-circuit on first condition, usually types match
-        if ($lookaheadType !== $token) {
-            // If parameter is not identifier (1-99) must be exact match
-            if ($token < Lexer::T_IDENTIFIER) {
-                $this->syntaxError($this->lexer->getLiteral($token));
-            }
+        if ($lookaheadType === $token) {
+            $this->lexer->moveNext();
+            return;
+        }
 
-            // If parameter is keyword (200+) must be exact match
-            if ($token > Lexer::T_IDENTIFIER) {
-                $this->syntaxError($this->lexer->getLiteral($token));
-            }
+        // If parameter is not identifier (1-99) must be exact match
+        if ($token < Lexer::T_IDENTIFIER) {
+            $this->syntaxError($this->lexer->getLiteral($token));
+        }
 
-            // If parameter is T_IDENTIFIER, then matches T_IDENTIFIER (100) and keywords (200+)
-            if ($token === Lexer::T_IDENTIFIER && $lookaheadType < Lexer::T_IDENTIFIER) {
-                $this->syntaxError($this->lexer->getLiteral($token));
-            }
+        // If parameter is keyword (200+) must be exact match
+        if ($token > Lexer::T_IDENTIFIER) {
+            $this->syntaxError($this->lexer->getLiteral($token));
+        }
+
+        // If parameter is T_IDENTIFIER, then matches T_IDENTIFIER (100) and keywords (200+)
+        if ($token === Lexer::T_IDENTIFIER && $lookaheadType < Lexer::T_IDENTIFIER) {
+            $this->syntaxError($this->lexer->getLiteral($token));
         }
 
         $this->lexer->moveNext();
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -1280,7 +1281,15 @@ public function DeleteClause()
             $this->match(Lexer::T_AS);
         }
 
-        $aliasIdentificationVariable = $this->AliasIdentificationVariable();
+        try {
```
lcobucci: Maybe we should a have a comment here to explain the 💩 and specially mention that we should raise a deprecation notice in `v2.7.0`

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -304,21 +304,24 @@ public function match($token)
         $lookaheadType = $this->lexer->lookahead['type'];
 
         // Short-circuit on first condition, usually types match
-        if ($lookaheadType !== $token) {
-            // If parameter is not identifier (1-99) must be exact match
-            if ($token < Lexer::T_IDENTIFIER) {
-                $this->syntaxError($this->lexer->getLiteral($token));
-            }
+        if ($lookaheadType === $token) {
```
lcobucci: I can drop this commit if you prefer, it was just to make things a bit more sane.

lcobucci:
@Ocramius @guilhermeblanco I was checking JPA 2.2 specs and maybe we should rather update the grammar for bulk delete:

> delete_statement ::= delete_clause [where_clause]
delete_clause ::= DELETE FROM entity_name [[AS] identification_variable]

**Source:** http://download.oracle.com/otndocs/jcp/persistence-2_2-mrel-spec/index.html (4.10 Bulk Update and Delete Operations)

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -1280,7 +1284,15 @@ public function DeleteClause()
             $this->match(Lexer::T_AS);
         }
 
-        $aliasIdentificationVariable = $this->AliasIdentificationVariable();
+        try {
```
carnage: Could you use $this->lexer->isNextToken for this instead of trying to handle the exception?

lcobucci: Yes indeed, updating

lcobucci: @carnage thanks... it was a way too gambiarra before HAHAHA

lib/Doctrine/ORM/Query/Parser.php
Ocramius: Works as-is

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -1280,7 +1284,9 @@ public function DeleteClause()
             $this->match(Lexer::T_AS);
         }
 
-        $aliasIdentificationVariable = $this->AliasIdentificationVariable();
+        $aliasIdentificationVariable = $this->lexer->isNextToken(Lexer::T_IDENTIFIER)
+            ? $this->AliasIdentificationVariable()
+            : 'alias_should_have_been_set';
```
Ocramius: We should likely raise a deprecation notice here, to clarify that this isn't supported DQL (according to BNF)

tests/Doctrine/Tests/ORM/Query/DeleteSqlGenerationTest.php
```diff
@@ -36,6 +36,14 @@ public function assertSqlGeneration($dqlToBeTested, $sqlToBeConfirmed)
         }
     }
 
+    public function testSupportsDeleteWithoutWhereAndAlias() : void
```
Ocramius: Please add an `@group` to this

Ocramius (changes requested):
Another location to check is `Doctrine\ORM\QueryBuilder#delete()`, which allows for an optional alias. We should probably raise a deprecation notice there too...

lib/Doctrine/ORM/Query/Parser.php
lcobucci: It works indeed, it's just not so good as it could be... I can send another PR to make you happier 😂

lcobucci:
@Ocramius and what's your opinion regarding changing the grammar?

Ocramius:
> what's your opinion regarding changing the grammar?

Only after we move to https://github.com/hoaproject/Compiler

Documenting grammar BC breaks becomes trivial after that. No changes before that, IMO.

lib/Doctrine/ORM/Query/Parser.php
lcobucci: Didn't we say that deprecation notices were going to be added in `v2.7` only?

Ocramius: Works for me as long as you please create an issue tracking this specific deprecation 👍

lcobucci: Sure, I was going to do that anyway 😄

Ocramius:
This has been forward-ported to `master` in doctrine/doctrine2@92445d095e8b11a98507438215b60cc6bab500a7

Issue #7062, opened 2018-02-13
BasicEntityPersister confuses association types with PDO datatypes


mariusklocke:
Hey folks,

i think i found a bug when combining one-to-one associations with association-key identifiers. My application deals with soccer leagues and the issue should become visible with the following 3 related entities:

```xml
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                    http://raw.github.com/doctrine/doctrine2/master/doctrine-mapping.xsd">

    <entity name="HexagonalPlayground\Domain\Ranking" table="rankings">
        <id name="season" association-key="true" />
        <field name="updatedAt" column="updated_at" type="datetime_immutable" nullable="true" />
        <one-to-one field="season" target-entity="HexagonalPlayground\Domain\Season" inversed-by="ranking">
            <join-column name="season_id" referenced-column-name="id" />
        </one-to-one>
        <one-to-many target-entity="HexagonalPlayground\Domain\RankingPosition" mapped-by="ranking" field="positions" index-by="team_id" orphan-removal="true">
            <cascade>
                <cascade-all/>
            </cascade>
        </one-to-many>
    </entity>
</doctrine-mapping>
```

```xml
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                    http://raw.github.com/doctrine/doctrine2/master/doctrine-mapping.xsd">

    <entity name="HexagonalPlayground\Domain\RankingPosition" table="ranking_positions">
        <indexes>
            <index columns="sort_index" />
        </indexes>
        <id name="ranking" association-key="true" />
        <id name="team" association-key="true" />
        <field name="sortIndex" type="integer" column="sort_index" />
        <field name="number" type="integer" />
        <field name="matches" type="integer" />
        <field name="wins" type="integer" />
        <field name="draws" type="integer" />
        <field name="losses" type="integer" />
        <field name="scoredGoals" type="integer" column="scored_goals" />
        <field name="concededGoals" type="integer" column="conceded_goals" />
        <field name="points" type="integer" />
[...]

PR #7082, opened 2018-02-20
Add failing test for issue #7062 - BasicEntityPersister confuses association types with PDO datatypes

mariusklocke:
Providing a failing testcase as requested in issue #7062

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7062Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7062Test.php
@@ -1,0 +1,212 @@
+<?php
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH7062Test extends OrmFunctionalTestCase
+{
+    private const SEASON_ID = 'season_18';
+    private const TEAM_ID   = 'team_A';
+
+    protected function setUp() : void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema(
+            [
+                GH7062Team::class,
+                GH7062Season::class,
+                GH7062Ranking::class,
+                GH7062RankingPosition::class
+            ]
+        );
+    }
+
+    /**
+     * @group 7062
+     */
+    public function testEntityWithAssociationKeyIdentityCanBeUpdated() : void
+    {
+        $this->createInitialRankingWithRelatedEntities();
+        $this->modifyRanking();
+        $this->verifyRanking();
+    }
+
+    private function createInitialRankingWithRelatedEntities() : void
+    {
+        $team    = new GH7062Team(self::TEAM_ID);
+        $season  = new GH7062Season(self::SEASON_ID);
+
+        $season->ranking = new GH7062Ranking($season, [$team]);
+
+        $this->_em->persist($team);
+        $this->_em->persist($season);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        foreach ($season->ranking->positions as $position) {
+            self::assertSame(0, $position->points);
+        }
+    }
+
+    private function modifyRanking() : void
+    {
+        /** @var GH7062Ranking $ranking */
+        $ranking = $this->_em->find(GH7062Ranking::class, self::SEASON_ID);
+
+        foreach ($ranking->positions as $position) {
+            $position->points += 3;
+        }
+
+        $this->_em->flush();
+        $this->_em->clear();
+    }
+
+    private function verifyRanking() : void
+    {
+        /** @var GH7062Season $season */
+        $season = $this->_em->find(GH7062Season::class, self::SEASON_ID);
+        self::assertInstanceOf(GH7062Season::class, $season);
+
+        $ranking = $season->ranking;
+        self::assertInstanceOf(GH7062Ranking::class, $ranking);
+
+        foreach ($ranking->positions as $position) {
+            self::assertSame(3, $position->points);
+        }
+    }
+}
+
+/**
+ * Simple Entity whose identity is defined through another Entity (Season)
+ *
+ * @Entity
+ * @Table(name="soccer_rankings")
+ */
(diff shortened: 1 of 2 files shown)


Ocramius:
@mariusklocke maybe you intended to send this to `2.6`?

mariusklocke:
Changed to 2.6
Got a little confused when reading the contribution guidelines :)

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC7062Test.php
```diff
@@ -0,0 +1,75 @@
...
+        $ranking = $this->createInitialRankingWithRelatedEntities();
+        $this->modifyRanking($ranking);
+
+        // Clear the EntityManager to make sure we fetch entities from database
+        $this->_em->clear();
+
+        // Now try to read entities again and check if all changes have been made persistent correctly
+        /** @var Season $season */
+        $season = $this->_em->find(Season::class, self::SEASON_ID);
+        $this->assertInstanceOf(Season::class, $season);
```
lcobucci: Please use `self::assert*` instead since the method is static.

lcobucci (changes requested):
@mariusklocke can you please combine the models in the test file just like https://github.com/doctrine/doctrine2/blob/30a063ef9d7b14d0c74a192d87dc31c600503a4b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6531Test.php ? 

This would also make things easier for us to review, you to fix coding standard issues using `phpcbf`, and also removes all the changes you've made in `OrmFunctionalTestCase` (which are not really necessary).

mariusklocke:
Thanks for your quick feedback. Will do that in the next days.

lcobucci:
@mariusklocke @Ocramius wow, this is a tough one! The cause of the issue is that for some reason the ORM is not using the correct type for the `season_id` on the update query... I'll continue my investigation here

mariusklocke:
@lcobucci Did you read my analysis where i reported the [issue](https://github.com/doctrine/doctrine2/issues/7062)? Can you state if my assumptions there are correct?

lcobucci (approved):
@mariusklocke I've changed things a bit in the test to always fetch things from the database, just to ensure that things are isolated.

The fix itself was pretty simple (with a little help from @guilhermeblanco).

Thanks for your contribution!

lcobucci:
> @lcobucci Did you read my analysis where i reported the issue? Can you state if my assumptions there are correct?

@mariusklocke I read it, after I've posted my comment... you're assumptions are indeed correct

mariusklocke:
I'm glad you found a fix. Debugging that took me a while :)

lcobucci:
@mariusklocke the only thing is that your mapping quite complex and it uncovers some scenarios that we haven't thought about (the fact that `RankingPosition` has `Ranking` as id, but that id is actually `Season` is making the `IdentifierFlattener` a bit crazy - only when running with L2C enabled)...

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7062Test.php
```diff
@@ -0,0 +1,255 @@
...
+ *
+ * @Entity
+ * @Table(name="soccer_rankings")
+ */
+class GH7062Ranking
+{
+    /**
+     * @Id
+     * @OneToOne(targetEntity=GH7062Season::class, inversedBy="ranking")
+     * @JoinColumn(name="season", referencedColumnName="id")
```
lcobucci: @Ocramius @guilhermeblanco this test revealed a nice unrelated issue (when loading things with L2C the identifier flattener tries to use the [column name in an array that doesn't have that index](https://travis-ci.org/doctrine/doctrine2/jobs/346051424#L959)).
Forcing the entity to use the name of the field as the name of the column was the only way to isolate the issue that @mariusklocke described... what are your thoughts about this workaround?

guilhermeblanco: @lcobucci as long as we can open a new issue with the new problem, I'm good to isolate issues in 2 separate PRs. =)

lcobucci: @guilhermeblanco that's the idea

Ocramius: Yeah, that's a separate issue - let's not creep it into here.

Ocramius (changes requested):
I'm fine with the patch overall, but the test uses a DSL from a different domain, and that makes things quite complex to follow when reasoning about ORM details.

I'd rather get rid of all the public methods in these entities, and modify all states via `public` properties instead, making the classes ORM-naming-specific instead of flavored with `Season` and `Team`

lcobucci:
@Ocramius how about now?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7062Test.php
lcobucci: https://github.com/doctrine/doctrine2/issues/7096 created to handle that

Ocramius:
@lcobucci if the test still fails with the patch not applied, then :+1:

I'd rather completely get rid of the team/season/etc thing...

lcobucci:
> @lcobucci if the test still fails with the patch not applied, then 👍

@Ocramius it does fail.

> I'd rather completely get rid of the team/season/etc thing...

I removed all the methods but decided to keep the whole thing due to the complicated setup of FKs (which is exactly what affects the bug).

Ocramius (approved):

Ocramius:
:ship:

Ocramius:
Btw, just noticed that recent patches aren't in `2.7`. Will need to merge `2.6` up (not nice, but not as important as `master` right now)

Ocramius:
Also cherry-picking this into `master`

Ocramius:
Interestingly, the test passes on `master` without actually needing to apply the patch - accidentally fixed by fixing our mapping system :-)

Ocramius:
In `master` as per doctrine/doctrine2@bbc1ed93b14e079cd403cb5b2b1134da35af431e
In `2.7` as per doctrine/doctrine2@4f6d47b88a84896dadaadd18212df6bb504f2e97

PR #7085, opened 2018-02-22
[RFC] Tighten owning side resolution

guilhermeblanco:
Fixed bug with potential double owning side in OneToOne when JoinColumns were mapped on both association sides. In that case, the ownership was determined based on association flow (in A -> B, A:b was owning, while in B -> A, B:a was owning).

### DIFF
--- a/doctrine-mapping.xsd
+++ b/doctrine-mapping.xsd
@@ -169,7 +169,7 @@
 
   <xs:simpleType name="tablename" id="tablename">
     <xs:restriction base="xs:token">
-      <xs:pattern value="[a-zA-Z_u01-uff.]+" id="tablename.pattern">
+      <xs:pattern value="[a-zA-Z\-_u01-uff.]+" id="tablename.pattern">
       </xs:pattern>
     </xs:restriction>
   </xs:simpleType>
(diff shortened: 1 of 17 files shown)


Ocramius (changes requested):
@guilhermeblanco I only see changes in existing tests, yet no new test showing the actual issue (with a single assertion) here

tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
```diff
@@ -359,6 +360,62 @@ public function testBooleanValuesForOptionIsSetCorrectly()
         return $class;
     }
 
+    public function testOwningSideResolution() : void
```
lcobucci: Please split this into multiple tests, each `//` can be a separate method. Things would be much simpler to understand.

doctrine-mapping.xsd
```diff
@@ -169,7 +169,7 @@
 
   <xs:simpleType name="tablename" id="tablename">
     <xs:restriction base="xs:token">
-      <xs:pattern value="[a-zA-Z_u01-uff.]+" id="tablename.pattern">
+      <xs:pattern value="[a-zA-Z\-_u01-uff.]+" id="tablename.pattern">
```
lcobucci: Do we have more chars that should be allowed here?

lcobucci (changes requested):

lcobucci (approved):

Ocramius (approved):

Ocramius:
:ship:

Majkl578:
Shouldn't this go to UPGRADE as well?

PR #7088, opened 2018-02-22
Drop explicit PDO dependency

Majkl578:
After doctrine/dbal#2958, #7076 and #7065, there is no longer reason for an explicit dependency on PDO in ORM.
DBAL dropped it already, ORM should too.


Ocramius (approved):
LGTM :ship:

PR #7090, opened 2018-02-24
Use proper grammar

greg0ire:
Not sure what idiot wrote it that way.


Ocramius:
> Not sure what idiot wrote it that way.

I think I know :D

PR #7091, opened 2018-02-25
Fix from CS for lib and tools

carusogabriel:
There are only `Empty CATCH statement detected` warnings now, but there aren't fixable ATM.

@Ocramius @Majkl578 I'm up to exclude `tools/*` from the `Class X should not be referenced via a fully qualified name but via a use statement`. What do you think?


Ocramius (approved):
:ship:

PR #7092, opened 2018-02-25
Declare some types

lcobucci:
Since now we're require PHP 7.2 we can start to add `object`

### DIFF
--- a/lib/Doctrine/ORM/Sequencing/Generator.php
+++ b/lib/Doctrine/ORM/Sequencing/Generator.php
@@ -11,11 +11,9 @@ interface Generator
     /**
      * Generates an identifier for an entity.
      *
-     * @param object $entity
-     *
-     * @return \Generator
+     * @return string|int
      */
-    public function generate(EntityManagerInterface $em, $entity);
+    public function generate(EntityManagerInterface $em, ?object $entity);
 
     /**
      * Gets whether this generator is a post-insert generator which means that
@@ -24,8 +22,6 @@ interface Generator
      *
      * By default, this method returns FALSE. Generators that have this requirement
      * must override this method and return TRUE.
-     *
-     * @return bool
      */
-    public function isPostInsertGenerator();
+    public function isPostInsertGenerator() : bool;
 }
--- a/lib/Doctrine/ORM/Sequencing/SequenceGenerator.php
+++ b/lib/Doctrine/ORM/Sequencing/SequenceGenerator.php
@@ -38,13 +38,7 @@ class SequenceGenerator implements Generator, Serializable
      */
     private $maxValue;
 
-    /**
-     * Initializes a new sequence generator.
-     *
-     * @param string $sequenceName   The name of the sequence.
-     * @param int    $allocationSize The allocation size of the sequence.
-     */
-    public function __construct($sequenceName, $allocationSize)
+    public function __construct(string $sequenceName, int $allocationSize)
     {
         $this->sequenceName   = $sequenceName;
         $this->allocationSize = $allocationSize;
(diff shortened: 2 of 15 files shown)


lcobucci:
I've tried to minimise the impact so I could already use this in https://github.com/doctrine/doctrine2/pull/7054

lib/Doctrine/ORM/Sequencing/Generator.php
```diff
@@ -11,11 +11,9 @@
     /**
      * Generates an identifier for an entity.
      *
-     * @param object $entity
-     *
-     * @return \Generator
+     * @return string|int
```
Ocramius: What happened here? :|

lib/Doctrine/ORM/Sequencing/SequenceGenerator.php
```diff
@@ -70,41 +64,31 @@ public function generate(EntityManagerInterface $em, $entity)
...
-    public function getCurrentMaxValue()
+    public function getCurrentMaxValue() : ?int
     {
         return $this->maxValue;
     }
 
     /**
      * Gets the next value that will be returned by generate().
-     *
-     * @return int
```
Ocramius: Was this previously wrong?

lib/Doctrine/ORM/Sequencing/SequenceGenerator.php
```diff
@@ -70,41 +64,31 @@ public function generate(EntityManagerInterface $em, $entity)
...
+                'sequenceName'   => $this->sequenceName,
             ]
         );
     }
 
-    /**
-     * @param string $serialized
-     */
-    public function unserialize($serialized)
+    public function unserialize($serialized) : void
```
Ocramius: Can a `string` type be added here? Or does it collide with `Serializable`?

Ocramius (changes requested):

lib/Doctrine/ORM/Sequencing/SequenceGenerator.php
lcobucci: It collides with `Serializable`

lib/Doctrine/ORM/Sequencing/SequenceGenerator.php
lcobucci: yeap...

lcobucci: I mean, it's is not needed with the type hint =)

lib/Doctrine/ORM/Sequencing/Generator.php
lcobucci: It was just wrong before, you can see based on the implementation...

lib/Doctrine/ORM/Sequencing/SequenceGenerator.php
Ocramius: The hint added nullability

Ocramius: Hmm, no, I think my comment was misplaced

lib/Doctrine/ORM/Sequencing/SequenceGenerator.php
lcobucci: I'll put the docblock back, my intention was to add the type declaration

lcobucci: done

Ocramius (approved):

Ocramius:
:ship:

Majkl578:
@Ocramius Hmm so now documentation of type changes is not required anymore? I am confused.

Ocramius:
@Majkl578 these types are compatible with existing ones, and as we now support PHP 7.2, then subclasses and implementors do not break if we add a type in a superclass or abstract type.

Majkl578:
> subclasses and implementors do not break if we add a type in a superclass or abstract type

That is only true for parameter types, but return types were also added here.

Ocramius:
@lcobucci worth addid `UPGRADE.md` docs then

lcobucci:
@Ocramius @Majkl578 yeap, I completely forgot to do it... sending a PR in a bit

PR #7093, opened 2018-02-25
Fix updating entities with quoted identifier association

lcobucci:
Replaces #7012 (since author didn't allow changes from maintainers).


lcobucci:
@janlanger 🚢 thanks again!

PR #7095, opened 2018-02-25
[EntityRepository] Cleanup named native features

beberlei:
Named DQL queries were already removed before, this PR removes named native queries as well  before doing further changes to ER interface.


Ocramius (approved):
:+1:

PR #7104, opened 2018-03-01
Adds possibility to overwrite EntityManager

CloudrexxTeam:
In order to customize the EntityManager class it is necessary to use "static" in its create() method.

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -831,7 +831,7 @@ final class EntityManager implements EntityManagerInterface
 
         $connection = static::createConnection($connection, $config, $eventManager);
 
-        return new EntityManager($connection, $config, $connection->getEventManager());
+        return new static($connection, $config, $connection->getEventManager());
     }
 
     /**


Ocramius:
Whilst the `EntityManager` is not marked as `final` yet, it is our intention to do that in the near future (before 3.x).

Opened https://github.com/doctrine/doctrine2/issues/7105 to make this explicit.

CloudrexxTeam:
Thanks for your quick reply. Could you elaborate on the reasons for this?

Ocramius:
@CloudrexxTeam the `EntityManager` has always been "soft-final":

https://github.com/doctrine/doctrine2/blob/v2.6.1/lib/Doctrine/ORM/EntityManager.php#L63

Please consider reading https://ocramius.github.io/blog/when-to-declare-classes-final/

CloudrexxTeam:
Ok. So if I understood correctly we are left with the following options:

1. Create a "new" EntityManager, using the Interface.
2. Propose the change in a generalized manner as a pull-request.

Assuming that the above is correct and concluding (otherwise any input is welcome):
We want to enable result cache for queries for all SELECT queries by default. Currently we have overwritten the createQuery() method to achieve this.
For such a change, should we preferably create a pull-request which solves this in a generic way (probably by introducing some kind of option whether to use result cache for all (select) queries by default) or are we to create a complete copy of EntityManager with just the change in createQuery()?

Thanks for your patience with my questions!

Ocramius:
> We want to enable result cache for queries for all SELECT queries by default.

This is already done by the L2 cache, but enabling by default for all queries is NOT endorseable.

What you can do is writing an `EntityManagerInterface` implementation that overrides `createQuery()` and `createQueryBuilder()` by adding a call to `AbstractQuery#setCacheable()`: https://github.com/doctrine/doctrine2/blob/bbc1ed93b14e079cd403cb5b2b1134da35af431e/lib/Doctrine/ORM/AbstractQuery.php#L174

This wouldn't be something to be merged in core though, as "cache-everything-by-default" is dangerous and domain-specific.

CloudrexxTeam:
Ok, just to make sure: This equals the first option I listed (create an almost exact copy of EntityManager)?

Ocramius:
Yes, you'd compose/decorate an existing `EntityManager` instance within your own implementation.

PR #7112, opened 2018-03-04
Apply Doctrine CS 4.0, version composer.lock

Majkl578:
* Exclude abstract/exception naming sniff
  * Abstracts either need to be reworked separately or ignored.
  * Exceptions are being handled in #6743.
* Exclude EntityManagerInterface from interface naming sniff
  * Do we want to rename this in 3.0?
* Use inline comments for single `@var` annotations
* Use short type casts
* Remove empty comments
* Apply other fixes (and fix previous breakages in master)

-----

Controversially, after the discussions in/around #30, we should start versioning the composer.lock file and use it to run CS/SA checks on CI. I am still not convinced it's a good idea to have it in the root though, but Composer doesn't have a switch to specify alternative lock file path...

-----

CS stage green, PHPStan stage fixed in #7113.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/FieldMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/FieldMetadata.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Mapping;
 
-use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Reflection\ReflectionService;
 
 class FieldMetadata extends LocalColumnMetadata implements Property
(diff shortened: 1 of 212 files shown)


carusogabriel (approved):

lib/Doctrine/ORM/Mapping/FieldMetadata.php
```diff
@@ -18,21 +17,10 @@ class FieldMetadata extends LocalColumnMetadata implements Property
...
-     */
-    /*public function __construct(string $name, string $columnName, Type $type)
-    {
-        parent::__construct($columnName, $type);
-
-        $this->name = $name;
-    }*/
-
-    public function __construct(string $name)
+    public function __construct(string $name/*, string $columnName, Type $type*/)
```
Ocramius: Weird one - is there an issue or WIP for this, @guilhermeblanco?

phpcs.xml.dist
```diff
@@ -20,6 +20,8 @@
         <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingReturnTypeHint"/>
         <exclude name="SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly.ReferencedGeneralException"/>
         <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming"/>
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming"/>
```
Ocramius: @greg0ire relevant for you

Ocramius (approved):
LGTM :+1:

phpcs.xml.dist
greg0ire: How will that work :thinking: ? Will that warn on new classes only?

Ocramius: Will run on everything

lib/Doctrine/ORM/Mapping/FieldMetadata.php
guilhermeblanco: WIP

PR #7119, opened 2018-03-07
DDC-2842: Leave out discriminator part of Doctrine' generated SQL on root entity

11ya:
Fix for [#2842](https://github.com/doctrine/doctrine2/issues/3596)

### DIFF
--- a/lib/Doctrine/ORM/Mapping/DiscriminatorColumn.php
+++ b/lib/Doctrine/ORM/Mapping/DiscriminatorColumn.php
@@ -51,4 +51,9 @@ final class DiscriminatorColumn implements Annotation
      * @var string
      */
     public $columnDefinition;
+
+    /**
+     * @var bool
+     */
+    public $strict;
 }
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -394,7 +394,15 @@ class SchemaTool
             $options['columnDefinition'] = $discrColumn['columnDefinition'];
         }
 
-        $table->addColumn($discrColumn['name'], $discrColumn['type'], $options);
+        $column = $table->addColumn($discrColumn['name'], $discrColumn['type'], $options);
+
+        if (isset($discrColumn['strict']) && $discrColumn['strict']) {
+            $values = [];
+            foreach ($class->discriminatorMap as $value => $className) {
+                $values[] = $this->platform->quoteStringLiteral($value);
+            }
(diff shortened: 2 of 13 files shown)


Ocramius (changes requested):
The reason why the `t.discriminator_column IN()` condition is added is that we do not assume totality of record ownership on that table.

Records may have an unknown discriminator or belong to other inheritance trees.

While this change is possibly an improvement, it also represents a BC break, as hydration would just choke when encountering an unknown discriminator value.

11ya:
@Ocramius Ok, if I understand you correctly, check on root entity class name should solve the problem of "belong to other inheritance trees". As for "unknown discriminator", I see only one decision - add checking discriminator value constraint to the table definition and separate option in discriminator map for that.

Ocramius:
> I see only one decision - add checking discriminator value constraint to the table definition and separate option in discriminator map for that.

No way: this is a lot of added complexity for something that DB indexes solved already

lib/Doctrine/ORM/Mapping/DiscriminatorColumn.php
```diff
@@ -51,4 +51,9 @@
      * @var string
      */
     public $columnDefinition;
+
+    /**
+     * @var bool
+     */
+    public $strict;
```
Ocramius: As I mentioned before, an addition here is not to be considered unless an actual problem is being solved by it. So far, it seems like this only leads to removal of an `IN()` condition on discrimitor values when set to `false`

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -394,7 +394,15 @@ private function addDiscriminatorColumnDefinition($class, Table $table)
...
 
-        $table->addColumn($discrColumn['name'], $discrColumn['type'], $options);
+        $column = $table->addColumn($discrColumn['name'], $discrColumn['type'], $options);
+
+        if (isset($discrColumn['strict']) && $discrColumn['strict']) {
+            $values = [];
+            foreach ($class->discriminatorMap as $value => $className) {
+                $values[] = $this->platform->quoteStringLiteral($value);
+            }
+            $column->setCustomSchemaOption('check', 'CHECK (' . $discrColumn['name'] . ' IN (' . implode(',', $values) . '))');
```
Ocramius: Don't think we'll manage to make this work cross-platform

Ocramius (changes requested):

Ocramius:
Taking a decision not to pursue this further: there is some improvement in the generated SQL, yes, but most decent SQL query optimisers can deal with these details anyway.

Therefore, closing as `won't fix` here, sorry @11ya!

PR #7122, opened 2018-03-08
Tests: Return types for setUp, tearDown and all test methods

Majkl578:
`(\S+ function (?:setUp|tearDown|test\w+\([^)]*\)))$` -> `$1 : void`
+ some manual fixes for dependent non-void tests.

further scope in upcoming PRs:
* native return types for other methods
* native parameter types


carusogabriel (approved):
👍  for this step in use return types

Ocramius (approved):
:+1:

PR #7130, opened 2018-03-14
Strict typing in Doctrine\ORM, disconnect Doctrine\Common

Majkl578:
Initial round of strict typing.

* Disconnected Doctrine\Common persistence for now to allow typehinting changes and Doctrine\Persistence migration.
* Typehints for classes in the root Doctrine\ORM namespace.
* Conflict to prophecy versions without PHP 7.2 support (to be dropped in future).

TODO
* [ ] document signature changes - @Ocramius can you please suggest some format? Maybe outside UPGRADE to keep it clean, i.e. something like `UPGRADE-STRICT-API.md` with diff-like list of changes.
* [ ] squash?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,9 @@
         "phpstan/phpstan-shim": "^0.9.2",
         "phpunit/phpunit": "^7.0"
     },
+    "conflict": {
+        "phpspec/prophecy": "<1.7.5"
+    },
     "autoload": {
         "psr-4": { "Doctrine\\ORM\\": "lib/Doctrine/ORM" }
     },
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -4,10 +4,11 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM;
 
+use Doctrine\Common\Cache\Cache as CommonCache;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
-use Doctrine\DBAL\Driver\Statement;
+use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\ORM\Cache\Logging\CacheLogger;
 use Doctrine\ORM\Cache\QueryCacheKey;
 use Doctrine\ORM\Cache\TimestampCacheKey;
@@ -90,7 +91,7 @@ abstract class AbstractQuery
     /**
      * The hydration mode.
      *
-     * @var int
+     * @var int|string
      */
     protected $hydrationMode = self::HYDRATE_OBJECT;
 
@@ -157,11 +158,9 @@ abstract class AbstractQuery
     /**
      * Enable/disable second level query (result) caching for this query.
      *
-     * @param bool $cacheable
-     *
      * @return static This query instance.
      */
-    public function setCacheable($cacheable)
+    public function setCacheable(bool $cacheable) : self
     {
         $this->cacheable = (bool) $cacheable;
 
@@ -171,19 +170,17 @@ abstract class AbstractQuery
     /**
      * @return bool TRUE if the query results are enable for second level cache, FALSE otherwise.
      */
-    public function isCacheable()
+    public function isCacheable() : bool
     {
         return $this->cacheable;
     }
 
     /**
-     * @param string $cacheRegion
-     *
      * @return static This query instance.
      */
-    public function setCacheRegion($cacheRegion)
+    public function setCacheRegion(string $cacheRegion) : self
     {
-        $this->cacheRegion = (string) $cacheRegion;
+        $this->cacheRegion = $cacheRegion;
 
         return $this;
     }
@@ -193,7 +190,7 @@ abstract class AbstractQuery
      *
      * @return string|null The cache region name; NULL indicates the default region.
      */
-    public function getCacheRegion()
+    public function getCacheRegion() : ?string
     {
         return $this->cacheRegion;
     }
@@ -201,15 +198,12 @@ abstract class AbstractQuery
     /**
      * @return bool TRUE if the query cache and second level cache are enabled, FALSE otherwise.
      */
-    protected function isCacheEnabled()
+    protected function isCacheEnabled() : bool
     {
         return $this->cacheable && $this->hasCache;
     }
 
-    /**
-     * @return int
-     */
-    public function getLifetime()
+    public function getLifetime() : int
     {
         return $this->lifetime;
     }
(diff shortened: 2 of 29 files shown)


tests/Doctrine/Performance/Mock/NonProxyLoadingEntityManager.php
```diff
@@ -126,15 +128,15 @@ public function rollback() : void
     /**
      * {@inheritDoc}
      */
-    public function createQuery($dql = '') : Query
+    public function createQuery(string $dql = '') : Query
```
SenseException: While introducing BC breaks for the next major release, should we keep the default empty string? Does it make sense to have that? Or is this part of a later discussion?

composer.json
```diff
@@ -33,6 +33,9 @@
         "phpstan/phpstan-shim": "^0.9.2",
         "phpunit/phpunit": "^7.0"
     },
+    "conflict": {
+        "phpspec/prophecy": "<1.7.5"
+    },
```
Taluu: You are blocking potential users of prophecy < 1.7.5 even if they don't intend to test doctrine

Taluu: So instead of how requiring it in the `require-dev` section ?

tests/Doctrine/Performance/Mock/NonProxyLoadingEntityManager.php
Majkl578: Adding parameter type is not a BC break. 
At this time I don't want to change logic or accepted types too much where not needed, otherwise the PR(s) would be pain to review.

composer.json
Majkl578: Technically legit, but no one is supposed to use dev-master now so it's not a real issue. This is mostly to make CI work with lowest dependencies as well (too bad there is no `conflict-dev` section).

tests/Doctrine/Performance/Mock/NonProxyLoadingEntityManager.php
SenseException: Oh, right. Type widening in PHP 7.2. PHP 7.2 will be minimum version. I totally forgot that.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -870,11 +821,10 @@ public function execute($parameters = null, $hydrationMode = null)
      * Execute query ignoring second level cache.
      *
      * @param ArrayCollection|array|Parameter[]|mixed[]|null $parameters
-     * @param int|null                                       $hydrationMode
      *
      * @return mixed
      */
-    private function executeIgnoreQueryCache($parameters = null, $hydrationMode = null)
+    private function executeIgnoreQueryCache(?iterable $parameters = null, ?int $hydrationMode = null)
```
ostrolucky: $hydrationMode can be string for custom hydrators

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -661,11 +620,9 @@ public function getHydrationMode()
      *
      * Alias for execute(null, $hydrationMode = HYDRATE_OBJECT).
      *
-     * @param int $hydrationMode
-     *
      * @return mixed
      */
-    public function getResult($hydrationMode = self::HYDRATE_OBJECT)
+    public function getResult(int $hydrationMode = self::HYDRATE_OBJECT)
```
ostrolucky: $hydrationMode can be string for custom hydrators

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -857,7 +808,7 @@ public function iterate($parameters = null, $hydrationMode = null)
      *
      * @return mixed
      */
-    public function execute($parameters = null, $hydrationMode = null)
+    public function execute(?iterable $parameters = null, ?int $hydrationMode = null)
```
ostrolucky: $hydrationMode can be string for custom hydrators

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -929,11 +879,10 @@ private function executeIgnoreQueryCache($parameters = null, $hydrationMode = nu
      * Load from second level cache or executes the query and put into cache.
      *
      * @param ArrayCollection|array|Parameter[]|mixed[]|null $parameters
-     * @param int|null                                       $hydrationMode
      *
      * @return mixed
      */
-    private function executeUsingQueryCache($parameters = null, $hydrationMode = null)
+    private function executeUsingQueryCache(?iterable $parameters = null, ?int $hydrationMode = null)
```
ostrolucky: $hydrationMode can be string for custom hydrators

ostrolucky (changes requested):
Be careful about $hydrationMode, it can be a string for custom hydrators, see [here](https://github.com/doctrine/doctrine2/blob/0aac49572a17b8631b045a26753564563cc59490/lib/Doctrine/ORM/EntityManager.php#L782). Yes, int in docblock was wrong all the time.

I've not highlighted all references, it became too much and would just clutter this PR

Majkl578:
@ostrolucky Thanks, adressed. Probably something to look into later to remove `int|string` union.

composer.json
Taluu: It doesn't make any sense to have a conflict-dev, as if you install the dev dependencies, it means you cloned the doctrine repo and working directly on it (you can't install dev dependencies of a dependency...). So making it into the dev section of the requirement and removing the conflict makes sense IMO.

lib/Doctrine/ORM/Event/LifecycleEventArgs.php
```diff
@@ -15,23 +15,41 @@
...
+    {
+        return $this->entity;
+    }
+
     /**
      * Retrieves associated Entity.
-     *
-     * @return object
      */
     public function getEntity()
```
ostrolucky: You removed return type annotation without adding return type. Same for following methods in this class

Ocramius:
> * [ ] document signature changes - @Ocramius can you please suggest some format? Maybe outside UPGRADE to keep it clean, i.e. something like `UPGRADE-STRICT-API.md` with diff-like list of changes.

No idea how to make this clearer for readers. `UPGRADE.md` is the location read by people upgrading the lib: why would this be documented outside it?

composer.json
Ocramius: > So instead of how requiring it in the `require-dev` section ?

Nope, let's please not do that.

Users will be perfectly fine as long as they upgrade all their dependencies **as they should be doing**

Ocramius:
> > * [ ] document signature changes - @Ocramius can you please suggest some format? Maybe outside UPGRADE to keep it clean, i.e. something like `UPGRADE-STRICT-API.md` with diff-like list of changes.

 > No idea how to make this clearer for readers. `UPGRADE.md` is the location read by people upgrading the lib: why would this be documented outside it?

Errata: I got a better idea here.

Let's try this upgrade path:

 1. all touched files must become either `interface` or `final class` - nothing in-between except for known edge-cases (`EntityManager`?)
 2. let's list all classes where `final` was added
 3. let's list all the interfaces in the diff

That's a sufficient `UPGRADE.md` entry.

Ocramius:
Also: I'm working on `roave/api-compare` this weekend: hope that will help in automating things. This patch is a good test-bed for that library too :+1:  /cc @asgrim

PR #7135, opened 2018-03-16
Getting Started examples do not work. Fixes via StackOverflow answers.

hostyle:
http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/getting-started.html

These docs are either out of date or simply untested:

With help from: https://stackoverflow.com/questions/17473225/doctrine2-no-metadata-classes-to-process and: https://stackoverflow.com/questions/48168310/doctrine-annotation-doctrine-orm-annotation-entity-in-class-does-not-exist-or I was able to get the first examples running.

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -132,7 +132,7 @@ step:
 
     // Create a simple "default" Doctrine ORM configuration for Annotations
     $isDevMode = true;
-    $config = Setup::createAnnotationMetadataConfiguration(array(__DIR__."/src"), $isDevMode);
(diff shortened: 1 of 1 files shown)


docs/en/tutorials/getting-started.rst
```diff
@@ -274,7 +274,7 @@ but you only need to choose one.
         <?php
         // src/Product.php
 
-        use Doctrine\ORM\Annotation as ORM;
+        use Doctrine\ORM\Mapping as ORM;
```
Ocramius: This is because `master` has a completely different structure. Please refer to the `2.6` branch for the current docs :-)

PR #7143, opened 2018-03-20
Removing the metadata instantiation from the EM

mikeSimonson:
The metadata instantiation is kept in the create method for BC for now.
A new static method is added to be able to pass the metadata to the EM.

## TODOs

 * [ ] Document BC Breakages
 * [ ] Add lazy collection implementation (populated as-we-go)
 * [ ] Ensure type aliases are correctly cached (create key copies in the metadata)
 * [ ] Ensure the `\\` `ltrim()` is gone


mikeSimonson:
@Ocramius Is this what you  meant when we discussed at phpbenelux ?
Is the next step refactoring all the call to create to the new createWithMetadata function ?

Ocramius:
@mikeSimonson the idea was to make the metadata a `ClassMetadata[]` lazy collection (injectable) rather than a `ClassMetadataFactory` - is this a middle-ground before we go there?

mikeSimonson:
I just wanted to make sure that I was going in the right direction.
I should pull lazymap as a dep then I suppose ?

Ocramius:
@mikeSimonson I'd suggest making it an eager map for now, and later we fix it with `LazyMap`

mikeSimonson:
Does the new commit goes in the right direction ?

Can I
    get rid of the [cache configuration in the EM constructor](https://github.com/doctrine/doctrine2/pull/7143/files#diff-cfa931dd1195fd8c5e36f3706b028dbaR153) ?
    change the behavior of [getAllMetadata](https://github.com/doctrine/doctrine2/pull/7143/files#diff-dfe17e5e229e8d404df56b3c4849d511R135) for simplicity ?
    remove [the initialize thingy in the ClassMetadataFactory](https://github.com/doctrine/doctrine2/pull/7143/files#diff-ecaa162f62bde19c758286c766911141R83) ?

How do I proceed to remove [the entity manager from the OnClassMetadataNotFoundEventArgs in the ClassMetadataFactory](https://github.com/mikeSimonson/doctrine2/blob/bc4133e3e4f530285168afacc95c591a4679eccc/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L97) and [LoadClassMetadataEventArgs](https://github.com/mikeSimonson/doctrine2/blob/bc4133e3e4f530285168afacc95c591a4679eccc/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L166) ?

Any other remarks / comment / direction ?
@Ocramius

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -155,7 +155,13 @@ protected function __construct(Connection $conn, Configuration $config, EventMan
         }
     }
 
-
+    /**
+     * @return MetadataCollection
+     */
+    public function getMetadatas(): MetadataCollection
```
Ocramius: Would avoid exposing it here at all.

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -134,52 +128,47 @@
...
+            //$this->cache  = $cacheFactory->createCache($this);
         }
     }
 
     /**
-     * {@inheritDoc}
+     * @return MetadataCollection
      */
-    public function getConnection()
+    public function getMetadatas(): MetadataCollection
```
Ocramius: Avoid exposing this

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -273,7 +262,8 @@ public function rollback()
      */
     public function getClassMetadata($className) : Mapping\ClassMetadata
     {
-        return $this->metadataFactory->getMetadataFor($className);
+        $className = StaticClassNameConverter::getRealClass($className);
```
Ocramius: This can likely be dealt with in a `MetadataCollection` decorator instead

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -134,52 +128,47 @@
...
-        $metadataFactoryClassName = $config->getClassMetadataFactoryName();
-
-        $this->metadataFactory = new $metadataFactoryClassName();
-
-        $this->metadataFactory->setEntityManager($this);
-        $this->metadataFactory->setCacheDriver($this->config->getMetadataCacheImpl());
+        $this->conn            = $conn;
+        $this->config          = $config;
+        $this->eventManager    = $eventManager;
+        $this->metadatas       = $metadataCollection;
```
Ocramius: `metadata` is irregular: the plural of `metadata` is `metadata`. A better way to name this is probably `mappings`

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -382,8 +372,13 @@ public function find($entityName, $id, $lockMode = null, $lockVersion = null)
         }
 
         foreach ($id as $i => $value) {
-            if (is_object($value) && $this->metadataFactory->hasMetadataFor(StaticClassNameConverter::getClass($value))) {
-                $id[$i] = $this->unitOfWork->getSingleIdentifierValue($value);
+            if (is_object($value)) {
+                try{
+                    $this->metadatas->get(StaticClassNameConverter::getClass($value));
+                    $id[$i] = $this->unitOfWork->getSingleIdentifierValue($value);
+                } catch (\Exception $e) {
```
Ocramius, line 363: Most likely wrong here

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -382,8 +372,13 @@ public function find($entityName, $id, $lockMode = null, $lockVersion = null)
         }
 
         foreach ($id as $i => $value) {
-            if (is_object($value) && $this->metadataFactory->hasMetadataFor(StaticClassNameConverter::getClass($value))) {
-                $id[$i] = $this->unitOfWork->getSingleIdentifierValue($value);
+            if (is_object($value)) {
+                try{
+                    $this->metadatas->get(StaticClassNameConverter::getClass($value));
```
Ocramius: Same here: move `StaticClassNameConverter` into a detail of `MetadataCollection`

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -479,8 +474,13 @@ public function getReference($entityName, $id)
         foreach ($id as $i => $value) {
             $scalarId[$i] = $value;
 
-            if (is_object($value) && $this->metadataFactory->hasMetadataFor(StaticClassNameConverter::getClass($value))) {
-                $scalarId[$i] = $this->unitOfWork->getSingleIdentifierValue($value);
+            if (is_object($value)) {
+                try{
+                    $this->metadatas->get(StaticClassNameConverter::getClass($value));
+                    $scalarId[$i] = $this->unitOfWork->getSingleIdentifierValue($value);
+                } catch (\Exception $e) {
```
Ocramius, line 465: Likely wrong here - why did you introduce these try-catches? I know you are still playing around with the patch: just wondering why this got added to make things work

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -541,8 +541,13 @@ public function getPartialReference($entityName, $id)
         }
 
         foreach ($id as $i => $value) {
-            if (is_object($value) && $this->metadataFactory->hasMetadataFor(StaticClassNameConverter::getClass($value))) {
-                $id[$i] = $this->unitOfWork->getSingleIdentifierValue($value);
+            if (is_object($value)) {
+                try{
+                    $this->metadatas->get(StaticClassNameConverter::getClass($value));
+                    $id[$i] = $this->unitOfWork->getSingleIdentifierValue($value);
+                } catch (\Exception $e) {
```
Ocramius, line 532: Likely wrong here - why did you introduce these try-catches? I know you are still playing around with the patch: just wondering why this got added to make things work

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -829,7 +834,23 @@ public static function create($connection, Configuration $config, ?EventManager
 
         $connection = static::createConnection($connection, $config, $eventManager);
 
-        return new EntityManager($connection, $config, $connection->getEventManager());
+        $metadataFactoryClassName = $config->getClassMetadataFactoryName();
+        /** @var \Doctrine\Common\Persistence\Mapping\ClassMetadataFactory $metadataFactory */
+        $metadataFactory          = new $metadataFactoryClassName($config, $connection, $eventManager);
```
Ocramius: :+1: this is a decent BC-compliant approach for now :)

lib/Doctrine/ORM/Mapping/MetadataCollection.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function get($name)
+    {
+        if(!isset($this->classMetadatas[$name])){
+            throw new \Exception('No metadata found for ' . $name);
+        }
+        return $this->classMetadatas[$name];
+    }
+
+    public static function fromClassMetadatas(array $classMetadatas)
```
Ocramius: I suggest using `(ClassMetadata $firstClass, ClassMetadata ...$otherClasses)` here, which prevents instantiation if no metadata is available at all.

lib/Doctrine/ORM/Mapping/MetadataCollection.php
```diff
@@ -0,0 +1,43 @@
...
+                return $carry;
+            } ,
+            []
+        ) ;
+        $this->classMetadatas = $metadataByClassName;
+    }
+
+    public function get($name)
+    {
+        if(!isset($this->classMetadatas[$name])){
```
Ocramius: Add the class name normalisation here

lib/Doctrine/ORM/Mapping/MetadataCollection.php
```diff
@@ -0,0 +1,43 @@
...
+{
+
+    /**
+     * @var ClassMetadata[]
+     */
+    private $classMetadatas;
+
+    private function __construct(array $classMetadatas)
+    {
+        $metadataByClassName = array_reduce (
```
Ocramius: This is simpler with:

```php
array_combine(
    array_map(function (ClassMetadata $metadata) { return $metadata->getClassName(); }, $metadata),
    $metadata
);
```

Ocramius (changes requested):

Ocramius:
> Does the new commit goes in the right direction ?

:+1: 

 > Can I get rid of the [cache configuration in the EM constructor](https://github.com/doctrine/doctrine2/pull/7143/files#diff-cfa931dd1195fd8c5e36f3706b028dbaR153) ?

Most likely. Metadata caching can be moved to a `CachedMetadataCollection` or such

 > change the behavior of [getAllMetadata](https://github.com/doctrine/doctrine2/pull/7143/files#diff-dfe17e5e229e8d404df56b3c4849d511R135) for simplicity ?

That is a subtle BC break: please don't change that.

 > remove [the initialize thingy in the ClassMetadataFactory](https://github.com/doctrine/doctrine2/pull/7143/files#diff-ecaa162f62bde19c758286c766911141R83) ?

Yes: since metadata now has a lifecycle that is completely separate from the ORM itself, removal is fine here :+1: 

 > How do I proceed to remove [the entity manager from the OnClassMetadataNotFoundEventArgs in the ClassMetadataFactory](https://github.com/mikeSimonson/doctrine2/blob/bc4133e3e4f530285168afacc95c591a4679eccc/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L97) and [LoadClassMetadataEventArgs](https://github.com/mikeSimonson/doctrine2/blob/bc4133e3e4f530285168afacc95c591a4679eccc/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L166) ?

I suggest we completely remove these events, as the user can simply decorate the `MetadataCollection` in order to manipulate metadata information

 > Any other remarks / comment / direction ?

Provided above :-)

lib/Doctrine/ORM/EntityManager.php
mikeSimonson: just because for now get throws an Exception if the metadata is not found where it wasn't before.

lib/Doctrine/ORM/Mapping/MetadataCollection.php
```diff
@@ -0,0 +1,41 @@
...
+    public function get($name)
+    {
+        $name = StaticClassNameConverter::getRealClass($name);
+        if(!isset($this->metadata[$name])){
+            throw new \Exception('No metadata found for ' . $name);
+        }
+        return $this->metadata[$name];
+    }
+
+    public static function fromClassMetadatas(ClassMetadata $firstClass, ClassMetadata ...$otherClasses)
```
Majkl578, line 42: Why is this a static instead of constructor directly?

Also the signature is ugly, let's just have `__construct(ClassMetadata ...$metadata)` with `assert(count($metadata) !== 0)` if it really MUST be non-empty.

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -784,7 +773,7 @@ public function newHydrator($hydrationMode)
                 return new Internal\Hydration\SimpleObjectHydrator($this);
 
             default:
-                $class = $this->config->getCustomHydrationMode($hydrationMode);
+                $class = $this->config->getCustomHydrationMode((string) $hydrationMode);
```
Majkl578: This is most likely wrong, see https://github.com/doctrine/doctrine2/pull/7130#discussion_r175259739. It could be `int|string` currently.

mikeSimonson: getCustomHydrationMode expect a string in it's function definition.

Majkl578: :thinking:

Majkl578: Changed this in my PR since it's really `int|string` now. Something to re-do later on probably.

mikeSimonson:
@Ocramius My main question now is to know if passing the whole mapping collection to the BasicEntityPersister is the way to go ? (commit extract the MetadataFactory from the basic entity persister and the following one)

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -541,8 +525,13 @@ public function getPartialReference($entityName, $id)
         }
 
         foreach ($id as $i => $value) {
-            if (is_object($value) && $this->metadataFactory->hasMetadataFor(StaticClassNameConverter::getClass($value))) {
-                $id[$i] = $this->unitOfWork->getSingleIdentifierValue($value);
+            if (is_object($value)) {
+                try{
```
Ocramius, line 529: Same: try-catch should be removed overall. We possibly need a better abstraction.

lib/Doctrine/ORM/Mapping/MetadataCollection.php
```diff
@@ -0,0 +1,47 @@
...
+        $metadataByClassName = array_combine(
+            array_map(function (ClassMetadata $metadata) { return $metadata->getClassName(); }, $metadata),
+            $metadata
+        );
+        $this->metadata = $metadataByClassName;
+    }
+
+    public function get(string $name) : ClassMetadata
+    {
+        $name = StaticClassNameConverter::getRealClass($name);
```
Ocramius: Consider checking for existence before calling `StaticClassNameConverter`. Copying a metadata reference into a different key is also a good idea for performance reasons

lib/Doctrine/ORM/Mapping/MetadataCollection.php
```diff
@@ -0,0 +1,47 @@
...
+
+class MetadataCollection
+{
+
+    /**
+     * @var ClassMetadata[]
+     */
+    private $metadata;
+
+    private function __construct(array $metadata)
```
Ocramius: Document this please

lib/Doctrine/ORM/Mapping/MetadataCollection.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $metadataByClassName = array_combine(
+            array_map(function (ClassMetadata $metadata) { return $metadata->getClassName(); }, $metadata),
+            $metadata
+        );
+        $this->metadata = $metadataByClassName;
+    }
+
+    public function get(string $name) : ClassMetadata
```
Ocramius: The various APIs getting class metadata by object reference could need a `getByObject(object $object)`

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -743,8 +744,66 @@ protected function getEntityManager(?Connection $connection = null, ?MappingDriv
 
         $config->setMetadataDriverImpl(
             $mappingDriver ?? $config->newDefaultAnnotationDriver([
+                realpath(__DIR__ . '/Models/CMS'),
```
Ocramius: Wondering why this test changed to load everything - is this because we don't yet have a lazy collection?

Ocramius (changes requested):
> @Ocramius My main question now is to know if passing the whole mapping collection to the BasicEntityPersister is the way to go ? (commit extract the MetadataFactory from the basic entity persister and the following one)

@mikeSimonson I'd suggest that any component requiring metadata should just be given the collection. The `MetadataCollection` is now to be considered a "dumb" immutable structure, so it should be fine to pass it around.

Obviously, if only one entity metadata is required in a persister, then it should be passed in explicitly, but I don't know how big that change would be.

It will be easy to extract these scenarios when looking at classes that always `get()` on the collection with the same parameter.

Ocramius:
@mikeSimonson I've updated the PR description with the TODOs matching the discussion above.

tests/Doctrine/Tests/OrmFunctionalTestCase.php
mikeSimonson: Just because I am pocking around a bit everywhere to try and fix everything that I broke.

Majkl578:
@mikeSimonson Do you plan to continue with this?

mikeSimonson:
@Majkl578 yes. I don't have much time right now but I hope that I will have it in a good state for amsterdamphp

derrabus:

Issue #7140, opened 2018-03-20
Relation hydration fails, stays integer instead of creating Proxy


JKapitein:
In specific cases in a project I've inherited, an entity relationship can fail to hydrate. So far, we've been able to use stopgap solutions, mostly applying eager fetching, but this bug keeps popping up in other places.

We were able to make a reproduction on a clean Symfony Flex project, which you can find here
https://github.com/JKapitein/doctrine-hydration

Steps to reproduce:

1. Load in a single entity where the relations go (at least?) 2 deep. In the reproduction, we load it in through the Symfony ParamConverter
2. Within the controller action, call a repository function to load a list of the same type as the entity, joining the relations in the QueryBuilder. The list must include the entity that was already loaded.
3. The 2nd level relation will not get hydrated, and will only return the scalar ID, leading to errors if you try to fetch data from that relation

We're not certain if this behaviour stems from the relationship structure (as the ID on Person is a relationship), or if there's any other reason, but guidance would be appreciated.

Ocramius:
This requires an isolated test case to be added to https://github.com/doctrine/doctrine2/tree/41292e85dfadac771713882663ab5389e1070ecc/tests/Doctrine/Tests/ORM/Functional/Ticket

We can't do much with an example app.

JKapitein:
It passes on master but not on the 2.7 branch. I've created the pull request against 2.7, since master seems like a rewrite of sorts?

Ocramius:
@JKapitein yeah, `master` is `3.x` now. If you are fixing an issue affecting `2.x`, I suggest you open a PR against `2.6`

JKapitein:
Thank you @Ocramius, I've recreated the pull request against 2.6

Best I can tell is that the 'fetched' hint in UnitOfWork::createEntity gets set to true for the "account" relation, presumably because of the query builder, so the proxy is never created for the already existing entity?

PR #7145, opened 2018-03-21
Failing testcase for #7140

JKapitein:
See https://github.com/doctrine/doctrine2/issues/7140 

2nd level relation hydration fails when object is already loaded and is then "reloaded" by a separate query where that relation is joined in.


Ocramius:
This looks similar to https://github.com/doctrine/doctrine2/pull/625 in nature.

JKapitein:
Looks to be the same problem yeah. Nothing to be done about it until 3.0 then? Making the test case at least gave us some potential workarounds to use for now.

Ocramius:
It's not that nothing can be done, but rather that a decision about eager loading has to be taken, and it might result in a BC break.

We didn't find a solution on our end (yet).

PR #7154, opened 2018-03-23
Fix getting started syntax issue causing rendering glitches

ogizanagi:
consistent with `DTOs` in next sentence and fixing this glitch:

screenshot 2018-03-23 a 22 22 50


Ocramius (approved):
Thanks @ogizanagi!

PR #7162, opened 2018-03-28
Do not schedule detached entities for orphan removal

thexpand:
Detached entities should not be scheduled for orphan removal,
because they have not yet been persisted and will result in an exception
when the entity manager is flushed.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1991,9 +1991,9 @@ class UnitOfWork implements PropertyChangedListener
      */
     public function scheduleOrphanRemoval($entity)
     {
-        // Detached entities should not be scheduled for orphan removal,
-        // because they have not yet been persisted.
-        if ($this->getEntityState($entity) === self::STATE_DETACHED) {
+        // Entities should not be scheduled for orphan removal,
+        // if they do not exist in the identity map or if they are not about to be inserted.
+        if (!$this->isInIdentityMap($entity) && !$this->isScheduledForInsert($entity)) {
             return;
         }
 
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1991,7 +1991,13 @@ class UnitOfWork implements PropertyChangedListener
      */
     public function scheduleOrphanRemoval($entity)
     {
-        $this->orphanRemovals[spl_object_id($entity)] = $entity;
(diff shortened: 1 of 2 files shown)


Ocramius:
This makes sense, but requires a test addition. See https://github.com/doctrine/doctrine2/tree/master/tests/Doctrine/Tests/ORM/Functional/Ticket for examples

thexpand:
@Ocramius How should I name the file, and after it's done how can I run the test.. or it's being run automatically after pushing to the branch?

Ocramius:
You can name it GH7162Test. It will be executed once you push your branch again 👍

thexpand:
@Ocramius Now I see that some other tests are breaking, because of this, and it's quite logical. Actually, I'm trying to figure out why does a non-persisted entity have a detached state... Any ideas?

thexpand:
I think I caught my error.
In the constructor of my entities, I was setting a default value for each of them to `(int) 0`
If we take a closer look here: https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/UnitOfWork.php#L1390
we can see that it checks whether for an identifier value. In my case, it's an array with 1 element (the id), which has a value of zero, meaning the array is not empty, thus the condition will not return true and the state of the object in the collection will not be `STATE_NEW`, but rather `STATE_DETACHED`.

I can easily refactor my project and change the $id property to use null value as it's initial value.
@Ocramius is there anything else, that I'm missing? Do you think it's okay to leave it that way, or should we fix the `getEntityState` method to return `STATE_NEW` when an identifier value is zero?

Ocramius:
> I was setting a default value for the id column

Ouch, that can indeed lead to problems.

 > hould we fix the `getEntityState` method to return `STATE_NEW` when an identifier value is zero?

The correct value for "missing identifier" is `null`, nothing else. The reason is that `null` has a very specific meaning in DB design and in PHP, as it represents the "absence of value".

I explained it in more detail somewhere on stackoverflow, but can't find the link right now :-(

Similarly related: https://softwareengineering.stackexchange.com/questions/32578/sql-empty-string-vs-null-value

Ocramius:
@thexpand thanks a lot for expanding into this and actually digging to the core of the issue: I wish everyone was as determined as you were with this issue. Sorry if it ended up being invalid, but at least you found the culprit.

thexpand:
@Ocramius I'm actually quite happy with the result, despite the issue being invalid. Issues just like this one give me the missing knowledge and experience.

PR #7170, opened 2018-04-03
Document BC-break due to type declaration

lcobucci:
I forgot to add this in https://github.com/doctrine/doctrine2/pull/7092, so here it goes.


Ocramius (approved):
LGTM :+1:

PR #7172, opened 2018-04-03
Remove redundant isset

carusogabriel:
Looks like the previous logic was checking if the `key` exists in an `array`, allowing `null`.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -590,7 +590,7 @@ class UnitOfWork implements PropertyChangedListener
 
             foreach ($actualData as $propName => $actualValue) {
                 // skip field, its a partially omitted one!
-                if (! (isset($originalData[$propName]) || array_key_exists($propName, $originalData))) {
+                if (! array_key_exists($propName, $originalData)) {
                     continue;
                 }
 
--- a/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
@@ -248,7 +248,7 @@ class Paginator implements \Countable, \IteratorAggregate
         foreach ($parameters as $key => $parameter) {
             $parameterName = $parameter->getName();
 
-            if (! (isset($parameterMappings[$parameterName]) || array_key_exists($parameterName, $parameterMappings))) {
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -590,7 +590,7 @@ public function computeChangeSet(ClassMetadata $class, $entity)
 
             foreach ($actualData as $propName => $actualValue) {
                 // skip field, its a partially omitted one!
-                if (! (isset($originalData[$propName]) || array_key_exists($propName, $originalData))) {
+                if (! array_key_exists($propName, $originalData)) {
```
Ocramius: This is done for performance reasons (fcall is much slower than isset)

Ocramius (changes requested):

lib/Doctrine/ORM/UnitOfWork.php
carusogabriel: Ah :disappointed: 

Is something that we can measure at userland to see if this patch affects it, or should we leave as it is? :smile:

Ocramius:
You can probably already measure it with the existing performance test suite: just make sure to disable XDebug first

carusogabriel:
@Ocramius PHP output proving `Xdebug` is disabled:
```
PHP 7.2.3-1+ubuntu17.10.1+deb.sury.org+1 (cli) (built: Mar  6 2018 11:19:27) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.3-1+ubuntu17.10.1+deb.sury.org+1, Copyright (c) 1999-2018, by Zend Technologies
```
Average of 3 `vendor/bin/phpunit` locally:
```diff
-master: 7.22s
+patch: 7.44s
```
:open_mouth:

Ocramius:
That's the location where I'd like to add GeneratedHydrator too 😥

Majkl578:
Related: https://bugs.php.net/bug.php?id=76148

Ocramius:
Closing here meanwhile.

lcobucci:
@carusogabriel FYI we have configured `phpbench` to run the performance tests, it's more reliable to use it instead of `phpunit` 😄

lcobucci:
http://www.phpbench.com

PR #7173, opened 2018-04-04
Fix Criteria's orderBy example

malarzm:
This was reported in https://github.com/doctrine/collections/issues/148

### DIFF
--- a/docs/en/reference/query-builder.rst
+++ b/docs/en/reference/query-builder.rst
@@ -505,7 +505,7 @@ using ``addCriteria``:
     // ...
 
     $criteria = Criteria::create()
-        ->orderBy(['firstName', 'ASC']);
+        ->orderBy(['firstName' => Criteria::ASC]);
 
     // $qb instanceof QueryBuilder
     $qb->addCriteria($criteria);


docs/en/reference/query-builder.rst
```diff
@@ -505,7 +505,7 @@ using ``addCriteria``:
     // ...
 
     $criteria = Criteria::create()
-        ->orderBy(['firstName', 'ASC']);
+        ->orderBy(['firstName' => 'ASC']);
```
Ocramius: Can you please use `Criteria::ASC` here?

Ocramius (changes requested):

docs/en/reference/query-builder.rst
malarzm: Sure, gimme a moment though as now I need to clone :D

Ocramius: Edit your branch from the online editor :D

malarzm: And clone later because of squash request? :P

Ocramius: What squash request? I'm not THAT mad.

Ocramius (approved):

docs/en/reference/query-builder.rst
malarzm: I'd normally ask for one ;)

PR #7182, opened 2018-04-10
Update `getting-started.rst` - fix `$entityManager` variable reference

rtoledof:
Change CreateHelperSet parameter to array with the key 'em'


Ocramius:
@rtoledof the diff is empty, can you check that please?

PR #7188, opened 2018-04-10
[TASK] Add documentation generation info for MacOS

afoeder:
Besides, this adds some formatting and improves sectioning.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,12 +1,23 @@
 # Doctrine ORM Documentation
 
 ## How to Generate:
-Using Ubuntu 14.04 LTS:
+
+The following will generate the documentation into the `build` directory of the checkout.
+
+### Using Ubuntu 14.04 LTS:
 
 1. Run ./bin/install-dependencies.sh
 2. Run ./bin/generate-docs.sh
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1,12 +1,23 @@
...
+
+### Using Ubuntu 14.04 LTS:
 
 1. Run ./bin/install-dependencies.sh
 2. Run ./bin/generate-docs.sh
 
-It will generate the documentation into the build directory of the checkout.
+### Using macOS:
+
+Sphinx Installation instructions can be found at the
```
Ocramius: This is no longer needed with the new website, tbh /cc @jwage

afoeder: ah nice even better; just stumbled upon this when I wanted to verify #7185 and wanted to share my insights :) I realized that it gets rendered in the old-school style, but for veryfing the linking it was fair enough.

jwage: This is updated here https://github.com/doctrine/doctrine2/pull/7177

afoeder: ok, I guess mine can be closes then. Thanks!

PR #7190, opened 2018-04-10
Fix wrong type in phpdoc of `Doctrine\ORM\Id\AbstractIdGenerator#generate()`

Tobion:
\Doctrine\ORM\Mapping\Entity is the annotation class which is not correct. The entity object itself is meant here as tests also assume see https://github.com/doctrine/doctrine2/blob/2.6/tests/Doctrine/Tests/ORM/Id/AssignedGeneratorTest.php#L28

Found this when running phpstan on our code that used a custom generator.

### DIFF
--- a/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
+++ b/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
@@ -27,7 +27,7 @@ abstract class AbstractIdGenerator
      * Generates an identifier for an entity.
      *
      * @param EntityManager $em
-     * @param \Doctrine\ORM\Mapping\Entity $entity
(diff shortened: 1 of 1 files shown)


Ocramius:
@Tobion is this still occurring in `master`?

Tobion:
I don't find the ID generators in master. Did they get extracted somewhere?

Tobion:
It's fixed in master https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Sequencing/Generator.php#L16

The namespace got changed. This breaking change seems to be missing in the uprade file.

Ocramius (approved):
LGTM! Thanks @Tobion

PR #7195, opened 2018-04-13
Move orm exception up

greg0ire:
Not sure how I overlooked this, but I kept those interfaces for BC, and then… changed the namespace. This puts them back where they originally were.


greg0ire:
In fact this wasn't a mistake, if we want to deprecate `ORMException` in favor of `Exception\ORMException`, then we should not move back to `ORMException` in v3

PR #7196, opened 2018-04-13
Document exception-splitting-related BC-breaks

### LINKED ISSUE #7194: Document BC breaks introduced by #6743 (Split `ORMException` and clean up exception types)
Needed for #6743

greg0ire:
Closes #7194 

~Should be modified if #7195 is not merged~ done

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,42 @@
 # Upgrade to 3.0
 
-## BC BREAK: Repository classes have to extend `EntityRepository`
+## BC Break: `Exception\ORMException` is no longer a class, but an interface
+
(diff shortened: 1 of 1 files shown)


Ocramius (dismissed):
LGTM, thanks @greg0ire!

Majkl578:
Actually, how about we stop using "create" and use "new" instead?
```php
throw Exception\MismatchedEventManager::new()
```

https://3v4l.org/CZ1fl

greg0ire:
On one hand, that will probably give weird syntactic coloration like it does on 3v4l, OTOH, it's nice because people might be less confused about what it does, and on the third hand if you're an :alien: , are they really confused about what `create` does?

derrabus (approved):

PR #1181, opened 2014-11-11
Support fetching entities by aliased name

Ocramius:
See #385 (this PR supersedes it)

The final aim is allowing something like:

``` php
$user = $em->find(UserInterface::class, 123);
```

This PR depends also on doctrine/common#342

PR #7199, opened 2018-04-17
Inherit from homonym exception in parent package

### LINKED ISSUE #10591: `ORM\MappingException` does not extend `Persistence\MappingException`
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | any

Hello!

is it normal that `Doctrine\ORM\Mapping\MappingException` does not extend `Doctrine\Persistence\Mapping\MappingException`? 

(but `\Doctrine\ODM\MongoDB\Mapping\MappingException` does!)

greg0ire:
This means that a class implementing a `doctrine/persistence` interface or
extending a class from that package can throw a specialized exception
that will still be caught by code in the parent package.
Fixes #7199

### DIFF
--- a/lib/Doctrine/ORM/Mapping/MappingException.php
+++ b/lib/Doctrine/ORM/Mapping/MappingException.php
@@ -6,7 +6,7 @@ namespace Doctrine\ORM\Mapping;
 
 use BackedEnum;
 use Doctrine\ORM\Exception\ORMException;
-use Exception;
+use Doctrine\Persistence\Mapping\MappingException as PersistenceMappingException;
 use LibXMLError;
 use ReflectionException;
 use ValueError;
@@ -24,7 +24,7 @@ use const PHP_EOL;
 /**
  * A MappingException indicates that something is wrong with the mapping setup.
  */
-class MappingException extends Exception implements ORMException
+class MappingException extends PersistenceMappingException implements ORMException
 {
     /** @param class-string $entityName */
     public static function identifierRequired(string $entityName): self


Majkl578:
Can we maybe for now not wire anything from Common (anymore), it would complicate migration later on. I actually started disconnecting Common mostly in order to migrate to typed code in #7130.

Majkl578:
Good overall, but this is not happening anytime soon. First we need to disconnect Common and migrate to doctrine/persistence 2.0 (which doesn't even have roadmap yet).
For starters, we need to extract the _disconnect Common_ part from #7130.

greg0ire:
The migration to persistence has been done, let's resume this.
Here are some things I could have bothered clarifying when I originally wrote this 4 years ago:

## The use case

This is not theoretical, there is a `catch (MappingException …` inside `doctrine/persistence`:
https://github.com/doctrine/persistence/blob/f8776dd9a0bdcd838812951a75f4ada72065a82a/src/Persistence/Mapping/AbstractClassMetadataFactory.php#L281-L289
This results in calling a method that does nothing, but can be overridden, like here:
https://github.com/doctrine/orm/blob/f4d5283f705f6ac51e81c8299ace074a619b40d7/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L88-L101

This means that as an ORM 2.x user, if https://github.com/doctrine/orm/pull/6728 gets merged, you would have to choose between XML validation and getting an event

## Why I'm not targeting 2.12.x

On 2.x `ORMException` is a class. On 3.x, it's an interface that `MappingException` implements, which makes us free to extend what we want (and is probably why `MappingException` wasn't extended in the first place).

derrabus (dismissed):
I'm fine with this change in general. However, looking at the change and why we could not implement it in 2.x, I wonder if persistence should provide an interface for mapping exceptions instead of a base class.

greg0ire:
@derrabus :thinking: I'm sad I didn't think of that, but yeah, that's a great idea! The only issue I see with this is one of naming, since we try not to have an `Interface` suffix, it's going to be hard to find a name that does not collide with the `MappingException` class from persistence.

greg0ire:
After taking a closer look, I realized that exceptions from ORM probably do not fall in the use case mentioned above. The `catch` statement catches `MappingException` because there is nothing more precise to catch, but is meant to catch the exception that occurs when loading metadata for an interface: https://github.com/doctrine/orm/pull/1181

This can be improved in at least 2 ways:
- extract that exception in a separate class and narrow it down;
- deprecate that hook in favor on another that would kick in earlier if the name is an interface name.

Both fixes should occur in `doctrine/persistence` and there is nothing to fix in ORM for now.
This also mean this issue is not a blocker for https://github.com/doctrine/orm/pull/6728, cc @Kern046 :)

greg0ire:
Reopening: ~there is still no bug caused by this, but it looks weird to people (see #10591)~ this seems to cause issues outside Doctrine: https://github.com/zenstruck/foundry/pull/441

derrabus (approved):

derrabus:
I still believe that Persistence should provide an interface instead of a base class though.

SenseException (approved):

PR #7200, opened 2018-04-17
Failing test: loading cached collection breaks if EM is cleared/closed

janlanger:
Steps to reproduce:

- have SLC and cached one-to-many collection
- load the entity with 1:N association
- close/clear EM (intentionally or due to error)
- try to load items in the association
=> 💥

**Without SLC, it works**

The underlining cause is, that `AbstractEntityPersister::loadOneToManyCollection` tries to access entity identifiers in `UoW:$entityIdentifiers` - which is empty after `em::clear`.

Attached is failing test case.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheLoadAfterClearTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheLoadAfterClearTest.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use Doctrine\ORM\PersistentCollection;
+use Doctrine\Tests\Models\Cache\Attraction;
+use Doctrine\Tests\Models\Cache\Bar;
+use Doctrine\Tests\Models\Cache\Beach;
+use Doctrine\Tests\Models\Cache\City;
+use Doctrine\Tests\Models\Cache\Restaurant;
+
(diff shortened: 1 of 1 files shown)


janlanger:
I've tried to prepare fix for this issue, but only come up with this workaround in our fork https://github.com/slevomat/doctrine-orm/commit/9ce16b88ca7c9593d87586600ac72118caaef17f - if identity map was cleared entity will always be loaded from database.

Would this be a sufficent fix? Should I make a PR?

SenseException:
@janlanger Can you please describe your use case that requires the suggested behavior of your workaround? I'm not quite sure if your proposed fix is a fitting solution or not.

janlanger:
@SenseException it happens what something goes wrong (unique constrains violation for example), but you still want to render the error page/flash message containing some data from database (eg. "Product <name> is sold out").

My linked workaround solves our issue, we use it in production.

SenseException:
If a unique constraint violation happens in your database, a validation of the inserted data is missing. Unique constraint errors have to be catched in userland before they can occur. You can evaluate the validation constraints for your later rendering of the error message.

janlanger:
@SenseException yes, usually it should, sometimes it isn't. Check before insertion is not atomic and sometimes will fail. But thats completely unrelated to this issue.

Imagine here https://github.com/doctrine/orm/pull/7200/files#diff-50e8ab2ce1c08b50ad0da5538feec4b6R29 ANY code that will cause EM to be locked or cleared. Does not have to be related to City entity. Without SLC the following line will work, with SLC it will not.

PR #7207, opened 2018-04-25
connect to master for schema updates

wodka:
Force to connect to master in schema tool.

I think this fixes #4183

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/AbstractCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/AbstractCommand.php
@@ -35,6 +35,12 @@ abstract class AbstractCommand extends Command
 
         /** @var EntityManagerInterface $em */
         $em = $emHelper->getEntityManager();
+        $connection = $em->getConnection();
+
+        if ($connection instanceof MasterSlaveConnection) {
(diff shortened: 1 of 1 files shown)


Ocramius:
I'd rather say that the `MasterSlaveConnection` should correctly refuse to run DDL on slave servers.

wodka:
@Ocramius in theory the masterslave should switch to the master automagically when ddl is executed

do you know of a way to force a specific connection in the schema tool? (this is the underlying issue I'm facing)

wodka:
ok, found a way without having to change anything.

` vendor/bin/doctrine orm:schema-tool:update --force` uses the cli-config.php and there I can change the connection:

```php
<?php
// cli-config.php
use Doctrine\DBAL\Connections\MasterSlaveConnection;
use Zend\Mvc\Application;
use Zend\Stdlib\ArrayUtils;

$appConfig = require 'application.config.php';
if (file_exists('development.config.php'))
{
	$appConfig = ArrayUtils::merge($appConfig, require 'development.config.php');
}

/** @var \Doctrine\ORM\EntityManager $em */
$em = Application::init($appConfig)->getServiceManager()->get('doctrine.entitymanager.orm_default');

$connection = $em->getConnection();
if ($connection instanceof MasterSlaveConnection)
{
	$connection->connect('master');
}

return ConsoleRunner::createHelperSet($em);
```

PR #7208, opened 2018-04-27
Better PHPDoc for AbstractQuery cache methods

someniatko:
Updated useResultCache() and setResultCacheLifetime() docs to clearly state $lifetime is measured in seconds;
Added description to useResultCache()'s params. Renamed $bool => $useCache.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -528,8 +528,8 @@ abstract class AbstractQuery
      * Set whether or not to cache the results of this query and if so, for
      * how long and which ID to use for the cache entry.
      *
-     * @param bool   $useCache Whether or not to cache the results of this query.
-     * @param int    $lifetime How long the cache entry is valid, in seconds.
+     * @param bool   $useCache      Whether or not to cache the results of this query.
+     * @param int    $lifetime      How long the cache entry is valid, in seconds.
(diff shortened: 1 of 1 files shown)


Ocramius (dismissed):
LGTM :+1: thanks @someniatko!

Ocramius:
@someniatko please check the CS failures: https://travis-ci.org/doctrine/doctrine2/jobs/371982203

someniatko:
@Ocramius  I have discussed this method (`useResultCache()`) with my colleague, and we came to conclusion it can be improved by splitting into two - one for "true" case and another for "false".

names could look like this: `enableResultCache($lifetime, $resultCacheId)` and `disableResultCache()`

 This original method could be deprecated. What do you think?

jwage (approved):

Ocramius:
> names could look like this: `enableResultCache($lifetime, $resultCacheId)` and `disableResultCache()`

@someniatko while I agree that that's a MUCH better API, this is an additional BC break. If you feel like implementing this, we'd need two patches:

 * one for `master` that introduces the BC break and documents it in `UPGRADE.md`
 * one for `2.7` that introduces a deprecation notice

Ocramius:
Meanwhile, this patch is :+1:

someniatko:
@Ocramius Thank you for your feedback, i will prepare the pull requests soon.

PR #7222, opened 2018-05-12
Add lastPage getter to Paginator

atillay:


lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -107,6 +107,16 @@ public function count()
 
         return $this->count;
     }
+    
+    /**
+     * Returns last page number
+     *
+     * @return int
+     */
+    public function lastPage()
```
fezfez: getLastPageNumber would not be more appropriate?

jwage:
This would need tests added if it were to be accepted.

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -107,6 +107,16 @@ public function count()
...
+    {
+        return ceil($this->count() / $this->query->getMaxResults());
```
Majkl578: Since `$this->query->getMaxResults()` is `int|null`, it's prone to division by zero when the value is  `null` or `0`. Needs to be checked explicitly.

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
Majkl578: Or `getPagesCount()`.
Also needs a return type.

Majkl578 (changes requested):

atillay:
@Majkl578 I made the changes :)

stof:
I would rather be -1 on this. The API of the Doctrine Paginator tool is not built around pages, but around a limit and offset (you can get 10 items with an offset of 5 for instance).
So computing a page count here looks weird.

PR #7225, opened 2018-05-16
Fixed typehint

hgiesenow:
When calling `EntityNotFoundException::fromClassNameAndIdentifier(Entity::class, ['name' => 3]);` then `phpstan -l max` issues an error
```
Doctrine\ORM\EntityNotFoundException::fromClassNameAndIdentifier()  
         expects array<string>, array<string, int> given.
```

The implementation is obviously done right as an associative array is expected instead of an index one.

### DIFF
--- a/lib/Doctrine/ORM/EntityNotFoundException.php
+++ b/lib/Doctrine/ORM/EntityNotFoundException.php
@@ -15,8 +15,8 @@ class EntityNotFoundException extends \RuntimeException implements ORMException
     /**
      * Static constructor.
      *
-     * @param string   $className
-     * @param string[] $id
+     * @param string $className
+     * @param array  $id
      *
      * @return self
      */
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/EntityNotFoundException.php
```diff
@@ -15,8 +15,8 @@ class EntityNotFoundException extends \RuntimeException implements ORMException
     /**
      * Static constructor.
      *
-     * @param string   $className
-     * @param string[] $id
+     * @param string $className
+     * @param array  $id
```
Ocramius: Should be `mixed[]` then. We don't allow just hinting something as `array` anymore, and `EntityNotFoundException` can indeed be raised with any stringable identifier.

Ocramius (changes requested):

Ocramius (approved):
LGTM :+1:

Ocramius:
Rebased and manually merged via d6ec5a3861995a8a7523d0468b1af6798d0666b3, thanks!

PR #7226, opened 2018-05-16
fixed type hint

hgiesenow:
Values in `QueryBuilder::set()` can be `mixed` as they they are only passed to Expr\Comparison which accepts `mixed`.


Ocramius:
Closing as per https://github.com/doctrine/doctrine2/pull/7226#discussion_r188622357

PR #7230, opened 2018-05-18
Mention that lifecycle callbacks do not support Embeddables

holtkamp:
As discussed in https://github.com/doctrine/doctrine2/issues/6855

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -243,6 +243,11 @@ a relevant lifecycle event. More than one callback can be defined for each
 lifecycle event. Lifecycle Callbacks are best used for simple operations
 specific to a particular entity class's lifecycle.
 
+
+.. note::
+
+    Note that Licecycle Callbacks are not supported for Embeddables.
+
 .. code-block:: php
 
     <?php


docs/en/reference/events.rst
```diff
@@ -240,6 +240,8 @@ a relevant lifecycle event. More than one callback can be defined for each
 lifecycle event. Lifecycle Callbacks are best used for simple operations
 specific to a particular entity class's lifecycle.
 
+Note that Licecycle Callbacks can not be defined for Embeddables.
+
```
jwage: If you use this syntax it will render it as a note box:

    .. note::

        Note that Licecycle Callbacks can not be defined for Embeddables.

holtkamp: Yeah, I saw it. I was not sure if it is *that* important to mention this. Anyhow, added it!

naitsirch:
@jwage ping

holtkamp:
@jwage, can we wrap this up?

SenseException:
Maybe a rebase would be a good idea. Also: Since this is from 2018, is the target branch master still the right choice? Do we need to have this in 2.7 by now?

holtkamp:
@SenseException

>Maybe a rebase would be a good idea.

Done!

>Since this is from 2018, is the target branch master still the right choice? Do we need to have this in 2.7 by now?

Well, I do not know what the rules are for this aspect. I do know it is only about one line of documentation so it can do little harm?

SenseException:
> Well, I do not know what the rules are for this aspect. I do know it is only about one line of documentation so it can do little harm?

There's AFAIK an upmerge strategy that lower branches will be merged with master. This way your contribution will appear in the other versions too.

greg0ire:
I rebased on 2.7 for you :)

greg0ire (approved):

greg0ire:
Just triggering the Travis build, I think I did things in the wrong order and the link was not established, now things look good :)

greg0ire:
Thanks @holtkamp !

PR #7235, opened 2018-05-22
Bugfix with namespaces

aashmelev:
For a schema of metadata, namespaces were not created. Because of this, there was a difference between the current schema and the schema from the metadata when you run `doctrine:migrations:diff` command.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -149,7 +149,7 @@ class SchemaTool
         $metadataSchemaConfig   = $schemaManager->createSchemaConfig();
 
         $metadataSchemaConfig->setExplicitForeignKeyIndexes(false);
-        $schema = new Schema([], [], $metadataSchemaConfig, $this->getNamespaces());
+        $schema = new Schema([], [], $metadataSchemaConfig, $this->getPlatformNamespaces());
 
         $addedFks = [];
         $blacklistedFks = [];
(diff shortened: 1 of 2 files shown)


Ocramius (changes requested):
@aashmelev this kind of patch requires a test case addition

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -899,4 +899,19 @@ public function getUpdateSchemaSql(array $classes, $saveMode = false)
 
         return $schemaDiff->toSql($this->platform);
     }
+
+    /**
+     * Gets namespaces for the current platform
+     *
+     * @return string[]
+     */
+    private function getNamespaces() {
```
Ocramius: ` : array`

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -899,4 +899,19 @@ public function getUpdateSchemaSql(array $classes, $saveMode = false)
 
         return $schemaDiff->toSql($this->platform);
     }
+
+    /**
+     * Gets namespaces for the current platform
```
Ocramius: This description can be made the method name

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -899,4 +899,19 @@ public function getUpdateSchemaSql(array $classes, $saveMode = false)
...
+     *
+     * @return string[]
+     */
+    private function getNamespaces() {
+        if (! $this->platform->supportsSchemas()) {
+            return [];
+        }
+
+        $namespaces = $this->em->getConnection()->getSchemaManager()->listNamespaceNames();
```
Ocramius: Can return directly. I suggest one `->` per line though, which aids stack trace handling

Ocramius (changes requested):

aashmelev:
Do I need to add any more tests?

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -149,7 +149,8 @@ public function getSchemaFromMetadata(array $classes)
         $metadataSchemaConfig   = $schemaManager->createSchemaConfig();
 
         $metadataSchemaConfig->setExplicitForeignKeyIndexes(false);
-        $schema = new Schema([], [], $metadataSchemaConfig);
+        $namespaces = ! $withNamespaces ? [] : $this->getPlatformNamespaces();
```
brizzz: I think 
`$namespaces = $withNamespaces ? $this->getPlatformNamespaces() : [];`
would be more clear to read.

PR #7253, opened 2018-06-08
Mention that Doctrine does not use Entities public API

JarJak:
As discussed in https://github.com/doctrine/DoctrineBundle/issues/778

### DIFF
--- a/docs/en/reference/working-with-objects.rst
+++ b/docs/en/reference/working-with-objects.rst
@@ -25,6 +25,14 @@ Work that have not yet been persisted are lost.
     Not calling ``EntityManager#flush()`` will lead to all changes
     during that request being lost.
 
+.. note::
+
+    Doctrine does NEVER touch the public API of methods in your entity 
+    classes (getters and setters). 
+    Instead, it uses reflection to set/get data from your objects.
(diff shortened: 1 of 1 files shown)


localheinz (commented):
Maybe worth mentioning that entity constructors are also never invoked?

JarJak:
@localheinz updated

docs/en/reference/working-with-objects.rst
```diff
@@ -25,6 +25,14 @@ Work that have not yet been persisted are lost.
...
     during that request being lost.
 
+.. note::
+
+    Doctrine does NEVER touch the public API of methods in your entity 
+    classes (like getters and setters) nor the constructor method.
+    Instead, it uses reflection to set/get data from your objects.
+    When Doctrine fetches data from DB and saves it back,
+    any code put in your get/set methods won't be implicitly taken into account.
+
```
jwage: Remove the extra new line here. We only need 1 new line in-between sections.

JarJak:
@jwage updated

jwage (approved):

Ocramius (approved):
:ship: :+1: :100:

PR #7260, opened 2018-06-15
Fix the handling of circular references in the commit order calculator

### LINKED ISSUE #7259: Broken commit order in 2.6
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.6

#### Summary

Commit order is wrong in some cases, leading to broken inserts (inserting without the id for the foreign key). this is dependant on the order of `persist()` calls.

#### Previous behavior

Flushing was working fine

#### Current behavior

Flush is failing by trying to insert `NULL` in a non-nullable foreign key (probably planning to update the field later).

#### How to reproduce

See #7260 which provides some tests reproducing the issue.

stof:
This fixes #7259

On 2.5, both tests are passing. On 2.6.2, the first one is failing while the second one is passing. The only difference between them is the order of `persist($fileVersion)` and `persist($file)`

### DIFF
--- a/lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
+++ b/lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
@@ -164,6 +164,17 @@ class CommitOrderCalculator
                 case self::IN_PROGRESS:
                     if (isset($adjacentVertex->dependencyList[$vertex->hash]) &&
                         $adjacentVertex->dependencyList[$vertex->hash]->weight < $edge->weight) {
+
+                        // If we have some non-visited dependencies in the in-progress dependency, we
+                        // need to visit them before adding the node.
+                        foreach ($adjacentVertex->dependencyList as $adjacentEdge) {
+                            $adjacentEdgeVertex = $this->nodeList[$adjacentEdge->to];
+
+                            if ($adjacentEdgeVertex->state === self::NOT_VISITED) {
+                                $this->visit($adjacentEdgeVertex);
+                            }
+                        }
+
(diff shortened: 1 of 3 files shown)


stof:
Note that removing the call to persist `$feed` after them (which also depends on the `Space` class involved in the circular reference) or moving them before them also removes the bug.

stof:
@guilhermeblanco I added another reproducer in a unit test at the commit order calculator level, based on what the UnitOfWork would be configuring in it in the failing functional test.

I kept the 2 functional tests I added previously, as they show high-level usage (and so would be helpful in case of a future refactoring of this part to ensure things keep working).

stof:
@guilhermeblanco the last commit in this PR is fixing the bug, with the idea I suggested in https://github.com/doctrine/doctrine2/issues/7259#issuecomment-423111105

Can you validate whether this approach makes sense ?

guilhermeblanco:
@stof That totally makes sense. It took me a while to even revive the whole operation/approach, but your solution is appropriate. Very well done! Thanks!!!

guilhermeblanco (approved):

guilhermeblanco:
I'm allowing someone else merge as I got very detached from 2.X land and don't have full context on merge process now.

stof:
@guilhermeblanco 2.5 is not maintained anymore (except maybe for security patches, but I'm not even sure). regarding 2.6 vs 2.7, I don't know what is the merge process for maintainers (but this bugfix should definitely go in the next 2.6 patch release as it fixes a regression)

PR #7263, opened 2018-06-15
Add quoted column keys while hydrating column information.

michaelpetri:
We can't hydrate column information while using a quote strategy because all or the most internal doctrine mappings use the quoted column alias name.

While hydrating a result row data we know we've only got the unquoted column alias and there is no way to quote an existing column alias.
So we need to split the column alias back to the real column name and the alias counter.
This is easy because we know it's get generated in the format '<column_name>_<counter>' by the basic entity persister.

With this information we can use the current quote strategy to generate a quoted column alias.

This PR should solve at least the following issues:

* https://github.com/doctrine/doctrine2/issues/7262
* https://github.com/doctrine/doctrine2/issues/5654

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -115,13 +115,6 @@ class SimpleObjectHydrator extends AbstractHydrator
                 throw new \Exception(sprintf('Unable to retrieve association information for column "%s"', $column));
             }
 
-            // Generate quoted column alias for column info hydrating.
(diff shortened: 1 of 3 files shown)


Ocramius:
This patch requires a failing test case first. IIRC the persister doesn't use anything from the actual column name, but rather deterministic aliases, so the test needs to be very explicit about the failure.

michaelpetri:
I removed my fix and added a test case to show you how it fails with the custom quote strategy. I also extended my current fix with also quoting column names which was required to pass my test case. I will push the fix as soon as you have reviewed the test.

michaelpetri:
Okay seems that there is more work required. I will check this soon.

Issue #7249, opened 2018-06-05
 Eliminate dependency on doctrine/common


Majkl578:
What is used through `doctrine/common` dependency:
* `Doctrine\Common\Util\Debug` - likely to be dropped, only used in one console  command and one test
* `Doctrine\Common\Event*` to be extracted to `doctrine/event-manager` (for now)
* `Doctrine\Common\Persistence` -> needs `doctrine/persistence`
* `Doctrine\Common\Util\Inflector` -> should be switched to `doctrine/inflector`
* `Doctrine\Common\Lexer` -> should be switched to `doctrine/lexer` (later removed)
* `Doctrine\Common\{PropertyChangedListener,NotifyPropertyChanged}` -> likely to be merged into ORM

cross references:
DBAL: https://github.com/doctrine/dbal/issues/3176
Migrations: https://github.com/doctrine/migrations/issues/722

Jean85:
Will this be feasible in the 2.x branch? Or are we forced to wait for 3.x due to BC?

Ocramius:
Due to BC reasons, dropping the dependency on `Doctrine\Common` will be problematic here, so it will be done in such a way that the components are no longer used, but still added as dependency.

Jean85:
So requiring the split packages and using Composer's `provide` will be a feasible upgrade path? Great!

Majkl578:
I don't think it's possible in 2.x, especially because of the Doctrine\Common\Proxy dependency.

Jean85:
Umh... Maybe we can write down a shim/polyfill that uses ProxyManager under the hood, but provides aliases for Doctrine\Common\Proxy? Or the behavior is too different?

Ocramius:
That's *a lot* of work that really isn't worth it. I already tried working towards a BC-friendly way of doing it, and it is feasible, but its at least a couple of full days of work.

Majkl578:
Handled in #7497.

PR #7274, opened 2018-06-25
Use non-deprecated version of Lexer and Inflector

Majkl578:
As of https://github.com/doctrine/common/pull/845, `Doctrine\Common\Lexer` and `Doctrine\Common\Util\Inflector` trigger silent deprecations.

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -19,7 +19,7 @@
 
 namespace Doctrine\ORM;
 
-use Doctrine\Common\Util\Inflector;
+use Doctrine\Common\Inflector\Inflector;
 use Doctrine\ORM\Query\ResultSetMappingBuilder;
 use Doctrine\Common\Persistence\ObjectRepository;
 use Doctrine\Common\Collections\Selectable;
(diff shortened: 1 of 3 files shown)


Ocramius (approved):
LGTM :+1: 

@Majkl578 any `composer.json` changes required?

Majkl578:
Not necessarily, we still depend on Common which requires both of the packages. :) Do we want them explicitly listed here?

(Will address `master` later separately.)

lcobucci (approved):

lcobucci:
> Not necessarily, we still depend on Common which requires both of the packages. :) Do we want them explicitly listed here?

@Majkl578 as long we don't plan to modify common and break everything is good HAHAHAH

Let's get it merged!

stof:
@lcobucci we should indeed add direct requirements on these packages though. but this could be done as part of https://github.com/doctrine/doctrine2/issues/7249 instead.

stof:
@lcobucci this fix also needs to be applied in master btw, not only in 2.6

Majkl578:
> we should indeed add direct requirements on these packages though

We don't need to do this now in 2.6 / 2.x imho since `doctrine/common` 2.x will always ship them.

> this fix also needs to be applied in master btw

As said above:
> (Will address master later separately.)

Master will be updated to not depend on doctrine/common, only separate packages. I'll likely cherry-pick this into #7249.
The tricky part is `PropertyChangedListener` + `NotifyPropertyChanged`, until it's decided what do do with those, we can't simply drop common _yet_. Maybe temporarily disabling the feature in master would be sufficient short-term.

PR #7276, opened 2018-06-27
Add UPGRADE note for EntityRepository::count()

### LINKED ISSUE #6933: Better document addition of EntityRepository::count()
There are multiple reports about consumers being broken due to this new method. See https://github.com/doctrine/doctrine2/pull/6003#issuecomment-353442961 and linked issues.
Although adding new public method is not a BC break per se. we should at least document it somewhere in UPGRADE as a possible source of problems.

Majkl578:
closes #6933

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,12 @@
 # Upgrade to 2.6
 
+## Added `Doctrine\ORM\EntityRepository::count()` method
+
+`Doctrine\ORM\EntityRepository::count()` has been added. This new method has different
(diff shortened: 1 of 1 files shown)


lcobucci (approved):

lcobucci:
@Majkl578 🚢 !

lcobucci:
We might probably to sync this to `master`, correct?

Majkl578:
Yesm the question is how do we do it in some organized way so we don't forget some commits. :|

lcobucci:
@Majkl578 labels?

Majkl578:
Possibly. 🥇

PR #7283, opened 2018-07-01
Add a polyfill for ctype

BackEndTea:
A cytpe function is used here: https://github.com/doctrine/doctrine2/blob/2726636e55fdbdf6fb1ed466f22d34de116fb8f9/lib/Doctrine/ORM/Query/Lexer.php#L163

But ctype is not always enabled, especially for minimalist distributions like alpine.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "doctrine/instantiator": "~1.1",
         "ocramius/package-versions": "^1.1.2",
         "ocramius/proxy-manager": "^2.1.1",
-        "symfony/console": "~3.2|~4.0"
+        "symfony/console": "~3.2|~4.0",
+        "symfony/polyfill-ctype": "~1.8"
     },
     "require-dev": {
         "ext-pdo": "*",
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "ocramius/package-versions": "^1.1.2",
         "ocramius/proxy-manager": "^2.1.1",
         "symfony/console": "~3.2|~4.0",
-        "symfony/polyfill-ctype": "~1.8"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -25,7 +25,8 @@
         "doctrine/instantiator": "~1.1",
         "ocramius/package-versions": "^1.1.2",
         "ocramius/proxy-manager": "^2.1.1",
-        "symfony/console": "~3.2|~4.0"
+        "symfony/console": "~3.2|~4.0",
+        "symfony/polyfill-ctype": "~1.8"
```
Ocramius: Probably better to require `"ext-ctype"` and have consumers include `"symfony/polyfill-ctype"` or any other polyfill that implements the `"provides"` clause in their `composer.json`

Ocramius (changes requested):

BackEndTea:
I have changed it to require ext-ctype instead

Majkl578:
Two things here:
Can we you retarget this to 2.6 branch? It's affected too.
Can you please move it just below `php`? The `require` section is `platform -> exts -> dependencies`.

Majkl578:
Also do we have any other extensions with similar "issue"?

Ocramius:
Let's keep it to `master` - the backporting is up to us or to separate PRs targeting the older branches.

Ocramius (approved):

Ocramius:
Hmm, I can't pull this into `2.6` due to the added dependency. Since it's an added dependency, it may break security updates for people not touching this code path and still running stable ORM versions.

BackEndTea:
Thank you @Ocramius & @Majkl578

PR #7285, opened 2018-07-02
Removed outdated `$entity` passed phpdoc for `EntityManager::flush()`

mkurzeja:
Removes outdated info about the possibility to pass an entity that should be flushed instead of all queued changes


Ocramius (approved):
LGTM :+1:

PR #7288, opened 2018-07-02
Fix of variable typo in cache usage example

SenseException:


Majkl578:
Should probably target 2.6? Or is this issue only in master?

SenseException:
It is in every available version, including 2.5. I guess I should have target 2.6 instead of master.

Ocramius:
@SenseException if it's a bugfix it's worth backporting, but I'm really unsure if we should start backporting also documentation issues unless the gravity is high. Reason being that the amount of work increases.

Ocramius (approved):
Shipping to `master` for now

PR #7290, opened 2018-07-02
Fix compatibility with DBAL 2.8 (doctrine/dbal#3157)

Majkl578:
Forward-compat for doctrine/dbal#3157.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
@@ -823,7 +823,14 @@ class SelectSqlGenerationTest extends OrmTestCase
             ->setMaxResults(10)
             ->setFirstResult(0);
 
-        $this->assertEquals('SELECT c0_.id AS id_0, c0_.status AS status_1, c0_.username AS username_2, c0_.name AS name_3, c0_.email_id AS email_id_4 FROM cms_users c0_ LIMIT 10 OFFSET 0', $q->getSql());
+        // DBAL 2.8+ doesn't add OFFSET part when offset is 0
+        self::assertThat(
+            $q->getSql(),
+            self::logicalOr(
+                self::identicalTo('SELECT c0_.id AS id_0, c0_.status AS status_1, c0_.username AS username_2, c0_.name AS name_3, c0_.email_id AS email_id_4 FROM cms_users c0_ LIMIT 10'),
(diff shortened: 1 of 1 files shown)


lcobucci (dismissed):
Code LGTM, I'd suggest to add more details to the commit message or add a comment in the code. It might be complicated to understand this after some time.

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
```diff
@@ -823,7 +823,13 @@ public function testLimitAndOffsetFromQueryClass()
             ->setMaxResults(10)
             ->setFirstResult(0);
 
-        $this->assertEquals('SELECT c0_.id AS id_0, c0_.status AS status_1, c0_.username AS username_2, c0_.name AS name_3, c0_.email_id AS email_id_4 FROM cms_users c0_ LIMIT 10 OFFSET 0', $q->getSql());
+        self::assertThat(
+            $q->getSql(),
```
morozov: Does it use `DBAL\QueryBuilder` internally? If yes, maybe the test could be improved by only making sure that the values passed to `Query::setMaxResults()` and `Query::setFirstResult()` are passed down to the `QueryBuilder` as is.

E.g. if instead of `getSql()` we could get the internal `QueryBuilder` and check its state, we could get rid of the dependency on SQL.

If it's not possible, then I'm fine with this solution.

morozov (dismissed):

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
Majkl578: It's not using DBAL builder, but [SQL walker](https://github.com/doctrine/doctrine2/blob/2.6/lib/Doctrine/ORM/Query/SqlWalker.php#L7).

Majkl578:
@lcobucci Added comment + reworded.

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
morozov: I see: https://github.com/doctrine/doctrine2/blob/32c125def19fb24941698c6e7e4f1e3065d28484/lib/Doctrine/ORM/Query/SqlWalker.php#L545

lcobucci: @morozov it doesn't using the query builder internally but rather a compiles a DQL and translates it into SQL using a `Query` object.

lcobucci (approved):

Majkl578:
Meh, Scrutinizer is not working properly... :S

PR #7075, opened 2018-02-17
Fix for #7068: `EntityManager::find()` with pessimistic lock should check for transaction

BreiteSeite:
Fix for #7068

I put the checks earlier in the method, that fixes the bug and is also a better practice in my opinion.

Also it is somehow hinted by the CONTRIBUTING.md guidelines:

> * Prefer early exit over nesting conditions

PR #7291, opened 2018-07-03
[2.6] Fix for #7068: EntityManager::find() with pessimistic lock should check for transaction

### LINKED ISSUE #7068: EntityManager find with pessimistic lock does not check for transaction when cached
When a pessimistic lock mode is passed to EntityManager::find it normally throws a TransactionRequiredException if no transaction is currently active.

However if the entity is currently in the identity map, it will refresh the entity from the database and return it but not check if a transaction is running. The refresh query has 'FOR UPDATE' (in case of pessimistic write) but this has no effect since there is no transaction.

To reproduce:
```php
// $em is EntityManager, Order entity can be any entity
$em->find(Order::class, 1); // Load order id 1 in entity map
$em->find(Order::class, 1, LockMode::PESSIMISTIC_WRITE); // does not throw exception
$em->find(Order::class, 2, LockMode::PESSIMISTIC_WRITE); // throws TransactionRequiredException 
```
Tested with doctrine 2.5.14 but latest code seems to have the same issue.

Majkl578:
Rebased and adjusted #7075 for 2.6.
Fixes #7068.

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -23,6 +23,7 @@ use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\LockMode;
+use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\ORM\Proxy\ProxyFactory;
 use Doctrine\ORM\Query\FilterCollection;
@@ -380,6 +381,10 @@ use Throwable;
     {
         $class = $this->metadataFactory->getMetadataFor(ltrim($entityName, '\\'));
 
+        if ($lockMode !== null) {
+            $this->checkLockRequirements($lockMode, $class);
+        }
+
         if ( ! is_array($id)) {
             if ($class->isIdentifierComposite) {
                 throw ORMInvalidArgumentException::invalidCompositeIdentifier();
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/EntityManager.php
```diff
@@ -915,4 +912,26 @@ public function hasFilters()
...
         return null !== $this->filterCollection;
     }
+
+    /**
+     * @param int $lockMode
+     * @param ClassMetadata $class
+     * @throws OptimisticLockException
+     * @throws TransactionRequiredException
+     */
+    private function checkLockRequirements(int $lockMode, ClassMetadata $class): void
```
Ocramius: Nit: we use a space between `)` and `:`. Not important though.

Ocramius (approved):

Ocramius:
Thanks @Majkl578!

lib/Doctrine/ORM/EntityManager.php
Majkl578: Hmm, leftover from original PR.

timdev:
Why the intentional fallthrough here?  It causes a BC break when using optimistic locking outside of a tx.  I'm probably missing something, but it just bit me.

Majkl578:
I think you are right, the transaction shouldn't be required for an optimistic lock ([documentation examples sagree](https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/transactions-and-concurrency.html#optimistic-locking)), at least not in 2.x. Do you want to submit a PR? :+1:

timdev:
My pleasure: https://github.com/doctrine/doctrine2/pull/7367

PR #6500, opened 2017-06-13
Allow internal functions to be overridden.

ThePixelDeveloper:
Background
------------

This came about because we're writing a driver for [PrestoDB](https://prestodb.io) and wanted to implement windowing for the aggregate functions. This is fine for anything new, but we struggled when trying to add windowing support for count, avg, etc ... 

> All Aggregate Functions can be used as window functions by adding the OVER clause. The aggregate function is computed for each row over the rows within the current row’s window frame.

Source: [6.15. Window Functions — Presto 0.178 Documentation](https://prestodb.io/docs/current/functions/window.html)

Solution
--------

Obvious solution for us was to have the ability to override the internal aggregate functions. We moved the internal aggregate functions into the `$_NUMERIC_FUNCTIONS` property so they can be overridden; They were handled as a special case before.

The only part we weren't sure of was the `dctrn` alias used: https://github.com/SamKnows/doctrine2/pull/1/files#diff-0c8825d1265a66bb8b20f3d99c276961 

If that doesn't matter and we can go with `sclr`, then this is another path that can be removed.

Thoughts? 

PS. First big PR for the Doctrine internals 👍

PR #7296, opened 2018-07-04
Fix #7286: StringPrimary no longer accepts aggregate functions as argument

### LINKED ISSUE #7286: 2.6.1 StringPrimary no longer accepts aggregate functions as argument(AVG, SUM, COUNT, MIN and MAX)
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.6.1

#### Summary

After upgrading from 2.5.14 to 2.6.1 StringPrimary no longer accepts aggregate functions as argument(AVG, SUM, COUNT, MIN and MAX).
Which affects any functions depending on StringPrimary, for example CONCAT
#### Previous behavior
`CONCAT(teable.field1, MAX(table.field2))` => '{text from field1}{maximum value of field2}'

#### Current behavior
`CONCAT(teable.field1, MAX(table.field2))` => ERROR: Expected StateFieldPathExpression | string | InputParameter | FunctionsReturningStrings | AggregateExpression, got 'MAX'

#### How to reproduce
pass any aggregate function to CONCAT (for example) as argument

#### PS

Old `StringPrimary` function looked like:
``` 
            case Lexer::T_CASE:
            case Lexer::T_COALESCE:
            case Lexer::T_NULLIF:
                return $this->CaseExpression();

            default:
                if ($this->isAggregateFunction($lookaheadType)) {
                    return $this->AggregateExpression();
                }
        }
        $this->syntaxError(
            'StateFieldPathExpression | string | InputParameter | FunctionsReturningStrings | AggregateExpression'
        );
```
while new one missing part responsible for accepting aggregate expressions:
```
            case Lexer::T_CASE:
            case Lexer::T_COALESCE:
            case Lexer::T_NULLIF:
                return $this->CaseExpression();
        }
        $this->syntaxError(
            'StateFieldPathExpression | string | InputParameter | FunctionsReturningStrings | AggregateExpression'
        );
```

Creating this as separate issue from #7205, because this concerns any function relient on StringPrimary

Majkl578:
866418e40f38e33f2c5dcf16ab9bccd3cdffa563 (#6500) removed special-casing of aggregate functions.
This was mostly transparent for cases where `isAggregateFunction()` is used, but not in StringPrimary which doesn't use `isAggregateFunction()` to detect the function ahead.
Since these functions (still) have separate tokens, they're not recognized as T_IDENTIFIER in StringPrimary, thus not handled as generic FunctionDeclaration.

Fixes #7286.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -2924,6 +2924,10 @@ class Parser
             case Lexer::T_COALESCE:
             case Lexer::T_NULLIF:
                 return $this->CaseExpression();
+            default:
+                if ($this->isAggregateFunction($lookaheadType)) {
+                    return $this->AggregateExpression();
+                }
         }
 
         $this->syntaxError(
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7286Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7286Test.php
@@ -1,0 +1,135 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Query\AST\Functions\FunctionNode;
+use Doctrine\ORM\Query\AST\Node;
+use Doctrine\ORM\Query\Lexer;
+use Doctrine\ORM\Query\Parser;
+use Doctrine\ORM\Query\SqlWalker;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH7286Test extends OrmFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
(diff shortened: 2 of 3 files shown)


lcobucci (commented):
@Majkl578 this seems to be a clean patch but I'm wondering if everything will still work if we use an aggregation function (or any function) as argument to a custom function - which was what removed the code before 🤔

lcobucci:
@Ocramius any thoughts here?

Majkl578:
Bah, fails on Postgres, would have to test with pre-2.6 to see what worked or not.

lcobucci:
@Majkl578 that's exactly what I meant, thanks for adding the test 👍

lcobucci (approved):

lcobucci:
@Majkl578 can you please port it to 3.0-dev?

PR #7300, opened 2018-07-06
Fix SchemaTool to use executeUpdate

mcfedr:
executeUpdate should be used for write queries. Specifically if a master slave connection is used it wont switch to the master unless executeUpdate is used.


Majkl578:
Hi,

could you please try to create a functional test case that reproduces the issue?
You can find existing examples in https://github.com/doctrine/doctrine2/tree/v2.6.2/tests/Doctrine/Tests/ORM/Functional/Ticket.

Thanks.

PR #7305, opened 2018-07-11
Make the iterate() function to respect its parent prototype

sandvige:
The iterate() method defined in Query.php is not respecting the defined prototype in AbstractQuery.php with the second default parameter. This makes not possible to define a specific hydration mode when using the iterate() method default behavior with setHydrationMode() because the setter is called only if the given value is !== null.

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -658,7 +658,7 @@ final class Query extends AbstractQuery
      *
      * @return IterableResult
      */
-    public function iterate($parameters = null, $hydrationMode = self::HYDRATE_OBJECT)
+    public function iterate($parameters = null, $hydrationMode = null)
     {
         $this->setHint(self::HINT_INTERNAL_ITERATION, true);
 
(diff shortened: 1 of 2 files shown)


sandvige:
It looks like the errors on travis are NOT related at all with my changes, any hint?

Majkl578:
Apparently, build restarted.

sandvige:
@jwage do you think that kind of test in QueryTest.php is enough:
``` php
    public function testIterateWithCustomHydrationMode()
    {
        $iterable = $this
            ->_em
            ->createQuery("SELECT u FROM Doctrine\Tests\Models\CMS\CmsUser u")
            ->setHydrationMode(Query::HYDRATE_ARRAY)
            ->iterate()
        ;

        $element = null;
        foreach ($iterable as $element)
            break;

        // Assert that we have at least one element found through the foreach
        $this->assertNotNull($element);

        // Assert the fetched element is NOT an entity as an Query::HYDRATE_ARRAY have been requested
        $this->assertNotInstanceOf('Doctrine\Tests\Models\CMS\CmsUser', $element[0]);
    }
```

sandvige:
@jwage Done

LordWeedlle (approved):
Sounds a good idea =)

slgeay (approved):

sandvige:
Should I back port this PR to other branches?

Ocramius:
No need: we'll merge it back manually

PR #7203, opened 2018-04-18
onFlush and postFlush Entity Listeners

alexgurrola:
Simply check if an onFlush or postFlush entity listener is subscribed
for the current class in the entity manager during dispatch.

This is a followup to merge and unit testing issues in #6414.  This branch only contains the feature that was submitted previously and does not contain any unit testing.  I can attempt to rebuild some unit tests when I have extra time.

PR #7310, opened 2018-07-13
onFlush and postFlush Entity Listeners

alexgurrola:
Simply check if an onFlush or postFlush entity listener is subscribed
for the current class in the entity manager during dispatch.

----------------

This contains Unit Tests per request by @Ocramius.
This references the same code as PR #7203 in the [2.7](https://github.com/doctrine/doctrine2/tree/2.7) branch.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -3382,11 +3382,31 @@ class UnitOfWork implements PropertyChangedListener
         }
     }
 
+    private function dispatchFlushEntityListener($event)
+    {
+        foreach ($this->identityMap as $className => $entities) {
+            $class = $this->em->getClassMetadata($className);
+            $invoke = $this->listenersInvoker->getSubscribedSystems($class, $event) & ~ListenersInvoker::INVOKE_MANAGER;
+            if (ListenersInvoker::INVOKE_NONE === $invoke) {
+                continue;
+            }
+            foreach ($entities as $entity) {
+                $oid = spl_object_hash($entity);
+                if (! isset($this->entityChangeSets[$oid])) {
+                    continue;
+                }
+                $argClass = 'Doctrine\\ORM\\Event\\'.ucfirst($event).'EventArgs';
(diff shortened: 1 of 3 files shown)


ostrolucky:
New functionality must go to master branch (doctrine 3.0). Instead of creating new PR (again), try learning how to do rebases https://stackoverflow.com/questions/36967600/how-to-rebase-last-x-commits-on-different-branch, rebase these commits onto master and push to same branch.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3382,18 +3382,39 @@ private function performCallbackOnCachedPersister(callable $callback)
         }
     }
 
+    private function dispatchFlushEntityListener($event)
+    {
+        foreach ($this->identityMap as $className => $entities) {
+            $class = $this->em->getClassMetadata($className);
+            $invoke = $this->listenersInvoker->getSubscribedSystems($class, $event) & ~ListenersInvoker::INVOKE_MANAGER;
+            if (ListenersInvoker::INVOKE_NONE === $invoke) {
+                continue;
```
alexgurrola: This is an entity listener, and if there aren't any subscribed, it will just `continue`.

Ocramius:
Release it as an OSS package?

ostrolucky:
> I've also been using this logic for more than a year in my projects

Seems that does not guaranteed code is allright, since you dispatch event for every entity in identity map, instead of entities which were flushed. This means your patch is essentially broken anyway. See link above to see what actually needs to be done. It's also a proof you don't need this to be built into Doctrine. Going from project to project with your own copy of vendor lib, instead of having package for it is pretty shortsighted.

alexgurrola:
It only dispatches for entities that have a subscribed listener, and skips the classes that do not.  If you'd like me to change the loop to something other than the identityMap that only contains the flushed entities, I could see how that would be better.

Do you have a line that I could reference so I can look for a better route to create this feature in the `UnitOfWork`?  The code for this particular file contains many code paths, so trying to comprehend how to make this feature fit better has proven difficult for me.

I do appreciate your help in this endeavor.  Thanks.

ostrolucky:
See referenced issue

PR #7317, opened 2018-07-19
[XML] Fix default value of many-to-many order-by to ASC

### LINKED ISSUE #7316: XML many-to-many order-by default isn't set
Similar to #7141 - but #7141 didn't cover the fact that the issue also affected many-to-many relationships.

The fix provided in #7146 works for the one-to-many relationships, so I am opening this issue (with PR to follow shortly) to apply the same fix for many-to-many

alexdenvir:
Fixes #7316 

Props to @Awkan for the original fix applied to one-to-many relationships - this just applied the same fix to many-to-many


Ocramius:
Does this also apply to `master`?

alexdenvir:
I haven't specifically tested with master, but a quick read of the code suggests it does. It looks like the one-to-many fix hasn't been applied to master either

lcobucci (approved):
@aaronmu :+1:

PR #7318, opened 2018-07-20
Original entity data is missing id when computing change set for already managed entity

supersmile2009:
I'm targeting 2.6 branch because it's a bugfix.

The bug happens when you persist again a managed entity.
- if entity has auto-generated id, original entity data will be missing id after computing change set.
- if entity owns a collection, that collection will be also missing

As you can see from the test case - steps to reproduce are pretty simple.

**For id:**
1. Have an entity with auto-generated id. This entity should be previously persisted (has an id, is managed).
2. Change any mapped property in the object.
3. Persist and commit changed entity or call `UnitOfWork::computeSingleEntityChangeSet()` or `UnitOfWork::recomputeSingleEntityChangeSet()` directly.
4. Result: original entity data for that object contains object data including the changes from step 2, but is missing the id.

**For collections:**
1. Have an entity A which contains collection of B.
2. Persist A, ArrayCollection becomes PersistentCollection. Original entity data is correct.
3. Change any mapped property in A object.
4. Persist it again.
5. Collection is missing from original entity data.

Id is missing because when populating `$actualData` array, auto-generated id value is ignored. It makes sense, because user shouldn't be able to set it manually.
I made a simple fix, which preserves id contained in original entity data when overwriting original data with `$actualData`.


Majkl578:
Hi, thanks for the PR. Can you please add a functional test (in [`tests/Doctrine/Tests/ORM/Functional/Ticket/`](https://github.com/doctrine/doctrine2/tree/2.6/tests/Doctrine/Tests/ORM/Functional/Ticket)) that covers this scenario, failing before the change and passing afterwards?
Also please avoid introducing unrelated CS changes (like `\in_array()`).

supersmile2009:
Hi @Majkl578 
I moved the tests into `tests/Doctrine/Tests/ORM/Functional/Ticket/` and cleaned up unrelated changes.

I did this fix some time ago and totally forgot that this bug affects collections as well, so I also added a third test case for collections. And I'll update the PR description with the details.

Webonaute:
hi. @supersmile2009 is in vacation. I am doing a followup, anything else need from our team to push this fix? thank.

lcobucci:
@supersmile2009 @Webonaute @Ocramius I've rebased this branch to sync with `v2.6` but I'm not being able to understand exactly why you think that the current behaviour is wrong.

You see, `UnitOfWork#originalEntityData` is used to calculate the changeset in managed entities in order to update the tables accordingly. This means that having data that should **never** change there is invalid IMHO. Since associations' data also don't affect the changeset calculation, that should also **never** be there.

Could you please give us a better explanation on why you want to have that data in the UoW? From a very first glance this looks like you're trying to use it to do things that it wasn't designed for...

lcobucci:
For now I'll move this to `v2.6.4`...

supersmile2009:
@lcobucci
>This means that having data that should never change there is invalid IMHO
In that case current behaviour is invalid because `originalEntityData` contains identifier and reference to any related collections when you initially fetch an entity.
So the main issue here is inconsistent (therefore unreliable) behaviour. If you say that data which never changes (like id) should not be present there, then it should never be added to `originalEntityData`. But I believe such change can only be done only in the next major version, otherwise it's going to be a breaking change.

The practical reason why I had to do this patch was implementing entity audit with https://github.com/simplethings/EntityAuditBundle
It relies heavily on `originalEntityData`. If you
- have some listener in your app which re-computes changeset
- or flush changes multiple times
You'll get all kinds of exceptions, because audit listener won't find the data it expects to have in `originalEntityData`.

So my goal here is to make `originalEntityData` consistent. And the most backwards-compatible way to accomplish that is to maintain it in a similar state to what you get when you initially fetch an entity from DB.

lcobucci:
@supersmile2009 moving to 2.7.1 as we're planning to have only very critical things in 2.6.5 and release 2.7.0 ASAP. Sorry for keep pushing this back but this needs attention and I want to give your work the proper love.

tavlima:
The same is true for the `versionField`: it is also removed from original data map after computing the change set.

Issue #6075, opened 2016-10-07
Remove of joined entities not possible for custom id types


hcomnetworkers:
The delete-calls on the connection in BasicEntityPersister::deleteJoinTableRecords do not include the types for the fields:

https://github.com/doctrine/doctrine2/blob/cd11723e6310b5973f50e7d9819c58916721d290/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L540

A joined entity with a custom id type can therefor not be removed because the Type::convertToDatabaseValue method is never called to transform the id into its database value.

lcobucci:
Already reported in https://github.com/doctrine/orm/issues/5988

Issue #5988, opened 2016-08-23
JoinedSubclassPersister::delete does not pass $types argument to Connection::delete


fred-jan:
My entity uses both class table inheritance and a custom id type (`uuid_binary` type provided by the `ramsey/uuid-doctrine` package). Simplified example:

```
/**
 * @ORM\Entity
 * @ORM\Table(name="my_entity")
 * @ORM\InheritanceType("JOINED")
 * @ORM\DiscriminatorColumn(name="type", type="string")
 * @ORM\DiscriminatorMap({
 *     "a" = "MyEntityA",
 *     "b" = "MyEntityB"
 * })
 */
abstract class MyEntity
{
    /**
     * @ORM\Id
     * @ORM\Column(type="uuid_binary")
     * @ORM\GeneratedValue(strategy="NONE")
     *
     * @var UuidInterface
     */
    private $id;
}
```

When I attempt to delete such an entity using the `EntityManager`, the `JoinedSubclassPersister` class is leveraged internally to perform the deletion and calls `Connection::delete` without passing along any field-mapping information as `$types` argument. This results in the following query being executed:

```
DELETE FROM my_entity WHERE id = 'Object(Ramsey\\Uuid\\Uuid)';
```

For my specific platform (which supports FK's) I've made a temporary fix by copying the `$types` assignment code from the `BasicEntityPersister` class, which looks like this:

```
        // ...
        // If the database platform supports FKs, just
        // delete the row from the root table. Cascades do the rest.
        if ($this->platform->supportsForeignKeyConstraints()) {
            $rootClass  = $this->em->getClassMetadata($this->class->rootEntityName);
            $rootTable  = $this->quoteStrategy->getTableName($rootClass, $this->platform);
            $types      = array_map(function ($identifier) use ($rootClass) {
                if (isset($rootClass->fieldMappings[$identifier])) {
                    return $rootClass->fieldMappings[$identifier]['type'];
                }

                $targetMapping = $this->em->getClassMetadata($rootClass->associationMappings[$identifier]['targetEntity']);

                if (isset($targetMapping->fieldMappings[$targetMapping->identifier[0]])) {
                    return $targetMapping->fieldMappings[$targetMapping->identifier[0]]['type'];
                }

                if (isset($targetMapping->associationMappings[$targetMapping->identifier[0]])) {
                    return $targetMapping->associationMappings[$targetMapping->identifier[0]]['type'];
                }

[...]

PR #7322, opened 2018-07-24
JoinedSubclassPersister pass identifier types on delete

dennisenderink:
Had trouble deleting entities that implemented the class inheritance strategy in combination with [ramsey' uuid custom type](https://github.com/ramsey/uuid-doctrine). Turned out the identifier types weren't passed to the connection delete method in the JoinedSubclassPersister class.

It seems I wasn't the only one hence #6075 and #5988 so here is a fix.

This is my first PR into a public repository so hopefully I've done everything according the rules. If not, let me know!


lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -2089,4 +2072,33 @@ protected function switchPersisterContext($offset, $limit)
...
+     * Retrieves class identifiers types
+     *
+     * @param ClassMetadata $class
+     * @return array
+     */
+    protected function getClassIdentifiersTypes(ClassMetadata $class)
+    {
+        $self = $this;
+
+        return array_map(function ($identifier) use ($class, $self) {
```
Ocramius: The `$self` trick should no longer be needed, since we don't support older PHP versions anyway

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -2089,4 +2072,33 @@ protected function switchPersisterContext($offset, $limit)
...
         $this->currentPersisterContext = $this->limitsHandlingContext;
     }
+
+    /**
+     * Retrieves class identifiers types
+     *
+     * @param ClassMetadata $class
+     * @return array
+     */
+    protected function getClassIdentifiersTypes(ClassMetadata $class)
```
Ocramius: Please make this `private` for new methods. Add a return type too

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -2089,4 +2072,33 @@ protected function switchPersisterContext($offset, $limit)
 
         $this->currentPersisterContext = $this->limitsHandlingContext;
     }
+
+    /**
+     * Retrieves class identifiers types
+     *
+     * @param ClassMetadata $class
+     * @return array
```
Ocramius: Please document the returned type in more detail - is this a `Type[]`? `string[]`?

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
Ocramius: As far as I can see, this was just moved out, right?

tests/Doctrine/Tests/DbalTypes/CustomIdObjectType.php
```diff
@@ -14,15 +14,15 @@ class CustomIdObjectType extends Type
      */
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
-        return $value->id;
+        return $value->id . '_test';
```
Ocramius: Please use new test assets for your tests: modifying existing test assets is extremely risky.

Ocramius (changes requested):
Patch and approach are good - will need to improve some details, but overall the approach is correct :+1:

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
dennisenderink: I made it `protected` so that the `JoinedSubclassPersister` could make use of it too. I'm not sure if it was moved out, I added it so that it can be used repeatedly.

Ocramius: Hmm, I'm not sure about using inheritance for this. It makes sense, as long as the component stays internal and documented as such.

Looking at what the logic actually does, this seems like something that could live on its own though.

Ocramius: @guilhermeblanco if you have time, can you check this and see what would be best (considering ORM 3.x changes)?

dennisenderink:
Updated patch as requested!

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
dennisenderink: OK, @guilhermeblanco, let me know if you want me to change this bit!

oleg-andreyev:
Today I've caught same issue (v2.5.14)

dennisenderink:
@Ocramius @guilhermeblanco what can I do to get this fix merged? (will update code with upstream soon)

Ocramius (dismissed):
LGTM 👍

Ocramius:
Note to self: tests are to be ported to `master` too.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -2089,4 +2072,31 @@ protected function switchPersisterContext($offset, $limit)
 
         $this->currentPersisterContext = $this->limitsHandlingContext;
     }
+
+    /**
+     * Retrieves class identifiers types
+     *
+     * @param ClassMetadata $class
+     * @return string[]
```
Majkl578: This doesn't seem to be accurate. When I changed the array_map as such:
```php
function (string $identifier) use ($class) : string {
```
I got test failures coming from Doctrine\Tests\ORM\Functional\Ticket\DDC2579Test.

Can you please investigate? Thanks. :)

Majkl578 (changes requested):

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
dennisenderink: I'll look into it!

lcobucci:
I've move this to `v2.6.4` since we still have some investigation to be done

r3m4k3:
@Ocramius @Majkl578 Any update on this?

SenseException:
@dennisenderink Anything that needs to be done from your side or is this ready for another review?

dennisenderink:
> @Ocramius @Majkl578 Any update on this?



> @dennisenderink Anything that needs to be done from your side or is this ready for another review?

I just came home after a few months of traveling, didn't have the time to check @Majkl578 feedback before that. I will have some time later this week!

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
dennisenderink: I updated the return types!

lukikrk:
@dennisenderink @Ocramius @Majkl578 
Is something already known?

kubapyla:
@dennisenderink @Ocramius @Majkl578 @lukikrk 
Are there any new updates on this?

dennisenderink:
> @dennisenderink @Ocramius @Majkl578 @lukikrk
> Are there any new updates on this?

I updated the code last week but it doesn't seem to pass the tests because it is unable to install MySQL, which I think has nothing to do with the changes.

Will this be fixed in a separate PR?

r3m4k3:
@Ocramius @Majkl578 
What can we do with this issue?
As I understand @dennisenderink fixed the issue, but there's an error on Travis CI, could you please help us to sort this out?

dennisenderink:
> @Ocramius @Majkl578
> What can we do with this issue?
> As I understand @dennisenderink fixed the issue, but there's an error on Travis CI, could you please help us to sort this out?

That's correct!

SenseException:
The builds using `DB=mysql MYSQL_VERSION=5.7` still show the following error after I restarted the build:

```
WARNING: The following packages cannot be authenticated!
  mysql-common libmysqlclient20 mysql-community-client mysql-client
  mysql-community-server libmysqlclient-dev mysql-server
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
The command "./tests/travis/install-mysql-$MYSQL_VERSION.sh" failed and exited with 100 during .
```

lukikrk:
Could you recommend any workaround to solve this problem?

lukikrk:
Detected the same situation during select with join operation.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -2089,4 +2072,31 @@ protected function switchPersisterContext($offset, $limit)
...
+
+    /**
+     * Retrieves class identifiers types
+     *
+     * @param ClassMetadata $class
+     * @return int[]|string[]
+     */
+    protected function getClassIdentifiersTypes(ClassMetadata $class): array
+    {
+        return array_map(function($identifier) use ($class) {
```
lcobucci: This looks like a good candidate for the `PersisterHelper::getgetTypeOfField()`

tests/Doctrine/Tests/ORM/Functional/ClassTableInheritanceCustomIdObjectTypeTest.php
```diff
@@ -0,0 +1,50 @@
...
+{
+    protected function setUp()
+    {
+        if (DBALType::hasType(CustomIdObjectHashType::NAME)) {
+            DBALType::overrideType(CustomIdObjectHashType::NAME, CustomIdObjectHashType::class);
+        } else {
+            DBALType::addType(CustomIdObjectHashType::NAME, CustomIdObjectHashType::class);
+        }
+
+        $this->useModelSet('class_table_inheritance_custom_id_object_type');
```
lcobucci: We should put all the classes here and not rely on the model set.

lcobucci (changes requested):
I'll be applying these changes :+1:

lcobucci (approved):
@dennisenderink thanks a lot for your contribution (and patience)!

dennisenderink:
> @dennisenderink thanks a lot for your contribution (and patience)!

No worries, glad I could help! Thanks for the changes, approval and merge! 👍

Issue #7001, opened 2018-01-23
Doctrine 3: Auto-escape column names


weaverryan:
Hi guys!

In #5874, Ocramius mentioned that the auto-escaping of column names for reserved keywords couldn't happen for Doctrine 2, but it *could* happen in Doctrine 3. I couldn't find any issue in the Doctrine 3 milestone, so I wanted to make sure the idea didn't get lost: I just saw a beginner hit this and get confused today (so 3.0 is our chance).

Is this still a "go" (i.e. yes, something we at least *can* do, assuming someone does the work) for Doctrine 3?

Thanks!

Ocramius:
Yeah, I think this should happen for 3.x.

We will have bugs with people doing mixed windows/mac/linux deployments, but less problems with reserved keywords used in mappings, and we can also get rid of the awkward "quoting in mappings" syntax.

ostrolucky:
This already happened in master

PowerKiKi:
This has been implemented in 08e2f779e3881f97c8136129fe9f951a9de2cb0d (and following commits) and the issue should closed.

GregOriol:
I'm maybe missing something but with a column definition as `#[Column(type: Types::BOOLEAN, name: 'manual')` and MySQL 8.4 which has "manual" as reserved keyword, an INSERT request made through the ORM fails on the word. It works by escaping it in the name definition, but shouldn't it be automatic since it belongs to the known reserved keywords?

Issue #6883, opened 2017-12-12
xml-mapping: custom-id-generator class not a valid value for nmtoken


IMM0rtalis:
Hello together,

```xml
<id name="id" column="id" type="uuid">
    <generator strategy="CUSTOM"/>
    <custom-id-generator class="Ramsey\Uuid\Doctrine\UuidGenerator"/>
</id>
```
Gives `'Ramsey\Uuid\Doctrine\UuidGenerator' is not a valid value for 'NMTOKEN'`

Somebody reported it here already: https://groups.google.com/forum/#!topic/doctrine-user/q9aHfy5l30M

But no update so far. I wanted to try to make a PR but i am struggling a bit with creating the test-case or more like "how to create an sample-xml-mapping-file" for this case.

All the best,
Florian

PR #7324, opened 2018-07-26
Fix XSD regex for fqcn and tablename patterns

guilliamxavier:
Follow-up to #6389's discussion. Mainly: "`[u01-uff]`" is just *wrong* (in any regex flavor, BTW).

Recap of the currently incorrect cases _(and inconsistencies)_:

- `tablename`:
    - rejects `` `foo-bar` `` [does not allow the hyphen even if quoted] _(while accepting `` `foo;bar` ``)_
    - accepts `foo;bar` [allows the semicolon although unquoted] _(while rejecting `foo-bar`)_
- `fqcn`:
    - rejects `C` [requires at least 2 chars]
    - accepts `Foo;Bar` [allows the semicolon] _(while rejecting `Foo-Bar`)_
    - accepts `3lephant` [allows digits for the first char]

---

Note: the pattern for `fqcn` is rather straightforward, but not so for `tablename`:
- different SQL database vendors allow different extra characters for [un]quoted identifiers...
- it seems that Doctrine supports the form "`myschema.mytable`", where `mytable` *may* be quoted whereas `myschema` may *not* but *will become* if `mytable` is ([here](https://github.com/doctrine/doctrine2/blob/96f166a7e90149adb1357c5db73c615407ea394e/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php#L80-L91) -> [here](https://github.com/doctrine/doctrine2/blob/96f166a7e90149adb1357c5db73c615407ea394e/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L2297-L2309), then [here](https://github.com/doctrine/doctrine2/blob/96f166a7e90149adb1357c5db73c615407ea394e/lib/Doctrine/ORM/Mapping/DefaultQuoteStrategy.php#L49-L61) -> [there](https://github.com/doctrine/dbal/blob/a7dc38d87bc5b89704b7c3388b1438701a4dfb0a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L1922-L1925))...
- cannot be merged into master *as-is*, because Doctrine 3.x will auto-quote everything...

Also, this adds many small XML test files (mostly because invalid cases have to be tested separately [and I split the valid cases a bit], but not all are strictly necessary)...

(But still, I think that at least those "`u01-uff`" should be fixed.)

Found related:
- #6728
- #6883
- #7001
- [#7085's second commit](https://github.com/doctrine/doctrine2/pull/7085/commits/0b04aa900063f3afe5a11e9697de01dd7c5a3987)

### DIFF
--- a/doctrine-mapping.xsd
+++ b/doctrine-mapping.xsd
@@ -424,7 +424,7 @@
 
   <xs:simpleType name="fqcn" id="fqcn">
     <xs:restriction base="xs:token">
-      <xs:pattern value="[a-zA-Z_u01-uff][a-zA-Z0-9_u01-uff]+" id="fqcn.pattern">
+      <xs:pattern value="([a-zA-Z_&#x7F;][a-zA-Z0-9_&#x7F;]*\\)*[a-zA-Z_&#x7F;][a-zA-Z0-9_&#x7F;]*" id="fqcn.pattern">
       </xs:pattern>
     </xs:restriction>
   </xs:simpleType>
--- a/doctrine-mapping.xsd
+++ b/doctrine-mapping.xsd
@@ -202,7 +202,7 @@
 
   <xs:simpleType name="tablename" id="tablename">
     <xs:restriction base="xs:token">
-      <xs:pattern value="[a-zA-Z_u01-uff.]+" id="tablename.pattern">
+      <xs:pattern value="([&#x20;-&#x7F;-[`.]]+\.)?([a-zA-Z][a-zA-Z0-9_]*|`[&#x20;-&#x7F;]+`)" id="tablename.pattern">
       </xs:pattern>
     </xs:restriction>
   </xs:simpleType>
--- a/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
@@ -275,6 +275,7 @@ class XmlMappingDriverTest extends AbstractMappingDriverTest
 
         foreach ([
             'fqcn' => ['custom-id-generator', 'class'],
+            'tablename' => ['entity', 'table'],
         ] as $type => [$element, $attribute]) {
             $errorMessagePrefix = sprintf("Element '%s', attribute '%s': ", $namespaced($element), $attribute);
 
--- a/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
@@ -206,7 +206,15 @@ class XmlMappingDriverTest extends AbstractMappingDriverTest
         $invalid = self::getInvalidXmlMappingMap();
 
         $list = array_filter($list, function($item) use ($invalid){
-            return ! array_key_exists(pathinfo($item, PATHINFO_FILENAME), $invalid);
+            $matchesInvalid = false;
+            foreach ($invalid as $filenamePattern => $unused) {
+                if (fnmatch($filenamePattern, pathinfo($item, PATHINFO_FILENAME))) {
+                    $matchesInvalid = true;
+                    break;
+                }
+            }
+
+            return ! $matchesInvalid;
         });
 
         return array_map(function($item){
@@ -220,19 +228,17 @@ class XmlMappingDriverTest extends AbstractMappingDriverTest
         $invalid = self::getInvalidXmlMappingMap();
 
         $map = [];
-        foreach ($invalid as $filename => $errorMessageRegexes) {
-            $foundItem = null;
-            foreach ($list as $item) {
-                if (pathinfo($item, PATHINFO_FILENAME) === $filename) {
-                    $foundItem = $item;
-                    break;
-                }
-            }
+        foreach ($invalid as $filenamePattern => $errorMessageRegexes) {
+            $foundItems = array_filter($list, function($item) use ($filenamePattern){
+                return fnmatch($filenamePattern, pathinfo($item, PATHINFO_FILENAME));
+            });
 
-            if ($foundItem !== null) {
-                $map[$foundItem] = $errorMessageRegexes;
+            if (count($foundItems) > 0) {
+                foreach ($foundItems as $foundItem) {
+                    $map[$foundItem] = $errorMessageRegexes;
+                }
             } else {
-                throw new \RuntimeException(sprintf('Found no XML mapping with filename "%s".', $filename));
+                throw new \RuntimeException(sprintf('Found no XML mapping with filename pattern "%s".', $filenamePattern));
             }
         }
 
@@ -250,15 +256,30 @@ class XmlMappingDriverTest extends AbstractMappingDriverTest
     }
 
     /**
-     * @return array<string, string[]> ($filename => $errorMessageRegexes)
+     * @return array<string, string[]> ($filenamePattern => $errorMessageRegexes)
      */
     static private function getInvalidXmlMappingMap()
     {
+        $namespaced = function ($name) {
+            return sprintf('{%s}%s', 'http://doctrine-project.org/schemas/orm/doctrine-mapping', $name);
+        };
+
         $invalid = [
             'Doctrine.Tests.Models.DDC889.DDC889Class.dcm' => [
-                "Element '{http://doctrine-project.org/schemas/orm/doctrine-mapping}class': This element is not expected. Expected is %s.",
+                sprintf("Element '%s': This element is not expected. Expected is %%s.", $namespaced('class')),
             ],
         ];
+
+        foreach ([
+            'fqcn' => ['custom-id-generator', 'class'],
+        ] as $type => [$element, $attribute]) {
+            $errorMessagePrefix = sprintf("Element '%s', attribute '%s': ", $namespaced($element), $attribute);
+
+            $invalid[sprintf('pattern-%s-invalid-*', $type)] = [
+                $errorMessagePrefix . "[facet 'pattern'] The value '%s' is not accepted by the pattern '%s'.",
+                $errorMessagePrefix . sprintf("'%%s' is not a valid value of the atomic type '%s'.", $namespaced($type)),
+            ];
+        }
 
         // Convert basic sprintf-style formats to PCRE patterns
         return array_map(function ($errorMessageFormats) {
--- a/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
@@ -205,17 +205,17 @@ class XmlMappingDriverTest extends AbstractMappingDriverTest
         $list    = self::getAllXmlMappingPaths();
         $invalid = self::getInvalidXmlMappingMap();
 
-        $list = array_filter($list, function($item) use ($invalid){
+        $list = array_filter($list, function($filename) use ($invalid){
             $matchesInvalid = false;
             foreach ($invalid as $filenamePattern => $unused) {
-                if (fnmatch($filenamePattern, pathinfo($item, PATHINFO_FILENAME))) {
+                if (fnmatch($filenamePattern, $filename)) {
                     $matchesInvalid = true;
                     break;
                 }
             }
 
             return ! $matchesInvalid;
-        });
+        }, ARRAY_FILTER_USE_KEY);
 
         return array_map(function($item){
             return [$item];
@@ -229,30 +229,33 @@ class XmlMappingDriverTest extends AbstractMappingDriverTest
 
         $map = [];
         foreach ($invalid as $filenamePattern => $errorMessageRegexes) {
-            $foundItems = array_filter($list, function($item) use ($filenamePattern){
-                return fnmatch($filenamePattern, pathinfo($item, PATHINFO_FILENAME));
-            });
+            $foundItems = array_filter($list, function($filename) use ($filenamePattern){
+                return fnmatch($filenamePattern, $filename);
+            }, ARRAY_FILTER_USE_KEY);
 
             if (count($foundItems) > 0) {
-                foreach ($foundItems as $foundItem) {
-                    $map[$foundItem] = $errorMessageRegexes;
+                foreach ($foundItems as $filename => $foundItem) {
+                    $map[$filename] = [$foundItem, $errorMessageRegexes];
                 }
             } else {
                 throw new \RuntimeException(sprintf('Found no XML mapping with filename pattern "%s".', $filenamePattern));
             }
         }
 
-        return array_map(function($item, $errorMessageRegexes){
-            return [$item, $errorMessageRegexes];
-        }, array_keys($map), $map);
+        return $map;
     }
 
     /**
-     * @return string[]
+     * @return array<string, string> ($filename => $path)
      */
     static private function getAllXmlMappingPaths()
     {
-        return glob(__DIR__ . '/xml/*.xml');
+        $list = [];
+        foreach (glob(__DIR__ . '/xml/*.xml') as $path) {
+            $list[pathinfo($path, PATHINFO_FILENAME)] = $path;
+        }
+
+        return $list;
     }
 
     /**
--- a/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
@@ -162,7 +162,7 @@ class XmlMappingDriverTest extends AbstractMappingDriverTest
      * @dataProvider dataValidSchemaInvalidMappings
      * @group 6389
      */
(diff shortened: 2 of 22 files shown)


Ocramius:
@guilliamxavier 2.5 is closed for bugfixing - if this applies to 2.6 or newer we can likely merge it, but 2.5 only receives security fixes.

doctrine-mapping.xsd
```diff
@@ -424,7 +424,7 @@
 
   <xs:simpleType name="fqcn" id="fqcn">
     <xs:restriction base="xs:token">
-      <xs:pattern value="[a-zA-Z_u01-uff][a-zA-Z0-9_u01-uff]+" id="fqcn.pattern">
+      <xs:pattern value="([a-zA-Z_&#x7F;-&#x10FFFF;][a-zA-Z0-9_&#x7F;-&#x10FFFF;]*\\)*[a-zA-Z_&#x7F;-&#x10FFFF;][a-zA-Z0-9_&#x7F;-&#x10FFFF;]*" id="fqcn.pattern">
```
Ocramius: Class names must follow `^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+$` (https://github.com/Ocramius/ProxyManager/blob/99542a2c0fef18e0b460cbed93c5cde63cc2be91/tests/ProxyManagerTest/Generator/Util/UniqueIdentifierGeneratorTest.php#L42)

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -149,27 +149,69 @@ public function testInvalidMappingFileException()
...
+    {
+        $this->assertTrue($this->doValidateXmlSchema($xmlMappingFile));
+    }
+
+    /**
+     * @param string $xmlMappingFile
+     * @dataProvider dataInvalidSchema
+     * @group 6389
+     */
+    public function testInvalidateXmlSchema($xmlMappingFile)
```
Ocramius: ` : void`

Add type declarations too (can remove them in docblock)

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -149,27 +149,69 @@ public function testInvalidMappingFileException()
...
+    {
+        // Avoid PHP Warning errors for expected invalid mappings
```
Ocramius: What errors did come up here?

Ocramius: Asking because this part should probably be moved into the XML driver itself

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -149,27 +149,69 @@ public function testInvalidMappingFileException()
...
+        try {
+            $this->assertFalse($this->doValidateXmlSchema($xmlMappingFile), 'Invalid XML mapping should not pass XSD validation.');
+        } finally {
+            // Restore previous setting
+            libxml_clear_errors();
+            libxml_use_internal_errors($savedUseErrors);
+        }
+    }
+
+    private function doValidateXmlSchema($xmlMappingFile)
```
Ocramius: Type declarations

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -149,27 +149,69 @@ public function testInvalidMappingFileException()
...
+        return $dom->schemaValidate($xsdSchemaFile);
     }
 
     static public function dataValidSchema()
     {
-        $list    = glob(__DIR__ . '/xml/*.xml');
+        return self::doDataValidSchema(true);
+    }
+
+    static public function dataInvalidSchema()
```
Ocramius: This sort of API is quite confusing: what's going on? Why is `doDataValidSchema()` used also for invalid data?

Ocramius (changes requested):

doctrine-mapping.xsd
guilliamxavier: First, they must follow `^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$` (with a `*`, not a `+`, on the character class for after the first char) according to http://php.net/manual/en/language.oop5.basic.php and verified at https://3v4l.org/klVjj.
=> Do you want to consider single-char names as invalid?

Second, that is a *mono-byte* regex, as explained at https://stackoverflow.com/a/5358944 (modulo the "*accidentally*") and verified at https://3v4l.org/XqJGf which shows that for UTF-8 encoded strings, the equivalent of
`/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/` is
`/^[a-zA-Z_\x7f-\x{10ffff}][a-zA-Z0-9_\x7f-\x{10ffff}]*$/u`.
Now, XML Regex is _Unicode mode_ (and all the files are in UTF-8 encoding), so I translated `\x7f-\x{10ffff}` to `&#x7F;-&#x10FFFF;`, because `&#x7F;-&#xFF;` (translated from `\x7f-\xff`) only matches Unicode characters in the range U+7F..U+FF, and I have indeed tested that it would then consider non-Latin1 characters as invalid (e.g. "The value 'Слон' is not accepted by the pattern ...").
=> Do you want to consider non-Latin1 (or even non-ASCII) names as invalid?

Ocramius: Hmm, indeed about `*` (probably lucky that nobody hit the pattern before - the library always generates 2+ chars for that particular use-case).

As for the multi-byte, I'm inclined to say that supporting multibyte strings for class names is a no-go for a lot of stuff, because most of the `get_class()`-based operations (not just in ORM, but in other libraries) simply don't operate under those circumstances, and you will get someone doing a nice unsafe `substr()` somewhere.

I'd say that we should keep those out of support: that would be an opinionated stance (by me), not the "full support for everything".

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
guilliamxavier:
```
$ ./vendor/bin/phpunit --group 6389
PHPUnit 6.5.9 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.7-2+0~20180714182139.1+stretch~1.gbp3fcba8
Configuration: /z/phpunit.xml.dist

.................................................EEEEEEEEEEEEEE   63 / 63 (100%)

Time: 1.11 seconds, Memory: 58.00MB

There were 14 errors:

1) Doctrine\Tests\ORM\Mapping\XmlMappingDriverTest::testInvalidateXmlSchema with data set #0 ('/z/tests/Doctrine/Tests/ORM/M...cm.xml')
DOMDocument::schemaValidate(): Element '{http://doctrine-project.org/schemas/orm/doctrine-mapping}class': This element is not expected. Expected is one of ( {http://doctrine-project.org/schemas/orm/doctrine-mapping}mapped-superclass, {http://doctrine-project.org/schemas/orm/doctrine-mapping}entity, {http://doctrine-project.org/schemas/orm/doctrine-mapping}embeddable, ##other{http://doctrine-project.org/schemas/orm/doctrine-mapping}* ).

/z/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php:185
/z/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php:170

2) Doctrine\Tests\ORM\Mapping\XmlMappingDriverTest::testInvalidateXmlSchema with data set #1 ('/z/tests/Doctrine/Tests/ORM/M...20.xml')
DOMDocument::schemaValidate(): Element '{http://doctrine-project.org/schemas/orm/doctrine-mapping}custom-id-generator', attribute 'class': [facet 'pattern'] The value 'Foo Bar' is not accepted by the pattern '([a-zA-Z_-􏿿][a-zA-Z0-9_-􏿿]*\\)*[a-zA-Z_-􏿿][a-zA-Z0-9_-􏿿]*'.

/z/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php:185
/z/tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php:170

*SNIP*

ERRORS!
Tests: 63, Assertions: 49, Errors: 14.
```

Alternatively it works with `@expectedException \PHPUnit\Framework\Error\Warning` (or do you prefer `$this->expectException(\PHPUnit\Framework\Error\Warning::class);`?) and reducing the body to the simple call
`$this->doValidateXmlSchema($xmlMappingFile);`
but that's depending both on `DOMDocument::schemaValidate()` emitting an `E_WARNING` and on PHPUnit translating the Warning to an exception...

This is only for testing XSD validation (actually testing that the schema definition is correct), how would you move it into the XML driver (without merging #6728)?

doctrine-mapping.xsd
guilliamxavier: So the `&#x7F;-&#x10FFFF;` part should become just `&#x7F;` (because `&#x7F;-&#xFF;` would still accept e.g. the character `é` = `&#xE9;` which is 2-bytes `\xC3\xA9` in UTF-8), or even be removed entirely (do you accept the `DEL` char)?
And I should remove the test file `pattern-fqcn-valid-unicode.xml`, or alternatively rename it to `pattern-fqcn-invalid-multibyte.xml` and keep only the `"Éléphant"` case?

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
guilliamxavier: Would it be enough to rename it, or do you prefer that I factorize the common parts differently (like the test functions)?

BTW, would it be a BC-break to rename an existing public test function?
E.g. `testValidateXmlSchema` to, say, `testValidateXmlSchemaWithValidMapping` (so that `testInvalidateXmlSchema` could be `testValidateXmlSchemaWithInvalidMapping`)? (maybe you have better ideas?)

doctrine-mapping.xsd
```diff
@@ -202,7 +202,7 @@
 
   <xs:simpleType name="tablename" id="tablename">
     <xs:restriction base="xs:token">
-      <xs:pattern value="[a-zA-Z_u01-uff.]+" id="tablename.pattern">
+      <xs:pattern value="([^`.][^.]*\.)?([a-zA-Z0-9_$#&#x80;-&#x10FFFF;]+|`.+`)" id="tablename.pattern">
```
guilliamxavier: Note: the proposed part for an unquoted table name, `[a-zA-Z0-9_$#&#x80;-&#x10FFFF;]+`, is (sort of) the _union_ of formats allowed by different vendors (mainly MySQL, PostgreSQL, Oracle).
Would you on the contrary want the _intersection_, e.g. `[a-zA-Z][a-zA-Z0-9_]*` (very restrictive)?

Do you also want that multibyte characters be considered as invalid (like for `fqcn`), even in _quoted_ table names?

And maybe also limit the maximum length? (to e.g. 128 or even 30 chars?)

Ocramius:
@mikeSimonson can you take over here?

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
Ocramius: Having completely separate methods, even if some code is duplicated, is preferable here

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
Ocramius: @greg0ire ping ^

doctrine-mapping.xsd
Ocramius: I don't know if we perform binary-unsafe operations on table names in the codebase (specifically: naming strategy, quoting), so being restrictive is OK.

Maximum length is to be avoided, because it is highly platform-dependent, and I've already seen DBs with horrendous generated table names in OracleDB (ascii-only): those would be surely over 64 chars. Unsure if any of the supported platforms goes over 128 chars though.

mikeSimonson (changes requested):
So my point of view.

We should remove the mutlibyte char entirely I think.

I am not sure a maximum length is necessary as it's probably super vendor/engine dependent.

Rename the test with valid unicode to invalid unicode.

Do we keep the "é" as valid is an open question.

You can add the type delcaration in your function in one commit and if you fell like it you can add them everywhere else in another commit or preferably PR.

#6728 should be finished and merged so that you don't need that logic in the tests.

Thanks for all the tests

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
greg0ire: @guilliamxavier You could reuse the `libxml_use_internal_errors` yadda yadda trick from https://github.com/doctrine/doctrine2/pull/6728/files#diff-f96b6283e2653400c7e5fbb845f50748R867 here.

@Ocramius not 100% sure why you ping me, if it is to know the status of #6728, it is blocked by #7199, which is blocked until further notice

doctrine-mapping.xsd
```diff
@@ -202,7 +202,7 @@
 
   <xs:simpleType name="tablename" id="tablename">
     <xs:restriction base="xs:token">
-      <xs:pattern value="[a-zA-Z_u01-uff.]+" id="tablename.pattern">
+      <xs:pattern value="([&#x20;-&#x7F;-[\.]]+\.)?([a-zA-Z][a-zA-Z0-9_]*|`[&#x20;-&#x7F;]+`)" id="tablename.pattern">
```
guilliamxavier: Note: the backslash before the dot in the subtracted character class is "redundant", but without it PhpStorm flags a syntax error "Unclosed character class"

guilliamxavier:
@Ocramius & @mikeSimonson I have reworked the commits according to your reviews and updated the PR, is it OK now?

doctrine-mapping.xsd
guilliamxavier: Wait, wouldn't ``"([a-zA-Z][a-zA-Z0-9_]*\.)?[a-zA-Z][a-zA-Z0-9_]*|([&#x20;-&#x7F;-[`\.]][&#x20;-&#x7F;-[\.]]*\.)?`[&#x20;-&#x7F;]+`"`` be better? (the schema part cannot be backquoted, and will be SQL-quoted only if the table part is backquoted)

guilliamxavier: I have opened a separate #7329 about the peculiar schema part quoting behavior. But for now this PR should be fine as it is.

ostrolucky (approved):
Feels over-engineered at first, but valid. Would use more verbose variable names next time.

guilliamxavier:
Sorry, only now do I understand that this is blocked by #6728 because merging both PRs would lead to logic duplication in the XML driver and in the test class (`libxml_use_internal_errors()` etc.), and since [enforced XSD validation is a desirable feature](https://github.com/doctrine/doctrine2/issues/6444#issuecomment-302871369) (for 3.0), the tests should wait for it and then probably use `XmlDriver::loadMetadataForClass()` rather that low-level DOM/libxml stuff (BTW, even the current `testValidateXmlSchema()` should be refactored then).

I also agree that this feels over-engineered (and as for the too-terse variable names, I guess I let myself influenced by `$list`/`$invalid`/`$item`, and agree should find better).

But since #6728 is itself blocked by #7199 which in turn is "blocked until further notice", and this PR maybe mixes too many things, I am going to extract independent parts into separate (and hopefully immediately-applicable) PRs :slightly_smiling_face:

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -153,7 +153,7 @@ public function testInvalidMappingFileException()
     public function testValidateXmlSchema($xmlMappingFile)
     {
         $xsdSchemaFile  = __DIR__ . '/../../../../../doctrine-mapping.xsd';
-        $dom            = new \DOMDocument('UTF-8');
+        $dom            = new \DOMDocument('1.0', 'UTF-8');
```
guilliamxavier: Actually better here: just `new \DOMDocument();` (done in #7345)

tests/Doctrine/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -149,33 +149,128 @@ public function testInvalidMappingFileException()
...
+     */
+    static private function getInvalidXmlMappingMap()
+    {
+        $invalid = [
+            'Doctrine.Tests.Models.DDC889.DDC889Class.dcm' => [
+                "Element '{http://doctrine-project.org/schemas/orm/doctrine-mapping}class': This element is not expected. Expected is %s.",
+            ],
+        ];
+
+        // Convert basic sprintf-style formats to PCRE patterns
```
guilliamxavier: Maybe better: use `assertStringMatchesFormat()` instead of `assertRegeExp()` at line 180 (and e.g. `%S`, `%a` or `%A` rather than `%s` in the format)

PR #7327, opened 2018-07-31
Fix inaccurate description of the `ORMException` interface

greg0ire:


Ocramius (approved):
LGTM :+1:

carnage:
1. Not an interface in 2.x branches, so not applicable
2. Still seems not to be an interface in 3.x branches (class was also moved to a new namespace), so perhaps the underlying change was dropped. 

Suggested resolution: move to rejected

greg0ire:
Yes, let's reject it indeed 👍 
The underlying change (moving from class to interface) will still happen, I'm waiting for 2.10.0 to be released before working on it, in case changes are needed on 2.10.x first.

PR #7336, opened 2018-08-07
Problem wit composite primary keys and ManyToMany-Relations

areanet-schmid:
See https://github.com/area-net-gmbh/doctrine2/issues/1


Ocramius:
@areanet-schmid this sort of change requires test case additions

PR #7347, opened 2018-08-10
Fix XSD regex for FQCN

guilliamxavier:
Extracted from #7324

---

Also applicable to `master` as is.


ostrolucky (approved):

lcobucci (dismissed):
@guilliamxavier can you please add tests which were not passing with the previous regex?

ostrolucky:
@lcobucci see referenced PR. Point is to merge tests separately so let's not hold this up

lcobucci:
@ostrolucky we don't have to hurry things too, let's think things properly and get this merged.

Issue #7329, opened 2018-08-01
Quoting problems with schema and dots in table name mapping


guilliamxavier:
Test entity:

```php
/** @Entity @Table(TEST_VALUES_HERE) */
class Qux
{
    /** @Id @Column */
    public $id;
}
```

Test code:

```php
echo $entityManager->createQuery('SELECT e FROM ' . Qux::class . ' e')->getSQL();
```

Full results:
- [table_schema-2.6.xlsx](https://github.com/doctrine/doctrine2/files/2250386/table_schema-2.6.xlsx)
- [table_schema-master_3.0.xlsx](https://github.com/doctrine/doctrine2/files/2250388/table_schema-master_3.0.xlsx)

Recap (`master` is simpler):

### `2.6` (current stable):

**Issues:**
- both ``name="`other-b`.foo"`` and ``schema="`other-b`", name="foo"`` generate ``SELECT ... FROM `other-b`.foo`` _(invalid SQL)_
    and one must do ``name="other-b.`foo`"`` or ``schema="other-b", name="`foo`"`` to get a ``FROM "other-b"."foo"`` (which is still not exactly the wanted ``FROM "other-b".foo``)
- ``name="`foo.x`"`` generates ``SELECT ... FROM `foo.x` `` _(invalid SQL)_
    and there is no way to get a ``FROM "foo.x"``
- ``name="other.`foo.x`"`` generates ``SELECT ... FROM "other"."foo"."x"`` and
    ``schema="other", name="`foo.x`"`` generates ``SELECT ... FROM other.x` `` _(invalid SQL)_

**Thoughts:**
- Option 1: Do not change anything, just document the current (surprising) behavior.
- Option 2: Disallow dots in table names/schemas (via mapping validation).
- Option 3: Handle backquotes on the schema [part] too, and properly handle backquoted dots:
    - both ``name="`other-b`.foo"`` and ``schema="`other-b`", name="foo"`` would generate a ``FROM "other-b".foo``
        while ``name="other.`foo-a`"`` and ``schema="other", name="`foo-a`"`` would generate a ``FROM other."foo-a"``
    - ``name="`foo.x`"`` would generate a ``FROM "foo.x"``
        (``name="other.foo"`` and ``schema="other", name="foo"`` still generating a ``FROM other.foo``)
    - both ``name="other.`foo.x`"`` and ``schema="other", name="`foo.x`"`` would generate a ``FROM other."foo.x"``

### `master` (future `3.0`):

**Issues:**
[...]

PR #7348, opened 2018-08-10
Fix XSD regex for table name

guilliamxavier:
Extracted from #7324 (see also #7329 about the optional schema part).

---

Note: for `master` (3.x, with auto-quoting) the patch would rather be:

```diff
-      <xs:pattern value="[a-zA-Z\-_u01-uff.]+" id="tablename.pattern">
+      <xs:pattern value="[&#x20;-&#x7F;]+" id="tablename.pattern">
```

### DIFF
--- a/doctrine-mapping.xsd
+++ b/doctrine-mapping.xsd
@@ -202,7 +202,7 @@
 
   <xs:simpleType name="tablename" id="tablename">
     <xs:restriction base="xs:token">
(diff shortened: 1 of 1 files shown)


ostrolucky (approved):

lcobucci (dismissed):
@guilliamxavier can you please add tests which were not passing with the previous regex?

PR #7350, opened 2018-08-10
Update doctrine/coding-standard in 2.x branch

ostrolucky:
See https://github.com/doctrine/doctrine2/pull/7328#discussion_r209202305

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "symfony/console": "~3.0|~4.0"
     },
     "require-dev": {
-        "doctrine/coding-standard": "^1.0",
+        "doctrine/coding-standard": "^4.0",
         "phpunit/phpunit": "^6.5",
         "squizlabs/php_codesniffer": "^3.2",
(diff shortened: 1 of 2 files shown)


Majkl578:
Any particular reason for this? We don't enforce CS in 2.x at all. For 2.7 we could at least consider the approach DBAL uses with CS checked on patch level.

ostrolucky:
See referenced comment. Reason is to allow contributors to check new code if they want, without team member having to point them out.

Having CI not enforcing this does not mean we don't waste time by back and forth talk about CS in PRs by humans.

Majkl578:
I don't see how this helps when you can't even run vendor/bin/phpcs.

ostrolucky:
What do you mean I can't? I can and I did. I add `NewFile.php`, I run `vendor/bin/phpcs NewFile.php` and fix errors.

ostrolucky:
Replaced with https://github.com/doctrine/doctrine2/pull/7473

PR #7354, opened 2018-08-12
IsNull function for MYSQL

userator:
IsNull function fo MYSQL


Ocramius (changes requested):
Why is a function required when `IS NULL` already exists? What's the difference?

ostrolucky:
This is MySQL specific. Doctrine provides abstraction layer, these functions must work in all other supported database platforms.

PR #7360, opened 2018-08-18
Document getPartialReference() properly

lcobucci:
According to the current implementation that method also returns `null`, the interface's documentation was incorrect.

https://github.com/doctrine/doctrine2/blob/d2b4dd71d2a276edd65d0c170375b445f8a4a4a8/lib/Doctrine/ORM/EntityManager.php#L514-L516


Majkl578:
:+1: Should we also document `@throws`?

lcobucci:
@Majkl578 what exception exactly? The method itself doesn't throw exceptions, I'm not sure about the methods that are called there.

Majkl578:
Ah, it does throw in master, not in 2.6, okay then. :+1:

Majkl578 (approved):

PR #7363, opened 2018-08-21
Fix compatibility with phan

philippe-unitiz:
See https://github.com/phan/phan/issues/1897

### DIFF
--- a/lib/Doctrine/ORM/Query/Expr/Base.php
+++ b/lib/Doctrine/ORM/Query/Expr/Base.php
@@ -56,7 +56,7 @@ abstract class Base
     protected $parts = [];
 
     /**
-     * @param array $args
+     * @param mixed $args
      */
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Query/Expr/Base.php
```diff
@@ -56,7 +56,7 @@
     protected $parts = [];
 
     /**
-     * @param array $args
+     * @param mixed $args
```
Ocramius: Probably `mixed[]`?

stof: the Base class indeed does not enforce passing an array to `addMultiple` (and so to the constructor), as it will cast it (which will wrap scalars into an array containing them)

Majkl578 (approved):

Issue #7321, opened 2018-07-23
Documentation about Yaml, mention it as discouraged to use?


iltar:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6

#### Summary

The documentation page about [Yaml](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/yaml-mapping.html) does not mention anything about it being removed/unsupported in the future. I think it's a good idea to mention on here that it's discouraged to use Yaml as configuration.

Whenever I see people suggest to not use Yaml because it's going to be removed as core supported configuration, the response is usually "oh I did not know that".

Ocramius:
@doctrine/team-documentation can we do something about this one?

weaverryan:
I would be +1 for removing it on the latest minor version docs. Basically, because it *is* now deprecated, why show it at all? Older versions of the docs will still exist for existing users to reference. We can cleanup the new version right now.

iltar:
It's already known to be deprecated and 2.6 is the current documentation. If a developer takes a look at the documentation, and doesn't see a deprecation, chances are that everything is going to be created in Yaml. This is a waste of time for the developer because it's already known to disappear with 3.0. Not letting the developer know this as soon as possible, will thus result in extra work while it was not necessary.

SenseException:
@iltar I've created PR #7374. Does it cover this issue?

Majkl578:
#7374 has been merged into 2.6.
https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/yaml-mapping.html#yaml-mapping

PR #7374, opened 2018-08-25
Deprecation message in documentation for YAML

SenseException:
This PR is supposed to handle #7321. Because YAML is removed in 3.0, the docs are still missing the message, that YAML shouldn't be used anymore. I found no information when the deprecation should be put in place, so I created the branch from the current 2.6 version. This commit needs to be added to 2.7 too.

Does this deprecation message need to be put in other places too? "Getting started" is using YAML, so it is strange to have it in use while we are trying to move users to other mapping solutions of Doctrine.

### DIFF
--- a/docs/en/reference/yaml-mapping.rst
+++ b/docs/en/reference/yaml-mapping.rst
@@ -1,6 +1,10 @@
 YAML Mapping
 ============
 
+.. note::
+    The YAML driver is deprecated and will be removed in version 3.0.
+    It is strongly recommended to switch to one of the other mappings.
+
 The YAML mapping driver enables you to provide the ORM metadata in
 form of YAML documents.
 
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -103,17 +103,20 @@ Install Doctrine using the Composer Dependency Management tool, by calling:
     $ composer install
 
 This will install the packages Doctrine Common, Doctrine DBAL, Doctrine ORM,
-Symfony YAML and Symfony Console into the `vendor` directory. The Symfony 
-dependencies are not required by Doctrine but will be used in this tutorial.
+into the `vendor` directory.
 
 Add the following directories:
 ::
 
     doctrine2-tutorial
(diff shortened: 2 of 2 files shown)


docs/en/reference/yaml-mapping.rst
```diff
@@ -1,6 +1,10 @@
 YAML Mapping
 ============
 
+.. attention::
```
jwage: Do we have an `attention` directive? I don't see it in https://github.com/doctrine/rst-parser/tree/master/lib/Directives or https://github.com/doctrine/doctrine-website/tree/master/lib/RST/Directive

SenseException: I didn't know that not all rst-syntax is available with rst-parser. Better I'll generate my local files with rst-syntax before I introduce changes instead of using other rst-tools.

SenseException: I've changed it to "note". This one is available in doctrine/rst-parser.

Majkl578: Can we have a `::big-blinking-red-alert` for this specific purpose? :smiley:

greg0ire (dismissed):

docs/en/tutorials/getting-started.rst
```diff
@@ -103,18 +103,21 @@ Install Doctrine using the Composer Dependency Management tool, by calling:
     $ composer install
 
 This will install the packages Doctrine Common, Doctrine DBAL, Doctrine ORM,
-Symfony YAML and Symfony Console into the `vendor` directory. The Symfony
```
Majkl578: We should also drop `symfony/yaml` from the composer.json snippet above.

Majkl578: nevermind, it's still used in the tutorial

Majkl578 (approved):

nitso:
Just for history: discussion on that topic is in https://github.com/doctrine/DoctrineBundle/issues/776 and in linked issues. Could not find links there.

PR #7399, opened 2018-09-12
Update DatabaseDriver.php

sof1105:
Adding an id column if it's messing and setting it as primary key instead of throwing an Exception


Ocramius (changes requested):
This change is invalid, since this metadata driver is supposed to reflect the database state, and not invent new details.

Issue #3037, opened 2013-03-05
DDC-2332: [UnitOfWork::doPersist()] The spl_object_hash() generate not unique hash!


doctrinebot:
Jira issue originally created by user fchris82:

I created fixtures and some data was inserted many times without calling the `Task` entity PrePersist event listener.

I printed the used and generated hash and I saw a `Proxies\*_CG_*\Asitly\ProjectManagementBundle\Entity\User` hash equal a `Task` entity hash!

tomaszmadeyski:
I can confirm having this issue. My case is working on quite big collection of entities, and code (simplified) looks like this:

```
//$idsArr is about 2k elements
foreach ($idsArr as $id) {
    $entityA = new A($id);
    $em->persist($entityA);
}
$em->flush();

//some code happening in between but everything is happening during single request

//$anotherIdsArr is about 2k items
foreach ($anotherIdsArr as $anotherId) {
    $entityB = new B($someConstValue, $anotherId);
    $em->persis($entityB);
}
$em->flush();
```
Class A definition:
```
/**
 * @ORM\Entity()
 * @ORM\Table(name="a")
 */
class A
{
    /**
     * @var integer
     *
     * @ORM\Id
     * @ORM\Column(type="integer")
     */
    protected $l1;

    /**
     * @param integer $l1
     */
    public function __construct($l1)
    {
        $this->l1 = $l1;
    }

} 
```
Class B definition
```
/**
 * @ORM\Entity()
 * @ORM\Table(name="b")
 */
class B
{
    /**
     * @var integer
     *
     * @ORM\Id
     * @ORM\Column(type="integer")
     */
    protected $l1;

    /**
     * @var integer
     *
     * @ORM\Id
     * @ORM\Column(type="integer")
     */
    protected $l2;

    /**
     * @param integer $l1
     * @param integer $l2
     */
    public function __construct($l1, $l2)
    {
        $this->l1 = $l1;
        $this->l2 = $l2;
    }

}
```

While trying to persist B instance 
`$entityB = new B(12, 8995); ` 
I get 
`$oid == '0000000013204a0200007fc5ffc03559'` 
and such element already exists in `$this->entityIdentifiers` so `getEntityState` method returns `STATE_MANAGED` so `$entityB` is not persisted.

`$this->entityIdentifiers['0000000013204a0200007fc5ffc03559']` stores instance of A entity: `$a = new A(6148)`

[...]

PR #7407, opened 2018-09-24
Adding unit tests showing #3037

sandvige:
#3037

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -105,6 +105,15 @@ class UnitOfWork implements PropertyChangedListener
     private $identityMap = [];
 
     /**
+     * Associate entities with OIDs to ensure the GC won't recycle a managed entity
+     *
+     * DDC-2332 / #3037
+     *
+     * @var array
+     */
+    private $oidMap = array();
+
+    /**
      * Map of all identifiers of managed entities.
      * Keys are object ids (spl_object_hash).
      *
@@ -1498,7 +1507,8 @@ class UnitOfWork implements PropertyChangedListener
     public function addToIdentityMap($entity)
     {
         $classMetadata = $this->em->getClassMetadata(get_class($entity));
-        $identifier    = $this->entityIdentifiers[spl_object_hash($entity)];
+        $oid           = spl_object_hash($entity);
+        $identifier    = $this->entityIdentifiers[$oid];
 
         if (empty($identifier) || in_array(null, $identifier, true)) {
             throw ORMInvalidArgumentException::entityWithoutIdentity($classMetadata->name, $entity);
@@ -1507,6 +1517,7 @@ class UnitOfWork implements PropertyChangedListener
         $idHash    = implode(' ', $identifier);
         $className = $classMetadata->rootEntityName;
 
+        $this->oidMap[$oid] = $entity;
         if (isset($this->identityMap[$className][$idHash])) {
             return false;
         }
@@ -1622,6 +1633,7 @@ class UnitOfWork implements PropertyChangedListener
 
         $className = $classMetadata->rootEntityName;
 
+        unset($this->oidMap[$oid]);
         if (isset($this->identityMap[$className][$idHash])) {
             unset($this->identityMap[$className][$idHash]);
             unset($this->readOnlyObjects[$oid]);
@@ -2482,6 +2494,7 @@ class UnitOfWork implements PropertyChangedListener
     {
         if ($entityName === null) {
             $this->identityMap =
+            $this->oidMap =
             $this->entityIdentifiers =
             $this->originalEntityData =
             $this->entityChangeSets =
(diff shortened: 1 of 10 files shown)


sandvige:
All tests are passing, except the one related to code coverage (it appears to be stuck). It is possible this dirty fix is implying a few quirks on performances (It's keep ALL entities, ever tracked), so it is possible this array is pretty big and the coverage is not very happy.

tests/Doctrine/Tests/ORM/Functional/IdentityMapTest.php
```diff
@@ -254,5 +255,50 @@ public function testCollectionValuedAssociationIdentityMapBehaviorWithRefresh()
...
+     */
+    public function testHashCollision() {
+        $user = new CmsUser();
+        $user->username = "Test";
+        $user->name     = "Test";
+        $this->_em->persist($user);
+        $this->_em->flush();
+
+        $articles = [];
+        for($i=0;$i<100;$i++) {
```
Ocramius: Can we reproduce this with just two?

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/IdentityMapTest.php
sandvige: The problem is the underlying GC implementation. With PHP 5.6, it is triggering the assert at the 85th entity. With PHP7, it's at the 100th. I think the problem is based on what the internal implementation of PHP is choosing among the available memory slots. When I try with 2, nothing happen.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1622,9 +1633,10 @@ public function removeFromIdentityMap($entity)
 
         $className = $classMetadata->rootEntityName;
 
+        unset($this->oidMap[$oid]);
+        unset($this->readOnlyObjects[$oid]);
```
sandvige: There's exactly the same issue with the read only objects: they can points at dead entities. We have to clear the oid even if the entity is not found into the identityMap.

sandvige:
I don't understand why the code coverage is stuck with my implementation. Is there something asserting (when using coverage only) something on the object layout of the UOW?

tests/Doctrine/Tests/ORM/Functional/IdentityMapTest.php
Ocramius: PHP 5.6 is generally not relevant - what's important is to verify PHP 7. The fact that it acts on the very last iteration, yet we can't reduce the iteration count to just 2 means that we don't fully understand the problem yet.

sandvige: Well, we're talking about how PHP is recycling memory. What if PHP is storing available memory slots into an hashmap? It would result into something more or less random, this is why having a lot of slots having the same size being available (due to recycling) is making this error to happens more frequently. This unit tests is just showing the problem, but there's too much random factor in this bug. Keep in mind that this is happening, and why: the documentation about spl_object_hash is clear: ids are recycled, thus not unique in time. My fix is pretty simple, it just ensure the UOW is still holding a reference on managed entities to avoid this pitfall. To be honest, I don't understand why we need a unit tests for this case, this is part of the way PHP is handling memory. And we DO have a unit test that is showing the issue, what a chance :). I think we do understand what is happening here, this is why the unit test is "statically working". I think this unit test should NOT be merged into the core as PHP might change the way it handles memory in the future. This fix is important as long as the UOW is using spl_object_hash (and spl_object_id for PHP7).

sandvige:
@Ocramius a build failed on travis because of this:
```
$ phpenv global 7.2 2>/dev/null
7.2 is not pre-installed; installing
Downloading archive: https://s3.amazonaws.com/travis-php-archives/binaries/ubuntu/14.04/x86_64/php-7.2.tar.bz2
404.01s$ curl -s -o archive.tar.bz2 $archive_url && tar xjf archive.tar.bz2 --directory /
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
0.01s0.03s$ phpenv global 7.2
rbenv: version `7.2' not installed
The command "phpenv global 7.2" failed and exited with 1 during .
Your build has been stopped.
```

Do you want to restart the build, because as a project maintainer, I'm pretty sure you have this button, or should I bump the PR with a fake commit?

lcobucci:
@sandvige I've restarted the build. Could you please rebase this branch on top of `2.7` instead of merging things in here? It would be slightly better for reviews if we don't have all these unnecessary merge commits.

sandvige:
@lcobucci Done.

bigfoot90:
Are there any news on this?
I'm also facing this issue, tried to apply this changes on my project and it's working very well for me, I'm using `doctrine/orm:2.9.5` on php7.4.
Should I create a new MR from branch `2.9.x`?
Please tell me if I ca do something to help.
Thanks

greg0ire:
>  Should I create a new MR from branch 2.9.x?

Yes, that would be helpful I think! I need to read the comments on this PR to see what's missing, but let's try this 👍

bigfoot90:
@greg0ire Created new PR of this changes from `2.9.x` in https://github.com/doctrine/orm/pull/8994

mpdude:
@sandvige maybe you want to join #10791

PR #7413, opened 2018-09-25
Apply Doctrine CS 5.0

Majkl578:


lib/Doctrine/ORM/Annotation/GeneratedValue.php
```diff
@@ -14,7 +14,6 @@
      * The type of Id generator.
      *
      * @var string
-     *
```
Majkl578: This may be something to fix in Doctrine CS, but IMO not a big deal.

Ocramius (approved):
Besides spacing in docblocks (https://github.com/doctrine/doctrine2/pull/7413/files#r230605819), seems :+1:

PR #7412, opened 2018-09-25
Some formatting improvements

ThomasLandauer:

PR #7420, opened 2018-10-07
Update getting-started.rst

Stanzi1791:
I propose `"doctrine/orm": "2.4.*",` is changed to `"doctrine/orm": "*",` in the example JSON file at the start of the Getting Started with Doctrine tutorial.

Without the change, I got a missing AnnotationFactory.php class error when running the `orm:schema-tool:create` command, and changing it to `"*"` fixed that for me.

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -87,7 +87,7 @@ the following contents:
 
     {
         "require": {
-            "doctrine/orm": "2.4.*",
+            "doctrine/orm": "*",
             "symfony/yaml": "2.*"
         },
         "autoload": {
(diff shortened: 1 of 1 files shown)


Majkl578:
Using `*` constraint is wrong as it would allow ORM 3.0 (when released) and break your application.
But you're correct that `2.4.*` is wrong too, it should be `^2.6` - can you use that instead of `"*"`?
Thanks.

Ocramius:
Can also be a `composer require` command: composer will automatically
select the latest stable release, and lock it with the caret operator.

Stanzi1791:
Changed to `^2.6`.

Majkl578:
Hi, sorry for bad news - #7412 got merged yesterday and it changed the constraint to `^2.6.2` so this PR is now redundant. :/ But thank you anyway. 👍

PR #7430, opened 2018-10-14
Use `in_array` instead of `array_search` + comparison

carusogabriel:
As we don't need the array index here, we can use the `in_array` function.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -1741,7 +1741,7 @@ class BasicEntityPersister implements EntityPersister
             if (is_array($value)) {
                 $in = sprintf('%s IN (%s)', $column, $placeholder);
 
-                if (array_search(null, $value, true) !== false) {
+                if (in_array(null, $value, true)) {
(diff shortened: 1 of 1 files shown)


Ocramius (approved):
:+1:

Majkl578:
https://travis-ci.org/doctrine/doctrine2/jobs/448594099#L558

carusogabriel:
@Majkl578 What is the state of Doctrine's CS here in the ORM? Strange the CI didn't catch it.

Majkl578:
Master requires Doctrine CS 5.0 with no errors. It would be caught but someone didn't let the stage run. 😏

Ocramius:
@Majkl578 that stage won't run if everything else is red

Majkl578:
It's broken because of this PR and BC break in PHPUnit 7.4 which has been addressed by the recent discriminator autodiscovery removal. So right now this is the only thing that breaks master: https://travis-ci.org/doctrine/doctrine2/builds/450559469

Ocramius:
Right, then fixup PR please. As I mentioned above, I'd rather keep things moving forward. Here, I based my merge on top of https://travis-ci.org/doctrine/doctrine2/jobs/441387813

PR #7436, opened 2018-10-19
Сheck platform supports foreign key constraints

nighten:
Сheck platform supports foreign key constraints when create schema from metadata


Ocramius (changes requested):
A test case is strictly required for this change

PR #7437, opened 2018-10-19
Update description and variable

martindelloro:


docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -332,7 +332,7 @@ With COUNT DISTINCT:
 
     <?php
     $query = $em->createQuery('SELECT COUNT(DISTINCT u.name) FROM CmsUser');
-    $users = $query->getResult(); // array of ForumUser objects
+    $countUsers = $query->getSingleResult(); // array of ForumUser count
```
Majkl578: Maybe `getSingleScalarResult()` would be even more appropriate for `COUNT()`? Also please name the variable `$usersCount`.

Majkl578 (changes requested):

PR #7438, opened 2018-10-19
change return type of doctrine query

EliuTimana:
the return type of the query is not correct.


Majkl578:
Hi, thanks for your PR, I'm goint o close it in favor of #7437 which changes the `get*Result()` method to a more appropriate one.

PR #7441, opened 2018-10-24
$hydrationMode throughout can be a string as well as int (for custom modes)

asgrim:
As per docs: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/dql-doctrine-query-language.html#custom-hydration-modes

You can add custom hydration modes. Whilst running phpstan on a project, it discovered that whilst `addCustomHydrationMode` accepts a `string`, when you call a method like `getResult` the `$hydrationMode` parameter only accepts `int` and therefore phpstan complains.

I've added `string|` to all existing type hints.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -98,7 +98,7 @@ abstract class AbstractQuery
     /**
      * The hydration mode.
      *
-     * @var integer
+     * @var string|int
      */
     protected $_hydrationMode = self::HYDRATE_OBJECT;
 
@@ -680,8 +680,8 @@ abstract class AbstractQuery
     /**
      * Defines the processing mode to be used during hydration / result set transformation.
      *
-     * @param integer $hydrationMode Doctrine processing mode to be used during hydration process.
-     *                               One of the Query::HYDRATE_* constants.
+     * @param string|int $hydrationMode Doctrine processing mode to be used during hydration process.
+     *                                  One of the Query::HYDRATE_* constants.
      *
      * @return static This query instance.
      */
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -91,7 +91,7 @@
     /**
      * The hydration mode.
      *
-     * @var int
+     * @var int|string
```
Ocramius: @Majkl578 WDYT about proposing hydration modes as enumerable types here as well? (Not related with this patch)

Ocramius (dismissed):
LGTM :+1:

Majkl578:
Ah nevermind, didn't read the diff properly. :|
Still needs to be retargeted to 2.6 though.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -662,7 +662,7 @@ public function getHydrationMode()
      *
      * Alias for execute(null, $hydrationMode = HYDRATE_OBJECT).
      *
-     * @param int $hydrationMode
+     * @param int|string $hydrationMode
```
malukenho: May we use `string|int` as in the rest of the file?

Majkl578:
Base branch changed, but rebase is required - @asgrim do you want to do it (and keep GPG sig) or want me to handle it?

asgrim:
I'm on it :+1:

asgrim:
Rebased and use `string|int` consistently instead of `int|string` :+1:

Majkl578 (approved):

lcobucci:
This LGTM just a tiny question: do we have a test using a custom hydration mode using strings? I'm on mobile now so can't check at the moment. It's also fine to move this to another issue/pr

lcobucci:
> This LGTM just a tiny question: do we have a test using a custom hydration mode using strings? I'm on mobile now so can't check at the moment. It's also fine to move this to another issue/pr

We do have tests :+1:

PR #7448, opened 2018-10-30
Throw exception in case of calling flush inside flush

aivus:
Restrict using `EntityManager::flush` inside `EntityManager::flush`

Ref: https://github.com/doctrine/doctrine2/issues/6822#issuecomment-429305777


aivus:
@Ocramius added exception in case of calling `flush` during `flush`. Covered with test.

Should we use custom exception instead of `LogicException`?

Ocramius:
@aivus yes, please see the work done by @greg0ire in https://github.com/doctrine/doctrine2/pull/6743

We're putting a lot of effort in this, so that exceptions make more sense to consumers.

This is more like an "unsupported operation" type of exception.

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -349,12 +357,21 @@ public function detach($object)
      * @throws OptimisticLockException If a version check on an entity that
      *         makes use of optimistic locking fails.
      * @throws ORMException
+     * @throws LogicException
      */
     public function flush()
     {
+        if ($this->flushInProgress) {
+            throw new LogicException('EntityManager::flush must not be called inside another EntityManager::flush');
```
Ocramius: We will need an entry in `UPGRADE.md`, since this is still a BC break for ORM v3

tests/Doctrine/Tests/ORM/EntityManagerTest.php
```diff
@@ -288,4 +291,34 @@ public function testGetReferenceRetrievesReferencesWithGivenProxiesAsIdentifiers
...
+            {
+                if ($this->alreadyInPreFlush) {
+                    $this->testCase->fail('Flush inside flush called. It should not happened. Fail test to prevent infinity loop');
+                }
+                $this->alreadyInPreFlush = true;
+                $eventArgs->getEntityManager()->flush();
+            }
+        };
+
+        $this->expectException(LogicException::class);
```
Ocramius: As discussed in the thread, a custom exception is to be expected here

tests/Doctrine/Tests/ORM/EntityManagerTest.php
```diff
@@ -288,4 +291,34 @@ public function testGetReferenceRetrievesReferencesWithGivenProxiesAsIdentifiers
         self::assertTrue($this->em->contains($simpleIdReferenceB));
         self::assertTrue($this->em->contains($nestedIdReference));
     }
+
+    public function testCallingFlushInsideFlush()
```
Ocramius: `: void`

tests/Doctrine/Tests/ORM/EntityManagerTest.php
```diff
@@ -288,4 +291,34 @@ public function testGetReferenceRetrievesReferencesWithGivenProxiesAsIdentifiers
...
+    {
+        $listener = new class($this)
+        {
+            /** @var OrmTestCase */
+            private $testCase;
+
+            /** @var bool This flag is used to prevent infinity loop in case of regression */
```
Ocramius: An infinite loop in case of a regression here is OK, since it would lead to an infinite stack trace, which (with xDebug enabled) would simply lead to a crash.

Ocramius: So all of this can be dropped :+1:

Ocramius (changes requested):

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -349,12 +357,21 @@ public function detach($object)
...
+        }
+
+        $this->flushInProgress = true;
+
         $this->errorIfClosed();
 
         $this->unitOfWork->commit();
+
+        $this->flushInProgress = false;
```
Majkl578: This must go into a `finally` block otherwise it won't be reset in case of exception.

lib/Doctrine/ORM/EntityManager.php
Majkl578: This will need to be a specific exception, not SPL one.
Also please use `()` for the method names.

Majkl578 (changes requested):
I'm not sure whether EntityManager is the proper place for this check - doesn't it rather belong to UoW?

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -145,6 +146,13 @@
     /** @var Cache The second level cache regions API. */
     private $cache;
 
+    /**
+     * Flag shows that flush in progress
```
greg0ire: Either "This flag shows that flush is in progress" or "Flag that shows that flush is in progress"… or maybe nothing, this comment is a bit tautological

tests/Doctrine/Tests/ORM/EntityManagerTest.php
```diff
@@ -288,4 +291,34 @@ public function testGetReferenceRetrievesReferencesWithGivenProxiesAsIdentifiers
...
+
+            public function __construct(OrmTestCase $testCase)
+            {
+                $this->testCase = $testCase;
+            }
+
+            public function preFlush(PreFlushEventArgs $eventArgs)
+            {
+                if ($this->alreadyInPreFlush) {
+                    $this->testCase->fail('Flush inside flush called. It should not happened. Fail test to prevent infinity loop');
```
greg0ire: This line is too long, let's use linebreaks.
```suggestion
                    $this->testCase->fail(
                        'Flush inside flush called. It should not happened. Fail test to prevent infinity loop'
                    );
```

greg0ire (changes requested):

aivus:
> I'm not sure whether EntityManager is the proper place for this check - doesn't it rather belong to UoW?

I thought about this also.
Because now it still possible to use `$em->getUnitOfWork()->commit();` in lifecycle callbacks without any doubts.

lib/Doctrine/ORM/EntityManager.php
aivus: * Added info to `UPGRADE.md`
* Custom exception used. Let me know if we need to rename it or reuse `NotSupported` instead.
* Added `()` to the method defitions

tests/Doctrine/Tests/ORM/EntityManagerTest.php
aivus: done, used new one but can reuse any existing

lib/Doctrine/ORM/EntityManager.php
aivus: Good catch. 👍  Fixed and covered by test.

lib/Doctrine/ORM/EntityManager.php
greg0ire: As you wish, I don't know what the current stance is towards this :)

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -145,6 +146,13 @@
     /** @var Cache The second level cache regions API. */
     private $cache;
 
+    /**
+     * This flag shows that flush is in progress
+     *
+     * @var bool
+     */
+    private $flushInProgress = false;
```
Ocramius: As discussed in https://github.com/doctrine/doctrine2/pull/7448#pullrequestreview-173711503, this should be moved into the `UnitOfWork`.

Ocramius (changes requested):

lib/Doctrine/ORM/EntityManager.php
lcobucci: Please also adapt the naming, we don't "flush" in the UoW but rather "commit"

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
```diff
@@ -707,6 +711,48 @@ public function testCanInstantiateInternalPhpClassSubclassFromUnserializedMetada
...
+
+        $this->eventManager->removeEventListener(Events::preFlush, $listener);
+
+        try {
+            $this->unitOfWork->commit();
+        } catch (CommitInsideCommit $e) {
+            $this->fail('UnitOfWork::$commitInProgress flag must be reset in case of exception during commit process');
+        }
+
+        $this->assertTrue(true);
```
greg0ire: Interesting thread about this: https://github.com/sebastianbergmann/phpunit-documentation/issues/171

greg0ire (dismissed):

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
Ocramius: A better approach to this would be to expect the listener to be called a certain number of times: that would remove the need for the assertion

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
aivus: I'll add a counter inside the listener class and check it value here.

aivus: I've added listener call counter and check, but it looks overcomplicated. Not sure that this is better than solution from sebastianbergmann/phpunit-documentation#171 or `assertTrue(true)`

https://github.com/doctrine/doctrine2/pull/7448/commits/0b3427bcc245c468f48878a0d70dd3c0a11069c9

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
```diff
@@ -729,10 +729,19 @@ public function preFlush(PreFlushEventArgs $eventArgs)
 
     public function testCommitWithExceptionClearsInProgressFlag() : void
     {
-        $listener = new class()
+        $listenerCallCounter = 0;
+        $listener = new class($listenerCallCounter)
         {
+            private $listenerCallCounter;
+
+            public function __construct(int &$listenerCallCounter)
```
Majkl578: Please use public property instead of a reference.

Majkl578 (changes requested):

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
aivus: done

greg0ire (dismissed):

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
```diff
@@ -707,6 +711,50 @@ public function testCanInstantiateInternalPhpClassSubclassFromUnserializedMetada
...
+
+        $this->eventManager->removeEventListener(Events::preFlush, $listener);
+
+        try {
+            $this->unitOfWork->commit();
+        } catch (CommitInsideCommit $e) {
+            $this->fail('UnitOfWork::$commitInProgress flag must be reset in case of exception during commit process');
+        }
+
+        $this->assertEquals(1, $listener->listenerCallCounter);
```
SenseException: Please use the static call like in the other tests and use `self::assertSame()` instead of `$this->assertEquals()`.

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
```diff
@@ -707,6 +711,50 @@ public function testCanInstantiateInternalPhpClassSubclassFromUnserializedMetada
...
+            $this->unitOfWork->commit();
+        } catch (RuntimeException $e) {
+        }
+
+        $this->eventManager->removeEventListener(Events::preFlush, $listener);
+
+        try {
+            $this->unitOfWork->commit();
+        } catch (CommitInsideCommit $e) {
+            $this->fail('UnitOfWork::$commitInProgress flag must be reset in case of exception during commit process');
```
SenseException: Please use the static call of this method `self::fail()`

SenseException (changes requested):

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
SenseException: `self::assertSame()`

aivus: What is the difference in case when we are comparing scalars?

SenseException: `self::assertSame()` is stricter than `self::assertEquals()` and preferable. Similar to `===` compared with `==`.

aivus: I know difference but there is no any for scalars.

As I see `assertEquals` is ok for scalars in existing code:
https://github.com/doctrine/orm/search?q=assertEquals&unscoped_q=assertEquals

SenseException: Scalar values should use `self::assertSame()` too, if there's no reason for a less stricter check.

Thank you for showing the other occurrences in existing tests. Those can be handled with a stricter check in another PR.

aivus:
Hey @greg0ire @beberlei 

Any thoughts about this?

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
```diff
@@ -705,6 +709,50 @@ public function testCanInstantiateInternalPhpClassSubclassFromUnserializedMetada
...
+        $listener = new class()
+        {
+            public function preFlush(PreFlushEventArgs $eventArgs)
+            {
+                $eventArgs->getEntityManager()->getUnitOfWork()->commit();
+            }
+        };
+
+        $this->expectException(CommitInsideCommit::class);
+        $this->eventManager->addEventListener(Events::preFlush, $listener);
```
greg0ire, line 724: It's a best practice to move the `expectException` calls right before the calls that throw the exception.

```suggestion
        $this->eventManager->addEventListener(Events::preFlush, $listener);
        $this->expectException(CommitInsideCommit::class);
```

greg0ire:
It looks like you have a git conflict now.

aivus:
Conflict has been resolved.

Tests on master branch are red also.

derrabus:
Do we still want to do this? If so, we should add a proper deprecation layer first before we introduce this breaking change. I'm removing this from the 3.0.0 milestone until someone picks up the work.

wadjeroudi:
@derrabus @beberlei @greg0ire This PR seems still relevant to me. Is it time to add a deprecate documentation to use flush inside flush ? Or will you change the whole approach ?
I still know a lot of devs who are unaware of the side effects and a lot of projects have issues because of that.

Thank you.

derrabus:
Please feel free to submit a PR for a deprecation.

PR #7449, opened 2018-10-30
Fix build: Disable test assertion upon a non-mockable ClassMetadataFactory method

Majkl578:
`ClassMetadataFactory::addDefaultDiscriminatorMap()` is a private method and as such  is not mockable. Older PHPUnit versions ignored this.

This is a BC break in PHPUnit 7.4: https://github.com/sebastianbergmann/phpunit/blob/7.4.0/ChangeLog-7.4.md#740---2018-10-05

PR #7451, opened 2018-10-31
Fix failing tests. Remove useless mock

aivus:
- this mock didn't work well before because phpunit didn't mock method addDefaultDiscriminatorMap.
 - since phpunit 7.4.0 it throws warning in case of mocking method which is not really mocked (sebastianbergmann/phpunit#3127)


Majkl578:
See  #7449.

PR #7457, opened 2018-11-02
Drop automatic discriminator map discovery

### LINKED ISSUE #7271: automatic discriminator discovery not namespace aware
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | latest

#### Current behavior

Right now there has been one PR closed, and 2 still open for a `@DiscriminatorValue` since the entire discriminator map currently needs to go into the base class, which means extending classes in a different package will not work with it.

Currently we have an automatic discriminator map, just by removing the annotation. However, two classes with the same name regardless of the namespace will not work.

#### How to reproduce

```php
<?php

namespace Del\Entity;

/**
 * @ORM\Entity(repositoryClass="Del\Repository\UserRepository")
 * @ORM\Table(name="User",uniqueConstraints={@ORM\UniqueConstraint(name="email_idx", columns={"email"})})
 * @ORM\DiscriminatorColumn(name="class", type="string")
 */
class User 
{
 // code
}
```
then try and extend it with another User class in a different namespace:

```php
<?php

namespace OAuth;

use Del\Entity\User as BaseUser;
use League\OAuth2\Server\Entities\UserEntityInterface;

/**
 * @ORM\Entity(repositoryClass="OAuth\Repository\UserRepository")
 */
class User extends BaseUser implements UserEntityInterface
{
  // code
}
```
#### Expected behavior
Migrations will generate.

### LINKED ISSUE #7294: Drop automatic discriminator map discovery in 3.0?
### Proposal

|    Q        |   A
|------------ | ------
| RFC         | yes
| BC Break    | yes

#### Summary

Metadata factory is currently performing an auto-discovery of discriminator members. This is often problematic and leads to multiple serious issues in long-run:
1) since discovered entries are not namespaced, there is high probability of collision i.e. #7271),
2) since class name is part of the discriminator map, the class must never be renamed.

We should deprecate and remove this behavior and replace it by `@DiscriminatorValue`: #6612

Majkl578:
Metadata factory is currently performing an auto-discovery of discriminator members. This is often problematic and leads to multiple serious issues in long-run:

* since discovered entries are not namespaced, there is high probability of collision i.e. #7271),
* since class name is part of the discriminator map, the class must never be renamed.

#6612 will be a better and more reliable way of building _dynamic_ discriminator maps.

closes #7294
closes #7271

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,20 @@
 # Upgrade to 3.0
 
+## BC Break: Dropped automatic discriminator map discovery
+
+Automatic discriminator map discovery exhibited multiple flaws
+that can't be reliably addressed and supported:
+
+ * discovered entries are not namespaced which leads to collisions,
+ * the class name is part of the discriminator map, therefore the class
+   must never be renamed.
+
+As a consequence this feature has been dropped.
+
+If your code relied on this feature, please build the discriminator map for
+your inheritance tree manually where each entry is an unqualified lowercase
+name of the member entities.
+
 ## BC Break: Missing type declaration added for identifier generators
 
 The interfaces `Doctrine\ORM\Sequencing\Generator` and
--- a/docs/en/reference/inheritance-mapping.rst
+++ b/docs/en/reference/inheritance-mapping.rst
@@ -131,9 +131,7 @@ Things to note:
    be fully qualified if the classes are contained in the same
    namespace as the entity class on which the discriminator map is
    applied.
--  If no discriminator map is provided, then the map is generated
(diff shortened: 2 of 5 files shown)


Majkl578:
CS failure unrelated

Ocramius (approved):
:ship:

jremmurd:
The mentioned PR #6612 is still not merged, what's the recommeded way to inherit from an entity which parent's definition cannot be changed? For an example use case please see https://stackoverflow.com/questions/58699076/how-to-extend-symfony-entity-in-bundle

PR #7471, opened 2018-11-13
Fix parameter value processing for objects with unloaded metadata

alcaeus:
When binding an object as query parameter, the query object normally tries to convert this object into its identifier. This however only happened if class metadata for the object was already loaded. If it wasn't loaded, conversion was skipped leading to wrong results. While `getMetadataFor` is more performance sensitive than `hasMetadataFor`, we may not assumed that metadata for the bound object has been loaded yet.

The code has been changed to no longer check the metadata factory and call `UnitOfWork::getSingleIdentifierValue` directly, ignoring any mapping exception in the process. This preserves behaviour where binding any unmapped object to a query will send the entire object to the database.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -19,10 +19,12 @@
 
 namespace Doctrine\ORM;
 
+use Doctrine\Common\Persistence\Mapping\MappingException;
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Collections\ArrayCollection;
 
+use Doctrine\ORM\Mapping\MappingException as ORMMappingException;
 use Doctrine\ORM\Query\Parameter;
 use Doctrine\ORM\Cache\QueryCacheKey;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
@@ -410,16 +412,24 @@ abstract class AbstractQuery
             return $value;
         }
 
-        if (is_object($value) && $this->_em->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
+        if ($value instanceof Mapping\ClassMetadata) {
+            return $value->name;
+        }
+
+        if (! is_object($value)) {
+            return $value;
+        }
+
+        try {
             $value = $this->_em->getUnitOfWork()->getSingleIdentifierValue($value);
 
             if ($value === null) {
                 throw ORMInvalidArgumentException::invalidIdentifierBindingEntity();
             }
-        }
-
-        if ($value instanceof Mapping\ClassMetadata) {
-            return $value->name;
+        } catch (MappingException | ORMMappingException $e) {
+            // Silence any mapping exceptions. These can occur if the object in
+            // question is not a mapped entity, in which case we just don't do
+            // any preparation on the value.
         }
 
         return $value;
--- a/tests/Doctrine/Tests/ORM/Query/QueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/QueryTest.php
@@ -189,6 +189,25 @@ class QueryTest extends OrmTestCase
         );
     }
 
+    public function testProcessParameterValueObject() : void
+    {
+        $query = $this->_em->createQuery('SELECT a FROM Doctrine\Tests\Models\CMS\CmsAddress a WHERE a.user = :user');
+        $user = new CmsUser();
+        $user->id = 12345;
+
+        self::assertSame(
+            12345,
+            $query->processParameterValue($user)
+        );
+    }
+
+    public function testProcessParameterValueNull() : void
+    {
+        $query = $this->_em->createQuery('SELECT a FROM Doctrine\Tests\Models\CMS\CmsAddress a WHERE a.user = :user');
+
+        self::assertNull($query->processParameterValue(null));
+    }
+
     public function testDefaultQueryHints()
     {
         $config = $this->_em->getConfiguration();


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -410,16 +412,21 @@ public function processParameterValue($value)
...
             $value = $this->_em->getUnitOfWork()->getSingleIdentifierValue($value);
 
             if ($value === null) {
                 throw ORMInvalidArgumentException::invalidIdentifierBindingEntity();
             }
-        }
-
-        if ($value instanceof Mapping\ClassMetadata) {
-            return $value->name;
+        } catch (MappingException|ORMMappingException $e) {
```
lcobucci: Can you please add a comment explaining why these exceptions are ignored?

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -410,16 +412,21 @@ public function processParameterValue($value)
             return $value;
         }
 
-        if (is_object($value) && $this->_em->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
+        if ($value instanceof Mapping\ClassMetadata) {
+            return $value->name;
+        }
+
+        if ( ! is_object($value)) {
+            return $value;
```
lcobucci: It seems like this path is not being covered by tests, can you please add one?

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -189,6 +189,18 @@ public function testProcessParameterValueClassMetadata()
         );
     }
 
+    public function testProcessParameterValueObject()
```
lcobucci: (nitpick) ` : void` missing

lcobucci (changes requested):
Just some tiny things :+1:

lcobucci:
This was the PHPBench report but it's not much relevant since that bit is not covered:

```diff
  24 subjects, 240 iterations, 24 revs, 0 rejects, 0 failures, 0 warnings
- (best [mean mode] worst) = 370.000 [39,908.388 39,463.004] 450.000 (μs)
+ (best [mean mode] worst) = 361.000 [39,836.125 39,382.889] 563.000 (μs)
- ⅀T: 9,578,013.000μs μSD/r 1,161.745μs μRSD/r: 5.875%
+ ⅀T: 9,560,670.000μs μSD/r 1,335.070μs μRSD/r: 5.475%
- suite: 133f079af8e76a725b3ce35962ff28c7851c24f3, date: 2018-11-13, stime: 10:34:20
+ suite: 133f079e185b4011810cb48680765ebf38c321c3, date: 2018-11-13, stime: 10:36:57
  +-----------------------------------------------------------+-----------------------------------------------------+-------------+---------------+---------------+---------------+---------------+-------------+--------+---------+
  | benchmark                                                 | subject                                             | mem_peak    | best          | mean          | mode          | worst         | stdev       | rstdev | diff    |
  +-----------------------------------------------------------+-----------------------------------------------------+-------------+---------------+---------------+---------------+---------------+-------------+--------+---------+
- | SimpleInsertPerformanceBench                              | benchHydration                                      | 20,343,704b | 262,975.000μs | 271,148.300μs | 270,203.227μs | 276,966.000μs | 4,075.729μs | 1.50%  | 667.85x |
+ | SimpleInsertPerformanceBench                              | benchHydration                                      | 20,343,704b | 269,992.000μs | 278,530.300μs | 275,499.687μs | 292,689.000μs | 7,040.146μs | 2.53%  | 677.19x |
- | SimpleHydrationBench                                      | benchHydration                                      | 67,667,720b | 132,631.000μs | 136,102.900μs | 134,341.123μs | 141,458.000μs | 2,834.684μs | 2.08%  | 335.23x |
+ | SimpleHydrationBench                                      | benchHydration                                      | 67,667,720b | 133,249.000μs | 142,360.300μs | 141,051.824μs | 158,326.000μs | 6,311.889μs | 4.43%  | 346.12x |
- | MixedQueryFetchJoinFullObjectHydrationPerformanceBench    | benchHydration                                      | 17,732,512b | 43,909.000μs  | 45,938.900μs  | 44,994.566μs  | 48,775.000μs  | 1,657.508μs | 3.61%  | 113.15x |
+ | MixedQueryFetchJoinFullObjectHydrationPerformanceBench    | benchHydration                                      | 17,732,408b | 43,624.000μs  | 46,903.700μs  | 45,546.769μs  | 52,319.000μs  | 2,578.879μs | 5.50%  | 114.04x |
- | SimpleQueryPartialObjectHydrationPerformanceBench         | benchHydration                                      | 25,305,536b | 45,104.000μs  | 47,931.300μs  | 46,234.548μs  | 56,894.000μs  | 3,508.759μs | 7.32%  | 118.06x |
+ | SimpleQueryPartialObjectHydrationPerformanceBench         | benchHydration                                      | 25,305,432b | 46,920.000μs  | 49,631.300μs  | 49,180.235μs  | 53,714.000μs  | 1,779.756μs | 3.59%  | 120.67x |
- | SimpleQueryScalarHydrationPerformanceBench                | benchHydration                                      | 14,337,104b | 15,268.000μs  | 16,512.400μs  | 16,114.051μs  | 18,995.000μs  | 1,040.037μs | 6.30%  | 40.67x  |
+ | SimpleQueryScalarHydrationPerformanceBench                | benchHydration                                      | 14,337,104b | 14,850.000μs  | 16,111.500μs  | 15,718.751μs  | 20,012.000μs  | 1,390.834μs | 8.63%  | 39.17x  |
- | SingleTableInheritanceHydrationPerformanceBench           | benchHydrateFixContracts                            | 6,363,736b  | 1,718.000μs   | 1,784.700μs   | 1,754.564μs   | 1,891.000μs   | 51.852μs    | 2.91%  | 4.40x   |
+ | SingleTableInheritanceHydrationPerformanceBench           | benchHydrateFixContracts                            | 6,363,736b  | 1,724.000μs   | 1,869.800μs   | 1,851.573μs   | 2,134.000μs   | 109.495μs   | 5.86%  | 4.55x   |
- | SingleTableInheritanceHydrationPerformanceBench           | benchHydrateFlexContracts                           | 6,501,880b  | 2,107.000μs   | 2,282.500μs   | 2,161.462μs   | 3,372.000μs   | 364.587μs   | 15.97% | 5.62x   |
+ | SingleTableInheritanceHydrationPerformanceBench           | benchHydrateFlexContracts                           | 6,501,776b  | 2,110.000μs   | 2,326.600μs   | 2,176.536μs   | 2,960.000μs   | 270.234μs   | 11.61% | 5.66x   |
- | SingleTableInheritanceHydrationPerformanceBench           | benchHydrateUltraContracts                          | 6,368,440b  | 1,843.000μs   | 1,976.300μs   | 1,917.980μs   | 2,238.000μs   | 129.170μs   | 6.54%  | 4.87x   |
+ | SingleTableInheritanceHydrationPerformanceBench           | benchHydrateUltraContracts                          | 6,368,336b  | 1,796.000μs   | 1,842.900μs   | 1,813.178μs   | 1,929.000μs   | 43.441μs    | 2.36%  | 4.48x   |
- | SingleTableInheritanceHydrationPerformanceBench           | benchHydrateAllContracts                            | 6,588,232b  | 2,499.000μs   | 2,682.500μs   | 2,616.534μs   | 3,269.000μs   | 208.184μs   | 7.76%  | 6.61x   |
+ | SingleTableInheritanceHydrationPerformanceBench           | benchHydrateAllContracts                            | 6,588,128b  | 2,532.000μs   | 2,769.000μs   | 2,621.636μs   | 3,226.000μs   | 231.238μs   | 8.35%  | 6.73x   |
- | SingleTableInheritanceInsertPerformanceBench              | benchInsertFixContracts                             | 5,937,680b  | 3,629.000μs   | 3,862.100μs   | 3,764.526μs   | 4,746.000μs   | 312.493μs   | 8.09%  | 9.51x   |
+ | SingleTableInheritanceInsertPerformanceBench              | benchInsertFixContracts                             | 5,937,680b  | 3,560.000μs   | 3,767.100μs   | 3,728.082μs   | 3,969.000μs   | 118.555μs   | 3.15%  | 9.16x   |
- | SingleTableInheritanceInsertPerformanceBench              | benchInsertFlexContracts                            | 5,950,472b  | 3,829.000μs   | 4,250.100μs   | 3,996.123μs   | 5,537.000μs   | 531.776μs   | 12.51% | 10.47x  |
+ | SingleTableInheritanceInsertPerformanceBench              | benchInsertFlexContracts                            | 5,950,472b  | 3,644.000μs   | 3,787.300μs   | 3,743.315μs   | 3,934.000μs   | 87.044μs    | 2.30%  | 9.21x   |
- | SingleTableInheritanceInsertPerformanceBench              | benchInsertUltraContracts                           | 5,996,000b  | 3,885.000μs   | 4,271.500μs   | 4,079.004μs   | 5,434.000μs   | 444.057μs   | 10.40% | 10.52x  |
+ | SingleTableInheritanceInsertPerformanceBench              | benchInsertUltraContracts                           | 5,996,000b  | 3,660.000μs   | 3,836.700μs   | 3,728.689μs   | 4,128.000μs   | 145.543μs   | 3.79%  | 9.33x   |
- | SingleTableInheritanceInsertPerformanceBench              | benchInsertAllContracts                             | 6,342,696b  | 5,409.000μs   | 5,712.300μs   | 5,597.472μs   | 6,192.000μs   | 254.228μs   | 4.45%  | 14.07x  |
+ | SingleTableInheritanceInsertPerformanceBench              | benchInsertAllContracts                             | 6,342,696b  | 4,952.000μs   | 5,249.600μs   | 5,084.094μs   | 6,202.000μs   | 369.274μs   | 7.03%  | 12.76x  |
- | SimpleQueryArrayHydrationPerformanceBench                 | benchHydration                                      | 14,071,568b | 24,375.000μs  | 26,385.600μs  | 25,465.411μs  | 32,335.000μs  | 2,298.519μs | 8.71%  | 64.99x  |
+ | SimpleQueryArrayHydrationPerformanceBench                 | benchHydration                                      | 14,071,568b | 23,559.000μs  | 24,327.500μs  | 23,941.935μs  | 26,005.000μs  | 749.271μs   | 3.08%  | 59.15x  |
- | MixedQueryFetchJoinPartialObjectHydrationPerformanceBench | benchHydration                                      | 11,838,584b | 25,009.000μs  | 26,833.900μs  | 26,635.149μs  | 28,892.000μs  | 1,032.904μs | 3.85%  | 66.09x  |
+ | MixedQueryFetchJoinPartialObjectHydrationPerformanceBench | benchHydration                                      | 11,838,480b | 25,895.000μs  | 26,653.500μs  | 26,537.241μs  | 28,190.000μs  | 574.283μs   | 2.15%  | 64.80x  |
- | SimpleQueryFullObjectHydrationPerformanceBench            | benchHydration                                      | 62,116,592b | 145,482.000μs | 149,224.200μs | 148,192.372μs | 152,407.000μs | 2,294.034μs | 1.54%  | 367.55x |
+ | SimpleQueryFullObjectHydrationPerformanceBench            | benchHydration                                      | 62,116,488b | 138,155.000μs | 143,178.600μs | 142,548.703μs | 148,454.000μs | 3,421.227μs | 2.39%  | 348.11x |
- | MixedQueryFetchJoinArrayHydrationPerformanceBench         | benchHydration                                      | 31,152,520b | 47,166.000μs  | 48,191.100μs  | 47,810.384μs  | 49,910.000μs  | 861.185μs   | 1.79%  | 118.70x |
+ | MixedQueryFetchJoinArrayHydrationPerformanceBench         | benchHydration                                      | 31,152,520b | 42,573.000μs  | 44,350.600μs  | 43,558.933μs  | 50,699.000μs  | 2,271.497μs | 5.12%  | 107.83x |
- | UnitOfWorkComputeChangesBench                             | benchChangeSetComputation                           | 4,891,920b  | 793.000μs     | 819.800μs     | 805.945μs     | 856.000μs     | 23.151μs    | 2.82%  | 2.02x   |
+ | UnitOfWorkComputeChangesBench                             | benchChangeSetComputation                           | 4,891,920b  | 721.000μs     | 790.400μs     | 774.605μs     | 977.000μs     | 69.230μs    | 8.76%  | 1.92x   |
- | ProxyInitializationTimeBench                              | benchCmsUserInitialization                          | 16,826,024b | 35,593.000μs  | 36,650.200μs  | 36,714.742μs  | 37,523.000μs  | 484.413μs   | 1.32%  | 90.27x  |
+ | ProxyInitializationTimeBench                              | benchCmsUserInitialization                          | 16,826,024b | 31,842.000μs  | 33,874.300μs  | 33,847.315μs  | 36,278.000μs  | 1,169.117μs | 3.45%  | 82.36x  |
- | ProxyInitializationTimeBench                              | benchCmsEmployeeInitialization                      | 16,826,024b | 5,880.000μs   | 6,717.000μs   | 6,852.634μs   | 7,296.000μs   | 389.934μs   | 5.81%  | 16.54x  |
+ | ProxyInitializationTimeBench                              | benchCmsEmployeeInitialization                      | 16,826,024b | 6,034.000μs   | 6,536.100μs   | 6,271.213μs   | 7,928.000μs   | 554.531μs   | 8.48%  | 15.89x  |
- | ProxyInitializationTimeBench                              | benchInitializationOfAlreadyInitializedCmsUsers     | 16,826,048b | 441.000μs     | 492.800μs     | 465.845μs     | 625.000μs     | 54.834μs    | 11.13% | 1.21x   |
+ | ProxyInitializationTimeBench                              | benchInitializationOfAlreadyInitializedCmsUsers     | 16,826,048b | 456.000μs     | 508.600μs     | 506.673μs     | 563.000μs     | 29.114μs    | 5.72%  | 1.24x   |
- | ProxyInitializationTimeBench                              | benchInitializationOfAlreadyInitializedCmsEmployees | 16,826,048b | 370.000μs     | 406.000μs     | 391.292μs     | 450.000μs     | 26.866μs    | 6.62%  | 1.00x   |
+ | ProxyInitializationTimeBench                              | benchInitializationOfAlreadyInitializedCmsEmployees | 16,826,048b | 361.000μs     | 411.300μs     | 386.550μs     | 633.000μs     | 75.777μs    | 18.42% | 1.00x   |
- | ProxyInstantiationTimeBench                               | benchCmsUserInstantiation                           | 4,828,112b  | 65,987.000μs  | 68,989.800μs  | 67,795.219μs  | 79,987.000μs  | 3,798.960μs | 5.51%  | 169.93x |
+ | ProxyInstantiationTimeBench                               | benchCmsUserInstantiation                           | 4,828,112b  | 65,510.000μs  | 66,710.400μs  | 65,979.785μs  | 69,511.000μs  | 1,285.225μs | 1.93%  | 162.19x |
- | ProxyInstantiationTimeBench                               | benchCmsEmployeeInstantiation                       | 4,558,264b  | 47,301.000μs  | 48,635.100μs  | 48,207.912μs  | 51,673.000μs  | 1,204.027μs | 2.48%  | 119.79x |
+ | ProxyInstantiationTimeBench                               | benchCmsEmployeeInstantiation                       | 4,558,264b  | 47,777.000μs  | 49,739.600μs  | 49,092.022μs  | 51,925.000μs  | 1,366.084μs | 2.75%  | 120.93x |
  +-----------------------------------------------------------+-----------------------------------------------------+-------------+---------------+---------------+---------------+---------------+-------------+--------+---------+
```

Ocramius (dismissed):
Besides existing feedback, this looks good! :+1:

alcaeus:
Build failures are due to MySQL not being installed at the moment...

lib/Doctrine/ORM/AbstractQuery.php
lcobucci: I know, it looks much better now. However, in the previous algorithm this scenario was implicitly covered and now it's not.

image

lcobucci: @alcaeus yes and that's exactly why I'm asking for one extra unit test to consider this PR ready to go. We're doing a lot of refactoring for v3 and it would be a shame if the this scenario doesn't work anymore...

alcaeus: Final test pushed.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -410,16 +412,24 @@ public function processParameterValue($value)
             return $value;
         }
 
-        if (is_object($value) && $this->_em->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
+        if ($value instanceof Mapping\ClassMetadata) {
+            return $value->name;
+        }
+
+        if ( ! is_object($value)) {
```
Majkl578:
```suggestion
        if (! is_object($value)) {
```

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -189,6 +189,25 @@ public function testProcessParameterValueClassMetadata()
         );
     }
 
+    public function testProcessParameterValueObject(): void
```
Majkl578:
```suggestion
    public function testProcessParameterValueObject() : void
```

tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -189,6 +189,25 @@ public function testProcessParameterValueClassMetadata()
...
+        $user = new CmsUser();
+        $user->id = 12345;
+
+        self::assertSame(
+            12345,
+            $query->processParameterValue($user)
+        );
+    }
+
+    public function testProcessParameterValueNull(): void
```
Majkl578:
```suggestion
    public function testProcessParameterValueNull() : void
```

lib/Doctrine/ORM/AbstractQuery.php
Majkl578:
```suggestion
        } catch (MappingException | ORMMappingException $e) {
```

lib/Doctrine/ORM/AbstractQuery.php
alcaeus: Take a look around the code: the old Doctrine standard (which is still being used in this branch) uses a space before the negation operator.

lib/Doctrine/ORM/AbstractQuery.php
alcaeus: @Majkl578 we currently don't have a rule for that in Doctrine CS, so I'd like to have that discussion in `doctrine/coding-standard` before committing to it.

lib/Doctrine/ORM/AbstractQuery.php
carusogabriel: Yeah, our new CS is only for `master`.

lib/Doctrine/ORM/AbstractQuery.php
carusogabriel: @Majkl578 Do you know any Sniff for that? I agree with @alcaeus, we should discuss it on the repo.

lcobucci (approved):
@alcaeus this is perfect now, I've addressed the minor CS things and we're good to go. Thanks for squashing this :bug:

lcobucci:
@alcaeus :ship:

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -410,16 +412,24 @@ public function processParameterValue($value)
             return $value;
         }
 
-        if (is_object($value) && $this->_em->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
+        if ($value instanceof Mapping\ClassMetadata) {
```
flaushi: Just found out via kcachegrind that this caused massive performance regressions when binding any non-scalar value as parameter.

For example:

```php
$thing = new DateTime();

$query->setParameter('foo', $thing);
$query->getResult();
```

Regardless of `$query`, this will lead to an internal class metadata lookup (and internal failure), plus exceptions being allocated and GC'd.

The following won't help either:

```php
$thing = new DateTime();

$query->setParameter('foo', $thing, DateTimeType::NAME);
$query->getResult();
```

That's because `processParameterValue($value)` is called even when the type is known.

@alcaeus @Ocramius would it be a good idea to skip the entire `processParameterValue()`when the type is explicitly passed in?

Ocramius: Moved to https://github.com/doctrine/doctrine2/issues/7527

PR #7488, opened 2018-11-26
Update DBAL 3.x version from 2018-04-17 to 2018-11-26

Majkl578:
Some BC breaks happened in DBAL 3.0.x-dev in meantime: https://github.com/doctrine/dbal/blob/snapshot/develop/2018-11-26/UPGRADE.md#upgrade-to-30, these have been fíxed on our side, notably:
- `NULL` for `$offset` in `LIMIT`
- removed `dbal:import`
- relying on `Type::__toString()`)

This is also pre-requisite for dropping doctrine/common.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
@@ -856,7 +856,7 @@ class ManyToManyPersister extends AbstractCollectionPersister
         $limit  = $criteria->getMaxResults();
         $offset = $criteria->getFirstResult();
         if ($limit !== null || $offset !== null) {
-            return $this->platform->modifyLimitQuery('', $limit, $offset);
+            return $this->platform->modifyLimitQuery('', $limit, $offset ?? 0);
         }
         return '';
     }
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -1165,7 +1165,7 @@ class BasicEntityPersister implements EntityPersister
             . $where
             . $orderBySql;
 
-        return $this->platform->modifyLimitQuery($query, $limit, $offset) . $lockSql;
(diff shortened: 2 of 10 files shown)


lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
```diff
@@ -856,7 +856,7 @@ private function getLimitSql(Criteria $criteria)
         $limit  = $criteria->getMaxResults();
         $offset = $criteria->getFirstResult();
         if ($limit !== null || $offset !== null) {
-            return $this->platform->modifyLimitQuery('', $limit, $offset);
+            return $this->platform->modifyLimitQuery('', $limit, $offset ?? 0);
```
morozov: Should `$criteria->getFirstResult()` return a non-nullable integer instead?

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1165,7 +1165,7 @@ public function getSelectSQL(
             . $where
             . $orderBySql;
 
-        return $this->platform->modifyLimitQuery($query, $limit, $offset) . $lockSql;
+        return $this->platform->modifyLimitQuery($query, $limit, $offset ?? 0) . $lockSql;
```
morozov: Should `$offset` have the default value of `0` instead of `null`?

lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
```diff
@@ -259,7 +259,7 @@ public function getSelectSQL(
                     . $where
                     . $orderBySql;
 
-        return $this->platform->modifyLimitQuery($query, $limit, $offset) . $lockSql;
+        return $this->platform->modifyLimitQuery($query, $limit, $offset ?? 0) . $lockSql;
```
morozov: Should `$offset` have the default value of `0` instead of `null`?

Majkl578:
@morozov Yes to all your comments, I just didn't want to break BC in this PR (later).

Ocramius (approved):
:ship: as per https://github.com/doctrine/doctrine2/pull/7488#issuecomment-441889207

PR #7493, opened 2018-11-29
Add warning about using CLI tool with APC(u) cache

shadowhand:
Because [this happens][1] and the APC(u) documentation does not help matters.

This PR really isn't enough to sufficiently document the "gotchas" when using APC(u) cache but hopefully it gets the conversation started.

[1]: https://twitter.com/shadowhand/status/1068192392761622529

### DIFF
--- a/docs/en/reference/caching.rst
+++ b/docs/en/reference/caching.rst
@@ -400,6 +400,16 @@ cache drivers to manually delete cache entries. For your
 convenience we offer command line tasks to help you with
 clearing the query, result and metadata cache.
 
+.. note::
+
+    Because APCu caches are stored in a memory segment owned by a php
+    process and not shared with other php processes, it is not possible
+    to clear the APCu cache of the webserver from the CLI context. The
+    following commands will not work as intended and will clear a
+    separate APCu cache that is owned by the PHP CLI process, provided
+    ``apcu.enable_cli`` is set to 1, `which it probably should not
+    <https://php.net/manual/en/apcu.configuration.php#ini.apcu.enable-cli>`_.
+
 From the Doctrine command line you can run the following commands:
 
 To clear the query cache use the ``orm:clear-cache:query`` task.


greg0ire:
You link to a tweet saying:

> Today I learned that you cannot flush the #Doctrine metadata cache from the CLI when using #APCu cache. Totally not obvious anywhere in the APCu or Doctrine documentation. #FML

So "you cannot"… can you please describe a bit more what happens exactly?

shadowhand:
@greg0ire the APC(u) cache exists in memory shared owned by a single PHP process. In my case, that process is owned by Apache. Running `vendor/bin/doctrine orm:clear-cache:metadata` does not error but it will not clear the cache that Apache uses.

greg0ire:
Thanks for the explanation, do you get an exception, or some CLI warning?

docs/en/reference/caching.rst
```diff
@@ -395,6 +395,9 @@ cache drivers to manually delete cache entries. For your
 convenience we offer command line tasks to help you with
 clearing the query, result and metadata cache.
 
+**NOTE: It is not possible to use these commands when APC(u)
```
greg0ire:
```suggestion
**NOTE: It is not possible to use these commands when APCu
```
APCu is now the default, APC has been deprecated for a long time.

greg0ire (changes requested):

shadowhand:
@greg0ire no, there is no warning because the APC(u) cache is initialized, assuming that [`apcu.enable-cli`][1] is active, but it contains no data. Doctrine CLI tool doesn't give any warning about this because it assumes there are no cache entries.

[1]: http://php.net/manual/en/apcu.configuration.php#ini.apcu.enable-cli

docs/en/reference/caching.rst
shadowhand: The problem applies to both, even if APC is deprecated.

greg0ire:
Makes sense, technically people could be willing to use a separate APCu cache just for the CLI, and should not get any error when trying to clear it. Note that the doc you linked says that having it enabled is not very likely though:

> Mostly for testing and debugging. Setting this enables APC for the CLI version of PHP. Under normal circumstances, it is not ideal to create, populate and destroy the APC cache on every CLI request, but for various test scenarios it is useful to be able to enable APC for the CLI version of PHP easily.

shadowhand:
@greg0ire right, my PR does not really reflect the complicated nature of what is happening. The APC(u) cache works just fine for CLI and for Apache/PHP-FPM, _**but** the cache is not shared_. This is the point that I want to see reflected in the documentation.

The language in this PR definitely needs to be updated, I am just not sure what the best way to explain this is.

greg0ire:
It could be nice to produce a warning to users that are trying to clear an APCu cache from the cli and have `apcu.enable_cli` set to "0"?

Regarding your message how about:

> It is not possible to clear the APCu cache of the webserver from the CLI context. The following commands will not work as intended and will clear a separate, empty cache that is owned by the PHP CLI process, provided `apcu.enable_cli` is set to `1`, which it probably shouldn't

with a nice link to the php docs.

shadowhand:
Perhaps the documentation should be updated to use `PhpFileCache` instead of `APCu` for metadata? At least then `vendor/bin/doctrine orm:clear-cache:metadata` would work correctly. 

That doesn't solve the overall problem of clearly documenting that APC(u) cache is not shared between CLI and webserver.

greg0ire:
There have been similar attempts on the sf docs, and they were reverted so that the docs stay compatible with read-only filesystems, which are getting traction.

shadowhand:
The problem is exacerbated by the fact that <http://php.net/apcu> says absolutely nothing about how APCu works with regard to visibility of cache between different PHP processes. Obviously that cannot be solved by Doctrine, but I think it is worth mentioning because linking to the PHP docs won't help people understand the root of the issue.

shadowhand:
> There have been similar attempts on the sf docs, and they were reverted so that the docs stay compatible with read-only filesystems, which are getting traction.

I understand that point, but the current documentation defaults imply that using the command line tools to manage cache will work as expected with APCu and that simply isn't true.

shadowhand:
This is (somewhat) related to symfony/symfony#24545

docs/en/reference/caching.rst
greg0ire: Oh right, didn't notice we still have docs about APC, my bad.

greg0ire:
>  linking to the PHP docs won't help people understand the root of the issue.

Ok so let's insist even more then:

> Because APCu caches are stored in a memory segment owned by a php process and not shared with other php processes, it is not possible to clear the APCu cache of the webserver from the CLI context. The following commands will not work as intended and will clear a separate APCu cache that is owned by the PHP CLI process, provided apcu.enable_cli is set to 1, which it probably shouldn't.

shadowhand:
@greg0ire that makes sense. Do I need to make separate PRs to target older versions of documentation?

greg0ire:
Sorry, I'm a bit new to reviewing on that repository. From my understanding, you should create one PR, and (I'm less sure here), target the stable branch (so 2.6 ?), see https://github.com/doctrine/doctrine2/blob/master/CONTRIBUTING.md#choosing-the-branch

docs/en/reference/caching.rst
greg0ire:
```suggestion
.. note::

    Because APCu caches are stored in a memory segment owned by a php
    process and not shared with other php processes, it is not possible
    to clear the APCu cache of the webserver from the CLI context. The
    following commands will not work as intended and will clear a
    separate APCu cache that is owned by the PHP CLI process, provided
    ``apcu.enable_cli`` is set to 1, `which it probably should not
    <https://php.net/manual/en/apcu.configuration.php#ini.apcu.enable-cli>`_.
```

shadowhand:
@greg0ire updated target and applied your wording.

greg0ire (approved):

Ocramius (changes requested):
As we previously did with the APC cache (when it existed), this should be a warning on the CLI command itself, instead of a documentation issue.

I'd also add a few things:

  1. This is doctrine ORM, not the APCu project. We shouldn't explain users how their caching backend works, since they installed it and should (hopefully) have read how it works. If we start nursing them through every quirk of every caching backend, we are moving some unrelated stuff into this project.
 2. APCu is relatively unstable, and really only suiyed for extremely low cache write throughout/concurrency. I would rather warn users about their use of the APCu backend, and steer them to the more efficient PhpFileCache
 3. Could really not care less about the symfony defaults: whole different project again, and we can't let scopes leak constantly.

shadowhand:
@Ocramius 

> this should be a warning on the CLI command itself

Do you mean that `doctrine orm:cache*` commands should have a warning when using the APCu driver?

> We shouldn't explain users how their caching backend works, since they installed it and should (hopefully) have read how it works.

As I mentioned above, there is zero mention of how memory sharing works in <http://php.net/apcu>. I understand your position but surely there is a duty to make it obvious to users that `doctrine` tool won't work as expected with APCu cache?

> I would rather warn users about their use of the APCu backend, and steer them to the more efficient PhpFileCache.

Does this mean that it would be better for the documentation to suggest `PhpFileCache` as the default cache implementation, as I [suggested above][1]?

[1]: https://github.com/doctrine/doctrine2/pull/7493#issuecomment-442951065

Ocramius:
> > this should be a warning on the CLI command itself
> 
> Do you mean that `doctrine orm:cache*` commands should have a warning when using the APCu driver?

Yes

> > We shouldn't explain users how their caching backend works, since they installed it and should (hopefully) have read how it works.
> 
> As I mentioned above, there is zero mention of how memory sharing works in the <http://php.net/apcu> documentation. I understand your position but surely there is a duty to make it obvious to users that `doctrine` tool won't work as expected with APCu cache?

This is a **SEP** (Someone Else's Problem). The fact that upstream docs are lacking in content doesn't mean that anything downstream should fill the gap. The analogy would be to find a MySQL documentation issue, and reporting it here: not OK.

> > I would rather warn users about their use of the APCu backend, and steer them to the more efficient PhpFileCache.
> 
> Does this mean that it would be better for the documentation to suggest `PhpFileCache` as the default cache implementation, as I [suggested above][1]?

Yup 👌

[1]: https://github.com/doctrine/doctrine2/pull/7493#issuecomment-442951065

shadowhand:
@Ocramius thank you for the clarification. I will work on two separate pull requests:

1. Add a warning to `doctrine` tool when using the APC(u) driver.
2. Update ORM cache documentation to use `PhpFileCache` in examples.

Should both of these PRs should target `2.6` branch?

Ocramius:
Yeah, you can target 2.6 👍

greg0ire:
> Could really not care less about the symfony defaults: whole different project again, and we can't let scopes leak constantly.

I was just giving out reasons (namely ro filesystems getting traction) why APCu might still be the default, but yeah, we don't have to be consistent with Symfony at all, of course.

shadowhand:
@greg0ire your point is valid. When working with RO filesystems, it would be possible to generate the metadata cache as part of the build process and upload the cached metadata to the RO filesystem. For most use cases, I think that @Ocramius is right about using PhpFileCache as the default. Opcode caching is so fast in PHP 7+ that it probably outperforms APCu.

shadowhand:
Replaced by #7494 and #7500.

PR #7494, opened 2018-11-30
Do not allow cache clearing with APCu

shadowhand:
Refs https://github.com/doctrine/doctrine2/pull/7493#pullrequestreview-180007007

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,11 @@
 # Upgrade to 3.0
 
+## BC Break: Removed ability to clear cache via console with some cache drivers
+
+The console commands `orm:clear-cache:metadata`, `orm:clear-cache:result`,
+and `orm:clear-cache:query` cannot be used with the `ApcCache`, `ApcuCache`,
+or `XcacheCache` because the memory is only available to the webserver process.
+
 ## BC Break: `orm:run-dql` command's `$depth` parameter removed
 
 The `$depth` parameter has been removed, the dumping functionality
--- a/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/MetadataCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/MetadataCommand.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ORM\Tools\Console\Command\ClearCache;
 
 use Doctrine\Common\Cache\ApcCache;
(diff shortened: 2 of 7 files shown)


greg0ire (changes requested):

shadowhand:
@greg0ire all of the LogicExceptions have the same grammar error... do I fix all of them?

greg0ire:
It's up to you I think, but do not introduce more of them :P

Ocramius (changes requested):
This is a BC break which we need to document in `UPGRADE.md`

shadowhand:
@greg0ire wording for all exceptions updated.

@Ocramius `UPGRADE.md` has been updated.

greg0ire (dismissed):

Ocramius:
We'll need to target `master` here

shadowhand:
@Ocramius doesn't targeting `master` imply version 3.0? I was really hoping to avoid that.

shadowhand:
@Ocramius in my mind this is a bugfix release, since clearing cache with APCu driver did not actually perform the expected operation.

greg0ire:
It was not doing anything, but now it throws, which means every deploy script that calls it will fail under the right circumstances. That makes it a BC-break. Plus the cache is still not cleared, so not a bugfix either I think. It's just a nice DX improvement. It you want it to land on lower branches, maybe make the bugfix a simple call to one of the methods of `OutputInterface` or trigger a deprecation (to be confirmed, not sure what the Doctrine policy would be for this). That wouldn't be a BC-break I think.

Ocramius:
Yes, by producing just a warning, it would not be a BC break

Ocramius:
No, that would be yet another BC break 😜

shadowhand:
@Ocramius okay, and how do you want to handle `master` branch? Separate PR? I'm don't know how the merging from 2.6 to master is done.

Ocramius:
That's done by us later on

shadowhand:
@Ocramius okay, then how about I make this PR against 3.0 and then add a "backport" for 2.6 that replaces the exception with a warning?

Ocramius:
Since you are aiming for an improvement, then just re-target this patch to `master` and call it a day

shadowhand:
@Ocramius rebased to master.

shadowhand:
@Ocramius tests are passing now.

Ocramius (approved):
:+1: thanks @shadowhand!

PR #7497, opened 2018-12-01
Drop dependency on doctrine/common

Majkl578:
Common is no more. \o/

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/RunDqlCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/RunDqlCommand.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Tools\Console\Command;
 
-use Doctrine\Common\Util\Debug;
 use Doctrine\ORM\EntityManagerInterface;
 use LogicException;
 use RuntimeException;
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ORM/Tools/Console/Command/RunDqlCommand.php
```diff
@@ -101,6 +95,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $resultSet = $query->execute([], constant($hydrationMode));
 
-        $ui->text(Debug::dump($resultSet, $input->getOption('depth'), true, false));
+        $dumper = new CliDumper(static function (string $payload) use ($output) : void {
+            $output->write($payload);
+        });
+        $dumper->dump((new VarCloner())->cloneVar($resultSet));
```
Ocramius: This change specifically requires an integration test to be added

Ocramius (changes requested):

lib/Doctrine/ORM/Tools/Console/Command/RunDqlCommand.php
Majkl578: There already is a test, which was failing without this change:
https://github.com/doctrine/doctrine2/blob/master/tests/Doctrine/Tests/ORM/Tools/Console/Command/RunDqlCommandTest.php#L67

Majkl578:
Scrutinizer pointed out one more occurence of ClassLoader in `bin/doctrine-pear.php`, but afaik we don't use PEAR for ages?

Ocramius (approved):
:ship:

PR #7501, opened 2018-12-03
Add __debugInfo to entity manager

carnage:
Not expecting this to go in as is; opened this to start a discussion on what is valuable to dump and what isn't.

I've var_dump'd the entity manager waaaay too many times, this stops it from becoming a huge mess.

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -174,6 +174,11 @@ final class EntityManager implements EntityManagerInterface
         }
     }
 
+    public function __debugInfo()
+    {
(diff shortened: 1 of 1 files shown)


Ocramius:
`var_dump()` is not an operation to be performed in a production environment: it should only be performed when/if debugging, and only with XDebug enabled (which makes it bearable).

Debugging the `UnitOfWork` and `EntityManager` are perfectly valid scenarios that should not be hidden via `__debugInfo()`.

I'd even add that this is precisely the dark side of `__debugInfo()`: it is used to hide away details that are otherwise necessary to the reader.

Closing as `won't fix`.

Issue #7505, opened 2018-12-06
Hydratation error when two classes with the same parent class both define the same property with different types.


m-r-r:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

#### Summary

`SimpleObjectHydrator` produces a type error when two classes with the same parent class both define a property with the same name and different types.

I think this may be related to issue #5989.

#### Current behavior

When two classes inherit from the same parent using `SINGLE_TABLE` inheritance, and both classes define the same property with different types, `SimpleObjectHydrator` produce a type error if one of the properties is of type `simple_array`.

#### How to reproduce

Create an abstract class with two child classes:
```php
use Doctrine\ORM\Annotation as ORM;

/**
 * @ORM\Entity()
 * @ORM\Table(name="test_responses")
 * @ORM\InheritanceType("SINGLE_TABLE")
 * @ORM\DiscriminatorColumn(name="discr", type="string")
 * @ORM\DiscriminatorMap({
 *     "array"       = ArrayResponse::class,
 *     "text"        = TextResponse::class,
 * })
 */
abstract class AbstractResponse {
    /**
     * @ORM\Id @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    public $id;
}
```

Create two child classes. Each class defines the same property with different types.
The first class use the type `simple_array` and the second class has a `string` property :
```php
/**
 * @ORM\Entity()
 */
class ArrayResponse extends AbstractResponse {
    /**
     * @ORM\Column(name="value_array", type="simple_array")
     *
     * @var array
     */
    public $value = [];
}

/**
 * @ORM\Entity()
 */
class TextResponse extends AbstractResponse {
    /**
     * @ORM\Column(name="value_string", type="string")
     *
     * @var string|null
     */
    public $value = null;
}
```

Create an instance of the second class and flush the `EntityManager`:
```php
$textResponse = new TextResponse();
$em->persist($textResponse);
$em->flush();
```

Clear the `EntityManager` and fetch the instance:
```php
$em->clear();

/** @var TextResponse $arrayResponse */
$textResponse = $em->getRepository(AbstractResponse::class)
                   ->find($textResponse->id);

var_dump($textResponse->value); // Display an empty array (type error)
```

#### Expected behavior

The value of `$textResponse->value` should be NULL, because the property `$value` of class `TextResponse` has type `string|null`.

m-r-r:
[...]

PR #7506, opened 2018-12-06
Ensure hydrator uses properties of the correct class when using inheritance

m-r-r:
I added a test case for issue #7505 and fixed the issue.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -78,8 +78,9 @@ class SimpleObjectHydrator extends AbstractHydrator
      */
     protected function hydrateRowData(array $sqlResult, array &$result)
     {
-        $entityName = $this->class->name;
-        $data       = [];
+        $entityName       = $this->class->name;
+        $data             = [];
+        $discrColumnValue = null;
 
         // We need to find the correct entity class name if we have inheritance in resultset
         if ($this->class->inheritanceType !== ClassMetadata::INHERITANCE_TYPE_NONE) {
@@ -104,7 +105,8 @@ class SimpleObjectHydrator extends AbstractHydrator
                 throw HydrationException::invalidDiscriminatorValue($sqlResult[$discrColumnName], array_keys($discrMap));
             }
 
-            $entityName = $discrMap[$sqlResult[$discrColumnName]];
+            $entityName       = $discrMap[$sqlResult[$discrColumnName]];
+            $discrColumnValue = $sqlResult[$discrColumnName];
 
             unset($sqlResult[$discrColumnName]);
         }
@@ -134,6 +136,11 @@ class SimpleObjectHydrator extends AbstractHydrator
 
             // Prevent overwrite in case of inherit classes using same property name (See AbstractHydrator)
             if ( ! isset($data[$fieldName]) || ! $valueIsNull) {
+                // If we have inheritance in resultset, make sure the field belongs to the correct class
+                if (isset($cacheKeyInfo['discriminatorValues']) && ! in_array($discrColumnValue, $cacheKeyInfo['discriminatorValues'], true)) {
+                    continue;
+                }
+
                 $data[$fieldName] = $value;
             }
         }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7505Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7505Test.php
@@ -1,0 +1,92 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Annotation as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH7505Test extends OrmFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function setUp() : void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH7505AbstractResponse::class,
+            GH7505ArrayResponse::class,
+            GH7505TextResponse::class,
+        ]);
+    }
+
+    public function testSimpleArrayTypeHydratedCorrectly() : void
+    {
+        $arrayResponse = new GH7505ArrayResponse();
+        $this->em->persist($arrayResponse);
+
+        $textResponse = new GH7505TextResponse();
+        $this->em->persist($textResponse);
+
+        $this->em->flush();
+        $this->em->clear();
+
+        $repository = $this->em->getRepository(GH7505AbstractResponse::class);
+
+        /** @var GH7505ArrayResponse $arrayResponse */
+        $arrayResponse = $repository->find($arrayResponse->id);
+        self::assertSame([], $arrayResponse->value);
+
+        /** @var GH7505TextResponse $textResponse */
+        $textResponse = $repository->find($textResponse->id);
+        self::assertNull($textResponse->value);
+    }
+}
+
+/**
+ * @ORM\Entity()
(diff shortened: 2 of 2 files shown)


Ocramius (changes requested):
This is not a valid schema: please make sure that the schema validator reports an invalid schema due to different types on the same single inheritance column

tests/Doctrine/Tests/Models/GH7505/TextResponse.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Doctrine\ORM\Annotation as ORM;
+
+/**
+ * @ORM\Entity()
+ */
+class TextResponse extends AbstractResponse
+{
+    /**
+     * @ORM\Column(name="value_string", type="string")
```
Ocramius: In order to avoid confusing this issue with what discussed in the original thread, please use the same `type` for both entities here, so that we first verify only the hydration issues, and not the schema validator ones.

tests/Doctrine/Tests/Models/GH7505/TextResponse.php
```diff
@@ -0,0 +1,20 @@
...
+     *
+     * @var string
+     */
+    public $value = null;
```
Ocramius: `= null` is redundant. Type should also be annotated as `string|null`

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -328,6 +331,11 @@
             Models\Issue5989\Issue5989Employee::class,
             Models\Issue5989\Issue5989Manager::class,
         ],
+        'gh7505' => [
```
Ocramius: Since this is for just one ticket, consider squashing the entire model into the ticket test case class, as it is done with other similar tests

Ocramius (changes requested):

tests/Doctrine/Tests/Models/GH7505/TextResponse.php
m-r-r: Hello,

I don't quite understand what you mean: if both columns have the same type, the problem doesn't happen.

Ocramius: As discussed in https://github.com/doctrine/doctrine2/issues/7505#issuecomment-444924119, the test scenarios are two:

 * one with failed schema validation (same column name in STI, but with different types)
 * one with failed hydration (this patch - different column, same reflection property, different types)

I think this particular discussion point can therefore be dropped.

lcobucci:
@m-r-r since this is a bugfix , can you please rebase this against `2.6.x-dev`?

lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
```diff
@@ -129,6 +132,11 @@ protected function hydrateRowData(array $sqlResult, array &$result)
 
             // Prevent overwrite in case of inherit classes using same property name (See AbstractHydrator)
             if (! isset($data[$fieldName]) || ! $valueIsNull) {
+                // If we have inheritance in resultset, make sure the field belongs to the correct class
+                if (isset($cacheKeyInfo['discriminatorValues']) && ! in_array($discrColumnValue, $cacheKeyInfo['discriminatorValues'])) {
```
lcobucci: It would be safe to perform a strict `in_array()` here, right? If so we need to add `true` as third argument, otherwise ensure that non-strict comparison here is tested.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7505Test.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->em->persist($textResponse);
+
+        $this->em->flush();
+        $this->em->clear();
+
+        $repository = $this->em->getRepository(GH7505AbstractResponse::class);
+
+        /** @var GH7505ArrayResponse $arrayResponse */
+        $arrayResponse = $repository->find($arrayResponse->id);
+        static::assertEquals($arrayResponse->value, []);
```
lcobucci: The order of arguments here are wrong, the expectation comes first. Please also use `self` instead of `static` (the class is final and won't never be inherited). Lastly, `assertSame()` would be better for this case than `assertEquals()`.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7505Test.php
```diff
@@ -0,0 +1,92 @@
...
+
+        /** @var GH7505TextResponse $textResponse */
+        $textResponse = $repository->find($textResponse->id);
+        static::assertEquals($textResponse->value, null);
```
lcobucci: `self` instead of `static`. Using `assertNull()` would be better too.

lcobucci (changes requested):
@m-r-r just some minor things, thanks :+1: 
@Ocramius I think we should also compare the results of the hydration benchmark to see the impact here.

lcobucci:
@m-r-r @Ocramius even though the code should just work, does it make sense to extend the test case to add an example which covers JTI?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7505Test.php
lcobucci: @m-r-r thanks :smile:

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7505Test.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->em->persist($textResponse);
+
+        $this->em->flush();
+        $this->em->clear();
+
+        $repository = $this->em->getRepository(GH7505AbstractResponse::class);
+
+        /** @var GH7505ArrayResponse $arrayResponse */
+        $arrayResponse = $repository->find($arrayResponse->id);
+        self::assertEquals([], $arrayResponse->value);
```
lcobucci: I think you forgot to change from `assertEquals()` to `assertSame()`

lcobucci (commented):
Tiny thing :)

greg0ire:
Travis did not report back… let me rebase this.

greg0ire:
The tests fail, please fix them.

lcobucci (changes requested):
Apart from the failing build this should not be supported as @Ocramius said in https://github.com/doctrine/orm/issues/7505 and should rather be a validation error.

lcobucci:
@m-r-r we're working to get `2.6.5` released, enabling us to release `2.7.0` ASAP. I'll move this to `2.7.1` to give us time to solve this issue.

Thanks for your time and effort!

lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
```diff
@@ -134,6 +136,11 @@ protected function hydrateRowData(array $sqlResult, array &$result)
 
             // Prevent overwrite in case of inherit classes using same property name (See AbstractHydrator)
             if ( ! isset($data[$fieldName]) || ! $valueIsNull) {
+                // If we have inheritance in resultset, make sure the field belongs to the correct class
+                if (isset($cacheKeyInfo['discriminatorValues']) && ! in_array($discrColumnValue, $cacheKeyInfo['discriminatorValues'], true)) {
```
beberlei: Its probably better to test for `$discrColumnValue !== null` here first, because its a more performant check, or?

beberlei: maybe overkill, isset is fast enough. I am a bit worried about this check in the happy hotpath though, but i see no other option.

beberlei (approved):
This code is essentially the same than what AbstractHydrator checks for in lines 301-305, so its good to get that merged imho.

beberlei:
Closing in favor of #7974

Issue #7508, opened 2018-12-07
phpdoc for SQLFilter::setParameter


chucky2305:
Running phpstan (https://github.com/phpstan/phpstan) with the following command `php bin/phpstan analyse -l 5 src tests`  i came up to this error message:

```
 ------ ------------------------------------------------------------------------------------------------------------------- 
  Line   src/Filter/Configurator.php                                                                                        
 ------ ------------------------------------------------------------------------------------------------------------------- 
  45     Parameter #2 $value of method Doctrine\ORM\Query\Filter\SQLFilter::setParameter() expects string, int|null given.  

```

The second parameter ($value) is described as `string`. But it is pretty obvious, due to the code that follows, that not only a string can be passed to this method. 

I have not worked / participated in open source that much yet, so here are my questions:
1) Is this a bug?
2) Is this covered by a bugfix of somebody else?
3) Could i create a bugfix for this on my own and contribute it to the project?

chucky

Ocramius:
> Is this a bug?

Most likely


 > Is this covered by a bugfix of somebody else?

Not aware of it

 > Could i create a bugfix for this on my own and contribute it to the project?

Absolutely! Start by forking this repository, then clone your fork locally, create a branch, commit a fix, push it, and then open a pull request (github will suggest it)

PR #7509, opened 2018-12-07
#7508: changed phpdoc for parameter `$value` of `SQLFilter#setValue()` to match static analysis results

chucky2305:
#7508: changed phpdoc for parameter `$value` to match the actual code


Ocramius (approved):
LGTM :+1: 

Thanks @chucky2305!

Issue #7512, opened 2018-12-07
findBy broken for inherited entity property lookups


themizzi:
### Bug Report

| Q.                 |   A
|------------ | ------
| BC Break     | no
| Version        | 2.6.3

#### Summary

findBy does not work correctly for inherited entities. proof here: https://github.com/themizzi/doctrine-find-by-metadata-bug

#### Current behavior

If EntityB has a OneToMany relationship to EntityA, and EntityC inherits (via join) from EntityA. If there is no cache, calling `$entityBRepo->findBy('entityA' => $newEntityCInstance)` works fine. On a subsequent request after cache has already been generated and metadata is lazy loaded, the request breaks. on sqlite this just fails, on MySQL this potentially loads up an incorrect entity.

#### How to reproduce

Please see link above for an example and reproduction steps.

#### Expected behavior

Find by can find classes by properties with an inheritance structure at all times.

Ocramius:
@themizzi please reproduce this bug in isolation within an integration test, like in https://github.com/doctrine/doctrine2/tree/master/tests/Doctrine/Tests/ORM/Functional/Ticket.

themizzi:
will try to do this weekend. other priorities call for now.

themizzi:
Created regression test. Might have some time to looking into an appropriate fix this weekend.

PR #7514, opened 2018-12-08
Fix findBy metadata loading to be consistent

themizzi:
Adds regression test for issue #7512.


beberlei (approved):
This looks good to me. I will clean this up a little and rebase onto 2.7 before merging, thank you very much.

Issue #7518, opened 2018-12-12
`EntityRepository::__construct()` expects `Doctrine\ORM\EntityManager` instead of actual required `EntityManagerInterface`


koftikes:
### Bug Report

When I do phpstan check of my code.

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.6.x

#### Summary

```
Parameter #1 $em of method Doctrine\ORM\EntityRepository::__construct() expects Doctrine\ORM\EntityManager, Doctrine\ORM\EntityManagerInterface given.
```
#### How to reproduce
```
class RequestMasterRepository extends EntityRepository implements ServiceEntityRepositoryInterface
{
    /**
     * CompanyConfigRepository constructor.
     *
     * @param EntityManagerInterface $manager
     */
    public function __construct(EntityManagerInterface $manager)
    {
        parent::__construct($manager, $manager->getClassMetadata(RequestMaster::class));
    }
....
```
And run phpstam level 7 to this file.
```
vendor/bin/phpstan analyse -c phpstan.neon -l max src/AppBundle/Repository/
```

Ocramius:
Being handled in https://github.com/doctrine/doctrine2/pull/7519

Ocramius:
Handled in #7519

PR #7519, opened 2018-12-12
#7518 Fixed type mismatch between `EntityRepository#__construct()` and its documented constructor arguments

koftikes:


lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -58,8 +58,8 @@ class EntityRepository implements ObjectRepository, Selectable
     /**
      * Initializes a new <tt>EntityRepository</tt>.
      *
-     * @param EntityManager         $em    The EntityManager to use.
-     * @param Mapping\ClassMetadata $class The class descriptor.
+     * @param EntityManagerInterface    $em    The EntityManager to use.
```
Ocramius: Can be dropped completely

lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -58,8 +58,8 @@ class EntityRepository implements ObjectRepository, Selectable
...
+     * @param Mapping\ClassMetadata     $class The class descriptor.
```
Ocramius: Can be dropped completely (comment is not really telling anything useful)

Ocramius (changes requested):

lib/Doctrine/ORM/EntityRepository.php
koftikes: You mean fix to that:
```
    /**
     * Initializes a new <tt>EntityRepository</tt>.
     *
     * @param EntityManagerInterface $em
     * @param Mapping\ClassMetadata  $class
     */
```

lcobucci: @koftikes nope, drop both `@param`. They aren't needed since types are declared properly.

koftikes: @lcobucci nope it's not correct params declarations need for normal code 
image

Ocramius: That's an IDE warning that we ignore: our coding standards enforce dropping any docblock entry that doesn't describe the actual purpose of the referenced symbol.

koftikes: really? you have all descriptions on each constructor. lol.

Ocramius (approved):
:+1:

Ocramius:
Thanks @koftikes!

PR #7526, opened 2018-12-16
Add support metadata

greg0ire:
This should result in more links available on Packagist.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,14 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"},
         {"name": "Marco Pivetta", "email": "ocramius@gmail.com"}
     ],
+    "support": {
+        "chat": "https://www.doctrine-project.org/slack",
+        "docs": "https://www.doctrine-project.org/projects/orm.html",
+        "email": "doctrine-user@googlegroups.com",
+        "issues": "https://github.com/doctrine/doctrine2/issues",
+        "rss": "https://github.com/doctrine/doctrine2/releases.atom",
+        "source": "https://github.com/doctrine/doctrine2"
+    },
     "config": {
         "sort-packages": true
     },


composer.json
```diff
@@ -22,6 +22,13 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"},
         {"name": "Marco Pivetta", "email": "ocramius@gmail.com"}
     ],
+    "support": {
+        "email": "doctrine-user@googlegroups.com",
+        "issues": "https://github.com/doctrine/doctrine2/issues",
+        "source": "https://github.com/doctrine/doctrine2",
```
greg0ire: Here I hesitated with https://github.com/doctrine/doctrine2/tree/master/lib/Doctrine/ORM, and I notice that Packagist already displays a link pointing to https://github.com/doctrine/doctrine2/tree/master. Tell me what you think is best.

composer.json
```diff
@@ -22,6 +22,13 @@
...
+        "rss": "https://github.com/doctrine/doctrine2/releases.atom",
```
greg0ire: I wasn't sure what to put here, @Majkl578 what were you thinking about? Something else maybe?

Majkl578: Should be fine.

composer.json
Majkl578: Leave it like this, otherwise it'd be another maintenance burden.

Majkl578:
Can you also add `docs`?

greg0ire:
Sure thing! I also sorted all this alphabetically.

Majkl578 (dismissed):

composer.json
```diff
@@ -22,6 +22,14 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"},
         {"name": "Marco Pivetta", "email": "ocramius@gmail.com"}
     ],
+    "support": {
+        "chat": "https://www.doctrine-project.org/slack",
+        "docs": "https://www.doctrine-project.org/projects/orm.html",
+        "email": "doctrine-user@googlegroups.com",
+        "issues": "https://github.com/doctrine/doctrine2/issues",
+        "rss": "https://github.com/doctrine/doctrine2/releases.atom",
+        "source": "https://github.com/doctrine/doctrine2",
```
lcobucci: Invalid json here, comma not allowed

lcobucci (changes requested):

composer.json
greg0ire: Woops, good catch!

lcobucci (approved):

lcobucci:
@greg0ire :ship:

lcobucci:
`rss` seems to not be used though :smile: https://packagist.org/packages/doctrine/orm

PR #7543, opened 2018-12-27
Clarify error message for entity without identifier

GKFX:
I'm assuming this is the most common cause of the error; the message just seemed very cryptic to me otherwise!


Ocramius (changes requested):
Possibly an improvement for 3.x, not 2.x. See https://github.com/doctrine/doctrine2/pull/6743

derrabus:
Closing as there's no progress on this PR for several years now.

PR #7546, opened 2018-12-28
Fix previous exception rethrow on transaction rollback

simPod:
"Fixes" https://github.com/doctrine/doctrine2/issues/7545

Not sure if proper fix, was checking DBAL's code and everything is IMO just fine there. So my goal was to keep both exceptions in ORM's code.

Or do you have ideas for a better way to fix?

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -28,6 +28,7 @@ use Doctrine\ORM\Repository\RepositoryFactory;
 use Doctrine\ORM\Utility\IdentifierFlattener;
 use Doctrine\ORM\Utility\StaticClassNameConverter;
 use InvalidArgumentException;
+use ReflectionClass;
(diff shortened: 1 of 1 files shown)


Ocramius (changes requested):
Absolutely no-go. 

As described in https://github.com/doctrine/doctrine2/issues/7545#issuecomment-450353241, the exception coming from a failed rollback is indeed much more important here.

simPod:
🤔 and how can ORM's user fix failing rollbacks then when it's impossible to find its origin? Only thing user knows is that rollback failed. Why, is a mystery.

Ocramius:
You can design an exception with two previous traces. See https://github.com/doctrine/doctrine2/pull/6743 for examples on how we improved this for ORM 3.x.

Right now, changing the exception type being thrown for a failed rollback would be a subtle BC break, so please look at that patch and propose a separate solution for 3.x, if possible.

PR #7554, opened 2019-01-04
Allow all datetime types to be used as version

nico-incubiq:
I am working on an existing database with fields of type `timestamp`. In order to map that to a DateTime object in Doctrine, I've found the solution of adding the `@Version` annotation. But I'd really prefer to get a DateTimeImmutable, which is impossible.

At the moment a field of type `datetime_immutable` with annotation `@Version` will trigger this following error:
```
Locking type "datetime_immutable" (specified in "App\Entity\Bakery", field "createdAt") is not supported by Doctrine.
```

I cannot see any reason not to allow any type of datetime to be mapped to a timestamp in database, so this PR allows all of them. Please tell me if i'm missing something.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -3042,9 +3042,9 @@ class ClassMetadataInfo implements ClassMetadata
         $this->versionField = $mapping['fieldName'];
 
         if ( ! isset($mapping['default'])) {
-            if (in_array($mapping['type'], ['integer', 'bigint', 'smallint'])) {
+            if (in_array($mapping['type'], ['integer', 'bigint', 'smallint'], true)) {
                 $mapping['default'] = 1;
-            } else if ($mapping['type'] == 'datetime') {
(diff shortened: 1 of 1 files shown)


SenseException (changes requested):
Thank you for your contribution. Can you please add tests for all the supported types?

I'm not sure about if the ORM version 2.6 should introduce these changes. I'll wait what others have to say about the version.

lcobucci:
@nico-incubiq according to our docs `datetime` is supported, have you tried that? https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/transactions-and-concurrency.html#optimistic-locking

lcobucci:
As @SenseException this would be ok for v3 (`master`)

nico-incubiq:
Yeah i'm currently working with "datetime", but i'd rather get a DatetimeImmutable rather than just a DateTime. When is v3 due? Since this isn't a breaking change I expected it to be fine in 2.6.

lcobucci:
@nico-incubiq this is considered a new feature, which should go in a new `minor` release - not `patch`. You can propose this to `v2.7` but we were planning to only have deprecation notices there, however there's no due date set for `v3`.

nico-incubiq:
I've opened a new PR targeting v3 and added tests to it, I'll close this one then.

PR #7556, opened 2019-01-04
Fix incorrect handling of transactions using deferred constraints

grongor:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #7555

#### Summary

Rollback called when outside of a transaction when using deferred constraints (PostgreSQL).

_(I wasn't sure what branch to target, please tell me if I should change it :) Thanks )_

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -236,12 +236,19 @@ use function trigger_error;
             $return = call_user_func($func, $this);
 
             $this->flush();
+        } catch (Throwable $e) {
+            $this->close();
+            $this->conn->rollBack();
+
+            throw $e;
+        }
+
+        try {
             $this->conn->commit();
 
             return $return ?: true;
         } catch (Throwable $e) {
             $this->close();
-            $this->conn->rollBack();
 
             throw $e;
         }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7555Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7555Test.php
@@ -1,0 +1,133 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Functional\Ticket;
+
+use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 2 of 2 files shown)


grongor:
Build will fail until we merge the DBAL changes. You can see successful build here https://travis-ci.org/grongor/doctrine2/builds/475297482 (it uses patched DBAL from here https://github.com/doctrine/dbal/pull/3424 )

Ocramius:
This looks like a much better approach than the one we discussed before: nice!

Ocramius:
Assuming doctrine/dbal#3424 can target `2.9.3`, we need to bump the dependency constraint here. Since we can only bump it in a minor release, and 2.7 is closed for improvements (besides deprecations), I'm not sure if this should target 3.0.

/cc @doctrine/doctrinecore?

grongor:
Shouldn't this be merged into older versions too, as this fixes a bug and is non-BC? Just saying ... I'm not familiar with your release schedules :)

Ocramius:
Can't merge in older versions due to the dependency bump, and dependency bumps are off-limits for patch releases, unless it's a security issue.

lcobucci:
This is quite good for 2.7.0 just needs a rebase and the dependency bump on DBAL.

lcobucci:
@grongor @Ocramius I was wrong in my statement, DBAL still needs to be adjusted.

After talking to @morozov we've decided to pursue a simpler solution than the one proposed in https://github.com/doctrine/dbal/pull/3424 (basically a `try/finally` in the `Connection#commit()`). I'll create a PR for that adding the necessary cross-platform tests.

However, DBAL 2.9 isn't maintained any longer and 2.10 requires PHP 7.2. Since we want to have ORM 2.7.0 out ASAP due to SF5 support I'll push this to ORM 2.8.0.

simPod:
@lcobucci is this superseded by https://github.com/doctrine/dbal/pull/3424?

lcobucci:
@simPod it's been a while but that's correct. The challenge is to find a cross-platform implementation (that's well tested).

simPod:
Fixed in https://github.com/doctrine/dbal/pull/6545

PR #7558, opened 2019-01-05
Allow all datetime types to be used as version

nico-incubiq:
Replaces https://github.com/doctrine/orm/pull/7554

---

I am working on an existing database with fields of type timestamp. In order to map that to a DateTime object in Doctrine, I've found the solution of adding the @Version annotation. But I'd really prefer to get a DateTimeImmutable, which is impossible.

At the moment a field of type datetime_immutable with annotation @Version will trigger this following error:

Locking type "datetime_immutable" (specified in "App\Entity\Bakery", field "createdAt") is not supported by Doctrine.
I cannot see any reason not to allow any type of datetime to be mapped to a timestamp in database, so this PR allows all of them. Please tell me if i'm missing something.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadata.php
@@ -526,7 +526,7 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
             return;
         }
 
-        if ($property->getTypeName() === 'datetime') {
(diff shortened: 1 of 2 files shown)


Ocramius (approved):

Ocramius:
:ship: thanks @nico-incubiq!

yethee:
The PR has the merged status, but the `datetime_immutable` type is not supported in the latest versions. Changes reverted for some reasons?

PR #7560, opened 2019-01-08
Test that cache is created when using single scalar hydrator

goetas:
Checks that https://github.com/doctrine/dbal/pull/3381 works with the ORM.

This will require most likley to put `doctrine/dbal:^2.9.4` (`2.9.4` is not yet released)

So not sure if it makes sense (but since orm 2.7 requires at least php 7.2, do not see anything wrong with it)

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,10 @@
+# Upgrade to 2.7
+
+## Deprecated: `Doctrine\ORM\EntityManagerInterface#copy()`
+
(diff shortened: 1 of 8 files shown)


goetas:
hmm, something is wrong with my test, https://travis-ci.org/doctrine/orm/jobs/476824447 should be failing

greg0ire:
> This will require most likley to put doctrine/dbal:^2.9.4 (2.9.4 is not yet released) So not sure if it makes sense (but since orm 2.7 requires at least php 7.2, do not see anything wrong with it)

IMO it makes sense.

PR #7571, opened 2019-01-17
Fix typo in inheritance mappings docs

batwolf:
`has to allow null values` vs `has to allows null values`


SenseException (approved):

Ocramius:
@SenseException can you prepare a PR for `master` with this, if applicable?

PR #7581, opened 2019-01-24
Fix LimitSubqueryOutputWalker : Add default order by to sql wrapper if not set.

khaireddine-ghanmi:
I got a problem when testing a huge volume of data : greater than offset (of the order of 450000), the query by range returns always the same row (the last one in the list).
Exemple of query : 
`SELECT distinct id_0 FROM (SELECT o0_.id AS id_0 FROM myTable o0_) dctrn_result LIMIT 1 OFFSET 450001`
reeturns same row of :
`SELECT distinct id_0 FROM (SELECT o0_.id AS id_0 FROM myTable o0_) dctrn_result LIMIT 1 OFFSET 450100`

Because mysql is not able to assign a default row_num to results, the solution is to define a default `order by` in wrap query.
Result : 
`SELECT distinct id_0 FROM (SELECT o0_.id AS id_0 FROM myTable o0_) dctrn_result order by id_0 ASC LIMIT 1 OFFSET 450001`

Hope it helps..


Ocramius (changes requested):
An Integration test running against the tested DBa is needed

PR #7586, opened 2019-01-26
#7583 fix

### LINKED ISSUE #7583: UnitOfWork incorrect compute change set when entity have property with \DateTime type

vyacheslav-startsev:
#7583

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -693,7 +693,11 @@ class UnitOfWork implements PropertyChangedListener
                 $orgValue = $originalData[$propName];
 
                 // skip if value haven't changed
-                if ($orgValue === $actualValue) {
+                if (is_object($orgValue) &&
+                    is_object($actualValue) &&
+                    $orgValue == $actualValue ||
+                    $orgValue === $actualValue
+                ) {
                     continue;
                 }
 
--- a/tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
+++ b/tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Tests\ORM;
 
+use DateTime;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\EventManager;
 use Doctrine\Common\NotifyPropertyChanged;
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -693,7 +693,11 @@ public function computeChangeSet(ClassMetadata $class, $entity)
                 $orgValue = $originalData[$propName];
 
                 // skip if value haven't changed
-                if ($orgValue === $actualValue) {
+                if (is_object($orgValue) &&
+                    is_object($actualValue) &&
+                    $orgValue == $actualValue ||
+                    $orgValue === $actualValue
```
greg0ire: I don't think this could evaluate to `true` if the previous line does not also evaluate to `true`. IMO this line is useless.

vyacheslav-startsev: The previous line will be evaluated only when `$orgValue` and `$actualValue` are objects.
Conditions may be written anotherly:
```
(is_object($orgValue) && is_object($actualValue) && $orgValue == $actualValue) || ($orgValue === $actualValue)
```

SenseException: This example would return true despite having different values. A `==` check isn't accurate for objects.

``` php
class Foo
{
    private $i;
    public function __construct(?int $i)
    {
        $this->i = $i;
    }
}

$f1 = new Foo(0);
$f2 = new Foo(null);

var_dump($f1 == $f2);
```

This would break, for example, with custom Doctrine types that are objects. I think this is a BC break and should target to master for the next mayor release, if this is considered as a feature.

lcobucci:
As @Ocramius pointed out in the issue, this is the expected behaviour. I'll close this as invalid.

Issue #7589, opened 2019-02-01
Persisted Then Removed Entity Tries to Persist


bobdercole:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

#### Summary

I found an issue while attempting to instantiate a new entity, persist, and remove it all in the same flush. This results in an exception on flush.

This entity is the owning side of a one-to-many bidirectional relationship. It gets added to the inverse association collection when it's instantiated.

#### Current behavior

As mentioned above, the entity gets added to the inverse association collection when it's instantiated. If the entity is then removed with ```EntityManager::remove()```, it does not immediately get removed from the inverse association collection. This is fine, since the removal should occur within ```UnitOfWork::computeAssociationChanges()```. However at this point the entity state is identified as ```UnitOfWork::STATE_NEW```, even though it's been removed via ```EntityManager::remove()```. This causes the entity to remain in the inverse association collection. It also results in an exception due to unexpected persistence by reachability. 

```
Uncaught Doctrine\\ORM\\ORMInvalidArgumentException: A new entity was found through the relationship...
```

I should note that this strange behavior only occurs if ```EntityManager::persist()``` and ```EntityManager::remove()``` are called within the same ```EntityManager::flush()```. The issue does not occur if a second ```EntityManager::flush()``` is added between ```EntityManager::persist()``` and ```EntityManager::remove()```.

#### How to reproduce

1. Declare an entity as the owning side of a one-to-many bidirectional relationship. The constructor needs to add itself to the inverse association collection.
2. Instantiate the entity.
3. Persist the entity via ```EntityManager::persist()```.
4. Remove the entity via ```EntityManager::remove()```.
5. Flush the entity manager.

Here is a test case: #7590.

#### Expected behavior

I believe that an entity should not be persisted if ```EntityManager::remove()``` is called on it. Furthermore, the entity should be removed from the inverse association collection on flush.

bobdercole:
Closed due to user error (see #7590 review conversion).

PR #7590, opened 2019-02-01
#7589 Test Case - Persisted Then Removed Entity Tries to Persist

bobdercole:
Test case for #7589.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7589Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7589Test.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH7589Test extends OrmFunctionalTestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+        $this->_schemaTool->createSchema(
+            [
+                $this->_em->getClassMetadata(GH7589User::class),
+                $this->_em->getClassMetadata(GH7589Article::class),
+            ]
+        );
+    }
+
+    public function testPersistedThenRemovedEntityIsRemoved()
+    {
+        $user    = new GH7589User(1);
+        $article = new GH7589Article(1, $user);
+
+        $this->_em->persist($user);
+        $this->_em->persist($article);
+
+        $this->_em->remove($article);
+
+        $this->_em->flush();
+
+        $this->assertNull($this->_em->find(GH7589Article::class, 1), 'Entity should not be persisted.');
+        $this->assertCount(0, $user->getArticles(), 'Entity should be removed from inverse association collection.');
+    }
+}
+
+/**
+ * @Entity
+ */
+class GH7589User
+{
+    /**
+     * @Id
+     * @Column(type="integer")
+     */
+    private $id;
+
+    /** @OneToMany(targetEntity="GH7589Article", mappedBy="user") */
+    private $articles;
+
+    public function __construct(int $id)
+    {
+        $this->id       = $id;
+        $this->articles = new ArrayCollection();
+    }
+
+    public function addArticle(GH7589Article $article) : void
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH7589Test.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $user    = new GH7589User(1);
+        $article = new GH7589Article(1, $user);
+
+        $this->_em->persist($user);
+        $this->_em->persist($article);
+
+        $this->_em->remove($article);
+
+        $this->_em->flush();
```
Ocramius: A crash here is expected due to incomplete object graph reference. I'd certainly expect the ORM to reject such an incomplete `UnitOfWork` at `commit()` time. What behaviour were you expecting?

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7589Test.php
Ocramius: The above example is:

 1. `flush()` call: consistent graph, all entities persisted
 2. `flush()` call: only removal scheduled, no changes on the association (hence skipped in update), since the managed entity is still in the `UnotOfWork`


I'd expect a third flush (with a change) to lead to a crash.

bobdercole: A third flush with a change, meaning a change to the article? This example does not crash since the first article is removed from ```$user::articles``` within ```UnitOfWork::computeAssociationChanges()```, [right here](https://github.com/doctrine/orm/blob/e003bb2bb4199fc00dfc89f5a98a8e8a2be3fc7f/lib/Doctrine/ORM/UnitOfWork.php#L900), during the second flush. I've added a string property, ```GH7589Article::title```, to demonstrate the update.

```php
$user    = new GH7589User(1);
$article = new GH7589Article(1, $user, 'Welding Tutorial');

$this->_em->persist($user);
$this->_em->persist($article);

$this->_em->flush();

$this->_em->remove($article);

$this->_em->flush();

$article->setTitle('Soldering Tutorial'); // No longer managed anyway?

$this->_em->flush();
```

I believe the object graph is made consistent after the second flush; ```$user::articles``` is empty.

Since I wasn't quite sure which change you suggested would crash, I tested this too and it works (except this would fail my original test case assertions since ```$user::articles``` is no longer empty):

```php
$user    = new GH7589User(1);
$article = new GH7589Article(1, $user);

$this->_em->persist($user);
$this->_em->persist($article);

$this->_em->flush();

$this->_em->remove($article);

$this->_em->flush();

$another_article = new GH7589Article(2, $user);
$this->_em->persist($another_article);

$this->_em->flush();
```

Ocramius: Ah, indeed, it wouldn't crash at all. Still, your object graph in the test s invalid anyway, since the `User` and the `Article` are in a bi-directional association that isn't kept consistent by your code (even at instantiation time).

Here's what I meant before:

```php
class A {public $id;}
class B {public $id; public $field = 'default'; public $a;}

$a = new A;
$b = new B;
$b->a = $a;

$em->persist($a);
$em->persist($b);

$em->flush();

$em->remove($a);

$em->flush(); // works (even if the association is still set);

$b->field = 'changed;

$em->flush(); // crashed (reference to non-managed $a)
```

Does that make more sense?

Ocramius: > How should the consistency on removal be ensured? Would it be recommended to use a ```preRemove``` lifecycle callback on ```Article``` to internally call ```User::removeArticle()```?

Avoid using listeners and magic for this: make it explicit, part of your code in first place.

Do you think the issue is cleared with this discussion? Or is there still buggy behavior that wasn't discussed?

bobdercole: I see. I think we can call it resolved. Thank you for your time. 👍

bobdercole:
Closed due to user error (see review conversion).

PR #7603, opened 2019-02-11
`AbstractQuery::getSingleScalarResult()` never conditionally returns `null`: corrected docblock accordingly

creocoder:
This method never return `null`.


Ocramius (approved):
LGTM :+1:

PR #7606, opened 2019-02-11
CASE WHEN as left operand of an IN comparison. Closes #7605

### LINKED ISSUE #7605: Condition using "CASE WHEN" not expecting "IN" keyword
### Bug Report

DQL:
SELECT cg FROM \App\Entity\Common\CompanyGroup cg WHERE CASE WHEN cg.id = 1 THEN 0 WHEN cg.id = 2 THEN 1 ELSE 3 END IN (3)
Error:
[Syntax Error] line 0, col 116: Error: Expected =, <, <=, <>, >, >=, !=, got 'IN'

If I change "IN (3)" for "= 3" it works.

Note: the DQL above is not my real need, it's just a simple snippet of code to demonstrate the bug. The entity used can have just one field called "id" to run the example.

BC Break: no (I guess)
Version: doctrine/orm 2.6.3

#### Summary
Doctrine should accept "IN" operator when the left operand IS a "CASE WHEN" (DQL)

#### Current behavior
Doctrine is not accepting "IN" when the left operand IS a "CASE WHEN"

#### How to reproduce

$q = $em->createQuery('SELECT cg FROM \App\Entity\Common\CompanyGroup cg WHERE CASE WHEN cg.id = 1 THEN 0 WHEN cg.id = 2 THEN 1 ELSE 3 END IN (3)');
$cgs = $q->getResult();

#### Expected behavior
The DQL above run without any error

tonyfarney:
Case when as left operand of an IN comparison was generating syntax error.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -2558,6 +2558,18 @@ class Parser
             }
 
             $this->lexer->resetPeek();
+        } elseif ($token['type'] === Lexer::T_CASE || ($token['type'] === Lexer::T_OPEN_PARENTHESIS && $peek['type'] === Lexer::T_CASE)) {
+            // Need to discover if it's a regular or an "in" comparison
+            do {
+                $token = $this->lexer->peek();
+            } while ($token['type'] != Lexer::T_END);
+            while ($token = $this->lexer->peek()) {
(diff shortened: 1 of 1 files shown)


tonyfarney:
According to orientation I received, I need to commit in the master branch, not in the 2.6. So... I'm going to close this PR.

PR #7610, opened 2019-02-14
Change APC to OPcache in improving-performance.rst 

smtchahal:
OPcache is a lot better supported now than APC, which is apparently not maintained anymore.

### DIFF
--- a/docs/en/reference/improving-performance.rst
+++ b/docs/en/reference/improving-performance.rst
@@ -4,7 +4,7 @@ Improving Performance
 Bytecode Cache
 --------------
 
-It is highly recommended to make use of a bytecode cache like APC.
+It is highly recommended to make use of a bytecode cache like OPcache.
 A bytecode cache removes the need for parsing PHP code on every
 request and can greatly improve performance.
 


holtkamp:
>OPcache is a lot better supported now than APC

Well, I guess it depends.. OPCache can not be easily purged using a "control panel" interface. APCu can be purged this way

>which is apparently not maintained anymore

Last release was 6 days ago? https://github.com/krakjoe/apcu/releases

or am I missing something...

apfelbox:
@holtkamp APCu and APC are two different tools.

holtkamp:
aah, yeah, that is true. In that case I would describe both `OPCache` and `APCu`, each has (dis)advantages

apfelbox:
APCu certainly has advantages, but it just isn't an opcode cache. So in this case it would be incorrect, because saying "use a bytecode cache like [...] APCu" is wrong – it isn't one. It is for storing user data, not internally generated opcodes.

holtkamp:
Ahh, ok, is that the intent of this commit. Then nevermind my remarks! Sorry 😄

greg0ire (dismissed):

greg0ire:
I approve the changes, but maybe you could improve the commit message subject? "Update improving-performance.rst" does not really tell us what you did. The commit message body is good though 👍

smtchahal:
@greg0ire Just made another amended commit with a different commit title.

greg0ire (approved):
It looks better!

SenseException (approved):

Majkl578:
> OPCache can not be easily purged using a "control panel" interface

@holtcamp This is not true, there are multiple _admin_ interfaces for OPcache as well. Some examples:
- https://github.com/rlerdorf/opcache-status
- https://github.com/carlosbuenosvinos/opcache-dashboard
- https://github.com/PeeHaa/OpCacheGUI
- https://github.com/amnuts/opcache-gui

>  In that case I would describe both `OPCache` and `APCu`, each has (dis)advantages

Note that APCu is not a bytecode/opcode cache but only a key-value store, so it's not an alternative to OPCache.

---

Generally speaking this article needs some work to make it reflect current state of PHP ecosystem, e.g. next paragraph also mentions obsolete APC and Memcache.

Majkl578 (approved):

PR #7614, opened 2019-02-17
Allow to escape reserved words in xml mapping

mintobit:
When escaping reserved words in XML mapping XSD validation fails. This PR uses regex to allow backticks. This regex is also much more restrictive than `xs:NMTOKEN`, let me know if that's not okay.
```
<field name="email" column="`user_email`" type="string" column-definition="CHAR(32) NOT NULL" />
```

Any advice in terms of tests? I went over #7324, #6728 and got a bit confused. As far as I understand, #6728 will introduce functionality that can be used in tests, and all XSD related PRs are waiting for that to be merged. Did I get it right?

### DIFF
--- a/doctrine-mapping.xsd
+++ b/doctrine-mapping.xsd
@@ -207,6 +207,13 @@
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:simpleType name="columnname" id="columnname">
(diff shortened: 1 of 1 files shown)


greg0ire:
> all XSD related PRs are waiting for that to be merged

I don't think so…

PR #7621, opened 2019-02-21
Issue #6009

salimimani:
Issue #6009 : https://github.com/doctrine/orm/issues/6009
The info is in my comment on the issue


Ocramius:
This sort of change requires test case additions demonstrating the issue being fixed (tests must fail before patch is applied)

PR #7624, opened 2019-02-22
Fixes #7623 : SqlWalker discriminator column condition optimization

### LINKED ISSUE #7623: [Perf] useless discriminator column filtering for top level classes
Given the following Address mapping : 
```
/**
 * @Entity
 * @Table(name="`quote-address`")
 * @InheritanceType("SINGLE_TABLE")
 * @DiscriminatorColumn(name="type", type="string")
 * @DiscriminatorMap({"simple" = Address::class, "full" = FullAddress::class})
 */
class Address
```

It seems to me that the SQL translation of the DQL `SELECT a, FROM Address` should be the simple `SELECT q0_...  FROM "quote-address" q0_` instead of the actual `SELECT q0_...  FROM "quote-address" q0_ WHERE q0_.type IN ('simple', 'full')`.

It is certainly not wrong but sometimes not so optimal : the condition is only needed when you ask for a subclass.

rbs-asimon:
when you select a class that is on the top of the hierarchy, there is no need to add a where "filtering" on ALL the subclasses: just do not do any filter
This fixes #7623

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -444,7 +444,7 @@ class SqlWalker implements TreeWalker
         foreach ($dqlAliases as $dqlAlias) {
             $class = $this->queryComponents[$dqlAlias]['metadata'];
 
-            if ( ! $class->isInheritanceTypeSingleTable()) continue;
+            if ( ! $class->isInheritanceTypeSingleTable() || ! $class->parentClasses) continue;
 
             $conn   = $this->em->getConnection();
             $values = [];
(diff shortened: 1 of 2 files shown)


Ocramius:
Same as in #7623, this is not necessary, as the discriminator is to be covered by indexing anyway.

In addition to that, the forced discriminator filtering was introduced to handle scenarios where the discriminator would go out of the supported types in a STI/JTI, which, while uncommon, is something that may happen.

Therefore, closing as "won't fix" here.

rbs-asimon:
@Ocramius could you elaborate the "out of the supported types in a STI/JTI" ? Here the proposal focuses on single table inheritance ? Do you have some example illustrating the point ? I feel silly... :)
Maybe a new unit test should exist so we all see the uncommon scenarios.

Ocramius:
Insert a record that has a discriminator value that is not in your discriminator map.

rbs-asimon:
@Ocramius  Yes sure, this is a case but this case seems to me like "someone choosed not to work with doctrine at all" and this case should be "someone" problem instead of doctrine problem ?

PR #7630, opened 2019-02-28
Fix #7629 - `scheduledForSynchronization` leaks memory when using `@ORM\ChangeTrackingPolicy("DEFERRED_EXPLICIT")`

### LINKED ISSUE #7629: `scheduledForSynchronization` leaks memory when using `@ORM\ChangeTrackingPolicy("DEFERRED_EXPLICIT")`
### Bug Report

`scheduledForSynchronization` not erases when no changes on the flush.

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 2.6.3

#### Summary

I use `DEFERRED_EXPLICIT` policy for entities in my app. And app crashes with out of memory error during batch processing. Because `scheduledForSynchronization` keeps references to entities when no changes for commit:

https://github.com/doctrine/orm/blob/434820973cadf2da2d66e7184be370084cc32ca8/lib/Doctrine/ORM/UnitOfWork.php#L359

In this case `UnitOfWork::postCommitCleanup()` not called.

#### How to reproduce

```php

/**
 * @ORM\Entity()
 * @ORM\ChangeTrackingPolicy("DEFERRED_EXPLICIT")
 */
class Entity
{
}

$entity = $em->find(Entity::class, 123);

$em->persist($entity);
$em->flush();
$em->detach($entity);

var_dump($em->getUnitOfWork()->size()); // int(0)
var_dump($em->getUnitOfWork()->isScheduledForDirtyCheck($entity)); // bool(true)
```

Was expected that `isScheduledForDirtyCheck()` return `false` after commit (flush) in any case.

yethee:
Fixes #7629

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -356,6 +356,8 @@ class UnitOfWork implements PropertyChangedListener
             $this->dispatchOnFlushEvent();
             $this->dispatchPostFlushEvent();
 
+            $this->postCommitCleanup($entity);
+
             return; // Nothing to do.
         }
 
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7629Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7629Test.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH7629Test extends OrmFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH7629Entity::class,
+        ]);
+
+        $this->_em->persist(new GH7629Entity());
+        $this->_em->flush();
+        $this->_em->clear();
+    }
+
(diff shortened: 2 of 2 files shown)


Ocramius:
Possibly enough to call `postCommitCleanup()` internally there? Does the memleak appears also in non-empty UoW scenarios?

yethee:
> Possibly enough to call `postCommitCleanup()` internally there?

Yes, this should solve the issue. I can update PR. But method `postCommitCleanup()` has been absent in the master branch.

> Does the memleak appears also in non-empty UoW scenarios?

I did not quite understand, that you mean on *non-empty UoW scenarios*? In theory, a next commit with changes (not empty) should reset `scheduledForSynchronization`.

Some info from debugging my app (used change tracking policy `DEFERRED_EXPLICIT` for all entities):

Chunk 1, before flush and detach (with exception of entities of `EligibleAdGroup`, they already detached):

image

Chunk 1, after flush and detach:

image

Chunk 2...

Chunk 3, before flush and detach:

image

Chunk 3, after flush and detach:

image

Commits (flush) for all chunks are empty - no changes.

Ocramius:
> Yes, this should solve the issue.

Then let's apply it here, and then we look for a solution for `master`.

Ocramius:
I think patch looks fine as-is now, but we'll likely need a separate one (first commit only) for `master`, where we'll most likely need a separate approach.

Ocramius (approved):
This one is good to go as-is, thanks @yethee!

yethee:
I create a separate PR #7632 for merge to `master`

PR #7632, opened 2019-03-01
Fix #7629 - `scheduledForSynchronization` leaks memory when using `@ORM\ChangeTrackingPolicy("DEFERRED_EXPLICIT")

### LINKED ISSUE #7629: `scheduledForSynchronization` leaks memory when using `@ORM\ChangeTrackingPolicy("DEFERRED_EXPLICIT")`

yethee:
Test to demonstrate unexpected behavior, related to issue #7629

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -349,6 +349,8 @@ class UnitOfWork implements PropertyChangedListener
             $this->dispatchOnFlushEvent();
             $this->dispatchPostFlushEvent();
 
+            $this->postCommitCleanup();
+
             return; // Nothing to do.
         }
 
@@ -426,7 +428,11 @@ class UnitOfWork implements PropertyChangedListener
 
         $this->dispatchPostFlushEvent();
 
-        // Clean up
+        $this->postCommitCleanup();
+    }
+
+    private function postCommitCleanup() : void
+    {
         $this->entityInsertions            =
         $this->entityUpdates               =
         $this->entityDeletions             =
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7629Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7629Test.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Annotation as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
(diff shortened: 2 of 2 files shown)


Ocramius:
So, I think that on `master` we need to take this bit and isolate it to a `private` method to be called also in case of a clean `UnitOfWork`:

https://github.com/doctrine/orm/blob/e8eaf8386de1497863cb400fbc4fddda5878db7e/lib/Doctrine/ORM/UnitOfWork.php#L429-L439

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -426,7 +428,11 @@ public function commit()
 
         $this->dispatchPostFlushEvent();
 
-        // Clean up
+        $this->postCommitCleanup();
+    }
+
+    private function postCommitCleanup()
```
Ocramius:
```suggestion
    private function postCommitCleanup() : void
```

Ocramius (changes requested):

lib/Doctrine/ORM/UnitOfWork.php
yethee: I did not add `void` because other methods without return type-hints.

Ocramius: `master` has enforced CS fixes for diffs anyway, so if this isn't applied you'll get a red build anyway

yethee: Fixed

Ocramius (approved):
Thanks @yethee!

PR #7639, opened 2019-03-06
Fixed break/continue error with PHP7.3

arkiski:
Error was produced using Laravel 5.4 with laravel-doctrine package. Updated server to php 7.3 and composer 1.8.4, and when i tried to use the doctrine extension the break/continue warning throwed the error.


Ocramius:
ORM 2.5.x is closed for fixes: please update to ORM 2.6.x (current stable).

7ochem:
This one bug/fix is causing people not being able to run Symfony 2.8 with Doctrine on PHP 7.3 / 7.4.

PR #7641, opened 2019-03-07
fix: COUNT() being casted as string

Grafikart:
Some functions have a predefined type and could be resolved correctly by the SQLWalker. I don't want to change things too much but it could be possible to do the same thing for other functions like SUM for instance that would be resolved to float. 

A much better solution would be to retrieve the aggregateExpression, then the pathExpression to resolve the type from the field but it's not feasable without an accessor from FunctionNode class.


Grafikart:
A much better solution (but I need to know if it would be accepted before implementing it) would be to add an interface `Typeable` on the `FunctionNode` class that would define a `getReturnType(): string`

Then checking for this interface would be sufficient to resolve the type and every function could inform the walker of the type returned.

Ocramius (changes requested):
This looks like a decent idea, but needs tests to back it.

Also, I'd say that we need a `private` method that identifies the type for built-in function nodes.

Grafikart:
I can try to write test but the SqlWalkerTest seems a bit empty ^^
You want me to try the interface way ?

Ocramius:
I suggest looking at integration tests that verify the generated `ResultSetMapping` instance from a full parser run.

Grafikart:
Good idea, can you point me to the file where I should add this test ?

Ocramius:
Possibly need a new file, similar to https://github.com/doctrine/orm/blob/master/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php, but focused on resultset mapping

Grafikart:
I added the tests and altered more functions

tests/Doctrine/Tests/ORM/Query/SelectSqlMappingTest.php
```diff
@@ -0,0 +1,81 @@
...
+            $method = $r->getMethod('getResultSetMapping');
+            $method->setAccessible(true);
+            /** @var ORMQuery\ResultSetMapping $mapping */
+            $mapping = $method->invoke($query);
+            foreach ($types as $key => $expectedType) {
+                $alias = array_search($key, $mapping->scalarMappings);
+                self::assertInstanceOf(
+                    $expectedType,
+                    $mapping->typeMappings[$alias],
+                    sprintf('The field "%s" was expected as a $expectedType', $key)
```
flohw: `$expectedType` value would not be printed, you should use another `%s` parameter

tests/Doctrine/Tests/Models/CMS/CmsProduct.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\CMS;
+
+use Doctrine\ORM\Annotation as ORM;
+
+/**
+ * CmsAddress
```
flohw: Only for consistency, it should be `CmsProduct`.

flohw (commented):
:slightly_smiling_face:

tests/Doctrine/Tests/ORM/Query/SelectSqlMappingTest.php
```diff
@@ -0,0 +1,81 @@
...
+                );
+            }
+            $query->free();
+        } catch (Exception $e) {
```
Ocramius: Please don't `catch` in tests

Ocramius (changes requested):

lib/Doctrine/ORM/Query/AST/TypableNode.php
```diff
@@ -0,0 +1,13 @@
+<?php
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Query\AST;
+
+use Doctrine\DBAL\Types\Type;
+
+interface TypableNode
```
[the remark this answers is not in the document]
Grafikart: As you wish I can rename it

Grafikart:
Everything should be fixed now

lib/Doctrine/ORM/Query/AST/Functions/AbsFunction.php
```diff
@@ -41,4 +42,12 @@ public function parse(Parser $parser)
 
         $parser->match(Lexer::T_CLOSE_PARENTHESIS);
     }
+
+    /**
+     * @inheritDoc
+     */
```
greg0ire: There is no doc to inherit

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -1388,7 +1388,12 @@ public function walkSelectExpression($selectExpression)
                 if (! $hidden) {
                     // Conceptually we could resolve field type here by traverse through AST to retrieve field type,
                     // but this is not a feasible solution; assume 'string'.
-                    $this->rsm->addScalarResult($columnAlias, $resultAlias, Type::getType('string'));
+                    if ($expr instanceof Query\AST\TypedExpression) {
```
greg0ire: Maybe the comment above is no longer 100% accurate? Should probably be "assume 'string' when the expression is not typed"?

greg0ire (commented):
Please squash your commits together and make the commit message more accurate, this is no longer just about `COUNT()`.

lib/Doctrine/ORM/Query/AST/TypedExpression.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Query\AST;
+
+use Doctrine\DBAL\Types\Type;
+
+interface TypedExpression
```
greg0ire: Is this interface for internal use only? If yes, please mark it as such with `@internal` + a comment. If not, please document it.

Grafikart: I don't think so it can be used by custom function created by the user (for instance COS() could implement this interface)

greg0ire: OK, then maybe document it?

greg0ire:
@Grafikart why did you delete your repository? I hope I didn't discourage you with my nitpicks?

lib/Doctrine/ORM/Query/AST/Functions/FunctionNode.php
```diff
@@ -43,4 +45,9 @@ public function dispatch($sqlWalker)
      * @return void
      */
     abstract public function parse(Parser $parser);
+
+    public function getReturnType() : Type
+    {
+        return Type::getType(Type::STRING);
```
ostrolucky: Why defining this when it's default fallback anyway?

ostrolucky (commented):
It's a nice, simple patch solving the problem, however I'm not sure we should bite the bullet with floats. COUNT() and LENGTH() as integer types are fine, but casting to float in PHP will mean losing accuracy. It's also quite strange to make eg. `ABS(1)` return `1.0`. Also, this a BC break and as such should be documented.

ostrolucky (commented):
It's a nice, simple patch solving the problem, however I'm not sure we should bite the bullet with floats. COUNT() and LENGTH() as integer types are fine, but casting to float in PHP will mean losing accuracy. It's also quite strange to make eg. `ABS(1)` return `1.0`. Also, this a BC break and as such should be documented.

Grafikart:
@greg0ire I wanted to clean my repositories and I ended up deleting all my forks with an automated tool :D. I have no problem with the nitpicking since it makes the code better for everyone and helps me learning how to be more rigorous.

To be honest I forgot about this PR but I could send a new PR with the same fixes if you still want this functionnality.

ostrolucky:
Please do so. I think there is an interest from core team to have this functionality in some way

Grafikart:
I did a new PR #7941. I'm closing this one since I deleted the fork by mistake 👎

PR #7645, opened 2019-03-12
update phpstan to 0.11

adaamz:
There is issue which I dont know how you want to resolve:
`Doctrine\ORM\Persisters\Entity\BasicEntityPersister.php:`
```
Array has 2 duplicate keys with value '=' (\Doctrine\Common\Collections\Expr\Comparison::EQ, \Doctrine\Common\Collections\Expr\Comparison::IS).
```

Just tell me which one I should delete and I fix it in this PR.

### DIFF
--- a/lib/Doctrine/ORM/Query/TreeWalkerChain.php
+++ b/lib/Doctrine/ORM/Query/TreeWalkerChain.php
@@ -6,6 +6,7 @@ namespace Doctrine\ORM\Query;
 
 use function array_diff;
 use function array_keys;
+use Doctrine\ORM\AbstractQuery;
 
 /**
  * Represents a chain of tree walkers that modify an AST and finally emit output.
@@ -55,7 +56,7 @@ class TreeWalkerChain implements TreeWalker
     /**
      * {@inheritdoc}
      */
-    public function __construct($query, $parserResult, array $queryComponents)
+    public function __construct(AbstractQuery $query, ParserResult $parserResult, array $queryComponents)
     {
         $this->queryComponents = $queryComponents;
         $this->walkers         = new TreeWalkerChainIterator($this, $query, $parserResult);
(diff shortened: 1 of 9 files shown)


carusogabriel:
@CZechBoY I believe that one should be ignored. Whoever is using `Comparison::EQ` or `Comparison::IS` is gonna be a BC 😕

CZechBoY:
@carusogabriel The collision is in private static property, but keys and values are the same.
```
class BasicEntityPersister implements EntityPersister
{
    /** @var string[] */
    private static $comparisonMap = [
        Comparison::EQ          => '= %s',
        Comparison::IS          => '= %s',
        Comparison::NEQ         => '!= %s',
        Comparison::GT          => '> %s',
        Comparison::GTE         => '>= %s',
        Comparison::LT          => '< %s',
        Comparison::LTE         => '<= %s',
        Comparison::IN          => 'IN (%s)',
        Comparison::NIN         => 'NOT IN (%s)',
        Comparison::CONTAINS    => 'LIKE %s',
        Comparison::STARTS_WITH => 'LIKE %s',
        Comparison::ENDS_WITH   => 'LIKE %s',
    ];
```

So should I remove IS/EQ? Or change constant value?

lib/Doctrine/ORM/Tools/Pagination/CountOutputWalker.php
```diff
@@ -49,7 +50,7 @@ class CountOutputWalker extends SqlWalker
      * @param ParserResult $parserResult
```
carusogabriel: You can drop these two phpDocs.

lib/Doctrine/ORM/Query/TreeWalkerChain.php
```diff
@@ -6,6 +6,7 @@
 
 use function array_diff;
 use function array_keys;
+use Doctrine\ORM\AbstractQuery;
```
carusogabriel: This should go before the `use function *` section.

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryOutputWalker.php
```diff
@@ -89,7 +90,7 @@ class LimitSubqueryOutputWalker extends SqlWalker
      * @param ParserResult $parserResult
```
carusogabriel: You can remove these two phpDocs.

carusogabriel (changes requested):
> So should I remove IS/EQ? Or change constant value?

Let's add it to the `phpstan.neon` file under `ignoreErrors` for now.

PR #7660, opened 2019-03-27
Fixed commit order

tomaszgaw:
Fixes visit method of CommitOrderCalculator.

Without the fix the insertion order is not properly calculated when a cycle within the graph involves more than 2 classes, leading to a violation of not null constraints on the owning side of ManyToOne relation. 

With the fix when a cycle is detected a node is revisited ignoring edges of weight 0.


SenseException:
Can you please cover your changes with tests?

I'm also not sure yet if this introduces changes in the behaviour on first look.

tomaszgaw:
The test I added passes when runs against the fixed code and fails when runs against the original code.

tests/Doctrine/Tests/ORM/CommitOrderCalculatorTest.php
```diff
@@ -100,6 +100,33 @@ public function testCommitOrdering3()
         // We want to perform a strict comparison of the array
         $this->assertContains($sorted, $correctOrders, '', false, true, true);
     }
+    
+    public function testCommitOrdering4()
```
greg0ire: I can't say I really understand this PR, but here is some help for those who can:

```
       +---------+
       |         |
    +--+ Class 4 <---+
    |  |         |   |
  1 |  +---------+   | 0
    |                |
    v                |
+---+-----+    +-----+---+
|         |    |         |
| Class 1 +--->+ Class 2 |
|         | 0  |         |
+---+-----+    +-----+---+
    ^                |
    |                |
  0 |                | 0
    |  +---------+   |
    |  |         |   |
    +--+ Class 3 +<--+
       |         |
       +---------+

```

Maybe embed this as a comment if you find it useful?

greg0ire:
There should be only one commit, the test does not make sense without the code, and the code should not come without tests. Please improve the commit message to make as interesting as the pull request text.

lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
```diff
@@ -149,11 +149,15 @@ public function sort()
      *
      * @param \stdClass $vertex
      */
-    private function visit($vertex)
+    private function visit($vertex, $minimumWeight = 0)
```
greg0ire: This branch requires 7.2, you can use a type hint:

```suggestion
    private function visit($vertex, int $minimumWeight = 0)
```

tests/Doctrine/Tests/ORM/CommitOrderCalculatorTest.php
tomaszgaw: Thank you for the suggestion. I embedded this graph in a comment within the test.

lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
tomaszgaw: I applied the suggested change in the last (squashed) commit. Thank you.

greg0ire:
Both commits explain what they fix, but not how. Please include the explanation and follow the guidelines at https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches

greg0ire:
Closing in favor of #8349

PR #7662, opened 2019-03-30
Fix test on Postgres when running on 30/31 march

mbeccati:
Adjusted the delta to account for March 30-31 -> Feb 28th + 1 extra day for DST changes. I guess it doesn't hurt to be extra careful.

Today is march 30th and postgres returns "feb 28th" when doing now() - '1 month', for reference:
```
Doctrine\Tests\ORM\Functional\QueryDqlFunctionTest::testDateSub with data set "month" ('month', 1, 86400)
Failed asserting that two DateTime objects are equal.
--- Expected
+++ Actual
@@ @@
-2019-03-02T16:50:58.973200+0100
+2019-02-28T16:50:58.973200+0100
```

### DIFF
--- a/tests/Tests/ORM/Functional/QueryDqlFunctionTest.php
+++ b/tests/Tests/ORM/Functional/QueryDqlFunctionTest.php
@@ -369,7 +369,7 @@ class QueryDqlFunctionTest extends OrmFunctionalTestCase
 
         return [
             'year'   => ['year', 1, $secondsInDay],
-            'month'  => ['month', 1, $secondsInDay],
+            'month'  => ['month', 1, 4 * $secondsInDay],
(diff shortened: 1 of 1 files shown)


mbeccati:
@greg0ire Apologies for pinging you directly, but you seem to be merging most of the PRs nowadays ;-)

derrabus:
Is this PR still relevant? If so, it needs to be rebased against 2.20.x.

mbeccati:
@derrabus Yep, the path of the file had changed, but the issue is still relevant: the test will always fail on PG during those two days each year.

mbeccati:
Strangely enough the test is not failing this year, so I'll just close the issue.

greg0ire:
I remember working on similar stuff in the past years, I might have fixed it.

PR #7663, opened 2019-04-01
fixed exception class name

OskarStark:


Majkl578:
Since the code is not inside namespace, I¨d argue the current snippet is correct and `\` is suprfluous (has no effect).

greg0ire:
Maybe, but it does not hurt and makes this piece of code more copy/pastable in context where namespaces are used, which is the most likely.

greg0ire (approved):

OskarStark:
> which is the most likely.

thats what I thought about, most users doing the copy paste will use it in a namespaced environment.

SenseException (commented):
-

SenseException (changes requested):
I would like this to be consistent through the docs. It's either all having a `\` or none. Just changing one class doesn't do the trick.

OskarStark:
@SenseException thank you for your feedback, which target branch would you prefer when doing such "global fix"?

SenseException:
2.6 and master diverges in the docs, but I guess there aren't much changes in the code examples yet. I suggest master.

OskarStark:
Sorry I don't have some free time to do this 😢

PR #7672, opened 2019-04-04
Added cross-links to relevant documentation

jschaedl:


docs/en/reference/improving-performance.rst
```diff
@@ -26,6 +26,8 @@ Doctrine will need to load your mapping information on every single
 request and has to parse each DQL query on every single request.
 This is a waste of resources.
 
+See :ref:`caching#integrating-with-the-orm`
```
greg0ire: https://github.com/doctrine/orm/blob/b243ea1940d65d9ec3cc3045286726b7e79145b7/docs/en/reference/caching.rst#integrating-with-the-orm

docs/en/reference/improving-performance.rst
```diff
@@ -42,6 +44,8 @@ for updates, which means when you call flush on the EntityManager these entities
 even if properties changed. Read-Only allows to persist new entities of a kind and remove existing
 ones, they are just not considered for updates.
 
+See :ref:`annref_entity`
```
greg0ire: https://github.com/doctrine/orm/blob/b243ea1940d65d9ec3cc3045286726b7e79145b7/docs/en/reference/annotations-reference.rst#entity

docs/en/reference/improving-performance.rst
```diff
@@ -61,6 +65,7 @@ Apply Best Practices
 A lot of the points mentioned in the Best Practices chapter will
 also positively affect the performance of Doctrine.
 
+See :ref:`best-practices`
```
greg0ire: I'm not very familiar with rst, so I read the docs, shouldn't it be `:doc:` instead of `:ref`?

greg0ire (commented):
I added links for the lazy

docs/en/reference/improving-performance.rst
jschaedl: Mhh, I followed the `:ref` documentation (http://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-objects) which says:

> Using ref is advised over standard reStructuredText links to sections (like `Section title`_) because it works across files, when section headings are changed, will raise warnings if incorrect, and works for all builders that support cross-references. 

But you are right, I link to a document and not to a specific section inside it. So I will change it.

docs/en/reference/improving-performance.rst
jschaedl: After checking the docs again. I will add a reST label to the caching document so I can use `:ref:integrating-with-the-orm` here.

docs/en/reference/improving-performance.rst
```diff
@@ -26,6 +26,8 @@ Doctrine will need to load your mapping information on every single
 request and has to parse each DQL query on every single request.
 This is a waste of resources.
 
+See :ref:`integrating-with-the-orm`
```
jschaedl: https://github.com/doctrine/orm/blob/b243ea1940d65d9ec3cc3045286726b7e79145b7/docs/en/reference/caching.rst#integrating-with-the-orm

docs/en/reference/improving-performance.rst
```diff
@@ -61,6 +65,7 @@ Apply Best Practices
 A lot of the points mentioned in the Best Practices chapter will
 also positively affect the performance of Doctrine.
 
+See :doc:`Best Practices <reference/best-practices>`
```
jschaedl: https://github.com/doctrine/orm/blob/b243ea1940d65d9ec3cc3045286726b7e79145b7/docs/en/reference/best-practices.rst

docs/en/reference/improving-performance.rst
greg0ire: Can you test your changes locally? I'm not an expert in rst, so it's best if you just generate the docs and open them with a browser.

docs/en/reference/improving-performance.rst
```diff
@@ -61,8 +65,9 @@ Apply Best Practices
 A lot of the points mentioned in the Best Practices chapter will
 also positively affect the performance of Doctrine.
 
+See :doc:`Best Practices <best-practices>`
 
 Change Tracking policies
 ------------------------
 
-See: :doc:`Change Tracking Policies <reference/change-tracking-policies>`
+See: :doc:`Change Tracking Policies <change-tracking-policies>`
```
jschaedl: The folder name ist actually not needed.

docs/en/reference/improving-performance.rst
jschaedl: I created the documentation locally and it worked the new links worked for me.

docs/en/reference/improving-performance.rst
greg0ire: Oh wow that's weird… what if 2 files in different directories have the same name?

SenseException: As long as we don't know why it works, we should keep the folder name like in the other links.

jschaedl: Ok, I reverted that change and adjusted the other link.

lcobucci (approved):
@jschaedl awesome, thanks!

Issue #6032, opened 2016-09-14
Wrong sequence name generator


andreas4all:
Hi,

I created entity like this:

```
/**
 * @ORM\Entity
 */
class WirelessStatisticsServiceShortTerm
{
    /**
     * @var int
     *
     * @ORM\Column(name="wireless_statistic_id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="IDENTITY")
     */
    protected $id;

... shortened ...
}
```

Doctrine use this sequence:
```
wireless_statistics_service_short_term_wireless_statistic_id_se
```

and PostgreSQL generate this:
```
wireless_statistics_service_short_ter_wireless_statistic_id_seq
```

lcobucci:
Are you sure that "IDENTITY" is the strategy you want: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/basic-mapping.html#identifier-generation-strategies

andreas4all:
Yes. I want to use `IDENTITY`, because I want to use `SERIAL` in PostgreSQL.

lcobucci:
If you take a look on `PostgreSqlPlatform` you'll see that's correct: https://github.com/doctrine/dbal/blob/9f8c05cd5225a320d56d4bfdb4772f10d045a0c9/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php#L196

However `wireless_statistics_service_short_term_wireless_statistic_id_se` contains exactly 63 chars which is the value that `AbstractPlatform#getMaxIdentifierLength()` returns and `DBAL\Schema\Table` uses it to trim the identifier names.

I don't know exactly where's the problem (and if it's a problem) but it's definitely related to `DBAL` and not the `ORM`.

Can you reopen this on `DBAL` (maybe with some test cases)?

lcobucci:
BTW 63 is the max identifier length for PostgreSQL too: https://www.postgresql.org/docs/9.1/static/sql-syntax-lexical.html (so it's not a bug)

andreas4all:
yes, thats right. but doctrine strip last character of sequence name to limit. Postgresql generate it from table name and column name, but PG strip characters from table name not column name (or last character from generated sequence name). You can see it there is `_ter_` in Postgresql name, not `_term_`.

lcobucci:
Indeed, but it's still related to the `DBAL` and not the `ORM`, open that bug there to help us finding a way to organize things https://github.com/doctrine/dbal

PR #7673, opened 2019-04-08
Add support for SequenceGenerator in strategy IDENTITY

la-costa:
Doctrine already supports the SequenceGenerator annotation for GeneratedValue(strategy="SEQUENCE"), however it is not possible to indicate the sequence name when using strategy=IDENTITY.

Unfortunately, the sequence name is not generated correctly for PostgreSQL. #6032

This workaround allow indication of the sequence name, without breaking any compatibility.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
-  - nightly
+  - 7.4snapshot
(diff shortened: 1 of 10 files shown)


Ocramius (changes requested):
Changes like these require automation tests to be added

Issue #5763, opened 2016-04-08
Bug(?) when resetting IdGenerator(Type)


ghost:
Hi,

We need to have some fixture entities with set ids and therefore we do the following in the specified order:
- save old idGenerator and idGeneratorType
- change idGenerator and idGeneratorType to to `AssignedGenerator` and `ClassMetadata::GENERATOR_TYPE_NONE`,  respectively (see [here](http://stackoverflow.com/questions/5301285/explicitly-set-id-with-doctrine-when-using-auto-strategy) why)
- execute fixtures
- restore idGenerator and idGeneratorType to the old saved values (we want the default behavior from here on)
- remaing test code gets executed

And here is the problem: When test code is storing an entity, it now must assign a id (which is only wanted for fixtures), otherwise the following error gets thrown:

`An exception occurred while executing 'INSERT INTO foos (id, firstname, lastname) VALUES (?, ?, ?)' with params ["yo", "ya"]: SQLSTATE[HY000]: General error: 20 datatype mismatch`

It appears that id and it's value placeholder are still there, although the idGenerator and type have been reset and the params have adapted. Without executing the fixtures, the code works without any problems.

To help you reproduce this, I've setup a repo which reproduces exactly this error: https://github.com/whatda/bug

Just do a composer install and try it with phpunit. There's a `var_dump` and `die` in the single test, which currently isn't reached because of the error I pointed out.

Any help is really appreciated. Thanks for reading.

Majkl578:
This will be related to how mapping is built. ID generators are [completed in metadata factory](https://github.com/doctrine/doctrine2/blob/v2.5.10/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L612) and are not intended directly for changes at runtime. You would have to manually change the idGenerator instance of the class metadata, by calling i.e. `$classMetadata->setIdGenerator(new \Doctrine\ORM\Id\AssignedGenerator())`.
The code around generators has been heavily rewritten in develop (3.0), but this is still to be solved afaik.

b4cedev:
I'm having exactly the same problem. 
Changing idGenerator and generatorType has the desired effect so the fixtures can create entities using specified ids. 
However, restoring idGenerator and generatorType (by using the setters) to the previously saved values doesn't restore the previous behaviour, but triggers the error mentioned by @whatda.
As the tests are run against an in-memory sqlite db i can't simply load fixtures and run tests in separate processes. 
[...]

PR #7677, opened 2019-04-11
Fix: Allows changing the idGenerator in between persists

arnegroskurth:
MR for issue #5763

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -163,9 +163,17 @@ class BasicEntityPersister implements EntityPersister
      * The INSERT SQL statement used for entities handled by this persister.
      * This SQL is only generated once per request, if at all.
      *
-     * @var string
+     * @var null|string
      */
     private $insertSql;
+
+    /**
+     * List of columns of the currently cached insertSql.
+     * The insertSql is reset if another column list is given which happens e.g. when changing the id generator.
+     *
+     * @var null|string[]
+     */
+    private $insertSqlColumns;
 
     /**
      * The quote strategy.
@@ -1390,11 +1398,12 @@ class BasicEntityPersister implements EntityPersister
      */
     public function getInsertSQL()
     {
-        if ($this->insertSql !== null) {
+        $columns = $this->getInsertColumnList();
+
+        if ($this->insertSql !== null && $this->insertSqlColumns === $columns) {
             return $this->insertSql;
         }
 
-        $columns   = $this->getInsertColumnList();
         $tableName = $this->quoteStrategy->getTableName($this->class, $this->platform);
 
         if (empty($columns)) {
@@ -1419,6 +1428,8 @@ class BasicEntityPersister implements EntityPersister
 
             $values[] = $placeholder;
         }
+
+        $this->insertSqlColumns = $columns;
 
         $columns = implode(', ', $columns);
         $values  = implode(', ', $values);
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5763Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5763Test.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Id\AssignedGenerator;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH5763Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(GH5763Entity::class),
+        ]);
+    }
+
+    public function testInsertSqlCacheIsBustedOnColumnSetChange()
+    {
+        $metaData = $this->_em->getClassMetadata(GH5763Entity::class);
+
+        list($originalIdGenerator, $originalIdGeneratorType) = [$metaData->idGenerator, $metaData->generatorType];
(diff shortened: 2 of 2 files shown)


arnegroskurth:
@Ocramius ping - The build failure does not seem to be related to these changes

greg0ire:
Please format your commit message according to these rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

Bad example :

```
Fixed bug #989 by removing call to defraculate()
```

Good example:

```
Remove call to defraculate()

Calling this function caused a bug because it interferes 
with calls to getSchmeckles().
Fixes #989
```

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -167,6 +167,14 @@ class BasicEntityPersister implements EntityPersister
      */
     private $insertSql;
 
+    /**
+     * List of columns of the currently cached insertSql.
+     * The insertSql is reset if another column list is given which happens e.g. when changing the id generator.
+     *
+     * @var array
```
greg0ire: It this an array of column names, if yes, it should probably be
```suggestion
     * @var string[]`
```

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1420,6 +1429,8 @@ public function getInsertSQL()
             $values[] = $placeholder;
         }
 
+        $this->insertSqlColumns = $columns;
```
greg0ire: Would it make sense to sort that array?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH5763Test.php
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Id\AssignedGenerator;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH5763Test extends OrmFunctionalTestCase
+{
+    protected function setUp()
```
greg0ire: Let's be FW-compatible with phpunit 8
```suggestion
    protected function setUp() : void
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH5763Test.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(GH5763Entity::class),
+        ]);
+    }
+
+    public function testIssue()
```
greg0ire: Let's try to be more descriptive?
```suggestion
    public function testCacheIsBustedOnColumnChange()
```

Or is this the recommended format for tests starting with GH?

greg0ire (changes requested):

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
arnegroskurth: I'm not sure about potential problems but what would be the benefit? If the column order changed in between calls then the current implementation would now work as it kept the column order fixed and if the order does not change than sorting does not have any benefit.

greg0ire: If the column order changed then the cache would be reused. That's the benefit.

arnegroskurth: @greg0ire Sure, but that would not work as the column order has to match the given parameter order as the question-mark placeholders are used.

greg0ire: This answers my question, thanks!

arnegroskurth:
@greg0ire Then it seems that everything is resolved!

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -167,6 +167,14 @@ class BasicEntityPersister implements EntityPersister
      */
     private $insertSql;
 
+    /**
+     * List of columns of the currently cached insertSql.
+     * The insertSql is reset if another column list is given which happens e.g. when changing the id generator.
+     *
+     * @var string[]
```
Ocramius: There's still a problem about the state of this: this is `string[]|null`, as otherwise line 1403 is controversial (checking against `null`)

Ocramius (changes requested):

greg0ire (dismissed):

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
arnegroskurth: @Ocramius I changed the typehint for `$this->insertSql` as well to be consistent.

SenseException (changes requested):

arnegroskurth:
@SenseException Resolved your issues

SenseException (approved):

arnegroskurth:
@Ocramius ping

lcobucci:
Requesting @guilhermeblanco's review here to see how does this PR fit in the newer metadata API too.

guilhermeblanco:
Class Metadata mapping (and its derivatives, such as Property mapping, Identity mapping, etc) are immutable and any intent of modifying it at runtime is strictly not supported.

-1 from me.

guilhermeblanco (changes requested):
-1 from me

vpArth:
> any intent of modifying it at runtime is strictly not supported

Then need to restrict or add some warnings on ClassMetadataInfo mutation methods.

PR #7684, opened 2019-04-17
only replace '_id' at end of columnName

rharink:
I've encountered a problem with foreign-key column-names that have a '_id' (e.g. _identity) in their name.
These columns don't get renamed properly (`partner_identity_card` -> `partnerentityCard`).

The fix could be a `preg_replace` instead of a `string_replace` that only replaces the _id at the end of a column name.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7684Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7684Test.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\Tests\ORM\Functional\DatabaseDriverTestCase;
+
+/**
+ * Verifies that associations/columns with an inline '_id' get named properly
+ *
+ * Github issue: 7684
+ */
+class GH7684 extends DatabaseDriverTestCase
+{
+    public function testIssue() : void
+    {
+        if (! $this->_em->getConnection()->getDatabasePlatform()->supportsForeignKeyConstraints()) {
+            $this->markTestSkipped('Platform does not support foreign keys.');
+        }
+
+        $table1 = new Table('GH7684_identity_test_table');
+        $table1->addColumn('id', 'integer');
+        $table1->setPrimaryKey(['id']);
+
+        $table2 = new Table('GH7684_identity_test_assoc_table');
+        $table2->addColumn('id', 'integer');
+        $table2->addColumn('gh7684_identity_test_id', 'integer');
+        $table2->setPrimaryKey(['id']);
+        $table2->addForeignKeyConstraint('GH7684_identity_test', ['gh7684_identity_test_id'], ['id']);
+
+        $metadatas = $this->convertToClassMetadata([$table1, $table2]);
+        $metadata  = $metadatas['Gh7684IdentityTestAssocTable'];
+
+        $this->assertArrayHasKey('gh7684IdentityTest', $metadata->associationMappings);
+    }
+}
--- a/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
@@ -29,6 +29,7 @@ use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Mapping\MappingException;
+use function preg_replace;
 
 /**
  * The DatabaseDriver reverse engineers the mapping metadata from a database.
@@ -548,7 +549,7 @@ class DatabaseDriver implements MappingDriver
 
         // Replace _id if it is a foreignkey column
         if ($fk) {
-            $columnName = str_replace('_id', '', $columnName);
+            $columnName = preg_replace('/_id$/', '', $columnName);
         }
 
         return Inflector::camelize($columnName);


rharink:
Could someone point me in the right direction for creating a test-case? I would need a (mock) driver that supports foreign-keys

Ocramius:
@rharink check https://github.com/doctrine/doctrine2/blob/master/tests/Doctrine/Tests/ORM/Functional/Ticket for examples of test cases.

I'd suggest a scenario that runs on SQLite only and:

 1. creates a table
 2. instantiates the `DatabaseDriver`
 3. verifies the schema information coming from `DatabaseDriver`

rharink:
@Ocramius Thanks for the reply,

The SqlLite driver doesn't support foreign keys. The `DatabaseDriver` class will not resolve foreign keys, thus the method `getFieldNameForColumn` is never called in the testcase. 

returns `false` for Sqllite
```
$this->_em->getConnection()->getDatabasePlatform()->supportsForeignKeyConstraints())
```

Ocramius:
True, would need to design this with MySQL or equivalent

rharink:
Should I use a `DbalFunctionalTestCase` or is there something else I can use to setup a MySQL test?
Thanks for the help!

rharink:
Sorry should read more before asking ;) I found documentation on how to run tests on mysql.

rharink:
@Ocramius Checks have passed, should / can I do anything else?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7684Test.php
```diff
@@ -0,0 +1,47 @@
...
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\ORM\Mapping\MappingException;
+use Doctrine\Tests\ORM\Functional\DatabaseDriverTestCase;
+
+/**
+ * Verifies that associations/columns with an inline '_id' get named properly
+ * 
+ * Github issue: 7684
+ */
+class GH7684 extends DatabaseDriverTestCase
```
Ocramius: Where are other `DatabaseDriverTest` implementations located?

Ocramius: (asking because we may get to structure this better, now that we know that the `DatabaseDriver` already has existing tests in place)

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7684Test.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    public function testIssue() : void
+    {
+        if (!$this->_em->getConnection()->getDatabasePlatform()->supportsForeignKeyConstraints()) {
+            $this->markTestSkipped('Platform does not support foreign keys.');
+        }
+
+        $card = new Table('partner_identity_card');
```
Ocramius: Could we drop the domain naming around this, and use test-specific naming for columns/tables? This would remove focus from your specific domain, and onto the issue details

Ocramius (changes requested):
Overall looking good: needs adjustments in test naming and placing (need to check if we can turn this into a full unit tests, eventually - not sure if that is feasible)

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7684Test.php
```diff
@@ -0,0 +1,47 @@
...
+        $metadata = $this->convertToClassMetadata([$card, $translation]);
+
+        $this->assertNotNull($metadata['PartnerIdentityCard'] ?? null);
+        $this->assertNotNull($metadata['PartnerIdentityCardTranslation'] ?? null);
+        
+        $partnerIdentityCardTranslation = $metadata['PartnerIdentityCardTranslation'];
+        
+        try {
+            $partnerIdentityCardTranslation->getAssociationMapping('partnerIdentityCard');
+        } catch (MappingException $e) {
```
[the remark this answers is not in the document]
rharink: Done

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7684Test.php
rharink: Yeah, removed the domain specific stuff

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7684Test.php
rharink: Should I create a test in `DatabaseDriverTests.php` something like the existing `testLoadMetadataWithForeignKeyFromDatabase` ?

rharink:
@Ocramius can I do something else? maybe add a unit test in the databaseDriverTest.php? Just let me know

Ocramius (dismissed):
LGTM :+1: 

@rharink can you rebase on top of `upstream/2.6` so that we can see if the CI failures are related to this patch at all?

Besides that, just needs a merge from our end.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7684Test.php
```diff
@@ -0,0 +1,43 @@
...
+        $metadatas = $this->convertToClassMetadata([$card, $translation]);
+
+        $this->assertNotNull($metadatas['Gh7684IdentityTest'] ?? null);
+        $this->assertNotNull($metadatas['Gh7684IdentityTestAssoc'] ?? null);
+        
+        $metadata= $metadatas['Gh7684IdentityTestAssoc'];
+        
+        $metadata->getAssociationMapping('gh7684IdentityTest');
+    }
+}
```
greg0ire: Please configure your IDE to always add a newline at end of file, this will make the red pictogram disappear.

rharink: Added a newline, will do, thanks!

rharink:
Ok, I've merged upstream.

greg0ire:
I rebased, I fixed the build in #7731, this means the tests might now pass.

ostrolucky (changes requested):
Test case is unfortunately not correct. It passes for previous behaviour too.

greg0ire:
Wow good catch! Well at least the build is green now :sweat_smile:
 
@rharink please look into this :pray:

rharink:
Good one! Ill look in to it.

lcobucci:
@ostrolucky nice catch indeed, thanks!

@rharink since we're aiming at releasing the last patch of the `2.6.x` series and release `2.7.0` I'm moving this to `2.7.1` so we don't put a lot of pressure on your shoulders :smile: 

Thanks for putting effort into this and I hope we can solve it together in the next patch :+1:

rharink:
I've improved the unit-test. it didn't pass for previous behaviour as the key `gh7684IdentityTest` in the `associationMapping` would not have been present. Or am i interpreting 'previous behaviour' wrong :sweat_smile:

ostrolucky (dismissed):
Test is correct now. @lcobucci I would say we can merge this now to 2.6.5 what do you say?

greg0ire:

How to do that?

1. `git rebase -i origin/2.6`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

lcobucci:
@rharink @ostrolucky awesome, thanks for being that quick. Please follow @greg0ire's instructions and let's ship this!

rharink:
@lcobucci @ostrolucky done, lets go!

lcobucci (approved):

lcobucci:
@rharink :ship:

PR #7691, opened 2019-04-23
Add another static analysis tool, grandfathering in existing issues

muglug:
There are a bunch of different ways to improve the code quality of a large project.

One way that's worked reasonably well at `$workplace` is to grandfather in a bunch of existing issues, such that any new code must abide by higher standards than existing code.

We do this with a baseline file - it records the current issues with the codebase, with some sort of marker to identify where in the file the issue happened. New code that triggers similar issues presumably has different markers, so a new specific error is emitted.

One benefit is that you don't have to fix a single type of issue all in one go - you can make incremental changes, and see the size of the generated baseline file shrink over time.

This approach isn't for everyone, or every project. I leave it up to you to decide if this is a good fit for Doctrine.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,3 +12,4 @@ phpunit.xml.dist export-ignore
 run-all.sh export-ignore
 phpcs.xml.dist export-ignore
 composer.lock export-ignore
+psalm-baseline.xml export-ignore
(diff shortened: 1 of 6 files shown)


SenseException:
Is this a WIP or can this PR be reviewed? AFAIK this is supposed to replace your previous PR you've closed yourself.

greg0ire:
Done in #8627, thanks for the suggestion!

PR #7692, opened 2019-04-23
Test to reproduce bug related to composite keys and EAGER fetch

mnapoli:
This PR adds a test that reproduces a bug. I have no idea yet on how to fix it, suggestions are welcome.

### Scenario

`A` (manyToOne)-> `B`.

- a row (table `A`) exists in database with foreign keys pointing to a row `B` that doesn't exist
- `B` has a composite primary key
- the relation has `fetch="EAGER"`

### Expected result

Since the target entity doesn't exist `null` should be loaded instead. That is the case for all the other scenarios.

### Actual result

The following error:

```
Warning: spl_object_hash() expects parameter 1 to be object, null given
```

in lib/Doctrine/ORM/UnitOfWork.php:2809

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7692Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7692Test.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH7692Test extends OrmFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH7692Project::class,
+            GH7692Contact::class,
+        ]);
+    }
+
+    public function testWrongForeignKeysInDatabaseAreHandledByDoctrine(): void
+    {
+        // Create a row that references missing rows
+        $this->_em->getConnection()->insert('project', [
+            // This composite foreign key doesn't exist
+            'contact_category' => 999,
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH7692Test.php
```diff
@@ -0,0 +1,75 @@
...
+            GH7692Project::class,
+            GH7692Contact::class,
+        ]);
+    }
+
+    public function testWrongForeignKeysInDatabaseAreHandledByDoctrine(): void
+    {
+        // Create a row that references missing rows
+        $this->_em->getConnection()->insert('project', [
+            // This composite foreign key doesn't exist
```
Ocramius: This seems to be an impossible scenario from a relational DB perspective. In practice, you'd get a proxy lazy-loading exception when not operating with `EAGER` loading

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7692Test.php
mnapoli: Understood, I was under the false impression that it worked without eager. It actually throws an exception like you say.

However the exception makes sense: it's a `EntityNotFoundException`.

I have updated the PR:

- added an additional test for a scenario that currently works (throws `EntityNotFoundException`)
- updated the existing test to assert that a `EntityNotFoundException` is thrown

The test still fails as the current `spl_object_hash` warning isn't what we want.

mnapoli: I have given a try to a fix (to have a proper exception).

mnapoli:
Cleaning up my pending PRs 😄

PR #7213, opened 2018-05-06
Split and deprecate AbstractQuery#useResultCache()

someniatko:
Split `useResultCache($boolFlag, ...)` into `enableResultCache()` and `disableResutCache()`.

To create tests for `enableResultCache()` and `disableResultCache()`, `useResultCache()`'s tests were copy-pasted, but I think that's fine.

PR #7214, opened 2018-05-07
Split and remove AbstractQuery#useResultCache()

someniatko:
Ported changes from https://github.com/doctrine/doctrine2/pull/7213, but also removed the deprecated useResultCache() method and added an entry in UPGRADE.md.

PR #7701, opened 2019-05-07
Split and deprecate AbstractQuery#useResultCache()

someniatko:
Split `useResultCache($boolFlag, ...)` into `enableResultCache()` and `disableResutCache()`.

To create tests for `enableResultCache()` and `disableResultCache()`, `useResultCache()`'s tests were copy-pasted, but I think that's fine.

This is proper PR, original one is https://github.com/doctrine/orm/pull/7213

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -583,21 +583,45 @@ abstract class AbstractQuery
      * Set whether or not to cache the results of this query and if so, for
      * how long and which ID to use for the cache entry.
      *
-     * @param boolean $bool
-     * @param integer $lifetime
-     * @param string  $resultCacheId
+     * @deprecated 2.7 Use {@see enableResultCache} and {@see disableResultCache} instead.
+     *
+     * @param bool   $useCache
+     * @param int    $lifetime
+     * @param string $resultCacheId
      *
      * @return static This query instance.
      */
-    public function useResultCache($bool, $lifetime = null, $resultCacheId = null)
+    public function useResultCache($useCache, $lifetime = null, $resultCacheId = null)
     {
-        if ($bool) {
-            $this->setResultCacheLifetime($lifetime);
-            $this->setResultCacheId($resultCacheId);
+        return $useCache
+            ? $this->enableResultCache($lifetime, $resultCacheId)
+            : $this->disableResultCache();
+    }
 
-            return $this;
-        }
+    /**
+     * Enables caching of the results of this query, for given or default amount of seconds
+     * and optionally specifies which ID to use for the cache entry.
+     *
+     * @param int|null    $lifetime      How long the cache entry is valid, in seconds.
+     * @param string|null $resultCacheId ID to use for the cache entry.
+     *
+     * @return static This query instance.
+     */
+    public function enableResultCache(?int $lifetime = null, ?string $resultCacheId = null) : self
+    {
+        $this->setResultCacheLifetime($lifetime);
+        $this->setResultCacheId($resultCacheId);
 
+        return $this;
+    }
+
+    /**
+     * Disables caching of the results of this query.
+     *
+     * @return static This query instance.
+     */
+    public function disableResultCache() : self
+    {
         $this->_queryCacheProfile = null;
 
         return $this;
--- a/tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php
@@ -89,7 +89,7 @@ class ResultCacheTest extends OrmFunctionalTestCase
         $this->assertTrue($cache->contains('testing_result_cache_id'));
     }
 
-    public function testUseResultCache()
+    public function testUseResultCacheTrue()
     {
         $cache = new ArrayCache();
         $query = $this->_em->createQuery('select ux from Doctrine\Tests\Models\CMS\CmsUser ux');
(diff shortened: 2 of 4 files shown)


greg0ire:
IMO you should fixup the first commit into the new one. Do you know how to do this without creating a new PR? Otherwise I can help.

someniatko:
@greg0ire yes. Last time i recreated the PR because a) initial branch was wrong (i committed directly to 2.7 of my fork), and b) interactive rebase rendered in merge conflicts for some reason, even if i tried to rebase without changing anything in commits order. It was easier to just start over and cherry-pick.

Okay, will rebase and force push now.

someniatko:
@greg0ire ping. I noticed you've put "Deprecation" label and "2.7.0" milestone to the previous PR, https://github.com/doctrine/orm/pull/7213. I think you should at least do the same for this one.

tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php
```diff
@@ -133,6 +149,65 @@ public function testUseResultCacheParams()
...
+        $this->assertCount($sqlCount + 2, $this->_sqlLoggerStack->queries, "Two non-cached queries.");
+
+        $query->setParameter(1, 1);
+        $query->enableResultCache();
+        $query->getResult();
+
+        $query->setParameter(1, 2);
+        $query->getResult();
+
+        $this->assertCount($sqlCount + 2, $this->_sqlLoggerStack->queries, "The next two sql should have been cached, but were not.");
```
greg0ire: - The next two sql what? queries?
- Why "next"? There is nothing after this line of code…

someniatko: @greg0ire These tests were copied from original test for `useResultCache()` method

greg0ire: Ok, I took more time to read and I understand now: there are 2 queries, then a first assertion, and then 2 queries, so "next" means "after the 2 first queries" (it's the next set of 2 queries). I think you should add the missing word:

```suggestion
        $this->assertCount(
            $sqlCount + 2,
            $this->_sqlLoggerStack->queries,
            "The next two sql queries should have been cached, but were not."
        );
```

Another thing I don't understand about this test (and yours) is why `useResultCache(true)` and `enableResultCache()` are called twice on the same object. Do you maybe know?

someniatko: @greg0ire as long as removing the line with the second `enableResultCache()` (or `useResultCache(true)` does not change the behavior of the test, apparently it is just a mistake made by original test writer during copy-paste. I'd also reposition the first `$query->setParameter()` call to be closer to `$query->getResult()` as it is probably sematically more correct. Will do both yours and mine suggested changes in the next commit.

greg0ire (dismissed):
It reads far better now, great work!

greg0ire:
Are there no docs that need to be updated?

jwage:
The deprecation should be added to the UPGRADE.md document too?

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -583,21 +583,45 @@ public function getResultCacheDriver()
      * Set whether or not to cache the results of this query and if so, for
      * how long and which ID to use for the cache entry.
      *
-     * @param boolean $bool
+     * @deprecated 3.0 Use {@see enableResultCache} and {@see disableResultCache} instead.
```
SenseException:
```suggestion
     * @deprecated Will be removed in 3.0. Use {@see enableResultCache} and {@see disableResultCache} instead.
```

In case this will be removed in 3.0.

SenseException (changes requested):
`useResultCache` appears in https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/caching.html#result-cache and needs to be mentioned/adapted for 2.7.

lib/Doctrine/ORM/AbstractQuery.php
someniatko: @SenseException https://docs.phpdoc.org/references/phpdoc/tags/deprecated.html

someniatko:
Got it, will add the requested changes to documentation and `UPGRADE.md`

someniatko:
Made the requested changes to documentation. Also i'd like to reject @SenseException 's suggestion for the reasons described above.

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: TIL. Is there tooling that takes this into account? Symfony's profiler maybe?

someniatko: I have no idea. Anyway, IMO it's at least future-proof to stick to existing standard.

SenseException: I've looked into other Doctrine projects and found a few deprecation entries that follow the documented phpdoc pattern. I think we can keep it as you proposed in your PR.

SenseException (dismissed):
Looks good to me.

Will those changes be merged or picked into master or is another PR for master needed?

someniatko:
@SenseException there is a pull request to master: https://github.com/doctrine/orm/pull/7214
It will need some work as well, like updating documentation. However I'd wait for this pull request to be merged before working on master's one.

lcobucci (approved):
@someniatko thanks!

PR #1219, opened 2014-12-10
Comparison like/notlike support

Fedik:
Support `Comparison::LIKE` and `Comparison::NOTLIKE` additionally to doctrine/collections#50

as alternative for #1150

PR #7705, opened 2019-05-08
Add info on Criteria missing methods

voltel:
Add info from StackOverflow post on missing "isNotNull" and "isLike" methods.

### DIFF
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -725,3 +725,26 @@ methods:
     There is a limitation on the compatibility of Criteria comparisons.
     You have to use scalar values only as the value in a comparison or
     the behaviour between different backends is not the same.
+
+
+.. note::
+
+   To have an **is not null** that acts like a *Criteria::expr()->isNotNull('field')* you can use
+
+.. code-block:: php
+
+   <?php
+   
+   $criteria = Criteria::create();
+   $expr = $criteria::expr();
+   $collection->matching($criteria->where($expr->neq('field', null)));
+
+For the *Criteria::expr()->like()* functionality, **Criteria::expr()->contains('property', 'value')** is the equivalent of the SQL property LIKE %value%. However it does not allow for changing to value% or %value (may be changed in future versions).
+
+Unfortunately for *Criteria::expr()->notLike()* and the other LIKE variants, the \Doctrine\Common\Collections\ExpressionBuilder used by Criteria does not support them. 
+
+.. note::
+
+   Read this StackOverflow post(https://stackoverflow.com/questions/26653651/collections-criteriaexpr-isnotnull-and-notlike) for filtering collections. 
+
+


SenseException (changes requested):
Thank you for your contribution.

The `Criteria` class is part of the Doctrine Collections project/repository and information about it should be written in the [Doctrine Collections documentation](https://www.doctrine-project.org/projects/doctrine-collections/en/current/expression-builder.html#expression-builder). The methods there lack a description you offer in this PR, which is why I would appreciated it if you could put the descriptions as a new PR in [doctrine/collections](https://github.com/doctrine/collections).

Please don't use terms like "may be changed in future versions" because software is always changing and a documentation is about "what currently is" and not "what might be". Don't write function names as code that don't exist in a documentation.

* Please move descriptions of the `Criteria` methods to the collections documentation (new PR).
* Put a link to the doctrine/collections documentation I provided earlier here to prevent redundancy between documentations of different Doctrine projects. This ORM PR of yours will only add the referring link to the doctrine/collections documentation with a hint like "for more information about Criteria (and so on)".

docs/en/reference/working-with-associations.rst
```diff
@@ -725,3 +725,26 @@ methods:
...
+
+   <?php
+   
+   $criteria = Criteria::create();
+   $expr = $criteria::expr();
+   $collection->matching($criteria->where($expr->neq('field', null)));
+
+For the *Criteria::expr()->like()* functionality, **Criteria::expr()->contains('property', 'value')** is the equivalent of the SQL property LIKE %value%. However it does not allow for changing to value% or %value (may be changed in future versions).
+
+Unfortunately for *Criteria::expr()->notLike()* and the other LIKE variants, the \Doctrine\Common\Collections\ExpressionBuilder used by Criteria does not support them.
```
Ocramius: Do we have an issue for this?

docs/en/reference/working-with-associations.rst
```diff
@@ -725,3 +725,26 @@ methods:
...
+   $expr = $criteria::expr();
+   $collection->matching($criteria->where($expr->neq('field', null)));
+
+For the *Criteria::expr()->like()* functionality, **Criteria::expr()->contains('property', 'value')** is the equivalent of the SQL property LIKE %value%. However it does not allow for changing to value% or %value (may be changed in future versions).
+
+Unfortunately for *Criteria::expr()->notLike()* and the other LIKE variants, the \Doctrine\Common\Collections\ExpressionBuilder used by Criteria does not support them. 
+
+.. note::
+
+   Read this StackOverflow post(https://stackoverflow.com/questions/26653651/collections-criteriaexpr-isnotnull-and-notlike) for filtering collections.
```
Ocramius: Unlikely to be added here - instead, we should expand this as docs in here, if relevant.

Ocramius (changes requested):

docs/en/reference/working-with-associations.rst
Fedik: even a pull request #1219

docs/en/reference/working-with-associations.rst
```diff
@@ -725,3 +725,26 @@ methods:
     There is a limitation on the compatibility of Criteria comparisons.
     You have to use scalar values only as the value in a comparison or
     the behaviour between different backends is not the same.
+
+
+.. note::
+
+   To have an **is not null** that acts like a *Criteria::expr()->isNotNull('field')* you can use
```
greg0ire: I think you miss 2 words here:

To have an **is not null** $word1 that acts like a *Criteria::expr()->isNotNull('field')* $word2 you can use

Also, a colon at the end would be nice IMO

docs/en/reference/working-with-associations.rst
```diff
@@ -725,3 +725,26 @@ methods:
...
+
+.. code-block:: php
+
+   <?php
+   
+   $criteria = Criteria::create();
+   $expr = $criteria::expr();
+   $collection->matching($criteria->where($expr->neq('field', null)));
```
greg0ire: Does this php code block render inside the note?

docs/en/reference/working-with-associations.rst
greg0ire: Also, please use double backticks around pieces of code

docs/en/reference/working-with-associations.rst
```diff
@@ -725,3 +725,26 @@ methods:
...
+
+For the *Criteria::expr()->like()* functionality, **Criteria::expr()->contains('property', 'value')** is the equivalent of the SQL property LIKE %value%. However it does not allow for changing to value% or %value (may be changed in future versions).
```
greg0ire:
```suggestion
For the ``Criteria::expr()->like()`` functionality, ``Criteria::expr()->contains('property', 'value')`` is the equivalent of the SQL syntax ``LIKE %value%``. However it does not allow for changing to `value%` or `%value` (may be changed in future versions).
```

docs/en/reference/working-with-associations.rst
greg0ire:
```suggestion
Unfortunately for ``Criteria::expr()->notLike()`` and the other ``LIKE`` variants, the ``\Doctrine\Common\Collections\ExpressionBuilder`` used by Criteria does not support them. 
```

greg0ire (changes requested):

PR #7716, opened 2019-05-22
Prevents unexpected errors from bubbling up when attempting to 

paul-court:
autodiscover a cache provider.

The worst case scenario for this situation is when the cache extension *is* installed, but the service is not running on the localhost. For example, our app was attempting to setup doctrine as follows:-

```php
            $config = Setup::createAnnotationMetadataConfiguration([
                __DIR__ . DIRECTORY_SEPARATOR,
            ], $isDevMode);

            if ($isDevMode) {
                // In development we don't want anything being cached for longer than a single request.
                $cache = new ArrayCache();
            } else {
                // Allow Doctrine to cache queries and metadata using Redis
                $redisCredentials = $this->getRedisCredentials();

                $redis = new \Redis();
                $redis->connect((string)$redisCredentials->getHost(), $redisCredentials->getPort()->value());
                $redis->setOption(\Redis::OPT_PREFIX, 'AUTHAPI:' . $this->getAppVersion() . ':');

                $cache = new \Doctrine\Common\Cache\RedisCache();
                $cache->setRedis($redis);
            }

            $config->setQueryCacheImpl($cache);
            $config->setResultCacheImpl($cache);
            $config->setMetadataCacheImpl($cache);
```

Given that the error message in our stack trace was:-

```
PHP Warning:  Redis::connect(): connect() failed: Connection refused in /var/www/vhosts/auth-api/vendor/doctrine/orm/lib/Doctrine/ORM/Tools/Setup.php on line 185
PHP Fatal error:  Uncaught RedisException: Redis server went away in /var/www/vhosts/auth-api/vendor/doctrine/cache/lib/Doctrine/Common/Cache/RedisCache.php:28
```

It took us a long time to discover that it was the call to `Setup::createAnnotationMetadataConfiguration` which was attempting to create a connection to a local redis instance which doesn't exist (We're using ElastiCache on AWS).

An alternate approach might be to throw a more helpful error message.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Setup.php
+++ b/lib/Doctrine/ORM/Tools/Setup.php
@@ -165,29 +165,47 @@ class Setup
             return new ArrayCache();
         }
 
-        if (extension_loaded('apcu')) {
-            return new \Doctrine\Common\Cache\ApcuCache();
-        }
+        try {
 
+            /*
+            For backwards-compatability, attempt to create to a caching provider with
+            default settings if that providers extension is installed.
+            */
 
-        if (extension_loaded('memcached')) {
-            $memcached = new \Memcached();
-            $memcached->addServer('127.0.0.1', 11211);
+            if (extension_loaded('apcu')) {
+                return new ApcuCache();
+            }
 
-            $cache = new \Doctrine\Common\Cache\MemcachedCache();
-            $cache->setMemcached($memcached);
+            if (extension_loaded('memcached')) {
+                $memcached = new Memcached();
+                $memcached->addServer('127.0.0.1', 11211);
 
-            return $cache;
-        }
+                $cache = new MemcachedCache();
+                $cache->setMemcached($memcached);
 
-        if (extension_loaded('redis')) {
-            $redis = new \Redis();
-            $redis->connect('127.0.0.1');
+                return $cache;
+            }
 
-            $cache = new \Doctrine\Common\Cache\RedisCache();
-            $cache->setRedis($redis);
+            if (extension_loaded('redis')) {
+                $redis = new Redis();
+                $redis->connect('127.0.0.1');
 
-            return $cache;
+                $cache = new RedisCache();
+                $cache->setRedis($redis);
+
+                return $cache;
+            }
+            
+        } catch (\Exception $ex) {
+
+            /*
+            Deliberate catch of top level \Exception in the event that an extension IS installed, but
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Tools/Setup.php
```diff
@@ -165,29 +165,47 @@ private static function createCacheInstance(bool $isDevMode, ?Cache $cache) : Ca
             return new ArrayCache();
         }
 
-        if (extension_loaded('apcu')) {
-            return new \Doctrine\Common\Cache\ApcuCache();
-        }
+        try {
 
+            /*
+            For backwards-compatability, attempt to create to a caching provider with
```
```suggestion
            For backwards-compatability, attempt to create a caching provider with
```

greg0ire:
> An alternate approach might be to throw a more helpful error message.

I would like that approach way better, because the attempt to connect will still be there and waste time.

lib/Doctrine/ORM/Tools/Setup.php
```diff
@@ -165,29 +165,47 @@ private static function createCacheInstance(bool $isDevMode, ?Cache $cache) : Ca
...
+                $redis->connect('127.0.0.1');
 
-            return $cache;
+                $cache = new \Doctrine\Common\Cache\RedisCache();
+                $cache->setRedis($redis);
+
+                return $cache;
+            }
+            
+        } catch (\Exception $ex) {
```
```suggestion
        } catch (\Throwable $ex) {
```

greg0ire (changes requested):
Please create a new exception class just for this, with a clear error message. Also, I think that feature is bad and should be deprecated.

gargoyle:
If you want to keep it fully BC, you are going to have to attempt the connections anyway to allow for the cases where it might have worked.

I'll see what I can do from just the internals of this class. I really have no knowledge about the other internal workings of Doctrine.

Ocramius (changes requested):
I'd probably deprecate any public API on this class that is also leading to that `private` method to be called. This kind of defaults cannot be changed easily (sneaky and very problematic BC break), and defaults like these would be rejected with the current code review process that we have.

Deprecation for 2.x, and removal for 3.x would be best.

gargoyle:
I've wired up a custom exception. This should be the closest to BC as possible - Instead of a PHP fatal error a custom exception is thrown.

I have not marked anything as deprecated because all but one of the public methods lead to `createCacheInstance` being called and I do not know enough about Doctrine to figure out what the proposed alternative should be.

I think someone else will need to pick-up if more needs to be done.

Hope this helps. :-)

PR #7719, opened 2019-05-22
Fix a few examples: Start i var from 1 instead of 0

bocharsky-bw:
Because `(0 % $batchSize) === 0` but we don't want to execute `flush()` and `clear()` on the first iteration

### DIFF
--- a/docs/en/reference/batch-processing.rst
+++ b/docs/en/reference/batch-processing.rst
@@ -74,7 +74,7 @@ with the batching strategy that was already used for bulk inserts:
 
     <?php
     $batchSize = 20;
-    $i = 0;
+    $i = 1;
     $q = $em->createQuery('select u from MyProject\Model\User u');
     $iterableResult = $q->iterate();
     foreach ($iterableResult as $row) {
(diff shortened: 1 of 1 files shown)


malukenho:
Maybe you could add that extra piece of information that you've put on the PR description to the commit message itself?

bocharsky-bw:
I could, but do you think we really need it? I mean, it's kinda obvious, and we actually start from 1 not 0 on the 1st example but somehow use 0 in further examples.

greg0ire:
> it's kinda obvious

Is it really though? Since you felt the need to add it to the PR description (which was great IMO)

> I could, but do you think we really need it?

We don't need it, but it would be nice to amend your commit message to include it. Also, in case you do not know, Github will usually reuse the commit message *body* as a pull request body.

greg0ire (dismissed):

greg0ire:
Ah apparently I have the right to do this for you :)

greg0ire (approved):

malukenho (approved):

bocharsky-bw:
OK, thanks for your help on it, guys :)

Let me know if you think we need more changes here

Ocramius (approved):
LGTM :+1:

bocharsky-bw:
It also should be done in 2.6 and 2.7 branches, I created a new PR for 2.6: #7742 - should I create one more for 2.7? Or it will be merged from 2.6?

greg0ire:
We will be able to automate merging up soon, so don't worry, I think 2.6 is enough.

bocharsky-bw:
Good to know! Sorry for sending the initial PR to master instead of 2.6

greg0ire:
No problem, we will probably also change the default branch.

Issue #7721, opened 2019-05-24
Unique index generated for one-to-one relation causes errors when replacing entities.


LukasKrc:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

#### Summary

A unique index error is caused when replacing entities (that is removing an existing relation and creating a new one) from the inverse side of a one-to-one relation.

#### Current behavior

When replacing an entity from the inverse side of a one-to-one relation, a unique index error is caused. This happens because the previously persisted related entity is not deleted before an attempt to insert the new one is made.

This was supposedly fixed in a previous issue: #2310. There is another issue which is partially related to this one: #6776.

#### How to reproduce

Please see the referenced pull request with an example test case.

#### Expected behavior

The expected behavior is that no unique index error is caused during the mentioned scenario.

LukasKrc:
#7722

PR #7722, opened 2019-05-24
Adding a test case indicating that certain scenarios involving one-to…

LukasKrc:
…-one relations cause unique index errors.

#7721

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7719Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7719Test.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Annotation as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH7719Test extends OrmFunctionalTestCase
+{
+    protected function setUp() : void
+    {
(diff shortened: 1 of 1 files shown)


beberlei:
Preliminary findings and potential fix approach:

1. yes this bug always existed, we usually recommend double flush before (sucks yeah).
2. problem here is mysql checking unique constraints every row, not during commit: https://dev.mysql.com/doc/refman/5.7/en/innodb-foreign-key-constraints.html
3. solution would need to have a flag on AbstractPlatform (`supportsDeferredUniqueProcessingDuringCommit()`). I believe Postgresql does it that way.
4. If deferred unique processing not supported, then for a bidirectional one to one, computeChangeSet should detect the change and schedule an extra update to set the foreign key empty.

lcobucci:
> 3. solution would need to have a flag on AbstractPlatform (supportsDeferredUniqueProcessingDuringCommit()). I believe Postgresql does it that way.
> 4. If deferred unique processing not supported, then for a bidirectional one to one, computeChangeSet should detect the change and schedule an extra update to set the foreign key empty.

That's quite correct but I think we can add the support for deferred stuff for v2.8.0 (since that would require a bump on DBAL). It would be awesome if we could solve the update/delete during `computeChangeSet` but I'm not sure if that's feasible since both updates and deletes are processed afterwards in `UoW#commit().`

What we could do is try to handle this similar to how collection deletes are treated (as in pulling the deletion before of the insertion).

PR #7724, opened 2019-05-29
Add UnderscorePluralNamingStrategy

zorn-v:
Converts 'MyEntity' to 'my_entities' or 'MY_ENTITIES'.


SenseException (changes requested):
Thank you for your contribution. I'm not sure if this feature will be added to the core code of ORM. It is basically the `UnderscoreNamingStrategy` code with a pluralisation.

I would expect it to allow other strategies to be used for pluralisation too instead for only the underscore one. This PR is also missing tests.

Further reviews are required before the new strategy gets accepted by other reviewers.

zorn-v:
Helooooo. No missing tests about month

SenseException:
> I would expect it to allow other strategies to be used for pluralisation too instead for only the underscore one.

Thanks for adding the tests. As mentioned before this class is a copy of the UnderscoreNamingStrategy and therefore duplicated code. Can this pluralisation also work for other naming strategies?

zorn-v:
> Can this pluralisation also work for other naming strategies?

You can ask it of architect designer of doctrine. I think no. You must create another *NamingStrategy.
But who names tables with no "underscore plural" now ? 
Yep, philosophical question :smile:

zorn-v:
Main purpose of this PR is to make possible configurable `things` instead of `thing` (table has MANY entities) default table names in symfony doctrine bundle. And do not provide `@ORM\Table(name="things")` in EACH entity

greg0ire:
> You can ask it of architect designer of doctrine ORM. I think no. You must create another *NamingStrategy.

I think it can and should be done, by injecting an optional dependency in the existing strategies and delegating the `classToTableName` conversion responsability to it. Does that make me an architect designer?

zorn-v:
Yep, my fail. In doctrine bundle we can make service like "doctrine.orm.naming_strategy.underscore_plural" with same class (%doctrine.orm.naming_strategy.underscore.class%) and constructor parameters then

zorn-v:
Add plural constructor parameter for naming strategies

lib/Doctrine/ORM/Mapping/Factory/UnderscoreNamingStrategy.php
```diff
@@ -100,18 +121,37 @@ public function joinColumnName($propertyName, $className = null)
...
+        if (strpos($className, '\\') !== false) {
+            $className = substr($className, strrpos($className, '\\') + 1);
+        }
+
+        if ($pluralize) {
+            $className = Inflector::pluralize($className);
+        }
+
+        return $this->underscore($className);
+    }
```
greg0ire: Can we avoid this duplication of code?

zorn-v: You mean something like protected method in default strategy and `return $this->underscore(parent::_classToTableName($className, $pluralize));` in underscore strategy, and inherit underscore from default ?
Is it really needed ?

greg0ire: I mean a new dependency you would inject in the constructors of the existing strategies, and initialize to a default implementation if not provided.

zorn-v: Well `$plural = false` as constructor argument - is not what you mean or what ?

zorn-v: Oh, seems I got your point - remove "UnderscoreNamingStrategy" and add constructor parameter for default one like "$underscore = false" ? But it breaks BC very... much ) I think

zorn-v: But... for BC we can make `parent::__construct` with proper parameters :smile: 
Do I understand you correctly ?

greg0ire: Almost… I mean this:
```
    public function __construct(/*parentParams*/, ?ClassToTableNameTransformer $classToTableNameTransformer = null)
    {
        parent::__construct($parentParams);

        if ($classToTableNameTransformer) {
            $this->classToTableNameTransformer = $classToTableNameTransformer;

            return;
        }
        $this->classToTableNameTransformer = new DefaultClassToTableNameTransformer();
    }
```

zorn-v: One more class just for "class for one more class" ? Java bite you :smile:

beberlei:
This could possibly be implemented as a decorator;

```php
$strategy = new PluralizeTableNameStrategy(new DefaultNamingStrategy);
```
The benefit would be the existing code wouldn't need to change at all (less risk of breakage). Another benefit then it wouldn't need the isPlural and setPlural methods, because that follows from using the class.

Also the tests added are without significane now, because even though plural=true none of them actually have a term that gets pluralized. Please use something meaningful like the MyEntity => my_entities example you gave in the PR description.

PR #7069, opened 2018-02-16
Test for Second level cache with Version, DDC #7067

wtorsi:

PR #7725, opened 2019-05-30
Sync bug fixes from v2.6

guilhermeblanco:


lcobucci (changes requested):
@guilhermeblanco we're missing recent changes and we have some commits that aren't necessary:

46cd05f2f8e5864fbf3f1b1e8d041dcbe781e2fe and 9b04b0c1e9d7a008ca55cbb01c18a6aa801f5c08 come from the backported version https://github.com/doctrine/orm/pull/7069, which is not needed to be applied again since it was originally made against `master`.

The same seems to happen for 73eaa1941477bdf99257bc142432c51d7d3acf5e and 042804cc7bd48544c9c23233991c2dcb72a43765

772426d3da2baa8c8e32dc811e45774bba6c4b34 seems to be missing the [test case](https://github.com/doctrine/orm/pull/7291/files#diff-81252fa90b8e754dc458a8133ac94290).

Commits from https://github.com/doctrine/orm/pull/7471 and https://github.com/doctrine/orm/pull/7528 should be present here too (among other PRs).

Is it possible that you used an old reference of the branch `2.6` to do this sync?

guilhermeblanco:
@lcobucci merged+rebased again. Unit test suite now runs smoothly. =)

lcobucci (dismissed):

lcobucci (approved):
@guilhermeblanco great work, thanks!

I removed many commits that were already present in `master`, which were mostly incorrect. Now PHPCS and PHPBench are also passing :tada:

PR #7727, opened 2019-05-31
[doc] Finish incomplete definition of class UTCDateTimeType

madand:
Method `convertToDatabaseValue()` invokes method `self::getUtc()`, but no actual definition of that method were present. On the other hand, method `convertToPHPValue()` contained the code that should have been inside of the missing `self::getUtc()`. This commit fixes the issue.

### DIFF
--- a/docs/en/cookbook/working-with-datetime.rst
+++ b/docs/en/cookbook/working-with-datetime.rst
@@ -90,7 +90,10 @@ the UTC time at the time of the booking and the timezone the event happened in.
 
     class UTCDateTimeType extends DateTimeType
     {
-        static private $utc;
+        /**
+         * @var \DateTimeZone
+         */
+        private static $utc;
 
         public function convertToDatabaseValue($value, AbstractPlatform $platform)
         {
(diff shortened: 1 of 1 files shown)


docs/en/cookbook/working-with-datetime.rst
```diff
@@ -123,6 +123,11 @@ the UTC time at the time of the booking and the timezone the event happened in.
 
             return $converted;
         }
+        
+        private static function getUtc()
```
greg0ire: How about adding a type hint here?

```suggestion
        private static function getUtc() : \DateTimeZone
```

madand: Force-commited proposed change. Also added dock-block for `self::$utc`.

madand:
I noticed that PR applies to the branch `2.6`. Please let me know whether I should also create PR for `master`.

greg0ire (dismissed):

SenseException (dismissed):

lcobucci (approved):

lcobucci:
@madand thanks :+1:

PR #7730, opened 2019-06-03
Use alternate keyserver

greg0ire:
All the keys at repo.mysql.com expired. Also, it is recommended to check
packages with keys published on a different server in case the first one
is compromised.


greg0ire:
See also #7731: way better alternative?

PR #7742, opened 2019-06-11
Start i var from 1 instead of 0

bocharsky-bw:
Because (0 % $batchSize) === 0 but we don't want to execute flush() and clear() on the first iteration.

Related PR: #7719 where it was done in master only, but it should be done in all supported branches.


greg0ire (approved):

SenseException (approved):

azjezz (approved):

greg0ire:
Restarting the build

greg0ire:
Thanks @bocharsky-bw !

PR #7743, opened 2019-06-12
fix: enable usage of findBy with Criteria[]

bmdevel:
This pull request shall enable something like

```
$criteria = new Criteria();
$criteria->where(Criteria::expr()->lt('id', $param));

return $this->entityManager
    ->getRepository(MyEntity::class)
    ->findBy([$criteria]);
```

Criteria is currently ignored in `\Doctrine\ORM\Persisters\Entity\BasicEntityPersister::getSelectConditionSQL()` and `\Doctrine\ORM\Persisters\Entity\BasicEntityPersister::getSelectConditionStatementSQL()`

### DIFF
--- a/tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterCompositeTypeSqlTest.php
+++ b/tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterCompositeTypeSqlTest.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Tests\ORM\Persisters;
 
+use Doctrine\Common\Collections\Criteria;
 use Doctrine\ORM\Persisters\Entity\BasicEntityPersister;
 use Doctrine\Common\Collections\Expr\Comparison;
 use Doctrine\Tests\Models\GeoNames\Admin1AlternateName;
@@ -54,5 +55,18 @@ class BasicEntityPersisterCompositeTypeSqlTest extends OrmTestCase
     public function testSelectConditionStatementIn()
     {
         $this->_persister->getSelectConditionStatementSQL('admin1', [], [], Comparison::IN);
+    }
+
+    /**
+     * this represents the case that occurs when getSelectConditionSQL() is called with Criteria[] as $criteria
+     * as used by findBy([$criteris]) with Critia[] given as argument
+     */
+    public function testSelectConditionStatementWithCriteria()
+    {
+        $criteria = Criteria::create();
+        $criteria->where(Criteria::expr()->gt('id', 1));
+
+        $statement= $this->_persister->getSelectConditionStatementSQL(0, $criteria);
+        $this->assertEquals('t0.id > ?', $statement);
     }
 }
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterCompositeTypeSqlTest.php
```diff
@@ -55,4 +56,17 @@ public function testSelectConditionStatementIn()
     {
         $this->_persister->getSelectConditionStatementSQL('admin1', [], [], Comparison::IN);
     }
+
+    /**
+     * this represents the case that occurs when getSelectConditionSQL() is called with Criteria[] as $criteria
+     * as used by findBy([$criteris]) with Critia[] given as argument
```
greg0ire:
```suggestion
     * as used by findBy([$criterias]) with Criteria[] given as argument
```

SenseException (changes requested):
``` php
$criteria = new Criteria();
$criteria->where(Criteria::expr()->lt('id', $param));

return $this->entityManager
    ->getRepository(MyEntity::class)
    ->findBy([$criteria]);
```

I would like to see a documentation entry about this. We have a special case here that a `Criteria` isn't fully supported, but only its (only one) WHERE condition.

Your proposal with this PR is that a repository can use a `Criteria` object with the `findBy` method. If this is the wanted feature, we should have a test, that guarantees it works with that repository method to prevent possible regressions.

I was also looking into the usage of `getSelectConditionStatementSQL()` in the project, because the values should be fetched from the `Criteria` and injected into `getSelectConditionStatementSQL()`, instead of overriding the original arguments inside the method. I have no overview about the complexity of such a approach yet.

So far:
* Please add (a) test(s) for the `findBy` usage (and maybe other methods that will be able to use `Criteria`
* Please extend the documentation to add the `Criteria` feature with its limitation

tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterCompositeTypeSqlTest.php
lcobucci: Isn't "criteria" already in plural form?

lcobucci:
Isn't this already provided by https://github.com/doctrine/orm/blob/434820973cadf2da2d66e7184be370084cc32ca8/lib/Doctrine/ORM/EntityRepository.php#L284-L289

tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterCompositeTypeSqlTest.php
greg0ire: You're right, fixed! I hope people will not get confused though :sweat_smile: Also, I removed the brackets, is that correct?

lcobucci: Seems correct, although I think that this change doesn't make sense because we already have `EntityRepository#matching()` that provides such functionality

greg0ire: Oh that's right!

lcobucci:
Closing as won't fix since `EntityRepository#matching()` should be used instead.

Issue #7735, opened 2019-06-05
Null in fields value in Cached Entity several times on day on high-load project.


AlexSmerw:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.6.3

#### Summary
We have a problem with the cache on a high-load project. Several times a day, the entity is retrieved from the cache with the field value = null, and these fields cannot be null
```
/ **
    * @var curator
    *
    * @ManyToOne (targetEntity = "SD \ Domain \ PersistModel \ Curator \ Curator")
    * @JoinColumn (name = "curator_id", referencedColumnName = "id")
    * /
   private $ curator ;
```

#### Current behavior

This situation arises from related entities. When the parent entity gets out of the cache, and the related is already expired.
See [DefaultEntityHydrator function loadCacheEntry](https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Cache/DefaultEntityHydrator.php#L182) 
Then 
[BasicEntityPersister function load](https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L745) will return Entity with null fields value.
Because [AbstractEntityPersister](https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php#L469)
$entity = $this->persister->loadById($identifier, $entity);
It put $entity = null to function loadById($identifier, $entity) Instead Entity.

#### How to reproduce
Set $assocEntry  = null on [DefaultEntityHydrator line180](https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Cache/DefaultEntityHydrator.php#L180) for child entity

#### Expected behavior
I think this can be corrected by editing the code in function loadById(array $identifier, $entity = null)
[loadById](https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php#L447)

Replace $ entity with $ cachedEntity
https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php#L458

```
$cachedEntity = $this->persister->loadById($identifier, $entity);
        if ($cachedEntity === null) {
            return null;
         }
```

Ocramius:
We'd need a test case for this: as the issue is presented right now, this isn't really usable by maintainers.

[...]

PR #7749, opened 2019-06-20
null values in cached entity fix

AlexSmerw:
This is fix for issue [7735](https://github.com/doctrine/orm/issues/7735)


Ocramius (changes requested):
This kind of change requires a test case that fails before the patch, and passes after the patch.

Ocramius:
@AlexSmerw could you also target `2.6` with this fix, please?

AlexSmerw:
> @AlexSmerw could you also target `2.6` with this fix, please?

Done https://github.com/doctrine/orm/pull/7750

lcobucci:
Closing here to keep the fix only for 2.6, we'll port it to master once things are integrated.

PR #7750, opened 2019-06-20
Fix incorrect return of null values in L2C

AlexSmerw:
Bug fix for issue [7735](https://github.com/doctrine/orm/issues/7735) for doctrine 2.6

### DIFF
--- a/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
+++ b/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
@@ -452,12 +452,14 @@ abstract class AbstractEntityPersister implements CachedEntityPersister
                 $class = $this->metadataFactory->getMetadataFor($cacheEntry->class);
             }
 
-            if (($entity = $this->hydrator->loadCacheEntry($class, $cacheKey, $cacheEntry, $entity)) !== null) {
+            $cachedEntity = $this->hydrator->loadCacheEntry($class, $cacheKey, $cacheEntry, $entity);
+
+            if ($cachedEntity !== null) {
                 if ($this->cacheLogger) {
                     $this->cacheLogger->entityCacheHit($this->regionName, $cacheKey);
                 }
 
-                return $entity;
+                return $cachedEntity;
             }
         }
 
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7735Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7735Test.php
@@ -1,0 +1,172 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+use function assert;
+
+final class GH7735Test extends OrmFunctionalTestCase
+{
+    public function setUp() : void
+    {
+        $this->enableSecondLevelCache();
+        parent::setUp();
+
+        $this->_schemaTool->createSchema(
+            [
+                $this->_em->getClassMetadata(GH7735Car::class),
+                $this->_em->getClassMetadata(GH7735Power::class),
+                $this->_em->getClassMetadata(GH7735Engine::class),
+            ]
+        );
+
+        $this->_em->persist(new GH7735Car(1, new GH7735Engine(1, 'turbo', new GH7735Power(1))));
+        $this->_em->flush();
+        $this->_em->clear();
+    }
+
+    /**
+     * @test
+     * @group GH7735
+     */
+    public function findByReturnsCachedEntity() : void
+    {
+        $this->_em->getCache()->evictEntityRegion(GH7735Power::class);
+
+        $car = $this->_em->find(GH7735Car::class, 1);
+        assert($car instanceof GH7735Car);
+
+        self::assertSame('turbo', $car->getEngine()->getModel());
+        self::assertSame(1, $car->getEngine()->getPower()->getId());
+    }
+}
+
+/**
+ * @Entity @Cache(usage="READ_ONLY")
+ */
+class GH7735Car
+{
+    /**
+     * @Id
+     * @Column(type="integer")
+     * @var int
+     */
+    private $id;
+
+    /**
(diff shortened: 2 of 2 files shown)


lcobucci (commented):
@AlexSmerw thanks for your contribution! 

This does look sane from a very quick look but could you please add a functional that reproduces the issue and proves that the fix solves it?

AlexSmerw:
> @AlexSmerw thanks for your contribution!
> 
> This does look sane from a very quick look but could you please add a functional that reproduces the issue and proves that the fix solves it?

This is not enough -  [description with link on code](https://github.com/doctrine/orm/issues/7735#issuecomment-503555655) ?

lcobucci:
Not really, we need a phpunit test that gets integrated to the codebase, which also ensures that this problem will never happen again. 

You can find examples on https://github.com/doctrine/doctrine2/tree/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket

AlexSmerw:
> Not really, we need a phpunit test that gets integrated to the codebase, which also ensures that this problem will never happen again.
> 
> You can find examples on https://github.com/doctrine/doctrine2/tree/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket

I have added unitTest in my branch. If you use this test **Issue7735Test** without my edits, you will get Exception `TypeError: Return value of Doctrine\Tests\Models\Issue7735\Engine::getModel() must be of the type string, null returned`. Look, I set 1000000 iterations for getting this bug. I think this count depends on you PC

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -724,7 +732,10 @@ protected function _getEntityManager(
 
             $cacheConfig    = new CacheConfiguration();
             $cache          = $this->getSharedSecondLevelCacheDriverImpl();
-            $factory        = new DefaultCacheFactory($cacheConfig->getRegionsConfiguration(), $cache);
+
+            $regionConfig = $cacheConfig->getRegionsConfiguration();
+            $regionConfig->setDefaultLifetime(static::DEFAULT_CACHE_REGION_LIFE_TIME);
```
lcobucci: Is this change really required for the test?

lcobucci (changes requested):
Thank you very much for adding the test, I've got a tiny adjustment request: can you please put the models in the same file as the test and remove all attributes which might be irrelevant to reproducible the issue?

tests/Doctrine/Tests/OrmFunctionalTestCase.php
AlexSmerw: For reproducing this problem,  I set second level cache time limit  = 1 second. This code need for this.

lcobucci: Sorry, what I meant to say was: can't we keep this contained to the functional test only? This is now changing the default for every test in the suite, which might create undesired side-effects 😁

AlexSmerw:
> Thank you very much for adding the test, I've got a tiny adjustment request: can you please put the models in the same file as the test and remove all attributes which might be irrelevant to reproducible the issue? 
Hi! I have done this.

lcobucci:
@AlexSmerw it seems like we still have separate files for each class... Check this example: https://github.com/doctrine/orm/blob/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6362Test.php

tests/Doctrine/Tests/OrmFunctionalTestCase.php
AlexSmerw: > Sorry, what I meant to say was: can't we keep this contained to the functional test only? This is now changing the default for every test in the suite, which might create undesired side-effects grin

Done

AlexSmerw:
> @AlexSmerw it seems like we still have separate files for each class... Check this example: https://github.com/doctrine/orm/blob/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6362Test.php

Oh! It's Evil!

AlexSmerw:
> @AlexSmerw it seems like we still have separate files for each class... Check this example: https://github.com/doctrine/orm/blob/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6362Test.php

Done. Check it, please.

lcobucci:
@AlexSmerw I'll take a look at it ASAP, thanks for applying the changes!

AlexSmerw:
> @AlexSmerw I'll take a look at it ASAP, thanks for applying the changes!
Hi!
My request still with label Missing Tests. Is it ok?
I worry about its fate :)

greg0ire:
Please associate your commit email address with your Github account, or change the email in your commits to an address already associated with it.

lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
```diff
@@ -452,12 +452,12 @@ public function loadById(array $identifier, $entity = null)
...
             }
 
-            if (($entity = $this->hydrator->loadCacheEntry($class, $cacheKey, $cacheEntry, $entity)) !== null) {
+            if (($cachedEntity = $this->hydrator->loadCacheEntry($class, $cacheKey, $cacheEntry, $entity)) !== null) {
                 if ($this->cacheLogger) {
                     $this->cacheLogger->entityCacheHit($this->regionName, $cacheKey);
                 }
 
-                return $entity;
+                return $cachedEntity;
```
greg0ire: A good commit message might be:

```
Avoid reusing variable name

The same variable name is used below, and that causes a bug etc.
Fixes #issueNumber
```

lcobucci: Commit message updated :+1:

lcobucci (dismissed):
@AlexSmerw I've managed to clean up your test and remove the need for the cache TTL configuration and the `for` loop.

Thank you very much for fixing this bug!

lcobucci (approved):

PR #7761, opened 2019-06-26
Do not modify UOW on PersistentCollection::clear() when owner has DEFFERED_EXPLICIT change tracking policy

### LINKED ISSUE #7758: PersistentCollection::clear() modifies uow for DEFERRED_EXPLICIT change tracking policy
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

#### Summary

Calling `clear()` on a Many-to-many association collection should not modify uow as long as `persist()` is not called on owner.
This is not the same behavior as calling `removeElement()` on each element, which sounds weird.

#### Current behavior

Calling `clear()` generates a `DELETE FROM table WHERE owner_id = ?` statement on a `flush()`.

#### How to reproduce

Mapping : 
```xml
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping">
    <entity name="Foo"
            change-tracking-policy="DEFERRED_EXPLICIT">
        <id name="id" type="integer" column="id">
            <generator strategy="IDENTITY"/>
        </id>
        <many-to-many field="bars" target-entity="Bar">
            <cascade>
                <cascade-all/>
            </cascade>
            <join-table name="foo_bar">
                <join-columns>
                    <join-column name="foo_id" nullable="false"/>
                </join-columns>
                <inverse-join-columns>
                    <join-column name="bar_id" nullable="false"/>
                </inverse-join-columns>
            </join-table>
        </many-to-many>
    </entity>
    <entity name="Bar"
            change-tracking-policy="DEFERRED_EXPLICIT">
        <id name="id" type="integer" column="id">
            <generator strategy="IDENTITY"/>
        </id>
    </entity>
</doctrine-mapping>
```

with PHP code : 
```php

class Bar { public $id; }
class Foo {
    public $id;
    public $bars;
    public function __construct()
    {
        $this->bars = new ArrayCollection();
    }
}

// Create new entity
$foo = new Foo();
// Attach Bar
$foo->bars->add(new Bar());
// Persist the whole thing
$em->persist($foo);
$em->flush();
// Retrieve assigned id
$fooId = $foo->id;
// Clear em
$em->clear();

// Fetch from DB
$foo = $em->find(Foo::class, $fooId);
// Clear bars item per item
foreach ($foo->bars as $bar) {
    $foo->bars->removeElement($bar);
}
// Does nothing as $foo is not persisted
$em->flush();

$foo->bars->clear();
// Removes all elements even though $foo is not persisted
$em->flush();
```

#### Expected behavior

No statement should be executed.

paxal:
This should fix #7758 for 2.6 branch ‒ might be ok to be forwarded on 2.7 but not tested.

I also worked on a fix for 3.0 branch here https://github.com/paxal/doctrine2/commit/82bb914fbcd7b66cea3b48a8c4d9253204d5ec16

### DIFF
--- a/lib/Doctrine/ORM/PersistentCollection.php
+++ b/lib/Doctrine/ORM/PersistentCollection.php
@@ -25,6 +25,7 @@ use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Selectable;
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\ORM\Mapping\ClassMetadata;
+use function get_class;
 
 /**
  * A PersistentCollection represents a collection of elements that have persistent state.
@@ -565,7 +566,9 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
         if ($this->association['isOwningSide'] && $this->owner) {
             $this->changed();
 
-            $uow->scheduleCollectionDeletion($this);
+            if (! $this->em->getClassMetadata(get_class($this->owner))->isChangeTrackingDeferredExplicit()) {
+                $uow->scheduleCollectionDeletion($this);
+            }
 
             $this->takeSnapshot();
         }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7761Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7761Test.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH7761Test extends OrmFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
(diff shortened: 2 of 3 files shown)


SenseException:
@doctrine/doctrinecore Should we add a functional test for this fix or are the test sufficient?

paxal:
@SenseException I added a functional test ‒ that fails without the patch :)

paxal:
Should I close this PR ? Or open one for 2.7 or 3.0 instead of 2.6 ? Or did I miss something else ?
It would be nice to have some feedback :sweat_smile:

guilhermeblanco:
@paxal please do not close the PR... I quickly reviewed it and it seems valid (the issue and the patch). We just need to coordinate merging strategy across all branches. =)

lcobucci (approved):

paxal:
Thanks ! How to apply this to 2.7 as well ?
For 3.0, I will open a new PR.

lcobucci:
@paxal you don't need to worry about the other branches, we'll handle it from here. Thanks :+1:

paxal:
Ok, thanks for clarification :+1:

PR #7766, opened 2019-07-01
Respect collection orderBy meta when matching()

stephanschuler:


tests/Doctrine/Tests/ORM/Functional/OneToManyMatchingCriteriaTest.php
```diff
@@ -0,0 +1,96 @@
...
+
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\ORM\Mapping\FetchMode;
+use Doctrine\Tests\Models\ECommerce\ECommerceCategory;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * Tests sorting mechanism of Selectable::matching()
+ */
+class OneToManyMatchingCriteriaTest extends OrmFunctionalTestCase
```
lcobucci: Please use the github issue id as class name, to help us identify things better (GH*)

tests/Doctrine/Tests/ORM/Functional/OneToManyMatchingCriteriaTest.php
```diff
@@ -0,0 +1,96 @@
...
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\ORM\Mapping\FetchMode;
+use Doctrine\Tests\Models\ECommerce\ECommerceCategory;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * Tests sorting mechanism of Selectable::matching()
```
lcobucci: Please add `@group` with the id of the github issue here

tests/Doctrine/Tests/ORM/Functional/OneToManyMatchingCriteriaTest.php
```diff
@@ -0,0 +1,96 @@
...
+class OneToManyMatchingCriteriaTest extends OrmFunctionalTestCase
+{
+    private $parent;
+    private $firstChild;
+    private $secondChild;
+    private $thirdChild;
+
+    protected function setUp() : void
+    {
+        $this->useModelSet('ecommerce');
```
lcobucci: On newer tests we decided to no longer use the model sets, instead we're putting the required entities together in the same file of the test (with only the necessary properties and methods to reproduce the issue).

This helps us to more easily analyse the system under test and find out if the solution is the best. Could you please change this test in that way?

tests/Doctrine/Tests/ORM/Functional/OneToManyMatchingCriteriaTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $this->secondChild->setName('Php books');
+        $this->thirdChild = new ECommerceCategory();
+        $this->thirdChild->setName('BBQ books');
+    }
+
+    public function testMatchingWithNoOrderingUsesDatabaseDefault() : void
+    {
+        $this->createFixture();
+
+        $metadata = $this->em->getClassMetadata(ECommerceCategory::class);
```
lcobucci: By providing a set of entities for the test we might avoid changing the metadata in runtime, which makes things slightly confusing 😁

tests/Doctrine/Tests/ORM/Functional/OneToManyMatchingCriteriaTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $metadata->getProperty('children')->setFetchMode(FetchMode::LAZY);
+
+        $query    = $this->em->createQuery('select c from Doctrine\Tests\Models\ECommerce\ECommerceCategory c order by c.id asc');
+        $result   = $query->getResult();
+        $parent   = $result[0];
+        $children = $parent->getChildren()->matching(Criteria::create());
+        $this->assertEquals('Java books', $children[0]->getName());
```
lcobucci: Please use `self::assertSame()` instead. PHPUnit assertions are static methods and, for these checks, we can use a more strict comparison.

lcobucci (changes requested):
@stephanschuler thank you very much for your contribution. I've found some small things for us to address/discuss.

Apart from that, could you please send modify your patch to target the branch `2.6`? This is a bug and we'll port to master once it's released.

tests/Doctrine/Tests/ORM/Functional/OneToManyMatchingCriteriaTest.php
stephanschuler: Done, it's GH7767Test now. Moved it to "Functional/Ticket" because that's where other tests seem to live.

tests/Doctrine/Tests/ORM/Functional/OneToManyMatchingCriteriaTest.php
stephanschuler: Done, added `@group 7767`

tests/Doctrine/Tests/ORM/Functional/OneToManyMatchingCriteriaTest.php
stephanschuler: Done, added a GH7767ParentEntity and a GH7767ChildEntity.

tests/Doctrine/Tests/ORM/Functional/OneToManyMatchingCriteriaTest.php
stephanschuler: Fixed.

lcobucci (approved):
@stephanschuler thanks a lot for your patch!

I've made a few modifications to simplify the test and not break other tests.

:ship:

PR #7770, opened 2019-07-07
#7769 Define getProxyDir in Configuration class

### LINKED ISSUE #7769: GenerateProxiesCommand getConfiguration()->getProxyDir() is undefined
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0.x-dev

#### Summary

I just started to use the doctrine console. With a simple "php vendor/bin/doctrine orm:generate-proxies" it fails with the following message: "Call to undefined method Doctrine\ORM\Configuration::getProxyDir() in .../vendor/doctrine/orm/lib/Doctrine/ORM/Tools/Console/Command/GenerateProxiesCommand.php on line 57"

#### Current behavior

Command failes when to proxies directory is given as an argument.

Use the doctrine console with following call:

"php vendor/bin/doctrine orm:generate-proxies"

#### Expected behavior

That the command generates the proxy classes even without giving an explicit directory.

alirezameskin:
This is my attempt to resolve https://github.com/doctrine/orm/issues/7769

### DIFF
--- a/tests/Doctrine/Tests/ORM/ConfigurationTest.php
+++ b/tests/Doctrine/Tests/ORM/ConfigurationTest.php
@@ -328,6 +328,14 @@ class ConfigurationTest extends DoctrineTestCase
         self::assertSame($namespace, $this->configuration->getProxyManagerConfiguration()->getProxiesNamespace());
     }
 
+    public function testGetProxyDir(): void
+    {
+        $proxyDir = $this->makeTemporaryValidDirectory();
+
+        $this->configuration->setProxyDir($proxyDir);
+        self::assertSame($proxyDir, $this->configuration->getProxyDir());
+    }
+
     /**
      * @param int|bool $proxyAutoGenerateFlag
      *
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -109,6 +109,16 @@ class Configuration extends DBALConfiguration
     }
 
     /**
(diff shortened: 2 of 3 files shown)


tests/Doctrine/Tests/ORM/ConfigurationTest.php
```diff
@@ -328,6 +328,14 @@ public function testProxyManagerConfigurationContainsGivenProxyNamespace() : voi
         self::assertSame($namespace, $this->configuration->getProxyManagerConfiguration()->getProxiesNamespace());
     }
 
+    public function testGetProxyDir(): void
+    {
+        $proxyDir = $this->makeTemporaryValidDirectory();
```
greg0ire: If the directory is temporary, it might not exist on subsequent executions, by definition. I don't know if that command is supposed to be used on production systems, but if it is, I'd rather have the directory become a mandatory argument than having it be temporary

lcobucci:
Asking a review from @guilhermeblanco since this has probably been removed for a reason.

Ocramius:
Please just use https://github.com/doctrine/orm/blob/d528f7056886fa2ea114b3179946cc6a5e8d08ee/lib/Doctrine/ORM/Configuration.php#L591-L595 - do not call `Configuration#getProxyDir()`, since it may not even be used anymore (Depending on how ProxyManager was set up)

Ocramius:
Note: patch still needs resolving, but it needs to consider whether codegen is to be performed in first place at all.

PR #7771, opened 2019-07-08
Remove year from license

Arman-Hosseini:


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2006-2018 Doctrine Project
+Copyright (c) 2006-2019 Doctrine Project
```
Ocramius: Probably better to just drop the year overall

Ocramius: Probably better to just drop the year overall

Ocramius (dismissed):

LICENSE
Arman-Hosseini: Some people think that 2006 was the beginning of the project.
It's good to always be mentioned this year as the end.

Arman-Hosseini: Do you mean to remove the end of this year in general?

Ocramius: I'd remove the whole range - it is something that can be always extracted via our signed releases and commits.

greg0ire: Unresolving because the range is still here… did you forget to push your changes, maybe?

Arman-Hosseini: I did not get you.
@greg0ire

greg0ire: That conversation was marked as resolved, but the issue is still here. You may mark it as resolved once this is fixed, like so:
```suggestion
Copyright (c) Doctrine Project
```

Arman-Hosseini: Oh, yes, my changes have not been saved.
I added the changes again.
@greg0ire  Thank you.

lcobucci (approved):
@Arman-Hosseini awesome, thanks!

PR #7772, opened 2019-07-09
Fix build

lcobucci:
Doing the minimum things to make PHPCS and PHPStan happy - the fixed string there is nasty, though, but no test is checking it yet.


lib/Doctrine/ORM/Mapping/Builder/FieldMetadataBuilder.php
```diff
@@ -37,10 +31,7 @@ public function build() : Mapping\FieldMetadata
     {
         // Validate required fields
         assert($this->componentMetadata !== null);
-        assert($this->cacheAnnotation !== null);
 
-        $fieldMetadata = new Mapping\FieldMetadata();
-
-        return $fieldMetadata;
+        return new Mapping\FieldMetadata('We need a name here Mr. Blanco');
```
guilhermeblanco: This one is not yet ready.... =)

guilhermeblanco (approved):

lib/Doctrine/ORM/Mapping/Builder/FieldMetadataBuilder.php
lcobucci: @guilhermeblanco I imagined :laughing:

Issue #7266, opened 2018-06-19
Call to a member function resolveAssociationEntries() on boolean {"detail":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Call to a member function resolveAssociationEntries() on boolean at /www/vendor/doctrine/orm/lib/Doctrine/ORM/Cache/DefaultQueryCache.php:140)"}


mingmingxianseng:
**The following mistakes occur occasionally，and I can't find the solution**

```
Call to a member function resolveAssociationEntries() on boolean {"detail":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Call to a member function resolveAssociationEntries() on boolean at /www/vendor/doctrine/orm/lib/Doctrine/ORM/Cache/DefaultQueryCache.php:140)"}
```

>  version: doctrine/orm V2.6.0 
>  in  DefaultQueryCache.php at line 140 

```php
        $entries   = $region->getMultiple($cacheKeys);
        // @TODO - move to cache hydration component
        foreach ($cacheEntry->result as $index => $entry) {
            $entityEntry = is_array($entries) && array_key_exists($index, $entries) ? $entries[$index] : null;

            if ($entityEntry === null) {
                if ($this->cacheLogger !== null) {
                    $this->cacheLogger->entityCacheMiss($regionName, $cacheKeys->identifiers[$index]);
                }

                return null;
            }

```

I think the parameter $entityEntry sometimes will be  false , then occur the mistakes. 

change `if ($entityEntry === null) `  to `if ($entityEntry === null || $entityEntry === false ) ` . 
Is this a good solution ? 

I need your help. Thank you very much.

javabudd:
+1

This issue started happening to me when I switched my redis `maxmemory-policy` to `allkeys-lru`

Ocramius:
Also on newer releases?

javabudd:
I'm currently on 2.6.2

Ocramius:
Fixing the conditional is incorrect there: would need to change the fetch operation so that multiple misses don't become `false[]`

Ocramius:
Can this be reproduced with a failing test case?

javabudd:
I'll try and reproduce

mingmingxianseng:
```

   ## in  Doctrine\Common\Cache\RedisCache.php  +72

    protected function doFetchMultiple(array $keys)
    {
[...]

PR #7778, opened 2019-07-23
Guard L2C regions against corrupted data

umpirsky:
Fixes https://github.com/doctrine/orm/issues/7266.

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -124,7 +124,7 @@ class DefaultQueryCache implements QueryCache
         foreach ($cacheEntry->result as $index => $entry) {
             $entityEntry = is_array($entries) && array_key_exists($index, $entries) ? $entries[$index] : null;
 
-            if ($entityEntry === null) {
+            if (! $entityEntry instanceof EntityCacheEntry) {
                 if ($this->cacheLogger !== null) {
                     $this->cacheLogger->entityCacheMiss($regionName, $cacheKeys->identifiers[$index]);
                 }
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -118,11 +118,11 @@ class DefaultQueryCache implements QueryCache
         };
 
         $cacheKeys = new CollectionCacheEntry(array_map($generateKeys, $cacheEntry->result));
-        $entries   = $region->getMultiple($cacheKeys);
+        $entries   = $region->getMultiple($cacheKeys) ?? [];
 
         // @TODO - move to cache hydration component
         foreach ($cacheEntry->result as $index => $entry) {
-            $entityEntry = is_array($entries) && array_key_exists($index, $entries) ? $entries[$index] : null;
+            $entityEntry = $entries[$index] ?? null;
 
             if (! $entityEntry instanceof EntityCacheEntry) {
                 if ($this->cacheLogger !== null) {
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -106,7 +106,7 @@ class DefaultQueryCache implements QueryCache
 
         $result      = [];
         $entityName  = reset($rsm->aliasMap);
-        $hasRelation = ( ! empty($rsm->relationMap));
+        $hasRelation = ! empty($rsm->relationMap);
         $persister   = $this->uow->getEntityPersister($entityName);
         $region      = $persister->getCacheRegion();
         $regionName  = $region->getName();
@@ -267,7 +267,7 @@ class DefaultQueryCache implements QueryCache
         $rootAlias   = key($rsm->aliasMap);
         $persister   = $this->uow->getEntityPersister($entityName);
 
-        if ( ! ($persister instanceof CachedPersister)) {
+        if (! $persister instanceof CachedEntityPersister) {
             throw CacheException::nonCacheableEntity($entityName);
         }
 
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -22,6 +22,7 @@ namespace Doctrine\ORM\Cache;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\ORM\Cache\Persister\CachedPersister;
+use Doctrine\ORM\Cache\Persister\Entity\CachedEntityPersister;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\ORM\Mapping\ClassMetadata;
@@ -29,6 +30,7 @@ use Doctrine\ORM\PersistentCollection;
 use Doctrine\Common\Proxy\Proxy;
 use Doctrine\ORM\Cache;
 use Doctrine\ORM\Query;
+use function assert;
 
 /**
  * Default query cache implementation.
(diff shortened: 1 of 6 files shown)


greg0ire:

umpirsky:
@greg0ire Improved, thanks.

greg0ire:
Your now exemplary commit message says:

> On cache miss the parameter `$entityEntry` sometimes will be false. 

But apparently, `false` is not supposed to happen at all:

https://github.com/doctrine/orm/blob/a896421be2eb9f9b6afd2c358f725462c234cefb/lib/Doctrine/ORM/Cache/MultiGetRegion.php#L20

IMO the right fix would be to make sure you cannot get `false` instead of handling it afterwards. Make the implementation conform to what the contract specified in the phpdoc says.

umpirsky:
I agree, this was a quick way for me to turn error into cache miss. But the root of the problem is in `getMultiple()` in deed, and I don't see the way to fix that.

greg0ire:
> `getMultiple()`

Which implementation are we talking about here? Whatever it is, shouldn't we ensure it never returns `false` as an element of the returned array?

umpirsky:
> Which implementation are we talking about here? Whatever it is, shouldn't we ensure it never returns `false` as an element of the returned array?

It's [RedisCache](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/RedisCache.php). It looks like [this condition](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/RedisCache.php#L60) can allow `false` values if `$this->redis->exists($key)` is `true`.

greg0ire:
`RedisCache` does not implement `MultiGetRegion`, does it? We're looking for an implementation of `MultiGetRegion` that returns a collection of `CacheEntry` objects, and lets a `false` slip in.

greg0ire:
I think the issue might be here: https://github.com/doctrine/orm/blob/2726636e55fdbdf6fb1ed466f22d34de116fb8f9/lib/Doctrine/ORM/Cache/Region/DefaultMultiGetRegion.php#L53-L55

See the difference with the other implementation: https://github.com/doctrine/orm/blob/2726636e55fdbdf6fb1ed466f22d34de116fb8f9/lib/Doctrine/ORM/Cache/Region/DefaultRegion.php#L85-L90

umpirsky:
Yes, that looks wrong, it's worth trying.

umpirsky:
I can't simulate it in the test case:

```diff
--- a/tests/Doctrine/Tests/ORM/Cache/MultiGetRegionTest.php
+++ b/tests/Doctrine/Tests/ORM/Cache/MultiGetRegionTest.php
@@ -38,4 +38,26 @@ class MultiGetRegionTest extends AbstractRegionTest
         self::assertEquals($value1, $actual[0]);
         self::assertEquals($value2, $actual[1]);
     }
+
+    public function testGetMultiCacheMiss() : void
+    {
+        $key1   = new CacheKeyMock('key.1');
+        $value1 = new CacheEntryMock(['id' => 1, 'name' => 'bar']);
+
+        $key2   = new CacheKeyMock('key.2');
+        $value2 = new CacheEntryMock(['id' => 2, 'name' => 'bar']);
+
+        self::assertFalse($this->region->contains($key1));
+        self::assertFalse($this->region->contains($key2));
+
+        $this->region->put($key1, $value1);
+        $this->region->put($key2, $value2);
+
+        self::assertTrue($this->region->contains($key1));
+        self::assertTrue($this->region->contains($key2));
+
+        $actual = $this->region->getMultiple(new CollectionCacheEntry([$key1, $key2]));
+
+        self::assertNull($actual);
+    }
```

I don't know how to create `CacheEntryMock` in a way to simulate a cache miss. It also looks like this is prevented in https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Cache/Region/DefaultRegion.php#L89-L91 so theoretically `DefaultMultiGetRegion` can't receive `false` value.

Am I missing something?

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -99,7 +99,7 @@ public function get(QueryCacheKey $key, ResultSetMapping $rsm, array $hints = []
         foreach ($cacheEntry->result as $index => $entry) {
             $entityEntry = is_array($entries) ? ($entries[$index] ?? null) : null;
 
-            if ($entityEntry === null) {
+            if (!$entityEntry) {
```
lcobucci: This should rather be `! $entityEntry instanceof CacheEntry` (or something more specific to this usage)

lcobucci:
@umpirsky I think we're looking this from the wrong angle... I mean, if the key exists in Redis it means that it was stored as `false` instead of the expected object and we should look into the write side instead.

lcobucci:
Also, please send the bugfix to `2.6` branch instead of `master`. We'll handle the porting :)

umpirsky:
That's the problem, when I try to create `new CacheEntryMock(false)` I get `InvalidArgumentException: Passed variable is not an array or object`. That's why I can't create failing test.

lcobucci:
@umpirsky I was looking at the write side and had some thoughts that might be related to max value size for strings or serialisation issues... which redis and redis client versions are you using and do you have igbinary installed?

lcobucci:
@umpirsky I just rebased your branch so that we can make this part of the next bugfix release. I'll continue investigating things here.

lcobucci (dismissed):
I'd say things are alright now, although this change doesn't fix the **generation** of such corrupted data (which I couldn't really find :disappointed:).

It would be nice to get another pair of eyes here, just to avoid potential issues.

@umpirsky thanks for your effort and time :+1:

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -106,25 +108,27 @@ public function get(QueryCacheKey $key, ResultSetMapping $rsm, array $hints = []
 
         $result      = [];
         $entityName  = reset($rsm->aliasMap);
-        $hasRelation = ( ! empty($rsm->relationMap));
+        $hasRelation = ! empty($rsm->relationMap);
         $persister   = $this->uow->getEntityPersister($entityName);
-        $region      = $persister->getCacheRegion();
-        $regionName  = $region->getName();
+        assert($persister instanceof CachedEntityPersister);
```
guilhermeblanco: not super happy with these asserts...

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -145,9 +149,11 @@ public function get(QueryCacheKey $key, ResultSetMapping $rsm, array $hints = []
             $data = $entityEntry->data;
 
             foreach ($entry['associations'] as $name => $assoc) {
-                $assocPersister  = $this->uow->getEntityPersister($assoc['targetEntity']);
-                $assocRegion     = $assocPersister->getCacheRegion();
-                $assocMetadata   = $this->em->getClassMetadata($assoc['targetEntity']);
+                $assocPersister = $this->uow->getEntityPersister($assoc['targetEntity']);
+                assert($assocPersister instanceof CachedEntityPersister);
```
guilhermeblanco: not super happy with these asserts...

guilhermeblanco (dismissed):
Patch seems fine for me aside from the asserts. They should be either removed or try/catch'ed.

lcobucci:
@guilhermeblanco these are meant for development only and replace the not so useful doc blocks for variables.

There's an extensive discussion on that in https://github.com/doctrine/coding-standard/pull/47

lcobucci (approved):

umpirsky:
Will this be released in `2.6.4` or `2.7`?

lcobucci:
`2.6.4` since it's a bug fix. I plan to release this week.

umpirsky:
Next weekend I guess? :)

lcobucci:
@umpirsky sorry, life happened and didn't allow me to handle things. Will do it ASAP (but won't promise any date :disappointed:)

PR #7782, opened 2019-07-28
Update to PHPStan v0.11

carusogabriel:
Although we're not in a higher level of PHPStan, let at least keep with the last version of it :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
     "require-dev": {
         "ext-pdo": "*",
         "doctrine/coding-standard": "^6.0",
-        "phpstan/phpstan-shim": "^0.9.2",
+        "phpstan/phpstan": "^0.11",
         "phpunit/phpunit": "^7.0"
     },
     "autoload": {
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -93,7 +93,6 @@ class BasicEntityPersister implements EntityPersister
 {
     /** @var string[] */
     private static $comparisonMap = [
-        Comparison::EQ          => '= %s',
         Comparison::IS          => '= %s',
         Comparison::NEQ         => '!= %s',
         Comparison::GT          => '> %s',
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -53,7 +53,7 @@
     "require-dev": {
         "ext-pdo": "*",
         "doctrine/coding-standard": "^6.0",
-        "phpstan/phpstan-shim": "^0.9.2",
+        "phpstan/phpstan": "^0.11",
```
greg0ire: I don't understand how it was working before… shouldn't it have been `vendor/bin/phpstan.phar` in `.travis.yml`? Also, why this change?

carusogabriel: `phpstan-shim` is the binary version and I think we're using it to avoid conflicts. I've changed it to align the way we do in other repositores.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -93,7 +93,6 @@ class BasicEntityPersister implements EntityPersister
 {
     /** @var string[] */
     private static $comparisonMap = [
-        Comparison::EQ          => '= %s',
```
Ocramius: Why is this being removed? Eq/Is being the same?

Ocramius (changes requested):

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
carusogabriel: Yeap, here's the PHPStan error:

```
 ------ ----------------------------------------------------------------------------------------------
  Line   Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
 ------ ----------------------------------------------------------------------------------------------
  96     Array has 2 duplicate keys with value '=' (\Doctrine\Common\Collections\Expr\Comparison::EQ,
         \Doctrine\Common\Collections\Expr\Comparison::IS).
 ------ ----------------------------------------------------------------------------------------------
```

Ocramius (approved):

PR #7796, opened 2019-08-12
Do not modify UOW on PersistentCollection::clear() when owner has DEFFERED_EXPLICIT change tracking policy 

paxal:
Same as #7761


lcobucci:
@paxal thanks for sending this too but we'll port it to `master` while releasing v2.6.4. Will close this one to avoid confusion :)

PR #7800, opened 2019-08-16
Query expr with array alias

maxhelias:
I would like to propose for the next version, a simplification of `select` & `groupBy` with an alias.

This can be interesting for a consequent query

Here a little example : 
```
$qb = $em->createQueryBuilder()
    ->select(['u' => ['id', 'name'], 'p' => ['number'])
    ->from('User', 'u')
    ->leftJoin('u.Phonenumbers', 'p')
    ->groupBy(['u' => ['lastLogin']])
    ->addGroupBy('u.createdAt');
```

Note : The example for the `groupBy` is bad but it's the idea

### DIFF
--- a/lib/Doctrine/ORM/Query/Expr/BaseAlias.php
+++ b/lib/Doctrine/ORM/Query/Expr/BaseAlias.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Query\Expr;
+
+use InvalidArgumentException;
+use function is_array;
+use function is_string;
+use function sprintf;
+
+/**
+ * Abstract base Expr class for building DQL parts.
+ */
+abstract class BaseAlias extends Base
+{
+    /**
+     * @param mixed $args
+     *
+     * @return Base
+     */
+    public function addMultiple($args = [])
+    {
+        foreach ((array) $args as $alias => $arg) {
+            if (is_string($alias)) {
+                $this->add($arg, $alias);
+
+                continue;
+            }
+            parent::add($arg);
+        }
+
+        return $this;
+    }
+
+    /**
+     * @param mixed       $arg
+     * @param string|null $alias
+     *
+     * @return Base
+     *
+     * @throws InvalidArgumentException
+     */
+    public function add($arg, $alias = null)
+    {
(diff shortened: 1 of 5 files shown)


SenseException:
Hi @maxhelias,

thank you for your contribution. Can you please evaluate the benefit of this alias style? I'm not sure I understand what the advantage is.

I also suggest not to create a new abstract class for implementing the new feature and put it into the base class.

maxhelias:
Hi @SenseException,

For the case of an application with a large and segmented data model, we find ourselves regularly with multiple join sql. This kind of query is use for fill a table in realtime (so no cache, just one shot)
To optimize the query, we only select the fields we need instead of `*` but the syntax becomes less readable.

Ex: 
```
->select(
    'op.id',
    'op.name',
    'a.street_number',
    'a.street_extension',
    'a.street_name',
    'a.additional_address',
    'a.distribution_mention',
    'a.latitude',
    'a.longitude',
    'a.establishment_id as establishment_id',
    'city.name as city_name',
    'city.zipcode as zipcode',
    'e'
)
```

Could become for better readability : 
```
->select([
    'op' => [
        'id',
        'name',
    ],
    'a' => [
        'street_number',
        'street_extension',
        'additional_address',
        'distribution_mention',
        'latitude',
        'longitude',
        'establishment_id',
    ],
    'city' => [
        'name',
        'zipcode',
    ],
    'e',
])
``` 
Note : This example is still relatively short but at first we can easily see the different tables.

For the implementation, I created a new abstract class to limit the performance impact only to the selectors concerned (select, groupBy and futur other). I could change it if the feature interests you.

delirehberi (approved):
seems right

beberlei:
You can write yourself a helper function that does this:

```php
function selectZip(array $select) {
    $result = [];
    foreach ($select as $entity => $fields) {
        foreach ($fields as $field) {
            $result[] = $entity . "." . $field;
    }
    return $result;
}
```
$qb->select(...selectZip([
    'op' => [
        'id',
        'name',
    ],
    'a' => [
        'street_number',
        'street_extension',
        'additional_address',
        'distribution_mention',
        'latitude',
        'longitude',
        'establishment_id',
    ],
    'city' => [
        'name',
        'zipcode',
    ],
    'e',
]));
```

PR #7803, opened 2019-08-19
Fix when SimpleArithmetic is used in Pagination order by

Seb33300:
This fix Pagination limit subquery when a `SimpleArithmetic` is used in the `ORDER BY`.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
@@ -23,10 +23,10 @@ use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\Query\AST\Functions\IdentityFunction;
+use Doctrine\ORM\Query\AST\Node;
 use Doctrine\ORM\Query\AST\PathExpression;
 use Doctrine\ORM\Query\AST\SelectExpression;
 use Doctrine\ORM\Query\AST\SelectStatement;
-use Doctrine\ORM\Query\AST\SimpleArithmeticExpression;
 use Doctrine\ORM\Query\TreeWalkerAdapter;
 
 /**
@@ -99,7 +99,7 @@ class LimitSubqueryWalker extends TreeWalkerAdapter
         }
 
         foreach ($AST->orderByClause->orderByItems as $item) {
-            if ($item->expression instanceof PathExpression || $item->expression instanceof SimpleArithmeticExpression) {
+            if ($item->expression instanceof Node) {
                 $AST->selectClause->selectExpressions[] = new SelectExpression(
                     $this->createSelectExpressionItem($item->expression),
                     '_dctrn_ord' . $this->_aliasCounter++
@@ -159,11 +159,11 @@ class LimitSubqueryWalker extends TreeWalkerAdapter
     /**
      * Retrieve either an IdentityFunction (IDENTITY(u.assoc)) or a state field (u.name).
      *
-     * @param mixed $expression
+     * @param Node $expression
      *
-     * @return IdentityFunction
+     * @return Node
      */
-    private function createSelectExpressionItem($expression)
+    private function createSelectExpressionItem(Node $expression)
     {
         if ($expression instanceof PathExpression && $expression->type === PathExpression::TYPE_SINGLE_VALUED_ASSOCIATION) {
             $identity = new IdentityFunction('identity');
--- a/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
@@ -22,11 +22,12 @@ namespace Doctrine\ORM\Tools\Pagination;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Query;
-use Doctrine\ORM\Query\TreeWalkerAdapter;
 use Doctrine\ORM\Query\AST\Functions\IdentityFunction;
 use Doctrine\ORM\Query\AST\PathExpression;
 use Doctrine\ORM\Query\AST\SelectExpression;
 use Doctrine\ORM\Query\AST\SelectStatement;
+use Doctrine\ORM\Query\AST\SimpleArithmeticExpression;
+use Doctrine\ORM\Query\TreeWalkerAdapter;
 
 /**
  * Replaces the selectClause of the AST with a SELECT DISTINCT root.id equivalent.
@@ -98,11 +99,29 @@ class LimitSubqueryWalker extends TreeWalkerAdapter
         }
 
         foreach ($AST->orderByClause->orderByItems as $item) {
-            if ($item->expression instanceof PathExpression) {
-                $AST->selectClause->selectExpressions[] = new SelectExpression(
-                    $this->createSelectExpressionItem($item->expression), '_dctrn_ord' . $this->_aliasCounter++
-                );
+            switch (true) {
+                case $item->expression instanceof PathExpression:
+                    $expressions = [$item->expression];
+                    break;
+                case $item->expression instanceof SimpleArithmeticExpression:
+                    $expressions = $item->expression->arithmeticTerms;
+                    break;
+                default:
+                    $expressions = [];
+            }
 
+            $hasPathExpression = false;
+
+            foreach ($expressions as $expression) {
+                if ($expression instanceof PathExpression) {
+                    $AST->selectClause->selectExpressions[] = new SelectExpression(
+                        $this->createSelectExpressionItem($expression), '_dctrn_ord' . $this->_aliasCounter++
+                    );
+                    $hasPathExpression = true;
+                }
+            }
+
+            if ($hasPathExpression) {
                 continue;
             }
 
--- a/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
@@ -159,8 +159,6 @@ class LimitSubqueryWalker extends TreeWalkerAdapter
     /**
      * Retrieve either an IdentityFunction (IDENTITY(u.assoc)) or a state field (u.name).
      *
-     * @param Node $expression
-     *
      * @return Node
      */
     private function createSelectExpressionItem(Node $expression)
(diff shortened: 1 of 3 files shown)


lcobucci:
@Seb33300 would please also add some functional tests that cover that use case?

You can find examples on https://github.com/doctrine/doctrine2/tree/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket

Seb33300:
@lcobucci done

Seb33300:
It looks like PostgreSQL require the full ORDER BY expression in the SELECT to work...

> SQLSTATE[42P10]: Invalid column reference: 7 ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
```diff
@@ -157,20 +159,20 @@ private function validate(SelectStatement $AST)
     /**
      * Retrieve either an IdentityFunction (IDENTITY(u.assoc)) or a state field (u.name).
      *
-     * @param \Doctrine\ORM\Query\AST\PathExpression $pathExpression
+     * @param mixed $expression
```
Ocramius: This should not be `mixed`: let's try avoiding `mixed` overall

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7805Test.php
```diff
@@ -0,0 +1,40 @@
...
+        $article = new CmsArticle();
+
+        $article->topic   = 'Test SimpleArithmetic ORDER BY';
+        $article->text    = 'This test fails on MySQL.';
+        $article->version = 1;
+
+        $this->_em->persist($article);
+        $this->_em->flush();
+
+        $query = $this->_em->createQuery('SELECT a FROM Doctrine\Tests\Models\CMS\CmsArticle a ORDER BY a.version + 0 ASC');
```
Ocramius: Does this now also work with custom DQL functions?

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7805Test.php
Seb33300: Originally, only `PathExpression` found in the `ORDER BY` part was added to the `SELECT`, everything else was skipped.

I just added another condition to also accept `SimpleArithmetic` (see line 102 upper in this file) so i don't think custom DQL functions will work unless they are considered as `PathExpression` or `SimpleArithmetic`.

Maybe we should consider that every `Node` found in the `ORDER BY` part could be added in the `SELECT` part and remove the condition on line 102?

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
Seb33300: I can replace it by `Doctrine\ORM\Query\AST\Node`?

Ocramius: That would be much better, thanks

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7805Test.php
Ocramius: I'm not sure why the `Identity` function wrapper was added in first place, but if you could add an SQL generation code using something like `SQRT()`, that would be sufficient to check how we need to modify the code.

Seb33300: Done

Seb33300:
Can someone please remove the `Missing Tests` label?

tests/Doctrine/Tests/ORM/Tools/Pagination/LimitSubqueryWalkerTest.php
```diff
@@ -47,7 +47,21 @@ public function testLimitSubqueryWithSortFunction() : void
         $limitQuery->setHint(Query::HINT_CUSTOM_TREE_WALKERS, [LimitSubqueryWalker::class]);
 
         self::assertSame(
-            'SELECT DISTINCT m0_.id AS id_0 FROM MyBlogPost m0_ INNER JOIN Category c1_ ON m0_.category_id = c1_.id GROUP BY m0_.id ORDER BY COUNT(c1_.id) ASC',
+            'SELECT DISTINCT m0_.id AS id_0, COUNT(c1_.id) AS sclr_1 FROM MyBlogPost m0_ INNER JOIN Category c1_ ON m0_.category_id = c1_.id GROUP BY m0_.id ORDER BY COUNT(c1_.id) ASC',
```
greg0ire: This line is getting long, how about adding some linebreaks?
```suggestion
            <<<SQL
      SELECT DISTINCT m0_.id AS id_0, COUNT(c1_.id) AS sclr_1 
         FROM MyBlogPost m0_ 
INNER JOIN Category c1_ ON m0_.category_id = c1_.id 
 GROUP BY m0_.id 
 ORDER BY COUNT(c1_.id) ASC
 SQL
            ,
```

tests/Doctrine/Tests/ORM/Tools/Pagination/LimitSubqueryWalkerTest.php
```diff
@@ -47,7 +47,21 @@ public function testLimitSubqueryWithSortFunction() : void
...
+            $limitQuery->getSQL()
+        );
+    }
+
+    public function testLimitSubqueryWithSortSimpleArithmetic() : void
+    {
+        $dql   = 'SELECT p FROM Doctrine\Tests\ORM\Tools\Pagination\MyBlogPost p JOIN p.category c GROUP BY p.id ORDER BY p.title + 0';
```
greg0ire:
```suggestion
        $dql   = <<<'DQL'
     SELECT p 
        FROM Doctrine\Tests\ORM\Tools\Pagination\MyBlogPost p 
          JOIN p.category c 
GROUP BY p.id 
ORDER BY p.title + 0
DQL;
```

tests/Doctrine/Tests/ORM/Tools/Pagination/LimitSubqueryWalkerTest.php
```diff
@@ -47,7 +47,21 @@ public function testLimitSubqueryWithSortFunction() : void
...
+        $query = $this->entityManager->createQuery($dql);
+
+        $limitQuery = clone $query;
+        $limitQuery->setHint(Query::HINT_CUSTOM_TREE_WALKERS, [LimitSubqueryWalker::class]);
+
+        self::assertSame(
+            'SELECT DISTINCT m0_.id AS id_0, m0_.title + 0 AS sclr_1 FROM MyBlogPost m0_ INNER JOIN Category c1_ ON m0_.category_id = c1_.id GROUP BY m0_.id ORDER BY m0_.title + 0 ASC',
```
greg0ire:
```suggestion
            <<<'SQL'
    SELECT DISTINCT m0_.id AS id_0, m0_.title + 0 AS sclr_1
      FROM MyBlogPost m0_
INNER JOIN Category c1_ ON m0_.category_id = c1_.id
  GROUP BY m0_.id
  ORDER BY m0_.title + 0 ASC
SQL
```

tests/Doctrine/Tests/ORM/Tools/Pagination/LimitSubqueryWalkerTest.php
Seb33300: I wanted to keep consistency with others queries.
I we want to rewrite queries on multilines, we should do it everywhere.

greg0ire: Not all queries should be multiline, just the long ones.

Seb33300: Yes, but if you look at others existing queries in the same file, they are all long queries.
That's why I think it should be done in a separate PR if you want to rewrite all queries on multiline.

greg0ire: That's an issue I often encounter as a maintainer: we make a style mistake, and then it's repeated over and over for consistency's sake, until someone goes through the painful task of rewriting it all properly. And the longer you wait, the more painful the task is (for instance here 10 queries to rewrite before, 13 after). And often, there is no actual rule in the style guide that says we should write things in some kind of way. In that case, people probably did not know any better. 

Luckily, someone already broke the unworded rule:
https://github.com/doctrine/orm/blob/21bc805dc9662ff05f3038b30a29e05591838a29/tests/Doctrine/Tests/ORM/Functional/LifecycleCallbackTest.php#L190-L199
So don't worry with being inconsistent with other occurrences, the project is already inconsistent.

lcobucci (changes requested):
@Seb33300 thanks for sending this PR but I believe this to be invalid.

As documented in https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/dql-doctrine-query-language.html#doctrine-query-language users are supposed to use `HIDDEN` fields to be able to have arithmetic expression in ORDER clause.

By changing the queries of your test to use this feature the test passes without requiring any modification in the source.

<details>

```php
<?php

declare(strict_types=1);

use Doctrine\ORM\Tools\Pagination\Paginator;
use Doctrine\Tests\Models\CMS\CmsArticle;
use function count;
use function iterator_to_array;

class GH7805Test extends OrmFunctionalTestCase
{
    protected function setUp()
    {
        $this->useModelSet('cms');
        parent::setUp();
    }

    public function testPaginationWithFunction()
    {
        $article = new CmsArticle();

        $article->topic   = 'Test ORDER BY with Function';
        $article->text    = 'This test fails on MySQL if ORDER BY part is not added to SELECT.';
        $article->version = 1;

        $this->_em->persist($article);
        $this->_em->flush();

        $query = $this->_em->createQuery('SELECT a, SQRT(a.version) AS HIDDEN function_result FROM Doctrine\Tests\Models\CMS\CmsArticle a ORDER BY function_result ASC');
        $query->setFirstResult(0);
        $query->setMaxResults(1);

        $paginator = new Paginator($query, true);
        $paginator->setUseOutputWalkers(false);
        $this->assertEquals(1, count(iterator_to_array($paginator)));
    }

    public function testPaginationWithSimpleArithmetic()
    {
        $article = new CmsArticle();

        $article->topic   = 'Test ORDER BY with SimpleArithmetic';
        $article->text    = 'This test fails on MySQL if ORDER BY part is not added to SELECT.';
        $article->version = 1;

        $this->_em->persist($article);
        $this->_em->flush();

        $query = $this->_em->createQuery('SELECT a, a.version + 0 AS HIDDEN simple_arithmetic FROM Doctrine\Tests\Models\CMS\CmsArticle a ORDER BY simple_arithmetic ASC');
        $query->setFirstResult(0);
        $query->setMaxResults(1);

        $paginator = new Paginator($query, true);
        $paginator->setUseOutputWalkers(false);
        $this->assertEquals(1, count(iterator_to_array($paginator)));
    }
}
```
</details>

lcobucci:
I'll close this as invalid as per my comment. We can discuss things further if you want :+1:

Issue #7804, opened 2019-08-19
Using DTOs with groups


core23:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

#### Summary

Since Version 2.4, it is possible to create queries with DTO result objects using `NEW` operator. Unfortunately it is not possible to use field aliases with this.

#### Current behavior

There is a `RuntimeException`:

```
An exception has been thrown during the rendering of a template ("[Syntax Error] line 0, col 95: Error: Expected Doctrine\ORM\Query\Lexer::T_CLOSE_PARENTHESIS, got 'as'").
```

#### How to reproduce

I use the https://github.com/beberlei/DoctrineExtensions library for `MONTH`, `YEAR`, `COUNT` and `SUM` support, but this should not be related to the actual problem.

Create the following query:

```php
$repository
    ->createQueryBuilder('i')
    ->select(
        sprintf('new %s(
        MONTH(i.invoiceDate) as mn,
        COUNT(DISTINCT i.id),
        SUM(p.productPrice * p.quantity * (100 - i.discount) / 100),
        SUM(p.productPrice * p.quantity * (100 - i.discount) / 100 * t.tax),
        YEAR(i.invoiceDate) as yr
        )
        ', StatisticViewModel::class)
    )
    ->leftJoin('i.positions', 'p')
    ->leftJoin('p.tax', 't')
    ->andWhere('i.invoiceDate is not null')
    ->addGroupBy('mn') 
    ->addGroupBy('yr') 
```

#### Expected behavior

I can use aliases in the select method and reuse them later when creating groups.

lcobucci:
@core23 I'd say that it would be more logical to have an alias for the entire DTO and use its properties. Like:

```php
$repository
    ->createQueryBuilder('i')
    ->select(
        sprintf('new %s(
        MONTH(i.invoiceDate),
        COUNT(DISTINCT i.id),
        SUM(p.productPrice * p.quantity * (100 - i.discount) / 100),
        SUM(p.productPrice * p.quantity * (100 - i.discount) / 100 * t.tax),
        YEAR(i.invoiceDate)
        ) as dto
        ', StatisticViewModel::class)
    )
    ->leftJoin('i.positions', 'p')
    ->leftJoin('p.tax', 't')
    ->andWhere('i.invoiceDate is not null')
    ->addGroupBy('dto.month') 
    ->addGroupBy('dto.year') 
```

[...]

PR #7807, opened 2019-08-20
Test to reproduce using field alias in NEW syntax

core23:
This PR adds a test that reproduces a bug that occurs when using an alias inside the `SELECT NEW` syntax. Refs issue #7804 

**Expected result**

The same query is created like without using the `NEW` operator.

**Actual result**

A `QueryException` occurs:
```
[Syntax Error] line 0, col 89: Error: Expected Doctrine\ORM\Query\Lexer::T_CLOSE_PARENTHESIS, got 'as'
```


tests/Doctrine/Tests/ORM/Functional/Ticket/GH7804.php
```diff
@@ -0,0 +1,88 @@
...
+
+    public function testPositiveQuery(): void
+    {
+        $qb = $this->em->createQueryBuilder();
+
+        $qb->from(GH7804Entity::class, 'te1')
+            ->select('te1.title as t')
+            ->addGroupBy('t');
+
+        self::assertSQLEquals(
```
lcobucci: Please test the result of the query instead of the generated query to avoid platform incompatibility issues.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7804.php
```diff
@@ -0,0 +1,88 @@
...
+        );
+    }
+
+    public function testIssue(): void
+    {
+        $qb = $this->em->createQueryBuilder();
+
+        $qb->from(GH7804Entity::class, 'te1')
+            ->select(
+                sprintf('new %s(te1.title as t)', GH7804Dto::class)
```
lcobucci: Can you use the hidden field as we discussed before?

lcobucci (changes requested):
Thanks for submitting the test, can you address these points? And since this might be a bug please change the target branch to 2.6 and rebase your branch with it.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7804.php
```diff
@@ -0,0 +1,126 @@
...
+        self::assertEquals('foo', $result[0]->getTitle());
+    }
+
+    public function testQueryWithNewAlias(): void
+    {
+        $qb = $this->_em->createQueryBuilder();
+
+        $result = $qb->from(GH7804Entity::class, 'a')
+            ->select(
+                sprintf('new %s(a.title as t)', GH7804Dto::class)
```
lcobucci: As we said before, this is not supported at all. So it's expected to fail.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7804.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    public function testQueryWithNewAndHidden(): void
+    {
+        $qb = $this->_em->createQueryBuilder();
+
+        $result = $qb->from(GH7804Entity::class, 'a')
+            ->select(
+                sprintf('new %s(a.title)', GH7804Dto::class),
+                'a.title as HIDDEN h1'
```
lcobucci: This works fine, though (your test passes).

lcobucci (changes requested):
@core23 it looks to me that the issue is `invalid`. We can't have aliases inside of the `NEW` operator, that's just not supported.

Did you manage to have your app running using the hidden field? If not, can you modify this test to make that scenario fail?

PR #7818, opened 2019-09-15
Add note into docs about not using SimpleAnnotationReader

SenseException:
This is a followup PR to #7753 adding a note about not using the `SimpleAnnotationReader` in ORM 2.6.


lcobucci (approved):
Great, thanks!

PR #7821, opened 2019-09-17
Bug: #7820 paginator ignores dbal type conversions in identifiers

### LINKED ISSUE #7820: Paginator won't take into account entity identifiers that use custom DBAL types
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

#### Summary

When using a `Doctrine\ORM\Tools\Pagination\Paginator` to iterate over a query that has entities with a custom DBAL type used in the identifier, then `$id->__toString()` is used implicitly by PDO, instead of being converted by the `Doctrine\DBAL\Types` system.

In order to reproduce this, you must have identifiers implementing `#__toString()` (to allow the `UnitOfWork` to hash them) and other accessors that are used by the custom DBAL type during DB/PHP conversions. If `#__toString()` and the DBAL type conversions are asymmetric, then the paginator will fail to find records.

Tricky situation, but this very much affects `ramsey/uuid-doctrine` and anyone relying on the `uuid_binary`.

#### Current behavior

```php
class MyEntity {
    public TheId $id;
}

class TheId {
    public function toId() : string
    {
        return 'yep';
    }
    public function __toString() : string
    {
        return 'nope';
    }
}

class TheIdType extends StringType
{
    public function convertToDatabaseValue($value, $platform)
    {
        return $value->toId();
    }
}

$entity = new MyEntity();
$entity->id = new TheId();

$em->persist($entity);
$em->flush();

$paginator = (new Paginator($em->createQuery('SELECT f FROM ' . MyEntity::class . ' f'));

foreach ($paginator as $entry) {
    die('this won't be reached with current ORM state');
}
```

#### Expected behavior

The code above should loop over one found record.

Ocramius:
Fixes #7820

Props to @lyska for giving me the opportunity to hunt this one down!

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -2017,7 +2017,7 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param string $fieldName
      *
-     * @return \Doctrine\DBAL\Types\Type|string|null
+     * @return string|null
      *
      * @todo 3.0 Remove this. PersisterHelper should fix it somehow
      */
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -2033,7 +2033,7 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param string $columnName
      *
-     * @return \Doctrine\DBAL\Types\Type|string|null
+     * @return string|null
      *
      * @deprecated 3.0 remove this. this method is bogus and unreliable, since it cannot resolve the type of a column
      *             that is derived by a referenced field on a different entity.
--- a/lib/Doctrine/ORM/Query/ResultSetMapping.php
+++ b/lib/Doctrine/ORM/Query/ResultSetMapping.php
@@ -19,6 +19,12 @@
 
 namespace Doctrine\ORM\Query;
 
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Mapping\MappingException;
+use function array_keys;
+use function array_values;
+use function assert;
+
 /**
  * A ResultSetMapping describes how a result set of an SQL query maps to a Doctrine result.
  *
@@ -583,5 +589,40 @@ class ResultSetMapping
 
         return $this;
     }
-}
 
+    /**
+     * Retrieves the DBAL type name for the single identifier column of the root of a selection.
+     * Composite identifiers not supported!
+     *
+     * @internal only to be used by ORM internals: do not use in downstream projects! This API is a minimal abstraction
+     *           that only ORM internals need, and it tries to make sense of the very complex and squishy array-alike
+     *           structure inside this class. Some assumptions are coded in here, so here be dragons.
+     *
+     * @throws MappingException If the identifier is not a single field, or if metadata for its
+     *                          owner is incorrect/missing.
+     */
+    final public function getTypeOfSelectionRootSingleIdentifierColumn(EntityManagerInterface $em) : string
+    {
+        assert($this->isSelect);
+
+        if ($this->isIdentifierColumn !== []) {
+            // Identifier columns are already discovered here: we can use the first one directly.
+            assert($this->typeMappings !== []);
+
+            return $this->typeMappings[array_keys(array_values($this->isIdentifierColumn)[0])[0]];
+        }
+
+        // We are selecting entities, and the first selected entity is our root of the selection.
+        if ($this->aliasMap !== []) {
+            $metadata = $em->getClassMetadata($this->aliasMap[array_keys($this->aliasMap)[0]]);
+
+            return $metadata->getTypeOfField($metadata->getSingleIdentifierFieldName());
+        }
+
+        // We are selecting scalar fields - the first selected field will be assumed (!!! assumption !!!) as identifier
+        assert($this->scalarMappings !== []);
+        assert($this->typeMappings !== []);
+
+        return $this->typeMappings[array_keys($this->scalarMappings)[0]];
+    }
+}
--- a/lib/Doctrine/ORM/Query/ResultSetMapping.php
+++ b/lib/Doctrine/ORM/Query/ResultSetMapping.php
@@ -19,12 +19,6 @@
 
 namespace Doctrine\ORM\Query;
 
-use Doctrine\ORM\EntityManagerInterface;
-use Doctrine\ORM\Mapping\MappingException;
-use function array_keys;
-use function array_values;
-use function assert;
-
 /**
  * A ResultSetMapping describes how a result set of an SQL query maps to a Doctrine result.
  *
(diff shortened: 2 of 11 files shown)


lib/Doctrine/ORM/Query/ResultSetMapping.php
```diff
@@ -583,5 +589,40 @@ public function addMetaResult($alias, $columnName, $fieldName, $isIdentifierColu
 
         return $this;
     }
-}
 
+    /**
+     * Retrieves the DBAL type name for the single identifier column of the root of a selection.
+     * Composite identifiers not supported!
```
lcobucci: Is there any test covering exceptions when trying to paginate entities with composite identifiers? Perhaps it might be important to add one, especially because we're strengthening this restriction.

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -150,14 +155,21 @@ public function getIterator()
...
             // don't do this for an empty id array
-            if (count($ids) === 0) {
+            if ($foundIdRows === []) {
                 return new \ArrayIterator([]);
             }
 
+            $whereInQuery = $this->cloneQuery($this->query);
+            $em           = $subQuery->getEntityManager();
+            $connection   = $em->getConnection();
+            $idType       = $this->getIdentifiersQueryScalarResultType($subQuery, $em);
```
lcobucci: Wouldn't we keep things a bit more contained if we perform this conversion in `WhereInWalker`? There's a lot going on there too but we already have the class metadata and it might be easier to detect the type?

lcobucci: And I think we would also mitigate possible performance impact due to double query parsing (although query cache reduces that)

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+}
+
+/** @Entity */
+class GH7820Line
+{
+    /**
+     * @var GH7820LineText
+     * @Id()
+     * @Column(type="Doctrine\Tests\ORM\Functional\Ticket\GH7820LineTextType")
```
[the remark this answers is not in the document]
Ocramius: Yeap, will need to re-visit this, since it didn't like it last time I tried it.

lib/Doctrine/ORM/Query/ResultSetMapping.php
Ocramius: I'll add one test specific to `ResultSetMapping`

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
Ocramius: > Wouldn't we keep things a bit more contained if we perform this conversion in `WhereInWalker`? 

Initially attempted this, but `WhereInWalker` does not have access to the DQL parameters, only to the DQL string/AST.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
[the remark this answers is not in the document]
Ocramius: Not applicable: seems like I hit an annotation bug in this case, since the class is ignored (not part of `use` statements, after all.

Won't fix for this patch though.

Ocramius:
Green! @lcobucci this should be much better now :+1:

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -2017,7 +2017,7 @@ public function isIdentifierUuid()
      *
      * @param string $fieldName
      *
-     * @return \Doctrine\DBAL\Types\Type|string|null
```
lcobucci: Is this change necessary/correct? Doesn't it affect `ClassMetadataInfo#getTypeOfColumn()` too?

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -150,14 +151,16 @@ public function getIterator()
...
-            if (count($ids) === 0) {
+            if ($foundIdRows === []) {
                 return new \ArrayIterator([]);
             }
 
+            $whereInQuery = $this->cloneQuery($this->query);
+            $ids          = array_map('current', $foundIdRows);
+
             $this->appendTreeWalker($whereInQuery, WhereInWalker::class);
             $whereInQuery->setHint(WhereInWalker::HINT_PAGINATOR_ID_COUNT, count($ids));
```
lcobucci: I know it's not really related to your fix, but shouldn't we keep the matched IDs as a hint too? Would increase the cohesion a bit more too.

lcobucci: We might leave this for v3, though.

lcobucci (changes requested):

lcobucci:
@Ocramius It's muuuuch better indeed. Great job there!

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Ocramius: It is correct though, since this endpoint always returns a `string`, and the code below does indeed rely on a `string` being retrieved.

If a DBAL type was returned (which is never the case), a string cast would not be acceptable.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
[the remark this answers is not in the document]
lcobucci: That's quite unexpected! Well, nothing we can sort out now :smile:

lcobucci (approved):
Awesome!

ksowa:
@Ocramius It looks like this PR caused another bug. 

I use `ramsey/uuid` library. When trying to get collection of objects with UUID as a Primary Key, it fails just after  `Doctrine\ORM\Tools\Pagination\Paginator::getIterator()`, which pulls PKs as scalar:

```
$foundIdRows = $subQuery->getScalarResult();
// ...
$whereInQuery = $this->cloneQuery($this->query);
$ids = array_map('current', $foundIdRows);
```
In the `v2.6.3` the value was properly converting from the string to PHP value and then back to database value (`convertToPHPValue()` -> `convertToDatabaseValue()`). Since `v2.6.4` the `convertToDatabaseValue` is called first, which causes an error. 

I'm not sure whether this is issue in `ramsey/uuid` or Doctrine, but I think the change caused BC break.

Ocramius:
Note that I'm working specifically on this issue (and #7865) because of `ramsey/uuid`

akorz:
@Ocramius @ksowa the same issue for me

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
mpdude: Just for the record, #8797 works in the other direction to replace the `count()` with a bool, since the information is not actually needed (at this time) and it improves cache efficiency.

PR #7825, opened 2019-09-22
Improved support for Mapped Superclasses

### LINKED ISSUE #5998: table inheritance 'JOINED' issue with selecting reference fields from base table
I have a base task and task which is derived off of basetask using CTI

``` php
/*
 * @Entity
 * @InheritanceType("JOINED")
 * @DiscriminatorMap({"base_task" = "BaseTask",
 *     "task" = "Task"`
 * })
 */
class BaseTask 
{
 ...

   /**
     * @var User
     * @ManyToOne(targetEntity="User")
     * @JoinColumn(name="owner_id", referencedColumnName="id", onDelete="SET NULL")
     */
    protected $owner;
 ...//getter and setter
}
```

``` php
class Task extends BaseTask 
{
}
```

For any query on Task involving owner field using doctrine query builder results in this issue

```
An exception occurred while executing 'SELECT o0_.id AS id_0 FROM xyz_task w1_ INNER JOIN xyz_base_task w2_ ON w1_.id = w2_.id INNER JOIN xyz_user o0_ ON w1_.owner_id = o0_.id WHERE o0_.id = ? LIMIT 1' with params [2]:

SQLSTATE[42703]: Undefined column: 7 ERROR: column w1_.owner_id does not exist
LINE 1: ...w2_ ON w1_.id = w2_.id INNER JOIN xyz_user o0_ ON w1_.owner_...
^
HINT: Perhaps you meant to reference the column "w2_.owner_id".
```

Seems like there is an issue with walkjoin in correctly dereferencing the field name to the correct table in the inheritance hierarchy for reference fields only. It works as expected for normal fields.

Any help please?

Doctrine version is 2.5.1 and I tried 2.5.4 as well and its the same behavior.

andrews05:
This PR makes improvements to support Mapped Superclasses in a number of ways that didn't work previously.
- Support for version property declared in mapped superclass
- Better support for to-one relationships declared in mapped superclass
- Support for mapped superclasses both above and within JTI/STI hierarchies

Fixes #5998.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadata.php
@@ -720,20 +720,46 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
     public function isInheritedProperty($fieldName)
     {
         $declaringClass = $this->properties[$fieldName]->getDeclaringClass();
-        if ($declaringClass->className === $this->className) {
-            return false;
-        } else if (! $declaringClass->isMappedSuperclass) {
-            return true;
+
+        return $declaringClass->className !== $this->className;
+    }
+
+    /**
+     * Checks whether a mapped column is inherited from a parent table.
+     *
+     * @param string $fieldName
+     *
+     * @return bool TRUE if the field is inherited, FALSE otherwise.
+     */
+    public function isInheritedColumn($fieldName)
+    {
+        $declaringClass = $this->getPropertyTableClass($fieldName);
+
+        return $declaringClass->className !== $this->className;
+    }
+
+    /**
+     * Gets the class for the table that a field belongs to.
+     *
+     * @param string $fieldName
+     *
+     * @return ClassMetadata.
+     */
+    public function getPropertyTableClass($fieldName) : ClassMetadata
+    {
+        $declaringClass = $this->properties[$fieldName]->getDeclaringClass();
+        if (! $declaringClass->isMappedSuperclass) {
+            return $declaringClass;
         }
 
-        $parentClass = $this;
-        while (($parentClass = $parentClass->getParent()) !== $declaringClass) {
+        $tableClass = $parentClass = $this;
+        while (($parentClass = $parentClass->getParent()) && $parentClass !== $declaringClass) {
             if (! $parentClass->isMappedSuperclass) {
-                return true;
+                $tableClass = $parentClass;
             }
         }
 
-        return false;
+        return $tableClass;
     }
 
     /**
@@ -861,7 +887,13 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
             $this->fieldNames[$property->getColumnName()] = $property->getName();
         } elseif ($inheritedProperty instanceof AssociationMetadata) {
             if ($declaringClass->isMappedSuperclass) {
-                $inheritedProperty->setSourceEntity($this->className);
+                $tableClass = $parentClass = $this;
+                while (($parentClass = $parentClass->getParent()) && $parentClass !== $declaringClass) {
+                    if (! $parentClass->isMappedSuperclass) {
+                        $tableClass = $parentClass;
+                    }
+                }
+                $inheritedProperty->setSourceEntity($tableClass->className);
             }
 
             // Need to add inherited fieldNames
--- a/lib/Doctrine/ORM/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadata.php
@@ -263,9 +263,13 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
 
     public function getRootClassName() : string
     {
-        return $this->parent instanceof ClassMetadata && ! $this->parent->isMappedSuperclass
-            ? $this->parent->getRootClassName()
-            : $this->className;
+        $rootClass = $parentClass = $this;
+        while (($parentClass = $parentClass->getParent()) !== null) {
+            if (! $parentClass->isMappedSuperclass) {
+                $rootClass = $parentClass;
+            }
+        }
+        return $rootClass->className;
     }
 
     /**
--- a/lib/Doctrine/ORM/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadata.php
@@ -720,7 +720,7 @@ class ClassMetadata extends ComponentMetadata implements TableOwner
     public function isInheritedProperty($fieldName)
     {
         $declaringClass = $this->properties[$fieldName]->getDeclaringClass();
-        if ($declaringClass === $this) {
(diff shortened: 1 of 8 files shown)


lcobucci:
@andrews05 thanks for sending this patch. Since this is a bug fix it must be sent to 2.6 instead of master and we need a functional test showing the problem and this fixes it.

andrews05:
@lcobucci I did look at v2.6 but the code looked quite different in some places so I wasn't sure if I could fix it the same way. (I also wasn't sure if it would really be classified as bug, it's perhaps more a matter of supporting a certain combination of features that wasn't previously supported).

I could try a v2.6 patch as well if you like though. And I will work on putting together a test...

Ocramius:
@andrews05 can you maybe open a separate patch against `2.6` to just highlight the bug? A failing test case would suffice.

lcobucci:
@Ocramius @andrews05 brings a fair point, though.

I think that the bug relates to even allowing to mix a mapped superclass in an STI/JTI and that should trigger a schema validation error. We didn't design for this use case and I don't believe we should support it.

What do you think about it @guilhermeblanco?

andrews05:
I've added a test. I had to also change ClassMetadata::isInheritedProperty to check for MappedSuperclasses in order for schema creation to work.
The test is currently only testing the case where the root table extends a mappedsuperclass. There are some issues still remaining if a mappedsuperclass exists between the root table and a joined subclass.
Much of the problem revolves around the isInheritedProperty/getDeclaringClass methods. It might be helpful to have another method like `ClassMetadata::getTableClass($fieldName)` which would return the class of the table containing the given column.

@lcobucci MappedSuperclass with STI already works fine, as far as I can tell. (With the exception of schema creation which is fixed with my latest commit.)

lcobucci:
> @lcobucci MappedSuperclass with STI already works fine, as far as I can tell. (With the exception of schema creation which is fixed with my latest commit.)

Can you please provide a test for that too? If we're going to support this scenario it can't break ever again.

andrews05:
@lcobucci I've updated the test to include STI as well, and also fixed the case where a mapped superclass exists between the root table and a joined subclass.

andrews05:
I've fixed up coding style and resolved a couple of failing tests. One failure still remains though:
In BasicInheritanceMappingTest it is asserting that a property declared in a mappedsuperclass returns true for isInheritedProperty(). This seems odd to me because as I best I can tell, all other uses of isInheritedProperty are only interested in table inheritance. Any advice?

lib/Doctrine/ORM/Mapping/ClassMetadata.php
```diff
@@ -715,8 +720,20 @@ public function isRootEntity()
...
     {
         $declaringClass = $this->properties[$fieldName]->getDeclaringClass();
+        if ($declaringClass->className === $this->className) {
+            return false;
+        } else if (! $declaringClass->isMappedSuperclass) {
+            return true;
+        }
+
+        $parentClass = $this;
+        while (($parentClass = $parentClass->getParent()) !== $declaringClass) {
```
SenseException: You override `$parentClass` here while above you set it to `$this`. Should it be `$this->getParent()` to be more readable or is there a reason for this choice? Something similar can be found in `ClassMetadata::getRootClassName()`.

andrews05: This is just a means of iterating up the hierarchy. The same method is used a number of times already in JoinedSubclassPersister.php.

guilhermeblanco:
> @Ocramius @andrews05 brings a fair point, though.
> 
> I think that the bug relates to even allowing to mix a mapped superclass in an STI/JTI and that should trigger a schema validation error. We didn't design for this use case and I don't believe we should support it.
> 
> What do you think about it @guilhermeblanco?

Actually we originally designed to have any mix of inheritance together, but for simplicity we killed the support for mixing the JTI and STI together, but was always intended to support JTI + MappedSuperclass and STI + MappedSuperclass.

guilhermeblanco:
I feel that most of the parent skipping would be gone once we break the ClassMetadata into more specialized units... I am still curious if there's a better way to go through check of isInheritedProperty and make the test suite pass somehow... maybe check against the getDeclaringClass() ClassMetadata table presence? I don't know yet... would have to dig more to answer this one.

andrews05:
@guilhermeblanco I guess the question is, what is the correct behaviour of isInheritedProperty()? As best I can tell, all existing uses of this method are wanting to know if the property belongs to the current table or not. Since any properties declared in a mapped superclass actually do belong to the table of the subclass, these will likely break if the method returns true for a mapped superclass. Is the failing test perhaps actually invalid?

guilhermeblanco:
@andrews05 maybe a good alternative would be to start differentiating between inheritedProperty (tracking classes) and inheritedColumn (tracking tables).

andrews05:
@guilhermeblanco Just to prove a point: Currently, if you declare a to-one relationship in a mapped superclass and then try to find a child class (without STI/JTI) using the relationship as a condition, it will error. This is fixed by the change to isInheritedProperty here. (I could submit a test case but I don't think it's particularly relevant to this PR).

Anyway, your suggestion does seem sensible. I could implement that new method if you like and change all the existing calls to use it. Although the old method would then only exist for the sake of that one test.

andrews05:
So I'll admit this has turned out to be somewhat more complicated than I initially thought when I first submitted this PR. I've added more tests for relationships and version properties and fixed issues with them via a new method `ClassMetadata::getPropertyTableClass()`.
I've also added the isInheritedColumn() method as @guilhermeblanco suggested, so the BasicInheritanceMappingTest now passes.

guilhermeblanco:
@andrews05 Wow, that is impressive work! Thanks a lot for the work on this. I'm feeling you earned some good internals knowledge while building this patch!
I think I might be able to assume this from now on. I see a few things I could change that would make this patch much easier, so I might just go ahead and do it and then incorporate your PR together. I'll keep this PR open, and add to my todo list to have this supported earlier. Keep in mind I'm working on lazy data types for now, but will soon get back to this.

In case this is something really critical for your team's operation, please let me know and I can try something out.

andrews05:
Thanks! Yeah, I've definitely been learning a lot more than I intended to. Quite happy to leave it to the pros now :)
Some other thoughts I had that might make things easier:
- A getParentTableClass() method, like getParent() but would skip mapped superclasses.
- Set the correct table name on each property during the setTable() method.

No hurry for this, take your time!

mpdude:
I think I've just stumbled across #5998 mentioned in the initial comment, and created #8415 to fix it.

Everyone involved in this discussion here seems to have a deeper understanding of how things work together internally, so I'd really appreciate your reviews over at #8415 to see if it suffices, misses edge cases or could help to get this one here along as well?

mpdude:
@andrews05 I've cherry-picked your functional test over at #7825 and it passes.

Issue #7827, opened 2019-09-23
2.6.4. breaks queries using objects as parameters


kdambekalns:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.6.4

#### Summary

Upgrading from 2.6.3 to 2.6.4 triggers an error related to obejct parameters in queries.

#### Previous behavior

It worked ;)

#### Current behavior

```
Catchable Fatal Error: Object of class Neos\ContentRepository\Domain\Model\Workspace could not be converted to string in /application/Packages/Libraries/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php line 77 - See also: 20190923101155fe8fe7.txt
[...]

PR #7828, opened 2019-09-24
Fix for TransientMetadata on joined subclass

### LINKED ISSUE #7824: Exception when inserting a joined subclass
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | don't know
| Version     | 3.0-dev

#### Summary

When persisting a new entity of a joined subclass, an exception will be thrown if the subclass has defined any properties that are not columns.

#### Current behavior

The following exception will be thrown:
`Call to undefined method Doctrine\ORM\Mapping\TransientMetadata::getColumnName()  in Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php on line 468`

#### How to reproduce

```PHP
/**
 * @ORM\Entity
 * @ORM\DiscriminatorMap({"child" = "Child"})
 */
abstract class Main {
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue
     */
    protected $id;
}

/**
 * @ORM\Entity
 */
class Child extends Main {
    private $someProperty; // Not a column
}

$child = new Child();
$em->persist($child);
$em->flush();
```

#### Expected behavior

Successful insert with no exception.

andrews05:
This PR changes JoinedSubclassPersister to check for and skip TransientMetadata properties while inserting. This prevents an exception being thrown if a non-ORM property is declared on a joined subclass.
Fixes #7824.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
@@ -15,6 +15,7 @@ use Doctrine\ORM\Mapping\JoinColumnMetadata;
 use Doctrine\ORM\Mapping\ManyToManyAssociationMetadata;
 use Doctrine\ORM\Mapping\ToManyAssociationMetadata;
 use Doctrine\ORM\Mapping\ToOneAssociationMetadata;
+use Doctrine\ORM\Mapping\TransientMetadata;
 use Doctrine\ORM\Utility\PersisterHelper;
 use function array_combine;
 use function array_keys;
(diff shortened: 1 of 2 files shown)


SenseException (changes requested):
Thank you for your test case and your first attempt to fix the issue of #7827 

Can you please fix the coding style issues that are listed in travis.ci: https://travis-ci.org/doctrine/orm/jobs/588738823?

It also seems that your first try for a fix broke some other tests.

andrews05:
Thanks, I've fixed the coding style. I'm not too familiar with travis, however I'm pretty sure the other test results are unchanged from before and after my fix. Are you able to tell if there's anything that has broken as a result of this change?

SenseException (changes requested):
Thank you for the fixes. Some of the tests fail, but are temporary excluded from a successful build. I'm not sure about those but it might have something to do with the FK constraints errors.

Can you please be more specific with your commit messages of tests and coding styles? "Add tests" and "Fix CS" are too generic.
https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches

andrews05:
@SenseException sure, I've reworded them now.

SenseException (dismissed):

greg0ire:
Please squash both of your commits together, as there would be no reason to revert one and not revert the other too. When doing so, please use the following as a commit message:

```
Skip transient metadata while inserting

This prevents an exception being thrown if a non-ORM property is declared on a joined subclass.
Fixes #7824.
```

andrews05:
@greg0ire sure, done.

greg0ire (approved):

SenseException (approved):

guilhermeblanco (approved):
LGTM

PR #7831, opened 2019-09-25
Correct condition to detect iteration many association

### LINKED ISSUE #5868: Iterate with fetch join in subquery
There are problem with query->iterate() method when using subquery with fetch join. E.g. there are a User and Group classes, User has a collection of group. When using query like SELECT u from User u WHERE u.id IN (SELECT DISTINCT u2.id from User u2 join u2.groups g WHERE g.name='Admin')
there are exception using iterate() method: Iterate with fetch join in class User using association groups not allowed.
But this query still return one row per User object, and there should be no problems with hydrating. 
It's possible to iterate using such queries?

peter-gribanov:
Maby fix #5868


peter-gribanov:
In its current form does not solve the problem.

Issue #7841, opened 2019-09-30
SchemaTool generates extra diff for platforms without FK support


vpArth:
### Bug Report

SchemaTool generates extra diff for platforms without FK support(ex: SqlitePlatform)

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | all

#### Summary

SchemaTool::getUpdateSchemaSql is never empty, because MetaData contains foreign keys adding, but SqlitePlatform has not support of them

#### How to reproduce

```php
/** @ORM\Entity */
class A {
    /** @ORM\Id @ORM\Column(type="integer") @ORM\GeneratedValue */
    public $id;
    /** @ORM\OneToMany(targetEntity=B::class, mappedBy="parent") */
    public $children;
}
/** @ORM\Entity */
class B {
    /** @ORM\Id @ORM\Column(type="integer") @ORM\GeneratedValue */
    public $id;
    /** @ORM\ManyToOne(targetEntity=A::class) */
    public $parent;
}
// ...
    $class = $em->getClassMetadata(B::class);
    $schemaTool->updateSchema([$class]);
    $diff = $schemaTool->getUpdateSchemaSql([$class]);
```

#### Current behavior
```php
count($diff) > 0;
```

#### Expected behavior
```php
count($diff) === 0;
```

lcobucci:
Handled by #7842

PR #7842, opened 2019-09-30
#7841 SchemaTool generates extra diff for platforms without FK support

vpArth:
Add breaking test for #7841

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -709,6 +709,11 @@ class SchemaTool
         }
 
         $compositeName = $theJoinTable->getName().'.'.implode('', $localColumns);
+
+        if (! $this->platform->supportsForeignKeyConstraints()) {
+            return;
+        }
+
         if (isset($addedFks[$compositeName])
             && ($foreignTableName != $addedFks[$compositeName]['foreignTableName']
             || 0 < count(array_diff($foreignColumns, $addedFks[$compositeName]['foreignColumns'])))
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC2138Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC2138Test.php
@@ -14,6 +14,9 @@ class DDC2138Test extends OrmFunctionalTestCase
     public function testForeignKeyOnSTIWithMultipleMapping()
     {
         $em = $this->_em;
+        if (! $em->getConnection()->getDatabasePlatform()->supportsForeignKeyConstraints()) {
+            $this->markTestSkipped('Platform does not support foreign keys.');
+        }
         $schemaTool = new SchemaTool($em);
 
         $classes = [
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7841Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7841Test.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group GH7841
+ */
+class GH7841Test extends OrmFunctionalTestCase
+{
+    public function testForeignKeysNotCompare() : void
+    {
+        if ($this->_em->getConnection()->getDatabasePlatform()->supportsForeignKeyConstraints()) {
+            $this->markTestSkipped('Test for platforms without foreign keys support');
+        }
+        $class = $this->_em->getClassMetadata(GH7841Child::class);
+        $this->_schemaTool->updateSchema([$class], true);
+        $diff = $this->_schemaTool->getUpdateSchemaSql([$class], true);
+
+        self::assertEmpty($diff);
+
+        $this->_schemaTool->dropSchema([$class]);
+    }
+}
+
+/**
+ * @Entity
+ */
+class GH7841Parent
+{
+    /** @Id @Column(type="integer") @GeneratedValue */
+    public $id;
+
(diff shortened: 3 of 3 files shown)


vpArth:
> `assertEmpty()`

ok, changed

vpArth:
Commits are squashed after both(scrutinizer, travis) checks passed

tests/Doctrine/Tests/ORM/Functional/SchemaTool/SqliteSchemaToolTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\SchemaTool;
```
lcobucci: Please move this to the `Doctrine\Tests\ORM\Functional\Ticket` namespace

tests/Doctrine/Tests/ORM/Functional/SchemaTool/SqliteSchemaToolTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ORM\Annotation as ORM;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Tools\SchemaTool;
+use Doctrine\Tests\OrmTestCase;
+use Doctrine\Tests\TestUtil;
+
+class SqliteSchemaToolTest extends OrmTestCase
```
lcobucci: Please name this test class, and all related classes, after the issue it fixes (`GH7841Test`, `GH7841Parent`, `GH7841Child`) and add a `@group GH7841` annotation to the test class.

lcobucci: I think that using the `OrmFunctionalTestCase` as parent class would simplify this class a bit (remove the need for the properties and the setup method)

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -737,33 +737,35 @@ private function gatherRelationJoinColumns(
...
+                    'foreignColumns' => $foreignColumns,
+                ];
+
+                $theJoinTable->addForeignKeyConstraint(
+                    $foreignTableName,
+                    $localColumns,
+                    $foreignColumns,
+                    $fkOptions
+                );
+            }
```
lcobucci: This portion of the code was already a bit complex and adding an if here just increases the cyclomatic complexity. Would you please try to use some early returns to make this a bit more readable and simpler? Like:

```php
if (! $this->platform->supportsForeignKeyConstraints()) {
    return;
}

if (isset($addedFks[$compositeName]) && ($foreignTableName !== $addedFks[$compositeName]['foreignTableName'] || array_diff($foreignColumns, $addedFks[$compositeName]['foreignColumns']) !== [])) { // expression simplified as well
    // (...)
}

if (isset($blacklistedFks[$compositeName])) {
    return;
}

$addedFks[$compositeName] = ['foreignTableName' => $foreignTableName, 'foreignColumns' => $foreignColumns];
$theJoinTable->addForeignKeyConstraint($foreignTableName, $localColumns, $foreignColumns, $fkOptions);
```

lcobucci (changes requested):
@vpArth thanks for sending this patch! Here are a few comments :+1:

I also believe that this was intended as a bug fix, right? If so, it must use `2.6` as base instead of `master.

tests/Doctrine/Tests/ORM/Functional/SchemaTool/SqliteSchemaToolTest.php
vpArth: I get a bunch of errors in other functional tests when I use OrmFunctionalTest.  
Looks like they are used some dirty global state... 
It was only a reason to use OrmTestCase.
I'll try, ok.

lib/Doctrine/ORM/Tools/SchemaTool.php
vpArth: I was believe, it should be in refactoring stage unrelated to any issue. But ok, will do it.

vpArth:
> it must use `2.6` as base instead of `master.

vendor/bin/phpcs fails every single file on 2.6 branch. is it ok?

lcobucci:
> > it must use `2.6` as base instead of `master.
> 
> vendor/bin/phpcs fails every single file on 2.6 branch. is it ok?

@vpArth we have configured it to validate the diff in PRs because of that. You can focus on your changes only and ensure they're correct.

vpArth:
Should I migrate to new PR to switch target branch to 2.6?

lcobucci (approved):
Awesome :tada:
Thanks!

beberlei:
This is causing a regression in 2.6.5 and 2.7 for SQLite users:

beberlei:
This needs to be reverted. It fixes the symptom and not the problem.

1. SQLite does not support foreign keys, but we emulate them. A check for `supportsForeignKeyConstraints()` is the wrong place, because we interact with DBAL SchemaManager API in SchemaTool that abstracts the platform and if foriegn keys are emulated or supported.

2. Debugging into the code, the culprit is `Comparator::diffTable()` detects an "added foreign key", even though it already exists. That means that the SQLite abstraction does not populate `Table#foreignKeys` with the emulated keys.

3. The real fix is probably that we let `SQLitePlatform#supportsForeignKeyConstraints()` return true.

beberlei:
Fix in DBAL: https://github.com/doctrine/dbal/pull/3762

lcobucci:
@vpArth we had to revert the changes as @beberlei explained. Unfortunately the fix can only be applied for ORM v2.8.0 since it requires us to bump up DBAL version (to 2.10).

vpArth:
But what about #7841? 
The SchemaTool bug for all platforms with false in `supportsForeignKeyConstraints` is still here then.

Adding true into SqlitePlatform could fix the regression without reverting that pr, right?

beberlei:
@vpArth please see doctrine/dbal#3762 - it ads the missing support in SQLite platform when it gets merged.

PR #7325, opened 2018-07-26
Make code php 7.3 lint-compatible

paxal:
Current v2.6.2 raised a warning when ran on php 7.3 : 
```
$ php7.3 -v ; find . -name '*.php' -exec php7.3 -l {} \;| grep -v '^No syntax errors'
PHP 7.3.0alpha4 (cli) (built: Jul 23 2018 11:02:37) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.0-dev, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.3.0alpha4, Copyright (c) 1999-2018, by Zend Technologies
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in ./lib/Doctrine/ORM/UnitOfWork.php on line 2718
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in ./lib/Doctrine/ORM/UnitOfWork.php on line 2747
```
I changed the `continue` into `break` ([meaning of `continue;` inside a switch](https://secure.php.net/continue)), although `continue 2` seems to be equivalent as there is no code after the `switch` statement.

PR #7843, opened 2019-10-01
Ported PR #7325 (making code php 7.3 lint-compatible) to v2.5

stollr:
Although branch 2.5 is not maintained anymore I would like to port the PR #7325 to this branch, because it is the last one that supports PHP 7.0 which is still needed in one of our projects.

If you don't like to merge this PR I can understand. But I thought it's worth sharing it.


Ocramius:
`2.5` is closed for modification: please upgrade to `2.6`.

`7.0` is EOL, and you should prioritize upgrading ASAP.

Issue #7844, opened 2019-10-01
Remove PHP 7.4 from `allow_failures`


carusogabriel:
PHP 7.4 will be released [soon](https://wiki.php.net/todo/php74) and we need to ensure that `Doctrine` is compatible with it.

All we need to do is to remove this following line

https://github.com/doctrine/orm/blob/d528f7056886fa2ea114b3179946cc6a5e8d08ee/.travis.yml#L95

 and verify that everything is working.

This is a very good issue for those who want to participate in https://hacktoberfest.digitalocean.com.

carusogabriel:
This was already handled via https://github.com/doctrine/orm/pull/7785 for `v2.6.*`.

PR #7845, opened 2019-10-01
Update .travis.yml

Klizzy:
removed PHP-7.4 from .travis.yml allow_failures to make it compatible with PHP 7.4
See #7844


carusogabriel:
Please see https://github.com/doctrine/orm/issues/7844#issuecomment-537013195.

PR #7849, opened 2019-10-01
Mention SQL logger impact on batch processing

axi:
Looking for a way to improve one of our bulk update treatment, I went back to this page then found elsewhere that setting logger to null was a really effective way to improve time and memory consumption. Might be a right place to state it ? Don't know if my edit style is ok

### DIFF
--- a/docs/en/reference/batch-processing.rst
+++ b/docs/en/reference/batch-processing.rst
@@ -18,8 +18,8 @@ especially what the strategies presented here provide help with.
 
 .. note::
 
-    Logger can have a serious memory and time impact on bulk treatments. 
-    Deactivating it can be a solution.
+    Having an SQL logger enabled when processing batches can have a serious impact on performance and resource usage.
+    To avoid that you should disable it in the DBAL configuration:
 .. code-block:: php
 
     <?php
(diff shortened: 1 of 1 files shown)


docs/en/reference/batch-processing.rst
```diff
@@ -16,6 +16,15 @@ especially what the strategies presented here provide help with.
...
 
 
+.. note::
+
+    Logger can have a serious memory and time impact on bulk treatments. 
+    Deactivating it can be a solution.
+.. code-block:: php
+
+    <?php
+    $em->getConnection()->getConfiguration()->setSQLLogger(null);
```
lcobucci: The SQL logger isn't enabled by default in the ORM, so it might be worth mentioning that too. My suggestion would be something like this:

```
Having an SQL logger enabled when processing batches can have a serious impact on performance and resource usage.

To avoid that you should disable it in the DBAL configuration:

.. code-block:: php
    <?php
    $em->getConnection()->getConfiguration()->setSQLLogger(null);
```

lcobucci (changes requested):
@axi thanks for your contribution, what do you think about this?

docs/en/reference/batch-processing.rst
axi: I think it's perfect ;-)

lcobucci (approved):
@axi thanks for your contribution :+1:

axi:
I knew I would do something for doctrine someday ;-) Keep going !

Issue #7855, opened 2019-10-08
Numbers in a string using underscore naming strategy


vuras:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.4

#### Summary

If a string has numbers the naming strategy fails to convert the property name.

#### Current behavior

base64Encoded converts to base64encoded

#### How to reproduce

Add a property to entity that has a number in camelcased property name.

#### Expected behavior

base64Encoded converts to base64_encoded

Ocramius:
Possibly a bug in https://github.com/doctrine/inflector?

lcobucci:
Not really:

https://github.com/doctrine/orm/blob/a36809db72124000edb2b93171b330ac1ec09f7b/lib/Doctrine/ORM/Mapping/UnderscoreNamingStrategy.php#L136-L145

lcobucci:
Seems an easy pick, though

Ocramius:
Was totally convinced it would be using the inflector there :D 

@vuras feel free to send a patch (and test) for `UnderscoreNamingStrategy`

lcobucci:
Sorry, I got the patch already :grimacing:

lcobucci:
Handled by #7856, thanks for reporting it @vuras!

vuras:
Thank you! When should I expect the 2.6.5 released?

lcobucci:
I'd like to say "really soon" but realistically it should take us a few weeks since there're some bugs that should be included IMHO.

igoel:
> Sorry, I got the patch already 
> 
> There's just one problem, though: it affects table names like `DDC1476EntityWithDefaultFieldType` (which gets converted to `DDC1476_ENTITY_WITH_DEFAULT_FIELD_TYPE` instead of `DDC1476ENTITY_WITH_DEFAULT_FIELD_TYPE`).
> 
> Possible BC break, then, so I'll force it to only work for properties.

Even this, is breaking everything, so why do this in a minor version?

lcobucci:
@igoel @cobyl @sGy1980de we overlooked things and this is indeed a BC break. I'll rework it and send a new patch version `2.6.6` to address that.

I'm sorry for affecting your app and thank you for letting us know of this error.

lcobucci:
@igoel @cobyl @sGy1980de new release made to address the BC break. We apologise, again, for the inconvenience.

igoel:
Thx, for the fast response and action :+1:

PR #7856, opened 2019-10-08
Fix underscore naming strategy behaviour with numbers

lcobucci:
Fixes https://github.com/doctrine/orm/issues/7855


Ocramius (dismissed):
LGTM - thanks for having separate commits for the added tests and the formatting changes :+1:

cobyl:
:/ broke my code. understand that behaviour before was wrongly, but that very deeply affect existing codebase :/

Issue #7854, opened 2019-10-07
Hydration not working for Embeddables with php 7.4 typed properties


sspat:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.4

#### Summary

There is an error when using php 7.4 and trying to fetch an entity from the database that has typed properties containing value objects that are mapped as embeddables.

#### Current behavior

There is a fatal error.

```
Fatal error: Uncaught Error: Typed property App\Domain\Entity\City\City::$id must not be accessed before initialization in /var/www/vendor/doctrine/or
m/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php:89
Stack trace:
#0 /var/www/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php(89): ReflectionProperty->getValue(Object(App\Domain\Entity\Cit
y\City))
#1 /var/www/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php(2670): Doctrine\ORM\Mapping\ReflectionEmbeddedProperty->setValue(Object(App\Domain\Ent
ity\City\City), '29df4356-b832-4...')
#2 /var/www/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php(271): Doctrine\ORM\UnitOfWork->createEntity('App\\Domain\\Enti.
..', Array, Array)
#3 /var/www/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php(492): Doctrine\ORM\Internal\Hydration\ObjectHydrator->getEntity
(Array, 'c')
#4 /var/www/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php(162): Doctrine\ORM\Internal\Hydration\ObjectHyd in /var/www/ven
dor/doctrine/orm/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php on line 89
```

#### How to reproduce

```php
$this
            ->entityManager
            ->createQueryBuilder()
            ->select('c')
            ->from(City::class, 'c')
            ->where('c.id.id = :id')
            ->setParameter('id', $id)
            ->getQuery()
            ->getOneOrNullResult();
```

```php
<?php
namespace App\Domain\Entity\City;
use App\Domain\Entity\City\ValueObject\CityId;

class City
{
[...]

PR #7857, opened 2019-10-08
Fix typed property without default value support

sspat:
Added testcase for bug described in issue #7854

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -47,7 +47,6 @@ use InvalidArgumentException;
 use Throwable;
 use UnexpectedValueException;
 use function get_class;
-use function method_exists;
 
 /**
  * The UnitOfWork is responsible for tracking changes to objects during an
@@ -632,11 +631,7 @@ class UnitOfWork implements PropertyChangedListener
         $actualData = [];
 
         foreach ($class->reflFields as $name => $refProp) {
-            if (method_exists($refProp, 'isInitialized') && $refProp->getDeclaringClass()->isInstance($entity)) {
-                $value = $refProp->isInitialized($entity) ? $refProp->getValue($entity) : null;
-            } else {
-                $value = $refProp->getValue($entity);
-            }
+            $value = $refProp->getValue($entity);
 
             if ($class->isCollectionValuedAssociation($name) && $value !== null) {
                 if ($value instanceof PersistentCollection) {
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -47,6 +47,7 @@ use InvalidArgumentException;
 use Throwable;
 use UnexpectedValueException;
 use function get_class;
+use function method_exists;
 
 /**
  * The UnitOfWork is responsible for tracking changes to objects during an
@@ -631,7 +632,11 @@ class UnitOfWork implements PropertyChangedListener
         $actualData = [];
 
         foreach ($class->reflFields as $name => $refProp) {
-            $value = $refProp->getValue($entity);
+            if (method_exists($refProp, 'isInitialized') && $refProp->getDeclaringClass()->isInstance($entity)) {
+                $value = $refProp->isInitialized($entity) ? $refProp->getValue($entity) : null;
+            } else {
+                $value = $refProp->getValue($entity);
+            }
 
             if ($class->isCollectionValuedAssociation($name) && $value !== null) {
                 if ($value instanceof PersistentCollection) {
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,7 @@
     "autoload-dev": {
         "psr-4": {
             "Doctrine\\Tests\\": "tests/Doctrine/Tests",
+            "Doctrine\\Tests_PHP74\\": "tests/Doctrine/Tests_PHP74",
             "Doctrine\\Performance\\": "tests/Doctrine/Performance"
         }
     },
--- a/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php
+++ b/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php
@@ -21,7 +21,6 @@ namespace Doctrine\ORM\Mapping;
 
 use Doctrine\Instantiator\Instantiator;
 use ReflectionProperty;
-use function method_exists;
 
 /**
  * Acts as a proxy to a nested Property structure, making it look like
@@ -73,7 +72,7 @@ class ReflectionEmbeddedProperty extends ReflectionProperty
      */
     public function getValue($object = null)
     {
-        $embeddedObject = $this->getParentPropertyValue($object);
+        $embeddedObject = $this->parentProperty->getValue($object);
 
         if (null === $embeddedObject) {
             return null;
@@ -87,7 +86,7 @@ class ReflectionEmbeddedProperty extends ReflectionProperty
      */
     public function setValue($object, $value = null)
     {
-        $embeddedObject = $this->getParentPropertyValue($object);
+        $embeddedObject = $this->parentProperty->getValue($object);
 
         if (null === $embeddedObject) {
             $this->instantiator = $this->instantiator ?: new Instantiator();
@@ -98,19 +97,5 @@ class ReflectionEmbeddedProperty extends ReflectionProperty
         }
 
         $this->childProperty->setValue($embeddedObject, $value);
-    }
-
-    /**
-     * @param object $object
-     *
-     * @return mixed
-     */
-    private function getParentPropertyValue($object)
-    {
-        if (method_exists($this->parentProperty, 'isInitialized') && $this->parentProperty->getDeclaringClass()->isInstance($object)) {
-            return $this->parentProperty->isInitialized($object) ? $this->parentProperty->getValue($object) : null;
-        }
-
-        return $this->parentProperty->getValue($object);
     }
 }
--- a/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php
+++ b/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php
@@ -21,6 +21,7 @@ namespace Doctrine\ORM\Mapping;
 
 use Doctrine\Instantiator\Instantiator;
 use ReflectionProperty;
+use function method_exists;
 
 /**
  * Acts as a proxy to a nested Property structure, making it look like
@@ -72,7 +73,7 @@ class ReflectionEmbeddedProperty extends ReflectionProperty
      */
     public function getValue($object = null)
     {
-        $embeddedObject = $this->parentProperty->getValue($object);
+        $embeddedObject = $this->getParentPropertyValue($object);
 
         if (null === $embeddedObject) {
             return null;
@@ -86,7 +87,7 @@ class ReflectionEmbeddedProperty extends ReflectionProperty
      */
     public function setValue($object, $value = null)
     {
-        $embeddedObject = $this->parentProperty->getValue($object);
+        $embeddedObject = $this->getParentPropertyValue($object);
 
         if (null === $embeddedObject) {
             $this->instantiator = $this->instantiator ?: new Instantiator();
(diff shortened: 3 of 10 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH7854Test.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group GH7854
```
lcobucci: Please add [`@requires PHP >= 7.4`](https://phpunit.readthedocs.io/en/8.4/incomplete-and-skipped-tests.html#skipping-tests-using-requires) to make sure we don't run this test for other PHP versions.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7854Test.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+/**
+ * @Entity()
+ */
+class GH7854Entity
+{
+    /**
+     * @Embedded(class = "GH7854ValueObject")
```
lcobucci: This should use `::class` syntax instead

```suggestion
     * @Embedded(class=GH7854ValueObject::class)
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7854Test.php
lcobucci: @Ocramius do you have any suggestion on how to avoid syntax errors for PHP < 7.3 here? The only thing I can think of is moving the classes to an external file or using `eval()` :nauseated_face:

lcobucci (changes requested):
@sspat thanks for this test! A few things we need to consider apart from the fix :+1:

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7854Test.php
Ocramius: I wouldn't mind bumping the dependency :trollface:

sspat:
@lcobucci added suggested changes, also made my try on implementing the fix for this issue.
However, there are a lot of places elsewhere in the codebase where reflection is used and the same issue could arise. If my implementation is going in the right direction, I could investigate further.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7854Test.php
greg0ire: I just had a look at [the PHPUnit docs](https://phpunit.readthedocs.io/en/8.4/incomplete-and-skipped-tests.html#skipping-tests-using-requires)

> If you are using syntax that doesn’t compile with a certain PHP Version look into the xml configuration for version dependent includes in The <testsuites> Element

Following the link leads to reading about a way to add a directory to a test suite only when matching some constraints:

```
<testsuites>
  <testsuite name="unit">
    <directory phpVersion="8.0.0" phpVersionOperator=">=">tests/unit</directory>
  </testsuite>
</testsuites>
```

greg0ire:
Please associate your email address with your Github account, or change the email in your commits to an address already associated with it.

sspat:
> Please associate your email address with your Github account, or change the email in your commits to an address already associated with it.

Fixed, thank you for pointing this out!

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7854Test.php
greg0ire: I like the first solution better, only one directory to create.

SenseException:
I've restarted the Scrutinizer build. Checks are all green now.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7854Test.php
greg0ire: Wait the tests pass now, I thought there were an issue with PHP < 7.3 as @lcobucci mentioned. Why is it not the case?

greg0ire: Ah that's because the class was defined in the same file at first, isn't it?

sspat: Yes, then I moved it to a separate file.

greg0ire: So I guess we should now decide if we stay with the separate file or if we go with the solution I described above. @lcobucci , @Ocramius , any opinions?

lcobucci: Didn't know about that PHPUnit feature :scream: It's much better than having a separate file. @sspat sorry about the back and forth but could you change it? :grimacing:

sspat: @lcobucci no problem, done!

tests/Doctrine/PHP_7_4/Tests/ORM/Functional/Ticket/GH7854Test.php
```diff
@@ -0,0 +1,56 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\PHP_7_4\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group GH7854
+ * @requires PHP >= 7.4
```
greg0ire: Is this still needed?

sspat: Probably not, removed id

lcobucci:
@sspat I've just rebased your branch to sync things and rearrange the commits. My hope was to ship this in v2.7.0 but, as you said, we should cover other scenarios not only with embeddable but entity properties as well.

Let's aim for v2.7.1 :+1:

sspat:
> @sspat I've just rebased your branch to sync things and rearrange the commits. My hope was to ship this in v2.7.0 but, as you said, we should cover other scenarios not only with embeddable but entity properties as well.
> 
> Let's aim for v2.7.1 👍

Extended the existing test-case, now it contains:
- Typed embeddable value object
- Typed relation with other entity
- Typed entity propety

These all work without any issues.

Also did a search in `/lib/Doctrine/ORM/Mapping` namespace for all instances of `->getValue(`

Apart from `\Doctrine\ORM\Mapping\ReflectionEmbeddedProperty` which is fixed, it is used only in 
`\Doctrine\ORM\Mapping\ClassMetadataInfo::getFieldValue` and `\Doctrine\ORM\Mapping\ClassMetadataInfo::getIdentifierValues`  - both of which use `\Doctrine\Common\Reflection\RuntimePublicReflectionProperty` for reflection which has this problem with accessing uninitialized properties fixed too, but in another way.

Instead of calling `\ReflectionProperty::isInitialized` like I did in `\Doctrine\ORM\Mapping\ReflectionEmbeddedProperty` and `\Doctrine\ORM\UnitOfWork`:
```php
if (method_exists($this->parentProperty, 'isInitialized') && $this->parentProperty->getDeclaringClass()->isInstance($object)) {
    return $this->parentProperty->isInitialized($object) ? $this->parentProperty->getValue($object) : null;
}
```
It does the following:
```php
return isset($object->$name) ? parent::getValue($object) : null;
```
When checking with `isset` on a non-initialized typed property it returns false and no `Error: Typed property must not be accessed before initialization` error is thrown.

Should I commit this extended test?
ShouId I replace my previous solution with `isset`?
Should I continue investigating outside of the `Doctrine\ORM\Mapping` namespace?

beberlei:
:-1: from me because of the performance implications. Refleciton setValue/getvalue are in the high performance path. This should be fixed with inheritance by creating different instances for PHP 7.4 if the property is typed, otherwise keep old code.

beberlei (changes requested):

beberlei:
This looks like it could be complicated to fix :) It needs additional class in doctrine/reflection, a change to RuntimeReflectionService in doctrine/persistence so that getAccessibleProperty returns the correct reflection property.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -631,7 +632,11 @@ public function computeChangeSet(ClassMetadata $class, $entity)
         $actualData = [];
 
         foreach ($class->reflFields as $name => $refProp) {
-            $value = $refProp->getValue($entity);
+            if (method_exists($refProp, 'isInitialized') && $refProp->getDeclaringClass()->isInstance($entity)) {
```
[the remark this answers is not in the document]
sspat: @beberlei Thank you very much for the review, it gave me a lot of missing insight.

The current state of the test for this issue seems to be broken, as the issue does not arise for public properties, because for them an instance of `\Doctrine\Common\Reflection\RuntimePublicReflectionProperty` is created, which is checking the object property with `isset` prior to accessing it's value:
```php
return isset($object->$name) ? parent::getValue($object) : null;
```
So the error happens only for protected and private non-initialized properties without a default value.

I reverted my previous fix and updated the test accordingly, it is again failing.

If I understand correctly, we need to first update doctrine/reflection, then update the dependency in doctrine/persistence and implement the fix there, and then update the dependencies in doctrine/orm and the test will pass.

I created https://github.com/doctrine/reflection/pull/22 in doctrine/reflection to fix some PHP 7.4 issues that currently block me from running the tests, then I will proceed with the implementation.

semaex:
How is this issue? I have the same problem and don't know if it is already fixed. Thanxs

Metabor:
@semaex My Workaround ist to convert the property to allow null (even if it will never be after loading) and set default to null.

So like this
private ?\DateTime $date = null;

kuraobi:
@Metabor that wasn't really an option for us, because then we would have to add checks everywhere to ensure the value is not null, otherwise PHPStan throws errors. So for now we are using https://github.com/doctrine/orm/pull/7950 until an official fix is released. It's been working fine in production for 10 days now, although it may not cover all the use cases.

semaex:
> @sspat I've just rebased your branch to sync things and rearrange the commits. My hope was to ship this in v2.7.0 but, as you said, we should cover other scenarios not only with embeddable but entity properties as well.
> 
> Let's aim for v2.7.1 👍

Hi, when will be a 2.7.1 release? I'm so crazy with this bug. Thanxs!

beberlei:
No 2.7.1 needed here, updating doctrine/persistence to 1.3.6 will fix this issue.

greg0ire:
Should we close this PR?

sspat:
Yes, the issue is fixed by upgrading a dependency, no changes needed in the orm

cbichis:
Hi,

Looks like with 2.7.5 and doctrine/persistence at 1.3.8 this is still an issue.

beberlei:
@cbichis csn you be more specific please with exception message and full stack trace?

cbichis:
I am running Doctrine ORM 2.7.5, doctrine/persistence 1.3.8, doctrine/reflextion at 1.2.2.

$entity = new Supplier();
$this->form->bind($entity);
triggering:

File:
    /project/module/Application/src/Entity/Traits/IdTrait.php:18
Message:
    Typed property Supplier::$id must not be accessed before initialization
Stack trace:
    #0 /project/vendor/doctrine/doctrine-module/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php(189): Supplier->getId()
    #1 /project/vendor/doctrine/doctrine-module/src/DoctrineModule/Stdlib/Hydrator/DoctrineObject.php(87): DoctrineModule\Stdlib\Hydrator\DoctrineObject->extractByValue()
    #2 /project/vendor/laminas/laminas-form/src/Fieldset.php(659): DoctrineModule\Stdlib\Hydrator\DoctrineObject->extract()
    #3 /project/vendor/laminas/laminas-form/src/Form.php(979): Laminas\Form\Fieldset->extract()
    #4 /project/vendor/laminas/laminas-form/src/Form.php(314): Laminas\Form\Form->extract()
    #5 /project/module/Application/src/Service/DoctrineCrudTrait.php(163): Laminas\Form\Form->bind()
And the entity has:
/**
 * @ORM\Id
 * @ORM\Column(name="id", type="integer", nullable=false, options={"unsigned"=true})
 * @Annotation\Exclude()
 */
protected int $id;

beberlei:
This is a bug in Doctrine Laminas Module their Hydrator support, please open an issue on their repo with the code and stacktrace from your comment.

cbichis:
Thanks. Even after updating DoctrineModule/DoctrineORMModule the error stays so yes... I will open issue.

PR #7861, opened 2019-10-09
Delete statements will not be created using `clear` 

ferrastas:
This is just a demo of an opened issue.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -46,6 +46,7 @@ use Doctrine\ORM\Utility\IdentifierFlattener;
 use InvalidArgumentException;
 use Throwable;
 use UnexpectedValueException;
+use function get_class;
 
 /**
  * The UnitOfWork is responsible for tracking changes to objects during an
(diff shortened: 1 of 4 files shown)


lcobucci (dismissed):
Please provide a new test class with its own set of entities instead of modifying an existing one.

ferrastas:
Yep, looks like a missing test case of the original fix :P

ostrolucky:
I've pushed commit with better solution to fix #7758, without this regression

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7761Test.php
```diff
@@ -31,6 +31,15 @@ protected function setUp() : void
         $this->_em->clear();
     }
 
+    protected function tearDown()
+    {
+        $this->_em->clear();
```
lcobucci: This is already handled in the parent, so not needed :+1:

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -380,7 +380,12 @@ public function commit($entity = null)
         try {
             // Collection deletions (deletions of complete collections)
             foreach ($this->collectionDeletions as $collectionToDelete) {
-                $this->getCollectionPersister($collectionToDelete->getMapping())->delete($collectionToDelete);
+                if (!$collectionToDelete instanceof PersistentCollection ||
+                    // deferred explicit tracked collections can be removed only when owning relation was persisted
+                    $this->em->getClassMetadata(get_class($owner = $collectionToDelete->getOwner()))->isChangeTrackingDeferredImplicit() || 
+                    $this->isScheduledForDirtyCheck($owner)) {
+                    $this->getCollectionPersister($collectionToDelete->getMapping())->delete($collectionToDelete);
+                }
```
lcobucci: I think there's a bit too much happening here. We can split things, do an early exit for non-persistent collections, and remove the assignment in condition.

lcobucci (changes requested):
@ostrolucky thanks a lot for handling it. It's indeed a much better solution. I'll address these two tiny things and we're good to ship this.

lib/Doctrine/ORM/UnitOfWork.php
ostrolucky: We can't do early exit, statement must be executed for non-persistent collections. And assignment needs to happen for persistent collections only.

lcobucci: `continue` is an early-exit for the loop :)
I just pushed the changes :+1:

lcobucci (approved):

lib/Doctrine/ORM/UnitOfWork.php
ostrolucky: Yeah I know. I meant we can't do it without duplicating code. If that's fine, then ok

lcobucci:
@ferrastas @ostrolucky thank you very much :+1:

ferrastas:
Thanks to fix it @ostrolucky 🏆

Issue #7829, opened 2019-09-24
Skip Paginator LIMIT subquery and WHERE IN if query do not have LIMIT


Seb33300:
I ran into an issue with the `Paginator` which I think could be handled to prevent it.

I am using the `Paginator` to paginate a table.
The user can choose the number of items by page (10 ,20, ... **but also _All items_**).

In the case the user choose to display `All items`, the `LIMIT` clause will NOT be added to the query.
Even if the `Paginator` is not required in this scenario, the same part of code handle this case so we also give the query to the `Paginator`.

This is working property unless the query return too many results.
Because the `Paginator` will use a `WHERE IN` query and it could reach the limit of allowed number of parameters by the database engine.
See: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/tutorials/pagination.html

This is what we are encountering on tables with thousands of items on SQL Server.

lcobucci:
@Seb33300 I'd say that handle this case isn't an ORM's responsibility. The paginator component is quite generic and having a flag to enable its pagination behaviour or not smells like a bad design for the library.

I'd suggest you to encapsulate this in your application (perhaps decorating the paginator?).

TomHAnderson:
It sounds to me like you already know if you don't need a limit clause.  So, if you're not using a limit don't use a Paginator.

Seb33300:
I know this is useless to use the `Paginator` feature if we are not intended to paginate results.
But I felt it more like an optimization, since this could prevent to execute useless queries.

Btw even if we do not use it to paginate results, it could also be used to count distinct entries.

@lcobucci not sure what you mean by a flag, but it just require to check if a limit has been added to the query. No extra parameter or something else required.

Seb33300:
Yes, this is exactly what I mean.

Not sure about the example you want.

An example of use case or a pull request to implement it?

lcobucci:
> @lcobucci not sure what you mean by a flag, but it just require to check if a limit has been added to the query. No extra parameter or something else required.

@Seb33300 setting `setMaxResults()` to `null` is kind of a flag :smile:.

> Not sure about the example you want.
> 
> An example of use case or a pull request to implement it?

A PR with (at least) a **functional test** showing what you expect from the library. 

You can use this test as example: 
https://github.com/doctrine/orm/blob/b52ef5a1002f99ab506a5a2d6dba5a2c236c5f43/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php

[...]

PR #7863, opened 2019-10-10
Paginator: Skip limit subquery if not required

Seb33300:
This fixes https://github.com/doctrine/orm/issues/7829

### DIFF
--- a/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
@@ -139,7 +139,7 @@ class Paginator implements \Countable, \IteratorAggregate
         $offset = $this->query->getFirstResult();
         $length = $this->query->getMaxResults();
 
-        if ($this->fetchJoinCollection) {
+        if ($this->fetchJoinCollection && $length !== null) {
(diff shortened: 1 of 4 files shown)


SenseException:
@lcobucci Are these `$query->setMaxResults(1);` in the older tests not an indicator for a breaking change that needs to target master?

Seb33300:
I feel more like I fixed issues in older tests.

They are both related to the `Paginator` and tests was failing because they was expecting the Limit Subquery to be applied, but without adding a LIMIT to the query...
And an empty limit will return all data, so this is why my PR is intended to skip the limit query in this case.

The only BC that could happen is if you use the SQLLogger to count how many queries have been executed when using a Paginator without LIMIT.

lcobucci:
@Seb33300 there's a small change which is not BC compatible here as highlighted by @SenseException: no length + usage of output walkers.

I'm working on solving this now because I think we can still put it in 2.7.0.

lcobucci (approved):

lcobucci:
@Seb33300 :ship: 

I marked this with `BC Break` because behaviour might change a tiny bit. I just forgot to document it in `UPGRADE.me`, will do it separately. Thanks for your contribution.

PR #7865, opened 2019-10-10
#7837 paginate with custom identifier types even with enabled DQL query cache

### LINKED ISSUE #7837: Paginator breaks when DQL query is re-used with different identifier parameters that require conversion
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.4

#### Summary

#7820 and #7821 fix pagination issues happening when the paginator is relying on identifiers that require custom DBAL type conversion.

The type conversion works, but since it has been performed inside the `WhereInWalker` (to keep things a bit simpler), it is only performed when the query is in `Doctrine\ORM\Query::STATE_DIRTY`.

If the parameters are changed on a `Doctrine\ORM\Query`, the query doesn't change parser state, and therefore the `WhereInWalker` is completely skipped.

I will try to write a small reproducer and fix ASAP, hoping to not cause too big performance regressions in the ORM. I think it may come down to forcing query parsing to be repeated.

#### Current behavior

```php
$id = generateCustomId();
$a  = new A($id);

$this->em->persist($a);
$this->em->flush();

$query = $this->em->createQuery('SELECT a FROM ' . A::class);

$query->setFirstResult(1);

Assert::count(0, \iterator_to_array($paginator));

$query->setFirstResult(0);

Assert::same([$a], \iterator_to_array($paginator));
```

#### Expected behavior

The last assertion in the test above fails: this is an unfortunate case of mutable state shared inside `Doctrine\ORM\Query`.

Ocramius:
This patch enforces re-parsing of DQL queries used in the paginator internals, so that the `WhereInWalker` can correctly process query parameters after evaluating the AST structure (root expression, specifically).

While this isn't nice for performance, it is also true that the alternative approach would be to run the parsing manually on the DQL query (to extract the root of the selection) and then operating on that (this was already attempted in #7821 in doctrine/orm@39d21135490472ba31755a0259fbaf38ac453b9b).

Fixes #7837

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
@@ -98,7 +98,7 @@ class GH7820Test extends OrmFunctionalTestCase
             array_map(static function (GH7820Line $line) : string {
                 return $line->toString();
             }, iterator_to_array(new Paginator($query))),
-            'Paginator runs once, query cache is populated with DQL -> SQL translation'
+            'Expected to return expected data before query cache is populated with DQL -> SQL translation. Were SQL parameters translated?'
         );
 
         $query = $this->_em->getRepository(GH7820Line::class)
@@ -110,7 +110,7 @@ class GH7820Test extends OrmFunctionalTestCase
             array_map(static function (GH7820Line $line) : string {
                 return $line->toString();
             }, iterator_to_array(new Paginator($query))),
-            'Paginator runs again, SQL parameters are translated again, even with cached DQL -> SQL translation'
+            'Expected to return expected data even when DQL -> SQL translation is present in cache. Were SQL parameters translated again?'
         );
     }
 }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\ORM\Functional\Ticket;
 
+use Doctrine\Common\Cache\ClearableCache;
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\StringType;
@@ -11,6 +12,7 @@ use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Tools\Pagination\Paginator;
 use Doctrine\Tests\OrmFunctionalTestCase;
 use function array_map;
+use function assert;
 use function is_string;
 use function iterator_to_array;
 
@@ -53,6 +55,9 @@ class GH7820Test extends OrmFunctionalTestCase
 
         $this->setUpEntitySchema([GH7820Line::class]);
 
+        $this->_em->createQuery('DELETE FROM ' . GH7820Line::class . ' l')
+            ->execute();
+
         foreach (self::SONG as $index => $line) {
             $this->_em->persist(new GH7820Line(GH7820LineText::fromText($line), $index));
         }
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
```diff
@@ -73,6 +78,41 @@ public function testWillFindSongsInPaginator() : void
...
+
+    /** @group GH7837 */
+    public function testWillFindSongsInPaginatorEvenWithCachedQueryParsing() : void
+    {
+        $cache = $this->_em->getConfiguration()
+            ->getQueryCacheImpl();
+
+        assert($cache instanceof ClearableCache);
+
+        $cache->deleteAll();
```
ostrolucky: this is not needed

Ocramius: Sadly needed, otherwise the two tests interfere with each other

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
```diff
@@ -73,6 +78,41 @@ public function testWillFindSongsInPaginator() : void
...
+        $query = $this->_em->getRepository(GH7820Line::class)
+            ->createQueryBuilder('l')
+            ->orderBy('l.lineNumber', Criteria::ASC);
+
+        self::assertSame(
+            self::SONG,
+            array_map(static function (GH7820Line $line) : string {
+                return $line->toString();
+            }, iterator_to_array(new Paginator($query))),
+            'Paginator runs once, query cache is populated with DQL -> SQL translation'
```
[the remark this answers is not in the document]
Ocramius: Had these as comments initially: moved to assertion message on purpose, since it makes failures much more descriptive

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
ostrolucky: which ones? Running whole GH7820Test class runs fine for me without this statement

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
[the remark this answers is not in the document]
ostrolucky: But this says what happens, not what should have happened

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
```diff
@@ -73,6 +78,41 @@ public function testWillFindSongsInPaginator() : void
...
         );
     }
+
+    /** @group GH7837 */
+    public function testWillFindSongsInPaginatorEvenWithCachedQueryParsing() : void
+    {
+        $cache = $this->_em->getConfiguration()
+            ->getQueryCacheImpl();
+
+        assert($cache instanceof ClearableCache);
```
Pierstoval: Why not a PHPUnit assertion? 🤔

SenseException: It is not part of the test.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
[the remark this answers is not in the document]
SenseException: I agree. Messages like "null is given, 1 was expected" are the usual messages that are shown on failing assertions.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
```diff
@@ -73,6 +78,41 @@ public function testWillFindSongsInPaginator() : void
...
+        $query = $this->_em->getRepository(GH7820Line::class)
+            ->createQueryBuilder('l')
+            ->orderBy('l.lineNumber', Criteria::ASC);
+
+        self::assertSame(
+            self::SONG,
+            array_map(static function (GH7820Line $line) : string {
+                return $line->toString();
+            }, iterator_to_array(new Paginator($query))),
+            'Paginator runs again, SQL parameters are translated again, even with cached DQL -> SQL translation'
```
SenseException: Same as mentioned for the assert-message above.

SenseException (dismissed):

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
ostrolucky: After another look, I can confirm this is neeed. Otherwise removing the fix would cause failure on first assertion

ostrolucky:
I have taken liberty to improve these messages. Let me know if it's ok now

lcobucci (approved):

PR #7869, opened 2019-10-15
UnitOfWork::clear() misses $eagerLoadingEntities

BenMorel:
AFAICS, eager loading entities should be cleared as well when the UnitOfWork is cleared.

Actually, this is [already the case](https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/UnitOfWork.php#L1982) on `master`!

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7869Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7869Test.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Decorator\EntityManagerDecorator;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\UnitOfWork;
+use Doctrine\Tests\Mocks\ConnectionMock;
+use Doctrine\Tests\Mocks\DriverMock;
+use Doctrine\Tests\Mocks\EntityManagerMock;
+use Doctrine\Tests\OrmTestCase;
+
+/**
+ * @group GH7869
+ */
+class GH7869Test extends OrmTestCase
+{
+    public function testDQLDeferredEagerLoad()
+    {
+        $decoratedEm = EntityManagerMock::create(new ConnectionMock([], new DriverMock()));
+
+        $em = $this->getMockBuilder(EntityManagerDecorator::class)
+            ->setConstructorArgs([$decoratedEm])
+            ->setMethods(['getClassMetadata'])
+            ->getMock();
+
+        $em->expects($this->exactly(2))
+            ->method('getClassMetadata')
+            ->willReturnCallback([$decoratedEm, 'getClassMetadata']);
+
+        $hints = [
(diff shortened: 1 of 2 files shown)


greg0ire:
`phpcs` has a weird complaint I don't understand… can you try running `vendor/bin/phpcbf`?

BenMorel:
I'm not sure why this happens either... Locally, `phpcs` reports, and `phpcbf` fixes, thousands (!) of CS issues.

I'm not familiar with `phpcs`, but I suspect from the command:

    php git-phpcs.phar origin/$TRAVIS_BRANCH...$TRAVIS_PULL_REQUEST_SHA

that it only reports issues on the line(s) modified by the PR. And due to the surrounding lines, there is not much I can do to fix this.

greg0ire:
Maybe you could try aligning all equal signs, including in the surrounding lines?

BenMorel:
Let's try this!

greg0ire:
Hey! That actually worked! :)

greg0ire:
Original commit: d070fc62a76cfe7b9ae35c3236335ab888762833

greg0ire (dismissed):

greg0ire:
The commit above contains a similar change for `$persisters`, should it be applied too?

BenMorel:
I have unfortunately no idea. For sure anything that contains a reference to an entity must be cleared, hence this PR, for the rest I'm not the right person to ask ;)

greg0ire:
Let's wait for feedback from contributors that know the UOW better than I do.

lcobucci (changes requested):
@BenMorel thanks for the patch. The fix is 100% valid but we need a test to be able to get it merged :+1:

BenMorel:
@lcobucci Adding a test does not look like a trivial task, as there is no setter/getter for `$eagerLoadingEntities`.

Suggestions:

- test using reflection 🤢
- side-test using `triggerEagerLoads()`; this would require a complex setting, where you have to somehow set up an eager loading entity, and check that its `EntityPersister` is not invoked when calling `triggerEagerLoads()` after `clear()`; *really* not trivial, in comparison to the fix!

SenseException:
Suggestion: You could use an EntityManager in a functional test to setup eagerloading entities and test the clear-method of the UOW.

ostrolucky:
Test case added

BenMorel:
@ostrolucky Thank you!

greg0ire (approved):

guilhermeblanco (approved):
This was an issue I found when i was working on ephemeral unit of work support. It also helps on freeing resources as eager loaded entities hang there forever. Patch on its own is great.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7869Test.php
```diff
@@ -0,0 +1,67 @@
...
+use Doctrine\Tests\Mocks\DriverMock;
+use Doctrine\Tests\Mocks\EntityManagerMock;
+use Doctrine\Tests\OrmTestCase;
+
+/**
+ * @group GH7869
+ */
+class GH7869Test extends OrmTestCase
+{
+    public function testDQLDeferredEagerLoad()
```
ostrolucky: I forgot to rename test case 😞 Let me know if I Should create new PR with just this change

PR #7872, opened 2019-10-17
#7498 - Drop legacy PEAR stuff

### LINKED ISSUE #7498: Drop legacy PEAR stuff
PEAR distribution is deprecated since ORM 2.4 and our pear channel on doctrine-project.org doesn't work anymore anyway. It should be dropped:
- any occurences in docs,
- support in `bin/doctrine.bat`,
- `bin/doctrine-pear.php` altogether.

ztroudi:
Fix : https://github.com/doctrine/orm/issues/7498

### DIFF
--- a/bin/doctrine.bat
+++ b/bin/doctrine.bat
@@ -3,7 +3,5 @@
 if "%PHPBIN%" == "" set PHPBIN=@php_bin@
 if not exist "%PHPBIN%" if "%PHP_PEAR_PHP_BIN%" neq "" goto USE_PEAR_PATH
 GOTO RUN
-:USE_PEAR_PATH
-set PHPBIN=%PHP_PEAR_PHP_BIN%
 :RUN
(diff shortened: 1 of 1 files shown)


greg0ire:
This one is a work in progress, will be tested under Windows. Some questions arose, like how are people to run the doctrine binary under Windows? Should they use the `bat` file or should they use `bin/doctrine` directly, in which case, the bat file should be removed entirely? Can anyone give some context about this bat file? Also, we found a https://github.com/doctrine/orm/blob/master/bin/doctrine-pear.php file, should be removed too?

beberlei:
@greg0ire many good questions I also have some more ;)

- Doesn't ocmposer automatically generate bat files now on Windows? Do we even need this file at all anymore?

I would target 2.8.x for this cleanup not master

derrabus:
Closing in favor of #9661 and #9662.

PR #7873, opened 2019-10-18
fix typo

akshay-kh:
* fix typo in getting-started doc.


greg0ire:
This is not a typo, it's the way [the english possessive](https://en.wikipedia.org/wiki/English_possessive) works:

> In the case of plural nouns ending in -s, the possessive is spelled by only adding an apostrophe and is pronounced the same (for example: Peasants' Revolt)

Issue #7717, opened 2019-05-22
isNull criteria doesn't work


BenWaNH:
Hi,

i've an issue with isNull method from Expression class

this is an example i use in my entity :
```
public function getValidDiscountOffers(): Collection
{

        $expr = Criteria::expr();
        $criteria = Criteria::create()
            ->where($expr->eq('active', true))
            ->andWhere($expr->isNull('validityStart'))
        ;
        return $this->discountOffers->matching($criteria);
}
```

And this is the result SQL from log :

> WHERE t.user_id = ? AND te.active = ? [145,false] []

No trace about IS NULL
And If i try to replace my validityStart variable by another whose doesn't exist, no error returned
But if i replace isNull method by eq method an error is trigger ... It seems to skip my expression when isNull...

Thanks for reading
BenWa

BenWaNH:
Yes, I will write it next week. I have to stay my time for this week and I will use the filter in the meantime.
Thanks

thicolares:
Hey @BenWaNH , I've created a test case for that, but I was not able to reproduce the error (version 2.7). Looks good to me. I took the opportunity to create a functional test that demonstrates that. See #7874.

If so, maybe we could close this issue.

hason:
@thicolares This issue is valid for ManyToManyPersister and SqlValueVisitor https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Persisters/SqlValueVisitor.php#L32-L36

ndench:
I can confirm that this issue exists for ManyToMany relationships but not for ManyToOne relationships.
Our entity model looks like this:

```
Organisation:
    - OneToMany TemplateTypes
Template:
    - ManyToMany TemplateTypes
TemplateType:
    - ManyToOne Organisation
    - ManyToMany Template
    - ?DateTime archivedAt
```

Doing a `Criteria::expr()->isNull('archivedAt')` on `Organisation::templateTypes` correctly filters out archived TemplateTypes.

Doing a `Criteria::expr()->isNull('archivedAt')` on `Template::templateTypes` when `templateTypes` is uninitialized does not filter correctly because `archived_at IS NULL` is left off the SQL query.

However, if `templateTypes` is initialized, then Criteria uses `array_filter` and it works as expected.

So the issue is only present when calling `PersistentCollection::matching` with an uninitialized collection with an many to many association.

ndench:
For anyone else who stumbles across this, I've been manually filtering these as a workaround:

```diff
[...]

PR #7874, opened 2019-10-20
#7717 Test Criteria isNull query

thicolares:
I wanted to demonstrate that I could not reproduce #7717 and it looks good.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/EntityRepositoryCriteriaTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/EntityRepositoryCriteriaTest.php
@@ -105,6 +105,21 @@ class EntityRepositoryCriteriaTest extends OrmFunctionalTestCase
         $this->assertEquals(1, count($dates));
     }
 
+    public function testIsNullComparisonQuery()
+    {
+        $this->loadNullFieldFixtures();
+        $repository = $this->_em->getRepository(DateTimeModel::class);
+
+        $repository->matching(new Criteria(
+            Criteria::expr()->isNull('time')
+        ))->first();
+
+        $this->assertContains(
+            'col_time IS NULL',
+            $this->_sqlLoggerStack->queries[$this->_sqlLoggerStack->currentQuery]['sql']
+        );
+    }
+
     public function testEqNullComparison()
     {
         $this->loadNullFieldFixtures();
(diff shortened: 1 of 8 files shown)


greg0ire:
Read [this slide](https://www.greg0ire.fr/git-gud/#/23) to learn how to avoid committing a .DS_Store file for life.

thicolares:
omg. I've looked at the commit and haven't noticed. I'd set other things to my global gitignore (like .idea etc), but I think I hadn't problem with .DS_store in the past. Well, I'll add. Thank you!

thicolares:
Fixed for good and rebased :)

greg0ire (dismissed):

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryCriteriaTest.php
```diff
@@ -105,6 +105,21 @@ public function testIsNullComparison()
...
+        $this->loadNullFieldFixtures();
+        $repository = $this->_em->getRepository(DateTimeModel::class);
+
+        $repository->matching(new Criteria(
+            Criteria::expr()->isNull('time')
+        ))->first();
+
+        $this->assertContains(
+            'col_time IS NULL',
+            $this->_sqlLoggerStack->queries[$this->_sqlLoggerStack->currentQuery]['sql']
```
SenseException: Checking for a part of a string in an SQL query is kind of a foggy test compared to the other tests in this class. I would prefer to work and compare with a result.

SenseException (changes requested):

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryCriteriaTest.php
thicolares: I see your point. But my intention was to demonstrate the query is created as expected given this criteria (due to #7717). What do you think?

SenseException: Despite that the query part was created I like the test to show that `isNull` is also returning the expected result when the query gets executed. Checking the SQL itself isn't the same as having the expected result of the query. This way regressions in the functionality can be covered too for any other unwanted changes that might be introduced besides `'col_time IS NULL'`.

PR #7875, opened 2019-10-21
Whilelist existing assets we know about from metadata in SchemaTool::getUpdateSchemaSql()

nicolas-grekas:
This PR is a sidekick of https://github.com/doctrine/DoctrineBundle/pull/1037

When generating schema diffs, if a source table is filtered out by a `schema_filter` expression, then a `CREATE TABLE` is always generated, even if the table already exists.

This PR makes `SchemaTool::getUpdateSchemaSql()` whitelist tables that are listed in the metadata we are generating the diff for, thus fixing the issue.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -10,6 +10,12 @@ Method `Doctrine\ORM\AbstractQuery#useResultCache()` which could be used for bot
 To optimize DB interaction, `Doctrine\ORM\Tools\Pagination\Paginator` no longer fetches identifiers to be able to
 perform the pagination with join collections when max results isn't set in the query.
 
+## Minor BC BREAK: tables filtered with `schema_filter` are no longer created
+
+When generating schema diffs, if a source table is filtered out by a `schema_filter` expression, then a `CREATE TABLE` was
+always generated, even if the table already existed. This has been changed in this release and the table will no longer
+be created.
+
 ## Deprecated number unaware `Doctrine\ORM\Mapping\UnderscoreNamingStrategy`
 
 In the last patch of the `v2.6.x` series, we fixed a bug that was not converting names properly when they had numbers
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -19,7 +19,7 @@
 
 namespace Doctrine\ORM\Tools;
 
-use Doctrine\ORM\ORMException;
+use Doctrine\DBAL\Schema\AbstractAsset;
 use Doctrine\DBAL\Schema\Comparator;
 use Doctrine\DBAL\Schema\Index;
 use Doctrine\DBAL\Schema\Schema;
@@ -28,6 +28,7 @@ use Doctrine\DBAL\Schema\Visitor\DropSchemaSqlCollector;
 use Doctrine\DBAL\Schema\Visitor\RemoveNamespacedAssets;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\ORMException;
 use Doctrine\ORM\Tools\Event\GenerateSchemaTableEventArgs;
 use Doctrine\ORM\Tools\Event\GenerateSchemaEventArgs;
 
@@ -891,10 +892,8 @@ class SchemaTool
      */
     public function getUpdateSchemaSql(array $classes, $saveMode = false)
     {
-        $sm = $this->em->getConnection()->getSchemaManager();
-
-        $fromSchema = $sm->createSchema();
         $toSchema = $this->getSchemaFromMetadata($classes);
+        $fromSchema = $this->createSchemaForComparison($toSchema);
 
         $comparator = new Comparator();
         $schemaDiff = $comparator->compare($fromSchema, $toSchema);
@@ -904,5 +903,36 @@ class SchemaTool
         }
 
         return $schemaDiff->toSql($this->platform);
+    }
+
+    /**
+     * Creates the schema from the database, ensuring tables from the target schema are whitelisted for comparison.
+     */
+    private function createSchemaForComparison(Schema $toSchema) : Schema
+    {
+        $connection    = $this->em->getConnection();
+        $schemaManager = $connection->getSchemaManager();
+
+        // backup schema assets filter
+        $config         = $connection->getConfiguration();
+        $previousFilter = $config->getSchemaAssetsFilter();
+
+        if ($previousFilter === null) {
+            return $schemaManager->createSchema();
+        }
+
+        // whitelist assets we already know about in $toSchema, use the existing filter otherwise
+        $config->setSchemaAssetsFilter(static function ($asset) use ($previousFilter, $toSchema) : bool {
+            $assetName = $asset instanceof AbstractAsset ? $asset->getName() : $asset;
+
+            return $toSchema->hasTable($assetName) || $toSchema->hasSequence($assetName) || $previousFilter($asset);
+        });
+
+        try {
+            return $schemaManager->createSchema();
+        } finally {
+            // restore schema assets filter
+            $config->setSchemaAssetsFilter($previousFilter);
+        }
     }
 }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7875Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7875Test.php
@@ -1,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\DBAL\Configuration;
+use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
+use Doctrine\ORM\Tools\SchemaTool;
+use Doctrine\Tests\OrmFunctionalTestCase;
+use function array_filter;
+use function current;
+use function method_exists;
+use function sprintf;
+use function strpos;
+
+/** @group GH7875 */
+final class GH7875Test extends OrmFunctionalTestCase
+{
+    /** @after */
+    public function cleanUpSchema() : void
+    {
+        $connection = $this->_em->getConnection();
+
+        $connection->exec('DROP TABLE IF EXISTS gh7875_my_entity');
+        $connection->exec('DROP TABLE IF EXISTS gh7875_my_other_entity');
+
+        if ($connection->getDatabasePlatform() instanceof PostgreSqlPlatform) {
+            $connection->exec('DROP SEQUENCE IF EXISTS gh7875_my_entity_id_seq');
+            $connection->exec('DROP SEQUENCE IF EXISTS gh7875_my_other_entity_id_seq');
+        }
+    }
+
+    /**
+     * @param string[] $sqls
+     *
+     * @return string[]
+     */
+    private function filterCreateTable(array $sqls, string $tableName) : array
+    {
+        return array_filter($sqls, static function (string $sql) use ($tableName) : bool {
+            return strpos($sql, sprintf('CREATE TABLE %s (', $tableName)) === 0;
+        });
+    }
+
+    public function testUpdateSchemaSql() : void
+    {
+        $classes = [$this->_em->getClassMetadata(GH7875MyEntity::class)];
+
+        $tool = new SchemaTool($this->_em);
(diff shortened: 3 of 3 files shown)


jwage:
I understand the problem and how it manifests itself to the end user. I've been running in to this in Symfony with the messenger tables. Kudos to you for taking a stab at fixing it. I always felt like the filter functionality was sort of half baked.

Can you think of any better way we could implement this to eliminate the nested closures?

nicolas-grekas:
Passing `$toFilter` to `createSchema()`could be made to produce the desired behavior, or a new method dedicated to the need.

But that's an API change so it's out of reach for a big fix on the ORM side. ORM 2.7 could bump to dbal 2.10 and clean the code afterwards.

nicolas-grekas:
Now with tests, courtesy of @l-vo, big thank you to him!

alcaeus:
Looks good to me. As for the nested closures, there isn't much we can do since DBAL only takes a single closure as a filter. If you want to expand on that filter later on, there's no other option than to wrap it. In one of our bundles we have a chain to take care of that, but DBAL itself does not provide any helpers for this purpose.

Introducing a chain filter class would require this to go into 2.7 as it's a new API, so I'm ok with keeping the nested closures and treating this as a (long-awaited) bug fix in 2.6.

alcaeus (dismissed):

Ocramius (dismissed):
LGTM - discussed directly with @nicolas-grekas, and the use-case seems clear to me now :+1:

Ocramius:
What I'm still puzzled on is BC compliance of this patch: we're now allowing the schema diffing to kick in for known symbols, which may break existing use-cases (where mappings were kept only as documentation, never really used by the ORM).

@doctrine/doctrinecore ideas?

jwage:
It is definitely a "BC Break" but I am not sure in what ways it would manifest itself in desirable ways that someone would depend on, but who knows. In the problem Symfony team is trying to fix, the resulting SQL we produce just has to be manually edited by the developer before reviewing and executing to remove the tables that shouldn't be there.

jwage (dismissed):

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -910,4 +911,55 @@ public function getUpdateSchemaSql(array $classes, $saveMode = false)
...
+            return $sm->createSchema();
+        } finally {
+            if ($hasFilterMethod && $filter !== null) {
+                // restore schema assets filter and filter-expression
+                if ($filterExpression !== null) {
+                    $config->setFilterSchemaAssetsExpression($filterExpression);
+                } else {
+                    $config->setSchemaAssetsFilter($filter);
+                }
+            }
```
lcobucci: We released the last version of the `2.6.x` series to be able to better focus on the `2.7.x` series, which means that we're making `2.7.x` the target of this PR.

I was actually hoping to release this on `2.7.0` but the complexity of this method makes me hesitate. The DBAL 2.9+ check isn't needed since ORM 2.7 requires DBAL 2.9+, which means we can clean this up.

tests/Doctrine/Tests/ORM/Functional/SchemaTool/MySqlSchemaToolTest.php
```diff
@@ -89,6 +105,89 @@ public function testGetCreateSchemaSql4()
         $this->assertEquals(0, count($sql));
     }
 
+    public function testUpdateSchemaSql()
```
lcobucci: ` : void` missing

tests/Doctrine/Tests/ORM/Functional/SchemaTool/MySqlSchemaToolTest.php
```diff
@@ -89,6 +105,89 @@ public function testGetCreateSchemaSql4()
...
+    {
+        $classes = [
+            $this->_em->getClassMetadata(MyEntityToRemove::class),
+        ];
+        $tool    = new SchemaTool($this->_em);
+        $sqls    = $tool->getUpdateSchemaSql($classes);
+        $sqls    = $this->filterSqls($sqls, ['entity_to_remove']);
+        $this->assertCount(1, $sqls);
```
lcobucci: `self::assertCount()` since the method is static (other places should be fixed as well)

tests/Doctrine/Tests/ORM/Functional/SchemaTool/MySqlSchemaToolTest.php
```diff
@@ -89,6 +105,89 @@ public function testGetCreateSchemaSql4()
...
+        $this->assertContains('CREATE TABLE entity_to_remove (id INT AUTO_INCREMENT NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB', $sqls);
```
lcobucci: `self::assertContains()` since the method is static (other places should be fixed as well)

lcobucci: I'm afraid this will also fail on DBAL 2.10+ due to quoted collation

lcobucci (changes requested):
Just some small changes, we should be good to go soon.

lcobucci:
I'll plan this for `v2.7.1` so we don't miss this.

nicolas-grekas:
Comments addressed @lcobucci 
I'll let you deal with the UPGRADE notes if you don't mind.

alcaeus:
@lcobucci 2.7.1 sounds good, as does mentioning this in UPGRADE.md. I've pushed a commit to do this. Thanks @nicolas-grekas!

alcaeus (dismissed):

UPGRADE.md
```diff
@@ -10,6 +10,12 @@ Method `Doctrine\ORM\AbstractQuery#useResultCache()` which could be used for bot
 To optimize DB interaction, `Doctrine\ORM\Tools\Pagination\Paginator` no longer fetches identifiers to be able to
 perform the pagination with join collections when max results isn't set in the query.
 
+## Minor BC BREAK: tables filtered with `schema_fiter` are no longer created
```
```suggestion
## Minor BC BREAK: tables filtered with `schema_filter` are no longer created
```

UPGRADE.md
```diff
@@ -10,6 +10,12 @@ Method `Doctrine\ORM\AbstractQuery#useResultCache()` which could be used for bot
...
+  
+When generating schema diffs, if a source table is filtered out by a schema_filter expression, then a CREATE TABLE was
```
```suggestion
When generating schema diffs, if a source table is filtered out by a `schema_filter` expression, then a `CREATE TABLE` was
```

alcaeus (dismissed):

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -910,4 +911,42 @@ public function getUpdateSchemaSql(array $classes, $saveMode = false)
...
     }
+
+    /**
+     * Creates the schema from the database, ensuring tables from the target schema are whitelisted for comparison.
+     */
+    private function createSchemaForComparison(Schema $toSchema) : Schema
+    {
+        // check and memorize if we are using dbal >= 3.0
+        static $hasFilterExpressionMethod;
+        $hasFilterExpressionMethod = $hasFilterExpressionMethod ?? method_exists(Configuration::class, 'getFilterSchemaAssetsExpression');
```
beberlei: Is this really needed? This function should be only called a few times, this looks like unnecessary optimization to me. Also we don't use `static $foo` in our code to declare global variables.

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -910,4 +911,42 @@ public function getUpdateSchemaSql(array $classes, $saveMode = false)
...
+
+        // backup schema assets filter and filter-expression
+        $sm               = $this->em->getConnection()->getSchemaManager();
+        $config           = $this->em->getConnection()->getConfiguration();
+        $filterExpression = $hasFilterExpressionMethod ? $config->getFilterSchemaAssetsExpression() : null;
```
beberlei: This code is not needed as `filterExpressions` are deprecated and are transalted to the new assets filter internally, and since you also have the old `$filter` there and set it anyways, this shouldn't be needed.

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -910,4 +911,42 @@ public function getUpdateSchemaSql(array $classes, $saveMode = false)
...
+            $assetName = $asset instanceof AbstractAsset ? $asset->getName() : $asset;
+
+            return $toSchema->hasTable($assetName) || $toSchema->hasSequence($assetName) || $filter($asset);
+        });
+
+        try {
+            return $sm->createSchema();
+        } finally {
+            // restore schema assets filter and filter-expression
+            if ($filterExpression !== null) {
```
beberlei: This can be cleaned up if we agree that filterExpression is actually not needed here.

UPGRADE.md
```diff
@@ -10,6 +10,12 @@ Method `Doctrine\ORM\AbstractQuery#useResultCache()` which could be used for bot
 To optimize DB interaction, `Doctrine\ORM\Tools\Pagination\Paginator` no longer fetches identifiers to be able to
 perform the pagination with join collections when max results isn't set in the query.
 
+## Minor BC BREAK: tables filtered with `schema_filter` are no longer created
```
beberlei: I would consider this a bugfix, and not a BC break.

beberlei (dismissed):

lib/Doctrine/ORM/Tools/SchemaTool.php
nicolas-grekas: The check is required to me as that's the only way to get the state of the config object when considering legacy and non legacy usages.

UPGRADE.md
lcobucci: https://github.com/doctrine/orm/pull/7875#issuecomment-546242165 and https://github.com/doctrine/orm/pull/7875#issuecomment-546615794 give a bit more explanation on whey we also consider this a behavioural BC-break.

lib/Doctrine/ORM/Tools/SchemaTool.php
nicolas-grekas: static variable removed

lib/Doctrine/ORM/Tools/SchemaTool.php
lcobucci: Wait... @beberlei seems right, though. ORM v2.7 requires DBAL 2.9, which is the version that deprecated these methods and introduced the new ones...

@nicolas-grekas can you explain a bit more why the check is required to you?

I mean, even though `setSchemaAssetsFilter()` resets the filter expression, the callable is the source of truth so this seems unnecessary to me. Perhaps @morozov would like to weight in too?

nicolas-grekas: When `setSchemaAssetsFilter()` is called to setup the decorating closure, we change the state of the config object.
We then have to restore this object to the same exact state it was before the call in the `finally` block.
And there is no other way to do so that calling `setFilterSchemaAssetsExpression` or `setSchemaAssetsFilter` conditionally as done in the PR...

lcobucci: Right, that's what I meant with my comment. But the expression is irrelevant since DBAL takes care of converting it into a callable, which is what's being used to revert stuff...

What I'm trying to say here is that we're trying to manage DBAL internals in the ORM and that's not required.

nicolas-grekas: Not sure to get teh reasoning, but it doesn't matter :)
Updated.

nicolas-grekas: But note that we had tests checking for the situation.
Please advise if I should remove them.

lcobucci: @nicolas-grekas don't worry. I'll check the tests and see what can be cleaned up there. Thanks for updating things here.

lcobucci (approved):
I've unified and simplified the test case too, LGTM.

Thanks @nicolas-grekas and @l-vo!

mvorisek:
@greg0ire why is this not merged in 2.8.x nor master?

mvorisek:
@greg0ire why is this not merged in master?

greg0ire:
@mvorisek because the last merge up to master is a few years old.

PR #7876, opened 2019-10-23
Fix compat of commands with Symfony 5

nicolas-grekas:


lcobucci (approved):
@nicolas-grekas thanks!

Issue #7151, opened 2018-03-22
PersistentCollection::matching() lacks support for inheritance


arnaud-lb:
`PersistentCollection::matching()` uses the entity persister, which in turn uses ResultSetMappingBuilder, which doesn't support inheritance. Trying to call `matching()` throws an exception with the following message:

    ResultSetMapping builder does not currently support your inheritance scheme

Is this fully intentional, or is this feature missing because of a lack of resources ? If I made a PR, would you be willing to accept it ?

Ocramius:
> or is this feature missing because of a lack of resources ?

Likely the case

 > If I made a PR, would you be willing to accept it ?

Depends on the test case that you can provide for this scenario. It would target `3.x` though.

arnaud-lb:
`3.x` seems far away, is `2.x` completely freezed ?

Ocramius:
Yes, only non-breaking bug fixes in `2.x`

ndench:
It certainly looks like the exception was added because of a lack of resources. You can see when the [ResultSetMappingBuilder was created](https://github.com/doctrine/orm/pull/42/files#diff-1f533f556074d8d7e49f661ba316b41bR56) the exception exists which leads me to believe that handling inheritance was considered an edge case they didn't want to deal with at the time.

It appears that over time, the ResultSetMappingBuilder has been updated to a point where it largely supports inheritance, you can [the check was loosened in this PR](https://github.com/doctrine/orm/pull/858/files) to support single table inheritance but only for concrete classes (ie. not the abstract base class).

Very superficial testing for me locally say that we can also support join table inheritance for concrete classes as well so we should be able to loosen the check as long as we add test cases to prove it works.

PR #7878, opened 2019-10-24
remove rsmBuilder restriction on joined table inheritance schemes

jnfry:
See https://github.com/doctrine/orm/issues/7151#issuecomment-545289066

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/NativeQueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/NativeQueryTest.php
@@ -12,8 +12,10 @@ use Doctrine\Tests\Models\CMS\CmsAddress;
 use Doctrine\Tests\Models\CMS\CmsEmail;
 use Doctrine\Tests\Models\CMS\CmsPhonenumber;
 use Doctrine\Tests\Models\CMS\CmsUser;
+use Doctrine\Tests\Models\Company\CompanyAuction;
 use Doctrine\Tests\Models\Company\CompanyContract;
 use Doctrine\Tests\Models\Company\CompanyEmployee;
+use Doctrine\Tests\Models\Company\CompanyEvent;
 use Doctrine\Tests\Models\Company\CompanyFixContract;
 use Doctrine\Tests\Models\Company\CompanyFlexContract;
 use Doctrine\Tests\Models\Company\CompanyPerson;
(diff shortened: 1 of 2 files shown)


SenseException (dismissed):
In would like to see another reviewer take a look at this PR. I'm not sure if the tests are sufficient for this change.

tests/Doctrine/Tests/ORM/Functional/NativeQueryTest.php
```diff
@@ -370,6 +372,31 @@ public function testAbstractClassInSingleTableInheritanceSchemaWithRSMBuilderThr
         $rsm->addRootEntityFromClassMetadata(CompanyContract::class, 'c');
     }
 
+    /**
+     * @group rsm-sti
```
beberlei:
```suggestion
     * @group GH-7151
```

tests/Doctrine/Tests/ORM/Functional/NativeQueryTest.php
```diff
@@ -370,6 +372,31 @@ public function testAbstractClassInSingleTableInheritanceSchemaWithRSMBuilderThr
...
+    {
+        $rsm = new ResultSetMappingBuilder($this->em);
+
+        $rsm->addRootEntityFromClassMetadata(CompanyAuction::class, 'c');
+
+        self::assertSame(CompanyAuction::class, $rsm->getClassName('c'));
+    }
+
+    /**
+     * @group rsm-sti
```
beberlei:
```suggestion
     * @group GH-7151
```

beberlei (changes requested):
This is missing a test in `LazyCriteriaCollectionTest` that triggers the code in `ManyToManyPersister::loadCriteria` with a joined table inheritance entity example.

chq81:
@jnfry Hey there. Thanks for your work. Did you have the possibility to look over the missing tests?

jnfry:
Hey there @chq81, sorry for taking so long to get back to this.

I've had a look at what @beberlei is asking and am unsure what to do. The current `LazyCriteriaCollectionTest` appears to effectively cover the responsibilities of the class (ie. only calling methods on the `Persister` when necessary and caching the result). It doesn't care which type of `Persister` it wraps, and adding a test which asserts it calls `Persister::loadCriteria()` would not test the change I've made.

I agree that there likely needs to be more tests to cover this change, but the `LazyCriteriaCollection` appears to be unaffected. Did you perhaps mean a different test case @beberlei?

I think adding a PersisterTest with a join table inheritance example would be better? Though I'm not sure which of the `*PersisterTest`'s to add this to. If you could perhaps elaborate on what exactly needs to be tested that would greatly help me out. This is my first contribution to Doctrine so I'm a bit lost :sweat_smile:

SenseException:
@jnfry The current tests are covering a `ManyToOne` relation case with `CompanyEvent` / `CompanyAuction`. Please take a look at the content of the mapping of these classes. What's now needed are `ManyToMany` cases.

beberlei:
@jnfry your change suggests this just works, but I am not convinced. What I am missing here is a functional test of the scenario end to end. In one of the joined table inheritance tests you need to add a test case for ->matching that triggers this code.

Your tests at the moment only check small things on ResultSetMapping, but what if the contents of the RSM are wrong so that *other* parts of the system fail, mainly the persisters?

PR #7880, opened 2019-10-28
Update documentation to recommend DQL over QueryBuilder when possible

### LINKED ISSUE #7520: Update docs about QueryBuilder?
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | -

#### Summary

In [this Reddit comment](https://www.reddit.com/r/PHP/comments/a58yt9/phpstorm_doctrine_query_language_support/eblvfmd/), @Ocramius recommends using DQL with NOWDOC strings instead of QueryBuilder. In fact, he says that *"query builder is less safe and less readable"*

#### Current behavior

[The main article of QueryBuilder](https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/query-builder.html) in Doctrine docs says things like this:

> QueryBuilder helper methods are considered the standard way to build DQL queries.

Maybe some docs need some update to recommend DQL instead? (except when building queries dynamically, when using QueryBuilder will be better)

Thanks!

kuraobi:
Fixes #7520 

As explained in the related issue, the idea here is to promote usage of DQL over `QueryBuilder` when writing static queries. While the `QueryBuilder` helps dynamically building DQL, it brings complexity to the code, so it should only be used when it actually brings value compared to DQL.
While the `QueryBuilder` could be argued to bring abstraction, the real abstraction is DQL itself, and the `QueryBuilder` does not bring more abstraction to writing queries.

I tried to clarify the intent of both DQL and the `QueryBuilder`, so please tell me if you think my update is clear enough or not, and I'll update the PR accordingly.

### DIFF
--- a/docs/en/reference/faq.rst
+++ b/docs/en/reference/faq.rst
@@ -198,6 +198,21 @@ No, it is not supported to sort by function in DQL. If you need this functionali
 use a native-query or come up with another solution. As a side note: Sorting with ORDER BY RAND() is painfully slow
 starting with 1000 rows.
 
+Is it better to write DQL or to generate it with the query builder?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The purpose of the ``QueryBuilder`` is to generate DQL dynamically,
+which is useful when you have optional filters, conditional joins, etc.
+
+But the ``QueryBuilder`` is not an alternative to DQL, it actually generates DQL
+queries at runtime, which are then interpreted by Doctrine. This means that
+using the ``QueryBuilder`` to build and run a query is actually always slower
+than only running the corresponding DQL query.
+
+So if you only need to generate a query and bind parameters to it,
+you should use plain DQL, as this is a simpler and much more readable solution.
+You should only use the ``QueryBuilder`` when you can't achieve what you want to do with a DQL query.
+
 A Query fails, how can I debug it?
 ----------------------------------
 
--- a/docs/en/reference/query-builder.rst
+++ b/docs/en/reference/query-builder.rst
@@ -9,6 +9,12 @@ programmatically build queries, and also provides a fluent API.
 This means that you can change between one methodology to the other
 as you want, or just pick a preferred one.
 
+.. note::
+
+    The ``QueryBuilder`` is not an abstraction of DQL, but merely a tool to dynamically build it.
+    You should still use plain DQL when you can, as it is simpler and more readable.
+    More about this in the :doc:`FAQ <faq>`_.
+
 Constructing a new QueryBuilder object
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -80,7 +86,7 @@ Working with QueryBuilder
 High level API methods
 ^^^^^^^^^^^^^^^^^^^^^^
 
-To simplify even more the way you build a query in Doctrine, you can take
+The most straightforward way to build a dynamic query with the ``QueryBuilder`` is by taking
 advantage of Helper methods. For all base code, there is a set of
 useful methods to simplify a programmer's life. To illustrate how
 to work with them, here is the same example 6 re-written using
@@ -97,10 +103,9 @@ to work with them, here is the same example 6 re-written using
        ->orderBy('u.name', 'ASC');
 
 ``QueryBuilder`` helper methods are considered the standard way to
-build DQL queries. Although it is supported, using string-based
-queries should be avoided.  You are greatly encouraged to use
-``$qb->expr()->*`` methods. Here is a converted example 8 to
-suggested standard way to build queries:
+use the ``QueryBuilder``. The ``$qb->expr()->*`` methods can help you
+build conditional expressions dynamically. Here is a converted example 8 to
+suggested way to build queries with dynamic conditions:
 
 .. code-block:: php
 
--- a/docs/en/reference/working-with-objects.rst
+++ b/docs/en/reference/working-with-objects.rst
@@ -800,7 +800,9 @@ DQL and its syntax as well as the Doctrine class can be found in
 :doc:`the dedicated chapter <dql-doctrine-query-language>`.
 For programmatically building up queries based on conditions that
 are only known at runtime, Doctrine provides the special
-``Doctrine\ORM\QueryBuilder`` class. More information on
+``Doctrine\ORM\QueryBuilder`` class. While this a powerful tool,
+it also brings more complexity to your code compared to plain DQL,
+so you should only use it when you need it. More information on
 constructing queries with a QueryBuilder can be found
 :doc:`in Query Builder chapter <query-builder>`.
 
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -1210,8 +1210,7 @@ The console output of this script is then:
     throw your ORM into the dumpster, because it doesn't support some
     the more powerful SQL concepts.
 
-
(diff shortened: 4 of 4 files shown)


docs/en/reference/query-builder.rst
```diff
@@ -80,7 +85,7 @@ Working with QueryBuilder
 High level API methods
 ^^^^^^^^^^^^^^^^^^^^^^
 
-To simplify even more the way you build a query in Doctrine, you can take
+The simplest way to build a dynamic query with the ``QueryBuilder`` is by taking
```
Pierstoval:
```suggestion
The most straightforward way to build a dynamic query with the ``QueryBuilder`` is by taking
```

I try as much as possible to avoid the word "simple", because nothing we do is _actually_ simple, and it depends on the point of view 😄

docs/en/reference/working-with-objects.rst
```diff
@@ -800,7 +800,9 @@ DQL and its syntax as well as the Doctrine class can be found in
 :doc:`the dedicated chapter <dql-doctrine-query-language>`.
 For programmatically building up queries based on conditions that
 are only known at runtime, Doctrine provides the special
-``Doctrine\ORM\QueryBuilder`` class. More information on
+``Doctrine\ORM\QueryBuilder`` class. While this a powerful tool,
+it also brings more complexity to your code compared to simple DQL,
```
Pierstoval:
```suggestion
it also brings more complexity to your code compared to common DQL,
```

Pierstoval (changes requested):
Some small wording reviews, nothing fundamental, but I hope it helps :)

docs/en/reference/query-builder.rst
kuraobi: To be fair, "simplest" does not imply simple: the simplest way to build a spaceship is not simple 😆 but I see your point

docs/en/reference/faq.rst
```diff
@@ -198,6 +198,15 @@ No, it is not supported to sort by function in DQL. If you need this functionali
...
 starting with 1000 rows.
 
+Is it better to write DQL or to generate it with the QueryBuilder?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The purpose of the QueryBuilder is to generate DQL dynamically,
+which is useful when you have optional filters, conditional joins, etc.
+But if you only need to generate a query and bind parameters to it,
+you should use plain DQL, as this is a simpler and much more readable solution.
+You should only use the QueryBuilder when you can't achieve what you want to do with a DQL query.
```
greg0ire: I'd either surround `QueryBuilder` with double backticks or use `query builder`

docs/en/reference/query-builder.rst
```diff
@@ -9,6 +9,11 @@ programmatically build queries, and also provides a fluent API.
 This means that you can change between one methodology to the other
 as you want, or just pick a preferred one.
 
+.. note::
+
+    The ``QueryBuilder`` is not an abstraction of DQL, but merely a tool to dynamically build it.
+    You should still use plain DQL when you can, as it is simpler and more readable.
```
greg0ire: How about this?

```suggestion
    Dynamically built DQL queries usually are the exception though, not
    the rule. For the majority of queries that do not need to change
    based on some condition, you should default to using DQL strings,
    because Doctrine would always build a DQL string behind the scenes
    anyway, and because it is simpler and more readable.
```

greg0ire (commented):
Great work!

docs/en/reference/query-builder.rst
kuraobi: While your text is fine, I'm hesitant because I don't want to over-explain it, because then people might not read it or read it and fail to get the point. Maybe I can add this to the FAQ part instead, because I don't think people will have the required attention span here. IMHO it's better to write "just STFU and do this because it's better" here, which is what most people want to read in the documentation. And for people who really want to know why or settle a debate between developer gurus, there is the FAQ.

Pierstoval: Agree, the explanation is fine, but going straightforward and appending `For more details, go there(FAQ link)`

SenseException: You might also put a link to the section of the faq into this note.

docs/en/reference/query-builder.rst
SenseException: Still "simple" and "simplest" suggest an easy to understand method. I try to avoid words like "simple" or "then you ONLY need to".

docs/en/reference/query-builder.rst
```diff
@@ -97,10 +102,9 @@ to work with them, here is the same example 6 re-written using
        ->orderBy('u.name', 'ASC');
 
 ``QueryBuilder`` helper methods are considered the standard way to
-build DQL queries. Although it is supported, using string-based
-queries should be avoided.  You are greatly encouraged to use
-``$qb->expr()->*`` methods. Here is a converted example 8 to
-suggested standard way to build queries:
+use the ``QueryBuilder``. The ``$qb->expr()->*`` methods can help you
+build conditional expressions dynamically. Here is a converted example 8 to
```
SenseException: The examples aren't counted. Does it make sense to call it "converted example 8" and can we remove the number from the docs?

SenseException (dismissed):

docs/en/reference/query-builder.rst
kuraobi: I didn't want to change this part of the sentence as it wasn't in the scope of the PR, but yes, these example numbers are confusing, I don't even know where/how to find the so called "example 8". Maybe something that existed before and was removed at some point?

docs/en/reference/query-builder.rst
kuraobi: Actually your text didn't fit well with my paragraph in the FAQ so I wrote it differently.
@greg0ire please tell me if the new paragraph in the FAQ covers your point.

Ocramius (approved):

docs/en/reference/query-builder.rst
SenseException: Let's keep this for another PR in the future. :-)

docs/en/reference/faq.rst
```diff
@@ -198,6 +198,21 @@ No, it is not supported to sort by function in DQL. If you need this functionali
...
 
+Is it better to write DQL or to generate it with the query builder?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The purpose of the ``QueryBuilder`` is to generate DQL dynamically,
+which is useful when you have optional filters, conditional joins, etc.
+
+But the ``QueryBuilder`` is not an alternative to DQL, it actually generates DQL
+queries at runtime, which are then interpreted by Doctrine. This means that
+using the ``QueryBuilder`` to build and run a query is actually always slower
```
Majkl578: While I understand you want to convince people to not use QueryBuilder, using performance as a reason sounds funny in the context of an ORM. A few method calls is totally negligible and not measurable.

greg0ire: Totally, but I think some people might be using the QB because they think it's more performant.

greg0ire: That paragraph was my idea, but I'll make a PR to amend it if more people think it's irrelevant. I I still like it because it teaches people about the internals.

kuraobi: The point here is to explain that the QB is an extra layer that comes at a cost. Coding choices are based on trade offs, and I'm just emphasizing that performance wouldn't be in the "+" column. A lot of people think the QB is preferable because Doctrine doesn't need to parse DQL, but it's not true.

PR #7882, opened 2019-10-30
Fix #7877

### LINKED ISSUE #7877: Extra update for self-referencing Many-To-One with NONE generator strategy during persist
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.4

#### Summary

Using a self-referencing entity with a Many-To-One relationship set during the `__construct()` call with a `NONE` strategy for id generation leads to two queries:
- first an INSERT
- then an UPDATE for the relationship

#### Current behavior

- this requires the relationship related column to accept `NULL` for the first `INSERT`

`[2019-10-23 16:04:21] doctrine.DEBUG: INSERT INTO organization (id, legal_parent_id) VALUES (?, ?, ?, ?, ?, ?, ?) {"1":"[object] (Ramsey\\Uuid\\Uuid: \"c68555b0-f5ae-11e9-914e-38baf82a0624\")","2":null} []`

- this leads to a useless `UPDATE` query

`[2019-10-23 16:04:21] doctrine.DEBUG: UPDATE organization SET legal_parent_id = ? WHERE id = ? ["[object] (Ramsey\\Uuid\\Uuid: \"c68555b0-f5ae-11e9-914e-38baf82a0624\")","[object] (Ramsey\\Uuid\\Uuid: \"c68555b0-f5ae-11e9-914e-38baf82a0624\")"] []`

#### How to reproduce

```sql

CREATE TABLE `organization` (
  `id` binary(16) NOT NULL COMMENT '(DC2Type:uuid_binary_ordered_time)',
  `legal_parent_id` BINARY(16) NOT NULL COMMENT '(DC2Type:uuid_binary_ordered_time)',
  PRIMARY KEY (`id`),
  CONSTRAINT FK_C1EE637C9EB3F521 FOREIGN KEY (legal_parent_id) REFERENCES organization (id)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;

```

```php
<?php

declare(strict_types=1);

namespace App\Entity;

/**
 * An Organization
 * @ORM\Entity(repositoryClass="App\Repository\OrganizationRepository")
 */
class Organization
{
    public function __construct()
    {
        $factory = clone \Ramsey\Uuid\Uuid::getFactory();
        $codec = new \Ramsey\Uuid\Codec\OrderedTimeCodec(
            $factory->getUuidBuilder()
        );
        $factory->setCodec($codec);
		$this->id = $factory->uuid1();
        $this->legalParent = $this;
        parent::__construct();
    }

    /**
     * Unique ID of the entity
     *
     * @ORM\Id()
     * @ORM\Column(type="uuid_binary_ordered_time", unique=true)
     */
    private $id;

    public function getId(): ?string
    {
        return $this->id->__toString();
    }

    /**
     * @var self
     * @ORM\ManyToOne(targetEntity="App\Entity\Organization")
     * @ORM\JoinColumn(fieldName="legal_parent_id", referencedColumnName="id", nullable=false)
     */
    protected $legalParent;

    public function getLegalParent(): self
    {
        return $this->legalParent;
    }

    public function setLegalParent(self $legalParent): self
    {
        $this->legalParent = $legalParent;

        return $this;
    }
}

```

#### Expected behavior

- Only the `INSERT` query with `legal_parent_id` not `NULL`

`[2019-10-23 16:04:21] doctrine.DEBUG: INSERT INTO organization (id, legal_parent_id) VALUES (?, ?, ?, ?, ?, ?, ?) {"1":"[object] (Ramsey\\Uuid\\Uuid: \"c68555b0-f5ae-11e9-914e-38baf82a0624\")","2":[object] (Ramsey\\Uuid\\Uuid: \"c68555b0-f5ae-11e9-914e-38baf82a0624\")} []`

#### Possible solution

This comes from `vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php`

image

We could check if `spl_object_hash($newVal) === spl_object_hash($entity)` and if `ClassMetadataInfo::GENERATOR_TYPE_NONE === $this->class->generatorType`. If yes, then do not set `$newVal = null;` and do not schedule extra update.

I can provide a PR if you want

sylfabre:
Fixes https://github.com/doctrine/orm/issues/7877

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -725,29 +725,29 @@ class BasicEntityPersister implements EntityPersister
      */
     private function isExtraUpdateRequired($entity, $newVal): bool
     {
+        // The association is left empty
         if ($newVal === null) {
             return false;
         }
 
-        $oid = spl_object_id($newVal);
         $uow = $this->em->getUnitOfWork();
-        if (! (isset($this->queuedInserts[$oid]) || $uow->isScheduledForInsert($newVal))) {
+        $oid = spl_object_id($newVal);
+
+        // The value is not scheduled for insert so an extra update is not required
+        if (! isset($this->queuedInserts[$oid]) && ! $uow->isScheduledForInsert($newVal)) {
             return false;
         }
 
+        // The entity is not self-referencing
         if ($newVal !== $entity) {
             return true;
         }
 
-        $identifiers = $this->class->getIdentifier();
-        // Only single-column identifiers are supported
+        // An extra update is useless for application-generated single-column identifiers
+        $identifiers     = $this->class->getIdentifier();
         $entityChangeSet = $uow->getEntityChangeSet($entity);
-        if (count($identifiers) === 1 && isset($entityChangeSet[$identifiers[0]])) {
-            // Extra update is required if the current entity does not have yet a value for its identifier
-            return $entityChangeSet[$identifiers[0]][1] === null;
-        }
 
-        return true;
+        return ! isset($entityChangeSet[$identifiers[0]]) || $entityChangeSet[$identifiers[0]][1] === null;
     }
 
     /**
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -680,17 +680,12 @@ class BasicEntityPersister implements EntityPersister
                 continue;
             }
 
-            if ($newVal !== null) {
-                $oid = spl_object_id($newVal);
-
-                if (isset($this->queuedInserts[$oid]) || $uow->isScheduledForInsert($newVal)) {
-                    // The associated entity $newVal is not yet persisted, so we must
-                    // set $newVal = null, in order to insert a null value and schedule an
-                    // extra update on the UnitOfWork.
-                    $uow->scheduleExtraUpdate($entity, [$field => [null, $newVal]]);
-
-                    $newVal = null;
-                }
+            if ($this->isExtraUpdateRequired($entity, $newVal)) {
+                // The associated entity $newVal is not yet persisted, so we must
+                // set $newVal = null, in order to insert a null value and schedule an
+                // extra update on the UnitOfWork.
+                $uow->scheduleExtraUpdate($entity, [$field => [null, $newVal]]);
+                $newVal = null;
             }
 
             $newValId = null;
@@ -716,6 +711,43 @@ class BasicEntityPersister implements EntityPersister
         }
 
         return $result;
+    }
+
+    /**
+     * Decides if an extra update is required for the entity being persisted
+     * This is only required if the associated entity is different from the current one,
+     * and if the current entity relies on the database to generate its id
+     *
+     * @param object      $entity
+     * @param object|null $newVal
+     *
+     * @return bool Returns true is an extra update is required
+     */
+    private function isExtraUpdateRequired($entity, $newVal): bool
+    {
+        if ($newVal === null) {
+            return false;
+        }
+
+        $oid = spl_object_id($newVal);
+        $uow = $this->em->getUnitOfWork();
+        if (! (isset($this->queuedInserts[$oid]) || $uow->isScheduledForInsert($newVal))) {
+            return false;
+        }
+
+        if ($newVal !== $entity) {
+            return true;
+        }
+
+        $identifiers = $this->class->getIdentifier();
+        // Only single-column identifiers are supported
+        $entityChangeSet = $uow->getEntityChangeSet($entity);
+        if (count($identifiers) === 1 && isset($entityChangeSet[$identifiers[0]])) {
+            // Extra update is required if the current entity does not have yet a value for its identifier
+            return $entityChangeSet[$identifiers[0]][1] === null;
+        }
+
+        return true;
     }
 
     /**
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -680,12 +680,18 @@ class BasicEntityPersister implements EntityPersister
                 continue;
             }
 
-            if ($this->isExtraUpdateRequired($entity, $newVal)) {
-                // The associated entity $newVal is not yet persisted, so we must
-                // set $newVal = null, in order to insert a null value and schedule an
-                // extra update on the UnitOfWork.
-                $uow->scheduleExtraUpdate($entity, [$field => [null, $newVal]]);
-                $newVal = null;
+            $oid = spl_object_id($newVal);
+
+            if (isset($this->queuedInserts[$oid]) || $uow->isScheduledForInsert($newVal)) {
+                // Optimization: Self-referencing entities with application-provided IDs
+                // do not need extra updates
+                if (! ($newVal === $entity && $this->class->getIdentifierValues($newVal))) {
+                    // The associated entity $newVal is not yet persisted, so we must
+                    // set $newVal = null, in order to insert a null value and schedule an
+                    // extra update on the UnitOfWork.
+                    $uow->scheduleExtraUpdate($entity, [$field => [null, $newVal]]);
+                    $newVal = null;
+                }
             }
 
             $newValId = null;
@@ -711,43 +717,6 @@ class BasicEntityPersister implements EntityPersister
         }
 
         return $result;
-    }
-
-    /**
-     * Decides if an extra update is required for the entity being persisted
-     * This is only required if the associated entity is different from the current one,
-     * and if the current entity relies on the database to generate its id
-     *
-     * @param object      $entity
-     * @param object|null $newVal
-     *
-     * @return bool Returns true is an extra update is required
-     */
-    private function isExtraUpdateRequired($entity, $newVal): bool
-    {
-        // The association is left empty
-        if ($newVal === null) {
-            return false;
-        }
-
-        $uow = $this->em->getUnitOfWork();
-        $oid = spl_object_id($newVal);
-
-        // The value is not scheduled for insert so an extra update is not required
-        if (! isset($this->queuedInserts[$oid]) && ! $uow->isScheduledForInsert($newVal)) {
-            return false;
-        }
-
-        // The entity is not self-referencing
-        if ($newVal !== $entity) {
-            return true;
-        }
-
-        // An extra update is useless for application-generated single-column identifiers
-        $identifiers     = $this->class->getIdentifier();
-        $entityChangeSet = $uow->getEntityChangeSet($entity);
-
-        return ! isset($entityChangeSet[$identifiers[0]]) || $entityChangeSet[$identifiers[0]][1] === null;
     }
 
     /**
--- a/tests/Doctrine/Tests/ORM/Functional/GH7877Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/GH7877Test.php
@@ -35,13 +35,7 @@ class GH7877Test extends OrmFunctionalTestCase
         }
 
         $this->_em->flush();
-        if ($this->isQueryLogAvailable()) {
-            if ($this->getQueryLog()->queries[0]['sql'] === '"START TRANSACTION"') {
-                self::assertQueryCount(3);
-            } else {
-                self::assertQueryCount(1);
-            }
-        }
+        $this->checkQueryCount(false);
 
         $this->_em->clear();
 
@@ -49,7 +43,7 @@ class GH7877Test extends OrmFunctionalTestCase
         $this->assertSame($entityId, $child->parent->id);
     }
 
-    public function textExtraUpdateWithDatabaseGeneratedId(): void
+    public function testExtraUpdateWithDatabaseGeneratedId(): void
     {
         $entity         = new GH7877DatabaseGenerated();
         $entity->parent = $entity;
@@ -60,19 +54,26 @@ class GH7877Test extends OrmFunctionalTestCase
         }
 
         $this->_em->flush();
-        if ($this->isQueryLogAvailable()) {
-            if ($this->getQueryLog()->queries[0]['sql'] === '"START TRANSACTION"') {
-                self::assertQueryCount(4);
-            } else {
-                self::assertQueryCount(2);
-            }
-        }
+        $this->checkQueryCount(
+            $this->_em->getClassMetadata(GH7877DatabaseGenerated::class)->idGenerator->isPostInsertGenerator()
+        );
 
         $entityId = $entity->id;
         $this->_em->clear();
 
         $child = $this->_em->find(GH7877DatabaseGenerated::class, $entityId);
         $this->assertSame($entityId, $child->parent->id);
+    }
+
+    private function checkQueryCount(bool $extra): void
+    {
+        if ($this->isQueryLogAvailable()) {
+            if ($this->getQueryLog()->queries[0]['sql'] === '"START TRANSACTION"') {
+                self::assertQueryCount($extra ? 4 : 3);
+            } else {
+                self::assertQueryCount($extra ? 2 : 1);
+            }
+        }
     }
 }
 
--- a/tests/Doctrine/Tests/ORM/Functional/GH7877Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/GH7877Test.php
@@ -1,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+use function uniqid;
+
+/**
+ * @group GH7877
+ */
+class GH7877Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->createSchemaForModels(
+            GH7877ApplicationGenerated::class,
+            GH7877DatabaseGenerated::class
+        );
+    }
+
+    public function testNoExtraUpdateWithApplicationGeneratedId(): void
+    {
+        $entity         = new GH7877ApplicationGenerated($entityId = uniqid());
+        $entity->parent = $entity;
+        $this->_em->persist($entity);
+
+        if ($this->isQueryLogAvailable()) {
+            $this->getQueryLog()->reset()->enable();
+        }
+
+        $this->_em->flush();
+        if ($this->isQueryLogAvailable()) {
+            if ($this->getQueryLog()->queries[0]['sql'] === '"START TRANSACTION"') {
+                self::assertQueryCount(3);
+            } else {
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -665,9 +666,26 @@ protected function prepareUpdateData($entity)
...
             if ($newValue !== null && $uow->isScheduledForInsert($newValue)) {
-                $uow->scheduleExtraUpdate($entity, [$propertyName => [null, $newValue]]);
+                // This is only required if the associated entity is different from the current one,
+                // and if the current entity relies on the database to generate its id
+                if ($newValue !== $entity) {
+                    $scheduleExtraUpdate = true;
+                } else {
+                    $identifiers = $this->class->getIdentifier();
+                    // Only single-column identifiers are supported
+                    if (1 === count($identifiers) && isset($entityChangeSet[$identifiers[0]])) {
```
Ocramius: There should be an explicit method to ask if the identifier is single-valued? Otherwise, I'd suggest using the identifier flattener (it's a class lying around in internals)

Ocramius (changes requested):

greg0ire:

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
sylfabre: @Ocramius actually the identifier flattener returns an array too, not a hash as string.

And I've found any method to ask if the identifier is single-valued.

tests/Doctrine/Tests/ORM/Functional/SelfReferencingTest.php
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue(strategy="NONE")
+     */
+    private $id;
+
+    public function getId(): ?int
```
Ocramius: Use public properties, avoid any getter/setter interactions in test scope (as they add possible issues)

tests/Doctrine/Tests/ORM/Functional/SelfReferencingTest.php
```diff
@@ -0,0 +1,173 @@
...
+    public function providerDifferentEntity()
+    {
+        yield [Issue7877ApplicationGenerated::class];
+        yield [Issue7877DatabaseGenerated::class];
+    }
+
+    /**
+     * @dataProvider providerDifferentEntity
+     */
+    public function testDifferentEntity(string $class)
```
Ocramius: What's being tested specifically? Can you clarify that via method names? (applies to all methods in this test)

Ocramius (changes requested):

tests/Doctrine/Tests/ORM/Functional/SelfReferencingTest.php
sylfabre: fixed

sylfabre:
@greg0ire Which coding standard do you use with PHPCS?

I see on Travis that it is checked with `php git-phpcs.phar origin/$TRAVIS_BRANCH...$TRAVIS_PULL_REQUEST_SHA` 

So I'm expecting the default one of PHPCS but I don't get the same violations when I run it.

I've looked on https://www.doctrine-project.org/contribute/index.html but I cannot find any info about the coding standard.

Thanks!

greg0ire:
We use https://github.com/doctrine/coding-standard
You should be able to fix cs issues by running `vendor/bin/phpcbf`

sylfabre:
@greg0ire @Ocramius I've fixed tests and updated according to your comments

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -681,6 +678,42 @@ protected function prepareUpdateData($entity)
...
 
+    /**
+     * Decides if an extra update is required for the entity being persisted
+     * This is only required if the associated entity is different from the current one,
+     * and if the current entity relies on the database to generate its id
+     *
+     * @param mixed $entity
+     * @param mixed $newVal
+     *
+     * @return bool Returns true is an extra update is required
```
greg0ire: I think these lines are unnecessary

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -681,6 +678,42 @@ protected function prepareUpdateData($entity)
...
+     */
+    private function isExtraUpdateRequired($entity, $newVal) : bool
```
greg0ire: I might be wrong, but both arguments are objects, right?
```suggestion
    private function isExtraUpdateRequired(object $entity, object $newVal) : bool
```

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -681,6 +678,42 @@ protected function prepareUpdateData($entity)
...
+    {
+        if ($newVal === null) {
+            return false;
+        }
+        $oid = spl_object_hash($newVal);
+        $uow = $this->em->getUnitOfWork();
+        if (isset($this->queuedInserts[$oid]) || $uow->isScheduledForInsert($newVal)) {
+            if ($newVal !== $entity) {
+                return true;
+            } else {
```
greg0ire: This `else` is useless, since the `if` contains a `return` statement.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -681,6 +678,42 @@ protected function prepareUpdateData($entity)
...
+                $identifiers = $this->class->getIdentifier();
+                // Only single-column identifiers are supported
+                $entityChangeSet = $uow->getEntityChangeSet($entity);
+                if (count($identifiers) === 1 && isset($entityChangeSet[$identifiers[0]])) {
+                    // Extra update is required if the current entity does not have yet a value for its identifier
+                    return $entityChangeSet[$identifiers[0]][1] === null;
+                } else {
```
greg0ire: Same here

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -681,6 +678,42 @@ protected function prepareUpdateData($entity)
...
+                    return true;
+                }
+            }
+        } else {
```
greg0ire: Same here

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -681,6 +678,42 @@ protected function prepareUpdateData($entity)
...
+     * @return bool Returns true is an extra update is required
+     */
+    private function isExtraUpdateRequired($entity, $newVal) : bool
+    {
+        if ($newVal === null) {
+            return false;
+        }
+        $oid = spl_object_hash($newVal);
+        $uow = $this->em->getUnitOfWork();
+        if (isset($this->queuedInserts[$oid]) || $uow->isScheduledForInsert($newVal)) {
```
greg0ire: You could negate this condition and use an early return.

sylfabre: fixed

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
sylfabre: @greg0ire `$newVal` might be `null` too.

This type hint is not used in the `prepareUpdateData()` method and it makes PHP7.1 tests fail, so I'll leave it without the type hint

greg0ire: Oh right, your PR is on 7.1, nevermind then. Please document both parameters, but with `object` rather than `mixed`

tests/Doctrine/Tests/ORM/Functional/SelfReferencingTest.php
```diff
@@ -0,0 +1,121 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+use function count;
+
+class SelfReferencingTest extends OrmFunctionalTestCase
```
lcobucci: Please use `GH7877` as the prefix for the classes (as we have on the newer test cases). Also, add `@group GH7877` to the class docblock to be able to easily execute all tests related to that issues.

Novynn:
Any maintainers able to check this again? Seems that @sylfabre has completed all the requests.

greg0ire:
@Novynn 2.6 is no longer maintained, so I retargeted on 2.15.x, but now there are conflicts that need to be addressed. Feel free to open a separate PR in case @sylfabre does not respond.

sylfabre:
@Novynn Thank you for resuscitating this PR!

@Ocramius @greg0ire I rebased and fixed the conflict so you're good to go!

tests/Doctrine/Tests/ORM/Functional/GH7877Test.php
```diff
@@ -0,0 +1,124 @@
...
+
+        $this->_em->clear();
+
+        $child = $this->_em->find(GH7877DatabaseGenerated::class, $entityId);
+        $this->assertSame($entityId, $child->parent->id);
+    }
+}
+
+/**
+ * @Entity
```
greg0ire: You will need an import statement for any annotation you use.

sylfabre:
@greg0ire can you please approve the workflow run so I can ensure my PR passes them?
Thanks 🙏

Ocramius:
@sylfabre run failed: related to your changes?

greg0ire:
@sylfabre you do run the tests locally first, right?

sylfabre:
@greg0ire I know how to run locally the `phpcs` check but `phpunit` fails with this message

```

Using DB driver Doctrine\DBAL\Logging\Driver

Doctrine\DBAL\Exception\DriverException : An exception occurred in the driver: could not find driver
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Driver/API/SQLite/ExceptionConverter.php:83
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:1935
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:1884
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:377
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:450
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:408
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:316
 /home/dev/repositories/orm/tests/Doctrine/Tests/TestUtil.php:128
 /home/dev/repositories/orm/tests/Doctrine/Tests/TestUtil.php:65
 /home/dev/repositories/orm/tests/Doctrine/Tests/OrmFunctionalTestCase.php:705
 /home/dev/repositories/orm/tests/Doctrine/Tests/ORM/Functional/GH7877Test.php:19
 
 Caused by
 Doctrine\DBAL\Driver\PDO\Exception: could not find driver
 
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Driver/PDO/Exception.php:28
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Driver/PDO/SQLite/Driver.php:50
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Driver/Middleware/AbstractDriverMiddleware.php:29
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Logging/Driver.php:34
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:375
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:450
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:408
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:316
 /home/dev/repositories/orm/tests/Doctrine/Tests/TestUtil.php:128
 /home/dev/repositories/orm/tests/Doctrine/Tests/TestUtil.php:65
 /home/dev/repositories/orm/tests/Doctrine/Tests/OrmFunctionalTestCase.php:705
 /home/dev/repositories/orm/tests/Doctrine/Tests/ORM/Functional/GH7877Test.php:19
 
 Caused by
 PDOException: could not find driver
 
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Driver/PDO/SQLite/Driver.php:44
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Driver/Middleware/AbstractDriverMiddleware.php:29
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Logging/Driver.php:34
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:375
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:450
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:408
 /home/dev/repositories/orm/vendor/doctrine/dbal/src/Connection.php:316
 /home/dev/repositories/orm/tests/Doctrine/Tests/TestUtil.php:128
 /home/dev/repositories/orm/tests/Doctrine/Tests/TestUtil.php:65
 /home/dev/repositories/orm/tests/Doctrine/Tests/OrmFunctionalTestCase.php:705
 /home/dev/repositories/orm/tests/Doctrine/Tests/ORM/Functional/GH7877Test.php:19
 ```

The error makes sense as I don't have a running SQLite server on my machine, and I haven't found how to set up one from the contributing documentation (https://www.doctrine-project.org/contribute/index.html)

**Comment by greg0ire** [comment]
SQLite does not work with a server, it works with a file. What you are missing is the driver. Either you don't have `pdo-sqlite` installed, or, and I think that's more likely, it's not enabled. Look at your `php.ini`, it might simply be commented out.

**Comment by sylfabre** [comment]
Thanks @greg0ire 
IMO, that's worth documenting it in the contributing guidelines :)

**Comment by greg0ire** [comment]
Yes, definitely! It should IMO go to https://github.com/doctrine/orm/blob/2.15.x/CONTRIBUTING.md

Feel free to start something, otherwise I might do it later.

File: lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php [hunk 12]
```diff
@@ -718,6 +713,43 @@ protected function prepareUpdateData($entity, bool $isInsert = false)
...
+
+        $identifiers = $this->class->getIdentifier();
+        // Only single-column identifiers are supported
+        $entityChangeSet = $uow->getEntityChangeSet($entity);
+        if (count($identifiers) === 1 && isset($entityChangeSet[$identifiers[0]])) {
+            // Extra update is required if the current entity does not have yet a value for its identifier
+            return $entityChangeSet[$identifiers[0]][1] === null;
+        }
+
+        return true;
```
Ocramius: This execution branch not being reached here seems to indicate a missing composite key test

greg0ire:
Documented the SQLite requirement at https://github.com/doctrine/orm/pull/10734

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
sylfabre: Thank you for your insight 👍 
I changed the code to remove this case as I'm only targeting self-referencing entities for this optimization.
So it doesn't make sense to check if this is composite key

sylfabre:
@greg0ire I think we are good to go now.

A test was mis-prefixed (`text` vs `test`) so wasn't executed by PHPUnit and was failing with Postgres. I install a Docker image to test this driver and it works.
May you please allow for another run? Thanks!

greg0ire:
Is this something different than what @mpdude has been working on? @mpdude , if that's the case, please review.

mpdude:
Is my understanding correct that you want to optimize only for the case that you have a self-referencing entity with application-generated IDs? In that case, avoid scheduling the extra update, since the ID is available at insert time?

mpdude:
The following patch seems to work as well.

In general, I prefer dedicated methods like the one you introduced.

In this case, I somehow find it easier to see what the optimization is about. WDYT?

By the way, do you see any cases apart from self-referencing entities where the extra update might be unnecessary?

```patch
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -680,12 +680,16 @@ class BasicEntityPersister implements EntityPersister
                 $oid = spl_object_id($newVal);

                 if (isset($this->queuedInserts[$oid]) || $uow->isScheduledForInsert($newVal)) {
-                    // The associated entity $newVal is not yet persisted, so we must
-                    // set $newVal = null, in order to insert a null value and schedule an
-                    // extra update on the UnitOfWork.
-                    $uow->scheduleExtraUpdate($entity, [$field => [null, $newVal]]);
-
-                    $newVal = null;
+                    // Optimization: Self-referencing entities with application-provided IDs
+                    // do not need extra updates
+                    if (!($newVal === $entity && $this->class->getIdentifierValues($newVal))) {
+                        // The associated entity $newVal is not yet persisted, so we must
+                        // set $newVal = null, in order to insert a null value and schedule an
+                        // extra update on the UnitOfWork.
+                        $uow->scheduleExtraUpdate($entity, [$field => [null, $newVal]]);
+
+                        $newVal = null;
+                    }
                 }
             }

```

sylfabre:
Hello @mpdude 

Thank you for your review

You correctly identified the corner case I'm targeting here.
I guess more aggressive optimizations are possible if entities referencing each other are inserted in the right order (cf your current work in https://github.com/doctrine/orm/pull/10547) while still respecting foreign key constraints.

I implemented the change you suggested: it is indeed easier to understand 👍 
I'm not aware of all available Doctrine internal methods so my solution was kind of hacky!

sylfabre:
Following-up on https://github.com/doctrine/orm/issues/7877#issuecomment-1571417944, this PR still needs some work

sylfabre:
@mpdude @greg0ire I updated the code and the test as suggested, may you please allow for another CI run?
Thanks!

mpdude:
@sylfabre my alternate suggestion is in #10735.

sylfabre:
@mpdude Alright, I'm trusting your judgment on this one as I'm not a Doctrine expert!

Are you taking the lead with your other PR? If yes, I may close mine

mpdude:
As the tests in #10735 show, only considering the self-referencing entity case is not enough. We also have to avoid extra updates for associations between different entities.

The extra update is not only about knowing database-provided IDs, but also about making sure that the depended-upon entity is already in the database so we can make a foreign key reference to it.

That, in turn, requires that the commit order is good enough to insert depended-upon entities first. The current implementation is not, at least not for references between entities of the same class.

In addition to that, the `BasicEntityPersister` needs a way to figure out if the referenced entity has already been inserted during the current commit phase. Only then it may skip the extra update.

Looking at IDs is not sufficient, since for application-provided IDs, they are set even though the insert is still pending.

sylfabre:
Abandoned in favor of https://github.com/doctrine/orm/pull/10735

Issue #3238, opened 2013-06-19
DDC-2516: Undefined offset in ObjectHydrator while working with iterableResult


doctrinebot:
Jira issue originally created by user johanny:

Following this documentation page http://doctrine-orm.readthedocs.org/en/2.0.x/reference/batch-processing.html#iterating-large-results-for-data-processing I tried to create a batch process on 75k entities. But an error came in the ObjectHydrator on line 511 there's an Undefined offset. Now, i just use the SimpleObjectHydrator and it solve the problem. but i'm wondering if it's a real bug or just a miss use of Doctrine.
Reproduce this error could be hard so ask if you want me to give you more informations about it.

beberlei:
Closing in favor of https://github.com/doctrine/orm/issues/7496

PR #7885, opened 2019-11-02
Introduce `getIterable()` on AbstractQuery

simPod:
I do not fancy accessing index 0 (`[0]`) while iterating `iterable` result from `iterate()` to get the entity.

```php
$results = $em->select('e')->from('entity')->getQuery()->iterate();
foreach($resuls as $result) {
   $entityIsHere = $result[0];
}
```

So  was thinking to introduce new method that gets rid of the index and returns the result same way as `getResult()` does. Only with the change that the new behaviour would keep row-by-row hydration.

```php
$generator = $em->select('e')->from('entity')->getQuery()->toIterable();
foreach($generator as $result) {
   $entityIsHere = $result;
}
```

This change deprecates `AbstractQuery::iterate` and follow up PRs renamed the new method `toIterable()`.

- [x]  Implement new row by row hydrating iterator
- [x]  Test the shapes are the same
- [x]   Add separate tests
- [x]   Find name for new method

Follow ups:
- https://github.com/doctrine/orm/pull/8268
- https://github.com/doctrine/orm/pull/8293

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -19,15 +19,28 @@
 
 namespace Doctrine\ORM;
 
-use Doctrine\Common\Util\ClassUtils;
+use Countable;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\ORM\Mapping\MappingException as ORMMappingException;
 use Doctrine\ORM\Query\Parameter;
 use Doctrine\ORM\Cache\QueryCacheKey;
-use Doctrine\Persistence\Mapping\MappingException;
 use Traversable;
+use function array_map;
+use function array_shift;
+use function count;
+use function is_array;
+use function is_numeric;
+use function is_object;
+use function is_scalar;
+use function iterator_count;
 use function iterator_to_array;
+use function ksort;
+use function reset;
+use function serialize;
+use function sha1;
+use function trigger_error;
+use const E_USER_DEPRECATED;
 
 /**
  * Base contract for ORM queries. Base class for Query and NativeQuery.
@@ -428,7 +441,7 @@ abstract class AbstractQuery
             if ($value === null) {
                 throw ORMInvalidArgumentException::invalidIdentifierBindingEntity();
             }
-        } catch (MappingException | ORMMappingException $e) {
+        } catch (ORMMappingException $e) {
             // Silence any mapping exceptions. These can occur if the object in
             // question is not a mapped entity, in which case we just don't do
             // any preparation on the value.
@@ -911,6 +924,8 @@ abstract class AbstractQuery
      * Executes the query and returns an IterableResult that can be used to incrementally
      * iterate over the result.
      *
+     * @deprecated
+     *
      * @param ArrayCollection|array|null $parameters    The query parameters.
      * @param string|int|null            $hydrationMode The hydration mode to use.
      *
@@ -918,6 +933,11 @@ abstract class AbstractQuery
      */
     public function iterate($parameters = null, $hydrationMode = null)
     {
+        @trigger_error(
+            'Method ' . __METHOD__ . '() is deprecated and will be removed in Doctrine ORM 3.0. Use getIterable() instead.',
+            E_USER_DEPRECATED
+        );
+
         if ($hydrationMode !== null) {
             $this->setHydrationMode($hydrationMode);
         }
@@ -930,6 +950,33 @@ abstract class AbstractQuery
         $stmt = $this->_doExecute();
 
         return $this->_em->newHydrator($this->_hydrationMode)->iterate($stmt, $rsm, $this->_hints);
+    }
+
+    /**
+     * Executes the query and returns an iterable that can be used to incrementally
+     * iterate over the result.
+     *
+     * @param ArrayCollection|mixed[] $parameters    The query parameters.
+     * @param string|int|null         $hydrationMode The hydration mode to use.
+     *
+     * @return iterable<mixed>
+     */
+    public function getIterable(iterable $parameters = [], $hydrationMode = null) : iterable
+    {
+        if ($hydrationMode !== null) {
+            $this->setHydrationMode($hydrationMode);
+        }
+
+        if (($this->isCountable($parameters) && count($parameters) !== 0)
+            || ($parameters instanceof Traversable && iterator_count($parameters) !== 0)
+        ) {
+            $this->setParameters($parameters);
+        }
+
+        $rsm  = $this->getResultSetMapping();
+        $stmt = $this->_doExecute();
+
+        return $this->_em->newHydrator($this->_hydrationMode)->getIterable($stmt, $rsm, $this->_hints);
     }
 
     /**
@@ -1163,5 +1210,11 @@ abstract class AbstractQuery
         ksort($hints);
 
         return sha1($query . '-' . serialize($params) . '-' . serialize($hints));
+    }
+
+    /** @param iterable<mixed> $subject */
+    private function isCountable(iterable $subject) : bool
+    {
+        return $subject instanceof Countable || is_array($subject);
     }
 }
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -19,14 +19,20 @@
 
 namespace Doctrine\ORM\Internal\Hydration;
 
+use Doctrine\DBAL\Driver\Statement;
+use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Events;
 use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\ORM\Tools\Pagination\LimitSubqueryWalker;
 use PDO;
+use const E_USER_DEPRECATED;
 use function array_map;
+use function end;
 use function in_array;
+use function trigger_error;
 
 /**
  * Base class for all hydrators. A hydrator is a class that provides some form
@@ -110,6 +116,8 @@ abstract class AbstractHydrator
     /**
      * Initiates a row-by-row hydration.
      *
+     * @deprecated
+     *
      * @param object $stmt
      * @param object $resultSetMapping
      * @param array  $hints
@@ -118,6 +126,11 @@ abstract class AbstractHydrator
      */
     public function iterate($stmt, $resultSetMapping, array $hints = [])
     {
+        @trigger_error(
+            'Method ' . __METHOD__ . '() is deprecated and will be removed in Doctrine ORM 3.0. Use getIterable() instead.',
+            E_USER_DEPRECATED
+        );
+
         $this->_stmt  = $stmt;
         $this->_rsm   = $resultSetMapping;
         $this->_hints = $hints;
@@ -129,6 +142,42 @@ abstract class AbstractHydrator
         $this->prepare();
 
         return new IterableResult($this);
+    }
+
+    /**
+     * Initiates a row-by-row hydration.
+     *
+     * @param mixed[] $hints
+     *
+     * @return iterable<mixed>
+     */
+    public function getIterable(Statement $stmt, ResultSetMapping $resultSetMapping, array $hints = []) : iterable
+    {
+        $this->_stmt  = $stmt;
+        $this->_rsm   = $resultSetMapping;
+        $this->_hints = $hints;
+
+        $evm = $this->_em->getEventManager();
+
+        $evm->addEventListener([Events::onClear], $this);
+
+        $this->prepare();
+
+        $result = [];
+
+        while (true) {
+            $row = $this->_stmt->fetch(FetchMode::ASSOCIATIVE);
+
+            if ($row === false || $row === null) {
+                $this->cleanup();
+
+                break;
+            }
+
+            $this->hydrateRowData($row, $result);
+
+            yield end($result);
+        }
     }
 
     /**
@@ -159,7 +208,7 @@ abstract class AbstractHydrator
 
     /**
      * Hydrates a single row returned by the current statement instance during
-     * row-by-row hydration with {@link iterate()}.
+     * row-by-row hydration with {@link iterate()} or {@link getIterable()}.
      *
      * @return mixed
      */
@@ -167,7 +216,7 @@ abstract class AbstractHydrator
     {
         $row = $this->_stmt->fetch(PDO::FETCH_ASSOC);
 
-        if ( ! $row) {
+        if ($row === false || $row === null) {
             $this->cleanup();
 
             return false;
--- a/tests/Doctrine/Tests/ORM/Functional/QueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/QueryTest.php
@@ -8,6 +8,7 @@ use Doctrine\ORM\NonUniqueResultException;
 use Doctrine\ORM\Proxy\Proxy;
 use Doctrine\ORM\Query\QueryException;
 use Doctrine\ORM\UnexpectedResultException;
+use Doctrine\Tests\GetIterableTester;
 use Doctrine\Tests\Models\CMS\CmsUser,
     Doctrine\Tests\Models\CMS\CmsArticle,
     Doctrine\Tests\Models\CMS\CmsPhonenumber;
@@ -15,6 +16,8 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\Query\Parameter;
 use Doctrine\Tests\OrmFunctionalTestCase;
+use function count;
+use function iterator_to_array;
 
 /**
  * Functional Query tests.
@@ -217,26 +220,27 @@ class QueryTest extends OrmFunctionalTestCase
         $query = $this->_em->createQuery('select a from ' . CmsArticle::class . ' a WHERE a.topic = ?1');
         $articles = $query->iterate(new ArrayCollection([new Parameter(1, 'Doctrine 2')]), Query::HYDRATE_ARRAY);
 
-        $found = [];
+        $expectedArticle = [
+            'id'      => $articleId,
+            'topic'   => 'Doctrine 2',
+            'text'    => 'This is an introduction to Doctrine 2.',
+            'version' => 1,
+        ];
 
-        foreach ($articles AS $article) {
-            $found[] = $article;
-        }
+        $articles = iterator_to_array($articles, false);
 
-        $this->assertEquals(1, count($found));
-        $this->assertSame(
-            [
-                [
-                    [
-                        'id'      => $articleId,
-                        'topic'   => 'Doctrine 2',
-                        'text'    => 'This is an introduction to Doctrine 2.',
-                        'version' => 1,
-                    ],
-                ],
-            ],
-            $found
+        self::assertCount(1, $articles);
+        self::assertEquals([[$expectedArticle]], $articles);
+
+        $articles = $query->getIterable(
+            new ArrayCollection([new Parameter(1, 'Doctrine 2')]),
+            Query::HYDRATE_ARRAY
         );
+
+        $articles = GetIterableTester::iterableToArray($articles);
+
+        self::assertCount(1, $articles);
+        self::assertEquals([$expectedArticle], $articles);
     }
 
     public function testIterateResult_IterativelyBuildUpUnitOfWork()
@@ -275,8 +279,23 @@ class QueryTest extends OrmFunctionalTestCase
         $this->assertSame(["Doctrine 2", "Symfony 2"], $topics);
         $this->assertSame(2, $iteratedCount);
 
-        $this->_em->flush();
-        $this->_em->clear();
+        $articles = $query->getIterable();
+
+        $iteratedCount = 0;
+        $topics        = [];
+
+        foreach ($articles as $article) {
+            $topics[] = $article->topic;
+
+            $identityMap      = $this->_em->getUnitOfWork()->getIdentityMap();
+            $identityMapCount = count($identityMap[CmsArticle::class]);
+            self::assertGreaterThan($iteratedCount, $identityMapCount);
+
+            $iteratedCount++;
+        }
+
+        self::assertSame(['Doctrine 2', 'Symfony 2'], $topics);
+        self::assertSame(2, $iteratedCount);
     }
 
     public function testIterateResultClearEveryCycle()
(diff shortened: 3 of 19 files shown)


Ocramius:
I think it would be better to have a new method for `iterate()`, and deprecating or removing the old one: the BC break is too subtle otherwise

lcobucci:
I think you're missing the scenario where you have multiple things in the select clause.

simPod:
@lcobucci Thanks. I assumed I handled it via `array_key_exists(0, $this->current) && count($this->current) === 1`. If the condition is not fullfiled it behaves in the original way. But I have no tests here anyway, will try to do what @Ocramius suggested by writing tests first and then will see.

lib/Doctrine/ORM/Internal/Hydration/IterableNewResult.php
```diff
@@ -0,0 +1,91 @@
...
+namespace Doctrine\ORM\Internal\Hydration;
+
+use Iterator;
+use function array_key_exists;
+use function count;
+use function is_array;
+
+/**
+ * Represents a result structure that can be iterated over, hydrating row-by-row
+ * during the iteration. An IterableResult is obtained by AbstractHydrator#iterate().
```
greg0ire:
```suggestion
 * during the iteration. An IterableResult is obtained by AbstractHydrator#getIterable().
```

simPod:
What version should this eventually be targeted? v3? If so, can't we BC break and replace original `IterableResult` with `IterableNewResult`? I was thinking how to name it and am kind of out of ideas. `IterableResult` is the best name for it IMO.

lcobucci:
@simPod it's v3 indeed. However, it would be quite nice if we could provide a good migration path on v2.7 :+1:

simPod:
@lcobucci so if I understand correctly, we would like to introduce this into v2.7 (if there are no gotchas in implementation) under new namimg. Therefore original naming is reserved and can't be used neither for v2.7 or v3.  
Amirite?

lcobucci:
@simPod correct. After a few internal discussions, we decided to be a bit more flexible with the 2.x series regarding improvements and this can go to `2.8.0` - since we're going to release `2.7.0` soon.

beberlei:
The logic in the new iterator is wrong when the hydration mode is changed to something else than object. The logic must be bullet proof, which is probably achieved by looking at the `ResultSetMapping` for this query to determine that only one entity is fetched and what its alias name is and 0 if none is used. The following additional tests must be made:

- `SELECT u AS u FROM CmsUser u` with hydration mode object and array
- `SELECT u AS u, a AS a FROM CmsUser u JOIN u.address a` with hydration mode object and array

Additionally when you deprecate functionality all existing tests using the old behavior should be checked for regression tests and other complex tests and then be tested on the new API as well.

simPod:
Thank you, just started to dig through it now, your timing is perfect 👌

simPod:
@beberlei I added those tests but still cannot produce behaviour that would break current logic. Testing both object and array hydrations.

Overall I reworked the PR, now it seems kind of solid for me.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -94,6 +94,8 @@ public function __construct(EntityManagerInterface $em)
     /**
      * Initiates a row-by-row hydration.
      *
+     * @deprecated
```
beberlei: Can you also trigger a user error here? comments are usually not seen by users:

```php
@trigger_error('Method ' . __METHOD__ . '() is deprecated and will be removed in Doctrine ORM 3.0. Use getIterable() instead.', E_USER_DEPRECATED);
```

beberlei (changes requested):
Very thorough work on the testing, thank you!

I had one idea to use Generators instead of an iterator, I believe that would make this even nicer. and the deprecation needs a trigger_error as well. Then its finished from my POV.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -115,6 +117,31 @@ public function iterate($stmt, $resultSetMapping, array $hints = [])
...
+        $this->rsm   = $resultSetMapping;
+        $this->hints = $hints;
+
+        $evm = $this->em->getEventManager();
+
+        $evm->addEventListener([Events::onClear], $this);
+
+        $this->prepare();
+
+        return new RowByRowResult($this);
```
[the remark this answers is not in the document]
simPod: Yes, that came to my mind as well few hours after I marked this as ready for review. RowByRow iterator is not rewindable anyway so with generator we probably only gain. Will rework it, thanks!

greg0ire:
Can you please write docs for this?

simPod:
@greg0ire can you please guide me to a place where such docs should exist?

greg0ire:
@simPod : I think this place should be updated, for instance: https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/batch-processing.html#iterating-large-results-for-data-processing

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -859,6 +859,31 @@ public function iterate($parameters = null, $hydrationMode = null)
...
+    /**
+     * Executes the query and returns an iterable that can be used to incrementally
+     * iterate over the result.
+     *
+     * @param ArrayCollection|mixed[]|null $parameters    The query parameters.
+     * @param string|int|null              $hydrationMode The hydration mode to use.
+     *
+     * @return iterable<mixed>
+     */
+    public function getIterable($parameters = null, $hydrationMode = null) : iterable
```
greg0ire: I think a type declaration will not hurt:

```suggestion
    public function getIterable(?iterable $parameters = null, $hydrationMode = null) : iterable
```

simPod: Unfortunately it would not comply with `setParameters()` that accepts `ArrayCollection|mixed[]` so `iterable` is no covariant with it. Or am I missing something?

greg0ire: I think `ArrayCollection|mixed[]` is iterable, isn't it? So it's compatible, I think…

simPod: This could be changed to work with `iterable` easily:

```diff
-       if ($parameters !== null && count($parameters) !== 0) {
+       if ($parameters !== null) {
            $this->setParameters($parameters);
        }
```

But I think that if I pass eg. something implementing `Iterator` it would not comply with `ArrayCollection<Parameter>` as it does not have its methods (`clear()` for example).

```php
    /**
     * The parameter map of this query.
     *
     * @var ArrayCollection|Parameter[]
     */
    protected $parameters;
```

greg0ire: Nothing tells you that an iterable that is passed will not be countable, so there is nothing to change here IMO. Using `iterable` is not a sufficient contract for using `count()`, but it's better than no type declaration at all I think.

simPod: You can't `count()` `iterable`. For that I'd also have to check for `is_countable()` but that's available since 7.3 so it's not applicable anyway. 

But still, the main issue is incompatibility between `ArrayCollection` and `iterable`. You cannot assign `iterable` to `ArrayCollection`.

https://phpstan.org/r/90dedd75-2907-46fc-897e-1d368d8dc75c

```php
<?php

declare(strict_types=1);

/**
 * @template K
 * @template V
 */
final class X
{
    /** @var ArrayIterator<K, V> */
    private $iterator;

    public function __construct()
    {
        $this->iterator = new ArrayIterator([]);
    }

    /** @param iterable<K, V> $x */
    public function setIterable(iterable $x) : void
    {
        $this->iterator = $x;
    }
}
```

greg0ire: I was not suggesting you change the phpdoc, I was suggesting you do just this, which phpstan is fully OK with: https://phpstan.org/r/c20675d0-8b8c-4ba6-8ca3-5812bb371265

simPod: Ah ok, now it clicked. Thanks

greg0ire:
Please kindly squash your commits together.

How to do that?

1. `git rebase -i origin/master`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

simPod:
@greg0ire should I also get rid of #662a71b08aa7bf60e989b2ae476f3ad178c0a1be? The build won't pass without it tho.

greg0ire:
No keep it, it's good like this, don't squash it with the other one… I only meant to get rid of the commit messages like "Fix me" or just ":)" :sweat_smile:

greg0ire (dismissed):

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -859,6 +859,31 @@ public function iterate($parameters = null, $hydrationMode = null)
...
+     *
+     * @return iterable<mixed>
+     */
+    public function getIterable(?iterable $parameters = null, $hydrationMode = null) : iterable
+    {
+        if ($hydrationMode !== null) {
+            $this->setHydrationMode($hydrationMode);
+        }
+
+        if ($parameters !== null && count($parameters) !== 0) {
```
SenseException: I'm not sure about using count here for an `iterable`. The doc comment reduces the type to `ArrayCollection` and `array`, but technically every iterable can be injected. Should an extra complexity be added with `iterator_count()`?

simPod: AFAIK `iterator_count` takes only `\Iterator` and not `iterable`. 

But you're correct that it's technically possible to pass every `iterable` and then it would crash. However, removing that type hint would have same effect, right? It would allow to pass even more types and then we would have to if-check for proper type anyway. So I still think the proper parameter type is defined in PHPDoc and `iterable` typehint is just kind of weak guarantee.

WDYT? I don't have a preferred way to handle it but would like to avoid something like `if not instance of arraycollection and is not array and is not null then throw`

SenseException: `?iterable` is fine, an improvement and omitting the type leads to the mentioned type misbehaviour, which is why I propose to add complexity and use both counts. `Traversable` can't be used alone without `Iterator` and `IteratorAggregate` for an instantiable custom class and even `IteratorAggregate` can be used in `iterator_count()` as seen in this example: https://3v4l.org/8j5mf

`count()` on an iterable instance implementing `Countable` should be preferred before `iterator_count()` to keep more BC to the previous ORM versions, if there's no argument against this suggestion.

Disadvantage: `setParameters()` itself, which is called there, checks if the set parameters variable is an array. In case of an `array`, it creates an `ArrayCollection` instance and set the elements of the array in there. In the current implementation of `setParameters()` it would set any other iterable to `$this->parameters`. Sure, it also did it in the versions before as long as the parameters were countable. Does `setParameters()` need to be adapted too? This PR would become bigger if that method is also going to be adapted. :thinking:

SenseException: I tend to keep `setParameters()` as it is for now and focus only on `getIterable()`.

simPod: @SenseException is this what you meant?

```php
        if ($parameters !== null
            && (
                (($parameters instanceof Countable || is_array($parameters)) && count($parameters) !== 0)
                || ($parameters instanceof Traversable && iterator_count($parameters) !==0)
            )
        ) {
```

SenseException:
``` php
        if ($parameters !== null
            && (
                (($parameters instanceof Countable || is_array($parameters)) && count($parameters) !== 0)
                || iterator_count($parameters) !== 0
            )
        ) {
```

Too bad we don't have `is_countable()` in the lowest supported PHP version, but I guess you could move `is_array()` and the `Countable` check into a private method. Because of the `?iterable` type hint, all that is left are iterators after the array check. What do you think about this?

greg0ire (dismissed):

SenseException (dismissed):
Thank you.

simPod:
I rebased onto current master and squashed commits. I suppose now it should be complete, would appreciate any feedback/approvals ;) Thanks!

greg0ire (dismissed):

SenseException (dismissed):

beberlei:
@simPod fyi i am going to merge this for 2.8, but i have one further question while triaging tickets, if you have time maybe to check if your new implementation is prone to this problem: https://github.com/doctrine/orm/issues/3238 and other referenced issues that report undefined index

simPod:
@beberlei it did not but I managed to fix it, will push tmrw. Can also do for `iterable()` I suppose. Should I?

beberlei:
@simPod thank you so much!

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -115,6 +126,51 @@ public function iterate($stmt, $resultSetMapping, array $hints = [])
...
+
+            if ($row === false || $row === null) {
+                $this->cleanup();
+
+                break;
+            }
+
+            $this->hydrateRowData($row, $result);
+
+            if (is_array($result)) {
```
ostrolucky, line 164: I don't understand why is this check here. When is it not an array?

ostrolucky (changes requested):

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -102,6 +108,11 @@ public function __construct(EntityManagerInterface $em)
      */
     public function iterate($stmt, $resultSetMapping, array $hints = [])
     {
+        @trigger_error(
+            'Method ' . __METHOD__ . '() is deprecated and will be removed in Doctrine ORM 3.0. Use getIterable() instead.',
```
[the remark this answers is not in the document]
simPod: I expected this to go to 2.x https://github.com/doctrine/orm/pull/7885#issuecomment-555274491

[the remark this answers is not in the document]
simPod: Pls discuss this and let me know how Doctrine want things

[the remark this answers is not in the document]
ostrolucky: I wasn't trying to imply it should be targetted to 3.0, I agree with others. It's just that your pull request is set to target master. This needs to change and branch rebased on 2.8.x

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -1088,4 +1120,10 @@ protected function getHash()
 
         return sha1($query . '-' . serialize($params) . '-' . serialize($hints));
     }
+
+    /** @param iterable<mixed> $subject */
+    private function isCountable(iterable $subject) : bool
```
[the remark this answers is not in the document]
simPod: I changed the target to 2.8.x

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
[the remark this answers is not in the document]
simPod: Changed the base, thx

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -94,6 +98,8 @@ public function __construct(EntityManagerInterface $em)
     /**
      * Initiates a row-by-row hydration.
      *
+     * @deprecated
```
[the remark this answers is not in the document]
simPod: I guess we should and was just forgotten.

tests/Doctrine/Tests/ORM/Functional/QueryTest.php
```diff
@@ -227,16 +234,24 @@ public function testIterateResultAsArrayAndParams() : void
...
+        );
+
+        $articles = $query->getIterable(
+            new ArrayCollection([new Parameter(1, 'Doctrine 2')]),
+            Query::HYDRATE_ARRAY
+        );
+
+        $found = [];
+
+        foreach ($articles as $article) {
```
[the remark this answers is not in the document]
simPod: I kept the logic as above. Will change both.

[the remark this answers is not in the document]
simPod: ah no, can use it above but not here as I can't pass `iterable` to `iterator_to_array()`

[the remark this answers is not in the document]
SenseException: `[...$iterable]` :clown_face:
But that's not possible with the currently supported PHP versions.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
SenseException: Hard to say with all these references of what's supposed to be an array. In the `ArrayHydrator` there's also this:

https://github.com/doctrine/orm/blob/6dc46e78ccf4fe75c80884e7abe41b811f2204b6/lib/Doctrine/ORM/Internal/Hydration/ArrayHydrator.php#L253-L255

and code like `$this->resultPointers[$dqlAlias] =& $coll;`

It's hard to tell if `null` or anything else would be possible.

simPod: I ran test suite and it was always an array 🤔

ostrolucky: Yeah let's go with assumption it's always array. That's the only way how to verify this.

tests/Doctrine/Tests/ORM/Functional/QueryTest.php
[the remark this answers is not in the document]
simPod: But I guess I can call the `iterableToArray` method from my utility class in tests

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -859,6 +861,36 @@ public function iterate($parameters = null, $hydrationMode = null)
...
+    public function getIterable(?iterable $parameters = null, $hydrationMode = null) : iterable
+    {
+        if ($hydrationMode !== null) {
+            $this->setHydrationMode($hydrationMode);
+        }
+
+        if ($parameters !== null
+            && (
+                ($this->isCountable($parameters) && count($parameters) !== 0)
+                || iterator_count($parameters) !== 0
```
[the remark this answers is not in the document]
simPod: This allows parameters to be set in advance and not to be overwritten by calling this method. Eg. see `ClassTableInheritanceTest::testLazyLoading2`. So `setParameters` should not be called every time.

I've changed nullability in the last commit, not sure if api is better this way tho. Passing empty iterable seems less explicit to me.

ostrolucky:
LGTM so far, except those unfortunate CS issues. Looks like some of those are bugs of sniffs...

simPod:
@ostrolucky should be fixed except those bugy ones

ostrolucky:
They are solvable though: `iterable<mixed>` -> `mixed[]`. Second sniff error is correct I believe, just some spaces need to be removed, even if it didn't confirm cs before too.

simPod:
Though `mixed[]` is incorrect as it's not `array`. I'd rather keep correct type hint and upgrade CS later than the other way. This bug is already fixed in CS, I remember reporting it long time ago.

> Second sniff error is correct I believe, just some spaces need to be removed, even if it didn't confirm cs before too.

This is the second one, how to fix it by removing spaces?

image

I see only 3 errors of `does not specify type hint for items of its traversable` type in diff reported by GA

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -930,6 +950,33 @@ public function iterate($parameters = null, $hydrationMode = null)
         return $this->_em->newHydrator($this->_hydrationMode)->iterate($stmt, $rsm, $this->_hints);
     }
 
+    /**
+     * Executes the query and returns an iterable that can be used to incrementally
+     * iterate over the result.
+     *
+     * @param ArrayCollection|mixed[]|null $parameters    The query parameters.
```
faizanakram99, line 957:
```suggestion
     * @param ArrayCollection|mixed[] $parameters    The query parameters.
```

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -954,7 +954,7 @@ public function iterate($parameters = null, $hydrationMode = null)
      * Executes the query and returns an iterable that can be used to incrementally
      * iterate over the result.
      *
-     * @param ArrayCollection|mixed[]|null $parameters    The query parameters.
```
faizanakram99: wow, that was fast :smile:

greg0ire:
Letting 73f0d893d493105a0beb996135714bae6564e599 land on 2.8 should fix the build

greg0ire:
I'm working on that here: https://github.com/doctrine/orm/pull/8240

greg0ire:
There you go! You now have comments in the changes tab you can address :)

simPod:
@greg0ire these seem to be invalid

> @return annotation of method ... does not specify type hint for items of its traversable return value.

greg0ire:
Apparently it's due to `SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableReturnTypeHintSpecification`

```
  962 | ERROR | [ ] @return annotation of method \Doctrine\ORM\AbstractQuery::getIterable() does not specify type hint for items of its traversable return value.
      |       |     (SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableReturnTypeHintSpecification)
```

simPod:
Yes, the old version is bugged. So I guess there's no change needed from me?

greg0ire:
The check is required, so you will have to try an upgrade I suppose.

simPod:
Yes, I remember the old issue https://github.com/slevomat/coding-standard/issues/251

ostrolucky (approved):

SenseException (approved):

greg0ire (approved):

beberlei:
Having a purely cosmetic thought here, if the naming "toIterable" would be preferrable over "getIterable".

ostrolucky:
I agree. Feel free to change in followup PR.

simPod:
What do you think about returning Generator (more specific return type) instead / in addition to the current getIterable() ?

ostrolucky:
That would be quite limiting for no reason

simPod:
Why? It's contra variant, what does it limit then?

greg0ire:
No, return types are covariant in php: https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters

simPod:
yes, covariant I meant. So what's the issue? `: Generator` is still compliant with `:iterable`

🤨

greg0ire:
The method is not final, which means subclasses can override it but would be forced to return a `Generator`

simPod:
We can add `final` then. `final public function toGenerator() : Generator` 🤔 Though `toIterable()` can be kept as well.

greg0ire:
If we name it `toIterable()` why do we want to commit (as in make a BC-promise) to return a `\Generator`?

simPod:
The point was to have both `toIterable()` and `toGenerator()` so consumers would be able to choose. Eg. if they choose Generator, they can rely on it not being "rewindable".

greg0ire:
:thinking: how can you rely on the absence of a feature?

simPod:
I guess I can statically analyse that the Generator value is iterated more than once

greg0ire:
Ah so that would help detect misuses, I see… that's a good point. I guess you can make a PR, it should draw more attention to the discussion.

PR #7890, opened 2019-11-05
Set of tests for paginator with entities where id is custom object

akorz:
See more #7837

### DIFF
--- a/tests/Doctrine/Tests/Models/CMSCustomId/CmsAddress.php
+++ b/tests/Doctrine/Tests/Models/CMSCustomId/CmsAddress.php
@@ -1,0 +1,250 @@
+<?php
+
+namespace Doctrine\Tests\Models\CMSCustomId;
+
+/**
+ * CmsAddress
+ *
+ * @author Roman S. Borschel
+ * @Entity
+ * @Table(name="cms_addresses_customid")
+ *
+ * @NamedNativeQueries({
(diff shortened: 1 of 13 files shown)


akorz:
@Ocramius Why? Do you have tests for paginator in your master branch for entities where id is custom object? No. I added them. Along that tests you can see 1 error. That's it. Let's imaging, I added just 1 test. But how can you be sure that all other things with paginator are still ok?

Ocramius:
I had added them to GH7820 if I remember correctly?

The problem is that a lot of unrelated stuff got committed here, which makes it really hard to understand which detail to look for.

akorz:
Do you mean #7820 is a reason of error? Yes. You tried to create patch https://github.com/doctrine/orm/pull/7865 but with no success in my case.

Please check Paginator tests in 2.6 branch. You can see a lot of entities are related with that tests. I created similar entities with custom id. Thats why diff is huge. But if it's necessary for getting things done what else should I do? :)

akorz:
@Ocramius Hello, did you have any chance to check it out? Thanks.

Ocramius:
Heya, as mentioned above, the diff is too big to evaluate properly. If you take #7821, for example, I tried isolating everything into GH7820Test

ostrolucky:
continued in https://github.com/doctrine/orm/pull/7903

PR #7894, opened 2019-11-14
Fix boolean properties default value when generating entities.

TomckySan:
Generating entities by a generate entity tool, entity properties are set "1" that default value, despite field type is boolean.  
So, if a field type is boolean and a default value is "1", replace to "true".

### DIFF
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -26,6 +26,7 @@ use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use const E_USER_DEPRECATED;
 use function str_replace;
 use function trigger_error;
+use function var_export;
 
 /**
  * Generic class used to generate PHP5 entity classes from ClassMetadataInfo instances.
@@ -1331,7 +1332,7 @@ public function __construct(<params>)
             $defaultValue = '';
             if (isset($fieldMapping['options']['default'])) {
                 if ($fieldMapping['type'] === 'boolean' && $fieldMapping['options']['default'] === '1') {
-                    $defaultValue = " = true";
+                    $defaultValue = ' = true';
                 } else {
                     $defaultValue = ' = ' . var_export($fieldMapping['options']['default'], true);
                 }
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -1328,9 +1328,17 @@ public function __construct(<params>)
                 continue;
             }
 
+            $defaultValue = '';
+            if (isset($fieldMapping['options']['default'])) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -1322,9 +1322,17 @@ protected function generateEntityFieldMappingProperties(ClassMetadataInfo $metad
                 continue;
             }
 
+            $defaultValue = '';
+            if (isset($fieldMapping['options']['default'])) {
+                if ($fieldMapping['type'] === 'boolean' && $fieldMapping['options']['default'] === '1') {
```
Ocramius: I'd rather say that the default value should be passed through the DBAL type conversion system before being used in default value mapping, so this patch fixes the symptom, not the cause of the issue.

Ocramius (changes requested):

lib/Doctrine/ORM/Tools/EntityGenerator.php
TomckySan: Thanks for your reply!  
Okay. I'll try to fix the DBAL type conversion system.

TomckySan: This means that this patch is unacceptable, right?

SenseException: DBAL contains the logic of the different types that are also used for the ORM entities. It should provide you with the functionality to get `true`.

SenseException (changes requested):
Next to the mentioned requested changes, please also fix the coding style issues: https://travis-ci.org/doctrine/orm/jobs/612435677

lcobucci:
Since 2.7.0 has been released the 2.6.x isn't maintained anymore. Would please rebase this branch on top of 2.7 to sync things properly?

lib/Doctrine/ORM/Tools/EntityGenerator.php
beberlei: The mentioned API can be accessed this way:

```php
$defaultValue = Type::get($fieldMapping['type'])->convertToPhpValue($fieldMapping['options']['default']);
$defaultValue = ' = ' . var_export($defaultValue, true);
```

TomckySan: Thanks for your comment.  

The `convertToPHPValue` method is had to pass `AbstractPlatform` value as a second argument parameter, isn't it?  
https://www.doctrine-project.org/api/dbal/2.8/Doctrine/DBAL/Types/Type.html#method_convertToPHPValue

I don't know how to get a `AbstractPlatform` value in `EntityGenerator` class.

beberlei (approved):
Since there is no access to `AbstractPlatform` in entity generator this is indeed a sane fix.

beberlei:
@TomckySan Thank you for contributing this!

TomckySan:
@beberlei Thank you for merging, too!

PR #7895, opened 2019-11-14
Use the hydrationMode provided in function call

Piskvor:
- less stateful
- fixes NoResultException thrown
    from AbstractQuery::getSingleScalarResult() invocation


Ocramius:
> fixes `NoResultException` thrown from `AbstractQuery::getSingleScalarResult()` invocation

Needs added tests showing the changed behavior

PR #7901, opened 2019-11-15
Add deprecation warnings for 2.7.x

lcobucci:
This recovers the great work done by @Majkl578 in #6869 but without using phpunit-bridge stuff to do the assertions.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,103 @@
 # Upgrade to 2.7
 
+## Deprecated code generators and related console commands
+ 
+These console commands have been deprecated:
+
+ * `orm:convert-mapping`
+ * `orm:generate:entities`
+ * `orm:generate-repositories`
+
+These classes have been deprecated:
+
+ * `Doctrine\ORM\Tools\EntityGenerator`
+ * `Doctrine\ORM\Tools\EntityRepositoryGenerator`
+
+Whole Doctrine\ORM\Tools\Export namespace with all its members have been deprecated as well.
+
+## Deprecated `Doctrine\ORM\Proxy\Proxy` marker interface
+
+Proxy objects in Doctrine 3.0 will no longer implement `Doctrine\ORM\Proxy\Proxy` nor
+`Doctrine\Common\Persistence\Proxy`: instead, they implement
+`ProxyManager\Proxy\GhostObjectInterface`.
+
+These related classes have been deprecated:
+
+ * `Doctrine\ORM\Proxy\ProxyFactory`
+ * `Doctrine\ORM\Proxy\Autoloader` - we suggest using the composer autoloader instead
+ 
+These methods have been deprecated:
+
+ * `Doctrine\ORM\Configuration#getAutoGenerateProxyClasses()`
+ * `Doctrine\ORM\Configuration#getProxyDir()`
+ * `Doctrine\ORM\Configuration#getProxyNamespace()`
+
+## Deprecated `Doctrine\ORM\Version`
+
+The `Doctrine\ORM\Version` class is now deprecated and will be removed in Doctrine 3.0:
+please refrain from checking the ORM version at runtime or use
+[Ocramius/PackageVersions](https://github.com/Ocramius/PackageVersions/).
+
+## Deprecated `EntityManager#merge()` and `EntityManager#detach()` methods
+
+Merge and detach semantics were a poor fit for the PHP "share-nothing" architecture.
+In addition to that, merging/detaching caused multiple issues with data integrity
+in the managed entity graph, which was constantly spawning more edge-case bugs/scenarios.
+
+The following API methods were therefore deprecated:
+
+* `EntityManager#merge()`
+* `EntityManager#detach()`
+* `UnitOfWork#merge()`
+* `UnitOfWork#detach()`
+
+Users are encouraged to migrate `EntityManager#detach()` calls to `EntityManager#clear()`.
+
+In order to maintain performance on batch processing jobs, it is endorsed to enable
+the second level cache (http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/second-level-cache.html)
+on entities that are frequently reused across multiple `EntityManager#clear()` calls.
+
+An alternative to `EntityManager#merge()` will not be provided by ORM 3.0, since the merging
+semantics should be part of the business domain rather than the persistence domain of an
+application. If your application relies heavily on CRUD-alike interactions and/or `PATCH`
+restful operations, you should look at alternatives such as [JMSSerializer](https://github.com/schmittjoh/serializer).
+
+## Extending `EntityManager` is deprecated
+
+Final keyword will be added to the `EntityManager::class` in Doctrine 3.0 in order to ensure that EntityManager
+ is not used as valid extension point. Valid extension point should be EntityManagerInterface.
+
+## Deprecated `EntityManager#flush($entity)` and `EntityManager#flush($entities)`
+
+If your code relies on single entity flushing optimisations via
+`EntityManager#flush($entity)`, the signature has been changed to
+`EntityManager#flush()`.
+
+Said API was affected by multiple data integrity bugs due to the fact
+that change tracking was being restricted upon a subset of the managed
+entities. The ORM cannot support committing subsets of the managed 
+entities while also guaranteeing data integrity, therefore this
+utility was removed.
+
+The `flush()` semantics will remain the same, but the change tracking will be performed
+on all entities managed by the unit of work, and not just on the provided
+`$entity` or `$entities`, as the parameter is now completely ignored.
+
+The same applies to `UnitOfWork#commit($entity)`, which will simply be
+`UnitOfWork#commit()`.
+
+If you would still like to perform batching operations over small `UnitOfWork`
+instances, it is suggested to follow these paths instead:
+
+ * eagerly use `EntityManager#clear()` in conjunction with a specific second level
+   cache configuration (see http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/second-level-cache.html)
+ * use an explicit change tracking policy (see http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/change-tracking-policies.html)
+
+## Deprecated `YAML` mapping drivers.
+
+If your code relies on `YamlDriver`  or `SimpleYamlDriver`, you **MUST** change to
+annotation or XML drivers instead.
+
 ## Deprecated: `Doctrine\ORM\EntityManagerInterface#copy()`
 
 Method `Doctrine\ORM\EntityManagerInterface#copy()` never got its implementation and is deprecated.
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "c3d70a08d75cd3995dd4c72864f3c880",
+    "content-hash": "084cd36de51fd4aa52c912b3e9c7c799",
     "packages": [
         {
             "name": "doctrine/annotations",
@@ -76,16 +76,16 @@
         },
         {
             "name": "doctrine/cache",
-            "version": "v1.8.0",
+            "version": "1.9.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/cache.git",
-                "reference": "d768d58baee9a4862ca783840eca1b9add7a7f57"
+                "reference": "c15dcd24b756f9e52ea7c3ae8227354f3628f11a"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/cache/zipball/d768d58baee9a4862ca783840eca1b9add7a7f57",
-                "reference": "d768d58baee9a4862ca783840eca1b9add7a7f57",
+                "url": "https://api.github.com/repos/doctrine/cache/zipball/c15dcd24b756f9e52ea7c3ae8227354f3628f11a",
+                "reference": "c15dcd24b756f9e52ea7c3ae8227354f3628f11a",
                 "shasum": ""
             },
             "require": {
@@ -96,7 +96,7 @@
             },
             "require-dev": {
                 "alcaeus/mongo-php-adapter": "^1.1",
-                "doctrine/coding-standard": "^4.0",
+                "doctrine/coding-standard": "^6.0",
                 "mongodb/mongodb": "^1.1",
                 "phpunit/phpunit": "^7.0",
                 "predis/predis": "~1.0"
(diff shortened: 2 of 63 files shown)


Ocramius (approved):
Build failures affect MySQL's quoting of collations: unrelated to this patch.

composer.lock
```diff
@@ -141,46 +141,52 @@
...
+                "riak",
+                "xcache"
             ],
-            "time": "2018-08-21T18:01:43+00:00"
+            "time": "2019-11-11T10:31:52+00:00"
         },
         {
             "name": "doctrine/collections",
-            "version": "v1.6.2",
+            "version": "v1.5.0",
```
Ocramius: Some downgrades happened in here - @lcobucci do we want to fix them, or is this stuff for another patch?

lcobucci: The composer.lock was previously edited without setting the platform to the lowest supported version.

michaljusiega:
Somebody Can tell me, why you want to remove that a importants methods like clear/merge/detach in EM? It will be some replacement methods? 

Second question. PHPDoc tag @deprecated mean from which version of package the specified method will be removed. So deprecated and removed in 3.0 ? I dont think so.

Ocramius:
> Somebody Can tell me, why you want to remove that a importants methods like clear/merge/detach in EM? It will be some replacement methods?

Mostly complexity and edge cases that are better solved via simpler `EntityManager#clear()` and smaller `UnitOfWork` scopes. We have closed dozens and dozens of issues related to `EntityManager#merge()`, which is something that was initially built after the JPA design, but is a bad fit for PHP's share-nothing architecture, and also introduces broken entity graphs in the `UnitOfWork`. This patch is mostly about documenting deprecations that were already agreed upon in @doctrine/doctrinecore. 

 > PHPDoc tag `@deprecated` mean from which version of package the specified method will be removed. So deprecated and removed in 3.0 ? I dont think so.

The deprecated API will be removed (and we also already did remove a lot of it) in `master`, which will become `3.0.0`

michaljusiega:
I will never understand your approach. You don't have a good standard of changes that you can change for anything ... in my humble opinion I think that these methods are very important, they work very well and getting rid of them will prevent you from raising to version 3.0 - put yourself in the places of many programmers for whom the problem there will be a raise to 3.0. In my case, rather not possible because I use merge / clear / detach myself - and since you have stated in PHPDoc that there will be no alternative methods - praise to you!

In Symfony, if they remove something, they at least give a replacement method.

Ocramius:
In many of the removed endpoints, the alternative is effectively "stop doing X", where "X" is identified as a problem.

Yes, this means that upgrading will require design changes in your code: that's something that was considered.

We don't take decisions in a rushed nor uninformed way: this has been discussed and agreed upon when keeping in mind that there are thousands (potentially millions) of consumers of the library.

michaljusiega:
Okey. Last one. Can you show and explain me a equivalent code to replace a merge/clear/detach methods? Thanks.

UPGRADE.md
```diff
@@ -1,5 +1,103 @@
...
+## Deprecated `Doctrine\ORM\Proxy\Proxy` marker interface
+
+Proxy objects in Doctrine 3.0 will no longer implement `Doctrine\ORM\Proxy\Proxy` nor
+`Doctrine\Common\Persistence\Proxy`: instead, they implement
+`ProxyManager\Proxy\GhostObjectInterface`.
+
+These related classes have been deprecated:
+
+ * `Doctrine\ORM\Proxy\ProxyFactory`
+ * `Doctrine\ORM\Proxy\Autoloader` - we suggest using the composer autoloader instead
```
stof: using the composer autoloader for generated proxy classes ? This won't work well. One of the point of the `Doctrine\ORM\Proxy\Autoloader` is that it would generate the proxy class if needed (allowing to unserialize your entities even in case the proxies are not generated yet, which is very handy during dev where you probably generate the proxies on demand rather than dumping them).

Ocramius:
* `clear()` will stay as-is: what's deprecated is clearing sub-sections of the entity graph (not supported, since it breaks the `UnitOfWork` internal state). If you need `EntityManager#clear()`, clear the entire `UnitOfWork`
 * `merge()` is no longer supported: instead, if you know that you have an in-memory object that is in sync with DB state, use `UnitOfWork#registerManaged()`. That's very close to internal API though, so that's discouraged too.
 * `detach()` is not to be used: if you need to store information such as which user is logged in, do not store the entire `User` entity (for example in a session), but rather only the user identifier.

There is no *equivalent* code: we removed these methods because people running into very bad bugs due to very very complex state permutations that shouldn't happen in first place.

UPGRADE.md
```diff
@@ -1,5 +1,103 @@
...
+
+ * `Doctrine\ORM\Configuration#getAutoGenerateProxyClasses()`
+ * `Doctrine\ORM\Configuration#getProxyDir()`
+ * `Doctrine\ORM\Configuration#getProxyNamespace()`
+
+## Deprecated `Doctrine\ORM\Version`
+
+The `Doctrine\ORM\Version` class is now deprecated and will be removed in Doctrine 3.0:
+please refrain from checking the ORM version at runtime or use
+[Ocramius/PackageVersions](https://github.com/Ocramius/PackageVersions/).
```
stof: I suggest writing the text as `ocramius/package-versions` (the package name) rather than `Ocramius/PackageVersions` (a substring of the repo URL)

UPGRADE.md
```diff
@@ -1,5 +1,103 @@
...
+on entities that are frequently reused across multiple `EntityManager#clear()` calls.
+
+An alternative to `EntityManager#merge()` will not be provided by ORM 3.0, since the merging
+semantics should be part of the business domain rather than the persistence domain of an
+application. If your application relies heavily on CRUD-alike interactions and/or `PATCH`
+restful operations, you should look at alternatives such as [JMSSerializer](https://github.com/schmittjoh/serializer).
+
+## Extending `EntityManager` is deprecated
+
+Final keyword will be added to the `EntityManager::class` in Doctrine 3.0 in order to ensure that EntityManager
```
stof: Currently, DoctrineBundle generates a proxy class (using `ocramius/proxy-manager`) for the entity manager, to allow resetting the entity manager (allowing to inject the entity manager in other services without having to kill the whole PHP process to rebuild all services in case an error occurs in the ORM). Will Doctrine ORM 3.0 allow reopening an entity manager instance (which of course involve it will be cleared) ? If no, this change means that using the EntityManager in a service used by a long-running process (a RabbitMQ consumer for instance) won't be possible anymore, which is quite impactful.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -37,6 +37,8 @@
  * @author Giorgio Sironi <piccoloprincipeazzurro@gmail.com>
  * @author Marco Pivetta  <ocramius@gmail.com>
  * @since 2.0
+ *
+ * @deprecated 3.0 This class is being removed from the ORM and won't have any replacement
```
stof: Well, actually, deprecating it in 2.7 looks wrong there, as it is impossible to avoid using it in 2.7 (as the migration to proxy-manager is not available in 2.7)

stof:
`clear` is not deprecated. Partial clears (clearing only some entity classes) is, because it is unreliable (if a non-cleared entity has a relation pointing to a cleared entity, you might have managed entities referencing cleared objects, which is broken). And once you get there, you cannot recover things in any other way than clearing everything as the unit of work became unusable.

for `merge`, the solution would be to do a `find` and then build your own logic for how the state of your object and the state of the entity managed by Doctrine should be merged together

stof:
@Ocramius my existing use case for detach is a process where I'm iterating over object results for a big result set, and detaching objects once I don't need them anymore (this process is never flushing the unit of work btw, I'm only reading in there). This detaching allows to remove the memory reference to this object from the unit of work so that the object can be destructed by PHP.
But I have some other objects (some of them being used by relations in this object btw) for which I still benefit from the entity map (and for which I'm fine with keeping them in the identity map until the end of the process, because they have a much lower cardinality).

What is the expected replacement for such usage of `detach` ? Would it be worth having a way to run a query which would hydrate results without registering the root entity of the results as managed by the unit of work if it is not already in it (this would of course need to be opt-in) ?
And if yes, would 2.7 provide this feature as well, alongside the deprecation of `detach` ?

UPGRADE.md
lcobucci: The architecture in v3 is quite different. The design we have been discussing is that each transaction will use a different UoW instance. This promotes better isolation and removes the need for em to be closed (in theory). There are still things to be addressed on this but that's the gist.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
lcobucci: That's completely fair. I'll address that in another PR, thanks

UPGRADE.md
lcobucci: Good point, thanks

UPGRADE.md
Majkl578: I disagree, you can just use `eval()` in dev.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
Majkl578: Deprecating removed stuff with no replacement (which is exactly the case) is absolutely fine IMO.

lcobucci: I completely misread the comments. What I thought was said was that the version is wrong in the deprecation notice (it's being deprecated in 2.7 not 3.0).

About it being deprecated now, it's completely fine since it's used internally and no error is being triggered.

lcobucci:
> What is the expected replacement for such usage of detach ? Would it be worth having a way to run a query which would hydrate results without registering the root entity of the results as managed by the unit of work if it is not already in it (this would of course need to be opt-in) ?
And if yes, would 2.7 provide this feature as well, alongside the deprecation of detach ?

@stof `Query#iterate()` + `EM#clear()` addresses that scenario, already in v2.6. Like:

```php
foreach ($query->iterate() as $result) {
    // Do whatever is needed with $result
    // ...

    // Free resources:
    unset($result[0]);
    $this->_em->clear();
}
```

UPGRADE.md
stof: well, if an em does not get closed on failure, then that's fine.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
stof: well, "deprecated" means you should stop using it because it will be removed. But here, there is no way to stop using it.

stof:
@lcobucci no it does not, because it also clears any other entity that might have been loaded earlier for this processing (and SLC is not an option for me, because some of these objects are also modified by direct SQL in some batch processing, so the SLC would contain the wrong info unless I also manage it myself which seems insane).

michaljusiega:
I try myself to replace a detach method in my projects but it's hard :/

lcobucci:
@stof have you considered enabling the SLC using the `ArrayCache` provider to contain the cached data to only the execution lifecycle?

That almost fully mimics the behaviour of the identity map, the only missing piece is the hydration. However, hydrating in an iteration like that should be quite cheap.

Our main goal with that move is to protect the integrity of the UoW, which is an internal component of the ORM. No application should rely on it for optimisation purposes, caching is meant to address that concern.

UPGRADE.md
stof: The strategy being used is totally unrelated to that. The `Doctrine\ORM\Proxy\Autoloader` is about calling the proxy generator, not about changing how the proxy class is defined.

When the ORM wants to instantiate a proxy class, it will take care of generating it first. The autoloader is solving the case where you call `unserialize()`, and the serialized entity was actually an initialized proxy (and so the proxy class name appears in the serialized string, not the entity class name).

stof:
@lcobucci but in my case, I want to stop tracking only my entities in the big list that I process. For other objects that I use in my processing, keeping them in the identity map is a good idea, as it ensures that I can use `===` being 2 variables holding them (if I clear, the second time might rehydrate a different instance).
What about my suggestion of an opt-in way to hydrate entities without adding them in the entity map for a specific query and entity (related objects would still be taken from the entity map) ? That would not corrupt the UoW, as that object would not be part of it at all (removing the need to detach it), but I would still get an instance of my class in the code (allowing to use it in my typed APIs).
In the case where I use `detach`, I don't actually need to have the returned instance in the UoW. but I need to have them hydrated, and having their relations being taken from the identity map.

Ocramius:
> What about my suggestion of an opt-in way to hydrate entities without adding them in the entity map

That's a very interesting idea, and could be done in future for sure. As @lcobucci stated though: 

 > No application should rely on it for optimisation purposes, caching is meant to address that concern.

stof:
Well, I'm not relying on the identity map for optimization purposes. that's precisely why using an ArrayCache as SLC does not fit me. I need to rely on the identity for other entities than the one I'm detaching during my processing. clearing+ in-memory SLC will optimize the loading of data, but won't preserve identity.

beberlei:
@stof Read only sort of exists, but not as much as I remembered.

I thought the following was possible:

```php
$query->setHint(Query::HINT_READ_ONLY, true); 
```
But that was removed during beta. 

I added read only objects later, but they are required to be managed manually:

```php
$entityManager->getUnitOfWork()->markReadOnly(entity);
```

Essentially what we just need to make the query hint automatically set entities as read only.

stof:
@beberlei but they are still referenced in the identity map. My issue is not about the size of computing the change set (this code does not flush at all). It is about avoiding to keep unnecessary objects in memory after processing them is done.

thePanz:
In one of my project we are facing the same situation described by @stof in https://github.com/doctrine/orm/pull/7901#issuecomment-554479904, and loved the suggestion in https://github.com/doctrine/orm/pull/7901#issuecomment-555950316
> What about my suggestion of an opt-in way to hydrate entities without adding them in the entity map for a specific query and entity (related objects would still be taken from the entity map)? [..]

Is there any work being done in that direction that we can help with? How did you solve it @stof ?

dbu:
the usecase of @thePanz and me is that we need to read (potentially lots of) entities that we know we don't intend to change, and we also don't need doctrine to cache them for us. at the same time, we have entities that we want to update and flush.

if i read the code in the 2.9.x branch correctly, there are mainly 2 places that might matter to us: 
* the query cache that caches results based on the dbal query. this cache can be disabled per query (but as it is a cache, i assume it to discard things when it overflows, and all of it should be cheap)
* the uow identity map. the ObjectHydrator invokes UnitOfWork::createEntity, which makes the entity tracked. there are hints for this method, but there is no hint to mark that the entity should be created in detached mode. could it be an option to add a hint for that and respect it in createEntity? 

we are motivated to contribute the feature, but of course we first need to figure out if the maintainer consider it a use case they want to support. if so, i suggest we create a new issue to discuss where to implement the functionality.

stof:
> Is there any work being done in that direction that we can help with? How did you solve it @stof ?

I haven't solved it yet. For now, I'm still using the deprecated `detach` method.

dbu:
i chreated #8429 to further discuss the issue and think about solutions

PR #7904, opened 2019-11-16
Make sure composer files are valid

greg0ire:
The composer.lock is put under version control and it often happens to
be out of sync with the composer.json, which could lead to
hard-to-understand issues.
Using the --strict option here because we might as well aim for a
perfectly valid composer.json


greg0ire:
The build should fail. If it does, I will rebase over a commit that puts the composer.lock and composer.json in sync.

greg0ire:
Fails with:

> The lock file is not up to date with the latest changes in composer.json, it is recommended that you run `composer update`.

ostrolucky (approved):

lcobucci (approved):

PR #7907, opened 2019-11-18
Add missing imports of annotations.

michaljusiega:
Hello. 

It's my first PR for ORM project.
I found (I think) a some little issue with annotations. Using defualt new AnnotationReader(), we can get a error about missing import like: 

`Doctrine\Common\Annotations\AnnotationException : [Semantical Error] The annotation "@Enum" in property Doctrine\ORM\Mapping\Cache::$usage was never imported. Did you maybe forget to add a "use" statement for this annotation?`

This PR fix this on 2.6

### DIFF
--- a/lib/Doctrine/ORM/Mapping/AssociationOverride.php
+++ b/lib/Doctrine/ORM/Mapping/AssociationOverride.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\ORM\Mapping;
 
+use Doctrine\Common\Annotations\Annotation\Enum;
+
 /**
  * This annotation is used to override association mapping of property for an entity relationship.
  *
--- a/lib/Doctrine/ORM/Mapping/Cache.php
+++ b/lib/Doctrine/ORM/Mapping/Cache.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\ORM\Mapping;
 
+use Doctrine\Common\Annotations\Annotation\Enum;
+
 /**
  * Caching to an entity or a collection.
  *
--- a/lib/Doctrine/ORM/Mapping/ChangeTrackingPolicy.php
+++ b/lib/Doctrine/ORM/Mapping/ChangeTrackingPolicy.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\ORM\Mapping;
 
+use Doctrine\Common\Annotations\Annotation\Enum;
+
 /**
  * @Annotation
  * @Target("CLASS")
--- a/lib/Doctrine/ORM/Mapping/GeneratedValue.php
+++ b/lib/Doctrine/ORM/Mapping/GeneratedValue.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\ORM\Mapping;
 
+use Doctrine\Common\Annotations\Annotation\Enum;
+
 /**
  * @Annotation
  * @Target("PROPERTY")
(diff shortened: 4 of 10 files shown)


greg0ire:
I haven't made my mind about this PR yet, but there is no mention of that import in the docs: https://www.doctrine-project.org/projects/doctrine-annotations/en/latest/custom.html#enumerated-values

michaljusiega:
So we should leave it as it is still generating an exception?

greg0ire:
I haven't made my mind about this PR yet. Please provide a full stack trace, and tell us which version of the ORM and the annotations library you use. Maybe a recent update broke something? The files you propose to change are quite old, hence this hunch.

michaljusiega:
A complicated matter. Every ORM  update since 2.6.0 I have to add these lines of code to make the whole project live for me. It's difficult for me to determine stack-trace and the cause of the error.

But it probably wouldn't hurt to add a few lines of code that theoretically do not change anything directly in the project.

But what's up, it's not my project. I reported such a need, because I was tired of modifications in this library so that everything was okay.

Yeah, the files are old ... because you are messing with this ORM 3.0 :D

greg0ire:
> It's difficult for me to determine stack-trace and the cause of the error.

Why?

> But what's up, it's not my project. I reported such a need, because I was tired of modifications in this library so that everything was okay.

And that's great, thank you ! I hope you can understand that we want to fully understand why there is a need for what me merge though. I even think that as a user, you would be quite afraid of using Doctrine if we didn't try to get to the bottom of things, wouldn't you?

michaljusiega:
No, I'm not afraid of using Doctrine. I have written tests for this in my own projects for which this error occurs - so if I do not notice the error after the update by Composer, the tests will tell the truth.

I added a test, but there is another error.

Instead: 
`Doctrine\Common\Annotations\AnnotationException : [Semantical Error] The annotation "@Enum" in property Doctrine\ORM\Mapping\Cache::$usage was never imported. Did you maybe forget to add a "use" statement for this annotation`

got now:

`Doctrine\Common\Annotations\AnnotationException : [Semantical Error] The annotation "@Doctrine\Common\Annotations\Annotation\Enum" in property Doctrine\ORM\Mapping\Cache::$usage does not exist, or could not be auto-loaded.`

I'm confused :/

greg0ire:
:thinking: meanwhile, I had a look at this, and now my understanding is that we have been wrongly assuming 100% of people are using https://github.com/doctrine/annotations/blob/1.9/lib/Doctrine/Common/Annotations/SimpleAnnotationReader.php, when in fact they could be using https://github.com/doctrine/annotations/blob/1.9/lib/Doctrine/Common/Annotations/AnnotationReader.php

I have yet to understand why there are 2 classes and the difference(s) between the 2 (git doesn't help much this time, and the commit that introduced this file was really hard to find: https://github.com/doctrine/common/pull/49)

SenseException:
The `SimpleAnnotationReader` was deprecated and will be removed in later versions because of its "magic" kind of working: 

* https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/tutorials/getting-started.html#obtaining-the-entitymanager
* https://github.com/doctrine/orm/pull/7753#issuecomment-530869183

beberlei:
Why is the import needed for `@Enum` but not for the others like `@Annotation` and `@Target`? This feels like a fix for the symptom and not the problem.

michaljusiega:
So ... what are we doing with it? We close and forget? :)

PR #7908, opened 2019-11-18
Fix BC-break on underscore naming strategy

lcobucci:
We broke our BC promises on the last patch release by changing [how the underscore naming strategy parses values with numbers](https://github.com/doctrine/orm/issues/7855#issuecomment-555190431).

This makes it possible to configure whether or not to make the underscore naming strategy aware of numbers, keeping the old configuration as default value.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/UnderscoreNamingStrategy.php
+++ b/lib/Doctrine/ORM/Mapping/UnderscoreNamingStrategy.php
@@ -40,22 +40,26 @@ use function substr;
  */
 class UnderscoreNamingStrategy implements NamingStrategy
 {
-    private const DEFAULT_PATTERN        = '/(?<=[a-z])([A-Z])/';
-    private const PATTERN_FOR_PROPERTIES = '/(?<=[a-z0-9])([A-Z])/';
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Mapping/UnderscoreNamingStrategy.php
```diff
@@ -40,22 +40,26 @@
...
+    private $pattern;
+
     /**
      * Underscore naming strategy construct.
      *
-     * @param integer $case CASE_LOWER | CASE_UPPER
+     * @param int $case CASE_LOWER | CASE_UPPER
      */
-    public function __construct($case = CASE_LOWER)
+    public function __construct($case = CASE_LOWER, bool $numberAware = false)
```
greg0ire: We should probably deprecate not setting this to `true` and warn our users that the class will behave as if it were set to true in the next major version

lcobucci: Yeap, planned for `2.7.0` didn't want to add deprecations in a patch.

greg0ire (approved):

lcobucci:
@greg0ire thanks for the review!

PR #7909, opened 2019-11-19
Add deprecation messages

lcobucci:


jwage (approved):

lib/Doctrine/ORM/Mapping/UnderscoreNamingStrategy.php
```diff
@@ -58,6 +60,13 @@ class UnderscoreNamingStrategy implements NamingStrategy
      */
     public function __construct($case = CASE_LOWER, bool $numberAware = false)
     {
+        if (! $numberAware) {
+            @trigger_error(
+                'Creating ' . self::class . ' without making it number aware is deprecated and will be removed in Doctrine 3.0.',
```
alcaeus: Noticed this in other pull requests, we should be talking about Doctrine **ORM** 3.0 to avoid confusion with other Doctrine libraries.

For consistency, feel free to change this in a separate PR for a subsequent patch release, nothing to hold up 2.7.0 for.

alcaeus (approved):

lib/Doctrine/ORM/Mapping/UnderscoreNamingStrategy.php
lcobucci: Agreed, I almost introduced an inconsistency here. I can quickly address this before releasing.

lcobucci:
@jwage @alcaeus thanks for the review

Issue #7661, opened 2019-03-29
Cannot index-by related entity


crtl:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

#### Summary

When trying to index by a property that is a mapped relation the following error is thrown:

```
Cannot match on App\Entity\Entity::thirdEntity with a non-object value. 
Matching objects by id is not compatible with matching on an in-memory collection, which compares objects by reference.
```

#### Current behavior

Exception is thrown

#### How to reproduce

Create the following mapping:

```xml

<entity name="App\Entity\User">
    <id name="id" />
</entity>

<entity name="App\Entity\Event">
    <id name="id" />

    <one-to-many field="participants" target-entity="App\Entity\Event\Participant" mapped-by="event" index-by="user" fetch="EXTRA_LAZY" >
        <cascade>
            <cascade-all/>
        </cascade>
    </one-to-many>
</entity>

<entity name="App\Entity\Event\Participant">
    <id name="id" />

    <many-to-one field="user" target-entity="App\Entity\User" />
    <many-to-one field="event" target-entity="App\Entity\Event" />

</entity>
```

```php

class Event {

    protected $participants;

    public function __construct() {
        $this->participants = new ArrayCollection();
    }

    public function isParticipating($user) {
        $id = $user instanceof User ? $user->getId() : $user;

        return $this->participants->containsKey($id); //Throws exception, passing $user (entity) throws illegal offset
    }
}
```

#### Expected behavior

The participants should be correctly mapped by their related users id or the schema validator should throw an error when validating the schema (schema is valid).

andrews05:
Have you tried using the name of the join column, e.g. "user_id" instead of just "user"? This is working for me in v3.

andrews05:
What I can't do is use INDEX BY in a query in the same manner.
`SELECT p FROM Participant INDEX BY p.user_id`: "Class Participant has no field or association named related_id"
`SELECT p FROM Participant INDEX BY p.user`: "Invalid PathExpression. Must be a StateFieldPathExpression."

beberlei:
Fixed and targeted for 2.9.0

PR #7915, opened 2019-11-22
Add support for INDEX BY an associated entity

andrews05:
This allows specifying an association in the INDEX BY clause of a query
which will index by the association's join column.

Related to #7661.

Note I haven't added support for specifying the name of a related entity in an indexed association ("indexBy" annotation) as I couldn't see a good way of doing so. Using the name of the join column does work here though, so it may not be necessary.

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -883,9 +883,38 @@ class SqlWalker implements TreeWalker
      */
     public function walkIndexBy($indexBy)
     {
-        $pathExpression = $indexBy->simpleStateFieldPathExpression;
-        $alias          = $pathExpression->identificationVariable;
-        $field          = $pathExpression->field;
+        $pathExpr = $indexBy->singleValuedPathExpression;
+        $alias    = $pathExpr->identificationVariable;
+
+        /** @var Query\AST\PathExpression $pathExpr */
+        switch ($pathExpr->type) {
+            case AST\PathExpression::TYPE_STATE_FIELD:
+                $field = $pathExpr->field;
+                break;
+
+            case AST\PathExpression::TYPE_SINGLE_VALUED_ASSOCIATION:
+                // 1- the owning side:
+                //    Just use the foreign key, i.e. u.group_id
+                $fieldName   = $pathExpr->field;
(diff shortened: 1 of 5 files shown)


beberlei:
I like this very much

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -883,9 +883,39 @@ public function walkIdentificationVariableDeclaration($identificationVariableDec
...
+                $class       = $this->queryComponents[$alias]['metadata'];
+                $association = $class->getProperty($fieldName);
+
+                if (! $association->isOwningSide()) {
+                    throw QueryException::associationPathInverseSideNotSupported($pathExpr);
+                }
+
+                $joinColumns = $association->getJoinColumns();
+
+                // COMPOSITE KEYS NOT (YET?) SUPPORTED
```
beberlei: they can't be supported for this, so its ok to remove the comment. The exception is exactly showing users this is not supported.

beberlei (dismissed):

lib/Doctrine/ORM/Query/SqlWalker.php
andrews05: The switch here is copied from walkPathExpression(), including the comments. Happy to remove the comment if you like though.

beberlei: Yes i guess removing the comment makes sense then, also at the other location.

beberlei:
@andrews05 can you make the PR open against the 2.8 branch instead? This is not a feature that requires to wait for 3.0 release

beberlei:
You should also add a sentence to the index by section in this doc page that both single value field and assocations are supported: https://github.com/doctrine/orm/blob/master/docs/en/reference/dql-doctrine-query-language.rst

andrews05:
@beberlei I've made the changes suggested (removed comment, note in docs).
I opened another PR #7918 for the 2.8 branch. The code is a little different, so should I keep this one open as well?

beberlei:
@lcobucci what do you think w.r.t to merge strategy of this feature? how do you upmerge changes from 2.8 to master?

lcobucci:
@beberlei we're going to organise things so that we can merge things up to master more easily. I'd say we can keep this in 2.8.x only.

PR #7918, opened 2019-11-24
Add support for INDEX BY an associated entity (2.8.x)

andrews05:
This allows specifying an association in the INDEX BY clause of a query
which will index by the association's join column.

Related to #7661.


beberlei:
@andrews05 awesome :-) Can you fix the coding standard violatoins listed here? 

https://travis-ci.org/doctrine/orm/jobs/616171121?utm_medium=notification&utm_source=github_status

This should fix automatically when you run `php vendor/bin/phpcbf lib/Doctrine/ORM/Query/SqlWalker.php` and commit the changes.

andrews05:
I did run phpcs but it seems the <3.0 branches don't really adhere to the coding standard? It showed 301 errors in that file alone and phpcbf "failed to fix". I could fix the issues in my changes manually, if you think it's worth it? (Keeping in my mind that code is copied from elsewhere in the file)

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -854,9 +856,41 @@ public function walkIdentificationVariableDeclaration($identificationVariableDec
      */
     public function walkIndexBy($indexBy)
     {
-        $pathExpression = $indexBy->simpleStateFieldPathExpression;
```
greg0ire: Why did you rename this variable?

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -854,9 +856,41 @@ public function walkIdentificationVariableDeclaration($identificationVariableDec
...
+        $alias    = $pathExpr->identificationVariable;
+
+        /** @var Query\AST\PathExpression $pathExpr */
+        switch ($pathExpr->type) {
+            case AST\PathExpression::TYPE_STATE_FIELD:
+                $field = $pathExpr->field;
+                break;
+
+            case AST\PathExpression::TYPE_SINGLE_VALUED_ASSOCIATION:
+                // 1- the owning side:
```
greg0ire: Why put a number if that comment?

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -854,9 +856,41 @@ public function walkIdentificationVariableDeclaration($identificationVariableDec
...
+                //    Just use the foreign key, i.e. u.group_id
+                $fieldName = $pathExpr->field;
+                $class     = $this->queryComponents[$alias]['metadata'];
+
+                if (isset($class->associationMappings[$fieldName]['inherited'])) {
+                    $class = $this->em->getClassMetadata($class->associationMappings[$fieldName]['inherited']);
+                }
+
+                $assoc = $class->associationMappings[$fieldName];
```
greg0ire: Please refrain from using abbrevations, `association` is fine.

greg0ire (commented):
I have a few nitpicks, but this looks great 👍

Should this be documented somewhere?

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -1613,7 +1613,7 @@ From, Join and Index by
     RangeVariableDeclaration                   ::= AbstractSchemaName ["AS"] AliasIdentificationVariable
     JoinAssociationDeclaration                 ::= JoinAssociationPathExpression ["AS"] AliasIdentificationVariable [IndexBy]
     Join                                       ::= ["LEFT" ["OUTER"] | "INNER"] "JOIN" (JoinAssociationDeclaration | RangeVariableDeclaration) ["WITH" ConditionalExpression]
-    IndexBy                                    ::= "INDEX" "BY" StateFieldPathExpression
```
beberlei: Please also add an example in the docs below after the language description in the `Using INDEX BY ` section.

lib/Doctrine/ORM/Query/SqlWalker.php
andrews05: To answer all three of your comments, the switch statement here was simply copied from walkPathExpression(). I renamed this variable only because it was easier than changing it throughout the switch.

docs/en/reference/dql-doctrine-query-language.rst
andrews05: I've added a second example. Does it look okay?
I think the existing example is helpful for a more complex result with nested arrays and multiple levels of indexing, so I left that one there.

lib/Doctrine/ORM/Query/SqlWalker.php
greg0ire: Found the commit that introduced it: 639718e95c1ee483d0c5f889b0326ec39aa187c5

greg0ire (dismissed):

beberlei (dismissed):

SenseException (dismissed):

andrews05:
@lcobucci Any chance of this getting approved?

lib/Doctrine/ORM/Query/AST/IndexBy.php
```diff
@@ -33,14 +33,14 @@ class IndexBy extends Node
     /**
      * @var PathExpression
      */
-    public $simpleStateFieldPathExpression = null;
+    public $singleValuedPathExpression = null;
```
lcobucci, line 36: Since this class isn't marked as `@internal` renaming the property is a BC-break

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -854,9 +856,40 @@ public function walkIdentificationVariableDeclaration($identificationVariableDec
      */
     public function walkIndexBy($indexBy)
     {
-        $pathExpression = $indexBy->simpleStateFieldPathExpression;
+        $pathExpression = $indexBy->singleValuedPathExpression;
         $alias          = $pathExpression->identificationVariable;
-        $field          = $pathExpression->field;
+
+        /** @var Query\AST\PathExpression $pathExpression */
```
lcobucci, line 862: We prefer `assert()` over `/** @var */`

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -854,9 +856,40 @@ public function walkIdentificationVariableDeclaration($identificationVariableDec
...
+                if (count($association['sourceToTargetKeyColumns']) > 1) {
+                    throw QueryException::associationPathCompositeKeyNotSupported();
+                }
+
+                $field = reset($association['targetToSourceKeyColumns']);
+                break;
+
+            default:
+                throw QueryException::invalidPathExpression($pathExpression);
+        }
```
lcobucci, line 892: This seems like something that could be extracted to a private method and possibly use `if` and early exits (allowing for us to have a more strict comparison).

lcobucci: Also, is this done elsewhere?

lcobucci:
Would you please rebase this to resolve the conflicts? I'm slightly concerned with the BC-break but I trust @beberlei and @greg0ire to make a decision here and ship this :+1:

lib/Doctrine/ORM/Query/SqlWalker.php
andrews05: This is largely a copy of the switch from walkPathExpression(), though not identical. The walkPathExpression() one is building a sql string wheras this one is just determining the appropriate field.
Should I try combining them both into a private method?

andrews05:
@lcobucci Conflicts resolved. Should I squash the commits also?

lib/Doctrine/ORM/Query/SqlWalker.php
andrews05: Sorry, I wasn't entirely clear of the purpose of some of the docblocks like this. Perhaps the `@var` is redundant as the variable isn't a function parameter?

greg0ire:
> Should I squash the commits also?

Yes IMO

lib/Doctrine/ORM/Query/SqlWalker.php
greg0ire: If you don't need it for auto-complete or SA tools like PHPStan or Psalm, it should be removed, otherwise please use `assert($pathExpression instanceof PathExpression)` (and import that class of course). It will have the same effect but might be checked in some setups, which makes it better (less likely to rot than a comment).

andrews05:
@greg0ire Commits squashed. I removed the @var comment, it's redundant for autocomplete.

beberlei (approved):

PR #7926, opened 2019-11-27
Add docs note that FQCN is needed for an Embeddable configured via annotations

Pictor13:
By default PHP users will try to omit the namespace, importing it with `use`.
But that is not supported, at the moment.

Is possible to spare developers the time for figuring this out.

### DIFF
--- a/docs/en/tutorials/embeddables.rst
+++ b/docs/en/tutorials/embeddables.rst
@@ -79,6 +79,10 @@ In terms of your database schema, Doctrine will automatically inline all
 columns from the ``Address`` class into the table of the ``User`` class,
 just as if you had declared them directly there.
 
+.. note::
+
+    The Embeddable class must be specified as fully qualified (e.g. ``Embeddable\Address``). Relying on `use` and omitting the FQCN will not work.
+
 Initializing embeddables
 ------------------------
 


lcobucci:
> But that is not supported, at the moment.

@Pictor13 that should work when using `::class` syntax, have you tried that?

Pictor13:
Inside the annotation/comment? 😳
Had no idea it would work

Pictor13:
It does.
I discovered something new and unexpected. 🙂

Thanks, and sorry for taking your time.
Closing here.

Pictor13:
Ehmm... pardon me for reopening again @lcobucci , I found out to have some caching issue
(since I recently updated the *doctrine-migration-bundle*, that removed the deprecated *doctrine-cache-bundle*).

I reported it as working, but it was because of my previous attempt with the full namespace, that was running correctly and got cached.

```
// WORKING
namespace My\Entity;

// ...
/**
 * @ORM\Embedded(class="My\ValueObject\Address")
 **/
private $address;
```

when I tried with `::class` as suggested it appeared to be working

```
// NOT WORKING
namespace My\Entity;

use My\ValueObject\Address;
// ...

/**
 * @ORM\Embedded(class="Address::class")
 **/
private $address;
```

but reality is that it was running over the cache from the previous correct case.
When I cleared the cache this morning, it blew up 😣.
> Unknown Entity namespace alias 'My\Entity\Address'.

**Note:**
Also, the other attempt that I tried was _without_ `::class` (as mentioned in the pull request)

```
// NOT WORKING
namespace My\Entity;

use My\ValueObject\Address;
// ...

/**
 * @ORM\Embedded(class="Address")
 **/
private $address;
```
> Class 'My\Entity\Address' does not exist

is not working.

---

So, if I am not **still** doing something wrong, I think the addendum to the documentation might be useful and more explicit about how the FQCN **must** be specified.

Consider that I am pretty n00b with Symfony & Doctrine, and I passed the night coding with not much sleep (😴 my brain is in zombie mode); so please, let me know if there is something else that I might be missing; or if the documentation can actually be updated.

Pictor13:
Also, I am not sure how documentation versioning is handled for this repo, but probably (if accepted) the note should be added also for versions < doctrine:2.7 ?

lcobucci:
The `::class` syntax should be used without quotes, like:

```php
namespace My\Entity;

use My\ValueObject\Address;

class Blah
{
    /**
     * @ORM\Embedded(class=Address::class)
     **/
    private $address;
```

Have you tried this?

lcobucci:
We don't maintain < v2.7 so don't worry about them.

Pictor13:
> The `::class` syntax should be used without quotes, like:
> Have you tried this?

Nope... 😬
And it works!

Closing again 😅

---
Ok, I am clearly missing some knowledge about annotations (cause I usually avoid them).
Is it, using quotes or not, defined in some kind of PHP annotations standard/spec?
Or is it some specific Doctrine way reading annotations?

PR #7928, opened 2019-11-27
Deprecate EntityRepository#clear()

### LINKED ISSUE #7922: Deprecate EntityRepository#clear()
That method calls `EntityManager#clear()` with an argument, which won't be allowed anymore on v3.0. 

We must add a `@deprecated` annotation (without a `trigger_error()` since it's already handled by `EntityManager#clear()`).

--- 

Context on the **why**: https://github.com/doctrine/orm/issues/7925#issuecomment-558997382

### LINKED ISSUE #7925: Calling Doctrine\ORM\EntityManager::clear() with any arguments to clear specific entities is deprecated
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    |no
| Version     | 2.7.0

#### Summary

Calling `Doctrine\ORM\EntityRepository::clear` triggers deprecation message:
```
User Deprecated: Calling Doctrine\ORM\EntityManager::clear() with any arguments to clear specific entities is deprecated and will not be supported in Doctrine ORM 3.0. {"exception":"[object] (ErrorException(code: 0): User Deprecated: Calling Doctrine\\ORM\\EntityManager::clear() with any arguments to clear specific entities is deprecated and will not be supported in Doctrine ORM 3.0. at /var/www/app/vendor/doctrine/orm/lib/Doctrine/ORM/EntityManager.php:562)"} []
```

#### Current behavior

`Doctrine\ORM\EntityRepository::clear` does not any input and passes the call to `$this->_em->clear($this->_class->rootEntityName);`. 

I think we can just remove the `$this->_class->rootEntityName` argument here:
https://github.com/doctrine/orm/blob/a416a9a8b2cd2ea0f7b7b5638986e8b398f34df8/lib/Doctrine/ORM/EntityRepository.php#L133-L136

kokspflanze:
clear with entity class, generate a deprecated warning.

maybe better to remove the class name.

Closes #7925 , Closes #7922

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -24,6 +24,8 @@ use Doctrine\ORM\Query\ResultSetMappingBuilder;
 use Doctrine\Common\Persistence\ObjectRepository;
 use Doctrine\Common\Collections\Selectable;
 use Doctrine\Common\Collections\Criteria;
+use const E_USER_DEPRECATED;
+use function trigger_error;
 
 /**
  * An EntityRepository serves as a repository for entities with generic as well as
@@ -129,9 +131,13 @@ class EntityRepository implements ObjectRepository, Selectable
      * Clears the repository, causing all managed entities to become detached.
      *
      * @return void
+     *
+     * @deprecated 2.8 This method is being removed from the ORM and won't have any replacement
      */
     public function clear()
     {
+        @trigger_error('Method ' . __METHOD__ . '() is deprecated and will be removed in Doctrine ORM 3.0.', E_USER_DEPRECATED);
+
         $this->_em->clear($this->_class->rootEntityName);
     }
 
--- a/tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
@@ -1123,5 +1123,14 @@ class EntityRepositoryTest extends OrmFunctionalTestCase
             $this->assertTrue(in_array($user, [$user1, $user2]));
         }
     }
+
+    public function testDeprecatedClear() : void
+    {
+        $repository = $this->_em->getRepository(CmsAddress::class);
+
+        $this->expectDeprecationMessage('Method Doctrine\ORM\EntityRepository::clear() is deprecated and will be removed in Doctrine ORM 3.0.');
+        $this->expectDeprecationMessage('Calling Doctrine\ORM\EntityManager::clear() with any arguments to clear specific entities is deprecated and will not be supported in Doctrine ORM 3.0.');
+        $repository->clear();
+    }
 }
 


lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -132,7 +132,7 @@ public function createNativeNamedQuery($queryName)
      */
     public function clear()
     {
-        $this->_em->clear($this->_class->rootEntityName);
+        $this->_em->clear();
```
greg0ire: I would rather deprecated `EntityRepository::clear()`

greg0ire: That's what #7922 suggests anyway, and I added it to the list of issues your PR closes because I think you should go with that solution.

lcobucci: Exactly, removing the argument is also a bc break since it will clear everything.

lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -129,9 +131,13 @@ public function createNativeNamedQuery($queryName)
      * Clears the repository, causing all managed entities to become detached.
      *
      * @return void
+     *
+     * @deprecated 2.7 This method is being removed from the ORM and won't have any replacement
```
lcobucci: This should be 2.8 since we don't deprecate things in patch releases.

lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -129,9 +131,13 @@ public function createNativeNamedQuery($queryName)
...
      */
     public function clear()
     {
+        @trigger_error('Method ' . __METHOD__ . '() is deprecated and will be removed in Doctrine ORM 3.0.', E_USER_DEPRECATED);
```
lcobucci: This will create two deprecation messages since the called method would also trigger an error, not really a huge issue but might confuse people (I think). 

We could add an error handler and then restore it back to avoid it or could not add a message here (but that might be confusing for users). What do you think is the most appropriate one?

lcobucci (changes requested):
Thank you for your contribution here! Just a few minor things and we should also add a test verifying the deprecation message (if we keep it).

lib/Doctrine/ORM/EntityRepository.php
lcobucci: For this reason I'm going to change the destination branch here, please rebase your work against the branch 2.8.x 👍

lib/Doctrine/ORM/EntityRepository.php
kokspflanze: for my case it would be better to remove the parameter, but you already say that its is a bc break for you.

to add an error handler, could create much more problems in custom use cases, as the double notification.

so i think its better to trigger 2 messages as possible problems with error handlers.

lib/Doctrine/ORM/EntityRepository.php
kokspflanze: atm 2.8 and 2.7 is the same, so i dont know why to rebase.

greg0ire: This is a minor contributions, and 2.7 only accepts patches.

kokspflanze: @greg0ire yes, but still the branches are same, wo why to git rebase?

SenseException (approved):

lib/Doctrine/ORM/EntityRepository.php
greg0ire: Oh right sorry, if it's the same there's no point indeed

lcobucci (approved):

PR #7929, opened 2019-11-28
docs: detach deprecated

pierreboissinot:
>      * @deprecated Detach operation is deprecated and will be removed in Persistence 2.0. Please use @see ObjectManager::clear()} instead.

### DIFF
--- a/docs/en/reference/batch-processing.rst
+++ b/docs/en/reference/batch-processing.rst
@@ -183,7 +183,7 @@ problems using the following approach:
         // do stuff with the data in the row, $row[0] is always the object
     
         // detach from Doctrine, so that it can be Garbage-Collected immediately
-        $this->_em->detach($row[0]);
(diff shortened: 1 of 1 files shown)


docs/en/reference/batch-processing.rst
```diff
@@ -183,7 +183,7 @@ problems using the following approach:
         // do stuff with the data in the row, $row[0] is always the object
     
         // detach from Doctrine, so that it can be Garbage-Collected immediately
-        $this->_em->detach($row[0]);
+        $this->_em->clear($row[0]);
```
lcobucci: Passing arguments to clear is also deprecated 😄
We possibly need to check more things in this doc.

SenseException: Detaching an entity won't be possible afaik, so I guess we will remove those parts from the documentation.

PR #7931, opened 2019-11-28
Fix behaviour when using CONTAINS on json fields

alberthorta:
Fix issues when using CONTAINS on json fields related with the json_encode during the hydration phase acting as if the field was string instead.

Basically when using CONTAINS Doctrine Criteria on a json field the behaviour and the query generated was not correct since it's encoded with json_encode.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -928,8 +928,8 @@ class BasicEntityPersister implements EntityPersister
 
         foreach ($types as $type) {
             [$field, $value, $operator] = $type;
-            $sqlType = $this->getTypes($field, $value, $this->class);
-            if($operator === Comparison::CONTAINS && $sqlType[0] === Type::JSON) {
+            $sqlType                    = $this->getTypes($field, $value, $this->class);
+            if ($operator === Comparison::CONTAINS && $sqlType[0] === Type::JSON) {
                 $sqlType[0] = Type::STRING;
             }
             $sqlTypes = array_merge($sqlTypes, $sqlType);
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -920,15 +920,19 @@ class BasicEntityPersister implements EntityPersister
 
         $valueVisitor->dispatch($expression);
 
-        [$params, $types] = $valueVisitor->getParamsAndTypes();
+        list($params, $types) = $valueVisitor->getParamsAndTypes();
 
         foreach ($params as $param) {
             $sqlParams = array_merge($sqlParams, $this->getValues($param));
         }
 
         foreach ($types as $type) {
-            [$field, $value] = $type;
-            $sqlTypes        = array_merge($sqlTypes, $this->getTypes($field, $value, $this->class));
+            list ($field, $value, $operator) = $type;
+            $sqlType = $this->getTypes($field, $value, $this->class);
+            if($operator === Comparison::CONTAINS && $sqlType[0] === Type::JSON) {
+                $sqlType[0] = Type::STRING;
+            }
+            $sqlTypes = array_merge($sqlTypes, $sqlType);
         }
 
         return [$sqlParams, $sqlTypes];
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -920,14 +920,14 @@ class BasicEntityPersister implements EntityPersister
 
         $valueVisitor->dispatch($expression);
 
-        list($params, $types) = $valueVisitor->getParamsAndTypes();
+        [$params, $types] = $valueVisitor->getParamsAndTypes();
 
         foreach ($params as $param) {
             $sqlParams = array_merge($sqlParams, $this->getValues($param));
         }
 
         foreach ($types as $type) {
-            list ($field, $value, $operator) = $type;
+            [$field, $value, $operator] = $type;
             $sqlType = $this->getTypes($field, $value, $this->class);
             if($operator === Comparison::CONTAINS && $sqlType[0] === Type::JSON) {
                 $sqlType[0] = Type::STRING;


greg0ire:
Can you please tell us what specifically was not correct and how do your changes fix that?

alberthorta:
Hi,
When you're using something like that:

```
$criteria = new DoctrineCriteria()
$criteria->where(
  DoctrineCriteria::expr()->contains(
    "field",
    "something"
  )
);
$this->entityManager->getRepository("name")
            ->matching($criteria)->getValues();
```

and the ```field``` is of type ```json``` I would expect the WHERE on the SQL launched something like this:
```
WHERE field LIKE '%something%'
```
and due to the json parsing this is what's sent to SQL (and that's basically because doctrine is using a json_encode when you send a value on a json field, and while that's generally true, it is not in that specific case, so basically I'm parsing the value as if it was of a type string):
```
WHERE field LIKE '\"%something\"'
```
Making impossible to do a proper CONTAINS over a json field.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -927,8 +927,12 @@ public function expandCriteriaParameters(Criteria $criteria)
         }
 
         foreach ($types as $type) {
-            [$field, $value] = $type;
-            $sqlTypes        = array_merge($sqlTypes, $this->getTypes($field, $value, $this->class));
+            [$field, $value, $operator] = $type;
+            $sqlType                    = $this->getTypes($field, $value, $this->class);
+            if ($operator === Comparison::CONTAINS && $sqlType[0] === Type::JSON) {
+                $sqlType[0] = Type::STRING;
+            }
```
greg0ire: I understand the problem better now, but it doesn't feel right to have it solved here… I mean what if there are some other, similar cases with other types? Is this going to become a growing piece of code full of special cases like this one?

alberthorta: Will try to contain the code change on the JsonType class to sort out this kind of issues. Anyway the behaviour right now it's faulty under my point of view. I will try to change the PR to have this logic on every Type Class.

Do you think it's worth it?

greg0ire: I don't know, I mean when someone asks me to check if a piece of json contains `foo"`, I think I should look for `foo\"` inside json values, it's not really like a normal string…

alberthorta: That's not the issue, the issue is that you're searching if ```foo"``` is contained in a json field and what doctrine is checking is if it conains ```\"foo\\"\"``right now.

greg0ire: I'm not describing the issue, I'm describing how I think it should behave when using contains with `'foo\"'`

1. You say it's currently looking for `\"foo\\"\"`
2. I think your code will result in it looking for `foo"`
3. I think it should instead look for `foo\"` (json special chars should be escaped)

alberthorta: Sorry It was my mistake, since not the value is following the string rules it will look for ```foo\"```exactly as you said.

greg0ire: I doubt that… why would string rules imply escaping for JSON?

alberthorta: The thing here is that the kind of escape you want to do on a LIKE WHERE condition should be only the one that escape the string to prevent injections and this kind of stuff but when you're doing a LIKE on a WHERE field you never want (at least for json types) to do any kind of escape because if you do it you're basically disallowing to use LIKE on those kind of fields.
In the end, the DB is storing those fields as a strings so when you're looking for a specific portion of the content you should not (I think) do any kind of json_escape, because if you do it you're not allowing CONTAINS on any kind of json field. If it have to be that way, then Doctrine should disallow the use of CONTAINS on any json field because it's basically broken right now.
If you only can use CONTAINS on a json field if you're checking if the whole content it's equal to the thing you're comparing, then use EQ instead, but when I'm using CONTAINS I want to check if those chars are inside the json field.

greg0ire: I'll let another contributor chime in on this one, because I'm really not sure what the right solution is here

alberthorta: For me is either disallow contains on json (because they're not working anyway) or allowing them with the behaviour that anybody will expect when using a like over a string field. I'm with you that maybe that's not the best place for the patch. Again if you think that this can be fixed I will try to see if there's a better place for this change allowing those kind of exceptions in a less (let's say) patchy way.

greg0ire: To me it should be search keys or values, I mean if the json is

```
{
    "foo": "bar"
}
```
, what should happen if we look for `{"foo"   :`?

It does not look like formatting is preserved when using a `jdoc` datatype in MySQL, for instance: https://www.db-fiddle.com/f/8GBrXECD5TkhTXPaxWRp3m/0
Same goes for Postgres: https://www.db-fiddle.com/f/umBxSdr7LSuoMjeH74Uebc/0

Note that containment is something that Postgres support with a special JSON operator: https://www.postgresql.org/docs/current/datatype-json.html#JSON-CONTAINMENT

MySQL uses a special function: https://dev.mysql.com/doc/refman/5.7/en/json-search-functions.html#function_json-contains

And it could be that there is something out there to help you solve your uuid searching issue with something more robust and efficient than the `LIKE` operator: https://packagist.org/?query=doctrine%20json

alberthorta: On a table with those contents:
```
+----------------+
|        field     |
+----------------+
+  {"foo": "bar"}  |
+----------------+
```
If I do :
```
$criteria = new DoctrineCriteria();
$criteria->where(
  DoctrineCriteria::expr()->contains(
    "field",
    "foo"
  )
);
$this->entityManager->getRepository("name")
            ->matching($criteria)->getValues();
```

On your example I WILL EXPECT this DB entry on the results. (Check https://www.db-fiddle.com/f/rY1rhWJD9mxAYR5MWHmPEB/0 , as you can see it works exactly that way, but Doctrine don't) Anything different than that it's clearly an error on the ORM. If the ORM allows me to use LIKE on json fields should work exactly as I describe. MySQL allows LIKE on json fields and work exactly as I mentioned but DOCTRINE not, that's misleading and took me some time to understand if I was doing something wrong, because even doctrine's LOG is not showing the real search, DOCTRINE log was showing me as if the prepared value for the statement was ```%foo%``` and clearly it is not, I had to activate the logs on MYSQL to understand what was wrong.

On MYSQL I can do something like this:
```SELECT * FROM TABLE WHERE field LIKE 'foo'```
Being field a json one. And MYSQL is returning exactly the row that contains a json with this foo somewhere in the middle. I expect the ORM to do the same or at least make it sense in any way. And right now CONTAINS on json fields is completely useless.

I know about JSON_CONTAINS, but once you have a query built using doctrine criterias it's not trivial to just go back and rewrite the whole logic just because CONTAINS is not doing what any person that is used to work with DB's is expecting. Not to mention that in most cases if the field content it's not too big it's much more performant a LIKE than a JSON_SEARCH that's impossible to index in any way at least in MYSQL. I know that postgress have much more advanced functionallity on JSON fields but MYSQL it is not.

greg0ire: > I can agree with you that maybe that is not the best place for the patch, but I cannot agree that DOCTRINE is working correctly at all in that specific case.

To clarify, I wasn't suggesting that Doctrine is working correctly either. I was suggesting that the solution might be to have it result in usage of `JSON_CONTAINS`, or the Postgres operator, or just an exception that contains is not supported on json fields.

alberthorta: The funny thing is that IT IS SUPPORTED on the real deal (the DB engine). Not supporting it on Doctrine it's something I can't understand when you can do it. But again, not on my hands.

greg0ire: That's a good point you're making there. As I already said, I'm probably not the best person for this, let's see what other contributors think, ok?

alberthorta: Perfect, as I said I'm happy to rethink what's the best place for this kind of things, I know that the PR fixes my specific problem. Now that I'm thinking the most probable is that ```startsWith``` and ```endsWith``` will have the same problem since they're end up using ```LIKE``` like ```contains```. So even if my patch fixes my specific issue it can get more and more complicated if we start adding exceptions...

beberlei:
I started writing a reply suggesting this needs generic fixing in DBAL Types abstraction, but looking at `SqlValueVisitor` again I blieve that the contains operator (also starts + ends with) only make sense on strings.

The type system would need to have an abstraction "supportsContainsOperators(): bool" or something alike though, so that Types that extend our type system can make sure they work.

Plan here should by in my opinion:

1. don't use contains() in your code unless its a string type.
2. add new method to `Doctrine\DBAL\Types\Type::supportsContainsOperators()` that returns true for `string` and `text` and false for all other types in the next version of DBAL.
3. Make use of this operator when DBAL dependency gets upgraded in ORM.

webda2l:
IMO this PR can be closed.

But the idea to better handling JSON field, that will work with filtering collection will be nice.
I don't see discussion about it.

As quick thinking, maybe a basic DQL (and qb fn) `JSON_GET(string col, string field, ?string cast = null)` fn, adding a new Criteria ExpressionBuilder jsonGet as well, and allowing all the existing Criteria ExpressionBuilder to have another Criteria ExpressionBuilder as field (to allow doing: `Criteria::expr()->eq(Criteria::expr()->jsonGet('cfg', 'max'), 1)`)
?

PR #7934, opened 2019-11-29
Fix Trying to access array offset on value of type null

BenMorel:
This error occurs on PHP 7.4, for example when you're running this (invalid) DQL from the Symfony console:

```
bin/console doctrine:query:dql "SELECT CURRENT_TIMESTAMP()"
```

> In Parser.php line 305:
> Notice: Trying to access array offset on value of type null

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -302,7 +302,7 @@ class Parser
      */
     public function match($token)
     {
-        $lookaheadType = $this->lexer->lookahead['type'];
+        $lookaheadType = $this->lexer->lookahead['type'] ?? null;
 
         // Short-circuit on first condition, usually types match
         if ($lookaheadType === $token) {
--- a/tests/Doctrine/Tests/ORM/Query/ParserTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/ParserTest.php
@@ -134,6 +134,24 @@ class ParserTest extends OrmTestCase
         ];
     }
 
+    /**
+     * PHP 7.4 would fail with Notice: Trying to access array offset on value of type null.
+     *
+     * @see https://github.com/doctrine/orm/pull/7934
+     *
+     * @group GH7934
+     */
+    public function testNullLookahead() : void
+    {
+        $query = new Query($this->_getTestEntityManager());
+        $query->setDQL('SELECT CURRENT_TIMESTAMP()');
+
+        $parser = new Parser($query);
+
+        $this->expectException(QueryException::class);
+        $parser->match(Lexer::T_SELECT);
+    }
+
     private function createParser($dql)
     {
         $query = new Query($this->_getTestEntityManager());


lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -302,7 +302,7 @@ public function getAST()
      */
     public function match($token)
     {
-        $lookaheadType = $this->lexer->lookahead['type'];
+        $lookaheadType = $this->lexer->lookahead ? $this->lexer->lookahead['type'] : null;
```
lcobucci: Maybe using `??` would be a bit more cleaner?

lcobucci (commented):
We'd also need a test case here

lib/Doctrine/ORM/Query/Parser.php
BenMorel: True. I didn't now that this operator would also handle attempting to access array offset on null without erroring. Fixed.

BenMorel:
Not sure I wrote the best test here, but it does catch the notice on PHP 7.4!

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -302,7 +302,7 @@ public function getAST()
      */
     public function match($token)
     {
-        $lookaheadType = $this->lexer->lookahead['type'];
+        $lookaheadType = $this->lexer->lookahead['type'] ?? null;
```
beberlei: So lookahead is not an array in your case?

BenMorel: Yep, `$this->lexer->lookahead` [is null](https://github.com/doctrine/lexer/blob/1.2.0/lib/Doctrine/Common/Lexer/AbstractLexer.php#L59) when executing the CLI command I posted in the intro. Is this a bug in itself?

beberlei: Not according to the typehint, it is `array|null` - so this change is right.

beberlei (dismissed):

lib/Doctrine/ORM/Query/Parser.php
BenMorel: Yes, but it could have been nullable only under certain conditions, that could be assumed would never be met at that point. Anyway if this is the expected behaviour, then yes the PR just fixes the notice on PHP 7.4.

tests/Doctrine/Tests/ORM/Query/ParserTest.php
```diff
@@ -134,6 +134,21 @@ public function invalidMatches()
         ];
     }
 
+    /**
+     * PHP 7.4 would fail with Notice: Trying to access array offset on value of type null.
+     * https://github.com/doctrine/orm/pull/7934
+     *
+     * @expectedException \Doctrine\ORM\Query\QueryException
```
SenseException: Can you please use `$this->expectException(QueryException::class);` in the test code instead of the annotation? Put it before that line of code that is throwing the expected exception and import `Doctrine\ORM\Query\QueryException` with `use`.

SenseException (changes requested):

tests/Doctrine/Tests/ORM/Query/ParserTest.php
BenMorel: Good point. Done in 000d923363423476bf4a1f0cd940f31548bdeae0.

SenseException (dismissed):

lcobucci (approved):

BenMorel:
👍

PR #7935, opened 2019-11-29
PHP 7.4 compat: check if a property is initialized

firstred:
...before accessing it with `ReflectionProperty`

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -775,7 +775,9 @@ class UnitOfWork implements PropertyChangedListener
 
         // Look for changes in associations of the entity
         foreach ($class->associationMappings as $field => $assoc) {
-            if (($val = $class->reflFields[$field]->getValue($entity)) === null) {
+            if ((method_exists($class->reflFields[$field], 'isInitialized') && ! $class->reflFields[$field]->isInitialized($entity))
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1020,7 +1022,9 @@ public function recomputeSingleEntityChangeSet(ClassMetadata $class, $entity)
         foreach ($class->reflFields as $name => $refProp) {
             if (( ! $class->isIdentifier($name) || ! $class->isIdGeneratorIdentity())
                 && ($name !== $class->versionField)
-                && ! $class->isCollectionValuedAssociation($name)) {
+                && ! $class->isCollectionValuedAssociation($name)
+                && (method_exists($refProp, 'isInitialized') && $refProp->isInitialized($entity))
```
Ocramius: Probably best to have a check on whether it's an uninitialized proxy or not: an entity with uninitialized fields landing in the `UnitOfWork` is very very likely a bug...

Ocramius (changes requested):

lib/Doctrine/ORM/UnitOfWork.php
firstred: Looks like it. I noticed that this PR might also be redundant and is missing some code, so I am closing it for now.
Thanks for having a look!

PR #7936, opened 2019-11-30
[GH-5202] Implement Query::HINT_READ_ONLY flag

### LINKED ISSUE #5202: DDC-691: doctrine.readOnly query hint
Jira issue originally created by user romanb:

Setting such a query hint to TRUE should result in all entities being retrieved by that query to be read-only for the purposes of change-tracking. Note that the entities themselves need not necessarily be read-only in general.

This feature is a flush performance tweak that can be used to query for objects but not let the returned objects run through change-tracking on flush. Any other managed objects are tracked as usual so you can do a read-only query for 100 entities and persist a new entity in the same unit of work with optimal flushing performance.

beberlei:
It was already possible to set entities as read only through 1.) metadata 2.) calling UnitOfWork->markReadOnly().

This PR adds functionality to automatically make objects read only if they are hydrated through a query that has a query hint `Query::HINT_READ_ONLY` set. 

This also changes objects to read only that have already been hydrated to the identity map from previous queries.

Fixes #5202


theofidry (dismissed):

SenseException (changes requested):
Aside from the coding styles that need to be fixed, is there a defined behaviour that a fetched read only entity can become writable with another query? Depending on the behaviour, can this be covered with a test to prevent regressions?

beberlei:
@SenseException objects can only become writeable again by either:

1. calling `EntityManager::clear`
2. removing entity form identityMap through detach

and then refetch it.

`ReadOnlyTest` covers this already.

SenseException:
Does it make sense to have a test like `ReadOnlyTest::testReadOnlyEntityNeverChangeTracked()` for the hint as a redundant test? I know that the "is readonly" tests suggest that writing a readonly object won't work, but because ORM isn't a simple piece of software maybe another test that tries to write an object can give extra protection for regressions.

beberlei:
I dont like redundant tests, but we can move the tests here to ReadOnlyTest maybe

SenseException:
> but we can move the tests here to ReadOnlyTest maybe

Okay, please move them to `ReadOnlyTest`

stof:
> This also changes objects to read only that have already been hydrated to the identity map from previous queries.

This one seems risky to me. The code that loaded these objects with a previous query might expect them to be writable. If we have a writable query and a read-only query, the final result should be writable IMO.

SenseException:
> the final result should be writable IMO

With partial objects you still get a partial object if one was pre-fetched by another query before. It is an expected behaviour to get an object in a state just like from the last query that fetched an entity before (if no one cleared the entity manager in between). It can confuse users.

malarzm:
Not sure how feasible it is for ORM, but in ODM (https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/query-builder-api.html#fetching-documents-as-read-only, PR for reference https://github.com/doctrine/mongodb-odm/pull/1403) we took the approach of omitting the identity map altogether: effectively READ_ONLY hint forces the UnitOfWork to totally forget about the object after hydration (also it omits identity map lookup in the first place). Thanks to this we don't have a conflict of interest between objects managed by UoW and those intended to be read only. This however conflicts with identity map lookup for readOnly objects ORM already has, IIRC we didn't have them at a time of writing the feature so this was neglected.

stof:
@malarzm there is a feature request in #8429 about getting results as untracked objects, which seems to correspond to your concept.

beberlei:
@stof @malarzm I believe the problem with an approach where the read only flag would not put entities in the identity map are two things:

1. what to do about proxies? Are they put in the identity map when loaded from a read only entity?
2. what about persistent collections? can we initialize them in a way that they don't blow up on changes and flush? Specifically orphan removal behavior could be tricky.

malarzm:
@beberlei reading old comments proxies and collections seemed doable, I've lost the battle to inverse side of a reference using repository method. So instead we took a shortcut and stated in documentation that the feature is half baked 🙈 

> Read-only mode is not deep, i.e. any references (be it owning or inverse) of fetched WILL be managed by Doctrine. This is a shortcoming of current implementation, may change in future and will not be considered a BC break (will be treated as a feature instead).

Nobody complained about it over the years though.

stof:
@beberlei if the feature is only about read-only entities, not about detached hydration, I would vote for the fact that entities which are already in the identity map before the hydration are not turned into read-only ones. This is safer regarding other code that fetched them as writeable and might not even be aware of another query being done in read-only mode.
Having to reason about the whole project to be sure that no read-only query can break your write code by changing the tracking of your entity would make things quite hard to manage (and would probably be a reason for projects to ban the usage of read-only queries altogether).
There is also a question about what happens if you fetch results for a write use case and the object was already in the identity map as read-only. If this does not make it writeable again, we have the same kind of issue (a read-only query anywhere in the project can poison other parts of the codebase expecting writeable objects).

beberlei:
@stof i changed the behavior to work like you described, when an object is known either as entity or proxy in the identity map already, then a query with read only hint will not make that object read only. Only if the object is fetched beforehand as read only will it stay as read only.

@SenseException i moved the tests to the `ReadOnlyTest`.

Issue #7765, opened 2019-07-01
[Documentation] Extra lazy docs do not mention removeElement eagerness


malarzm:
TIL that not initialized collection in `EXTRA_LAZY` mode will remove element with `->removeElement()` from the database without waiting for a `flush`. I think it should be mentioned in the docs https://github.com/doctrine/orm/blob/04762015c6b4fe1dc2f3c1043c307710b2c59fa7/lib/Doctrine/ORM/PersistentCollection.php#L355-L365

DonCallisto:
As `$persister` here 

https://github.com/doctrine/orm/blob/04762015c6b4fe1dc2f3c1043c307710b2c59fa7/lib/Doctrine/ORM/PersistentCollection.php#L362

could be of two types `ManyToManyPersister` and `OneToMany` persister this is the relevant code

https://github.com/doctrine/orm/blob/e1bb9e005d2676c42ec19521db9569a28da179d3/lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php#L159
https://github.com/doctrine/orm/blob/1ab9d0b1942a6b70d01640d51056a71916255ecf/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php#L234

I think the docs should point out this situation - where applicable - both in https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/working-with-associations.html#orphan-removal and https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/tutorials/extra-lazy-associations.html#extra-lazy-associations

BTW the fact that `ManyToManyPersister` does not contemplate `orphanRemoval` makes me wonder if this is a bug or is done by design.

As said in #7852 that's now closed, I'm willing to work on this PR if think is worthy.

flaushi:
If I see this correctly, one viable workaround is - and could well be mentioned in the docs - that if one needs the event listener to be called, one can replace
```
$entity->manyToManyCollection->removeElement($other);
```
with
```
$index = $entity->manyToManyCollection->indexOf($other);
if($index !== false)
    $entity->manyToManyCollection->remove($index);
```

DonCallisto:
The intent of `EXTRA_LAZY` is to prevent the whole collection to be loaded so I don't think it has too much sense in this context.
Moreover how can one change that section of code?
Lastly, this is the intended behaviour.

malarzm:
Closing as behaviour was changed in #7940

PR #7940, opened 2019-12-01
[GH-7864] Bugfix in PersistentCollection::removeElement for EXTRA_LAZY.

### LINKED ISSUE #7864: Bug fetch EXTRA_LAZY, removeElement doesn't remove the element from the collection
Hello,

When working with a realtion `OneToMany` with `fetch="EXTRA_LAZY"` when removing an element of the relation doesn't remove the element from the collection.

Example, create an entity `Person` that has a `OneToMany` association with an entity `PhoneNumber`.

Crate the content of the database, inserting a person with 3 phone numbers.

```
$person = (new Person())->setName('David');
$person->addPhone((new PhoneNumber())->setPhoneNumber('11111111'));
$person->addPhone((new PhoneNumber())->setPhoneNumber('22222222'));
$person->addPhone((new PhoneNumber())->setPhoneNumber('33333333'));
$this->em->persist($person);
$this->em->flush();
```

The following code removes the phone number 22222222: 

```
/** @var Person $person */
$person = $this->personRepository->findOneByName('David');
/** @var PhoneNumber $number */
$number = $this->phoneNumberRepository->findOneByPhoneNumber('22222222');
$person->removePhone($number);
dump($person->getPhones()->map(function(PhoneNumber $p) { return $p->getPhoneNumber(); }));

====>
$ bin/console app:test
Doctrine\Common\Collections\ArrayCollection^ {#1520
  -elements: array:2 [
    0 => "11111111"
    2 => "33333333"
  ]
}
```

But if I change in the Person entity the relation to `EXTRA_LAZY`:

```
/**
 * @ORM\OneToMany(targetEntity="App\Entity\PhoneNumber", mappedBy="person", fetch="EXTRA_LAZY", cascade={"persist"})
 */
private $phones;

```

The code above produces the following output:

```
$ bin/console app:test
Doctrine\Common\Collections\ArrayCollection^ {#1583
  -elements: array:3 [
    0 => "11111111"
    1 => "22222222"
    2 => "33333333"
  ]
}
```

As you can see, the phone number `22222222` has not been removed from the collection.

If I iterate the relation before the removing the phone, then the phone is removed.

```
foreach($person->getPhones() as $phone){};
$person->removePhone($number);
===>
$ bin/console app:test
Doctrine\Common\Collections\ArrayCollection^ {#1501
  -elements: array:2 [
    0 => "11111111"
    2 => "33333333"
  ]
}
```

It looks like removeElement does't fetch the relation from the database so it's not able to remove anything.

This is the implementation of th entities:

```
namespace App\Entity;

/**
 * @ORM\Entity(repositoryClass="App\Repository\PersonRepository")
 */
class Person
{
    /**
     * @ORM\Id()
     * @ORM\GeneratedValue()
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @ORM\Column(type="string", length=255)
     */
    private $name;

    /**
     * @ORM\OneToMany(targetEntity="App\Entity\PhoneNumber", mappedBy="person", cascade={"persist"})
     */
    private $phones;

    public function __construct()
    {
        $this->phones = new ArrayCollection();
    }

    public function getId(): ?int
    {
        return $this->id;
    }

    public function getName(): ?string
    {
        return $this->name;
    }

    public function setName(string $name): self
    {
        $this->name = $name;

        return $this;
    }

    /**
     * @return Collection|PhoneNumber[]
     */
    public function getPhones(): Collection
    {
        return $this->phones;
    }

    public function addPhone(PhoneNumber $phone): self
    {
        if (!$this->phones->contains($phone)) {
            $this->phones[] = $phone;
            $phone->setPerson($this);
        }

        return $this;
    }

    public function removePhone(PhoneNumber $phone): self
    {
        if ($this->phones->contains($phone)) {
            $this->phones->removeElement($phone);
            // set the owning side to null (unless already changed)
            if ($phone->getPerson() === $this) {
                $phone->setPerson(null);
            }
        }

        return $this;
    }
}

```

```
namespace App\Entity;

/**
 * @ORM\Entity(repositoryClass="App\Repository\PhoneNumberRepository")
 */
class PhoneNumber
{
    /**
     * @ORM\Id()
     * @ORM\GeneratedValue()
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @ORM\Column(type="string", length=255)
     */
    private $phoneNumber;

    /**
     * @ORM\ManyToOne(targetEntity="App\Entity\Person", inversedBy="phones")
     */
    private $person;

    public function getId(): ?int
    {
        return $this->id;
    }

    public function getPhoneNumber(): ?string
    {
        return $this->phoneNumber;
    }

    public function setPhoneNumber(string $phoneNumber): self
    {
        $this->phoneNumber = $phoneNumber;

        return $this;
    }

    public function getPerson(): ?Person
    {
        return $this->person;
    }

    public function setPerson(?Person $person): self
    {
        $this->person = $person;

        return $this;
    }
}

```

beberlei:
`PersistentCollection::removeElement` cannot be `EXTRA_LAZY` as it breaks the assumption that writes are only ever done during the flush operation

Yet so far calling `removeElement` on an EXTRA_LAZY collection deletes the related entity immediately.

We revert this functionality completely to its pre 2.4.0 state, because it breaks the ORM "contract".

Fixes #7864


alcaeus (dismissed):

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7864Test.php
```diff
@@ -0,0 +1,51 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\Models\Tweet\Tweet;
+use Doctrine\Tests\Models\Tweet\User;
+use Doctrine\Tests\OrmFunctionalTestCase;
+use function array_values;
+
+class GH7864Test extends OrmFunctionalTestCase
```
lcobucci: Please add `@group gh7864` to more easily allow us to run this in the future

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7864Test.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    protected function setUp() : void
+    {
+        $this->useModelSet('tweet');
```
lcobucci: We're avoiding using model sets and putting all necessary classes in the test file. That helps a lot to keep the SUT more explicit. Would you please do the same here?

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -372,9 +372,17 @@ public function removeElement($element)
...
+            if ($this->em->getUnitOfWork()->getEntityState($element) === UnitOfWork::STATE_NEW) {
+                return true;
+            }
+
             $persister = $this->em->getUnitOfWork()->getCollectionPersister($this->association);
 
-            return $persister->removeElement($this, $element);
+            if ($persister->removeElement($this, $element) === true) {
+                // only  owning side or orphan removal triggers successful
+                // delete so that we can keep this collection unitialized.
```
lcobucci: Typo: unitialized -> uninitialized

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -372,9 +372,17 @@ public function removeElement($element)
...
             }
 
+            if ($this->em->getUnitOfWork()->getEntityState($element) === UnitOfWork::STATE_NEW) {
+                return true;
+            }
+
             $persister = $this->em->getUnitOfWork()->getCollectionPersister($this->association);
 
-            return $persister->removeElement($this, $element);
+            if ($persister->removeElement($this, $element) === true) {
```
lcobucci: It's kind of risky to leave this outside of a transaction. That creates confusion and would be lovely if we could avoid it IMHO

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7864Test.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->_em->persist($user);
+        $this->_em->persist($tweet1);
+        $this->_em->persist($tweet2);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $user  = $this->_em->find(User::class, $user->id);
+        $tweet = $this->_em->find(Tweet::class, $tweet1->id);
+
+        $user->tweets->removeElement($tweet);
```
lcobucci: Since this PR also removes the tests which were asserting that the collection was/wasn't initialised shouldn't we add the assertion here as well?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7864Test.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $tweets = $user->tweets->map(static function (Tweet $tweet) {
+            return $tweet->content;
+        });
+
+        $this->assertEquals(['Goodbye, and thanks for all the fish'], array_values($tweets->toArray()));
```
lcobucci: Please use `self::assert*` since these methods are static.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7864Test.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->_em->persist($tweet2);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $user  = $this->_em->find(User::class, $user->id);
+        $tweet = $this->_em->find(Tweet::class, $tweet1->id);
+
+        $user->tweets->removeElement($tweet);
+
+        $tweets = $user->tweets->map(static function (Tweet $tweet) {
```
lcobucci: ` : string` missing here

lcobucci (changes requested):
A few tiny things... ~I'm not sure if we should put this in a patch since it's kind of a minor bc-break because of the addition of a delete query not handled during flush.~

lib/Doctrine/ORM/PersistentCollection.php
beberlei: @lcobucci I agree, but its too late now. We can't move this behavior out of a transaction without completly removing the extra lazy from `remoteElement`.

beberlei:
@lcobucci i think we have a misudnerstanding. The delete query is already done outside of a flush operation in the old code / now. Nothing changes with respect to the delete query, only thing that changes is that it could still trigger a full load of the collection if no entry was deleted.

lib/Doctrine/ORM/PersistentCollection.php
lcobucci: Even if we track the IDs of removed elements and filter them from the select query?

lcobucci: Nerver mind, you're right about it being too late.

lcobucci:
@beberlei you're right, sorry. Slow brain on a Monday morning.

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -372,9 +372,17 @@ public function removeElement($element)
                 return $this->collection->removeElement($element);
             }
 
+            if ($this->em->getUnitOfWork()->getEntityState($element) === UnitOfWork::STATE_NEW) {
+                return true;
```
lcobucci: Do we have a test covering this scenario? If not, can you provide one as well?

beberlei: Yes, i had to change a lot in `ExtraLazyCollecitonTest.php` as this change introduced many failures, its fully tested after the adjustment of tests.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7864Test.php
beberlei: The tests all use $this->assert* at the moment. Shouldn't we enforce this via coding standard instead so that we have a phpcbf fix for it instead of requiring us to do this kind of remark manually?

docs/en/tutorials/extra-lazy-associations.rst
```diff
@@ -19,6 +19,7 @@ can be called without triggering a full load of the collection:
 -  ``Collection#count()``
 -  ``Collection#get($key)``  (available with Doctrine 2.4)
 -  ``Collection#slice($offset, $length = null)``
+-  ``Collection#removeElement($element)`` for either owning side collections or inverse collections with orphanRemoval=true
```
malarzm: With this addition we should be fine closing https://github.com/doctrine/orm/issues/7765. Could you maybe change line below (I can't suggest change there) to mention DELETE query? Something along "For each call, if the Collection is not yet loaded, issue a straight SELECT/DELETE statement against the database."

stof:
IMO, doing the DELETE query outside a flush transaction should be considered as a bug. What the collection should do is register this query as an extra query to run during the next flush.

beberlei:
@stof but that is the problem of the bug. If we keep allowing removeElement to be lazy, then we registered the delete, initialize the collcetion and it would suddenly have the element in it again that was actually removed from it.

Imho removeElement should never have been made lazy, we could built it back. Its not really a BC break, but potentially a massive performance hog.

stof:
I vote for removeElement to become non-lazy. The performance cost of loading the collection is IMO better than the surprise of having the change being flushed separately from the ORM flush.

Btw, in case you use such entity in a Symfony form, it means you execute the delete in the DB even before running the validation, so that means you might even decide that you don't want to save the removal.

beberlei:
@stof thats a very good point.

stof:
Note that I was not aware of the fact that `removeElement` was not respecting the flush semantic for extra-lazy collections. And I'm sure that this means that some parts of my project actually have bugs related to the few extra-lazy collections I have...

stof:
Thinking about it, the bug might not impact Symfony forms as much as I thought, because the way they work make it likely that the collections gets loaded anyway before removing elements from it (which would work around the bug AFAIU). But it is still possible to be impacted.

docs/en/tutorials/extra-lazy-associations.rst
```diff
@@ -38,9 +37,11 @@ easily using a combination of ``count`` and ``slice``.
 
 .. warning::
 
-   Using ``removeElement`` will directly issue DELETE queries to the database.
-   This circumvents the flush operation and might run outside a transactional
-   boundary if you don't create one yourself.
+   ``removeElement`` directly issued DELETE queries to the database from
+   version 2.4.0 to 2.7.0.  This circumvents the flush operation and might run
+   outside a transactional boundary if you don't create one yourself. We
+   consider this a critical bug in the assumptio of how the ORM works and
```
stloyd: Typo: `assumption`

lcobucci:
Didn't @guilhermeblanco say that he was going to look into an alternative to removing support for it? I'd say it might be better to wait or coordinate things with him...

guilhermeblanco:
I'm a bit slow catching up on work, but this is very high on priority list... =)

beberlei:
@guilhermeblanco @lcobucci still, wouldn't it be necessary to revert this for the 2.7 branch as a bug and re-introduce in 2.8 with a working solution? Especially since a solution is going to add a good amount of extra complecity with a new type of local entity cache in persister or UoW.

I feel this is a critical bug doing a delete outside flush, as its completely against expectations. I wouldn't want to drag this out.

lcobucci:
I wouldn't touch it, tbqh. It's a known (non-ideal) behaviour. But many people do explicit transactions already. Do agree that we might have to target 2.8.x

beberlei:
@lcobucci what do you mean doing explicit transactions? I don't do explicit transactions and let flush handle everything. And i doubt this is known behavior outside of our circle of people wokring on the bug and reporting it.

beberlei:
I am merging this now, because it is a critical bugfix that violates our contract of no write operations outside flush.

ureimers:
Hi @beberlei,

first, thank you for the fix. In my case the direct removal let to update listeners not firing and workaround code like this:

```
// Make sure to initialize the collection so that the campaign relation isn't
// removed directly from the DB (because of possible extra-laziness). Only this
// way we can be sure, that the campaign's update listener will be called.
$this->campaigns->toArray();
$this->campaigns->removeElement($campaign);
```

But as your fix isn't in the master branch I'm worried, that it might be added back at a later time. Can you confirm, that I can remove my workarounds and that `removeElement` won't introduce the deletion outside of flush in the future again?

acasademont:
Wow, this fix made a longstanding bug in our app appear.

We had an EXTRA_LAZY associacion with an "orphanRemoval" attribute that should had never been there. This immediate `removeElement` before the main transaction was preventing the orphanRemoval from kicking in and creating a mess. Thank god our e2e tests cached it before going to production...

acasademont:
Maybe a bigger warning in the Release Notes would be appropiate. Composer makes updates too easy these days :p

alanhartless:
I have to agree with @acasademont :-D

I understand why the behavior was changed but just mentioning that this caused a bit of a "fire" in our application and took some sleuth work to figure out what was happening (resulted in an attempt to hydrate a very large collection that caused the process to run out of memory). 

https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/tutorials/extra-lazy-associations.html noted that this behavior was reverted because it was a considered a "critical bug" but has been the expected behavior for years (since 2.4.0 according to the tutorial) and then changed in a .1 patch release. It know it was mentioned on the release notes as a bug but I'd argue it was a pretty decent BC break.

beberlei:
@acasademont @alanhartless Sorry for the late reply on your comments. This popped up again in #8739. I have posted a temporary workaround if you need the old behavior in this comment:

https://github.com/doctrine/orm/issues/8739#issuecomment-853783135

PR #7941, opened 2019-12-01
Allow DQL functions to specify return type

Grafikart:
Some functions have a predefined type and could be resolved correctly by the SQLWalker. I don't want to change things too much but it could be possible to do the same thing for other functions like SUM for instance that would be resolved to float.

This is the follow up of the #7641 PR (I won't delete it this time sorry for the inconvenience)

### DIFF
--- a/lib/Doctrine/ORM/Query/AST/Functions/FunctionNode.php
+++ b/lib/Doctrine/ORM/Query/AST/Functions/FunctionNode.php
@@ -24,7 +24,7 @@ use Doctrine\ORM\Query\AST\Node;
 /**
  * Abstract Function Node.
  *
- * 
+ *
  * @link    www.doctrine-project.org
  * @since   2.0
  * @author  Guilherme Blanco <guilhermeblanco@hotmail.com>
--- a/lib/Doctrine/ORM/Query/AST/TypedExpression.php
+++ b/lib/Doctrine/ORM/Query/AST/TypedExpression.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Query\AST;
+
+use Doctrine\DBAL\Types\Type;
+
+/**
+ * Provides an API for resolving the type of a Node
+ */
+interface TypedExpression
+{
+    public function getReturnType() : Type;
+}
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7941Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7941Test.php
@@ -1,0 +1,101 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use DateTimeImmutable;
+use Doctrine\Tests\OrmFunctionalTestCase;
+use function ltrim;
+use function strlen;
+
+/** @group GH7941 */
+final class GH7941Test extends OrmFunctionalTestCase
+{
+    private const PRODUCTS = [
+        ['name' => 'Test 1', 'price' => '100', 'square_root' => '/^10(\.0+)?$/'],
+        ['name' => 'Test 2', 'price' => '100', 'square_root' => '/^10(\.0+)?$/'],
+        ['name' => 'Test 3', 'price' => '100', 'square_root' => '/^10(\.0+)?$/'],
+        ['name' => 'Test 4', 'price' => '25', 'square_root' => '/^5(\.0+)?$/'],
+        ['name' => 'Test 5', 'price' => '25', 'square_root' => '/^5(\.0+)?$/'],
+        ['name' => 'Test 6', 'price' => '-25', 'square_root' => '/^5(\.0+)?$/'],
+    ];
+
+    protected function setUp() : void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([GH7941Product::class]);
+
+        foreach (self::PRODUCTS as $product) {
+            $this->_em->persist(new GH7941Product($product['name'], $product['price']));
+        }
+
+        $this->_em->flush();
+        $this->_em->clear();
+    }
+
+    /** @test */
+    public function typesShouldBeConvertedForDQLFunctions() : void
+    {
+        $query = $this->_em->createQuery(
+            'SELECT
+                 COUNT(product.id) as count,
+                 SUM(product.price) as sales,
+                 AVG(product.price) as average
+             FROM ' . GH7941Product::class . ' product'
+        );
+
+        $result = $query->getSingleResult();
+
+        self::assertSame(6, $result['count']);
+        self::assertSame('325', $result['sales']);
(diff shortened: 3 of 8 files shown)


lib/Doctrine/ORM/Query/AST/Functions/FunctionNode.php
```diff
@@ -43,4 +45,9 @@ public function dispatch($sqlWalker)
      * @return void
      */
     abstract public function parse(Parser $parser);
+
+    public function getReturnType(): Type
+    {
+        return Type::getType(Type::STRING);
```
ostrolucky: So copying my comment here: 

Why defining this when it's default fallback anyway?

ostrolucky (commented):
Copying my review here 

It's a nice, simple patch solving the problem, however I'm not sure we should bite the bullet with floats. COUNT() and LENGTH() as integer types are fine, but casting to float in PHP will mean losing accuracy. It's also quite strange to make eg. ABS(1) return 1.0. Also, this a BC break and as such should be documented.

If more dpeople could share their opinion about what I writtent, that would be nice too

lib/Doctrine/ORM/Query/AST/Functions/FunctionNode.php
Grafikart: I copied this here so child classes automatically implement the TypedExpression interface.

Grafikart:
@ostrolucky If you want I could postpone the change on ABS() so the new feature can be merged without BC (knowing we would have to change it in the future).

ostrolucky:
Before doing that, let's see what other doctrine maintainers think. WDYT @doctrine/doctrinecore ?

lib/Doctrine/ORM/Query/AST/Functions/FunctionNode.php
ostrolucky: Why should they automatically implement it though? Then they need to override it, instead of cleanly implement it.

Grafikart: This is an opinionated decision (I didn't want to specify "implements ...." on subclass). If it really bother I won't argue and fix my code according to your review

SenseException:
master is the next major release. Isn't this a fitting time to introduce a BC break?

lib/Doctrine/ORM/Query/AST/Functions/FunctionNode.php
lcobucci: @Grafikart I understand your decision. However, having the concrete classes implementing the interface makes things much more explicit - especially if you consider that we have many extensions providing platform-specific functions.

Please do apply the changes to have the type only on the concrete classes :+1:

lib/Doctrine/ORM/Query/AST/TypedExpression.php
```diff
@@ -0,0 +1,16 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Query\AST;
+
+use Doctrine\DBAL\Types\Type;
+
+/**
+ * Provides an API for resolving the type of a Node
+ *
+ * @internal
```
lcobucci: As mentioned in another comment we have extensions that could benefit from having DBAL handling the type conversion, so I think this doesn't necessarily need to be marked as `@internal`.

tests/Doctrine/Tests/ORM/Query/SelectSqlMappingTest.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public function assertScalarTypes(string $dqlToBeTested, array $types)
+    {
+        $query = $this->em->createQuery($dqlToBeTested);
+        $query
+            ->setHint(ORMQuery::HINT_FORCE_PARTIAL_LOAD, true)
+            ->useQueryCache(false);
+        $r = new ReflectionObject($query);
+        $method = $r->getMethod('getResultSetMapping');
+        $method->setAccessible(true);
```
lcobucci: I think that we should turn this test into a functional one, avoiding the use of reflection - which makes the test somewhat fragile.

lcobucci (changes requested):
This changeset is quite nice indeed. The title is a bit misleading since the previous behaviour was working as intended, I'd consider this as an improvement of the API.

The usage of floats for SUM(), AVG(), and ABS() might indeed cause some trouble with precision (which is why we don't do casts for decimal columns). I'd say we should leave them as string for now and have this targeting v2.8.0.

lib/Doctrine/ORM/Query/AST/TypedExpression.php
Grafikart: This is fixed

tests/Doctrine/Tests/ORM/Query/SelectSqlMappingTest.php
Grafikart: I don't know how to solve this test. I can't find another entry point to reach the resultSetMapping to check if everything is working as expected.

lcobucci: Don't worry, we can take care of that. Thanks a lot for your help!

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7941Test.php
```diff
@@ -0,0 +1,108 @@
...
+                 SUM(product.price) as sales,
+                 AVG(product.price) as average
+             FROM ' . GH7941Product::class . ' product'
+        );
+
+        $result = $query->getSingleResult();
+
+        self::assertSame(6, $result['count']);
+        self::assertSame('325', $result['sales']);
+        self::assertRegExp('/^54\.16+7$/', $result['average']);
```
lcobucci: @Grafikart this is what I meant with a more functional approach :smile: 

The main advantage is that we only rely on the public API here, which makes it easier to detect BC-breaks.

By using `self::assertSame()` we check that both value and type match expected value and `self::assertRegExp()` only accepts strings.

lcobucci (approved):
This is good to go IMHO, it would be nice to have another pair of eyes here.

It might be good to document a minor behaviour change regarding the return type of `COUNT()` and `LENGTH()` - there's no need for casting it to integer anymore. What do you think @doctrine/team-orm-documentation?

ostrolucky (approved):

lcobucci:
@Grafikart thanks! :ship:

SenseException:
In some cases functions like `COUNT` are already documented with an integer return type: https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/dql-doctrine-query-language.html#pure-and-mixed-results
This is now some kind of self-fulfilling prophecy (even though the change is for 2.8).

VincentLanglet:
@Grafikart When I use `COUNT` with `getSingleResult`, I now get
```
[ 0 => 42 ]
```
But if I use `getSingleScalarResult` I get `'42'`.

Shouldn't the result be consistent ?

Issue #7893, opened 2019-11-08
Commits silently fail (Percona cluster + Galera plugin) 


Hamael18:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.4

#### Summary
DBAL have been recently updated. It's now 2.10 (https://github.com/doctrine/dbal/pull/3588).
It fix issue with the commit() method. But it's not handled in ORM.

#### Current behavior

If commit() fails without exception (commit() return false but no exception), there is no information about that. 

#### Expected behavior

Throw an exception if commit() fails and return false.

lcobucci:
@Hamael18 thanks. This seems to be an easy thing to solve and to isolate in a unit test.
Planning it for the next patch milestone :+1:

guilhermeblanco:
@lcobucci Unfortunately this needs to be moved to 2.7, since it requires a bump in the doctrine/dbal version from ^2.6 to ^2.10 and we're not able to do this on a patch level release.

I'm moving this to 2.7, which can tolerate dependency bumps like the required one here, and we can work on a quick/simple patch after we have 2.6.5 out of the door.

chosroes:
```php
    if (!$conn->commit()) {
        throw new OptimisticLockException('Commit failed', $entity);
    }
} catch (Throwable $e) {
    $this->em->close();

    if ($conn->isTransactionActive()) {
        $conn->rollBack();
    }
```
Small proposal for UnitOfWork::commit()
I'll not have time to made a PR with unit tests, but hope this will help someone else for a better proposal

lcobucci:
@chosroes thanks!

lcobucci:
I'm moving this to version 2.8.0 because DBAL requires PHP 7.2+ and we'll release 2.7.0 still supporting PHP 7.1+.

cs278:
Looks like this should have been closed by 1da002ca2f84d9d5da11b4c60e24d957cb8f6c16?

beberlei:
Yes, fixed by https://github.com/doctrine/orm/pull/7946

PR #7946, opened 2019-12-06
Throw OptimisticLockException when connection::commit() returns false

chosroes:
Proposal for #7893 in order to handle silently failed commits


SenseException (changes requested):
Thank you for your PR. Please add unit tests that are covering your changes.

chosroes:
> Thank you for your PR. Please add unit tests that are covering your changes.

I'd like to but there are currently no tests for exception in commit process, or I should create another connection mock with commit that return false ?

chosroes:
Tests added and checks passed

SenseException:
Thank you for the changes. You introduced changes in .scrutinizer.yml, the travis file and composer.json. Are these changes relevant to your feature?

chosroes:
Yes since doctrine/dbal now needs at least php 7.2

tests/Doctrine/Tests/Mocks/ConnectionCommitFailMock.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Doctrine\Tests\Mocks;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\Statement;
+
+/**
+ * Mock class for Connection.
+ */
+class ConnectionCommitFailMock extends ConnectionMock
```
ostrolucky: Not sure why nobody commented here yet, but this was talked about a bit internally and consensus seemed to be that doctrine team doesn't like dedicated mock class for such simple, non-reusable mock like is a case here. PHPunit mock was recommended to be used instead, like
```php
$this->_connectionMock = $this->getMockBuilder(ConnectionMock::class)
  ->setConstructorArgs([[], new DriverMock()])
  ->setMethods(['commit'])
  ->getMock();

$this->_connectionMock->method('commit')->willReturn(false);
```

although if you prefer, I would also suggest anonymous class such as
```php
$this->_connectionMock = new class([], new DriverMock()) extends ConnectionMock {
  public function commit()
    {
        return false;
    }
};
```
I think complaint was mainly about polluting project/namespace with one-off classes.

ostrolucky (changes requested):
Lot of tests fail for me locally. Not sure why travis did not start here, but tests must run

tests/Doctrine/Tests/Mocks/ConnectionCommitFailMock.php
SenseException: I wouldn't use an anonymous class extending one of those Mock classes because this will still have the disadvantages mentioned in doctrine/dbal#3470. I would prefer a cut and that we break with the usage of custom Mock classes, so that it will be easier to remove them in another PR.

ostrolucky: Only first point from https://github.com/doctrine/dbal/pull/3470 apply here. And refactoring of signatures via phpstorm is piece of cake, unlike handling stuff like method renames and change of constructors in phpunit mocks. And readability, SA analysis support. But since we already know type hints are going to be added, it indeed kinda makes sense to use phpunit mock here to avoid having to touch this in upper branches.

SenseException: The third point also apply here for implicit dependencies between tests using `ConnectionMock`. Invocations won't be tracked on it either. I see more advantages between those two solutions in the phpunit mock.

ostrolucky: that point applies for external classes, not for anonymous ones used in single test case

chosroes: Ok for me

chosroes:
@ostrolucky I think there is a problem with travis. The request is in status "Abuse detected"...
I had no account in travis CI. I created one just now...
https://travis-ci.org/doctrine/orm/requests?utm_medium=notification&utm_source=github_status

By the way, I changed my unit test. There was already lot of tests failures in the 2.8.x branch for me locally. So I don't know what to do with that.

greg0ire:
Apparently the issue was transient, here is your build!

ostrolucky:
Aand it shows failures I mentioned.

chosroes:
I can't get travis to work with me...

chosroes:
I've got -0.164% code coverage 😒

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -424,10 +425,18 @@ public function commit($entity = null)
                 }
             }
 
-            $conn->commit();
+            // Commit failed silently
+            if ($conn->commit() === false) {
```
SenseException: In the next major release of DBAL `$conn->commit()` won't return a boolean anymore. Will a silent failure still be a problem there?

chosroes: In 3.x branch of dbal, the commit function is still returning a boolean

SenseException: You're right. I was looking in the master branch. https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L1083

SenseException:
Is this still considered a WIP?

chosroes:
@SenseException Not for me

chosroes:
Any update ?

SenseException:
I'm undecided for the usage of `$conn->commit() === false`. It introduces the usage of an already existing boolean return value in that method. It is true that 3.x is still using it, but master has changed it to `void`. Maybe I'm planning to far into the future, because we can deprecate it in the last release of 3.x or respectively implement that optimistic lock case differently in master. //cc @ostrolucky 

Otherwise I'm fine with the changes.

chosroes:
@SenseException I don't know for master, the current commit without error may have be solve. But meanwhile, we have to do with the current interface that can return false

ostrolucky (approved):
> master has changed it to void

Good point. Fortunately that won't break this code, just make bool condition dead. I expect phpstan will alert us on it in future so we can remove it then.

SenseException (approved):
If postponing the issue with bool -> void is acceptable for this PR, then I'll give my approval.

PR #7947, opened 2019-12-06
Update custom-mapping-types.rst

glhms:
If you run tests, it will add type multiple times, because its static.

you must check if its already added.

### DIFF
--- a/docs/en/cookbook/custom-mapping-types.rst
+++ b/docs/en/cookbook/custom-mapping-types.rst
@@ -73,7 +73,10 @@ method. See the following example:
     // ...
 
     // Register my type
-    Type::addType('mytype', 'My\Project\Types\MyType');
+    
+    if(!array_key_exists('mytype', Type::getTypesMap())){
+        Type::addType('mytype', 'My\Project\Types\MyType');
+    }
 
 To convert the underlying database type of your
 new "mytype" directly into an instance of ``MyType`` when performing


greg0ire:
It says "in bootstrapping code", and I think that implies this piece of code is only supposed to be run once.

glhms:
Lets say you put it in the kernel constructor, which should run only once.

If you run multiple test cases, it makes new kernel, and everything is running again.

But the types are saved between test cases, because its static.

In the second test you will get a mapping exception because your type is already added.

Also, we can simply check if it is already added with Type::hasType(), then we do not need array_key_exists().

greg0ire:
> Lets say you put it in the kernel constructor, which should run only once.

As you yourself said, it does not run only once, and thus does not qualify as bootstrapping code. In a phpunit context, bootstrapping code will be understood as running once too: https://phpunit.readthedocs.io/en/8.4/configuration.html#the-bootstrap-attribute

What you are describing is code running on `setUp`, not because of `bootstrap`.

SenseException (changes requested):
As @greg0ire mentioned before this code is supposed to be put into a bootstrapper, that is only run once. This should be enough information for the ORM documentation. But it is only mentioned as a code comment and I suppose not everyone might be familiar with a "bootstrap" concept.

@glhms Can you please add a description that this has to be added into bootstrapping code and that bootstrapping code is supposed to run only once? This way we don't need to adapt the code example with an if-block like you introduced first. The new description addition should be added before the code example.

PR #7948, opened 2019-12-07
Add stage that runs before Test to allow fast failures on Sqlite/phpcs

beberlei:
Similar to how DBAL runs Smoke Testing stage, allows Travis exit early when failures existing either SQLite testsuite or Coding Standards as found by git-differ (not against whole codebase).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ dist: trusty
 sudo: false
 language: php
 
+stages:
+    - Smoke Testing
+    - Test
+    - Code Quality
+
 php:
   - 7.1
   - 7.2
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -27,6 +32,34 @@ script:
 
 jobs:
   include:
+    - stage: Smoke Testing
+      php: 7.3
```
lcobucci: Is there a particular reason for using `7.3` here?

.travis.yml
```diff
@@ -27,6 +32,34 @@ script:
...
+      env: DB=sqlite
+
+    - stage: Smoke Testing
+      if: type = pull_request
+      env: DB=none PULL_REQUEST_CODING_STANDARDS
+      php: 7.1
```
lcobucci: We could remove this line since Travis uses the lowest PHP version in the matrix by default. I know you just moved stuff but that would save us a bit of work in the future (when dropping PHP 7.1 for example :smile:)

.travis.yml
```diff
@@ -27,6 +32,34 @@ script:
...
+      install: travis_retry composer install --prefer-dist
+      script:
+        - |
+          if [ $TRAVIS_BRANCH != "master" ]; then
+            git remote set-branches --add origin $TRAVIS_BRANCH;
+            git fetch origin $TRAVIS_BRANCH;
+          fi
+        - git merge-base origin/$TRAVIS_BRANCH $TRAVIS_PULL_REQUEST_SHA || git fetch origin +refs/pull/$TRAVIS_PULL_REQUEST/merge --unshallow
+        - wget https://github.com/diff-sniffer/git/releases/download/0.2.0/git-phpcs.phar
```
lcobucci: We can bump this to the latest diff-sniffer/git release (0.3.2) - separate commit would be fine to me but it can be done in another PR if you prefer :+1:

lcobucci (commented):
This looks like an awesome thing, thanks @beberlei :+1: 

Just a few questions before we get it merged in.

lcobucci:
@beberlei thanks! :ship:

PR #7950, opened 2019-12-09
fixed hydrating relations with typed, non nullable properties

### LINKED ISSUE #7945: PHP 7.4: Object Hydrator is trying to access related entity without initialization when fetch mode = "EAGER"
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7.0

#### Summary
Error when calling ```\Doctrine\ORM\EntityManager::find```:
```
[Error] Typed property App\TestEnt::$relEnt must not be accessed before initialization  
                                                                                                             
#1  /var/www/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php:432
#2  /var/www/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php:162
#3  /var/www/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php:153
#4  /var/www/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php:726
#5  /var/www/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php:736
#6  /var/www/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/EntityManager.php:469
```

#### How to reproduce
<details>
  <summary>Entities</summary>
 
```php
<?php

declare(strict_types=1);

namespace App\test;

/**
 * @ORM\Entity()
 */
class TestEnt
{
    /**
     * @ORM\Id
     * @ORM\Column(name="id", type="guid")
     */
    private string $id;

    /**
     * @ORM\OneToOne(targetEntity="TestRelEnt", fetch="EAGER", cascade={"persist"})
     */
    private TestRelEnt $relEnt;

    public function __construct(string $id, TestRelEnt $relEnt)
    {
        $this->id = $id;
        $this->relEnt = $relEnt;
    }
}
  ```

```php
<?php

declare(strict_types=1);

namespace App\test;

/**
 * @ORM\Entity()
 */
class TestRelEnt
{
    /**
     * @ORM\Id
     * @ORM\Column(name="id", type="integer")
     * @ORM\GeneratedValue()
     */
    private string $id;
}
  ```
</details>

<details>
  <summary>Code</summary>
 
```php
$id = Uuid::uuid4()->toString();
$testEnt = new TestEnt($id, new TestRelEnt());
$this->entityManager->persist($testEnt);
$this->entityManager->clear();
$this->entityManager->find(TestEnt::class, $id);
  ```
</details>

kuraobi:
When initializing entities with relations, the ObjectHydrator calls ReflectionProperty's `->getValue()` method to get the existing value. This will fail on PHP7.4 if a the property is not initialized, for example if you have a blog entity with a `private Collection $articles;` property. Thus, we must call the new `->isInitialized()` method to check if the property is initialized before calling `->getValue()`.

May fix #7945

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
@@ -26,6 +26,9 @@ use Doctrine\ORM\PersistentCollection;
 use Doctrine\ORM\Query;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\ORM\Proxy\Proxy;
+use function assert;
+use function method_exists;
+use function spl_object_hash;
 
 /**
  * The ObjectHydrator constructs an object graph out of an SQL result set.
@@ -182,9 +185,14 @@ class ObjectHydrator extends AbstractHydrator
      */
     private function initRelatedCollection($entity, $class, $fieldName, $parentDqlAlias)
     {
-        $oid      = spl_object_hash($entity);
-        $relation = $class->associationMappings[$fieldName];
-        $value    = $class->reflFields[$fieldName]->getValue($entity);
+        $oid       = spl_object_hash($entity);
+        $relation  = $class->getAssociationMapping($fieldName);
+        $reflField = $class->getReflectionProperty($fieldName);
+        $value     = null;
+
+        if (! method_exists($reflField, 'isInitialized') || $reflField->isInitialized($entity)) {
+            $value = $reflField->getValue($entity);
+        }
 
         if ($value === null || is_array($value)) {
             $value = new ArrayCollection((array) $value);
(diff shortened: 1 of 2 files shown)


lcobucci (changes requested):
This looks similar to the approach that was discussed in https://github.com/doctrine/orm/pull/7857? Can you please try to read the history there and perhaps send a reproducing test case to that fork so that we can be sure it will address everything?

leongersen:
As per [#7857 (comment)](https://github.com/doctrine/orm/pull/7857#issuecomment-575537801), this was fixed in `doctrine/persistence`, so this can probably be closed.

kuraobi:
OK I've tested again with the fix in `doctrine/persistence` and it seems to be fixed. My unit test still fails because `PersistenceCollection::initialized` is `true` in the original entity and `false` after hydrating, but this seems normal.

PR #7953, opened 2019-12-12
Address deprecations from persistence

greg0ire:
A backwards-compatibility layer has been added to persistence to help
consumers move to the new namespacing. It is based on class aliases,
which means the type declaration changes should not be a BC-break: types
are the same.
See https://github.com/doctrine/persistence/pull/71
    
This means:
- using the new namespaces
- adding autoload calls for new types to types that may be extended and
use persistence types in type declarations of non-constructor methods,
so that signature compatibility is recognized by old versions of php.
More details on this at
https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf

# How can I test this on my project?

```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/doctrine-orm
composer require doctrine/orm "dev-address-persistence-deprecations as v2.7.0"
```

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -47,7 +47,7 @@ Whole Doctrine\ORM\Tools\Export namespace with all its members have been depreca
 ## Deprecated `Doctrine\ORM\Proxy\Proxy` marker interface
 
 Proxy objects in Doctrine ORM 3.0 will no longer implement `Doctrine\ORM\Proxy\Proxy` nor
-`Doctrine\Common\Persistence\Proxy`: instead, they implement
+`Doctrine\Persistence\Proxy`: instead, they implement
 `ProxyManager\Proxy\GhostObjectInterface`.
 
 These related classes have been deprecated:
@@ -439,17 +439,17 @@ above you must implement these new methods.
 
 ## Metadata Drivers
 
-Metadata drivers have been rewritten to reuse code from Doctrine\Common. Anyone who is using the
+Metadata drivers have been rewritten to reuse code from `Doctrine\Persistence`. Anyone who is using the
 `Doctrine\ORM\Mapping\Driver\Driver` interface should instead refer to
-`Doctrine\Common\Persistence\Mapping\Driver\MappingDriver`. Same applies to
+`Doctrine\Persistence\Mapping\Driver\MappingDriver`. Same applies to
 `Doctrine\ORM\Mapping\Driver\AbstractFileDriver`: you should now refer to
-`Doctrine\Common\Persistence\Mapping\Driver\FileDriver`.
+`Doctrine\Persistence\Mapping\Driver\FileDriver`.
 
 Also, following mapping drivers have been deprecated, please use their replacements in Doctrine\Common as listed:
 
- *  `Doctrine\ORM\Mapping\Driver\DriverChain`       => `Doctrine\Common\Persistence\Mapping\Driver\MappingDriverChain`
- *  `Doctrine\ORM\Mapping\Driver\PHPDriver`         => `Doctrine\Common\Persistence\Mapping\Driver\PHPDriver`
- *  `Doctrine\ORM\Mapping\Driver\StaticPHPDriver`   => `Doctrine\Common\Persistence\Mapping\Driver\StaticPHPDriver`
+ *  `Doctrine\ORM\Mapping\Driver\DriverChain`       => `Doctrine\Persistence\Mapping\Driver\MappingDriverChain`
+ *  `Doctrine\ORM\Mapping\Driver\PHPDriver`         => `Doctrine\Persistence\Mapping\Driver\PHPDriver`
+ *  `Doctrine\ORM\Mapping\Driver\StaticPHPDriver`   => `Doctrine\Persistence\Mapping\Driver\StaticPHPDriver`
 
 # Upgrade to 2.2
 
@@ -538,7 +538,7 @@ Previously EntityManager#find(null) returned null. It now throws an exception.
 
 ## Interface for EntityRepository
 
-The EntityRepository now has an interface Doctrine\Common\Persistence\ObjectRepository. This means that your classes that override EntityRepository and extend find(), findOneBy() or findBy() must be adjusted to follow this interface.
+The EntityRepository now has an interface Doctrine\Persistence\ObjectRepository. This means that your classes that override EntityRepository and extend find(), findOneBy() or findBy() must be adjusted to follow this interface.
 
 ## AnnotationReader changes
 
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "doctrine/dbal": "^2.9.3",
         "doctrine/event-manager": "^1.1",
         "doctrine/instantiator": "^1.3",
-        "doctrine/persistence": "^1.2",
+        "doctrine/persistence": "^1.3.3",
         "ocramius/package-versions": "^1.2",
         "symfony/console": "^3.0|^4.0|^5.0"
     },
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -25,8 +25,6 @@ use Doctrine\Common\Annotations\CachedReader;
 use Doctrine\Common\Annotations\SimpleAnnotationReader;
 use Doctrine\Common\Cache\ArrayCache;
 use Doctrine\Common\Cache\Cache as CacheDriver;
-use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
-use Doctrine\Common\Persistence\ObjectRepository;
 use Doctrine\Common\Proxy\AbstractProxyFactory;
 use Doctrine\ORM\Cache\CacheConfiguration;
 use Doctrine\ORM\Mapping\ClassMetadataFactory;
@@ -39,6 +37,9 @@ use Doctrine\ORM\Mapping\NamingStrategy;
 use Doctrine\ORM\Mapping\QuoteStrategy;
 use Doctrine\ORM\Repository\DefaultRepositoryFactory;
 use Doctrine\ORM\Repository\RepositoryFactory;
+use Doctrine\Persistence\Mapping\Driver\MappingDriver;
+use Doctrine\Persistence\ObjectRepository;
+use function interface_exists;
 
 /**
  * Configuration container for all configuration options of Doctrine.
@@ -676,7 +677,7 @@ class Configuration extends \Doctrine\DBAL\Configuration
      *
      * @return void
      *
-     * @throws ORMException If not is a \Doctrine\Common\Persistence\ObjectRepository
+     * @throws ORMException If $classname is not an ObjectRepository.
      */
     public function setDefaultRepositoryClassName($className)
     {
@@ -918,3 +919,5 @@ class Configuration extends \Doctrine\DBAL\Configuration
         $this->_attributes['defaultQueryHints'][$name] = $value;
     }
 }
+
+interface_exists(MappingDriver::class);
--- a/lib/Doctrine/ORM/Event/OnClassMetadataNotFoundEventArgs.php
+++ b/lib/Doctrine/ORM/Event/OnClassMetadataNotFoundEventArgs.php
@@ -19,9 +19,10 @@
 
 namespace Doctrine\ORM\Event;
 
-use Doctrine\Common\Persistence\Event\ManagerEventArgs;
-use Doctrine\Common\Persistence\Mapping\ClassMetadata;
-use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\Persistence\Event\ManagerEventArgs;
+use Doctrine\Persistence\Mapping\ClassMetadata;
+use Doctrine\Persistence\ObjectManager;
+use function interface_exists;
 
 /**
  * Class that holds event arguments for a `onClassMetadataNotFound` event.
@@ -84,3 +85,4 @@ class OnClassMetadataNotFoundEventArgs extends ManagerEventArgs
     }
 }
 
+interface_exists(ClassMetadata::class);
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "22be7b4b42da2931c1033f8818a0caa9",
+    "content-hash": "3ecc0039f7919112197e4fb876a5184d",
     "packages": [
         {
             "name": "doctrine/annotations",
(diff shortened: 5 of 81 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
         "doctrine/dbal": "^2.9.3",
         "doctrine/event-manager": "^1.1",
         "doctrine/instantiator": "^1.3",
-        "doctrine/persistence": "^1.2",
+        "doctrine/persistence": "^1.3@dev",
```
alcaeus: This will be changed in the coming days once we release 1.3, but we need to dogfood here to make sure the deprecation layer works as expected.

alcaeus (dismissed):

composer.json
alcaeus: 1.3.0 has been released, this can be changed back to stable.

greg0ire: I've put 1.3.1, please restart the build after releasing it.

alcaeus: 1.3.1 has been released.

alcaeus (dismissed):
LGTM. @nicolas-grekas do you want to give this another look before we merge it?

lib/Doctrine/ORM/Event/OnClassMetadataNotFoundEventArgs.php
```diff
@@ -84,3 +85,4 @@ public function getClassName()
     }
 }
 
+interface_exists(ClassMetadata::class);
```
ostrolucky: This warrants explanation

lib/Doctrine/ORM/Configuration.php
```diff
@@ -918,3 +919,5 @@ public function setDefaultQueryHint($name, $value)
         $this->_attributes['defaultQueryHints'][$name] = $value;
     }
 }
+
+interface_exists(MappingDriver::class);
```
ostrolucky: this too

lib/Doctrine/ORM/Event/OnClassMetadataNotFoundEventArgs.php
greg0ire: I'm going to amend my commit message

ostrolucky (commented):
I think PR description needs to be updated here

lib/Doctrine/ORM/Configuration.php
```diff
@@ -676,7 +677,7 @@ public function getFilterClassName($name)
      *
      * @return void
      *
-     * @throws ORMException If not is a \Doctrine\Common\Persistence\ObjectRepository
+     * @throws ORMException If not is a ObjectRepository.
```
lcobucci:
```suggestion
     * @throws ORMException If not is an ObjectRepository.
```

lib/Doctrine/ORM/Configuration.php
lcobucci: This is a side-effect of a non-optimal decision done in the upstream that I'd love to avoid by all means necessary - yes I read the blog post and I disagree with the approach.

lcobucci (changes requested):
This should target v2.8 since deprecations aren't bugs nor errors, they're just warnings telling us that we have to change things before upgrading and we don't bump up dependencies in patches.

lib/Doctrine/ORM/Configuration.php
greg0ire: I think it could be avoided by bumping the version of php to something that does not require that piece of black magic… I didn't manage to figure out which one work yet though.

greg0ire: Do you disagree with the whole approach or just the last part? If you know of other approaches, please share them.

lib/Doctrine/ORM/Configuration.php
greg0ire: I didn't notice until now, but this is still a bit weird :sweat_smile:

alcaeus:
> This should target v2.8 since deprecations aren't bugs nor errors

@lcobucci is there a timeline for 2.8?

lib/Doctrine/ORM/Configuration.php
alcaeus: > This is a side-effect of a non-optimal decision done in the upstream that I'd love to avoid by all means necessary

The team a while back decided to go ahead with the namespace change from `Doctrine\Common\Persistence` to `Doctrine\Persistence`, against the advice from a few people warning of the impact of such a change (not specifically for persistence, but for annotations where the impact is pretty much the same, see also https://github.com/doctrine/annotations/pull/75#discussion_r56872787). This polyfill layer is the only option we have to avoid a hard BC break that will either keep people from upgrading to v2 or other projects (e.g. Symfony) having to maintain their own compatibility layers (which previously caused protests from other library authors).

So while you may disagree with the approach (as others have done), there has yet to be a specific suggestion that allows for such a significant change in a widely used library without the need for this kind of glue code. If you have a suggestion, by all means, let us know. Until then, this is the way this will have to be done.

greg0ire: @lcobucci as I said at the end of my blog post, I'm not a huge fan of this approach either: it's ugly. But IMO it's far better than having no BC-layer at all. I think there should be way more of them in Doctrine and I'm afraid several Doctrine packages are going to be quite a pain to migrate to their next major version. That being said, I don't know that much about the differences between branches in the ORM, for instance, so I hope I'm wrong.

lcobucci: I agree on the need for a BC-layer (no need for discussions there).

The problem is the usage of class/interface aliases, having explicit interfaces/classes extending the new stuff is much cleaner, is cacheable by opcache (and preoloaded properly), doesn't force us to have these calls, and doesn't break anything on 3rd party.

lcobucci:
> @lcobucci is there a timeline for 2.8?

@alcaeus nope.

lib/Doctrine/ORM/Configuration.php
nicolas-grekas: Note that class aliases *are* compatible with opcache/preloading.

greg0ire: > having explicit interfaces/classes extending the new stuff is much cleaner

Ok so let's say you have `LegacyFoo extends NewShinyFoo`.

If you have an interface that has a method with a type declaration, and you change it to `NewShinyFoo`, you have a BC break, because any implementor will restrict the type to `LegacyFoo` instead of widening it, which is possible in theory but only in php 7.4

lcobucci: @nicolas-grekas not according to the RFC:

> Only classes without unresolved parent, interfaces, traits and constant values may be preloaded. If a class doesn't satisfy to this condition, it's stored in opcache SHM as a part of corresponding PHP script in the same way as without preloading. Also, only top-level entities that are not nested within control structures (e.g. if()…) may be preloaded.

I get a `Notice` when trying to preload stuff, like `[16-Dec-2019 16:53:42 UTC] PHP Warning:  Can't preload unlinked class TestC: Unknown interface TestB in /app/C.php on line 3`. Please let me know if I'm doing something wrong.

<details>
  <summary>Reproducible script</summary>

*A.php*
```php
<?php

interface TestA
{
}

class_alias(TestA::class, TestB::class);
``` 

*C.php*
```php
<?php

final class TestC implements TestB
{
    public function run(): void
    {
        echo 'blah', PHP_EOL;
    }
}
```

*D.php*
```php
<?php

require 'A.php';
require 'C.php';

(new TestC())->run();
```

*preload.php*
```php
<?php

if (! opcache_compile_file(__DIR__ . '/A.php')) {
    trigger_error("Preloading Failed", E_USER_ERROR);
}

if (! opcache_compile_file(__DIR__ . '/C.php')) {
    trigger_error("Preloading Failed", E_USER_ERROR);
}
```

*Dockerfile*
```Dockerfile
FROM php:7.4-alpine

WORKDIR /app

RUN { \
    echo 'zend.assertions=-1'; \
    echo 'assert.exception=1'; \
    echo 'opcache.enable_cli=1'; \
    echo 'opcache.max_wasted_percentage=10'; \
    echo 'opcache.validate_timestamps=0'; \
    echo 'opcache.enable_file_override=1'; \
    echo 'opcache.optimization_level=0xFFFFFFFF'; \
    echo 'opcache.preload=/app/preload.php'; \
    echo 'opcache.preload_user=root'; \
    echo 'error_reporting=E_ALL & ~E_DEPRECATED & ~E_USER_DEPRECATED & ~E_STRICT'; \
    echo 'display_errors=1'; \
    echo 'log_errors=1'; \
    echo 'error_log=/dev/stderr'; \
} | tee /usr/local/etc/php/conf.d/10-prod.ini

RUN apk add --no-cache ${PHPIZE_DEPENDENCIES} \
    && docker-php-ext-install opcache

COPY A.php C.php D.php preload.php ./

CMD ["php", "D.php"]
```

Run:

```bash
$ docker build -t preload-test . && docker run --rm preload-test
```
</details>

nicolas-grekas: The commit is quite new:
https://github.com/php/php-src/commit/baf3a9133bf0bc7069e5191b780c221485b2dda6

If you already have a recent enough build, maybe it works with "require" but not with "opcache_compile_file()"?

lcobucci: @nicolas-grekas that targets PHP 7.4.1, so we don't have something that works with stable the version. Good to know it's being addressed, though.

lcobucci: [Nikita confirmed that `class_alias()` will never work with `opcache_compile_file()`](https://github.com/php/php-src/commit/baf3a9133bf0bc7069e5191b780c221485b2dda6#commitcomment-36450273). Using `require` isn't an option since it executes the code, which is quite undesired.

Anyhow, thanks for your input @nicolas-grekas.

nicolas-grekas: > Nikita confirmed that class_alias() will never work with opcache_compile_file().

Great, thanks for the hint. I think this can be resolved in the future - opcache is already able to resolve some functions at compiler time. I'm sure it will happen for class_alias() if it's worth it at some point :)

> Using require isn't an option since it executes the code, which is quite undesired.

But you have zero control over this (you = any lib author). Actually, "require" is the preferred way of configuring preloading in Symfony, *because* it is executing the code, thus it works with conditional classes or similar exotic situations that happen in real life.

lcobucci:
@greg0ire is now targeting 2.8.x since it requires a dependency change :+1:

lib/Doctrine/ORM/Configuration.php
lcobucci: > But you have zero control over this (you = any lib author)

You're completely correct that I don't have control over what users do. However, I do have control if the code I produce is fully compatible with `opcache_compile_file()`.

> Actually, "require" is the preferred way of configuring preloading in Symfony, because it is executing the code, thus it works with conditional classes or similar exotic situations that happen in real life.

Yeah... I rather keep the preoloading of my apps clean of execution, speeding up such process and avoiding unintended side-effects. Especially because is the intended way for as far as the RFC tells (but maybe @nikic or @dstogov can present arguments disproving my understanding).

@nicolas-grekas thanks sharing your PoV (despite our disagreement) :+1:

greg0ire:
I rebased 👍

lib/Doctrine/ORM/Configuration.php
nikic: @lcobucci I think at this point I would recommend the require model. The opcache_compile_file() variant may stop being tenable as projects add property types (and thus introduce circular dependencies). I'm not sure at this point if it's worth investing in making that work (quite hard) when the require based approach can handle this and more transparently.

cybernet:
can it be merged ?

UPGRADE.md
```diff
@@ -435,15 +435,15 @@ above you must implement these new methods.
 
 Metadata drivers have been rewritten to reuse code from Doctrine\Common. Anyone who is using the
```
sstok: Doctrine\Common is no longer used, this should be updated.

sstok (changes requested):

garak:
> # How can I test this on my project?
> 
> ```shell
> composer config repositories.greg0ire vcs https://github.com/greg0ire/doctrine-orm
> composer require doctrine/orm "dev-address-persistence-deprecations as v2.7.0"
> ```

This is not working, it looks like repository is named "greg0ire/doctrine2" instead of "greg0ire/doctrine-orm"

greg0ire:
Thanks for the heads up, I just renamed it!

lib/Doctrine/ORM/Configuration.php
beberlei: For me this is a good sign that the namespace changes away from Common don't add value, but only problems. Its unfortunate that we have to deal with this stuff, but we can still roll back on it to simplify our lives.

greg0ire: They do add a little value IMO: it's now easier to troubleshoot what package is involved when getting error messages that include that kind of FQCN. It's less confusing. Shorter namespaces also help fuzzy matchers perform better when finding types. And it's not like this BC layer should stay here forever.

nicolas-grekas (dismissed):
That's the way to go.
The BC layer will require no maintenance and will fade out with v2 anyway.
Please consider merging asap so we can get free of those deprecation reports that pop up on our CIs or on our repos.

SenseException (dismissed):

beberlei:
We released a 1.3.4 of persistence and rolled back the deprecation messages, and will only emit them when 1.4 is released. The idea is that all external code depends on `doctrine/persistence` version `^1.3` and then the ORM and/or `mongodb-odm` will raise to 1.4 in the near future when they are compatible. This way we will defer the deprecation messages until we ourselves are compatible.

cybernet:
#OMG i’m 20 errors free 👍

composer.json
```diff
@@ -25,7 +25,7 @@
         "doctrine/dbal": "^2.9.3",
         "doctrine/event-manager": "^1.1",
         "doctrine/instantiator": "^1.3",
-        "doctrine/persistence": "^1.2",
+        "doctrine/persistence": "^1.3.3",
```
alcaeus:
```suggestion
        "doctrine/persistence": "^1.4 || ^2.0",
```

greg0ire: I'd do this in a separate PR, once 1.4 is effectively published, cause I'm not going to get a green build otherwise, am I?

alcaeus: Um, yeah. Absolutely.

alcaeus (dismissed):

lugus:
Thanks @greg0ire !

Any news when it will be merged and released?

greg0ire:
If you ask because of the deprecations, upgrade `doctrine/persistence` to get rid of them.

lugus:
Thank you @greg0ire

I asked for the deprecation yes, because i use `Doctrine\ORM\Event\LifecycleEventArgs` in my subscribers and got deprecation messages.

I thought this MR could resolve my issues? 

But maybe i should use `
Doctrine\Persistence\Event\LifecycleEventArgs` instead.

greg0ire:
This MR will definitely resolve the issue

oojacoboo:
Any expectations on when we'll see this merged in and a release?

alcaeus:
> Any expectations on when we'll see this merged in and a release?

Due to the impact this namespace change has on the PHP and Symfony ecosystems, we are re-evaluating whether it is actually worth the pain and suffering it will cause to downstream consumers.

nicolas-grekas:
The pain is past now: we already tagged releases with the new namespace, not sure there is any way back...

alcaeus:
> The pain is past now: we already tagged releases with the new namespace, not sure there is any way back...

Yeah, that's the problem here: can't really go back, don't really want to go through with it. I think we'll just have to bite the bullet, take care of this one and then never do something like this again.

nicolas-grekas:
Yes, can't say otherwise...
TIme to merge here? @greg0ire, rebase needed :)

alcaeus:
> TIme to merge here?

Absolutely. Ping @beberlei - any chance we can move this into 2.7.2?

greg0ire:
Rebased. IMO we should release persistence 2 with only the removal of the BC layer, allow it in all consuming packages, and then requiring it in all consuming packages, removing BC layers in these packages at that moment,  in quick succession. I know Doctrine is against bumping a major version of persistence in a minor, but maybe if we are changing our workflow towards majors that are less painful to upgrade to this could re-considered.

func0der:
I really appreciate all the hard work that went into this.

Do we have a timetable on this? 
The split of `doctrine/common` and the use of `class_alias` actually do break code, because it can not ber properly parsed by php parsers (e.g. `nikic/php-parser` and furthermore `goaop/parser-reflection`). This might be just one of the problems, that is out there as other might still hold them back with `"doctrine/common": "^2.8 <2.9",` :D

How can we help to give this a go?

beberlei (approved):

PR #7957, opened 2019-12-17
Fix version information

lcobucci:
Since `Doctrine\ORM\Version` is now deprecated it shall not be updated on future releases.

This ensures that our CLI tool will present the correct version number.


lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
```diff
@@ -22,7 +22,7 @@
 use Doctrine\DBAL\Tools\Console as DBALConsole;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Tools\Console\Helper\EntityManagerHelper;
-use Doctrine\ORM\Version;
+use PackageVersions\Versions;
```
greg0ire:
```suggestion
use OutOfBoundsException;
use PackageVersions\Versions;
```

lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
```diff
@@ -70,10 +70,11 @@ public static function run(HelperSet $helperSet, array $commands = []) : void
      * @param array                                       $commands
      *
      * @return \Symfony\Component\Console\Application
+     * @throws \OutOfBoundsException
```
greg0ire:
```suggestion
     * @throws OutOfBoundsException
```

greg0ire (commented):
Looks good, I added suggestions to fix the build

greg0ire (approved):

alcaeus:
> Since `Doctrine\ORM\Version` is now deprecated it shall not be updated on future releases.

This is a significant BC break: the only purpose of the class is to expose the currently installed version so it be updated on subsequent releases.

Since the version is exposed as a constant, there's no easy way to dynamically generate this. A workaround could be to create the class dynamically in an autoloader file (reading the version with the help of the ocramius/package-versions package), but that could be considered hacky. @lcobucci would you be open to merging that solution?

Ocramius:
> This is a significant BC break: the only purpose of the class is to expose the currently installed version so it be updated on subsequent releases.

It's not a BC break: the version is frozen, and new consumers (relying on `version_compare()`) will immediately see the deprecation. Old consumers will be fine (no changes there).

PR #7960, opened 2019-12-20
Update advanced-configuration.rst

fernando-henrique-bling:
Removing unnecessary code, the property had already been set at line 31


SenseException (changes requested):
Thank you for your contribution for this outdated part of the docs. Both `setAutoGenerateProxyClasses()` are setting different values. Currently, this method allows a boolean value, which is casted to int and therefore doing something like `setAutoGenerateProxyClasses(ProxyFactory::AUTOGENERATE_ALWAYS)`. This is okay for the development case we have here.

Can you also please remove the `use` for `ProxyFactory`? This old namespace and now ORM  interface isn't used in the code example anymore.

PR #7966, opened 2019-12-27
Fix JoinColumn documentation

### LINKED ISSUE #6184: Default values for JoinColumn->referencedColumnName
Hi guys,

not sure if this is an technical issue or just outdated documentation...

In http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/annotations-reference.html#annref-joincolumn the documentation says
> If it is not specified the attributes name and referencedColumnName are inferred from the table and primary key names.

So a annotation like this seems totally valid to me:
```php
class Entity {
/**
 * @ORM\ManyToOne(targetEntity="Foo\Bar\Entity\Customer", fetch="LAZY")
 * @ORM\JoinColumn(name="created_by", nullable=true)
 * @var \Foo\Bar\Entity\Customer
 */
protected $createdBy;
}
```
At this point i would expect doctrine to figure out which is the primary key of the Customer entity is and resolve everything correctly.
But, sadly it doesn't.
When validating the mapping via command line i got the following error:
```
The referenced column name 'id' has to be a primary key column on the target entity class 'Foo\Bar\Entity\Customer'.
```

In Annotation Mapping documentation (http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/association-mapping.html#mapping-defaults) the behaviour is described completely different than in the first document but exactly the way it really works.

> The referencedColumnName always defaults to “id”, just as in one-to-one or many-to-one mappings.

So, is the first document outdated or is it a bug and the second document is outdated?

MatTheCat:
Fixes https://github.com/doctrine/orm/issues/6184

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -619,22 +619,17 @@ Examples:
 This annotation is used in the context of relations in
 :ref:`@ManyToOne <annref_manytoone>`, :ref:`@OneToOne <annref_onetoone>` fields
 and in the Context of :ref:`@JoinTable <annref_jointable>` nested inside
-a @ManyToMany. This annotation is not required. If it is not
-specified the attributes *name* and *referencedColumnName* are
-inferred from the table and primary key names.
+a @ManyToMany. If this annotation or both *name* and *referencedColumnName*
+are missing they will be computed considering the field's name and the current
+:doc:`naming strategy <namingstrategy>`.
 
-Required attributes:
-
+Optional attributes:
 
 -  **name**: Column name that holds the foreign key identifier for
    this relation. In the context of @JoinTable it specifies the column
    name in the join table.
 -  **referencedColumnName**: Name of the primary key identifier that
-   is used for joining of this relation.
-
-Optional attributes:
-
-
+   is used for joining of this relation. Defaults to *id*.
 -  **unique**: Determines whether this relation is exclusive between the
    affected entities and should be enforced as such on the database
    constraint level. Defaults to false.


docs/en/reference/annotations-reference.rst
```diff
@@ -610,20 +610,17 @@ Examples:
 This annotation is used in the context of relations in
 :ref:`@ManyToOne <annref_manytoone>`, :ref:`@OneToOne <annref_onetoone>` fields
 and in the Context of :ref:`@JoinTable <annref_jointable>` nested inside
-a @ManyToMany. This annotation is not required. If it is not
-specified the attributes *name* and *referencedColumnName* are
-inferred from the table and primary key names.
+a @ManyToMany. If this annotation or both *name* and *referencedColumnName*
```
SenseException: Does the missing name and referenceColumnName mean a JoinColumn like this?

```
    /**
     * @ORM\OneToOne(targetEntity="App\Entity\Bar")
     * @ORM\JoinColumn(nullable=true)
     */
    public $bar;
```

MatTheCat: Yep

SenseException:
This also seems to fit for the latest Doctrine 2 versions. Can you please target your PR to the branch of 2.7?

MatTheCat:
PR rebased

SenseException (dismissed):

MatTheCat:
Is there something left to do?

SenseException (approved):

greg0ire:
Thanks @MatTheCat !

PR #7970, opened 2020-01-07
Fix countBy inconsistency

zorn-v:
Currently if you use something like `$repo->countBy(['visible' => true])` you'll get inarticulate error, but `$repo->findBy(['visible' => true])` will work fine. After investigation you'll find that method should named `count` not `countBy`
It is weird

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -158,6 +158,11 @@ class EntityRepository implements ObjectRepository, Selectable
      */
     public function count(array $criteria)
     {
+        return $this->countBy($criteria);
+    }
+    
+    public function countBy(array $criteria)
+    {
         return $this->em->getUnitOfWork()->getEntityPersister($this->entityName)->count($criteria);
     }
 


SenseException:
I'm not sure what exactly is the issue you're trying to fix here, because of a missing error and error description.

The `EntityRepository` supports many magic methods that are supposed to give the user more convenient method names. While `findBy` is a valid and existing method name that can be used, there are magic methods based on a findBy-prefix that can be combined with an existing property of the entity the called repository belongs to. If the entity has a `name` property, you can use `findByName('my-name')`.

While a magic `countBy`-prefix method name exists to use names like `countByName`, a physical `countBy` method doesn't exist and an exception will happen when you won't use a property name at the end of it.

`$repo->count(['visible' => true])` should be working fine in your case and I don't think we should add a physical `countBy` just for consistency to `findBy`.

zorn-v:
To find find that there is no `countBy` method, you need to dig documentation or source.
Error sounds like `Entity 'App\Entity\Entity' has no field ''. You can therefore not call 'count' on the entities' repository` which just confuse. I know about `countByVisible` magic and does not understand why `findByVisible(true)` and `findBy(['visible' => true])` is correct, but not with `count`

lcobucci:
@zorn-v so perhaps we could look into improving the error message? Adding that method is quite to me based on repository semantics.

beberlei:
An IDE would hint that the method is count not countBy. This inconsistency is not too big a problem enough to consider adding this in my opinion.

PR #7974, opened 2020-01-08
[GH-7505] Bug in SimpleObjectHydrator when using inheritance with same field

### LINKED ISSUE #7505: Hydratation error when two classes with the same parent class both define the same property with different types.

beberlei:
Fixes #7505 

Supersedes https://github.com/doctrine/orm/pull/7506

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -78,8 +78,9 @@ class SimpleObjectHydrator extends AbstractHydrator
      */
     protected function hydrateRowData(array $sqlResult, array &$result)
     {
-        $entityName = $this->class->name;
-        $data       = [];
+        $entityName       = $this->class->name;
+        $data             = [];
+        $discrColumnValue = null;
 
         // We need to find the correct entity class name if we have inheritance in resultset
         if ($this->class->inheritanceType !== ClassMetadata::INHERITANCE_TYPE_NONE) {
@@ -104,7 +105,8 @@ class SimpleObjectHydrator extends AbstractHydrator
                 throw HydrationException::invalidDiscriminatorValue($sqlResult[$discrColumnName], array_keys($discrMap));
             }
 
-            $entityName = $discrMap[$sqlResult[$discrColumnName]];
+            $entityName       = $discrMap[$sqlResult[$discrColumnName]];
+            $discrColumnValue = $sqlResult[$discrColumnName];
 
             unset($sqlResult[$discrColumnName]);
         }
@@ -134,6 +136,11 @@ class SimpleObjectHydrator extends AbstractHydrator
 
             // Prevent overwrite in case of inherit classes using same property name (See AbstractHydrator)
             if ( ! isset($data[$fieldName]) || ! $valueIsNull) {
+                // If we have inheritance in resultset, make sure the field belongs to the correct class
+                if (isset($cacheKeyInfo['discriminatorValues']) && ! in_array($discrColumnValue, $cacheKeyInfo['discriminatorValues'], true)) {
+                    continue;
+                }
+
                 $data[$fieldName] = $value;
             }
         }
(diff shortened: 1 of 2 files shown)


SenseException (dismissed):

lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
```diff
@@ -134,6 +136,11 @@ protected function hydrateRowData(array $sqlResult, array &$result)
 
             // Prevent overwrite in case of inherit classes using same property name (See AbstractHydrator)
             if ( ! isset($data[$fieldName]) || ! $valueIsNull) {
+                // If we have inheritance in resultset, make sure the field belongs to the correct class
+                if (isset($cacheKeyInfo['discriminatorValues']) && ! in_array($discrColumnValue, $cacheKeyInfo['discriminatorValues'], true)) {
```
lcobucci: Would it be possible to combine this with the previous if statement? It would simplify things, wouldn't it?

lcobucci (changes requested):
Fix LGTM, just a few nitpicks... 

The original issue also mention ensuring that the schema validation detects an issue when the column name is the same, do we have tests for this scenario?

About joined tables, does it behaves the same way? Could we also cover it with tests?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7505Test.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->setUpEntitySchema([
+            GH7505AbstractResponse::class,
+            GH7505ArrayResponse::class,
+            GH7505TextResponse::class,
+        ]);
+    }
+
+    public function testSimpleArrayTypeHydratedCorrectly() : void
+    {
+        $arrayResponse = new GH7505ArrayResponse();
```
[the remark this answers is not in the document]
beberlei: Not sure it matters, the entity is needed for the test-case

lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
beberlei: No its not possible.

beberlei:
@lcobucci Joined tables behave the same, this code is unaware of type of inheritance. The schema validator should not make it an issue if its the same column name, because that is supported behavior as per the code.

Issue #7972, opened 2020-01-08
ORM should load platform lazily on some places


beberlei:
To avoid triggering the database platform detection, we should use the platform more lazily instead of in the constructor in a few places.

PR #7975, opened 2020-01-09
Make console commands lazy

andrew-demb:
Link to doc: https://symfony.com/doc/current/console/commands_as_services.html#lazy-loading

Will be helpful for doctrine/orm#7972

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/CollectionRegionCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/CollectionRegionCommand.php
@@ -24,13 +24,15 @@ use function sprintf;
  */
 class CollectionRegionCommand extends AbstractEntityManagerCommand
 {
+    /** @var string|null */
+    protected static $defaultName = 'orm:clear-cache:region:collection';
+
     /**
      * {@inheritdoc}
      */
     protected function configure()
     {
-        $this->setName('orm:clear-cache:region:collection')
-             ->setDescription('Clear a second-level cache collection region')
+        $this->setDescription('Clear a second-level cache collection region')
              ->addArgument('owner-class', InputArgument::OPTIONAL, 'The owner entity name.')
              ->addArgument('association', InputArgument::OPTIONAL, 'The association collection name.')
              ->addArgument('owner-id', InputArgument::OPTIONAL, 'The owner identifier.')
--- a/lib/Doctrine/ORM/Tools/Console/Command/ConvertMappingCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ConvertMappingCommand.php
@@ -35,13 +35,15 @@ use function strtolower;
  */
 class ConvertMappingCommand extends AbstractEntityManagerCommand
 {
+    /** @var string|null */
+    protected static $defaultName = 'orm:convert-mapping';
+
     /**
      * {@inheritdoc}
      */
     protected function configure()
     {
-        $this->setName('orm:convert-mapping')
-             ->setAliases(['orm:convert:mapping'])
+        $this->setAliases(['orm:convert:mapping'])
              ->setDescription('Convert mapping information between supported formats')
              ->addArgument('to-type', InputArgument::REQUIRED, 'The mapping type to be converted.')
              ->addArgument('dest-path', InputArgument::REQUIRED, 'The path to generate your entities classes.')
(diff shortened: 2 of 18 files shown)


lib/Doctrine/ORM/Tools/Console/Command/ConvertMappingCommand.php
```diff
@@ -43,6 +43,9 @@
  */
 class ConvertMappingCommand extends Command
 {
+    /** @var string */
+    protected static $defaultName = 'orm:convert-mapping';
```
andrew-demb: As a minimal version of `symfony/console` dependency in `composer.json` is `3.0` and laziness supports introduced in `3.4`, I duplicate the same command name in `$defaultName` and `->setName()`

theofidry: what about re-using `self::$defaultName` in `setName()` then and mention that `setName()` can be removed when dependencies are bumped to 3.4 minimum in a TODO?

andrew-demb: Applied. Thank you

lcobucci: We can bump the dependency too since this is a minor release, makes things much simpler IMHO

andrew-demb: Applied. Thank you @lcobucci.

SenseException (dismissed):

andrew-demb:
Added 'laziness' to all commands within `doctrine/orm`, retargeted to 2.10.x, and rebased with squash.

There is one command with a different way to control command name - https://github.com/doctrine/orm/blob/f6e1dd44f0fac0c59c540723bd24303df49f7b93/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/UpdateCommand.php#L32
I'm not sure how to preserve BC (as It was added for BC 11 years ago - https://github.com/doctrine/orm/commit/64687409155f5e547b2611db1eed4948fba847f9)

Also, `symfony/console:5.3+` added a way to configure the default description for the command. Should I add this feature usage?

lib/Doctrine/ORM/Tools/Console/Command/ClearCache/CollectionRegionCommand.php
```diff
@@ -24,13 +24,15 @@
  */
 class CollectionRegionCommand extends AbstractEntityManagerCommand
 {
+    /** @var string|null */
```
andrew-demb, line 27: Unfortunately I cannot put here `@var string`, because psalm fails

```

Error: lib/Doctrine/ORM/Tools/Console/Command/ClearCache/CollectionRegionCommand.php:28:22: NonInvariantDocblockPropertyType: Property Doctrine\ORM\Tools\Console\Command\ClearCache\CollectionRegionCommand::$defaultName has type string, not invariant with Symfony\Component\Console\Command\Command::$defaultName of type null|string (see https://psalm.dev/267)
```

derrabus:
I'm not sure if we need this. In a Symfony full-stack application, DoctrineBundle takes care of making the commands lazy. And when using the commands standalone, we need to provide a readily configured entity manager anyway, don't we? We wouldn't benefit from the laziness then.

andrew-demb:
@derrabus I have no answer :(

I agree that commands from doctrine/orm are not ready to use as is and require additional configuration (like inheritance in Symfony bundle, or instantiation commands like in Laminas module with injecting EM helper).

Also, that laziness for commands with `$defaultName` property works only in case of using `\Symfony\Component\Console\DependencyInjection\AddConsoleCommandPass`, which is not always can be used (non-Symfony app).

So the only reason I can suggest is "why not add if we can?")

PS: for doctrine bundle, I can send PR with adding `$defaultName` property to avoid duplicating the same command name in XML configs and PHP command code.

derrabus:
> So the only reason I can suggest is "why not add if we can?")

Symfony is moving away from that property and towards attributes for configuring the commands. So we would use a mechanism that is legacy already and without any real benefit. Unless we have a good reason to make this change, I'd rather not.

> PS: for doctrine bundle, I can send PR with adding `$defaultName` property to avoid duplicating the same command name in XML configs and PHP command code.

Doesn't DoctrineBundle rename the commands anyway?

andrew-demb:
> Symfony is moving away from that property and towards attributes for configuring the commands.

Didn't know this before...

> Doesn't DoctrineBundle rename the commands anyway?

Yes. They are.
I mean sync a way for configuring command name - using only static $defaultName, without duplication same name in DI and PHP
https://github.com/doctrine/DoctrineBundle/blob/2.5.2/Resources/config/dbal.xml#L86
https://github.com/doctrine/DoctrineBundle/blob/2.5.2/Command/CreateDatabaseDoctrineCommand.php#L30

___
So, if this way will be considered as a  deprecated by Symfony soon - I'm going to close this PR

PR #7978, opened 2020-01-12
Stable PHP 7.4 in Travis

alexeyshockov:


alexeyshockov:
OK, so there were a lot `Trying to access array offset on value of type null` errors that I fixed.

But the one related to `ReflectionProperty` serialization looks not so easy to fix. The issue is with serializing [`AssociationMetadata::$reflection`](https://github.com/doctrine/orm/blob/ff65cae17083266c43f091fff03289197db9e4fb/lib/Doctrine/ORM/Mapping/AssociationMetadata.php#L19) as far as I can see, but it cannot be excluded in as simple as in [`ClassMetadata`](https://github.com/doctrine/orm/blob/ff65cae17083266c43f091fff03289197db9e4fb/lib/Doctrine/ORM/Mapping/ClassMetadata.php#L297-L309). I tried to play around, but no luck so far.

IMO we can merge it and fix this test in a separate PR.

alexeyshockov:
So the serialization error happens because of [a BC break in 7.4](https://github.com/php/php-src/blob/PHP-7.4/UPGRADING#L112-L115):

> Reflection objects will now generate an exception if an attempt is made
    to serialize them. Serialization for reflection objects was never
    supported and resulted in corrupted reflection objects. It has been
    explicitly prohibited now.

greg0ire:
Thanks for your contribution. I think you should target a lower branch (2.8 ?)

alexeyshockov:
@greg0ire, thanks. This fix is actual for both `master` and `2.8.x` branches IMO.

Is it better to change target of this PR to `2.8.x` (and merge `2.8.x` then to `master`), or do a separate PR to `2.8.x` (if you don't plan to merge it back to master)?

greg0ire:
We are supposed to use up-merge now. It's not the case for the ORM yet but I'd contribute this to 2.8.x first, since this is where we're headed. Once it is merged, and if it does not look like we do upmerges then, you can try a PR on master.

alexeyshockov:
OK, thanks, then I will change the target to `2.8.x`

alexeyshockov:
OK, `2.8.x` doesn't have this bug with reflection, that I found in `master`, so currently the PR is simple.

@greg0ire, shall I create another PR to `master`, because it seems to have its own (unique) issues on PHP 7.4 (because of the big changes in mapping metadata)?

greg0ire:
You probably should indeed, but please focus on 2.8.x first, it's more important IMO :)

alexeyshockov:
For `2.8.x` it's just a simple change of the version in Travis, ready for review and for merging

greg0ire (approved):

SenseException (approved):

.travis.yml
```diff
@@ -124,7 +124,7 @@ jobs:
     - stage: Code Quality
       if: NOT type = pull_request
       env: DB=none CODING_STANDARDS
-      php: 7.4snapshot
+      php: 7.4
```
greg0ire: I think you forgot this 3rd occurence.

greg0ire:
We're working on fixing the build, it is currently broken on 2.8.x

greg0ire:
Restarting the build

greg0ire:
Thanks @alexeyshockov !

PR #7979, opened 2020-01-12
Private is not necessary

oojacoboo:
There is no reason for the EntityManager to not be accessible within filters.  It causes extension writers to do code gymnastics to get the EntityManager.

Example:
https://github.com/Atlantic18/DoctrineExtensions/blob/master/src/SoftDeleteable/Filter/SoftDeleteableFilter.php

### DIFF
--- a/lib/Doctrine/ORM/Query/Filter/SQLFilter.php
+++ b/lib/Doctrine/ORM/Query/Filter/SQLFilter.php
@@ -112,6 +112,11 @@ abstract class SQLFilter
         return serialize($this->parameters);
     }
 
+    final protected function getEntityManager() : EntityManagerInterface
+    {
+        return $this->em;
+    }
+
     /**
      * Returns the database connection used by the entity manager
      *


lib/Doctrine/ORM/Query/Filter/SQLFilter.php
```diff
@@ -24,7 +24,7 @@ abstract class SQLFilter
      *
      * @var EntityManagerInterface
      */
-    private $em;
+    protected $em;
```
greg0ire: There is no reason for the entity manager to be writable. I would be more ok with a `final protected` accessor.

greg0ire: :thinking: it's an object, so I guess it would be writable anyway, unless we clone it in the accessor, which is probably not desirable.

greg0ire:
`master` is probably not the branch you want to target, it represents version 3

oojacoboo:
Version 3 is good enough.  There are already hacks in place for most extensions anyway via reflection.

beberlei:
We initially designed this class to have very little possibilities, because you can easily make mistakes depending on outside state. I can see how reusable filters need more information than what it provides at the moment.

In this case you need access to the database platform and custom mapping schema. The first we could easily provide to a custom protected getter, the second might require us to think about extension metadata, maybe as a first class citizen. 

But maybe we should provide access to EntityManager and pass responsibility about performance to users.

oojacoboo:
I can add a protected getter if that'll suffice.  I'm writing an extension now where I need access to the EntityManager, but just using a reflection hack.  I don't need access to the mapping schema, but I can see how that'd be commonly needed.

You're mentioning performance, but having to do hacks like the ones in this extension are degrading performance as well.

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
greg0ire: But I guess at least you will not be able to replace `$em` with another instance, and since that object is already shared with a lot of things, and you access it in your filter, it's not a big deal to give access to it. Please use a `final protected` accessor.

oojacoboo:
@greg0ire and @beberlei I've committed a final protected accessor now.

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
```diff
@@ -112,6 +112,16 @@ abstract class SQLFilter
         return serialize($this->parameters);
     }
 
+    /**
+     * Returns the entity manager
+     *
+     * @return EntityManagerInterface
+     */
+    final protected function getEntityManager()
```
greg0ire:
```suggestion
    final protected function getEntityManager() : EntityManagerInterface
```

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
```diff
@@ -112,6 +112,16 @@ abstract class SQLFilter
         return serialize($this->parameters);
     }
 
+    /**
+     * Returns the entity manager
+     *
+     * @return EntityManagerInterface
+     */
```
greg0ire: Please remove the whole comment.

oojacoboo:
@greg0ire done.  Didn't know it was being typed - the rest of the methods weren't.

greg0ire (dismissed):

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
```diff
@@ -112,6 +112,11 @@ abstract class SQLFilter
         return serialize($this->parameters);
     }
 
+    final protected function getEntityManager(): EntityManagerInterface
```
SenseException:
```suggestion
    final protected function getEntityManager() : EntityManagerInterface
```

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
```diff
@@ -112,6 +112,11 @@ abstract class SQLFilter
...
+    {
+        return $this->em;
+    }
+
```
SenseException:
```suggestion

```

SenseException (changes requested):
There are just a few coding style issues that need to be fixed.

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
oojacoboo: I didn't look to see what the standard is for Doctrine here, but having an additional empty line at the end of a file is recommended for portability reasons.  Unix/POSIX systems generally assume a trailing empty line as files are concatenated together.  Without the final line, the last line will be merged with the first line of the next file, sometimes causing syntax errors.

greg0ire: This is not the end of the file, is it? There should definitely be a blank line between methods though.

oojacoboo: Sorry, I thought it was.  I have no idea what this is then.  This snippet is one line.  I see no way to view the full file with requested changes here.  I'm not very familiar with this interface and am finding it to be incredibly frustrating to work with at this depth.

Feel free to commit it.  I thought I added a line between the method.  At least when I click to "view file", that's what I see.

greg0ire: I think there is a line between the methods, so all is well for me. I was saying that for @SenseException , not for you.

SenseException: It's not about the line but empty spaces. "Whitespace found at end of line" is more that the whole line consists of empty spaces.

greg0ire: I think they must have been removed cause I don't see the spaces in the diff when trying to highlight them with my cursor.

SenseException: I see that it is still mentioned by phpcs in the current build result: https://travis-ci.org/doctrine/orm/jobs/641467131#L562. Does the IDE not show them?

greg0ire: vim does show them, not sure why Github doesn't. I just force pushed.

SenseException (approved):

oojacoboo:
Is there any reason why this was never merged?

oojacoboo:
@SenseException @greg0ire any reason why we couldn't merge this?

greg0ire:
Yes: it never targeted the correct branch.

PR #7994, opened 2020-01-20
Update setParameters to conform to DocBlock for PHP 7.4

pavlakis:
The `setParameters` method accepts `* @param ArrayCollection|mixed[] $parameters` and as such the `array` is no longer a valid type as of `2.7.0`

### DIFF
--- a/docs/en/reference/query-builder.rst
+++ b/docs/en/reference/query-builder.rst
@@ -265,7 +265,7 @@ following syntax:
     // $qb instanceof QueryBuilder
 
     // Query here...
-    $qb->setParameters(array(1 => 'value for ?1', 2 => 'value for ?2'));
+    $qb->setParameters(new ArrayCollection(array(1 => 'value for ?1', 2 => 'value for ?2')));
 
 Getting already bound parameters is easy - simply use the above
 mentioned syntax with "getParameter()" or "getParameters()":


greg0ire:
I'm confused… `array` is compatible with `mixed[]` and therefore compatible with `ArrayCollection|mixed[]`

pavlakis:
@greg0ire 
That was my take on it too. So, I raised it as an [issue with PHPStan](https://github.com/phpstan/phpstan-doctrine/issues/105) where I found this issue from, and apparently `array` is not compatible with `mixed[]`.

greg0ire (commented):
In theory, I'm ok with this PR, since we are going to drop arrays here if I believe the comment found at https://github.com/doctrine/orm/blob/ff65cae17083266c43f091fff03289197db9e4fb/lib/Doctrine/ORM/AbstractQuery.php#L323
In practice, this has been decided 8 years ago (!) in https://github.com/doctrine/orm/commit/161ae31a7e044a7f0dda357ed271215f2317e9a5, and the BC layer is still present on master, so it might be a bit early to recommend that, I don't know. I notice that it's already present in some docblocks: https://github.com/doctrine/orm/blob/ff65cae17083266c43f091fff03289197db9e4fb/lib/Doctrine/ORM/QueryBuilder.php#L549-L556

Also, I think there are more occurences that can and should be changed: https://github.com/doctrine/orm/search?q=setParameters&unscoped_q=setParameters

greg0ire:
I think we would accept a PR changing the docblock so that it's compatible with phpstan

Screenshot_2020-01-20 Slack static-analysis Symfony Devs

ondrejmirtes:
I just realized I could add a phpDoc stub file to phpstan-doctrine that changes the signature so that PHPStan understands how it’s meant 😊

pavlakis:
@greg0ire I have created PR https://github.com/doctrine/orm/pull/7995 with the DocBlock update.

PR #7995, opened 2020-01-20
Change DocBlock to be compatible with PHPStan

pavlakis:
The current ArrayCollection|mixed[] DocBlock causes PHPStan to fail the current behaviour of
passing an array on setParameters()

Related issue: https://github.com/doctrine/orm/pull/7994

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -336,7 +336,7 @@ abstract class AbstractQuery
     /**
      * Sets a collection of query parameters.
      *
-     * @param ArrayCollection|mixed[] $parameters
+     * @param ArrayCollection|array<mixed> $parameters
      *
      * @return static This query instance.
      */


greg0ire:
Should https://github.com/doctrine/orm/blob/fdad48278b6c0d0d7075de584e6982969bf430c6/lib/Doctrine/ORM/QueryBuilder.php#L562 be "fixed" too?

Using quotes around fixed, and I'm curious to see what other maintainers will say, because the syntax we currently use looks like it should be supported.

pavlakis:
It might be obvious, but I just realised that passing `new ArrayCollection([$params])` although it will allow PHPStan to pass, it won't support the same functionality as the elements in the collection won't be `new Parameter($key, $value)`.

So without making the above change, the options at present are:
* pass the array but ignore the PHPStan error, or
* add each parameter individually using `setParameter`

greg0ire:
> It might be obvious, but I just realised that passing new ArrayCollection([$params]) although it will allow PHPStan to pass, it won't support the same functionality as the elements in the collection won't be new Parameter($key, $value).

Link for the lazy:

https://github.com/doctrine/orm/blob/e77091399a4475630a833fc33bff0d54c194511a/lib/Doctrine/ORM/QueryBuilder.php#L563-L581

PR #8002, opened 2020-01-25
Remove ORM 2.0 version constraints from docs

SenseException:
master is going to be ORM 3.0 and the documentation is still referencing information about version 2 features and when they were introduced. This PR is removing them because they are irrelevant for 3.0.

The docs also still name the ORM project "Doctrine 2" which isn't accurate since the renaming of the repository/project. This PR rephrases those occurrences with "Doctrine ORM" or just "ORM".

### DIFF
--- a/docs/en/cookbook/working-with-datetime.rst
+++ b/docs/en/cookbook/working-with-datetime.rst
@@ -3,7 +3,7 @@ Working with DateTime Instances
 
 There are many nitty gritty details when working with PHPs DateTime instances. You have to know their inner
 workings pretty well not to make mistakes with date handling. This cookbook entry holds several
-interesting pieces of information on how to work with PHP DateTime instances in Doctrine 2.
+interesting pieces of information on how to work with PHP DateTime instances in ORM.
 
 DateTime changes are detected by Reference
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -58,7 +58,7 @@ Handling different Timezones with the DateTime Type
 
 If you first come across the requirement to save different timezones you may be still optimistic about how
 to manage this mess,
-however let me crush your expectations fast. There is not a single database out there (supported by Doctrine 2)
+however let me crush your expectations fast. There is not a single database out there (supported by Doctrine ORM)
 that supports timezones correctly. Correctly here means that you can cover all the use-cases that
 can come up with timezones. If you don't believe me you should read up on `Storing DateTime
 in Databases <http://derickrethans.nl/storing-date-time-in-database.html>`_.
@@ -67,7 +67,7 @@ The problem is simple. Not a single database vendor saves the timezone, only the
 However with frequent daylight saving and political timezone changes you can have a UTC offset that moves
 in different offset directions depending on the real location.
 
-The solution for this dilemma is simple. Don't use timezones with DateTime and Doctrine 2. However there is a workaround
+The solution for this dilemma is simple. Don't use timezones with DateTime and Doctrine ORM. However there is a workaround
 that even allows correct date-time handling with timezones:
 
 1. Always convert any DateTime instance to UTC.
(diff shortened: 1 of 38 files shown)


docs/en/cookbook/working-with-datetime.rst
```diff
@@ -3,7 +3,7 @@ Working with DateTime Instances
 
 There are many nitty gritty details when working with PHPs DateTime instances. You have to know their inner
 workings pretty well not to make mistakes with date handling. This cookbook entry holds several
-interesting pieces of information on how to work with PHP DateTime instances in Doctrine 2.
+interesting pieces of information on how to work with PHP DateTime instances in ORM.
```
greg0ire:
```suggestion
interesting pieces of information on how to work with PHP DateTime instances in Doctrine ORM.
```

docs/en/reference/advanced-configuration.rst
```diff
@@ -317,7 +317,7 @@ Proxy Objects
 
 A proxy object is an object that is put in place or used instead of
 the "real" object. A proxy object can add behavior to the object
-being proxied without that object being aware of it. In Doctrine 2,
+being proxied without that object being aware of it. In ORM,
```
greg0ire:
```suggestion
being proxied without that object being aware of it. In Doctrine ORM,
```

greg0ire (commented):
How are these changes going to land in lower branches?

docs/en/cookbook/working-with-datetime.rst
SenseException: This was intentional. I wanted to reduce "Doctrine ORM" appearing everywhere to prevent it becoming too redundant. I actually thinking about reducing its occurrence in annotations-reference.rst where it is used multiple times.

docs/en/reference/advanced-configuration.rst
SenseException: This was intentional. I wanted to reduce "Doctrine" appearing everywhere to prevent it becoming too redundant. I actually thinking about reducing its occurrence in annotations-reference.rst where it is used multiple times.

SenseException:
> How are these changes going to land in lower branches?

They aren't going to be in the lower branches. This PR is set for ORM 3.0 and replacing "Doctrine 2" is just a decision between this or replacing it with "Doctrine 3". Because all the differences inside the Doctrine 2 versions aren't relevant for 3 anymore, I decided to remove the number entirely.

greg0ire (approved):
Looks good, but I wouldn't be against doing this on stable branches, because it makes sense here too IMO, and it will make merging up less painful.

SenseException:
I'd like to hear more opinions about that.

ostrolucky:
Another vote for targetting lower branch

SenseException:
@ostrolucky Do you expect any differences for the current changed code when it would be on a 2.x branch? Some lines that don't need to be removed?

ostrolucky:
I don't. Merging it lower will just help slowing down a big diverge betwen 2.x and master. But if you are not comfortable rebasing it on 2.8.x, feel free to merge, I'm not extremely opposed to merge this type of change just for master

SenseException:
I just don't want to have false information in the lower branches if there are still one in the current changes. If these are still working for 2.7 or 2.8, I would rebase it to the lower branches. I assume you mentioned 2.8.x because it's the preferred branch for this?

ostrolucky:
> I assume you mentioned 2.8.x because it's the preferred branch for this?

It's just branch I checked. I don't know what's the policy for target branches in ORM repository. I would like to see it in 2.7.x, but I would personally target it for 2.8.x as it's less controversional than trying to pass it as bug fix to 2.7x.

greg0ire:
> I would personally target it for 2.8.x as it's less controversional than trying to pass it as bug fix to 2.7x.

We shouldn't merge in 2.7 if there is a risk of making the code less stable. This does not apply to tests, CI improvements, or documentation changes, IMO there is no issue in targetting 2.7.x

SenseException:
I've rebased it onto 2.8.x.

ostrolucky (approved):

greg0ire:
Thanks @SenseException !

PR #8006, opened 2020-01-30
Make Embeddable not transient

malarzm:
I believe this was an oversight during original implementation but let's see what tests will tell us. Also this made PHPStan ignore Embeddables (https://github.com/phpstan/phpstan-doctrine/pull/108#pullrequestreview-349722039)

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
@@ -44,6 +44,7 @@ class AnnotationDriver extends AbstractAnnotationDriver
     protected $entityAnnotationClasses = [
         Mapping\Entity::class => 1,
         Mapping\MappedSuperclass::class => 2,
+        Mapping\Embeddable::class => 3,
     ];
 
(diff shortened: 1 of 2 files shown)


lcobucci:
@malarzm @beberlei I believe that embedded objects were transient by design (at least it makes sense to me) and this should be reverted.

An embedded object is merely a value object and we shouldn't worry about its life cycle or identity, because that's handled by the entity/aggregate itself.

goetas:
hi. I beilive that this change makes sense, but it broke JMS serializer (see https://github.com/schmittjoh/JMSSerializerBundle/issues/790). How can we now distinguish if a class can be managed by the entity manager?

malarzm:
@goetas I believe the question should be "How can we now distinguish if a class should be loaded from db?" and the answer would be then "Check if something is an embeddable through [$metadata->isEmbeddedClass](https://github.com/doctrine/orm/blob/2.7/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L270)" :)

goetas:
ok :)


thanks!

Issue #7969, opened 2020-01-03
Child entity not retrieved from array cache


peterkeatingie:
### Bug Report

BC Break    | no
Version     | 2.7.0

#### Summary

Entity which is extended from a base class is not retrieved from array cache when using the second level cache.
ZF3 application.

#### Current behavior

Base class is abstract, with extending class as a child, using single table inheritance and second level cache.
When retrieving the child class using the findBy repository method, calls subsequent to the first one are not found in the cache. The entity is fetched from the database each time.

It looks to me like the entity is put into the cache with a key which is based on its child name but retrieval is attempted using the base class name.

#### How to reproduce

Entity definitions:

Base class:

```
<?xml version="1.0" encoding="utf-8"?>
<doctrine-mapping xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
    <entity name="Gts\Base\Entity\Base" table="base" inheritance-type="SINGLE_TABLE">
        <id name="id" type="integer" column="id">
            <generator strategy="IDENTITY"/>
        </id>
        <field name="someField"/>
    </entity>
</doctrine-mapping>
```

Child class. Here the cache usage is set:

```
<?xml version="1.0" encoding="utf-8"?>
<doctrine-mapping xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
    <entity name="Gts\Base\Entity\BaseChild" extends="base">
        <cache usage="NONSTRICT_READ_WRITE" />
        <field name="someNewField"/>
    </entity>
</doctrine-mapping>

```

Doctrine configuration:

```
'second_level_cache' => [
                    'enabled' => true,
                ],
```

Calling fetch several times leads to a new db call each time (this is BaseChild repo):

```
// Doctrine\ORM\EntityRepository
        $this->repository->findBy([
            'someField' => 'Some Field Value',
        ]);

        $this->repository->findBy([
            'someField' => 'Some Field Value',
        ]);
```

Query log:

```
[...]

PR #8009, opened 2020-01-31
Put into cache using root entity name

peterkeatingie:
The cache key for reading is created based on the entity root name but the cache was written with a key based on the found entity (the child in this case) resulting in cache miss.

Using the same root name when putting fixes the issue. I do believe that any extending instances should have unique identifiers as they will have to share an identifier from the base class. This will mean the key won't be ambiguous - if my thinking is correct.

https://github.com/doctrine/orm/issues/7969

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -279,12 +279,12 @@ class DefaultQueryCache implements QueryCache
 
         $region = $persister->getCacheRegion();
 
-        $cm             = $this->em->getClassMetadata($entityName);
-        $rootEntityName = $cm->getMetadataValue('rootEntityName');
+        $cm = $this->em->getClassMetadata($entityName);
+        assert($cm instanceof ClassMetadata);
 
         foreach ($result as $index => $entity) {
             $identifier = $this->uow->getEntityIdentifier($entity);
-            $entityKey  = new EntityCacheKey($rootEntityName, $identifier);
+            $entityKey  = new EntityCacheKey($cm->rootEntityName, $identifier);
 
             if (($key->cacheMode & Cache::MODE_REFRESH) || ! $region->contains($entityKey)) {
                 // Cancel put result if entity put fail
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -279,11 +279,12 @@ class DefaultQueryCache implements QueryCache
 
         $region = $persister->getCacheRegion();
 
-        $cm = $this->em->getClassMetadata($entityName);
+        $cm             = $this->em->getClassMetadata($entityName);
+        $rootEntityName = $cm->getMetadataValue('rootEntityName');
 
         foreach ($result as $index => $entity) {
-            $identifier                     = $this->uow->getEntityIdentifier($entity);
-            $entityKey  = new EntityCacheKey($cm->getMetadataValue('rootEntityName'), $identifier);
+            $identifier = $this->uow->getEntityIdentifier($entity);
+            $entityKey  = new EntityCacheKey($rootEntityName, $identifier);
 
             if (($key->cacheMode & Cache::MODE_REFRESH) || ! $region->contains($entityKey)) {
                 // Cancel put result if entity put fail
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -279,9 +279,11 @@ class DefaultQueryCache implements QueryCache
 
         $region = $persister->getCacheRegion();
 
+        $cm = $this->em->getClassMetadata($entityName);
+
         foreach ($result as $index => $entity) {
             $identifier                     = $this->uow->getEntityIdentifier($entity);
-            $entityKey                      = new EntityCacheKey($entityName, $identifier);
+            $entityKey                      = new EntityCacheKey($cm->rootEntityName, $identifier);
 
             if (($key->cacheMode & Cache::MODE_REFRESH) || ! $region->contains($entityKey)) {
                 // Cancel put result if entity put fail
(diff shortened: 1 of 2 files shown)


SenseException (changes requested):

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -279,9 +279,11 @@ public function put(QueryCacheKey $key, ResultSetMapping $rsm, $result, array $h
 
         $region = $persister->getCacheRegion();
 
+        $cm = $this->em->getClassMetadata($entityName);
+
         foreach ($result as $index => $entity) {
             $identifier                     = $this->uow->getEntityIdentifier($entity);
-            $entityKey                      = new EntityCacheKey($entityName, $identifier);
+            $entityKey  = new EntityCacheKey($cm->getMetadataValue('rootEntityName'), $identifier);
```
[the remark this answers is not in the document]
peterkeatingie: At first I called the property of the metadata directly and then changed to the method call in situ, hence remaining inside the loop. I see your point though, no need to call each time. Updated now.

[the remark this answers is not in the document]
peterkeatingie: Fixed the alignment too.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC7969Test.php
```diff
@@ -0,0 +1,48 @@
...
+namespace tests\Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Cache\Region\DefaultMultiGetRegion;
+use Doctrine\Tests\Models\Cache\Attraction;
+use Doctrine\Tests\Models\Cache\Bar;
+use Doctrine\Tests\ORM\Functional\SecondLevelCacheAbstractTest;
+
+class DDC7969Test extends SecondLevelCacheAbstractTest
+{
+    public function testChildEntityRetrievedFromCache()
```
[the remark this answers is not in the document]
peterkeatingie: Updated now.

peterkeatingie:
Is there anything I can do to trigger the travis build again without updating the PR? I don't think the failure is related to anything I added.

SenseException (approved):
I'd like to have more reviewers from the Doctrine team accepting this PR in case I overlooked a detail in the functionality.

peterkeatingie:
> I'd like to have more reviewers from the Doctrine team accepting this PR in case I overlooked a detail in the functionality.

No problem, do I need to add more reviewers?

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -279,9 +279,12 @@ public function put(QueryCacheKey $key, ResultSetMapping $rsm, $result, array $h
...
         $region = $persister->getCacheRegion();
 
+        $cm             = $this->em->getClassMetadata($entityName);
+        $rootEntityName = $cm->getMetadataValue('rootEntityName');
+
         foreach ($result as $index => $entity) {
-            $identifier                     = $this->uow->getEntityIdentifier($entity);
-            $entityKey                      = new EntityCacheKey($entityName, $identifier);
+            $identifier = $this->uow->getEntityIdentifier($entity);
+            $entityKey  = new EntityCacheKey($rootEntityName, $identifier);
```
beberlei:
```suggestion
            $entityKey  = new EntityCacheKey($cm->rootEntityName, $identifier);
```

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -279,9 +279,12 @@ public function put(QueryCacheKey $key, ResultSetMapping $rsm, $result, array $h
 
         $region = $persister->getCacheRegion();
 
+        $cm             = $this->em->getClassMetadata($entityName);
+        $rootEntityName = $cm->getMetadataValue('rootEntityName');
```
beberlei:
```suggestion
```

You should directly use `$cm->rootEntityName` in line 287, `getMetadataValue` has a different use-case.

beberlei (changes requested):

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
peterkeatingie: Hi @beberlei , I originally did that but then the codesniffer job failed due to accessing a property when presented only with an interface - should we suppress or ignore that?

beberlei: @peterkeatingie yes, you can add `assert($cm instanceof ClassMetadata);` in line 283, this code gets compiled away in production when assertions are off.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC7969Test.php
```diff
@@ -0,0 +1,48 @@
...
+        if ($region instanceof DefaultMultiGetRegion) {
+            $region->getCache()->flushAll();
+        }
+
+        /** @var Bar $bar */
+        $bar = $this->attractions[0];
+
+        $repository = $this->_em->getRepository(Bar::class);
+
+        $this->assertFalse($this->cache->containsEntity(Bar::class, $bar->getId()));
```
beberlei: Can you add a check for the same using the root class name as well?

beberlei (approved):

peterkeatingie:
Thanks @beberlei  @SenseException. Is there anything I can do expedite this PR or just wait now?

beberlei:
@peterkeatingie i am waiting from someone with more SLC experience to ack. One question I have though, you made this PR towards 2.8.x branch, but I would consider this a bugfix, so this would be good for the 2.7 branch. What was your approach here?

peterkeatingie:
> @peterkeatingie i am waiting from someone with more SLC experience to ack. One question I have though, you made this PR towards 2.8.x branch, but I would consider this a bugfix, so this would be good for the 2.7 branch. What was your approach here?

No problem. Actually I might have this wrong - I thought our project had 2.8 using composer show but maybe got that wrong or maybe that's for dev, I can't remember. Anyway checked the lock file now and it's 2.7 alright. Also was following this advice: https://github.com/doctrine/orm#which-branch-should-i-choose.

So if that should be against 2.7 then I guess I need to branch that and make a new PR?!

peterkeatingie:
Closing this in favour of https://github.com/doctrine/orm/pull/8023

PR #8013, opened 2020-02-05
[FIX] Add custom exception for parameters parsing

KartaviK:
Simple reproduce example

```php
$qb->
    ...
    ->setParameters(new ArrayCollection([
        'key' => 'value',
    ])
;
```

We will catch not semantic exception `Error: Call to a member function getName() on string`.
This exception will correctly give information about valid data

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -218,7 +218,10 @@ final class Query extends AbstractQuery
         $types = [];
 
         foreach ($this->parameters as $parameter) {
-            /** @var Query\Parameter $parameter */
+            if (!$parameter instanceof Parameter) {
+                throw new QueryException('Set parameter must be instance of \Doctrine\ORM\Query\Parameter');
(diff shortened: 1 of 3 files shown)


SenseException:
Thanks for your pull request. Instead of throwing an exception I would prefer that the parameters would be in the correct format when they are set with `setParameters()`. This way we never would have an invalid parameters state that needs to be checked.

KartaviK:
> Thanks for your pull request. Instead of throwing an exception I would prefer that the parameters would be in the correct format when they are set with `setParameters()`. This way we never would have an invalid parameters state that needs to be checked.

So how in future build will look signature for `setParameters` to avoid this case?

SenseException:
I would expect similar to #7885 an `iterable` and internally it would still create an `ArrayCollection` with `Parameter`. It shouldn't pass blindly an `ArrayCollection` without checking for `Parameter`, so I expect the `is_array()` check will be obsolete.

Issue #8014, opened 2020-02-06
Non-nullable typed property for Embeddable should be ignored when hydrating 


sneakyvv:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7.0

#### Summary

Required typed properties of nullable embeddables can not be hydrated in PHP 7.4.

#### Current behavior

Take this example:

```
/**
 * @ORM\Entity()
 */
class Foo
{    
    /**
     * @ORM\Embedded(class="App\Domain\Model\Bar")
     */
    private ?Bar $bar = null;
}
```
and the Embeddable
```
/**
 * @ORM\Embeddable()
 */
class Bar
{
    /**
     * @ORM\Column(type="datetime_immutable", nullable=true)
     */
    private \DateTimeImmutable $startDate;
}
```
So, the `$bar` property on `Foo` can be null, but when it is provided it should have a `$startDate`. However, since it's part of the Embeddable, it could be `null` in the database for the table `foo`. 
But, when it's hydrated it should not set that null value, otherwise you get this error: 
```
Typed property App\Domain\Model\Bar::$startDate must be an instance of DateTimeImmutable, null used - #0 ..lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php(99): ReflectionProperty->setValue(Object(App\Domain\Model\Bar), NULL)
```

#### How to reproduce

See example above, create an instance of Foo, save it and load/hydrate it again.

#### Expected behavior

IMO, a null value, for non-nullable typed properties should be ignored in Embeddables.

PS: Kinda related to #4568, because when all properties of the Embeddable are null, the parent property should also be null (which I fix now post-hydrating), but that occurs even later obviously, after all properties of the Embeddable are set.

beberlei:
@sneakyvv which version of doctrine/persistence and doctrine/reflection are you on? The latest ones contain fixes for typed propreties, I hoped they work on embeddables as well.

sneakyvv:
@beberlei, I saw the ticket handling that, and upgrade my packages, but the mentioned bug still happened. 

My current versions are:
doctrine/persistence: 1.3.6
doctrine/reflection: 1.1.0

sneakyvv:
@beberlei I've made the test work (or be skipped) for older PHP versions.

Travis is now passing, since I've conditionally included a file containing the php7.4-style classes. 
Is there a better, more elegant way to include such conditional code in the test file itself? The purpose is to avoid syntax errors for older versions.

Issue #4568, opened 2015-05-08
DDC-3730: Embeddable hydrates to object instead of null


doctrinebot:
Jira issue originally created by user pcnc:

Good afternoon,

When hydrating an Embeddable with nullable attributes the result is an instance of the Embeddable class , this is obviously correct and expected behavior. 

If all the attributes are null the hydrator will still return an instance of the class with all of its properties null , even if I persist and flush my Entity with the Embeddable being set as null . 

For clarification :

```

class MyEntity
{
    protected $myEmbeddable;

    public function setMyEmbeddable(MyEmbeddable $myEmbeddable = null)
    {
        $this->myEmbeddable = $myEmbeddable;
    }
    [...]
}

$newEntity = new MyEntity();
$newEntity->setMyEmbeddable(null);

$em->persist($newEntity);
$em->flsuh($newEntity);

```

Calling $newEntity->getMyEmbeddable() will return an instance of the MyEmbeddable object with all of it's attributes set to null .

I expected $newEntity->getMyEmbeddable() to be NULL . 

Can someone clarify is this is expected behaviour ? In case it is , how can I achieve what I'm looking for ? 

Best regards

boekkooi:
This is actually really confusing definitely when you combine it with the [docs](http://doctrine-orm.readthedocs.org/projects/doctrine-orm/en/latest/tutorials/embeddables.html#initializing-embeddables)

Ocramius:
Related: #4670

Ocramius:
Related: #4568

afoeder:
I must say I find this rather a bug than an improvement... agree with @boekkooi regarding the docs...

Ocramius:
Can anyone confirm if this is a dupe of #4670 ?

afoeder:
well it *is* already 4568 and 4670 is rather a duplicate of 1275, both actually "features" etc.

This here however is a Bug because embeddables whose all properties are nullable and being null hydrate to an empty object when being stored as null. This is misleading (saving != retrieving) and not according to the docs.

Ocramius:
@afoeder the order of the tickets is scrambled due to the fact that they were imported from Jira in December, heh

Harrisonbro:
Hi @Ocramius 

Has there been any further discussion on this topic? We've just hit into the same problem as described here on our first Doctrine project. 

Let me know if there's anything we can do to help — provide usage examples, code samples, discussions, etc.

Harrisonbro:
[...]

PR #8015, opened 2020-02-06
Ignore non-nullable typed property for Embeddable when hydrating

sneakyvv:
This is the PR requested for issue #8014.

Following the explanation in the bug report, when you have a null value in the database (which is valid because the parent property containing the Embeddable can be null, i.e. all the properties o f the Embeddable are null in the database), but the typed property is required (does not allow null), the hydrator should skip setting that null value to the typed property.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php
+++ b/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php
@@ -96,6 +96,8 @@ class ReflectionEmbeddedProperty extends ReflectionProperty
             $this->parentProperty->setValue($object, $embeddedObject);
         }
 
-        $this->childProperty->setValue($embeddedObject, $value);
+        if (null !== $value || null === $this->childProperty->getType() || true === $this->childProperty->getType()->allowsNull()) {
+            $this->childProperty->setValue($embeddedObject, $value);
+        }
     }
 }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8014Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8014Test.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Doctrine\Test\ORM\Functional\Ticket;
+
+use DateTimeImmutable;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH8014Test extends OrmFunctionalTestCase
+{
+    protected function setUp()
+    {
+        parent::setUp();
+
+        if (PHP_VERSION_ID >= 70400) {
+            $this->_schemaTool->createSchema(
+                [
+                    $this->_em->getClassMetadata(GH8014Foo::class),
+                ]
+            );
+        }
+    }
+
+    public function testNonNullablePropertyOfEmbeddableCanBeHydratedWithNullValueInDatabase()
+    {
+        if (PHP_VERSION_ID < 70400) {
+            self::markTestSkipped('This test only applies to PHP versions higher than 7.4');
+
+            return;
+        }
+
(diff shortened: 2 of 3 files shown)


SenseException:
@doctrine/doctrinecore Should this PR target the 2.7 branch?

alcaeus:
Since 2.6 is no longer maintained, this indeed needs to target 2.7.

sneakyvv:
target updated

PS: the [CONTRIBUTING.md](https://github.com/doctrine/orm/blame/master/CONTRIBUTING.md#L27) needs updating :-)

> I am submitting a bugfix for a stable release
> Your PR should target the lowest active stable branch (2.6).

SenseException:
> target updated
> 
> PS: the [CONTRIBUTING.md](https://github.com/doctrine/orm/blame/master/CONTRIBUTING.md#L27) needs updating :-)
> 
> > I am submitting a bugfix for a stable release
> > Your PR should target the lowest active stable branch (2.6).

Thank you. I guess we could change the link to the documentation page where the supported versions are listed and remove versions from the CONTRIBUTING.md.

beberlei:
This is the wrong bugfix. The problem is that `$foo->bar` is created, even though its saved as NULL.

If you add the following to line 36, it already fails:

```php
$this->assertNull($foo->bar);
```
The fix should be that this is already NULL. But thats actually hard to detect, because if all embeddable properities are null, should the embeddable be null? or is the embeddable present?

beberlei:
Maybe we can even use this, but its actually necessary to check this before line 89:

```php
if ($value === null) {
    return
}
```

sneakyvv:
@beberlei I totally agree, and think the embeddable _should_ be null, if all embeddable properties are null. But that's not the current behaviour. The current behaviour it to have the embeddable present (with all null values). 
That's what I said in the bug report (#8014)

> PS: Kinda related to #4568, because when all properties of the Embeddable are null, the parent property should also be null (which I fix now post-hydrating), but that occurs even later obviously, after all properties of the Embeddable are set.

Since #4568 ended rather badly... I decided to fix that myself like this with a PostLoad-listener:

```
<?php

namespace App\Infrastructure\Persistence\Doctrine;

use Doctrine\ORM\Event\LifecycleEventArgs;
use Doctrine\ORM\Mapping\ReflectionEmbeddedProperty;
use Doctrine\Persistence\ObjectManager;

class NullableEmbeddableFixer
{
    private ObjectManager $objectManager;

    public function postLoad(LifecycleEventArgs $eventArgs)
    {
        $this->objectManager = $eventArgs->getObjectManager();

        $this->fixNullEmbeddables($eventArgs->getObject());
    }

    private function fixNullEmbeddables($object, \ReflectionProperty $parentProperty = null, $parentObject = null)
    {
        $hasNonNullProperties = false;
        /** @var ClassMetadata $embeddedMetadata */
        $embeddedMetadata = $this->objectManager->getClassMetadata(get_class($object));
        $embeddedClasses = $embeddedMetadata->embeddedClasses;
        /** @var \ReflectionProperty $property */
        foreach ($embeddedMetadata->getReflectionProperties() as $property) {
            if ($property instanceof ReflectionEmbeddedProperty) {
                continue;
            }
            if (array_key_exists($property->getName(), $embeddedClasses)) {
                $this->fixNullEmbeddables($property->getValue($object), $property, $object);
            }

            if (null !== $property->getValue($object)) {
                $hasNonNullProperties = true;
            }
        }

        if (!$hasNonNullProperties) {
            $parentProperty->setValue($parentObject, null);
        }
    }
}
```

I'm not familiar enough with Doctrine's inner workings, and how the same might be achieved in an other way. So if you see a solution, and I might be of any help, let me know!

beberlei:
@sneakyvv see #8022 please

Issue #8016, opened 2020-02-12
[AbstractIdGenerator] - generate() methods : EntityManagerInterface instead EntityManager


youssefbenhssaien:
### Feature Request

Replace `EntityManager` parameter by `EntityManagerInterface` in `AbstractIdGenerator:generate()` methods

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | yes

#### Summary

Working with `EntityManagerDecorator`, break at `AbstractIdGenerator::generate(EntityManager $em)`  regarding an instance of class extending `EntityManagerDecorator` passed instead of `EntitManager`

PR #8017, opened 2020-02-12
#8016 - AbstractIdGenerator:generate() - EntityManagerInterface inste…

ybenhssaien:
Use `AbstractIdGenerator:generate(EntityManagerInterface $em)` instead of `AbstractIdGenerator:generate(EntityManager $em)`

> Resolve : #8016

### DIFF
--- a/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
+++ b/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
@@ -19,18 +19,18 @@
 
 namespace Doctrine\ORM\Id;
 
-use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 
 abstract class AbstractIdGenerator
 {
     /**
      * Generates an identifier for an entity.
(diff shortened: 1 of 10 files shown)


SenseException (changes requested):
Changing the type hint to `EntityManagerInterface` would be a BC break for the 2.7 branch. I would expect this break in the next major 3.0 version.

youssefbenhssaien:
Unfortunately yes, unless a minor version is planned between 2.7 and 2.8 (already include those changes) ...

youssefbenhssaien:
Since most of Generator are instantiated directly by the framework, only CUSTOM generators will be impacted (< PHP 7.4)

greg0ire:
> Unfortunately yes, unless a minor version is planned between 2.7 and 2.8 (already include those changes) ...

What do you mean by that? I don't understand…

> Since most of Generator are instantiated directly by the framework, only CUSTOM generators will be impacted (< PHP 7.4)

I don't think it makes the BC-break OK.

PR #8020, opened 2020-02-13
use unitOfWork::identityFlattener

ybenhssaien:
# Modifications
- Use UnitOfWork instanciated identityFlattener instead of reinstanciate it in `BasicEntityPersister`.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -314,7 +314,7 @@ class UnitOfWork implements PropertyChangedListener
     /**
      * @return IdentifierFlattener
      */
-    public function getIdentifierFlattener(): IdentifierFlattener
+    public function getIdentifierFlattener() : IdentifierFlattener
     {
         return $this->identifierFlattener;
     }
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -311,9 +311,6 @@ class UnitOfWork implements PropertyChangedListener
         $this->reflectionPropertiesGetter = new ReflectionPropertiesGetter(new RuntimeReflectionService());
     }
 
-    /**
-     * @return IdentifierFlattener
-     */
     public function getIdentifierFlattener() : IdentifierFlattener
     {
         return $this->identifierFlattener;
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -312,6 +312,14 @@ class UnitOfWork implements PropertyChangedListener
     }
 
     /**
+     * @return IdentifierFlattener
+     */
+    public function getIdentifierFlattener(): IdentifierFlattener
+    {
+        return $this->identifierFlattener;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -311,6 +311,14 @@ public function __construct(EntityManagerInterface $em)
         $this->reflectionPropertiesGetter = new ReflectionPropertiesGetter(new RuntimeReflectionService());
     }
 
+    /**
+     * @return IdentifierFlattener
+     */
+    public function getIdentifierFlattener(): IdentifierFlattener
```
SenseException:
```suggestion
    public function getIdentifierFlattener() : IdentifierFlattener
```

Also please remove the method doc-block to fix the coding style issues mentioned in the travis-ci build.

SenseException (changes requested):

greg0ire:
Hi! Can you please explain why this is needed? It makes the public API of UnitOfWork bigger, not sure we want that.

youssefbenhssaien:
@greg0ire for two reasons :
- Working on an already instantiated IdentityFlattener instead doing it twice.
- Allow hacking UnitOfWork to add some custom Persister (like BulkInsertPersister on MySQL) since there is no way to custom persisters.

greg0ire:
> Allow hacking UnitOfWork to add some custom Persister (like BulkInsertPersister on MySQL) since there is no way to custom persisters.

Can you please elaborate on this? It's not obvious for me.

youssefbenhssaien:
> > Allow hacking UnitOfWork to add some custom Persister (like BulkInsertPersister on MySQL) since there is no way to custom persisters.
> 
> Can you please elaborate on this? It's not obvious for me.

UnitOfWork creates a Persister per inheritance type (cannot add a custom persister), Adding a custom BulkUnitOfWork inherit from UnitOfWork (this is not a good solution but had to) then return the new custom persister (BulkInsertPersister)
> Using ReflectionApi to bypass many inheritance limits (Some class constructs excpect EntityManager instead of EntityManagerInterface + Others expect instance of UnitOfWork instead the custom BulkUnitOfWork....)

This is a basic example : https://github.com/youssefbenhssaien/doctrine-bundle/blob/master/src/Doctrine/BulkUnitOfWork.php

> This example (only experimental) works only with one case when persisting an entity without inheritance Map.

greg0ire:
Thanks I understand better now. Here is a related issue: https://github.com/doctrine/orm/issues/4767

youssefbenhssaien:
> Thanks I understand better now. Here is a related issue: #4767

Thank you, but still not resolved and cannot inject a custom persister ! the only workaround is doing it with Reflection (but with BulkInsert it is more complicated)

greg0ire:
I know it's not resolved, I was including here in case other maintainers were wondering why there is no support for custom persisters, and maybe give it another look.

IMO there could maybe be a `PersisterRegistry` object that would be configurable and be able to provide the right persister for each case.

I'm not super sure about that design, but I don't think allowing hacks that involve inheritance + making public things that should probably not be is the way.

I hope a more seasoned ORM maintainer sees this and gives a better recommendation regarding what should be done here.

PR #8023, opened 2020-02-15
Put into cache using root entity name

peterkeatingie:
The cache key for reading is created based on the entity root name but the cache was written with a key based on the found entity (the child in this case) resulting in cache miss.
Using the same root name when putting fixes the issue. I do believe that any extending instances should have unique identifiers as they will have to share an identifier from the base class. This will mean the key won't be ambiguous - if my thinking is correct.

https://github.com/doctrine/orm/issues/7969


beberlei (approved):
Just checked the Cache code myself, everywhere cache keys are created the root entity name is used, except here. So that is definately a bug and the right fix.

PR #8025, opened 2020-02-16
[GH-4641] IDENTITY() field type conversion

beberlei:
This fixes a bug with a missing type conversion for `IDENTITY()` function values and data was always returned with type "string". Now the specified type is used and the new typed function node functionality #7941 allows converting it to the real PHP type.

Fixes #4641


lcobucci (approved):
Changes LGTM, I'm just unsure if we have tests covering both paths (`$this->fieldMapping === null` and `$this->fieldMapping !== null`)

PR #8027, opened 2020-02-17
Fixes #8026 - IterableResult has wrong "valid" behavior

### LINKED ISSUE #8026: IterableResult has wrong "valid" behavior
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7.0

#### Summary

To implement iterator's method `valid()`, `\Doctrine\ORM\Internal\Hydration\IterableResult`, [uses](/doctrine/orm/blob/ca57222010de9dfa57b978be37ca4571178df67c/lib/Doctrine/ORM/Internal/Hydration/IterableResult.php#L108) a loosely comparison (`==`) instead of a strict one (`===`): 
```php
// ...
private $_current = null;
// ...
public function valid()
{
    return ($this->_current!=false);
}
// ...
```

#### Current behavior

Because of that, the very first validation (before revinding or iteration) is always wrong.

#### How to reproduce

Just try to validate an iterator before any iteration methods are called. 

#### Expected behavior

Should return `(bool) true` if the current position isn't filled yet. 

#### How to fix

The master-branch, already has this issue resolved (in the 5221e999 commit).

awd-studio:
(cherry picked from commit 5221e999c2057dbf25d1a60dd6c82ffb30bf1f1c)

closes #8026

### DIFF
--- a/lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
+++ b/lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
@@ -1,22 +1,6 @@
 <?php
 
-/*
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
- * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
- * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
- * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
- * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
- * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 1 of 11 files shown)


greg0ire:
The commit you cherry pick is huge and starts with `[CS]`. Please make it smaller and use a commit message that shows better how this is a fix. Also, please have a look at Travis, these changes make the test suite fail.

beberlei:
I believe the fix is only to use `===`. Not sure the "fix" was intentional btw, just making everything stricter.

awd-studio:
Thanks, guys, for tips. 
I'll make another PR. (Is this one suppose to be closed?) 

But the tests, they're not running on the raw 2.7 branch:
```sh
ERRORS!
Tests: 3415, Assertions: 7425, Errors: 1153, Failures: 21, Skipped: 100, Incomplete: 2.
```
So that, I can't even check them before changes. Maybe, I need some extra environment except PHP (I use 7.3 locally)?

SenseException:
> I'll make another PR. (Is this one suppose to be closed?)
You can also undo your commit and create a new one.

You don't need to have all the databases for your test of the iterator.

beberlei:
@awd-studio did you run "composer install" after switching to 2.7 branch?

awd-studio:
Yes, of course. Same errors were in travis.

PR #8032, opened 2020-02-20
Throw exception with more meaningful details

oojacoboo:
Currently an undefined index is thrown.  This error is confusing and not helpful, so we perform a check to see if the key is set and throw a more meaningful exception.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2478,7 +2478,7 @@ class UnitOfWork implements PropertyChangedListener
         if (!isset($this->entityIdentifiers[$id])) {
             throw new EntityNotFoundException(sprintf(
                 'Unable to find "%s" entity identifier associated with the UnitOfWork',
-                get_class($entity),
+                get_class($entity)
             ));
         }
 
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2474,7 +2474,15 @@ class UnitOfWork implements PropertyChangedListener
      */
     public function getEntityIdentifier($entity)
     {
-        return $this->entityIdentifiers[spl_object_id($entity)];
+        $id = spl_object_id($entity);
+        if (!isset($this->entityIdentifiers[$id])) {
+            throw new EntityNotFoundException(sprintf(
+                'Unable to find "%s" entity identifier associated with the UnitOfWork',
+                get_class($entity),
+            ));
+        }
+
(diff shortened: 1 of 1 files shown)


oojacoboo (commented):
Removed trailing comma

SenseException (changes requested):
Thank you for your contribution. Can you please ~~target your pull request to the 2.7 branch and~~ add a test that covers the newly the added code?

beberlei:
@oojacoboo when do you see this exception being a notice? Please post a stack trace and maybe some code what you are doing.

oojacoboo:
@beberlei the issue was related to the EntityManager being cleared and an entity subsequently being associated with another entity as a relation.

```php
$foo = $repo->find(1);
$em->clear();
$bar = new Bar($foo);
$em->persist($bar);
$em->flush();
```

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2474,7 +2474,15 @@ public function setOriginalEntityProperty($oid, $property, $value)
      */
     public function getEntityIdentifier($entity)
     {
-        return $this->entityIdentifiers[spl_object_id($entity)];
+        $id = spl_object_id($entity);
```
beberlei, line 2477: This can be optimized to:
```suggestion
        return $this->entityIdentifiers[spl_object_id($entity)] ?? EntityNotFoundException::noIdentifierFound(get_class($entity));
```
This keeps the number of executed opcodes lower in this hot path.

beberlei (changes requested):
Please refactor as commented and add a test to `UnitOfWorkTest.php` test-case and change branch to 2.8 then I can merge it.

If you are ok, then we can do these changes as well.

oojacoboo:
Okay, I had no idea how this pull requests gets put onto another branch.  I've submitted a new pull request.

https://github.com/doctrine/orm/pull/8201

greg0ire:
> Okay, I had no idea how this pull requests gets put onto another branch. I've submitted a new pull request.

That can be achieved by clicking the Edit button next to the title of your PR, and then force pushing your rebased commit on the branch.

oojacoboo:
@greg0ire and all the merge conflicts?

greg0ire:
That's what the "rebased" part is about. In your case, you would run `git rebase patch-2 --onto origin/2.8.x origin/master patch-2`, fix them, and then force push.

oojacoboo:
@greg0ire yea, that makes sense.  However, if there are many commits, that could take forever.  It's far easier to just cherry pick IMO.

PR #8036, opened 2020-02-23
Fix inherited embeddables and nesting after AnnotationDriver change #8006

malarzm:
Changing `AnnotationDriver` to mark embeddables as "entities"  in #8006 caused a bunch of bugs with inheritance and nested embeddables that are fixed by this PR.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -776,8 +776,9 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
      */
     protected function isEntity(ClassMetadataInterface $class)
     {
-        return isset($class->isMappedSuperclass) && $class->isMappedSuperclass === false
-            && isset($class->isEmbeddedClass) && $class->isEmbeddedClass === false;
+        assert($class instanceof ClassMetadata);
+
+        return $class->isMappedSuperclass === false && $class->isEmbeddedClass === false;
     }
 
     /**
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -401,10 +401,10 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
     private function addInheritedFields(ClassMetadata $subClass, ClassMetadata $parentClass)
     {
         foreach ($parentClass->fieldMappings as $mapping) {
-            if ( ! isset($mapping['inherited']) && ! $parentClass->isMappedSuperclass && ! $parentClass->isEmbeddedClass) {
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -277,6 +277,8 @@ public function loadMetadataForClass($className, ClassMetadata $metadata)
         /* @var $property \ReflectionProperty */
         foreach ($class->getProperties() as $property) {
             if ($metadata->isMappedSuperclass && ! $property->isPrivate()
```
beberlei: I believe the time has come to move the whole expression of this if condition into a function :-)

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -401,7 +401,7 @@ private function getShortName($className)
     private function addInheritedFields(ClassMetadata $subClass, ClassMetadata $parentClass)
     {
         foreach ($parentClass->fieldMappings as $mapping) {
-            if ( ! isset($mapping['inherited']) && ! $parentClass->isMappedSuperclass) {
+            if ( ! isset($mapping['inherited']) && ! $parentClass->isMappedSuperclass && ! $parentClass->isEmbeddedClass) {
```
beberlei: Couldn't you use the `isEntity` method here? This function is used during mapping construction time, so its not critical here to avoid too many function calls.

beberlei (changes requested):
Yay! Good catch

lcobucci:
It seems like these changes are treating embeddable objects as mapped superclasses... can we also make sure we don't add support to associations for embeddable? This would be quite odd design-wise since they're value objects.

beberlei:
@lcobucci I am afraid adding relationships to embeddables was possible before or not? 

If it was possible before I don't think we should prevent adding associations though, since Doctrine doesn't prescribe a paradigm such as DDD, we don't really have value objects here but embeddables.

lcobucci:
It wasn't possible before

malarzm:
I confirm using references wasn't possible as I was really sad when I found out ;)

lcobucci:
If you want to expand the concept of an embeddable, that's fine. I'd just suggest to avoid, as much as, doing that on a patch release.

We'd also have to make sure that DQL works properly, btw.

malarzm:
> If you want to expand the concept of an embeddable, that's fine. I'd just suggest to avoid, as much as, doing that on a patch release.

I agree, that'd be a new feature anyway

> We'd also have to make sure that DQL works properly, btw.

Is this about current patch or still references?

lcobucci:
References 😁

beberlei:
There isn't a specific error on 2.7.0 with associations on embedded objects though. They are just not stored. The embeddable has the association, but its not copied over to the Entity. The same is true of 2.7.2 with this patch. So there is nothing we need to change.

beberlei:
Added this issue to track improvement to schema validator to disallow this: https://github.com/doctrine/orm/issues/8052

beberlei (approved):

PR #8041, opened 2020-02-26
Improve documentation about naming strategy usage

RosemaryOrchard:
Add a note/warning that annotations override the naming strategy.

### DIFF
--- a/docs/en/reference/namingstrategy.rst
+++ b/docs/en/reference/namingstrategy.rst
@@ -4,8 +4,12 @@ Implementing a NamingStrategy
 .. versionadded:: 2.3
 
 Using a naming strategy you can provide rules for generating database identifiers,
-column or table names when the column or table name is not given. This feature helps
+column or table names. This feature helps
 reduce the verbosity of the mapping document, eliminating repetitive noise (eg: ``TABLE_``).
+
+.. warning
+
+    The naming strategy is always overridden by entity mapping such as the `Table` annotation.
 
 Configuring a naming strategy
 -----------------------------


docs/en/reference/namingstrategy.rst
```diff
@@ -21,6 +21,12 @@ You can specify a different strategy by calling ``Doctrine\ORM\Configuration#set
     <?php
     $namingStrategy = new MyNamingStrategy();
     $configuration->setNamingStrategy($namingStrategy);
+    
+ 
+
+Warning
+
+The naming strategy is always overridden by annotations on an entity.
```
lcobucci: To render a nice warning block we should format this like:

```rst
.. warning

    The naming strategy is always overridden by annotations on an entity.
```

---

I'd also suggest moving this block to the intro (right before of the title`Configuring a naming strategy`) to make things a bit more explicit.
Text-wise I think we could a bit more verbose since I noticed that we do mention this on the [intro "when the column or table name is not given"](https://github.com/doctrine/orm/pull/8041/files#diff-88b1a8bc0ff7545ed3ba546bf87a03d4R7).

Perhaps removing the "when the column or table name is not given" and saying something like "Naming strategies are completely ignored when table or column names are provided in the entity mapping", what do you think?

lcobucci (changes requested):
Thanks for your contribution, here are some thoughts on how to make it better :)

RosemaryOrchard:
I've moved it up, changed the formatting and removed the part of the intro that implied this but didn't explicitly state it.

docs/en/reference/namingstrategy.rst
```diff
@@ -4,9 +4,13 @@ Implementing a NamingStrategy
 .. versionadded:: 2.3
 
 Using a naming strategy you can provide rules for generating database identifiers,
-column or table names when the column or table name is not given. This feature helps
+column or table names. This feature helps
 reduce the verbosity of the mapping document, eliminating repetitive noise (eg: ``TABLE_``).
 
+.. warning
+
+    The naming strategy is always overridden by annotations on an entity.
```
lcobucci: Nice, quite straightforward. The only issue here is that entities can be mapped by XML files, hence my suggestion of saying "entity mapping".

greg0ire: Might be perceived as a bit abstract for new users though, adding an example might help:

```suggestion
    The naming strategy is always overridden by entity mapping such as the `Table` annotation.
```

RosemaryOrchard: I agree, I also suspect that people will be more likely to remember they've used XML mapping than an annotation copied from an example/another file and just updated.

greg0ire (approved):

SenseException (approved):
Thank you

PR #8044, opened 2020-02-28
No docs in distribution

Spomky:
This PR adds the folder `/docs` in .gitattributes so that it will not be present in distribution


Spomky:
Other files like UPGRADE.md or phpbench.json could be added as well.
Just ask and I will update that PR.

Regards

SenseException:
Thank you. phpbench.json sounds like a candidate too, but I'm unsure about `UPGRADE.md`. It could be relevant for those who update their project.

Spomky:
> Thank you. phpbench.json sounds like a candidate too

Thank you. I will add it soo.

> I'm unsure about `UPGRADE.md`. It could be relevant for those who update their project.

This page is supposed to by used during the development process. I don't see any reason why it should be part of the distributed package.
I may be wring, but except the code and package stuff I only keep the license file.

SenseException:
README and the licence are distributed too and for the code-surfing developers it might be handy to have the `UPGRADE.md` there.

Spomky:
Hi @SenseException,

Sorry I was not clear in my last comment. For me `distributed package` means the package in production i.e. when the composer option `--prefer-dist` is used. In that case the documentation files (except the license that shall be present at all time) are useless. The package is supposed to run on a server.

For code-surfing developers (on Github or with composer option `--prefer-source`), I agree with you and the documentation should be present.

So I still think we could add the md  files in the list.

SenseException:
If you cover that case, you can add `UPGRADE.md` too. Thanks for clarification. :+1:

Spomky:
I am not sure failing tests come from this PR.
Any thoughts?

greg0ire:
It comes from the branch you are targeting, you should target 2.8.x

beberlei:
Every regular composer require using a stable vetsion uses the distributed packages, why UPGRADE.md must be part of the distributed release

Spomky:
> UPGRADE.md must be part of the distributed release

Ok 👌 
Please note that I closed the PR as it was not on the correct branch. I will create a new one based on the `2.8` or `2.9` one.

PR #8046, opened 2020-02-28
feat(schema-tools): support name space on schema creation

### LINKED ISSUE #8038: Name spaces not supported when creating a Schema instance from a given set of metadata classes
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7.1

#### Current behavior
Name spaces not supported during schema creation.

#### How to reproduce

1.  Create an entity with `@Table(schema="mySchema", name="myTable")`
2. Launch the command `doctrine:schema:update --force`
3. you'll have an SQL error `the schema mySchema doesn't exists`
4. To understand exactly what happen, lunch the command `doctrine:schema:update --dump-sql`, you'll see that an SQL instruction is missing `create schema mySchema;`

#### Expected behavior
Support name spaces during schema creation.

abdel-geniusespro:
Closes #8038

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -19,10 +19,12 @@
 
 namespace Doctrine\ORM\Tools;
 
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Schema\AbstractAsset;
 use Doctrine\DBAL\Schema\Comparator;
 use Doctrine\DBAL\Schema\Index;
 use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Schema\SchemaException;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\Visitor\DropSchemaSqlCollector;
 use Doctrine\DBAL\Schema\Visitor\RemoveNamespacedAssets;
@@ -141,7 +143,9 @@ class SchemaTool
      *
      * @return Schema
      *
-     * @throws \Doctrine\ORM\ORMException
+     * @throws ORMException
+     * @throws DBALException
+     * @throws SchemaException
      */
     public function getSchemaFromMetadata(array $classes)
     {
@@ -164,6 +168,7 @@ class SchemaTool
             }
 
             $table = $schema->createTable($this->quoteStrategy->getTableName($class, $this->platform));
+            $this->addSchemaNameSpaceFromTable($schema, $table);
 
             if ($class->isInheritanceTypeSingleTable()) {
                 $this->gatherColumns($class, $table);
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -164,6 +168,7 @@ public function getSchemaFromMetadata(array $classes)
             }
 
             $table = $schema->createTable($this->quoteStrategy->getTableName($class, $this->platform));
+            $this->addSchemaNameSpaceFromTable($schema, $table);
```
greg0ire:
```suggestion
            $this->addSchemaNamespaceFromTable($schema, $table);
```

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -368,6 +373,27 @@ function (ClassMetadata $class) use ($idMapping) : bool {
...
     }
 
+    /**
+     * @param Schema $schema
+     * @param Table  $table
+     *
+     * @throws DBALException
+     * @throws SchemaException
+     */
+    private function addSchemaNameSpaceFromTable(Schema $schema, Table $table): void
```
greg0ire:
```suggestion
    private function addSchemaNamespaceFromTable(Schema $schema, Table $table): void
```

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -368,6 +373,27 @@ function (ClassMetadata $class) use ($idMapping) : bool {
         return $schema;
     }
 
+    /**
+     * @param Schema $schema
+     * @param Table  $table
+     *
```
greg0ire: No need for such phpdoc when you have type declarations.

```suggestion
```

greg0ire (commented):
There probably should be tests for this, and maybe docs.

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -141,7 +143,9 @@ private function processingNotRequired($class, array $processedClasses)
      *
      * @return Schema
      *
-     * @throws \Doctrine\ORM\ORMException
+     * @throws ORMException
+     * @throws DBALException
+     * @throws SchemaException
```
SenseException: `DBALException`and `SchemaException` are part of DBAL. Are those exceptions caught in ORM by using this method? If not, I would expect that an `ORMException` would be thrown if internally the DBAL exceptions occur.

lib/Doctrine/ORM/Tools/SchemaTool.php
abdel-geniusespro: I did it to keep the file homogeneous

abdel-geniusespro:
Thank's @SenseException & @greg0ire for the review, but this is not the best way to solve this problem.

PR #8049, opened 2020-03-01
Run Static analysis and coding standards in Github Actions

beberlei:
This moves static analysis runs from Travis-CI to Github Actions, because it allows sending violations directly into the PR files as annotations very easily.

This will look like this:

Screenshot_2020-03-01 Run Static analysis and coding standards in Github Actions by beberlei · Pull Request #8049 · doctrin

I haven't moved phpbench yet, because it looks like the history.db isn't even cached on Travis, so I don't know if this step can fail the build?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,27 +6,7 @@ on:
 jobs:
   coding-standards:
     name: "Coding Standards"
-    runs-on: ubuntu-latest
-    steps:
-    - name: Checkout
-      uses: actions/checkout@v2
-
-    - name: Setup PHP
-      uses: shivammathur/setup-php@v2
-      with:
-        php-version: '7.4'
-        extensions: mbstring
-        tools: composer, cs2pr
-
-    - name: composer install
-      run: "composer install --no-progress"
-
-    - name: phpcs
-      run: "php vendor/bin/phpcs -q --report=checkstyle --no-colors | cs2pr"
-
-  static-analysis:
-    name: "Static Analysis"
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     steps:
     - name: Checkout
       uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


greg0ire:
Weird, the names don't seem to have been taken into account… let's try to close and reopen.

beberlei:
@greg0ire they are taken into account in the check list below the comments for me.

greg0ire:
Ah wait it did work, it's just that it's still prepended with `.github/workflows/ci.yml` for some reason.

greg0ire:
There are no prefixes on this PR for instance: https://github.com/doctrine/coding-standard/pull/161

greg0ire (dismissed):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+    name: "Coding Standards"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '7.4'
```
lcobucci: We should run tools using the lowest supported version to avoid compatibility issues - e.g.: our new CS release adds property type declaration when running on PHP 7.4

SenseException: Do new PHP versions enable new sniffs that aren't needed to be checked in older versions?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+jobs:
+  coding-standards:
+    name: "Coding Standards"
+    runs-on: ubuntu-latest
```
SenseException: Using the latest version might break builds in the future. What about
```suggestion
    runs-on: ubuntu-18.04
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    - name: composer install
+      run: "composer install --no-progress"
+
+    - name: phpcs
+      run: "php vendor/bin/phpcs -q --report=checkstyle --no-colors | cs2pr"
+
+  static-analysis:
+    name: "Static Analysis"
+    runs-on: ubuntu-latest
```
SenseException: As mentioned above.
```suggestion
    runs-on: ubuntu-18.04
```

SenseException (changes requested):

.github/workflows/ci.yml
lcobucci: Don't enable new sniffs but modify sniffs behaviour, like the property type declaration I mentioned.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+      with:
+        php-version: '7.4'
+        extensions: mbstring
+        tools: composer, cs2pr
+
+    - name: composer install
+      run: "composer install --no-progress"
+
+    - name: phpcs
+      run: "php vendor/bin/phpcs -q --report=checkstyle --no-colors | cs2pr"
```
lcobucci: We should also run the diff sniffer in v2.x since we don't have all files following the CS in that branch.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '7.2'
+        extensions: mbstring
+        tools: composer, cs2pr
+
+    - name: composer install
+      run: "composer install --no-progress"
```
carusogabriel:
```suggestion
      run: "composer install --no-progress --no-suggest --no-interaction --prefer-dist --optimize-autoloader"
```

Issue #7633, opened 2019-03-02
Partial objects get cached and break "actual" objects


Amunak:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

#### Summary

When using partial objects and second-level cache, the partial objects can get cached and then they get retrieved from the cache instead of querying for the "full" object.

#### Current behavior
Doctrine doesn't properly recognize query as fetching a partial object, caches it and then later when asked to query the actual object it only retreives the partial from cache.

#### How to reproduce

The following snippet would save a partial object to your second-level cache:

```php
class Object {
	public $id;
	public $field1;
	public $field2;
}

/** @var $qb  Doctrine\ORM\QueryBuilder */
$qb
	->select('partial o.{id, field1}')
	->from(Object::class, 'o')
	->setCacheable(true)
;

// when second level cache is enabled, this caches the resulting partial object(s)
$partial = $qb->getQuery()->getResult();
```

Then later running this code:
```php
/** @var $om Doctrine\Common\Persistence\ObjectManager */
$real = $om->find(Object::class, 1);
```

`$real` would actually contain only the partial object data - property `$real->field2` would be null and if you could compare the `$partial` and `$real` objects you would notice that they are essentialy the same.

#### Expected behavior
Ideally Doctrine would cache partial objects separately. While you could argue "why use partial objects when you cache the actual objects anyways and save on speed there" they are great for optimizations and in some cases you don't want to load hundreds of potentially big objects as a whole when you need only a small portion of them.

However at the very least it should recognize that the query above is indeed a partial query and throw CacheException - as it does when you modify the snippet like this:

```php
/** @var $qb  Doctrine\ORM\QueryBuilder */
$qb
	->select('partial o.{id, field1}')
	->from(Object::class, 'o')
	->setCacheable(true)
;
$query = $qb->getQuery();
$query->setHint(Query::HINT_FORCE_PARTIAL_LOAD, 1);

$partial = $query->getResult();
```

Obviously that makes the query unusable as it now throws an exception, so the actual solution is to just not mark it as cacheable.

No matter what solution will be implemented I would strongly advice to also mention this caveat of partial objects in their [documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/partial-objects.html) - there is no mention of them not being cacheable anywhere.

---

[...]

PR #8050, opened 2020-03-01
[GH-7633] disallow cache partial objects

beberlei:
There was a check in DefaultQueryCache that prevented partial queries,
because they are not supported. However the checked hint
Query::HINT_FORCE_PARTIAL_LOAD is optional, so cant be used to prevent
caching partial DQL queries.

Introduce a new hint that the SqlWalker sets on detecing a PARTIAL
query and throw an exception in the DefaultQueryCache if thats found.

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -260,7 +260,7 @@ class DefaultQueryCache implements QueryCache
             throw new CacheException("Second-level cache query supports only select statements.");
         }
 
-        if (isset($hints[Query::HINT_FORCE_PARTIAL_LOAD]) && $hints[Query::HINT_FORCE_PARTIAL_LOAD]) {
+        if (isset($hints[Query\SqlWalker::HINT_PARTIAL]) && $hints[Query\SqlWalker::HINT_PARTIAL]) {
             throw new CacheException("Second level cache does not support partial entities.");
         }
 
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -260,7 +260,7 @@ class DefaultQueryCache implements QueryCache
             throw new CacheException("Second-level cache query supports only select statements.");
         }
 
-        if (isset($hints[Query\SqlWalker::HINT_PARTIAL]) && $hints[Query\SqlWalker::HINT_PARTIAL]) {
+        if (($hints[Query\SqlWalker::HINT_PARTIAL] ?? false) === true || ($hints[Query::HINT_FORCE_PARTIAL_LOAD] ?? false) === true) {
             throw new CacheException("Second level cache does not support partial entities.");
         }
 
--- a/tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheQueryCacheTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheQueryCacheTest.php
@@ -1095,7 +1095,6 @@ class SecondLevelCacheQueryCacheTest extends SecondLevelCacheAbstractTest
         $this->loadFixturesCountries();
 
         $this->_em->createQuery("SELECT PARTIAL c.{id} FROM Doctrine\Tests\Models\Cache\Country c")
-            ->setHint(Query::HINT_FORCE_PARTIAL_LOAD, true)
             ->setCacheable(true)
             ->getResult();
     }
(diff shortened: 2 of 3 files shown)


lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -260,7 +260,7 @@ public function put(QueryCacheKey $key, ResultSetMapping $rsm, $result, array $h
             throw new CacheException("Second-level cache query supports only select statements.");
         }
 
-        if (isset($hints[Query::HINT_FORCE_PARTIAL_LOAD]) && $hints[Query::HINT_FORCE_PARTIAL_LOAD]) {
+        if (isset($hints[Query\SqlWalker::HINT_PARTIAL]) && $hints[Query\SqlWalker::HINT_PARTIAL]) {
```
lcobucci: Maybe `if ($hints[Query\SqlWalker::HINT_PARTIAL] ?? false === true) {` is simpler?

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -260,7 +260,7 @@ public function put(QueryCacheKey $key, ResultSetMapping $rsm, $result, array $h
             throw new CacheException("Second-level cache query supports only select statements.");
         }
 
-        if (isset($hints[Query::HINT_FORCE_PARTIAL_LOAD]) && $hints[Query::HINT_FORCE_PARTIAL_LOAD]) {
```
lcobucci: I have never used this hint but is there any context where it would be used and, with this change, would start caching things in the L2C?

lcobucci (dismissed):
Awesome! It's great that we have this being handled automatically now, thanks 👍

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
beberlei: You are right, that might be the case. Both hints need to lead to exception.

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -260,7 +260,7 @@ public function put(QueryCacheKey $key, ResultSetMapping $rsm, $result, array $h
             throw new CacheException("Second-level cache query supports only select statements.");
         }
 
-        if (isset($hints[Query\SqlWalker::HINT_PARTIAL]) && $hints[Query\SqlWalker::HINT_PARTIAL]) {
+        if (($hints[Query\SqlWalker::HINT_PARTIAL] ?? false) === true || ($hints[Query::HINT_FORCE_PARTIAL_LOAD] ?? false) === true) {
```
lcobucci: Then we can do this:

```suggestion
        if (($hints[Query\SqlWalker::HINT_PARTIAL] ?? $hints[Query::HINT_FORCE_PARTIAL_LOAD] ?? false) === true) {
```

tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheQueryCacheTest.php
```diff
@@ -1099,6 +1099,21 @@ public function testCacheablePartialQueryException()
...
+    public function testCacheableForcePartialLoadHintQueryException()
+    {
+        $this->evictRegions();
+        $this->loadFixturesCountries();
+
+        $this->_em->createQuery("SELECT c FROM Doctrine\Tests\Models\Cache\Country c")
+            ->setCacheable(true)
+            ->setHint(Query::HINT_FORCE_PARTIAL_LOAD, true)
+            ->getResult();
+    }
```
lcobucci: `@expectedException` and `@expectedExceptionMessage` have been deprecated in PHPUnit 8.x and removed [in PHPUnit v9.0](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-9.0.md#removed). We should avoid using them.

We might need to import `Doctrine\ORM\Cache\CacheException` but this should be the way to go:

```suggestion
    public function testCacheableForcePartialLoadHintQueryException()
    {
        $this->evictRegions();
        $this->loadFixturesCountries();
        
        $this->expectedException(CacheException::class);
        $this->expectedExceptionMessage('Second level cache does not support partial entities.');

        $this->_em->createQuery("SELECT c FROM Doctrine\Tests\Models\Cache\Country c")
            ->setCacheable(true)
            ->setHint(Query::HINT_FORCE_PARTIAL_LOAD, true)
            ->getResult();
    }
```

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
beberlei: I like to keep it this way now.

tests/Doctrine/Tests/ORM/Functional/SecondLevelCacheQueryCacheTest.php
beberlei: 99 occurrences of this in tests on 2.7, i move this to a mainteance task for 2.8

beberlei: 99 occurrences of this in tests on 2.7, i move this to a mainteance task for 2.8

PR #8055, opened 2020-03-04
Fix regression in 2.7.1 when `mysqli` is used with discriminator column that is not a string

mpdude:
In 2.7.1, #7974 was included as the fix for #7505. Basically, the change made was that when hydrating entities with inheritance, only the fields/data that belong to the particular inheritance type must be passed to `UoW::createEntity()`.

To do so, the `SimpleObjectHydrator` will compare the discriminator column value against the list of known types/mappings.

https://github.com/doctrine/orm/blob/927305764930d714f7c6d8e3727bdfdc910c9703/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php#L140-L143

While the descriminator types in `$cacheKeyInfo['discriminatorValues']` are always strings, the actual `$discrColumnValue` may be an `int`, at least when the discriminator column is defined as int and the `mysqli` driver is used. 

The problem seems not to occur with `pdo_mysql`: In that case, the values returned from the result set/query are (always?) strings.

My suggested fix is to `(string)` cast the discriminator value, as it already happens in the `AbstractHydrator`:

https://github.com/doctrine/orm/blob/927305764930d714f7c6d8e3727bdfdc910c9703/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php#L300-L306

*Update:* 

It seems Travis is running MySQL tests with `pdo_mysql` only, so the test case does not even show the issue. I am also sure that the issue would have surfaced with the existing tests.

I still believe the fix is correct, given the similarity to the code in `AbstractHydrator`. 

Anyway, could someone advise if this `pdo_mysql`/`mysqli` discrepancy is something worth looking into?

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -138,7 +138,7 @@ class SimpleObjectHydrator extends AbstractHydrator
             // Prevent overwrite in case of inherit classes using same property name (See AbstractHydrator)
             if ( ! isset($data[$fieldName]) || ! $valueIsNull) {
                 // If we have inheritance in resultset, make sure the field belongs to the correct class
-                if (isset($cacheKeyInfo['discriminatorValues']) && ! in_array($discrColumnValue, $cacheKeyInfo['discriminatorValues'], true)) {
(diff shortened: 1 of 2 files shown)


beberlei (approved):

mpdude:
What about running tests for `mysqli`?

I understand that the test matrix is already very complex and resource intensive.

But unless we run `mysqli`, the regression test is useless and we’re not going to find such bugs.

PR #8057, opened 2020-03-04
Sets EntityManagerInterface instead of EntityManager when needed

studiocaramia:
When implementing a Custom EntityManager, all IdGenerators was typed with EntityManager instead of EntityManagerInterface.

We also updated some doc blocks

### DIFF
--- a/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
+++ b/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
@@ -19,18 +19,18 @@
 
 namespace Doctrine\ORM\Id;
 
-use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 
 abstract class AbstractIdGenerator
 {
     /**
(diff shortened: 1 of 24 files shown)


lib/Doctrine/ORM/Id/AbstractIdGenerator.php
```diff
@@ -19,18 +19,18 @@
...
 {
     /**
      * Generates an identifier for an entity.
      *
-     * @param EntityManager $em
+     * @param EntityManagerInterface $em
      * @param object|null $entity
      * @return mixed
      */
-    abstract public function generate(EntityManager $em, $entity);
```
lcobucci: This is a BC-break, that's why we didn't do it on v2.x. v3.x (`dev-master`) already solved this AFAIK.

studiocaramia: Ok

SenseException:
Yes, master contains the `EntityManagerInterface`. Nevertheless, thank you for your effort.

Issue #5544, opened 2015-12-10
$qb->expr()->countDistinct() fails with multiple values


greeny:
I have $qb (QueryBuilder), and i call this:

`$qb->expr()->countDistinct('e.name', 'e.value')`

Generated SQL / DQL is valid, but Parser complains about syntax error:

`Error: Expected Doctrine\ORM\Query\Lexer::T_CLOSE_PARENTHESIS, got ','`

In tests, there is a test for multiple arguments in `countDistinct()`, but no test for Parser accepting multiple arguments for `COUNT(DISTINCT`

oskargunther:
Still not solved? Any work around?

lart2150:
I found a major hack of a workaround.  If you combine the two fields and then select distinct of the two it works. 
`COUNT(DISTINCT CONCAT(ps.date, s.id))`

lcobucci:
It would be amazing if someone could send us a failing test case that reproduces that behaviour. It would help us a lot to identify and fix the issue you're describing.

You can find examples on https://github.com/doctrine/doctrine2/tree/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket

peter-gribanov:
> It would be amazing if someone could send us a failing test case that reproduces that behaviour. It would help us a lot to identify and fix the issue you're describing.

@lcobucci necro post #8058

PR #8058, opened 2020-03-04
Сreate test for bug #5544

peter-gribanov:
In tests, this bug reproduce only if entity used as identifier.

Related #5544

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5544Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5544Test.php
@@ -1,0 +1,89 @@
+<?php
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Query;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH5544Test.php
```diff
@@ -0,0 +1,176 @@
...
+        $this->_em->flush();
+        $this->_em->clear();
+    }
+
+    private function createEntityIdentifierQuery(bool $distinct) : Query
+    {
+        return $this->_em
+            ->createQueryBuilder()
+            ->select(\sprintf(
+                'COUNT(%s CONCAT(ub.user.id, :concat_separator, ub.browser)) cnt',
```
peter-gribanov: It’s strange. Both of these queries do not work.
```sql
SELECT COUNT(DISTINCT CONCAT(ub.user, :concat_separator, ub.browser)) as cnt FROM GH5544UserBrowser ub
```
```sql
SELECT COUNT(DISTINCT CONCAT(ub.user.id, :concat_separator, ub.browser)) as cnt FROM GH5544UserBrowser ub
```

PR #8060, opened 2020-03-04
Fix: Give workflow a name

localheinz:
This PR

* [x] gives the workflow a name

### Before

Screen Shot 2020-03-04 at 22 20 39

### After

Screen Shot 2020-03-04 at 22 20 48


SenseException (approved):

localheinz:
Thank you, @beberlei and @SenseException!

Issue #8061, opened 2020-03-09
New object expression does not convert fields before creating object


jeroenvdheuvel:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.7.1

#### Summary

New object expression is not converting PHP value to SQL using the `convertToPHPValueSQL` method in a given type. This can break new object construction because wrong arguments are provided (due to not executing a SQL function on the database).

#### Current behavior

Object creation failed because the wrong argument type is provided, due to not running the convert function on the database.

#### How to reproduce

```php
Type::addType('FooType', FooType::class);
$dql           = 'SELECT NEW ' . FooNewObject::class . '(e.field) FROM ' . FooEntity::class . ' e WHERE e.id = :id';
$entityManager = $this->_getTestEntityManager();
$query         = $entityManager->createQuery($dql);

self::assertRegExp('/SELECT DatabaseFunction\(\w+\.field\) AS /', $query->getSQL());

/**
 * @Entity
 */
final class FooEntity
{
    /** @Id @Column(type="integer") @GeneratedValue */
    public $id;

    /**@Column(type="FooType") */
    public $field;
}

final class Type extends Type
{
    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform) : string
    {
        return $platform->getVarcharTypeDeclarationSQL($fieldDeclaration);
    }

    public function getName() : string
    {
        return 'Foo';
    }

    public function canRequireSQLConversion() : bool
    {
        return true;
    }

    public function convertToPHPValueSQL($sqlExpr, $platform) : string
    {
        return sprintf('DatabaseFunction(%s)', $sqlExpr);
    }
}

final class FooNewObject
{
    /** @var string */
    public $field;

    public function __construct(string $field)
    {
        $this->field = $field;
    }
}
```

#### Expected behavior

When using the new object expression, `convertToPHPValueSQL` is being called on types that are being used.

PR #8062, opened 2020-03-09
Convert PHP to SQL for new object expression

jeroenvdheuvel:
Fields used in the new object expression. Were not yet converted to via
their configured type/convertToPHPValueSQL. This can lead to incorrect
arguments being injected into the new object.

This is fixed by calling convertToPHPValueSQL while creating the SQL
query.

This fixes issue #8061

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -27,6 +27,7 @@ use Doctrine\ORM\OptimisticLockException;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\Utility\HierarchyDiscriminatorResolver;
 use Doctrine\ORM\Utility\PersisterHelper;
+use function trim;
 
 /**
  * The SqlWalker is a TreeWalker that walks over a DQL AST and constructs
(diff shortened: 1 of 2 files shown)


SenseException:
Thank you for your addition to the NEW object syntax. The implementation looks good, but I'm not sure if this belongs into the 2.7 branch. This looks more like an improvement than a bugfix. @doctrine/team-orm

ostrolucky:
Hard to say. I'm ok with both.

jeroenvdheuvel:
It might be both. I considered it a bug because for me this is unexpected behavior. `convertToPHPValueSQL` is called when hydrating an object or an array but not when creating a new object (the odd one out).

The [custom types mapping documentation](convertToPHPValueSQL) also doesn't mention anything specific about not supporting the new object syntax.

jeroenvdheuvel:
@SenseException or @ostrolucky what should I do to get this changed merged?

ostrolucky (approved):

SenseException (approved):
If 2.7 is the right branch for this, I'm :+1: with it.

jeroenvdheuvel:
Thanks @ostrolucky and @SenseException !

PR #8064, opened 2020-03-10
fix-abstract-query-embeddable

### LINKED ISSUE #8063: Processes an individual parameter value with embeddables
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.7.1

#### Summary

Queries with embeddable parameters has an error.
trying to access to a hardcoded undefined index 0.

message = "Notice: Undefined offset: 0"
line = {int} 730
file = "/app
Screen Shot 2020-03-10 at 13 42 26
/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php"
severity = {int} 8
*Exception*string = ""
code = {int} 0

**trace:**
  file = "/app/vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php"
  line = {int} 424
  function = "getSingleIdentifierValue"
  class = "Doctrine\ORM\UnitOfWork"
  type = "->"

#### Current behavior

Imposible to create a query with query builder using embeddable value objects.

#### Expected behavior

execute the query

amelendres:
fixes #8063 This fix to create a custom query with embeddable value objects

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -416,7 +416,7 @@ abstract class AbstractQuery
             return $value->name;
         }
 
-        if (! is_object($value)) {
+        if (is_object($value)) {
             return $value;
         }
 


SenseException:
Your change will break the further processing of `$value` for doing the opposite of the previous `if (! is_object($value)) {` and can't be approved. The code after this if-block is still expecting an object and therefore will run into errors.

Issue #7999, opened 2020-01-22
PHP 7.4 Typed Ids musts be nullable to remove an entity


leongersen:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7.0

#### Summary

This is closely related to #7940, which reports the UnitOfWork accessing properties before initialization. This was resolved in `doctrine/persistence` 1.3.6.

With that update, Ids must still be nullable to be able to remove the entity, as the `UnitOfWork` will be setting the id value back to null in `\Doctrine\ORM\UnitOfWork::executeDeletions`:

```php
// Entity with this $oid after deletion treated as NEW, even if the $oid
// is obtained by a new entity because the old one went out of scope.
//$this->entityStates[$oid] = self::STATE_NEW;
if ( ! $class->isIdentifierNatural()) {
	$class->reflFields[$class->identifier[0]]->setValue($entity, null);
}
```

Adding a non-nullable type to a property and trying to remove it results in the following error:

```
Typed property App\Entity\MyEntity::$id must be int, null used
```

I'm not sure this necessarily qualifies as a bug in `orm`, but I'm not sure what else to call it. A potential solution would be to check if the `reflField` allows null (`->getType()->allowsNull()`).

beberlei:
Checking for allows null seems like the sensible approach.

beberlei:
Fix should be in doctrine/persistence `TypedNoDefaultReflectionProperty` class with a new setValue overwrite that handles the special case of setting to NULL by unitializing with unset().

enumag:
@beberlei isn't this fixed by https://github.com/doctrine/reflection/releases/tag/1.2.0?

Alexandre-T:
@enumag I still have this bug, I use doctrine/reflection 1.2.1. The bug appears as soon as I added a ManyToMany relation between the same entity.

enumag:
@Alexandre-T What is the version of your doctrine/persistence? If it is 1.3.6+, can you post the exact error message and relevant part of stack trace? It seems your case is elsewhere than what was fixed.

Alexandre-T:
@enumag doctrine/persistence 1.3.7
I'm using a Symfony5 backend application with api-platform.
[...]

PR #8065, opened 2020-03-10
 Added check for non-nullable property when deleting a record

joaojacome:
Solves #7999

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1210,12 +1210,16 @@ class UnitOfWork implements PropertyChangedListener
             // Entity with this $oid after deletion treated as NEW, even if the $oid
             // is obtained by a new entity because the old one went out of scope.
             //$this->entityStates[$oid] = self::STATE_NEW;
-            if (! $class->isIdentifierNatural() &&
-                (
-                    ! method_exists($class->reflFields[$class->identifier[0]], 'getType') ||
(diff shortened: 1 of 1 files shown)


beberlei:
This needs to be fixed in doctrine/persistence instead. We already have a `TypedNoDefaultReflectionProperty` overwrite for `ReflectionProperty` there and it only needs an addition to `setValue()` to handle the `NULL` case as `unset()`. The problem is your unset() code here only works on public properties, it will cause an error on protected or private ones.

A workaround needs to be found to unset any kind of visibility property. The following is a prototype using closure and binding to this, but it doesn't seem to work: https://gist.github.com/beberlei/f28142a1b15f011bbcd0156c7aa64430

beberlei:
Have the right approach thanks to @lcobucci now, updated the gist.

prof-web-coding:
@joaojacome Will you update the code according to @beberlei 's input?

joaojacome:
> @joaojacome Will you update the code according to @beberlei 's input?

I could, but I think @beberlei fixed it here:
https://github.com/doctrine/reflection/issues/34

prof-web-coding:
Thanks. It does not work for me. I have a
`/**
     * @ORM\Id()
     * @ORM\GeneratedValue()
     * @ORM\Column(type="integer")
     */
    private int $id;` in an `Doctrine\ORM\Mapping\Entity` and I'm getting`Error: Cannot access private property App\Entity\User::$id`. It seems that the newly added setValue in [lib/Doctrine/Common/Reflection/TypedNoDefaultReflectionProperty.php](url) is not called, also not without #23b57f6

Issue #7006, opened 2018-01-24
Wrong commit order in some relation cases


contentrail:
Issue was found after updating to v.2.6 and associated with improvements in CommitOrderCalculator class.
Here is the example to reproduce.
Objects:

- **Book**
- **PCT** with mandatory relation to Book (ManyToOne).
Book has optional relation to the PCT object. Some Bookes has PCT and some not. But all PCT has relation to Book.
- **PCTFee** as a child object of PCT (ManyToOne).

![db_scheme](https://user-images.githubusercontent.com/18163201/35378065-1b7f00d8-01ba-11e8-818b-d8053ed3df34.PNG)

```
/**
 * @ORM\Entity
 * @ORM\Table(name="book")
 */
class Book
{
    /**
     * @var int
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="NONE")
     */
    private $id;

    /**
     * @var string
     * @ORM\Column(type="string", length=255, nullable=true)
     */
    private $exchangeCode;

    /**
     * @var PCT
     * @ORM\OneToOne(targetEntity="PCT", cascade={"persist", "remove"})
     * @ORM\JoinColumn(name="paymentCardTransactionId", referencedColumnName="id")
     */
    private $paymentCardTransaction;

    public function __construct(int $id)
    {
        $this->id = $id;
    }

    public function getId(): int
    {
        return $this->id;
    }

    public function getExchangeCode(): ?string
    {
        return $this->exchangeCode;
    }

    public function setExchangeCode(string $exchangeCode = null): void
    {
        $this->exchangeCode = $exchangeCode;
    }

    public function getPaymentCardTransaction(): ?PCT
    {
        return $this->paymentCardTransaction;
    }

    public function setPaymentCardTransaction(PCT $paymentCardTransaction = null): void
    {
        $this->paymentCardTransaction = $paymentCardTransaction;
    }
}
```

```
/**
 * @ORM\Entity
 * @ORM\Table(name="pct")
 */
class PCT
{
    /**
     * @var int
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="NONE")
     */
    private $id;

    /**
     * @var Book
     * @ORM\ManyToOne(targetEntity="Book")
     * @ORM\JoinColumn(name="bookingId", referencedColumnName="id", nullable=false)
     */
    private $book;

    /**
     * @var PCTFee[]
[...]

PR #8067, opened 2020-03-11
[GH-7006] Wrong commit order in some relation cases

Nitro-N:
Just reverted previous vertex checking method

Related to #7006


SenseException:
Can you please describe what this PR is supposed to accomplish in the commit order?

* Please add a test that is covering your fix
* Please don't break the other tests with your changes. There probably is a BC break in there.

PR #8068, opened 2020-03-12
Fix documentation of default generated value behavior

jorrit:
When reading about the default behavior of `@GeneratedValue` I noticed a mismatch in the basic-mapping docs. The introduction tells that PostgreSQL uses SERIAL by default, while later on it says sequence (which is the right one). This small PR fixes this.


SenseException (approved):
Yes, with the example above in the docs it should result in a sequence in PostgreSQL. Thank you for your fix.

PR #8074, opened 2020-03-18
Fix phpdoc

VincentLanglet:
When running phpstan, I discover that an error was coming from a wrong phpDoc in these files.

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -624,7 +624,7 @@ final class Query extends AbstractQuery
     /**
      * Sets the position of the first result to retrieve (the "offset").
      *
-     * @param integer $firstResult The first result to return.
+     * @param integer|null $firstResult The first result to return.
      *
      * @return Query This query object.
      */
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -624,7 +624,7 @@ final class Query extends AbstractQuery
     /**
      * Sets the position of the first result to retrieve (the "offset").
      *
-     * @param integer|null $firstResult The first result to return.
+     * @param int|null $firstResult The first result to return.
      *
      * @return Query This query object.
      */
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Query.php
```diff
@@ -624,7 +624,7 @@ public function contains($dql)
     /**
      * Sets the position of the first result to retrieve (the "offset").
      *
-     * @param integer $firstResult The first result to return.
+     * @param integer|null $firstResult The first result to return.
```
greg0ire:
```suggestion
     * @param int|null $firstResult The first result to return.
```

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -100,7 +100,7 @@ class QueryBuilder
     /**
      * The index of the first result to retrieve.
      *
-     * @var integer
+     * @var integer|null
```
greg0ire:
```suggestion
     * @var int|null
```

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -616,7 +616,7 @@ function (Query\Parameter $parameter) use ($key) : bool {
     /**
      * Sets the position of the first result to retrieve (the "offset").
      *
-     * @param integer $firstResult The first result to return.
+     * @param integer|null $firstResult The first result to return.
```
greg0ire:
```suggestion
     * @param int|null $firstResult The first result to return.
```

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -631,7 +631,7 @@ public function setFirstResult($firstResult)
      * Gets the position of the first result the query object was set to retrieve (the "offset").
      * Returns NULL if {@link setFirstResult} was not applied to this QueryBuilder.
      *
-     * @return integer The position of the first result.
+     * @return integer|null The position of the first result.
```
greg0ire:
```suggestion
     * @return int|null The position of the first result.
```

greg0ire (commented):
In case you are wondering what's up with phpcs, we have so many issues with cs on this project that checks are only enabled on PR diffs. That's why `integer` was ok before but isn't now.

VincentLanglet:
@greg0ire Thanks

greg0ire (approved):

greg0ire:
Thanks @VincentLanglet !

PR #8083, opened 2020-03-28
Undefined index on Hydrating Object fix

ghost:
Hi all, i m working on a project, the developer before me don't make the migration and i have a difference between the entities and the bd.
I got this errors : Undefined index ... and that crashed the script.
i fix it with an update of the condition on lib/Doctrine/ORM/UnitOfWork.php
at line 2842
the update would be : 
` if ($assoc['inversedBy'] && $assoc['type'] && ClassMetadata::ONE_TO_ONE  && isset($targetClass->associationMappings[$assoc['inversedBy']])) { .... }`
that for allowing to make the entities and skip the undefined fields on the db, that can be great to make a migrations from an old database to a new one (the new database is valid for the entity but not necessary the old one)

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2624,6 +2624,7 @@ class UnitOfWork implements PropertyChangedListener
      *
      * @return object The managed entity instance.
      *
+     * @throws ORMException
      * @todo Rename: getOrCreateEntity
      */
     public function createEntity($className, array $data, &$hints = [])
@@ -2847,9 +2848,11 @@ class UnitOfWork implements PropertyChangedListener
 
                     $this->originalEntityData[$oid][$field] = $newValue;
                     $class->reflFields[$field]->setValue($entity, $newValue);
+                    if ( ! isset($targetClass->associationMappings[$assoc['inversedBy']])) {
+                        throw new ORMException("The column " . $assoc['inversedBy'] . " is needed by " . $targetClass->getName() . " entity and should be are on the database");
+                    }
 
-		    if ($assoc['inversedBy'] && $assoc['type'] && ClassMetadata::ONE_TO_ONE  &&
-			    isset($targetClass->associationMappings[$assoc['inversedBy']])) {
+                    if ($assoc['inversedBy'] && $assoc['type'] & ClassMetadata::ONE_TO_ONE) {
 
                         $inverseAssoc = $targetClass->associationMappings[$assoc['inversedBy']];
                         $targetClass->reflFields[$inverseAssoc['fieldName']]->setValue($newValue, $entity);
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2848,7 +2848,7 @@ class UnitOfWork implements PropertyChangedListener
                     $this->originalEntityData[$oid][$field] = $newValue;
                     $class->reflFields[$field]->setValue($entity, $newValue);
 
-                    if ($assoc['inversedBy'] && $assoc['type'] & ClassMetadata::ONE_TO_ONE && is_string($targetClass->associationMappings[$assoc['inversedBy']] )) {
+                    if ($assoc['inversedBy'] && $assoc['type'] & ClassMetadata::ONE_TO_ONE && array_key_exists($assoc['inversedBy'], $targetClass->associationMappings)) {
                         $inverseAssoc = $targetClass->associationMappings[$assoc['inversedBy']];
                         $targetClass->reflFields[$inverseAssoc['fieldName']]->setValue($newValue, $entity);
                     }
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -48,6 +48,7 @@ use Throwable;
 use UnexpectedValueException;
 use function get_class;
 use function is_object;
+use function array_key_exists;
 
 /**
  * The UnitOfWork is responsible for tracking changes to objects during an
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2709,6 +2709,10 @@ class UnitOfWork implements PropertyChangedListener
             return $entity;
         }
 
+        if ($class instanceof ClassMetadataInfo) {
+            return $entity;
+        }
+
         foreach ($class->associationMappings as $field => $assoc) {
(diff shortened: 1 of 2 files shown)


greg0ire:
> Hi all, i m working on a project, the developer before me don't make the migration and i have a difference between the entities and the bd.

What migration?

> I got this errors : Undefined index ... and that crashed the script.

What script? A migration script?


> that for allowing to make the entities

You mean hydrate them, right?

> and skip the undefined fields on the db

So there are fields defined in your entity that map to columns that do not exist in your db, is that it? If yes then maybe a good solution for your issue would be to have one mapping for the old db, and one mapping for the new one, that way you don't map to columns that don't exist in the mapping for the old db.

LeGouffre:
i m agree but the problem steel for me, if i mapped  the db ( i did it ) the relation 1-n or n-n is not taken in consideration. and for the errors from continuous-integration/travis-ci/pr, i understand the phpsc and i wil fix it for  #10367.1 test i dont understand for me the logs says  got and errors. i dont think, can you explain me this error : Error: Call to undefined method Doctrine\Tests\Models\CMS\CmsArticle::isInitialized()

greg0ire:
> i m agree but the problem steel for me, if i mapped the db ( i did it ) the relation 1-n or n-n is not taken in consideration.

You mapped it once, I'm telling you should probably map it twice, but I'm not even sure that is possible, I don't know if an entity can be mapped in several EMs. Probably not a great idea after all :thinking: 

What about my suggestion to use pure SQL?

LeGouffre:
that is difficulte because the database is very big (2Gb), with relation in all directions, so pure SQL, is can take so much time.

LeGouffre:
But i dont understand why my idea to check if is column is set on the database (i speak about doctrine 2.8.x) is bad, for me that add stability to doctrine.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2848,7 +2848,9 @@ public function createEntity($className, array $data, &$hints = [])
                     $this->originalEntityData[$oid][$field] = $newValue;
                     $class->reflFields[$field]->setValue($entity, $newValue);
 
-                    if ($assoc['inversedBy'] && $assoc['type'] & ClassMetadata::ONE_TO_ONE) {
+		    if ($assoc['inversedBy'] && $assoc['type'] && ClassMetadata::ONE_TO_ONE  &&
```
greg0ire: Changing `&&` to `&` is introducing a big bug.

greg0ire:
To me, stability does not mean sweeping things under the rug. If anything, I would throw a more helpful exception explaining to the developer that we expect the field to be set.

greg0ire:
> that is difficulte because the database is very big (2Gb), with relation in all directions, so pure SQL, is can take so much time.The new database is not the same as the old one ( thank to the previous dev on my project)

Actually, pure SQL is probably the fastest way to import a big dataset. Also, 2 Gb is not big at all.

LeGouffre:
Agree  with u a throw is more readable, but that isnt throw but a crash that is the point, i  can recode the "if" with the throw exception. and make an other pull request. I will use my forked for my project its more simple for me to use like this, thank u for u explanation

greg0ire:
> make an other pull request.

Or you can force push to this branch

LeGouffre:
yeah i see it after my message (its my first pull request) :)

LeGouffre:
just a last question about the throwing error and the status code send, who exit status will be used to passed the travis check ?

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2847,8 +2848,12 @@ public function createEntity($className, array $data, &$hints = [])
 
                     $this->originalEntityData[$oid][$field] = $newValue;
                     $class->reflFields[$field]->setValue($entity, $newValue);
+                    if ( ! $assoc['inversedBy'] && ! $assoc['type'] && ! isset($targetClass->associationMappings[$assoc['inversedBy']])) {
+                        throw new ORMException("The column " . $assoc['inversedBy'] . " is needed by " . $targetClass->getName() . " entity and should be are on the database", 1);
```
greg0ire: There should be unit tests for this piece of code, but did you at least test it manually? (Beware, it's a trick question)

LeGouffre: yeah i test it manually it work with the good error message but travis test didn't expect a throw error, then i add the two negation for target more precisely the error. Then i think the code error should be a great idea

greg0ire: Also, I think this should be moved inside the if below, and you should only check for `isset($targetClass->associationMappings[$assoc['inversedBy']])`

greg0ire:
> just a last question about the throwing error and the status code send, who exit status will be used to passed the travis check ?

I don't understand the question.

LeGouffre:
ho sorry, its when i test it with travis, he make an error because he didnt expect a throw error and i cant passed the unit test. And i was think to add a code its a good idea (but i don't know if this code is an exit status for the script)

lib/Doctrine/ORM/UnitOfWork.php
LeGouffre: are u french ? (in this case will be more practical to speak in french cae me too)

greg0ire: Sorry, we can't, this needs to be understandable for everyone. If you're going to contribute to open source, using english is a must, so let's start as soon as possible, shall we?

>  did i something wrong ?

Well `ORMException` is an interface, for starters, so you can't throw it. Also, checking `! $assoc['type']` has no meaning, while the original expression `$assoc['type'] & ClassMetadata::ONE_TO_ONE` you copied it from has some. I suggest you read about the `&` operator to understand what it does. https://www.php.net/manual/en/language.operators.bitwise.php

LeGouffre: Ok i forgot binary operator ( i don't really use it except in a game ), for the throw, i see how  to do it like add a method in the ORMExeption. to throw the exeption with a static method like line 897 UnitOfWork.php

LeGouffre:
Hello, some one could tell me why i got new phpcs errors on the pull request (update ?, or any think)  form unchanged files ?

beberlei:
This was never needed before by anyone, it looks like its a check because of invalid mappings. You describe as much with the missing migration. Doctrine cannot really hydrate "unexistant columns", because it puts every column into the SELECT satement, so the query would fail on itself before. There must be something else wrong.

Can you open a new PR with a fialing test case or with your mappings, and query? We need much more details. I close this PR for now, because its not mergable without understanding the problem.

PR #8086, opened 2020-03-29
setParameters in docs example should use new syntax

vladysta:
Since there is only compatibility with 2.3, we should point in docs how to add params with `setParameters` in the new way.

### DIFF
--- a/docs/en/reference/dql-doctrine-query-language.rst
+++ b/docs/en/reference/dql-doctrine-query-language.rst
@@ -319,11 +319,11 @@ With Nested Conditions in WHERE Clause:
 
     <?php
     $query = $em->createQuery('SELECT u FROM ForumUser u WHERE (u.username = :name OR u.username = :name2) AND u.id = :id');
-    $query->setParameters(array(
-        'name' => 'Bob',
-        'name2' => 'Alice',
-        'id' => 321,
-    ));
+    $query->setParameters(new \Doctrine\Common\Collections\ArrayCollection([
+        new \Doctrine\ORM\Query\Parameter('name', 'Bob'),
+        new \Doctrine\ORM\Query\Parameter('name2', 'Alice'),
+        new \Doctrine\ORM\Query\Parameter('id', '321'),
+    ]));
     $users = $query->getResult(); // array of ForumUser objects
 
 With COUNT DISTINCT:
@@ -1299,8 +1299,8 @@ pass parameters to the query the following methods can be used:
 
 -  ``AbstractQuery::setParameter($param, $value)`` - Set the
    numerical or named wildcard to the given value.
--  ``AbstractQuery::setParameters(array $params)`` - Set an array
-   of parameter key-value pairs.
+-  ``AbstractQuery::setParameters(ArrayCollection $params)`` - Set a ``\Doctrine\Common\Collections\ArrayCollection``
+   of ``\Doctrine\ORM\Query\Parameter``s.
 -  ``AbstractQuery::getParameter($param)``
 -  ``AbstractQuery::getParameters()``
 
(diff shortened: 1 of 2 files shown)


greg0ire:
? This looks a lot like #8085, doesn't it?

greg0ire:
Also, I'd like to have other contributors chime in about [this](https://github.com/doctrine/orm/pull/7994#pullrequestreview-345470061)

vladyslavstartsev:
@greg0ire yes, it looks the same as  #8085
But I've messed up there. Sorry for the double notification

greg0ire:
> Sorry for the double notification

Triple, you mean, there's also #8084 :wink: 

next time, please use a force push / the edit button on your PR

vladyslavstartsev:
how can I request review? so I can get decline (with review fixes) or approve this change

greg0ire:
You're the author, why would you decline or approve this change? Also, please rebase, as you see, new checks have been added to the repository, this can't be merged unless they are run against your PR.

vladyslavstartsev:
@greg0ire sorry, I meant get approve from reviewers

greg0ire:
I will request reviews after the build is green.

vladyslavstartsev:
done

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -1299,7 +1299,7 @@ pass parameters to the query the following methods can be used:
 
 -  ``AbstractQuery::setParameter($param, $value)`` - Set the
    numerical or named wildcard to the given value.
--  ``AbstractQuery::setParameters(array $params)`` - Set an array
+-  ``AbstractQuery::setParameters(ArrayCollection $params)`` - Set an array
    of parameter key-value pairs.
```
ostrolucky, line 1303: Description is misleading now, not sure if it's needed now. It's either no description, or something along the lines of "Set a collection of `\Doctrine\ORM\Query\Parameter`". But that's just repeating what's supposed to be obvious from method docblock.

vladyslavstartsev: I did "Set a `\Doctrine\Common\Collections\ArrayCollection` of `\Doctrine\ORM\Query\Parameter`s."

is that what you want?

ostrolucky (approved):

SenseException (approved):

greg0ire (changes requested):
I'd really like to have other contributors chime in about [this](https://github.com/doctrine/orm/pull/7994#pullrequestreview-345470061). Your changes reflect what seems to be the plan, devised 8 years ago, and the new syntax seems super cumbersome to me (is it really just to me?). Is the BC layer even removed on master?

ostrolucky:
Well I'm with you, but opening this topic will possibly create another long discussion, meanwhile having this PR stuck here. It's not a fault of OP we don't have clear direction in core. Perhaps let's try opening in it Slack and if we see it doesn't go anywhere, let's merge this. Sounds good?

greg0ire:
> It's not a fault of OP we don't have clear direction in core

Absolutely not. We have to have this discussion and be clear with either yes, we will do it, or no, this API looks weird.

Note that I just had a look at master, and the BC layer is still here after all these years: https://github.com/doctrine/orm/blob/72bc09926df1ff71697f4cc2e478cf52f0aa30d8/lib/Doctrine/ORM/AbstractQuery.php#L321-L337

ostrolucky:
Consensus in our internal discussion was that requiring ArrayCollection doesn't make sense and we should go in future with array or iterable of Parameter[] instead. Hence we cannot merge this PR and promote passing ArrayCollection further. Let's close.

PR #8087, opened 2020-03-29
Updated CONCAT to support SimpleArithmeticExpression

oleg-andreyev:
While working on a project noticed that `CONCAT` does not support any expressions at the moment it only supports`StringPrimary`

P.S. Decided not to rename/remove `$firstStringPrimary` and `$secondStringPrimary` because they are public... while they seem as useless fields.

My CONCAT example
```dql
    CONCAT(
        i.price
        *
        IFELSE(
          i.priceOrigin > 0.01,
          1,
          IFELSE(
            dd.value > 0 AND dd.value != 1,
            dd.value,
            1
          
        )
      ),
      ' EUR'
    ) AS price,
```

### DIFF
--- a/lib/Doctrine/ORM/Query/AST/Functions/ConcatFunction.php
+++ b/lib/Doctrine/ORM/Query/AST/Functions/ConcatFunction.php
@@ -22,7 +22,7 @@ namespace Doctrine\ORM\Query\AST\Functions;
 use Doctrine\ORM\Query\Lexer;
 
 /**
- * "CONCAT" "(" StringPrimary "," StringPrimary {"," StringPrimary }* ")"
+ * "CONCAT" "(" SimpleArithmeticExpression "," SimpleArithmeticExpression {"," SimpleArithmeticExpression }* ")"
  *
  *
  * @link    www.doctrine-project.org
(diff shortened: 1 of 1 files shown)


oleg-andreyev:
Failed stages are not related to this PR

greg0ire:
I'm restarting the failing jobs

lib/Doctrine/ORM/Query/AST/Functions/ConcatFunction.php
```diff
@@ -22,7 +22,7 @@
 use Doctrine\ORM\Query\Lexer;
 
 /**
- * "CONCAT" "(" StringPrimary "," StringPrimary {"," StringPrimary }* ")"
+ * "CONCAT" "(" SimpleArithmeticExpression "," SimpleArithmeticExpression {"," SimpleArithmeticExpression }* ")"
```
greg0ire: From https://www.postgresql.org/docs/9.1/functions-string.html:

> Note: Before PostgreSQL 8.3, these functions would silently accept values of several non-string data types as well, due to the presence of implicit coercions from those data types to text. Those coercions have been removed because they frequently caused surprising behaviors. However, the string concatenation operator (||) still accepts non-string input, so long as at least one input is of a string type, as shown in Table 9-6. For other cases, insert an explicit coercion to text if you need to duplicate the previous behavior.

`CONCAT` translates to `||` for PostgreSQL: https://github.com/doctrine/dbal/blob/ff1501a12ae604dd3fc910713a3b9c25d30663db/src/Platforms/AbstractPlatform.php#L754

I think this might be too permissive: `1 || 1` will lead to a crash on postgreSQL I'm afraid.

oleg-andreyev: Hmm. Interesting. I'll take a look at it

greg0ire: Thanks! Not sure if it should be a blocker though… maybe we already allow similar cases: if the crash you get from postgreSQL is easy to understand, maybe that's ok. Or maybe we use the lowest common denominator and allow only things that work everywhere with every allowed value, I don't know. I hope another maintainer can chime in on this.

SenseException: `||` was also mentioned in doctrine/dbal#3887. A switch to `CONCAT` can happen in the next major DBAL release and for the next ORM release that is using that DBAL version.

PR #8095, opened 2020-04-06
Make _parseOptions function protected

griniv:
It would be nice to be able to override _parseOptions function and add some custom features to it (for example, nullable options support)

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -672,7 +672,7 @@ class XmlDriver extends FileDriver
      *
      * @return array The options array.
      */
-    private function _parseOptions(SimpleXMLElement $options)
+    protected function _parseOptions(SimpleXMLElement $options)
     {
         $array = [];
 


greg0ire:
Why not contribute nullable options support directly to Doctrine? Not sure what it means BTW.

griniv:
Basically I've suggested a solution that is way more flexible (and simple!) than just implementing one missing feature I've stumbled upon.

Also its kinda hard to explain nullable options use. Here's a mapping example: 
```
      <unique-constraint name="example" columns="foo,bar,baz">
        <options>
          <option name="lengths">
            <option />
            <option />
            <option>200</option>
          </option>
        </options>
      </unique-constraint>
```
Right now it is impossible to define third column's index length because first two are parsed into empty strings, not nulls.

greg0ire:
Oh right. I understand better.

> Basically I've suggested a solution that is way more flexible (and simple!) than just implementing one missing feature I've stumbled upon.

It's more flexible for everyone, because people can implement whatever they want, but it's definitely less simple when compared to _having absolutely nothing to do_ to get that missing feature in the case you contribute it.

PR #8098, opened 2020-04-09
Update andx annotation

### LINKED ISSUE #8096: PHPStan warning
### Bug Report

|    Q         |   A
|------------ | ------
| Version  | 2.7.1

#### Summary
When im run phpstan analyser with 6 level i get error:
*Parameter #1 $x of method Doctrine\ORM\Query\Expr::andX() expects Doctrine\ORM\Query\Expr\Comparison|Doctrine\ORM\Query\Expr\Func|Doctrine\ORM\Query\Expr\Orx|string|null, Doctrine\ORM\Query\Expr\Andx given.*

#### Current behavior

https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/Query/Expr.php#L31

#### How to reproduce

Run ` vendor/bin/phpstan analyse -c phpstan.neon  -l 6`

#### Expected behavior

No warnings

Just replace annotation on mixed?

kirya-dev:


greg0ire:
Can you tell me why you choosed to ignore [my comment](https://github.com/doctrine/orm/issues/8096#issuecomment-611089876)?

kirya-dev:
In latest version of branch 2.7 here is `Andx::class`, but not `self::class`
https://github.com/doctrine/orm/blob/2.7/lib/Doctrine/ORM/Query/Expr/Andx.php#L41-L46

greg0ire:
Isn't it the same thing?

kirya-dev:
True

lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -41,8 +41,8 @@ class Expr
      *     // (u.type = ?1) AND (u.role = ?2)
      *     $expr->andX($expr->eq('u.type', ':1'), $expr->eq('u.role', ':2'));
      *
-     * @param Expr\Comparison|Expr\Func|Expr\Orx|string $x Optional clause. Defaults to null, but requires at least one
-     *                                                     defined when converting to string.
+     * @param mixed $x Optional clause. Defaults to null, but requires
```
greg0ire, line 44:
```suggestion
     * @param Expr\Comparison|Expr\Func|Expr\Orx|Expr\Andx|string $x Optional clause. Defaults to null, but requires
```

kirya-dev: hmm. other functuions in this class has `mixed` param annotaion. Maybe there must be equal `mixed`?

greg0ire: If there is a finite number of types that can be accepted here, I'd rather document them. This will help SA tools spot bugs.

kirya-dev: Ok, im do this later

PR #8105, opened 2020-04-14
Update AndX and OrX methods to use variadic parameter notation

renorram:
Use variadic parameter notation to update the method's signature but without causing BC. Also, update the PHPDOC of the methods.

### DIFF
--- a/lib/Doctrine/ORM/Query/Expr.php
+++ b/lib/Doctrine/ORM/Query/Expr.php
@@ -28,7 +28,7 @@ class Expr
      *     // (u.type = ?1) AND (u.role = ?2)
      *     $expr->andX($expr->eq('u.type', ':1'), $expr->eq('u.role', ':2'));
      *
-     * @param Expr\Comparison|Expr\Func|Expr\Orx|string ...$clauses Optional clauses. Defaults to an empty array, but requires at least one
+     * @param Expr\Comparison|Expr\Func|Expr\Orx|Expr\Andx|string ...$clauses Optional clauses. Defaults to an empty array, but requires at least one
      *                                                     defined when converting to string.
      *
      * @return Expr\Andx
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -28,14 +28,14 @@ class Expr
...
      *     $expr->andX($expr->eq('u.type', ':1'), $expr->eq('u.role', ':2'));
      *
-     * @param Expr\Comparison|Expr\Func|Expr\Orx|string $x Optional clause. Defaults to null, but requires at least one
+     * @param Expr\Comparison|Expr\Func|Expr\Orx|string ...$clauses Optional clauses. Defaults to an empty array, but requires at least one
      *                                                     defined when converting to string.
      *
      * @return Expr\Andx
      */
-    public function andX($x = null)
+    public function andX(...$clauses)
```
greg0ire, line 36: This change is a BC-break for extending classes: https://3v4l.org/IkqbQ, and this class is not `final`

lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -47,14 +47,14 @@ public function andX($x = null)
      *     // (u.type = ?1) OR (u.role = ?2)
      *     $q->where($q->expr()->orX('u.type = ?1', 'u.role = ?2'));
      *
-     * @param mixed $x Optional clause. Defaults to null, but requires
+     * @param Expr\Comparison|Expr\Func|Expr\Orx|string ...$clauses Optional clauses. Defaults to an empty array, but requires
```
greg0ire, line 50: Please also add `Andx` to the list, see https://github.com/doctrine/orm/pull/8098/files

renorram: Done! Added to orX method too.

greg0ire:
Not sure what our policy is exactly but I noticed that BC-breaks are documented here: https://github.com/doctrine/orm/blob/master/UPGRADE.md

renorram:
@greg0ire does this really classify as BC? because variadic parameters are available since 5.6  and the doctrine 3 is already targeted to 7.2  and 2.7 is targeted 7.1  so I think this change wouldn't cause any BC and the methods' invocations syntax don't change since you can still call with or without parameters.

greg0ire:
I thought you were targeting because you spotted the BC-break, which is the following:https://3v4l.org/tZChB

renorram:
> I thought you were targeting because you spotted the BC-break, which is the following: https://3v4l.org/tZChB

Oh got it, thanks for the heads up, didn't thought about that, I'll add the BC ASAP, any tips on how should I describe this will be welcomed

greg0ire:
# BC Break: `Expr::andX()` and `Expr::orX()` now use variadics

If you extend `Expr` and override these methods, you have to use the new signature.

SenseException:
Looks good to me, but we need a rebase after the build of master is fixed.

PR #8107, opened 2020-04-15
Fix QueryBuilder::getParameter() on parameter names with colons

### LINKED ISSUE #8106: QueryBuilder::setParameter() breaks if parameter name with ':' is set multiple times
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | technically yes, broke in 2.3.0 commit [1635e0af4](https://github.com/doctrine/orm/commit/1635e0af4b06ef3015205563b59b505ae3fac69d#diff-6e1fe274092a9dc90ad2f6728df45c6aL358)
| Version     | 2.3.0, v2.7.2, 2.8.x @ 3c91792dd, master @ 8c259ea5c

#### Summary

`QueryBuilder::setParameter()` breaks if a parameter name that starts with `:` is set multiple times.
(Parameter names that don't start with a colon can be set multiple times without issues, and normally it is allowed to prefix a parameter name with a colon.)

#### Current behavior

An exception is thrown when the query is executed:
> Doctrine\ORM\Query\QueryException: Too many parameters

#### How to reproduce

```php
// without colon (succeeds)
$qb = $em->createQueryBuilder();
$qb
    ->select('u')
    ->from('User', 'u')
    ->where('u.id = :id')
    ->setParameter('id', 1)  // <--
    ->setParameter('id', 1); // <--
dump($qb->getQuery()->getResult());
// success
```

```php
// with colon (throws exception)
$qb = $em->createQueryBuilder();
$qb
    ->select('u')
    ->from('User', 'u')
    ->where('u.id = :id')
    ->setParameter(':id', 1)  // <--
    ->setParameter(':id', 1); // <--
dump($qb->getQuery()->getResult());
// Doctrine\ORM\Query\QueryException: Too many parameters: the query defines 1 parameters and you bound 2 in [...]
```

#### Expected behavior

The second snippet (with colon) should succeed and produce the same result as the first snippet.

#### Cause

`setParameter($key, ...)` checks if $key already exists using `getParameter($key)`, which compares $key to the names of the existing Query\Parameter objects:
https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/QueryBuilder.php#L606

When $key starts with a colon, the comparison always fails because the names of the existing Query\Parameter objects are trimmed of colons in the constructor:
https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/Query/Parameter.php#L49

#### Fix

`getParameter()` should trim colons from $key to match the behaviour of the Query\Parameter constructor.
To avoid code duplication, the existing trimming code could be moved to a separate function such as `Query\Parameter::normalizeName()`.
(Another approach is for `getParameter()` create a dummy instance of Query\Parameter to normalize $key, but that's ugly and requires a dummy value and type.)

#### Pull request

#8107

#### Background / original use case

<details>

When building complex queries with conditional parts, it can be useful to call `setParameter()` multiple times with the same parameter.
Here is a contrived example that lists users, optionally filtered by a pattern. It can be configured to search for users by first name, last name, or both:

```php
$qb = $em->createQueryBuilder();
$qb
    ->select('u')
    ->from('User', 'u');
if ($searchFirstName) {
    $qb
        ->orWhere('u.first_name LIKE :pattern')
        ->setParameter('pattern', $pattern);
}
if ($searchLastName) {
    $qb
        ->orWhere('u.last_name LIKE :pattern')
        ->setParameter('pattern', $pattern);
}
```

Note that the `setParameter()` calls cannot be combined and moved outside the `if` statements, because then when both conditions are false the parameter would be unused, resulting in the following error:
```
Doctrine\ORM\Query\QueryException: Too many parameters: the query defines 0 parameters and you bound 1 in [...]
```

The [QueryBuilder documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/query-builder.html) doesn't use leading colons in parameter names passed to `setParameter()`, but they are supported and are popular in some projects (maybe the practice has been adopted from [PDO](https://www.php.net/manual/en/pdostatement.bindparam.php)).

</details>

tom93:
Fixes #8106.

This change can be backported to the stable release (2.7), but will cause a small merge conflict because it builds on commit [17e7c2a42](https://github.com/doctrine/orm/commit/17e7c2a42e0ee49781e60d6facdd2efc9e1f86b3#diff-01ac4deedc780ada53097bd3bd8d5be2) which hasn't been backported.
*Update: has been backported from master to 2.7 as requested.*

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -321,11 +321,13 @@ abstract class AbstractQuery
      */
     public function getParameter($key)
     {
+        $key = Query\Parameter::normalizeName($key);
+
         $filteredParameters = $this->parameters->filter(
             function (Query\Parameter $parameter) use ($key) : bool {
                 $parameterName = $parameter->getName();
 
-                return $key === $parameterName || (string) $key === (string) $parameterName;
+                return $key === $parameterName;
             }
         );
 
--- a/lib/Doctrine/ORM/Query/Parameter.php
+++ b/lib/Doctrine/ORM/Query/Parameter.php
@@ -31,6 +31,18 @@ use function trim;
 class Parameter
 {
     /**
+     * Returns the internal representation of a parameter name.
+     *
+     * @param string|int $name The parameter name or position.
+     *
+     * @return string The normalized parameter name.
+     */
+    public static function normalizeName($name)
+    {
+        return trim((string) $name, ':');
+    }
+
+    /**
      * The parameter name.
      *
      * @var string
@@ -67,7 +79,7 @@ class Parameter
      */
     public function __construct($name, $value, $type = null)
     {
-        $this->name          = trim((string) $name, ':');
+        $this->name          = self::normalizeName($name);
         $this->typeSpecified = $type !== null;
 
         $this->setValue($value, $type);
--- a/lib/Doctrine/ORM/Query/Parameter.php
+++ b/lib/Doctrine/ORM/Query/Parameter.php
@@ -67,7 +67,7 @@ class Parameter
      */
     public function __construct($name, $value, $type = null)
     {
-        $this->name          = trim($name, ':');
+        $this->name          = trim((string) $name, ':');
         $this->typeSpecified = $type !== null;
 
         $this->setValue($value, $type);
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -602,11 +602,13 @@ class QueryBuilder
      */
     public function getParameter($key)
     {
+        $key = Query\Parameter::normalizeName($key);
+
         $filteredParameters = $this->parameters->filter(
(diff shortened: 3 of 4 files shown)


SenseException:
Hi. You should target 2.7 instead of master for your PR. Please also make sure that the build doesn't break and is green.

tom93:
@SenseException Done.
I added a separate commit to which partially backports the relevant type cast from commit 17e7c2a42. Hopefully this will make it easier to merge/cherry-pick my change into 'master' later on.
(By the way, the broken build on 'master' is not related to my change, e.g. https://travis-ci.org/github/doctrine/orm/builds/675106837)

beberlei (changes requested):
Can you move the test to `QueryTest` instead please? Only testing for `Query::getParameter()` is already enough as well, the test case covers too much lines for the simple thing it asserts.

beberlei:
Simplified test in 181114f2c

PR #8109, opened 2020-04-16
Add alert to "avoid persisting detached entities"

Pictor13:
The alert is hidden into the code

https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/EntityManager.php#L619-L620

while it's useful to be mentioned in the documentation, the same way it is done for the [documentation from Doctrine ODM](https://github.com/doctrine/mongodb-odm/blob/2.0.x/docs/en/reference/working-with-objects.rst#persisting-documents), as caution alert at the end of the section.

### DIFF
--- a/docs/en/reference/working-with-objects.rst
+++ b/docs/en/reference/working-with-objects.rst
@@ -250,6 +250,12 @@ as follows:
 -  If X is a detached entity, an exception will be thrown on
    flush.
 
+.. caution::
+
+    Do not pass detached entities to the persist operation. The persist operation always
+    considers entities that are not yet known to the ``EntityManager`` as new entities
+    (refer to the ``STATE_NEW`` constant inside the ``UnitOfWork``).
+
 Removing entities
 -----------------
 


docs/en/reference/working-with-objects.rst
```diff
@@ -250,6 +250,10 @@ as follows:
 -  If X is a detached entity, an exception will be thrown on
    flush.
 
+.. caution::
+
+    Do not pass detached entities to the persist operation.
```
SenseException, line 255: Thank you for the addition to the docs. You actually could use the same sentence for the caution-block.

```suggestion
    Do not pass detached entities to the persist operation. The persist operation always
    considers entities that are not yet known to this EntityManager as NEW.
```

SenseException (changes requested):

docs/en/reference/working-with-objects.rst
Pictor13: Right, it's more clear and spares from having to go read the comment in-code 👍

docs/en/reference/working-with-objects.rst
```diff
@@ -250,6 +250,11 @@ as follows:
 -  If X is a detached entity, an exception will be thrown on
    flush.
 
+.. caution::
+
+    Do not pass detached entities to the persist operation. The persist operation always
+    considers entities that are not yet known to this EntityManager as NEW.
```
greg0ire, line 256: I don't think NEW was capitalized for emphasis, but because it is a reference to this:

https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/UnitOfWork.php#L87

Let's make that clearer by using code markup:
```suggestion
    considers entities that are not yet known to this EntityManager as ``NEW``.
```

SenseException: Yepp, that makes sense.

Pictor13: Yes I agree.

Am wondering if there are policies about linking the code in the documentation? Or is it to be avoided?

Would it makes sense to link ``NEW`` to the [2.7:UnitOfWork.php#L76](https://github.com/doctrine/orm/blob/2.7/lib/Doctrine/ORM/UnitOfWork.php#L76), to give more context?
My only concern is about eventual commits moving the code to another line and breaking the link.

In that last case the link is to be avoided and maybe we could add details regarding that ``NEW`` word, thrown out without much context.

> Do not pass detached entities to the persist operation. The persist operation always considers entities that are not yet known to the EntityManager as ``NEW`` (refer to the available states  inside ``UnitOfWork``).

or keep consistency with the rest of the documentation and be more discorsive 

> Do not pass detached entities to the persist operation. The persist operation always considers entities that are not yet known to the ``EntityManager`` as new entities (refer to the ``STATE_NEW``  inside the ``UnitOfWork``).


Also note changing to *the* rather than *this EntityManager*, since we are not inside the class.

greg0ire: A quick grep shows no links to the code, probably for the reason you stating, although using permalink might mitigate that. I like the second version you propose better.

Pictor13: I though about permalinks but that would point a specific commit, while the (impossible) ideal would be to point the branch.

I'll go for the second version then; I like it more too, and it gives enough hints on how to find the correct line.

docs/en/reference/working-with-objects.rst
```diff
@@ -250,6 +250,12 @@ as follows:
 -  If X is a detached entity, an exception will be thrown on
    flush.
 
+.. caution::
+
+    Do not pass detached entities to the persist operation. The persist operation always
+    considers entities that are not yet known to the ``EntityManager`` as new entities
+    (refer to the ``STATE_NEW`` inside the ``UnitOfWork``).
```
greg0ire, line 257:
```suggestion
    (refer to the ``STATE_NEW`` constant inside the ``UnitOfWork``).
```

greg0ire (dismissed):

greg0ire:

How to do that?

1. `git rebase -i origin/2.7`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

greg0ire (approved):

Pictor13:
Thank you for the instructions! (even though I usually don't like to force push on PRs 😅)

SenseException (approved):

PR #8110, opened 2020-04-16
ExtraLazy OneToMay associations indexed by metacolumn fails

ajgarlag:
When you call `$collection->containsKey($key)` on a collection defined by a OneToMany extra lazy association. the `OneToManyPersister` fails to generate the correct criteria to execute the count statement.


ajgarlag:
Closed in favor of #8111

PR #8114, opened 2020-04-17
Warn users about performance issues with DateTime objects in setParameter()

maks-rafalko:
The full explanation of performance impact is explained in this bug report: https://github.com/doctrine/orm/issues/8113

Here is an attempt to warn developers about this potential issue, and to show how the third parameter can be used to avoid performance issue with DateTime* instances

PR #8115, opened 2020-04-18
Do not try loading ClassMetadata for DateTimeInterface parameters

### LINKED ISSUE #8113: Massive performance issue with setParameter() and DateTime* objects
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 2.7.2

Related doc PR to warn users: https://github.com/doctrine/orm/pull/8114

#### Summary

Using `$qb->setParameter('date', new \DateTimeImmutable())` without specifying 3rd argument (the type) leads to massive performance issues in long running processes.

This is a follow up for https://github.com/doctrine/orm/issues/7527 (originally posted here https://github.com/doctrine/orm/pull/7471#discussion_r241949045)

#### Current behavior

As it's stated in the official docs ([binding parameters to query](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/query-builder.html#binding-parameters-to-your-query))

> Calling `setParameter()` automatically infers which type you are setting as value. This works for integers, arrays of strings/integers, **DateTime instances** and for managed entities.

So you can just pass a datetime object and it will work:

```php
$qb->setParameter('date', new \DateTimeImmutable())
```

However, this leads to a) search this metadata in a cache, it fails and then b) load class metadata for every parameter of date types (e.g. `\DateTimeImmutable`), and such metadata does not exist of course. 

### Performance impact

We have a consumer that continuously reads data from the queue and executes queries, passing datetime object to `setParameter()`.

Consumer is able to handle ~250 messages per secong, producing thousands of queries per second when it is fresh / is just started.

However, after some time there is a massive performance degradation (from 250/s, 230/s, ... 100/s, ... and so on). And this is exactly because of all the code that is executed just to load metadata for `\DateTimeImmutable` class which never exist).

So there are thousands of cache misses, thrown but caught exceptions, attempts to load metadata.

After changing this

```php
$qb->setParameter('date', new \DateTimeImmutable())
```

to this

```php
use Doctrine\DBAL\Types\Types;

$qb->setParameter('date', new \DateTimeImmutable(), Types::DATE_IMMUTABLE)
```

we are no longer experiencing performance issues. After 30 minutes of working, consumer still handles ~250 records per second without degradation.

#### How to reproduce

Just create any entity with `datetime` / `datetime_immutable` mapped property. 

```php
$em->getRepository(User::class)->createQueryBuilder('u');

$qb->select('u')
        ->where('u.createdAt > :now')
        ->setParameter(
            'now',
            new \DateTimeImmutable()
            // ,Types::DATE_IMMUTABLE uncomment to fix performance issue
        )
        ->getQuery()
        ->getResult();
```

#### Expected behavior

Expected behavior would be to not try loading metadata for datetime objects, but return the type immediately somewhere here:

https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/Query.php#L411-L413

*before trying to load metadata*.

At this point, `$parameter` already have `datetime` type inside:

datetime

maks-rafalko:
Fixes #8113

* Happy with the test that is failing now
* Not happy with the implementation. This is more like a call to discuss a possible solution, if it's needed, keeping in mind the following tweets:

https://twitter.com/Ocramius/status/1251463806770872322?s=20

> Object-alike types are always checked for metadata existence first (by design). If a DBAL type is needed for an object, it is to be passed in as parameter (also by design).

> Otherwise, the ORM doesn't know if we're talking about a mapped type or a DBAL type.

> Note: this is a flaw in the architecture of DBAL types. DBAL types were initially sufficient, but as more complex mappings (such as association-as-identifier) came up, they became mostly obsolete.

So, @Ocramius, do you mean that we **can't** do optimization for `DateTimeInterface` implemented in this PR?

I'm not sure I got it, because `ParameterTypeInferer` in case of date time objects returns DBAL type right from the start (early).

So, we immediately know that `DateTimeInterface` has DBAL's `Doctrine\DBAL\Types\Type::DATETIME` type.

tl;dr

* Do we want a fix for #8113? If we do, could you please advice with the implementation, as I had many of them but not sure what is the right one
* If we don't, probably makes sense to merge Doc PR to warn users about this performance issue: https://github.com/doctrine/orm/pull/8114

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -432,7 +432,7 @@ final class Query extends AbstractQuery
     /** @return mixed[] tuple of (value, type) */
     private function resolveParameterValue(Parameter $parameter) : array
     {
-        if ($parameter->typeWasSpecified()) {
+        if ($parameter->typeWasSpecified() || $this->isEarlyInferredType($parameter)) {
             return [$parameter->getValue(), $parameter->getType()];
         }
 
@@ -790,5 +790,15 @@ final class Query extends AbstractQuery
         parent::__clone();
 
         $this->_state = self::STATE_DIRTY;
+    }
+
+    /**
+     * @param Parameter $parameter
+     *
+     * @return bool
+     */
+    private function isEarlyInferredType(Parameter $parameter)
+    {
+        return $parameter->getValue() instanceof \DateTimeInterface;
     }
 }
--- a/tests/Doctrine/Tests/ORM/Query/QueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/QueryTest.php
@@ -421,6 +421,24 @@ class QueryTest extends OrmTestCase
         self::assertEmpty($query->getResult());
     }
 
+    /** @group 8113 */
+    public function testValuesAreNotBeingResolvedForDateTimeParameterTypes() : void
+    {
+        $unitOfWork = $this->createMock(UnitOfWork::class);
+
+        $this->_em->setUnitOfWork($unitOfWork);
+
+        $unitOfWork
+            ->expects(self::never())
+            ->method('getSingleIdentifierValue');
+
+        $query = $this->_em->createQuery('SELECT d FROM ' . DateTimeModel::class . ' d WHERE d.datetime = :value');
+
+        $query->setParameter('value', new DateTime());
+
+        self::assertEmpty($query->getResult());
+    }
+
     /** @group 7982 */
     public function testNonExistentExecutor()
     {


maks-rafalko:
Sorry for posting messages from twitter, but it's better for future readers to keep them here:

https://twitter.com/Ocramius/status/1251478780381716480
from @Ocramius 
> Not to be fixed: passing in an object will always lead to a metadata lookup, unless an explicit type is given.

> I patched it to avoid repeated failed lookups back in 2.6.x (performance impact was way bigger)

PR #8116, opened 2020-04-19
Static analysis with Psalm

greg0ire:
Setup static analysis with Psalm. For now, we are using the most lenient level.

Also, this moves PHPStan to Github Actions.

I'm targeting 2.7 because I think CI pipelines should not differ too much between branches, otherwise it can be confusing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,37 @@
+name: CI
+
+on:
+  pull_request:
+
+jobs:
+  static-analysis-psalm:
+    name: "Static Analysis with Psalm"
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.4"
+
+    steps:
+      - name: "Checkout code"
+        uses: "actions/checkout@v2"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Run a static analysis with vimeo/psalm"
+        run: "vendor/bin/psalm --show-info=false --stats --output-format=github --threads=$(nproc)"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,39 @@ on:
   pull_request:
 
 jobs:
+  static-analysis-phpstan:
+    name: "Static Analysis with PHPStan"
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.4"
+
+    steps:
+      - name: "Checkout code"
+        uses: "actions/checkout@v2"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+          tools: cs2pr
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-progress --no-suggest --no-interaction --prefer-dist"
+
+      - name: "Run a static analysis with phpstan/phpstan"
+        run: "php vendor/bin/phpstan analyse --error-format=checkstyle | cs2pr"
+
   static-analysis-psalm:
     name: "Static Analysis with Psalm"
     runs-on: "ubuntu-latest"
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,9 @@
         {"name": "Marco Pivetta", "email": "ocramius@gmail.com"}
     ],
     "config": {
+        "platform": {
+            "php": "7.1.3"
+        },
         "sort-packages": true
     },
     "require": {
@@ -32,7 +35,8 @@
     "require-dev": {
         "doctrine/coding-standard": "^5.0",
         "phpunit/phpunit": "^7.5",
-        "symfony/yaml": "^3.4|^4.0|^5.0"
+        "symfony/yaml": "^3.4|^4.0|^5.0",
+        "vimeo/psalm": "^3.11"
     },
     "suggest": {
         "symfony/yaml": "If you want to use YAML Metadata Mapping Driver"
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^5.0",
+        "phpstan/phpstan": "^0.12.18",
         "phpunit/phpunit": "^7.5",
         "symfony/yaml": "^3.4|^4.0|^5.0",
         "vimeo/psalm": "^3.11"
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="8"
+    resolveFromConfigFile="true"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="https://getpsalm.org/schema/config"
+    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+>
+    <projectFiles>
+        <directory name="lib/Doctrine/ORM" />
+        <ignoreFiles>
+            <directory name="vendor" />
+        </ignoreFiles>
+    </projectFiles>
+</psalm>
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,16 +85,6 @@ jobs:
         - php ocular.phar code-coverage:upload --format=php-clover build/logs/clover.xml
 
     - stage: Code Quality
-      env: DB=none STATIC_ANALYSIS
-      php: 7.4
-      install: travis_retry composer install --prefer-dist
-      before_script:
-        - echo "extension=memcached.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
-        - echo "extension=redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
-        - travis_retry composer require --dev --prefer-dist --prefer-stable phpstan/phpstan:^0.9
-      script: vendor/bin/phpstan analyse -l 1 -c phpstan.neon lib
-
-    - stage: Code Quality
       env: DB=none BENCHMARK
       before_script: wget https://phpbench.github.io/phpbench/phpbench.phar https://phpbench.github.io/phpbench/phpbench.phar.pubkey
       script: php phpbench.phar run -l dots --report=default
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "3ecc0039f7919112197e4fb876a5184d",
+    "content-hash": "2b24c11061bfc0216397bfa7e2c17b6e",
     "packages": [
         {
             "name": "doctrine/annotations",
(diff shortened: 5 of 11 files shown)


greg0ire:
:thinking: for some reason PHPStan is refusing to run now… https://travis-ci.org/github/doctrine/orm/jobs/676946612 … not sure what it is afraid of.

composer.json
```diff
@@ -13,6 +13,9 @@
         {"name": "Marco Pivetta", "email": "ocramius@gmail.com"}
     ],
     "config": {
+        "platform": {
+            "php": "7.1.3"
```
greg0ire, line 17: I picked this version because it is require by recent versions of Psalm, I hope this is not too much of a big deal.

SenseException: It is described as "root-only" so in context of projects using ORM as a dependency this shouldn't be a problem. But I'm not so sure for the ORM-build itself which may fetch older packages listed in composer's `require`.

greg0ire: Removing it makes it really harder to use composer and be sure the `composer.lock` you generate is usable on build jobs that use old versions of php (e.g. if you install a version of a package that requires 7.2)

greg0ire:
@bendavies @localheinz I copied things from your work, please review

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+          php-version: "${{ matrix.php-version }}"
+          extensions: mbstring
+          tools: composer, cs2pr
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
```
greg0ire, line 33: @localheinz used a `php-` prefix in coding-standard, @bendavies didn't in https://github.com/doctrine/dbal/pull/3979 . Not 100% sure what's best, but I'm more leaning towards the prefixed version, it's more explicit.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: composer install
```
localheinz, line 35:
```suggestion
      - name: "Install dependencies with composer"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: "composer install --no-progress --no-suggest --no-interaction --prefer-dist --optimize-autoloader"
+
+      - name: phpstan
```
localheinz, line 38:
```suggestion
      - name: "Run a static analysis with phpstan/phpstan"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Psalm"
```
localheinz, line 71:
```suggestion
      - name: "Run a static analysis with vimeo/psalm"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          path: "~/.composer/cache"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Run a static analysis with vimeo/psalm"
+        run: "vendor/bin/psalm --output-format=checkstyle | cs2pr"
```
bendavies, line 72: psalm has native support, no need for `cs2pr`
`vendor/bin/psalm --show-info=false --stats --output-format=github --threads=4`

.github/workflows/ci.yml
bendavies: as long as it's standarised.

greg0ire: Ok then let's use the prefix :)

greg0ire:
@bendavies @localheinz @goetas @beberlei any comments about the name and the layout? Currently sql-formatter and coding-standard use a workflow named `Continuous Integration`, in a file named continuous-integration.yml, while the master branch of the ORM has a workflow named `CI` in a `ci.yml` file (that's probably why I went with this for this PR). I think I'm going to rename it to the longer counterpart, but maybe you have input on this? As you can see, this results in a `CI / ` prefix, and I don't think we plan to have anything else than continuous integration… maybe it would make more sense to have a `static-analysis.yml` file so that we have `Static Analysis / PHPStan` and `Static Analysis / Psalm`?

localheinz:
@greg0ire 

Perhaps take a look at https://github.com/ergebnis/php-library-template/tree/master/.github/workflows for some inspiration?

greg0ire:
> Perhaps take a look at https://github.com/ergebnis/php-library-template/tree/master/.github/workflows for some inspiration?

So you would leave CI-related things to leave space for bots? We have a bot for releases, and it doesn't dependen on a Github Action. Also, we don't need a job for renewing the license: https://github.com/doctrine/orm/commit/8220bb3cedab14eeb8b5701890fdf77c4788ae18

psalm.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="8"
+    resolveFromConfigFile="true"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="https://getpsalm.org/schema/config"
+    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+>
+    <projectFiles>
+        <directory name="lib/Doctrine/ORM" />
```
bendavies, line 10: tests?

greg0ire: I might contribute it once we drop php 7.1, which will allow us to install phpunit 8, which is understood by Psalm with far less effort on our side.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
+name: CI
+
+on:
+  pull_request:
+
+jobs:
+  static-analysis-phpstan:
+    name: "Static Analysis with PHPStan"
+    runs-on: ubuntu-18.04
```
bendavies, line 9: ubuntu-latest

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-progress --no-suggest --no-interaction --prefer-dist --optimize-autoloader"
+
+      - name: "Run a static analysis with phpstan/phpstan"
+        run: "php vendor/bin/phpstan analyse --error-format=checkstyle --no-progress | cs2pr"
+
+  static-analysis-psalm:
+    name: "Static Analysis with Psalm"
+    runs-on: ubuntu-18.04
```
bendavies, line 43: ubuntu-latest

.github/workflows/ci.yml
bendavies: quote for consistency

.github/workflows/ci.yml
bendavies: quote for consistency

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+    steps:
+      - name: "Checkout code"
+        uses: "actions/checkout@v2"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+          extensions: mbstring
```
bendavies, line 25: quote for consistency

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+          tools: composer, cs2pr
```
bendavies, line 26: quote for consistency

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-progress --no-suggest --no-interaction --prefer-dist --optimize-autoloader"
+
+      - name: "Run a static analysis with phpstan/phpstan"
+        run: "php vendor/bin/phpstan analyse --error-format=checkstyle --no-progress | cs2pr"
```
bendavies, line 39:
```suggestion
        run: "vendor/bin/phpstan analyse --error-format=checkstyle --no-progress | cs2pr"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-progress --no-suggest --no-interaction --prefer-dist --optimize-autoloader"
```
bendavies, line 36:
```suggestion
        run: "composer install --no-progress --no-suggest --no-interaction --prefer-dist"
```

.github/workflows/ci.yml
bendavies: no need for composer. GA has it.

.github/workflows/ci.yml
bendavies, line 25: is mb string needed??

.github/workflows/ci.yml
bendavies, line 39:
```suggestion
        run: "vendor/bin/phpstan analyse --error-format=checkstyle | cs2pr"
```

.github/workflows/ci.yml
SenseException: I'm not a fan of using the "latest" tag, because it can break a build just because of some changes after "latest" becomes silently a newer version.

greg0ire: cc @goetas who expressed the same opinion here: https://github.com/doctrine/sql-formatter/pull/51#discussion_r421962085

.github/workflows/ci.yml
greg0ire: That's something I blindly copied from the master branch: https://github.com/doctrine/orm/pull/8049

greg0ire: @beberlei maybe you know?

bendavies: i don't think you need it. the version of php on github runners comes with mbstring intsalled

bendavies: https://github.com/actions/virtual-environments/blob/48f3964fe67be7ba20761c4c186ba00418c3bb86/images/linux/scripts/installers/1804/php.sh#L156

greg0ire: Maybe that was not the case in the past, which would explain why it was needed in the master branch. I will remove it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Run a static analysis with vimeo/psalm"
+        run: "vendor/bin/psalm --show-info=false --stats --output-format=github --threads=4"
```
carusogabriel, line 71: Can we use `nproc` here? Is that possible within GA/yml?

```suggestion
        run: "vendor/bin/psalm --show-info=false --stats --output-format=github --threads=$(nproc)"
```

greg0ire: It seems to work, but then again you can use `foo` as a value here and Psalm does not complain, so I'm not sure.

greg0ire: I added a commit to display the number of proc, let's see…

greg0ire: The commit works (and says 2 lol)

bendavies: Depends if Psalm is cpu or io bound

greg0ire: > Can we use nproc here? Is that possible within GA/yml?

Looks like we can and it's possible

> Depends if Psalm is cpu or io bound

Maybe it's IO bound sometimes, and then CPU bound during other phases.

carusogabriel (approved):

composer.json
SenseException: I hope this won't give us any trouble in the future. :crossed_fingers:

.github/workflows/ci.yml
SenseException: If this is only for static analysis, it should be fine. I don't recommend it in general though.

SenseException (approved):

PR #8118, opened 2020-04-21
Do not pass entity name in \Doctrine\ORM\EntityManager::clear call

antavelis:
Hello,

Since passing the `entityName` parameter in `\Doctrine\ORM\EntityManager::clear` is deprecated, this call should be also updated to not pass the `entityName` when calling the method. In my CI/CD it triggers the following error:

```
Calling Doctrine\ORM\EntityManager::clear() with any arguments to clear specific entities is deprecated and will not be supported in Doctrine ORM 3.0.
```

Thank you 💪

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -84,7 +84,7 @@ class EntityRepository implements ObjectRepository, Selectable
      */
     public function clear()
     {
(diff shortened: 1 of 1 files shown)


beberlei:
The solution here is not to call `EntityRepository::clear()` anymore inyour code instead of clearing all entities, which is a BC break.

PR #8120, opened 2020-04-22
EntityPersisterFactory

MartkCz:


beberlei:
Entity Persisters are not extendable by design, they are too dangerous to change.

PR #8122, opened 2020-04-23
Prohibits class typo in the discriminator map

gquemener:
When a defined subclass has a case typo, the query builder will be lost
and will generate exotic table alias. This commit fixes the issue at the
root by prohibiting case typo in discriminator map.

See https://github.com/doctrine/orm/pull/8112 for the consequence of
such typo.


gquemener:
Related to https://github.com/doctrine/orm/pull/8112

gquemener:
This PR is better than https://github.com/doctrine/orm/pull/8112 in term of DX as it will raise errors way before any sql query will hit the DB. Pointing the user exactly where the issue is.

lib/Doctrine/ORM/Mapping/ClassMetadata.php
```diff
@@ -1002,6 +1003,11 @@ public function addDiscriminatorMapClass($name, string $className) : void
             throw MappingException::invalidClassInDiscriminatorMap($className, $this->className);
         }
 
+        $refl = new ReflectionClass($className);
+        if ($refl->name !== $className) {
+            throw MappingException::invalidClassInDiscriminatorMap($className, $this->className);
+        }
```
gquemener, line 1009: I'm opened to suggestion if you see a more efficient way to grab actual class name without using reflection.

gquemener:
This change could probably be considered as a BC break.

lib/Doctrine/ORM/Mapping/ClassMetadata.php
gquemener: However, I'm not sure efficiency is really researched here, as metadata are cached.

lib/Doctrine/ORM/Mapping/ClassMetadata.php
```diff
@@ -1002,6 +1003,11 @@ public function addDiscriminatorMapClass($name, string $className) : void
             throw MappingException::invalidClassInDiscriminatorMap($className, $this->className);
         }
 
+        $refl = new ReflectionClass($className);
```
gquemener, line 1006: > Throws ReflectionException if the class to reflect does not exist. [Source](https://www.php.net/manual/en/reflectionclass.construct.php)

It won't as we check for class existance on line 1001.

lib/Doctrine/ORM/Mapping/ClassMetadata.php
greg0ire: I don't think the reflection API performance are that bad, and since this is about metadata, there is probably caching after this, right? So it's not a hot path, and it's cached, isn't it?

greg0ire (dismissed):
A+ commit message, and nice tests! I'm unsure if this should really be considered a BC-break. If it is, we should probably trigger a deprecation but that sounds weird to me. In any case, I would target the bugfix branch (so 2.7)

lib/Doctrine/ORM/Mapping/ClassMetadata.php
gquemener: That's my belief as well :+1:

greg0ire:
Why did you change it back?

gquemener:
Switching the base branch to 2.7 showed 500+ commits to merge. Let me work on that.

greg0ire:
`git rebase --onto origin/2.7 origin/master prevents-sub-class-case-difference && git push --force` ;)

gquemener:
Done, thanks for the hint :+1: 

I'm curious, how would you transpose this fix into master? I've noticed quite a few changes that'll make it non trivial (ClassMetadataInfo has been renamed ClassMetadata and a few metadata drivers have been removed).

greg0ire:
I'm curious about that too to be honest :sweat_smile:

greg0ire (dismissed):

tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
```diff
@@ -1344,6 +1352,47 @@ public static function loadMetadata(ClassMetadataInfo $metadata)
     }
 }
 
+/**
+ * @Entity
+ * @InheritanceType("SINGLE_TABLE")
+ * @DiscriminatorMap({"cube" = cube::class})
+ * @DiscriminatorColumn(name="discr", length=32, type="string")
+ */
+abstract class Shape
```
SenseException, line 1361: Should this class be moved out of `AbstractMappingDriverTest`? The class already has more than 1000 LoC and it has to start somewhere. :-)

SenseException (changes requested):

tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
gquemener: Sure thing, any preference for its namespace?

greg0ire:
The build fail, please run `vendor/bin/phpcbf`

gquemener:
Done :ok_hand:

greg0ire (dismissed):
LGTM

>  In any case, I would target the bugfix branch (so 2.7)

Correction: if someone feels there is a BC-break and this should instead be a deprecation, then 2.8 should be targeted. Otherwise, 2.7 is fine.

SenseException (dismissed):
:+1:

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -2816,6 +2816,11 @@ public function addDiscriminatorMapClass($name, $className)
             throw MappingException::invalidClassInDiscriminatorMap($className, $this->name);
         }
 
+        $refl = new ReflectionClass($className);
```
beberlei, line 2819: Sorry to swoop in so late, but loading ReflectionClass inside ClassMetadataInfo is not allowed from architecture POV. This code doesn't work with static runtime reflection, which will break detached metadata use cases (such as everything with the EntityGenerator).

Instead this check is something you should put either into ClassMetadataFactory if we want this to fail at runtime, or in the SchemaValidator if we want this to fail only in debug/dev mode.

beberlei (changes requested):
The right place for validation is `ClassMetadataFactory::validateRuntimeMetadata`

gquemener:
> The right place for validation is `ClassMetadataFactory::validateRuntimeMetadata`

Thanks for pointing this out.
I'll check this out, hopefully soon :)

gquemener:
I'm wondering if this case sensitivity check should not be performed more globally as `class_exists` and `interface_exists` are used in other places.

The underlying issue is that once the class has been included once (FYI, composer autoload will not autoload class with incorrect case, as it does not respect psr-0/4 AFAIK), `class_exists` and `interface_exists` will behave in a case insensitive manner.

In the case of the discriminator map, this will generate incorrect sql statement because the system won't be able to determine the correct table alias. I'm wondering if a similar issue could not appear anywhere else where `(class|interface)_exists` is used :thinking: 

I'm thinking about associations (targetEntity with incorrect case), entity definition (name with incorrect case), ...

I need to take some more time to investigate, any hint is welcome in the meantime :)

PR #8125, opened 2020-04-29
removed repository class from metadata

binarygondola:
The annotation is required only in the next chapter of the tutorial, specifically the "Entity Repositories"


SenseException (approved):
True. This information and annotation is too early in use here. :+1:

PR #8147, opened 2020-05-16
[GH-8137] Update doctrine/inflector dependency to 1.4|2.0

### LINKED ISSUE #8137: Deprecation notices from doctrine/inflector 1.4
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7/2.8.x

#### Summary

The old inflector API is deprecated since `doctrine/inflector` 1.4.

The new API is available in inflector 1.4. The new API should be used for 2.7/2.8 and master. See https://github.com/doctrine/common/issues/889

#### Current behavior

The following deprecation is triggered by https://github.com/doctrine/orm/blob/2.7/lib/Doctrine/ORM/EntityRepository.php#L305

    The "Doctrine\Common\Inflector\Inflector::classify" method is deprecated and will be dropped in doctrine/inflector 2.0. Please update to the new Inflector 
API.

#### How to reproduce

Just install this package together with `doctrine/inflector` 1.4.

#### Expected behavior

No deprecation should be triggered

beberlei:
Resolve the deprecations by instantiating default inflector.

This cannot go into 2.7, because it would raise PHP dependency from 7.1 to 7.2.

Fixes #8137

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -58,7 +58,7 @@ class EntityRepository implements ObjectRepository, Selectable
      */
     protected $_class;
 
-    /** @var Inflector */
+    /** @var \Doctrine\Inflector\Inflector */
     private static $inflector;
 
     /**
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -25,6 +25,7 @@ use Doctrine\Inflector\InflectorFactory;
 use Doctrine\ORM\Query\ResultSetMappingBuilder;
 use Doctrine\Persistence\ObjectRepository;
 use const E_USER_DEPRECATED;
+use function lcfirst;
 use function trigger_error;
 
 /**
@@ -57,10 +58,8 @@ class EntityRepository implements ObjectRepository, Selectable
      */
     protected $_class;
 
-    /**
-     * @var \Doctrine\Inflector\Inflector
-     */
-    static private $inflector;
+    /** @var Inflector */
+    private static $inflector;
 
     /**
      * Initializes a new <tt>EntityRepository</tt>.
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -21,6 +21,7 @@ namespace Doctrine\ORM;
 
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\Common\Collections\Selectable;
+use Doctrine\Inflector\Inflector;
 use Doctrine\Inflector\InflectorFactory;
 use Doctrine\ORM\Query\ResultSetMappingBuilder;
 use Doctrine\Persistence\ObjectRepository;
@@ -58,7 +59,7 @@ class EntityRepository implements ObjectRepository, Selectable
      */
     protected $_class;
 
-    /** @var \Doctrine\Inflector\Inflector */
+    /** @var Inflector */
     private static $inflector;
 
     /**
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -19,7 +19,7 @@
 
 namespace Doctrine\ORM;
 
-use Doctrine\Common\Inflector\Inflector;
+use Doctrine\Inflector\InflectorFactory;
 use Doctrine\ORM\Query\ResultSetMappingBuilder;
 use Doctrine\Common\Persistence\ObjectRepository;
 use Doctrine\Common\Collections\Selectable;
@@ -56,6 +56,11 @@ class EntityRepository implements ObjectRepository, Selectable
      * @var \Doctrine\ORM\Mapping\ClassMetadata
      */
     protected $_class;
+
+    /**
+     * @var \Doctrine\Inflector\Inflector
+     */
+    static private $inflector;
 
     /**
      * Initializes a new <tt>EntityRepository</tt>.
@@ -308,7 +313,11 @@ class EntityRepository implements ObjectRepository, Selectable
             throw ORMException::findByRequiresParameter($method . $by);
         }
 
-        $fieldName = lcfirst(Inflector::classify($by));
+        if (self::$inflector === null) {
+            self::$inflector = InflectorFactory::create()->build();
+        }
+
+        $fieldName = lcfirst(self::$inflector->classify($by));
 
         if (! ($this->_class->hasField($fieldName) || $this->_class->hasAssociation($fieldName))) {
             throw ORMException::invalidMagicCall($this->_entityName, $fieldName, $method . $by);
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -57,6 +57,11 @@ class EntityRepository implements ObjectRepository, Selectable
      */
     protected $_class;
 
+    /**
+     * @var \Doctrine\Inflector\Inflector
```
greg0ire, line 61:
```suggestion
     * @var Inflector
```

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -79,12 +80,18 @@ class DatabaseDriver implements MappingDriver
      */
     private $namespace;
 
+    /**
+     * @var \Doctrine\Inflector\Inflector
```
greg0ire, line 84:
```suggestion
     * @var Inflector
```

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -79,12 +80,18 @@ class DatabaseDriver implements MappingDriver
...
+     */
+    protected $inflector;
```
greg0ire, line 86: Since there is an accessor, it can (and should?) be private

```suggestion
    private $inflector;
```

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -168,6 +175,16 @@ public function setTables($entityTables, $manyToManyTables)
...
     }
 
+    public function setInflector(Inflector $inflector) : void
+    {
+        $this->inflector = $inflector;
+    }
+
+    public function getInflector() : Inflector
+    {
+        return $this->Inflector;
```
greg0ire, line 185: Inflector is badly inflected here :P
```suggestion
        return $this->inflector;
```

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -168,6 +175,16 @@ public function setTables($entityTables, $manyToManyTables)
         }
     }
 
+    public function setInflector(Inflector $inflector) : void
+    {
+        $this->inflector = $inflector;
+    }
+
+    public function getInflector() : Inflector
```
greg0ire, line 183: Shouldn't it be final?
```suggestion
    final public function getInflector() : Inflector
```

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -168,6 +175,16 @@ public function setTables($entityTables, $manyToManyTables)
...
+    {
+        return $this->Inflector;
+    }
```
greg0ire, line 186: Do we really need both of these accessors?

lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -335,6 +336,11 @@ public function __construct(<params>)
 }
 ';
 
+    /**
+     * @var \Doctrine\Inflector\Inflector
```
greg0ire, line 340:
```suggestion
     * @var Inflector
```

lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -335,6 +336,11 @@ public function __construct(<params>)
...
+     */
+    protected $inflector;
```
greg0ire, line 342:
```suggestion
    private $inflector;
```

lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -592,6 +600,16 @@ public function setBackupExisting($bool)
...
     }
 
+    public function setInflector(Inflector $inflector) : void
+    {
+        $this->inflector = $inflector;
+    }
+
+    public function getInflector() : Inflector
+    {
+        return $this->Inflector;
```
greg0ire, line 610:
```suggestion
        return $this->inflector;
```

lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -592,6 +600,16 @@ public function setBackupExisting($bool)
         $this->backupExisting = $bool;
     }
 
+    public function setInflector(Inflector $inflector) : void
+    {
+        $this->inflector = $inflector;
+    }
+
+    public function getInflector() : Inflector
```
greg0ire, line 608:
```suggestion
    final public function getInflector() : Inflector
```

lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -592,6 +600,16 @@ public function setBackupExisting($bool)
...
+    {
+        return $this->Inflector;
+    }
```
greg0ire, line 611: Same question here: why expose this internal to the outside world?

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
beberlei: i removed the getters now

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
beberlei: No, in this case its used from protected methods, maybe someone overwrites them and needs access.

lib/Doctrine/ORM/Tools/EntityGenerator.php
beberlei: No, in this case its used from protected methods, maybe someone overwrites them and needs access.

lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -57,6 +58,9 @@ class EntityRepository implements ObjectRepository, Selectable
      */
     protected $_class;
 
+    /** @var Inflector */
```
beberlei, line 61: @greg0ire i really have to stop using Vim to do my Doctrine changes :) I don't have inflector type imported already, is it really necessary to make the use statement just for the docblock? seems excessive :)

lib/Doctrine/ORM/Tools/EntityGenerator.php
greg0ire: Maybe we could make the protected methods final?

lib/Doctrine/ORM/EntityRepository.php
greg0ire: I use Vim too, and it is absolutely terrific! Just make sure you use `phpactor` with it, and use `:PhpactorImportClass` ;) My vimrc is here if you want to take a look: https://github.com/greg0ire/dotvim/blob/master/vimrc

> I don't have inflector type imported already, is it really necessary to make the use statement just for the docblock? seems excessive :)

I don't make the rules, it's our coding standard… anyway, the docblock will become a type in php 7.4, which means we will end up importing the use statement if we didn't already. Let's do it now, it's just less work.

greg0ire (approved):

Issue #8097, opened 2020-04-08
Change GetSingleScalarResult phpDoc to return string|null


VincentLanglet:
#### Summary

When using `getSingleScalarResult` is a `COUNT` in my query, I always thought this would return an `int`, but I recently discovered it return a `string` instead.

I checked:
- int are return as string.
- bool are return as '0' or '1'.
- null is return as null.
- object are return with __toString.

Am I missing something or the `GetSingleScalarResult` phpDoc could be improved to `return string|null` ?

PR #8149, opened 2020-05-16
Improve phpdoc of getSingleScalarResult method

VincentLanglet:
Cf https://github.com/doctrine/orm/issues/8097

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -845,7 +845,7 @@ abstract class AbstractQuery
      *
      * Alias for getSingleResult(HYDRATE_SINGLE_SCALAR).
      *
-     * @return mixed The scalar result.
+     * @return string|null The scalar result.
      *
      * @throws NoResultException        If the query returned no result.
      * @throws NonUniqueResultException If the query result is not unique.


greg0ire (approved):

greg0ire:
Hey buddy! I reviewed your PR about return types, it's only fair that you review mine now: https://github.com/doctrine/orm/pull/8150 :trollface:

SenseException (changes requested):
``` php
$this->entityManager->createQuery(sprintf('SELECT f.num FROM %s f', Foo::class))
    ->setMaxResults(1)
    ->getSingleScalarResult()
```

`Foo::num` is an integer in my example and an PHP integer is returned by that query.

VincentLanglet:
@SenseException When I'm doing 
```
$queryBuilder = $this->createQueryBuilder('foo')
            ->select('foo.id')
            ->setMaxResults(1);
```

With
```
    /**
     * @var int|null
     *
     * @ORM\Column(type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;
```

I get 
image

So even if it's reported as an integer, the query is returning a string.

SenseException:
Can you please test this with PostgreSQL? I used an 11 version in my case.

VincentLanglet:
I was using 
```
doctrine:
    dbal:
        # configure these for your database server
        driver: pdo_mysql
        server_version: mariadb-10.2.17
```

I tried with another project
```
doctrine:
    dbal:
        # configure these for your database server
        driver: 'pgsql'
        server_version: '10.9'
```
And indeed I get an integer in this case.

I was starting to debate about yes or not this should be considered as a bug and then I discovered that
```
$queryBuilder = $this->createQueryBuilder('contractOperation')
            ->select('contractOperation.id')
            ->setMaxResults(1)
            ->getQuery()
            ->getResult();
```
Is returning
image
With MariaDB.

Since `getResult` is returning an int, I would expect that `getSingleScalarResult` returns an `int` too.

Do you agree that this could be considered as bug @SenseException and that I can open an issue ? Any idea about how to fix this ?

SenseException:
It was previously typed as `mixed` in the docs so it's hard to tell if this is a bug or not. If it is a bug, it's a minor one. For now, the typing in the phpdocs would be enough for this PR.

SenseException:
It was previously typed as `mixed` in the docs so it's hard to tell if this is a bug or not. If it is a bug, it's a minor one. For now, the typing in the phpdocs would be enough for this PR. Another PR can handle a consistent `int` return type.

VincentLanglet:
This PR should not be merge indeed.

I'll create an issue for the consistent `int` type.

SenseException:
This PR is IMO still valid, because I think fixing the return type behaviour might need to happen on a newer version of ORM.

VincentLanglet:
> This PR is IMO still valid, because I think fixing the return type behaviour might need to happen on a newer version of ORM.

`string|null` is a valid typehint with MariaDB. But this PR shouldn't be merged since we succeed to get an integer as return type with PostgreSQL. So `string|null` is not a valid typehint in general.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -845,7 +845,7 @@ public function getSingleResult($hydrationMode = null)
      *
      * Alias for getSingleResult(HYDRATE_SINGLE_SCALAR).
      *
-     * @return mixed The scalar result.
+     * @return string|null The scalar result.
```
SenseException, line 848:
```suggestion
     * @return int|string|null The scalar result.
```

It's not perfect, but it fits to the current situation for 2.7.

SenseException (changes requested):

lib/Doctrine/ORM/AbstractQuery.php
VincentLanglet: It doesn't fit.

I just tried
@ORM\Column(type="boolean") is return as boolean
@ORM\Column(type="datetime") is return as DateTime object
@ORM\Column(type="array") is return as an Array.

We certainly can get a float too.

I never used binary/blob but this is certainly convert to resource
Same for date_immutable and DateTimeImmutbale or dateinterval and DateInterval

And there is an object type, I never used it but maybe it return an object ?

Which lead to
```
boolean|int|float|string|\DateTime|\DateTime|\DateInterval|resource|array|object|null
```
Which is mixed, isn't it ?

The only type I don't know is resource and object, since I never used them.

Source:
https://www.doctrine-project.org/projects/doctrine-dbal/en/2.10/reference/types.html

SenseException: Oh, right :-D

PR #8144, opened 2020-05-15
Automated fixes with phpcbf + manual fixes

greg0ire:
~Blocked by https://github.com/doctrine/orm/pull/8150~

I often tell people to run `vendor/bin/phpcbf` to fix their PRs, by habit,
which sometimes results in confusion because it produces a huge unwanted diff.

Also, and this is key, this reduces greatly the diff between branches, which should mean far fewer conflicts when merging up, or when rebasing between 2.7. and 2.8, so maybe this will make @guilhermeblanco 's work easier as per https://github.com/orgs/doctrine/teams/doctrinecore/discussions/14 . Once this is merged, I think we should vow never to change the CS config unless we do it from the lowest branch and merge up.

There are still a lot of cs issues that need to be fixed before we can switch from git-phpcs to just phpcs.

The rules that required the most work are rules about specifying the types of elements instead of just specifying `array`. I thought that we could ignore it, but then that would mean that it wouldn't be required for new code, which would be bad, so I tried to describe it when I could figure out easily and fast enough, and in other cases, I just used `mixed[]`. These kind of changes are in the third commit, and if you are unhappy with that, I can drop that commit that has manual fixes. That will still leave us with a `vendor/bin/phpcbf` that runs without errors. 

Next steps, before merging this: creating merge up PRs, so that they can be pushed atomically all at once. That way we have no unknowns about the merge up and how hard they are. I gave merging into 2.8.x a quick try and there are not that many files conflicting.

I still would like a review, or at least comments about the above plan before going ahead, because this took me a lot of time, so I'd like to have everyone on board before I invest even more time on this.

PR #8150, opened 2020-05-17
Infer return type information with Psalter

greg0ire:
These changes are a subset of changes done with `vendor/bin/psalm --alter
--issues=LessSpecificReturnType
--allow-backwards-incompatible-changes=false`

If you want to see changes I rejected, you can have a look at https://github.com/greg0ire/doctrine-orm/pull/2

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -240,7 +240,7 @@ abstract class AbstractQuery
      *
      * @param integer $lifetime
      *
-     * @return \Doctrine\ORM\AbstractQuery This query instance.
+     * @return static This query instance.
      */
     public function setLifetime($lifetime)
     {
@@ -260,7 +260,7 @@ abstract class AbstractQuery
     /**
      * @param integer $cacheMode
      *
-     * @return \Doctrine\ORM\AbstractQuery This query instance.
+     * @return static This query instance.
      */
     public function setCacheMode($cacheMode)
     {
@@ -1068,7 +1068,7 @@ abstract class AbstractQuery
      * Will return the configured id if it exists otherwise a hash will be
      * automatically generated for you.
      *
-     * @return array ($key, $hash)
+     * @return array<string, string> ($key, $hash)
      */
     protected function getHydrationCacheId()
     {
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -348,7 +348,9 @@ class DefaultQueryCache implements QueryCache
      * @param array                             $assoc
      * @param mixed                             $assocValue
      *
-     * @return array|null
+     * @return mixed[]|null
+     *
+     * @psalm-return array{targetEntity: string, type: mixed, list?: array[], identifier?: array}|null
      */
     private function storeAssociationCache(QueryCacheKey $key, array $assoc, $assocValue)
     {
--- a/lib/Doctrine/ORM/Event/OnClearEventArgs.php
+++ b/lib/Doctrine/ORM/Event/OnClearEventArgs.php
@@ -38,7 +38,7 @@ class OnClearEventArgs extends \Doctrine\Common\EventArgs
     private $em;
 
     /**
-     * @var string
+     * @var string|null
      */
     private $entityClass;
 
--- a/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
+++ b/lib/Doctrine/ORM/Id/AbstractIdGenerator.php
@@ -40,7 +40,7 @@ abstract class AbstractIdGenerator
      * By default, this method returns FALSE. Generators that have this requirement
      * must override this method and return TRUE.
      *
-     * @return boolean
+     * @return bool
      */
     public function isPostInsertGenerator()
     {
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -260,8 +260,19 @@ abstract class AbstractHydrator
      * @param array &$id                 Dql-Alias => ID-Hash.
      * @param array &$nonemptyComponents Does this DQL-Alias has at least one non NULL value?
      *
-     * @return array  An array with all the fields (name => value) of the data row,
-     *                grouped by their component alias.
+     * @return array<string, array<string, mixed>> An array with all the fields
+     *                                             (name => value) of the data
+     *                                             row, grouped by their
+     *                                             component alias.
+     *
+     * @psalm-return array{
+     *                   data: array<array-key, array>,
+     *                   newObjects?: array<array-key, array{
+     *                       class: mixed,
+     *                       args?: array
+     *                   }>,
+     *                   scalars?: array
+     *               }
      */
     protected function gatherRowData(array $data, array &$id, array &$nonemptyComponents)
     {
--- a/lib/Doctrine/ORM/Mapping/Builder/AssociationBuilder.php
+++ b/lib/Doctrine/ORM/Mapping/Builder/AssociationBuilder.php
@@ -58,7 +58,7 @@ class AssociationBuilder
     /**
      * @param string $fieldName
      *
-     * @return AssociationBuilder
+     * @return static
      */
     public function mappedBy($fieldName)
     {
@@ -70,7 +70,7 @@ class AssociationBuilder
     /**
      * @param string $fieldName
      *
-     * @return AssociationBuilder
+     * @return static
      */
     public function inversedBy($fieldName)
     {
@@ -80,7 +80,7 @@ class AssociationBuilder
     }
 
     /**
-     * @return AssociationBuilder
+     * @return static
      */
     public function cascadeAll()
     {
@@ -90,7 +90,7 @@ class AssociationBuilder
     }
 
     /**
-     * @return AssociationBuilder
+     * @return static
      */
     public function cascadePersist()
     {
@@ -100,7 +100,7 @@ class AssociationBuilder
     }
 
     /**
-     * @return AssociationBuilder
+     * @return static
      */
     public function cascadeRemove()
     {
@@ -110,7 +110,7 @@ class AssociationBuilder
     }
 
     /**
-     * @return AssociationBuilder
+     * @return static
      */
     public function cascadeMerge()
     {
@@ -120,7 +120,7 @@ class AssociationBuilder
     }
 
     /**
-     * @return AssociationBuilder
+     * @return static
      */
     public function cascadeDetach()
     {
@@ -130,7 +130,7 @@ class AssociationBuilder
     }
 
     /**
-     * @return AssociationBuilder
+     * @return static
      */
     public function cascadeRefresh()
     {
@@ -140,7 +140,7 @@ class AssociationBuilder
     }
 
     /**
-     * @return AssociationBuilder
+     * @return static
      */
     public function fetchExtraLazy()
     {
@@ -150,7 +150,7 @@ class AssociationBuilder
     }
 
     /**
-     * @return AssociationBuilder
+     * @return static
      */
     public function fetchEager()
     {
@@ -160,7 +160,7 @@ class AssociationBuilder
     }
 
     /**
-     * @return AssociationBuilder
+     * @return static
      */
     public function fetchLazy()
     {
@@ -179,7 +179,7 @@ class AssociationBuilder
      * @param string|null $onDelete
      * @param string|null $columnDef
      *
-     * @return AssociationBuilder
+     * @return static
      */
     public function addJoinColumn($columnName, $referencedColumnName, $nullable = true, $unique = false, $onDelete = null, $columnDef = null)
     {
@@ -198,7 +198,7 @@ class AssociationBuilder
     /**
      * Sets field as primary key.
      *
-     * @return self
+     * @return static
      */
     public function makePrimaryKey()
     {
@@ -210,7 +210,7 @@ class AssociationBuilder
     /**
      * Removes orphan entities when detached from their parent.
      *
-     * @return self
+     * @return static
      */
     public function orphanRemoval()
     {
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -28,6 +28,7 @@ use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\ReflectionService;
 use InvalidArgumentException;
 use ReflectionClass;
+use ReflectionProperty;
 use RuntimeException;
 
 /**
@@ -640,7 +641,7 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * The ReflectionProperty instances of the mapped class.
      *
-     * @var \ReflectionProperty[]
+     * @var ReflectionProperty[]
      */
     public $reflFields = [];
 
@@ -667,7 +668,7 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * Gets the ReflectionProperties of the mapped class.
      *
-     * @return array An array of ReflectionProperty instances.
+     * @return ReflectionProperty[] An array of ReflectionProperty instances.
      */
     public function getReflectionProperties()
     {
@@ -679,7 +680,7 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param string $name
      *
-     * @return \ReflectionProperty
+     * @return ReflectionProperty
      */
     public function getReflectionProperty($name)
     {
@@ -689,7 +690,7 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * Gets the ReflectionProperty for the single identifier field.
      *
-     * @return \ReflectionProperty
+     * @return ReflectionProperty
      *
      * @throws BadMethodCallException If the class has a composite identifier.
      */
@@ -805,7 +806,7 @@ class ClassMetadataInfo implements ClassMetadata
      *      - reflClass (ReflectionClass)
      *      - reflFields (ReflectionProperty array)
      *
-     * @return array The names of all the fields that should be serialized.
+     * @return string[] The names of all the fields that should be serialized.
      */
     public function __sleep()
     {
@@ -1096,7 +1097,10 @@ class ClassMetadataInfo implements ClassMetadata
      * @param string $fieldName
      * @param array  $cache
      *
-     * @return array
+     * @return mixed[]
+     *
+     * @psalm-param array{usage: mixed, region: mixed} $cache
+     * @psalm-return array{usage: mixed, region: mixed}
      */
     public function getAssociationCacheDefaults($fieldName, array $cache)
     {
@@ -1445,9 +1449,25 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param array $mapping The mapping.
      *
-     * @return array The updated mapping.
+     * @return mixed[] The updated mapping.
      *
      * @throws MappingException If something is wrong with the mapping.
+     *
+     * @psalm-return array{
+     *                   mappedBy: mixed,
+     *                   inversedBy: mixed,
+     *                   isOwningSide: bool,
+     *                   sourceEntity: string,
+     *                   targetEntity: string,
+     *                   fieldName: mixed,
+     *                   fetch: mixed,
+     *                   cascade: array<array-key,string>,
+     *                   isCascadeRemove: bool,
+     *                   isCascadePersist: bool,
+     *                   isCascadeRefresh: bool,
+     *                   isCascadeMerge: bool,
+     *                   isCascadeDetach: bool
+     *               }
      */
     protected function _validateAndCompleteAssociationMapping(array $mapping)
     {
@@ -1565,10 +1585,12 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param array $mapping The mapping to validate & complete.
      *
-     * @return array The validated & completed mapping.
+     * @return mixed[] The validated & completed mapping.
      *
      * @throws RuntimeException
      * @throws MappingException
+     *
+     * @psalm-return array{isOwningSide: mixed, orphanRemoval: bool, isCascadeRemove: bool}
      */
     protected function _validateAndCompleteOneToOneMapping(array $mapping)
     {
@@ -1656,10 +1678,27 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param array $mapping The mapping to validate and complete.
      *
-     * @return array The validated and completed mapping.
+     * @return mixed[] The validated and completed mapping.
      *
      * @throws MappingException
      * @throws InvalidArgumentException
+     *
+     * @psalm-return array{
+     *                   mappedBy: mixed,
+     *                   inversedBy: mixed,
+     *                   isOwningSide: bool,
+     *                   sourceEntity: string,
+     *                   targetEntity: string,
+     *                   fieldName: mixed,
+     *                   fetch: int|mixed,
+     *                   cascade: array<array-key,string>,
+     *                   isCascadeRemove: bool,
+     *                   isCascadePersist: bool,
+     *                   isCascadeRefresh: bool,
+     *                   isCascadeMerge: bool,
+     *                   isCascadeDetach: bool,
+     *                   orphanRemoval: bool
+     *               }
      */
     protected function _validateAndCompleteOneToManyMapping(array $mapping)
     {
@@ -1683,9 +1722,11 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param array $mapping The mapping to validate & complete.
      *
-     * @return array The validated & completed mapping.
+     * @return mixed[] The validated & completed mapping.
      *
      * @throws \InvalidArgumentException
+     *
+     * @psalm-return array{isOwningSide: mixed, orphanRemoval: bool}
      */
     protected function _validateAndCompleteManyToManyMapping(array $mapping)
     {
@@ -1862,7 +1903,9 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param array|null $fieldNames
      *
-     * @return array
+     * @return mixed[]
+     *
+     * @psalm-return list<string>
      */
     public function getColumnNames(array $fieldNames = null)
     {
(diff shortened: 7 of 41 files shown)


lib/Doctrine/ORM/Event/OnClearEventArgs.php
```diff
@@ -67,7 +67,7 @@ public function getEntityManager()
     /**
      * Name of the entity class that is cleared, or empty if all are cleared.
      *
-     * @return string|null
+     * @return string
```
VincentLanglet, line 70: There is a check `$this->entityClass === null` line 84.

Are you sure this value can't be null ?

greg0ire: Right, I'll revert this, what is wrong here is the property's phpdoc, let's fix it.

lib/Doctrine/ORM/Id/AbstractIdGenerator.php
```diff
@@ -40,7 +40,7 @@ abstract class AbstractIdGenerator
      * By default, this method returns FALSE. Generators that have this requirement
      * must override this method and return TRUE.
      *
-     * @return boolean
+     * @return false
```
VincentLanglet, line 43: I think it's better to keep `return boolean` to show all possible values.
It can avoid some `return type no compatible with parent` error message too.

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
```diff
@@ -74,7 +74,7 @@ public function __construct($queryAliases)
      * Gets bound parameters.
      * Filled after {@link dispach()}.
      *
-     * @return \Doctrine\Common\Collections\Collection
+     * @return ArrayCollection
```
VincentLanglet, line 77: Isn't better to keep the Interface as return type ?

lib/Doctrine/ORM/Id/AbstractIdGenerator.php
greg0ire: Can it really? `false` is a boolean, and is not really a subtype of that. We have both Psalm and Phpstan running on the ORM and neither complained…

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
greg0ire: I'm not sure, let see what other people say.

lib/Doctrine/ORM/Id/AbstractIdGenerator.php
VincentLanglet: https://phpstan.org/r/bde04f1f-84a8-4ad9-888d-91b617d29e24

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -28,6 +28,7 @@
 use Doctrine\Persistence\Mapping\ReflectionService;
 use InvalidArgumentException;
 use ReflectionClass;
+use ReflectionProperty;
```
ostrolucky, line 31: It's odd that we are mixing import and fcqn. Property uses \ReflectionProperty

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1096,7 +1097,9 @@ public function enableAssociationCache($fieldName, array $cache)
      * @param string $fieldName
      * @param array  $cache
      *
-     * @return array
+     * @return mixed[]
+     *
+     * @psalm-return array{usage: mixed, region: mixed}
```
ostrolucky, line 1102: We should probably apply same for $cache `@param`

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
kuraobi: People will code against your advertised return type. The question is: do you need this method to return an ArrayCollection for it to be used properly, or would any Collection be the same, meaning the choice of ArrayCollection is an implementation detail internal to the method itself?

kuraobi: People will code against your advertised return type. The question is: do you need this method to return an ArrayCollection for it to be used properly, or would any Collection be the same, meaning the choice of ArrayCollection is an implementation detail internal to the method itself?

lib/Doctrine/ORM/Tools/ConvertDoctrine1Schema.php
```diff
@@ -170,9 +172,10 @@ private function convertColumns($className, array $model, ClassMetadataInfo $met
      * @param string|array      $column
      * @param ClassMetadataInfo $metadata
      *
-     * @return array
+     * @return mixed[]
      *
      * @throws ToolsException
+     *
```
ostrolucky, line 178: ?

ostrolucky (changes requested):
This is similar to https://github.com/doctrine/orm/pull/8144, we should think about complications when upmerging to 3.x, so let's merge it after it's clear what to do there

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
ostrolucky: I agree with @kuraobi

greg0ire: I don't _need_ it to be advertised as such, but as a bonus, people can have methods specific to ArrayCollection show up in their auto-completion. Also, their SA tools will not complain if they do use such methods.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
greg0ire: That shall be fixed in #8144

lib/Doctrine/ORM/Tools/ConvertDoctrine1Schema.php
greg0ire: That's me rejecting something badly, I will remove this.

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
kuraobi: Sure, but the cost is that you are committing to return an ArrayCollection until at least the next major release, so you have to be sure that you're ok with that. If you're fine with the trade-off, go for it.

greg0ire: @kuraobi @ostrolucky would you consider suddenly returning something else than `ArrayCollection` here a BC-break?

ostrolucky: It's BC breaking and irrelevant change for goal of this PR. We shouldn't be changing public API in this PR, that's not its goal. Previously, API said this method needs to return Collection, now we say it needs concrete implementation of Collection.

kuraobi: For people using SA, @return ArrayCollection is the same as `function f(): ArrayCollection`, because their code relies on that. So changing the return type would be a BC break IMHO.

greg0ire: No I mean changing the actual returned type:

```diff
/** @return Collection */
function ()
{
-    return ArrayCollection();
+    return SomeOtherCollection();
}
```

If this is a BC-break, then we are already commited to return `ArrayCollection`. If this isn't, I should revert this change.

ostrolucky: It's a LSP breaking change. I don't know why are we considering this change here, API changes are not a goal of this PR. Return type was defined here and was correct. Did Psalm do it automatically? I would report it as a psalm bug.

lib/Doctrine/ORM/Id/AbstractIdGenerator.php
greg0ire: Ah it's probably that we don't use a strict enough level yet… I will revert this.

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
ostrolucky: No, your last diff isn't a bc break

greg0ire: Ok I will revert this. Maybe the right change would be to have this:

```
/**
 * @return Collection
 * @psalm-return ArrayCollection
 */

File: lib/Doctrine/ORM/EntityRepository.php [hunk 7]
```diff
@@ -276,7 +276,7 @@ protected function getClassMetadata()
      *
      * @param \Doctrine\Common\Collections\Criteria $criteria
      *
-     * @return \Doctrine\Common\Collections\Collection
+     * @return LazyCriteriaCollection
```
greg0ire, line 279: I'm reverting this too

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
ostrolucky: I don't think so, psalm-return overrides return. What we are trying to tell consumers here is that they cannot rely on us returning ArrayCollection, we can change it any time to different instance of Collection.

beberlei: ArrayCollection implements both Collection and Selectable, so it might be a better typehint. However @ostrolucky is right that it could theoretically return a PersistentCollection as well (it does not yet I believe)

greg0ire: I could use `Collection&Selectable`, WDYT?

ostrolucky: It's still narrowing the type, which is a BC break. Please propose it for master if you want that

kuraobi: Narrowing a return type in a concrete class is not a BC break AFAIK, unless you have children classes and no interface to enforce a common contract.

greg0ire: Well there is no interface here, and the class is not final…

ostrolucky:
```
>>> interface a{function a():IteratorAggregate;}
>>> class b implements a{function a():ArrayObject {}}
>>> class c extends b {function a(): IteratorAggregate {}}
PHP Fatal error:  Declaration of c::a(): IteratorAggregate must be compatible with b::a(): ArrayObject
```

kuraobi: Yeah so we're all saying the exact same thing :laughing:

greg0ire:
> This is similar to #8144, we should think about complications when upmerging to 3.x, so let's merge it after it's clear what to do there

Let take this with a grain of salt (lol): the order of magnitude is not really the same (42 files changed vs more than 1000). I can definitely cherry-pick this myself in master afterwards, and I can't say the same about #8144

greg0ire:
I pushed a new version that should make the `@psalm-return` annotations easier on the eye

ostrolucky:
I appreciate you volunteer to cherry-pick this to master, but I'm not sure it will solve issues when rebasing it. What I worry would happen is that even if you cherry pick this, during rebasing all conflicts would have to be solved again. But of course, I acknowledge this is much smaller PR. I would merge it, but let's see what @guilhermeblanco has to say.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -240,7 +240,7 @@ public function getLifetime()
      *
      * @param integer $lifetime
      *
-     * @return \Doctrine\ORM\AbstractQuery This query instance.
+     * @return self This query instance.
```
lcobucci, line 243: Shouldn't we use `static` instead? Since this is an abstract class and all....

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -1068,7 +1068,7 @@ private function getTimestampKey()
      * Will return the configured id if it exists otherwise a hash will be
      * automatically generated for you.
      *
-     * @return array ($key, $hash)
+     * @return string[] ($key, $hash)
```
lcobucci, line 1071:
```suggestion
     * @return array<string, string> ($key, $hash)
```

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -348,7 +348,9 @@ public function put(QueryCacheKey $key, ResultSetMapping $rsm, $result, array $h
      * @param array                             $assoc
      * @param mixed                             $assocValue
      *
-     * @return array|null
+     * @return mixed[]|null
+     *
+     * @psalm-return array{targetEntity: string, type: mixed, list?: array<array-key, array>, identifier?: array}|null
```
lcobucci, line 353: AFAIK psalm, phpstan, and phpcs understand that if we use it on `@return`.

Also `array<array-key, array>` is a bit vague, isn't it?

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1862,7 +1903,9 @@ public function hasField($fieldName)
      *
      * @param array|null $fieldNames
      *
-     * @return array
+     * @return mixed[]
+     *
+     * @psalm-return list<array-key>
```
lcobucci, line 1908:
```suggestion
     * @psalm-return list<string>
```

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -626,7 +636,20 @@ private function joinColumnToArray(Mapping\JoinColumn $joinColumn)
...
+     *
+     * @psalm-return array{
+     *                   fieldName: string,
+     *                   type: mixed,
+     *                   scale: int,
+     *                   length: int,
+     *                   unique: bool,
+     *                   nullable: bool,
+     *                   precision: int,
+     *                   options?: array,
```
lcobucci, line 649: Array of what?

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -488,7 +505,9 @@ private function buildToOneAssociationMappings(ClassMetadataInfo $metadata)
      *
      * @param \Doctrine\DBAL\Schema\Table $table
      *
-     * @return array
+     * @return ForeignKeyConstraint[]
+     *
+     * @psalm-return array<array-key, ForeignKeyConstraint>
```
lcobucci, line 510: Not really needed, isn't it?

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -795,7 +816,9 @@ private function columnToArray(SimpleXMLElement $fieldMapping)
      *
      * @param SimpleXMLElement $cacheMapping
      *
-     * @return array
+     * @return mixed[]
+     *
+     * @psalm-return array{usage: mixed, region: null|string}
```
lcobucci, line 821: If I'm not mistaken CS enforces `null` to be the last type.

```suggestion
     * @psalm-return array{usage: mixed, region: string|null}
```

lib/Doctrine/ORM/Mapping/Driver/YamlDriver.php
```diff
@@ -789,7 +813,10 @@ private function columnToArray($fieldName, $column)
      *
      * @param array $cacheMapping
```
lcobucci, line 814: This should have been changed too, right?

lib/Doctrine/ORM/Mapping/Reflection/ReflectionPropertiesGetter.php
```diff
@@ -79,8 +79,10 @@ public function getProperties($className)
      * @param string $className
      *
      * @return ReflectionClass[]
+     *
+     * @psalm-return list<ReflectionClass>
```
lcobucci, line 83: Is it really needed?

lib/Doctrine/ORM/Mapping/Reflection/ReflectionPropertiesGetter.php
```diff
@@ -101,8 +103,10 @@ private function getHierarchyClasses($className)
      * @param ReflectionClass $reflectionClass
      *
      * @return ReflectionProperty[]
+     *
+     * @psalm-return array<string, ReflectionProperty>
```
lcobucci, line 107: Same here

lib/Doctrine/ORM/NativeQuery.php
```diff
@@ -37,9 +37,9 @@ final class NativeQuery extends AbstractQuery
      *
      * @param string $sql
      *
-     * @return NativeQuery This query instance.
+     * @return self This query instance.
      */
-    public function setSQL($sql)
+    public function setSQL($sql) : self
```
lcobucci, line 42: Final classes are just :heart:

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -360,6 +360,11 @@ protected function fetchVersionValue($versionedClass, array $id)
         return Type::getType($fieldMapping['type'])->convertToPHPValue($value, $this->platform);
     }
 
+    /**
+     * @return mixed[]
```
lcobucci, line 364: Wouldn't this be something like `\Doctrine\DBAL\Types\Type[]`?

lib/Doctrine/ORM/Query/Expr/Base.php
```diff
@@ -66,7 +66,7 @@ public function __construct($args = [])
     /**
      * @param array $args
      *
-     * @return Base
+     * @return self
```
lcobucci, line 69: `static`?

lib/Doctrine/ORM/Query/ParserResult.php
```diff
@@ -65,8 +65,8 @@ public function __construct()
     /**
      * Gets the ResultSetMapping for the parsed query.
      *
-     * @return ResultSetMapping|null The result set mapping of the parsed query or NULL
-     *                               if the query is not a SELECT query.
+     * @return ResultSetMapping The result set mapping of the parsed query or
+     *                          NULL if the query is not a SELECT query.
```
lcobucci, line 69: Comment and change don't make sense together

phpcs.xml.dist
```diff
@@ -125,6 +125,10 @@
         <exclude-pattern>*/tests/*</exclude-pattern>
     </rule>
 
+    <rule ref="SlevomatCodingStandard.Classes.UnusedPrivateElements.UnusedMethod">
+        <exclude-pattern>lib/Doctrine/ORM/Mapping/Reflection/ReflectionPropertiesGetter.php</exclude-pattern>
+    </rule>
```
lcobucci, line 130: I'd suggest using `// phpcs:ignore SlevomatCodingStandard.Classes.UnusedPrivateElements.UnusedMethod` only on the method that's being ignored instead of excluding the entire file.

lcobucci (commented):
There're plenty of `array<array-key...` things there that might be relevant for us to alter. Despite that, thanks a lot :+1:

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
greg0ire: > Also array<array-key, array> is a bit vague, isn't it?

woops, will change that indeed

> AFAIK psalm, phpstan, and phpcs understand that if we use it on @return.

I'm doing this out of consideration for PhpStorm users (and also it's what Psalter did naturally). I'm told PhpStorm does not understand this yet, cc @goetas with who I had that debate.

lcobucci: That's probably true... 😭

guilhermeblanco: My general feel on arrays matters are:
1- If array key is ignorable, then use `type[]` syntax
2- If array key is relevant, then use `array<type, type>` syntax
3- For maps of maps, repeat the process `array<type, array<type, type>>`

Anything more complex than 2 requires a comment that describes the return. Also, this hints me that it would be a good candidate for a wrapper object (that encapsulates this logic, make them immutable if needed, only allow pin-point modifications, etc).

Finally, avoid tools specific (aka. non-standardized) markers, such as phpcbf, psalm, phpstan, etc, but rather backed by official standard published publicly, such as xUnit, coverage, etc.

lib/Doctrine/ORM/Event/OnClearEventArgs.php
```diff
@@ -38,7 +38,7 @@ class OnClearEventArgs extends \Doctrine\Common\EventArgs
     private $em;
 
     /**
-     * @var string
+     * @var string|null
```
guilhermeblanco, line 41: Please review this, I think the string is always true, not nullable.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -260,8 +260,17 @@ protected function hydrateRowData(array $data, array &$result)
      * @param array &$id                 Dql-Alias => ID-Hash.
      * @param array &$nonemptyComponents Does this DQL-Alias has at least one non NULL value?
      *
-     * @return array  An array with all the fields (name => value) of the data row,
-     *                grouped by their component alias.
+     * @return mixed[][] An array with all the fields (name => value) of the
```
guilhermeblanco, line 263: Sharing same thing here as mentioned on https://github.com/doctrine/orm/pull/8150/files#r427314368

guilhermeblanco: Copying same comments shared on Slack:
> I’m against using these types of annotations, but rather provide relevant comments on the actual data structure. Also, we should be doing better at describing the data structure (mixed[][] is far from ideal), and rely on array<string, array<string, mixed>> in this scenario. Also, meaningful comments should exist describing the data structure… and finally, complex arrays hierarchies like this one screams the needs of wrapper objects.

lib/Doctrine/ORM/Mapping/Builder/AssociationBuilder.php
```diff
@@ -58,7 +58,7 @@ public function __construct(ClassMetadataBuilder $builder, array $mapping, $type
     /**
      * @param string $fieldName
      *
-     * @return AssociationBuilder
+     * @return self
```
guilhermeblanco, line 61: `static` as return type everywhere in this file, since it's inherited by more specialized AssociationBuilders

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1445,9 +1449,25 @@ protected function _validateAndCompleteFieldMapping(array &$mapping)
      *
      * @param array $mapping The mapping.
      *
-     * @return array The updated mapping.
+     * @return mixed[] The updated mapping.
```
guilhermeblanco, line 1452: Oh... now I remember why we started 3.0...
Maybe we should do something on psalm to ignore these return mappings and only enforce on 3.0. The entire 2.X series we are only relying on complex array structures as they were 1- more memory effective and 2- faster than objects.
This is no longer true on PHP 7.

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
guilhermeblanco: Of "both"! Ya... if you use `['foo' => true]` or `['foo']`, it'll lead to the same result and will become `FOO` on column definition.

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
guilhermeblanco: Should be `array<string, ForeignKeyConstraint>`, since it returns the column name as keys IIRC.

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
greg0ire: > Anything more complex than 2 requires a comment that describes the return

From the original author though, I will have a hard time commenting on that after the fact. I agree that we could require that from contributors.

> Also, this hints me that it would be a good candidate for a wrapper object (that encapsulates this logic, make them immutable if needed, only allow pin-point modifications, etc).

I love how this PR makes that kind of code smell more visible

> Finally, avoid tools specific (aka. non-standardized) markers, such as phpcbf, psalm, phpstan, etc, but rather backed by official standard published publicly, such as xUnit, coverage, etc.

My hope was to remove the prefixes once Phpstorm has caught up.

lib/Doctrine/ORM/Event/OnClearEventArgs.php
greg0ire: `clear()` can (and should?) be called without an entity name. And from the code it looks like that really happens.

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
orklah: > Finally, avoid tools specific (aka. non-standardized) markers, such as phpcbf, psalm, phpstan

My two cents: psalm prefixes are actually compatible with Phpstan and Phan for the most part. Some syntax used is inspired by the unfinished PSR-5 about PHPDoc and it's quite probable that any syntax used in static analysis will make it's way into standards in the end.
Phpstorm is lagging behind because they make the mistake of waiting for standards. It slows adoption, which in turn slows the publication of standards. Adoption by big tools is exactly what we need to help everything move.

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
greg0ire: Let's use `mixed[]` here I guess :sweat_smile:

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
greg0ire: You are correct :) , it just doesn't enforce it here, let's change that.

lib/Doctrine/ORM/Mapping/Reflection/ReflectionPropertiesGetter.php
greg0ire: I don't know, but it's more information

lib/Doctrine/ORM/Query/ParserResult.php
greg0ire: :thinking: indeed. The property is private, and from reading its accessor and the setter, it does not look like it could be null, so let's remove that part of the comment.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
greg0ire: Glad to learn all these `mixed[]` will go away in 3.x :)

beberlei (approved):
What could go wrong? only comments :D

PR #8154, opened 2020-05-21
Fix deprecation warning when using doctrine/inflector 1.4

W0rma:
https://github.com/doctrine/orm/pull/8147 fixed the deprecation warnings for 2.8.

This PR fixes deprecation warnings in 2.7 while keeping compatibility with PHP 7.1.

Fixed https://github.com/doctrine/orm/issues/8137

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -21,7 +21,9 @@ namespace Doctrine\ORM;
 
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\Common\Collections\Selectable;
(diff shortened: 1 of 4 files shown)


beberlei:
A fix for this was merged into 2.8 already. 2.7 doesnt work because it would increase php platform requirement.

W0rma:
> 2.7 doesnt work because it would increase php platform requirement.

@beberlei Are you sure? This package requires at least `doctrine/common` 2.11 which itself requires `"doctrine/inflector` 1.0. My changes are compatible with inflector 1.0-1.4 - the new api is already part of 1.4.

PR #8165, opened 2020-06-02
Manually merge 2.7 into 2.8.x

greg0ire:
Not sure why doctrinebot didn't give it a try. I'm using the draft status so that this doesn't get merge by accident, which would result in 2 merges (not a big deal, but let's not needlessly confuse people). Once I get a green build and an approval, I will be able to push.


greg0ire:
Not sure why we suddenly require `ext-dom` in GA… :thinking: `phar-io/manifest` wasn't upgraded.

greg0ire:
Apparently the root cause is https://github.com/actions/virtual-environments/issues/981

beberlei (dismissed):

greg0ire:
I think I will wait for it to be fixed and revert 6393f8b00ab5e2facd996e511825685f82c46426

greg0ire:
[It's supposed to be fixed now](https://github.com/actions/virtual-environments/issues/981#issuecomment-637724378)

PR #8166, opened 2020-06-02
Require persistence 2

greg0ire:
This allows us to drop the autoload-calls-based BC layer.


greg0ire:
I think this would not be accepted in 2.7, tell me if you think otherwise and I will happily retarget.

greg0ire:
Woops didn't mean to push to `origin`, sorry :sweat_smile:

ostrolucky:
Isn't this BC break? Injecting common\persistence objects won't cause errors now?

greg0ire:
:thinking: how would you inject `Common\Persistence` object if you can't even build them because the classes do not exist anymore?

greg0ire:
:thinking: to be on the safe side, let's drop `doctrine/common` 2 too, so that people don't expect `Common\Persistence` types to be provided by `doctrine/common`.

ostrolucky (approved):
LGTM, it just irks me that we still have some remnaints of common\persistencene in test cases, but that's because that class is in doctrine/persistence 2

greg0ire:
Ah yes you're referring to that one class that disappears in Persistence 3, right?

ostrolucky:
yes

beberlei (approved):

ivanbogomoloff:
> Isn't this BC break? Injecting common\persistence objects won't cause errors now?

Yes it is.
This changes break some client code who use type hinted with non strict package version.
For example:
composer.json out box in symfony 5 project have only doctrine/orm like this.
```json
..
 "doctrine/doctrine-migrations-bundle": "^2.0",
 "doctrine/migrations": "^2.2",
 "doctrine/orm": "^2.7",
..
```
and no `doctrine/common` package defined, because it's orm dependency.

Then after this upgrade orm will get common with 3.0 version and will break with 
Invalid type hint for `Doctrine\Common\Persistence\ManagerRegistry`  like this

```bash
 - Updating doctrine/common (2.13.3 => 3.1.0): 
  - Updating doctrine/inflector (1.4.3 => 2.0.3): 
  - Updating doctrine/orm (2.7.4 => 2.8.1):
```
and this code will break

```php
    use Doctrine\Common\Persistence\ManagerRegistry;
    
public function __construct(ManagerRegistry $registry) // <-- HERE BREAK by invalid type
    {
        parent::__construct($registry, Entity::class);
    }
```

Ocramius:
> This changes break some client code who use type hinted with non strict package version.


That's up to downstream consumers to handle themselves: there are tools, for that, such as [`maglnet/composer-require-checker`](https://github.com/maglnet/ComposerRequireChecker).

beberlei:
I believe @greg0ire can chime in here, but the fix is that you also upgrade doctrine-bundle or downgrade common explicitly. While we do our best to find the right constraints we leave some space to avoid locking versions down too much, and this requires your intervention

greg0ire:
> and no doctrine/common package defined, because it's orm dependency.

That sounds fine. But then you write this (in your project, right? Can't really tell from the code because you did not include the namespace) :

```
   use Doctrine\Common\Persistence\ManagerRegistry;
    
public function __construct(ManagerRegistry $registry) // <-- HERE BREAK by invalid type
    {
        parent::__construct($registry, Entity::class);
    }
```

As soon as you write that first line, you are supposed to require `doctrine/common` explicitly.

PR #8182, opened 2020-06-16
Updated annotations-reference.rst @GeneratedValue

CodyCodeman:
Clarified the possibly unwanted effects of using @GeneratedValue more than once or anywhere else within the entity besides an @Id field.


SenseException:
@CodyCodeman Thanks for your PR. Can you please elaborate what this new paragraph describes? Is this about having the two `GeneratedValue` in the same property?

PR #8184, opened 2020-06-17
Allow using on PHP 7.1 with Composer 2

nicolas-grekas:
Similar to https://github.com/doctrine/dbal/pull/4084

In order to support Composer 2, `ocramius/package-versions` requires PHP 7.4.
This effectively means that using it in "require" forces ppl on PHP 7.1 to use Composer 1.

@seldaek created the `composer/package-versions-deprecated` to unlock this situation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
     "require": {
         "php": "^7.1",
         "ext-pdo": "*",
+        "composer/package-versions-deprecated": "^1.8",
         "doctrine/annotations": "^1.8",
         "doctrine/cache": "^1.9.1",
         "doctrine/collections": "^1.5",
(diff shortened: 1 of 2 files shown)


beberlei:
@nicolas-grekas any idea what the travis failure is? weid console undefined function thing.

Otherwise +1 for going forward with this.

nicolas-grekas:
> any idea what the travis failure is? weid console undefined function thing

Nope, I can't reproduce locally. The call to stream_isatty is guarded by a function_exists check. That's strange.

greg0ire:
[The cron build fails with the same message](https://travis-ci.org/github/doctrine/orm/builds/699328588), so it's not related to your changes.

greg0ire:
Fixed in https://github.com/doctrine/orm/pull/8186

greg0ire (approved):

greg0ire:
Thanks @nicolas-grekas !

cminovici:
Hello there. Could you also create a new tag for this?

PR #8187, opened 2020-06-17
 Use CodeCov instead of Scrutinizer 

greg0ire:

PR #8190, opened 2020-06-18
add dev files to .gitattribute

vladysta:
those files probably are not used by end user of the lib.
And they where added after the last change of `.gitattributres`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 /tests export-ignore
 /tools export-ignore
+/.github export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
 .gitmodules export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,13 +1,18 @@
 /tests export-ignore
 /tools export-ignore
+/.github export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
 .gitmodules export-ignore
 .travis.yml export-ignore
+.scrutinizer.yml export-ignore
```
greg0ire, line 8:
```suggestion
```

I just removed that file in #8187

vladyslavstartsev: fixed

greg0ire (approved):

vladyslavstartsev:
@greg0ire can you please request review from doctrine orm team?

Issue #8193, opened 2020-06-24
Associated entities are persisted when no direct persist is called or cascade is not configured


phillipsnick:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 2.7.3

#### Summary

If associated entities fetched using lazy loading are modified, they will be persisted along with the parent entity even when no persist has been called on the lazy loaded entity or any cascade configuration has been set.

#### Current behavior

Taking the example where a User can have many Events

- Fetch a single User 
- Lazy load all Events
- Modify the User
- Modify the Event (noting that Doctrine will in some cases pick up modifications as discovered under https://github.com/doctrine/orm/issues/5594)
- Persist ONLY the User
- Flush
- The Event will have been persisted to the database too

Query log

```
'"START TRANSACTION"' Params:
'UPDATE users SET email = ? WHERE userId = ?' Params: 'updated@test.com', 1
'UPDATE events SET amount = ? WHERE eventId = ?' Params: 20.0, 1
'"COMMIT"' Params:
'SELECT t0.eventId AS eventId_1, t0.amount AS amount_2, t0.userId AS userId_3 FROM events t0 LIMIT 1' Params:
```

#### How to reproduce

Please see pull request here: https://github.com/doctrine/orm/pull/8194

Run test with

```bash
vendor/bin/phpunit --filter GH8193
```

#### Expected behavior

- Fetch a single User 
- Lazy load all Events
- Modify the User
- Modify the Event 
- Persist ONLY the User
- Flush
- Event should not have been updated

Expected query log

```
'"START TRANSACTION"' Params:
'UPDATE users SET email = ? WHERE userId = ?' Params: 'updated@test.com', 1
'"COMMIT"' Params:
'SELECT t0.eventId AS eventId_1, t0.amount AS amount_2, t0.userId AS userId_3 FROM events t0 LIMIT 1' Params:
```

beberlei:
Expected behavior, see PR comment

PR #8194, opened 2020-06-24
Associated entities are persisted when no direct persist is called or cascade is not configured #8193

phillipsnick:
If associated entities fetched using lazy loading are modified, they will be persisted along with the parent entity even when no persist has been called on the lazy loaded entity or any cascade configuration has been set.

As outlined under issuer here: https://github.com/doctrine/orm/issues/8193

Test run with

```bash
vendor/bin/phpunit --filter GH8193
```

### DIFF
--- a/tests/Doctrine/Tests/Models/GH8193/Event.php
+++ b/tests/Doctrine/Tests/Models/GH8193/Event.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Doctrine\Tests\Models\GH8193;
+
+/**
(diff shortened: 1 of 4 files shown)


beberlei:
This is the correct behavior, persist is only required to make new entities known to EntityManager. The flush operation will store all changes to all entities that are known to the unit of work. All loaded entities are known to UnitOfWork.

You can look into the following things:

- change tracking policy „persist“ does what you want
- read only entities

PR #7989, opened 2020-01-16
Fix issues detected by psalm on 2.7

orklah:
This PR is basically #7986 but targetted on 2.8 as suggested.

I'll copy the text from there:

> Hi,
> 
> This PR is the result of fixing easy issues from the two first levels of Psalm.
> 
> This is the starting point of trying to begin #7762. If this goes smoothly, my goal is to continue fixing errors by slowly increasing the Psalm's levels. Eventually, the idea would be to add Psalm to the CI and to add template annotation to interfaces and exposed classes to help tools like phpstan and psalm understand Doctrine without the need of stubs or plugins.
> 
> I'm always trying to learn so don't hesitate to ask for changes or anything!
> 
> Thanks!

I didn't expect such difference between master and 2.8! Great job guys, I can't wait for future versions!

It seems that the CI is broken on 2.8. I tried to run phpcbf locally but there are massive changes, I didn't dare apply them here, so there may be some CS errors. If the CI is fixed and the errors are still there in the future, I'll correct them.

I'll close #7986, some changes there were specific to the master branch due to code movements but let's fix that later.

PR #8199, opened 2020-07-05
add/fix more types (checked by psalm)

orklah:
Following #7989 and #8150 here's a little batch of changes in type documentation.

I mainly focused on parsing error detected by psalm (missing type, unparsable syntax, and false information). I also fixed some errors I made in #7989 

EDIT: I believe after this, Psalm can be upgraded to lvl 7

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -22,6 +22,9 @@ namespace Doctrine\ORM;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\DBAL\LockMode;
 use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Query\AST\DeleteStatement;
+use Doctrine\ORM\Query\AST\SelectStatement;
+use Doctrine\ORM\Query\AST\UpdateStatement;
 use Doctrine\ORM\Query\Exec\AbstractSqlExecutor;
 use Doctrine\ORM\Query\Parameter;
 use Doctrine\ORM\Query\ParameterTypeInferer;
@@ -213,9 +216,7 @@ final class Query extends AbstractQuery
     /**
      * Returns the corresponding AST for this DQL query.
      *
-     * @return \Doctrine\ORM\Query\AST\SelectStatement |
-     *         \Doctrine\ORM\Query\AST\UpdateStatement |
-     *         \Doctrine\ORM\Query\AST\DeleteStatement
+     * @return SelectStatement|UpdateStatement|DeleteStatement
      */
     public function getAST()
     {
@@ -364,11 +365,11 @@ final class Query extends AbstractQuery
     {
         $AST = $this->getAST();
 
-        if ($AST instanceof \Doctrine\ORM\Query\AST\SelectStatement) {
+        if ($AST instanceof SelectStatement) {
             throw new QueryException('The hint "HINT_CACHE_EVICT" is not valid for select statements.');
         }
 
-        $className = ($AST instanceof \Doctrine\ORM\Query\AST\DeleteStatement)
+        $className = $AST instanceof DeleteStatement
             ? $AST->deleteClause->abstractSchemaName
             : $AST->updateClause->abstractSchemaName;
 
--- a/lib/Doctrine/ORM/Query/TreeWalkerChain.php
+++ b/lib/Doctrine/ORM/Query/TreeWalkerChain.php
@@ -33,6 +33,7 @@ class TreeWalkerChain implements TreeWalker
      * The tree walkers.
      *
      * @var TreeWalker[]
+     * @psalm-var TreeWalkerChainIterator<TreeWalker>
      */
     private $_walkers;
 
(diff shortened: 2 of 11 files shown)


orklah:
Marking as draft while fixing CI

orklah:
Great, the only CI fail is because of https://github.com/doctrine/orm/pull/8199/files#diff-3aff39fbb5b16ba9be6df33f2fcde380R2951
Before this PR, the syntax did not allow parsing and one of there is a unknown class here. Should it be removed?

lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -506,7 +506,7 @@ public function setGenerateAnnotations($bool)
     /**
      * Sets the class fields visibility for the entity (can either be private or protected).
      *
-     * @param bool $visibility
+     * @param string $visibility
```
SenseException, line 509: Since only specific constants are allowed, would the use of the psalm enum be a fitting choice?
https://psalm.dev/docs/annotating_code/typing_in_psalm/#specifying-stringint-options-aka-enums

SenseException (changes requested):
> Before this PR, the syntax did not allow parsing and one of there is a unknown class here. Should it be removed?

It seems to be gone, so I suggest to remove it.

lib/Doctrine/ORM/Tools/EntityGenerator.php
orklah: Well, I wasn't sure how far to push it for this PR, but if you suggest it, Psalm allows that sort of syntax
`@psalm-param self::FIELD_VISIBLE_*` to restrict usage to existing properties following the pattern. I pushed the change

lib/Doctrine/ORM/Query.php
```diff
@@ -326,7 +326,7 @@ protected function _doExecute()
             $this->evictEntityCacheRegion();
         }
 
-        [$sqlParams, $types] = $this->processParameterMappings($paramMappings);
+        list($sqlParams, $types) = $this->processParameterMappings($paramMappings);
```
SenseException: Was psalm complaining about this? `[$sqlParams, $types] = $this->processParameterMappings($paramMappings);` works with PHP 7.1.

SenseException (changes requested):

lib/Doctrine/ORM/Query.php
orklah: No, but it was an unintended change. PhpStorm has a bug when refactoring a file, it replace those automatically.

If it's something we want to change, I can propose another PR to change every occurence

SenseException (dismissed):

lib/Doctrine/ORM/Query/TreeWalkerChain.php
```diff
@@ -33,6 +33,7 @@ class TreeWalkerChain implements TreeWalker
      * The tree walkers.
      *
      * @var TreeWalker[]
+     * @psalm-var TreeWalkerChainIterator<TreeWalker>
```
greg0ire, line 36: This looks repetitive. I think it would be better to use just `TreeWalkerChainIterator`, and use `class-implements Iterator<TreeWalker>` in `TreeWalkerChainIterator`

orklah: You're right. Actually, I have pretty much this in a changeset. I wasn't sure it was a good idea to bring up template for generics so soon :)

lib/Doctrine/ORM/Query/TreeWalkerChain.php
```diff
@@ -33,6 +33,7 @@ class TreeWalkerChain implements TreeWalker
      * The tree walkers.
      *
      * @var TreeWalker[]
+     * @psalm-var TreeWalkerChainIterator
```
greg0ire, line 36: :thinking: why not use `@var TreeWalkerChainIterator` here?

orklah: To maintain auto-completion in PHPStorm. If we put in `@var`, each foreach loop will result in a mixed object.
It's become a reflex for me when I add psalm properties that will cause this kind of issue but I can adapt if needed.

greg0ire: I've seen on Twitter that integration with Psalm and PHPStorm was coming and that it included using some annotations: https://twitter.com/psalmphp/status/1280848618895880197

greg0ire: Doesn't look like it's there yet though: https://github.com/JetBrains/phpstorm-psalm-plugin/issues/16, so I suppose it's fine to have both.

orklah: Yeah, they expect to include it in the EAP starting September. So it should be released as stable by the end of the year. And we will have to see if generics are included and work well. I'll gladly submit a PR then to remove excess information if everything goes well :)

beberlei (approved):

PR #8201, opened 2020-07-06
Throw Exception when unable to locate identifier

oojacoboo:
When UnitOfWork is unable to locate an entity by the `spl_object_hash`, we throw a custom Exception that includes the name of the entity that was unable to be located.

### DIFF
--- a/lib/Doctrine/ORM/EntityNotFoundException.php
+++ b/lib/Doctrine/ORM/EntityNotFoundException.php
@@ -21,6 +21,7 @@
 namespace Doctrine\ORM;
 
 use function implode;
+use function sprintf;
 
 /**
  * Exception thrown when a Proxy fails to retrieve an Entity result.
@@ -46,5 +47,18 @@ class EntityNotFoundException extends ORMException
         return new self(
             'Entity of type \'' . $className . '\'' . ($ids ? ' for IDs ' . implode(', ', $ids) : '') . ' was not found'
         );
+    }
+
+    /**
+     * Instance for which no identifier can be found
+     *
+     * @psalm-param class-string $className
+     */
+    public static function noIdentifierFound(string $className): self
+    {
+        return new self(sprintf(
+            'Unable to find "%s" entity identifier associated with the UnitOfWork',
+            $className
+        ));
     }
 }
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -3038,7 +3038,8 @@ class UnitOfWork implements PropertyChangedListener
      */
     public function getEntityIdentifier($entity)
     {
-        return $this->entityIdentifiers[spl_object_hash($entity)];
+        return $this->entityIdentifiers[spl_object_hash($entity)]
+            ?? EntityNotFoundException::noIdentifierFound(get_class($entity));
     }
 
     /**
--- a/tests/Doctrine/Tests/ORM/EntityNotFoundExceptionTest.php
+++ b/tests/Doctrine/Tests/ORM/EntityNotFoundExceptionTest.php
@@ -32,4 +32,12 @@ class EntityNotFoundExceptionTest extends TestCase
         $this->assertInstanceOf(EntityNotFoundException::class, $exception);
         $this->assertSame('Entity of type \'foo\' was not found', $exception->getMessage());
     }
+
+    public function testNoIdentifierFound(): void
+    {
+        $exception = EntityNotFoundException::noIdentifierFound('foo');
+
+        $this->assertInstanceOf(EntityNotFoundException::class, $exception);
+        $this->assertSame('Unable to find "foo" entity identifier associated with the UnitOfWork', $exception->getMessage());
+    }
 }


oojacoboo:
This is in reference to:
https://github.com/doctrine/orm/pull/8032

lib/Doctrine/ORM/EntityNotFoundException.php
```diff
@@ -48,4 +50,18 @@ public static function fromClassNameAndIdentifier($className, array $id)
...
         );
     }
+
+
+    /**
+     * Instance where no identifier can be found
+     *
+     * @param string $className
+     */
+    public static function noIdentifierFound($className)
```
greg0ire:
```suggestion
    public static function noIdentifierFound(string $className): self
```

lib/Doctrine/ORM/EntityNotFoundException.php
```diff
@@ -48,4 +50,18 @@ public static function fromClassNameAndIdentifier($className, array $id)
             'Entity of type \'' . $className . '\'' . ($ids ? ' for IDs ' . implode(', ', $ids) : '') . ' was not found'
         );
     }
+
+
+    /**
+     * Instance where no identifier can be found
+     *
+     * @param string $className
```
greg0ire: When using type declaration, this becomes unnecessary, let's change it for something that adds information.

```suggestion
     * @psalm-param class-string $className
```

lib/Doctrine/ORM/EntityNotFoundException.php
```diff
@@ -48,4 +50,18 @@ public static function fromClassNameAndIdentifier($className, array $id)
             'Entity of type \'' . $className . '\'' . ($ids ? ' for IDs ' . implode(', ', $ids) : '') . ' was not found'
         );
     }
+
+
+    /**
+     * Instance where no identifier can be found
```
greg0ire:
```suggestion
     * Instance for which no identifier can be found
```

tests/Doctrine/Tests/ORM/EntityNotFoundExceptionTest.php
```diff
@@ -30,4 +30,14 @@ public function testFromClassNameAndIdentifier()
         $this->assertInstanceOf(EntityNotFoundException::class, $exception);
         $this->assertSame('Entity of type \'foo\' was not found', $exception->getMessage());
     }
+
+    public function testNoIdentifierFound()
```
greg0ire:
```suggestion
    public function testNoIdentifierFound(): void
```

oojacoboo:
Accepted all suggestions, but it's not clear from the codebase that strict typing is being used.  Also, `EntityNotFoundException` does not `declare(strict_types = 1);`.  Is that intentional or not?

tests/Doctrine/Tests/ORM/EntityNotFoundExceptionTest.php
```diff
@@ -30,4 +30,14 @@ public function testFromClassNameAndIdentifier()
         $this->assertInstanceOf(EntityNotFoundException::class, $exception);
         $this->assertSame('Entity of type \'foo\' was not found', $exception->getMessage());
     }
+
+    public function testNoIdentifierFound() : void
+    {
+        $exception = EntityNotFoundException::noIdentifierFound(
+            'foo',
```
greg0ire, line 37:
```suggestion
            'foo'
```

tests/Doctrine/Tests/ORM/EntityNotFoundExceptionTest.php
```diff
@@ -30,4 +30,14 @@ public function testFromClassNameAndIdentifier()
...
+        );
```
greg0ire: Trailing commas in method calls require php 7.3, and there is no need for that linebreak anyway IMO
```suggestion
        $exception = EntityNotFoundException::noIdentifierFound('foo');
```

oojacoboo: I'm fine with this.  However, this project needs to make up it's mind on what version is being targeted.  It's extremely confused.  `composer.json` says 7.3+.  Most of the code doesn't use typing either.

SenseException: The branch you're targeting to has `^7.2` as constraint, so 7.3 features are not usable.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3018,7 +3019,8 @@ public function setOriginalEntityProperty($oid, $property, $value)
      */
     public function getEntityIdentifier($entity)
     {
-        return $this->entityIdentifiers[spl_object_hash($entity)];
+        return $this->entityIdentifiers[spl_object_hash($entity)]
```
SenseException, line 3041: Can you please create a test that will cover the exception part?

SenseException (changes requested):

lib/Doctrine/ORM/UnitOfWork.php
oojacoboo: Meaning what?  A test that will cause the Exception to be thrown?

SenseException: Yes. By making `getEntityIdentifier($entity)` fail with an entity that isn't in `$this->entityIdentifiers`

smilesrg: ping @oojacoboo

oojacoboo: @smilesrg I've added a test.  Trying to run `phpunit` locally failed with `Interface 'Doctrine\Persistence\NotifyPropertyChanged' not found`.  Sounds like a dependency to another repo not included as a vendor lib.

smilesrg: I can't run unit tests locally at all: phpunit reports `No tests executed!`

smilesrg: @oojacoboo try to rebase onto origin, then run `composer update` and then execute tests again

oojacoboo: Tests are passing now

SenseException:
Please also fix the Coding Standards, so that every check is green.

oojacoboo:
@SenseException fixed.  It doesn't help that tests don't run automatically on PRs now.

SenseException (dismissed):

greg0ire:
:thinking: this is not a bugfix, let's target 2.9.x.

greg0ire (approved):

greg0ire:
Thanks @oojacoboo !

PR #8203, opened 2020-07-07
Add SchemaIgnoreClasses property for #8195.

### LINKED ISSUE #8195: Schema exclude annotation
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

When using migrations or schema operations, there are facilities to ignore tables via schema_filter setting, see [Ignoring Custom Tables](https://www.doctrine-project.org/projects/doctrine-migrations/en/2.2/reference/generating-migrations.html#ignoring-custom-tables).

However there are many circumstances where entities should not be included for consideration in the schema. For example entities based on views, multiple entities for the same table or view for different sections of a project or serialisation requirements. This list can be quite long. Relying on excluding items just by the table name does not work for a lot of real world situations.

Currently there is no way to prevent a normal entity from being included in the schema list, this can cause schema failures and the inability to use the migration or schema tools in Doctrine.

I would like to propose adding a @SchemaExclude annotation which could be tested for in the SchemaTool::procesingNotRequired method and prevent the entity table from being added to the schema table list.

Many thanks.
Ian.

ianef:
Add schemaIgnoreClasses property, this list of entity classes is ignored by the schema tool when building a schema from metadata.

Fixes #8195

### DIFF
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -936,4 +936,28 @@ class Configuration extends \Doctrine\DBAL\Configuration
     {
         $this->_attributes['defaultQueryHints'][$name] = $value;
     }
+
+    /**
+     * Gets a list of entity class names to be ignored by the SchemaTool
+     *
+     * @since 2.8
+     *
+     * @return array
+     */
+    public function getSchemaIgnoreClasses() : array
+    {
+        return isset($this->_attributes['schemaIgnoreClasses']) ? $this->_attributes['schemaIgnoreClasses'] : [];
+    }
+    
+    /**
+     * Sets a list of entity class names to be ignored by the SchemaTool
+     *
+     * @since 2.8
+     *
+     * @param array $schemaIgnoreClasses Array of entity class names
+     */
+    public function setSchemaIgnoreClasses(array $schemaIgnoreClasses)
+    {
+        $this->_attributes['schemaIgnoreClasses'] = $schemaIgnoreClasses;
+    }
 }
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Configuration.php
```diff
@@ -936,4 +936,24 @@ public function setDefaultQueryHint($name, $value)
     {
         $this->_attributes['defaultQueryHints'][$name] = $value;
     }
+
+    /**
+     * Gets a list of entity class names to be ignored by the SchemaTool
+     *
+     * @return string[]
+     */
+    public function getSchemaIgnoreClasses() : array
```
beberlei, line 945: I would like to make the naming a bit more precise, otherwise everything else is great. What do you think of `get/setEntitiesIgnoredDuringSchemaGeneration`. I first thought it should be classes instead of entities, but embeddables and other non-entities are skipped already, so its really about skipping entities here.

ianef: OK, I'm happy with that. Apologies for the delay, just committed the changes here so fingers crossed.

mnkii:
This feature would be very handy - is it just waiting approval?

greg0ire:
@mnkii : 
- Some checks haven't completed yet
- This branch has conflicts that must be resolved

ianef:
I'll take a look in a few days and try and resolve the conflicts. I wasn't aware!

simPod:
Can I eventually help with it? Just run into the same need.

greg0ire:
@simPod you can sure open another PR targeting the right branch and fixing the conflicts 👍

simPod:
@greg0ire is it 2.10?

greg0ire:
Yes, and the method to determine that is to take a look at the latest tag, then drop the patch number.

derrabus:
Closing in favor of #9202.

ianef:
Thanks @simPod, I've been really busy recently. Much appreciated.

PR #8204, opened 2020-07-07
change list() syntax to array destructuring (short list syntax)

orklah:
After a quick discussion here: https://github.com/doctrine/orm/pull/8199#discussion_r450444924, I started a PR to propose changing this syntax. It's supported starting from PHP 7.1 which is the lowest requirement of ORM 2.7


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @orklah !

PR #8162, opened 2020-05-31
Add iterable support for IN criterias

l-vo:
Add support for iterables (including generators) for queries with IN/NOT IN criteria

PR #8208, opened 2020-07-11
Fix ambiguous case where an entity is also a Traversable

l-vo:
I've just realized that #8162 could cause a BC break when an entity passed as parameter is also traversable. This PR aims to fix that, trying to process a parameter as a Traversable only if the parameter isn't an entity.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\ORM;
 
+use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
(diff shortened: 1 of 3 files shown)


greg0ire:
Maintainers, do we usually assign a milestone to such bugfixes that fix unreleased features? Should it appear in the changelog?

greg0ire (dismissed):

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -430,11 +430,30 @@ public function processParameterValue($value)
...
 
         return $value;
     }
 
+    /**
+     * Process a parameter value which was previously identified as an array
+     *
+     * @param mixed[] $value
+     */
+    private function processArrayParameterValue(array &$value) : void
```
SenseException, line 449: Can you please pass this argument by value instead of by reference and return a value?

SenseException (changes requested):

lib/Doctrine/ORM/AbstractQuery.php
l-vo: Of course... Updated :)

SenseException (approved):

PR #8209, opened 2020-07-11
Fix psalm errors and upgrade strictness

orklah:
Hi,

This PR fix a new batch of issues reported by Psalm or PHPStorm.

It also upgrade the errorLevel from 8 to 7.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -72,6 +72,7 @@ abstract class AbstractQuery
      * The parameter map of this query.
      *
      * @var ArrayCollection|Parameter[]
+     * @psalm-var ArrayCollection<int, Parameter>
      */
     protected $parameters;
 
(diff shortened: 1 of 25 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -337,9 +338,11 @@ function (Query\Parameter $parameter) use ($key) : bool {
     /**
      * Sets a collection of query parameters.
      *
-     * @param ArrayCollection|mixed[] $parameters
+     * @param ArrayCollection|Parameter[] $parameters
```
greg0ire, line 341: This does not look right… on line 354, `$value` is could be a `Parameter` instance if I look at the right hand part of this phpdoc type, and I don't think `new Parameter($key, $value)` makes sense in that case.

greg0ire:
It's weird that github says you are a first-time contributor to this repository :thinking:

lib/Doctrine/ORM/AbstractQuery.php
orklah: You're right. I was confused by `ArrayCollection|mixed[]` that I assumed was a way to express `ArrayCollection<mixed>` and then, I saw that it was assigned to `$this->parameters` who is indeed an `ArrayCollection<Parameter>`. In fact, the param is `ArrayCollection<Parameter>|mixed[]`. I'll fix this

orklah:
> It's weird that github says you are a first-time contributor to this repository 🤔

Maybe it's only counting the main branch?

greg0ire (approved):

SenseException (approved):

PR #8210, opened 2020-07-11
Property access on metadata Interface

orklah:
this PR replace invalid access to name property on the Metadata Interface. The property is actually defined in its implementation MetadataInfo. The replacement is getName() function that is defined in the interface so it's valid.

There are still some read access to name that can't be replace as easily that I will look into if this PR is well received. Also there are write access and more properties that will have to be replaced.

Note: this is a good chunk of issues in the next level of PHPStan and future levels of Psalm


orklah:
I'll close this for now. Will reopen once #8144 is done. There is too many issues with CS that are unrelated to the actual fix

PR #8216, opened 2020-07-22
Fix deleting an element with index 0 from select expression in query AST due to casting false to 0

hason:
@greg0ire If `identVariableExpression`doesn't exist in `$AST->selectClause->selectExpressions`, function `array_search` returns false. Then, the `unset` function deletes the invalid item (see https://3v4l.org/t1Ybf).

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -437,6 +437,10 @@ class Parser
             $expr = $this->identVariableExpressions[$dqlAlias];
             $key  = array_search($expr, $AST->selectClause->selectExpressions);
 
+            if ($key === false) {
+                continue;
+            }
+
             unset($AST->selectClause->selectExpressions[$key]);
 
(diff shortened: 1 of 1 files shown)


greg0ire:
Please describe the potential bug in your commit message. Where would the cast happen?

greg0ire:
@hason the same with more emphasis:

> Please describe the potential bug in your _commit message_

hason:
@greg0ire I have changed the commit message.

greg0ire:
Are you aware commit messages can have a body? You are not forced to put everything on one line.
Read more on this in [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches), then please improve your commit message to explain the bug. Right now it's all condensed in one sentence and it is still hard to understand (although it's better than before). Adding a test would also help understanding in what situation this can happen.

SenseException:
Please also add a test that covers your fix and that would break if your newly added lines of code are removed.

ostrolucky:
We need a test indeed.

PR #8228, opened 2020-08-07
Update SchemaTool.php

krimchenko:
Added schema support for sequence

Now the sequence is created in the schema (namespace) of the owner table.


SenseException:
Please fix your pull request and cover your changes with a test.

Issue #8229, opened 2020-08-07
Joined table inheritance with AttributeOverride (rename) of the identifier column fails on insert


cziegenberg:
In the following example, the automatically created insert into the resource table fails when creating a new user, because the id field got another column name in the user table, using the AttributeOverride annotation:

// src/Entity/Resource.php
```
/**
 * @ORM\Entity()
 * @ORM\Table(name="app_resource")
 * @ORM\InheritanceType("JOINED")
 * @ORM\DiscriminatorColumn(name="resource_type", type="string", length=191)
 * @ORM\DiscriminatorMap({"resource"="Resource", "user"="User"})
 */
abstract class Resource
{
    /**
     * @ORM\Id()
     * @ORM\Column(name="resource_id", type="guid")
     */
    protected string $id;
}
```

// src/Entity/User.php
```
/**
 * @ORM\Entity()
 * @ORM\Table(name="app_user")
 * @ORM\AttributeOverrides({@ORM\AttributeOverride(name="id", column=@ORM\Column(name="user_id", type="guid"))})
 */
class User extends Resource
{
    //...
}
```

Error (meaning only one parameter instead of two given for the prepared statement, because the id value is not set):
```
  [Doctrine\DBAL\Driver\PDOException (08P01)]                                                                                      
  SQLSTATE[08P01]: <<Unknown error>>: 7 FEHLER:  Binden-Nachricht enthält 1 Parameter, aber vorbereitete Anweisung »« erfordert 2 
```

Also reading the entity, updating it and deleting it fails. I have a working fix for it, tested with the 2.8.x and the 2.7.x branch.

cziegenberg:
That's not enough. There are other parts effected by this (imho) bug, i.e. when creating joins to the root table... I will have a look into it, but I'm totally new to Doctrine, so it could take some time. Is it generally allowed to create a pull request for a 2.x branch for this type of fix or would this be treated as a new feature?

cziegenberg:
Also fixed in master. See pull requests.

greg0ire:
Reopening since it isn't fixed until your PR is merged.

rogregoire:
@ostrolucky @greg0ire : 
[...]

PR #8233, opened 2020-08-08
Fix for bug #8229 (id column from parent class renamed in child class)

cziegenberg:


greg0ire:
Closing as per https://github.com/doctrine/orm/pull/8235#issuecomment-670951626

PR #8234, opened 2020-08-08
 Fix for bug #8229 (id column from parent class renamed in child class)

### LINKED ISSUE #8229: Joined table inheritance with AttributeOverride (rename) of the identifier column fails on insert

cziegenberg:

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -37,6 +37,7 @@ use Doctrine\ORM\Query;
 use Doctrine\ORM\UnitOfWork;
 use Doctrine\ORM\Utility\IdentifierFlattener;
 use Doctrine\ORM\Utility\PersisterHelper;
+use function array_key_exists;
 use function array_map;
 use function array_merge;
 use function assert;
@@ -447,14 +448,26 @@ class BasicEntityPersister implements EntityPersister
             $types[]    = $this->columnTypes[$columnName];
         }
 
-        $where      = [];
-        $identifier = $this->em->getUnitOfWork()->getEntityIdentifier($entity);
+        $where                = [];
+        $identifier           = $this->em->getUnitOfWork()->getEntityIdentifier($entity);
+        $quotedClassTableName = $this->quoteStrategy->getTableName($this->class, $this->platform);
 
         foreach ($this->class->identifier as $idField) {
             if ( ! isset($this->class->associationMappings[$idField])) {
-                $params[]   = $identifier[$idField];
-                $types[]    = $this->class->fieldMappings[$idField]['type'];
-                $where[]    = $this->quoteStrategy->getColumnName($idField, $this->class, $this->platform);
+                $params[] = $identifier[$idField];
+                $types[]  = $this->class->fieldMappings[$idField]['type'];
+
+                // Fix for bug GH-8229 (id column from parent class renamed in child class):
+                // This method is called with the updated entity, but with different table names
+                // (the entity table name or a table name of an inherited entity). In dependence
+                // of the used table, the identifier name must be adjusted.
+                $class = $this->class;
+                if (isset($class->fieldMappings[$idField]['inherited']) && $quotedTableName !== $quotedClassTableName) {
+                    $className = $this->class->fieldMappings[$idField]['inherited'];
+                    $class     = $this->em->getClassMetadata($className);
+                }
+
+                $where[] = $this->quoteStrategy->getColumnName($idField, $class, $this->platform);
 
                 continue;
             }
@@ -632,7 +645,18 @@ class BasicEntityPersister implements EntityPersister
             $newVal = $change[1];
 
             if ( ! isset($this->class->associationMappings[$field])) {
-                $fieldMapping = $this->class->fieldMappings[$field];
+                $class = $this->class;
+
+                // Fix for bug GH-8229 (id column from parent class renamed in child class):
+                // Get the correct class metadata
+                foreach ($class->parentClasses as $parentClassName) {
+                    $parentClass = $this->em->getClassMetadata($parentClassName);
+                    if (array_key_exists($field, $parentClass->fieldMappings)) {
+                        $class = $parentClass;
+                    }
+                }
+
+                $fieldMapping = $class->fieldMappings[$field];
                 $columnName   = $fieldMapping['columnName'];
 
                 $this->columnTypes[$columnName] = $fieldMapping['type'];
@@ -1672,11 +1696,17 @@ class BasicEntityPersister implements EntityPersister
     private function getSelectConditionStatementColumnSQL($field, $assoc = null)
     {
         if (isset($this->class->fieldMappings[$field])) {
-            $className = (isset($this->class->fieldMappings[$field]['inherited']))
-                ? $this->class->fieldMappings[$field]['inherited']
-                : $this->class->name;
+            // Fix for bug GH-8229 (id column from parent class renamed in child class):
+            // Use the correct metadata and name for the id column
+            if (isset($this->class->fieldMappings[$field]['inherited'])) {
+                $className = $this->class->fieldMappings[$field]['inherited'];
+                $class     = $this->em->getClassMetadata($className);
+            } else {
+                $className = $this->class->name;
+                $class     = $this->class;
+            }
 
-            return [$this->getSQLTableAlias($className) . '.' . $this->quoteStrategy->getColumnName($field, $this->class, $this->platform)];
+            return [$this->getSQLTableAlias($className) . '.' . $this->quoteStrategy->getColumnName($field, $class, $this->platform)];
         }
 
         if (isset($this->class->associationMappings[$field])) {
--- a/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
@@ -26,6 +26,8 @@ use Doctrine\DBAL\Types\Type;
 
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\ORM\Utility\PersisterHelper;
+use function array_combine;
+use function array_reverse;
 
 /**
  * The joined subclass persister maps a single entity instance to several tables in the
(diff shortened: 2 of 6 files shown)


greg0ire:
Please associate your email address with your Github account, or change the

cziegenberg:
Done I think. Not sure about the e-mail - I had to convert my account to an organization last week and switched to a new account, but GitHub Desktop didn't update the name/e-mail after switching.

PS: Thanks for the rebase instruction. I also had to add `~` + [number of commits] to the first command to get it working.

cziegenberg:
Just recognized another bug related to this one. The foreign key for the id column in the child table is not set, if it's renamed via AttributeOverride. I think this is an ORM bug, not a migrations problem, right? Will look into it...

cziegenberg:
The foreign key generation is also fixed now (the "inherited" property of the column has been set correctly, but accidentally removed later, so that the foreign key was not created). After this I also had to fix the generation of the WHERE condition for the id column (which depends on the "inherited" property).

lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
```diff
@@ -282,7 +283,11 @@ public function delete($entity)
             $rootTable = $this->quoteStrategy->getTableName($rootClass, $this->platform);
             $rootTypes = $this->getClassIdentifiersTypes($rootClass);
 
-            return (bool) $this->conn->delete($rootTable, $id, $rootTypes);
+            // Fix for bug GH-8229 (id column from parent class renamed in child class):
+            // Use the correct name for the id column as named in the root class.
+            $rootId = array_combine($rootClass->getIdentifierColumnNames(), $identifier);
+
+            return (bool) $this->conn->delete($rootTable, $rootId, $rootTypes);
```
ostrolucky, line 290: Why similar line at L297 still uses $id, is that correct?

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1676,7 +1696,14 @@ private function getSelectConditionStatementColumnSQL($field, $assoc = null)
                 ? $this->class->fieldMappings[$field]['inherited']
                 : $this->class->name;
 
-            return [$this->getSQLTableAlias($className) . '.' . $this->quoteStrategy->getColumnName($field, $this->class, $this->platform)];
+            // Fix for bug GH-8229 (id column from parent class renamed in child class):
+            // Use the correct name for the id columns as named in the inherited class.
+            $class = $this->class;
```
ostrolucky, line 1701: WDYT?
```suggestion
            $nonInheritedClass = $this->class;
            if ($className !== $nonInheritedClass->name) {
```

Doing `$className !== $this->class->name` is kinda confusing

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -632,12 +633,31 @@ protected function prepareUpdateData($entity)
...
+                    }
+                }
+
+                $fieldMapping = $class->fieldMappings[$field];
                 $columnName   = $fieldMapping['columnName'];
 
                 $this->columnTypes[$columnName] = $fieldMapping['type'];
 
+                // Fix for bug GH-8229 (id column from parent class renamed in child class):
+                // The getOwningTable() method is overwritten in class JoinedSubclassPersister
```
ostrolucky, line 653: Should it be fixed there, then? Current approach of fixing it means people using getOwningTable standalone will still have old, buggy behaviour, right? These fixes are for some reason localized to few methods only.

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -248,14 +248,23 @@ function (ClassMetadata $class) use ($idMapping) : bool {
                     }
 
                     if ( ! empty($inheritedKeyColumns)) {
+                        // Fix for bug GH-8229 (id column from parent class renamed in child class):
+                        // Use the correct name for the id columns as named in the parent class.
+                        $rootClass                 = $this->em->getClassMetadata($class->rootEntityName);
+                        $rootIdentifierColumnNames = $rootClass->getIdentifierColumnNames();
+                        $rootIdentifierColumns     = [];
+                        foreach ($inheritedKeyColumns as $index => $inheritedKeyColumn) {
```
ostrolucky, line 256: Can we use `array_intersect_key` here?

lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
cziegenberg: Yes, that's correct. The variable naming is a bit confusing, because it's also called $rootTable, although it isn't the root table, it's the current one (see comment "starting from this class' table up to the root table"), so $id is correct in this case. I didn't want to unnecessarily change existing code, but could rename it of course, to make it clearer.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
cziegenberg: You're right, that's confusing. I will optimize this part and merge it with the condition in the lines above.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
cziegenberg: That's right. I will modify it and move this part to an own method that will always be called by getOwningTable().

lib/Doctrine/ORM/Tools/SchemaTool.php
cziegenberg: Yes, I will change it.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
cziegenberg: Just realized that this isn't necessary anymore, because the correction of the "inherited" property in the column metadata (added alter to fix missing foreign keys) also fixed that problem, so I wll simply remove this part. :)

cziegenberg:
@ostrolucky I implemented all suggested changes.

ostrolucky (changes requested):

lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
```diff
@@ -296,7 +301,11 @@ public function delete($entity)
             $parentTable    = $this->quoteStrategy->getTableName($parentMetadata, $this->platform);
             $parentTypes    = $this->getClassIdentifiersTypes($parentMetadata);
 
-            $this->conn->delete($parentTable, $id, $parentTypes);
+            // Fix for bug GH-8229 (id column from parent class renamed in child class):
+            // Use the correct name for the id column as named in the parent class.
+            $parentId = array_combine($parentMetadata->getIdentifierColumnNames(), $identifier);
```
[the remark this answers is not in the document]
cziegenberg: That depends on the platform you use. If you use a database with foreign key support you're right, then you won't get there. If you use SQLite instead, the test would fail there.

IMHO the whole block starting in line 281 ("if ($this->platform->supportsForeignKeyConstraints()) {") should be removed, because this is a potential bug: If a platform supports foreign keys in general (or you decide to support foreign keys for a platform later), the entries are only deleted from the parent table, assuming that the foreign keys are really set and handle the deletion of the entries from the child table, but it's never checked if they exist. So if the foreign key is missing the entry in the child table will never be deleted (and perhabs result in other errors).

Removing this prevent this potential error, and the test would always fail without the change.  

WDYT?

lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
```diff
@@ -604,8 +615,11 @@ private function getJoinSql($baseTableAlias)
             $tableAlias  = $this->getSQLTableAlias($subClassName);
             $joinSql    .= ' LEFT JOIN ' . $this->quoteStrategy->getTableName($subClass, $this->platform) . ' ' . $tableAlias . ' ON ';
 
-            foreach ($identifierColumn as $idColumn) {
-                $conditions[] = $baseTableAlias . '.' . $idColumn . ' = ' . $tableAlias . '.' . $idColumn;
+            // Fix for bug GH-8229 (id column from parent class renamed in child class):
+            // Use the correct name for the id column as named in the parent class.
+            $subClassIdentifierColumn = $subClass->getIdentifierColumnNames();
```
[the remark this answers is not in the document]
cziegenberg: You're right. I will extend the test.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -632,7 +633,18 @@ protected function prepareUpdateData($entity)
...
             if ( ! isset($this->class->associationMappings[$field])) {
-                $fieldMapping = $this->class->fieldMappings[$field];
+                $class = $this->class;
+
+                // Fix for bug GH-8229 (id column from parent class renamed in child class):
+                // Get the correct class metadata
+                foreach ($class->parentClasses as $parentClassName) {
+                    $parentClass = $this->em->getClassMetadata($parentClassName);
+                    if (array_key_exists($field, $parentClass->fieldMappings)) {
+                        $class = $parentClass;
```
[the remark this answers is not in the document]
cziegenberg: You're right. I extended the test for this case.

lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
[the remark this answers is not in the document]
cziegenberg: There was another potential bug: When a foreign key was set but not expected, because not supported by DBAL, i.e. when you use an existing SQLite database - SQLite supports foreign keys, by DBAL doesn't (for good reason I think, because not updateable). In this case the deletion was done in wrong order (child -> root) and would have failed.

I refactored this method to remove the previous duplicate code and the resulting duplicate bug fix as well as the two potential bugs.

lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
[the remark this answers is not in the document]
cziegenberg: Done.

cziegenberg:
Just realized an error in the annotations after the last change. Will checke that. Sorry.

cziegenberg:
I fixed the wrong annotation, and I also added a test for the possible foreign key problem when deleting entries from the tables of a joined inheritance entity.

ostrolucky (dismissed):
Nice work! :) I'm not comfortable to merge this on my own though, let's see if somebody else will pop in. In case nobody will, ping me in a few weeks and we will proceed anyways.

cziegenberg:
Hmm, there seems to be a problem with Travis CI:

Fatal error: Uncaught InvalidArgumentException: Extension class "PhpBench\Extensions\Dbal\DbalExtension" does not exist in phar:///home/travis/build/doctrine/orm/phpbench.phar/vendor/phpbench/container/lib/Container.php:54

greg0ire:
Blocked by #8252

cziegenberg:
Thanks for looking into the build problem... 

Latest changes:
- Fixed a problem with the SQL generation from DQL (wrong columns used in the SELECT, the JOIN and the ORDER BY part)

cziegenberg:
I just fixed another special case with wrong column names when creating a pessimistic write lock.

ostrolucky (approved):

ostrolucky:
Alright I see still nobody merged this, so let's do it. Thanks for your effort :)

beberlei:
I just stumbled over this and i am considering to revert this change, because i am unsure about the consequences of such a radical change in a patch version, and because the Java Persistence API doesn't allow AttributeOverride to be used on STI or JTI. I am unsure about the consequences of this change and it should be broader discussed and not be part of a patch release.

Unrelated, the change to `JoinedSubclassPersister::delete` is a behavior change, it looks like a good idea, but is also not acceptable in a patch version.

cziegenberg:
Thanks @beberlei for reviewing this change.

Following the Doctrine documentation, the attribute overrides feature can be used to override all attribute settings, explicitly including the name (see the [provided example](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/inheritance-mapping.html)). The feature existed before, but it didnt't work in all cases. So I cannot see a behavior change here, but a bugfix.

---

I checked the change in JoinedSubclassPersister::delete() again and I will try to explain it:

Before:
1) When foreign keys were supported by DBAL:
Delete the entry from the root table (and let the delete cascade).
2) When foreign keys were not supported by DBAL:
Delete all entries separately (the order of deletion didn't matter, because no support for foreign keys).
3) Bug: When foreign keys are supported by DBAL but missing in the DB:
The delete in 1) succeeded, but wasn't cascaded (some entries remained).

After:
1) The first delete from the root table cascades, when a foreign key exists.
2) If the Nth delete didn't cascade, the (N+1)th delete removes the related entries. If a cascade happended before, the Nth delete doesn't do anything. I changed the order, because I tried to keep the old behavior, but avoid duplicate code (and because the order of deletion didn't matter anyway in this case).
3) Bug fixed.

So IMHO it's also just a bugfix and no behavior change, BUT I think, that this can be improved, because I missed another potential bug (which also existed before):
Doctrine ORM always expects foreign keys to be cascading here, but what if the foreign key action is changed to "no action" or "restrict"? The the delete would fail (and it would have falied in the previous version).

So I'd suggest to fix this as follows:
Do not rely on any expected, unverified foreign key or foreign key action, but delete all entries in the order child to root separately.

beberlei:
@cziegenberg ah thanks for linking the documentation it actually says this is only working with mapped superclasses:

> Used to override a mapping for an entity field or relationship. May be applied to an entity that extends a mapped superclass to override a relationship or field mapping defined by the mapped superclass.

As for the JTI and delete. It was the previous behavior that a foreign key with delete cascade had to be configured. For 10 years I cannot remember a single discussion or bug related to people expecting this to work different. Your approach is a bit safer, but its also inefficient, because it can potentially lead to increase number of DELETE statements.

cziegenberg:
Yes, additional delete statements that IMHO should perform the same (because in the end they do the same), but in a safer way.

ostrolucky:
Documentation also provides example where attribute overrides are used without mapped superclass https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/tutorials/override-field-association-mappings-in-subclasses.html

beberlei:
Just to clarify, this documentation example is for a trait, which is copied into the same entity, so it is not using inheritance either.

@cziegenberg Attribute overrides for regular field mappings in JTI are broken though. You cannot query these fields across the class structure anymore and if the column is not nullable on the parent class, then it breaks during inserting. "It works" for me just means it accidently does what it should in 50% of the use cases, but we forgot to prevent it.

cziegenberg:
I think what @ostrolucky wanted to say is that a mapped superclass isn't always required, while you said it's "only working with mapped superclasses". This doesn't mean other types of inheritance were planned to be supported, but that the documentation isn't that clear.

I also misunderstood the documentation, and yes, it worked for me (after the fix) and it passed all tests (also for other use cases), so from my point of view everything was fine. You have an overview of all the use cases and features you want to support. If you say it's too risky or causes problems with other use cases that cannot be fixed, then you have to revert the changes (and prevent the wrong usage of this feature). I have to accept this.

PR #8235, opened 2020-08-08
 Fix for bug doctrine#8229 (id column from parent class renamed in child class)

cziegenberg:

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -621,6 +621,15 @@ class BasicEntityPersister implements EntityPersister
                     break;
 
                 case $property instanceof LocalColumnMetadata:
+                    // Fix for bug GH-8229 (id column from parent class renamed in child class):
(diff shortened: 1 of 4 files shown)


cziegenberg:
Fixed all coding standard issues, but the check still reports errors in a file which is not part of this pull request?!

greg0ire:
You don't need to open one PR per branch, this is a bugfix, and as such should go on 2.7.

cziegenberg:
Sorry, it's my first week with Doctrine and my first fix.

My problem was that your guidelines require a commit to the master, but 1) it wasn't clear if it's treated as a bugfix or feature (because it never worked?) and 2) in the master branch the fix works much different (due to identifier quoting etc.).

Nevertheless I learned a lot about Doctrine and the changes in future versions. ;)

greg0ire:
> My problem was that your guidelines require a commit to the master

Ah maybe this needs to be fixed? We had a backport workflow in the past, and we are trying to go towards merge up… Anyway, let's try to get it merge on 2.7, and we'll see about other branches then. In all cases, to me it's a bugfix because as you pointed out, it doesn't work.

PR #8236, opened 2020-08-08
Use more accurate terminology

greg0ire:
I think this was a mistake when writing this documentation, and that the
original author meant to use attribute here, columns do not have
columns, they have attributes.

Saw this while trying to understand https://github.com/doctrine/orm/pull/8234


SenseException (approved):
Thank you. But did you mean another issue/PR? You're referencing this one.

greg0ire:
Haha yeah I messed up, thanks for pointing that out, I fixed it.

PR #8237, opened 2020-08-10
Add two new events: initializePersistentCollection and initializeProxy

danydev:
This pull request adds two new events so that they can be observed from userland. The two new events are
- `initializeProxy`: called each time a proxy is initialized
- `initializePersistentCollection`: called each time a persistent collection is initialized.

The use case behind this PR is to provide an hook to be able to detect n+1. See [n-plus-one-detector](https://github.com/danydev/doctrine-n-plus-one-detector)

### DIFF
--- a/lib/Doctrine/ORM/Event/InitializePersistentCollectionEventArgs.php
+++ b/lib/Doctrine/ORM/Event/InitializePersistentCollectionEventArgs.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Event;
+
+use Doctrine\Common\EventArgs;
+use Doctrine\ORM\PersistentCollection;
+
+/**
+ * Provides event arguments for the initializePersistentCollection event.
+ */
+class InitializePersistentCollectionEventArgs extends EventArgs
+{
+    /** @var PersistentCollection */
+    private $collection;
+
+    public function __construct(PersistentCollection $collection)
+    {
+        $this->collection = $collection;
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


Ocramius:
Considering the amount of misuse of such listeners, I am very much against increasing the capabilities of such deep internal ORM functionality.

This should not become an extension point, and should be retained as a purely internal detail.

danydev:
So, do you think that would make sense to integrate an opt-in for an 'n plus one detector' directly in the orm? If so, I can give it a try

Ocramius:
It would make sense to add a logging endpoint (PSR-3, perhaps), but not an event trigger

danydev:
Cool, I was thinking about something inspectable at any given time (so exposing a new class with a method like `getNPlusOne()`), so that user could do something with it, maybe just log it, or maybe enriching the log (with a route name for example), or even aggregate it somehow (e.g. log only if I did not log this info within 7 day). If we log directly from the orm we would loose a bit of flexibility.

I can do a PR in the next days, and we can start from there. Thanks for your feedback

beberlei:
Thank you for your contribution, but it is unlikely we are going to merge this, especially when the use-case is so narrow. 

You can see N+1 queries by using an SQL Logger and inspecting the backtrace already, or by using a Profiler and looking at the execution graphs.

danydev:
I agree that the event way is not going to be merged, but I think we could do something directly in the ORM itself maybe. I'll try to propose something when I can dedicate some more time.
I don't think sql logger + backtrace way could work, too much performance hog, also profiler is not the answer since ideally such a tool should be used in production env and not only in dev.

Issue #8249, opened 2020-08-26
PHPBench: use of DBAL storage extension


dantleech:
Hey

I was going to remove the DBAL extension from PHPBench for 1.0:

- it's not fully implemented.
- it's complicated.

I notice that the ORM is using it:

- Does it provide any value to this project over the default XML storage driver? (i'd imagine the file size is smaller with sqlite f.e.)
- If so, is it worth splitting to a separate repo?

Ocramius:
> Does it provide any value to this project over the default XML storage driver? (i'd imagine the file size is smaller with sqlite f.e.)

No, not really: XML works just fine for this sort of project.

 > If so, is it worth splitting to a separate repo?

If people need it, they will open issues, and then the pre-1.0 code can be recovered and split off.

dantleech:
Great, then I'll just remove it for now.

PR #8251, opened 2020-08-27
Resolving relationships with interface and Abstract

### LINKED ISSUE #7670: Resolving target entities from abstract class seems to not working as expected
Hello, according to the [How to Define Relationships with Abstract Classes and Interfaces](https://symfony.com/doc/current/doctrine/resolve_target_entity.html), I should be able to use the `ResolveTargetEntityListener` to resolve an abstract class to a specific entity.

**However, this "feature" seems to not working properly.**

## 1. Using an interface
Here is the PHP code used to resolve the interface to the entity class;
```php
$registry
    ->getManager()
    ->getMetadataFactory()
    ->getMetadataFor(RedirectRouteInterface::class)
    ->getName();
```
> **$registry** is an instance of `\Doctrine\Common\Persistence\ManagerRegistry`

When configuring doctrine using an interface like:
```yaml
doctrine:
    orm:
        resolve_target_entities:
            Symfony\Cmf\Component\Routing\RedirectRouteInterface: App\Entity\RedirectRoute
```
everything is working perfectly and the name is equals to `App\Entity\RedirectRoute`

## 2. Using an abstract class
Now when I'm trying to resolve an entity from an abstract class, the resolver seems to not working properly.
Here is the PHP code used to resolve the abstract class to the entity class:
```php
$registry
    ->getManager()
    ->getMetadataFactory()
    ->getMetadataFor(AbstractRedirectRoute::class)
    ->getName();
```
> **$registry** is an instance of `\Doctrine\Common\Persistence\ManagerRegistry`

When configuring doctrine using an abstract class like:
```yaml
doctrine:
    orm:
        resolve_target_entities:
            App\Model\AbstractRedirectRoute: App\Entity\RedirectRoute
```
In that case, the resolver is returning the abstract class FQCN: `App\Model\AbstractRedirectRoute`.

Weird thing is, to be able to resolve the abstract class I need to make the call to `getAllMetadata()` method like:
```php
$registry
    ->getManager()
    ->getMetadataFactory()
    ->getAllMetadata();
$registry
    ->getManager()
    ->getMetadataFactory()
    ->getMetadataFor(AbstractRedirectRoute::class)
    ->getName();
```
In that case, the entity is resolved correctly!

---
**Is this an issue or more a miss-configuration on my side?**

Zul3s:
Fixes #7670

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -22,6 +22,8 @@ namespace Doctrine\ORM\Tools;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\DBAL\Types\Type;
+use function class_exists;
+use function interface_exists;
 
 /**
  * Performs strict validation of the mapping schema
@@ -97,7 +99,12 @@ class SchemaValidator
         }
 
         foreach ($class->associationMappings as $fieldName => $assoc) {
-            if (!class_exists($assoc['targetEntity']) || $cmf->isTransient($assoc['targetEntity'])) {
+            /**
+             * Resolving target entities from Interface or Abstract ex. resolve_target_entities
+             */
+            $targetEntity = isset($cmf->getLoadedMetadata()[$assoc['targetEntity']])
+                ? $cmf->getLoadedMetadata()[$assoc['targetEntity']]->name : $assoc['targetEntity'];
+            if ((! class_exists($assoc['targetEntity']) && ! interface_exists($assoc['targetEntity'])) || $cmf->isTransient($targetEntity)) {
                 $ce[] = "The target entity '" . $assoc['targetEntity'] . "' specified on " . $class->name . '#' . $fieldName . ' is unknown or not an entity.';
 
                 return $ce;


kuronana (dismissed):

Zul3s:
Hi, sorry, but what can i do to fix CI problem ?

greg0ire:
The failure looks related to #8249

Zul3s:
Ok thank you. So if i understand good, i can do nothing, just waiting :) .

greg0ire (commented):
Please add some tests

Zul3s:
Hi, 
I am sorry, but it's realy hard to  make tests with sense. 
The problem case is with Symfony validation schema (toolbar and command) and Bundle Entities. I can not understand why, but bundle entites is not in associationMappings on \Doctrine\ORM\Tools\ResolveTargetEntityListener::loadClassMetadata. So the remapAssociation will be not execute. In my tests, remapAssociation is always execute. 

The problem is not present on Doctrine Migration (ex. doctrine:migration:diff nothing to do). 

Have you an advise for me plz ? 

Without fix 
image
With fix 
image

greg0ire:
> Have you an advise for me plz ?

Yes: show us the tests you have tried to write, maybe we will be able to understand more what makes the tests hard to write.

PR #8253, opened 2020-08-28
Add schema support for sequence.

krimchenko:
Now the sequence is created in the schema (namespace) of the owner table.


krimchenko:
Wrong way.

PR #8259, opened 2020-09-04
WIP: Parse Traversable into array in for Collection fields

thiagocordeiro:


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -642,6 +643,11 @@ public function computeChangeSet(ClassMetadata $class, $entity)
                     $value = new ArrayCollection($value->getValues());
                 }
 
+                // If $value is not a Collection and is a Traversable then use an ArrayCollection.
+                if ( ! $value instanceof Collection && $value instanceof \Traversable) {
```
thiagocordeiro, line 647:
```suggestion
                if ( ! $value instanceof Collection && $value instanceof Traversable) {
```

PR #8260, opened 2020-09-04
Improve insert performance with batched insert queries

mvorisek:
Non-versioned not-post incremented inserts can be grouped together and performed with much less query count.

Based on my measurements batched inserts are about 50% faster with local DB and several times faster with DB on another machine due less total RTTs (round trip times) needed. This change largely increases real-world performace with larger imports.

The provided implementation is fully back compatible.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -37,11 +37,17 @@ use Doctrine\ORM\Query;
 use Doctrine\ORM\UnitOfWork;
 use Doctrine\ORM\Utility\IdentifierFlattener;
 use Doctrine\ORM\Utility\PersisterHelper;
+use function array_chunk;
+use function array_fill;
 use function array_key_exists;
 use function array_map;
 use function array_merge;
+use function array_unique;
 use function assert;
+use function count;
+use function implode;
 use function reset;
+use function sprintf;
 
 /**
  * A BasicEntityPersister maps an entity to a single table in a relational database.
@@ -162,14 +168,6 @@ class BasicEntityPersister implements EntityPersister
     protected $quotedColumns = [];
 
     /**
-     * The INSERT SQL statement used for entities handled by this persister.
-     * This SQL is only generated once per request, if at all.
-     *
-     * @var string
-     */
-    private $insertSql;
-
-    /**
      * The quote strategy.
      *
      * @var \Doctrine\ORM\Mapping\QuoteStrategy
@@ -273,38 +271,43 @@ class BasicEntityPersister implements EntityPersister
         $stmt       = $this->conn->prepare($this->getInsertSQL());
         $tableName  = $this->class->getTableName();
 
-        foreach ($this->queuedInserts as $entity) {
-            $insertData = $this->prepareInsertData($entity);
+        $canGroup = ! $isPostInsertId && ! $this->class->isVersioned;
 
-            if (isset($insertData[$tableName])) {
-                $paramIndex = 1;
+        foreach (array_chunk($this->queuedInserts, $canGroup ? 50 : 1) as $queuedInsertsChunk) {
+            $stmt = $this->conn->prepare($this->getInsertSQL(count($queuedInsertsChunk)));
 
-                foreach ($insertData[$tableName] as $column => $value) {
-                    $stmt->bindValue($paramIndex++, $value, $this->columnTypes[$column]);
+            $paramIndex = 1;
+            foreach ($queuedInsertsChunk as $entity) {
+                $insertData = $this->prepareInsertData($entity);
+
+                if (isset($insertData[$tableName])) {
+                    foreach ($insertData[$tableName] as $column => $value) {
+                        $stmt->bindValue($paramIndex++, $value, $this->columnTypes[$column]);
+                    }
                 }
             }
 
             $stmt->execute();
 
-            if ($isPostInsertId) {
-                $generatedId = $idGenerator->generate($this->em, $entity);
-                $id = [
-                    $this->class->identifier[0] => $generatedId
-                ];
-                $postInsertIds[] = [
-                    'generatedId' => $generatedId,
-                    'entity' => $entity,
-                ];
-            } else {
-                $id = $this->class->getIdentifierValues($entity);
+            foreach ($queuedInsertsChunk as $entity) {
+                if ($isPostInsertId) {
+                    $generatedId     = $idGenerator->generate($this->em, $entity);
+                    $id              = [$this->class->identifier[0] => $generatedId];
+                    $postInsertIds[] = [
+                        'generatedId' => $generatedId,
+                        'entity' => $entity,
+                    ];
+                } else {
+                    $id = $this->class->getIdentifierValues($entity);
+                }
+
+                if ($this->class->isVersioned) {
+                    $this->assignDefaultVersionValue($entity, $id);
+                }
             }
 
-            if ($this->class->isVersioned) {
-                $this->assignDefaultVersionValue($entity, $id);
-            }
+            $stmt->closeCursor();
         }
-
-        $stmt->closeCursor();
         $this->queuedInserts = [];
 
         return $postInsertIds;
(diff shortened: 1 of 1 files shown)


mvorisek:
ping @greg0ire, @SenseException can you please review/merge?

greg0ire:
I don't think you should target 2.7, this does not qualify as a bugfix.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -273,38 +271,41 @@ public function executeInserts()
         $stmt       = $this->conn->prepare($this->getInsertSQL());
         $tableName  = $this->class->getTableName();
 
-        foreach ($this->queuedInserts as $entity) {
-            $insertData = $this->prepareInsertData($entity);
+        foreach (array_chunk($this->queuedInserts, $this->getMaxBatchedInsertQueries()) as $queuedInsertsChunk) {
```
greg0ire, line 274: As soon as you do that, you now have the contents of `$this->queuedInserts` twice in memory. Wouldn't it be better to extract the first `$this->getMaxBatchedInsertQueries()` elements out of that array with `array_slice()` until it is empty? I see it will be emptied anyway after the loop, on line 308.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1402,27 +1403,34 @@ protected function getSelectManyToManyJoinSQL(array $manyToMany)
...
+
     /**
      * {@inheritdoc}
      */
-    public function getInsertSQL()
+    public function getInsertSQL(int $recordsCount = 1)
     {
-        if ($this->insertSql !== null) {
-            return $this->insertSql;
-        }
```
greg0ire, line 1388: So now if we call this method several times in a row and there is no batching, we will re-compute the insertSQL every time?

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1402,27 +1403,34 @@ protected function getSelectManyToManyJoinSQL(array $manyToMany)
...
+    public function getMaxBatchedInsertQueries()
+    {
+        $idGenerator    = $this->class->idGenerator;
+        $isPostInsertId = $idGenerator->isPostInsertGenerator();
+
+        if ($isPostInsertId || $this->class->isVersioned) {
+            return 1;
+        }
+
+        return 50;
```
greg0ire, line 1415: Magic number! Why 50? Shouldn't it be configurable?

greg0ire (commented):
Shouldn't there be extra tests, and documentation?

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
mvorisek: I think this is ok, as values in php are not duplicated if not modified (aka copy on modify)

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
mvorisek: This is correct, I was not able to measure any significant performance difference. Also, other queries like UPDATE and DELETE also rebuild the query from scratch every time.

ostrolucky (changes requested):
Not a fan of doing this logic in ORM directly. Batch insert should be contributed to DBAL, only then used by ORM. We don't even know if this syntax is supported by all DBMSes.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
mvorisek: Yes, it should. Please advise how.

beberlei:
This needs to target 2.8, and the batch size should be globally configurable via Configuration class

beberlei:
@guilhermeblanco does this work with goals for Next?

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
greg0ire: Doesn't look ok to me: https://3v4l.org/1LYhg
`array_chunk` does not create a simple copy, it creates another array entirely

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
greg0ire: I would use a constructor argument

lcobucci:
I agree with @ostrolucky here. There was an initial work on DBAL. We should look into solving things there IMHO.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
SenseException: It would be helpful if you can provide the method how you run the performance tests to see any improvements. If this PR gets merged, we will need a way to prevent regressions in the future that will drop any performance improvements of this.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
SenseException: I agree here. Any fixed number may perform differently depending on the original amount of entities that need to be processed.

SenseException (changes requested):
If there is a big gain of performance by this PR, I would like to see how it was calculated/measured and if this measurement can be used by CI to avoid performance regressions.

ostrolucky:
> Can Doctrine/ORM/Persisters/Entity/BasicEntityPersister::getInsertSQL() method be dropped in 2.8.x then?

Logic for creating SQL query can be dropped from method, not drop whole method.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
guilhermeblanco: This information should be set by a new annotation `@BatchSize(size = 100)` that is configured at the entity level.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
guilhermeblanco: This information should be set by a new annotation @BatchSize(size = 100) that is configured at the entity level.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1378,27 +1379,34 @@ protected function getSelectManyToManyJoinSQL(array $manyToMany)
         return ' INNER JOIN ' . $joinTableName . ' ON ' . implode(' AND ', $conditions);
     }
 
+    public function getMaxBatchedInsertQueries()
+    {
+        $idGenerator    = $this->class->idGenerator;
+        $isPostInsertId = $idGenerator->isPostInsertGenerator();
+
+        if ($isPostInsertId || $this->class->isVersioned) {
+            return 1;
```
guilhermeblanco, line 1388: We can start naively here with 1, but we could be smarter depending of post generated identifier. If using auto incremented, certain RDBMS (such as MySQL and MariaDB, very likely PgSQL too) guarantee the ordering of incremental identifiers on a per transaction basis. This means in case of concurrent writes, the first transaction would be 1 - X as ID incrementally, and second transaction would generate X+1 - Y, also incrementally.

But as an initial implementation, this is more than enough.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
guilhermeblanco: I completely agree that full insert SQL cache is not possible, however, we should be able to store columns and placeholders to prevent us from recomputing every time. I recall a 2s slowdown in the whole testing suite by dropping out this cache.
Actually, now that I think about it... we should still store the full insert SQL referring to a single entry. Also, we should store the column count, so we can easily build new sets of placeholders.

Now for update and delete, the reason why they are not cached is because we're not doing full updates. If we were to be more inline with EclipseLink, Hibernate and EntityFramework, we would do full updates (where you update all columns, excluding LOB fields) and also cache this computation. For delete, id based removals would be ok too...
But this is another micro-optimization that could be done in a separate PR.

guilhermeblanco:
> @guilhermeblanco does this work with goals for Next?

Yes, this is one of the goals of ORM 3. The insert logic could be either defined linearly as we do today (using strings) or using SQLStatement value objects like `Statement\Insert`, `Statement\Update` and `Statement\Delete`.

I remember talking about this with Steve Müller (deeky666, not @ 'ing him) in the past, as most of this logic could be ideally moved to DBAL, but very likely it was lost on removal of old develop branch there.

Fedik:
In my opinion this should be done on application level, and on a library level it is unneeded complication.
I often use similar approach for batch import ("chunk" insert / update), but in regular routine it does not make much sense.

mvorisek:
I plan to look into this.

PR #8266, opened 2020-09-13
[GH-8265] Prototype for Attribute Metadata Driver

### LINKED ISSUE #8265: Add New Attribute Metadata Driver
For Doctrine ORM 2.8 we should add a metadata driver using Attributes, based off the AnnotationDriver with some adjustment to account for the missing nested support.

Idea:
- Re-use `Doctrine\ORM\Mapping\*` annotation classes and add `#[Attribute]` as needed.
- Flatten `@JoinTable` by introducing new `InverseJoinColumn` attribute
- Flatten `@Table` by allowing `Index` and `UniqueConstraint` on the entity/class level.
- Drop support for named queries and named native queries in Attribute based mapping driver due to their nesting complexity.
- No support for association overrides in Attribute based mapping driver.

beberlei:
Support for Attribute Metadata Driver with PHP 8+

Note: For now the `AttributeDriver` does not support attribute overrides, association overrides, named queries, named native queries and sql resultset mapping, because of missing nested attributes support. This is not a big problem, because these are quite edge case features some of which we have thought about removing anyways. As for the overwrites, we should think of a solution in the near future.

A few other deviations from the Annotation based Driver:
- Instead of `@JoinTable` holding join columns and inverse join columns as a property, you use `#[ORM\JoinColumn]` and `#[ORM\InverseJoinColumn]` on the top level.
- Instead of `@Table` holding indexes and unique constraints, these are also attributes on the top level of a declaration.

Todos:
- [x] Actually run this against a php 8 build
- [x] Adjust tests to make this running
- [x] Adjust annotation / attribute classes to make use of named annotation support in doctrine/annotations v1.11
- [x] Add reference section in the docs
- [x] Integrate with https://github.com/doctrine/annotations/pull/402

Fixes #8265


lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Doctrine\ORM\Mapping\Driver;
+
+use Doctrine\Common\Annotations\Reader;
+use Doctrine\ORM\Mapping\Annotation;
+
+// TODO: Should we move this to doctrine/annotations?
```
SenseException: It doesn't look like this class has any hard relations to ORM at the moment, so it seems to be a good candidate for Annotations. Your improvements will have to coordinate with the Annotations repository and package.

TomasVotruba:
There is related PR in Symfony annotations class reader. I share it here, because it might bring some inspiration or path to find united way with parsing and working with attributes for both Symfony and Dotrine.

> I have modified the existing AnnotationClassLoader so that it merges attributes and annotations. This way, an application could transparently switch from annotations to attributes. Moreover, the AnnotationClassLoader does not require an annotation reader anymore.

The file https://github.com/symfony/symfony/pull/37474/files#diff-90cbecfd4505e6e8af5842db11add3b5

beberlei:
@TomasVotruba i don't believe it makes sense to integrate AnnotationClassLoader with attributes for more than a few simple cases, because attributes will not be 100% compatible with annotations. Tieing them together makes everything complicated instead.

TomasVotruba:
@beberlei Thanks for sharing. I haven't really digged into it, as the migration path is my only focus here. It would be better to commen to the author there, if you find it useful.

lib/Doctrine/ORM/Mapping/AttributeOverride.php
```diff
@@ -19,6 +19,8 @@
 
 namespace Doctrine\ORM\Mapping;
 
+use Attribute;
```
beberlei, line 23: @carusogabriel do you have a plan or a workaround for me how to get this working for now? How would we handle this in PHPCS?

carusogabriel: No, sorry, I haven't been looking at that. After PHP 8.0's release, I might have some time. For now, I'd just put this file on `phpcs.xml` ignoring whatever rule is causing it, with a comment "waiting for PHP 8.0 compatibility" :)

SenseException (dismissed):

lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php
```diff
@@ -0,0 +1,96 @@
...
+use Doctrine\ORM\Mapping\Annotation;
+use ReflectionAttribute;
+use ReflectionClass;
+use ReflectionMethod;
+use ReflectionProperty;
+
+use function count;
+use function is_subclass_of;
+
+class AttributeReader
```
malarzm: I haven't grasped these changes fully yet, but will this class be useful to ODM? Or is it only an implementation detail and convenient way for ORM to read annotations?

beberlei: It depends on how much work you want to put into `AttributeDriver`. I copied the code over from `AnnotationDriver` and the way it was setup made it more convenient to use this helper class here. Essentially a smaller version of this can be embedded into `AttributeDriver` after some more refactoring. For example the fact if an attribute is repeatable or not is known, so the driver code could be changed to be explicit.

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+
+.. _annref_column:
+
+#[Column]
+~~~~~~~~~
+
+Marks an annotated instance variable as "persistent". It has to be
+inside the instance variables PHP DocBlock comment. Any value hold
+inside this variable will be saved to and loaded from the database
+as part of the lifecycle of the instance variables entity-class.
```
greg0ire, line 60: I don't understand that last bit :sweat_smile:

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+
+Required attributes:
+
+-  **type**: Name of the Doctrine Type which is converted between PHP
```
greg0ire:
```suggestion
-  **type**: Name of the DBAL Type which does the conversion between PHP
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+   and Database representation.
+
+Optional attributes:
+
+-  **name**: By default the property name is used for the database
+   column name also, however the 'name' attribute allows you to
```
greg0ire:
```suggestion
   column name also, however the ``name`` attribute allows you to
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+   determine the column name.
+
+-  **length**: Used by the "string" type to determine its maximum
```
greg0ire:
```suggestion
-  **length**: Used by the ``string`` type to determine its maximum
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+
+-  **scale**: The scale for a decimal (exact numeric) column (applies
+   only for decimal column), which represents the number of digits
+   to the right of the decimal point and must not be greater than
+   *precision*.
+
+-  **unique**: Boolean value to determine if the value of the column
+   should be unique across all rows of the underlying entities table.
+
+-  **nullable**: Determines if NULL values allowed for this column. If not specified, default value is false.
```
greg0ire:
```suggestion
-  **nullable**: Determines if NULL values allowed for this column.
    If not specified, default value is ``false``.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+      all vendors).
+
+   -  ``fixed``: Boolean value to determine if the specified length of
+      a string column should be fixed or varying (applies only for
+      string/binary column and might not be supported by all vendors).
+
+   -  ``comment``: The comment of the column in the schema (might not
+      be supported by all vendors).
+
+   -  ``collation``: The collation of the column (only supported by Drizzle, Mysql, PostgreSQL>=9.1, Sqlite and SQLServer).
```
greg0ire: Support for Drizzle has been dropped from the DBAL. Also PostgreSQL 9.1 is very old, so there's probably no need to be this precise.
```suggestion
   -  ``collation``: The collation of the column (only supported by Mysql, PostgreSQL, Sqlite and SQLServer).
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+
+   -  ``check``: Adds a check constraint type to the column (might not
+      be supported by all vendors).
+
+-  **columnDefinition**: DDL SQL snippet that starts after the column
+   name and specifies the complete (non-portable!) column definition.
+   This attribute allows to make use of advanced RMDBS features.
+   However you should make careful use of this feature and the
+   consequences. SchemaTool will not detect changes on the column correctly
```
greg0ire:
```suggestion
   consequences. ``SchemaTool`` will not detect changes on the column correctly
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+   anymore if you use "columnDefinition".
```
greg0ire:
```suggestion
   anymore if you use ``columnDefinition``.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+
+   Additionally you should remember that the "type"
```
greg0ire:
```suggestion
   Additionally you should remember that the ``type``
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+    #[Column(type: "string", length: 32, unique: true, nullable: false)]
+    protected $username;
+
+    #[Column(type: "string", columnDefinition: "CHAR(2) NOT NULL")]
+    protected $country;
+
+    #[Column(type: "decimal", precision: 2, scale: 1)]
+    protected $height;
+
+    #[Column(type: "string", length: 2, options: ["fixed" => true, "comment" => "Initial letters of first and last name"])]
```
greg0ire:
```suggestion
    #[Column(type: "string", length: 2, options: [
        "fixed" => true,
        "comment" => "Initial letters of first and last name"
    ])]
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+    protected $initials;
+
+    #[Column(type: "integer", name: "login_count", nullable: false, options: ["unsigned" => true, "default" => 0])]
```
greg0ire:
```suggestion
    #[Column(
        type: "integer",
        name: "login_count",
        nullable: false,
        options: ["unsigned" => true, "default" => 0]
    )]
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+-  **usage**: One of ``READ_ONLY``, ``READ_WRITE`` or ``NONSTRICT_READ_WRITE``, By default this is ``READ_ONLY``.
+-  **region**: An specific region name
+
+.. _annref_changetrackingpolicy:
+
+#[ChangeTrackingPolicy]
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The Change Tracking Policy attribute allows to specify how the
+Doctrine ORM UnitOfWork should detect changes in properties of
```
greg0ire:
```suggestion
Doctrine ORM ``UnitOfWork`` should detect changes in properties of
```
or unit of work

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+entities during flush. By default each entity is checked according
+to a deferred implicit strategy, which means upon flush UnitOfWork
```
greg0ire:
```suggestion
to a deferred implicit strategy, which means upon flush ``UnitOfWork``
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+    #[GeneratedValue(strategy: "CUSTOM")]
+    #[CustomIdGenerator(class: MyIdGenerator::class)]
+    public $id;
+
+.. _annref_discriminatorcolumn:
+
+#[DiscriminatorColumn]
+~~~~~~~~~~~~~~~~~~~~~~
+
+This attribute is an optional and set on the root entity
```
greg0ire:
```suggestion
This attribute is optional and set on the root entity
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+    class User
+    {
+        #[Embedded(class: Address::class)]
+        private $address;
+
+
+.. _annref_embedded:
+
+#[Embedded]
+~~~~~~~~~~~~~~~~~~~~~
```
greg0ire:
```suggestion
~~~~~~~~~~~
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+#[Entity]
+~~~~~~~~~
+
+Required attribute to mark a PHP class as an entity. Doctrine manages
+the persistence of all classes marked as entities.
+
+Optional attributes:
+
+-  **repositoryClass**: Specifies the FQCN of a subclass of the
+   EntityRepository. Use of repositories for entities is encouraged to keep
```
greg0ire:
```suggestion
   ``EntityRepository``. Use of repositories for entities is encouraged to keep
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+
+#[GeneratedValue]
+~~~~~~~~~~~~~~~~~
+
+Specifies which strategy is used for identifier generation for an
+instance variable which is annotated by :ref:`#[Id] <annref_id>`. This
+attribute is optional and only has meaning when used in
+conjunction with #[Id].
+
+If this attribute is not specified with #[Id] the NONE strategy is
```
greg0ire:
```suggestion
If this attribute is not specified with ``#[Id]`` the ``NONE`` strategy is
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+used as default.
+
+Optional attributes:
+
+-  **strategy**: Set the name of the identifier generation strategy.
+   Valid values are AUTO, SEQUENCE, TABLE, IDENTITY, UUID, CUSTOM and NONE.
```
greg0ire:
```suggestion
   Valid values are ``AUTO``, ``SEQUENCE``, ``TABLE``, ``IDENTITY``,
   ``UUID``, ``CUSTOM`` and ``NONE``.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+   If not specified, default value is AUTO.
```
greg0ire:
```suggestion
   If not specified, the default value is ``AUTO``.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+
+#[HasLifecycleCallbacks]
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+This attribute has to be set on the entity-class to
+notify Doctrine that this entity has entity lifecycle callback
+attributes set on at least one of its methods. Using #[PostLoad],
+#[PrePersist], #[PostPersist], #[PreRemove], #[PostRemove], #[PreUpdate] or
+#[PostUpdate] without this marker attribute will make Doctrine
+ignore the callbacks.
```
greg0ire:
```suggestion
``#[PrePersist]``, ``#[PostPersist]``, ``#[PreRemove]``, ``#[PostRemove]``,
``#[PreUpdate]`` or ``#[PostUpdate]`` without this marker attribute will
make Doctrine ignore the callbacks.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+    }
+
+.. _annref_index:
+
+#[Index]
+~~~~~~~~
+
+Attribute is used on the entity-class level. It provides a hint to the SchemaTool to
+generate a database index on the specified table columns. It only
+has meaning in the SchemaTool schema generation context.
```
greg0ire:
```suggestion
has meaning in the ``SchemaTool`` schema generation context.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+#[JoinColumn], #[InverseJoinColumn]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This attribute is used in the context of relations in
+:ref:`#[ManyToOne] <annref_manytoone>`, :ref:`#[OneToOne] <annref_onetoone>` fields
+and in the Context of a :ref:`#[ManyToMany] <annref_manytomany>`. If this attribute or both *name* and *referencedColumnName*
+are missing they will be computed considering the field's name and the current
+:doc:`naming strategy <namingstrategy>`.
+
+The #[InverseJoinColumn] is the same as #[JoinColumn] and is used in the context
```
greg0ire:
```suggestion
The ``#[InverseJoinColumn]`` is the same as ``#[JoinColumn]`` and is used in the context
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+of a #[ManyToMany] attribute declaration to specifiy the details of the join table's
```
greg0ire:
```suggestion
of a ``#[ManyToMany]`` attribute declaration to specifiy the details of the join table's
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+column information used for the join to the inverse entity.
+
+Optional attributes:
+
+-  **name**: Column name that holds the foreign key identifier for
+   this relation. In the context of @JoinTable it specifies the column
+   name in the join table.
+-  **referencedColumnName**: Name of the primary key identifier that
+   is used for joining of this relation. Defaults to *id*.
```
greg0ire:
```suggestion
   is used for joining of this relation. Defaults to ``id``.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+:doc:`naming strategy <namingstrategy>`.
+
+The #[InverseJoinColumn] is the same as #[JoinColumn] and is used in the context
+of a #[ManyToMany] attribute declaration to specifiy the details of the join table's
+column information used for the join to the inverse entity.
+
+Optional attributes:
+
+-  **name**: Column name that holds the foreign key identifier for
+   this relation. In the context of @JoinTable it specifies the column
```
greg0ire:
```suggestion
   this relation. In the context of ``#[JoinTable]`` it specifies the column
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+-  **unique**: Determines whether this relation is exclusive between the
+   affected entities and should be enforced as such on the database
+   constraint level. Defaults to false.
+-  **nullable**: Determine whether the related entity is required, or if
+   null is an allowed state for the relation. Defaults to true.
+-  **onDelete**: Cascade Action (Database-level)
+-  **columnDefinition**: DDL SQL snippet that starts after the column
+   name and specifies the complete (non-portable!) column definition.
+   This attribute enables the use of advanced RMDBS features. Using
+   this attribute on @JoinColumn is necessary if you need slightly
```
greg0ire:
```suggestion
   this attribute on ``#[JoinColumn]`` is necessary if you need slightly
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+   different column definitions for joining columns, for example
+   regarding NULL/NOT NULL defaults. However by default a
+   "columnDefinition" attribute on :ref:`#[Column] <annref_column>` also sets
+   the related #[JoinColumn]'s columnDefinition. This is necessary to
```
greg0ire:
```suggestion
   the related ``#[JoinColumn]``'s columnDefinition. This is necessary to
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+.. _annref_jointable:
+
+#[JoinTable]
+~~~~~~~~~~~~
+
+Using
+:ref:`#[ManytoMany] <annref_manytomany>` on the owning side of the relation
+requires to specify the #[JoinTable] attribute which describes the
+details of the database join table. If you do not specify
+#[JoinTable] on these relations reasonable mapping defaults apply
```
greg0ire:
```suggestion
``#[JoinTable]`` on these relations reasonable mapping defaults apply
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+using the affected table and the column names.
+
+A notable difference to the annotation metadata support, #[JoinColumn]
```
greg0ire:
```suggestion
A notable difference to the annotation metadata support, ``#[JoinColumn]``
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+and #[InverseJoinColumn] are specified at the property level and are not
```
greg0ire:
```suggestion
and ``#[InverseJoinColumn]`` are specified at the property level and are not
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+nested within the #[JoinTable] attribute.
```
greg0ire:
```suggestion
nested within the ``#[JoinTable]`` attribute.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+Optional attributes:
+
+
+-  **mappedBy**: This option specifies the property name on the
+   targetEntity that is the owning side of this relation. It is a
+   required attribute for the inverse side of a relationship.
+-  **inversedBy**: The inversedBy attribute designates the field in the
+   entity that is the inverse side of the relationship.
+-  **cascade**: Cascade Option
+-  **fetch**: One of LAZY, EXTRA_LAZY or EAGER
```
greg0ire:
```suggestion
-  **fetch**: One of ``LAZY``, ``EXTRA_LAZY`` or ``EAGER``
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+-  **indexBy**: Index the collection by a field on the target entity.
+
+.. note::
+
+    For ManyToMany bidirectional relationships either side may
```
greg0ire:
```suggestion
    For ``ManyToMany`` bidirectional relationships either side may
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+    be the owning side (the side that defines the @JoinTable and/or
```
greg0ire:
```suggestion
    be the owning side (the side that defines the ``#[JoinTable]`` and/or
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+
+#[MappedSuperclass]
+~~~~~~~~~~~~~~~~~~~
+
+A mapped superclass is an abstract or concrete class that provides
+persistent entity state and mapping information for its subclasses,
+but which is not itself an entity. This attribute is specified on
+the Class level and has no additional settings.
+
+The #[MappedSuperclass] attribute cannot be used in conjunction with
```
greg0ire:
```suggestion
The ``#[MappedSuperclass]`` attribute cannot be used in conjunction with
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+#[Entity]. See the Inheritance Mapping section for
```
greg0ire:
```suggestion
``#[Entity]``. See the Inheritance Mapping section for
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+    {
+        // ... fields and methods
+    }
+
+.. _annref_onetoone:
+
+#[OneToOne]
+~~~~~~~~~~~
+
+The #[OneToOne] attribute works almost exactly as the
```
greg0ire:
```suggestion
The ``#[OneToOne]`` attribute works almost exactly as the
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+Example:
+
+.. code-block:: php
+
+    <?php
+    #[ManyToMany(targetEntity: "Group")]
+    #[OrderBy(["name" => "ASC"])]
+    private $groups;
+
+The key in OrderBy is only allowed to consist of
```
greg0ire:
```suggestion
The key in ``OrderBy`` is only allowed to consist of
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+unqualified, unquoted field names and of an optional ASC/DESC
```
greg0ire:
```suggestion
unqualified, unquoted field names and of an optional ``ASC``/``DESC``
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+positional statement. Multiple Fields are separated by a comma (,).
+The referenced field names have to exist on the ``targetEntity``
+class of the ``#[ManyToMany]`` or ``#[OneToMany]`` attribute.
+
+.. _annref_postload:
+
+#[PostLoad]
+~~~~~~~~~~~~~~
+
+Marks a method on the entity to be called as a #[PostLoad] event.
```
greg0ire:
```suggestion
Marks a method on the entity to be called as a ``#[PostLoad]`` event.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+Only works with #[HasLifecycleCallbacks] in the entity class PHP
```
greg0ire:
```suggestion
Only works with ``#[HasLifecycleCallbacks]`` in the entity class PHP
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+level.
+
+.. _annref_postpersist:
+
+#[PostPersist]
+~~~~~~~~~~~~~~
+
+Marks a method on the entity to be called as a #[PostPersist] event.
```
greg0ire:
```suggestion
Marks a method on the entity to be called as a ``#[PostPersist]`` event.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+Only works with #[HasLifecycleCallbacks] in the entity class PHP
```
greg0ire:
```suggestion
Only works with ``#[HasLifecycleCallbacks]`` in the entity class PHP
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+level.
+
+.. _annref_postremove:
+
+#[PostRemove]
+~~~~~~~~~~~~~~
+
+Marks a method on the entity to be called as a #[PostRemove] event.
```
greg0ire:
```suggestion
Marks a method on the entity to be called as a ``#[PostRemove]`` event.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+Only works with #[HasLifecycleCallbacks] in the entity class PHP
```
greg0ire:
```suggestion
Only works with ``#[HasLifecycleCallbacks]`` in the entity class PHP
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+level.
+
+.. _annref_preupdate:
+
+#[PreUpdate]
+~~~~~~~~~~~~~~
+
+Marks a method on the entity to be called as a #[PreUpdate] event.
+Only works with #[HasLifecycleCallbacks] in the entity class PHP
+level.
```
greg0ire:
```suggestion
Marks a method on the entity to be called as a ``#[PostUpdate]`` event.
Only works with ``#[HasLifecycleCallbacks]`` in the entity class PHP
level.

.. _annref_prepersist:

#[PrePersist]
~~~~~~~~~~~~~~

Marks a method on the entity to be called as a ``#[PrePersist]`` event.
Only works with ``#[HasLifecycleCallbacks]`` in the entity class PHP
level.

.. _annref_preremove:

#[PreRemove]
~~~~~~~~~~~~~~

Marks a method on the entity to be called as a ``#[PreRemove]`` event.
Only works with ``#[HasLifecycleCallbacks]`` in the entity class PHP
level.

.. _annref_preupdate:

#[PreUpdate]
~~~~~~~~~~~~~~

Marks a method on the entity to be called as a ``#[PreUpdate]`` event.
Only works with ``#[HasLifecycleCallbacks]`` in the entity class PHP
level.
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+
+.. _annref_sequencegenerator:
+
+#[SequenceGenerator]
+~~~~~~~~~~~~~~~~~~~~~
+
+For use with #[GeneratedValue(strategy: "SEQUENCE")] this
```
greg0ire:
```suggestion
For use with ``#[GeneratedValue(strategy: "SEQUENCE")]`` this
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+    #[Table(name: "user", schema: "schema_name")]
+    class User { }
+
+.. _annref_uniqueconstraint:
+
+#[UniqueConstraint]
+~~~~~~~~~~~~~~~~~~~
+
+Attribute is used on
+the entity-class level. It allows to hint the SchemaTool to
```
greg0ire:
```suggestion
the entity-class level. It allows to hint the ``SchemaTool`` to
```

docs/en/reference/attributes-reference.rst
```diff
@@ -0,0 +1,1023 @@
...
+        ChangeTrackingPolicy("NOTIFY")
+    ]
+    class User {}
+
+.. _annref_customidgenerator:
+
+#[CustomIdGenerator]
+~~~~~~~~~~~~~~~~~~~~
+
+This attribute allows you to specify a user-provided class to generate identifiers. This attribute only works when both :ref:`@Id <annref_id>` and :ref:`@GeneratedValue(strategy="CUSTOM") <annref_generatedvalue>` are specified.
```
greg0ire: Needs to be converted

greg0ire (commented):
Just a docs review for now, it's big enough as is :stuck_out_tongue:

walva:
Good luck guys ! You rock !

docs/en/reference/attributes-reference.rst
beberlei: This is copied from the annotation docs :)

lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php
greg0ire: Why is `Reader` not explicitly implemented here? Is it because `*Annotations` methods are going to be renamed?

lib/Doctrine/ORM/Mapping/InverseJoinColumn.php
```diff
@@ -0,0 +1,72 @@
...
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\ORM\Mapping;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
+final class InverseJoinColumn implements Annotation
```
greg0ire, line 28: :thinking: why is there a new "annotation"?

beberlei: I use this for filtering in `if (! is_subclass_of($attribute->getName(), Annotation::class)) {` in `AnnotationReader`.

lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php
beberlei: Because its not supposed to be used with Annotations, its just the same API because I copy adapted the code from AnnotationDriver for AttributesDriver

greg0ire (approved):

albe:
Great work on getting native attribute support into doctrine! @beberlei we (neos/flow framework) however now suffer from a slight change in expectations for some annotation classes - namely the `ManyToOne`, `ManyToMany` and `Embedded` now have a required first constructor argument. While those arguments were always considered "required" in a pure doctrine environment, we partially enhanced the DX by extracting some information for the annotations from other reflection data (like the `targetEntity`/`class` the annotation targets). Hence we have usage of doctrine annotations like:
```
@var SomeEntity
@ORM\ManyToMany
```
which now breaks, because the ManyToMany is attempted to be constructed with `null` as `targetEntity` in the DocParser. Until now we solved the "missing" targetEntity (and other things) via a custom `AnnotationDriver` and that worked out quite nicely. With this change we are at a loss though, because we'd need to hook into the DocParser now, which happens way too early.

See https://github.com/neos/flow-development-collection/issues/2487#issuecomment-850405128

While we can "solve" this by restricting doctrine/orm to <2.9, we would also like to move towards PHP8 support and attributes. For us the best solution would be if those three constructors would allow `null` for their first argument, but I understand that would mean you'd need to check for that `null` at runtime which is a step backward. Optimally the Annotation would not be instanciated until it is used for mapping, but that would be too great a change.
Are there any other possible ways to "enhance" the annotations and hook into the instanciation process?

PR #8277, opened 2020-09-20
Fix OrderByItem parser to proceed with SimpleArithmetic expression before function

### LINKED ISSUE #7626: Failed parse the condition in ORDER BY
### Bug Report

I want [calculate the number of occurrences of a char in string](http://bookofzeus.com/articles/mysql/count-the-number-of-occurrences-of-a-char-in-mysql/). But DQL is failed parse to SQL.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

### Summary

```
$ composer show --latest 'doctrine/*'
doctrine/annotations                v1.6.0  v1.6.0  Docblock Annotations Parser
doctrine/cache                      v1.8.0  v1.8.0  Caching library offering an object-oriented API for many cache backends
doctrine/collections                v1.5.0  v1.5.0  Collections Abstraction library
doctrine/common                     v2.10.0 v2.10.0 PHP Doctrine Common project is a library that provides additional functionality that other Doctrine p...
doctrine/dbal                       v2.9.0  v2.9.2  Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection a...
doctrine/doctrine-bundle            1.10.0  1.10.2  Symfony DoctrineBundle
doctrine/doctrine-cache-bundle      1.3.5   1.3.5   Symfony Bundle for Doctrine Cache
doctrine/doctrine-migrations-bundle v1.3.2  v2.0.0  Symfony DoctrineMigrationsBundle
doctrine/event-manager              v1.0.0  v1.0.0  Doctrine Event Manager component
doctrine/inflector                  v1.3.0  v1.3.0  Common String Manipulations with regard to casing and singular/plural rules.
doctrine/instantiator               1.1.0   1.1.0   A small, lightweight utility to instantiate objects in PHP without invoking their constructors
doctrine/lexer                      v1.0.1  v1.0.1  Base library for a lexer that can be used in Top-Down, Recursive Descent Parsers.
doctrine/migrations                 v1.8.1  v2.0.0  Database Schema migrations using Doctrine DBAL
doctrine/orm                        v2.6.3  v2.6.3  Object-Relational-Mapper for PHP
doctrine/persistence                v1.1.0  v1.1.0  The Doctrine Persistence project is a set of shared interfaces and functionality that the different D...
doctrine/reflection                 v1.0.0  v1.0.0  Doctrine Reflection component
```

### How to reproduce

**DQL**

```sql
SELECT (LENGTH(e.path) - LENGTH(REPLACE(e.path, :path_separator, :empty_string))) s FROM Category e
```

**Result SQL**

```sql
SELECT (LENGTH(e.path) - LENGTH(REPLACE(e.path, ".", ""))) AS s FROM shop__category e
```

**DQL**

```sql
SELECT LENGTH(e.path) - LENGTH(REPLACE(e.path, :path_separator, :empty_string)) FROM Category e
```

**Result SQL**

```sql
SELECT LENGTH(e.path) - LENGTH(REPLACE(e.path, ".", "")) FROM shop__category e
```

**DQL**

```sql
SELECT e FROM Category e ORDER BY LENGTH(e.path) - LENGTH(REPLACE(e.path, :path_separator, :empty_string)) ASC
```

**Current behavior**

```
[Syntax Error] line 0, col 82: Error: Expected end of string, got '-'
```

**Expected SQL**

```sql
SELECT e.* FROM shop__category e ORDER BY LENGTH(e.path) - LENGTH(REPLACE(e.path, ".", "")) ASC
```

**DQL**

```sql
SELECT e FROM Category e ORDER BY (LENGTH(e.path) - LENGTH(REPLACE(e.path, :path_separator, :empty_string))) ASC
```

**Current behavior**

```
[Syntax Error] line 0, col 99: Error: Expected Doctrine\ORM\Query\Lexer::T_IDENTIFIER, got '('
```

**Expected SQL**

```sql
SELECT e.* FROM shop__category e ORDER BY (LENGTH(e.path) - LENGTH(REPLACE(e.path, ".", ""))) ASC
```

ipernet:
Hi,

Fix #7626 

There is parser bug for `OrderByItem()`, where a **function** is detected before a **SimpleArithmeticExpression** while it should be the reverse: simple arithmetic expressions can start with a **function** too, and thus they should be matched first if present.

This fix enables the use of expression in sorting and with window functions.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -1510,10 +1510,6 @@ class Parser
         $glimpse = $this->lexer->glimpse();
 
         switch (true) {
-            case ($this->isFunction()):
-                $expr = $this->FunctionDeclaration();
-                break;
-
             case ($this->isMathOperator($peek)):
                 $expr = $this->SimpleArithmeticExpression();
                 break;
@@ -1524,6 +1520,10 @@ class Parser
 
             case ($this->lexer->peek() && $this->isMathOperator($this->peekBeyondClosingParenthesis())):
                 $expr = $this->ScalarExpression();
+                break;
+
+            case $this->isFunction():
+                $expr = $this->FunctionDeclaration();
                 break;
 
             default:
--- a/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
@@ -281,11 +281,20 @@ class SelectSqlGenerationTest extends OrmTestCase
             'SELECT f0_.id AS id_0, f0_.username AS username_1 FROM forum_users f0_ ORDER BY f0_.id ASC'
         );
     }
+
     public function testSupportsOrderByDesc()
     {
         $this->assertSqlGeneration(
             'SELECT u FROM Doctrine\Tests\Models\Forum\ForumUser u ORDER BY u.id desc',
             'SELECT f0_.id AS id_0, f0_.username AS username_1 FROM forum_users f0_ ORDER BY f0_.id DESC'
+        );
+    }
+
+    public function testSupportsOrderByWithSimpleArithmeticExpression() : void
+    {
+        $this->assertSqlGeneration(
+            'SELECT u FROM Doctrine\Tests\Models\Forum\ForumUser u ORDER BY LENGTH(u.username) + LENGTH(u.username) asc',
+            'SELECT f0_.id AS id_0, f0_.username AS username_1 FROM forum_users f0_ ORDER BY LENGTH(f0_.username) + LENGTH(f0_.username) ASC'
         );
     }
 


greg0ire:

Also, please associate your email address with your Github account, or change the

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -1510,22 +1510,22 @@ public function OrderByItem()
         $glimpse = $this->lexer->glimpse();
 
         switch (true) {
-            case ($this->isFunction()):
-                $expr = $this->FunctionDeclaration();
-                break;
-
-            case ($this->isMathOperator($peek)):
+            case $this->isMathOperator($peek):
```
greg0ire, line 1513: This and the change below should be one separate commit, since they have nothing to do with the issue at hand.

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
```diff
@@ -289,6 +290,22 @@ public function testSupportsOrderByDesc()
         );
     }
 
+    public function testSupportsOrderByStartingWithFunction()
```
greg0ire, line 293:
```suggestion
    public function testSupportsOrderByStartingWithFunction() : void
```

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
```diff
@@ -289,6 +290,22 @@ public function testSupportsOrderByDesc()
...
+    {
+        $this->assertSqlGeneration(
+            'SELECT u FROM Doctrine\Tests\Models\Forum\ForumUser u ORDER BY LENGTH(u.username) asc',
+            'SELECT f0_.id AS id_0, f0_.username AS username_1 FROM forum_users f0_ ORDER BY LENGTH(f0_.username) ASC'
+        );
+    }
+
+    public function testSupportsOrderByWithSimpleArithmeticExpression()
```
greg0ire, line 301:
```suggestion
    public function testSupportsOrderByWithSimpleArithmeticExpression() : void
```

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
```diff
@@ -289,6 +290,22 @@ public function testSupportsOrderByDesc()
         );
     }
 
+    public function testSupportsOrderByStartingWithFunction()
+    {
+        $this->assertSqlGeneration(
+            'SELECT u FROM Doctrine\Tests\Models\Forum\ForumUser u ORDER BY LENGTH(u.username) asc',
+            'SELECT f0_.id AS id_0, f0_.username AS username_1 FROM forum_users f0_ ORDER BY LENGTH(f0_.username) ASC'
+        );
+    }
```
greg0ire, line 299: This test passes already without a need for your patch, did you add it to make sure you were not breaking that feature? Might warrant a separate commit too.

lib/Doctrine/ORM/Query/Parser.php
ipernet: The coding standard build fails if I just move down the list the `case ($this->isFunction()):` because of the now useless parenthesis.

So for consistency in the switch block I have also removed the parentheses of the others cases. I will send a fix without altering the other cases. Thanks for looking into it!

greg0ire:
> Re-order OrderByIdem parser to proceed with SimpleArithmetic expressions before functions

- The subject line should be less than 50 chars, you can and should use the commit message body if you need more; in fact, the great explanation you put in your first Github message should go there. 
- `OrderByIdem` => `OrderByItem`


Here is how it could look like:

```
Parse simple arithmetic expressions before functions

There is parser bug for OrderByItem(), where a function is detected
before a SimpleArithmeticExpression while it should be the reverse:
simple arithmetic expressions can start with a function too, and thus
they should be matched first if present.

This fix enables the use of expressions in sorting and with window
functions.
```

lib/Doctrine/ORM/Query/Parser.php
greg0ire: You can add another commit in the same PR, that's ok!

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
ipernet: Yes, it's a complementary test that in fact is useless because it is not even impacted by the change (there is NO confusion possible with tsuch order by tokens), so I'll remove it.

lib/Doctrine/ORM/Query/Parser.php
greg0ire: Ah in fact, forget about this: we have temporarily enabled a squash-only policy on this repository, I forgot about it. You can keep it in one single commit.

ipernet:
Thanks @greg0ire for your feedback, they should now all be resolved

greg0ire (approved):

SenseException (approved):
I was playing around with DQL and these changes and couldn't find anything that broke with it. The order of cases could have had an impact on other use cases, but it didn't.

PR #8285, opened 2020-09-25
Findby joined lookup

### LINKED ISSUE #7512: findBy broken for inherited entity property lookups

beberlei:
Supersedes https://github.com/doctrine/orm/issues/7514

Problem with object parameters not recongized as entities when they are.

Fixes #7512

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -41,6 +41,7 @@ use function array_key_exists;
 use function array_map;
 use function array_merge;
 use function assert;
+use function is_object;
 use function reset;
 
 /**
@@ -1985,24 +1986,7 @@ class BasicEntityPersister implements EntityPersister
             return [$newValue];
         }
 
-        if (is_object($value)) {
-            $valueClass = ClassUtils::getClass($value);
-            if ( ! $this->em->getMetadataFactory()->isTransient($valueClass)) {
-                $class = $this->em->getClassMetadata($valueClass);
-                if ($class->isIdentifierComposite) {
-                    $newValue = [];
-
-                    foreach ($class->getIdentifierValues($value) as $innerValue) {
-                        $newValue = array_merge($newValue, $this->getValues($innerValue));
-                    }
-
-                    return $newValue;
-                }
-            }
-
-        }
-
-        return [$this->getIndividualValue($value)];
+        return $this->getIndividualValue($value);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1980,11 +1968,29 @@ private function getValues($value)
      */
     private function getIndividualValue($value)
```
greg0ire, line 1969: The return type seems to have changed, it should be like this now I think:

```
@return       array<mixed>
@psalm-return list<mixed>
```

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1980,11 +1968,29 @@ private function getValues($value)
...
     {
-        if ( ! is_object($value) || ! $this->em->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
-            return $value;
+        if ( ! is_object($value)) {
```
greg0ire, line 1971: I think git-phpcs is indicating the wrong line to cs2pr. This should fix the cs issue.

```suggestion
        if (! is_object($value)) {
```

greg0ire (dismissed):

greg0ire (dismissed):

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1955,36 +1956,43 @@ private function getValues($value)
...
      * Retrieves an individual parameter value.
      *
      * @param mixed $value
      *
-     * @return mixed
+     * @return       array<mixed>
+     *
+     * @psalm-return list<mixed>
      */
     private function getIndividualValue($value)
```
ostrolucky, line 1971: I think method name is no longer correct. It was used only as a fallback when it was clear there is only one value

beberlei: I am not sure, it still returns the parameters for an individual value (which can be an entity with a composite id), whereas the method "above" also handles the case of arrays of individual values.

ostrolucky (dismissed):

SenseException (dismissed):

PR #8288, opened 2020-09-29
[GH-8043] Make IdentityFunction handle primary keys when they are part of an association

### LINKED ISSUE #8043: Make IDENTITY function work with joined columns
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

### Summary

On https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/dql-doctrine-query-language.html#dql-select-examples it is said

> the IDENTITY() DQL function also works for composite primary keys

But if these keys are part of another association they won’t be found in fields but in associations mapping, [hence throwing](https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Query/AST/Functions/IdentityFunction.php#L46).

### Example

Given the following entities

```php
class Contract
{
    /**
     * @ORM\ManyToOne(targetEntity="Agent")
     * @ORM\Id
     */
    private $agent;

    /**
     * @ORM\ManyToOne(targetEntity="Agency")
     * @ORM\Id
     */
    private $agency;
}

class Estate
{
    /**
     * @ORM\ManyToOne(targetEntity="Contract")
     * @ORM\JoinColumns({
     *  @ORM\JoinColumn(name="contract_agent_id", referencedColumnName="agent_id"),
     *  @ORM\JoinColumn(name="contract_agency_id", referencedColumnName="agency_id"),
     * })
     */
    protected $contract;
}
```

I expected to be able to write `IDENTITY(estate.contract, 'agent_id')`.

Is this something that could be implemented?

MatTheCat:
Fix https://github.com/doctrine/orm/issues/8043

When primary keys are part of an association they won’t be found in `fieldMappings` but in `associationMappings` so I refactored `IdentityFunction` to handle this case.

I’m not completely sure about the logic (especially the `reset` which I copied from line 61) but it worked for every case I thought of.

### DIFF
--- a/lib/Doctrine/ORM/Query/AST/Functions/IdentityFunction.php
+++ b/lib/Doctrine/ORM/Query/AST/Functions/IdentityFunction.php
@@ -58,15 +58,18 @@ class IdentityFunction extends FunctionNode
         $joinColumn    = reset($assoc['joinColumns']);
 
         if ($this->fieldMapping !== null) {
-            if (! isset($targetEntity->fieldMappings[$this->fieldMapping])) {
-                throw new QueryException(sprintf('Undefined reference field mapping "%s"', $this->fieldMapping));
-            }
-
-            $field      = $targetEntity->fieldMappings[$this->fieldMapping];
             $joinColumn = null;
 
+            if (isset($targetEntity->fieldMappings[$this->fieldMapping])) {
+                $columnName = $targetEntity->fieldMappings[$this->fieldMapping]['columnName'];
+            } elseif (isset($targetEntity->associationMappings[$this->fieldMapping])) {
+                $columnName = reset($targetEntity->associationMappings[$this->fieldMapping]['joinColumns'])['name'];
+            } else {
+                throw new QueryException(sprintf('Undefined reference mapping "%s"', $this->fieldMapping));
+            }
+
             foreach ($assoc['joinColumns'] as $mapping) {
-                if ($mapping['referencedColumnName'] === $field['columnName']) {
+                if ($mapping['referencedColumnName'] === $columnName) {
                     $joinColumn = $mapping;
 
                     break;
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8043Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8043Test.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH8043Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH8043FirstEntity::class,
+            GH8043SecondEntity::class,
+            GH8043CompositeEntity::class,
+            GH8043ChildEntity::class,
+        ]);
+    }
+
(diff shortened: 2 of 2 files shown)


drupol:
Any news on this?

lib/Doctrine/ORM/Query/AST/Functions/IdentityFunction.php
```diff
@@ -61,16 +61,19 @@ public function getSql(SqlWalker $sqlWalker)
...
-                throw new QueryException(sprintf('Undefined reference field mapping "%s"', $this->fieldMapping));
-            }
-
-            $field      = $targetEntity->fieldMappings[$this->fieldMapping];
             $joinColumn = null;
 
+            if (isset($targetEntity->fieldMappings[$this->fieldMapping])) {
+                $columnName = $targetEntity->fieldMappings[$this->fieldMapping]['columnName'];
+            } elseif (isset($targetEntity->associationMappings[$this->fieldMapping])) {
+                $columnName = reset($targetEntity->associationMappings[$this->fieldMapping]['joinColumns'])['name'];
```
greg0ire, line 66: I think `reset()` is here because we don't support `IDENTITY` with composite keys?

greg0ire (commented):
Should docs be added?

lib/Doctrine/ORM/Query/AST/Functions/IdentityFunction.php
drupol: I guess this is the reason.

MatTheCat: In that case could we throw an exception if there are more than one key?

greg0ire: Maybe, not sure why it wasn't done like that in the first place. Right now, it will only return the value of the first part of the composite key, right?

MatTheCat: Hm from https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/dql-doctrine-query-language.html#:~:text=the%20IDENTITY()%20DQL I read

> the IDENTITY() DQL function also works for composite primary keys

So there must be another reason :thinking:

SenseException:
The bottom of https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/dql-doctrine-query-language.html#dql-select-examples shows some addition for every version change. This can be extended for this PR.

MatTheCat:
Hi everyone I’m back :wave: 

I changed the target branch to `2.8.x` and added a test. Is everything alright?

MatTheCat:
@SenseException I feel like my PR is a bugfix so I’m not sure if I should add an example in the documentation?

drupol:
> @SenseException I feel like my PR is a bugfix so I’m not sure if I should add an example in the documentation?

Bugfix or not, I think an example always worth :)

MatTheCat:
The documentation already has an example for

> The IDENTITY() DQL function also works for composite primary keys

Should I add `even if they’re part of an association`? Don’t really know what to add :thinking:

SenseException:
An addition that IdentityFunction can handle primary keys now would be sufficient.

MatTheCat:
@SenseException this PR allows `IDENTITY` to handle composite primary keys which also are foreign keys. From the documentation I expected this to work as it says 

> The IDENTITY() DQL function also works for composite primary keys

Now that it really works what could be to add?

MatTheCat:
@greg0ire I added a test and rebased

PR #8292, opened 2020-10-05
Introduce `toGenerator()` on AbstractQuery

simPod:
The point is to have both `toIterable()` and `toGenerator()` so consumers are able to rely on stricter type.

So if they choose `toGenerator()` in order to rely on `Generator` return type, they can assume it is not "rewindable", eg. they can analyze in code that the Generator value is not iterated more than once

### DIFF
--- a/docs/en/reference/batch-processing.rst
+++ b/docs/en/reference/batch-processing.rst
@@ -75,7 +75,7 @@ Iterating results
 ~~~~~~~~~~~~~~~~~
 
 An alternative solution for bulk updates is to use the
-``Query#getIterable()`` facility to iterate over the query results step
+``Query#getIterable()`` or ``Query#toGenerator()`` facility to iterate over the query results step
 by step instead of loading the whole result into memory at once.
 The following example shows how to do this, combining the iteration
 with the batching strategy that was already used for bulk inserts:
@@ -135,7 +135,7 @@ Iterating results
 ~~~~~~~~~~~~~~~~~
 
 An alternative solution for bulk deletes is to use the
-``Query#getIterable()`` facility to iterate over the query results step
+``Query#getIterable()`` or ``Query#toGenerator()`` facility to iterate over the query results step
 by step instead of loading the whole result into memory at once.
 The following example shows how to do this:
 
@@ -165,7 +165,7 @@ The following example shows how to do this:
 Iterating Large Results for Data-Processing
 -------------------------------------------
 
-You can use the ``getIterable()`` method just to iterate over a large
+You can use the ``getIterable()`` method or ``toGenerator()`` to iterate over a large
 result and no UPDATE or DELETE intention. ``$query->getIterable()`` returns ``iterable``
 so you can process a large result without memory
 problems using the following approach:
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -191,10 +191,10 @@ the life-time of their registered entities.
 
 .. warning::
 
-    Note that, when using ``Doctrine\ORM\AbstractQuery#getIterable()``, ``postLoad``
+    Note that, when using ``Doctrine\ORM\AbstractQuery#getIterable()`` or ``Doctrine\ORM\AbstractQuery#toGenerator()``, ``postLoad``
     events will be executed immediately after objects are being hydrated, and therefore
     associations are not guaranteed to be initialized. It is not safe to combine
-    usage of ``Doctrine\ORM\AbstractQuery#getIterable()`` and ``postLoad`` event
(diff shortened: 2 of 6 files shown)


SenseException:
On the other side you can't assume that an Iterator is rewindable (e.g. https://www.php.net/manual/en/class.norewinditerator.php). `NoRewindIterator` doesn't raise a fatal error like a Generator though.

These changes look like syntactic sugar to use a stricter return type for what `toIterator` is already doing. This could also be done in userland.

simPod:
@SenseException yes, this is only a syntactic sugar

beberlei:
I think we should decide to have either one or the other, not both. Would it be better to have a `toGenerator` than a `toIterable`?

simPod:
Personally I prefer `toGenerator()`. Can't come up with any disadvantages. 

Not sure what user base would think of that though.

SenseException:
`toGenerator()` would also suggest that the return value will have the typical PHP Generator behaviour.

jvasseur:
Personally I'm against using a `Generator` return type. To me using a `Generator` return type advertise that you can use the `send`/`throw`/`getReturn` on it. In this case what we want to provide is a way to iterate over result and `iterable` means exactly that.

To me the non-rewindable argument is a moot point since other iterators could display the same behavior and/or [php generators could become rewindable at some point](https://externals.io/message/108767).

simPod:
Generator is iterable so IMO we're providing a way to iterate. 🤔

docs/en/reference/batch-processing.rst
```diff
@@ -75,7 +75,7 @@ Iterating results
 ~~~~~~~~~~~~~~~~~
 
 An alternative solution for bulk updates is to use the
-``Query#getIterable()`` facility to iterate over the query results step
```
beberlei, line 78: The docs still need change as `getIterable` needs to be removed form the sentence.

docs/en/reference/batch-processing.rst
```diff
@@ -135,7 +135,7 @@ Iterating results
 ~~~~~~~~~~~~~~~~~
 
 An alternative solution for bulk deletes is to use the
-``Query#getIterable()`` facility to iterate over the query results step
+``Query#getIterable()`` or ``Query#toGenerator()`` facility to iterate over the query results step
```
beberlei, line 138: and here

docs/en/reference/batch-processing.rst
```diff
@@ -165,7 +165,7 @@ The following example shows how to do this:
 Iterating Large Results for Data-Processing
 -------------------------------------------
 
-You can use the ``getIterable()`` method just to iterate over a large
+You can use the ``getIterable()`` method or ``toGenerator()`` to iterate over a large
```
beberlei, line 168: and here.

docs/en/reference/events.rst
```diff
@@ -191,10 +191,10 @@ the life-time of their registered entities.
 
 .. warning::
 
-    Note that, when using ``Doctrine\ORM\AbstractQuery#getIterable()``, ``postLoad``
+    Note that, when using ``Doctrine\ORM\AbstractQuery#getIterable()`` or ``Doctrine\ORM\AbstractQuery#toGenerator()``, ``postLoad``
```
beberlei, line 194: remove getIterable here

docs/en/reference/events.rst
```diff
@@ -191,10 +191,10 @@ the life-time of their registered entities.
...
     events will be executed immediately after objects are being hydrated, and therefore
     associations are not guaranteed to be initialized. It is not safe to combine
-    usage of ``Doctrine\ORM\AbstractQuery#getIterable()`` and ``postLoad`` event
+    usage of ``Doctrine\ORM\AbstractQuery#getIterable()`` or ``Doctrine\ORM\AbstractQuery#toGenerator()`` with ``postLoad`` event
```
beberlei, line 197: remove getIterable here

beberlei (changes requested):
There are also references still to `getIterable` in the deprecation messaage of `iterate()` in AbstractQuery and AbstractHydrator

docs/en/reference/batch-processing.rst
simPod: I fixed the naming here https://github.com/doctrine/orm/pull/8293 though there are no reviews yet. It is not part of this PR.

beberlei:
Jus tsaw @jvasseur's argument and I think he is right, we should use `toIterable`, because while we do return a generator, we don't want users to use the generators API. Iterators are also not required to be rewindable. Sorry for the left, right, left jumping here.

beberlei:
Replaced by https://github.com/doctrine/orm/pull/8293

PR #8295, opened 2020-10-07
Add public getter for queryComponents

leipsfur:
I propose to add this getter to get access to the queryComponents property of the parser. The queryComponents property contains useful information about the corresponding entity class name of query aliases.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -232,6 +232,16 @@ class Parser
     }
 
     /**
+     * Gets the queryComponents of the parsed query
+     *
+     * @return array
+     */
+    public function getQueryComponents()
+    {
+        return $this->queryComponents;
+    }
+
+    /**
      * Gets the lexer used by the parser.
      *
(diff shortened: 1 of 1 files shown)


SenseException:
This is an access to the `Parser` interna. What do you try to accomplish with the access to that information?

leipsfur:
Here is my use-case: I have an API that is serializing doctrine collections with mapped entities. My API provides an "group by"-endpoint where results are grouped by a column. Since I only want to allow the user to group by certain columns I have an annotation in my entities that allow "group by"s. When the user wants to group by an property of a mapped entity the API parameter value looks like "child_entity.column1". I have to check the Annotation in the mapped entities class but at this point I do not have the fully qualified class name. The queryComponents property contains this information.

IMO the better solution would be to also provide a Getter for this property in the Query class for performance reasons but I was unsure if accessing this property at all is a valid solution

beberlei:
I don't really want to expose this, you can use an SQL Walker for this, it has access to the query components:

https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/cookbook/dql-custom-walkers.html

Issue #8294, opened 2020-10-06
Forgetting to add `indexes` keyword results in php error


JohJohan:
Forgetting to add keyword `indexes` will result in a php error example code:

``` php
<?php

declare(strict_types=1);

namespace App\Entity;

use Doctrine\ORM\Mapping as ORM;

/**
 * Class Blog.
 *
 * @ORM\Table(name="blog", {
 *     @ORM\Index(name="blog_id_idx", columns={"id"})
 *}
 *)
 * @ORM\Entity
 */
class Blog
{
}
```

This will cause this PHP error:
```
In ClassMetadataInfo.php line 2299:

Warning: strpos() expects parameter 1 to be string, array given
```

This is caused here:
https://github.com/doctrine/orm/blob/e0eb82a3b12a3fd0878b5d56c3c293c04c168329/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L2356

My suggestion is to add some validation and throw a exception maby a `MappingException`. 

I will like to do this let me know if i should.

PR #8296, opened 2020-10-07
8294 - Handle invalid table name

JohJohan:
See https://github.com/doctrine/orm/issues/8294

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -2347,11 +2347,16 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param array $table The table description.
      *
+     * @throws MappingException
+     *
      * @return void
      */
     public function setPrimaryTable(array $table)
     {
         if (isset($table['name'])) {
+            if (false === is_string($table['name'])) {
+                throw MappingException::invalidTableName($this->name, gettype($table['name']));
+            }
             // Split schema and table name from a table name like "myschema.mytable"
             if (strpos($table['name'], '.') !== false) {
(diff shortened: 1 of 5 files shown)


Ahummeling:
I think the problem here (programmer error) is solved in the wrong place. The annotations reader is responsible for reading the annotations. If the syntax is wrong (a keyword is missing), and if the reader is able to detect this as an error, then the reader should throw an exception to inform the programmer he is making a mistake. 

The problem I have with this solution is that checking whether the `$tabe['name']` is a string is that it still doesn't guarantee that it's actually the name, and not some other property of the table annotation that also happens to be a string. 

In short, I think the issue lies with incorrect mapping of annotation.

JohJohan:
@Ahummeling i can see where your comming from, maybe a doctrine maintainer can give a hint where the fix needs to be placed

greg0ire:
@Ahummeling already gave the correct hint IMO:

> The annotations reader

JohJohan:
@greg0ire i create a branch on doctrine/annotations see https://github.com/doctrine/annotations/pull/447 it looks like it is not a problem anymore for tag 1.13.x because we get a valid name see my test. Let me know what you think?

greg0ire:
@JohJohan sorry I didn't double check @Ahummeling 's hint more . The right place to do this is IMO inside the annotation _driver_ (so inside the ORM). The annotation reader doesn't know about ORM-specific annotations at all (nor should it).

https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php

JohJohan:
@greg0ire i made changes in the [AnnotationDriver](https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php) maybe you can do a CR?

PR #8302, opened 2020-10-11
Fix ClassMetadataInfo::$fieldMappings type

ondrejmirtes:
/cc @orklah


orklah:
Great! I tried to infer manually the content of this array, looks like I missed some content.

norkunas:
Isn't this also missing the `quoted` property? As it is set on the 1417 line when field name starts with backtick

ondrejmirtes:
It's possible, but that one isn't even documented in the PHPDoc above.

greg0ire (approved):

orklah:
@greg0ire I think this can be closed now

PR #8303, opened 2020-10-11
PHP8 Support

beberlei:
- [x] upgrade phpunit to 8+ (https://github.com/doctrine/orm/pull/8330)
- [x] update DBAL dependency (waiting for https://github.com/doctrine/dbal/pull/4347 and DBAL 2.12)
- [x] wait on https://github.com/doctrine/orm/pull/8317 for php 8 support in CI
- [x] update doctrine/coding-standard to v8+

## Attributes

Support for an Attribute Metadata Driver will be added seperately in #8266 

As discussed in #8266 we need a few small changes on `doctrine/annotations`. These are addressed in https://github.com/doctrine/annotations/pull/369 and we are next going for for annotations 1.11 release to include this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "phpstan/phpstan": "^0.12.18",
         "phpunit/phpunit": "^8.0",
         "symfony/yaml": "^3.4|^4.0|^5.0",
-        "vimeo/psalm": "^3.11"
+        "vimeo/psalm": "4.1.1"
     },
     "suggest": {
         "symfony/yaml": "If you want to use YAML Metadata Mapping Driver"
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "symfony/console": "^3.0|^4.0|^5.0"
     },
     "require-dev": {
-        "doctrine/coding-standard": "^6.0",
+        "doctrine/coding-standard": "^8.0",
         "phpstan/phpstan": "^0.12.18",
         "phpunit/phpunit": "^8.5|^9.4",
         "symfony/yaml": "^3.4|^4.0|^5.0",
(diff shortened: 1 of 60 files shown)


composer.json
```diff
@@ -37,7 +37,7 @@
     "require-dev": {
         "doctrine/coding-standard": "^6.0",
         "phpstan/phpstan": "^0.12.18",
-        "phpunit/phpunit": "^7.5",
+        "phpunit/phpunit": "^8.5|^9.5",
```
derrabus, line 40:
```suggestion
        "phpunit/phpunit": "^9.4",
```

There is no PHPUnit 9.5 yet. PHPUnit 9.4 should be compatible with php 7.3, so no need for PHPUnit 8 anymore.

composer.json
```diff
@@ -16,14 +16,14 @@
...
-        "php": "^7.2",
+        "php": "^7.2|^8.0",
         "ext-pdo": "*",
         "composer/package-versions-deprecated": "^1.8",
         "doctrine/annotations": "^1.11.1",
         "doctrine/cache": "^1.9.1",
         "doctrine/collections": "^1.5",
         "doctrine/common": "^3.0",
-        "doctrine/dbal": "^2.10.0",
+        "doctrine/dbal": "^2.12.0",
```
greg0ire, line 26: Why are we bumping this? Can't we let Composer figure things out and restore the 7.2 job?

beberlei: @greg0ire because PHP 8 requires 2.12

beberlei: I plan to merge the PHP 7.1 support for DBAL 2.12 in the next days, so that means we can add them in ORM as well. But that can be seperate: https://github.com/doctrine/dbal/pull/4386

greg0ire (approved):
:ship: :ship: :ship:

PR #8305, opened 2020-10-15
Exception when loading proxy does not mark proxy as initialized

lstrojny:
Similar issue as https://github.com/doctrine/orm/issues/3147 but with a random exception, e.g. when the connection is closed.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
@@ -19,6 +19,7 @@ use Doctrine\Tests\Models\Company\CompanyEmployee;
 use Doctrine\Tests\Models\Company\CompanyPerson;
 use Doctrine\Tests\Models\ECommerce\ECommerceFeature;
 use Doctrine\Tests\OrmTestCase;
+use Exception;
 use ReflectionProperty;
 use stdClass;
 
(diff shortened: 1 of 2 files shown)


greg0ire:

lstrojny:
@greg0ire done

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
```diff
@@ -139,6 +140,29 @@ public function testFailedProxyLoadingDoesNotMarkTheProxyAsInitialized(): void
...
+        $persister
+            ->expects($this->atLeastOnce())
+            ->method('load')
+            ->willThrowException(new Exception());
+
+        try {
+            $proxy->getDescription();
+            $this->fail('An exception was expected to be raised');
+        } catch (Exception $exception) {
+        }
```
SenseException, line 162:
```suggestion
        $this->expectException(Exception::class);
        $proxy->getDescription();
```

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
```diff
@@ -139,6 +140,29 @@ public function testFailedProxyLoadingDoesNotMarkTheProxyAsInitialized(): void
         $this->assertInstanceOf('Closure', $proxy->__getCloner(), 'The cloner wasn\'t removed');
     }
 
+    public function testExceptionWhenLoadingProxyDoesNotMarkTheProxyAsInitialized(): void
+    {
+        $persister = $this->getMockBuilder(BasicEntityPersister::class)->setMethods(['load'])->disableOriginalConstructor()->getMock();
```
SenseException:
```suggestion
        $persister = $this->getMockBuilder(BasicEntityPersister::class)
            ->setMethods(['load'])
            ->disableOriginalConstructor()
            ->getMock();
```

SenseException (changes requested):

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
lstrojny: Would make the test less precise: instead of expecting exactly `getDescription` to throw, the exception could happen anywhere, e.g. when calling `$proxy->__isInitialized()`

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
lstrojny: Formatting is identical to the one above. I guess if the Doctrine projects wants a different way of formatting fluent interfaces, introducing a CS rule and applying it across the whole project would make more sense. What do you think?

lstrojny:
Bump. Could you have another look?

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
greg0ire: > I guess if the Doctrine projects wants a different way of formatting fluent interfaces

This is not specifically about fluent interfaces: this line is 135 characters long, while PSR recommends 120.

> introducing a CS rule and applying it across the whole project would make more sense. What do you think?

It would make sense, but although it sounds simple, it sure isn't, see http://journal.stuffwithstuff.com/2015/09/08/the-hardest-program-ive-ever-written/

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
SenseException: `expectException()` does exactly that. The method is more precise than the outdated annotation from the older PHPUnit versions.

lstrojny:
Bump again

SenseException:
Please take a look at the open change request that needs to be referenced.

lstrojny:
@SenseException do you mean [this comment](https://github.com/doctrine/orm/pull/8305#discussion_r603647302)? If so, that would render the assertions after the line that renders the exceptions, useless. That's why we shouldn't do it.

PR #8317, opened 2020-10-25
Start moving travis phpunit runs to Github actions.

beberlei:
Move from Travis CI to Github for PHPUnit tests.


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,225 @@
...
+          - "13"
+
+    services:
+      postgres:
+        image: "postgres:${{ matrix.postgres-version }}"
+        env:
+          POSTGRES_PASSWORD: "postgres"
+
+        options: >-
+          --health-cmd "pg_isready || psql -U postgres -c 'SELECT 1'"
```
greg0ire, line 82: Not needed unless you test very old versions of PG where `pg_isready` does not exist
```suggestion
          --health-cmd "pg_isready"
```

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,225 @@
...
+
+    services:
+      mysql:
+        image: "mysql:${{ matrix.mysql-version }}"
+
+        options: >-
+          --health-cmd "mysqladmin ping --silent"
+          -e MYSQL_ALLOW_EMPTY_PASSWORD=yes
+          -e MYSQL_DATABASE=doctrine_tests
+          ${{ matrix.custom-entrypoint }}
```
greg0ire, line 192: There is no such key in the matrix (and it's a bit weird to me that things work without it, but maybe things are not setup the same way here?)

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,225 @@
...
+on:
+  pull_request:
+    branches:
+      - "*.x"
+      - "master"
+  push:
+    branches:
+      - "*.x"
+      - "master"
+
```
greg0ire, line 12: This will fail the build when an extension fails to install
```suggestion

env:
  fail-fast: true
```

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,225 @@
...
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v2"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Extensions"
+        run: "php -r 'var_dump(get_loaded_extensions());'"
```
greg0ire, line 219: Is that a remnant from debugging?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,225 @@
...
+        uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 2
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          extensions: "pdo, pdo_sqlite"
+          coverage: "pcov"
```
greg0ire, line 41: Please add a step that uses Codecov to upload the coverage files

.travis.yml
```diff
@@ -1,103 +0,0 @@
...
-before_install:
-  - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
-  - composer self-update
-
-install: travis_retry composer validate --strict && composer update --prefer-dist
-
-script:
-  - if [[ "$DB" == "mysql" || "$DB" == "mariadb" ]]; then mysql -e "CREATE SCHEMA doctrine_tests; GRANT ALL PRIVILEGES ON doctrine_tests.* to travis@'%'"; fi
-  - ENABLE_SECOND_LEVEL_CACHE=0 ./vendor/bin/phpunit -v -c tests/travis/$DB.travis.xml
-  - ENABLE_SECOND_LEVEL_CACHE=1 ./vendor/bin/phpunit -v -c tests/travis/$DB.travis.xml --exclude-group performance,non-cacheable,locking_functional
```
greg0ire, line 29: This seems to have been lost in translation

.travis.yml
```diff
@@ -1,103 +0,0 @@
...
-      script:
-        - ENABLE_SECOND_LEVEL_CACHE=0 ./vendor/bin/phpunit -v -c tests/travis/$DB.travis.xml --coverage-clover ./build/logs/clover.xml
-      after_success:
-        - bash <(curl -s https://codecov.io/bash) -f ./build/logs/clover.xml
-
-    - stage: Code Quality
-      env: DB=none BENCHMARK
-      php: 7.4
-      before_script: wget https://phpbench.github.io/phpbench/phpbench.phar https://phpbench.github.io/phpbench/phpbench.phar.pubkey
-      script: php phpbench.phar run -l dots --report=default
```
greg0ire, line 87: Lost in translation too

greg0ire (commented):
I see we no longer use `composer install`… should `composer.lock` be removed from version control?

.github/workflows/continuous-integration.yml
beberlei: its a leftover from going by the DBAL suite

beberlei:
@greg0ire its a good question about composer install/update, this probably goes hand in hand with pin tooling versions to specific versions to avoid failures suddenly starting due to new versions of psalm, phpunit, phpstan, phpcs...

greg0ire:
@beberlei see https://github.com/orgs/doctrine/teams/doctrinecore/discussions/21 if you want to dive into that rabbit hole :)

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -342,8 +342,6 @@ public function loadMetadataForClass($className, ClassMetadata $metadata)
                             'initialValue' => $seqGeneratorAnnot->initialValue
                         ]
                     );
-                } else if ($this->reader->getPropertyAnnotation($property, 'Doctrine\ORM\Mapping\TableGenerator')) {
-                    throw MappingException::tableIdGeneratorNotImplemented($className);
```
greg0ire, line 346: I used another fix in #8321

greg0ire:
Also, could you please target 2.7 for that kind of changes? Fewer conflicts when merging up.

greg0ire:
Guess I will have to upgrade to PHP 8 :sweat_smile:

greg0ire:
It's going to be too big, let's merge your PR as is, I'll do the PHPUnit 8 migration and coverage upload in a separate PR.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,232 @@
...
+    runs-on: "ubuntu-20.04"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.2"
+          - "7.3"
+          - "7.4"
+        deps:
+          - "fixed"
```
greg0ire, line 22: Should we use `"normal"` here? You are using `composer update`, so the dependencies are not fixed.

greg0ire:
Let's remove `composer.lock` from version control? As concluded with https://github.com/orgs/doctrine/teams/doctrinecore/discussions/21

PR #1333, opened 2015-03-15
Allow DateTimeImmutable as parameter value

janlanger:
This pull request allows passing instance of DateTimeImmutable as value to `setParameter` and its correct formating in query.

PR #8328, opened 2020-11-05
Infer `datetime_immutable` DBAL type for `\DateTimeImmutable` instance parameters

vhenzl:
The support for passing `\DateTimeImmutable` instance as a query parameter has been added to ORM in #1333 (the year 2015), a long time before immutable date types (`datetime_immutable` etc) were introduced to DBAL in doctrine/dbal#2450 (2017).

Back then, it made sense to treat `\DateTimeImmutable` (or any `\DateTimeInterface`) in the same way as `\DateTime` and infer parameter type as `datetime`. However, when immutable date types were later added to DBAL, it wasn't reflected anyhow in type inference in ORM and `\DateTimeImmmutable` instances are still inferred as `datetime` DBAL type.

This PR fixes this IMO incorrect behaviour of `ParameterTypeInferer::inferType()`: for a `\DateTimeImmmutable` parameter, it now returns `datetime_immutable` DBAL type; for `\DateTime` or any other types implementing `\DateTimeInterface`, it returns `datetime` DBAL type as it did before.

This behaviour is in line with [`DateTimeImmutableType` handling only `\DateTimeImmutable`](https://github.com/doctrine/dbal/blob/2.12.x/lib/Doctrine/DBAL/Types/DateTimeImmutableType.php#L32) and [`DateTimeType` handling any `\DateTimeInterface`](https://github.com/doctrine/dbal/blob/2.12.x/lib/Doctrine/DBAL/Types/DateTimeType.php#L41).

Why? In most cases, it doesn't matter and `datetime` works for `\DateTimeImmutable` parameters just fine. But it does matter if using custom implementation of `datetime_immutable` type like `UTCDateTimeImmutableType` from [simpod/doctrine-utcdatetime](https://packagist.org/packages/simpod/doctrine-utcdatetime). Then the broken type inference is revealed.

This is partially related to #6443, however, this PR isn't about custom DBAL types but about correct type inference for build-in types.

### DIFF
--- a/lib/Doctrine/ORM/Query/ParameterTypeInferer.php
+++ b/lib/Doctrine/ORM/Query/ParameterTypeInferer.php
@@ -53,7 +53,11 @@ class ParameterTypeInferer
             return Type::BOOLEAN;
         }
 
-        if ($value instanceof \DateTime || $value instanceof \DateTimeInterface) {
+        if ($value instanceof \DateTimeImmutable) {
+            return Type::DATETIME_IMMUTABLE;
+        }
+
+        if ($value instanceof \DateTimeInterface) {
             return Type::DATETIME;
         }
 
--- a/lib/Doctrine/ORM/Query/ParameterTypeInferer.php
+++ b/lib/Doctrine/ORM/Query/ParameterTypeInferer.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\ORM\Query;
 
+use DateTimeImmutable;
+use DateTimeInterface;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Types\Type;
 
@@ -53,11 +55,11 @@ class ParameterTypeInferer
             return Type::BOOLEAN;
         }
 
-        if ($value instanceof \DateTimeImmutable) {
+        if ($value instanceof DateTimeImmutable) {
             return Type::DATETIME_IMMUTABLE;
         }
 
-        if ($value instanceof \DateTimeInterface) {
+        if ($value instanceof DateTimeInterface) {
             return Type::DATETIME;
         }
 
--- a/tests/Doctrine/Tests/ORM/Query/ParameterTypeInfererTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/ParameterTypeInfererTest.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Tests\ORM\Query;
 
+use DateTimeImmutable;
 use Doctrine\ORM\Query\ParameterTypeInferer;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Types\Type;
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ORM/Query/ParameterTypeInferer.php
```diff
@@ -53,7 +53,11 @@ public static function inferType($value)
             return Type::BOOLEAN;
         }
 
-        if ($value instanceof \DateTime || $value instanceof \DateTimeInterface) {
+        if ($value instanceof \DateTimeImmutable) {
```
vhenzl, line 56: Why is this failing? The changed code used fully qualified names as well.

SenseException: Try to use `use` instead of having the FQCN directly in the code. The same with your other changes.

SenseException (changes requested):

lib/Doctrine/ORM/Query/ParameterTypeInferer.php
vhenzl: Well, yeah, that's obvious. My question is _why_? Because it doesn't make any sense.

The original code uses FQN and evidently don't adhere to the set coding standards. Why force the CS for the modified code then?

Now, I have two options - either I'll add `use`s only for the changed lines leaving the modified files inconsistent (one line will have, for instance, `DateTimeInterface` while another `\DateInterval`), or I will add `use`s for all FQN in all modified files making the surface of the PR much bigger with a lot of changes not related to the proposed change. And both options are wrong.

SenseException: ORM is a long running project and some parts, like the coding styles were introduced/updated late over time. This is why there is a coding style check that targets changed code that are part of a commit/diff. In your case the changes in the lines with `instanceof` are now affected by the style-checks and need to be adapted in this PR.

vhenzl: That's all nice but I still fail to see any benefits of that as it only clutters the code, must make merging between branches harder and leads to useless discussions like this one. IMO all active branches should be updated to follow CS rules and then makes sense to force CS checks in PRs. But whatever works for Doctrine maintainers, take this just as feedback from a random contributor that current state makes contribution unnecessary harder…

The code has been updated, all checks are green now.

SenseException (dismissed):

vhenzl:
Rebased and target changed to `2.8.x` branch now that v2.8 has been released.

vhenzl:
Again failing because of coding standard check… 🤬 How it should be fixed? These reported problems are totally unrelated to this PR!

SenseException (approved):

greg0ire:
Thanks @vhenzl !

PR #8329, opened 2020-11-06
Fix SQL alias generation regression for simple inheritance 

rogregoire:
As asked by @ostrolucky in the #8229 issue - I'm created a PR trying to fix a edge case of what I think may be a regression in the effort of supporting Joined table inheritance with AttributeOverride.

Context:
- Entity1 extends Entity0 through ResolveTargetEntities mecanism
- Entity1 has all its fieldsMapping marked as inherited=true 
- Entity1 inheritanceType is ClassMetadata::INHERITANCE_TYPE_NONE
- Some code calls $entityManager->getRepository($entity1)->findOneBy()
- The SQL generated in the where part of the query has a bad alias because it uses Entity0 as classname for getSQLTableAlias call in BasicEntityPersister::getSelectConditionStatementSQL method
- The query crash

Proposal:
  Excluding INHERITANCE_TYPE_NONE entities to use inherited classname as parameter for getSQLTableAlias generation in BasicEntityPersister::getSelectConditionStatementSQL

As said in original comment, I'm really not sure of what I'm doing, I'm new to doctrine 2

Referenced Issue comments: https://github.com/doctrine/orm/issues/8229#issuecomment-722942180

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -1698,7 +1698,7 @@ class BasicEntityPersister implements EntityPersister
         if (isset($this->class->fieldMappings[$field])) {
             // Fix for bug GH-8229 (id column from parent class renamed in child class):
             // Use the correct metadata and name for the id column
-            if (isset($this->class->fieldMappings[$field]['inherited'])) {
+            if (isset($this->class->fieldMappings[$field]['inherited']) && $this->class->inheritanceType !== ClassMetadata::INHERITANCE_TYPE_NONE) {
                 $className = $this->class->fieldMappings[$field]['inherited'];
                 $class     = $this->em->getClassMetadata($className);
             } else {
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8229Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8229Test.php
@@ -23,6 +23,8 @@ class GH8229Test extends OrmFunctionalTestCase
             [
                 GH8229Resource::class,
                 GH8229User::class,
+                GH8229EntityWithoutDiscriminator::class,
+                EntityExtendingGH8229EntityWithoutDiscriminator::class,
             ]
         );
     }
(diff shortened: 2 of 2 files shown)


ostrolucky:
Test case is also needed

ostrolucky:
There is lot of problems in this PR. Please edit this PR and check this checkbox, so I can help you.

Screenshot 2020-11-09 at 18 53 04

ostrolucky:
Have you checked the checkbox yet?

rogregoire:
image
That's all I found and it's already checked, Am I missing something ?

ostrolucky:
I have cleaned up git history and removed unnecessary try catch in a test and unnecessary usage of `ResolveTargetEntityListener`. This listener has no effect on reproducing this.

If you plan to push some more commits, you need to pull with rebase your branch: `git pull --rebase`

rogregoire:
Thanks, 
Well, I didn't test without ResolveTargetEntityListener because that wasn't my bug case, but I guess that mean this bug must have impacted a lot more of people that I initially thought ?

ostrolucky:
yes

ostrolucky (approved):

cziegenberg:
Thanks @rogregoire!

FYI: I just found two other untested use cases for which the #8229 fix doesn't work yet. I'm currently trying to create tests for it and will provide a PR soon.

ostrolucky:
Thanks! Ping me in a PR afterwads

cziegenberg:
@ostrolucky After some testing everything works fine. The problem was an outdated composer.lock file not matching the composer.json file and pointing to an old version.

ostrolucky:
👍

beberlei:
@cziegenberg can you please explain your use case? This looks like invalid mapping to me. But maybe I am wrong, are these two entities stored in one or two distinct tables in your case?

beberlei:
@rogregoire The ResolveEntityTargetListener does not work when you extend another entity with. Its specifically for implementing an interface. Your use-case of Entity1 extends Entity0 is only supported if you do Inheritence Mapping.

cziegenberg:
@beberlei:

This isn't my PR, so do you really mean my usecase? My use case is explained in #8229. The goal is a "everything is a resource"-structure:
- a global resource table "app_resource" with the columns "resource_id" and "resource_type"
- an own table for each specific resource type (i.e. "app_user") with additional, type-specific columns, and an adjusted name for the identifier column to clearly name it and differ references
- other tables referencing either any type of resource (-> "app_resource") or a specific one (i.e. "app_user")

A mapped super class wouldn't allow to reference "any type of resource", right?

PR #8330, opened 2020-11-07
Phpunit 8

greg0ire:


beberlei:
@greg0ire i did this work in #8303 already, should we spit that up?

greg0ire:
> @greg0ire i did this work in #8303 already, should we spit that up?

Argh, I didn't see… well, I'm almost done I think.

beberlei (dismissed):

greg0ire:
I had to push again @beberlei , the `filter` path missed a `../`, now coverage is generated properly and should be uploaded to Codecov :crossed_fingers:

beberlei (approved):

PR #8332, opened 2020-11-07
Upgrade Psalm to v4.1.1

beberlei:
This is required for PHP 8 support, because Psalm 3 has a composer installation conflict with a downstream dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "phpstan/phpstan": "^0.12.18",
         "phpunit/phpunit": "^8.0",
         "symfony/yaml": "^3.4|^4.0|^5.0",
-        "vimeo/psalm": "^3.11"
+        "vimeo/psalm": "4.1.1"
     },
     "suggest": {
         "symfony/yaml": "If you want to use YAML Metadata Mapping Driver"


composer.json
```diff
@@ -36,7 +36,7 @@
         "phpstan/phpstan": "^0.12.18",
         "phpunit/phpunit": "^8.0",
         "symfony/yaml": "^3.4|^4.0|^5.0",
-        "vimeo/psalm": "^3.11"
+        "vimeo/psalm": "4.1.1"
```
greg0ire, line 39: Let's allow bugfixes?

```suggestion
        "vimeo/psalm": "~4.1.1"
```

beberlei: what will bugfixes yield? all checks pass at the time of picking 4.1.1. The only benefit a bugfix brings would be either internal, or with respect to a feature we didn't use. In any case a bugfix could also introduce a new violation/warning and cause the spurious failures in unrelated PRs due to timing that we want to avoid.

greg0ire: :thinking: indeed, that makes sense 👍

greg0ire (approved):

PR #8334, opened 2020-11-09
ORM layer validation is not always sufficient

oojacoboo:
The example provided here discusses accounting level data design, something that should always strive to remain 100% accurate.  Doctrine is entirely not sufficient, alone, in ensuring this data integrity.  A notice should be provided in the docs regarding this.

### DIFF
--- a/docs/en/cookbook/aggregate-fields.rst
+++ b/docs/en/cookbook/aggregate-fields.rst
@@ -159,10 +159,19 @@ collection, which means we can compute this value at runtime:
 Now we can always call ``Account::getBalance()`` to access the
 current account balance.
 
-To enforce the max credit rule we have to implement the "Aggregate
+To enforce the max credit rule we implement the "Aggregate
 Root" pattern as described in Eric Evans book on Domain Driven
 Design. Described with one sentence, an aggregate root controls the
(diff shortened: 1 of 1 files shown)


SenseException:
Thank you for your contribution. I agree that data integrity checks with a db side validation can improve data handling and consistency, but I also think that this might be out of scope for this ORM specific example that has the focus in the docs. It hasn't the purpose to show the best way how to implement a bank account data model but just the usage of aggregates in DQL.

oojacoboo:
I don't see the harm in providing the "disclaimer"/"notice" to devs reviewing the docs.

SenseException:
What do you suggest as a sidenote? It should point out that it is not in the scope of the ORM or this example.

oojacoboo:
I'm not suggesting a change in the wording, but a stylistic effect on the paragraph.  I don't know how the formatting is done here, so maybe you can provide some insight into that.

beberlei:
I am unsure about this disclaimer, this should rather be in the architecture or overview section of Doctrine. As an ORM obviously the vlaidation logic is in code and not in the database, this is based on the abstraction used. Here at this point you choose it feels a bit lost, and a bit overkill for explaining something entirely different.

PR #8336, opened 2020-11-10
Update functionality of code examples

SenseException:


greg0ire (dismissed):

greg0ire:
We might even want to use type declarations here, WDYT?

SenseException:
@greg0ire I've added types to the methods. Do you think I should add those to properties too? It looks weitd that `addEntry` has a return type too.

docs/en/cookbook/aggregate-fields.rst
```diff
@@ -32,30 +32,39 @@ Our entities look like:
 .. code-block:: php
 
     <?php
+
     namespace Bank\Entities;
+
+    use use Doctrine\ORM\Mapping as ORM;
```
greg0ire, line 38:
```suggestion
    use Doctrine\ORM\Mapping as ORM;
```

docs/en/cookbook/aggregate-fields.rst
```diff
@@ -175,12 +190,13 @@ relation with this method:
     <?php
     class Account
     {
-        public function addEntry($amount)
+        public function addEntry(int $amount): Entry
```
greg0ire, line 193: That does look weird, not sure where it comes from.

greg0ire (commented):
> Do you think I should add those to properties too?

Yes, I think we should promote modern code.

docs/en/cookbook/aggregate-fields.rst
SenseException: I don't see any usage of the return value so I removed it.

greg0ire:
> SenseException requested a review from greg0ire 2 hours ago


What about adding types to properties?

SenseException:
I can add those too. PHP 7.3 is close to the end of bugfix support and adding property types would have been part of another PR, but sure: why not start now?

greg0ire (approved):

docs/en/cookbook/aggregate-fields.rst
```diff
@@ -64,31 +73,35 @@ Our entities look like:
...
-        public function __construct($account, $amount)
+        public function __construct(Account $account, int $amount)
         {
             $this->account = $account;
             $this->amount = $amount;
             // more stuff here, from/to whom, stated reason, execution date and such
         }
     
-        public function getAmount()
+        public function getAmount(): Amount
```
svycka, line 104: the return value should be int instead of Amount

Issue #4257, opened 2014-12-15
DDC-3449: Single scalar Result and Hidden field


doctrinebot:
Jira issue originally created by user tgallice:

I have try to get a single scalar result but my query contains a HIDDEN field. The result is the `NonUniqueResultException` exception.

gitbugr:
This is still an issue.

SenseException:
This issue needs a better description.

@gitbugr Would you like to add a summary, how to reproduce, the current behavior and the expected behavior to this issue? Having a code example or a PR with a failing test is helpful too.

gitbugr:
@SenseException Sure.

The most basic example of reproducing what is happening above would be: 

```php
$query = $em->createQuery('SELECT COUNT(u.id), u.age AS HIDDEN userAge FROM MyProject\Model\User u HAVING userAge > 20');

$result = $query->getResult();
/**
 *  $result = [
 *    0 => [ 
 *      1 => 1234 
 *    ]
 *  ]
 */

$singleScalarResult = $query->getSingleScalarResult(); // error due to structure of result above
```

The error returned when running `getSingleScalarResult();` would be 
```
Doctrine\ORM\NonUniqueResultException

The query returned a row containing multiple columns. Change the query or use a different result function like getScalarResult().
```

SenseException:
Thank you.

gitbugr:
it's worth noting that a somewhat stable "workaround" for this, that shouldn't break when this gets fixed would be to give the field you're selecting an alias, and returning it like so:

```
$query = $em->createQuery('SELECT COUNT(u.id) AS count, u.age AS HIDDEN userAge FROM MyProject\Model\User u HAVING userAge > 20');

$singleScalarResult = $query->getResult()[0]['count'];
```

knezmilos13:
Same problem here, using doctrine v2.10.1, calling `getSingleScalarResult` when loading `count`, while using another AS HIDDEN attribute, getting the following message:

`The query returned a row containing multiple columns. Change the query or use a different result function like getScalarResult().`

At `\Doctrine\ORM\Internal\Hydration\SingleScalarHydrator.php 51`

Mediagone:
I've submitted a fix to this issue: https://github.com/doctrine/orm/pull/8340

greg0ire:
The fix has been merged.

greg0ire:
Actually, it has been merged in the wrong branch :sweat_smile: 2.7 vs 2.7.x

[...]

PR #8340, opened 2020-11-13
Fix Hidden fields triggering error when using getSingleScalarResult()

Mediagone:
Using HIDDEN fields in a request was causing the "unicity" check to fail and throw a NonUniqueResultException, because we were counting raw data instead of counting gathered row data.

See https://github.com/doctrine/orm/issues/4257  for original issue.

_PS: sorry if I've done something wrong in the contribution procedure, it's my first pull request :)_

### DIFF
--- a/tests/Doctrine/Tests/ORM/Hydration/SingleScalarHydratorTest.php
+++ b/tests/Doctrine/Tests/ORM/Hydration/SingleScalarHydratorTest.php
@@ -64,7 +64,7 @@ class SingleScalarHydratorTest extends HydrationTestCase
      */
     public function testHydrateSingleScalar($name, $resultSet)
     {
-        $rsm = new ResultSetMapping;
+        $rsm = new ResultSetMapping();
         $rsm->addEntityResult(CmsUser::class, 'u');
         $rsm->addFieldResult('u', 'u__id', 'id');
         $rsm->addFieldResult('u', 'u__name', 'name');
@@ -124,21 +124,20 @@ class SingleScalarHydratorTest extends HydrationTestCase
     }
 
     /**
-     *
      * @dataProvider singleScalarResultSetWithHiddenFieldProvider
      */
     public function testHydrateSingleScalarWithHiddenField($name, $resultSet)
     {
-        $rsm = new ResultSetMapping;
+        $rsm = new ResultSetMapping();
         $rsm->addScalarResult('u__id', 'id', 'string');
 
-        $stmt = new HydratorMockStatement($resultSet);
+        $stmt     = new HydratorMockStatement($resultSet);
         $hydrator = new SingleScalarHydrator($this->_em);
 
         if ($name === 'result1') {
             $result = $hydrator->hydrateAll($stmt, $rsm);
             $this->assertEquals('1', $result);
-            
+
             return;
         }
 
@@ -156,5 +155,4 @@ class SingleScalarHydratorTest extends HydrationTestCase
             return;
         }
     }
-
 }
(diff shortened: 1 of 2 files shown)


Mediagone:
I didn't add "use function" statements because there was none in the original file, but I see errors in the commit. 
Should I add them?

beberlei:
@Mediagone i don't understand what you mean with HIDDEN fields? This behavior here is the same that DBAL and by extension PDO have when you use `Statement->fetchColumn()` on a row with more results. It returns just the first one, and that is to be expected at the moment.

If we want this to change, it would be a BC break.

Mediagone:
Hi @beberlei , let's take this example:
```
'SELECT 
COUNT(u.id) as count, 
(u.posts_count + u.likes_count) AS HIDDEN score
FROM CmsUser u
ORDER BY score'
```
In this case, `getScalarResult()` will return an array of 1 result ("count"), because "score" is only used internally and not returned since it uses the HIDDEN DQL keyword.
However, `getSingleScalarResult()` will throw a NonUniqueResultException because it counts selected fields, including hidden ones (even if the final result will filter out some fields, and return a single result).

tests/Doctrine/Tests/ORM/Hydration/SingleScalarHydratorTest.php
```diff
@@ -90,4 +90,69 @@ public function testHydrateSingleScalar($name, $resultSet)
...
+
+            return;
+        }
+
+        if ($name === 'result2') {
+            $result = $hydrator->hydrateAll($stmt, $rsm);
+            $this->assertEquals(null, $result);
+
+            return;
+        }
```
SenseException, line 149: Why not injecting `'1'` and `null` with the data provider instead of the name? Then you can drop the if-blocks.

tests/Doctrine/Tests/ORM/Hydration/SingleScalarHydratorTest.php
```diff
@@ -90,4 +90,69 @@ public function testHydrateSingleScalar($name, $resultSet)
...
+
+        if ($name === 'result3') {
+            $this->expectException(NonUniqueResultException::class);
+            $hydrator->hydrateAll($stmt, $rsm);
+
+            return;
+        }
```
SenseException, line 156: I'd like to see the exception case in its own test method.

tests/Doctrine/Tests/ORM/Hydration/SingleScalarHydratorTest.php
```diff
@@ -90,4 +90,69 @@ public function testHydrateSingleScalar($name, $resultSet)
...
+                    ['u__id' => '2'],
+                ],
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider singleScalarResultSetWithHiddenFieldProvider
+     */
+    public function testHydrateSingleScalarWithHiddenField($name, $resultSet)
```
SenseException, line 129:
```suggestion
    public function testHydrateSingleScalarWithHiddenField(?string $name, array $resultSet) : void
```

SenseException (changes requested):

tests/Doctrine/Tests/ORM/Hydration/SingleScalarHydratorTest.php
Mediagone: Because it's my first PR, I took the original test structure without changing it but I fully agree with you, these tests need separate methods. I'll do that tomorrow!

Mediagone:
Tests updated 🙂

SenseException (approved):
I'd like to see another reviewer checking the changes and if the tests are sufficient.

beberlei (approved):

beberlei:
thank you

PR #8341, opened 2020-11-13
Use LockMode::NONE as default lock mode

BenMorel:
As discussed in https://github.com/doctrine/dbal/issues/4391, the `$lockMode` parameter should not be nullable anywhere; what should be used as a default is `LockMode::NONE`.

There was a long standing confusion in the DBAL that can be summarized to:

- in all platforms but SQL Server / SQL Anywhere, `null` was equivalent to `LockMode::NONE`;
- in SQL Server / SQL Anywhere, `LockMode::NONE` resulted in using `WITH (NOLOCK)`, which changed the effective isolation level to `READ UNCOMMITTED` for the table; this is not the contract of `LockMode::NONE`, and was qualified as a bug.

This has been fixed in https://github.com/doctrine/dbal/pull/4400, and now `null` is not a valid value anymore for lock modes.

This PR is the sequel to the above:

- `$lockMode` defaults to `LockMode::NONE` everywhere, `null` is not allowed anymore
- no transaction is required anymore when using `LockMode::NONE`, which didn't make any sense

Notes:

- the transaction requirement for `LockMode::NONE` was highly inconsistent anyway: the `Query` required one, but `EntityManager::find()` didn't
- there were many inconsistencies in the documented types for `$lockMode` as well, which did document in many places that only an `int` was acceptable, but did accept or even expect `null` at some point: [1](https://github.com/doctrine/orm/blob/e8c4d7b177b52fd2cb75522c77d601caabbb4806/lib/Doctrine/ORM/EntityManagerInterface.php#L178), [2](https://github.com/doctrine/orm/blob/e8c4d7b177b52fd2cb75522c77d601caabbb4806/lib/Doctrine/ORM/Query.php#L690), [3](https://github.com/doctrine/orm/blob/e8c4d7b177b52fd2cb75522c77d601caabbb4806/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1732)

### DIFF
--- a/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
+++ b/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ORM\Cache\Persister\Entity;
 
 use Doctrine\Common\Collections\Criteria;
+use Doctrine\DBAL\LockMode;
 use Doctrine\ORM\Cache;
 use Doctrine\ORM\Cache\CollectionCacheKey;
 use Doctrine\ORM\Cache\EntityCacheKey;
@@ -105,7 +106,7 @@ abstract class AbstractEntityPersister implements CachedEntityPersister
     public function getSelectSQL(
         $criteria,
         ?AssociationMetadata $association = null,
-        $lockMode = null,
+        $lockMode = LockMode::NONE,
         $limit = null,
         $offset = null,
         array $orderBy = []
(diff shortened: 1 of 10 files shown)


BenMorel:
Looks like composer lock file is broken; CI can't run.

BenMorel:
@beberlei Could you please take a look at this PR? This should be an easy pick.

SenseException:
You need to re-target your PR to a different branch, but I'm not sure if this is supposed to be a bugfix or an improvement for a newer branch. `composer.lock` isn't in the versioning in 2.8.x and 2.9.x anymore and shouldn't be a problem.

BenMorel:
@SenseException This is not a bugfix, rather an improvement. Should this go into 2.9.x? What's master currently, is it 3.x?

SenseException:
If master is the correct target branch, then it needs a new upmerge. It might be a current blocker.

BenMorel:
That's the point, I'm not sure if `master` is the correct branch; is it 3.x?

SenseException:
@doctrine/team-orm Is there a suggestion about the target branch for this PR?

PR #8342, opened 2020-11-18
update DBAL constraint to move away from non-existing branch

xabbuh:


xabbuh:
This will allow other projects that require Doctrine ORM (for example the Symfony Doctrine bridge) to test for compatibility with Doctrine DBAL 3.

beberlei:
Dbal 3 is not supported with PRM at the moment. We will add support in an upcoming 2.x version, until then you cannot use dbal 3 if you use the ORM

xabbuh:
@beberlei Thank you for the information. Would you be open for a PR targetting `2.8.x` that allows to install DBAL 3 and fixes the incompatibilities? I'd be willing to give it a try.

beberlei:
@xabbuh honestly that iss something we want to do ourselves, since close interference is needed, and potentially learnings to improve the API and such.

xabbuh:
@beberlei Understood, I am going to find another solution to test the Doctrine bridge with DBAL 3 then. Thanks for your time. 👍

Issue #8337, opened 2020-11-12
PreFlush event is triggered on every flush even if already flushed


Seb33300:
I am encountering an issue where doctrine events are triggered every time the flush method is called, even if the entity was persisted only once.

I am using doctrine 2.7.4 with Symfony 4.4.

Example:
``` php

// $entity is an entity with a PreFlush event attached

$entityManager->persist($entity);
$entityManager->flush(); // the PreFlush event is triggered on $entity
$entityManager->flush(); // the PreFlush event is triggered a second time on $entity even if $entity was not persisted a second time before.
```

Is it normal that the second flush trigger the event a second time?

I thought it should be triggered only if we persist the entity a second time:
``` php

// $entity is an entity with a PreFlush event attached

$entityManager->persist($entity);
$entityManager->flush(); // the PreFlush event is triggered on $entity
$entityManager->flush(); // should not trigger the event
$entityManager->persist($entity);
$entityManager->flush(); // should trigger the event a second time
```

Thank you

flaushi:
The preFlush event only has PreFlushEventArgs, which in turn do not have any entity, only the EntityManager is passed with them. So I think you mixed something up.

lcobucci:
Closing as invalid based on the discussion in #8343

PR #8343, opened 2020-11-19
Failing test for #8337

Seb33300:
Failing test for #8337

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8337Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8337Test.php
@@ -1,0 +1,89 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\EventSubscriber;
+use Doctrine\ORM\Event\PreFlushEventArgs;
+use Doctrine\ORM\Events;
+use Doctrine\Tests\OrmFunctionalTestCase;
+use Exception;
+
+/**
+ * @group GH8337
+ */
+class GH8337Test extends OrmFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function setUp() : void
+    {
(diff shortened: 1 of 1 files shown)


lcobucci:
@Seb33300 the test expectations is incorrect. The preFlush event is dispatched in the beginning of the flush operation. If you call flush twice you will have two events.

Seb33300:
If you flush twice on an entity which has modified fields the `UPDATE` query will be executed only once unless you persist again the entity before the second flush. Right?
I was expecting the same behavior with events.

So, how can we prevent event to be executed multiple times on the same entity if we have several flush?

lcobucci:
The ORM is built on top of the [Unit of Work pattern](https://www.martinfowler.com/eaaCatalog/unitOfWork.html). The flush operation commits the UoW to finalise the business transaction.

Having multiple flushes in a single business transaction is a wrong usage of the pattern (and the ORM).

Seb33300:
Unfortunately the app was built like this and it could be quite complex to change it.

I am trying to understand why queries operations are executed only on the first `flush` call by the UoW but event operations are executed on every `flush` calls.
`flush` should flush all operations (queries & events), not only queries.

lcobucci:
> I am trying to understand why queries operations are executed only on the first flush call by the UoW but event operations are executed on every flush calls.
> flush should flush all operations (queries & events), not only queries.

@Seb33300 that's a misunderstanding of the architecture. As mentioned in our [docs](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/events.html#preflush), `preFlush` is called before anything happens (even the changeset calculation).

> Unfortunately the app was built like this and it could be quite complex to change it.

That's a common mistake but IMHO we shouldn't adapt the ORM because of misusages.

It's also worth mentioning that the events used by the ORM aren't meant to trigger business rules, so if your app is also coupled to them I'd strongly recommend dispatching your own domain events.

lcobucci:
Closing as invalid based on the discussion.

PR #8347, opened 2020-11-24
Add doctrine/dbal 3 support

acasademont:
Hi all,

This is a first attempt into having DBAL 3 support on the 2.x ORM branches. There are some things that should we discussed, namely:

- Some Platforms do no longer exist on DBAL (`PostgreSqlPlatfrom`, `PostgreSql91Platfrom`, `PostgreSql92Platfrom`, `SQLServerPlatform`, `SQLServer2005Platform`, `SQLServer2008Platform`). As DBAL is most of the times fetched as a dependency of ORM this can be a problem if someone still relies on those platforms. Not sure if this should be considered a BC-break, and hence DBAL 3 would not be possible on ORM 2. Some tests that relied on those platforms have then changed.

- Some methods that the ORM uses are no longer present on DBAL 3, namely `AbstractPlatform::prefersSequences()`,  `AbstractPlatform::getSQLResultCasing()` and `AbstractPlatform::fixSchemaElementName()`. For `prefersSequences` I think I have found an acceptable workaround that should not break anything. For `getSQLResultCasing` I have created a new `SQLResultCaser` class to do it, same with `fixSchemaElementName()` and a new `SchemaElementNameFixer`. See the discussion with @morozov on https://github.com/doctrine/dbal/pull/4229

- Should we trigger deprecation warnings somewhere?

Tests pass with both DBAL 2 and DBAL 3. I have raised the minimum DBAL version to 2.12.


acasademont:
The discussion  and PR about the LockModes in DBAL is also relevant here, I had to do some magick tricks to make the tests pass in both DBAL 2 and 3

https://github.com/doctrine/orm/pull/8341

lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
```diff
@@ -91,13 +91,16 @@ public static function createApplication(HelperSet $helperSet, array $commands =
      */
     public static function addCommands(Application $cli) : void
     {
-        $cli->addCommands(
-            [
+        if (class_exists(DBALConsole\Command\ImportCommand::class)) {
```
acasademont, line 94: Should the ORM expose DBAL console commands? The new commands are not really compatible, not sure how to make them.

acasademont:
Also DBAL removed the `$_attributes` array of the `ORM\Configuration` object, we should decide if we want to keep using our own private `$_attributes` property or do the same as DBAL and move everything to individual properties, with the BC-Break that exists if someone was extending that `Configuration` class.

https://github.com/doctrine/orm/blob/3d46e078875263370565beb28733931fce7a4a9f/lib/Doctrine/ORM/Configuration.php#L67

acasademont:
Also, PHP was bumped to 7.3 in order to support DBAL 2.12

acasademont:
Also, the `MySqlPlatform` was renamed to `MySQLPlatform`, not sure the impact this can have on case-sensitive filesystems.

beberlei:
@acasademont this is something we want to do ourselves to learn from the API changes. I would advise not to put too much work into this before we map this out, sorry :-(

acasademont:
@beberlei sure! I just wanted to give it a first try, happy to help you in any way I can. I think it's a good idea to branch ORM 3 and put DBAL support there.

PowerKiKi:
@beberlei, I couldn't find any other issue talking about DBAL 3.0 support. Should we create an issue for it in [milestone 3.0.0](https://github.com/doctrine/orm/milestone/1) and close this PR ? or should we track progress through this existing PR ?

greg0ire:
Closed by https://github.com/doctrine/orm/pull/8964

PR #8348, opened 2020-11-24
[GH-8229] Prevent Illegal Inheritance Override

beberlei:
The PRs #8234 and #8329 are based on a flawed understanding of inheritance in Doctrine and need to be reverted:

* Attribute/Association Overrides are only allowed to be used for MappedSuperclass and Traits. This is documented, albeit not phrased strongly enough: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/inheritance-mapping.html#overrides

* Entities are not allowed to extend from entities without an inheritence mapping relationship (Single Table or Joined Table inheritance). If you have this use case you should instead have two entities and they each extend a common `@MappedSuperclass`.

One excemption was made to use `@AssocationOVerride` only for changing the fetch strategy. Soft-deprecated this with a TODO.

While Doctrine so far allowed these things, they are fragile and will break on certain scenarios. Please give feedback here or in new issues on the impact of this change on your projects.


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -2193,6 +2193,12 @@ public function setAssociationOverride($fieldName, array $overrideMapping)
 
         $mapping = $this->associationMappings[$fieldName];
 
+        if (isset($mapping['inherited']) && (count($overrideMapping) !== 1 || ! isset($overrideMapping['fetch']))) {
+            // TODO: Deprecate overriding the fetch mode via association override for 3.0,
+            // users should do this with a listener and a custom attribute/annotation
+            throw MappingException::illegalOverrideOfInheritedProperty($this->name, $fieldName);
+        }
```
greg0ire, line 2200: Should the parts that are about throwing be contributed to 2.8 instead? If we think there are people currently using this, we should probably not have it blow up in their face in a patch

beberlei: You are right, i should remove them for 2.7, and re-introduce them for 2.8

greg0ire (dismissed):

greg0ire (approved):

Kalyse:
> Entities are not allowed to extend from entities without an inheritence mapping relationship (Single Table or Joined Table inheritance). If you have this use case you should instead have two entities and they each extend a common @MappedSuperclass.

I think this update just severely broke my application. I'm still debugging, however, I use a ClassMetadataSubscriber to override the [name] field on the Entities to match a prefix. ie. You know how wordpess prefixes based on `wp_`, I do similar. 

Anyway, I have a MappedSuperclass called.. "Building", and then I have on entity which extends it, called "Room". 

It appears now, as though the MappedSuperClass "Building" does not existing within the `LoadClassMetadataEventArgs` event passesd to my subscriber. 

Is this intended? It loosk like all MapperSuperClasses have been removed, so any of my superclasses now do not exist with the vendor prefix. 

What ended up happening on my development machine was 20 table deletions.. so it's quite a big issue which I didn't expect from a minor patch?

beberlei:
Please open up a new issue and improve the issue description, I don't understand mapped superclasses do not have tables, and should not have table prefixes. What means "not existing within LoadClassMetadataEventArgs"?

mpdude:
#10431 suggests adding a runtime check for missing inheritance declarations

Issue #7866, opened 2019-10-12
Incorrect commit order when inserting multiple entities


AssortedParrot:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.4

#### Summary

In some cases, Doctrine will insert entities into the database in the wrong order, causing constraint violations for non-nullable foreign keys.

#### Current behavior

As a simple test case with two entities, consider an application where users can upload files: it would have a `User` and an `UploadedFile` entity.

Other than its ID, an `UploadedFile` has two fields:

- `owner` (OneToOne `User`, *not* nullable): so we can keep track of which user uploaded the file. This is not nullable because all files need to belong to a valid user.
- `lastDownloadedBy` (OneToOne `User`, nullable): because for some reason we want to keep track of who last downloaded a file. This may be null in case nobody downloaded  the file yet.

A `User` only has one field:

- `lastUploadedFile` (OneToOne `UploadedFile`, nullable): stores the last file the user uploaded, or null if the user didn't upload a file yet

From the above mapping it can be seen that the `User` must be inserted before the `UploadedFile` can be inserted: no `UploadedFile` can exist without a `User`.

In my test case, a `User` and an `UploadedFile` are created, and all fields are populated. Then, `$em->persist()` is called on the `UploadedFile` and then the `User`. During the `$em->flush()`, Doctrine will try to insert the `UploadedFile` first, which will fail (because the `owner` field is set to `NULL` since the `User` is not inserted yet,) causing an exception.

#### How to reproduce

The test case mentioned above can be found in [this repository](https://github.com/AssortedParrot/doctrine-incorrect-commit-order). To reproduce it, clone the repository, run `composer install` and then run `php test.php`. It will create a SQLite database in the current directory, create the schema and then try to insert a `User` and an `UploadedFile` object into the database.

The SQL logs show that Doctrine is trying to `INSERT` the `UploadedFile` object first, which fails. It will then roll back the transaction and throw an exception.

#### Expected behavior

I would expect the following behavior:

- Insert the `User` object first (with the `lastUploadedFile` field set to `NULL`)
- Insert the `UploadedFile` object (the `owner` field can now be set to the ID of the `User` we just inserted)
- Update the `User` object (set `lastUploadedFile` to the ID of the `UploadedFile` we just inserted)

### Additional information

It should be noted that I found many small things that, when changed, caused Doctrine to exhibit the expected behavior:

[...]

PR #8349, opened 2020-11-25
Fix commit order

tomaszgaw:
Fixes visit method of CommitOrderCalculator.

Without the fix the insertion order is not properly calculated when a cycle within the graph involves more than 2 classes, leading to a violation of not null constraints on the owning side of ManyToOne relation.

With the fix when a cycle is detected a node is revisited ignoring edges of weight 0.


jhony1104:
This PR fixes [#7866](https://github.com/doctrine/orm/issues/7866).

@tomaszgaw: if you are interested in adding a test for this, it can be found here: https://gist.github.com/jhony1104/3c0793be7f1ee2fb9b933e1ee3734782

What is the status of this PR?

derrabus:
> What is the status of this PR?

Well, it has been open for a while. If it's still relevant, it needs a rebase and a functional test that reproduces the issues that it tries to fix.

beberlei:
We also have a bunch of commit order related PRs already. Its very complex algo with all rhe side effects and requirements, therefore it requires lots of maintainer time to review and trust to merge.

jhony1104:
The product I am working on runs in a lof of issues with the commit order calculator. Because of this we are using a modified fork of doctrine wich changes the commit order calculator.These changes were made to 2.9.6.

A few weeks ago I war updating doctrine and since the plain doctrine was still having the same issues and our old patch was also no longer working as expeted I tried some of the PRs on this repos.

I am sorry to report, that this PR can - in some cases - still calculate the wrong commit order. In fact, I am not sure if the algorithm used (modified topolocical sorting with weights) can - without larger changes than modifying weights - ever reliably calculate the correct commit order.

For the software I am working on, I did create a new patch that treats all edges as required (not nullable) and on detecting a loop backtracks to the last optional (nullable) edge, skips it an continues. This is of course much less performant, but should allways be correct. It can be found [here](https://github.com/jhony1104/doctrine-orm). I am not sure if I should create a PR for it (performance, larger changes, ...).

mpdude:
@jhony1104 and @tomaszgaw you might be valuable reviewing #10547.

Issue #8350, opened 2020-11-26
PHP 8 compatibility


BenMorel:
Hi, PHP 8 being released today, it would be nice to tag a release compatible with it.

I tried running tests on PHP 8 on the `2.7` branch, and a lot of tests fail.

Do you need help to fix those tests? I can do so today, if you wish.

BenMorel:
Oh, I can see that some work has been done on `2.8.x` for PHP 8 compatibility. I think we can expect PHP 8 support in 2.8.0 then?

Tests still fail in the `2.8.x` branch though; do you need help in fixing them?

Finally, when can be expect `2.8.0` to be released? Thanks in advance for your insights!

beberlei:
Yes 2.8, will be released within days

andig:
2.8 seems currently tied to dbal 2.10, but not dbal 3:

    - doctrine/orm 2.8.x-dev requires doctrine/dbal ^2.10.0 -> found doctrine/dbal[v2.10.0, ..., 2.12.x-dev] but it conflicts with your root composer.json require (^3.0).

fezfez:
> 2.8 seems currently tied to dbal 2.10, but not dbal 3:
> 
> ```
> - doctrine/orm 2.8.x-dev requires doctrine/dbal ^2.10.0 -> found doctrine/dbal[v2.10.0, ..., 2.12.x-dev] but it conflicts with your root composer.json require (^3.0).
> ```

Yep, i think declaring 2.8 compatibility with PHP 8.0 is useless because 2.x will not support doctrine/dbal:3.x

@beberlei what do you think ?

BenMorel:
> Yep, i think declaring 2.8 compatibility with PHP 8.0 is useless because 2.x will not support doctrine/dbal:3.x

DBAL 2.12 supports PHP 8, what’s the issue exactly?

fezfez:
> > Yep, i think declaring 2.8 compatibility with PHP 8.0 is useless because 2.x will not support doctrine/dbal:3.x
> 
> DBAL 2.12 supports PHP 8, what’s the issue exactly?

my bad, didnt see that

BenMorel:
@beberlei Is there anything I can do to help finalize `2.8`?

bpesch:
is there an eta on this?

BenMorel:
Doctrine ORM 2.8.0 [was just released](https://github.com/doctrine/orm/releases/tag/2.8.0)! Thank you :+1:

PR #8352, opened 2020-11-27
php 8

psohm:
doctrine/orm 2.7.4 requires php ^7.1 -> your PHP version (8.0.0) does not satisfy that requirement.


yethee:
In 2.8 will be added support PHP 8.0. See for details #8350

PR #8356, opened 2020-11-30
fix(persistent-collection): check "association" is not nullable before using it as an array

Kocal:
Hi, 

Today I've faced an issue with Doctrine 2.7 and https://github.com/KnpLabs/DoctrineBehaviors when configuring the parameter `doctrine_behaviors_translatable_fetch_mode` to `LAZY` (see [KnpDoctrineBehaviors Translatable configuration](https://github.com/KnpLabs/DoctrineBehaviors/blob/master/docs/translatable.md#configuration)):

image

By using xdebug, I found it was coming from the method `PersistentCollection#get()` which try to use `$this->association` when it is equal to `null`: 

There are already some methods that check if `$this->association !== null`: 
- `changed`: https://github.com/doctrine/orm/blob/e0eb82a/lib/Doctrine/ORM/PersistentCollection.php#L294
- `remove`: https://github.com/doctrine/orm/blob/e0eb82a/lib/Doctrine/ORM/PersistentCollection.php#L357
- `removeElement`: https://github.com/doctrine/orm/blob/e0eb82a/lib/Doctrine/ORM/PersistentCollection.php#L380
- `count`: https://github.com/doctrine/orm/blob/e0eb82a/lib/Doctrine/ORM/PersistentCollection.php#L443

So I've applied the same patch on all other methods that use `$this->association` without checking its value.

WDYT? Thanks!

### DIFF
--- a/lib/Doctrine/ORM/PersistentCollection.php
+++ b/lib/Doctrine/ORM/PersistentCollection.php
@@ -163,7 +163,7 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
 
         // If _backRefFieldName is set and its a one-to-many association,
         // we need to set the back reference.
-        if ($this->backRefFieldName && $this->association['type'] === ClassMetadata::ONE_TO_MANY) {
+        if ($this->backRefFieldName && $this->association !== null && $this->association['type'] === ClassMetadata::ONE_TO_MANY) {
             // Set back reference to owner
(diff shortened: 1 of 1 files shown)


Kocal:
Hum, I don't understand why the CI is failing. For me my changes are unrelated to the failed tests.. :/

Kocal:
Superseded by #10599

Issue #8365, opened 2020-12-04
Break in 2.8: Cannot bind QueryBuilder parameters with \IteratorAggregate entities: leads to incorrect SQL generated code


ambroisemaupate:
### Bug Report

Binding QueryBuilder query parameters to entities that implements \IteratorAggregate leads to incorrect DQL generated code.

Q | A
-- | --
BC Break | yes
Version | 2.8.x-dev `418587b`
Working on | 2.7 branch

It seems that QueryBuilder tries to serialize object instead of using its *primary key*:

```php
// Node is a valid Doctrine entity with primary key
// Node implements \IteratorAggregate, \Countable
$node = $this->_em->find(Node::class, $nodeId);

// …

$qb = $this->createQueryBuilder('t');
$qb->select('t')
    ->innerJoin('t.nodeSources', 'ns')
    ->andWhere($qb->expr()->eq('ns.node', ':node'))
    ->setParameter('node', $node);
```

leads into:

```
An exception occurred while executing […]
n1_.node_id = ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ? 
with params [7, false, false, false, false, false, false, false, false, false, false, false, false]:
```

To fix this I need to explicitely use primary key `$node->getId()`:

```php
$qb = $this->createQueryBuilder('t');
$qb->select('t')
    ->innerJoin('t.nodeSources', 'ns')
    ->andWhere($qb->expr()->eq('ns.node', ':node'))
    ->setParameter('node', $node->getId());
```

https://github.com/doctrine/orm/issues/8181

ambroisemaupate:
Possibly due to 

https://github.com/doctrine/orm/blob/2.8.x/lib/Doctrine/ORM/AbstractQuery.php#L423

ambroisemaupate:
Here is a test case to reproduce this issue: 

https://github.com/ambroisemaupate/orm/commit/eca3a3fb158e8d39de2b6db2d552d10c140544bd

beberlei:
Fixed by #8371

PR #8368, opened 2020-12-04
BC 2.8.x: process parameter value as Entity before checking if it implements Traversable

ambroisemaupate:
Binding AbstractQuery parameters to entities that implements \Traversable leads to incorrect DQL generated code.

Q | A
-- | --
BC Break | yes
Version | 2.8.x-dev `418587b`
Working on | 2.7 branch
Related to | https://github.com/doctrine/orm/issues/8365


beberlei:
Dropped in favor of https://github.com/doctrine/orm/pull/8371

PR #5571, opened 2015-12-22
#1168 Add support for array parameters on the SQLFilter

mnogales:

PR #8375, opened 2020-12-06
Support for Array parameters in SQL filters

### LINKED ISSUE #2624: DDC-1952: Add support for array parameters on the SQLFilter
Jira issue originally created by user holtkamp:

The SQLFilter currently only accepts string parameters which would result in SQL like:

```
"tableAlias.column = '$filterParameter'"
```

To filter an Entity that has a lifecycle, this can be usefull to filter Entities that are in a specific state, for example:

```
"tableAlias.state = 1"
```

To be able to apply the filter on an Entity that can be in multiple states, it is usefull to be able to assign an array of states using setParameter:

```
$allowedStates = array(1,2,3,4);
$filter->setParameter('allowedStatesParam', $allowedStates);
sprintf("tableAlias.state IN (%s)", implode(',', $this->getParameter('allowedStatesParam')));
```

to eventually result in:

```
"tableAlias.state IN (1,2,3,4)"
```

However, this is currently not supported, it seems to go wrong on the PDO::quote() of the parameter. The SQL works ok when setting it statically in the filter, not taking the parameter into account.

It would be nice to have support for arrays on the setParameter()

beberlei:
Supersedes https://github.com/doctrine/orm/pull/5571 by @mnogales

Fixes #2624 

The original PR had a conceptual problem by putting the array detection directly into `getParameter()`, but this could be a BC break for array DBAL types and it could be a risk, because the generated SQL must be written by the filter implementor, so magically returning a quoted list instead a quoted value could break things.

Instead you must combine your generated SQL IN query with `$this->getParameterList($name)`:

```php
class CMSCountryFilter extends SQLFilter
{
    public function addFilterConstraint(ClassMetadata $targetEntity, $targetTableAlias)
    {
        if ($targetEntity->name !== CmsAddress::class || !$this->hasParameter('country')) {
            return "";
        }

        return $targetTableAlias.'.country IN (' . $this->getParameterList('country') . ')';
    }
}
```
To use `getParameterList` you must also pass the values using the new `setParameterList` method. It does not support type inference as we don't need to inspect the array elements then (which could be different).

```php
$em->getFilters()->getFilter('country')->setParameter('country', ['de', 'es']);
```

Passing a list/single value and retrieving a single/list value (mixed up) will throw a new `FilterException`.


lib/Doctrine/ORM/Query/Filter/SQLFilter.php
```diff
@@ -111,6 +115,40 @@ final public function getParameter($name)
...
+     * @throws InvalidArgumentException
+     */
+    final public function getParameterList(string $name): string
+    {
+        if (! isset($this->parameters[$name])) {
+            throw new InvalidArgumentException("Parameter '" . $name . "' does not exist.");
+        }
+
+        $param         = $this->parameters[$name];
+        $isTraversable = is_array($param['value']); // WARNING: no Traversable support, because it might break on hashing
```
SenseException: > // WARNING: no Traversable support

Still, the variable is named `$isTraversable` :-D.

tests/Doctrine/Tests/ORM/Functional/SQLFilterTest.php
```diff
@@ -328,6 +328,36 @@ public function testSQLFilterSetParameterInfersType(): void
...
+        // Setup mock connection
+        $conn = $this->getMockConnection();
+        $conn->expects($this->at(0))
+            ->method('quote')
+            ->with($this->equalTo('en'))
+            ->will($this->returnValue("'en'"));
+        $conn->expects($this->at(1))
+            ->method('quote')
+            ->with($this->equalTo('es'))
+            ->will($this->returnValue("'es'"));
```
SenseException: Please use `returnValueMap()` here. The invocation order of the mock isn't relevant and is covered in the resulting assertion anyways.

tests/Doctrine/Tests/ORM/Functional/SQLFilterTest.php
```diff
@@ -401,7 +431,7 @@ public function testQueryGenerationDependsOnFilters(): void
         $conf = $this->_em->getConfiguration();
         $conf->addFilter('country', '\Doctrine\Tests\ORM\Functional\CMSCountryFilter');
         $this->_em->getFilters()->enable('country')
-            ->setParameter('country', 'en', DBALType::STRING);
+            ->setParameter('country', ['en'], DBALType::STRING);
```
SenseException: Why changing an existing test and filter?

SenseException (changes requested):
Please also cover the two exception cases in `getParameterList()`.

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
beberlei: Removed traversable support explicitly now.

tests/Doctrine/Tests/ORM/Functional/SQLFilterTest.php
beberlei: Because we have two filters in the test, one for locale one for country. By changing one to be a list we maximize test coverage without adding more tests.

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
```diff
@@ -108,7 +137,46 @@ final public function getParameter($name)
...
+            throw FilterException::cannotConvertListParameterIntoSingleValue($name);
+        }
+
+        $param = $this->parameters[$name];
+
+        return $this->em->getConnection()->quote($param['value'], $param['type']);
+    }
+
+    /**
+     * Gets a parameter to use in a query assuming its a list of entries.
```
greg0ire:
```suggestion
     * Gets a parameter to use in a query assuming it's a list of entries.
```

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
```diff
@@ -108,7 +137,46 @@ final public function getParameter($name)
...
+     *
+     * The function is responsible for the right output escaping to use the
+     * value in a query, seperating each entry by comma to inline it into
```
greg0ire:
```suggestion
     * value in a query, separating each entry by comma to inline it into
```

SenseException (dismissed):
After the typos that @greg0ire mentioned are fixed, this is :+1:

greg0ire (approved):

PR #8377, opened 2020-12-06
Add Doctrine\ORM\Query\Expr::mod()

### LINKED ISSUE #6739: Doctrine\ORM\Query::mod() is missing
While `mod()` is a *native* [DQL function](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html#dql-functions), it seems not supported by the ExpressionBuilder. 

Not a big issue, but as a user I would expect it there... 

Untested, but it would be probably something like:

```php
public function mod($x, $y)
{
    return new Expr\Func('MOD', array($x, $y));
}
```

beberlei:
Backports https://github.com/doctrine/orm/pull/6747 to 2.x branch, slated for 2.9

Fixes #6739

### DIFF
--- a/docs/en/reference/query-builder.rst
+++ b/docs/en/reference/query-builder.rst
@@ -513,6 +513,9 @@ complete list of supported helper methods available:
         // Example - $qb->expr()->sqrt('u.currentBalance')
         public function sqrt($x); // Returns Expr\Func
 
+        // Example - $qb->expr()->mod('u.currentBalance', '10')
+        public function mod($x); // Returns Expr\Func
+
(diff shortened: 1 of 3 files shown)


greg0ire (commented):
In https://github.com/doctrine/orm/blob/2.8.x/docs/en/reference/query-builder.rst#the-expr-class , we say we provide a complete list of helper methods, so I guess this one needs to be added to that list.

greg0ire (approved):

SenseException (approved):

PR #8378, opened 2020-12-06
[GH-6578] Add validation that inherited entity class is mapped in discriminator.

### LINKED ISSUE #6558: Invalid discriminator maps should be detected during schema/mapping validation
Given the following structure of entities, which are discriminated over two levels, fetching a _Developer_ using the repository from the root entity (_Person_) results in exceptions.

doctrine2

The issue is that an _Employee_ object is being instantiated, which fails since that class is abstract. So it seems that Doctrine is not following the tree until the leaf nodes. Fetching the _Developer_ using the repository from it's direct parent (so the _Employee_) works fine.

This may, or may not, give some insights as well:
Persisting a _Developer_ works ok, except that the order of insert queries executed are wrong (in my opinion). The _Person_ is inserted first, followed by the _Developer_ and the _Employee_ is inserted last. (So: root -> leaf -> 2nd layer). This forces us to not have foreign keys on the id columns of the _Employee_, _Staff_ and _Developer_.

Note:
I will submit some test cases which test the two scenarios of fetching using the different repositories.

Edit:
PR with the test case: https://github.com/doctrine/doctrine2/pull/6559

beberlei:
Without tests as not failing all the schema validator tests with our existing models is proof enough, i flipped a condition to see it breaking hard.

Supersedes #6578
Fixes #6558


lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -255,6 +256,12 @@ public function validateClass(ClassMetadataInfo $class)
             }
         }
 
+        if (! $class->isInheritanceTypeNone() && ! $class->isRootEntity() && array_search($class->name, $class->discriminatorMap) === false) {
```
vovkin, line 259: Hello @beberlei,

shouldn't we skip this for abstract classes?
They were allowed to not be a part of mapping, but just be parent class for classes that are mapped in the root entity.

Now with update to Doctrine version 2.9.1 we are getting this error by schema validator
```
Entity class 'App\Entity\Order\CardSuborder' is part of inheritance hierarchy, but is not mapped in the root entity 'App\Entity\Suborder' discriminator map. All subclasses must be listed in the discriminator map.
```

where `Suborder` is a root entity, and `CardSuborder` is an abstract class extended from it, and then we have other not abstract classes extended from `CardSuborder`, and mapped in `Suborder`.

BenMorel: Same issue for me in #8771; any solution for this, @beberlei?

PR #8380, opened 2020-12-07
[GH-7128] Improve OneToManyRequiresMappedBy Exception message

### LINKED ISSUE #7128: error message provides insufficient information for OneToMany with missing mappedBy
The error message provides only half of the necessary information. It tells which attribute is concerned, but not on which entity, between which two entities, or anything else giving the required hint as to which mapping is wrong. 

Adding the `sourceEntity` to the error message would provide that missing information and help locate the malformed mapping significantly. used in:
https://github.com/doctrine/doctrine2/blob/43c252458b1c46081b828555013bb3fb0bb3c389/lib/Doctrine/ORM/Mapping/ClassMetadata.php#L730

(since the called static function is public, I shy away from providing a pull request)

beberlei:
Fixes #7128

### DIFF
--- a/lib/Doctrine/ORM/Mapping/MappingException.php
+++ b/lib/Doctrine/ORM/Mapping/MappingException.php
@@ -278,14 +278,13 @@ class MappingException extends ORMException
         return new self(sprintf("Result set mapping name on entity class '%s' is not defined.", $className));
     }
 
-    /**
-     * @param string $fieldName
-     *
-     * @return MappingException
-     */
-    public static function oneToManyRequiresMappedBy($fieldName)
+    public static function oneToManyRequiresMappedBy(string $entityName, string $fieldName): MappingException
     {
-        return new self(sprintf("OneToMany mapping on field '%s' requires the 'mappedBy' attribute.", $fieldName));
+        return new self(sprintf(
+            "OneToMany mapping on entity '%s' field '%s' requires the 'mappedBy' attribute.",
+            $entityName,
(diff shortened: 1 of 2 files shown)


drupol (dismissed):

lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -251,14 +251,13 @@ public static function nameIsMandatoryForSqlResultSetMapping($className)
         return new self("Result set mapping name on entity class '$className' is not defined.");
     }
 
-    /**
-     * @param string $fieldName
-     *
-     * @return MappingException
-     */
-    public static function oneToManyRequiresMappedBy($fieldName)
+    public static function oneToManyRequiresMappedBy(string $entityName, string $fieldName): MappingException
```
greg0ire, line 281:
```suggestion
    public static function oneToManyRequiresMappedBy(string $entityName, string $fieldName): self
```

This is a BC-break. If you think that class is not part of the BC promise, then maybe let's explicitly mark it as `@internal`?

beberlei: Yes its a BC break, but this exception is only used in `ClassMetadataInfo` in a code block that cannot be overwritten. Users are not expected to use it in any extension points themselves, such as a custom `MappingDriver`. `MappingException` cannot be `@internal` and adding that to every method seems a bit overhead or? I think there is no risk to ignore this.

PR #8384, opened 2020-12-08
[GH-6394] Bugfix: IdentifierFlattener support for association non-object values.

### LINKED ISSUE #6393: Exception in `IdentifierFlattener::flattenIdentifier` - column name used instead of field name in identifier extraction

beberlei:
Fixes #6394 Fixes #6393

`BasicEntityPersister::update` passes the wrong representation of identifiers to `assignDefaultVersionValue`. Instead of grabbing the values from `UnitOfWork::getEntityIdentifier` the method `ClassMetadataInfo::getIdentifierValues` must be used as in all other places where the identify flattener is ultimately used.

The difference is that `UnitOfWork` identifiers contain the scalar values representing an association identifier, but `ClassMetadataInfo` representation returns the actual entity object.

/cc @goetas

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -394,7 +394,7 @@ class BasicEntityPersister implements EntityPersister
         $this->updateTable($entity, $quotedTableName, $data, $isVersioned);
 
         if ($isVersioned) {
-            $id = $this->em->getUnitOfWork()->getEntityIdentifier($entity);
+            $id = $this->class->getIdentifierValues($entity);
 
             $this->assignDefaultVersionValue($entity, $id);
         }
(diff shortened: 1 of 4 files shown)


SenseException (approved):

goetas:
It is beautify to see this fixed! Thanks @beberlei !

PR #1569, opened 2015-11-24
Add EAGER_BATCHED fetchmode to OneToMany association

woutervanvliet:
In regular EAGER mode, doctrine adds a JOIN for a OneToMany association
forcing the database to load more data in memory than essentially
needed. In a classic Article => Tag concept, each article would be
loaded one extra time for every addition tag.

LAZY mode issues a query for every article to load all associated tags,
also adding more overhead to the database than needed.

This new fetchmode addreses this issue by loading all tags associated to
an article into one additional query.

PR #8391, opened 2020-12-13
[GH-1569] Optimize eager fetch for collections to batch query

beberlei:
This supersedes and adjusts #1569 to current Doctrine 2.17 branch.

When you configure a one-to-many collection to be `fetch=EAGER`, then instead of immediately loading all collections it batches the fetch operation at the end by selecting all entities for all collections. Collections are loaded in batches of 100, configurable by `Configuration::setEagerFetchBatchSize`.

Changing the collection fetch mode can be done on a per query basis with `$query->setFetchMode('rootEntity', 'association', ClassMetadata::FETCH_EAGER);`. This can be done as an alternative to fetch join collection valued associations.

Open points:
- [x] Naming: Its not really using subselect, alternative name could be "Batch". It supersede EAGER.
- [x] requires removal of WITH keyword or prevention of WITH keyword when association is fetch=eager
- [x] The batch loading o collections requires handling of maximum number of items in WHERE IN clause and potentially chunk it into multiple findBy operations.
- [x] Should work across multiple layers, i.e. being recursive
- [x] For fetch=eager the loading is delegated to persisters, here its going through the whole repository stack. Wondering if that must be improved code wise to go directly to persisters.
- [x] There is no validation in ClassMetadataInfo if the selected fetch mode makes no sense for the entity.
- [x] Does not work with ManyToMany, because they don't have a mapped field. Logic needs to be deferred to CollectionPersister.

Fixes #4762


danydev:
Any update on this? It's something that could land in doctrine sooner or after or just a a PoC?. Can we help somehow to move it forward?

dmelskyi:
Hello everyone, any updates on this topic? Any problems besides existing Merge Conflict?

kgasienica:
Any details? It already took like 7 years :)

xepozz:
Up

mpdude:
To summarize the suggested changes with my own words, this hooks into the place where entitites are being created during hydration. Thus, it sees all "new" entities, i. e. at the are first re-created from database values.

It tracks all those entities and their one-to-many (also many-to-many?) collections that are configured with `fetch=SUBSELECT`.

After the entities have been loaded, for every such collection a new query is performed through the target (referred-to) entity class' `EntityRepository` to find all of those entities by means the foreign key implementing the to-many association.

Then, loop over all those entities and add them to their respective collections to initialize those.

Not sure if `indexBy`/`orderBy` need to be taken into account.

mpdude:
@beberlei Given your experience writing tricky solutions using subselects and doing AST manipulations like the Paginator component does, do you think it would be possible to write a helper component that could provide this feature and be used to do the "context sensitive optimization"? I also have the "multi-step hydration" explained at https://ocramius.github.io/blog/doctrine-orm-optimization-hydration/ in mind.

Idea along the lines of:

```php
// Assume User has to-many association to Address in User::$addresses
$query = $entityManager->createQuery('SELECT u FROM ' . User::class . ' u WHERE ...');
$result = $query->getResult();

InitializeCollection::for($query, 'u.addresses');
```

What this would be supposed to do:
* Figure out the class that `u` refers to
* Clone and modify `$query` to `SELECT u.id ...` only
* Run a `SELECT _tmp1, _tmp2 FROM {class of u} _tmp1 JOIN _tmp1.addresses _tmp2 WHERE _tmp1.id IN ({modified query})`

Might be more tricky to get right with composite primary keys?

tripany:
Any news on this?

belmeopmenieuwesim:
soon ™️

JanTvrdik:
> [Disallow use of fetch=SUBSELECT on to-one associations.](https://github.com/doctrine/orm/pull/8391/commits/ff28ba8080e378399c32237a41814cc2fd86d757)

Is there a technical reason to disallow this? Would it not still be benefical to fetch even has one associations in it's own query when the cardinality of the target entity is much lower to avoid fetching duplicated data?

beberlei:
@JanTvrdik yes, but that is what fetch=EAGER does for to one already, either by doing a join, or by triggering all laods at once.

JanTvrdik:
@beberlei I see, that makes sense. Will this work with [Doctrine\ORM\AbstractQuery::setFetchMode](https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/dql-doctrine-query-language.html#temporarily-change-fetch-mode-in-dql)?

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3176,6 +3241,30 @@ public function loadCollection(PersistentCollection $collection)
...
+            ];
+        }
+
+        $id     = $this->identifierFlattener->flattenIdentifier(
+            $sourceClass,
+            $sourceClass->getIdentifierValues($collection->getOwner())
+        );
+        $idHash = implode(' ', $id);
+
+        $this->eagerLoadingCollections[$name]['items'][$idHash] = $collection;
```
mpdude: Has nothing to do with this PR, but this is a pattern I've seen very often across the codebase. Maybe we should look into factoring this out (in general) in another PR.

beberlei:
@JanTvrdik yes, its now integrated with `setFetchMode`, this is the only way this becomes useful anyways :-)

tests/Doctrine/Tests/ORM/Functional/EagerFetchCollectionTest.php
```diff
@@ -0,0 +1,164 @@
...
+    {
+        parent::setUp();
+
+        try {
+            $this->_schemaTool->createSchema([
+                $this->_em->getClassMetadata(EagerFetchOwner::class),
+                $this->_em->getClassMetadata(EagerFetchChild::class),
+            ]);
+        } catch (Exception $e) {
+        }
```
greg0ire: You could leverage this method here: https://github.com/doctrine/orm/blob/0b7fe1862e2594fbc77eec8e4af66c42dabba9cd/tests/Doctrine/Tests/OrmFunctionalTestCase.php#L347-L354

tests/Doctrine/Tests/ORM/Functional/EagerFetchCollectionTest.php
```diff
@@ -0,0 +1,164 @@
...
+        $this->_em->clear();
+
+        $owner = $this->_em->find(EagerFetchOwner::class, $owner->id);
+
+        $afterQueryCount = count($this->getQueryLog()->queries);
+        $this->assertCount(2, $owner->children);
+
+        $anotherQueryCount = count($this->getQueryLog()->queries);
+
+        $this->assertEquals($anotherQueryCount, $afterQueryCount);
```
greg0ire:
```suggestion
        $this->assertQueryCount($afterQueryCount, 'no extra queries should have been performed');
```

tests/Doctrine/Tests/ORM/Functional/EagerFetchCollectionTest.php
```diff
@@ -0,0 +1,164 @@
...
+        $this->assertCount(3, $this->_em->find(EagerFetchOwner::class, $owner2->id)->children);
+
+        $this->_em->clear();
+
+        $beforeQueryCount  = count($this->getQueryLog()->queries);
+        $owners            = $this->_em->getRepository(EagerFetchOwner::class)->findAll();
+        $anotherQueryCount = count($this->getQueryLog()->queries);
+
+        // the findAll() + 1 subselect loading both collections of the two returned $owners
+        $this->assertEquals($beforeQueryCount + 2, $anotherQueryCount);
```
greg0ire:
```suggestion
        $this->assertQueryCount(
                $anotherQueryCount = $beforeQueryCount + 2,
                'there should have been 2 extra queries: the findAll() + 1 subselect loading both collections of the two returned $owners'
        );
```

tests/Doctrine/Tests/ORM/Functional/EagerFetchCollectionTest.php
```diff
@@ -0,0 +1,164 @@
...
+
+        $this->assertCount(2, $owners[0]->children);
+        $this->assertCount(3, $owners[1]->children);
+
+        // both collections are already initialized and count'ing them does not make a difference in total query count
+        $this->assertEquals($anotherQueryCount, count($this->getQueryLog()->queries));
```
greg0ire:
```suggestion
        $this->assertQueryCount(
                $anotherQueryCount,
                "there should be no extra queries: both collections are already initialized and count'ing them does not make a difference in total query count"
        );
```

tests/Doctrine/Tests/ORM/Functional/EagerFetchCollectionTest.php
```diff
@@ -0,0 +1,164 @@
...
+        $this->_em->clear();
+
+        $query = $this->_em->createQuery('SELECT o FROM ' . EagerFetchOwner::class . ' o');
+        $query->setFetchMode(EagerFetchOwner::class, 'children', ORM\ClassMetadata::FETCH_EAGER);
+
+        $beforeQueryCount = count($this->getQueryLog()->queries);
+        $owners           = $query->getResult();
+        $afterQueryCount  = count($this->getQueryLog()->queries);
+
+        $this->assertEquals($beforeQueryCount + 2, $afterQueryCount);
```
greg0ire, line 68: Please use `assertQueryCount` and add a message argument with an explanation.

greg0ire (commented):
Please add docs :pray:

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -194,6 +194,12 @@ class ClassMetadataInfo implements ClassMetadata
      */
     public const FETCH_EXTRA_LAZY = 4;
 
+    /**
+     * Specifies that all entities of a collection valued association are fetched using a single or more
+     * additional WHERE IN queries.
+     */
+    public const FETCH_SUBSELECT = 5;
```
greg0ire: I'm very late to this party, but I think I like `FETCH_EAGER_BATCHED` better. There seems to be 2 families of fetch mode, eager and lazy, indicating which one we are dealing with is important. The `BATCHED` part indicates that we are slicing it, which sounds like an implementation detail though… so maybe `FETCH_EAGER_SUBSELECT`? The fact that entities are fetched afterwards seems more important than the fact that it is sliced in batches, and specifying that this is done with a `SUBSELECT` a.k.a.`WHERE IN` will lead to understanding that. Or we could be explicit and name it something like `FETCH_EAGER_AFTERWARDS`

beberlei: Forgot to remove this, as the feature is now just a change of the FETCH_EAGER behavior for 1:n assoc.

docs/en/reference/working-with-objects.rst
```diff
@@ -782,6 +782,23 @@ and these associations are mapped as EAGER, they will automatically
...
+
+Eager loading for many-to-one associations uses a second query to load
+the collections for several entities at the same time.
+
+When many-to-many, one-to-one or one-to-many associations are eagerly loaded,
+then the global batch size configuration is used to avoid IN(?) queries with
+too many arguments. The default batch size is 100 and can be changed with
+``Configuration::setEagerFetchBatchSize()``.
+
+For eagerly laoded Many-To-Many associations one query has to made for each
```
greg0ire:
```suggestion
For eagerly loaded Many-To-Many associations one query has to made for each
```

stetodd: Should it read "..has to **be** made for each"?

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @beberlei !

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -1047,7 +1047,9 @@ public function walkJoinAssociationDeclaration($joinAssociationDeclaration, $joi
             }
         }
 
-        $targetTableJoin = null;
+        if ($relation['fetch'] === ClassMetadata::FETCH_EAGER && $condExpr !== null) {
+            throw QueryException::eagerFetchJoinWithNotAllowed($assoc['sourceEntity'], $assoc['fieldName']);
```
maxhelias, line 1051: Hi, I probably have a poor implementation or use of eager mode on my relationships but this exception creates a BC break on my side.
Maybe it should be replaced by a deprecation pointing to this ticket : https://github.com/doctrine/orm/issues/10978 to explain why this is a bad pratice ?

aprat84: Indeed this is a BC in v2... How was it not been cached before merging? :(

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1264,7 +1264,7 @@ protected function getSelectColumnsSQL()
             }
 
             $isAssocToOneInverseSide = $assoc['type'] & ClassMetadata::TO_ONE && ! $assoc['isOwningSide'];
-            $isAssocFromOneEager     = $assoc['type'] !== ClassMetadata::MANY_TO_MANY && $assoc['fetch'] === ClassMetadata::FETCH_EAGER;
```
krzysztof-ciszewski-tg, line 1267: What is the reason for this change?

PR #8392, opened 2020-12-13
[GH-8231] Bugfix: Missed dirty check synchronization check.

### LINKED ISSUE #8231: Lost Update when using Deferred Explicit change tracking, Remove followed by Persist.
... That's hard to describe, I'll try nevertheless :-)

## Background

- We use Deferred Explicit as change tracking policy; and we have a single entity which is enough to reproduce the issue.
- The entity is persisted in the database already, and we have loaded it as `$entity`.

## How to reproduce

- call `$entityManager->remove($entity);`
- modify the entity `$entity->setFoo('bar');`
- call `$entityManager->persist($entity);`
- call `$entityManager->flush();`
- EXPECTED: The entity should be updated in the database.
- ACTUAL: the entity is not updated.

## Root Cause Analysis:

- call `$entityManager->remove($entity);`
  - Internally, this calls `UnitOfWork::doRemove()`, where the entity is marked with `UnitOfWork::STATE_REMOVED`.
- modify the entity `$entity->setFoo('bar');`
- call `$entityManager->persist($entity);`
  - Internally, this calls `UnitOfWork::doPersist()`.
  - There, the entity is marked as `UnitOfWork::STATE_MANAGED` again (it was marked as removed beforehand).
  - **NOTE:** Doctrine does NOT call `UnitOfWork::scheduleForDirtyCheck()`, which it would call if the entity was already
    in `STATE_MANAGED`. We rely on this `scheduleForDirtyCheck` to be called, because we use DEFERRED_EXPLICIT as change tracking policy.
- call `$entityManager->flush();`
  - the entity does not appear in the changed entities.
- EXPECTED: The entity should be updated in the database.
- ACTUAL: the entity is not updated.

## Suggested fix:

in `UnitOfWork::doPersist`
```
private function doPersist($entity, array &$visited) {
  // ... 
     switch ($entityState) {
            case self::STATE_MANAGED:
                // Nothing to do, except if policy is "deferred explicit"
                if ($class->isChangeTrackingDeferredExplicit()) {
                    $this->scheduleForDirtyCheck($entity);
                }
                break;

            case self::STATE_NEW:
                $this->persistNew($class, $entity);
                break;

            case self::STATE_REMOVED:
                // Entity becomes managed again
                unset($this->entityDeletions[$oid]);
                $this->addToIdentityMap($entity);

                $this->entityStates[$oid] = self::STATE_MANAGED;

// BEGIN CHANGE
                // Nothing to do, except if policy is "deferred explicit"
                if ($class->isChangeTrackingDeferredExplicit()) {
                    $this->scheduleForDirtyCheck($entity);
                }
// END CHANGE

                break;

            case self::STATE_DETACHED:
                // Can actually not happen right now since we assume STATE_NEW.
                throw ORMInvalidArgumentException::detachedEntityCannot($entity, "persisted");

            default:
                throw new UnexpectedValueException("Unexpected entity state: $entityState." . self::objToStr($entity));
        }
    // ... 
}

```

Note: this issue appeared as part of https://github.com/neos/neos-development-collection/pull/3015 

Thanks for your great work :) 
Sebastian

beberlei:
When an entity with change tracking policy "deferred explicit" gets removed, then persisted again, it is not schedulded for a dirty check synchronization. This is not the case for entities that are persisted and are already in the managed state.

Fixes #8231

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1796,6 +1796,10 @@ class UnitOfWork implements PropertyChangedListener
                 $this->addToIdentityMap($entity);
 
                 $this->entityStates[$oid] = self::STATE_MANAGED;
+
+                if ($class->isChangeTrackingDeferredExplicit()) {
+                    $this->scheduleForDirtyCheck($entity);
+                }
                 break;
 
             case self::STATE_DETACHED:
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7629Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7629Test.php
@@ -31,6 +31,20 @@ class GH7629Test extends OrmFunctionalTestCase
 
         self::assertFalse($this->_em->getUnitOfWork()->isScheduledForDirtyCheck($entity));
     }
+
+    /**
+     * @group GH-8231
+     */
+    public function testPersistAfterRemoveSchedulesForSynchronization(): void
+    {
+        $entity = $this->_em->find(GH7629Entity::class, 1);
+
+        $this->_em->remove($entity);
+
+        $this->_em->persist($entity);
+
+        self::assertTrue($this->_em->getUnitOfWork()->isScheduledForDirtyCheck($entity));
+    }
 }
 
 /**


skurfuerst:
@beberlei thank you so much, sorry for not following up on this ❤️

PR #8393, opened 2020-12-16
Added getHash method in SQLFilter to avoid caching the filtered queries

alecszaharia:
I introduced a new final method getHash on SQLFilter that generates a hash based on the called class name and all the parameters. Also the FilterCollection::getHash() was modified to use the getHash method from SQLFilter.

This is implemented acording to @beberlei's recomandation see the discussion bellow.

https://github.com/doctrine/orm/issues/3955#issuecomment-593161750

### DIFF
--- a/lib/Doctrine/ORM/Query/Filter/SQLFilter.php
+++ b/lib/Doctrine/ORM/Query/Filter/SQLFilter.php
@@ -103,6 +103,27 @@ abstract class SQLFilter
     }
 
     /**
+     * Return a unique hash base on filter class and his parametters
+     *
+     * @return string
+     */
+    final public function getHash()
+    {
+        $parameterCount = count($this->parameters);
+        $hash           = get_class($this).$parameterCount;
+        foreach ($this->parameters as $name => $param) {
+            if (is_object($param['value'])) {
+                $valueHash = spl_object_hash($param['value']);
+            } else {
+                $valueHash = $param['value'];
+            }
+            $hash .= $name.$param['type'].$valueHash;
+        }
+
+        return $hash;
+    }
+
+    /**
      * Returns as string representation of the SQLFilter parameters (the state).
      *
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Query/Filter/SQLFilter.php
```diff
@@ -102,6 +102,27 @@ abstract class SQLFilter
...
+     *
+     * @return string
+     */
+    final public function getHash()
+    {
+        $parameterCount = count($this->parameters);
+        $hash           = get_class($this).$parameterCount;
+        foreach ($this->parameters as $name => $param) {
+            if (is_object($param['value'])) {
+                $valueHash = spl_object_hash($param['value']);
```
beberlei, line 116: `spl_object_hash` is not reliable within and across requests. You need to serialize the parameter. the API of `setParameter` only allows stirngs anyways, it is not enforced though. But that should allow us to do this.

tests/Doctrine/Tests/ORM/Query/FilterCollectionTest.php
```diff
@@ -95,6 +95,17 @@ public function testGetFilter() : void
 
         self::assertInstanceOf(MyFilter::class, $filterCollection->getFilter(self::TEST_FILTER));
     }
+
+    public function testGetHash()
```
beberlei, line 99: Please also add an object based parameter and see serialize.

beberlei (changes requested):
Few change requests.

lib/Doctrine/ORM/Query/Filter/SQLFilter.php
alecszaharia: How about changing this method to
```php
final public function getHash()
{
        return self::class.$this;
}
```
I don't think the serialize() will have a big performance impact here.

PR #8374, opened 2020-12-05
Fix invalid psalm annotation

micheh:
The psalm annotation is currently invalid and psalm outputs an error:
`ERROR: InvalidDocblock - vendor/doctrine/orm/lib/Doctrine/ORM/EntityRepository.php:74:5 - Badly-formatted @param in docblock for Doctrine\ORM\EntityRepository::__construct (see https://psalm.dev/008)`

PR #8398, opened 2020-12-18
remove `T` from class metadata

azjezz:
while there's no releasing containing #8374, applying it locally causes the following issue, as  `ClassMetadata` has no generic parameter `T`.

```
ERROR: MixedArgumentTypeCoercion - src/Repository/UserRepository.php:48:39 - Argument 2 of Doctrine\ORM\EntityRepository::__construct expects Doctrine\ORM\Mapping\ClassMetadata<App\Entity\User>, parent type Doctrine\ORM\Mapping\ClassMetadata provided (see https://psalm.dev/194)
        parent::__construct($manager, $metadata);

```


beberlei (approved):

beberlei:
thanks

PR #8401, opened 2020-12-19
Identifier type is not set when many2many relations are deleted

yoshz:
I was trying out the new Symfony Ulid doctrine type together with Postgres and found a small bug.
For this use case the Ulid is formatted as UUID so it can be stored using the native GUID type.
In Doctrine this works fine until you want to delete an entity that has many2many relations causing error
`SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type uuid`

This PR makes sure that the identifier types are set on the delete queries so the Ulid is mapped correctly.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -515,7 +515,7 @@ class BasicEntityPersister implements EntityPersister
 
     /**
      * @param array<mixed> $identifier
-     * @param string[] $types
+     * @param string[]     $types
      *
      * @todo Add check for platform if it supports foreign keys/cascading.
      */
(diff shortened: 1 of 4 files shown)


beberlei:
This looks good, however it also is confusing this wasn't found before.

SenseException:
This PR also needs tests for this fix.

yoshz:
Added the unittest and tried to fix the failing checks. 
The coding standards check is conflicting with support for < PHP 7.4 so can't fix that.

SenseException:
You're referring to property types, right? You can extend phpcs.xml.dist for those checks that are only supported by PHP 7.4.

yoshz:
Phpcs is complaining about native return type hint, native property type hint and more that doesn't make sense. 
I see that on other PR's the phpcs check also fails, so I am going to ignore that.

Hopefully the unittest covers the cahnge enough so this bugfix can be merged.

beberlei:
@yoshz we merged a huge PR cleaning up code style into 2.8 yesterday, so there are now conflics. However after merging 2.8.x again you can now run `php vendor/bin/phpcbf` for autofixes safely.

beberlei (approved):

PR #1257, opened 2015-01-15
Resolve target entity also in discriminator map (allows interfaces and custom names in discriminator map)

Ocramius:
This PR supersedes #1130 ([DDC-3300](http://www.doctrine-project.org/jira/browse/DDC-3300)).

The final aim of the PR is to be able to define following in discriminator maps:

``` php
/** @DiscriminatorMap({"foo" = "FooInterface", "bar" = "BarConcreteImpl"}) */
```

PR #8402, opened 2020-12-21
Resolving target entity in discriminator map omits fields from subtables

mpdude:
#1257 introduced the possibility to use interfaces instead of concrete classes also in the DiscriminatorMap. Here's an example config:

```php
/** @DiscriminatorMap({"foo" = "FooInterface", "bar" = "BarConcreteImpl"}) */
```

The `ResolveTargetEntityListener` would be used to resolve such interface references to concrete classes at runtime.

The feature is currently broken. Core of the issue seems to be that `\Doctrine\ORM\Mapping\ClassMetadataInfo::$subClasses` is not populated with the concrete classes. This proably has to do with the resolution taking place too late, although I do not fully understand where this happens.

Example test failure before the change: https://github.com/doctrine/orm/pull/8402/checks?check_run_id=1587762816#step:7:73. The test case failed because the columns necessary for the subclasses are not created, but if they were (in an existing schema), the loaded entities would also lack data for their respective fields.

PR is against 2.7.x where the feature was introduced, since I did not find any documentation on PR policies/target branches.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -329,24 +329,26 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
         }
 
         // minor optimization: avoid loading related metadata when not needed
-        foreach ($metadata->discriminatorMap as $discriminatorValue => $discriminatorClass) {
-            if ($discriminatorClass === $metadata->name) {
-                $metadata->discriminatorValue = $discriminatorValue;
+//        foreach ($metadata->discriminatorMap as $discriminatorValue => $discriminatorClass) {
+//            if ($discriminatorClass === $metadata->name) {
+//                $metadata->discriminatorValue = $discriminatorValue;
+//
+//                return;
+//            }
+//        }
+//
+//        // iterate over discriminator mappings and resolve actual referenced classes according to existing metadata
+//        foreach ($metadata->discriminatorMap as $discriminatorValue => $discriminatorClass) {
+//            if ($metadata->name === $this->getMetadataFor($discriminatorClass)->getName()) {
+//                $metadata->discriminatorValue = $discriminatorValue;
+//
+//                return;
+//            }
+//        }
 
-                return;
-            }
+        if (! in_array($metadata->name, $metadata->discriminatorMap)) {
+            throw MappingException::mappedClassNotPartOfDiscriminatorMap($metadata->name, $metadata->rootEntityName);
         }
-
-        // iterate over discriminator mappings and resolve actual referenced classes according to existing metadata
-        foreach ($metadata->discriminatorMap as $discriminatorValue => $discriminatorClass) {
-            if ($metadata->name === $this->getMetadataFor($discriminatorClass)->getName()) {
-                $metadata->discriminatorValue = $discriminatorValue;
-
-                return;
-            }
-        }
-
-        throw MappingException::mappedClassNotPartOfDiscriminatorMap($metadata->name, $metadata->rootEntityName);
     }
 
     /**
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -41,7 +41,6 @@ use Doctrine\Persistence\Mapping\ReflectionService;
 use ReflectionClass;
 use ReflectionException;
 
-use function array_map;
 use function assert;
 use function class_exists;
 use function count;
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -284,7 +284,7 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
                     }
                 }
             } else {
-                assert($parent instanceof ClassMetadataInfo); // make static code analysis happy, see https://github.com/doctrine/orm/pull/8402#issuecomment-855805708
+                assert($parent instanceof ClassMetadataInfo); // https://github.com/doctrine/orm/issues/8746
                 if ((! $class->reflClass || ! $class->reflClass->isAbstract()) && ! in_array($class->name, $parent->discriminatorMap)) {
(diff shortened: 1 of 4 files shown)


mpdude:
Apparently, this might be fixed if the `ResolveTargetEntityListener` also deals with the DiscriminatorMap during the `loadClassMetadata` event.

@Ocramius:

You added [this code](https://github.com/doctrine/orm/blob/01187c9260cd085529ddd1273665217cae659640/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L312-L341) in #1257. My guess is that, apart from the check that all classes are part of the DiscriminatorMap, it is no longer necessary when the `ResolveTargetEntityListener` deals with interface-to-class substitution: Since the concrete classes are now available much earlier, the `discriminatorValue` will already be setup correctly (happens in `\Doctrine\ORM\Mapping\ClassMetadataInfo::addDiscriminatorMapClass()`). 

Maybe this method could even be removed altogether and the check added in `\Doctrine\ORM\Mapping\ClassMetadataFactory::validateRuntimeMetadata()`?

mpdude:
Note to self: #8378 might be working on the same thing/might be fixed as well by the last change?

beberlei (dismissed):
I think this is a valid bugfix, because the `Events::loadClassMetadata` is triggered inside `AbstractClassMetadataFactory::loadMetadata` and the logic from `resolveDiscriminatorValue` seems to be replaced correctly by the simple foreach in `ResolveTargetEntityListener`.

I am wondering though `resolveDiscriminatorValue` chanes `$metadata->discrimintorValue`, does this happen elsewhere now?

beberlei:
@mpdude needs to be rebased on 2.8.x as this is the bugfix target at the moment.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -74,18 +75,6 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
      */
     private $embeddablesActiveNesting = [];
 
-    /**
-     * {@inheritDoc}
-     */
-    protected function loadMetadata($name)
```
SenseException, line 80: The removal of this method can break code that extends `ClassMetadataFactory` and overrides `loadMetadata()`.

SenseException (changes requested):

mpdude:
Rebased to `2.8.x`. The failing CS check applies to files not changed in this PR, so I'll ignore it.

@beberlei regarding your comment

> I am wondering though resolveDiscriminatorValue chanes $metadata->discrimintorValue, does this happen elsewhere now?

I am not exactly sure what you're out for... The only write access to `\Doctrine\ORM\Mapping\ClassMetadataInfo::$discriminatorValue` I could find was in `\Doctrine\ORM\Mapping\ClassMetadataInfo::addDiscriminatorMapClass()`. 

My guess is that this will be called on every `ClassMetadataInfo` instance during metadata load, and that it will determine the discriminator value to be used for the underlying class (`\Doctrine\ORM\Mapping\ClassMetadataInfo::$name`). 

Since the resolution of interfaces to concrete classes now happens before that, we should already be looking a the classes (not intefaces anymore) in this place... Not sure, does this answer your question?

mpdude:
Rebased against `2.8.x` again, hoping that this time this will resolve CS issues 🤞🏻

mpdude:
Could anybody give me hints what I could do to get this along? 🙏🏻

mpdude:
@beberlei I've rebased this onto `2.9.x`.

Given that you already had a look at this and [basically agreed](https://github.com/doctrine/orm/pull/8402#pullrequestreview-584756258), could you please advise me what changes, clarifications or additional tests you'd like to see to make it a :+1:?

mpdude:
How can I make PHPStan/Psalm happy with regards to `\Doctrine\ORM\Mapping\ClassMetadataFactory::validateRuntimeMetadata()`?

The second argument `$parent` is hinted as `ClassMetadataInterface`, but I need to access `$parent->discriminatorMap`, which is not possible on an interface.

Changing it to `ClassMetadata` (as it is for the first `$class` argument) seems to be not easily possible because that in turn cascades to other, central places like `\Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::doLoadMetadata()`.

greg0ire:
An easy way to fix this should be to use `assert($parent instanceof ClassMetadata)` if you are sure that's the case, and to create a Github issue to fix the introduced technical debt by handling that cascade you describe.

mpdude:
@greg0ire Thank you for the hint! I guess it makes only sense to open the other PR once this PR here is merged, right?

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -295,6 +283,11 @@ protected function validateRuntimeMetadata($class, $parent)
                         throw MappingException::invalidClassInDiscriminatorMap($subClass, $class->name);
                     }
                 }
+            } else {
+                assert($parent instanceof ClassMetadataInfo); // make static code analysis happy, see https://github.com/doctrine/orm/pull/8402#issuecomment-855805708
```
greg0ire: Please create a Github issue instead, it will be easier to track :pray:

greg0ire:
> @greg0ire Thank you for the hint! I guess it makes only sense to open the other PR once this PR here is merged, right?

I think both issues and PR are independent, feel free to work on it now if you want to.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
mpdude: Not sure I got you right... You mean remove this line here? Then PHPStan will complain.

I thought you asked me to add this code here to make PHPStan happy, and to open another issue/PR to remove it again at the same time as fixing the typehints/param docblocks?

greg0ire: I mean keep this line here, but change the comment to point to an issue clearly describing what technical debt we have, not a thread buried in an unrelated pull request :wink:

ostrolucky (approved):

greg0ire:
Thanks @mpdude ! I took the liberty of squashing your commits, as most of them were amending the main one.

mpdude:
ORM 2.8 is end of life and this will not be backported, right?

greg0ire:
Correct 👍

PR #8403, opened 2020-12-22
Fixed embeddable loading

DemianShtepa:
This fixes a bug that arises when you need to map your entity without existed appropriate class (by using DisconnectedClassMetadataFactory), which contains embeddable fields. It can help laravel-doctrine package be able to generate Entities, which have embeddable values, based on mapping files. But now there is an ErrorException, because StaticReflectionService returned by DisconnectedClassMetadataFactory returns null on getClass method in inlineEmbeddable method, which needs emmbeddable's argument refClass to not be a null too.


SenseException:
@DemianShtepa Thank you. Please add tests that cover the new changes of your PR.

PR #8407, opened 2020-12-29
change branch-alias version from 2.7 to 2.8

Gemorroj:


composer.json
```diff
@@ -53,7 +53,7 @@
     "bin": ["bin/doctrine"],
     "extra": {
         "branch-alias": {
-            "dev-master": "2.7.x-dev"
+            "dev-master": "2.8.x-dev"
```
greg0ire, line 56: The alias of the master branch is not 2.8 but 3.0.

Gemorroj: this pr for 2.8.x branch. master have correct alias.

greg0ire: What's the point of changing the alias of the `master` branch in a PR to another branch?

https://getcomposer.org/doc/articles/aliases.md#branch-alias

Gemorroj: dont understan you.
We have https://github.com/doctrine/orm/tree/2.7.x branch with alias `2.7.x-dev` (https://github.com/doctrine/orm/blob/2.7.x/composer.json#L55)
For https://github.com/doctrine/orm/tree/2.8.x branch i guess alias should be `2.8.x-dev`
For https://github.com/doctrine/orm/tree/master `3.0.x-dev` now

greg0ire: > For https://github.com/doctrine/orm/tree/master 3.0.x-dev now

I think that's the only one that matters, as suggested in the docs I linked above.

> The alias must be a comparable dev version, and the branch-alias must be present on the branch that it references. For dev-master, you need to commit it on the master branch.

The alias could be removed from other branches for all I care.

greg0ire:
We don't use branch aliases anymore.

PR #8409, opened 2021-01-03
bump psalm, fix some issues on level 6 and improve type coverage

orklah:
This PR does the following:
- bump psalm to last version
- fix some issues on level 6
- improve type coverage by adding types when needed

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -213,6 +213,7 @@ class ClassMetadataInfo implements ClassMetadata
      * READ-ONLY: The name of the entity class.
      *
      * @var string
+     * @psalm-var class-string
      */
     public $name;
 
@@ -231,6 +232,7 @@ class ClassMetadataInfo implements ClassMetadata
      * as {@link $name}.
      *
      * @var string
+     * @psalm-var class-string
      */
     public $rootEntityName;
 
@@ -398,7 +400,15 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @var array
      *
-     * @psalm-var array<string, array{type: string, fieldName: string, columnName: string, inherited: class-string}>
+     * @psalm-var array<string, array{
+     *      type: string,
+     *      fieldName: string,
+     *      columnName?: string,
+     *      inherited?: class-string,
+     *      nullable?: bool,
+     *      originalClass?: class-string,
+     *      originalField?: string
+     * }>
      */
     public $fieldMappings = [];
 
@@ -463,6 +473,7 @@ class ClassMetadataInfo implements ClassMetadata
      * uniqueConstraints => array
      *
      * @var array
+     * @psalm-var array{name: string, schema: string, indexes: array, uniqueConstraints: array}
      */
     public $table;
 
@@ -576,6 +587,7 @@ class ClassMetadataInfo implements ClassMetadata
      * </code>
      *
      * @var array
+     * @psalm-var array{sequenceName: string, allocationSize: int, initialValue: int}
      *
      * @todo Merge with tableGeneratorDefinition into generic generatorDefinition
      */
@@ -1461,10 +1473,10 @@ class ClassMetadataInfo implements ClassMetadata
      * @throws MappingException If something is wrong with the mapping.
      *
      * @psalm-return array{
-     *                   mappedBy: mixed,
-     *                   inversedBy: mixed,
+     *                   mappedBy: mixed|null,
+     *                   inversedBy: mixed|null,
      *                   isOwningSide: bool,
-     *                   sourceEntity: string,
+     *                   sourceEntity: class-string,
      *                   targetEntity: string,
      *                   fieldName: mixed,
      *                   fetch: mixed,
@@ -1474,6 +1486,9 @@ class ClassMetadataInfo implements ClassMetadata
      *                   isCascadeRefresh: bool,
      *                   isCascadeMerge: bool,
      *                   isCascadeDetach: bool
+     *                   type: int
+     *                   originalField: string
+     *                   originalClass: class-string
      *               }
      */
     protected function _validateAndCompleteAssociationMapping(array $mapping)
@@ -2322,6 +2337,9 @@ class ClassMetadataInfo implements ClassMetadata
         return isset($this->associationMappings[$fieldName]['inherited']);
     }
 
+    /**
+     * @return bool
+     */
     public function isInheritedEmbeddedClass($fieldName)
     {
         return isset($this->embeddedClasses[$fieldName]['inherited']);
--- a/lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
+++ b/lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
@@ -181,7 +181,7 @@ class ResultSetMappingBuilder extends ResultSetMapping
         }
     }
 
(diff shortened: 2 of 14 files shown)


VincentLanglet:
Can you also fix this: https://github.com/weirdan/doctrine-psalm-plugin/issues/84#issuecomment-753953610 ?

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -398,7 +398,7 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @var array
      *
-     * @psalm-var array<string, array{type: string, fieldName: string, columnName: string, inherited: class-string}>
+     * @psalm-var array<string, array{type: string, fieldName: string, columnName?: string, inherited?: class-string, nullable?: bool, originalClass?: class-string, originalField?: string}>
```
greg0ire, line 401: I think linebreaks are ok inside psalm annotations.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -212,7 +212,7 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * READ-ONLY: The name of the entity class.
      *
-     * @var string
+     * @var class-string
```
greg0ire, line 215: Should we use `class-string<object>` wherever we use `class-string` but cannot be more precise?

orklah: We could but this is implied. Psalm will interpret it exactly the same AFAIK: https://github.com/vimeo/psalm/blob/ab5ddb1514f7de0976b964e92e9bf51138aa1496/src/Psalm/Type/Atomic/TClassString.php#L31

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
orklah: I'll do that

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
VincentLanglet: But you should use `@psalm-var` for class-string, because it's not a real php type.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -230,7 +230,7 @@ class ClassMetadataInfo implements ClassMetadata
      * hierarchy. If the entity is not part of a mapped inheritance hierarchy this is the same
      * as {@link $name}.
      *
-     * @var string
+     * @var class-string
```
VincentLanglet, line 233:
```suggestion
     * @var string
     * @psalm-var class-string
```

orklah: What's the rationale for this? PHPDocumentor and PHPStorm both understand this notation now. Is there another tool we want compatibility with that doesn't support this?

VincentLanglet: Someone with an older version of PHPStorm ? Someone with another IDE ?

orklah: @greg0ire WDYT? Should we keep class-string in @psalm annotations? What's the policy for doctrine repos?

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
greg0ire: @orklah I know, my question is : should we say it anyway just like one might use the type declaration `mixed` in the future (https://wiki.php.net/rfc/mixed_type_v2)

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
greg0ire: For Doctrine repositories we use `@psalm` annotations historically, because of PHPStorm, but we might want to relax that, I hear that indeed PHPStorm understands these now.

orklah: Thanks! Maybe that's something that could be enforced via CS rules, one way or another

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
orklah: Well, this is a matter of preference. I guess it's kind of the same rationale as `mixed[]` (or `array<mixed>`) being prefered to `array`. I'll add it for now

VincentLanglet: I never saw `class-string<object>`

orklah: All things considered, we already have quite a few `class-string` in codebase and no `class-string<object>`. For coherence, I'll follow the de facto convention.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
greg0ire: After discussing this, we are not yet ready to enforce it, but we do think we should no longer require contributors to add prefixes.

orklah: Good to know, thanks!

I added the psalm prefix for now but I'll keep that in mind for next time :)

greg0ire: Yup, also, as pointed out internally by @ostrolucky, there are some tricky cases like cases involving template that should stay with prefixes.

orklah: yes. Also phpstorm doesn't yet read *all* annotations. Array shapes for example will be reported as an error, even more so with list (for example `list<string, array{string, string}>` will be a `undefined class list`)

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
```diff
@@ -181,6 +181,9 @@ protected function addAllClassFields($class, $alias, $columnAliasMap = [])
         }
     }
 
+    /**
+     * @return bool
+     */
```
greg0ire, line 186: This method is `private`, you can add a type declaration instead.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -230,7 +231,7 @@ class ClassMetadataInfo implements ClassMetadata
...
```
VincentLanglet, line 234: Here you write `@var class-string` but here https://github.com/doctrine/orm/pull/8409/commits/ef2260c211169286fff87949cf42adbe060be93e#diff-732e324167dd49e48b221477c3e0f6d7934f3eb5ec0970dbf1c06f6c7df15398R215-R216
you wrote used `@psalm-var`. Consistency ?

orklah: done

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
orklah: done

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @orklah !

Issue #8410, opened 2021-01-05
Batch Processing using toIterable()


stlrnz:
Hi all!

I've updated my application to Doctrine ORM 2.8 some days ago.

Before that, my Repository code to iterate through a large result set looked like the following and worked perfectly even on >500.000 rows without leaking any memory.

```
    public function iterateAll(): Generator
    {
        $iterator = $this->createQueryBuilder('e')->getQuery()->iterate(null, Query::HYDRATE_SIMPLEOBJECT);

        foreach ($iterator as $equipment) {
            // index 0 is always the object
            yield $equipment[0];
        }
    }
```

Since `Query::iterate()` is deprecated now, I tried to use `Query::toIterable()` as suggested.

```
    public function iterateAll(): Generator
    {
        $iterator = $this->createQueryBuilder('e')->getQuery()->toIterable([], Query::HYDRATE_SIMPLEOBJECT);

        foreach ($iterator as $equipment) {
            yield $equipment;
        }
    }
```

With this implementation I stumbled over a massive memory leak in my application. I debugged this and found out that the `AbstractHydrator` is never releasing the Objects in `AbstractHydrator::toIterable()`. The called method `AbstractHydrator::hydrateRowData()` (or in my case `SimpleObjectHydrator::hydrateRowData`) just adds new entries to `$result`.
https://github.com/doctrine/orm/blob/378944dd27c1953440c1797fc02ff4620ca956ea/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php#L177
https://github.com/doctrine/orm/blob/378944dd27c1953440c1797fc02ff4620ca956ea/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php#L157

Is that the intended behaviour? Wouldn't it be better to clear `$result` in every loop cycle to free memory?

If that's intended, is there a better way to loop through large result sets?

simPod:
@beberlei this looks like in order to fix https://github.com/doctrine/orm/issues/3238 we need to keep hydrated objects in memory so ObjectHydrator can look them up. Any idea? Seems like we can't have both to me 🤔

PR #8412, opened 2021-01-06
fixed memory leak while iterating results

stlrnz:
I've tried to fix the issue described in https://github.com/doctrine/orm/issues/8410

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -163,8 +163,6 @@ abstract class AbstractHydrator
 
         $this->prepare();
 
-        $result = [];
-
         while (true) {
             $row = $this->_stmt->fetch(FetchMode::ASSOCIATIVE);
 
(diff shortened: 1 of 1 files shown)


beberlei:
@simPod can you say something?

stlrnz:
I've checked why the test https://github.com/doctrine/orm/pull/8412/checks?check_run_id=1654953910 failed. 

It seems that the `ObjectHydrator` in contrast to the `SimpleObjectHydrator` does need `$result` to be persistent.
So maybe this is not the correct solution for the linked issue?

simPod:
@stirnz same test run fails for `iterate()` as well, it was not tested for that case though while `toIterable()` is. Looking into it.

simPod:
I remember being [tasked](https://github.com/doctrine/orm/pull/7885#issuecomment-587250359) with fixing https://github.com/doctrine/orm/issues/3238 which introduced the memory leak as we're keeping result in memory now. Looking for workaround.

beberlei:
Related https://github.com/doctrine/orm/pull/8414

PR #8414, opened 2021-01-06
Fix selecting multiple entities #8413

### LINKED ISSUE #8413: Issue with toIterable() and selecting multiple entities
My code was previously working and since the last update, all the application is down due to toIterable() returning only the last entity when selecting multiple entities.

At line 179 in AbstractHydrator, Doctrine is doing yield end($result); causing the issue, the call to end() function should not be used.
$result is a simple array containing all retrieved entities, calling end() is just returning the last one.

https://github.com/doctrine/orm/blob/378944dd27c1953440c1797fc02ff4620ca956ea/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php#L179

I tested and the solution is really simple: Just remove the end() call, only yield $result;

This may be related to #8410

Cartman34:
Fix #8413
Now allowing to select multiple entities and return all of them

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -176,7 +176,7 @@ abstract class AbstractHydrator
 
             $this->hydrateRowData($row, $result);
 
-            yield end($result);
+            yield $result;
         }
     }
 
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -163,8 +163,6 @@ abstract class AbstractHydrator
 
         $this->prepare();
 
-        $result = [];
-
         while (true) {
             $row = $this->_stmt->fetch(FetchMode::ASSOCIATIVE);
 
@@ -173,6 +171,8 @@ abstract class AbstractHydrator
 
                 break;
             }
+
+            $result = [];
 
             $this->hydrateRowData($row, $result);
 


Cartman34:
This fix is not enough to really fix issue, it creates another issue.

Cartman34:
Tested selecting one or more entity in one row through several rows. It seems to work in all cases, it only affects iterations.

SenseException:
Thanks for your contribution. Your changes are currently breaking the existing tests. You also need to create a test for your use case.

Cartman34:
> Thanks for your contribution. Your changes are currently breaking the existing tests. You also need to create a test for your use case.

There is no new code, no new function, why should I develop a unit test ?
It's an important fix and I don't know how to do, so anyone do it and push it 😅
The a faulting code was developed recently and pushed without any test...

SenseException:
> There is no new code, no new function, why should I develop a unit test ?

If changes don't fix or introduce anything, like in a refactoring, then there's no need for new tests. There's still the possibility that an assumed harmless change introduced a bug in a functionality, that wasn't covered by tests yet. New tests will close such gaps in the coverage too.

If your code introduces a change, a feature or a bugfix, then a new test has to guarantee that this newly code change is covered. This doesn't have anything to do with new classes or methods, but with code paths. The test should prevent that someone else is removing your fix accidentally (even after months or years). It's in your and the other affected developers' interest that this can be prevented.

Cartman34:
I have no doubt this is very useful and I should provide it but it's out of my knowledges. I developed and shared the fix in a short time but this is a complex test, with probably a solution to mock up the database. I don't have the skills to do that and no time.
So somebody else have to do that.

beberlei:
Superseded by #8467

PR #8415, opened 2021-01-06
Fix association handling when there is a MappedSuperclass in the middle of an inheritance hierarchy

### LINKED ISSUE #5998: table inheritance 'JOINED' issue with selecting reference fields from base table

mpdude:
This fixes two closely related bugs.

1. When inheriting a to-one association from a mapped superclass, update the `sourceEntity` class name to the current class only when the association is actually _declared_ in the mapped superclass.
2. Reject association types that are not allowed on mapped superclasses only when they are actually _declared_ in a mapped superclass, not when inherited from parent classes.

#### Background 

Currently, when a many-to-one association is inherited from a `MappedSuperclass`, mapping information will be updated so that the association has the current (inheriting) class as the source entity.

https://github.com/doctrine/orm/blob/2138cc93834cfae9cd3f86c991fa051a3129b693/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L384-L393

This was added in 7dc8ef1db9b55ab1a76f416d78e6bd882bbbd1d5 for [DDC-671](https://github.com/doctrine/orm/issues/5181).

The reason for this is that a mapped superclass is not an entity itself and has no table. 

So, in the database, associations can only be from the inheriting entities' tables towards the referred-to target. This is also the reason for the limitation that only to-one associations may be added in mapped superclasses, since for those the database foreign key can be placed on the table(s) of the inheriting entities (and there may be more than one child class).

#### Issue with the current code

Neither the decision to update the `sourceEntity` nor the validation check should be based on `$parent->isMappedSuperclass`.

This only works in the simple case where the class hierarchy is `Mapped Superclass → Entity`.

The check is wrong when we have an inheritance hierarchy set up and the class hierarchy is `Base Entity → Mapped Superclass → Child Entity`.

Bug 1: The association should keep the root entity as the source. After all, in a JTI, the root table will contain the foreign key, and we need to base joins on that table when traversing `FROM LeafClass l JOIN l.target`.

Bug 2: Do not reject the to-many association declared in the base class. It is ok to have the reverse (owning) side point back to the base entity, as it would be if there were no mapped superclasses at all. The mapped superclass does not declare, add or otherwise interfere with the to-many association at all.

#### Suggested fix

Base the decision to change the `sourceEntity` on `$mapping['inherited']` being set. This field points to the topmost _parent entity_ class in the ancestry tree where the relationship mapping appears for the first time.

When it is not set, the current class is the first _entity_ class in the hierarchy with that association. Since we are inheriting the relation, it must have been added in a mapped superclass above, but was not yet present in the nearest parent entity class.

In that case, it may only be a to-one association and the source entity needs to be updated.

(See #10396 for a clarification of the semantics of `inherited`.)

#### Example 

Here is a simplified example of the class hierarchy. 

See the two tests added for more details – one is for checking the correct usage of a to-one association against/with the base class in JTI. The other is to test that a to-many association on the base class is not rejected.

I am sure that there are other tests that (still) cover the update of `sourceEntity` is happening.

```php
/**
 * @Entity
 */
class AssociationTarget
{
    /**
     * @Column(type="integer")
     * @Id
     * @GeneratedValue
     */
    public $id;
}

/**
 * @Entity
 * @InheritanceType("JOINED")
 * @DiscriminatorColumn(name="discriminator", type="string")
 * @DiscriminatorMap({"1" = "BaseClass", "2" = "LeafClass"})
 */
class BaseClass
{
    /**
     * @Column(type="integer")
     * @Id
     * @GeneratedValue
     */
    public $id;

    /**
     * @ManyToOne(targetEntity="AssociationTarget")
     */
    public $target;
}

/**
 * @MappedSuperclass
 */
class MediumSuperclass extends BaseClass
{
}

/**
 * @Entity
 */
class LeafClass extends MediumSuperclass
{
}
```

When querying `FROM LeafClass l`, it should be possible to `JOIN l.target`. This currently leads to an SQL error because the SQL join will be made via `LeafClass.target_id` instead of `BaseClass.target_id`. `LeafClass` is considered the `sourceEntity` for the association – which is wrong–, and so the foreign key field is expected to be in the `LeafClass` table (using JTI here).

Fixes #5998, fixes #7825. 

#### Updated: 

I have removed the abstract entity class, since it is not relevant for the issue and took the discussion off course. Also, the discriminator map now contains all classes.

#### Updated 2:

Added the second variant of the bug, namely that a to-many association would wrongly be rejected in the same situation.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5998Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5998Test.php
@@ -1,0 +1,255 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\DBAL\LockMode;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group GH-5998
+ */
+class GH5998Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(GH5998JTI::class),
+            $this->_em->getClassMetadata(GH5998JTIChild::class),
+            $this->_em->getClassMetadata(GH5998STI::class),
+            $this->_em->getClassMetadata(GH5998Basic::class),
+            $this->_em->getClassMetadata(GH5998Related::class),
+        ]);
+    }
+
+    /**
+     * Verifies that MappedSuperclasses work within an inheritance hierarchy.
+     */
+    public function testIssue(): void
+    {
+        // Test JTI
+        $this->classTests(GH5998JTIChild::class);
+        // Test STI
+        $this->classTests(GH5998STIChild::class);
+        // Test Basic
+        $this->classTests(GH5998Basic::class);
+    }
+
+    private function classTests($className): void
+    {
+        // Test insert
+        $child      = new $className('Sam', 0, 1);
+        $child->rel = new GH5998Related();
+        $this->_em->persist($child);
+        $this->_em->persist($child->rel);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        // Test find by rel
+        $child = $this->_em->getRepository($className)->findOneBy(['rel' => $child->rel]);
+        self::assertNotNull($child);
+        $this->_em->clear();
+
+        // Test query by id with fetch join
+        $child = $this->_em->createQuery('SELECT t, r FROM ' . $className . ' t JOIN t.rel r WHERE t.id = 1')->getOneOrNullResult();
+        self::assertNotNull($child);
+
+        // Test lock and update
+        $this->_em->transactional(static function ($em) use ($child): void {
+            $em->lock($child, LockMode::NONE);
+            $child->firstName = 'Bob';
+            $child->status    = 0;
+        });
+        $this->_em->clear();
+        $child = $this->_em->getRepository($className)->find(1);
+        self::assertEquals($child->firstName, 'Bob');
+        self::assertEquals($child->status, 0);
+
+        // Test delete
+        $this->_em->remove($child);
+        $this->_em->flush();
+        $child = $this->_em->getRepository($className)->find(1);
+        self::assertNull($child);
+    }
+}
+
+/**
+ * @ORM\MappedSuperclass
+ */
+class GH5998Common
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue
+     *
+     * @var int
+     */
+    public $id;
+    /**
+     * @ORM\ManyToOne(targetEntity=GH5998Related::class)
+     * @ORM\JoinColumn(name="related_id", referencedColumnName="id")
+     *
+     * @var GH5998Related
+     */
+    public $rel;
+    /**
+     * @ORM\Version
+     * @ORM\Column(type="integer")
+     *
+     * @var int
+     */
+    public $version;
+
+    /** @var mixed */
+    public $other;
+}
+
+/**
+ * @ORM\Entity
+ * @ORM\InheritanceType("JOINED")
+ * @ORM\DiscriminatorMap({"child" = GH5998JTIChild::class})
+ */
+abstract class GH5998JTI extends GH5998Common
+{
+    /**
+     * @ORM\Column(type="string", length=255)
+     *
+     * @var string
+     */
+    public $firstName;
+}
+
+/**
+ * @ORM\MappedSuperclass
+ */
+class GH5998JTICommon extends GH5998JTI
+{
+    /**
+     * @ORM\Column(type="integer")
+     *
+     * @var int
+     */
+    public $status;
+}
+
+/**
+ * @ORM\Entity
+ */
+class GH5998JTIChild extends GH5998JTICommon
+{
+    /**
+     * @ORM\Column(type="integer")
+     *
+     * @var int
+     */
+    public $type;
+
+    public function __construct(string $firstName, int $type, int $status)
+    {
+        $this->firstName = $firstName;
+        $this->type      = $type;
+        $this->status    = $status;
+    }
+}
+
+/**
+ * @ORM\Entity
+ * @ORM\InheritanceType("SINGLE_TABLE")
+ * @ORM\DiscriminatorMap({"child" = GH5998STIChild::class})
+ */
+abstract class GH5998STI extends GH5998Common
+{
+    /**
+     * @ORM\Column(type="string", length=255)
+     *
+     * @var string
+     */
+    public $firstName;
+}
+
+/**
+ * @ORM\MappedSuperclass
+ */
+class GH5998STICommon extends GH5998STI
+{
+    /**
+     * @ORM\Column(type="integer")
+     *
+     * @var int
+     */
+    public $status;
+}
+
+/**
+ * @ORM\Entity
+ */
+class GH5998STIChild extends GH5998STICommon
+{
+    /**
+     * @ORM\Column(type="integer")
+     *
+     * @var int
+     */
+    public $type;
+
+    public function __construct(string $firstName, int $type, int $status)
+    {
+        $this->firstName = $firstName;
+        $this->type      = $type;
+        $this->status    = $status;
+    }
+}
+
+/**
+ * @ORM\Entity
+ */
+class GH5998Basic extends GH5998Common
+{
+    /**
+     * @ORM\Column(type="string", length=255)
+     *
+     * @var string
+     */
+    public $firstName;
+
+    /**
+     * @ORM\Column(type="integer")
+     *
+     * @var int
+     */
+    public $status;
+
+    /**
+     * @ORM\Column(type="integer")
+     *
+     * @var int
+     */
+    public $type;
+
+    public function __construct(string $firstName, int $type, int $status)
+    {
+        $this->firstName = $firstName;
+        $this->type      = $type;
+        $this->status    = $status;
+    }
+}
+
+/**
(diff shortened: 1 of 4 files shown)


mpdude:
It seems #5998 describes the same bug. #7825 might include the attempt to fix it, although it's a bigger (abandoned?) PR and seems to address additional things.

greg0ire:
Please add something to sum https://github.com/doctrine/orm/pull/7825#issuecomment-554214736 up in the docs, that kind of knowledge of what is supported and what isn't could come in handy to both users and maintainers.

greg0ire:
I don't think 2.7 is still maintained, but https://github.com/doctrine/orm/blob/2.8.x/.doctrine-project.json says it is. Maybe @beberlei or @SenseException will know?

mpdude:
@greg0ire Regarding additional documentation on `MappedSuperclass`: I have often missed that myself :+1:. Maybe someone who knows better than me could give me some starting pointers...? 

Is a `@MappedSuperclass` declaration *always* necessary when a class contains fields or associations that need to be persisted, but when that class is not an `@Entity`?

I sometimes had the impression that things also work without the declaration?

Also, what about `@Entity` classes that are `abstract`?

greg0ire:
> Maybe someone who knows better than me could give me some starting pointers...?

Sure, but it won't be me I'm afraid :sweat_smile:

SenseException:
For ORM it should the current version that is supported. In that case it's 2.8. I'm not sure if there are exceptions for 2.7, like security issues. I think this should target `2.8.x`. I can see confusion what the repository "says" and what the documentation shows, because the docs still label it as upcoming. This will be handled in a different PR.

@mpdude `@Entity` shouldn't be used in an abstract because Entities can't be abstract itself, while a MappedSuperclass can. There's probably a way to make things work in a way, but this might result in problems on ORM updates in the future.

From the docs: https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/inheritance-mapping.html
> Typically, the purpose of such a mapped superclass is to define state and mapping information that is common to multiple entity classes.

If there are entities with similar mapping, the MappedSuperclass can contain these common fields. There's not much extended info about other possible use cases and I'm not sure if there are some that are considered as a intentional feature. The sentence

> Mapped superclasses ... can appear in the middle of an otherwise mapped inheritance hierarchy (through Single Table Inheritance or Class Table Inheritance).

seems to open a lot of possibilities that might not be covered by ORM updates.

mpdude:
@SenseException I'm afraid I might not fully understand your comment.

Regarding the abstract base `@Entity`, that might be a leftover from the real life case I came from; I'll check that and think the problem persists if I remove it.

I will also rebase this PR to 2.8. Will it be backported to 2.7, or what is the exact state of that (security fixes only?).

But more in general, are you suggesting that it is not supported, discouraged or otherwise unsafe to use `@MappedSuperclass` in hierarchies below `@Entity` classes?

mpdude:
Whether or not the base class from my example is `abstract` does not make a difference for the issue. 

Just as a side note, when the class is _not_ `abstract`, for obvious reasons I get an error message 

> [base class] has to be part of the discriminator map of [base class] to be properly mapped in the inheritance hierarchy. Alternatively you can make [base class] an abstract class to avoid this exception from occurring

This seems to confirm that an `abstract` `@Entity` was not completely unexpected when that part of the ORM was designed. Also, at https://www.doctrine-project.org/projects/doctrine-orm/en/2.13/reference/architecture.html#entities it says

> Entities support inheritance, polymorphic associations, and polymorphic queries. Both abstract and concrete classes can be entities. Entities may extend non-entity classes as well as entity classes, and non-entity classes may extend entity classes.

But as I said, it does not matter here.

Regarding `MappedSuperclass` in the middle of a class hierarchy, apart from the documentation quoted above there has also been some discussion in #7825. If this approach were completely beyond what ORM was designed for, my guess is it would have been rejected over there already, which apparently wasn't the case.

andrews05:
Would be great to see this working!
I'm surprised at what a minor change it was to make it work and pass all those tests, considering everything it took to achieve this on the v3 branch.

mpdude:
@andrews05 Great to have you on board here! You must have spent a lot of time working on #7825, so I think you have some valuable expertise to add. From your point of view, are there any edge cases we should cover as well?

andrews05:
@mpdude Hm, I would say if it's passing all the tests I added, then that should cover all the edge cases I'm aware of.

mpdude:
Could anybody give me hints what I could do to get this along? 🙏🏻

vv12131415:
@mpdude since 2.8 is unmaintained, first thing to do, is to change the branch to `2.9`

then you can tag someone who can do review for this part of code


upd: you can also see the history https://github.com/doctrine/orm/commits/2.9.x/lib/Doctrine/ORM/Mapping
and tag those who have most recent commits there

mpdude:
@beberlei you last worked on this area of the code in 2010 when you did 7dc8ef1db9b55ab1a76f416d78e6bd882bbbd1d5.  @guilhermeblanco you were involved in the discussion of #7825, which should be fixed by this PR.

Would it be possible for either of you to review this PR? 💚

BenMorel:
Hi @mpdude, thank you for this PR! I have a closely related issue in #8771.

Could you please check if you get the same error as I do, when running `vendor/bin/doctrine orm:validate-schema` or `bin/console doctrine:schema:validate` (Symfony)? I expect that you get something like:

> [FAIL] The entity-class (...)\MediumSuperclass mapping is invalid:
> * Entity class '(...)\MediumSuperclass' is part of inheritance hierarchy, but is not mapped in the root entity '(...)\BaseClass' discriminator map. All subclasses must be listed in the discriminator map.

If so, this would make me doubt whether having a `@MappedSuperclass` in the middle of a hierarchy is supported at all.

mpdude:
If I get recent comments in #8771 right, this is a bug in the validator command and the MappedSuperclass in the middle of the inheritance hierarchy should be supported?

beberlei:
This is very difficult and time consuming to get into the state of mind to find out what the right behavior is and I haven't had time for it yet, sorry it will take a while longer until I can review this.

mpdude:
@beberlei sure, I fully understand. If there is anything I can do to make it easier, or if it would help to go through this in a real life use case on a screen shared in Slack, DM me and we‘ll find a way.

mpdude:
Rebased onto `2.11.x`. The fix itself is a three-line change, the rest are tests added ;-).

greg0ire:
@mpdude I removed the usage of the simple annotation loader a while ago. This means the annotations in your tests need to come with corresponding use statements.

mpdude:
Thank you @greg0ire for the heads-up.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH8415Test.php
```diff
@@ -0,0 +1,115 @@
...
+ * @ORM\DiscriminatorColumn(name="discriminator", type="string")
+ * @ORM\DiscriminatorMap({"1" = "Doctrine\Tests\ORM\Functional\Ticket\GH8415LeafClass"})
+ */
+abstract class GH8415BaseClass
+{
+    /**
+     * @ORM\Column(type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue
+     * @ORM\var int
```
greg0ire: Woopsie :P

tests/Doctrine/Tests/ORM/Functional/Ticket/GH5998Test.php
```diff
@@ -0,0 +1,241 @@
...
+    }
+}
+
+/**
+ * @ORM\MappedSuperclass
+ */
+class GH5998Common
+{
+    /**
+     * @ORM\Id
```
greg0ire, line 86: You can use `vendor/bin/phpcbf` to fix cs issues

mpdude:
I'd still like to get this resolved and merged, and I am willing to spend the time to port and this to the current 2.x branch.

@beberlei Do you see a chance to review this and what would you need to make this as easy as possible?

mpdude:
I have an idea how the check can be written even more clearly based on #10397 and #10398

mpdude:
@SenseException maybe you could give this another review? I have removed the parts that you questioned the last time.

mpdude:
@derrabus Would it be possible for you to review this? 

If it is too complex, has missing context or similar, let me know and I'll try to make it as easy as possible for you.

beberlei (dismissed):

greg0ire (approved):

greg0ire:
Thanks @mpdude !

mpdude:
Thank you all!

PR #8419, opened 2021-01-09
Introduce `EntityManagerInterface#wrapInTransaction()`

simPod:
This introduces `Doctrine\ORM\EntityManagerInterface#wrapInTransaction()` so we can return any value from Transaction.

### DIFF
--- a/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
+++ b/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
@@ -8,6 +8,13 @@ use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\Persistence\ObjectManagerDecorator;
 
+use function get_class;
+use function method_exists;
+use function sprintf;
+use function trigger_error;
+
+use const E_USER_NOTICE;
+
 /**
  * Base class for EntityManager decorators
  */
@@ -51,6 +58,23 @@ abstract class EntityManagerDecorator extends ObjectManagerDecorator implements 
     public function transactional($func)
     {
         return $this->wrapped->transactional($func);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function wrapInTransaction(callable $func)
+    {
+        if (! method_exists($this->wrapped, 'wrapInTransaction')) {
+            trigger_error(
+                sprintf('Calling `transactional()` instead of `wrapInTransaction()` which is not implemented on %s', get_class($this->wrapped)),
+                E_USER_NOTICE
+            );
+
+            return $this->wrapped->transactional($func);
+        }
+
+        return $this->wrapped->wrapInTransaction($func);
     }
 
     /**
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -251,6 +251,28 @@ use function sprintf;
     /**
      * {@inheritDoc}
      */
+    public function wrapInTransaction(callable $func)
+    {
+        $this->conn->beginTransaction();
+
+        try {
+            $return = $func($this);
+
+            $this->flush();
+            $this->conn->commit();
+
+            return $return;
+        } catch (Throwable $e) {
+            $this->close();
+            $this->conn->rollBack();
+
+            throw $e;
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     public function commit()
     {
         $this->conn->commit();
--- a/lib/Doctrine/ORM/EntityManagerInterface.php
+++ b/lib/Doctrine/ORM/EntityManagerInterface.php
@@ -80,11 +80,29 @@ interface EntityManagerInterface extends ObjectManager
      * If an exception occurs during execution of the function or flushing or transaction commit,
      * the transaction is rolled back, the EntityManager closed and the exception re-thrown.
      *
+     * @deprecated 2.10 Use {@link wrapInTransaction} instead.
+     *
      * @param callable $func The function to execute transactionally.
      *
      * @return mixed The non-empty value returned from the closure or true instead.
      */
     public function transactional($func);
+
+    /**
+     * Executes a function in a transaction.
+     *
+     * The function gets passed this EntityManager instance as an (optional) parameter.
+     *
+     * {@link flush} is invoked prior to transaction commit.
+     *
+     * If an exception occurs during execution of the function or flushing or transaction commit,
+     * the transaction is rolled back, the EntityManager closed and the exception re-thrown.
+     *
+     * @param callable $func The function to execute transactionally.
+     *
+     * @return mixed The value returned from the closure.
+     */
+    // public function wrapInTransaction(callable $func);
 
     /**
      * Commits a transaction on the underlying database connection.
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,6 +4,24 @@
 
 If you want to extend it now you have to provide your own validation schema.
 
+## New method `Doctrine\ORM\EntityManagerInterface#wrapInTransaction($func)`
+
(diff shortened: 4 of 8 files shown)


greg0ire:
> I believe I've seen this bug fixed on v3 like half a year ago. Now I've found this PR stale so I just wanted to update the test and I've found the fix is not there anymore. I can't tell what happened here, branches are shuffled etc.

The v3 you know has been renamed to old-prototype-3.x, and we are working on backporting some of the PRs here: https://github.com/doctrine/orm/projects/4

The difference this time is we are going to regularly merge up 2.x into 3.x .

We may backport more, feel free to find the commits in old-prototype-3.x and backport them 👍

greg0ire:
Is there a BC-break in this PR? If not, please target 2.10.x instead.

greg0ire:
Or even 2.9.x since it is a bugfix

simPod:
Thanks for explaining, I'll revert this and target the`old-prototype`. This was initially only about tests for the bug.

simPod:
@greg0ire I suppose CI is broken on old-prototype, right?

greg0ire:
Yes, it's no longer maintained, it's just a source of commits to port to maintained branches. Why are you targeting it? Does the bug only happen there?

simPod:
@greg0ire the bug is fixed only there 

https://github.com/doctrine/orm/commit/aa9f11e11d31590c6c346074f33ea7bf01468fd1#diff-09261ffd6eb75f4fda9de6c2fbf156761380cf13687e57a199878630890402c4R232

the previous version (e.g. the whole 2.x orm) returns `true` from `transactional()` when inner closure returns anything falsy (`[]`, `0`, `false`) instead of the falsy value.

greg0ire:
Could you please work on porting the new behavior to 3.x instead? Otherwise it's just more porting work for us in the future

simPod:
ah ok, I have just reverted that few h ago. Np. Gimme sec.

simPod:
It's done, test failures are irrelevant https://github.com/simPod/orm/runs/3093712747.

greg0ire:
I think they relate to https://github.com/doctrine/cache/pull/381

greg0ire:
Blocked by https://github.com/doctrine/orm/pull/8847 and 2 merges up

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -237,7 +237,9 @@ public function transactional($func)
             $this->flush();
             $this->conn->commit();
 
-            return $return ?: true;
+            // Can't use ?? because PHP 7.1 is supported
+            //phpcs:ignore SlevomatCodingStandard.ControlStructures.RequireNullCoalesceOperator
+            return $return === null ? true : $return;
```
greg0ire: This description looks wrong now:

https://github.com/doctrine/orm/blob/551018136bc0c4f53f047a33f7fdac128589c6b7/lib/Doctrine/ORM/EntityManagerInterface.php#L83

Also, BC-breaks should be documented in UPGRADE.md

greg0ire:
Regarding the PR itself: why did non-empty value deserve a special treatment, and why does `null` deserve one?

simPod:
@greg0ire I misinterpreted `non-empty`. Will fix.

lib/Doctrine/ORM/EntityManager.php
simPod: done

greg0ire:
Please fix the unit tests. The failures are now relevant.

beberlei:
This is a problematic BC break because we cannot offer a forward compatible way to migrate or a way to detect the deprecated behavior and warn about it.

I believe this requires a new method name.

simPod:
Hmm, would name should we choose? Do you have some Doctrine guidelines on method names? `transact()` / `transaction()` / `?`?

greg0ire:
How about `wrapInTransaction()`?

simPod:
Ok, initially I just wanted to add few more tests cases to ensure it returns values properly. This is gonna be slightly bigger tho 😄

greg0ire:
Also, since this should be backwards compatible, you should target 2.10.x

simPod:
So, should I introduce a new method, that returns all value types and
target 2.10.x? Am I right?

greg0ire:
Yes! It's better because:
- we have an upgrade path
- we can try this out earlier, and know if it's a good idea without having to wait for 3.0.0

simPod:
#8872

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -248,6 +248,28 @@ public function transactional($func)
         }
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function wrapInTransaction(callable $func)
```
greg0ire, line 254: Shouldn't `transactional()` be deprecated in favor of this method?

simPod: Aight, tried to deprecate it :)

lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
```diff
@@ -53,6 +53,14 @@ public function transactional($func)
         return $this->wrapped->transactional($func);
     }
```
greg0ire, line 61: This method probably needs to be deprecated too

simPod: It implements the interface so should be covered by that 🤔

greg0ire:
Please fix the conflict :pray: 

Note: this PR kind of backports https://github.com/doctrine/orm/pull/6147

simPod:
done

> Note: this PR kind of backports #6147

I believe I've seen it multiple times here so initially I only wanted to add test cases for it then I§ve found out it's not implemented 🤷🏾

greg0ire:
Hence the branch name!

greg0ire (dismissed):

SenseException:
The extended interface is a BC break. @greg0ire Is this considered as a "smaller" one?

lib/Doctrine/ORM/EntityManagerInterface.php
```diff
@@ -80,12 +80,30 @@ public function beginTransaction();
...
+     * {@link flush} is invoked prior to transaction commit.
+     *
+     * If an exception occurs during execution of the function or flushing or transaction commit,
+     * the transaction is rolled back, the EntityManager closed and the exception re-thrown.
+     *
+     * @param callable $func The function to execute transactionally.
+     *
+     * @return mixed The value returned from the closure.
+     */
+    public function wrapInTransaction(callable $func);
```
greg0ire: @SenseException maybe let's just do this:
```suggestion
    // public function wrapInTransaction(callable $func);
```

callers will have to check that this method exists before calling it though.

SenseException: This will do it for 2.10.

simPod: or I can target 3.0?

greg0ire: @simPod please don't retarget this whole PR to 3.0, that would mean we cannot remove `transactional` in 3.0

simPod: I'm waiting for instructions 🙃

greg0ire: Ok, here are the steps:

1. apply my suggestion
2. squash it with the previous commit
3. improve the upgrade notes by explaining how people are supposed to check that `wrapInTransaction` exists before calling it, if they have an instance of `EntityManagerInterface`

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @simPod !

lib/Doctrine/ORM/EntityManagerInterface.php
zoltan-schwarz: I'm really curious. What is the reason to comment this out? Why can't it be the interface?

PR #8420, opened 2021-01-09
Set 2.8 as current supported version

SenseException:


greg0ire:
There is a 2.9.x branch, should something be done about it?

SenseException:
Good point. It also should be added. :+1:

greg0ire (approved):

greg0ire:
Thanks @SenseException !

PR #8421, opened 2021-01-11
fix incorrect type annotation

dbu:
according to the constructor parameter type declaration, this is the interface and not necessarily BasicEntityPersister.


dbu:
i found that BasicEntityPersister is also explicitly referenced in SqlExpressionVisitor, even though the only thing the visitor uses is  getSelectConditionStatementSQL which is part of the EntityPersister interface. why is it not requesting the interface? 

i think i could change that as well - even if somebody extends the visitor, they would only narrow the constructor argument to a specific implementation and not break anything. the property is private, so afaict there is no BC break involved. still, might be more correct to only do that in the 2.9.x branch as its not a bugfix.

beberlei:
@dbu thanks

PR #8426, opened 2021-01-14
Fix --complete flag in orm:ensure-production-settings command

alexander-schranz:
$input->getOption('complete') returns `false` and not `null` if not set.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/EnsureProductionSettingsCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/EnsureProductionSettingsCommand.php
@@ -62,7 +62,7 @@ class EnsureProductionSettingsCommand extends Command
         try {
             $em->getConfiguration()->ensureProductionSettings();
 
-            if ($input->getOption('complete') !== null) {
+            if ($input->getOption('complete') !== false) {
                 $em->getConnection()->connect();
             }
         } catch (Throwable $e) {
--- a/lib/Doctrine/ORM/Tools/Console/Command/EnsureProductionSettingsCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/EnsureProductionSettingsCommand.php
@@ -62,7 +62,7 @@ class EnsureProductionSettingsCommand extends Command
         try {
             $em->getConfiguration()->ensureProductionSettings();
 
-            if ($input->getOption('complete') !== false) {
+            if (!$input->getOption('complete')) {
                 $em->getConnection()->connect();
             }
         } catch (Throwable $e) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Tools/Console/Command/EnsureProductionSettingsCommand.php
```diff
@@ -62,7 +62,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         try {
             $em->getConfiguration()->ensureProductionSettings();
 
-            if ($input->getOption('complete') !== null) {
+            if (!$input->getOption('complete')) {
```
alexander-schranz, line 65: current symfony returns `false` not sure if `null` was returned in older version so I did switch to using `!` instead for checking for false or null.

SenseException: There is a default value that can be set in `addOption`. Without an explicit value it is `null`.

https://github.com/symfony/console/blob/5.x/Command/Command.php#L400

It seems to be the same in 4.0 and 3.0.

alexander-schranz: Seems like its not possible to set a default value for InputOption::VALUE_NONE: https://github.com/symfony/console/blob/356c70659d6b48595f9f275cd7c4de0d5544c49c/Input/InputOption.php#L168

alexander-schranz: So I could go with:

```
$input->getOption('complete') === true
```

```
$input->getOption('complete') !== false
```

or the current version.

The value of the input is set here to `true`:
https://github.com/symfony/console/blob/356c70659d6b48595f9f275cd7c4de0d5544c49c/Input/ArgvInput.php#L248

And default value to `false`:

https://github.com/symfony/console/blob/356c70659d6b48595f9f275cd7c4de0d5544c49c/Input/InputOption.php#L182

SenseException: If it's a boolean then your current syntax should fit. You currently let it step into the if-block when the value is `false`. It also looks like there are no tests yet that covers this class. Can you please create one that covers your changes?

alexander-schranz: @SenseException created the missing test class hope it matches how tests are created in this repository.

SenseException:
@doctrine/team-orm Do we need `OrmFunctionalTestCase` in the test class?

ostrolucky:
I don't think so.

alexander-schranz:
Change to the DoctrineTestCase

SenseException (approved):

alexander-schranz:
Anything I can do to get this merged?

SenseException:
Currently not. This will be handled by another reviewer.

ostrolucky (approved):

alexander-schranz:
@SenseException @ostrolucky thx!

PR #8433, opened 2021-01-20
Adding note on `preFlush`

ThomasLandauer:
Taken from https://github.com/doctrine/orm/pull/8343#issuecomment-731038953 - to clarify the discussion there.

Question: Will this change be merged automatically into v3? https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/events.html#preflush

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -555,7 +555,8 @@ preFlush
 ~~~~~~~~
 
 ``preFlush`` is called at ``EntityManager#flush()`` before
-anything else. ``EntityManager#flush()`` should not be called inside
+anything else – even before computing the changeset.
+``EntityManager#flush()`` should not be called inside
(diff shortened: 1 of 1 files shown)


ThomasLandauer:
I just noticed that doing changes on associated entities inside preFlush results in a mess (but no exception!): Some data is written to the wrong fields, even to the wrong records.
=> I think this should be explained too! But I don't know: What are the *exact* things that should not be done?

ThomasLandauer:
One more question: On https://symfony.com/doc/current/doctrine/events.html#doctrine-entity-listeners it says:
> Entity listeners are defined as PHP classes that listen to a single Doctrine event **on a single entity class.**

Is this really true? Cause if `preFlush` occurs before Doctrine even knows what to do, the event is triggered for all classes - not just for this single one? So using `preFlush` in an **entity** listener just doesn't make sense?

PR #8434, opened 2021-01-21
fixed entity generation for numeric values

julianullrich99:
Originally doctrine adds integer and float default-values as 

`private $property = '12';`

With this fix it correctly inserts

`private $property = 12;`


SenseException:
Thanks for your contribution. Does this fix a specific bug?

jukisu23:
Yes, it fixes a bug that I discovered at work where when you declare a default value for a db field as an integer 0 it generates something like

```
/**
* @type INTEGER
*/
private $field = '0';
```

So it creates a String. With my fix it properly generates an integer for it

```
/**
* @type INTEGER
*/
private $field = 0;
```

So without this bug it hurts type safety.

SenseException:
In that case please create a test that covers your changes.

beberlei:
We can take our chances on this one without new tests, integer default values are already tested and should break if this code was wrong.

beberlei (approved):

PR #8435, opened 2021-01-21
Minor rewording

ThomasLandauer:
Emphasizing the (counter-intuitive) fact that preUpdate is called inside **flush** - cause this was causing me some confusion, see https://github.com/symfony/symfony/issues/39894

In general: This entire page is confusing, for two reasons:
* The list of events appears *twice* (as "overview" list on top, then as separate headings below), with neither being complete.
* The description of each event is inconsistent.

To improve this, I'm suggesting an **overview table** on top, with the events being the rows (linked to their respective heading below), and the following columns:
* called inside which of *my* function calls (i.e. `em->persist` or `$em->flush`)
* Is allowed to change the *current* entity?
* Is allowed to change *associated* entities?
* Is allowed to create *new* entities?
* Is calling `persist` or `flush` inside it allowed/required?
* Can access ID?
* ...?

I can start the table, but I'd need some help, since I don't know all the details myself.

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -659,9 +659,8 @@ postFlush
 preUpdate
 ~~~~~~~~~
 
-PreUpdate is the most restrictive to use event, since it is called
-right before an update statement is called for an entity inside the
-``EntityManager#flush()`` method. Note that this event is not
+PreUpdate is called inside the ``EntityManager#flush()`` method,
+right before an SQL ``UPDATE`` statement. This event is not
 triggered when the computed changeset is empty.
 
 Changes to associations of the updated entity are never allowed in


SenseException:
> To improve this, I'm suggesting an overview table on top, with the events being the rows (linked to their respective heading below), and the following columns:

Your suggestion sounds interesting and the redundant events can be confusing. Can you post an example ob such a row?
For which details do you need help?

ThomasLandauer:
Here's just some quick sketch:
Event | Called by | Can change current entity | Generated ID available
--| -- | -- | --
[preUpdate]() | `$em->flush()` | ? 
[postUpdate]() | `$em->flush()` | No
[postRemove]() | `$em->flush()` | No
[postPersist]() |  `$em->flush()` | No
[preRemove]() | `$em->remove()` | Yes, but useless | Yes
[prePersist]() | `$em->persist()` on *initial* persist | ? | No

Main advantage: Whenever I setup a listener, I have some requirements in mind (e.g.: I need the ID). So an overview table makes searching for events in question much faster and easier.

Since the table will get wide, maybe you could change the overall layout of https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/events.html and increase `.container { max-width: 1140px; }`

I wrote this with `entity` listeners in mind (therefore "Can change *current* entity"). So my first question would be: What's the difference to "general" listeners? IMHO, entity listeners should be mentioned first and foremost, cause doing stuff on *any* flush doesn't make much sense to me - or does it?

> For which details do you need help?

Probably for many ;-) It would be some work and it would take some time, so I wanted to find out first if you're interested at all.

SenseException:
A additional small matrix for the events can give a helpful overview if that table doesn't become too wide with too many columns. It would lose its lucidity. I would also expect the yes/no values you had in your example.

> What's the difference to "general" listeners?

Entity Listeners are bound to just one entity and will only be triggered for this entity alone. If you e.g. want to have a `prePersist` only for `Foo` entity, but not for the others, the Entity Listeners can help here. The "general" `prePersist` listener is for every entity.

> The list of events appears twice

I just realized that the events at the top are representing the Lifecycle Events while the bottom ones are Event Listeners. Maybe that table can combine those different event types.

ThomasLandauer:
Here's my plan: I'll take the list at https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/events.html#lifecycle-events as a starting point and set up a table with these rows, starting with *only one column*: "Dispatched by". So we can see if this works fine, then continue adding more columns.

If, at some point, this all-in-one table gets too wide, here's my plan B:
Under each heading of https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/events.html#implementing-event-listeners, start each event's description with an *identical* overview table:
|  |  |
| -- | -- |
| Dispatched by | `->foo()` |
| Can change current entity | Yes |
| Generated ID available | No |
| ... | ... |

Questions:
1. Should I reuse this PR or start a new one? If new: On the 2.8 branch?
2. Which table syntax? There are four at https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#table-directives

SenseException:
Yes, you can use this PR for your table. Your minor rewording was targeting an improvement of understanding anyway.

Used table syntax:

```
+--------+--------+--------+
| head 1 | head 2 | head 3 | 
+=================+========+
| row 1  | row 1  | row 1  |
+--------+--------+--------+
| row 2  | row 2  | row 2  |
+--------+--------+--------+
```

ThomasLandauer:
Please merge this; since some time has passed, I've started a separate PR for the overview table.

derrabus (approved):

PR #8438, opened 2021-01-25
DDC-2076 Remove useless join over target table of ManyToMany relationship

plfort:
Quite an old issue : https://github.com/doctrine/orm/issues/2758

I noticed this additional "join" while trying to optimize a query. As mentioned in the issue, I don't see a case when this join is useful.
In my case (query optimization), this PR provides a significant reduction of the query cost given by Postgresql

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -2001,25 +2001,9 @@ class SqlWalker implements TreeWalker
 
             // SQL table aliases
             $joinTableAlias   = $this->getSQLTableAlias($joinTable['name']);
-            $targetTableAlias = $this->getSQLTableAlias($targetClass->getTableName());
             $sourceTableAlias = $this->getSQLTableAlias($class->getTableName(), $dqlAlias);
 
-            // join to target table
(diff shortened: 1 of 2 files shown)


SenseException (dismissed):
Looks good to me. Maybe `SqlWalker` need some integration tests, but I'm not sure if this should be done in the scope of this PR.

beberlei:
@SenseException The `SelectSqlGenerationTest` is the integration test for `SqlWalker`, so this is fine. 

I think the patch is good, thanks.

unrelated however I realized this must provide wrong results on composite keys or not? It seems weird to iterate the columns and add multiple in conditions, maybe I am wrong.

beberlei (dismissed):

beberlei (dismissed):

beberlei:
@plfort i am a bit hesitant to merge a cosmetic PR into a branch in bugfix mode, the old code works, so this is an improvement and changing the SQL is a bit more risky, so I would like you to rebase this on 2.9 and we can merge it there.

plfort:
I agree with you about composite keys, it requires further tests. I thought about it but for this change I wanted to keep the behaviour.

plfort:
Rebase on 2.9 done BTW

beberlei (approved):

rvanlaak:
if this PR did fix #2758, that issue can get closed now

Issue #7939, opened 2019-12-01
Typed Properties and default column/association types


beberlei:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

When declaring a type on a property, this should act as a default for field type or ManyToOne/OneToOne targetEntity.

```php
class User
{
    /** @ORM\Column */
    public int $id; // defaults to type="integer"

   /** @ORM\ManyToOne */
   public Email $email; // defaults to targetEntity=Email::class
}
```

The relevant code would be in `ClassMetadataInfo::mapField`, `mapManyToOne` and `mapOneToOne`. Whenever `type` or `targetEntity` is missing, it could look this up in reflection property instead if it has a type.

renanbr:
Could we expect it'll also detect `nullable` properties?

```php
/** @ORM\Column */
public ?string $name;
```

Lustmored:
I have submitted #8439 as an attempt to implement features discussed in this issue. It probably still needs work, but it would be great to get some feedback :)

beberlei:
Fixed in #8439

PR #8439, opened 2021-01-26
Use typed properties for default metadata for #7939

Lustmored:
This is simple implementation of idea outlined in #7939. During completion step of a field mapping completion or association mapping completion it looks for typed property and, when existent, uses it to determine default for `type`, `targetEntity` and `nullable` (in fields and `joinTable`).

I have added simple tests covering all cases I found.

I am not a great fan of encapsulating code in `PHP_VERSION_ID >= 70400`, but at the time of writing got no better idea.

I couldn't get master branch to work locally and that's the only reason why I target '2.9.x' (Composer has problem resolving doctrine/dbal).

This is my first PR to Doctrine project, so if anything is done wrong please let me know and I will do my best to fix it :)

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -89,7 +89,7 @@ as part of the lifecycle of the instance variables entity-class.
 Required attributes:
 
 -  **type**: Name of the Doctrine Type which is converted between PHP
-   and Database representation.
+   and Database representation. Default to ``string`` or :ref:`Type from PHP property type <reference-php-mapping-types>`
 
 Optional attributes:
 
@@ -113,7 +113,7 @@ Optional attributes:
 -  **unique**: Boolean value to determine if the value of the column
    should be unique across all rows of the underlying entities table.
 
--  **nullable**: Determines if NULL values allowed for this column. If not specified, default value is false.
+-  **nullable**: Determines if NULL values allowed for this column. If not specified, default value is false. When using typed properties on entity class defaults to true when property is nullable.
 
 -  **options**: Array of additional options:
 
@@ -635,6 +635,8 @@ Optional attributes:
    constraint level. Defaults to false.
 -  **nullable**: Determine whether the related entity is required, or if
    null is an allowed state for the relation. Defaults to true.
+   When using typed properties on entity class defaults to false when
+   property is not nullable.
 -  **onDelete**: Cascade Action (Database-level)
 -  **columnDefinition**: DDL SQL snippet that starts after the column
    name and specifies the complete (non-portable!) column definition.
@@ -715,6 +717,7 @@ Required attributes:
 
 -  **targetEntity**: FQCN of the referenced target entity. Can be the
    unqualified class name if both classes are in the same namespace.
+   When typed properties are used it is inherited from PHP type.
    *IMPORTANT:* No leading backslash!
 
 Optional attributes:
@@ -923,6 +926,7 @@ Required attributes:
 
 -  **targetEntity**: FQCN of the referenced target entity. Can be the
    unqualified class name if both classes are in the same namespace.
+   When typed properties are used it is inherited from PHP type.
    *IMPORTANT:* No leading backslash!
 
 Optional attributes:
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -717,7 +717,7 @@ Required attributes:
 
 -  **targetEntity**: FQCN of the referenced target entity. Can be the
    unqualified class name if both classes are in the same namespace.
-   When typed properties are used it is inherited from PHP type.
+   You can omit this value if you use a PHP property type instead.
    *IMPORTANT:* No leading backslash!
 
 Optional attributes:
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -26,6 +26,7 @@ use DateTime;
 use DateTimeImmutable;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Types\Types;
 use Doctrine\Instantiator\Instantiator;
 use Doctrine\Instantiator\InstantiatorInterface;
 use Doctrine\ORM\Cache\CacheException;
@@ -1444,28 +1445,28 @@ class ClassMetadataInfo implements ClassMetadata
             ) {
                 switch ($type->getName()) {
                     case DateInterval::class:
-                        $mapping['type'] = 'dateinterval';
+                        $mapping['type'] = Types::DATEINTERVAL;
                         break;
                     case DateTime::class:
-                        $mapping['type'] = 'datetime';
+                        $mapping['type'] = Types::DATETIME_MUTABLE;
                         break;
                     case DateTimeImmutable::class:
-                        $mapping['type'] = 'datetime_immutable';
+                        $mapping['type'] = Types::DATETIME_IMMUTABLE;
                         break;
                     case 'array':
-                        $mapping['type'] = 'json';
+                        $mapping['type'] = Types::JSON;
                         break;
                     case 'bool':
-                        $mapping['type'] = 'boolean';
+                        $mapping['type'] = Types::BOOLEAN;
                         break;
                     case 'float':
-                        $mapping['type'] = 'float';
+                        $mapping['type'] = Types::FLOAT;
                         break;
                     case 'int':
-                        $mapping['type'] = 'integer';
+                        $mapping['type'] = Types::INTEGER;
                         break;
                     case 'string':
-                        $mapping['type'] = 'string';
+                        $mapping['type'] = Types::STRING;
                         break;
                 }
             }
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1429,9 +1429,9 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param  array  $mapping The field mapping to validate & complete.
      *
-     * @return void
+     * @return mixed[] The updated mapping.
      */
-    private function _validateAndCompleteTypedFieldMapping(array &$mapping)
+    private function _validateAndCompleteTypedFieldMapping(array $mapping)
     {
         $type     = $this->reflClass->getProperty($mapping['fieldName'])->getType();
 
@@ -1472,6 +1472,8 @@ class ClassMetadataInfo implements ClassMetadata
                 }
             }
         }
+
+        return $mapping;
     }
 
     /**
@@ -1479,9 +1481,9 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param array $mapping The mapping.
      *
-     * @return void
+     * @return mixed[] The updated mapping.
      */
-    private function _validateAndCompleteTypedAssociationMapping(array &$mapping)
+    private function _validateAndCompleteTypedAssociationMapping(array $mapping)
     {
         $type = $this->reflClass->getProperty($mapping['fieldName'])->getType();
 
@@ -1500,6 +1502,8 @@ class ClassMetadataInfo implements ClassMetadata
                 }
             }
         }
+
+        return $mapping;
     }
 
     /**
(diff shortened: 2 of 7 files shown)


Lustmored:
I don't know how to proceed with tests under PHP < 7.4 as there are no typed properties and entity class required for new test cases fails there. I probably could move it to other namespace, but that seems like complicating file structure without much benefit. Please help me with any suggestion :)

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1414,6 +1420,53 @@ protected function _validateAndCompleteFieldMapping(array &$mapping)
             throw MappingException::missingFieldName($this->name);
         }
 
+        if (
+            PHP_VERSION_ID >= 70400
+            && isset($this->reflClass)
```
beberlei: Just to understand, do you have a case where `$this->reflClass` is not set? I assume when `DisconnectedMetadataFactory` is used while converting an XML/YML mapping or generating entities?

beberlei: The condition is duplicated below as well, can you factor it out into a private method?

```php
private function isTypedProperty($name) : bool
{
}
```

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1414,6 +1420,53 @@ protected function _validateAndCompleteFieldMapping(array &$mapping)
...
+                        case DateInterval::class:
+                            $mapping['type'] = 'dateinterval';
+                            break;
+                        case DateTime::class:
+                            $mapping['type'] = 'datetime';
+                            break;
+                        case DateTimeImmutable::class:
+                            $mapping['type'] = 'datetime_immutable';
+                            break;
+                        case 'array':
```
beberlei: this is ambiguous, could also be `simple_array` or `object` type. Not sure what to do, since otherwise it will default to `"string"` later.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1414,6 +1420,53 @@ protected function _validateAndCompleteFieldMapping(array &$mapping)
             throw MappingException::missingFieldName($this->name);
         }
 
+        if (
+            PHP_VERSION_ID >= 70400
+            && isset($this->reflClass)
+            && $this->reflClass->hasProperty($mapping['fieldName'])
+        ) {
+            $property = $this->reflClass->getProperty($mapping['fieldName']);
```
beberlei: the body of the if statement should also be factored out into a private method `$mapping = $this->_validateAndCompleteTypedFieldMapping($mapping);`

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1511,6 +1564,31 @@ protected function _validateAndCompleteAssociationMapping(array $mapping)
         // the sourceEntity.
         $mapping['sourceEntity'] = $this->name;
 
+        if (
```
beberlei: Reuse the factored out condition method described above.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1511,6 +1564,31 @@ protected function _validateAndCompleteAssociationMapping(array $mapping)
...
+            PHP_VERSION_ID >= 70400
+            && isset($this->reflClass)
+            && $this->reflClass->hasProperty($mapping['fieldName'])
+        ) {
+            $property = $this->reflClass->getProperty($mapping['fieldName']);
```
beberlei: factor this whole block out as `$mapping = $this->_validateAndCompleteTypedAssociationMapping($mapping);`.

beberlei (changes requested):
Thank you for your PR, this is a very good first contribution, congrats :-)

I made some comments and have some questions inline.

Lustmored:
Thanks for the review. I will work on it next week 👍

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Lustmored: When I remove `isset($this->reflClass)` there are 56 test errors and 3 failures with `Error: Call to a member function hasProperty() on null` message. I have no idea if those tests should be fixed or it is expected behavior and don't know Doctrine internals enough yet to determine if it can occur in real usage.

I can work through those tests and enforce setting `$this->reflClass` if that's desired, but someone more experience would need to take a look for real cases where unset `reflClass` is desired.

Lustmored: I have also factored condition as requested

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Lustmored: My reasoning was to provide useful defaults instead of solving every possible use case. In this example having `"string"` as property type would simply fail at runtime as ORM would try to assign `string` value to `array` typed property. Therefore I believe useful default is better than nothing.

But the same problem comes with `\DateType` and many other fields. I believe the best long-term solution would be to create new interface, for example `DefaultTypeResolverInterface` that could look like this:
```
interface DefaultTypeResolverInterface
{
    public function determineTypeForProperty(\ReflectionType $type): ?string;
}
```
And provide at least one implementation (maybe working as here) and then allow users to choose implementation based on their config.

This way we could provide useful defaults for everyone and allow users to have simple method of providing own rules (for example I would benefit from overwriting `\DateTimeImmutable` in my projects).

I don't know how to approach such solution in ORM structure, but would love to do so. If this seems reasonable please just give me some directions and I will follow through.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Lustmored: Done

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Lustmored: Done

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Lustmored: Done

Lustmored:
I have rebased onto 2.9.x (still some conflicts pop out, don't know why) and handled suggestions :+1:

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
beberlei: It makes sense that it fails for when the `DisconnectedMetadataFactory` is used in case of EntityGenerator and exporter and such things. Of course this creates an inconsistency where the types default to "string" when reflClass does not exist. But there is no way to avoid that at the moment, so lets keep it this way.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
beberlei: Yes, lets stay with the hardcoded resolving rules for now and we'll leave the replacable resolving rules for a future PR.

You should document what PHP types get mapped to which Doctrine types in the documentation, see `docs/` folder.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1398,6 +1404,99 @@ public function getSqlResultSetMappings()
...
+    }
+
+    /**
+     * Validates & completes the given field mapping based on typed property.
+     *
+     * @param  array  $mapping The field mapping to validate & complete.
+     *
+     * @return void
+     */
+    private function _validateAndCompleteTypedFieldMapping(array &$mapping)
```
beberlei: don't pass by reference please but return the array again. This is how other `_validate` methods in this class do it already, so that stay consistent.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1398,6 +1404,99 @@ public function getSqlResultSetMappings()
...
+    }
+
+    /**
+     * Validates & completes the basic mapping information based on typed property.
+     *
+     * @param array $mapping The mapping.
+     *
+     * @return void
+     */
+    private function _validateAndCompleteTypedAssociationMapping(array &$mapping)
```
beberlei: same here, return the array again and don't pass by reference

beberlei (changes requested):

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Lustmored: OK, I have changed _validateAndCompleteFieldMapping too while at it - it was still using reference.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Lustmored: OK, done

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Lustmored: OK, I have added documentation where it seemed logical to me

Lustmored:
I think I addressed all current comments to some degree. I see that checks are somehow stuck and I see no way to "unstuck" them to see if my code passes. Is it possible somehow to run them?

beberlei:
@Lustmored this is because of the conflict in `ClassMetadataInfo.php`, you need to fix that before the checks run.

docs/en/reference/annotations-reference.rst
```diff
@@ -715,6 +717,7 @@ Required attributes:
 
 -  **targetEntity**: FQCN of the referenced target entity. Can be the
    unqualified class name if both classes are in the same namespace.
+   When typed properties are used it is inherited from PHP type.
```
beberlei: I would formulate this a bit differently: `You can omit this value if you use a PHP property type instead.`

tests/Doctrine/Tests/ORM/Mapping/ClassMetadataTest.php
```diff
@@ -92,6 +95,70 @@ public function testFieldIsNullable()
...
+        $cm->mapField(['fieldName' => 'status', 'length' => 50]);
+        $this->assertTrue($cm->isNullable('status'));
+
+        // Explicit Not Nullable
+        $cm->mapField(['fieldName' => 'username', 'length' => 50]);
+        $this->assertFalse($cm->isNullable('username'));
+
+        // Join table Nullable
+        $cm->mapOneToOne(['fieldName' => 'email', 'joinColumns' => [[]]]);
+        $this->assertFalse($cm->getAssociationMapping('email')['joinColumns'][0]['nullable']);
```
beberlei, line 133: Missing assertion on `target_entity`.

beberlei (changes requested):
Just two small changes, then I believe this is ready to go.

docs/en/reference/annotations-reference.rst
Lustmored: OK, changed

tests/Doctrine/Tests/ORM/Mapping/ClassMetadataTest.php
Lustmored: Right, missed it. Assertion added

Lustmored:
I think I have complied with all your comments. I have also rebased onto 2.9.x and pushed CS fixed.

Only remaining failing check is CS complaing about method names starting with underscore - I don't have the confidence to change them without asking first.

Lustmored:
I have rebased onto 2.9 and made a simple change to use `Types::` constants instead of strings for types. Is there anything else I can do to make this happen? :)

beberlei (approved):

beberlei:
@Lustmored thank you very much!

beberlei:
@Lustmored FYI, This doesn't work anymore with AttributeDriver (does it with Annotations?) Because of the default value of `Column::$type = "string"`. The typed mapping completion only works when `$mapping['type']` is not set. I am going to investigate more.

Lustmored:
@beberlei I see the problem and I see how I missed it in tests (I have created only tests for `ClassMetadataInfo` and not metadata obtained from mapping drivers). I am also looking into it. Probably simplest solution would be to change the rule from `isset` to `! isset($foo) || $foo === 'string'`, but that would lead to obvious problem of impossibility to overwrite type inherited from typed property with a simple `string` type.

beberlei:
@Lustmored i think setting Column::$type to nullable & default null is better

Lustmored:
@beberlei That's exactly what I'm experimenting with right now. Probably will come with a PR fixing the issue and introducing more tests to ensure it works with mapping later today :+1:

Lustmored:
@beberlei let's continue in #8589. Hopefully all cases are covered there this time :)

PR #8447, opened 2021-01-29
Removing "Naming convention"

ThomasLandauer:
...cause the example (which newbies will copy-paste anyway :-) shows it all. Explaining something like that so high up on the page and so detailed, gives the topic a perceived importance which it doesn't really have.

Besides, I suggest to move the entire header about custom events ("The Event System") at the very bottom of that page. Reason: Before anybody will go and hack anything themselves, they will certainly first take a look at what's included out of the box ;-)

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -121,25 +121,6 @@ Now you can test the ``$eventSubscriber`` instance to see if the
         echo 'pre foo invoked!';
     }
 
-Naming convention
-~~~~~~~~~~~~~~~~~
-
-Events being used with the Doctrine ORM EventManager are best named
-with camelcase and the value of the corresponding constant should
-be the name of the constant itself, even with spelling. This has
-several reasons:
-
-
--  It is easy to read.
--  Simplicity.
--  Each method within an EventSubscriber is named after the
-   corresponding constant's value. If the constant's name and value differ
-   it contradicts the intention of using the constant and makes your code
-   harder to maintain.
-
-An example for a correct notation can be found in the example
-``TestEvent`` above.
-
 .. _reference-events-lifecycle-events:
 
 Lifecycle Events


SenseException:
I don't see why the naming convention should be removed. Even if the code examples are showing implicit conventions, It is the explanation that shows context about the benefits.

ThomasLandauer:
For two reasons:
* It's patronizing. People usually know how to name their own stuff themselves.
* Hints like "It is easy to read" are ridiculous ;-) Camelcase is best-practice in PHP for years, and when somebody doesn't want to follow it - well it's up to them...

IMHO, people coming here are looking for details about the events. Not general hints for naming variables :-)

SenseException:
Naming conventions aren't unusual in frameworks and libraries and you'll find them in other projects too. It's okay if others don't want to follow them and that's why it is a naming convention instead of a naming law. A documentation is targeting many different types of people. Some are happy if they can only copy and paste code examples (and understand them more or less). Others like to read about details and reasons behind decisions in libraries. Some like to follow a naming convention. Others don't give a :duck: about documentation at all and prefer to read only code.

We surely can't cover every detail, but a documentation is trying to help different type of understanding. You might see the naming convention as unimportant. Others prefer to have one as a suggestion.

ThomasLandauer:
I'm certainly not against a naming convention per se! I just think that this lengthy explanation is too much, cause the code sample shows it all.

Anyway - it's not so important to continue discussing it ;-)

More important would be my other idea - what do you say to that?

> Besides, I suggest to move the entire header about custom events ("The Event System") at the very bottom of that page. Reason: Before anybody will go and hack anything themselves, they will certainly first take a look at what's included out of the box ;-)

SenseException:
It makes sense to explain the event system itself as introduction before the events itself are introduced. This structure of the documentation was created as an introduction instead of a reference work. I agree that someone who's familiar with the ORM event system would like to skip to the events itself, but the goal of this docs page is the introduction of the events.

I think your table suggestion in #8435 has more value to the event docs for new readers and for advanced users.

PR #8448, opened 2021-01-30
Relax contract of EntityListenerResolver so it doesn't require class name

ostrolucky:
There is actually no technical reason why interface should require FQCN

Rel: https://github.com/doctrine/DoctrineBundle/issues/1224

### DIFF
--- a/lib/Doctrine/ORM/Mapping/EntityListenerResolver.php
+++ b/lib/Doctrine/ORM/Mapping/EntityListenerResolver.php
@@ -29,24 +29,20 @@ namespace Doctrine\ORM\Mapping;
 interface EntityListenerResolver
 {
     /**
-     * Clear all instances from the set, or a specific class when given.
+     * Clear all instances from the set, or a specific instance when given its identifier.
      *
-     * @param string $className The fully-qualified class name
+     * @param string $className May be arbitrary string. Name kept for BC only.
      *
      * @return void
-     *
-     * @psalm-param class-string $className
      */
     function clear($className = null);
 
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Mapping/EntityListenerResolver.php
```diff
@@ -29,24 +29,20 @@
 interface EntityListenerResolver
 {
     /**
-     * Clear all instances from the set, or a specific class when given.
+     * Clear all instances from the set, or a specific instance when given its identifier.
      *
-     * @param string $className The fully-qualified class name
+     * @param string $className May be arbitrary string. Name kept for BC only.
```
greg0ire, line 34:
```suggestion
     * @param string $className May be any arbitrary string. Name kept for BC only.
```

lib/Doctrine/ORM/Mapping/EntityListenerResolver.php
```diff
@@ -29,24 +29,20 @@
...
-     * @psalm-param class-string $className
      */
     function clear($className = null);
 
     /**
-     * Returns a entity listener instance for the given class name.
+     * Returns a entity listener instance for the given identifier.
      *
-     * @param string $className The fully-qualified class name
+     * @param string $className May be arbitrary string. Name kept for BC only.
```
greg0ire, line 43:
```suggestion
     * @param string $className May be any arbitrary string. Name kept for BC only.
```

greg0ire (approved):

PR #8453, opened 2021-02-05
Fixes parenthesis when \n is involved

Warxcell:
```
JSON_CONTAINS(p.state, :positive_value, :ordered_state) = :positive_value 
OR 
JSON_CONTAINS(p.state, :positive_value, :completed_state) = :positive_value)
```

when we have OR like that (without space before OR) - it doesn't add parenthesis.

### DIFF
--- a/lib/Doctrine/ORM/Query/Expr/Composite.php
+++ b/lib/Doctrine/ORM/Query/Expr/Composite.php
@@ -22,7 +22,7 @@ namespace Doctrine\ORM\Query\Expr;
 
 use function implode;
 use function is_object;
-use function stripos;
+use function preg_match;
 
 /**
  * Expression class for building DQL and parts.
@@ -63,7 +63,7 @@ class Composite extends Base
         }
 
         // Fixes DDC-1237: User may have added a where item containing nested expression (with "OR" or "AND")
-        if (stripos($queryPart, ' OR ') !== false || stripos($queryPart, ' AND ') !== false) {
+        if (preg_match('/\s(OR|AND)\s/', $queryPart)) {
             return $this->preSeparator . $queryPart . $this->postSeparator;
         }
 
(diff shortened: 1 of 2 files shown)


greg0ire:
I guess there is the same issue with tabulations, right?

Warxcell:
> I guess there is the same issue with tabulations, right?

I don't know. Who uses tabs on queries anyways?

lib/Doctrine/ORM/Query/Expr/Composite.php
```diff
@@ -63,7 +66,8 @@ private function processQueryPart($part)
         }
 
         // Fixes DDC-1237: User may have added a where item containing nested expression (with "OR" or "AND")
-        if (stripos($queryPart, ' OR ') !== false || stripos($queryPart, ' AND ') !== false) {
+        $singleLineQueryPart = str_replace(PHP_EOL, ' ', $queryPart);
+        if (stripos($singleLineQueryPart, ' OR ') !== false || stripos($singleLineQueryPart, ' AND ') !== false) {
```
greg0ire, line 70: Maybe we would be better off with one single `preg_replace` here, with `\s(OR|AND)\s`. That way tabs are handled, and line feeds are preserved, that's what I'm getting at.

Warxcell: Yeah, but maybe `str_replace` is faster. I will try `preg_replace`.

greg0ire (approved):
> Yeah, but maybe str_replace is faster. If we will use RegEx, I can just use preg_match and not replace anything.

Probably, but here I don't think we are in the hot path: we are in dql parsing which:
- is cached
- does not happen a lot (a few queries vs other code that applies on every row of the result set)

Warxcell:
> * is cached
> * does not happen a lot (a few queries vs other code that applies on every row of the result set)

I think it's not the actual flow.

This is QueryBuilder, which builds DQL which is then parsed and after that cached.
So next time we have again QueryBuilder, which builds DQL, then it's returned by cache.

greg0ire:
Hmmm yes I think you're right, it's probably executed every time indeed 👍
Anyway, not significant at all IMO.

SenseException (approved):
The pattern is not a expensive one. I think this should be fine.

greg0ire:
Thanks @Warxcell ! I squashed your commits because the first one and the last one really should, and although you made an argument for squashing in another PR, in this case it's unlikely to apply.

PR #8466, opened 2021-02-07
Introduce doctrine/deprecations

beberlei:
Using alpha version of doctrine/deprecations (v0.2.0) replace all usages of `trigger_error` and replace tests using `VerifyDeprecations`. During implementation realized a few changes to API needed leading to these PRs that were merged into v0.2: 

- https://github.com/doctrine/deprecations/pull/7
- https://github.com/doctrine/deprecations/pull/8

The tests were cleaned up a lot with assertions about deprecations, it is not helpful to have the same deprecation asserted on in hundrets of different, unrelated tests.

Uncoditional deprecations are not tested anymore, because their code is trival and does not affect runtime behavior. Only conditional deprecations are still tested.

This PR also undeprecates `EntityManager::detach` for now, because unless we find some kind of replacement API it cannot be removed as relied on heavily by the batch processing use-cases with `AbstractQuery::toIterable`.


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -957,9 +955,12 @@ public function getHints()
      */
     public function iterate($parameters = null, $hydrationMode = null)
     {
-        @trigger_error(
-            'Method ' . __METHOD__ . '() is deprecated and will be removed in Doctrine ORM 3.0. Use toIterable() instead.',
-            E_USER_DEPRECATED
+        Deprecation::trigger(
+            'doctrine/orm',
```
greg0ire, line 959: It's always going to be `doctrine/orm` here, right? It would be cool to auto-detect that, I did it some time ago in the PHPUnit bridge: https://github.com/symfony/phpunit-bridge/blob/0a67187d0c23cc00a14b6af752d5f8b2412c4f16/DeprecationErrorHandler/Deprecation.php#L213

beberlei: This code looks expensive to execute at runtime/production, I am unsure how to do this "performantly" without resorting to a simple thing such as regexping for `vendor/{package}/` in triggering file. That wouldn't work when run on the root package for example.

greg0ire: Ok let's ignore my comment then. I think one of the main selling points of your implementation is to be performant in production, let's try to keep it that way 👍

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -707,7 +707,12 @@ public function detach($entity)
      */
     public function merge($entity)
     {
-        @trigger_error('Method ' . __METHOD__ . '() is deprecated and will be removed in Doctrine ORM 3.0.', E_USER_DEPRECATED);
+        Deprecation::trigger(
+            'doctrine/orm',
+            '2.7',
+            'https://github.com/doctrine/orm/issues/8461',
+            'Method EntityManager::merge() is deprecated and will be removed in Doctrine ORM 3.0.'
```
greg0ire, line 714: Why not use `__METHOD__` here and below?

tests/Doctrine/Tests/Mapping/UnderscoreNamingStrategyTest.php
```diff
@@ -4,20 +4,35 @@
...
+
         new UnderscoreNamingStrategy(CASE_LOWER, false);
+
+        $after = Deprecation::getTriggeredDeprecations()['https://github.com/doctrine/orm/pull/7908'] ?? 0;
+
+        $this->assertSame($before + 1, $after);
+    }
+
+    /** @test */
+    public function checNoDeprecationMessageWhenNumberAwareEnabled(): void
```
greg0ire, line 28:
```suggestion
    public function checkNoDeprecationMessageWhenNumberAwareEnabled(): void
```

lib/Doctrine/ORM/EntityManager.php
beberlei: Yse, my bad :) Fixed

greg0ire (approved):

PR #8467, opened 2021-02-08
[GH-8410] Fix memory leak in new toIterable and state bug.

### LINKED ISSUE #8387: Query::iterate not marked as deprecated
Just a minor missing deprecation annotation in 2.7 and 2.8.
`AbstractQuery::iterate` is marked as deprecated but `Query::iterate` isn't.
It still triggers a message through its parent call.

### LINKED ISSUE #8410: Batch Processing using toIterable()

### LINKED ISSUE #8413: Issue with toIterable() and selecting multiple entities

beberlei:
The new AbstractQuery::toIterable() had a memory leak that AbstractQuery::iterable() did not have. This leak is now fixed.

After fixing the leak, one test failed where the identity map in ObjectHydrator triggered and lead to a notice. Introduced a new AbstractHydrator::cleanupAfterRowIteration() that the ObjectHydrator uses to cleanup the state.

When the result contains multiple entities, there must be special handling, otherwise only the last entity is returned.

In addition mixed result queries with scalar mappings don't work with iteration and are now explicitly prevented by exception.

Fixes #8410 Fixes #8413 Fixes #8387
Supersedes #8412 #8414

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -999,7 +999,7 @@ abstract class AbstractQuery
             $this->setParameters($parameters);
         }
 
-        $rsm  = $this->getResultSetMapping();
+        $rsm = $this->getResultSetMapping();
 
         if ($rsm->isMixed && count($rsm->scalarMappings) > 0) {
             throw QueryException::iterateWithMixedResultNotAllowed();
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -31,6 +31,7 @@ use Doctrine\ORM\Cache\TimestampCacheKey;
 use Doctrine\ORM\Internal\Hydration\IterableResult;
 use Doctrine\ORM\Mapping\MappingException as ORMMappingException;
 use Doctrine\ORM\Query\Parameter;
+use Doctrine\ORM\Query\QueryException;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\Persistence\Mapping\MappingException;
 use Traversable;
@@ -999,6 +1000,11 @@ abstract class AbstractQuery
         }
 
         $rsm  = $this->getResultSetMapping();
+
+        if ($rsm->isMixed && count($rsm->scalarMappings) > 0) {
+            throw QueryException::iterateWithMixedResultNotAllowed();
+        }
+
         $stmt = $this->_doExecute();
 
         return $this->_em->newHydrator($this->_hydrationMode)->toIterable($stmt, $rsm, $this->_hints);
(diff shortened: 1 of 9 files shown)


simPod (dismissed):

snapshotpl (dismissed):

alex-dev:
> In addition mixed result queries with scalar mappings don't work with iteration and are now explicitly prevented by exception.

Do we have a follow-up fix for this or an issue tracking it?

kbond:
I created an issue (#8520) to at least track the problem this PR introduced.

PR #8469, opened 2021-02-09
CS batch 5/an estimated 30

greg0ire:
From 2718 to 2584 errors


SenseException (approved):
What are those batches where you split up your changes? Do you target specific sniffs for every PR?

greg0ire:
No, it's completely random (well it's usually the last files in the output, or files with most changes, depending on how I feel), but in case you want to participate, just tell me which directory you are working on, I'll avoid it 👍

PR #8472, opened 2021-02-12
[GH-8471] Deprecate Partial DQL syntax and forcing partial loads.

### LINKED ISSUE #8471: Deprecate Partial Objects
We plan to remove partial objects in the forseeable future. This doesn't have to be ORM 3, because it depends on a few alternatives being in place.

Partial objects are a bad fix for wanting to avoid loading too much data like text fields /blobs that are not needed in the request. Two solutions are proposed to migrate towards around this deprecation and removal:

- Use data transfer objects with the `SELECT new MyDTO()`
- Add support for lazy loading some properties through a better proxy manager (this is not supported yet) https://github.com/doctrine/orm/issues/8518

beberlei:
Deprecates when you `SELECT PARTIAL` and when you use the query hint `Query::HINT_FORCE_PARTIAL_LOAD`.

Fixes #8471

### DIFF
--- a/docs/en/reference/dql-doctrine-query-language.rst
+++ b/docs/en/reference/dql-doctrine-query-language.rst
@@ -1377,7 +1377,8 @@ userland:
    that contain char or binary data. Doctrine has no way of implicitly
    reloading this data. Partially loaded objects have to be passed to
    ``EntityManager::refresh()`` if they are to be reloaded fully from
-   the database.
+   the database. This query hint is deprecated and will be removed
+   in the future (`Details <https://github.com/doctrine/orm/issues/8471>`_)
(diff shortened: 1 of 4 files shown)


SenseException:
Should the message contain the `SELECT NEW` alternative that can replace partial objects?

beberlei:
@SenseException My idea would be that people click the link, which then shows the alternative. Its not so easy to explain in the short space available for the message.

SenseException:
What if we keep that information in the documentation? Your current doc-changes don't contain the alternative and I would expect to find it there if not in a deprecation message.

greg0ire (dismissed):

SenseException (approved):

PR #8475, opened 2021-02-12
Cs fixes1

beberlei:


greg0ire (commented):
I think we should postpone the `_doExecute()` rename to at least 2.9.x (or 3.0)

beberlei:
You are right, slightly more defensive is good

SenseException:
> I think we should postpone the _doExecute() rename to at least 2.9.x (or 3.0)

Isn't that renaming connected to a new deprecation? Otherwise it will break on an 2.8 -> 2.9 update. I'd expect `_doExecute()` completely disappear in 3.0.

lib/Doctrine/ORM/Query.php
```diff
@@ -751,7 +757,7 @@ public function getLockMode()
      *
      * @return string
      */
-    protected function _getQueryCacheId()
+    protected function getQueryCacheId()
```
greg0ire, line 760: Ah we have another BC-break here

greg0ire (approved):

PR #8478, opened 2021-02-14
CS Fixes 2 string interpolation to sprintf

beberlei:
Used PHP Storm Context Actions string to sprintf for this.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/MappingException.php
+++ b/lib/Doctrine/ORM/Mapping/MappingException.php
@@ -74,7 +74,7 @@ class MappingException extends ORMException
      */
     public static function invalidInheritanceType($entityName, $type)
     {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -843,7 +920,7 @@ public static function infiniteEmbeddableNesting($className, $propertyName)
         );
     }
 
-    public static function illegalOverrideOfInheritedProperty($className, $propertyName)
+    public static function illegalOverrideOfInheritedProperty($className, $propertyName): MappingException
```
greg0ire, line 923: That class is not final, this is a BC-break for extending classes.

greg0ire (approved):

greg0ire:
Thanks @beberlei !

PR #8479, opened 2021-02-15
CS batch 10/an estimated 27

greg0ire:
From 1834 to 1665 errors


SenseException (approved):
I'm not sure what to think about the missing newlines between two properties. I guess it's okay because cs isn't complaining about it? What do you think?

greg0ire:
> I'm not sure what to think about the missing newlines between two properties. I guess it's okay because cs isn't complaining about it? What do you think?

I can fix it afterwards with a well thought out search and replace, but maybe we should find a sniff for this and integrate it in our coding standard?

PR #8483, opened 2021-02-17
Fix single scalar hydrator memory leak on exception

### LINKED ISSUE #8482: Fix memory leak in AbstractHydrator - Follow up
I have encountered an issue with a memory leak in SingleScalarHydrator.

`AbstractHydrator::hydrateAll()` method does not take into account the possibility of an exception being thrown so when `SingleScalarHydrator::hydrateAllData` throws `NoResultException`/`NonUniqueResultException`, the event listener for `onClear event` is not removed and the hydrator is stuck in the memory.

 It's been solved here once but not merged due to missing test cases.

Original issue: [Fix memory leak in AbstractHydrator](https://github.com/doctrine/orm/pull/7315)

I'll try to provide a fix and appropriate test case for this issue so it can be finally merged.

olsavmic:
Resolves #8482, Closes https://github.com/doctrine/orm/pull/7315

> I have encountered an issue with a memory leak in SingleScalarHydrator.
> 
> AbstractHydrator::hydrateAll() method does not take into account the possibility of an exception being thrown so when SingleScalarHydrator::hydrateAllData throws NoResultException/NonUniqueResultException, the event listener for onClear event is not removed and the hydrator is stuck in the memory.
> 
> It's been solved here once but not merged due to missing test cases.
> 
> Original issue: Fix memory leak in AbstractHydrator
> 
> I'll try to provide a fix and appropriate test case for this issue so it can be finally merged.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Hydration/AbstractHydratorTest.php
+++ b/tests/Doctrine/Tests/ORM/Hydration/AbstractHydratorTest.php
@@ -10,6 +10,7 @@ use Doctrine\DBAL\Driver\Statement;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Events;
 use Doctrine\ORM\Internal\Hydration\AbstractHydrator;
+use Doctrine\ORM\ORMException;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\Tests\OrmFunctionalTestCase;
 use PHPUnit_Framework_MockObject_MockObject;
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Hydration/AbstractHydratorTest.php
```diff
@@ -106,4 +107,31 @@ public function testHydrateAllRegistersAndClearsAllAttachedListeners(): void
...
     }
+
+    /**
+     * @group #DDC-8482
+     */
+    public function testHydrateAllClearsAllAttachedListenersEvenOnError(): void
+    {
+        $this
+            ->mockEventManager
+            ->expects(self::at(0))
```
greg0ire, line 118: `self::at` is deprecated in recent versions of PHPUnit, please avoid it

greg0ire:

How to do that?

1. `git rebase -i origin/2.8.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

tests/Doctrine/Tests/ORM/Hydration/AbstractHydratorTest.php
olsavmic: I've changed it and rebased the MR although I did not find a clean replacement for the self::at method.

I want to keep the check of the method call order so please look into it and say if it seems ok to you.

I can update the other methods in the test as well although it's out of scope of this MR so I'm not sure if I should :)

olsavmic: Fixed and rebased, although I do not like the replacement (I did not find any other recommended way). Say if it seems ok to you please. I might replace the the rest of self::at occurrences in this Test class if requested (although it's out of scope of this MR)

tests/Doctrine/Tests/ORM/Hydration/AbstractHydratorTest.php
```diff
@@ -106,4 +107,40 @@ public function testHydrateAllRegistersAndClearsAllAttachedListeners(): void
 
         $this->hydrator->hydrateAll($this->mockStatement, $this->mockResultMapping);
     }
+
+    /**
+     * @group #DDC-8482
```
greg0ire, line 112: `DDC-` is a prefix we used when we were using JIRA: https://github.com/doctrine/jira-github-issues/blob/89d31b0477cd0ade65535659c92f173f9ca83abc/projects.php#L4

You should probably use `GH` here instead, and that's if a group is really needed (not sure we recommend doing that in our documentation).

tests/Doctrine/Tests/ORM/Hydration/AbstractHydratorTest.php
greg0ire: > I might replace the the rest of self::at occurrences in this Test class if requested (although it's out of scope of this MR)

It would certainly be appreciated, but it's not at all required.

greg0ire (approved):

beberlei (approved):

SenseException (approved):

greg0ire:
Thanks @olsavmic !

PR #8487, opened 2021-02-18
Update QueryBuilder::setParameters docs

Philosoft:
Use `ArrayCollection` instead of plain array (which is supported only for bc)


SenseException (approved):

SenseException:
Thank you @Philosoft

PR #8489, opened 2021-02-18
Property with same name as discriminator column returns wrong instance.

Warxcell:
This PR contains the test for it.
I will be happy to discuss how to fix that issue.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
@@ -230,7 +230,7 @@ class ObjectHydrator extends AbstractHydrator
      *
      * @throws HydrationException
      */
-    private function getEntity(array $data, $dqlAlias, array $rawData = [])
+    private function getEntity(array $data, $dqlAlias, array $rawData): object
     {
         $className = $this->_rsm->aliasMap[$dqlAlias];
 
--- a/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
@@ -229,37 +229,29 @@ class ObjectHydrator extends AbstractHydrator
      *
      * @throws HydrationException
      */
-    private function getEntity(array $data, $dqlAlias)
+    private function getEntity(array $data, $dqlAlias, array $rawData = [])
     {
         $className = $this->_rsm->aliasMap[$dqlAlias];
 
         if (isset($this->_rsm->discriminatorColumns[$dqlAlias])) {
             $fieldName = $this->_rsm->discriminatorColumns[$dqlAlias];
 
-            if (! isset($this->_rsm->metaMappings[$fieldName])) {
-                throw HydrationException::missingDiscriminatorMetaMappingColumn($className, $fieldName, $dqlAlias);
+            if (! isset($rawData[$fieldName])) {
+                throw HydrationException::missingDiscriminatorColumn($className, $fieldName, $dqlAlias);
             }
 
-            $discrColumn = $this->_rsm->metaMappings[$fieldName];
-
-            if (! isset($data[$discrColumn])) {
-                throw HydrationException::missingDiscriminatorColumn($className, $discrColumn, $dqlAlias);
-            }
-
-            if ($data[$discrColumn] === '') {
+            if ($rawData[$fieldName] === '') {
                 throw HydrationException::emptyDiscriminatorValue($dqlAlias);
             }
 
             $discrMap           = $this->_metadataCache[$className]->discriminatorMap;
-            $discriminatorValue = (string) $data[$discrColumn];
+            $discriminatorValue = (string) $rawData[$fieldName];
 
             if (! isset($discrMap[$discriminatorValue])) {
                 throw HydrationException::invalidDiscriminatorValue($discriminatorValue, array_keys($discrMap));
             }
 
             $className = $discrMap[$discriminatorValue];
-
-            unset($data[$discrColumn]);
         }
 
         if (isset($this->_hints[Query::HINT_REFRESH_ENTITY]) && isset($this->rootAliases[$dqlAlias])) {
@@ -364,7 +356,7 @@ class ObjectHydrator extends AbstractHydrator
                     $parentObject = $this->resultPointers[$parentAlias];
                 } else {
                     // Parent object of relation not found, mark as not-fetched again
-                    $element = $this->getEntity($data, $dqlAlias);
+                    $element = $this->getEntity($data, $dqlAlias, $row);
 
                     // Update result pointer and provide initial fetch data for parent
                     $this->resultPointers[$dqlAlias]               = $element;
@@ -403,7 +395,7 @@ class ObjectHydrator extends AbstractHydrator
                                     unset($this->resultPointers[$dqlAlias]);
                                 }
                             } else {
-                                $element = $this->getEntity($data, $dqlAlias);
+                                $element = $this->getEntity($data, $dqlAlias, $row);
 
                                 if (isset($this->_rsm->indexByMap[$dqlAlias])) {
                                     $indexValue = $row[$this->_rsm->indexByMap[$dqlAlias]];
@@ -435,7 +427,7 @@ class ObjectHydrator extends AbstractHydrator
                         // we only need to take action if this value is null,
                         // we refresh the entity or its an uninitialized proxy.
                         if (isset($nonemptyComponents[$dqlAlias])) {
-                            $element = $this->getEntity($data, $dqlAlias);
+                            $element = $this->getEntity($data, $dqlAlias, $row);
                             $reflField->setValue($parentObject, $element);
                             $this->_uow->setOriginalEntityProperty($oid, $relationField, $element);
                             $targetClass = $this->_metadataCache[$relation['targetEntity']];
@@ -491,7 +483,7 @@ class ObjectHydrator extends AbstractHydrator
 
                 // check for existing result from the iterations before
                 if (! isset($this->identifierMap[$dqlAlias][$id[$dqlAlias]])) {
-                    $element = $this->getEntity($data, $dqlAlias);
+                    $element = $this->getEntity($data, $dqlAlias, $row);
 
                     if ($this->_rsm->isMixed) {
                         $element = [$entityKey => $element];
--- a/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
@@ -224,6 +224,7 @@ class ObjectHydrator extends AbstractHydrator
      *
      * @param array  $data     The instance data.
      * @param string $dqlAlias The DQL alias of the entity's class.
+     * @param array  $rawData  Raw data directly from query.
      *
      * @return object The entity.
      *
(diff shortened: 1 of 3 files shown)


Warxcell:
Hm.... why tests are successful here?


```php
 ./vendor/bin/phpunit tests/Doctrine/Tests/ORM/Functional/JoinedTableWithPropertyAsDiscriminatorColumn.php 
PHPUnit 9.5.2 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)

Time: 00:00.034, Memory: 12.00 MB

There was 1 failure:

1) Doctrine\Tests\ORM\Functional\JoinedTableWithPropertyAsDiscriminatorColumn::testExplicitPolicy
Failed asserting that Doctrine\Tests\ORM\Functional\JoinedTableWithPropertyAsDiscriminatorColumnChild2 Object (...) is an instance of class "Doctrine\Tests\ORM\Functional\JoinedTableWithPropertyAsDiscriminatorColumnChild".

/home/bozhidar_hristov/projects/orm/tests/Doctrine/Tests/ORM/Functional/JoinedTableWithPropertyAsDiscriminatorColumn.php:36

```

greg0ire:
I have tried locally, the tests fails when run alone, and passes when you run the whole test suite.

Warxcell:
> I have tried locally, the tests fails when run alone, and passes when you run the whole test suite.

Weird. Any hints where to seek for the problem?

greg0ire:
> Weird. Any hints where to seek for the problem? EntityManager::clear is called just before the actual test.

I usually troubleshoot that kind of thing by building a list of all test files in `phpunit.xml`, under `testsuites > testsuite`, then I execute the test suite again. If it still fails, I remove all the files executed after the test from the list cause they are obviously not going to cause any issue.

Then I do a bisection: remove half the list. If it still fails remove half of that, if not restore the list and remove the other half, etc. etc. . You should end up with one test which does something that affects the global state `putenv`, things like that.

Warxcell:
Actually it seems that my test is not executed.

Warxcell:
Nevermind. Found it. Can be reviewed.

greg0ire:
Before we review, please fix the build, and reduce the number of commits `vendor/bin/phpcbf` can help.

Warxcell:
I won't fix anything yet, because i'm almost sure that this will be rejected. :D 

Can you please review and give advice if something needs to be refactored. After you approve the change itself I will fix the CS and squash it into single commit.

greg0ire:
> Add test to verify that using property with same name as discriminator column will fail.

The description of this PR is a bit succinct… and misleading. I thought you were adding a failing test, but in fact, you are also touching `src`.

greg0ire:
The added test is pretty clear though, so it's easy to understand what you want to do. Should there be docs about this new restriction? Also, do you think there is any chance people might be relying on this? Maybe some expose it in a property but never write to it?

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
```diff
@@ -224,42 +224,35 @@ private function initRelatedCollection($entity, $class, $fieldName, $parentDqlAl
...
+    private function getEntity(array $data, $dqlAlias, array $rawData): object
     {
         $className = $this->_rsm->aliasMap[$dqlAlias];
 
         if (isset($this->_rsm->discriminatorColumns[$dqlAlias])) {
             $fieldName = $this->_rsm->discriminatorColumns[$dqlAlias];
 
-            if (! isset($this->_rsm->metaMappings[$fieldName])) {
-                throw HydrationException::missingDiscriminatorMetaMappingColumn($className, $fieldName, $dqlAlias);
+            if (! isset($rawData[$fieldName])) {
```
greg0ire, line 240: Please explain why this fixes the issue. I don't know the ORM well, and even if I did understand it enough, I would still ask you to explain this in a commit message.

Warxcell: image
I guess this picture will tell everything.

greg0ire:
> Should there be docs about this new restriction?

Oh but I misunderstood, you don't want there to be any restriction, right? You want people to be able to do this (accessing the discriminator column from a child class… isn't it like a can of worms? I mean, what if you have instance A, and you write 'B' in its discriminator property?

Warxcell:
> > Should there be docs about this new restriction?
> 
> Oh but I misunderstood, you don't want there to be any restriction, right? You want people to be able to do this (accessing the discriminator column from a child class… isn't it like a can of worms? I mean, what if you have instance A, and you write 'B' in its discriminator property?

I don't write in discriminator property. (discriminator is not property at all).
I just write to a property with same name as discriminator column, and this leads to bug.

greg0ire:
Ok now I have a far better understanding on what this is about, you should really have written something like that when posting your PR.

> Discriminator should be something imaginary, right? There shouldn't be a property for it, so it's name should be available for usage. (Of course COLUMN should be restricted).

I agree with you on the theory, but in practice, I feel uneasy: wouldn't it be confusing for people reading the code if somebody happens to use a property with the very same name as the discriminator column? I'm torn between your solution and throwing an exception forbidding people to do that. Since you went through the trouble of writing a fix for this, I suppose you have an example showing how this can actually make sense?

> Furthermore currently ORM behaves differently depending on if you are using Query or directly Persister/Repository.
Hydrator(Query) returns wrong instance, Persister/Repository returns correct instance.

Ok so we definitely have a change to do here, that's for sure. Let's first establish what would be the best behavior and see how it can be implemented from here.

Warxcell:
> Ok now I have a far better understanding on what this is about, you should really have written something like that when posting your PR.
> 
> > Discriminator should be something imaginary, right? There shouldn't be a property for it, so it's name should be available for usage. (Of course COLUMN should be restricted).
> 
> I agree with you on the theory, but in practice, I feel uneasy: wouldn't it be confusing for people reading the code if somebody happens to use a property with the very same name as the discriminator column? I'm torn between your solution and throwing an exception forbidding people to do that. Since you went through the trouble of writing a fix for this, I suppose you have an example showing how this can actually make sense?
> 
> > Furthermore currently ORM behaves differently depending on if you are using Query or directly Persister/Repository.
> > Hydrator(Query) returns wrong instance, Persister/Repository returns correct instance.
> 
> Ok so we definitely have a change to do here, that's for sure. Let's first establish what would be the best behavior and see how it can be implemented from here.

Yes, sure. I accidentally found this bug when I had this structure/logic:

Parameter (joined table, discriminator: type)
- IntegerParameter
- DateParameter
- DateTimeParameter
- TimeParameter
- DecimalParameter
- BooleanParameter
- CalculatedParameter
-- expression
-- type <= this contains discriminator value of `ParameterValue`

ParameterValue: (joined table, discriminator: type)
- IntegerValue
-- value
- DateValue
-- value
- DateTimeValue
-- value
- TimeValue
-- value
- DecimalValue
-- value
- BooleanValue
-- value

So when my code meet `CalculatedParameter` it created corresponding type of `ParameterValue` depending on `type` column.
In my particular example so avoid the bug, I renamed `type` property to `valueType`.
But in my opinion - it's good to allow people having parameter with same name, since discriminator is a column definition, not property definition.

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
greg0ire: > I guess this picture will tell everything.

Sorry, but not really… what's the value of `$fieldName` here? And you talk a lot about the "wrong instance" but never say what the "wrong instance" here… is it an existing object, or a new instance of the wrong class, and if yes, which class? The parent one, another child?

Warxcell: `$data` is something middle between data returned by query (`$rawData`) and just before it's mapped to properties.

in `$data` discriminator column is overwritten by real property with same name. I don't know why discriminator is in data at all.

Warxcell: > And you talk a lot about the "wrong instance" but never say what the "wrong instance" here… is it an existing object, or a new instance of the wrong class, and if yes, which class? The parent one, another child?

The test shows where which one is wrong instance.

Warxcell:
```php
$object       = new JoinedTableWithPropertyAsDiscriminatorColumnChild();
$object->type = 'child2';
```
We persist Child as instanceof `Child`

then we retrieve it
```php
$q = $this->_em->createQuery('SELECT o FROM ' . JoinedTableWithPropertyAsDiscriminatorColumnRoot::class . ' o WHERE o.id = :id');
$q->setParameter('id', $child->id);
$object = $q->getSingleResult();
```
Here object is `Child2` instance, NOT `Child` as it's persisted.

Warxcell:
Other possible scenario for this is

`Property` (Joined table, discriminatorColumn: type)
- House
- - type (column: house_type) (studio, penthouse, etc.)
- Apartment
-

Warxcell:
Renamed tests to something more meaningful, so they will be easier to understand.

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
greg0ire: > The test shows where which one is wrong instance.

Does it? The test should show the right behavior, I wouldn't expect it to mention the wrong instance at all, unless maybe you assert that the returned instance is not the wrong one, rather than asserting that it is the right one…

greg0ire: At the time where you took your screenshot above, what was the value of `$fieldName`?

tests/Doctrine/Tests/ORM/Functional/JoinedTableWithPropertyAsDiscriminatorColumnTest.php
```diff
@@ -0,0 +1,212 @@
...
+
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class JoinedTableWithPropertyAsDiscriminatorColumnTest extends OrmFunctionalTestCase
+{
+    private const MODELS = [
+        Property::class,
+        Apartment::class,
+        House::class,
+        ReferencedProperty::class,
```
greg0ire, line 15: That's a nice change, it's easier to understand with something more concrete 👍

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
greg0ire: > in $data discriminator column is overwritten by real property with same name. I don't know why discriminator is in $data at all. That's why i pass $rawData and get discriminator from there,before it's overwritten.

Ok so to make the fix more clean, we should probably also make sure the discriminator never ends up in `$data` as well. That would make sense, since it is not supposed to be mapped in the object.

Warxcell: >  The test should show the right behavior
Exactly and it fails if I remove my fix. 
> Ok so to make the fix more clean, we should probably also make sure the discriminator never ends up in $data as well. That would make sense, since it is not supposed to be mapped in the object.

Exactly.

greg0ire: So maybe this should be done here:

https://github.com/doctrine/orm/blob/9836092602e7935966f4e5b0b354b78975568359/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php#L343

We should make sure not to gather row data from discriminator columns so as not to pollute `$rowData['data']` with it. You should be able to do that based on the contents of this variable IMO:

https://github.com/doctrine/orm/blob/9836092602e7935966f4e5b0b354b78975568359/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php#L344

beberlei:
This is not supported, a discriminator column is not visible outside of the class name, we should make this fail in either ClassMetadata or in SchemaValidator.

PR #8495, opened 2021-02-21
Fix bug when using Result Cache with Query::toIterable

Warxcell:
`Argument 1 passed to Doctrine\ORM\Internal\Hydration\AbstractHydrator::toIterable() must implement interface Doctrine\DBAL\Driver\Statement, instance of Doctrine\DBAL\Cache\ResultCacheStatement given, called in /app/vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php on line 1010`

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -20,7 +20,7 @@
 
 namespace Doctrine\ORM\Internal\Hydration;
 
-use Doctrine\DBAL\Driver\Statement;
+use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
@@ -93,7 +93,7 @@ abstract class AbstractHydrator
     /**
      * The statement that provides the data to hydrate.
      *
-     * @var Statement
+     * @var ResultStatement
      */
     protected $_stmt;
 
@@ -154,7 +154,7 @@ abstract class AbstractHydrator
      *
      * @return iterable<mixed>
      */
-    public function toIterable(Statement $stmt, ResultSetMapping $resultSetMapping, array $hints = []): iterable
+    public function toIterable(ResultStatement $stmt, ResultSetMapping $resultSetMapping, array $hints = []): iterable
     {
         $this->_stmt  = $stmt;
         $this->_rsm   = $resultSetMapping;
--- a/tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php
@@ -12,13 +12,14 @@ use Doctrine\Tests\OrmFunctionalTestCase;
 use ReflectionProperty;
 
 use function count;
+use function iterator_to_array;
 
 /**
  * ResultCacheTest
  */
 class ResultCacheTest extends OrmFunctionalTestCase
 {
-   /** @var ReflectionProperty */
+    /** @var ReflectionProperty */
     private $cacheDataReflection;
 
     protected function setUp(): void
@@ -164,6 +165,40 @@ class ResultCacheTest extends OrmFunctionalTestCase
         $query->getResult();
 
         $this->assertTrue($cache->contains('testing_result_cache_id'));
+
+        $this->_em->getConfiguration()->setResultCacheImpl(new ArrayCache());
+    }
+
+    public function testEnableResultCacheWithIterable(): void
+    {
+        $cache            = new ArrayCache();
+        $expectedSQLCount = count($this->_sqlLoggerStack->queries) + 1;
+
+        $query = $this->_em->createQuery('select ux from Doctrine\Tests\Models\CMS\CmsUser ux');
+        $query->enableResultCache();
+        $query->setResultCacheDriver($cache);
+        $query->setResultCacheId('testing_iterable_result_cache_id');
+        iterator_to_array($query->toIterable());
+
+        $this->_em->clear();
+
+        $this->assertCount(
+            $expectedSQLCount,
+            $this->_sqlLoggerStack->queries
+        );
+        $this->assertTrue($cache->contains('testing_iterable_result_cache_id'));
+
+        $query = $this->_em->createQuery('select ux from Doctrine\Tests\Models\CMS\CmsUser ux');
+        $query->enableResultCache();
+        $query->setResultCacheDriver($cache);
+        $query->setResultCacheId('testing_iterable_result_cache_id');
+        iterator_to_array($query->toIterable());
+
+        $this->assertCount(
+            $expectedSQLCount,
+            $this->_sqlLoggerStack->queries,
+            'Expected query to be cached'
+        );
 
         $this->_em->getConfiguration()->setResultCacheImpl(new ArrayCache());
     }


lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -154,7 +155,7 @@ public function iterate($stmt, $resultSetMapping, array $hints = [])
      *
      * @return iterable<mixed>
      */
-    public function toIterable(Statement $stmt, ResultSetMapping $resultSetMapping, array $hints = []): iterable
+    public function toIterable(ResultStatement $stmt, ResultSetMapping $resultSetMapping, array $hints = []): iterable
```
greg0ire, line 158: This looks a lot like a BC-break… under what circumstances do you observe the issue?

Warxcell: I added a test.

Warxcell: I doubt it's BC-break, because `Statement` extends `ResultStatement`. So it's the opposite of more restricting interface.

tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php
```diff
@@ -168,6 +167,23 @@ public function testEnableResultCache(): void
         $this->_em->getConfiguration()->setResultCacheImpl(new ArrayCache());
     }
 
+    public function testEnableResultCacheWithIterable(): void
```
Warxcell, line 170: @greg0ire this is the test case.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
greg0ire: It's a BC-break for extending classes, that now have to also relax their interface , and handle more types than they already do.

Oh but I now notice that `AbstractHydrator` is inside the `Internal` namespace, so that must be OK anyway I suppose :slightly_smiling_face:

tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php
```diff
@@ -168,6 +169,21 @@ public function testEnableResultCache(): void
...
+        $query = $this->_em->createQuery('select ux from Doctrine\Tests\Models\CMS\CmsUser ux');
+
+        $query->enableResultCache();
+        $query->setResultCacheDriver($cache);
+        $query->setResultCacheId('testing_iterable_result_cache_id');
+        iterator_to_array($query->toIterable());
+
+        $this->assertTrue($cache->contains('testing_iterable_result_cache_id'));
+
+        $this->_em->getConfiguration()->setResultCacheImpl(new ArrayCache());
```
greg0ire, line 184: :thinking: I don't understand why that last line is not an assertion… what's the point of doing something if you don't perform assertions afterwards, or setup expectations beforehand?

tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php
greg0ire: Wow, that's suprising, it looks like a very common scenario… kind of weird that nobody else reported the issue. Did it happen since a recent upgrade, or did you experiencing on new code you wrote?

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
Warxcell: And the method below this one is 

```public function hydrateAll($stmt, $resultSetMapping, array $hints = [])`

No typehint at all.

Comment (greg0ire) [reply lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php:157]: Yes but that's not suspicious: the git log shows that it was written at a time where type declarations did not even exist. Since the whole namespace is internal, I think it would be ok for you to add type hints if you want to. Maybe it was never reconsidered.

Comment (Warxcell) [reply lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php:157]: Actually it seems it can be nullable ?
`
Failed asserting that exception of type "TypeError" matches expected exception "Doctrine\ORM\Query\QueryException". Message was: "Mock_AbstractHydrator_836a32f6::hydrateAll(): Argument #1 ($stmt) must be of type Doctrine\DBAL\Driver\ResultStatement, null given, called in /home/bozhidar_hristov/projects/orm/lib/Doctrine/ORM/AbstractQuery.php on line 1082" at
`

Comment (Warxcell) [reply lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php:157]: It's might be because of the test itself:

 `$this->connection->method('executeQuery')->with($this->anything(), []);`

Comment (Warxcell) [reply lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php:157]: Also it seems some of Docblock typehints are wrong:

`     
* @return array
*/
public function hydrateAll
`

but then

`13) Doctrine\Tests\ORM\Hydration\SingleScalarHydratorTest::testHydrateSingleScalar with data set #0 ('result2', array(array('1')))
TypeError: Doctrine\ORM\Internal\Hydration\AbstractHydrator::hydrateAll(): Return value must be of type array, int returned`

Comment (Warxcell) [reply lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php:157]: Added typehints everywhere, just not on `$stmt` param on `hydrateAll`, because of failing tests.

File: tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php (hunk 3 above)
Comment (greg0ire) [reply tests/Doctrine/Tests/ORM/Functional/ResultCacheTest.php:184]: Oh ok, it makes sense then 👍

File: lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php [hunk 4]
```diff
@@ -121,13 +121,13 @@ public function __construct(EntityManagerInterface $em)
      *
      * @deprecated
      *
-     * @param object $stmt
+     * @param ResultStatement $stmt
```
greg0ire, line 124: `vendor/bin/phpcbf` is your friend :)

Warxcell: Yeah. just ran it.

Warxcell:
`Error: Method \Doctrine\ORM\Internal\Hydration\AbstractHydrator::gatherRowData() does not have @param annotation for its traversable parameter $id.
Error: Method \Doctrine\ORM\Internal\Hydration\AbstractHydrator::gatherRowData() does not have @param annotation for its traversable parameter $nonemptyComponents.`

Whyy???

```
     * @param array<string, string>  &$id                 Dql-Alias => ID-Hash.
     * @param array<string, boolean> &$nonemptyComponents Does this DQL-Alias has at least one non NULL value? 
 ```

Warxcell:
Adjusted a bit test, squashed it in 3 commits (add test, fix bug, add typehints).

I hope it's all good now. :)

greg0ire (dismissed):
It looks good to me, although I would squash the failing test and it fix in just one commit: it would not make sense to revert one without also reverting the other. Also, tests should pass on every commit IMO.

Warxcell:
> It looks good to me, although I would squash the failing test and it fix in just one commit: it would not make sense to revert one without also reverting the other. Also, tests should pass on every commit IMO.

My logic behind this is that one could provide better implementation of fix and that would required reverting the fix, but not and the test. :)

beberlei:
Isnt adding types a bc break? Hydrators are a regular extension point of the ORM so we need to be extra careful.

greg0ire:
@beberlei they all are in an `Internal` namespace, what about that? Is it not about the BC promise?

SenseException:
I understand and agree that internal classes can have changes that are usually considered a BC break, but all those other type hints and return types are out of scope for the bug fix and don't add anything to the bugfix.

greg0ire:
@Warxcell let's drop that third commit then?

Warxcell:
> @Warxcell let's drop that third commit then?

Done. :(

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @Warxcell !

PR #8501, opened 2021-02-22
Composite ID not working when having fields that are objects.

Warxcell:
In this example it's not working with DateTime.

Can you give me advice where the change should happen? I'm a bit confused, because Relations get the ID as value, not the whole object - so the conversion should happen somewhere else rathen than here  https://github.com/doctrine/orm/blob/2.8.x/lib/Doctrine/ORM/UnitOfWork.php#L1550 ?

Why is ORM using PHP values and not raw DB values?

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/CompositeIdWithDateTimeTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/CompositeIdWithDateTimeTest.php
@@ -1,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use DateTimeImmutable;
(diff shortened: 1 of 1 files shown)


beberlei:
The problem is probably that DateTimes cannot be converted to string, to make this work you need to create a DBAL Type for the datetime that has `__toString`. This is a long time usability problem which we haven't adressed yet.

Warxcell:
> The problem is probably that DateTimes cannot be converted to string, to make this work you need to create a DBAL Type for the datetime that has `__toString`. This is a long time usability problem which we haven't adressed yet.

Which object will have the `__toString` ? I think the type itself is not used as string anywhere?
I'm thinking of using `Metadata::getType($field)->convertToDatabaseValue($value)` ?

PR #8508, opened 2021-02-23
Proposed fix for #8499

### LINKED ISSUE #8499: Optimistic lock not working when used with DateTime objects.
I will try to be as concrete and specific as possible, I think there is a bug when comparing the DateTime objects used in the optimistic lock.

In the UnitOfWork (Line 2497) you are using a !== comparison that only resolves to true when the two instances of the compared object are the same.  This way it is totally impossible for $entityVersion and $lockVersion to be the same at any time since they are different objects. 

```php
    if ($entityVersion !== $lockVersion) {
        throw OptimisticLockException::lockFailedVersionMismatch($entity, $lockVersion, $entityVersion);
    }
```
Debugging to that point I get the following pair of objects, that should pass the validation but they don't:

[Debug variables](https://ibb.co/QXw6kvt)

I have checked the tests that cover the lock functionality and the only thing that is checked is that the appropriate exceptions are thrown, which is always the case since the result of the comparison is always false. There is no test that covers the success of the lock operation.

In older versions of the code the comparison was performed using the != operator with which the functionality works as expected. 

Possible solutions could be to revert to using the != operator or to directly compare the timestamps of DateTime objects using the !== operator.

I can send a PR with the proposed fix.

Thanks in advance.

diego-ninja:
Changed lock function to compare timestamps instead of DateTimeInterface objects directly.

When using optimistic lock with DateTimeInterface based version field a bug appears due to the use of the === operator for comparing the lock version and the entity version. This comparison always resolves to false because the === operator when comparing objects is only true when both sides are the exact same instance of the object.

To fix the issue I have decided to compare timestamps instead the DateTimeInterface based objects directly, calling getTimestamp() method and doing a strict comparison.

Modified OptimisticLockException to use DateTimeInterface instead of DateTime class.

Added test suite to cover case.

Fixes #8499

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8499Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8499Test.php
@@ -1,0 +1,166 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use DateTime;
+use DateTimeInterface;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\LockMode;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\Table;
+use Doctrine\ORM\Mapping\Version;
+use Doctrine\ORM\OptimisticLockException;
+use Doctrine\Tests\OrmFunctionalTestCase;
+use Exception;
+
+use function date;
+use function strtotime;
+
+class GH8499Test extends OrmFunctionalTestCase
+{
+    /** @var Connection */
+    protected $conn;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        try {
+            $this->_schemaTool->createSchema(
+                [$this->_em->getClassMetadata(GH8499VersionableEntity::class)]
+            );
+        } catch (Exception $e) {
+            // Swallow all exceptions. We do not test the schema tool here.
+        }
+
+        $this->conn = $this->_em->getConnection();
+    }
+
+    /**
+     * @group GH-8499
+     */
+    public function testOptimisticTimestampSetsDefaultValue(): GH8499VersionableEntity
+    {
+        $entity = new GH8499VersionableEntity();
+        $entity->setName('Test Entity');
+        $entity->setDescription('Entity to test optimistic lock fix with DateTimeInterface objects');
+        self::assertNull($entity->getRevision(), 'Pre-Condition');
+
+        $this->_em->persist($entity);
+        $this->_em->flush();
+
+        self::assertInstanceOf(DateTimeInterface::class, $entity->getRevision());
+
+        return $entity;
+    }
+
+    /**
+     * @group GH-8499
+     * @depends testOptimisticTimestampSetsDefaultValue
+     */
+    public function testOptimisticLockWithDateTimeForVersion(GH8499VersionableEntity $entity): void
+    {
+        $q = $this->_em->createQuery('SELECT t FROM Doctrine\Tests\ORM\Functional\Ticket\GH8499VersionableEntity t WHERE t.id = :id');
+        $q->setParameter('id', $entity->id);
+        $test = $q->getSingleResult();
+
+        $format       = $this->_em->getConnection()->getDatabasePlatform()->getDateTimeFormatString();
+        $modifiedDate = new DateTime(date($format, strtotime($test->getRevision()->format($format)) - 3600));
+
+        $this->conn->executeQuery('UPDATE GH8499VersionableEntity SET revision = ? WHERE id = ?', [$modifiedDate->format($format), $test->id]);
+
+        $this->_em->refresh($test);
+        $this->_em->lock($test, LockMode::OPTIMISTIC, $modifiedDate);
+
+        $test->setName('Test Entity Locked');
+        $this->_em->persist($test);
+        $this->_em->flush();
+
+        self::assertEquals('Test Entity Locked', $test->getName(), 'Entity not modified after persist/flush,');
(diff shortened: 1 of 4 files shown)


greg0ire:

diego-ninja:
> Please improve your commit message according to [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches).

Done.

greg0ire:
I meant the _commit_ message, so you can squash your current commits together, and reuse your PR message as a commit message. BTW, when you create a PR with just one commit, Github will reuse its message as a PR message.

diego-ninja:
> I meant the _commit_ message, so you can squash your current commits together, and reuse your PR message as a commit message. BTW, when you create a PR with just one commit, Github will reuse its message as a PR message.

I hope it's ok now.

greg0ire:
I will do it for you.

greg0ire:
`DDC-` is a prefix that was used when we were using JIRA a long time ago, I will replace it with `GH`. Sorry for breaking this BTW, that was due to a CS-fixing commit of mine :bow:

greg0ire:
:thinking: is this line wrong: https://github.com/doctrine/orm/blob/e40ac3e1d0f82e39c8ae9990866773372a4a069d/lib/Doctrine/ORM/EntityManagerInterface.php#L218 ? It does not mention `\DateTimeInterface` although the docs do mention using a `datetime` field as a revision: https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/transactions-and-concurrency.html#locking-support

This line might be wrong too: https://github.com/doctrine/orm/blob/e40ac3e1d0f82e39c8ae9990866773372a4a069d/lib/Doctrine/ORM/UnitOfWork.php#L2461

tests/Doctrine/Tests/ORM/Functional/Ticket/GH8499Test.php
```diff
@@ -0,0 +1,175 @@
...
+        $firstVersion = $entity->getRevision();
+        sleep(1);
+
+        $entity->setName('Test Entity Updated');
+        $this->_em->persist($entity);
+        $this->_em->flush();
+        $this->_em->refresh($entity);
+        $lastVersion = $entity->getRevision();
+
+        $this->assertNotEquals($firstVersion, $lastVersion, 'Version field value not updated on persist');
```
greg0ire: Let's be consistent with the test below?
```suggestion
        $this->assertNotEquals($firstVersion, $entity->getRevision(), 'Version field value not updated on persist');
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH8499Test.php
```diff
@@ -0,0 +1,175 @@
...
+    public function testOptimisticLockUpdatesVersionOnEntityUpdate(): void
+    {
+        $entity = new GH8499VersionableEntity();
+        $entity->setName('Test Entity');
+        $entity->setDescription('Entity to test optimistic lock fix with DateTimeInterface objects');
+        $this->_em->persist($entity);
+        $this->_em->flush();
+        $this->_em->refresh($entity);
+        $firstVersion = $entity->getRevision();
+        sleep(1);
```
greg0ire: I wish there was a better way to do this, maybe another maintainer will have a better idea. Right now, I think this test is probably the longest test of the whole test suite.

greg0ire: I have given it more thought:
That feature also works with integer Versions.
You already tested that the version was initialized on insert. If an entity's integer version is updated on update, I think we can reasonably think it will work too for a datetime version.

1. Are we sure such a test does not already exist?
2. Should we rewrite that particular test with an integer field so that we can avoid that call to `sleep`? And if yes, maybe it shouldn't be in `GH9499Test`. This test passes without your fix, doesn't it?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH8499Test.php
```diff
@@ -0,0 +1,173 @@
...
+        $firstVersion = $entity->getRevision();
+
+        $this->_em->lock($entity, LockMode::OPTIMISTIC, $firstVersion);
+        sleep(1);
+
+        $entity->setName('Test Entity Locked');
+        $this->_em->persist($entity);
+        $this->_em->flush();
+
+        $this->assertNotEquals($firstVersion, $entity->getRevision(), 'Version field value not updated on persist');
```
greg0ire: We already tested that above, right? IMO, we should test the locking mechanism independently from version update. Could we achieve that by adding a setter for the version field and calling it at the right time?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH8499Test.php
beberlei: Use dbal to update the row into the past

tests/Doctrine/Tests/ORM/Functional/Ticket/GH8499Test.php
diego-ninja: Yeah, your right, stritctly speaking I should only test two things, that the lock function works with a correct dataset and throws a exception when not. So the rest of the test cases are not needed here, I mean, the case of the version field being initailized and updated in a proper way should be tested anywhere else but not here. So, I'm going to remove the "extra" cases from the test suite.

sounds good to you?

diego-ninja: Okie, I'm checking the OptimisticTest.php included in Tests\ORM\Functional\Locking directory again, there are a bunch of tests which covers all cases except the one with lock mechanism working as expected, the one that would have detected this bug, So I'm going to write that test only. 

it's that okie?

greg0ire: It's super okie 👍 Thanks a lot for investigating this.

diego-ninja: It's done, I write the test without using sleep function, following the work done in OptimisticTest.php. Check it, I think is a good approach, I followed the @beberlei advice and modified the version datetime in the past using dbal.

greg0ire (approved):
Looks great 👍

beberlei (approved):

PR #8516, opened 2021-02-28
 CS batch 22/an estimated 26 

greg0ire:
499 to 398 errors

`vendor/bin/phpcs --report=summary` now fits on one screen !


beberlei (approved):
getting into the more juicy non tests files here, but it all looks good :)

PR #8521, opened 2021-03-01
 CS batch 23/an estimated 26 

greg0ire:
From 398 to 300 errors in 31 files

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -112,7 +112,7 @@ abstract class AbstractQuery
     /**
      * The map of query hints.
      *
-     * @var array
+     * @psalm-var array<string, mixed>
      */
     protected $_hints = [];
 
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -79,14 +79,14 @@ abstract class AbstractHydrator
     /**
      * Local ClassMetadata cache to avoid going to the EntityManager all the time.
      *
-     * @var array
+     * @var array<string, ClassMetadata>
      */
     protected $_metadataCache = [];
 
     /**
      * The cache used during row-by-row hydration.
      *
-     * @var array
+     * @var array<string, mixed[]|null>
      */
     protected $_cache = [];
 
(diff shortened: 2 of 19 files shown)


lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -100,7 +100,7 @@ abstract class AbstractHydrator
     /**
      * The query hints.
      *
-     * @var array
```
beberlei, line 103: This must be `array<string, mixed>` here same as `AbstractQuery` i suppose?

lib/Doctrine/ORM/Query.php
```diff
@@ -756,10 +756,8 @@ public function getLockMode()
 
     /**
      * Generate a cache id for the query cache - reusing the Result-Cache-Id generator.
-     *
-     * @return string
      */
-    protected function _getQueryCacheId()
+    protected function getQueryCacheId(): string
```
beberlei, line 760: this is fine because `Query` is final

beberlei (changes requested):

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -112,7 +112,7 @@ abstract class AbstractQuery
     /**
      * The map of query hints.
      *
-     * @var array
+     * @psalm-var array<string, mixed>
```
simonberger, line 115: What's the reason to use psalm-var instead of var for this kind of notation here and on several other places?

beberlei (approved):

lib/Doctrine/ORM/AbstractQuery.php
simonberger: To precise my question. I know not every IDE understands this syntax but the most used one PHPStorm does meanswhile.
Also I saw that in some changes of this PR the syntax is also used without the specific psalm-prefix.

SenseException (approved):

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: I'm myself not super interested in this since I don't use PHPStorm, so I try not to break auto-completion for people that do by using the prefix. I think it might be easier to search and replace `psalm-var` with `var` later on than the other way around.

>  Also I saw that in some changes of this PR the syntax is also used without the specific psalm-prefix.

Feel free to make a PR getting this exactly right once I have finished the batches.

Issue #8345, opened 2020-11-23
Add possibility to use fields instead of column for unique constraint


alexander-schranz:
Currently when using a unique constraint like:

```xml
<unique-constraints>
    <unique-constraint columns="locale,product_key" />
</unique-constraints>
```

you are forced to define the `column` name for the fields e.g.:

```xml
<field name="productKey" column="product_key" type="string" nullable="true" unique="true"/>
```

if you don't do this and somebody configure another `naming_strategy` for your entity it will not longer work as the column  does then not exist.

It would be a great feature if you could use unique-constraints based on field name and doctrine orm will read the column name from the configured field e.g.:

```xml
<unique-constraints>
    <unique-constraint fields="locale,productKey" />
</unique-constraints>

<!-- no column is needed here now -->
<field name="productKey" type="string" nullable="true" unique="true"/>
```

What do you think about this?

beberlei:
This is a great idea, there are some caviats. 

- What about associations? Is every join column added? That would make sense.
- Unique constraints need to be parsed after column/fields so that mapping info is available on CöassMetsdataInfo. This is not the case for all mapping drivers right now

beberlei:
The logic of resolving Unique Constraint (and Index as well) fields to columns would need to be done in `SchemaTool`, because access to `NamingStrategy` is needed.

Lustmored:
I took a shot at this and results are in #8523 - feedback is very welcome :)

beberlei:
Fixed in #8629

alexander-schranz:
@beberlei @Lustmored  This is really great to hear! Thank you for implementing this 🙏

PR #8523, opened 2021-03-04
Add possibility to use fields instead of column for unique constraint and indexes (#8345)

Lustmored:
This PR implements feature requested in #8345. It handles all existing drivers (probably upcoming Annotation will need a little change if it gets merged).

I based my work on last comment by @beberlei - it translates fields and columns to indexes only in `SchemaTool::getSchemaFromMetadata`. I haven't found the need to translate them beforehand, but maybe I've missed something.

I have adapted metadata tests to cover all cases when reading entities and added simple test for translating fields to columns.

In my implementation one can use both `fields` and `columns` at the same time and they will be merged.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -30,6 +30,7 @@ use Doctrine\DBAL\Schema\Visitor\DropSchemaSqlCollector;
 use Doctrine\DBAL\Schema\Visitor\RemoveNamespacedAssets;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\ORM\Mapping\QuoteStrategy;
 use Doctrine\ORM\ORMException;
 use Doctrine\ORM\Tools\Event\GenerateSchemaEventArgs;
@@ -148,6 +149,19 @@ class SchemaTool
     private function getIndexColumns($class, array $indexData)
     {
         $columns = [];
+
+        if (
+            isset($indexData['columns'], $indexData['fields'])
+            || (
+                ! isset($indexData['columns'])
+                && ! isset($indexData['fields'])
+            )
+        ) {
+            throw MappingException::invalidIndexConfiguration(
+                $class,
+                $indexData['name'] ?? 'unnamed'
+            );
+        }
 
         if (isset($indexData['columns'])) {
             $columns = $indexData['columns'];
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -138,6 +138,37 @@ class SchemaTool
     }
 
     /**
+     * Resolves fields in index mapping to column names
+     *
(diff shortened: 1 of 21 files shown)


lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -138,6 +138,37 @@ private function processingNotRequired($class, array $processedClasses)
...
+     * @param mixed[]       $indexData index or unique constraint data
+     *
+     * @return string[] Column names from combined fields and columns mappings
+     */
+    private function getIndexColumns($class, array $indexData)
+    {
+        $columns = [];
+
+        if (isset($indexData['columns'])) {
+            $columns = $indexData['columns'];
```
beberlei, line 167: I think its not a good idea to merge them, because in case of many the order of columns is significant for how a B-Tree index is created in the database. 

The only solution where we could support mixing them would be to have a `@IndexField` and `@IndexColumn` annotation and in xml nested children `<index-field name="foo"><index-column name="bar" />`. This is a bit overkill for this problem, so I would rather that we throw an exception if you define both instead of merging them.

What do you think?

beberlei, line 167: I think its not a good idea to merge them, because in case of many the order of columns is significant for how a B-Tree index is created in the database. 

The only solution where we could support mixing them would be to have a `@IndexField` and `@IndexColumn` annotation and in xml nested children `<index-field name="foo"><index-column name="bar" />`. This is a bit overkill for this problem, so I would rather that we throw an exception if you define both instead of merging them.

What do you think?

beberlei (changes requested):
I am unsure if merging is a good behavior, see inline comment.

Lustmored:
> I am unsure if merging is a good behavior, see inline comment.

Thank you for the review. I see the problem and agree with your point. My initial idea was to allow unmapped columns to be added to indexes, but in such corner case user would probably create unmapped index also, so this point seems invalid.

There are a few things here that will need refactor. I will do them as noted below and come back with updated PR unless there are objections:

- when to throw exception: I believe exceptions should be thrown as soon as wrong behavior is found. In this case it'd be in drivers, where it's very simple to determine that both attributes are set. I would also add exception in `SchemaTool` to future-proof the feature (if some driver in the future would allow setting both of them it should still throw)
- In XML schema as far as I know it's not possible to mark "one of 2 elements are required", therefore I would change schema to make both fields optional
- Another question is whether drivers should throw when neither `columns` nor `fields` are specified? I believe they should, as I see no use case for annotating indexes without specified columns.

lib/Doctrine/ORM/Tools/SchemaTool.php
Lustmored: I have adapted code to enforce exactly one of fields/columns and extended tests to cover all cases I could come up with.

Lustmored: I have adapted code to enforce exactly one of fields/columns and extended tests to cover all cases I could come up with.

Lustmored:
I have rebased onto 2.9 and added implementation to new attribute driver. I am not a fan of code duplication in metadata drivers when checking fields/columns configuration, but wasn't sure how to mitigate it. Any suggestions are welcome :)

PR #8524, opened 2021-03-06
[GH-8327] Deprecate EntityManagerHelper for a provider abstraction.

### LINKED ISSUE #8327: cli-config.php compatibility with DBAL
Since DBAL 2.11.0 and https://github.com/doctrine/dbal/pull/3956, it is now deprecated to use HelperSet in `cli-config.php`. This will become impossible in DBAL 3, following https://github.com/doctrine/dbal/pull/4059.

This means that, in order to keep sharing a single config file for both packages, this package should also accept the new `\Doctrine\DBAL\Tools\Console\ConnectionProvider`. It could actually be a new interface along the lines of:

```php
namespace \Doctrine\ORM\Tools\Console;

interface EntityManagerProvider
    public function getEntityManager(string $name) : EntityManagerInterface;
}
```

I suppose the support of HelperSet could be drop entirely in the next major version too. So long we can use a `HelperSetManagerProvider` that extends the `EntityManagerProvider`.

All commands must be adjusted to access the entity manager from a provider instead of the HelperSet. Deprecation messages when using HelperSet must be thrown appropriately.

See related https://github.com/doctrine/migrations/issues/1070

beberlei:
Deprecates the `EntityManagerHelper` in favor of a new `EntityManagerProvider` interface that comes with two implementation.

- A BC abstraction wrapping the Console HelperSet `HelpersetManagerProvider`
- a new `SingleManagerProvider` that provides access to a single entity manager.

In addition an implementation for the DBAL `ConnectionProvider` is provided that is based on any `EntityManagerProvider` and calling `getConnection()` on the EntityManager.

It is expected for the DoctrineBundle to create a new `EntityManagerProvider` that provides access to their multiple ORM entity managers via the Symfony DI Container.

Two deprecation messages are added:
- When the `HelperSetManagerProvider` must be used from within `ConsoleRunner`, because a `HelperSet` is passed.
- When an ORM command is used without explicitly passing the `EntityManagerProvider` as a dependency.

Todos:
- [x] Sort out update of DBAL and 7.2 dependency, waiting for https://github.com/doctrine/dbal/pull/4386
- [x] Update tests to pass provider

Fixes #8327


lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/SingleManagerProvider.php
```diff
@@ -0,0 +1,26 @@
...
+
+    public function __construct(EntityManagerInterface $entityManager)
+    {
+        $this->entityManager = $entityManager;
+    }
+
+    public function getManager(string $name = 'default'): EntityManagerInterface
+    {
+        if ($name !== 'default') {
+            throw UnknownManagerException::unknownManager($name, ['default']);
```
greg0ire, line 21: Is it possible that people define several managers but end up using `SingleManagerProvider`? If yes, that exception will be confusing I think.

tests/Doctrine/Tests/ORM/Tools/Console/ConsoleRunnerTest.php
```diff
@@ -60,4 +66,12 @@ public function testCreateApplicationShouldAppendGivenCommands(): void
 
         self::assertTrue($app->has($command));
     }
+
+    public function testCreateApplicationWithProvider(): void
+    {
+        $provider = $this->getMockBuilder(EntityManagerProvider::class)->getMock();
```
greg0ire:
```suggestion
        $provider = $this->createMock(EntityManagerProvider::class);
```

lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/SingleManagerProvider.php
greg0ire: Also, let's test this?

beberlei: No the `SingleManagerProvider` is for non Symfony context, where usually you have only one and you don't care about any of this. :)

lib/Doctrine/ORM/Tools/Console/Command/AbstractEntityManagerCommand.php
```diff
@@ -0,0 +1,39 @@
...
+    private $entityManagerProvider;
+
+    public function __construct(?EntityManagerProvider $entityManagerProvider = null)
+    {
+        parent::__construct();
+
+        $this->entityManagerProvider = $entityManagerProvider;
+    }
+
+    protected function getEntityManager(InputInterface $input): EntityManagerInterface
```
greg0ire:
```suggestion
    final protected function getEntityManager(InputInterface $input): EntityManagerInterface
```

lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/ConnectionFromManagerProvider.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Doctrine\ORM\Tools\Console\EntityManagerProvider;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Tools\Console\ConnectionProvider;
+use Doctrine\ORM\Tools\Console\EntityManagerProvider;
+
+class ConnectionFromManagerProvider implements ConnectionProvider
```
greg0ire:
```suggestion
final class ConnectionFromManagerProvider implements ConnectionProvider
```

lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/HelperSetManagerProvider.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Doctrine\Deprecations\Deprecation;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Tools\Console\EntityManagerProvider;
+use Doctrine\ORM\Tools\Console\Helper\EntityManagerHelper;
+use Symfony\Component\Console\Helper\HelperSet;
+
+use function assert;
+
+class HelperSetManagerProvider implements EntityManagerProvider
```
greg0ire:
```suggestion
final class HelperSetManagerProvider implements EntityManagerProvider
```

lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/SingleManagerProvider.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Doctrine\ORM\Tools\Console\EntityManagerProvider;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Tools\Console\EntityManagerProvider;
+
+class SingleManagerProvider implements EntityManagerProvider
```
greg0ire:
```suggestion
final class SingleManagerProvider implements EntityManagerProvider
```

lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/UnknownManagerException.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Doctrine\ORM\Tools\Console\EntityManagerProvider;
+
+use OutOfBoundsException;
+
+use function implode;
+use function sprintf;
+
+class UnknownManagerException extends OutOfBoundsException
```
greg0ire:
```suggestion
final class UnknownManagerException extends OutOfBoundsException
```

greg0ire (dismissed):
LGTM. The code suggestions I added are just that: suggestions

dominikzogg:
@beberlei adding `em` option probably breaks every glue code out there similar to the doctrine bundle. This change should be in 3.0 not 2.9

beberlei:
@dominikzogg we speifically added code to prevent breaking in doctrine Bundle

beberlei:
https://github.com/doctrine/DoctrineBundle/commit/d6b3c37804539a24ba8a7d647a6144cab2f13242

dominikzogg:
@beberlei as mentioned: there is code outside the symfony ecosystem out there

https://github.com/chubbyphp/chubbyphp-laminas-config-doctrine/pull/3/files and i am not even done yet => will lead to a major release on my side and when orm 3.0 gets released another major cause i am sure there will be (expected) bc's

beberlei:
Adding *any* parameter could have caused this, others might use `entity-manager` instead of `em`. Not sure how this could have been prevented, we won't change this now. Sorry to cause extra work on your end.

dominikzogg:
> Not sure how this could have been prevented

do it in 3.0, so nobody get forced to update code immediately

beberlei:
That ship has sailed :)

PR #8529, opened 2021-03-09
 CS batch 24/26 🤞

greg0ire:
From 300 to 200 errors in 26 files

### DIFF
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -36,6 +36,7 @@ use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
 use Doctrine\ORM\Mapping\EntityListenerResolver;
 use Doctrine\ORM\Mapping\NamingStrategy;
 use Doctrine\ORM\Mapping\QuoteStrategy;
+use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\ORM\Repository\DefaultRepositoryFactory;
 use Doctrine\ORM\Repository\RepositoryFactory;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
@@ -149,10 +150,11 @@ class Configuration extends \Doctrine\DBAL\Configuration
      * Adds a new default annotation driver with a correctly configured annotation reader. If $useSimpleAnnotationReader
      * is true, the notation `@Entity` will work, otherwise, the notation `@ORM\Entity` will be supported.
      *
-     * @param array $paths
-     * @param bool  $useSimpleAnnotationReader
+     * @param bool $useSimpleAnnotationReader
      *
      * @return AnnotationDriver
+     *
+     * @psalm-param string|list<string> $paths
      */
     public function newDefaultAnnotationDriver($paths = [], $useSimpleAnnotationReader = true)
     {
@@ -207,9 +209,9 @@ class Configuration extends \Doctrine\DBAL\Configuration
     /**
      * Sets the entity alias map.
      *
-     * @param array $entityNamespaces
-     *
      * @return void
+     *
+     * @psalm-param array<string, string> $entityNamespaces
      */
     public function setEntityNamespaces(array $entityNamespaces)
     {
@@ -219,7 +221,7 @@ class Configuration extends \Doctrine\DBAL\Configuration
     /**
      * Retrieves the list of registered entity namespace aliases.
      *
-     * @return array
+     * @psalm-return array<string, string>
      */
     public function getEntityNamespaces()
     {
@@ -348,10 +350,13 @@ class Configuration extends \Doctrine\DBAL\Configuration
      *
      * @param string $name The name of the query.
      *
-     * @return array A tuple with the first element being the SQL string and the second
-     *               element being the ResultSetMapping.
-     *
      * @throws ORMException
+     *
+     * @psalm-return array{string, ResultSetMapping} A tuple with the first
+     *                                               element being the SQL
+     *                                               string and the second
+     *                                               element being the
+     *                                               ResultSetMapping.
      */
     public function getNamedNativeQuery($name)
     {
@@ -439,9 +444,10 @@ class Configuration extends \Doctrine\DBAL\Configuration
      *
      * Any previously added string functions are discarded.
      *
-     * @param array $functions The map of custom DQL string functions.
-     *
      * @return void
+     *
+     * @psalm-param array<string, class-string> $functions The map of custom
+     *                                                     DQL string functions.
      */
     public function setCustomStringFunctions(array $functions)
     {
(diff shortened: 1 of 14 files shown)


SenseException (approved):
Nice improvements of already existing psalm annotations :+1:

lib/Doctrine/ORM/Configuration.php
```diff
@@ -561,7 +568,7 @@ public function setCustomDatetimeFunctions(array $functions)
     /**
      * Sets the custom hydrator modes in one pass.
      *
-     * @param array $modes An array of ($modeName => $hydrator).
+     * @param array<string, class-string> $modes An array of ($modeName => $hydrator).
```
vladysta, line 571:
```suggestion
     * @param array<\Doctrine\ORM\AbstractQuery::HYDRATE_*, class-string> $modes An array of ($modeName => $hydrator).
```

It's mostly `\Doctrine\ORM\AbstractQuery::HYDRATE_*` constants, BUT yes, you can set custom once. So I don't know if I'm fully right

lib/Doctrine/ORM/Configuration.php
```diff
@@ -816,7 +824,7 @@ public function getSecondLevelCacheConfiguration()
     /**
      * Returns query hints, which will be applied to every query in application
      *
-     * @return array
+     * @psalm-return array<string, mixed>
```
vladysta, line 827:
```suggestion
     * @psalm-return array<string, \Doctrine\ORM\Query::HINT_*>
```

Same as previous but for `\Doctrine\ORM\Query::HINT_*`

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -297,28 +297,28 @@ class ClassMetadataInfo implements ClassMetadata
...
-     * @var array
+     * @psalm-var list<class-string>
      */
     public $subClasses = [];
 
     /**
      * READ-ONLY: The names of all embedded classes based on properties.
      *
-     * @var array
+     * @psalm-var array<string, mixed[]>
```
vladysta, line 314:
```suggestion
     * @psalm-var array<string, array{class: class-string, columnPrefix: string|false, declaredField: ?string, originalField: ?string, inherited?: class-string, declared?: class-string}>
```
inferred from `\Doctrine\ORM\Mapping\ClassMetadataFactory::addInheritedEmbeddedClasses`
inferred from `\Doctrine\ORM\Mapping\ClassMetadataInfo::mapEmbedded`

----------------

This looks kind of ugly, and probably need to be refactored in ORM3

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -297,28 +297,28 @@ class ClassMetadataInfo implements ClassMetadata
...
      */
     public $embeddedClasses = [];
 
     /**
      * READ-ONLY: The named queries allowed to be called directly from Repository.
      *
-     * @var array
+     * @psalm-var array<string, array<string, mixed>>
```
vladysta, line 321:
```suggestion
     * @psalm-var array<string, array{name: string, query: string, dql: string}>
```

inferred from `\Doctrine\ORM\Mapping\ClassMetadataInfo::getNamedQuery`
inferred from `\Doctrine\ORM\Mapping\ClassMetadataInfo::addNamedQuery`

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -335,7 +335,7 @@ class ClassMetadataInfo implements ClassMetadata
      * )
      * </pre>
      *
-     * @var array
+     * @psalm-var array<string, array<string, mixed>>
```
vladysta, line 338:
```suggestion
     * @psalm-var array<string, array{name: string, query: string, isSelfClass: bool, resultSetMapping?: , resultClass?: class-string}>
```

as said in the above comment and also
inferred from `\Doctrine\ORM\Mapping\ClassMetadataFactory::addInheritedNamedNativeQueries` also I was not 100% sure about the `resultSetMapping` key of array 

the http://doctrine-project.org/schemas/orm/doctrine-mapping says that 
```xml
    <xs:attribute name="result-set-mapping" type="xs:string" />
```
the type is `string`
but if you look to this answer on stack overflow https://stackoverflow.com/a/37715649
it has different type - `array`

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -351,7 +351,11 @@ class ClassMetadataInfo implements ClassMetadata
      * )
      * </pre>
      *
-     * @var array
+     * @psalm-var array<string, array{
+     *                name: string,
+     *                entities: mixed[],
```
vladysta, line 356: I'm confused here ( what types it has)

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -468,7 +471,7 @@ class ClassMetadataInfo implements ClassMetadata
      * READ-ONLY: The definition of the discriminator column used in JOINED and SINGLE_TABLE
      * inheritance mappings.
      *
-     * @var array
+     * @psalm-var array<string, mixed>
```
vladysta, line 474:
```suggestion
     * @psalm-var array{name: string, type: string|class-string, length: int, columnDefinition: ?string}
```
inferred from `\Doctrine\ORM\Mapping\ClassMetadataInfo::setDiscriminatorColumn`

vv12131415 (approved):
I did only 7, but I think it's already enough

greg0ire:
@vladyslavstartsev thanks for the suggestions, but I think that should be a separate PR: mine is just about fixing cs issues, and I don't want the review to become hard because of all the additional checks we would have to perform. In fact, I think that when I'm done, we can work on raising Psalm and Phpstan level. These PRs only provide advanced types like you propose when it's really straightforward to check.

Maybe we should even try raising the level before adding more type annotations, so that the type annotations are actually checked by something?

vv12131415:
@greg0ire as you said, you are doing things perfectly. Just go step-by-step. And after CS fixes are finally done, ping if I can help you with rising Psalm and Phpstan level (not really sure, because I only have couple of hours in a month to spend). 

Thank you for doing this work and also for writing feedback comment.

lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -183,14 +181,12 @@ public function findAll()
...
-     * @param int|null   $limit
-     * @param int|null   $offset
-     *
-     * @return array The objects.
+     * @param int|null $limit
+     * @param int|null $offset
      *
-     * @psalm-return list<T>
+     * @psalm-param array<string, mixed> $criteria
+     * @psalm-param list<string>|null $orderBy
```
annuh, line 188: Not sure if I should create new issue for this, but after v2.8.3, Psalm throws a warning for this function in our codebase:

```
ERROR: InvalidArgument - src/SomeClass.php:25:13 - Argument 2 of Doctrine\ORM\EntityRepository::findBy expects list<string>\|null, array{id: "ASC"} provided (see https://psalm.dev/004)
12 | ['id' => 'ASC'],
```

I have a feeling this updated `@psalm-param` is not correct, because:
https://github.com/doctrine/orm/blob/2.8.x/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1145
```
@psalm-param array<string, string> $orderBy
```


EDIT: nevermind, there is already an open PR to fix this 👏  https://github.com/doctrine/orm/pull/8586

PR #8532, opened 2021-03-11
All CI runs are using the sqlite fallback connection instead of the expected driver

acoulton:
While writing a test for an unrelated issue, I noticed that the unit tests are not loading the expected database connection details. Regardless of the config file provided, the `TestUtil` class is providing the fallback sqlite driver. 

As a result a) none of the code is being tested against other platforms and b) tests that rely on platform-specific features (e.g. pessimistic locking) are being skipped on all builds.

It looks like this has been the case ever since the migration from Travis CI. Therefore some of the consistently-skipped tests are also not valid with phpunit 9 - e.g. `Doctrine\Tests\ORM\Functional\Locking\LockTest` has two tests that fail due to using `->assertContains` with a string argument, which need to be upgraded to `->assertStringContainsString`.

The root cause is that when the driver-specific phpunit.xml files were updated for github actions, the various `tmpdb_type` / `tmpdb_username` etc with the config for the privileged temporary connection were removed. It looks like this [might be valid for doctrine/dbal because it falls back to using the main db params](https://github.com/doctrine/dbal/blob/3.1.x/tests/TestUtil.php#L158) however the [TestUtil class in doctrine/orm does not](https://github.com/doctrine/orm/blob/2.8.x/tests/Doctrine/Tests/TestUtil.php#L162) and [it has a much stricter list of required parameters before it will use the specified driver](https://github.com/doctrine/orm/blob/2.8.x/tests/Doctrine/Tests/TestUtil.php#L162).

* [x] Update test configuration to prove the issue and hard-fail if the job is not using the expected driver
* [x] *Either* reinstate the extra variables in the phpunit.xml files *or* update TestUtil to match the behaviour of the doctrine/dbal version
* [x] Identify tests that fail once the actual drivers are being used and fix (upgrade to phpunit 9) as required.

I can finish this off, but could you advise whether you'd prefer me to add the missing vars to the phpunit.xml or update TestUtil to work as expected with the current configs?

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7941Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7941Test.php
@@ -14,12 +14,12 @@ use function strlen;
 final class GH7941Test extends OrmFunctionalTestCase
 {
     private const PRODUCTS = [
-        ['name' => 'Test 1', 'price' => '100', 'square_root' => '/^10(\.0+)?$/'],
-        ['name' => 'Test 2', 'price' => '100', 'square_root' => '/^10(\.0+)?$/'],
-        ['name' => 'Test 3', 'price' => '100', 'square_root' => '/^10(\.0+)?$/'],
-        ['name' => 'Test 4', 'price' => '25', 'square_root' => '/^5(\.0+)?$/'],
-        ['name' => 'Test 5', 'price' => '25', 'square_root' => '/^5(\.0+)?$/'],
-        ['name' => 'Test 6', 'price' => '-25', 'square_root' => '/^5(\.0+)?$/'],
+        ['name' => 'Test 1', 'price' => '100', 'square_root' => 10],
+        ['name' => 'Test 2', 'price' => '100', 'square_root' => 10],
+        ['name' => 'Test 3', 'price' => '100', 'square_root' => 10],
+        ['name' => 'Test 4', 'price' => '25', 'square_root' => 5],
+        ['name' => 'Test 5', 'price' => '25', 'square_root' => 5],
+        ['name' => 'Test 6', 'price' => '-25', 'square_root' => 5],
     ];
 
     protected function setUp(): void
@@ -51,7 +51,11 @@ final class GH7941Test extends OrmFunctionalTestCase
 
         self::assertSame(6, $result['count']);
         self::assertSame('325', $result['sales']);
-        self::assertRegExp('/^54\.16+7$/', $result['average']);
+
+        // Driver return type and precision is determined by the underlying php extension, most seem to return a string.
+        // pdo_mysql and mysqli both currently return '54.1667' so this is the maximum precision we can assert.
+        // See https://github.com/doctrine/orm/pull/8532#pullrequestreview-610037209
+        self::assertEqualsWithDelta(54.1667, $result['average'], 0.0001);
 
         $query = $this->_em->createQuery(
             'SELECT
@@ -66,7 +70,11 @@ final class GH7941Test extends OrmFunctionalTestCase
 
             self::assertSame(ltrim($product['price'], '-'), $item['absolute']);
             self::assertSame(strlen($product['name']), $item['length']);
-            self::assertRegExp($product['square_root'], $item['square_root']);
+
+            // Driver return types for the `square_root` column are inconsistent depending on the underlying
+            // database driver. Most return string (though some '10' and some '10.000000000000000') but at least mysqli
+            // returns a float.
+            self::assertEqualsWithDelta($product['square_root'], $item['square_root'], 0.00000001);
         }
     }
 }
--- a/tests/Doctrine/Tests/TestUtil.php
+++ b/tests/Doctrine/Tests/TestUtil.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests;
 
+use Doctrine\Common\EventSubscriber;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
 use UnexpectedValueException;
@@ -30,18 +31,16 @@ class TestUtil
      * Gets a <b>real</b> database connection using the following parameters
      * of the $GLOBALS array:
      *
-     * 'db_type' : The name of the Doctrine DBAL database driver to use.
-     * 'db_username' : The username to use for connecting.
+     * 'db_driver' : The name of the Doctrine DBAL database driver to use.
+     * 'db_user' : The username to use for connecting.
      * 'db_password' : The password to use for connecting.
      * 'db_host' : The hostname of the database to connect to.
      * 'db_server' : The server name of the database to connect to
      *               (optional, some vendors allow multiple server instances with different names on the same host).
-     * 'db_name' : The name of the database to connect to.
+     * 'db_dbname' : The name of the database to connect to.
      * 'db_port' : The port of the database to connect to.
      *
-     * Usually these variables of the $GLOBALS array are filled by PHPUnit based
-     * on an XML configuration file. If no such parameters exist, an SQLite
-     * in-memory database is used.
+     * These variables of the $GLOBALS array are filled by PHPUnit based on an XML configuration file.
      *
      * IMPORTANT:
      * 1) Each invocation of this method returns a NEW database connection.
@@ -51,87 +50,79 @@ class TestUtil
      */
     public static function getConnection(): Connection
     {
-        $params = self::getConnectionParams();
-        $conn   = DriverManager::getConnection($params);
-        // Note, writes direct to STDERR to prevent phpunit detecting output - otherwise this would cause either an
-        // "unexpected output" warning or a failure on the first test case to call this method.
-        fwrite(
-            STDERR,
-            sprintf(
-                "\nUsing DB driver %s\n",
-                get_class($conn->getDriver())
-            )
-        );
+        if (! self::$initialized) {
+            self::initializeDatabase();
+            self::$initialized = true;
+        }
+
+        $conn = DriverManager::getConnection(self::getTestConnectionParameters());
 
         self::addDbEventSubscribers($conn);
 
         return $conn;
     }
 
-    public static function getTempConnection(): Connection
+    public static function getPrivilegedConnection(): Connection
     {
-        return DriverManager::getConnection(self::getParamsForTemporaryConnection());
+        return DriverManager::getConnection(self::getPrivilegedConnectionParameters());
     }
 
-    /**
-     * @psalm-return array<string, mixed>
-     */
-    private static function getConnectionParams()
+    private static function initializeDatabase(): void
     {
-        $realDbParams = self::getParamsForMainConnection();
+        $testConnParams = self::getTestConnectionParameters();
+        $privConnParams = self::getPrivilegedConnectionParameters();
 
-        if (! self::$initialized) {
-            $tmpDbParams = self::getParamsForTemporaryConnection();
+        $testConn = DriverManager::getConnection($testConnParams);
 
-            $realConn = DriverManager::getConnection($realDbParams);
+        // Note, writes direct to STDERR to prevent phpunit detecting output - otherwise this would cause either an
+        // "unexpected output" warning or a failure on the first test case to call this method.
+        fwrite(STDERR, sprintf("\nUsing DB driver %s\n", get_class($testConn->getDriver())));
 
-            // Connect to tmpdb in order to drop and create the real test db.
-            $tmpConn = DriverManager::getConnection($tmpDbParams);
+        // Connect as a privileged user to create and drop the test database.
+        $privConn = DriverManager::getConnection($privConnParams);
 
-            $platform = $tmpConn->getDatabasePlatform();
+        $platform = $privConn->getDatabasePlatform();
 
-            if ($platform->supportsCreateDropDatabase()) {
-                $dbname = $realConn->getDatabase();
-                $realConn->close();
+        if ($platform->supportsCreateDropDatabase()) {
+            $dbname = $testConn->getDatabase();
+            $testConn->close();
 
-                $tmpConn->getSchemaManager()->dropAndCreateDatabase($dbname);
+            $privConn->getSchemaManager()->dropAndCreateDatabase($dbname);
 
-                $tmpConn->close();
-            } else {
-                $sm = $realConn->getSchemaManager();
+            $privConn->close();
+        } else {
+            $sm = $testConn->getSchemaManager();
 
-                $schema = $sm->createSchema();
-                $stmts  = $schema->toDropSql($realConn->getDatabasePlatform());
+            $schema = $sm->createSchema();
+            $stmts  = $schema->toDropSql($testConn->getDatabasePlatform());
 
-                foreach ($stmts as $stmt) {
-                    $realConn->exec($stmt);
-                }
+            foreach ($stmts as $stmt) {
+                $testConn->exec($stmt);
             }
-
-            self::$initialized = true;
         }
-
-        return $realDbParams;
     }
 
     private static function addDbEventSubscribers(Connection $conn): void
     {
-        if (isset($GLOBALS['db_event_subscribers'])) {
-            $evm = $conn->getEventManager();
-            foreach (explode(',', $GLOBALS['db_event_subscribers']) as $subscriberClass) {
-                $subscriberInstance = new $subscriberClass();
-                $evm->addEventSubscriber($subscriberInstance);
-            }
+        if (! isset($GLOBALS['db_event_subscribers'])) {
+            return;
+        }
+
+        $evm = $conn->getEventManager();
+        /** @psalm-var class-string<EventSubscriber> $subscriberClass */
+        foreach (explode(',', $GLOBALS['db_event_subscribers']) as $subscriberClass) {
+            $subscriberInstance = new $subscriberClass();
+            $evm->addEventSubscriber($subscriberInstance);
         }
     }
 
     /**
      * @psalm-return array<string, mixed>
      */
-    private static function getParamsForTemporaryConnection()
+    private static function getPrivilegedConnectionParameters(): array
     {
-        if (isset($GLOBALS['tmpdb_driver'])) {
-            return self::mapConnectionParameters($GLOBALS, 'tmpdb_');
+        if (isset($GLOBALS['privileged_db_driver'])) {
+            return self::mapConnectionParameters($GLOBALS, 'privileged_db_');
         }
 
         $parameters = self::mapConnectionParameters($GLOBALS, 'db_');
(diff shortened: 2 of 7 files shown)


acoulton:
hmm - apologies my second commit here doesn't seem to be setting / reading the expected DB properly in the github actions environment so it's not failing the build as expected but you can see from the build output logs that it's definitely not loading the right driver. I need to jump onto something else just now, I'll come back in a bit and debug why that commit isn't failing properly.

acoulton:
OK - I've now got this far enough to identify all the issues.

The more I think about it the more I think it's better to move towards structuring the config the way doctrine/dbal does so that the github actions and phpunit configs are equivalent (as they are now). So I've rejigged the logic and config key naming in TestUtil to work the same way. This will mean that if developers have a local phpunit.xml that was working with the older code, they'll have to rename some variables to get their connection working again. Obviously doesn't affect end-users at all.

There are still some naming differences between TestUtil here and in dbal, e.g. dbal has `getPrivilegedConnectionParameters` and `getTestConnectionParameters` where orm has `getParamsForTemporaryConnection` and `getParamsForMainConnection`.  These are all private / internal that I can see, not sure if there's any value in making the naming more consistent between the projects?

There were 3 very minor phpunit failures that I've fixed in https://github.com/doctrine/orm/pull/8532/commits/7388aa7417a6aa79255e5c6adbe062f3dca7a854.

The only thing I'm stuck on is the GH7941 test which I'll annotate separately.

Obviously I will deal with coding standards, squashing up commits sensibly etc, just wanted to get a steer first.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7941Test.php
```diff
@@ -64,6 +65,18 @@ public function typesShouldBeConvertedForDQLFunctions(): void
...
+            // !!!!! TEMPORARY HACK
+            // Just to make sure this is the only remaining issue to solve as I'm not sure if the issue here is the
+            // testcase or the inconsistent return type.
+            // mysqli returns the `square_root` column as a float but interestingly it doesn't return the average
+            // as a float...?
+            // All other drivers return the sqrt value as a string
+            if ($this->getEntityManager()->getConnection()->getDriver() instanceof MysqliDriver) {
+                $this->assertIsFloat($item['square_root']);
+                $this->addWarning('Manually casting inconsistent `float` value to string for test to pass');
+                $item['square_root'] = (string) $item['square_root'];
```
acoulton: This test was failing with the mysqli driver because assertRegExp now strictly requires a string value.

All other drivers are returning a string for the `square_root` column, but mysqli is returning a float. Oddly this is the only affected column, the others (including the result of AVG in the query above) are all strings.

The test was introduced in #7941 and I can't spot anything since that would be causing it to fail. My hunch is it was always getting a float on mysqli but it wasn't a problem due to the soft typing in older phpunit.

I've had a bit of a dig in internals and as far as I can see:

* SqlWalker::walkSelectExpression is happily calling ResultSetMapping::addScalarResult with a `string` type for all the columns in these queries
* AbstractHydrator::gatherRowData is correctly getting a `Doctrine\DBAL\Types\StringType` type object for the square_root column
* `StringType::convertToPhpValue` falls back to the parent `Type::convertToPhpValue` which just returns the incoming value
* With the mysqli extension, `->convertToPhpValue()` is called with a `double`, but with pdo_mysql (and all other drivers) it's called with a `string`.

So unless I'm missing something I think it's just that the `mysqli` extension returns a `double` for columns that call the `SQRT` function.

If that's right then there are two options:

* If the return type should be guaranteed to be a string across all drivers then I guess `StringType->convertToPhpValue` needs to explicitly cast the return value - but that feels like quite a major change that could cause problems elsewhere.
* If it's expected that the return type is inconsistent / undefined then I can just tidy up the fix in the test so that it will accept either float or string. Instead of regexp, I'd be tempted to compare them using `->assertEqualsWithDelta()` to match numerically with a reasonable level of precision. That would probably be more future proof, especially if the future plan is to manually cast these functions to float as discussed in #7941.

How should I proceed?

greg0ire: Since types are defined in the DBAL (so another package) I think it will be easier to go with option 2 (and `->assertEqualsWithDelta()`. @morozov, are you maybe aware of this discrepancy of returned type between extensions?

tests/Doctrine/Tests/TestUtil.php
```diff
@@ -43,7 +53,31 @@ class TestUtil
...
+        fwrite(
+            STDERR,
+            sprintf(
+                "\nUsing DB driver %s (from %s connection params)\n",
+                get_class($conn->getDriver()),
+                $params['is_fallback'] ? 'fallback' : 'specified'
+            )
+        );
+
+        $expectDriver = getenv('EXPECT_DB_DRIVER');
```
greg0ire: Maybe we should throw if the `CI` env var is true and this is not set? That way if we migrate to another CI system, we will have to implement that `EXPECT_DB_DRIVER` variable

greg0ire:
Well… this is embarrassing… thankfully you spotted it before things became too hard to fix, although there are already pretty weird things like this type difference between pdo_mysql and mysqli. Anyway, great job with this!

> I can finish this off, but could you advise whether you'd prefer me to add the missing vars to the phpunit.xml or update TestUtil to work as expected with the current configs?

I'm not sure why we had 2 sets of parameters in the first place, and I'd go with the second solution for the sake of simplicity.

> Successful in 1s — 88.32% (+0.88%) compared to 23dc804 

Nice coverage improvement!

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7941Test.php
morozov: @greg0ire I believe it's a common thing for many of the DB drivers in PHP. IIRC, `mysqli` may [cast](https://www.php.net/manual/en/mysqli.insert-id.php) an integer to a string if there's a risk of it causing integer overflow. So the discrepancy may exist even within a given extension.

acoulton: Thanks @morozov OK, makes sense - I'll update the test accordingly.

tests/Doctrine/Tests/TestUtil.php
acoulton: @greg0ire only thing is the `CI` env var is also just set by github actions so not necessarily guaranteed it would be present on a new CI system...?

The other possibility is to just remove the concept of the `fallback` connection, and require that the phpunit.xml you're using specifies a driver and connection details. The default phpunit.xml.dist (and the `sqlite.xml` used in CI) would just need to have explicit config for the sqlite db.

Since it'd be much harder for that to go wrong, the extra `EXPECT_DB_DRIVER` check might then not be required, which would reduce some duplication / coupling between the workflow files and the phpunit configs? Probably still worth displaying which driver was selected for debugging / visual confirmation if changing CI system.

That does diverge from how dbal do it - don't know how maintainers are working across projects / how important it is to have consistency?

acoulton:
> I'm not sure why we had 2 sets of parameters in the first place, and I'd go with the second solution for the sake of simplicity.

It looks like it's on the assumption you might have a less-privileged user for running the main tests and then a more privileged one to create / drop the databases and run setup etc. I'm not sure how crucial that is. Maybe dbal has some tests that need to run as a less-privileged user to assert permission handling or something and this came from there but isn't really necessary?

I'd thought maybe it was in case there were devs / environments where you were happy to provide limited-permission creds but couldn't give root access to the DB. But AFAICS the `temporary connection` is still actually required for every test run, and it needs root-like permissions to wipe and recreate if you're running more than once, so it looks a bit like complexity for no (remaining) purpose... 

I can't see any case where it would make sense to be using a different driver or speaking to a different database platform though.

Again if you're happy to diverge from dbal it looks like I could just remove the whole `temporary connection` concept and just have a single set of connection parameters. That would further simplify things / reduce the scope for future unexpected behaviour.

greg0ire:
@beberlei @morozov do you maybe have some context about this?

morozov:
The privileged credentials are used to create a test database. The regular credentials should be scoped to the test database to run tests there.

greg0ire:
> The regular credentials should be scoped to the test database to run tests there.

Looks like they are not, since they were apparently used to do that creation. Also, what is the point of this? I know it's good to be close to a real situation, and people will scope restrict credentials in real life to avoid disasters, but I can't imagine what bugs we want to catch here, and there certainly can't be a disaster in the case of the CI. Maybe it's in the case of a local run on the development environment?

morozov:
Another reason is that when the test starts, the test database will likely not exist, so if it's used as a parameter of the only connection, this only connection will fail. So we temporarily connect to some other database or don't use any (if the platform allows) and create the database to which the test connection will be able to connect.

greg0ire:
So the question is: why does it currently work? Currently, we only have one connection, and it specifies the database.

For mysql and mariadb, there is a simple answer, the database is created because we specify its name as a standard env variable of the docker image:

https://github.com/doctrine/orm/blob/93f31d2c33f6dae7309184bbf2b338ae411a3fd9/.github/workflows/continuous-integration.yml#L135

https://github.com/doctrine/orm/blob/93f31d2c33f6dae7309184bbf2b338ae411a3fd9/.github/workflows/continuous-integration.yml#L135

For PostgreSQL, I have no idea why it works. We have one set of credentials: https://github.com/doctrine/orm/blob/93f31d2c33f6dae7309184bbf2b338ae411a3fd9/ci/github/phpunit/pdo_pgsql.xml#L10-L14

And there is no mention on the database name in the github workflow.

UPD: apparently the database is dropped and re-created every time, so having it created with a Docker variable does not explain how it works for Mysql and MariaDB: https://github.com/doctrine/orm/blob/93f31d2c33f6dae7309184bbf2b338ae411a3fd9/tests/Doctrine/Tests/TestUtil.php#L38-L40

Ah I think I have a good explanation! Although we can indeed specify a database, it will not be passed on when creating the temporary connection: https://github.com/doctrine/orm/blob/93f31d2c33f6dae7309184bbf2b338ae411a3fd9/tests/Doctrine/Tests/TestUtil.php#L162-L171

@acoulton IMO you do not need to restore the `tmp_` variables, it's good as is and would only be useful to a developer willing to run the tests on an instance that has other databases that might be precious.

acoulton:
OK, so this is potentially a platform-specific thing, it did fail on the db not existing on an at least one of the builds, possibly postgres, earlier in this branch.

Then I updated TestUtil to the full method from dbal, which does this:

```
    private static function getParamsForTemporaryConnection()
    {
        if (isset($GLOBALS['tmpdb_driver'])) {
            return self::mapConnectionParameters($GLOBALS, 'tmpdb_');
        }

        $parameters = self::mapConnectionParameters($GLOBALS, 'db_');
        unset($parameters['dbname']);

        return $parameters;
    }
```

So basically if you don't provide explicit details for a privileged connection, it just uses the normal connection config but removes the dbname before connecting. That's how postgres is now able to connect before the database exists, but still only uses one set of credentials / config.

In orm, the `public static function getTempConnection()` is never called. The "temporary DB" params are *only ever* used to drop / create the database before the tests.

In dbal, the method is called `public static function getPrivilegedConnection()` and it is called from some of the Oracle tests. However it is *only* the oracle tests that actually provide [different config for the `tmpdb_` values](https://github.com/doctrine/dbal/blob/3.1.x/ci/github/phpunit/pdo_oci.xml#L18). All other drivers in theory could use them but are in practice falling back to use the `db_` values without the `dbname` and running all of the tests with root-level credentials.

acoulton:
@greg0ire sorry my post crossed with your edit.

So yeah, AFAICS the only theoretical use for the `tmpdb_` values would be to run tests locally against a db that contained other things. **However** as it stands you can run the tests with a less privileged-account but you *still* have to provide root permissions in the `tmpdb_` values for most platforms.

That's because TestUtil always attempts to drop and recreate the database at the start of the run - https://github.com/doctrine/orm/blob/2.8.x/tests/Doctrine/Tests/TestUtil.php#L110 - that will ignore a failure on the `DROP DATABASE` command but it will throw if `CREATE DATABASE` fails.

So there's no scenario at the moment where you can *actually* run anything without giving doctrine root-level permissions in one or both of those configs. I can't see a case where:
* You're sensitive about what else is in your database so you don't want doctrine to run the tests as root
* But you don't mind giving doctrine root to run DROP DATABASE / CREATE DATABASE at the start of the test run

In my mind either you don't care about the db and doctrine can be root, or you care and doctrine can't run at all?

acoulton:
I guess what I'm saying is, if this was my project I'd remove the entire `tmpdb_` concept and all the code that supports it from TestUtil. And just use the main `db_` config vars for everything, removing the dbname for the initial connection as now.

If the ambition is that a developer should also be able to run locally against a system they care about it would be better that they provide restricted `db_` creds and a flag to tell doctrine to skip the `DROP / CREATE` step entirely on the basis they will manage creating / resetting the database outside the doctrine test process. But I'm not sure how crucial that is in the world of docker...

Or I can just tidy up what's here and leave `tmpdb_` as optional but not required. Your call.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7941Test.php
acoulton: Done in https://github.com/doctrine/orm/pull/8532/commits/62bf9b83ca59d6552eedbc3582290a8151702eca (to be squashed up). Turns out the actual returned values vary quite a bit, even in the level of precision in the drivers that return strings. I was surprised mysql only gives 4 d.p. for the average...

greg0ire:
> I can't see a case 
> You're sensitive about what else is in your database so you don't want doctrine to run the tests as root
> But you don't mind giving doctrine root to run DROP DATABASE / CREATE DATABASE at the start of the test run


I think it hardly makes sense indeed. The only way I could see it having sense is if you decide to trust the code that does the create/drop because it rarely changes,and don't trust code from tests because it might not be as well vetted, and every new line makes it more prone to error. Or you don't want contributors to be able to write tests that perform admin tasks from a concrete tests. IMO these concerns are not super valid.

My call would be: let's remove it, but it might be just me. WDYT @beberlei @morozov ?

acoulton:
OK, I have squashed and tidied what I have so far, which solves the actual current problems.

The outstanding questions are whether I should:

* Remove the concept of `tmpdb_`
* Remove the silent/automatic fallback to sqlite in favour of always requiring phpunit.xml to specify the driver you want. I think this would avoid the need for the separate `EXPECT_DB_DRIVER` check in CI and be more future-proof. I'd update phpunit.xml.dist so it works as-now "out of the box".

tests/Doctrine/Tests/TestUtil.php
greg0ire: >  only thing is the CI env var is also just set by github actions so not necessarily guaranteed it would be present on a new CI system...?

Isn't it a standard? Because it looks a lot like it is:

https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
https://github.com/circleci/circleci-docs/blob/master/jekyll/_cci1/environment-variables.md
https://www.appveyor.com/docs/environment-variables/

> Since it'd be much harder for that to go wrong

It would be, but I like the idea of checking that kind of expectation a lot. In fact I thought a bit about it some time ago, because I was not fully at ease with the concept of skipped tests: how can we be sure that skipped tests are run on at least one job? Maybe there should be a coverage report for tests themselves and it should always be 100% ? I don't know if that's even possible. And I think that `CI` env var would help us being lenient on dev, and super strict on the CI.

> the extra EXPECT_DB_DRIVER check might then not be required, which would reduce some duplication / coupling between the workflow files and the phpunit configs?

I wonder if it might be possible to reference a matrix variable in the env section, so as not to have that much duplication. I'll try to push something to your branch soon.

> That does diverge from how dbal do it - don't know how maintainers are working across projects / how important it is to have consistency?

I'm not a developer as much as a maintainer in Doctrine, so I can answer that question: to me it is quite important, because many of the contributions I do to Doctrine are org-wide (and can be followed here: https://github.com/orgs/doctrine/projects/7). Having one way of doing things is better for me, and I think it is also better for contributors. Let's draft something and have both @beberlei and @morozov sign off on it.

greg0ire: >  I'll try to push something to your branch soon.

Or rather, I'll try on my fork since I seem to lack the necessary permissions

UPD: as I feared, it does not work

acoulton: > Isn't it a standard? Because it looks a lot like it is:

Ah, I hadn't realised. Sounds good.

> possible to reference a matrix variable in the env section

Yes, I think not possible in the overall `env` (not least as not all the jobs have a matrix.extension value) but should work at job level which is less duplication than I had.

greg0ire (dismissed):

greg0ire:
> IMO these concerns are not super valid.

I've given this more thought, and I think you can compare this to something we do often as developers, which is installing a piece of software like follows:

- `make`
- `sudo make install`

I think the issue is not one of security, we are not assuming malice here, we are just being cautious with privileges, which can sometimes spare some disasters like [this one](https://github.com/ValveSoftware/steam-for-linux/issues/3671).

I would be (bandwagon) fallacious of me to say that this suddenly makes these concerns valid, but we can't pretend this kind of practice is unheard of, quite the contrary. In the case of the CI, we are ourselves not following that best practice since apparently the main credentials are enough, but that mechanism gives contributors the option to follow it while developing. So I'd keep that `tmpdb_`, but maybe rename it to `privileged_` or something like that

> Remove the silent/automatic fallback to sqlite in favour of always requiring phpunit.xml to specify the driver you want. I think this would avoid the need for the separate EXPECT_DB_DRIVER check in CI and be more future-proof. I'd update phpunit.xml.dist so it works as-now "out of the box".

I agree it would be best to remove that fallback mechanism. 👍 I'm not sure about removing the `EXPECT_DB_DRIVER` check, hopefully other maintainers will chime in.

greg0ire:
Hey @acoulton ! I discussed internally with Benjamin and Sergei, and got the answers you needed:
- they agree with you on `EXPECT_DB_DRIVER` being overkill now, let's remove it
- about the privileged creds, let's keep them, they might be necessary if we add an Oracle build in the future (which is totally doable on GA, see the DBAL), or for devs that want to use Oracle
- they agree that the fallback mechanism can go 👍

acoulton:
@greg0ire great - that's all done and tidied up now. 

I thought it made sense to get rid of the redundant / arbitrary differences between the TestUtil class here and in doctrine/dbal (and their method etc naming is clearer) so have done that too - sorry that makes the diff a little bit harder to follow. The only real diff between those classes now is that we have removed the fallback connection and changed the naming of the `tmpdb_` params - probably worth porting that the other way so that the two projects are consistent? I can do that if you agree.

greg0ire (approved):

greg0ire:
Thanks a lot for the top-notch quality PR!

> probably worth porting that the other way so that the two projects are consistent? I can do that if you agree.

Yes please :pray:

Issue #8527, opened 2021-03-09
Breaking change to optimistic locking in 2.8.2


acoulton:
Version 2.8.2 introduced a breaking change to the way optimistic lock versions are compared.

This has been reported (and fixed) specifically for DateTime objects as #8499 but in reality it applies more broadly.

Prior to 2.8.2 the version was compared with a `!=` operator. However among the 25k lines changed in this [automated coding style fix commit](https://github.com/doctrine/orm/commit/68bc00b6c672349925c247951f13c7c90eb801e6#diff-55a900494fc8033ab498c53929716caf0aa39d6bdd7058e7d256787a24412ee4R2502) is a change from that to a `!==` operator.

Previously, this would work:

```php
// view.php
//... form stuff
<input type="hidden" name="version" value="<?=$entity->getVersion();?>">

// controller.php
$entity = $em->find('MyEntity', $id, LockMode::OPTIMISTIC, $_POST['version']);
```

However because the POST value is a `string` rather than an `int` it now consistently fails with an OptimisticLockException.

I appreciate that the documentation examples for this show the controller code explicitly casting the incoming value to `(int) $version` before passing it in. Also that the various Doctrine phpdocs have `@param int|null $lockVersion` - although this is technically incorrect, since it's also valid to pass a DateTimeInterface.

So there's an argument the old behaviour was wrong, and the caller should always have been forced to cast the value if required.

On the other hand, since there's only (incomplete/inaccurate) phpdoc rather than strict typehints, and it's always "just worked" it seems reasonable to assume a fair number of people will be just passing in the POST/GET value without explicitly casting.

And arguably it should actually be valid to pass the raw string here - AFAIK integer-versioned entities always start at version 1, so there's no case where you could get a false-positive-version match by passing in a string that casts to `0`. If you pass a string, either it contains the value of the correct integer (and the lock is valid), or you haven't (and the lock fails).

Either way the release changes previously-working behaviour, so should at very least have a changelog / commit message highlighting that even if you consider this a side-effect of a bugfix rather than a semver-level breaking change.

[...]

PR #8533, opened 2021-03-11
Add test coverage for passing optimistic lock version as string

acoulton:
As suggested by @greg0ire I have updated the tests to cover the current behaviour of passing a string `version` argument for optimistic locking.

Since #8527 was only-sort-of-a-bug and this is just a simple variation of how `->lock` is already tested, I thought it made sense to modify the existing tests to cover this case rather than the overhead of a whole separate test case. Particularly since I guess this will probably change again in a future release, and the goal is just to avoid that happening accidentally in the meantime.

If you'd prefer to have a standalone GH8527 ticket test instead I can happily do that, just let me know.


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @acoulton !

PR #8539, opened 2021-03-13
🎉 Final CS batch 🎉

greg0ire:
```
vendor/bin/phpcs --report=summary
............................................................ 60 / 80 (75%)
....................                                         80 / 80 (100%)


Time: 3.75 secs; Memory: 82MB
```

I will follow up with a merge-up to 2.9.x (already tried it, it's not difficult).

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -321,15 +321,15 @@ abstract class AbstractHydrator
      * field names during this procedure as well as any necessary conversions on
      * the values applied. Scalar values are kept in a specific key 'scalars'.
      *
-     * @param mixed[] $data                SQL Result Row.
-     * @param array   &$nonemptyComponents Does this DQL-Alias has at least one non NULL value?
+     * @param mixed[] $data SQL Result Row.
      *
      * @return array<string, array<string, mixed>> An array with all the fields
      *                                             (name => value) of the data
      *                                             row, grouped by their
      *                                             component alias.
      *
-     * @psalm-param array<string, string> &$id                 Dql-Alias => ID-Hash.
+     * @psalm-param array<string, string> $id                 Dql-Alias => ID-Hash.
+     * @psalm-param array<string, bool>   $nonemptyComponents Does this DQL-Alias has at least one non NULL value?
      * @psalm-return array{
      *                   data: array<array-key, array>,
      *                   newObjects?: array<array-key, array{
@@ -414,9 +414,8 @@ abstract class AbstractHydrator
      * values according to their types. The resulting row has the same number
      * of elements as before.
      *
-     * @param array $data
-     *
-     * @return array The processed row.
+     * @psalm-param array<string, mixed> $data
+     * @psalm-return array<string, mixed> The processed row.
      */
     protected function gatherScalarRowData(&$data)
     {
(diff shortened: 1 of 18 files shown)


lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -414,9 +414,8 @@ protected function gatherRowData(array $data, array &$id, array &$nonemptyCompon
      * values according to their types. The resulting row has the same number
      * of elements as before.
      *
-     * @param array $data
```
simonberger, line 417: This removes param information for non-psalm interpreters. I saw that also in older batches.
I you want I can add a PR to fix this as well as using psalm-param / psalm-var for all array<> notations.

greg0ire: I'm not 100% sure yet if `psalm-param` or `psalm-var` is the way to go or if we can use just `@param`. I decided to take no risk of providing information that cannot be understood and went with the prefix any time there were `<>`, maybe we should discuss this more in depth? What non-psalm interpreters are you thinking about? PHPStorm, probably, is there something else?

simonberger: I also don't know how well IDEs like eclipse or netbeans are working with array<> for example. At least I hope there are plugins to enable understanding them.
At least we should always have `@param` if a typing is missing.
I prefer removing the psalm prefix for this notation. but there might be more investigation needed to find out how well it is supported in more IDEs.

simonberger: After some research my conclusion is that array generic notation should be added to psalm-prefixed PHPDoc attributes only.
The adaption in tools other than intellij, Psalm or PHPStan seems not existent.
[It had been implemented in PHPStorm](https://blog.jetbrains.com/phpstorm/2018/09/phpstorm-2018-3-eap-183-2635-12/#psr-5-generics) according to [PSR-5](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md)

The PSR-5 chapter for collections had been removed from the draft a month later and did not find a way back yet.
[Here are some more information in the PR removing the collection part](https://github.com/php-fig/fig-standards/pull/1102)
The goal was to newly start the discussion of this topic. I did not try to find out how this proceeded.

greg0ire: I feel this is an important discussion we need to have regularly (because tooling will evolve, and new PSRs might come up). I know we have already discussed this in the past somewhere, but cannot find where.

I think there should either be an issue for this at https://github.com/doctrine/coding-standard/issues or a contribution should be done to https://github.com/doctrine/doctrine-website/tree/master/source/policies . Can you pick one and open an issue there? That way we can point Doctrine members and people that are knowledgeable about various IDE support to it.

In any case, my point is that we should have a central location where we record this decision and what it is grounded on so that it can be found and challenged in the future, when tooling has evolved. Having a table with syntaxes and how supported they are across IDEs would help a lot.

We usually use https://github.com/orgs/doctrine/teams/doctrinecore for that kind of thing, but it is internal, so that would prevent us from discussing it with you.

beberlei: I believe it makes sense that we have `@param array` + `@psalm-param array{....}`, probably a phpcs rule that can autofix the param for us, but it needs some work in the coding standard repo.

beberlei (approved):
yay!

Issue #8379, opened 2020-12-06
Deprecate type inference in `Query::setParameter`


beberlei:
We should assume every object passed *must* be an entity *or* specify a $type as third argument. Only primitive types should work with type inference. A deprecation is in order to allow users. to migrate of this behavior.

Problems with type inference:
- Performance problems due to trying to load mapping files and failing
- Ambiguities when inferring `DateTime`, to `datetime` or `datetime_utc`.

beberlei:
In addition we should think about differentiating between `setParameter` for single valued paraemters and `setParameterList` for an array of values of the specified type. Then this could allow support for arrays of more complex types:

```
$dates = [new DateTime, new DateTime];
$query->setParameterList(1, $dates, "datetime");
```

beberlei:
Needs more research

sips-richard:
The optional type argument on setParameter() and the baked in behaviour of inferType() for booleans and datetimes is confusing and sets the wrong expectation for consumers. I've just been digging into this to understand why these types were treated differently from other registered types. We will be updating our company coding standard to mandate that the type is always specified for consistency and clarity (obviously Connection::PARAM_STR_ARRAY etc. isn't ideal). I would welcome this change in Doctrine.

PR #8540, opened 2021-03-14
Deprecate type inference in `Query::setParameter`

abame:
Deprecated type inference for `Query::setParameter` mentioned in issue #8379

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -381,6 +381,13 @@ abstract class AbstractQuery
      */
     public function setParameter($key, $value, $type = null)
     {
+        if ($type === null) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/issues/8379',
+                'Consider adding the type as third argument since null will not be accepted anymore in Doctrine ORM 3.0.'
+            );
+        }
         $existingParameter = $this->getParameter($key);
 
         if ($existingParameter !== null) {
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -388,6 +388,7 @@ abstract class AbstractQuery
                 'Consider adding the type as third argument since null will not be accepted anymore in Doctrine ORM 3.0.'
             );
         }
+
         $existingParameter = $this->getParameter($key);
 
         if ($existingParameter !== null) {
--- a/lib/Doctrine/ORM/Query/ParameterTypeInferer.php
+++ b/lib/Doctrine/ORM/Query/ParameterTypeInferer.php
@@ -44,6 +44,9 @@ class ParameterTypeInferer
      * - Type (\Doctrine\DBAL\Types\Type::*)
      * - Connection (\Doctrine\DBAL\Connection::PARAM_*)
      *
+     * @deprecated Infer Type will be removed in the future for setParameter method,
+     * - Consider adding the type as third argument to setParameter method since null will not be accepted anymore
+     *
      * @param mixed $value Parameter value.
      *
      * @return mixed Parameter type constant.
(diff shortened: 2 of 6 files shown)


lib/Doctrine/ORM/Query/ParameterTypeInferer.php
```diff
@@ -44,6 +44,9 @@ class ParameterTypeInferer
      * - Type (\Doctrine\DBAL\Types\Type::*)
      * - Connection (\Doctrine\DBAL\Connection::PARAM_*)
      *
+     * @deprecated Infer Type will be removed in the future for setParameter method,
```
greg0ire: Will this not cause IDEs and static analyzers to mark _all_ calls to this method as deprecated?

abame: yes that's correct but isn't this the desired outcome of the opened issue or did I misunderstood the issue?

greg0ire: I think you misunderstood. IMO what's needed here is some usage of `doctrine/deprecations` in the cases that should be deprecated and only those.

abame: My bad then. Does this mean, if I understood it correctly now, that we just need to trigger a deprecation warning for the third argument of `setParameter` method by using `doctrine/deprecations`?

greg0ire: Yes, I believe that's indeed what's needed 👍

abame: thanks for the clarification, I'll update the PR later today

abame: Do I have to update the existing tests or create a new one? How is this usually handled?

greg0ire: It looks like you had the former anyway, but it might be good to do the latter too, to demonstrate that it affects all classes extending `AbstractQuery`, and not just `NativeQuery`

abame: I found only two classes that extend the `AbstractQuery` and the tests for both classes are updated. Can you think about any use-case for which we might need new tests for this?

greg0ire: Since `AbstractQuery` is neither final nor internal, anybody can extend it outside `doctrine/orm` if they want to, and they should get the deprecation too.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -381,6 +381,14 @@ public function setParameters($parameters)
      */
     public function setParameter($key, $value, $type = null)
     {
+        if ($type === null) {
```
greg0ire, line 384: > We should assume every object passed must be an entity or specify a $type as third argument. Only primitive types should work with type inference.

I think the deprecation should fire only in some cases, but I'm not sure how to do that detection:  @beberlei please help :pray:

abame: Based on the example of `Datetime` I guess we should also check if the value is an object and type is null, than we trigger the deprecation, but let's wait and see if we can get some more input

beberlei:
sorry didn't see tis earlier, i looked over the code for a few times as well, its sadly quite complicated and i haven't found the perfect location either. I believe `processParameterValue` should be it, but its a bit difficult.

abame:
> sorry didn't see tis earlier, i looked over the code for a few times as well, its sadly quite complicated and i haven't found the perfect location either. I believe `processParameterValue` should be it, but its a bit difficult.

@beberlei just for my understanding since from the issue description it says, `Only primitive types should work with type inference`, and based on this my question is that will it not be enough to check the following in `setParameter` method:
```php
if(is_object($value) && $type === null) {
   //trigger deprecation notice
}
```

This way we let the user know that the third argument has to be provided for non-primitive values. What are the implications that prevent us from doing so? In the `processParameterValue` method we only know about the value and we don't know if the `type` argument is provided or not in the `setParameter` method

beberlei:
What I want to prevent here is #8113 for example, in `AbstractQuery::processParameterValue` the `MappingException` is swallowed. I would say we should put the deprecation message there: "Passing a parameter value of type object that is not an entity is deprecated, unless you also specify its Doctrine\DBAL\Types\Type name."

PR #8544, opened 2021-03-14
Add type declarations where backwards-compatible

### LINKED ISSUE #8538: Add backwards-compatible type declarations
We should grep for:
- `private function`
- `final protected function`
- `final class`

And see if we can add type declarations.
Since the phpdoc might be inaccurate, this is a risky move, and as such should IMO target 2.9.x

greg0ire:
This includes:
- private methods
- return type declarations of final protected methods
- return type declarations of public and protected methods of final classes

Parameter type declarations are a more delicate matter and should
probably be handled separately to make it easier to catch issues during
code review.

Type declarations can be more trusted than simple phpdoc when it
comes to static analysis, having them means we can infer the phpdoc of
calling methods with confidence.

Note that it seems that some of the phpdoc I initially inferred these
declarations from were apparently wrong, in particular some mentioning
`Doctrine\Dbal\Statement` when was is really passed around is
`Doctrine\Dbal\Driver\Statement`.

Closes #8538

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -347,7 +347,7 @@ class DefaultQueryCache implements QueryCache
      * @return mixed[]|null
      * @psalm-return array{targetEntity: class-string, type: mixed, list?: array[], identifier?: array}|null
      */
-    private function storeAssociationCache(QueryCacheKey $key, array $assoc, $assocValue)
+    private function storeAssociationCache(QueryCacheKey $key, array $assoc, $assocValue): ?array
     {
         $assocPersister = $this->uow->getEntityPersister($assoc['targetEntity']);
         $assocMetadata  = $assocPersister->getClassMetadata();
@@ -397,13 +397,15 @@ class DefaultQueryCache implements QueryCache
     }
 
     /**
-     * @param string $assocAlias
      * @param object $entity
      *
      * @return array<object>|object
      */
-    private function getAssociationValue(ResultSetMapping $rsm, $assocAlias, $entity)
-    {
+    private function getAssociationValue(
+        ResultSetMapping $rsm,
+        string $assocAlias,
+        $entity
+    ) {
         $path  = [];
         $alias = $assocAlias;
 
--- a/lib/Doctrine/ORM/Cache/Persister/Entity/NonStrictReadWriteCachedEntityPersister.php
+++ b/lib/Doctrine/ORM/Cache/Persister/Entity/NonStrictReadWriteCachedEntityPersister.php
@@ -100,17 +100,14 @@ class NonStrictReadWriteCachedEntityPersister extends AbstractEntityPersister
 
     /**
      * @param object $entity
-     * @param bool   $isChanged
-     *
-     * @return bool
      */
-    private function updateCache($entity, $isChanged)
+    private function updateCache($entity, bool $isChanged): bool
     {
         $class     = $this->metadataFactory->getMetadataFor(get_class($entity));
         $key       = new EntityCacheKey($class->rootEntityName, $this->uow->getEntityIdentifier($entity));
         $entry     = $this->hydrator->buildCacheEntry($class, $key, $entity);
         $cached    = $this->region->put($key, $entry);
-        $isChanged = $isChanged ?: $cached;
+        $isChanged = $isChanged || $cached;
 
         if ($this->cacheLogger && $cached) {
             $this->cacheLogger->entityCachePut($this->regionName, $key);
--- a/lib/Doctrine/ORM/Internal/Hydration/ArrayHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ArrayHydrator.php
@@ -259,15 +259,16 @@ class ArrayHydrator extends AbstractHydrator
      * Updates the result pointer for an Entity. The result pointers point to the
      * last seen instance of each Entity type. This is used for graph construction.
      *
-     * @param mixed[]  $coll     The element.
-     * @param bool|int $index    Index of the element in the collection.
-     * @param string   $dqlAlias
-     * @param bool     $oneToOne Whether it is a single-valued association or not.
-     *
-     * @return void
+     * @param mixed[]|null $coll     The element.
+     * @param bool|int     $index    Index of the element in the collection.
+     * @param bool         $oneToOne Whether it is a single-valued association or not.
      */
-    private function updateResultPointer(array &$coll, $index, $dqlAlias, $oneToOne)
-    {
+    private function updateResultPointer(
+        ?array &$coll,
+        $index,
+        string $dqlAlias,
+        bool $oneToOne
+    ): void {
         if ($coll === null) {
             unset($this->_resultPointers[$dqlAlias]); // Ticket #1228
 
--- a/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
@@ -171,15 +171,16 @@ class ObjectHydrator extends AbstractHydrator
     /**
      * Initializes a related collection.
      *
-     * @param object        $entity         The entity to which the collection belongs.
-     * @param ClassMetadata $class
-     * @param string        $fieldName      The name of the field on the entity that holds the collection.
-     * @param string        $parentDqlAlias Alias of the parent fetch joining this collection.
-     *
-     * @return PersistentCollection
+     * @param object $entity         The entity to which the collection belongs.
+     * @param string $fieldName      The name of the field on the entity that holds the collection.
+     * @param string $parentDqlAlias Alias of the parent fetch joining this collection.
      */
-    private function initRelatedCollection($entity, $class, $fieldName, $parentDqlAlias)
-    {
+    private function initRelatedCollection(
+        $entity,
+        ClassMetadata $class,
+        string $fieldName,
+        string $parentDqlAlias
+    ): PersistentCollection {
         $oid      = spl_object_hash($entity);
         $relation = $class->associationMappings[$fieldName];
         $value    = $class->reflFields[$fieldName]->getValue($entity);
@@ -225,11 +226,11 @@ class ObjectHydrator extends AbstractHydrator
      * @param string $dqlAlias The DQL alias of the entity's class.
      * @psalm-param array<string, mixed> $data     The instance data.
      *
-     * @return object The entity.
+     * @return object
      *
      * @throws HydrationException
      */
-    private function getEntity(array $data, $dqlAlias)
+    private function getEntity(array $data, string $dqlAlias)
     {
         $className = $this->_rsm->aliasMap[$dqlAlias];
 
@@ -272,12 +273,12 @@ class ObjectHydrator extends AbstractHydrator
     }
 
     /**
-     * @param string $className
+     * @psalm-param class-string $className
      * @psalm-param array<string, mixed> $data
      *
      * @return mixed
      */
-    private function getEntityFromIdentityMap($className, array $data)
+    private function getEntityFromIdentityMap(string $className, array $data)
     {
         // TODO: Abstract this code and UnitOfWork::createEntity() equivalent?
         $class = $this->_metadataCache[$className];
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -61,7 +61,7 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
     /** @var EntityManagerInterface|null */
     private $em;
 
-    /** @var AbstractPlatform */
+    /** @var AbstractPlatform|null */
     private $targetPlatform;
 
     /** @var MappingDriver */
@@ -310,11 +310,9 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
      * Populates the discriminator value of the given metadata (if not set) by iterating over discriminator
      * map classes and looking for a fitting one.
      *
-     * @return void
-     *
      * @throws MappingException
      */
-    private function resolveDiscriminatorValue(ClassMetadata $metadata)
+    private function resolveDiscriminatorValue(ClassMetadata $metadata): void
     {
         if (
             $metadata->discriminatorValue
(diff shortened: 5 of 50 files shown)


lib/Doctrine/ORM/Internal/Hydration/ArrayHydrator.php
```diff
@@ -261,13 +261,14 @@ protected function hydrateRowData(array $row, array &$result)
...
      * @param bool|int $index    Index of the element in the collection.
-     * @param string   $dqlAlias
      * @param bool     $oneToOne Whether it is a single-valued association or not.
-     *
-     * @return void
      */
-    private function updateResultPointer(array &$coll, $index, $dqlAlias, $oneToOne)
-    {
+    private function updateResultPointer(
+        array &$coll,
```
simonberger: `null` is handled in the code for parameter `$coll`.

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -292,17 +277,15 @@ static function ($a, $b): int {
      *
      * @psalm-return array<string, mixed>
      */
-    public function getMapping()
+    public function getMapping(): array
```
simonberger: Result can be null. Property var attribute needs a change as well.

greg0ire:
Thanks for the careful review @simonberger , much appreciated 👍

simonberger (dismissed):

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -344,11 +344,9 @@ public function put(QueryCacheKey $key, ResultSetMapping $rsm, $result, array $h
      * @param array<string,mixed> $assoc
      * @param mixed               $assocValue
      *
-     * @return mixed[]|null
```
BackEndTea, line 347: I'd keep the annotation here. That at least makes it clear the return type wasn't forgotten here

lib/Doctrine/ORM/Cache/Persister/Entity/NonStrictReadWriteCachedEntityPersister.php
```diff
@@ -98,13 +98,7 @@ public function update($entity)
...
     }
 
-    /**
-     * @param object $entity
-     * @param bool   $isChanged
-     *
-     * @return bool
-     */
-    private function updateCache($entity, $isChanged)
+    private function updateCache(object $entity, bool $isChanged): ?bool
```
ostrolucky: rather than changing it from docblock version, i would prefer fixing the code here: `$isChanged = $isChanged || $cached;`

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
```diff
@@ -273,13 +272,11 @@ private function getEntity(array $data, $dqlAlias)
     }
 
     /**
-     * @param string $className
```
ostrolucky, line 275: isn't this a class-string? should be marked as such, if so

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -384,12 +382,8 @@ private function addDefaultDiscriminatorMap(ClassMetadata $class)
 
     /**
      * Gets the lower-case short name of a class.
-     *
-     * @param string $className
```
ostrolucky, line 393: class-string?

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -526,12 +522,8 @@ private function getTablePrimaryKeys(Table $table)
 
     /**
      * Returns the mapped class name for a table if it exists. Otherwise return "classified" version.
-     *
-     * @param string $tableName
-     *
-     * @return string
```
ostrolucky, line 537: class-string?

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -821,11 +819,9 @@ private function columnToArray(SimpleXMLElement $fieldMapping)
     /**
      * Parse / Normalize the cache configuration
      *
-     * @return mixed[]
```
ostrolucky, line 878: Sometimes you remove these, sometimes don't, like in yamldriver or PersistentCollection. So which way is it, what's the policy?

lib/Doctrine/ORM/NativeQuery.php
```diff
@@ -60,7 +60,7 @@ public function getSQL()
     }
 
     /**
-     * {@inheritdoc}
+     * {@inheritDoc}
```
ostrolucky: what was your motivation to change these? I fuzzily remember phpcs understood lowercase variant only

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -143,18 +141,13 @@ public function setOwner($entity, array $assoc)
...
+    public function getOwner(): ?object
     {
         return $this->owner;
     }
 
-    /**
-     * @return Mapping\ClassMetadata
-     */
-    public function getTypeClass()
+    public function getTypeClass(): Mapping\ClassMetadata
```
ostrolucky: This one is dangerous. We don't enforce type in constructor and there also is ClassMetadataInfo

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -348,20 +329,16 @@ public function setDirty($dirty)
...
         $this->initialized = $bool;
     }
 
     /**
      * {@inheritdoc}
-     *
-     * @return object
      */
-    public function remove($key)
+    public function remove($key): object
```
ostrolucky: should be `?object`. It calls parent::remove, ArrayCollection returns null in some cases and interface also says T|null.

please also take care of references to this method, like offsetUnset

lib/Doctrine/ORM/Query.php
```diff
@@ -279,7 +277,7 @@ private function parse()
     }
 
     /**
-     * {@inheritdoc}
+     * {@inheritDoc}
      */
     protected function _doExecute()
```
ostrolucky: return type here should probably be added

lib/Doctrine/ORM/Query.php
```diff
@@ -523,10 +521,8 @@ public function setQueryCacheLifetime($timeToLive): self
 
     /**
      * Retrieves the lifetime of resultset cache.
-     *
-     * @return int
      */
-    public function getQueryCacheLifetime()
+    public function getQueryCacheLifetime(): int
```
ostrolucky: Not sure why SA isn't complaining here. `queryCacheTTL` prop is uninitializd so I think it may return null

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
```diff
@@ -223,15 +225,13 @@ private function getColumnAlias($columnName, $mode, array $customRenameColumns)
      *
      * This depends on the renaming mode selected by the user.
      *
-     * @param string $className
```
ostrolucky, line 222: class-string?

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
```diff
@@ -223,15 +225,13 @@ private function getColumnAlias($columnName, $mode, array $customRenameColumns)
...
-     * @param int    $mode
-     *
      * @return string[]
      *
+     * @psalm-param self::COLUMN_RENAMING_* $mode
      * @psalm-param array<string, string> $customRenameColumns
      * @psalm-return array<array-key, string>
      */
-    private function getColumnAliasMap($className, $mode, array $customRenameColumns)
+    private function getColumnAliasMap(string $className, $mode, array $customRenameColumns): array
```
ostrolucky: `int $mode`? Either that or docblock should be fixed, shouldn't it?

lib/Doctrine/ORM/Tools/EntityRepositoryGenerator.php
```diff
@@ -95,24 +95,16 @@ public function generateEntityRepositoryClass($fullClassName)
 
     /**
      * Generates the namespace, if class do not have namespace, return empty string instead.
-     *
-     * @param string $fullClassName
```
ostrolucky, line 99: class-string

lib/Doctrine/ORM/Tools/EntityRepositoryGenerator.php
```diff
@@ -95,24 +95,16 @@ public function generateEntityRepositoryClass($fullClassName)
...
-    private function getClassNamespace($fullClassName)
+    private function getClassNamespace(string $fullClassName): string
     {
         return substr($fullClassName, 0, strrpos($fullClassName, '\\'));
     }
 
     /**
      * Generates the class name
-     *
-     * @param string $fullClassName
```
ostrolucky, line 111: class-string

lib/Doctrine/ORM/Tools/EntityRepositoryGenerator.php
```diff
@@ -129,22 +121,15 @@ private function generateClassName($fullClassName)
      * Generates the namespace statement, if class do not have namespace, return empty string instead.
      *
      * @param string $fullClassName The full repository class name.
```
ostrolucky: class-string

lib/Doctrine/ORM/Tools/EntityRepositoryGenerator.php
```diff
@@ -129,22 +121,15 @@ private function generateClassName($fullClassName)
...
-    private function generateEntityRepositoryNamespace($fullClassName)
+    private function generateEntityRepositoryNamespace(string $fullClassName): string
     {
         $namespace = $this->getClassNamespace($fullClassName);
 
         return $namespace ? 'namespace ' . $namespace . ';' : '';
     }
 
-    /**
-     * @param string $fullClassName
```
ostrolucky, line 143: class-string

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -219,10 +208,8 @@ private function useOutputWalker(Query $query)
 
     /**
      * Appends a custom tree walker to the tree walkers hint.
-     *
-     * @param string $walkerClass
```
ostrolucky, line 224: class-string

lib/Doctrine/ORM/Tools/ResolveTargetEntityListener.php
```diff
@@ -110,12 +109,9 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $args)
...
 
     /**
-     * @param ClassMetadataInfo $classMetadata
-     * @param mixed[]           $mapping
-     *
-     * @return void
+     * @param mixed[] $mapping
      */
-    private function remapAssociation($classMetadata, $mapping)
+    private function remapAssociation(ClassMetadata $classMetadata, $mapping): void
```
ostrolucky: any reason why not declaring type for $mapping?

ostrolucky (changes requested):

lib/Doctrine/ORM/Query.php
greg0ire: I think that's because we are at a dangerously lenient level right now.

lib/Doctrine/ORM/Tools/ResolveTargetEntityListener.php
greg0ire: That one slipped through :P

lib/Doctrine/ORM/NativeQuery.php
greg0ire: I don't remember, and we seem to be using a mix of both in the project :sweat: .

Will look into it later.

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
greg0ire: It's unclear to me what's best, will give it more thought later.

lib/Doctrine/ORM/Query.php
greg0ire: That one causes so many test failures :scream:

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
ostrolucky: If I may, I slightly prefer keeping them. There might be tools that don't respect psalm vendor specific annotations and mixed[] makes it clear to communicate we didn't forget to specify the type.

lib/Doctrine/ORM/Query.php
ostrolucky: So is it actually `mixed`?

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
greg0ire: Ok, I'll restore them then 👍

lib/Doctrine/ORM/Query.php
greg0ire: Indeed, it's not just the tests, the sql executor hierarchy has inconsistent return type declarations. `MultiTableDeleteExecutor` declares `int`, so does `SingleTableDeleteUpdateExecutor`.

ostrolucky (commented):
One comment remaining. Also, you are adding object type declarations, but these are being removed in https://github.com/doctrine/orm/pull/8613. That will cause mess of a conflict.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -24,9 +24,9 @@
 use Doctrine\Common\Collections\Expr\Comparison;
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\ResultStatement as DriverStatement;
 use Doctrine\DBAL\LockMode;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
-use Doctrine\DBAL\Statement;
```
greg0ire, line 29: Unsure if we should wait for https://github.com/doctrine/dbal/pull/4596 and revert this change in use statements. I think that change might be good because more forward compatible.

ostrolucky: I'll leave you decide. But to me, such change seems a bit unrelated to this PR

greg0ire:
I will wait until #8613 is resolved

ostrolucky (dismissed):

greg0ire:
The 7.1 PR has been merged, let's rebase to be 100% sure it still works

PR #8545, opened 2021-03-15
remove some empty() use when safe

orklah:
This PR propose replacing some empty() uses by a more explicit expression when the type of the variable is known statically from signature.

Those were generated by https://github.com/orklah/psalm-not-empty

### DIFF
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -1880,7 +1880,7 @@ public function __construct(<params>)
         $lines = explode("\n", $code);
 
         foreach ($lines as $key => $value) {
-            if ($value !== '' && $value !== '0') {
+            if ($value !== '') {
                 $lines[$key] = str_repeat($this->spaces, $num) . $lines[$key];
             }
         }
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -1880,7 +1880,7 @@ public function __construct(<params>)
         $lines = explode("\n", $code);
 
         foreach ($lines as $key => $value) {
-            if (! empty($value)) {
+            if ($value !== '' && $value !== '0') {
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -1876,7 +1876,7 @@ protected function prefixCodeWithSpaces($code, $num = 1)
         $lines = explode("\n", $code);
 
         foreach ($lines as $key => $value) {
-            if (! empty($value)) {
+            if (($value !== '' && $value !== '0')) {
```
greg0ire:
```suggestion
            if ($value !== '' && $value !== '0') {
```
Are you sure we want to test for `'0'` here?

orklah: That's what empty() does under the hood. I can remove the '0' test, but this may have unintended consequences. (That's exactly why I despise empty() :D )

greg0ire: Yeah me too, but now with your patch, people might think this `$value !== '0'` really is useful, and it might very well not be.

orklah: What's best here, risk the side effect or revert?

greg0ire: Reading the code and understanding what it does is best. Here, we have a method that indents a code block. Except it should not indent empty lines otherwise that will mean trailing spaces. So IMO we can drop that `&& $value !== '0'`, because in the super unlikely case we have a `0` sitting by itself on a single line in a code block, we want to indent it.

greg0ire (dismissed):

greg0ire:
Nice job and nice plugin BTW :) Not fully clear to me if this should go to 2.8.x or 2.9.x . Since there are code changes that are not bugfixes, maybe 2.9.x would be best? Let's see what other maintainers recommend.

beberlei:
2.9 please :)

greg0ire (approved):

greg0ire:
Thanks @orklah !

PR #8546, opened 2021-03-15
use constants when possible

orklah:
This PR proposes using constants instead of literal string when possible.

This was generated by https://github.com/orklah/psalm-use-constants and then checked by hand. It's possible that some constants could be used where the context does not match but I don't know doctrine's internal well enough to judge every case.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Builder/ClassMetadataBuilder.php
+++ b/lib/Doctrine/ORM/Mapping/Builder/ClassMetadataBuilder.php
@@ -20,6 +20,7 @@
 
 namespace Doctrine\ORM\Mapping\Builder;
 
+use Doctrine\DBAL\Types\Types;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 
@@ -225,7 +226,7 @@ class ClassMetadataBuilder
      *
      * @return static
      */
-    public function setDiscriminatorColumn($name, $type = 'string', $length = 255)
+    public function setDiscriminatorColumn($name, $type = Types::STRING, $length = 255)
     {
         $this->cm->setDiscriminatorColumn(
             [
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -23,6 +23,7 @@ namespace Doctrine\ORM\Mapping;
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Platforms;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\Types;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Event\LoadClassMetadataEventArgs;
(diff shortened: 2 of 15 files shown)


lib/Doctrine/ORM/Cache/QueryCacheEntry.php
```diff
@@ -58,6 +60,6 @@ public function __construct($result, $time = null)
      */
     public static function __set_state(array $values)
     {
-        return new self($values['result'], $values['time']);
+        return new self($values['result'], $values[Types::TIME_MUTABLE]);
```
greg0ire: Unrelated I think

lib/Doctrine/ORM/Cache/TimestampCacheEntry.php
```diff
@@ -53,6 +55,6 @@ public function __construct($time = null)
      */
     public static function __set_state(array $values)
     {
-        return new self($values['time']);
+        return new self($values[Types::TIME_MUTABLE]);
```
greg0ire: Same

lib/Doctrine/ORM/Tools/Console/Command/RunDqlCommand.php
```diff
@@ -54,7 +55,7 @@ protected function configure()
         $this->setName('orm:run-dql')
              ->setDescription('Executes arbitrary DQL directly from the command line')
              ->addArgument('dql', InputArgument::REQUIRED, 'The DQL to execute.')
-             ->addOption('hydrate', null, InputOption::VALUE_REQUIRED, 'Hydration mode of result set. Should be either: object, array, scalar or single-scalar.', 'object')
+             ->addOption('hydrate', null, InputOption::VALUE_REQUIRED, 'Hydration mode of result set. Should be either: object, array, scalar or single-scalar.', Types::OBJECT)
```
greg0ire: Unrelated: array ,  scalar and single-scalar are not Types constants are they?

lib/Doctrine/ORM/Cache/QueryCacheEntry.php
orklah: reverted :)

SenseException (dismissed):

orklah:
Build failure have to be unrelated but not sure where it came from

greg0ire:
It looks like an issue with the terminal width, which is governed by https://github.com/doctrine/orm/blob/3f3de70c3e7e03e2c2e67fb990eda401b387b5d3/phpunit.xml.dist#L67

My bet is this has to do with the last `symfony/console` version

EDIT: looks like it! https://github.com/symfony/symfony/pull/40348

greg0ire:
Blocked by https://github.com/doctrine/orm/pull/8573

greg0ire:
Ah it needs a merge up

greg0ire:
@orklah there are conflicts, please rebase.

orklah:
Done. Psalm errors seems unrelated

greg0ire:
> Psalm errors seems unrelated

Yep, this is due to my latest work on `doctrine/persistence` I'm afraid. I will look into this, but you can add them to the baseline in the meantime if you want this to move forward. They don't seem to affect 2.8.x, which is kind of weird.

orklah:
Baseline regenerated. Maybe 2.8 is not affected because doctrine/persistence is pinned to an older version?

greg0ire:
It does not seem to be… maybe the strictness is not the same?

PR #8547, opened 2021-03-16
Make phpdoc types correct

### LINKED ISSUE #8503: Incompatibility between `getArrayResult()` new type hint and use of `indexBy()`
Cs fixes1 #8475 brought this change of `AbstractQuery` which I find incompatible with the use of `indexBy()`:
https://github.com/doctrine/orm/commit/f83322201771fc8db6b0a18b7d82fd51e893cc02#diff-65b30cb7319fd94159051fe212c87565a4ffd15a796fa0db4983c477076b88c8R798

~~@return array~~
@return array<int,mixed>

The new type hint is incompatible with a result indexed with a string property:

```php
$qb = $userRepository
    ->createQueryBuilder('u')
    ->indexBy('u', 'u.firstName')
    ->getQuery()
    ->getArrayResult()
;
```

I noticed the same issue with the same change for `getScalarResult()`.

I have thought of no alternative but to revert the change.

What do you think about this?

greg0ire:
Fixes #8503

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -428,7 +428,7 @@ class DefaultQueryCache implements QueryCache
      * @param mixed        $value
      * @param array<mixed> $path
      *
-     * @return array<object>|object|null
+     * @return mixed
      */
     private function getAssociationPathValue($value, array $path)
     {
--- a/lib/Doctrine/ORM/Cache/Region/FileLockRegion.php
+++ b/lib/Doctrine/ORM/Cache/Region/FileLockRegion.php
@@ -59,7 +59,7 @@ class FileLockRegion implements ConcurrentRegion
     /** @var string */
     private $directory;
 
-    /** @var int */
+    /** @psalm-var numeric-string */
     private $lockLifetime;
 
     /**
--- a/lib/Doctrine/ORM/Tools/Export/Driver/YamlExporter.php
+++ b/lib/Doctrine/ORM/Tools/Export/Driver/YamlExporter.php
@@ -244,8 +244,8 @@ class YamlExporter extends AbstractExporter
     }
 
     /**
-     * @psalm-param array{entityListeners: array<class-string, array<string, array{string}>>} $array
-     * @psalm-return array{entityListeners: array<class-string, array<string, array{string}>>}
+     * @psalm-param array<string, mixed> $array
+     * @psalm-return array<string, mixed>&array{entityListeners: array<class-string, array<string, array{string}>>}
      */
     private function processEntityListeners(ClassMetadataInfo $metadata, array $array): array
     {
--- a/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
@@ -132,7 +132,7 @@ class Paginator implements Countable, IteratorAggregate
     /**
      * {@inheritdoc}
      *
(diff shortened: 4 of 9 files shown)


lib/Doctrine/ORM/Tools/Export/Driver/YamlExporter.php
```diff
@@ -244,8 +244,8 @@ protected function yamlDump($array, $inline = 2)
     }
 
     /**
-     * @psalm-param array{entityListeners: array<class-string, array<string, array{string}>>} $array
-     * @psalm-return array{entityListeners: array<class-string, array<string, array{string}>>}
+     * @psalm-param array<string, mixed> $array
+     * @psalm-return array<string, mixed>&array{entityListeners: array<class-string, array<string, array{string}>>}
```
greg0ire, line 248: This is the best way I have found to say that I'm not super sure about the shape of the array except that it will have an `entityListeners` key with that form. Can and should be improved in subsequent PRs with more details.

greg0ire:
@benjamintoussaint @stlrnz @orklah @simonberger please give this a review :pray:

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -428,7 +428,7 @@ private function getAssociationValue(ResultSetMapping $rsm, $assocAlias, $entity
      * @param mixed        $value
      * @param array<mixed> $path
      *
-     * @return array<object>|object|null
```
orklah, line 431: Why do you widen the type here? Was this wrong?

lib/Doctrine/ORM/Cache/Region/FileLockRegion.php
```diff
@@ -59,7 +59,7 @@ class FileLockRegion implements ConcurrentRegion
     /** @var string */
     private $directory;
 
-    /** @var int */
+    /** @var string */
```
orklah: https://github.com/doctrine/orm/blob/d6f926d8a17b572e0ab673a185036058b8e93afd/lib/Doctrine/ORM/Cache/Region/FileLockRegion.php#L111

Maybe you're looking for the type numeric-string?

lib/Doctrine/ORM/Tools/Export/Driver/YamlExporter.php
orklah: That's a perfectly valid syntax :)

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -132,7 +132,7 @@ public function count()
     /**
      * {@inheritdoc}
      *
-     * @return ArrayIterator<mixed, T>
+     * @return ArrayIterator<int|string, T>
```
orklah: maybe array-key would be more adapted here?

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
greg0ire: Yes, or at least the recursivity line 456 makes it hard to understand for psalm.

At level 6, you get `ERROR: InvalidReturnStatement - lib/Doctrine/ORM/Cache/DefaultQueryCache.php:459:16 - The inferred type 'list<array<array-key, object>|null|object>' does not match the declared return type 'array<array-
key, object>|null|object' for Doctrine\ORM\Cache\DefaultQueryCache::getAssociationPathValue (see http
s://psalm.dev/128)`.

beberlei (approved):

PR #8548, opened 2021-03-16
Adding details to types in PHPDoc

orklah:
This PR propose detailing phpdoc for the most part. Most types were generated directly by Psalm and refined by hand.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -312,6 +312,7 @@ abstract class AbstractQuery
      * Get all defined parameters.
      *
      * @return ArrayCollection The defined query parameters.
+     * @psalm-return ArrayCollection<int, Parameter>
      */
     public function getParameters()
     {
@@ -323,7 +324,7 @@ abstract class AbstractQuery
      *
      * @param mixed $key The key (index or name) of the bound parameter.
      *
-     * @return Query\Parameter|null The value of the bound parameter, or NULL if not available.
+     * @return Parameter|null The value of the bound parameter, or NULL if not available.
      */
     public function getParameter($key)
     {
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -31,6 +31,7 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\ORM\Tools\Pagination\LimitSubqueryWalker;
 use Doctrine\ORM\UnitOfWork;
+use Generator;
 use PDO;
 use ReflectionClass;
 
@@ -152,7 +153,7 @@ abstract class AbstractHydrator
      *
      * @psalm-param array<string, mixed> $hints
      *
-     * @return iterable<mixed>
+     * @return Generator<int, mixed>
      */
     public function toIterable(ResultStatement $stmt, ResultSetMapping $resultSetMapping, array $hints = []): iterable
     {
@@ -220,7 +221,7 @@ abstract class AbstractHydrator
      * Hydrates a single row returned by the current statement instance during
      * row-by-row hydration with {@link iterate()} or {@link toIterable()}.
      *
-     * @return mixed
+     * @return mixed[]|false
      */
     public function hydrateRow()
     {
@@ -539,6 +540,7 @@ abstract class AbstractHydrator
 
     /**
      * @return string[]
+     * @psalm-return non-empty-list<string>
      */
     private function getDiscriminatorValues(ClassMetadata $classMetadata): array
     {
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -381,6 +381,7 @@ class ClassMetadataInfo implements ClassMetadata
      * READ-ONLY: The inheritance mapping type used by the class.
      *
      * @var int
+     * @psalm-var self::$INHERITANCE_TYPE_*
      */
     public $inheritanceType = self::INHERITANCE_TYPE_NONE;
 
@@ -447,6 +448,8 @@ class ClassMetadataInfo implements ClassMetadata
      *      originalField?: string,
      *      quoted?: bool,
      *      requireSQLConversion?: bool,
+     *      declaredField?: string,
+     *      options: array<mixed>
      * }>
      */
     public $fieldMappings = [];
@@ -1133,7 +1136,7 @@ class ClassMetadataInfo implements ClassMetadata
 
     /**
      * @param string $fieldName
-     * @psalm-param array{usage?: mixed, region?: mixed} $cache
+     * @psalm-param array{usage?: int, region?: string} $cache
      *
      * @return void
      */
@@ -1145,10 +1148,10 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * @param string $fieldName
      * @param array  $cache
-     * @psalm-param array{usage?: mixed, region?: mixed} $cache
+     * @psalm-param array{usage?: int, region?: string} $cache
      *
-     * @return mixed[]
-     * @psalm-return array{usage: mixed, region: mixed}
+     * @return int[]|string[]
+     * @psalm-return array{usage: int, region: string|null}
      */
     public function getAssociationCacheDefaults($fieldName, array $cache)
     {
@@ -1274,7 +1277,19 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param string $fieldName The field name.
      *
-     * @psalm-return array<string, mixed> The field mapping.
+     * @return mixed[] The field mapping.
+     * @psalm-return array{
+     *      type: string,
+     *      fieldName: string,
+     *      columnName?: string,
+     *      inherited?: class-string,
+     *      nullable?: bool,
+     *      originalClass?: class-string,
+     *      originalField?: string,
+     *      scale?: int,
+     *      precision?: int,
+     *      length?: int
+     * }
      *
      * @throws MappingException
      */
@@ -1354,7 +1369,8 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * Gets all named queries of the class.
      *
-     * @psalm-return array<string, mixed>
+     * @return mixed[][]
+     * @psalm-return array<string, array<string, mixed>>
      */
     public function getNamedQueries()
     {
@@ -1398,7 +1414,8 @@ class ClassMetadataInfo implements ClassMetadata
      *
      * @param string $name The result set mapping name.
      *
-     * @psalm-return array<string, mixed>
+     * @return mixed[]
+     * @psalm-return array{name: string, entities: array, columns: array}
      *
      * @throws MappingException
      */
@@ -1414,7 +1431,8 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * Gets all sql result set mappings of the class.
      *
-     * @psalm-return array<string, array<string, mixed>>
+     * @return mixed[]
+     * @psalm-return array<string, array{name: string, entities: array, columns: array}>
      */
     public function getSqlResultSetMappings()
     {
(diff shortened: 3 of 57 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -323,7 +325,7 @@ public function getParameters()
      *
      * @param mixed $key The key (index or name) of the bound parameter.
      *
-     * @return Query\Parameter|null The value of the bound parameter, or NULL if not available.
+     * @return Parameter|false|null The value of the bound parameter, or NULL if not available.
```
greg0ire: I don't think `false` can actually be returned here: if `$filteredParmeters` is empty, we return `null`

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -151,9 +152,8 @@ public function iterate($stmt, $resultSetMapping, array $hints = [])
     /**
      * Initiates a row-by-row hydration.
      *
-     * @return iterable<mixed>
-     *
      * @psalm-param array<string, mixed> $hints
+     * @psalm-return Generator<int, (array|mixed), mixed, void>
```
greg0ire: What does this syntax mean? I'd expect no more than one comma inside the first level of `<>`

orklah: Generator is kinda special and Psalm support up to 4 templated types for it: https://psalm.dev/docs/annotating_code/type_syntax/object_types/
Strangely, Psalm's doc says that the last two are mixed by default, but Psalter generated a void in 4th position instead of mixed. It doesn't seem to provoke errors up to Psalm lvl 4 one way or another, I removed the last two parameters

lib/Doctrine/ORM/AbstractQuery.php
orklah: You're right. I guess we could make Psalm understand this somehow to solve this on future levels

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -151,9 +152,8 @@ public function iterate($stmt, $resultSetMapping, array $hints = [])
     /**
      * Initiates a row-by-row hydration.
      *
-     * @return iterable<mixed>
-     *
      * @psalm-param array<string, mixed> $hints
+     * @psalm-return Generator<int, (array|mixed)>
```
greg0ire: :thinking: the union of `array` and `mixed` is `mixed`, so what's the point?

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1138,12 +1139,11 @@ public function enableAssociationCache($fieldName, array $cache)
 
     /**
      * @param string $fieldName
-     * @param array  $cache
      *
-     * @return mixed[]
+     * @return (int|string)[]
```
greg0ire: Are parenthesis understood by IDEs?

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1409,7 +1425,9 @@ public function getSqlResultSetMapping($name)
     /**
      * Gets all sql result set mappings of the class.
      *
-     * @psalm-return array<string, array<string, mixed>>
+     * @return (array|string)[][]
```
greg0ire: :thinking: phpcs might be happy with this for now, but maybe this should become `(mixed[]|string)[][]` or something more precise to be future proof

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1869,7 +1911,8 @@ protected function _validateAndCompleteOneToManyMapping(array $mapping)
      * @throws InvalidArgumentException
      *
      * @psalm-param array<string, mixed> $mapping The mapping to validate & complete.
-     * @psalm-return array{isOwningSide: mixed, orphanRemoval: bool}
+     *
+     * @psalm-return array{mappedBy: (mixed|null), inversedBy: (mixed|null), isOwningSide: bool, sourceEntity: class-string, targetEntity: string, fieldName: mixed, fetch: mixed, cascade: array<string>, isCascadeRemove: bool, isCascadePersist: bool, isCascadeRefresh: bool, isCascadeMerge: bool, isCascadeDetach: bool, type: int, originalField: string, originalClass: class-string, joinTable?: (ArrayAccess|array{inverseJoinColumns: (array{0: array{name: string, referencedColumnName: string, onDelete: string}}|mixed)}|mixed), joinTableColumns?: list<mixed>, isOnDeleteCascade?: true, relationToSourceKeyColumns?: array, relationToTargetKeyColumns?: array, orphanRemoval: bool}
```
greg0ire: Please use linebreaks

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -821,9 +823,9 @@ private function columnToArray(SimpleXMLElement $fieldMapping)
     /**
      * Parse / Normalize the cache configuration
      *
-     * @return mixed[]
+     * @return (mixed|string|null)[]
```
greg0ire:
```suggestion
     * @return mixed[]
```

lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
```diff
@@ -399,7 +399,7 @@ protected function getDeleteSQL(PersistentCollection $collection)
     /**
      * Internal note: Order of the parameters must be the same as the order of the columns in getDeleteSql.
      *
-     * @return mixed[]
+     * @return list<mixed>
```
greg0ire, line 402: Should this be duplicated because of the `list` part?

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1856,7 +1856,7 @@ public function expandParameters($criteria)
      *
      * @return mixed[][]
      *
-     * @psalm-return array{0: array, 1: list<mixed>}
+     * @psalm-return array{0: array, 1: list<(int|string|null)>}
```
greg0ire: The parenthesis don't look necessary

lib/Doctrine/ORM/Persisters/SqlValueVisitor.php
```diff
@@ -84,7 +84,7 @@ public function walkValue(Value $value)
      *
      * @return mixed[][]
      *
-     * @psalm-return array{0: array, 1: array}
+     * @psalm-return array{0: array, 1: array<array>}
```
greg0ire: Let's use `mixed[]` here and there?

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
```diff
@@ -51,7 +50,7 @@ class QueryExpressionVisitor extends ExpressionVisitor
     /** @var Expr */
     private $expr;
 
-    /** @var mixed[] */
+    /** @var array<int, mixed> */
```
greg0ire: Maybe we should use `list` here and in the constructor?

lib/Doctrine/ORM/Query/ResultSetMapping.php
```diff
@@ -428,7 +430,7 @@ public function isScalarResult($columnName)
      *
      * @param string $alias
      *
-     * @return string
+     * @return class-string
```
greg0ire, line 433: This is not understood by IDEs is it?

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
greg0ire: Oh right, how did I forget about this 👍 Anyway here it doesn't matter indeed, removing them is good.

greg0ire:
How closer to Psalm and PHPStan's next levels does this bring us?

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
orklah: fixed

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
orklah: Understood by VS Code but not by phpstorm. Fixed

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
orklah: fixed

lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
orklah: list is now an alias of array in phpstorm. Not sure in VS Code

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
orklah: fixed

lib/Doctrine/ORM/Persisters/SqlValueVisitor.php
orklah: replaced by array<mixed> to keep coherence

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
orklah: replaced here, the property was not from constructor though

lib/Doctrine/ORM/Query/ResultSetMapping.php
orklah: It's now an alias of string in PHPStorm. Not sure in VS Code

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
orklah: fixed

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
orklah: fixed

greg0ire:
@vladyslavstartsev @simonberger please review :pray:

orklah:
> How closer to Psalm and PHPStan's next levels does this bring us?

It wasn't quite my metric, but with issues up to level 3, it fixes 63 issues, it revealed 13 new ones and upped the type coverage from 90.4980% to 90.5595%

greg0ire:
Why are there 2 commits? What is the difference between "psalm fixes" and just "fixes"?

orklah:
!'m still working on that, I'll squash them later

orklah:
This should be good now :)

This has been rebased and I fixed the new issues related to the baselines that were introduced.

I had to change a few things:
- I updated phpstan version. The version that was used before didn't support PHP8 so I couldn't work with it in my dev environment
- I introduced phpstan/phpstan-webmozart-assert. The library is used at some point and phpstan needs the plugin to be able to understand what it does (EDIT: not anymore, I must have added the only usage myself and removed it since)
- I added a set php version to phpstan config. This will ensure everybody runs phpstan against the same version (namely the oldest version supported by the branch). This is needed to have a coherent baseline generated no matter what php version used to run phpstan itself (I was getting issues with different errors in PHP8 than in PHP7.4 on CI and even in PHP 7.2). Note that this is not the recommandation for this: https://github.com/phpstan/phpstan/issues/4871#issuecomment-821778532 but going further is not in the scope of this PR at all. It may be adressed later.

greg0ire:
Your last message has bullet points containing interesting information. Could you please make one commit per bullet point, and reuse what you wrote to create a commit messages? Some of these commits should probably target 2.8.x. For instance, to avoid issues when merging up, it would be great to have the same versions of phpstan on all branches.

The second one can stay on the 2.9.x PR I think, because it will have no effect on the CI of the end user.

The third one should be contributed to 2.8.x, since it does not affect the code itself but rather the build process, and as you pointed out, it seems important to have the same messages, especially when merging up (we don't want to have complicated things to do when merging up). Since it is a temporary solution, it should probably come with a comment saying we should remove it when we successfully remove the baseline?

And finally, some changes in this PR are type improvements, while some others are type fixes (the types were just plain wrong before, for instance when we use `int` instead of `int|false` as a return type. Using `git reset origin/2.9.x && git add -p` please separate the changes in 2 sets, then cherry-pick one of those on a new branch created from 2.8.x, then we will merge it, and merge the result up in 2.9.x

greg0ire (approved):
I learnt today that 2.9.0 was going to be released really soon, so there is no need to backport the "fixes" to 2.8.x

greg0ire:
Thanks @orklah !

PR #8552, opened 2021-03-18
Fix remaining warnings from the phpunit9 upgrade

acoulton:
Some tests were still using deprecated assertion and mocking methods,
resulting in a long list of warnings in the phpunit output.

This commit resolves all the warnings:

* Fixes a couple of test names in `@depends` tags (presumably these
  tests were renamed at some point for coding standards).

* Changes how mocks are configured when asserting the same method
  is called multiple times with a sequence of arguments / sequence
  of return values. The old `->at` expectation is deprecated because
  it can be brittle and give unreliable results. Some of this
  mocking could probably still be refactored further, but my focus
  was on solving the deprecation with the existing setup.

* Removes one use of prophecy for mocking, in favour of using
  phpunit mock objects. Prophecy now requires an extra composer
  dependency and a trait which seems overkill given it was only
  used in one place.

* Updates to the new names for assertFileExists and assertRegExp
  (and their `not` versions) - these are functionally equivalent.

* Replaces the last few references to old PHPUnit_Framework_XXX
  classes with their namespaced equivalent. These were mostly in
  comments, or in native php `assert()` statements that were sanity
  checking mocked object types. These asserts are probably redundant
  (and are clearly not running in CI since the classes they referenced
  no longer exist).

### DIFF
--- a/tests/Doctrine/Tests/DoctrineTestCase.php
+++ b/tests/Doctrine/Tests/DoctrineTestCase.php
@@ -4,52 +4,9 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests;
 
-use PHPUnit\Framework\TestCase;
-use PHPUnit\Runner\Version;
-
 /**
  * Base testcase class for all Doctrine testcases.
  */
-abstract class DoctrineTestCase extends TestCase
+abstract class DoctrineTestCase extends PolyfillableTestCase
 {
-    public static function assertDoesNotMatchRegularExpression(
-        string $pattern,
-        string $string,
-        string $message = ''
-    ): void {
-        // Forward-compatibility wrapper for phpunit 9 : can be removed once phpunit 8 / php 7.2 support is dropped.
-        if (self::isPhpunit9()) {
-            parent::assertDoesNotMatchRegularExpression($pattern, $string, $message);
-        } else {
-            self::assertNotRegExp($pattern, $string, $message);
-        }
-    }
-
-    public static function assertMatchesRegularExpression(string $pattern, string $string, string $message = ''): void
-    {
-        // Forward-compatibility wrapper for phpunit 9 : can be removed once phpunit 8 / php 7.2 support is dropped.
-        if (self::isPhpunit9()) {
-            parent::assertMatchesRegularExpression($pattern, $string, $message);
-        } else {
-            self::assertRegExp($pattern, $string, $message);
-        }
-    }
-
-    public static function assertFileDoesNotExist(string $filename, string $message = ''): void
-    {
-        // Forward-compatibility wrapper for phpunit 9 : can be removed once phpunit 8 / php 7.2 support is dropped.
-        if (self::isPhpunit9()) {
-            parent::assertFileDoesNotExist($filename, $message);
-        } else {
-            self::assertFileNotExists($filename, $message);
-        }
-    }
-
-    /**
-     * Check if we're running in phpunit >= 9.0
-     */
-    private static function isPhpunit9(): bool
-    {
-        return (int) Version::series() >= 9;
-    }
 }
--- a/tests/Doctrine/Tests/DoctrineTestCase.php
+++ b/tests/Doctrine/Tests/DoctrineTestCase.php
@@ -5,10 +5,51 @@ declare(strict_types=1);
 namespace Doctrine\Tests;
 
 use PHPUnit\Framework\TestCase;
+use PHPUnit\Runner\Version;
 
 /**
  * Base testcase class for all Doctrine testcases.
  */
 abstract class DoctrineTestCase extends TestCase
 {
+    public static function assertDoesNotMatchRegularExpression(
+        string $pattern,
+        string $string,
+        string $message = ''
+    ): void {
+        // Forward-compatibility wrapper for phpunit 9 : can be removed once phpunit 8 / php 7.2 support is dropped.
+        if (self::isPhpunit9()) {
+            parent::assertDoesNotMatchRegularExpression($pattern, $string, $message);
+        } else {
+            self::assertNotRegExp($pattern, $string, $message);
+        }
+    }
+
+    public static function assertMatchesRegularExpression(string $pattern, string $string, string $message = ''): void
+    {
+        // Forward-compatibility wrapper for phpunit 9 : can be removed once phpunit 8 / php 7.2 support is dropped.
+        if (self::isPhpunit9()) {
+            parent::assertMatchesRegularExpression($pattern, $string, $message);
+        } else {
+            self::assertRegExp($pattern, $string, $message);
+        }
+    }
+
+    public static function assertFileDoesNotExist(string $filename, string $message = ''): void
+    {
+        // Forward-compatibility wrapper for phpunit 9 : can be removed once phpunit 8 / php 7.2 support is dropped.
+        if (self::isPhpunit9()) {
+            parent::assertFileDoesNotExist($filename, $message);
+        } else {
+            self::assertFileNotExists($filename, $message);
+        }
+    }
+
+    /**
+     * Check if we're running in phpunit >= 9.0
+     */
+    private static function isPhpunit9(): bool
+    {
+        return (int) Version::series() >= 9;
+    }
 }
--- a/tests/Doctrine/Tests/ORM/Functional/QueryCacheTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/QueryCacheTest.php
@@ -59,7 +59,7 @@ class QueryCacheTest extends OrmFunctionalTestCase
     /**
      * @param <type> $query
      *
-     * @depends testQueryCache_DependsOnHints
+     * @depends testQueryCacheDependsOnHints
      */
     public function testQueryCacheDependsOnFirstResult($query): void
     {
@@ -76,7 +76,7 @@ class QueryCacheTest extends OrmFunctionalTestCase
     /**
      * @param <type> $query
      *
-     * @depends testQueryCache_DependsOnHints
+     * @depends testQueryCacheDependsOnHints
      */
     public function testQueryCacheDependsOnMaxResults($query): void
     {
@@ -92,7 +92,7 @@ class QueryCacheTest extends OrmFunctionalTestCase
     /**
      * @param <type> $query
      *
-     * @depends testQueryCache_DependsOnHints
+     * @depends testQueryCacheDependsOnHints
      */
     public function testQueryCacheDependsOnHydrationMode($query): void
     {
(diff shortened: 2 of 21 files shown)


greg0ire:
> These asserts are probably redundant (and are clearly not running in CI since the classes they referenced no longer exist).

On one hand, it would be great to check that these are accurate, on the other hand, it's also great to check that the code works when the `assert` statements are not run. Maybe we should pick one of the jobs and enable assertions on it…

tests/Doctrine/Tests/DoctrineTestCase.php
```diff
@@ -5,10 +5,51 @@
...
+            parent::assertFileDoesNotExist($filename, $message);
+        } else {
+            self::assertFileNotExists($filename, $message);
+        }
+    }
+
+    /**
+     * Check if we're running in phpunit >= 9.0
+     */
+    private static function isPhpunit9(): bool
```
greg0ire: So PHPUnit 10 "is phpunit 9" :thinking:

tests/Doctrine/Tests/DoctrineTestCase.php
```diff
@@ -5,10 +5,51 @@
...
+        string $pattern,
+        string $string,
+        string $message = ''
+    ): void {
+        // Forward-compatibility wrapper for phpunit 9 : can be removed once phpunit 8 / php 7.2 support is dropped.
+        if (self::isPhpunit9()) {
+            parent::assertDoesNotMatchRegularExpression($pattern, $string, $message);
+        } else {
+            self::assertNotRegExp($pattern, $string, $message);
+        }
```
greg0ire: Let's do feature detection instead?


```suggestion
    $methodName = method_exists(TestCase::class, 'assertDoesNotMatchRegularExpression')?
        'assertDoesNotMatchRegularExpression',
        'assertNotRegExp';
        
    return parent::$methodName($pattern, $string, $message);
```

tests/Doctrine/Tests/DoctrineTestCase.php
```diff
@@ -5,10 +5,51 @@
...
  * Base testcase class for all Doctrine testcases.
  */
 abstract class DoctrineTestCase extends TestCase
 {
+    public static function assertDoesNotMatchRegularExpression(
+        string $pattern,
+        string $string,
+        string $message = ''
+    ): void {
+        // Forward-compatibility wrapper for phpunit 9 : can be removed once phpunit 8 / php 7.2 support is dropped.
```
greg0ire: I don't like traits much, but maybe this would be a good use case, that way we don't repeat that comment, and removing the methods means removing a file.

tests/Doctrine/Tests/ORM/Functional/QueryCacheTest.php
```diff
@@ -59,7 +59,7 @@ public function testQueryCacheDependsOnHints(): Query
     /**
      * @param <type> $query
      *
-     * @depends testQueryCache_DependsOnHints
+     * @depends testQueryCacheDependsOnHints
```
greg0ire, line 62: > presumably these tests were renamed at some point for coding standards

Ah yes that does ring a bell :sweat_smile:

tests/Doctrine/Tests/ORM/Utility/HierarchyDiscriminatorResolverTest.php
```diff
@@ -22,15 +22,17 @@ public function testResolveDiscriminatorsForClass(): void
...
-        $em = $this->prophesize(EntityManagerInterface::class);
-        $em->getClassMetadata($classMetadata->name)
-            ->shouldBeCalled()
-            ->willReturn($classMetadata);
-        $em->getClassMetadata($childClassMetadata->name)
-            ->shouldBeCalled()
-            ->willReturn($childClassMetadata);
-
-        $discriminators = HierarchyDiscriminatorResolver::resolveDiscriminatorsForClass($classMetadata, $em->reveal());
+        $em = $this->getMockBuilder(EntityManagerInterface::class)->getMock();
```
greg0ire:
```suggestion
        $em = $this->createMock(EntityManagerInterface::class);
```

tests/Doctrine/Tests/ORM/Utility/HierarchyDiscriminatorResolverTest.php
```diff
@@ -44,12 +46,13 @@ public function testResolveDiscriminatorsForClassWithNoSubclasses(): void
         $classMetadata->name               = 'Some\Class\Name';
         $classMetadata->discriminatorValue = 'discriminator';
 
-        $em = $this->prophesize(EntityManagerInterface::class);
-        $em->getClassMetadata($classMetadata->name)
-            ->shouldBeCalled()
+        $em = $this->getMockBuilder(EntityManagerInterface::class)->getMock();
```
greg0ire:
```suggestion
        $em = $this->createMock(EntityManagerInterface::class);
```

greg0ire (commented):
Once again, great contribution, and great commit messages, I should write more like these 👍

greg0ire:
@acoulton if you don't have time to take care of this, just tell me, and I will carry it to the finish line :slightly_smiling_face:

acoulton:
@greg0ire sorry - had a crazy couple of weeks. I'll try to get back to this tomorrow.

tests/Doctrine/Tests/DoctrineTestCase.php
acoulton: Thinking about it more, maybe it'd be better to use class_alias and/or a conditional include so that the methods are only defined if running on the older phpunit version. Then there's no risk of conflict with any future phpunit version and the polyfills are isolated and easy to delete as above.

Conceptually with a trait it could be:

```
if (Version::series() === 8) {
  abstract class DoctrineTestCase extends TestCase {
    use Phpunit8ForwardCompatibilityPolyfill;
  }
} else {
  abstract class DoctrineTestCase extends TestCase {
  }
}
```

Although I'm not wild about swapping the `DoctrineTestCase` itself in case that needs methods added in future, would probably be better to do something at the `TestCase` level - maybe introduce an intermediate class a bit like this (not tested yet):

```php
// tests/Doctrine/Tests/PolyfillableTestCase.php
if (Version::series() === 8) {
  class_alias('Doctrine\Tests\PhpUnit8PolyfilledTestCase', 'Doctrine\Tests\PolyfillableTestCase');
} else {
  class_alias('PhpUnit\Framework\TestCase', 'Doctrine\Tests\PolyfillableTestCase');
}

// tests/Doctrine/Tests/DoctrineTestCase.php
abstract class DoctrineTestCase extends PolyfillableTestCase {
}
```

This has the advantage that you can easily delete the PHPUnit 8 polyfill when you drop support (presumably relatively soon, since it and PHP 7.2 are EOL?), but keep the mechanism so you can easily add e.g. a PHPUnit 9 polyfill when you upgrade to version 10. Yet is totally transparent (bar an extra class name in the hierarchy) for the latest supported version each time.

tests/Doctrine/Tests/DoctrineTestCase.php
acoulton: :) fair point

greg0ire:
No worries, take your time 👍

acoulton:
@greg0ire OK I think this is better - I've gone for a virtual parent class and class_alias so that the polyfilled methods are only present when required (and can easily be removed in future). What do you think?

The other comment outstanding is this:

> > These asserts are probably redundant (and are clearly not running in CI since the classes they referenced no longer exist).
> 
> On one hand, it would be great to check that these are accurate, on the other hand, it's also great to check that the code works when the `assert` statements are not run. Maybe we should pick one of the jobs and enable assertions on it…

I see now that there's quite a lot of `assert(` calls dotted around the test suite, looks like they were introduced to replace `*@var` comments as part of fixing coding standards.

In my opinion they're a bit redundant - there are ones like these that are only really testing behaviour of phpunit itself:

https://github.com/doctrine/orm/blob/5ac036de0273c6fc819d1e1835ab4f548dc24727/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC2359Test.php#L34-L37

Or that are testing the actual code but duplicate a PHPUnit `assertInstanceOf` on the next line:

https://github.com/doctrine/orm/blob/5ac036de0273c6fc819d1e1835ab4f548dc24727/tests/Doctrine/Tests/ORM/UnitOfWorkTest.php#L601-L604

IMO they don't add an enormous amount and introduce a bit of duplication / future maintenance overhead e.g. if phpunit renames classes as here, or if doctrine's classes / behaviour change.

But equally appreciate may be a fair tradeoff if they're necessary to allow static analysis etc to run.

I agree if you want/need to keep them it would be worth having one of the jobs (maybe the sqlite smoketest) run them to make sure they stay in sync with reality....

tests/Doctrine/Tests/DoctrineTestCase.php
greg0ire: Yes, that does look perfect, thanks 👍

greg0ire:
I just took a look, and it seems that assertions are supposed to be enabled: https://github.com/doctrine/orm/blob/5ac036de0273c6fc819d1e1835ab4f548dc24727/.github/workflows/continuous-integration.yml#L40

Maybe that setting is not sufficient?

greg0ire (approved):
Looks great to me regardless of the point about `assert` that could be handled separately 👍

ostrolucky (approved):

greg0ire:
Thanks @acoulton !

acoulton:
@greg0ire grand - I'd been going to squash those up for you or I'd have written better commit messages today ;)

I have looked again at the asserts and of course they *are* running. It's just that the ones I had to fix here were all this type:

```php
assert($metadataFactory instanceof ClassMetadataFactory || $metadataFactory instanceof PHPUnit_Framework_MockObject_MockObject); 
```

And of course in this case the mock is **always** an instance of the mocked class, so the or case never executes, so the fact the interface name has changed makes no difference.

I guess in reality what you / a static analyser need to know is that the object is an instance of the mocked class **and** an instance of the mock interface.  So it ought really to be two separate `asserts` rather than an `||` condition.

Equally, since `createMock` is already strictly typed to return a `MockObject` instance it might be enough just to assert it's the mocked class and remove the phpunit class refs from those asserts altogether.

But it's not really a big deal, maybe not worth bothering with unless a future PHPUnit renames them.

greg0ire:
`createMock` also comes with Psalm annotations that would make all asserts unneeded, provided we use a recent enough version of PHPUnit  https://github.com/sebastianbergmann/phpunit/blob/1b3675760e1a6d3048b53eb092022a67951beaa4/src/Framework/TestCase.php#L1292

I think we will in fact be forced to remove these assertions at some point when setting PHPStan and/or Psalm to a stricter level.

acoulton:
Oh nice - that's very clever. Looks like that's been in since 8.2 so should be present in all the supported versions already. I just checked and in PHPStorm at least that annotation is already understood so typehinting / autocomplete / clickthrough just works out of the box. Really useful to know. I guess maybe worth leaving the asserts till they cause a problem in case there are other IDEs that haven't caught up yet.

tests/Doctrine/Tests/DoctrineTestCase.php
beberlei: @acoulton i have simplified this into the `DoctrineTestCase` and using `__call` and `__callStatic` for #8613 for a few reasons:

- This is only about 1/3 of the code, and since we are "only in test code" here, we don't need to go over the top with the perfect abstraction. I prefer "fewer elements" when the solution is otherwise comparable.
- this is the same way we do it in DBAL, so code would be easier to copy/adapt/share
- its similarly easy to remove or extend this code when we change the PHPUnit requirements and since we will support several PHPUnit versions anyways, this kind of abstraction will always be present, having it in one place is better and `DoctrineTestCase` was introduced for this reason way back.

PR #8556, opened 2021-03-24
Fix fieldMapping phpdoc

VincentLanglet:
See https://github.com/doctrine/orm/pull/8286#discussion_r502907339

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -420,7 +420,23 @@ class ClassMetadataInfo implements ClassMetadata
      * - <b>'unique'</b> (string, optional, schema-only)
      * Whether a unique constraint should be generated for the column.
      *
-     * @psalm-var array<string, array{type: string, fieldName: string, columnName: string, inherited: class-string}>
+     * @psalm-var array<string, array{
+     *      type: string,
+     *      fieldName: string,
+     *      columnName?: string,
+     *      length?: int,
+     *      id?: bool,
+     *      nullable?: bool,
+     *      columnDefinition?: string,
+     *      precision?: int,
+     *      scale?: int,
+     *      unique?: string,
+     *      inherited?: class-string,
+     *      originalClass?: class-string,
+     *      originalField?: string,
+     *      quoted?: bool,
+     *      requireSQLConversion?: bool,
+     * }>
      */
     public $fieldMappings = [];
 


VincentLanglet:
Friendly ping @greg0ire for the review :)

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -420,7 +420,19 @@ class ClassMetadataInfo implements ClassMetadata
...
+     *      columnName?: string,
+     *      length?: int,
+     *      id?: bool,
+     *      nullable?: bool,
+     *      columnDefinition?: string,
+     *      precision?: int,
+     *      scale?:int,
+     *      unique?: string,
+     *      inherited?: class-string
+     * }>
```
greg0ire, line 439: Do you think this list is exhaustive now? If not please do this:
```suggestion
     * }>&array<string,mixed>
```

greg0ire:
@orklah please review.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
VincentLanglet: I don't know at all, I just followed the phpdoc:
```
     * The mapping definition array has the following values:
     *
     * - <b>fieldName</b> (string)
     * The name of the field in the Entity.
     *
     * - <b>type</b> (string)
     * The type name of the mapped field. Can be one of Doctrine's mapping types
     * or a custom mapping type.
     *
     * - <b>columnName</b> (string, optional)
     * The column name. Optional. Defaults to the field name.
     *
     * - <b>length</b> (integer, optional)
     * The database length of the column. Optional. Default value taken from
     * the type.
     *
     * - <b>id</b> (boolean, optional)
     * Marks the field as the primary key of the entity. Multiple fields of an
     * entity can have the id attribute, forming a composite key.
     *
     * - <b>nullable</b> (boolean, optional)
     * Whether the column is nullable. Defaults to FALSE.
     *
     * - <b>columnDefinition</b> (string, optional, schema-only)
     * The SQL fragment that is used when generating the DDL for the column.
     *
     * - <b>precision</b> (integer, optional, schema-only)
     * The precision of a decimal column. Only valid if the column type is decimal.
     *
     * - <b>scale</b> (integer, optional, schema-only)
     * The scale of a decimal column. Only valid if the column type is decimal.
     *
     * - <b>'unique'</b> (string, optional, schema-only)
     * Whether a unique constraint should be generated for the column.
```

The `inherited` key already added wasn't in the phpdoc.
I don't know if someone can confirm there is nothing else... If nobody can, I will add your suggestion.

orklah: It's definitely incomplete. I think I added more in the 2.9.x branch:
https://github.com/doctrine/orm/blob/2.9.x/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L440

Also this PR should supersede https://github.com/doctrine/orm/pull/8302 (which was already kinda superseded by the PR on 2.9.x)

orklah: https://github.com/doctrine/orm/pull/8302#issuecomment-706965562 this one should be considered too

VincentLanglet: I have added 
```
originalClass?: class-string,
originalField?: string
```
from the 2.9 branch

And also
```
quoted?: bool
requireSQLConversion?: bool
```

I found nothing else in the code

orklah (approved):

greg0ire (approved):

greg0ire:
Thanks @VincentLanglet !

PR #8566, opened 2021-03-26
SimpleObjectHydrator: skip unsuit custom type before convert it

### LINKED ISSUE #8565: SimpleObjectHydrator: skip unsuit custom type before convert it
SimpleObjectHydrator tries to convert the value, and only then checks if we need it.
This results in an error if our custom type does not get expected data to initialize php value.

https://github.com/doctrine/orm/blob/c8f2f61ea14e05c8f9a0da23467f3ca6a07d0173/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php#L124-L156

This can be fixed if we check if it is the correct type before converting it
```php 
foreach ($row as $column => $value) {
    // An ObjectHydrator should be used instead of SimpleObjectHydrator
    if (isset($this->_rsm->relationMap[$column])) {
        throw new Exception(sprintf('Unable to retrieve association information for column "%s"', $column));
    }

    $cacheKeyInfo = $this->hydrateColumnInfo($column);

    if (! $cacheKeyInfo) {
        continue;
    }

    // If we have inheritance in resultset, make sure the field belongs to the correct class
    if (isset($cacheKeyInfo['discriminatorValues']) && ! in_array((string) $discrColumnValue, $cacheKeyInfo['discriminatorValues'], true)) {
        continue;
    }

    // Check if value is null before conversion (because some types convert null to something else)
    $valueIsNull = $value === null;


    // Convert field to a valid PHP value
    if (isset($cacheKeyInfo['type'])) {
        $type  = $cacheKeyInfo['type'];
        $value = $type->convertToPHPValue($value, $this->_platform);
    }

    $fieldName = $cacheKeyInfo['fieldName'];

    if (! isset($data[$fieldName]) || ! $valueIsNull) {
        $data[$fieldName] = $value;
    }
}
```

PR with fix #8566

tweet9ra:
SimpleObjectHydrator tries to convert the value, and only then checks if we need it.
This results in an error if our custom type does not get expected data to initialize php value.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Hydration/SimpleObjectHydratorTest.php
+++ b/tests/Doctrine/Tests/ORM/Hydration/SimpleObjectHydratorTest.php
@@ -4,11 +4,17 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\ORM\Hydration;
 
+use Doctrine\DBAL\Types\Type as DBALType;
 use Doctrine\ORM\Internal\Hydration\SimpleObjectHydrator;
 use Doctrine\ORM\Query\ResultSetMapping;
+use Doctrine\Tests\DbalTypes\GH8565EmployeePayloadType;
+use Doctrine\Tests\DbalTypes\GH8565ManagerPayloadType;
 use Doctrine\Tests\Mocks\HydratorMockStatement;
 use Doctrine\Tests\Models\CMS\CmsAddress;
 use Doctrine\Tests\Models\Company\CompanyPerson;
+use Doctrine\Tests\Models\GH8565\GH8565Employee;
+use Doctrine\Tests\Models\GH8565\GH8565Manager;
+use Doctrine\Tests\Models\GH8565\GH8565Person;
 use Doctrine\Tests\Models\Issue5989\Issue5989Employee;
 use Doctrine\Tests\Models\Issue5989\Issue5989Manager;
 use Doctrine\Tests\Models\Issue5989\Issue5989Person;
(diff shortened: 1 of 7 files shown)


ehunov (dismissed):

greg0ire:
I have repaired the test suite, let's try this again.

tests/Doctrine/Tests/DbalTypes/Issue8565EmployeePayloadType.php
```diff
@@ -0,0 +1,23 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DbalTypes;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\JsonType;
+use Exception;
+
+class Issue8565EmployeePayloadType extends JsonType
```
greg0ire: We usually use `GH` instead of `Issue` here. There is no guarantee that we aren't migrate to something else than Github in the future and that's a risk of collision.

tests/Doctrine/Tests/ORM/Hydration/SimpleObjectHydratorTest.php
```diff
@@ -122,4 +128,36 @@ public function testNullValueShouldNotOverwriteFieldWithSameNameInJoinedInherita
...
+        $rsm->addFieldResult('p', 'p__id', 'id');
+        $rsm->addFieldResult('p', 'm__type', 'type', Issue8565Manager::class);
+        $rsm->addFieldResult('p', 'e__type', 'type', Issue8565Employee::class);
+        $rsm->addMetaResult('p', 'discr', 'discr', false, 'string');
+        $rsm->setDiscriminatorColumn('p', 'type');
+        $resultSet = [
+            [
+                'p__id'   => '1',
+                'm__type' => 'type field',
+                'e__type' => 'type field',
```
greg0ire, line 148: To make sure I understand the PR: here we have a field with an incorrect value in the result set, but since we are going to build a manager object, this field should not even be examined, is that correct?
If yes, why is that field even present in the result set in the first place? And why does it have a value instead of something like `null`? It make no sense for a manager to have the employee `type` field filled in, right?

ehunov: If there is a statement like this:

```sql
SELECT p0.id   AS id_1,
       p0.type AS type_2,
       p0.type AS type_3,
       p0.type AS type_4
FROM person p0
WHERE p0.discr IN ('person', 'manager', 'employee');
```

In result set mapping will appear:

```php
$fieldMappings = [
    'type_2' => 'type',
    'type_3' => 'type',
    'type_4' => 'type',
];
```

So field has a value because there is only **one column** that has **different types for different entities**, and that's why it's not `null`.

tests/Doctrine/Tests/DbalTypes/Issue8565EmployeePayloadType.php
tweet9ra: Fixed

ehunov (approved):

ehunov:
@greg0ire, hi! any updates?

tests/Doctrine/Tests/ORM/Hydration/SimpleObjectHydratorTest.php
greg0ire: @tweet9ra please don't mark conversations you did not open as resolved, what if I didn't understand the comment above?
@toshksenof thanks for your comment, I will try to understand it later, right now I'm a bit too tired for that I'm afraid. Or maybe the people I have added as reviewers will beat me to it.

ostrolucky (approved):

tests/Doctrine/Tests/ORM/Hydration/SimpleObjectHydratorTest.php
greg0ire: After a good night's sleep, I think I finally got it:
Doctrine will select the same column several times, one for each mapping in each subclass.

greg0ire:
I squash merged with a commit message that contains a more complete explanation, thanks to you both!
Next time, try to insist more on what is peculiar here (having one column mapped to 2 properties that have the same name but that are declared in 2 different child classes).

PR #8569, opened 2021-03-26
Add ignoreDuplicates option for ManyToMany relations

malarzm:
Background: our e-commerce has something called "manual categories" where operators upload a literal ton of products (like 70k). From mapping's perspective it's a simple ManyToMany relation. The process is straightforward for the operators, they upload a CSV file with a list of product codes they want to have in said category. Uploaded file gets chunked into parts and then digested asynchronously. Sadly, it happens that there are either duplicates in different chunks, or people are too eager and are uploading same file twice. Anyhow, we end up having to check whether a product was already associated (which is slow and generates a lot of queries) and even that we can't be sure that there is no concurrent worker trying to do something similar. If there is, we get a `Duplicate entry` exception thrown and entire processed batch needs to be redone. We'd be more than happy to let ORM know that we don't give a flying duck about duplicates, we're just happy the association is there and so should be ORM. Effectively this means using `INSERT IGNORE` (or other flavours) for insertion to the join table.

Out of all platforms ORM supports officially it seems that Oracle and Microsoft SQL Server don't have an easy way of doing `INSERT IGNORE` but rather some wierd `MERGE` thingies. Since the feature is an opt-in we could throw a runtime error when `ignoreDuplicates` is used, or maybe there is an option to validate mapping against used platform. I'll appreciate an opinion on where to go with this.

Final note: Ignoring duplicates is crafted specifically for ManyToMany relationship and I have no intention of moving it elsewhere. It is done this way precisely because there is no use for `ON CONFLICT` clause and other things on a ManyToMany's join table as there are only two foreign keys - you can either have them inserted or not, no other fields to update.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Builder/ManyToManyAssociationBuilder.php
+++ b/lib/Doctrine/ORM/Mapping/Builder/ManyToManyAssociationBuilder.php
@@ -46,6 +46,18 @@ class ManyToManyAssociationBuilder extends OneToManyAssociationBuilder
     }
 
     /**
+     * @param bool $name
+     *
+     * @return static
+     */
+    public function setIgnoreDuplicates($ignoreDuplicates)
+    {
+        $this->mapping['ignoreDuplicates'] = $ignoreDuplicates;
+
+        return $this;
+    }
+
+    /**
      * Adds Inverse Join Columns.
      *
      * @param string      $columnName
--- a/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
@@ -498,8 +498,13 @@ class ManyToManyPersister extends AbstractCollectionPersister
             $types[]   = PersisterHelper::getTypeOfColumn($joinColumn['referencedColumnName'], $targetClass, $this->em);
         }
 
+        $ignore = '';
+        if ($mapping['ignoreDuplicates']) {
+            $ignore = 'IGNORE'; // @todo
+        }
+
         return [
-            'INSERT INTO ' . $this->quoteStrategy->getJoinTableName($mapping, $class, $this->platform)
+            'INSERT ' . $ignore . ' INTO ' . $this->quoteStrategy->getJoinTableName($mapping, $class, $this->platform)
             . ' (' . implode(', ', $columns) . ')'
             . ' VALUES'
             . ' (' . implode(', ', array_fill(0, count($columns), '?')) . ')',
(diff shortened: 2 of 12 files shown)


lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
```diff
@@ -498,8 +498,13 @@ protected function getInsertRowSQL(PersistentCollection $collection)
             $types[]   = PersisterHelper::getTypeOfColumn($joinColumn['referencedColumnName'], $targetClass, $this->em);
         }
 
+        $ignore = '';
+        if ($mapping['ignoreDuplicates']) {
+            $ignore = 'IGNORE'; // @todo
```
malarzm, line 503: I'll appreciate guidance here: I assume I need to open a PR for DBAL that will allow platforms specyfying the keyword that needs to be used (or return `null` if it's not supported)?

lib/Doctrine/ORM/Mapping/Builder/ManyToManyAssociationBuilder.php
```diff
@@ -45,6 +45,18 @@ public function setJoinTable($name)
         return $this;
     }
 
+    /**
+     * @param bool $name
```
greg0ire, line 49: That variable name looks wrong and anyway, you can use native type declarations here :)

lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
greg0ire: Maybe yeah… I see that there is a `AbstractPlatform::getForUpdateSQL()` method that looks designed that way.

lib/Doctrine/ORM/Mapping/Builder/ManyToManyAssociationBuilder.php
malarzm: Right, the name is off 👍 For types I was trying to follow how other methods look like, but I can surely add type here if that'll be the consensus :)

lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
malarzm: So looking at https://github.com/doctrine/dbal/blob/012569dfcc96478c061d3b529fecb5c8ac0405e5/src/Platforms/SqlitePlatform.php#L626-L629 the method would return empty string if the platform does not support `INSERT IGNORE` and generated query would be a casual `INSERT`, right?

greg0ire: Yup, sounds right 👍

lib/Doctrine/ORM/Mapping/Builder/ManyToManyAssociationBuilder.php
greg0ire: Yeah IMO we are going to migrate to types at some point, so I think we should make that as less work as possible.

PR #8571, opened 2021-03-27
Added strict PHP parameter and return types

satalaondrej:
Added strict parameter and return types to Persiters and SQLFilter


beberlei:
The change on SQLFilter is a BC break and the ones on persisters can wait until 3.0, i don't think these changes make sense st the moment and for 2.8 branch

satalaondrej:
> The change on SQLFilter is a BC break and the ones on persisters can wait until 3.0, i don't think these changes make sense st the moment and for 2.8 branch

I am closing this pull request and opening new one for 3.0.x branch.

PR #8572, opened 2021-03-27
3.0.x Added strict PHP parameter and return types

satalaondrej:
Added strict parameter and return types to Persiters and SQLFilter


beberlei:
While I appreciate the direction and your work on this PR, it would be better if we go about this strategically and in one or a few big steps and with automated tooling like rector instead of individually on certain files. Otherwise this potential change for 3.0 will take up too much mental space for reviewers. I prefer if we discuss the strategy in an issue first and decide on an approach before starting with actual changes.

Issue #8575, opened 2021-03-30
Exception when calling toIterable without parameters


flack:
I'm calling Doctrine\ORM\Query::toIterable() without any parameters like this:

```php
$resultset = $query->toIterable();
```

This produces the following exception:

```
TypeError: Argument 1 passed to Doctrine\ORM\Internal\Hydration\AbstractHydrator::toIterable() must implement interface Doctrine\DBAL\Driver\Statement, instance of Doctrine\DBAL\ForwardCompatibility\Result given, called in vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php on line 1010
```

The code used to work fine with the `iterate()` method, but I wanted to replace it with the new version, since it's deprecated. If I downgrade `doctrine/dbal` to 2.12.1, everything works. With 2.13.0 it breaks. Not sure if this means I should rather open a ticket for doctrine/dbal?

derrabus:
I have the same issue here. This is a blocker for upgrading to DBAL 2.13 for us.

beberlei:
Fixed with https://github.com/doctrine/orm/pull/8495 already, not released yet - will be in 2.8.3

flack:
@beberlei awesome, thanks for the quick fix!

PR #8577, opened 2021-03-31
Bump requirement to DBAL 2.13

beberlei:
This updates Doctrine DBAL requirement to 2.13 - because of the PHP 7.1 dependency and to prepare for first steps to migrate to new from deprecated APIs.


flack:
It would be good if #8575 gets sorted out before this is merged

beberlei:
@flack maybe, but i decide what i want to work on in my free time.

PR #8586, opened 2021-04-02
Additional psalm param typehint for orderBy argument in findBy method

KartaviK:
Because of `findBy([], ['somField' => 'DESC']...);`

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -185,7 +185,7 @@ class EntityRepository implements ObjectRepository, Selectable
      * @param int|null $offset
      *
      * @psalm-param array<string, mixed> $criteria
-     * @psalm-param list<string>|null $orderBy
+     * @psalm-param array<string, string>|null $orderBy
      * @psalm-return list<T> The objects.
      */
     public function findBy(array $criteria, ?array $orderBy = null, $limit = null, $offset = null)
--- a/lib/Doctrine/ORM/Persisters/Entity/EntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/EntityPersister.php
@@ -273,8 +273,8 @@ interface EntityPersister
      * @param int|null $limit
      * @param int|null $offset
      *
-     * @psalm-param list<string>|null $orderBy
-     * @psalm-param array<string, mixed> $criteria
+     * @psalm-param array<string, string>|null $orderBy
+     * @psalm-param array<string, mixed>       $criteria
      */
     public function loadAll(array $criteria = [], ?array $orderBy = null, $limit = null, $offset = null);
 


lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -185,7 +185,7 @@ public function findAll()
      * @param int|null $offset
      *
      * @psalm-param array<string, mixed> $criteria
-     * @psalm-param list<string>|null $orderBy
+     * @psalm-param list<string>|array<string, string>|null $orderBy
```
greg0ire: Good catch, my bad! Should `list<string>` be removed? Also, `EntityPersister::loadAll()` seems to have the same issue, and `EntityPersister::getSelectSQL()` could be improved.

KartaviK: If `list<string>` can not be, so it can be removed

KartaviK:
```suggestion
     * @psalm-param array<string, string>|null $orderBy
```

KartaviK:
@greg0ire Updated!

lib/Doctrine/ORM/EntityRepository.php
KartaviK: Should i update `EntityPersister::loadAll()`?

greg0ire: Yes please, and `EntityPersister::getSelectSQL()` too please :pray:

greg0ire (approved):

greg0ire:
@orklah @vladyslavstartsev @simonberger please review, and please also tell me if you would rather I don't ping you for such reviews.

orklah:
I don't mind being pinged :)

Ideally, this [line](https://github.com/doctrine/persistence/blob/2.1.x/lib/Doctrine/Persistence/ObjectRepository.php#L42) in doctrine/persistence should be updated to match `array<string, string>` accordingly as well.

It could be the occasion to update some other case of $orderBy as well like in AbstractEntityPersister::getHash

greg0ire:
Maybe we could go even further, see https://github.com/doctrine/persistence/pull/165 ?
I looked for parameters called `$orderBy` in the whole project, here is the list of signatures that should be updated:
- `EntityRepository::findBy` (wrong)
- `EntityRepository::findOneBy` (imprecise)
- `AbstractEntityPersister::getHash` (imprecise)
- `EntityPersister::getSelectSQL` (imprecise)
- `EntityPersister::load` (imprecise)
- `EntityPersister::loadAll` (wrong)

Maybe this PR should only fix what's wrong so `EntityRepository::findBy` and `EntityPersister::loadAll`, and the rest should be improved in a PR targeting 2.9.x

greg0ire (dismissed):

greg0ire:
I updated the PR according to that plan.

greg0ire (approved):

greg0ire:
Thanks @KartaviK !

PR #8589, opened 2021-04-04
Fix typed properties for default metadata (#7939)

Lustmored:
This is continuation of work from #7939. Tests for real life scenarios were missing therefore inheritance from typed properties wasn't running with metadata drivers.

This PR introduces such tests, adds metadata to test object for all drivers and makes small changes in drivers to adapt. Most needed changes were changing default value to null or removing `required` from not required XML attributes.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1763,7 +1763,7 @@ class ClassMetadataInfo implements ClassMetadata
                     [
                         'name' => $this->namingStrategy->joinColumnName($mapping['fieldName'], $this->name),
                         'referencedColumnName' => $this->namingStrategy->referenceColumnName(),
-                        'nullable' => true
+                        'nullable' => true,
                     ],
                 ];
             }
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1642,6 +1642,13 @@ class ClassMetadataInfo implements ClassMetadata
 
         if ($this->isTypedProperty($mapping['fieldName'])) {
             $mapping = $this->validateAndCompleteTypedAssociationMapping($mapping);
+        } elseif (isset($mapping['joinColumns'])) {
+            foreach ($mapping['joinColumns'] as &$joinColumn) {
+                if (! isset($joinColumn['nullable'])) {
+                    $joinColumn['nullable'] = true;
+                }
+            }
+            unset($joinColumn);
         }
 
         if (isset($mapping['targetEntity'])) {
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1642,14 +1642,6 @@ class ClassMetadataInfo implements ClassMetadata
 
         if ($this->isTypedProperty($mapping['fieldName'])) {
             $mapping = $this->validateAndCompleteTypedAssociationMapping($mapping);
-        } elseif (isset($mapping['joinColumns'])) {
-            foreach ($mapping['joinColumns'] as &$joinColumn) {
-                if (! isset($joinColumn['nullable'])) {
-                    $joinColumn['nullable'] = true;
-                }
-            }
-
-            unset($joinColumn);
         }
 
         if (isset($mapping['targetEntity'])) {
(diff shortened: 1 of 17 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1642,6 +1642,14 @@ protected function _validateAndCompleteAssociationMapping(array $mapping)
 
         if ($this->isTypedProperty($mapping['fieldName'])) {
             $mapping = $this->validateAndCompleteTypedAssociationMapping($mapping);
+        } elseif (isset($mapping['joinColumns'])) {
+            foreach ($mapping['joinColumns'] as &$joinColumn) {
```
beberlei: I believe this will not work here, because `_validateAndCompleteAssociationMapping` is called before the code for setting default joinColumns is called in `_validateAndCompleteOneToOneMapping`.

I believe, you need to put this code into `_validateAndCompleteOneToOneMapping`, it is also used for ManyToOne, and those two assocations are the only two that we are interested in completing anyways.

Otherwise it will not work for the case where you only specify `@ManyToOne` but not a corresponding `@JoinColumn`.

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -327,10 +327,6 @@ public function loadMetadataForClass($className, ClassMetadata $metadata)
             // @Column, @OneToOne, @OneToMany, @ManyToOne, @ManyToMany
             $columnAnnot = $this->reader->getPropertyAnnotation($property, Mapping\Column::class);
             if ($columnAnnot) {
-                if ($columnAnnot->type === null) {
-                    throw MappingException::propertyTypeIsRequired($className, $property->getName());
```
beberlei, line 331: is this method still used anywhere else? otherwise we can `@deprecated` it in `MappingException`.

tests/Doctrine/Tests/Models/TypedProperties/UserTyped.php
```diff
@@ -7,46 +7,121 @@
...
 
     /** @Column */
+    #[ORM\Column]
     public float $float;
 
     /**
      * @OneToOne(cascade={"persist"}, orphanRemoval=true)
      * @JoinColumn
      */
+    #[ORM\OneToOne(cascade: ["persist"], orphanRemoval: true), ORM\JoinColumn]
```
beberlei: Please add another `ManyToOne` association to another email that does not define a join column.

beberlei (changes requested):
This is great work, thank you very much! I have only a few comments.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
Lustmored: You are right and adding tests for `@ManyToOne` without `joinColumn` uncovered it. I have fixed it exactly as suggested.

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
Lustmored: It seems it is not. I have added `@deprecated`

tests/Doctrine/Tests/Models/TypedProperties/UserTyped.php
Lustmored: Great suggestion - it have shown a case not covered before and failing due to `ManyToOne` constructor dependency and wrong place of `nullable` checks. Fixed :)

Lustmored:
I have rebased and fix mentioned problems. Some tests now fail (despite showing `All checks have passed` here - take a look in checks tab) but it's withing cascade operation tests so I assume it's out of the scope of this PR and some change in `2.9.x` branch probably causes them.

beberlei:
@Lustmored that is a serious heisenbug, can reproduce locally unless I poke Xdebug at it, then it succeeds suddenly. Will investigate. It fails on 2.9.x as well.

beberlei:
For some unclear reason, sometimes the CommitOrderCalculator changes the order of `Entity0` and `EntityG` in the `CascadeRemoveOrderTest`. This leads to the foreign key failure during commit. The problem is two relationships in both directions, so each entity has a foreign key to the other. The CommitOrderCalculactor does not seem to prefer the order with the join column that has `onDelete`=`SET NULL`. This looks like a bug in the CommitOrderCalculator, however why this suddenly starts happening only in this branch is confusing.

beberlei:
Ah of course it happens because of this change: https://github.com/doctrine/orm/pull/8589/files#diff-20cad4158379b4e6d3f30045b627fd986d0b238212c8b0318bbfca13d09f0d3dR44

beberlei (approved):

PR #8591, opened 2021-04-05
Return case insensitive check

DmitriiBezborodnikov:
Return insensitive check after #8453 

Problem: `->andWhere("u.name = ?1 or u.username = ?1");` did not wrap part in parenthesis when `or` or `and` was written in lowercase anymore. It still worked for uppercase `OR` and `AND`.

Fixes #8595

PR #8594, opened 2021-04-05
Ensure test is suffixed with Test

greg0ire:
In #8591 , I noticed that some tests were not run, and found an exhaustive list of tests that need to be renamed. Some did not require much modifications to run again, but that last one does not pass in the CI, although it passes locally on my machine.

The assertion that fails is about an entity fetched through a one-to-one relationship. It is not supposed to be a proxy but on the CI, it seems to be.

The test that fails looks like this: https://github.com/doctrine/orm/blob/7de84537f61129a96658e50799131dec779c55bd/tests/Doctrine/Tests/ORM/Functional/ClassTableInheritanceTest2.php#L38-L62

The failure looks like this: 

```
1) Doctrine\Tests\ORM\Functional\ClassTableInheritanceSecondTest::testOneToOneAssocToBaseTypeBidirectional
Failed asserting that Doctrine\Tests\Proxies\__CG__\Doctrine\Tests\ORM\Functional\CTIChild Object (...) is not an instance of interface "Doctrine\ORM\Proxy\Proxy".
```

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/ClassTableInheritanceSecondTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/ClassTableInheritanceSecondTest.php
@@ -6,7 +6,6 @@ namespace Doctrine\Tests\ORM\Functional;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
-use Doctrine\ORM\Proxy\Proxy;
 use Doctrine\Tests\OrmFunctionalTestCase;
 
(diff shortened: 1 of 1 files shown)


stof:
why would we enforce that it is not a proxy ? Instantiating a proxy object or the entity class itself is internal details of the hydrator to me. It could instantiate a proxy even if it initializes it directly.

greg0ire:
Ok, works for me, I'll remove that assertion then 👍

ostrolucky (approved):

PR #8600, opened 2021-04-07
Remove internal tag from computeChangeSet

VincentLanglet:
In the doctrine https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/events.html#onflush

The `computeChangeSet` is recommended to be used
```
If you create and persist a new entity in onFlush, then calling EntityManager#persist() is not enough. You have to execute an additional call to $unitOfWork->computeChangeSet($classMetadata, $entity).
Changing primitive fields or associations requires you to explicitly trigger a re-computation of the changeset of the affected entity. This can be done by calling $unitOfWork->recomputeSingleEntityChangeSet($classMetadata, $entity).
```

So the method shouldn't be flag as internal:
- To provide a BC-promise
- To avoid psalm error with internal method

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -629,8 +629,6 @@ class UnitOfWork implements PropertyChangedListener
      * If a PersistentCollection has been de-referenced in a fully MANAGED entity,
      * then this collection is marked for deletion.
      *
-     * @internal Don't call from the outside.
-     *
      * @param ClassMetadata $class  The class descriptor of the entity.
      * @param object        $entity The entity for which to compute the changes.
      *


VincentLanglet:
Friendly ping @greg0ire, what do you think about this PR ?  :)

greg0ire:
It makes sense to me 👍
Can you find the author of this comment and ping them if they are still part of the team?

greg0ire (approved):

VincentLanglet:
It was @beberlei https://github.com/doctrine/orm/commit/6f3667201c981be7e4341435ddde5850edd3116a

greg0ire:
Ok, I think the doc you linked to was already present, so this must be a mistake.

ostrolucky (approved):

greg0ire:
Thanks @VincentLanglet !

PR #8602, opened 2021-04-07
Adjusted return type annotation for getOriginalEntityData

### LINKED ISSUE #8599: Wrong return type for UnitOfWork::getOriginalEntityData
In #8521, a new return type for `getOriginalEntityData` has been introduced (see https://github.com/doctrine/orm/blob/2.8.x/lib/Doctrine/ORM/UnitOfWork.php#L2992). `array<string, array<string, mixed>>` does not look correct to me: when calling that method, it returns `array<string, mixed>`

NicoHaase:
Fixes #8599

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2989,7 +2989,7 @@ class UnitOfWork implements PropertyChangedListener
      *
      * @param object $entity
      *
-     * @psalm-return array<string, array<string, mixed>>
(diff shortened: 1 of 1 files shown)


greg0ire (approved):

greg0ire:
Thanks @NicoHaase !

soullivaneuh:
@greg0ire May we please have a fix release for this? This would make PHPStan less crazy. :wink:

Thanks! :wave: :beers:

PR #8613, opened 2021-04-10
Reintroduce PHP 7.1 support

beberlei:
This reintroduces PHP 7.1 support, since the code still works (except a few auto fixed `object` typehints that are on private or unoverwritable methods).


tests/Doctrine/Tests/OrmTestCase.php
```diff
@@ -168,4 +168,43 @@ protected function getSharedSecondLevelCacheDriverImpl(): Cache
...
+
+    /**
+     * @param array<mixed> $arguments
+     *
+     * @return mixed
+     */
+    public static function __callStatic(string $method, array $arguments)
+    {
+        if ($method === 'assertMatchesRegularExpression') {
+            self::assertRegExp(...$arguments);
```
greg0ire: This one is already covered here: https://github.com/doctrine/orm/blob/e2e9f8fa97a2e8b6ad79f1b2f68c921262677cc1/tests/Doctrine/Tests/PHPUnit8PolyfilledTestCase.php#L19

That's super recent. Let's do similar things for `createStub`?

beberlei: why is this an extra test case? this would require to change all tests that need the polyfill, `OrmTestCase` is alrady the right abstraction that was added for this.

greg0ire: The reasoning is in this comment: https://github.com/doctrine/orm/pull/8552#discussion_r607402830

TL;DR : having it in a separate class makes it easy to drop support for PHPUnit 8 once we can do so: we delete the class and remove the class alias.

In your case, since you are adding support for PHPUnit 7, I think you should actually introduce a `PHPUnit7PolyfilledTestCase` that extends `PHPUnit8PolyfilledTestCase`, put things that are missing on PHPUnit 7 in it (so `createStub`?), and add an if branch here: https://github.com/doctrine/orm/blob/e2e9f8fa97a2e8b6ad79f1b2f68c921262677cc1/tests/Doctrine/Tests/PolyfillableTestCase.php#L18-L20

beberlei: I do not think its a good argument, because you need to modify the base classes of many tests to change the polyfill. So its not really just deleting a file. Putting it into OrmTestCase isolates this code in one location and does not spread knowledge across the whole code base

greg0ire: :thinking: the only test that changed its base class was `DoctrineTestCase`, see https://github.com/doctrine/orm/pull/8552/files

beberlei: Ah you are right, that makes sense i thought the hierachy was `PoliyfillTestCase` => `OrmTestCase`, but its `OrmTestCase` -> `DoctrineTestCase` -> `PolyfillTestCase`. That is better and i rework this.

acoulton:
@beberlei thanks for tagging me in #8552.

To be honest I'm not sure of the need for this? PHP 7.1 has been EOL for almost a year and a half, and nearly 2 and a half years since it was actively supported. I'm not aware of it being maintained in any LTS distros either, so people really shouldn't be still using it for anything that matters...

And since Doctrine is pretty stable, even if there is a minority of users determined to stay on 7.1 for a while, it doesn't seem likely there are any big bugs they've not already found, so they could just stick with whatever version of doctrine they're already using? If you were determined to offer security support, you could still backport specific fixes to a 2.7.x release if something comes up.

Likewise PHPUnit 7 ended life support a year ago. Sooner or later there are bound to be more conflicts with newer phpunit releases that someone will have to fix / polyfill.

My $0.02 is reverting to 7.1 support adds no real benefit, removes improvements that are already present, blocks a range of improvements that might be made in the future, and will before long create a maintenance overhead / headache for you all.

One thing I did notice is you haven't updated any of the CI build matrix to run tests on 7.1. 

That will obviously add a bunch more jobs - arguably there should be a 7.1 version of all the current ones, not just the smoke tests. 7.1 is so old that it's almost guaranteed someone will eventually put through a commit that uses >= 7.2 syntax without realising, so unless you have maximum possible test coverage on the dead version there's a fair risk of shipping a broken release. Obviously more jobs = more noise, more risk of encountering a runtime / flickering test failure on one of them, slower feedback if you hit the github actions concurrent job limit across the doctrine org.

It's your project, obviously, so not really any of my business. But if it was my project I 100% wouldn't do it. In fact if it was my project I'd be strongly considering dropping 7.2 support very soon as well - though I know that's in e.g. Ubuntu 18.04 LTS so slightly less dead.

beberlei:
@acoulton  i just wanted to notify you about the change in phpunit polyfills, not about PHP 7.1 in particular.

tests/Doctrine/Tests/DoctrineTestCase.php
```diff
@@ -4,9 +4,59 @@
...
+    /**
+     * @param array<mixed> $arguments
+     *
+     * @return mixed
+     */
+    public static function __callStatic(string $method, array $arguments)
+    {
+        if ($method === 'assertMatchesRegularExpression') {
+            return self::assertRegExp(...$arguments);
+        } elseif ($method === 'assertDoesNotMatchRegularExpression') {
```
acoulton: On the specific thing of how you support older phpunit versions, my thoughts are:

* Having to duplicate these stubs between `__callStatic` and `__call` is not ideal
* It's not at all clear which phpunit versions these are needed for, so although it's *possible* to remove them by editing a single file it's fairly likely one or both *won't* actually be removed when a given phpunit version is dropped. That could be helped by at least commenting each one to document the phpunit version they're for.
* It boxes you into a corner if a future phpunit release includes something that can't be fixed just by proxying a method with the same argument sequence.

More importantly, having the polyfill present at runtime in newer phpunit versions increases the risk for confusion & unexpected behaviour in future. When you eventually update to e.g. phpunit 10, it will be much easier to find all the things that need to be updated by iterating till you get a green run on the new version, then adding polyfills on the old versions if / as required. Otherwise there's a high chance some method calls in tests are missed, and don't break till you drop the old version some years later when people will have forgotten what they need to be changed to. 

It also allows contributors to add tests with outdated syntax, which just adds to the work of dropping the old version / polyfill eventually.

Or - worst case - the polyfill doesn't do what you expect (at first, or over time), so you get false positives on tests that should be failing, and would be if they were calling the (tested) phpunit methods directly.

acoulton:
> @acoulton i just wanted to notify you about the change in phpunit polyfills, not about PHP 7.1 in particular.

@beberlei understood, see my comment on that. But since then I saw this, just wanted to stick my oar in and let you know I disagree. :wink: There's always someone wrong on the internet :laughing:

beberlei:
@acoulton you are right about php 7.1 missing in the CI, it was missing since a dependency still needed update. I tested it locally.

As for your PHP 7.1 comments, we still see people locked on old ORM that last supported 7.1. While we did raise the PHP requirement, you can see from the handful of `object` typehint only changes, that the codebase is still working perfectly on 7.1. This does not block supporting any new features at all, we have support for typed properties, union types et al in ORM, its easy to support this with with the current code base. None of the PHP 7.2 - 8.1 features is a killer feature that improves ORM code base maintenance. We will be upgrading PHP dependency with ORM 3, but for 2.x since nothing much happens, we are conservative, since we start triggering deprecations, we want as many users as possible to see them.

greg0ire:
@beberlei this brand new line will have to be changed too: https://github.com/doctrine/orm/blob/da3a9fa3614c353608497c97827075494b7f7743/phpstan.neon#L11

beberlei:
@greg0ire it needs to be upmerged to 2.9.x first

greg0ire:
Ah sure, I'm on it!

beberlei:
@greg0ire done, no changes in phpstan output, all green.

greg0ire (dismissed):
Looks good to me, one things looks weird though, even before this PR: 

https://github.com/doctrine/orm/blob/da3a9fa3614c353608497c97827075494b7f7743/.github/workflows/continuous-integration.yml#L26

Why is that line 7.3 and not 7.2 right now? Should it be 7.1?

beberlei:
@greg0ire good question, `composer --prefer-lowest` with 7.1 works for me locally, so that is good. Should prefer lowest always run with the lowest supported php version?

greg0ire:
> Should prefer lowest always run with the lowest supported php version?

IMO yes: 7.1 is where you have the best chances of having the truly lowest versions.

acoulton:
> We will be upgrading PHP dependency with ORM 3, but for 2.x since nothing much happens, we are conservative, since we start triggering deprecations, we want as many users as possible to see them.

@beberlei fair enough, thanks for taking the time to respond.

PR #8617, opened 2021-04-13
Upgrade to doctrine/coding-standard 9

greg0ire:
Although this is no bugfix and touches `src`, I'm still targeting 2.8.x so that merges up are not too painful. The changes are about moving phpdoc comments around, so I don't think there is a risk of regression, even for static analysis.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "symfony/console": "^3.0|^4.0|^5.0"
     },
     "require-dev": {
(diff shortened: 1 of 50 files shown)


greg0ire:
cc @simonberger

ostrolucky (approved):

simonberger:
Thanks @greg0ire I'll rebase my PR in a couple of days. I expect a lot of conflicts.

Issue #8624, opened 2021-04-15
AbstractQuery::execute (and others) wrong typing


soullivaneuh:
# Summary

Since latest upgrade, PHPStan detect typing error related to the `AbstractQuery::execute` method:

https://github.com/doctrine/orm/blob/305e0d66641412731256c5cc4bb370027d8cf8af/lib/Doctrine/ORM/AbstractQuery.php#L1011-L1019

According to PHPStan owner, the typing is not correctly set: https://github.com/phpstan/phpstan-doctrine/issues/182

What do you think?

ondrejmirtes:
I'm also pretty sure that some places in Doctrine have it already correct with: `ArrayCollection|mixed[]|array|null`.

The reason why `ArrayCollection|Foo[]` is interpreted as `ArrayCollection of Foo` within PHPStan heuristics is because for many years, many developers have written it like that. To start interpreting it as `ArrayCollection or array` would break a lot of code because in a lot of code the array is not present.

greg0ire:
@soullivaneuh hi mate! I'm afraid that if we contribute it as @ondrejmirtes described, some well-minded contributor will simplify it back to how it is now. The way to fix this would be to have a test that makes a call like the one you performed, or if it exists already, to ensure the PHPStan level is high enough to detect that issue (right now we are pretty low).

@ondrejmirtes , is there a long-term plan to fix this situation (in a next major version of PHPStan)? PHP has union types now, and developers are probably going to be more and more confused if they have `|` meaning one thing in a context, and another thing in another context.

ondrejmirtes:
The easiest way to test this would be to have PHPStan analyse the codebase at least on level 5 (right now it's analysed on level 1, please consider increasing the level while employing a baseline) and then have some code somewhere in the codebase that calls the method with an array, and then have some code somewhere in the codebase that calls the method with an ArrayCollection.

I don't know how native union types fit into this scenario. I could probably make sure that `ArrayCollection|Foo[]` in the PHPDoc with native type `ArrayCollection|array` would result in the expected type which probably isn't the case right now. Feel free to open an issue about that :)

greg0ire:
I think you're right about the baseline, and it is now implemented :slightly_smiling_face: 

@soullivaneuh , would you like to contribute a fix for this now?

It should make https://github.com/doctrine/orm/blob/e2e9f8fa97a2e8b6ad79f1b2f68c921262677cc1/phpstan-baseline.neon#L2308-L2311 disappear


[...]

PR #8627, opened 2021-04-16
Add baseline files for static analyzers

greg0ire:
There are many CS and SA-related changes in the ORM as we pursue better
code quality, and easier contributions. These often involve huge
changes, which can be hard to review and inevitably lead to some
regressions. I know some of those could have been avoided if we were
using stricter levels for PHPStan and Psalm.

By adding baselines, we ensure new code is at level 5 for both tools,
which should allow us to catch the most interesting issues.

⚠️ I think what follows might be a fallacy know as "appeal to authority" , but both @ondrejmirtes and @muglug recommended this to us in #7691 and #8624 

Also note that is comparable to what we have been previously doing with https://github.com/diff-sniffer/diff-sniffer for the CS

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/MetadataFilter.php
+++ b/lib/Doctrine/ORM/Tools/Console/MetadataFilter.php
@@ -26,6 +26,7 @@ use Doctrine\Persistence\Mapping\ClassMetadata;
 use FilterIterator;
 use RuntimeException;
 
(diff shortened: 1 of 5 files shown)


greg0ire:
I don't reproduce the Psalm failure locally :confused:

ostrolucky (approved):

PR #8629, opened 2021-04-16
Fields for unique constraints

### LINKED ISSUE #8345: Add possibility to use fields instead of column for unique constraint

beberlei:
This replaces #8523 - since its not using a PR branch on the remote.

It fixes #8345 to implement support for field based configuration on `Index` and `UniqueConstraint` instead of column based.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/MappingException.php
+++ b/lib/Doctrine/ORM/Mapping/MappingException.php
@@ -933,4 +933,32 @@ class MappingException extends ORMException
             )
         );
     }
+
+    /**
+     * @return self
+     */
+    public static function invalidIndexConfiguration($className, $indexName)
+    {
+        return new self(
+            sprintf(
+                'Index %s for entity %s should contain columns or fields values, but not both.',
+                $indexName,
+                $className
+            )
+        );
+    }
+
+    /**
+     * @return self
+     */
+    public static function invalidUniqueConstraintConfiguration($className, $indexName)
+    {
+        return new self(
+            sprintf(
+                'Unique constraint %s for entity %s should contain columns or fields values, but not both.',
+                $indexName,
+                $className
+            )
+        );
+    }
 }
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -146,7 +146,7 @@ class SchemaTool
      *
      * @return string[] Column names from combined fields and columns mappings
      */
-    private function getIndexColumns($class, array $indexData)
+    private function getIndexColumns(ClassMetadata $class, array $indexData): array
     {
         $columns = [];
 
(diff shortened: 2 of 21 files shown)


lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -933,4 +933,32 @@ public static function illegalOverrideOfInheritedProperty($className, $propertyN
             )
         );
     }
+
+    /**
+     * @return self
+     */
```
greg0ire, line 939: Let's use a native return type declaration here?

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -137,6 +138,50 @@ private function processingNotRequired(
...
 
+    /**
+     * Resolves fields in index mapping to column names
+     *
+     * @param ClassMetadata $class
+     * @param mixed[]       $indexData index or unique constraint data
+     *
+     * @return string[] Column names from combined fields and columns mappings
+     */
+    private function getIndexColumns($class, array $indexData)
```
greg0ire:
```suggestion
    private function getIndexColumns(ClassMetadata $class, array $indexData): array
```

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -309,13 +354,13 @@ static function (ClassMetadata $class) use ($idMapping): bool {
                         $indexData['flags'] = [];
                     }
 
-                    $table->addIndex($indexData['columns'], is_numeric($indexName) ? null : $indexName, (array) $indexData['flags'], $indexData['options'] ?? []);
+                    $table->addIndex($this->getIndexColumns($class, $indexData), is_numeric($indexName) ? null : $indexName, (array) $indexData['flags'], $indexData['options'] ?? []);
```
greg0ire:
```suggestion
                    $table->addIndex(
                                        $this->getIndexColumns($class, $indexData),
                                        is_numeric($indexName) ? null : $indexName,
                                        (array) $indexData['flags'],
                                        $indexData['options'] ?? []
                    );
```

lib/Doctrine/ORM/Mapping/MappingException.php
beberlei: we can do that in one big step when we introduce typehints more consistently, this is how all other methods in `MappingException`, i prefer the consistency for now. We can rector everything at some point

greg0ire (approved):

PR #8630, opened 2021-04-16
Reduce baseline

Jean85:
This contains a couple of baseline reduction, without causing too much hassle in terms of code changes.
This was triggered by #8624 + https://twitter.com/greg0ire/status/1383108988259201026

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -123,7 +123,7 @@ abstract class AbstractQuery
      */
     protected $_hydrationMode = self::HYDRATE_OBJECT;
 
-    /** @var QueryCacheProfile */
+    /** @var QueryCacheProfile|null */
     protected $_queryCacheProfile;
 
     /**
@@ -289,7 +289,7 @@ abstract class AbstractQuery
     /**
      * Retrieves the associated EntityManager of this Query instance.
      *
-     * @return EntityManager
+     * @return EntityManagerInterface
      */
     public function getEntityManager()
     {
@@ -591,6 +591,7 @@ abstract class AbstractQuery
      */
     public function setResultCacheDriver($resultCacheDriver = null)
     {
+        /** @phpstan-ignore-next-line */
         if ($resultCacheDriver !== null && ! ($resultCacheDriver instanceof \Doctrine\Common\Cache\Cache)) {
             throw ORMException::invalidResultCacheDriver();
         }
@@ -721,7 +722,7 @@ abstract class AbstractQuery
     }
 
     /**
-     * @return QueryCacheProfile
+     * @return QueryCacheProfile|null
      */
     public function getQueryCacheProfile()
     {
@@ -979,8 +980,8 @@ abstract class AbstractQuery
      * Executes the query and returns an iterable that can be used to incrementally
      * iterate over the result.
      *
-     * @param ArrayCollection|mixed[] $parameters    The query parameters.
-     * @param string|int|null         $hydrationMode The hydration mode to use.
+     * @param ArrayCollection|array|mixed[] $parameters    The query parameters.
+     * @param string|int|null               $hydrationMode The hydration mode to use.
      *
      * @return iterable<mixed>
      */
@@ -1043,7 +1044,7 @@ abstract class AbstractQuery
             $this->setParameters($parameters);
         }
 
-        $setCacheEntry = static function (): void {
+        $setCacheEntry = static function ($data): void {
         };
 
         if ($this->_hydrationCacheProfile !== null) {
--- a/lib/Doctrine/ORM/Cache/CollectionHydrator.php
+++ b/lib/Doctrine/ORM/Cache/CollectionHydrator.php
@@ -30,9 +30,9 @@ use Doctrine\ORM\PersistentCollection;
 interface CollectionHydrator
 {
     /**
-     * @param ClassMetadata      $metadata   The entity metadata.
-     * @param CollectionCacheKey $key        The cached collection key.
-     * @param mixed[]|Collection $collection The collection.
+     * @param ClassMetadata            $metadata   The entity metadata.
+     * @param CollectionCacheKey       $key        The cached collection key.
+     * @param array|mixed[]|Collection $collection The collection.
      *
      * @return CollectionCacheEntry
      */
--- a/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
+++ b/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
@@ -111,6 +111,10 @@ class DefaultCacheFactory implements CacheFactory
         }
 
         if ($usage === ClassMetadata::CACHE_USAGE_READ_WRITE) {
+            if (! $region instanceof ConcurrentRegion) {
+                throw new InvalidArgumentException(sprintf('Unable to use access strategy type of [%s] without a ConcurrentRegion', $usage));
+            }
+
             return new ReadWriteCachedEntityPersister($persister, $region, $em, $metadata);
         }
 
@@ -134,6 +138,10 @@ class DefaultCacheFactory implements CacheFactory
         }
 
         if ($usage === ClassMetadata::CACHE_USAGE_READ_WRITE) {
+            if (! $region instanceof ConcurrentRegion) {
+                throw new InvalidArgumentException(sprintf('Unable to use access strategy type of [%s] without a ConcurrentRegion', $usage));
+            }
+
             return new ReadWriteCachedCollectionPersister($persister, $region, $em, $mapping);
         }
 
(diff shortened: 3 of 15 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -591,6 +591,7 @@ public function setResultCacheProfile(?QueryCacheProfile $profile = null)
      */
     public function setResultCacheDriver($resultCacheDriver = null)
     {
+        /** @phpstan-ignore-next-line */
```
greg0ire, line 594: What error are we ignoring here, and why do it that way? If future versions of PHPStan find more things here, we are going to ignore them too because of this, right?

lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php
```diff
@@ -30,10 +30,10 @@
 use Doctrine\ORM\Cache\Region;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
-use Doctrine\ORM\Mapping\ClassMetadataFactory;
 use Doctrine\ORM\PersistentCollection;
 use Doctrine\ORM\Persisters\Collection\CollectionPersister;
 use Doctrine\ORM\UnitOfWork;
+use Doctrine\Persistence\Mapping\ClassMetadataFactory;
```
greg0ire: This looks wrong… maybe the right fix here would be to assert that the value we are assigning is an instance of the ORM CMF… or can `$em->getMetadataFactory();` really return something else?

greg0ire:
@orklah please review :slightly_smiling_face:

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -979,8 +980,8 @@ public function iterate($parameters = null, $hydrationMode = null)
      * Executes the query and returns an iterable that can be used to incrementally
      * iterate over the result.
      *
-     * @param ArrayCollection|mixed[] $parameters    The query parameters.
-     * @param string|int|null         $hydrationMode The hydration mode to use.
+     * @param ArrayCollection|array|mixed[] $parameters    The query parameters.
```
orklah, line 983: |array seems useless given mixed[]

lib/Doctrine/ORM/Cache/CollectionHydrator.php
```diff
@@ -30,9 +30,9 @@
 interface CollectionHydrator
 {
     /**
-     * @param ClassMetadata      $metadata   The entity metadata.
-     * @param CollectionCacheKey $key        The cached collection key.
-     * @param mixed[]|Collection $collection The collection.
+     * @param ClassMetadata            $metadata   The entity metadata.
+     * @param CollectionCacheKey       $key        The cached collection key.
+     * @param array|mixed[]|Collection $collection The collection.
```
orklah, line 35: same

lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
```diff
@@ -201,7 +209,7 @@ public function getRegion(array $cache)
             }
 
             $directory = $this->fileLockRegionDirectory . DIRECTORY_SEPARATOR . $cache['region'];
-            $region    = new FileLockRegion($region, $directory, $this->regionsConfig->getLockLifetime($cache['region']));
+            $region    = new FileLockRegion($region, $directory, (string) $this->regionsConfig->getLockLifetime($cache['region']));
```
orklah, line 212: seems like the 3rd param should be an int instead of a string, it solves everything without changing the code

lib/Doctrine/ORM/Cache/Persister/Collection/CachedCollectionPersister.php
```diff
@@ -52,7 +52,7 @@ public function loadCollectionCache(PersistentCollection $collection, Collection
     /**
      * Stores a collection into cache
      *
-     * @param mixed[]|Collection $elements
+     * @param array|mixed[]|Collection $elements
```
orklah, line 55: same

lib/Doctrine/ORM/Cache/Region/FileLockRegion.php
```diff
@@ -63,8 +63,8 @@ class FileLockRegion implements ConcurrentRegion
     private $lockLifetime;
 
     /**
-     * @param string $directory
-     * @param string $lockLifetime
+     * @param string         $directory
+     * @param numeric-string $lockLifetime
```
orklah, line 67: as said before, this should be an int I think

psalm-baseline.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.3.2@57b53ff26237074fdf5cbcb034f7da5172be4524">
+<files psalm-version="4.1.1@16bfbd9224698bd738c665f33039fade2a1a3977">
```
orklah: please use the same version as the dependancy here

orklah (changes requested):

psalm-baseline.xml
orklah: My bad, this targets 2.8 and Psalm is indeed in 4.1.1 there. The baseline was generated with a more up to date version before.

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: @orklah see https://github.com/doctrine/orm/issues/8624 for why this is actually needed

psalm-baseline.xml
greg0ire: Woopsie

lib/Doctrine/ORM/AbstractQuery.php
orklah: yeah, I was just reading that, thanks for the info :)

lib/Doctrine/ORM/AbstractQuery.php
Jean85: We're ignoring the if because it's doing a FC check for the argument type, which is for the same type that is declared in the PHPDoc, so PHPStan is complaining that the `if` clause will be always true.

Maybe we can specify the kind of error in the annotation? Psalm has that functionality, I don't know if PHPStan does.

lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php
Jean85: I had the same doubt here. I don't know ORM's internals good enough to know if this is right or wrong.

lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
Jean85: Doesn't switching to an int be a breaking change though?

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: Well we could always add that error to the baseline, couldn't we?

lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php
greg0ire: Then please revert that one.

lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
orklah: Not sure how public FileLockRegion is. However, your PR already changed the type to numeric-string. In for a penny... :)

lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php
greg0ire: I found a good solution that just involves adding a comment to `EntityManagerInterface`, I'm going to push it to your branch soon

lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
greg0ire: `numeric-string` is still a string, and yes, switching to `int` would be a breaking change. Definitely not something you would want to do as part of a patch. As for `FileLockRegion`, it's not marked as internal, so in doubt, let's assume it's very much public.

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: I think I understand why you're doing it this way: there is no plan to remove that ignore rule ever. Sounds fair 👍

greg0ire (approved):

greg0ire:
Thanks @Jean85 !

PR #8631, opened 2021-04-16
Enforce read-only properties through Psalm annotation

Jean85:
As in title.

### DIFF
--- a/lib/Doctrine/ORM/Cache/AssociationCacheEntry.php
+++ b/lib/Doctrine/ORM/Cache/AssociationCacheEntry.php
@@ -26,14 +26,14 @@ namespace Doctrine\ORM\Cache;
 class AssociationCacheEntry implements CacheEntry
 {
     /**
-     * READ-ONLY: Public only for performance reasons, it should be considered immutable.
+     * @psalm-immutable Public only for performance reasons, it should be considered immutable.
      *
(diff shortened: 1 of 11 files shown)


greg0ire:
Maybe this should target 2.9.x, since it's not a bugfix, and might break some pipelines. Your other PR is not 100% bugfixes either (for instance the `numeric-string` thing is more precise, but `string` was technically not wrong). Maybe group things that are note bugfixes with this PR?

beberlei:
`@psalm-immutable` is a class level annotation, if you use `@psalm-readonly` however this will fail internally, because we don't modify these only the in the constructor. We cannot use Psalm annotations for this.

Ref: https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-immutable
https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-readonly-and-readonly

Jean85:
Ok, so maybe we can use `@psalm-readonly-allow-private-mutation` in those cases? 
https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-readonly-allow-private-mutation

greg0ire:
That sounds appropriate 👍

Jean85:
I've tried, it fails a lot, due to stuff like `ClassMetadataBuilder`, `ClassMetadataFactory` and a couple others...

PR #8633, opened 2021-04-17
Analyze test directory

greg0ire:
It is full of calls to our codebase. Analyzing that directory allows us
to "test" changes to the phpdoc even more.


greg0ire:
@orklah @Jean85 @ondrejmirtes please review

greg0ire:
> Adding tests will be a major pain, not sure we want to burden us this way

From @beberlei  - https://github.com/doctrine/orm/issues/8624#issuecomment-821803057

Do you mean it will make contributions to hard to make? Tests are supposed to be straightforward, and I think that if they are painful to write because of PHPStan and Psalm, then that means code will be painful to write for our users as well, doesn't it? And we should definitely work on that if that's the case.

orklah:
From my (limited) experience, tests often have to check things that will be considered errors by static analysis.

For example, if you want to check that a method that takes a string throws when given an int, Psalm will consider this an issue but it's a correct check (even more so if the type is only in phpdoc).

On the other hand, this highlights errors in tests and enable useful features in Psalm, like unused code, because any unused code on an API is untested code.

greg0ire:
> For example, if you want to check that a method that takes a string throws when given an int, Psalm will consider this an issue but it's a correct check (even more so if the type is only in phpdoc).

Does that seem to happen a lot here, based on the baseline?

ondrejmirtes:
Analyzing tests is often really useful - see for example https://twitter.com/OndrejMirtes/status/1368899254253740037

greg0ire:
It should be noted that this comes with a big performance hit:

Phpstan: 36s => ~59s~ 51s
Psalm: 28s => ~42s~ 30s

Update: I excluded some directories.

beberlei:
@ondrejmirtes ORM is not in a state that allows static analysis on the tests directory in a meaningful way that is not slowing us down massively. The code is 10 years old not primarly written with static analysis in mind. 

Alternatively we could introduce a single test-case that is used for PHPStan (and another one for Psalm) testing the APIs actually work.

The misinterpretation of `ArrayCollection|mixed[]` is unfortunate, because this has always been `ArrayCollection` or `mixed[]` in PHP docblocks. I feel we can't tackle this in ORM, because we would only put a band aid on a bug in phpstan.

Jean85 (approved):
I'm personally :+1: on analyzing the tests too, I've been doing since a long time in my private & OSS projects and it improves the DX a lot.

Obviously it has some "adoption curve" at the start, but I'm happy to help reducing this baseline too :smile:

greg0ire:
> Alternatively we could introduce a single test-case that is used for PHPStan (and another one for Psalm) testing the APIs actually work.

@Ocramius did exactly that recently on the DBAL, in case anyone wants to give that a try: https://github.com/doctrine/dbal/pull/4638

PR #8636, opened 2021-04-17
Update ignore rules to reflect current situation

greg0ire:
We no longer use Travis, we do not use git submodules as far as I know,
and we now use baseline files.


SenseException (dismissed):
You could insert `.doctrine-project.json` too while you're at it.

beberlei (approved):

PR #8640, opened 2021-04-18
FileLockRegion::__construct 3rd param is meant to be an int

orklah:
Following https://github.com/doctrine/orm/pull/8630#discussion_r615421792, this is a proper PR that propose changing the type of the param to an int.

This is kinda a BC break because while the code and behaviour doesn't change, it technically modify the expected type.

Note that before #8630, every value given internally was effectively an int.


greg0ire (approved):

ostrolucky:
just merge it, you don't need to ping me for such simple changes

PR #8646, opened 2021-04-20
[GH-8327] Make EntityManagerProvider compatible with expected DoctrineBundle usage

### LINKED ISSUE #8643: Doctrine console commands em default should be null
Doctrine console commands now have a default value as `default` instead of `null`.

When `em` is set to `null`, the `AbstractManagerRegistry`'s functionality kicks in and uses real default EntityManager (`$this->defaultManager`) https://github.com/doctrine/persistence/blob/2.2.x/lib/Doctrine/Persistence/AbstractManagerRegistry.php#L143-L146

Now's it looking for EM with name `default` which does not guarantee it's the default one.

https://github.com/doctrine/orm/blob/22413453daa872e121514bab1fe4c5fa982623ea/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/UpdateCommand.php#L50

> Doctrine ORM Manager named "default" does not exist.

beberlei:
This complication of the `EntityManagerProvider` API with a new method is necessary, because during integration in DoctrineBundle https://github.com/doctrine/DoctrineBundle/pull/1327 it showed that their `em` option defaults to `NULL`, which isn't compatible with ORM using `default` as the default value.

Related #8327 #8524

Fixes #8643

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/ConnectionFromManagerProvider.php
+++ b/lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/ConnectionFromManagerProvider.php
@@ -17,7 +17,7 @@ final class ConnectionFromManagerProvider implements ConnectionProvider
     public function __construct(EntityManagerProvider $entityManagerProvider, $defaultManagerName = 'default')
     {
         $this->entityManagerProvider = $entityManagerProvider;
-        $this->defaultManagerName = $defaultManagerName;
+        $this->defaultManagerName    = $defaultManagerName;
     }
 
     public function getDefaultConnection(): Connection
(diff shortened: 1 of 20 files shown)


lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/ConnectionFromManagerProvider.php
```diff
@@ -11,14 +11,18 @@ final class ConnectionFromManagerProvider implements ConnectionProvider
...
+    public function __construct(EntityManagerProvider $entityManagerProvider, $defaultManagerName = 'default')
     {
         $this->entityManagerProvider = $entityManagerProvider;
+        $this->defaultManagerName    = $defaultManagerName;
     }
 
     public function getDefaultConnection(): Connection
     {
-        return $this->entityManagerProvider->getManager('default')->getConnection();
+        return $this->entityManagerProvider->getManager($this->defaultManagerName)->getConnection();
```
simPod: I wonder whether this should leverage `EntityManagerProvider::getDefaultManager()` now when we have it (if `$defaultManagerName` was not provided).

beberlei: Yes you are right, this would be better.

simPod (approved):

PR #8649, opened 2021-04-22
ConvertDoctrine1Schema: Fix Doctrine 1 `notnull` field import

aweelex:


lib/Doctrine/ORM/Tools/ConvertDoctrine1Schema.php
```diff
@@ -202,7 +202,9 @@ private function convertColumn($className, $name, $column, ClassMetadataInfo $me
             throw ToolsException::couldNotMapDoctrine1Type($column['type']);
         }
 
-        $fieldMapping = [];
+        $fieldMapping = [
+            'nullable' => !$column['notnull'] ?? true, // Doctrine 1 columns are nullable by default
```
greg0ire:
```suggestion
            'nullable' => ! $column['notnull'] ?? true, // Doctrine 1 columns are nullable by default
```

beberlei (approved):
Lets just merge this, as we deprecated the code already and I don't expect more than a handful of people worldwide still needing this at this point in time.

PR #8651, opened 2021-04-22
Introduce PSR-6 for metadata caching

### LINKED ISSUE #8650: Deprecate using doctrine/cache for metadata caching
With doctrine/cache being sunset, and doctrine/persistence supporting PSR-6 caches in metadata factories, it's time to update the ORM configuration class to also accept PSR-6 caches. When setting up the metadata factory, the adapters provided in doctrine/cache should be used to avoid calling deprecated methods in doctrine/persistence.

alcaeus:
Closes #8650.

This bumps the doctrine/cache dependency to the just-released 1.11 version that includes the PSR-6 adapters needed for compatibility code. The dependency to doctrine/persistence does not to be bumped; instead we check if doctrine/persistence supports PSR-6 caching. I can update this if requested.

### DIFF
--- a/tests/Doctrine/Tests/ORM/ConfigurationTest.php
+++ b/tests/Doctrine/Tests/ORM/ConfigurationTest.php
@@ -19,7 +19,9 @@ use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Tests\Models\DDC753\DDC753CustomRepository;
 use PHPUnit\Framework\TestCase;
+use Psr\Cache\CacheItemPoolInterface;
 use ReflectionClass;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
 
 /**
  * Tests for the Configuration object
@@ -131,6 +133,14 @@ class ConfigurationTest extends TestCase
         $this->assertSame($queryCacheImpl, $this->configuration->getMetadataCacheImpl());
     }
 
+    public function testSetGetMetadataCache(): void
+    {
+        $this->assertNull($this->configuration->getMetadataCache());
+        $cache = $this->createMock(CacheItemPoolInterface::class);
+        $this->configuration->setMetadataCache($cache);
+        $this->assertSame($cache, $this->configuration->getMetadataCache());
+    }
+
     public function testAddGetNamedQuery(): void
     {
         $dql = 'SELECT u FROM User u';
@@ -182,7 +192,17 @@ class ConfigurationTest extends TestCase
         $this->addToAssertionCount(1);
     }
 
-    public function testEnsureProductionSettingsQueryCache(): void
+    public function testEnsureProductionSettingsWithNewMetadataCache(): void
+    {
+        $this->setProductionSettings('metadata');
+        $this->configuration->setMetadataCache(new ArrayAdapter());
+
+        $this->configuration->ensureProductionSettings();
+
+        $this->addToAssertionCount(1);
+    }
+
+    public function testEnsureProductionSettingsMissingQueryCache(): void
     {
         $this->setProductionSettings('query');
 
@@ -192,7 +212,7 @@ class ConfigurationTest extends TestCase
         $this->configuration->ensureProductionSettings();
     }
 
-    public function testEnsureProductionSettingsMetadataCache(): void
+    public function testEnsureProductionSettingsMissingMetadataCache(): void
     {
         $this->setProductionSettings('metadata');
 
@@ -213,7 +233,7 @@ class ConfigurationTest extends TestCase
         $this->configuration->ensureProductionSettings();
     }
 
-    public function testEnsureProductionSettingsMetadataArrayCache(): void
+    public function testEnsureProductionSettingsLegacyLegacyMetadataArrayCache(): void
     {
         $this->setProductionSettings();
         $this->configuration->setMetadataCacheImpl(new ArrayCache());
--- a/tests/Doctrine/Tests/ORM/ConfigurationTest.php
+++ b/tests/Doctrine/Tests/ORM/ConfigurationTest.php
@@ -17,8 +17,8 @@ use Doctrine\ORM\Mapping\QuoteStrategy;
 use Doctrine\ORM\ORMException;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
+use Doctrine\Tests\DoctrineTestCase;
 use Doctrine\Tests\Models\DDC753\DDC753CustomRepository;
-use PHPUnit\Framework\TestCase;
 use Psr\Cache\CacheItemPoolInterface;
 use ReflectionClass;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
@@ -26,7 +26,7 @@ use Symfony\Component\Cache\Adapter\ArrayAdapter;
 /**
  * Tests for the Configuration object
  */
-class ConfigurationTest extends TestCase
+class ConfigurationTest extends DoctrineTestCase
 {
     /** @var Configuration */
     private $configuration;
--- a/tests/Doctrine/Tests/ORM/ConfigurationTest.php
+++ b/tests/Doctrine/Tests/ORM/ConfigurationTest.php
@@ -136,7 +136,7 @@ class ConfigurationTest extends TestCase
     public function testSetGetMetadataCache(): void
     {
         $this->assertNull($this->configuration->getMetadataCache());
-        $cache = $this->createMock(CacheItemPoolInterface::class);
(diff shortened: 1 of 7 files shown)


greg0ire:
> The dependency to doctrine/persistence does not to be bumped; instead we check if doctrine/persistence supports PSR-6 caching. I can update this if requested.

Bumping to 2.2.0 does not look like a big deal dependency-wise (the only added dependency is `psr/cache`). Let's release 2.2.0 and bump the dependency?

alcaeus:
I wanted to release after 2.9 and the bundle releases to avoid unnecessary deprecation warnings users can't avoid

tests/Doctrine/Tests/ORM/ConfigurationTest.php
```diff
@@ -131,6 +133,14 @@ public function testSetGetMetadataCacheImpl(): void
         $this->assertSame($queryCacheImpl, $this->configuration->getMetadataCacheImpl());
     }
 
+    public function testSetGetMetadataCache(): void
+    {
+        $this->assertNull($this->configuration->getMetadataCache());
+        $cache = $this->createMock(CacheItemPoolInterface::class);
```
greg0ire:
```suggestion
        $cache = $this->createStub(CacheItemPoolInterface::class);
```

tests/Doctrine/Tests/ORM/ConfigurationTest.php
```diff
@@ -213,7 +233,7 @@ public function testEnsureProductionSettingsQueryArrayCache(): void
         $this->configuration->ensureProductionSettings();
     }
 
-    public function testEnsureProductionSettingsMetadataArrayCache(): void
+    public function testEnsureProductionSettingsLegacyLegacyMetadataArrayCache(): void
```
greg0ire:
```suggestion
    public function testEnsureProductionSettingsLegacyMetadataArrayCache(): void
```

alcaeus:
@greg0ire any guidance on those failures? I don't see how changing the metadata cache could cause these failures 🤔

greg0ire:
Another issue I do not reproduce locally :weary: … caching issues are the worst!

greg0ire:
Let's check if it is transient, but there are 2 more recent PRs that did not have this failure, so this time I think it really does come from the code in the PR.

greg0ire:
I don't understand, but I see that class metadata is involved at some point: https://github.com/doctrine/orm/blob/9a484504816be257303d4131ac37a2d1c5713c32/lib/Doctrine/ORM/UnitOfWork.php#L2181-L2199

greg0ire:
Ah I think it's starting to make sense:

https://github.com/doctrine/orm/blob/9a484504816be257303d4131ac37a2d1c5713c32/lib/Doctrine/ORM/UnitOfWork.php#L3153-L3158

Screenshot_2021-04-24 Introduce PSR-6 for metadata caching by alcaeus · Pull Request #8651 · doctrine orm

I can reproduce the issue locally with `ENABLE_SECOND_LEVEL_CACHE=1 vendor/bin/phpunit tests/Doctrine/Tests/ORM/Functional/IdentityMapTest.php`, and you probably can too :slightly_smiling_face:

alcaeus:
Thanks for investigating. I'll debug and see what breaks this. Thanks!

Nyholm (dismissed):
Thank you. 

I reviewed the implementation. Im happy with it. I did not dig into the tests.

alcaeus:
@greg0ire you're right, I can reproduce the failure locally, but I can also reliably reproduce this in 2.9.x. This looks very much unrelated to this patch.

greg0ire:
Indeed, me too!

UPDATE: I don't reproduce it when running the same phpunit command as in the CI though: `ENABLE_SECOND_LEVEL_CACHE=1 vendor/bin/phpunit -c ci/github/phpunit/sqlite.xml --exclude-group performance,non-cacheable,locking_functional --coverage-clover=coverage-cache.xml`

Notice the `--exclude-group non-cacheable` part, and how a lot of tests fail when dropping that group. Those were introduced along with the second level cache in https://github.com/doctrine/orm/pull/808

greg0ire:
Screenshot_2021-04-28 Second level cache by FabioBatSilva · Pull Request #808 · doctrine orm
Screenshot of the relevant replies since they are buried deep inside the 200 or so collapsed comments :point_up: 

Both links are 404s of course:
`https://github.com/FabioBatSilva/doctrine2/blob/slc/docs/en/reference/second-level-cache.rst#composite-primary-key`
`https://github.com/FabioBatSilva/doctrine2/blob/slc/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC117Test.php#L479`

But I think the gist of it is: if a test is about counting things, then we can put it in that group?

Note that when running `ENABLE_SECOND_LEVEL_CACHE=1 vendor/bin/phpunit --exclude-group performance,locking_functional,non-cacheable`, tests pass but the test that fails with the code in this PR still gets executed. It also is executed and passes if using the same dependencies as with your branch.

So some other test before that one makes it somehow pass on 2.9, and no longer does on your branch.

greg0ire:
I bisected it down to `tests/Doctrine/Tests/ORM/Functional/ExtraLazyCollectionTest.php`. Removing that test from the suite makes the test fail, so it's the only one that does what's necessary.

Reducing the test to just the following is enough to make `IdentityMapTest::testCollectionValuedAssociationIdentityMapBehaviorWithRefresh` pass on 2.9.x:

```php
<?php

declare(strict_types=1);

namespace Doctrine\Tests\ORM\Functional;

use Doctrine\Tests\Models\CMS\CmsArticle;
use Doctrine\Tests\Models\CMS\CmsGroup;
use Doctrine\Tests\Models\CMS\CmsPhonenumber;
use Doctrine\Tests\Models\CMS\CmsUser;
use Doctrine\Tests\Models\DDC2504\DDC2504ChildClass;
use Doctrine\Tests\Models\DDC2504\DDC2504OtherClass;

use function array_shift;
use function assert;
use function count;

/**
 * Description of ExtraLazyCollectionTest
 */
class ExtraLazyCollectionTest extends OrmFunctionalTestCase
{
    {
        parent::setUp();

        $class = $this->_em->getClassMetadata(CmsUser::class);

        // drop that line and IdentityMapTest::testCollectionValuedAssociationIdentityMapBehaviorWithRefresh fails
        unset($class->associationMappings['phonenumbers']['cache']); 

    }

    public function testContainsKeyNonExistentIndexByManyToMany(): void
    {
    }
}
```

Both tests deal with phonenumbers, and the line that makes the subsequent test pass has to do with metadata cache, so it's starting to make some sense.

greg0ire:
Blocked by https://github.com/doctrine/orm/pull/8659 (and a merge up)

alcaeus:
@greg0ire @beberlei codecov failures are due to not testing with a development version of doctrine/persistence. That can only be released after an ORM release containing this patch to avoid unnecessary deprecation warnings.

greg0ire:
> That can only be released after an ORM release containing this patch to avoid unnecessary deprecation warnings.

Or maybe revert those, add them back later?

tests/Doctrine/Tests/ORM/ConfigurationTest.php
```diff
@@ -131,6 +133,14 @@ public function testSetGetMetadataCacheImpl(): void
         $this->assertSame($queryCacheImpl, $this->configuration->getMetadataCacheImpl());
     }
 
+    public function testSetGetMetadataCache(): void
+    {
+        $this->assertNull($this->configuration->getMetadataCache());
+        $cache = $this->createStub(CacheItemPoolInterface::class);
```
greg0ire, line 139: Make sure to extend `DoctrineTestCase` for `createStub` to be available on PHP 7.1

greg0ire:
> alcaeus requested a review from greg0ire 20 minutes ago

Let's talk about the aforementioned issue first? I think the deprecations introduced in https://github.com/doctrine/persistence/pull/144 should be reverted if it is what we are talking about, until all calling projects have adapted. Then, in a subsequent PR, we can reintroduce them, and we might want to use `doctrine/deprecations` this time, might we not?

alcaeus:
Summary of discussion with @greg0ire:
* Deprecation will be done in doctrine/persistence 2.2. We're waiting on an unrelated bug fix, so the release is a few days out. Despite that, the deprecation is not in the hot path, so there's no requirement to release ORM before persistence
* The bundle will be released once its work has been completed. This ensures that the bundle itself configures a PSR-6 cache if the methods are available

With that in mind, I think this PR is ready for review.

greg0ire (commented):
Please summarize the deprecations in `UPGRADE.md`

beberlei:
I think i can do the deprecation summarization as part of the general 2.9 work, as this is not done for a few deprecations yet.

beberlei (approved):

greg0ire:
In that case, let's go!

greg0ire:
Thanks @alcaeus !

PR #8655, opened 2021-04-23
UnitOfWork: calling EntityManager::getClassMetadata() only when needed

smilesrg:
I noticed some unnecessary calls of `EntityManager::getClassMetadata()` in some cases. This PR improves that by calling EntityManager::getClassMetadata() only when needed.


greg0ire:
Maintainers: is this the right target branch? I'm don't think this qualifies as a bugfix…

Issue #8518, opened 2021-02-28
Use ProxyManager


beberlei:
We can replace our own proxy generator with https://github.com/FriendsOfPHP/proxy-manager-lts in two steps:

- Replace current behavior 1:1. This should target the next 2.x branch already and work alongside the existing proxy code, so that usres can choose. Use the Mongo ODM as example, they moved to proxy manager already. Tricky bits will be how to make sure something like `getClassMetadata(get_class($proxy))`, which needs a resolver or something alike.
- Add support for lazy loading single columns later than the whole entity such as large clobs/blobs.

smilesrg:
`composer require friendsofphp/proxy-manager-lts`
`composer require -W friendsofphp/proxy-manager-lts`

Leads to an error:

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires doctrine/coding-standard ^9.0, found doctrine/coding-standard[9.0.0] but the package is fixed to 8.2.1 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
  Problem 2
    - Root composer.json requires vimeo/psalm 4.7.0, found vimeo/psalm[4.7.0] but the package is fixed to 4.3.2 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
```
/cc @beberlei @greg0ire

greg0ire:
It works for me on b2f404b25 . Have you run `composer update` beforehand?

smilesrg:
`composer update` helped, thank you very much!

beberlei:
@alcaeus brought to my attention that our old prototype branch has done this also: https://github.com/doctrine/orm/blob/old-prototype-3.x/lib/Doctrine/ORM/Proxy/Factory/StaticProxyFactory.php - This could be used as a starting point, sorry I didn't see this before.

alcaeus:
> Tricky bits will be how to make sure something like `getClassMetadata(get_class($proxy))`, which needs a resolver or something alike.

We've had issues around that in ODM. We worked around it, but the upcoming release of doctrine/persistence includes configurable resolvers: https://github.com/doctrine/persistence/pull/145. This will allow a better implementation in ORM without hooking into a number of method calls to resolve proxy class names.

alcaeus:
[...]

PR #8660, opened 2021-04-28
Unit of work refactoring: extracted IdentityMap

smilesrg:
Hello!

This PR introduces UnitOfWork class refactoring. This PR introduces IdentityMap extraction from UnitOfWork. The ideas of refactoring are:

* Reduce the complexity of UnitOfWork by extracting several dependencies
* Make extracted dependencies reusable in other classes

That would make UnitOfWork better testable and maintainable.

Refactoring plan is to extract such classes from UnitOfWork:
* ChangesetComputer
* Scheduler
* EntityStateManager (move out switch/case `UnitOfWork::STATE_*` constants)

What do you think about this initiative? 

/cc @greg0ire @beberlei


composer.json
```diff
@@ -16,7 +16,7 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.1|^8.0",
+        "php": "^7.4|^8.0",
```
smilesrg: Bumped up PHP version to 7.4 LTS to use properties type hinting

lib/Doctrine/ORM/Internal/IdentityMap.php
```diff
@@ -0,0 +1,217 @@
...
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\ORMInvalidArgumentException;
+
+use function array_map;
+use function array_sum;
+use function get_class;
+use function implode;
+use function in_array;
+
+class IdentityMap implements Countable
```
smilesrg, line 16: I am thinking about implementing the `ArrayAccess` interface or extending from`SplObjectStorage`, but it's under discussion, the problem is that we need to use entity identifiers separately. If we can change the logic not to use entity identifieres, that could be possible. And since PHP8 we even can use `WeakMap`

lib/Doctrine/ORM/Internal/IdentityMap.php
```diff
@@ -0,0 +1,217 @@
...
+     * an identity. The entities are grouped by their class name.
+     * Since all classes in a hierarchy must share the same identifier set,
+     * we always take the root class name of the hierarchy.
+     *
+     * @var mixed[]
+     * @psalm-var array<class-string, array<string, object|null>>
+     */
+    private $identityMap = [];
+
+    public function addEntityIdentifier(string $oid, $identifier): void
```
smilesrg, line 46: I found just one place where this method is used without adding entity to an identity map. And probably we can re-write the logic to use just identity map methods, without entity identifiers. Please check UnitOfWork::persistNew()

```php
            // Some identifiers may be foreign keys to new entities.
            // In this case, we don't have the value yet and should treat it as if we have a post-insert generator
            if (! $this->hasMissingIdsWhichAreForeignKeys($class, $idValue)) {
                $this->identityMap->addEntityIdentifier($oid, $idValue);
            }
```

lib/Doctrine/ORM/Internal/ObjectIdFetcher.php
```diff
@@ -0,0 +1,31 @@
...
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\ORM\Internal;
+
+use function spl_object_hash;
+
+class ObjectIdFetcher
```
smilesrg, line 25: I introduced this class to wrap getting object hash by using `spl_object_hash`. Later, we can change this function to `spl_object_id` that is faster.

greg0ire:
> What do you think about this initiative?

Without reviewing yet, it sounds great to me! Maybe the extracted classes should be made `final` and marked as internal? Otherwise a lot more changes become BC-breaks.

Update: I see you extracted them to the `Internal` namespace, so I guess it's fine although IDE and SA tools will probably not detect that.

tests/Doctrine/Tests/ORM/Functional/DetachedEntityTest.php
```diff
@@ -96,7 +96,6 @@ public function testSerializeUnserializeModifyMerge(): void
         $user = $this->_em->merge($user); // merge cascaded to phonenumbers
         $this->assertInstanceOf(CmsUser::class, $user->phonenumbers[0]->user);
         $this->assertInstanceOf(CmsUser::class, $user->phonenumbers[1]->user);
-        $im = $this->_em->getUnitOfWork()->getIdentityMap();
```
smilesrg, line 99: that's not needed - not used in the test

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
```diff
@@ -454,19 +454,6 @@ public function invalidAssociationValuesDataProvider(): array
         ];
     }
 
-    /**
-     * @param object $entity
-     *
-     * @dataProvider entitiesWithValidIdentifiersProvider
-     */
-    public function testAddToIdentityMapValidIdentifiers($entity, string $idHash): void
```
smilesrg, line 462: Test is removed and covered in `IdentityMapTest`

composer.json
greg0ire: I'd rather have a BC way contributed to 2.9.x and then whatever requires 7.4 added to 3.0.x. The less divergence between branches, the easier the merge up. And we definitely want to be able to merge up, unlike, you know, previously :grimacing: 

May of the public methods you are removing are marked as internal (although not with `@internal`), so removing them should not be considered a BC-break.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1797,7 +1660,7 @@ private function doPersist($entity, array &$visited): void
             case self::STATE_REMOVED:
```
smilesrg, line 1660: These switch/cases are needed to be replaced with state pattern (EntityStateManager class)

smilesrg:
> Otherwise a lot more changes become BC-breaks.

I thought that Doctrine 3.0 will introduce BC-breaks :-)

>I'd rather have a BC way contributed to 2.9.x and then whatever requires 7.4 added to 3.0.x. The less divergence between branches, the easier the merge up. And we definitely want to be able to merge up, unlike, you know, previously 😬

😄 I guess after refactoring, all UnitOfWork changes need to be merged manually, as a lot of code will migrate to other classes. And that wouldn't be very easy 😢

composer.json
greg0ire: Are those required though? Again, I'd rather have a version without them contributed to 2.9.x, then we merge up, then you can contribute types to the properties.

greg0ire:
> I guess after refactoring, all UnitOfWork changes need to be merged manually, as a lot of code will migrate to other classes. And that wouldn't be very easy

I don't understand what you mean, please elaborate.

smilesrg:
>I don't understand what you mean, please elaborate.

For example, this change

https://github.com/doctrine/orm/pull/8201/files#diff-55a900494fc8033ab498c53929716caf0aa39d6bdd7058e7d256787a24412ee4R3041-R3042


...is done in UnitOfWork, but after merging this PR, should be transferred to `IdentityMap` manually.

greg0ire:
Oh right, but you are only talking about changes that happened between the time where you started working on this and the time when your PR gets merged to 2.9.x, right? Because I'm don't think that would be such a big deal to do that rebase, would it?

We would still have the problem you describe with 2.8.x, but that's probably fine, 2.8.x is a bugfix branch and is not supposed to change a lot. Also, git is able to figure a lot of this by itself :slightly_smiling_face:

smilesrg:
> Oh right, but you are only talking about changes that happened between the time where you started working on this and the time when your PR gets merged to 2.9.x, right? Because I'm don't think that would be such a big deal to do that rebase, would it?

You understood me right, I don't think there'll be many conflicts, and even if there will be, I'll resolve them :-)

greg0ire:
Consider waiting for the approval of other maintainers before doing that effort, let's see if they agree with me and also with merging this in theory :)

ostrolucky:
Not a move without careful performance benchmarking. UOW was designed that way because of performance reasons. It's often a bottleneck, so anything helps in those tight loops, including overhead of calling a function instead of isset directly.

beberlei:
@smilesrg I am sorry, but this kind of internal refactoring is not something we are looking to work on at the moment with everything else that needs work. While certainly code that can be improved the `UnitOfWork` itself is pretty stable code and well tested from many different angles. Without a long term goal, this and other refactoring is only going to create noise to review and risk to introduce bugs or performance changes and we are stretched very thin on time as-is, as you can see from the many open Pull Requests that go unmerged.

I see you have a lot of enthusiasm working on ORM issues, would it be helpful If I compiled a list of things that we plan to work on so that you can pick what interests you and it the likelihood of us merging it is much higher? I believe that would be less frustrating for both sides :-)

smilesrg:
@beberlei Thanks for information. I'll look at the other issues that are more important. 

>Without a long term goal, this and other refactoring is only going to create noise to review and risk to introduce bugs or performance changes and we are stretched very thin on time as-is

That's why I targeted this PR to 3.0 version, as it's a long-term project. I guess when you will have a roadmap for 3.0, it worth upgrading the codebase too, because right now it's not really easy to understand what's going on in UnitOfWork, it's very hard to debug. Also, introducing new changes and bug fixes is complicated too.

beberlei:
@smilesrg one issue that you could look into is https://github.com/doctrine/orm/issues/8518 - bulletpoint 1.

smilesrg:
@beberlei thanks, I'll take a look at the issue you mentioned

PR #8669, opened 2021-05-04
Validate all fields in schema validator

alexander-schranz:
I'm currently debugging some thing with resolve target entities and did saw that the validator only output the first error instead of all associations. The following changes should now return all errors on the target entity.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -103,7 +103,7 @@ class SchemaValidator
             if (! class_exists($assoc['targetEntity']) || $cmf->isTransient($assoc['targetEntity'])) {
                 $ce[] = "The target entity '" . $assoc['targetEntity'] . "' specified on " . $class->name . '#' . $fieldName . ' is unknown or not an entity.';
 
-                return $ce;
+                continue;
             }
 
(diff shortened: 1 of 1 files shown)


greg0ire:
So right now, `doctrine:schema:validate` only validates one field per entity, is that it?

alexander-schranz:
> So right now, doctrine:schema:validate only validates one field per entity, is that it?

Currently it will `break;` after the **first** invalid field and so output only the first invalid field instead of returning **all** invalid fields.

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -103,7 +103,7 @@ public function validateClass(ClassMetadataInfo $class)
             if (! class_exists($assoc['targetEntity']) || $cmf->isTransient($assoc['targetEntity'])) {
                 $ce[] = "The target entity '" . $assoc['targetEntity'] . "' specified on " . $class->name . '#' . $fieldName . ' is unknown or not an entity.';
 
-                return $ce;
+                continue;
```
greg0ire, line 106: Would it be possible to write a test for this? Or is it too involving?

alexander-schranz: Will try to have a look at it this week.

PR #8670, opened 2021-05-04
Bug/#8667 segmentation fault when column is nullable

Nicc0:
When property in Entity is defined as column with ManyToOne relation and have possibility to provide nullable values, during insertion to database original `\PDO` object returns PHP Core error `Segmentation fault`, because we trying to bind nullable parameter as integer. But this PullRequest is just proposition. Maybe someone have better idea how we can manage nullable relation and binding of parameters.

Tested with `doctrine\orm` 2.9.x and PHP 8.0.5


beberlei:
@Nicc0 what database driver is this with? a segfault means the problem is not Doctrine, but PHP related. Have you looked at bugs.php.net for similar ones? I saw that PHP 8.0.5 has a regression in PDO PgSQL, so maybe that is related?

beberlei:
Closing because no feedback was given and assumption is that was related to 8.0.5 regression.

PR #8671, opened 2021-05-04
Reflect that default EntityManager is not always named `default`

simPod:
Follows up #8646


simPod:
cc @beberlei

beberlei (approved):
@simPod thank you!

PR #8672, opened 2021-05-05
Allow installing doctrine/cache 2.0

alcaeus:
This PR is a draft for changes that would have to be made in ORM to allow installing doctrine/cache 2.0. The initial commit updating composer.json would have to be reworked - the `@dev` constraints are necessary until we tag releases allowing doctrine/cache 2.0 in those packages.

The changes mostly affect the test, but as is, this PR breaks applications that are relying on the setup tool. Without cache implementations, we won't be able to restore functionality, so we would either have to block doctrine/cache 2.0 until this has changed, or ship a different cache implementation which isn't really what we want.

I'm not sure what the best course of action is here. The Setup tool isn't necessary to run the ORM, so I'm a bit bummed that it's the only thing blocking us from allowing doctrine/cache 2.0. On the other hand, even if we were to simply not set a cache if doctrine/cache 2.0 is installed, this could cause severe issues for people relying on caches to be set automatically. @beberlei would you have any guidance on this specifically?


alcaeus:
After discussion with @beberlei, this PR now contains a minor BC break in the setup tool when doctrine/cache 2.0 is installed. In those cases, the setup tool will need to be provided with a cache or have the symfony/cache package installed. Other PSR-6 libraries can be supported as well; people who have an interest in having their cache of choice supported should create a PR adding it.

I've also improved the BC layer for `getMetadataCacheImpl` to return a wrapped PSR-6 cache when one was configured. If a doctrine/cache was configured, the method always returns that cache instead.

composer.json
```diff
@@ -19,17 +19,17 @@
         "php": "^7.1|^8.0",
         "ext-pdo": "*",
         "composer/package-versions-deprecated": "^1.8",
-        "doctrine/annotations": "^1.12",
-        "doctrine/cache": "^1.11.0",
+        "doctrine/annotations": "^1.12@dev",
```
beberlei: These need to be reverted to their stable counter parts, I assume they were only present for testing?

beberlei (changes requested):
Looks good, just one question about `composer.json` changes, I assume they need to be reverted to loose the `@dev`?

composer.json
alcaeus: Correct. If everything else looks good, I'd remove the dev portion to ensure I didn't break anything.

alcaeus:
Change applied, this now depends on doctrine/annotations 1.13 (which was just released) for the PsrCachedReader.

alcaeus:
PhpStan errors will reappear once doctrine/cache 2.0 becomes eligible again. What should we do here?

beberlei (approved):

greg0ire (approved):

Issue #8463, opened 2021-02-06
Deprecate AbstractQuery::iterate


beberlei:
The result format of `AbstractQuery::iterate` was always a bit clumsy as #7885 explains and subsequently changes by introducing `toIterable`. As such `iterate` is deprecated as of Doctrine ORM 2.8 in favor of `toIterable`, which returns a generator and doesn't return the result at key 0, instead directly.

Before:

```php
$results = $em->select('e')->from('entity')->getQuery()->iterate();
foreach ($resuls as $result) {
   $entityIsHere = $result[0];
}
```

After:

```php
$results = $em->select('e')->from('entity')->getQuery()->toIterable();
foreach($resuls as $result) {
    $entityIsHere = $result;
}
```

PR #8673, opened 2021-05-05
Add hint for ->iterate() deprecation

Seldaek:
Just going over deprecations and noticed this one just tells me it's deprecated without saying what to use instead.. Also pointed to https://github.com/doctrine/orm/issues/8463 as it clearly shows the usage change which is great to have 👍🏻


greg0ire (approved):
Indeed! Also, we should migrate from `trigger_error` to `doctrine/deprecations` on 2.9.x

greg0ire:
Thanks @Seldaek !

PR #8678, opened 2021-05-09
[Gh-8589] A new approach to non-nullable typed associations for BC

### LINKED ISSUE #8638: Check CommitOrderCalculator inconsistencies before 2.9.0
See last commits in https://github.com/doctrine/orm/pull/8589, changing the nullable on JoinColumn causes a regression that can be explained by the weird condition in UnitOfWork building up the CommitOrderCalculator.

beberlei:
This partially reverts #8589 because it was overzealously working on non typed properties as well, defining 'nullable' => true on all join columns. This made `CascadeRemoveOrderTest` fail on databases with foreign keys.

With this approach the code about join columns stays untouched, unless the typed property is defined as non-nullable *and* join columns are used. 

Fixes #8638

### DIFF
--- a/tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
@@ -278,13 +278,6 @@ abstract class AbstractMappingDriverTest extends OrmTestCase
 
         // Explicit Not Nullable
         $this->assertFalse($class->isNullable('username'));
-
-        // Join table Nullable
-        $this->assertFalse($class->getAssociationMapping('email')['joinColumns'][0]['nullable']);
-        $this->assertEquals(CmsEmail::class, $class->getAssociationMapping('email')['targetEntity']);
-
-        $this->assertTrue($class->getAssociationMapping('mainEmail')['joinColumns'][0]['nullable']);
-        $this->assertEquals(CmsEmail::class, $class->getAssociationMapping('mainEmail')['targetEntity']);
     }
 
     public function testFieldTypeFromReflection(): void
--- a/tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
@@ -278,6 +278,9 @@ abstract class AbstractMappingDriverTest extends OrmTestCase
 
         // Explicit Not Nullable
         $this->assertFalse($class->isNullable('username'));
+
+        $this->assertEquals(CmsEmail::class, $class->getAssociationMapping('email')['targetEntity']);
+        $this->assertEquals(CmsEmail::class, $class->getAssociationMapping('mainEmail')['targetEntity']);
     }
 
     public function testFieldTypeFromReflection(): void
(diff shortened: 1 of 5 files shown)


beberlei:
@Lustmored ping for a comment :-)

tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
```diff
@@ -278,13 +278,6 @@ public function testFieldIsNullableByType(): void
 
         // Explicit Not Nullable
         $this->assertFalse($class->isNullable('username'));
-
-        // Join table Nullable
-        $this->assertFalse($class->getAssociationMapping('email')['joinColumns'][0]['nullable']);
-        $this->assertEquals(CmsEmail::class, $class->getAssociationMapping('email')['targetEntity']);
-
-        $this->assertTrue($class->getAssociationMapping('mainEmail')['joinColumns'][0]['nullable']);
-        $this->assertEquals(CmsEmail::class, $class->getAssociationMapping('mainEmail')['targetEntity']);
```
Lustmored, line 287: I believe `targetEntity` tests should stay. My method naming is lacking here as it suggests only checks for `nullable`, while in fact there are 2 checks for `targetEntity` also. Probably method name change would be good?

tests/Doctrine/Tests/ORM/Mapping/ClassMetadataTest.php
```diff
@@ -127,15 +127,6 @@ public function testFieldIsNullableByType(): void
...
         $this->assertFalse($cm->isNullable('username'));
-
-        // Join table Nullable
-        $cm->mapOneToOne(['fieldName' => 'email', 'joinColumns' => [[]]]);
-        $this->assertFalse($cm->getAssociationMapping('email')['joinColumns'][0]['nullable']);
-        $this->assertEquals(CmsEmail::class, $cm->getAssociationMapping('email')['targetEntity']);
-
-        $cm->mapManyToOne(['fieldName' => 'mainEmail']);
-        $this->assertTrue($cm->getAssociationMapping('mainEmail')['joinColumns'][0]['nullable']);
-        $this->assertEquals(CmsEmail::class, $cm->getAssociationMapping('mainEmail')['targetEntity']);
```
Lustmored, line 138: I believe `targetEntity` tests should stay. My method naming is lacking here as it suggests only checks for `nullable`, while in fact there are 2 checks for `targetEntity` also. Probably method name change would be good?

Lustmored (dismissed):
It generally looks good and certainly solves the issue :+1:  I was certain that `nullable = true` should be the default hence the regression.

I would just leave tests for `targetEntity` (per code comments) as they seem useful.

rvanlaak:
I have a question while trying to understand the scope of this PR, and whether we will wait on 2.9 to release or try helping by contributing a PR.

Could this PR somehow be related to the long-standing https://github.com/doctrine/orm/issues/4090 issue from 2014 on `OneToOne` relationships having their FK set to non-nullable? 

In other words, is `2.9` going to allow the scenario that PR https://github.com/doctrine/orm/pull/6533 tried to solve earlier? 👏 If not, would there be anything to help, and/or could a fix get backported to make it in 2.8.x ?

PR #8689, opened 2021-05-14
Add phpstan-deprecation-rules violations to baseline

alcaeus:
This PR adds phpstan-deprecation-rules to statically look for deprecated method calls. All current violations have been added to the baseline and can then be selectively removed as people want to upgrade these. For example, one could remove all ignored errors referring to deprecated DBAL behaviour and start fixing errors until PHPStan (and the test suite) is happy.

PR #8691, opened 2021-05-15
Check for use of deprecated API with Psalm

alcaeus:
Follow-up to #8689.

This enables checks for usage of deprecated API in the psalm configuration and also updates the baseline to not report any errors. Deprecations will be fixed gradually.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,20 +1,104 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="4.7.0@d4377c0baf3ffbf0b1ec6998e8d1be2a40971005">
   <file src="lib/Doctrine/ORM/AbstractQuery.php">
+    <DeprecatedClass occurrences="1">
+      <code>IterableResult</code>
+    </DeprecatedClass>
+    <DeprecatedMethod occurrences="1">
+      <code>iterate</code>
+    </DeprecatedMethod>
+    <DocblockTypeContradiction occurrences="1">
+      <code>$resultCacheDriver !== null &amp;&amp; ! ($resultCacheDriver instanceof \Doctrine\Common\Cache\Cache)</code>
+    </DocblockTypeContradiction>
     <FalsableReturnStatement occurrences="1">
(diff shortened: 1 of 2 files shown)


ostrolucky:
DeprecatedMethod was just an example I gave. There are a lot more deprecated* issue handlers in Psalm https://github.com/vimeo/psalm/blob/master/docs/running_psalm/error_levels.md:

DeprecatedClass
DeprecatedConstant
DeprecatedFunction
DeprecatedInterface
DeprecatedMethod
DeprecatedProperty
DeprecatedTrait

Maybe would be easier to bump psalm to level 3?

alcaeus:
We'd have to bump to level 2, as the `Deprecated*` class of issues are treated as errors only from level 2 and lower. I'm ok with updating the baseline and require new code to be stricter, but then would definitely defer this change until 2.10. WDYT?

alcaeus:
As a follow-up, here's the output when moving to level 2:
```
No errors found!
------------------------------
5995 other issues found.
```

And just for shits and giggles, here's what happens when we include tests:
```
No errors found!
------------------------------
18562 other issues found.
```

ostrolucky:
Ok. What about other Deprecated* issue handlers? Do you want to include them in this PR, or you think DeprecatedMethod will do?

alcaeus:
I'd add all others, but as you said: it might be easier to go to level 2. This puts more restrictions on new codes with existing issues being safely ignored through the baseline.

greg0ire:
> here's what happens when we include tests:

About that, [@beberlei proposed to have some kind of sample calling code instead](https://github.com/doctrine/orm/pull/8633#issuecomment-821863690). That's looks a lot like what Marco did on the DBAL recently: https://github.com/doctrine/dbal/pull/4638

alcaeus:
Then let's not add tests. I've pushed a change to check level 2.

greg0ire (approved):
Looks good! Is there a Psalm command that allows to only see the deprecations?

ostrolucky (approved):

greg0ire:
Thanks @alcaeus !

PR #8692, opened 2021-05-15
Split the ORMException class

greg0ire:
This is a "backport" of https://github.com/doctrine/orm/pull/6743

I'm using quotes here because it's different in that it should be backwards-compatible, and a subsequent backport of #6473 should happen on 3.0.x if this gets merged.

# Todo

- [x] deprecate methods
- [x] address https://github.com/doctrine/orm/pull/8692#discussion_r636341387 (and generalize it if needed)
- [x] hunt for new `throw ORMException` occurrences


lib/Doctrine/ORM/Exception/RepositoryException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Exception;
+
+/**
+ * This interface should be implemented by all exceptions in the Repository
```
stof, line 10: this is not an interface.

And if you intend to turn it into an interface in 3.0, I suggets making it an abstract class so that users cannot instantiate it.

stof (commented):
Should the old ORMException be tagged as `@internal` or `@deprecated` (not sure which one is better) so that static analysis tools report its usage in catch clauses ? Otherwise, it will have to stay forever.
Also, existing factory methods in ORMException should be deprecated

lib/Doctrine/ORM/Exception/RepositoryException.php
stof: btw, this could be introduced as an interface directly, with the actual extensions extending from the old class (for BC) and implementing the new extension.

greg0ire:
I'd say let's mark it as `@deprecated`, marking it as `@internal` would make BC-breaks in it OK, when those BC-breaks shouldn't happen really.

ostrolucky (dismissed):

SenseException (dismissed):

greg0ire:
Rebased as I don't understand why there were 14 commits…

BenoitDuffez:
Hi, prob not the best place to ask this, but I'm wondering why the old exception was deprecated when the entity manager still has it in its annotation? For example `EntityManager::persist` has `@throws ORMException` which is not covered by a `use` statement, so the `Doctrine\ORM\ORMException` is used.

Now I have two options (or both at the same time)

* have my IDE warn me I'm trying to catch a deprecated exception (if I catch the old one)
* have my IDE warn me about an uncaught exception (if I catch the new one)

Either way I don't even know which exception is thrown, so I'm not very confident in all my try/catches around Doctrine.

Is this covered by a document that I would have missed? I checked the upgrade document at the root of this repo but didn't find the info I was looking for.

I'm currently using 2.11.2.

greg0ire:
@BenoitDuffez indeed not the best place.

> I'm wondering why the old exception was deprecated when the entity manager still has it in its annotation?

As you can see, this was a big PR I made ages ago, and I've probably made some mistakes. Please send a PR replacing that annotation with what seems correct from exploring all the possibilities, a quick look shows that there aren't that many.

BenoitDuffez:
@greg0ire sure! Sorry about the way I asked. Thanks for replying so rapidly.

Here's the PR: https://github.com/doctrine/orm/pull/9623. I did limited testing and now my IDE is happy.

Issue #5961, opened 2016-08-02
Useless join statements when using a discriminator map


sh4bang:
Hi !

I'm working in a big project in which I use a discriminatorMap to handle many entity type : [Class Table Inheritance](http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/reference/inheritance-mapping.html#class-table-inheritance)

If we have many mapped entity inside the discriminatorMap (over something like 61 in my case), it generate a MySQL error :

> SQLSTATE[HY000]: General error: 1116 Too many tables; MySQL can only use 61 tables in a join`

The problem is that Doctrine add a join statement for each known entity, even if I don't ask him any field of any sub entity.

My closure that return a QueryBuilder (for my form type), in the repository :
```php
public function getEntityTest()
    {
        return $this->getEntityManager()
            ->createQueryBuilder()
            ->select('t.id, t.slug')
            ->from('AcmeBackendCoreBundle:TestRoot', 't');
    }
```

Using the above QueryBuilder will generate that kind of SQL :
```sql
SELECT 
  t0_.id AS id0, 
  t0_.slug AS slug1 
FROM 
  test_root t0_ 
  LEFT JOIN test_sub1 t1_ ON t0_.id = t1_.id 
  LEFT JOIN test_sub2 t2_ ON t0_.id = t2_.id 
```
But I don't care about test_sub1 or test_sub2 entities... It makes no sense to use it in some cases.

Here is an easy sample to reproduce :
```php
use Doctrine\ORM\Mapping as ORM;
/**
 * TestRoot
 *
 * @ORM\Entity
 * @ORM\Table(name="test_root")
 *
 * @InheritanceType("JOINED")
 * @DiscriminatorColumn(name="type", type="string")
 * @ORM\Entity(repositoryClass="Acme\BackendCoreBundle\Repository\TestRootRepository")
 * @DiscriminatorMap({
 *      "sub_e1" = "\Acme\BackendCoreBundle\Entity\TestSub1",
 *      "sub_e2" = "\Acme\BackendCoreBundle\Entity\TestSub2",
 * })
 */
class TestRoot
{
    /**
     * @var integer
     *
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * @var String
     *
     * @ORM\Column(name="slug", type="string")
     */
    private $slug = "default_slug";

    /**
     * Get id
     *
[...]

PR #8704, opened 2021-05-21
Update inheritance-mapping.rst

juaiglesias:
I updated **Inheritance Mapping** documentation based on what we discussed in #5961. 
I added what I think is an important comment about how it is not possible to query directly the base class in a CTI schema without getting all the LEFT JOINs with the leaf entities tables.

### DIFF
--- a/docs/en/reference/inheritance-mapping.rst
+++ b/docs/en/reference/inheritance-mapping.rst
@@ -274,6 +274,9 @@ be a leaf entity in the inheritance hierarchy, (ie. have no subclasses).
 Otherwise Doctrine *CANNOT* create proxy instances
 of this entity and will *ALWAYS* load the entity eagerly.
 
+Also, there is another important performance consideration: It is *NOT POSSIBLE* 
+to query for the base entity avoiding all the LEFT JOINs with the sub-types.
+
 SQL Schema considerations
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


docs/en/reference/inheritance-mapping.rst
```diff
@@ -274,6 +274,9 @@ be a leaf entity in the inheritance hierarchy, (ie. have no subclasses).
 Otherwise Doctrine *CANNOT* create proxy instances
 of this entity and will *ALWAYS* load the entity eagerly.
 
+Also, there is another important performance consideration: It is *NOT POSSIBLE* 
+to query for the base entity avoiding all the LEFT JOINs with the sub-types.
```
SenseException: Not sure I understand this sentence considering the linked issue that led to this PR. Do you suggest to mention that it is impossible to query on a CTI without having no LEFT JOINS?

juaiglesias: Yes. What I tried to specify is that you can't query for the base entity of the inheritance without avoiding all the left joins with the leafs of the inheritance tree.

SenseException:
```suggestion
There is also another important performance consideration that it is *NOT POSSIBLE* 
to query for the base entity without any LEFT JOINs to the sub-types.
```

Maybe this is more understandable. What do you think?

SenseException (changes requested):

docs/en/reference/inheritance-mapping.rst
juaiglesias: Agreed! I will commit that

SenseException (dismissed):

greg0ire (approved):

PR #8708, opened 2021-05-22
Fix ClassMetadaInfo template inference

VincentLanglet:
Hi @greg0ire, I saw you fixed some psalm issue recently
https://github.com/doctrine/orm/commit/6de321cb09d7208ff3930db50b0bedaf7ca39d92

Currently the following code
```
    /**
     * @phpstan-template T of object
     * @phpstan-param class-string<T> $class
     * @phpstan-return ClassMetadata<T>
     */
    private function getMetadata(string $class): ClassMetadata
    {
        return new ClassMetadata($class);
    }
```
is not working, since `new ClassMetadata($class)` is inferred as `ClassMetadata<object>`.

Which lead to weird static analysis issue in 2.8.5 release.

I fix this in this PR.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -710,6 +710,7 @@ class ClassMetadataInfo implements ClassMetadata
      * metadata of the class with the given name.
      *
      * @param string $entityName The name of the entity class the new instance is used for.
+     * @psalm-param class-string<T> $entityName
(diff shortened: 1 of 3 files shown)


greg0ire:
I think 2.8.x is no longer maintained, or should not be, that's why I re-targeted this, let's see what happens to https://github.com/doctrine/orm/pull/8717

greg0ire (approved):

greg0ire:
Thanks @VincentLanglet !

PR #8713, opened 2021-05-24
Stop using Webmozart to fix error in Query/Parser

peter-gribanov:
The `Doctrine\ORM\Query\Parser` uses [`Webmozart\Assert\Assert`](https://github.com/doctrine/orm/blob/2.9.x/lib/Doctrine/ORM/Query/Parser.php#L81) but it not required in [composer.json](https://github.com/doctrine/orm/blob/2.9.x/composer.json#L18-L35).

This results in error:

```
Uncaught PHP Exception Symfony\Component\ErrorHandler\Error\ClassNotFoundError: "Attempted to load class "Assert" from namespace "Webmozart\Assert".
Did you forget a "use" statement for another namespace?" at /www/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php line 3629
```

This problem created by PR #8409

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -78,7 +78,6 @@ use Doctrine\ORM\Query\AST\UpdateStatement;
 use Doctrine\ORM\Query\AST\WhenClause;
 use Doctrine\ORM\Query\AST\WhereClause;
 use ReflectionClass;
-use Webmozart\Assert\Assert;
 
 use function array_intersect;
 use function array_search;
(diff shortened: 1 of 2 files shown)


beberlei:
@peter-gribanov thank you for making me aware of this, but the correct fix is to remove `Assert::notNull` and replace it with `assert(!is_null()))`

peter-gribanov:
There can obviously be problems with this code, since `$functionClass` can only be a `string` or `null`.

```php
$function = is_string($functionClass)
    ? new $functionClass($functionName)
    : call_user_func($functionClass, $functionName);
```

beberlei:
These problems always existed before, lets ignore them for now. A hotfix for this is important for now.

beberlei (approved):

beberlei:
@peter-gribanov thank you!

PR #8714, opened 2021-05-24
Fix links to attribute sections

mbabker:
Currently, the [attributes reference](https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/attributes-reference.html) page links to the annotations reference page when clicking one of the attributes in the index, it looks to be because the section IDs are duplicated across the two pages.  This should fix the issue by changing the reference IDs.


beberlei:
@mbabker thank you!

PR #8724, opened 2021-05-25
Add automatic type detection for Embedded.

Warxcell:
Great work on https://github.com/doctrine/orm/pull/8439 and https://github.com/doctrine/orm/pull/8589.
Following it - This adds same automatic type detection when using typed properties for Embedded. 

Needs help on PHPStan failure.

Edit: Figured it out.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -3605,9 +3605,16 @@ class ClassMetadataInfo implements ClassMetadata
     {
         $this->assertFieldNotMapped($mapping['fieldName']);
 
+        if (! isset($mapping['class']) && $this->isTypedProperty($mapping['fieldName'])) {
+            $type = $this->reflClass->getProperty($mapping['fieldName'])->getType();
+            if ($type instanceof ReflectionNamedType) {
+                $mapping['class'] = $type->getName();
+            }
+        }
+
         $this->embeddedClasses[$mapping['fieldName']] = [
             'class' => $this->fullyQualifiedClassName($mapping['class']),
-            'columnPrefix' => $mapping['columnPrefix'],
+            'columnPrefix' => $mapping['columnPrefix'] ?? null,
             'declaredField' => $mapping['declaredField'] ?? null,
             'originalField' => $mapping['originalField'] ?? null,
         ];
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -322,11 +322,9 @@ class XmlDriver extends FileDriver
                     ? $this->evaluateBoolean($embeddedMapping['use-column-prefix'])
                     : true;
 
-                $class = isset($embeddedMapping['class']) ? (string) $embeddedMapping['class'] : null;
-
                 $mapping = [
                     'fieldName' => (string) $embeddedMapping['name'],
-                    'class' => $class,
+                    'class' => isset($embeddedMapping['class']) ? (string) $embeddedMapping['class'] : null,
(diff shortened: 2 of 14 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -3605,9 +3605,16 @@ public function mapEmbedded(array $mapping)
     {
         $this->assertFieldNotMapped($mapping['fieldName']);
 
+        if (! isset($mapping['class']) && $this->isTypedProperty($mapping['fieldName'])) {
+            $type = $this->reflClass->getProperty($mapping['fieldName'])->getType();
+            if ($type) {
```
troytft:
```suggestion
            if ($type instanceof \ReflectionNamedType) {
```

to avoid  unsupported \ReflectionUnionType()

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -3605,9 +3605,16 @@ public function mapEmbedded(array $mapping)
...
+            $type = $this->reflClass->getProperty($mapping['fieldName'])->getType();
+            if ($type instanceof ReflectionNamedType) {
+                $mapping['class'] = $type->getName();
+            }
+        }
+
         $this->embeddedClasses[$mapping['fieldName']] = [
             'class' => $this->fullyQualifiedClassName($mapping['class']),
-            'columnPrefix' => $mapping['columnPrefix'],
+            'columnPrefix' => $mapping['columnPrefix'] ?? null,
```
greg0ire, line 3605: :thinking: what does this have to do with the rest of your PR?

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -322,9 +322,11 @@ public function loadMetadataForClass($className, ClassMetadata $metadata)
                     ? $this->evaluateBoolean($embeddedMapping['use-column-prefix'])
                     : true;
 
+                $class = isset($embeddedMapping['class']) ? (string) $embeddedMapping['class'] : null;
```
greg0ire: Let's inline this on line 327 and get rid of this one-time variable?

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -3605,9 +3605,16 @@ public function mapEmbedded(array $mapping)
     {
         $this->assertFieldNotMapped($mapping['fieldName']);
 
+        if (! isset($mapping['class']) && $this->isTypedProperty($mapping['fieldName'])) {
+            $type = $this->reflClass->getProperty($mapping['fieldName'])->getType();
+            if ($type instanceof ReflectionNamedType) {
+                $mapping['class'] = $type->getName();
+            }
+        }
+
```
beberlei, line 3602: You need to check here again that class isset or otherwise throw an exception. If not, then line 3616 will throw a notice accessing mapping class that does not exist.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
beberlei: I assume phpstan failure they mentioned in the PR description.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
beberlei: Actually the validation is done "late" in `ClassMetadataFactory::doLoadMetadata` as per https://github.com/doctrine/orm/pull/1133 - this should probably be changed, but not now.

beberlei (approved):

Warxcell:
@greg0ire since columnPrefix can be null, it makes no sense to make it required and pass null. So I just allow to omit it.

PR #8728, opened 2021-05-27
Remove flush($argument) from documentation as it's deprecated.

Warxcell:


SenseException (changes requested):
Thanks for pointing this out. I'd say that the 2.9,x branch should mention its deprecation while it should be removed from the docs in 3.0.x

docs/en/reference/unitofwork.rst
```diff
@@ -134,6 +134,10 @@ optimize the performance of the Flush Operation:
   explicit strategies of notifying the UnitOfWork what objects/properties
   changed.
 
+.. warning::
```
SenseException:
```suggestion
.. note::
```

Because `note` was used in other deprecation boxes.

SenseException (changes requested):

docs/en/reference/unitofwork.rst
Warxcell: Not quite I would say. 3 of 10 total deprecations are with warning.

image
image
image

Warxcell: I refactored all other deprecations to note, for sake of consistency.

SenseException (approved):
Thank you for fixing the other inconsistencies.

PR #8732, opened 2021-05-31
[GH-8723] Remove use of nullability to automatically detect nullable status

### LINKED ISSUE #8723: Upgrade from 2.8.x to 2.9.1: possible BC break for nullable behaviour
Hi everyone,

I just upgraded doctrine/orm to 2.9.1 (from 2.8.x) on a simple project.

It makes some columns that were previously not nullable, nullable (the Column annotation [does not have "nullable" set to any value in 2.9](https://github.com/doctrine/orm/blob/2.9.x/lib/Doctrine/ORM/Mapping/Column.php#L61), but from what I can see [it was false by default until 2.8.x included](https://github.com/doctrine/orm/blob/2.8.x/lib/Doctrine/ORM/Mapping/Column.php#L56)).

I think the changes happened since these PRs: https://github.com/doctrine/orm/pull/8439 & https://github.com/doctrine/orm/pull/8589

I guess the changes happen because the property in my code is nullable, and as nullable is not defined in my Column annotation, the driver thinks it's nullable.

Is it considered a BC break?

Regards,

beberlei:
The new typed properties support has a flaw around the the concept of nullable. Setting a type in the entity to be or not be nullable does not necessarily mean the databsae column has to be the same. They can differ from each other.

As such it is better to remove the the use of `allowsNull()` as the default value for a column's or joincolumn's nullable flag.

This supersedes #8726 

Fixes #8723

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1468,10 +1468,6 @@ class ClassMetadataInfo implements ClassMetadata
         $type = $this->reflClass->getProperty($mapping['fieldName'])->getType();
 
         if ($type) {
-            if (! isset($mapping['nullable'])) {
-                $mapping['nullable'] = $type->allowsNull();
-            }
-
             if (
                 ! isset($mapping['type'])
                 && ($type instanceof ReflectionNamedType)
@@ -1525,14 +1521,6 @@ class ClassMetadataInfo implements ClassMetadata
 
         if (! isset($mapping['targetEntity']) && $type instanceof ReflectionNamedType) {
             $mapping['targetEntity'] = $type->getName();
-        }
-
-        if (isset($mapping['joinColumns'])) {
-            foreach ($mapping['joinColumns'] as &$joinColumn) {
-                if ($type->allowsNull() === false) {
-                    $joinColumn['nullable'] = false;
-                }
-            }
         }
 
         return $mapping;
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -113,7 +113,7 @@ Optional attributes:
 -  **unique**: Boolean value to determine if the value of the column
    should be unique across all rows of the underlying entities table.
 
--  **nullable**: Determines if NULL values allowed for this column. If not specified, default value is false. When using typed properties on entity class defaults to true when property is nullable.
+-  **nullable**: Determines if NULL values allowed for this column. If not specified, default value is false.
 
 -  **options**: Array of additional options:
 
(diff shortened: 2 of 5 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1527,14 +1523,6 @@ private function validateAndCompleteTypedAssociationMapping(array $mapping): arr
             $mapping['targetEntity'] = $type->getName();
         }
 
-        if (isset($mapping['joinColumns'])) {
-            foreach ($mapping['joinColumns'] as &$joinColumn) {
-                if ($type->allowsNull() === false) {
```
zolex, line 1532: instead of removing it, do we maybe want the same logic as for `$mapping['nullable']`?
e.g. `if (! isset($joinColumn['nullable'])) $joinColumn['nullable'] = $type->allowsNull();`

beberlei: No, tthe column nullable behavior is also flawed, as there are legit cases where you want nullable behavior to be different between entity and database.

zolex: ah, at least for me the bug was only that I could not set the value at all in the JoinColumn annotation. it was always false even when providing it explicitly true.

but alright, I see setting the default based on the php type may be a bc where it is not defined in the annotation. so maybe this is really something for orm 3?

BusterNeece:
A question as a user: why is it preferable to remove this as a default instead of simply allowing the user to override the nullable status whenever it misaligns with the property type?

I considered the alignment with property type to be a hugely positive addition alongside the support for attributes, since it made it possible for, in a majority of cases, a simple variable declaration and single attribute to be all that is required to have a database column that corresponds intuitively to that property. That included the nullability of that property.

The changelog suggests that this fixes a bug with joinColumn nullability; rather than simply not depending on the nullable status _at all_, is there another way to resolve this problem?

zolex:
> A question as a user: why is it preferable to remove this as a default instead of simply allowing the user to override the nullable status whenever it misaligns with the property type?

because in an existing application that has an entity with a nullable typed php property and nullable not defined in the annotation/attribute, it changes the behavior, which is a breaking change. according to semantic versioning a bc is only allowed in a major version increment, so handling it strictly, it can not be in orm 2.x versions. 

let's hope we get this in 3.0, basically it's a good idea :)

Slamdunk:
@zolex sounds right, but god it feels so weird to revert everything I did alongside updating to `2.9.0`, it helped me to uncover a lot of unaligned types

BusterNeece:
Same; in upgrading to 2.9.0 I took all the nullable specifications _out_ in favor of the native PHP types, which helped me find numerous instances (particularly on `ManyToOne` joins) where nullability didn't match my expectations and I had to migrate the database to make the two match.

It's definitely a step back in the name of compatibility, but I would love to see it be something that's configurable to use now.

PR #8734, opened 2021-05-31
Fix metadata constructor inference by phpstan

### LINKED ISSUE #8709: The class ClassMetada is not correctly inferring the constructor of ClassMetadaInfo
Hi

Currently
- `new ClassMetadataInfo(\DateTime::class)` is inferred as `ClassMetadataInfo<\DateTime>` by both psalm and phpstan
- `new ClassMetadata(\DateTime::class)` is inferred as `ClassMetadata<\DateTime>` by psalm but as `ClassMetadata<object>` by phpstan

This is related to this phpstan issue https://github.com/phpstan/phpstan/issues/5057
I don't know how and when the issue will be fixed so one way to fix it temporary would be to repeat the constructor:
```
        /**
	 * @psalm-param class-string<T> $class
	 **/
	public function __construct(string $class)
	{
		parent::__construct($class);
	}
```

Since you're using phpstan in this project, you might end with this issue while bumping the phpstan level.
Do you prefer I create a PR here or in https://github.com/phpstan/phpstan-doctrine ?
(The second solution won't fix the issue here since you're not using the library here)

VincentLanglet:
Close https://github.com/doctrine/orm/issues/8709

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -5,6 +5,7 @@ parameters:
     level: 5
     paths:
         - %currentWorkingDirectory%/lib
+        - %currentWorkingDirectory%/tests/Doctrine/StaticAnalysis
     earlyTerminatingMethodCalls:
         Doctrine\ORM\Query\Parser:
             - syntaxError
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -4,8 +4,8 @@ includes:
 parameters:
     level: 5
     paths:
-        - %currentWorkingDirectory%/lib
-        - %currentWorkingDirectory%/tests/Doctrine/StaticAnalysis
+        - lib
+        - tests/Doctrine/StaticAnalysis
     earlyTerminatingMethodCalls:
         Doctrine\ORM\Query\Parser:
             - syntaxError
--- a/composer.json
+++ b/composer.json
@@ -52,6 +52,7 @@
     "autoload-dev": {
         "psr-4": {
             "Doctrine\\Tests\\": "tests/Doctrine/Tests",
+            "Doctrine\\StaticAnalysis\\": "tests/Doctrine/StaticAnalysis",
             "Doctrine\\Performance\\": "tests/Doctrine/Performance"
         }
     },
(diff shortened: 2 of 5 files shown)


phpstan.neon
```diff
@@ -5,6 +5,7 @@ parameters:
     level: 5
     paths:
         - %currentWorkingDirectory%/lib
+        - %currentWorkingDirectory%/tests/Doctrine/StaticAnalysis
```
greg0ire: Let's get rid of `%currentWorkingDirectory%` instead? It does not bring any benefit.

greg0ire (commented):
Do any errors go away if you regenerate Psalm's baseline? I think PHPStan would complain if it was the case for its baseline.

VincentLanglet:
> Do any errors go away if you regenerate Psalm's baseline? I think PHPStan would complain if it was the case for its baseline.

PHPStan does complain.
No error will be removed from Psalm baseline since Psalm already correctly inferred the constructor.

phpstan.neon
VincentLanglet: Done

greg0ire (approved):

greg0ire:
@orklah please review

ostrolucky (approved):

orklah:
It's a shame to fix that in Doctrine's code but there's not much else to do...

I'm quite puzzled by the tests though. I wonder if it's a good move to add tests to check that Psalm and PHPStan understand an arbitrary piece of code. I'm afraid future errors in those tests could lead to version stagnation for those tools while providing little value for the user (they will still upgrade their versions and their analysis will still be broken).

I'd much rather see PHPStan and Psalm add doctrine to their own CI instead (Psalm already does this for a few libraries that relies on static analysis)

greg0ire:
About the tests, this approach was suggested by both beberlei and Ocramius, see https://github.com/doctrine/orm/issues/8709#issuecomment-847303948 . IMO it's good to test that the annotations we write are correct and don't break downstream projects.

As for the deps stagnation, I'm not too worried about it, I think we will regularly upgrade just like we do now, to get bugfixes.

greg0ire:
Thanks @VincentLanglet !

PR #8738, opened 2021-06-02
EntityManager@transactional - Use return value or null

willpercey-gb:
Use expected return values or null instead of "falsey" values becoming true (empty arrays, strings)


beberlei:
We know this API of `transactional` is supoptimal, but we cannot change it this way without causing a backwards compatibility break. I hope to work on something for 3.0, but this has to wait for now, sorry :-(

As a workaround, maybe you can define your own service class wrapping the EntityManager that offers a better transactional:

```php
$transaction = new Transactional($entityManager);
$transaction->run(function () { });
```

PR #8740, opened 2021-06-03
Make ClassMetadata covariant

VincentLanglet:
The interface ClassMetada use `template-covariant`.
But the both implementations `ClassMetadaInfo` and `ClassMetada` doesn't.
I think it should. @greg0ire 

This would solve this kind of issue https://psalm.dev/r/d704dea38a
We're having this issue on SonataDoctrineORMAdminBundle for instance https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1454

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadata.php
@@ -24,7 +24,7 @@ namespace Doctrine\ORM\Mapping;
  * {@inheritDoc}
  *
  * @todo remove or rename ClassMetadataInfo to ClassMetadata
- * @template T of object
+ * @template-covariant T of object
  * @template-extends ClassMetadataInfo<T>
  */
 class ClassMetadata extends ClassMetadataInfo
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -81,7 +81,7 @@ use const PHP_VERSION_ID;
  *    get the whole class name, namespace inclusive, prepended to every property in
  *    the serialized representation).
  *
- * @template T of object
+ * @template-covariant T of object
  * @template-implements ClassMetadata<T>
  */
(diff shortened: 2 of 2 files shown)


greg0ire:
It's weird because the example in [the docs](https://psalm.dev/r/d704dea38a) mention that this is about fixing errors about `Collection<Dog>` vs `Collection<Animal>`, not `SubCollection<Animal>` vs `Collection<Animal>`…

VincentLanglet:
In my case, it's `SubCollection<Dog>` vs `Collection<Animal>`

```
Doctrine\ORM\Mapping\ClassMetadataInfo<object>
Doctrine\ORM\Mapping\ClassMetadata<Sonata\DoctrineORMAdminBundle\Tests\Fixtures\Entity\Embeddable\EmbeddedEntity>
```

greg0ire (approved):

greg0ire:
Thanks @VincentLanglet !

VincentLanglet:
You're welcome @greg0ire.

Do you think a patch could be release after the merge of https://github.com/doctrine/orm/pull/8734 ; or does this fixes are not "big enough" to trigger a release ? :)

greg0ire:
Well it would be great to get a non-CI related fix in there :sweat_smile:

VincentLanglet:
> Well it would be great to get a non-CI related fix in there 😅

I fully understand, but it will force all the static analysis to add the error to the baseline until the next patch. (like SonataDoctrineORMAdminBundle). And they could lose some time to understand the reason of the error.

But let's wait for the https://github.com/doctrine/orm/pull/8734 merge first :)

VincentLanglet:
Now https://github.com/doctrine/orm/pull/8734 is also merged.
Do you think a 2.9.3 can be considered or should we find a workaroung in SonataDoctrineORMAdminBundle ?

greg0ire:
Let me ask

greg0ire:
@VincentLanglet I asked, and the answer would be no :sweat_smile: … sorry

greg0ire:
But we might have some actual bugfixes coming soon, so maybe it will happen :)

PR #8742, opened 2021-06-05
Add ReturnTypeWillChange to ReflectionEmbeddedProperty

derrabus:
PHP 8.1 will introduce a process to add return types to internal methods, see https://wiki.php.net/rfc/internal_method_return_types

* Follows doctrine/dbal#4662
* Follows doctrine/persistence#189

This library extends `ReflectionProperty` and overrides public methods that are affected by this change. PHP 8.1 triggers deprecation warnings unless we declare that a future version of this library will add these return types as well. We can do so by adding the `ReturnTypeWillChange` attribute to those methods.

This PR suggests to do just that because adding the actual return types would break code that extends our classes. A future 3.0 release could add the actual return types.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php
+++ b/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php
@@ -22,6 +22,7 @@ namespace Doctrine\ORM\Mapping;
 
 use Doctrine\Instantiator\Instantiator;
 use ReflectionProperty;
+use ReturnTypeWillChange;
 
 /**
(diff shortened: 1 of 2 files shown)


derrabus:
The PHP CodeSniffer errors seem to be false positives.

edit: Reported as slevomat/coding-standard#1233

greg0ire:
@derrabus let's add an ignore rule with the link above then? Or is the plan to wait for an upstream fix?

derrabus:
> let's add an ignore rule with the link above then?

Done.

greg0ire (approved):

greg0ire:
Thanks @derrabus !

PR #8743, opened 2021-06-06
Update attributes-reference.rst

coldhu:
typo in UniqueConstraint attribute example "columns=" instead of "columns:"

PR #8744, opened 2021-06-07
Fix syntax

OskarStark:


beberlei:
Duplicate of #8743. a PR should fix all at once

PR #8756, opened 2021-06-09
fix: handle repeatable attributes

### LINKED ISSUE #8755: bug: AttributeDriver throws error for repeatable attributes
[`Doctrine\ORM\Mapping\Driver\AttributeDriver`](https://github.com/doctrine/orm/blob/2.9.x/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php) extends [`Doctrine\Persistence\Mapping\Driver\AnnotationDriver`](https://github.com/doctrine/persistence/blob/2.2.x/lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php).

`AttributeDriver::isTransient` resolves to `AnnotationDriver::isTransient`.

If a class has repeating attributes an error is thrown in [AnnotationDriver:182](https://github.com/doctrine/persistence/blob/79e97bcb1dfe3b242961253d81b8357a04e96ca1/lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php#L179-L182) because `$annot` is an `array`.

```php
namespace App\Entity;

#[ORM\Index(columns: ['foo'])]
#[ORM\Index(columns: ['bar', 'baz'])]
#[ORM\Entity]
class Foo
{
 //...
```

Workaround: When the repeatable attributes trail non-repeatable ones of `entityAnnotationClasses` the method returns before the error occurs.

ph-fritsche:
Introduces `isTransient` on `AttributeDriver` for better handling of repeatable attributes. Cleans up `AttributeReader` API a little.

Closes #8755

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
@@ -20,6 +20,8 @@ use function class_exists;
 use function constant;
 use function count;
 use function defined;
+use function get_class;
+use function is_array;
 
 class AttributeDriver extends AnnotationDriver
 {
@@ -36,6 +38,23 @@ class AttributeDriver extends AnnotationDriver
     public function __construct(array $paths)
     {
         parent::__construct(new AttributeReader(), $paths);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function isTransient($className)
+    {
+        $classAnnotations = $this->reader->getClassAnnotations(new ReflectionClass($className));
+
+        foreach ($classAnnotations as $a) {
+            $annot = is_array($a) ? $a[0] : $a;
+            if (isset($this->entityAnnotationClasses[get_class($annot)])) {
+                return false;
+            }
+        }
+
+        return true;
     }
 
     public function loadMetadataForClass($className, ClassMetadata $metadata): void
--- a/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
@@ -21,7 +21,6 @@ use function constant;
 use function count;
 use function defined;
 use function get_class;
-use function is_array;
 
 class AttributeDriver extends AnnotationDriver
 {
@@ -48,7 +47,7 @@ class AttributeDriver extends AnnotationDriver
         $classAnnotations = $this->reader->getClassAnnotations(new ReflectionClass($className));
 
         foreach ($classAnnotations as $a) {
-            $annot = is_array($a) ? $a[0] : $a;
+            $annot = $a instanceof RepeatableAttributeCollection ? $a[0] : $a;
             if (isset($this->entityAnnotationClasses[get_class($annot)])) {
                 return false;
             }
--- a/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
@@ -18,7 +18,6 @@ use ReflectionProperty;
 use function assert;
 use function class_exists;
 use function constant;
-use function count;
 use function defined;
 use function get_class;
 
--- a/tests/Doctrine/Tests/ORM/Mapping/AttributeDriverTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/AttributeDriverTest.php
@@ -4,9 +4,12 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\ORM\Mapping;
 
+use Attribute;
 use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\Mapping\Annotation;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
+use stdClass;
 
 use const PHP_VERSION_ID;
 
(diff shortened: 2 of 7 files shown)


lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
```diff
@@ -38,6 +40,23 @@ public function __construct(array $paths)
...
 
+    /**
+     * {@inheritDoc}
+     */
+    public function isTransient($className)
+    {
+        $classAnnotations = $this->reader->getClassAnnotations(new ReflectionClass($className));
+
+        foreach ($classAnnotations as $a) {
+            $annot = is_array($a) ? $a[0] : $a;
```
greg0ire: @ph-fritsche can `$a` really be an array?

ph-fritsche:
```php
            $annot = /* is_array($a) ? $a[0] : */ $a;
```
results in `phpunit` reporting:
```
There was 1 error:

1) Doctrine\Tests\ORM\Mapping\AttributeDriverTest::testIsTransientWithRepeatableAttributes
TypeError: get_class(): Argument #1 ($object) must be of type object, array given

/workspaces/misc/doctrine-orm/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php:52
/workspaces/misc/doctrine-orm/tests/Doctrine/Tests/ORM/Mapping/AttributeDriverTest.php:85
```
due to:
https://github.com/doctrine/orm/blob/332c314e9f055d729555f2861fcb81ee50e482b5/lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php#L26-L28
https://github.com/doctrine/orm/blob/332c314e9f055d729555f2861fcb81ee50e482b5/lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php#L66-L85

greg0ire: Indeed. The phpdoc of `convertToAttributeInstances` is wrong, this is why the build fails.

greg0ire: Also, the phpdoc of `Doctrine\Common\Annotations\Reader` is wrong, and that's in another package :grimacing:

greg0ire: Or rather `AttributeReader` does not implement the interface it says it implements. Widening the return type in the interface wouldn't be a BC break for classes implementing it, but it would be a BC break for calling code. Not sure what the best course of action would be here @beberlei . I think we should widen it in `doctrine/annotations`, but to be clean, that would require a major release :grimacing: or since it's just phpdoc, we pray it does not break to many static analysis builds and release it as a minor?

ph-fritsche: Maybe `Doctrine\ORM\AttributeDriver` should not extend `Doctrine\Persistence\Mapping\Driver\AnnotationDriver` at all.
How about we internally refactor `Doctrine\Persistence\Mapping\Driver\AnnotationDriver::getAllClassNames` into another class that could be reused by `Doctrine\ORM\AttributeDriver`?
This way `doctrine/persistence` and `doctrine/annotation` would not have any breaking changes.

If you want to be sentimental about the version increment on this package, one could argue that the declaration of `Doctrine\ORM\AttributeDriver` was broken from the beginning in `2.9` so breaking that inheritance here might be released as a fix.

beberlei: That is a course of action we could take for the next minor release. To fix this bug the small solution plus suppression of tooling errors is my preference.

greg0ire: I just thought of another quite simple solution that does not imply changing other packages: let's use an iterable object instead of an array to represent a bug of attributes. As a bonus it will make the code clearer and will answer `can $a really be an array?` because you will have `$a instanceof RepeatableAttributeCollection` instead of `is_array($a)`. WDYT?

greg0ire: Here is what it would look like: https://github.com/ph-fritsche/doctrine-orm/pull/1

beberlei:
Can you exclude AttributeReader from php-stan and psalm analysis for now? We should upgrade to check with 8.0 in 2.10 branch.

beberlei:
@ph-fritsche just thought the ping pong fixing some config changes might going to be tedious and made the change myself for now. Lets see what the CI says about this.

beberlei (dismissed):
This is good for now to fix the error.

I would say we could improve this in the future (2.10) by changing `AttributeReader` to have only three functions `getAnnotations() : array<Annotation>`, `getFilteredAnnotations(string $className) : array<Annotation>` and `getAnnotation(string $className) : Annotation` which throws when the class name is repeatable.

tests/Doctrine/Tests/ORM/Mapping/AttributeDriverTest.php
```diff
@@ -78,6 +78,13 @@ public function testOriginallyNestedAttributesDeclaredWithoutOriginalParent(): v
         );
         $this->assertEquals(['assoz_id', 'assoz_id'], $metadata->associationMappings['assoc']['joinTableColumns']);
     }
+
+    public function testIsTransientWithRepeatableAttributes(): void
+    {
+        $driver      = $this->loadDriver();
+        $isTransient = $driver->isTransient(AttributeEntityStartingWithRepeatableAttributes::class);
+        $this->assertFalse($isTransient);
+    }
```
SenseException: Covering the `true` case too would be nice.

SenseException (changes requested):

beberlei (approved):

Issue #8753, opened 2021-06-08
Edge-case regression with 2.9.x


albe:
We (neos/flow framework) suffer from a slight change in expectations for some annotation classes since #8266 - namely the `ManyToOne`, `ManyToMany` and `Embedded` now have a required first constructor argument. While those arguments were always considered "required" in a pure doctrine environment, we partially enhanced the DX by extracting some information for the annotations from other reflection data (like the `targetEntity`/`class` the annotation targets). Hence we have usage of doctrine annotations like:
```
@var Collection<SomeEntity>
@ORM\ManyToMany
```
which now breaks, because the ManyToMany is attempted to be constructed with `null` as `targetEntity` in the DocParser. Until now we solved the "missing" targetEntity (and other things) via a custom `AnnotationDriver` and that worked out quite nicely. With this change we are at a loss though, because we'd need to hook into the DocParser now, which happens way too early.

See https://github.com/neos/flow-development-collection/issues/2487#issuecomment-850405128

While we can "solve" this by restricting doctrine/orm to <2.9, we would also like to move towards PHP8 support and attributes. For us the best solution would be if those three constructors would allow `null` for their first argument, but I understand that would mean you'd need to check for that `null` at runtime which is a step backward. Optimally the Annotation would not be instanciated until it is used for mapping, but that would be too great a change.
Are there any other possible ways to "enhance" the annotations and hook into the instanciation process?

albe:
friendly ping @beberlei @greg0ire

greg0ire:
`@var SomeEntity`

Shouldn't that be `@var ArrayCollection<int, SomeEntity>`, or are you really using it that way?

Anyway, adding these constructors was technically a BC break, so it would make sense to me to allow `null` until next major to give you some wiggle room. A deprecation would have to be triggered when `null` is passed.

beberlei:
Yes that makes sense, can you make a PR for that change? Including the deprecation about nullability if the value is passed this way to the attribute/annotation class:

```php
Deprecation::trigger('doctrine/orm', 'https://github.com/doctrine/orm/issues/8753', 'Passing no target entity is deprecated.');
```

greg0ire:
I can give it a try, yes

greg0ire:
Wait, this has already been fixed in https://github.com/doctrine/orm/commit/a2230485b2f54b6046f7e1f736d15be18cfef2fa, hasn't it? The only thing that seems to be missing are deprecations…

[...]

PR #8758, opened 2021-06-09
Restore bc for annotations

greg0ire:
Some doubts I have about this:

- should deprecations be contributed to 2.10.x instead?
- are there more things to handle with deprecations (more require arguments, and more classes `OneToOne`?)

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ManyToMany.php
+++ b/lib/Doctrine/ORM/Mapping/ManyToMany.php
@@ -31,7 +31,7 @@ use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
 #[Attribute(Attribute::TARGET_PROPERTY)]
 final class ManyToMany implements Annotation
 {
-    /** @var string */
+    /** @var string|null */
     public $targetEntity;
 
     /** @var string */
@@ -61,7 +61,7 @@ final class ManyToMany implements Annotation
      * @param array<string> $cascade
      */
     public function __construct(
-        string $targetEntity,
+        ?string $targetEntity = null,
         ?string $mappedBy = null,
         ?string $inversedBy = null,
         ?array $cascade = null,
--- a/lib/Doctrine/ORM/Mapping/ManyToMany.php
+++ b/lib/Doctrine/ORM/Mapping/ManyToMany.php
@@ -22,6 +22,7 @@ namespace Doctrine\ORM\Mapping;
 
 use Attribute;
 use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
+use Doctrine\Deprecations\Deprecation;
 
 /**
  * @Annotation
@@ -69,6 +70,14 @@ final class ManyToMany implements Annotation
         bool $orphanRemoval = false,
         ?string $indexBy = null
     ) {
+        if ($targetEntity === null) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/issues/8753',
+                'Passing no target entity is deprecated.'
+            );
+        }
+
         $this->targetEntity  = $targetEntity;
         $this->mappedBy      = $mappedBy;
         $this->inversedBy    = $inversedBy;
--- a/lib/Doctrine/ORM/Mapping/ManyToOne.php
+++ b/lib/Doctrine/ORM/Mapping/ManyToOne.php
@@ -22,6 +22,7 @@ namespace Doctrine\ORM\Mapping;
 
 use Attribute;
 use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
+use Doctrine\Deprecations\Deprecation;
 
 /**
  * @Annotation
@@ -57,6 +58,14 @@ final class ManyToOne implements Annotation
         string $fetch = 'LAZY',
         ?string $inversedBy = null
     ) {
+        if ($targetEntity === null) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/issues/8753',
+                'Passing no target entity is deprecated.'
+            );
+        }
+
         $this->targetEntity = $targetEntity;
         $this->cascade      = $cascade;
         $this->fetch        = $fetch;
--- a/lib/Doctrine/ORM/Mapping/Embedded.php
+++ b/lib/Doctrine/ORM/Mapping/Embedded.php
@@ -22,6 +22,7 @@ namespace Doctrine\ORM\Mapping;
 
 use Attribute;
 use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
+use Doctrine\Deprecations\Deprecation;
 
 /**
  * @Annotation
@@ -39,6 +40,14 @@ final class Embedded implements Annotation
 
     public function __construct(?string $class = null, $columnPrefix = null)
     {
+        if ($class === null) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/issues/8753',
+                'Passing no class is deprecated.'
+            );
+        }
+
         $this->class        = $class;
         $this->columnPrefix = $columnPrefix;
     }


greg0ire:
@albe please review

lib/Doctrine/ORM/Mapping/ManyToMany.php
```diff
@@ -61,14 +62,22 @@ final class ManyToMany implements Annotation
      * @param array<string> $cascade
      */
     public function __construct(
-        string $targetEntity,
+        ?string $targetEntity = null,
```
SenseException, line 65: Why adding nullable when it gets deprecated?

greg0ire: See the commit message :)

> There used to be no constructor in this class, adding one with mandatory
arguments was technically a BC-break.

albe:
Thanks a lot! Looks good from my side, but haven't yet checked against Neos Flow if that makes things fully work. Would need to find some more time first.

> should deprecations be contributed to 2.10.x instead?

IMO as a consumer, yes, as right now that will create LOTS of warnings. But I do see how the deprecation conceptually already happened with introduction of the constructor in 2.9.0 and those deprecations are already included in some other annotation classes - sooo... guess that's what it is?

> are there more things to handle with deprecations (more require arguments, and more classes OneToOne?)

Other required arguments (no default value) are in:
- ChangeTrackingPolicy
- InheritanceType
- OrderBy

beberlei (approved):
This looks good! :ship: 

Deprecations are ok for 2.9 from my POV, especially they don't generate a lot messages as per doctrine/deprecations contract (not using `trigger_error`).

beberlei:
@albe what do you mean other required arguments are in OrderBy et al? Do they caus eproblems for Neos/Flow? If not, then some required arguments are fine imho.

albe:
No, not causing issues for us. Was just looking through all Annotation classes that introduced additional required arguments as @greg0ire asked in the PR message

greg0ire:
I guess if it's not causing issues for you, it's fine to have that "BC break" stay.

greg0ire:
Not sure which label to use here…

lib/Doctrine/ORM/Mapping/ManyToOne.php
```diff
@@ -57,6 +58,14 @@ public function __construct(
         string $fetch = 'LAZY',
         ?string $inversedBy = null
     ) {
+        if ($targetEntity === null) {
+            Deprecation::trigger(
```
mbabker, line 62: @greg0ire I realize I'm a little late to the party on this one, but I just noticed the deprecation in our dev logs today.

One of the features of 2.9 was automatic type detection when using typed properties, but these deprecations imply that with ORM 3.0 that the `$targetEntity` param will be required which negates the automatic type detection feature.  Is this intended?

greg0ire: I don't know how type detection works, but it looks kind of strange to me that the build passed prior to this, with a required constructor argument. Maybe type detection is done before building this class? If not, then something is off. Can you please find the PR which introduced automatic type detection?

mbabker: https://github.com/doctrine/orm/pull/8439 was the base PR and there are a couple follow-ons after it.

greg0ire: I could be wrong, but from this PR you linked, it looks like the detection happens in order to complete the mapping and make it valid (provide that `$targetEntity` argument) before instanciating `ManyToOne` rather than changing a `ManyToOne` instance after the fact. Also, I don't think we got complaints about this when the version with mandatory arguments was released. And also, no tests broke.

I'll stop the investigation here, if you still think it negates the automatic type detection feature, please send a failing test.

mbabker: The `ManyToMany` annotation was the only one that had the required constructor argument, which this PR did address from the B/C standpoint.

But, this PR also adds deprecations when `$targetEntity === null` for `@Embedded` (which just added type detection in https://github.com/doctrine/orm/pull/8724), `@ManyToMany`, and `@ManyToOne`, which the automatic type detection feature relies on as a DX improvement (basically, we can omit the target when using PHP 7.4+ and typed properties, the metadata mapping would infer it from the property type).  So my reading on the deprecations would be the automatic type detection feature goes away with ORM 3 because the `$targetEntity` argument would be required as basically [`ClassMetadataInfo::validateAndCompleteTypedAssociationMapping()`](https://github.com/doctrine/orm/blob/f233e4cf6b62a90a9d3ce5c5cad055dce944ce7c/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L1516-L1529) would never hit the check to assign `$mapping['targetEntity']`.

greg0ire: I'm under the impression that you think that `$mapping` is an object… the classes we are talking about do not implement `ArrayAccess` AFAIK, so the `$mapping['targetEntity']` should make you think that it is in fact an array. I also think that #8724 will make the deprecation you are observing in your logs disappear for `Embedded`.

mbabker: OK, I'm clearly not communicating my concern very well.

I have these two simplified entities in my app:

```php
<?php declare(strict_types=1);

namespace App\Entity;

use Doctrine\ORM\EntityRepository;

#[ORM\Entity(repositoryClass: EntityRepository::class)]
#[ORM\Table]
class UserDocument
{
    #[ORM\Column(nullable: false)]
    #[ORM\Id]
    #[ORM\GeneratedValue]
    private ?int $id = null;

    #[ORM\ManyToOne(cascade: ['persist', 'remove'], inversedBy: 'documents')]
    #[ORM\JoinColumn(nullable: false, onDelete: 'CASCADE')]
    public User $user;

    public function __construct(User $user)
    {
        $this->user = $user;
    }

    public function getId(): ?int
    {
        return $this->id;
    }
}
```

```php
<?php declare(strict_types=1);

namespace App\Entity;

use Doctrine\ORM\EntityRepository;

#[ORM\Entity(repositoryClass: EntityRepository::class)]
#[ORM\Table]
class User
{
    #[ORM\Column(nullable: false)]
    #[ORM\Id]
    #[ORM\GeneratedValue]
    private ?int $id = null;

    /**
     * @var Collection<UserDocument>
     */
    #[ORM\OneToMany(mappedBy: 'user', targetEntity: UserDocument::class, cascade: ['persist'], fetch: 'EXTRA_LAZY')]
    private Collection $documents;

    public function __construct()
    {
        $this->documents = new ArrayCollection();
    }

    public function getId(): ?int
    {
        return $this->id;
    }

    public function addDocument(UserDocument $document): void
    {
        $this->documents->add($document);
    }

    public function removeDocument(UserDocument $document): void
    {
        $this->documents->removeElement($document);
    }

    /**
     * @return Collection<UserDocument>
     */
    public function getDocuments(): Collection
    {
        return $this->documents;
    }

    public function hasDocument(UserDocument $document): bool
    {
        return $this->documents->contains($document);
    }

    public function getDocumentCount(): int
    {
        return $this->documents->count();
    }
}
```

This triggers the "Passing no target entity is deprecated. (ManyToOne.php:65 called by UserDocument.php:17, https://github.com/doctrine/orm/issues/8753, package doctrine/orm)" deprecation message.  The mapping works fine now because of the automatic type detection feature (ultimately `ClassMetadataInfo` handles assigning the target entity correctly because I have a typed property), but I am being told that not passing the `targetEntity` parameter is deprecated, which is the exact thing that the automatic type detection feature does for me.

Do I need to restore the `targetEntity` parameter on all of my mapping attributes or not?  Based on the deprecation coming from the `ManyToOne` class constructor, I am being told yes, but as I said in my original comment, that implies to me that the automatic type detection gets cancelled out by requiring the annotation to be configured.

greg0ire: Ah yes it makes more sense to me now, and I see that logically, a  `ManyToOne` object exists to build the mapping, not the other way around (since there are other drivers and all are meant to build the same kind of mapping array). I think you are right, the deprecations should be removed. I clearly did not have that autodetection feature in mind when writing https://github.com/doctrine/orm/issues/8753#issuecomment-857449470.

greg0ire: Working on it at https://github.com/doctrine/orm/pull/8784 :slightly_smiling_face:

greg0ire: > OK, I'm clearly not communicating my concern very well.

Thanks for being patient, I'm unfamiliar with the auto-detection feature and the fact that it's only implemented for some of the relationships did not help with understanding :sweat_smile:

PR #8764, opened 2021-06-13
Update UnitOfWork to be PHP 7.3 compatible in v2.5

zebraf1:
Fix issue on PHP 7.3 where this error is triggered.
Fatal error: Uncaught Symfony\Component\Debug\Exception\ContextErrorException: Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:2636

Doctrine/orm 2.5 is the latest version usable on Symfony 2.8 and must be compatible with PHP 7.3
Solves update conflict for Symfony 2.x projects with PHP 7.3+

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2633,7 +2633,7 @@ class UnitOfWork implements PropertyChangedListener
                         $class->reflFields[$field]->setValue($entity, $data[$field]);
                         $this->originalEntityData[$oid][$field] = $data[$field];
 
-                        continue;
+                        break;
                     }
 
                     $associatedId = array();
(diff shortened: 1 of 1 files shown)


beberlei:
2.5 is not maintained anymore please upgrade to 2.9

beberlei:
@zebraf1 You should work with your fork in your project, you can do that using composer:

```json
{
    "require": {
        "doctrine/orm": "2.5-dev"
    },
    "repositories": [
        {"type": "vcs", "url": "https://github.com/zebraf1/orm"}
    ]
}
```

It would be better if you create your own branch "php7.3-compatibility" for example and then load that via `dev-php7.3-compatibility as 2.5.10`.

See https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository for more infos.

PR #8767, opened 2021-06-15
Fix getClassMetadata phpdoc

VincentLanglet:
Hi @greg0ire, 2.9.3 fixed multiples issues but I also discovered another issue with the generics ^^'

```
@psalm-param string|class-string<T>
```
is reduce to `string` by phpstan. 

Which lead to this issue: https://phpstan.org/r/aaf083de-4264-424d-a40e-55aa6af4256c
It works fine with https://phpstan.org/r/91eed37e-b9ca-45b3-9145-fa237922d9f5

### DIFF
--- a/lib/Doctrine/ORM/EntityManagerInterface.php
+++ b/lib/Doctrine/ORM/EntityManagerInterface.php
@@ -320,7 +320,8 @@ interface EntityManagerInterface extends ObjectManager
     /**
      * {@inheritDoc}
      *
-     * @psalm-param string|class-string<T> $className
+     * @param string $className
+     * @psalm-param class-string<T> $className
      *
      * @return Mapping\ClassMetadata
      * @psalm-return Mapping\ClassMetadata<T>


lib/Doctrine/ORM/EntityManagerInterface.php
```diff
@@ -320,7 +320,8 @@ public function hasFilters();
     /**
      * {@inheritDoc}
      *
-     * @psalm-param string|class-string<T> $className
+     * @param string $className
+     * @psalm-param class-string<T> $className
```
greg0ire, line 324: This is wrong, `string` is also accepted here: https://github.com/doctrine/persistence/blob/79e97bcb1dfe3b242961253d81b8357a04e96ca1/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php#L226

VincentLanglet: Indeed, so it's a phpstan issue.

greg0ire: Probably yes. Note that I don't know what that `:` notation is useful for, maybe we should deprecate it if it's just an alternate way to do things…

PR #8781, opened 2021-06-22
Fix Psalm/PHPStan errors

derrabus:
Now that a DBAL release was tagged that is compatible with Doctrine Cache 2, PHPStan and Psalm complain about undefined legacy classes from the cache namespace. All occurrences look safe which is why I propose to ignore those errors.


greg0ire (approved):

greg0ire:
Thanks @derrabus !

PR #8784, opened 2021-06-24
Undeprecate targetEntity omission for some types

greg0ire:
Some relationship types are attached to properties that can be typed
since php 7.4.
To avoid repeating the property type in targetEntity, an autodetection
feature has been developed that allows building the mapping from that
type information and annotations.
Omitting the targetEntity stays deprecated for ManyToMany as there is no
auto-detection for that yet (although one could be built by analyzing
phpdoc annotations).


greg0ire:
@mbabker I quite tired and not sure I 100% understood the situation, so please review the code and the commit message and let me know if anything looks wrong.

mbabker (approved):
LGTM, thanks!

SenseException (approved):

derrabus (approved):

PR #8787, opened 2021-06-25
Make doctrine/annotations an optional dependency

### LINKED ISSUE #8785: Don't define Doctrine annotations as a hard dependency?
In #8266 @beberlei added support for defining mapping with PHP attributes. However, `doctrine/annotations` is still defined as a required dependency of this project:

https://github.com/doctrine/orm/blob/f233e4cf6b62a90a9d3ce5c5cad055dce944ce7c/composer.json#L22

Is there any technical reason to keep it as is, or is it just that we need to find someone willing to contribute this change? Thanks!

greg0ire:
First, there are other drivers than the annotations-based one, and
second, one of them is base on attributes, which are basically
annotations native to PHP.

Closes #8785

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -1,6 +1,11 @@
 Annotations Reference
 =====================
 
+.. note::
+
+    To be able to use annotations, you will have to install an extra
+    package called ``doctrine/annotations``.
+
 You've probably used docblock annotations in some form already,
 most likely to provide documentation metadata for a tool like
 ``PHPDocumentor`` (@author, @link, ...). Docblock annotations are a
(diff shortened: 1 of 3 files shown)


docs/en/reference/annotations-reference.rst
```diff
@@ -13,12 +13,15 @@ chances of clashes with other docblock annotations, the Doctrine ORM
...
-annotations support namespaces and nested annotations among other
-things. The Doctrine ORM ORM defines its own set of docblock
-annotations for supplying object-relational mapping metadata.
+The implementation of these enhanced docblock annotations is located in
+the ``doctrine/annotations`` package, but in the
+``Doctrine\Common\Annotations`` namespace for backwards compatibility
+reasons. Note that ``doctrine/annotations`` is not require by Doctrine
+ORM, and you will need to require that package if you want to use
+annotations. Doctrine ORM docblock annotations support namespaces and
+nested annotations among other things. The Doctrine ORM ORM defines its
```
SenseException:
```suggestion
nested annotations among other things. The Doctrine ORM defines its
```

SenseException (changes requested):
Maybe a `.. note::` at the top of docs/en/reference/annotations-reference.rst with the optional dependency information would help.

SenseException (approved):

PR #8794, opened 2021-06-27
Fix deprecated DBAL calls

derrabus:
This PR fixes some calls to deprecated DBAL methods.

### DIFF
--- a/lib/Doctrine/ORM/Id/SequenceGenerator.php
+++ b/lib/Doctrine/ORM/Id/SequenceGenerator.php
@@ -74,7 +74,7 @@ class SequenceGenerator extends AbstractIdGenerator implements Serializable
             $sql  = $conn->getDatabasePlatform()->getSequenceNextValSQL($this->_sequenceName);
 
             // Using `query` to force usage of the master server in MasterSlaveConnection
-            $this->_nextValue = (int) $conn->query($sql)->fetchColumn();
+            $this->_nextValue = (int) $conn->executeQuery($sql)->fetchOne();
             $this->_maxValue  = $this->_nextValue + $this->_allocationSize;
         }
 
--- a/tests/Doctrine/Tests/Mocks/ConnectionMock.php
+++ b/tests/Doctrine/Tests/Mocks/ConnectionMock.php
@@ -4,15 +4,19 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\Mocks;
 
+use BadMethodCallException;
 use Doctrine\Common\EventManager;
+use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Driver\Statement;
+use Doctrine\DBAL\ForwardCompatibility\Result as ForwardCompatibilityResult;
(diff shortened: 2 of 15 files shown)


tests/Doctrine/Tests/Mocks/ConnectionMock.php
```diff
@@ -74,7 +78,17 @@ public function insert($tableName, array $data, array $types = [])
      */
     public function executeUpdate($query, array $params = [], array $types = [])
     {
-        $this->_executeUpdates[] = ['query' => $query, 'params' => $params, 'types' => $types];
+        throw new BadMethodCallException(sprintf('Call to deprecated method %s().', __METHOD__));
+    }
```
greg0ire, line 82: Was this something you introduced to spot the deprecated calls and if yes, should we remove it? Or is it in order to prevent future contributors from adding more calls?

derrabus: Yes, no and yes. 🙂 

This is a mock class that is used in unit tests only. I think it‘s helpful for future contributions if the mock throws on method calls that we want to avoid.

greg0ire (approved):

beberlei (approved):
Good and simple start

greg0ire:
Thanks @derrabus !

lib/Doctrine/ORM/Id/SequenceGenerator.php
```diff
@@ -74,7 +74,7 @@ public function generate(EntityManager $em, $entity)
             $sql  = $conn->getDatabasePlatform()->getSequenceNextValSQL($this->_sequenceName);
 
             // Using `query` to force usage of the master server in MasterSlaveConnection
-            $this->_nextValue = (int) $conn->query($sql)->fetchColumn();
+            $this->_nextValue = (int) $conn->executeQuery($sql)->fetchOne();
```
stof, line 77: this change made the ORM incompatible with the `PrimaryReadReplicaConnection` (the new name of `MasterSlaveConnection`) by not respecting the comment just above it. `executeQuery` is the API where reads are allowed from the read replica
The alternative is to force connecting to the master here instead.

derrabus: Can your provide a PR to fix that? I don't use `PrimaryReadReplicaConnection` anywhere at the moment.

stof: I started a Slack discussion about that, to see what is the right fix for that.

Semantically, we can have 3 kind of execution:

- read-only queries, for which we want the result -> `executeQuery`, and `PrimaryReadReplicaConnection` let them go through the read replica (if not already connected to the primary)
- write statements without a result (e.g. `UPDATE` for instance) -> `executeStatement` (returning the number of affected rows), , and `PrimaryReadReplicaConnection` automatically ensures that they go to the primary
- statements needing write priviledges but returning a result (e.g. moving to the next value of a sequence like here) -> `query` and `PrimaryReadReplicaConnection` ensures that they go to the primary

The issue here is that `query` got deprecated in favor of `executeQuery`, meaning that we don't actually have a method for the third case anymore (which would force to add support for `PrimaryReadReplicaConnection` in each place needing that kind of query). To me, the right fix would be to undeprecate `query` (or add a new method name for that behavior, using the new return type) in DBAL, and then using it in the ORM

PR #8795, opened 2021-06-27
Update session examples in docs

SenseException:
As mentioned in https://github.com/doctrine/orm/discussions/8766 the session-section in the docs needed an update that discourages the usage of an entity in the session just as not using the methods `detach()` and `merge()`. Because users tend to copy-paste code, I struggled a bit with simple low-line examples and tried to keep them small.

### DIFF
--- a/docs/en/cookbook/entities-in-session.rst
+++ b/docs/en/cookbook/entities-in-session.rst
@@ -3,42 +3,69 @@ Entities in the Session
 
 There are several use-cases to save entities in the session, for example:
 
-1.  User object
+1.  User data
 2.  Multi-step forms
 
 To achieve this with Doctrine you have to pay attention to some details to get
 this working.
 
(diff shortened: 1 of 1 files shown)


docs/en/cookbook/entities-in-session.rst
```diff
@@ -3,42 +3,69 @@ Entities in the Session
...
 
-For a representative User object the code to get turn an instance from
-the session into a managed Doctrine object looks like this:
+For a representative User object the code to get data from the session into a
+managed Doctrine object can look like these examples:
+
+Working with scalars
+~~~~~~~~~~~~~~~~~~~~
+
+In simpler applications there is no need to work with objects in sessions and can use
```
greg0ire: "and you can use"? "and one can use"?

greg0ire (approved):

PR #8796, opened 2021-06-28
Migrate hydrators and queries to DBAL's new Result interface

derrabus:
This PR is attempt to cope with DBAL's switch from result statements to dedicated result objects.


lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -190,6 +190,27 @@ public function toIterable(ResultStatement $stmt, ResultSetMapping $resultSetMap
...
+    /**
+     * Initiates a row-by-row hydration.
+     *
+     * @deprecated
+     *
+     * @psalm-param array<string, mixed> $hints
+     *
+     * @return Generator<int, mixed>
+     */
+    public function toIterable(ResultStatement $stmt, ResultSetMapping $resultSetMapping, array $hints = []): iterable
```
derrabus: It saddens me that I've had to deprecated that shiny new method again, just because I wanted to switch the type declaration from `ResultStatement` to `Result`. Without this change, we cannot achieve DBAL 3 compatibility because `ResultStatement` has been removed in 3.0.

Changing or removing the type declaration would cause custom hydrators that override this very method to break.

If you know a better way, please advise.

greg0ire: It's indeed a bit sad but I'm unsure how commonplace it is to implement custom hydrators (although it's documented here: https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html#custom-hydration-modes). Maybe the impact is not that big?

derrabus: Plus, those custom hydrators would only break if they override `toIterable()`. If you think that this break would be tolerable, I can revert the deprecation. tbh, I feel like `toIterable()` should've been `final` anyway.

greg0ire: Let's mark it as final with `@final` then? And hope the backlash is small, and document the small BC-break.

derrabus: Done!

composer.json
```diff
@@ -30,7 +30,8 @@
         "doctrine/lexer": "^1.0",
         "doctrine/persistence": "^2.2",
         "psr/cache": "^1 || ^2 || ^3",
-        "symfony/console": "^3.0|^4.0|^5.0|^6.0"
+        "symfony/console": "^3.0|^4.0|^5.0|^6.0",
+        "symfony/polyfill-php80": "^1.23"
```
derrabus, line 34: Note: I did this to gain access to the `get_debug_type()` function. The polyfill is already a dependency of `symfony/console`, so I'm effectively not introducing a new dependency.

composer.json
```diff
@@ -30,7 +30,8 @@
         "doctrine/lexer": "^1.0",
         "doctrine/persistence": "^2.2",
         "psr/cache": "^1 || ^2 || ^3",
-        "symfony/console": "^3.0|^4.0|^5.0|^6.0"
+        "symfony/console": "^3.0|^4.0|^5.0|^6.0",
+        "symfony/polyfill-php80": "^1.15"
```
derrabus, line 34: Note: I did this to gain access to the `get_debug_type()` function. The polyfill is already a dependency of `symfony/console`, so I'm effectively not introducing a new dependency.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -151,12 +154,37 @@ public function iterate($stmt, $resultSetMapping, array $hints = [])
...
+    public function toIterable($stmt, ResultSetMapping $resultSetMapping, array $hints = []): iterable
     {
+        if (! $stmt instanceof Result) {
+            if (! $stmt instanceof ResultStatement) {
+                throw new TypeError(sprintf(
+                    '%s: Expected parameter $stmt to be an instance of %s or %s, got %s',
+                    __METHOD__,
+                    Result::class,
+                    ResultStatement::class,
+                    get_debug_type($stmt)
```
greg0ire, line 173: TIL :bulb:

composer.json
greg0ire: \* of `symfony/console` 5.3 :stuck_out_tongue:

tests/Doctrine/Tests/Mocks/HydratorMockResult.php
```diff
@@ -0,0 +1,141 @@
...
+use Traversable;
+
+use function count;
+use function current;
+use function next;
+use function reset;
+
+class HydratorMockResult implements Result
+{
+    /** @var array<int, array<string, mixed>> */
```
greg0ire: If the `int` does not matter, I'd use `list` here: although it's not required this be a list, cases where people would not pass a list are rare enough and IMO it's better if people reading that piece of code can understand that the key does not matter.

composer.json
derrabus: Console 4.4 also pulls the polyfill: https://packagist.org/packages/symfony/console#v4.4.25

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
derrabus: 🙂

tests/Doctrine/Tests/Mocks/HydratorMockResult.php
derrabus: `list<>`  is more accurate. I've changed the annotation.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @derrabus !

PR #7776, opened 2019-07-21
Remove hack to access class scope inside closures

carusogabriel:
Found some os these while working at #7688.

PR #8798, opened 2021-06-28
 Remove hack to access class scope inside closures

greg0ire:
Backport #7776 (Cc @carusogabriel)


carusogabriel (dismissed):
Thanks!

SenseException (dismissed):
Would those changes also work on 2.9.x?

greg0ire:
:thinking: since there is no risk of introducing a bug (only tests are being touched here), I think it would be better indeed on 2.9.x

PR #8800, opened 2021-06-29
Fix exception not thrown by "getEntityIdentifier"

jderusse:
In #8201, the behavior of `getEntityIdentifier` changed to return either an `array` **OR** an instance of `EntityNotFoundException`.

But everywhere else in the project, we assume the function to return an array (here is one of the 40 occurrences I found https://github.com/doctrine/orm/blob/233d9b027692cad87c970eb660958855daeb8043/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1014-L1015)

This PR change the method to throws the exception instead of returning it, and change the docblock to match the description saying ` The returned value is always an array of identifier values` (https://github.com/doctrine/orm/blob/233d9b027692cad87c970eb660958855daeb8043/lib/Doctrine/ORM/UnitOfWork.php#L3053)

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
@@ -192,14 +192,14 @@ class JoinedSubclassPersister extends AbstractEntityInheritancePersister
                 $paramIndex = 1;
                 $data       = $insertData[$tableName] ?? [];
 
-                foreach ((array) $id as $idName => $idVal) {
+                foreach ($id as $idName => $idVal) {
                     $type = $this->columnTypes[$idName] ?? Type::STRING;
 
                     $stmt->bindValue($paramIndex++, $idVal, $type);
                 }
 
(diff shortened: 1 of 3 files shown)


greg0ire:
@oojacoboo woops! Not sure how I didn't catch that during code review :sweat_smile:

greg0ire:
@jderusse I fixed the CS with `vendor/bin/phpcbf`, then addressed the static analysis issues that were about unnecessary casts and checks, now we know for sure we have an array, and added the unit test that was missing in #8201 .

jderusse:
> @jderusse I fixed the CS with `vendor/bin/phpcbf`, then addressed the static analysis issues that were about unnecessary casts and checks, now we know for sure we have an array, and added the unit test that was missing in #8201 .

ohh.. thank you so much

SenseException (approved):
Returning an exception instance is a bug, so I think this should be fine for 2.9.x.

oojacoboo:
@greg0ire yikes - don't recall that exactly, but good call on returning a single type @jderusse.

PR #8820, opened 2021-07-04
[GH-8818] Deprecate entity namespace short aliases.

### LINKED ISSUE #8818: Deprecate Foo:Bar entity short syntax
We want to deprecate and remove the short alias naming of entities in manager, repository and DQL. This feature is available through the following configuration:

```php
$configuration->addEntityNamespace('MyBundle', 'App\Entity');
$entityManager->find('MyBundle:User', $id);
$entityManager->createQuery('SELECT u FROM MyBundle:User');
```
Using short alias naming prevents Psalm/PHPStan static analysis from working and also generally is confusing for people to see that are only used to the fully qualified class name syntax.

The short alias syntax was invented when PHP did not have the magic class constant yet, so we recommend all users to change their code using short aliases to use them:

```diff
-  $entityManager->find('MyBundle:User', $id);
+  $entityManager->find(User::class, $id);
-  $entityManager->createQuery('SELECT u FROM MyBundle:User');
+  $entityManager->createQuery('SELECT u FROM ' . User::class');
```

This deprecation should be triggered only on using the aliases, not on configuring them. This way it does not get triggered if the alias is just defined but never used.

beberlei:
This PR deprecates entity namespace short aliases as advertised in issue.

Fixes #8818

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -29,6 +29,16 @@ Overriding this method is not recommended, which is why the method is documented
 + public function toIterable($stmt, ResultSetMapping $resultSetMapping, array $hints = []): iterable
 ```
 
+## Deprecated: Entity Namespace Aliases
+
+Entity namespace aliases are deprecated, use the magic ::class constant to abbreviate full class names
+in EntityManager, EntityRepository and DQL.
+
+```diff
+-  $entityManager->find('MyBundle:User', $id);
++  $entityManager->find(User::class, $id);
+```
+
 # Upgrade to 2.9
 
 ## Minor BC BREAK: Setup tool needs cache implementation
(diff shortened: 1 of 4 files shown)


greg0ire:
Blocked by https://github.com/doctrine/orm/pull/8821

greg0ire:
Let's document this in `UPGRADE.md`? Also, the wrong branch is being targeted here.

SenseException (changes requested):
https://github.com/doctrine/orm/blob/2.10.x/docs/en/reference/dql-doctrine-query-language.rst#terminals

AFAIK this seems the only occurrance of an alias in the docs. I think removing the last trails of aliases should be done with the deprecation.

SenseException (approved):

greg0ire:
Thanks @beberlei !

PR #8821, opened 2021-07-05
Adapt ignore rules to new version of PHPStan

greg0ire:


beberlei:
@greg0ire what happened here though, is phpstan not pinned to an exact version?

greg0ire:
It isn't and according to the history, has never been? I'm confused…

greg0ire:
Ah I think we did it only for Psalm. Pushed a new commit.

PR #8823, opened 2021-07-05
Referenced new support for PHP8 attributes

### LINKED ISSUE #8804: New annotation driver and typed properties not mentioned in docs
According to [this announcement post](https://www.doctrine-project.org/2021/05/24/orm2.9.html) from May 24 by @beberlei we now can use the new annotation syntax introduced in PHP 8 and also we can use typed properties to type hint associations between entities.

If I haven't missed anything, the docs make absolutely no mention of these features and only instruct the corresponding topics the "old way" (docblocks, yaml, XML, PHP).

Is this part of the docs an ongoing project? Or are these features not mature enough to be introduced into the docs?

fridde:
In reference to [this issue](https://github.com/doctrine/orm/issues/8804).

Since 2.9 there is support for attributes as referenced by @beberlei 's [blog post](https://www.doctrine-project.org/2021/05/24/orm2.9.html).

[The docs for this new feature](https://github.com/doctrine/orm/blob/2.9.x/docs/en/reference/attributes-reference.rst?rgh-link-date=2021-06-30T06%3A50%3A00Z) are written, but not referenced anywhere.

This is a attempt to correct this. There is also a slight reordering taking place to keep consistency.

Fixes #8804

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -41,6 +41,7 @@ Mapping Objects onto a Database
 
 * **Drivers**:
   :doc:`Docblock Annotations <reference/annotations-reference>` |
+  :doc:`Attributes <reference/attributes-reference>` |
   :doc:`XML <reference/xml-mapping>` |
   :doc:`YAML <reference/yaml-mapping>` |
   :doc:`PHP <reference/php-mapping>`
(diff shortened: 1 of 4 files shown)


greg0ire (approved):

beberlei (approved):

greg0ire:
Thanks @fridde !

PR #6600, opened 2017-08-03
Removing copyright headers in every file - LICENSE file is sufficient

mikeSimonson:

PR #8824, opened 2021-07-05
Drop license header

greg0ire:
Things to note:
- I'm not a lawyer
- `bin/doctrine-pear.php` mentioned `LGPL`
- all other php files mention `MIT`
- LICENSE does not mention either

Backports #6600


SenseException (approved):

greg0ire:
Github says the ORM uses the MIT license when you go on that link: https://github.com/doctrine/orm/blob/2.9.x/LICENSE

So I suppose the text is sufficient.

PR #8828, opened 2021-07-06
Clarified type hint nullability being the first default

fridde:
This adds the reference to the new mapping that uses explicit type hints and their nullability to the attribute reference doc.

It also clarifies that a property being typed as nullable will be mapped to a nullable column, in both attribute- as annotation-reference-doc.

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -113,9 +113,9 @@ Optional attributes:
 -  **unique**: Boolean value to determine if the value of the column
    should be unique across all rows of the underlying entities table.
 
--  **nullable**: Determines if NULL values allowed for this column. 
-    If not specified and a type hint is given (e.g. ?int or int), the nullability
-    of the type hint is used. Defaults to false.
+-  **nullable**: Determines if NULL values are allowed for this column. 
+   If not specified and a type hint is given (e.g. ?int or int), the nullability
+   of the type hint is used. Defaults to false.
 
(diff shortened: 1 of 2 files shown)


greg0ire (approved):

SenseException (approved):

beberlei:
Doesnt this contradict #8835?

SenseException:
@fridde Maybe you can take a look at the introduced changes in the mentioned PR and use it in yours?

fridde:
@SenseException Sorry for the delay, I was busy with other stuff.

After reading through a few threads, like [this](https://github.com/doctrine/orm/issues/8843) and [this](https://github.com/doctrine/orm/pull/8732) I managed to confuse myself and everyone else.

I agree with @beberlei that if inferring column nullability from the type's nullability (i.e. ```public string $user = null``` breaks backwards compatibility, introducing this part of the feature was a mistake and had to be reverted.

But it would be nice if someone from the core team could comment on why this inference is *not* a part of the 3.0-branch (as far as I can see)?

Is this a deliberate choice? Because the BC-break would definitely be worth it and spare future developers the headache of understanding this odd anomaly.

After the comment, I'll rewrite my PR to be less confusing and more correct.

beberlei:
@fridde Its purely an issue of time and not getting to it yet that this is not part of the 3.0 branch yet. But it will also require a very precisely located and crafted deprecation message in 2.10.x so that users can prepare for this change without false-positives.

PR #8830, opened 2021-07-06
Adding `fields` to Index

ThomasLandauer:
I guess this was just forgotten here.
Wording is taken from https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/annotations-reference.html#index


SenseException (changes requested):
Can you please also add an example with `fields`?

ThomasLandauer:
OK, like this? I think it's not worth an extra code block...

SenseException (approved):
I guess it's fine. Thank you.

PR #8831, opened 2021-07-06
refac: remove duplicate code from AbstractQuery

gimler:
`AbstractQuery::getSingleResult` is 90% the same code as `AbstractQuery::getOneOrNullResult` the only different is that `AbstractQuery::getOneOrNullResult` catch the `NoResultException` and return null.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -823,24 +823,10 @@ abstract class AbstractQuery
     public function getOneOrNullResult($hydrationMode = null)
     {
         try {
-            $result = $this->execute(null, $hydrationMode);
+            return $this->getSingleResult($hydrationMode);
         } catch (NoResultException $e) {
             return null;
         }
-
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -823,24 +823,10 @@ public function getScalarResult()
...
     {
         try {
-            $result = $this->execute(null, $hydrationMode);
+            return $this->getSingleResult($hydrationMode);
         } catch (NoResultException $e) {
             return null;
         }
-
-        if ($this->_hydrationMode !== self::HYDRATE_SINGLE_SCALAR && ! $result) {
-            return null;
```
SenseException, line 832: `getSingleResult` throws an exception here instead of returning `null`.

gimler: @SenseException yes the new code catch the NoResultException and return null same as before.

SenseException (approved):

PR #8835, opened 2021-07-08
[docs] Fix sentence about how nullable types affecting column definit…

### LINKED ISSUE #8834: Nullable property type does still not translate to nullable SQL column
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.3

#### Summary

It seems like making a property type nullable like this
```
 #[ORM\Column]
    protected ?string $Location = null;
```
does *not* translate to the SQL column being nullable, contrary to what [the docs say](https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/basic-mapping.html#php-types-mapping). It still has to be declared in the attribute as ```#[ORM\Column(nullable: true)]```.

#### How to reproduce

1. Create an entity using attributes, for example
```
#[ORM\Table(name: "users")]
class User
{
    #[ORM\Id, ORM\Column, ORM\GeneratedValue]
    protected int $id;

    #[ORM\Column]
    protected ?string $Name = null;
}
```
2. Set your mapping driver to "attribute".
3. Create a migration with ```.\bin\console make:migration```
4. Inspect the created migration and see that within the up()-method you see ```[...] Name VARCHAR(255) NOT NULL [...]```

#### Expected behavior
The last step should have contained ```[...] Name VARCHAR(255) DEFAULT NULL [...]```.

I think that @beberlei has been working on the implementation?

beberlei:
…ions.

Closes #8834

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -218,9 +218,11 @@ PHP Types Mapping
 _________________
 
 Since version 2.9 Doctrine can determine usable defaults from property types
-on entity classes. When property type is nullable the default for ``nullable``
-Column attribute is set to TRUE. Additionally, Doctrine will map PHP types
-to ``type`` attribute as follows:
+on entity classes. When property type is nullable this has no effect on
+``nullable`` Column attribute at the moment for backwards compatibility
+reasons.
+
+Additionally, Doctrine will map PHP types to ``type`` attribute as follows:
 
 - ``DateInterval``: ``dateinterval``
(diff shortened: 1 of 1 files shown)


greg0ire (approved):

SenseException (approved):

SenseException:
Thank you, @beberlei

fridde:
@beberlei This might be the wrong place to ask, but is this planned to be changed for 3.0? It would be worth the BC break, in my opinion.

Really great work on the attribute driver! My code is so much cleaner!

PR #8836, opened 2021-07-10
Make implicit type casts explicit

greg0ire:
Before we discuss what target branch the `strict_types` PR should use, here is a PR that makes 2.10.x compatible with strict types.
Everything was pretty straightforward except the sequence generator which is full of casts from int to string and _vice versa_. After reading https://github.com/doctrine/orm/pull/6683 , it seems to me that we want to be compatible with sequences greater than PHP_INT_MAX, but I doubt we are in practice.
Note how this reduces the baseline.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -614,7 +614,7 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
 
                 $sequenceGenerator = new SequenceGenerator(
                     $this->em->getConfiguration()->getQuoteStrategy()->getSequenceName($definition, $class, $this->getTargetPlatform()),
-                    $definition['allocationSize']
+                    (int) $definition['allocationSize']
                 );
                 $class->setIdGenerator($sequenceGenerator);
                 break;
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -610,13 +610,13 @@ class ClassMetadataInfo implements ClassMetadata
      * <code>
      * array(
      *     'sequenceName' => 'name',
-     *     'allocationSize' => 20,
-     *     'initialValue' => 1
+     *     'allocationSize' => '20',
+     *     'initialValue' => '1'
      * )
      * </code>
      *
      * @var mixed[]
-     * @psalm-var array{sequenceName: string, allocationSize: int, initialValue: int}
+     * @psalm-var array{sequenceName: string, allocationSize: string, initialValue: string, quoted?: mixed}
      * @todo Merge with tableGeneratorDefinition into generic generatorDefinition
      */
     public $sequenceGeneratorDefinition;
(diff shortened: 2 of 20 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -614,7 +614,7 @@ private function completeIdGeneratorMapping(ClassMetadataInfo $class): void
 
                 $sequenceGenerator = new SequenceGenerator(
                     $this->em->getConfiguration()->getQuoteStrategy()->getSequenceName($definition, $class, $this->getTargetPlatform()),
-                    $definition['allocationSize']
+                    (int) $definition['allocationSize']
```
greg0ire, line 617: Since the `+` operator is used here, this must be fine:

https://github.com/doctrine/orm/blob/95408cd8e41867c38ac9cc2395404e391b3df389/lib/Doctrine/ORM/Id/SequenceGenerator.php#L78

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -3306,14 +3306,17 @@ public function setSequenceGeneratorDefinition(array $definition)
...
         }
 
-        if (! isset($definition['initialValue']) || trim($definition['initialValue']) === '') {
+        if (! isset($definition['initialValue']) || trim((string) $definition['initialValue']) === '') {
             $definition['initialValue'] = '1';
         }
 
+        $definition['allocationSize'] = (string) $definition['allocationSize'];
+        $definition['initialValue']   = (string) $definition['initialValue'];
+
```
SenseException, line 3319: It would make sense to use a variable here now so you need to cast only one time. It increases readability  after the type casts were introduced or when changes are going to be introduced where the type casts are forgotten.

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -493,7 +493,7 @@ public function syntaxError($expected = '', $token = null)
         $message .= $expected !== '' ? sprintf('Expected %s, got ', $expected) : 'Unexpected ';
         $message .= $this->lexer->lookahead === null ? 'end of string.' : sprintf("'%s'", $token['value']);
 
-        throw QueryException::syntaxError($message, QueryException::dqlError($this->query->getDQL()));
+        throw QueryException::syntaxError($message, QueryException::dqlError($this->query->getDQL() ?? ''));
```
SenseException, line 496: When does `getDQL()` return null?

SenseException (commented):
I like how the psalm and phpstan config became smaller :+1:

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
greg0ire: Can you please show me how you would do it? It tried, and it's a bit too tricky IMO, since `$definition['allocationSize']` might not be defined.

lib/Doctrine/ORM/Query/Parser.php
greg0ire: It's `null` by default, when `setDQL()` is never called or when `free()` is called.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
SenseException: My bad, I confused keys during my review.

SenseException (approved):

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
greg0ire: No worries :)

PR #8841, opened 2021-07-13
Enable strict mode

### LINKED ISSUE #8819: Move to strict_types=1 for ORM 3
This is something we can and should do in the new major version.

greg0ire:
#8836 was not enough to catch all the needed changes, I only fixed things I detected in my local setup, but there were more casts need with PHP 8. It also seems that Psalm and PHPStan detect more things in strict mode.

If you want me to split this into 2 PRs, let me know.

Closes #8819 

Note that I'm targeting 2.x here, let me explain why:
- I don't want too much divergence between both branches, especially when they touch that many files
- there is no BC break AFAIK: strict mode is about function we call, not about our public API (and yes, the AFAIK is important, but if there is one, we can fix it with a cast IMO)
- 3.x should be about removing deprecated stuff IMO: I think risky stuff that can be fixed quickly should be diluted in minors rather than piled up in 3.x, because having risky stuff always go to 3.x will make the migration to 3.0.0 very hard for everyone. In my vision, 3.0.0 should be very boring for somebody having addressed all deprecations, and not a big deal to upgrade to.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Doctrine\ORM;
 
 use Countable;
--- a/lib/Doctrine/ORM/Cache.php
+++ b/lib/Doctrine/ORM/Cache.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Doctrine\ORM;
 
 use Doctrine\ORM\Cache\QueryCache;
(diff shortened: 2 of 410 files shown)


SenseException (dismissed):
I tried some combinations of strict type files and regular files and AFAIK2 this shouldn't be a problem, but I would wait for another reviewer to confirm this. 
The additional typecasts are manageable and okay for me.

greg0ire:
There is a conflict in the baseline, I'm going to push

SenseException (approved):

beberlei (approved):

PR #8844, opened 2021-07-14
Ditch SimpleAnnotationReader

greg0ire:
We want to get rid of it, because it is not really usable in context
when you use annotations from more than one namespace. This implies
importing classes for all annotations in use.

Backports #6549


tests/Doctrine/Tests/Models/Company/CompanyContract.php
```diff
@@ -71,7 +90,8 @@ abstract class CompanyContract
     /**
      * @var int
      * @Id
-     * @column(type="integer")
+     * @Column(type="integer")
+use Doctrine\ORM\Mapping\OrderBy;
```
SenseException: I think this is supposed to be on top of the file, but since there is no `OrderBy` needed, I assume it just needs to be removed.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7737Test.php
```diff
@@ -96,6 +102,7 @@ class GH7737Person
      * @var Collection<int, GH7737Group>
      * @ManyToMany(targetEntity=GH7737Group::class)
      * @JoinTable(inverseJoinColumns={@JoinColumn(name="group_id", referencedColumnName="id", unique=true)})
+use Doctrine\ORM\Mapping\OrderBy;
```
SenseException: Same as above.

SenseException (changes requested):

tests/Doctrine/Tests/Models/Company/CompanyContract.php
greg0ire: Oh yes, good catch!

SenseException (approved):

PR #8846, opened 2021-07-15
Use $strict param in functions that have it

greg0ire:
Backport of #6974


SenseException (dismissed):
Are those changes covered by the static analysis tools?

greg0ire:
@SenseException not sure what you mean (everything under `lib` is "covered" by static analysis)  , but this results in a small reduction of the Psalm baseline.

SenseException:
I ask this because after these changes were introduced, I would expect a build to fail when removed again. This would be the job of static analysis tools.

greg0ire:
@SenseException I pushed a new commit which uses `--generate-baseline` to remove even more things from it.
Regarding your question, not every omission of `true` causes a static analysis issue (in fact, only 2 issues were removed with this PR), which means it's not much covered by SA tools. To me, this is more akin to avoiding `empty`, and thus the job of a CS tool, and can have a beneficial impact on SA tools, but is not guaranteed to fix an SA issue every time it is applied.

SenseException (approved):

PR #7008, opened 2018-01-24
Remove unused exceptions

greg0ire:

PR #8851, opened 2021-07-20
Remove unused exceptions

greg0ire:
After discussing this with @beberlei , we picked 2.10.x as a target branch since these methods are very unlikely to be used outside the project.

Backports #7008


SenseException (approved):
Instances of ORM exceptions shouldn't be created outside of ORM.

PR #6812, opened 2017-11-06
Fix syntax error when join unrelated entity with discriminator entity

Deltachaos:
When joining a unrelated entity with a discriminator the join contains the table name and alias in brackets which is a syntax error (at least in SQL Server).

Exmaple:

```sql
SELECT
  c
FROM
  Doctrine\Tests\Models\Company\CompanyContract c
JOIN
  c.salesPerson s
LEFT JOIN
  Doctrine\Tests\Models\Company\CompanyEvent e WITH s.id = e.id
```
Results in:
```sql
SELECT
  c0_.id AS id_0,
  c0_.completed AS completed_1,
  c0_.fixPrice AS fixPrice_2,
  c0_.hoursWorked AS hoursWorked_3,
  c0_.pricePerHour AS pricePerHour_4,
  c0_.maxPrice AS maxPrice_5,
  c0_.discr AS discr_6
FROM
  company_contracts c0_
INNER JOIN
  company_employees c1_ ON c0_.salesPerson_id = c1_.id
LEFT JOIN
  company_persons c2_ ON c1_.id = c2_.id
LEFT JOIN
  (company_events c3_) ON (c2_.id = c3_.id)
WHERE
  c0_.discr IN ('fix', 'flexible', 'flexultra')
```

After the fix it results in:
```sql
SELECT
  c0_.id AS id_0,
  c0_.completed AS completed_1,
  c0_.fixPrice AS fixPrice_2,
  c0_.hoursWorked AS hoursWorked_3,
  c0_.pricePerHour AS pricePerHour_4,
  c0_.maxPrice AS maxPrice_5,
  c0_.discr AS discr_6
FROM
  company_contracts c0_
INNER JOIN
  company_employees c1_ ON c0_.salesPerson_id = c1_.id
LEFT JOIN
  company_persons c2_ ON c1_.id = c2_.id
LEFT JOIN
  company_events c3_ ON (c2_.id = c3_.id)
WHERE
  c0_.discr IN ('fix', 'flexible', 'flexultra')
```

I have created #6811 PR for backporting this to the 2.5 branch as well.

PR #8855, opened 2021-07-23
Failing test for GH8443

piowin:
A test suit for #8443.

Closes #8443

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8443Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8443Test.php
@@ -1,0 +1,140 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Query;
+use Doctrine\Tests\Models\Company\CompanyManager;
+use Doctrine\Tests\Models\Company\CompanyPerson;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+use function assert;
+use function count;
+
+final class GH8443Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        $this->useModelSet('company');
+        parent::setUp();
+        $this->_schemaTool->createSchema([$this->_em->getClassMetadata(GH8443Foo::class)]);
+    }
+
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+        $this->_schemaTool->dropSchema([$this->_em->getClassMetadata(GH8443Foo::class)]);
+    }
+
+    /**
+     * @group GH-8443
+     */
+    public function testJoinRootEntityWithForcePartialLoad(): void
+    {
+        $person = new CompanyPerson();
+        $person->setName('John');
+
+        $manager = new CompanyManager();
+        $manager->setName('Adam');
+        $manager->setSalary(1000);
+        $manager->setDepartment('IT');
+        $manager->setTitle('manager');
+
+        $manager->setSpouse($person);
+
+        $this->_em->persist($person);
+        $this->_em->persist($manager);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $manager = $this->_em->createQuery(
+            "SELECT m from Doctrine\Tests\Models\Company\CompanyManager m
+               JOIN m.spouse s
+               WITH s.name = 'John'"
+        )->setHint(Query::HINT_FORCE_PARTIAL_LOAD, true)->getSingleResult();
+        $this->_em->refresh($manager);
+
+        $this->assertEquals('John', $manager->getSpouse()->getName());
+    }
+
+    /**
+     * @group GH-8443
+     */
+    public function testJoinRootEntityWithOnlyOneEntityInHierarchy(): void
+    {
+        $bar = new GH8443Foo('bar');
+
+        $foo = new GH8443Foo('foo');
+        $foo->setBar($bar);
+
+        $this->_em->persist($bar);
+        $this->_em->persist($foo);
+        $this->_em->flush();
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH8443Test.php
```diff
@@ -0,0 +1,130 @@
...
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+use function count;
+
+final class GH8443Test extends OrmFunctionalTestCase
+{
+    /**
+     * @group GH-8443
+     */
+    public function testJoinRootEntityWithForcePartialLoad(): void
```
greg0ire, line 33: Running that test by itself fails with `SQLSTATE[HY000]: General error: 1 no such table: company_persons`.
Tests should be independent from each other

tests/Doctrine/Tests/ORM/Functional/Ticket/GH8443Test.php
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * @group GH-8443
+     */
+    public function testJoinRootEntityWithOnlyOneEntityInHierarchy(): void
+    {
+        try {
+            $this->_schemaTool->createSchema([$this->_em->getClassMetadata(GH8443Foo::class)]);
+        } catch (Exception $e) {
```
greg0ire: There should be an import for this class

tests/Doctrine/Tests/ORM/Functional/Ticket/GH8443Test.php
```diff
@@ -0,0 +1,130 @@
...
+        $manager->setTitle('manager');
+
+        $manager->setSpouse($person);
+
+        $this->_em->persist($person);
+        $this->_em->persist($manager);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $result = $this->_em->createQuery("select m from Doctrine\Tests\Models\Company\CompanyManager m JOIN m.spouse s WITH s.name = 'John'")->setHint(Query::HINT_FORCE_PARTIAL_LOAD, true)->getResult();
```
greg0ire: Why not use `getSingleResult`?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH8443Test.php
```diff
@@ -0,0 +1,130 @@
...
+        $this->assertEquals(count($result), 1);
+
+        $manager = $result[0];
+        $this->_em->refresh($manager);
+
+        $this->assertEquals($manager->getSpouse()->getName(), 'John');
```
greg0ire:
```suggestion
        $this->assertEquals('John', $manager->getSpouse()->getName());
```

greg0ire:
> The same problem appears if you add WITH condition to the joined entity with discriminator

As opposed to what? https://github.com/doctrine/orm/pull/6812 is also about `WITH`, so I'm confused…

piowin:
> 
> 
> > The same problem appears if you add WITH condition to the joined entity with discriminator
> 
> As opposed to what? #6812 is also about `WITH`, so I'm confused…

You're right. In both issues the problem was a result of using a with condition. I think the code of SqlWalker::walkJoinAssociationDeclaration could be simplfy a little and the #8443 wouldn't be necessary, but then the code doesn't pass some tests of the SelectSqlGeneratorTest, even though the sql syntax is correct.

greg0ire:
I'll rephrase my question: what is the difference between both issues? Since the first patch did not fix the second issue, there must be some difference somewhere, right?

piowin:
The sqls are generated in different methods with different logic.

greg0ire:
Ok, why is that?

piowin:
I haven't designed the SqlWalker, but I see that, in the terms of the EBNF of DQL, the logic of the generation of Join is more complex when we join JoinAssociationDeclaration than when we join RangeVariableDeclaration . Some of the logic is similar and maybe it would be possible to put that logic in a separate method.

greg0ire:
> when we join JoinAssociationDeclaration than when we join RangeVariableDeclaration

Ah, so that's the difference! How do each of these look like in DQL?

piowin:
A JoinAssociationDeclaration  is a path to an entity's field (the field can be a collection of entities, a single entity or a simple state field) plus an alias: p.friend AS f (p must be defined before and that's one of the differences with the next one).

A RangeVariableDeclaration is a path to a class plus an alias: \MyNamespace\Person AS p. A RangeVariableDeclaration is the obligatory component that must be placed immediately after the "FROM" clause.

A JoinAssociationDeclaration  can be only used as a joined component while A RangeVariableDeclaration can be used as a joined component and a starting point/base of an IdentificationVariableDeclaration which is a building block of the "FROM" clause.

How does the method responsible for generating joins work?
When the sql for a Join is being generated, if the Join is a JoinAssociationDeclaration, it simply delegates to the method that is responsible for generating JoinAssociationDeclaration, that's all. But if the Join is a RangeVariableDeclaration, it generates optional conditions for the RangeVariableDeclaration, like a filter condition, a "with" condition or a discriminator condition. The generation of each conidtion has its unique logic. Also the sql for the RangeVariableDeclaration is being generated by the method that is responsible for that (the method is invoked also when we genarate IdentificationVariableDeclaration, see above). The conditions are added to the sql and the result is returned.

The logic of the generation of the optional brackets (the subject of our discussion) is put in the method that is reponsible for generating JoinAssociationDeclaration, if the Join is a JoinAssociationDeclaration, and is put in the method that is responsible for generating joins, if the Join is a RangeVariableDeclaration.

greg0ire (approved):
Very clear! Thank you!

SenseException (approved):

greg0ire:
Thanks @piowin!

PR #8862, opened 2021-07-24
Change $this->assert* to self::assert* in unit tests

carnage:
As discussed in #6854 this PR updates all the unit tests to use self::assert* instead of $this->assert* 

This was already done in the v3 develop branch and it'll smooth merging one of the remaining backports. 

It was generated using this command: `php ./php-cs-fixer-v2.phar fix ./tests/ --rules='{"php_unit_test_case_static_method_calls":{"call_type": "self"}}' --allow-risky=yes` I would suggest if there is a delay in merging and conflicts come up, that someone just reruns that on a new branch taken from 2.10.x instead of trying to fix the conflicts.


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @carnage !

carnage:
Will try and find time at the weekend to have another go at backporting the test improvements PR now this should solve a lot of the conflicts

PR #8863, opened 2021-07-25
Fix loading metadata for cold reflection class

Deltachaos:
When using discriminator map attributes the reflection class can be null


beberlei:
Can you give much more infos please? What is the error? How do you get it? This change alone can't be merged, need to identify the cause

j-mahlke:
This already has been fixed in the Annotation, but is missing in the Attributes.
 https://github.com/doctrine/orm/blob/e313d012aea7a904ecceb78f70fcab1b66932393/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L56
 
 https://github.com/doctrine/orm/commit/e16803de618ef0b94b1fa0917bf0fefff23e07de

Issue #6429, opened 2017-05-08
GeneratedValue strategy AUTO should use IDENTITY on PostgreSQL


Seldaek:
As per http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/basic-mapping.html#identifier-generation-strategies - AUTO uses SEQUENCE by default on PostgreSQL. This seems sub-optimal as it forces INSERT queries to provide the id explicitly using `nextval('sequence_name')` which is not easy to do in ORM queries, and requires that you know the sequence name. On the other hand if it's set to IDENTITY it works better as it does it like MySQL's AUTO_INCREMENT, you can just omit the id field entirely.

As doing this would cause schema changes I can imagine it won't be fixed in a minor release, but should be considered for v3 I think.

Majkl578:
AFAIK this would only work with SERIAL fields (fields with default value being `nextval('...')`). This is not default behavior and IMHO should not be considered as such.

flaushi:
Hmm, I also stumpled upon this issue.
I have an id field with `@GeneratedValue`,  which gives `bin/console doctrine:schema:create --dump-sql`:
```
CREATE TABLE data_sample (id INT NOT NULL, hashvalue VARCHAR(64) DEFAULT NULL, date TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, refdate TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, value NUMERIC(20, 2) NOT NULL, PRIMARY KEY(id));

CREATE SEQUENCE data_sample_id_seq INCREMENT BY 1 MINVALUE 1 START 1;
``` 
So the sequence is not assigned as default value. However, when adding entitities, the id is incremented. I don't understand why.

DrummerKH:
@flaushi Because of doctrine getting ID from the sequence and add to the INSERT statement.

breitsmiley:
Since PostgreSQL 10 more pefered way for autoincrement id is GENERATED AS IDENTITY syntax

breitsmiley:
Since PostgreSQL 10 more prefered way for using autoincrement is **GENERATED AS IDENTITY** that is very similar to mysql AUTO_INCREMENT

It would be great to support by doctrine this features

```
CREATE TABLE color (
    color_id INT GENERATED ALWAYS AS IDENTITY,
    color_name VARCHAR NOT NULL
);
```
[...]

PR #8870, opened 2021-07-30
Remove calls to prefersSequences()

derrabus:
`AbstractPlatform::prefersSequences()` is deprecated and does not exist in DBAL 3. Let's not call it if it's not there.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -548,13 +548,7 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
     {
         $idGenType = $class->generatorType;
         if ($idGenType === ClassMetadata::GENERATOR_TYPE_AUTO) {
-            if ($this->getTargetPlatform()->prefersSequences()) {
-                $class->setIdGeneratorType(ClassMetadata::GENERATOR_TYPE_SEQUENCE);
-            } elseif ($this->getTargetPlatform()->prefersIdentityColumns()) {
-                $class->setIdGeneratorType(ClassMetadata::GENERATOR_TYPE_IDENTITY);
-            } else {
-                $class->setIdGeneratorType(ClassMetadata::GENERATOR_TYPE_TABLE);
-            }
+            $class->setIdGeneratorType($this->determineIdGeneratorStrategy($this->getTargetPlatform()));
         }
 
         // Create & assign an appropriate ID generator instance
@@ -656,6 +650,23 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
             default:
                 throw UnknownGeneratorType::create($class->generatorType);
         }
+    }
+
+    private function determineIdGeneratorStrategy(AbstractPlatform $platform): int
+    {
+        if ($platform->getName() === 'oracle' || $platform->getName() === 'postgresql') {
+            return ClassMetadata::GENERATOR_TYPE_SEQUENCE;
+        }
+
+        if ($platform->supportsIdentityColumns()) {
+            return ClassMetadata::GENERATOR_TYPE_IDENTITY;
+        }
+
+        if ($platform->supportsSequences()) {
+            return ClassMetadata::GENERATOR_TYPE_SEQUENCE;
+        }
+
+        return ClassMetadata::GENERATOR_TYPE_TABLE;
     }
 
     /**
--- a/tests/Doctrine/Tests/Mocks/DatabasePlatformMock.php
+++ b/tests/Doctrine/Tests/Mocks/DatabasePlatformMock.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\Mocks;
 
+use BadMethodCallException;
+use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 
 /**
@@ -11,29 +13,30 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
  */
 class DatabasePlatformMock extends AbstractPlatform
 {
-    /** @var string */
-    private $_sequenceNextValSql = '';
+    /** @var bool */
+    private $supportsIdentityColumns = true;
 
     /** @var bool */
-    private $_prefersIdentityColumns = true;
+    private $supportsSequences = false;
 
-    /** @var bool */
-    private $_prefersSequences = false;
-
-    /**
-     * {@inheritdoc}
-     */
-    public function prefersIdentityColumns()
+    public function prefersIdentityColumns(): bool
     {
-        return $this->_prefersIdentityColumns;
+        throw new BadMethodCallException('Call to deprecated method.');
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function prefersSequences()
+    public function supportsIdentityColumns(): bool
     {
-        return $this->_prefersSequences;
+        return $this->supportsIdentityColumns;
+    }
+
+    public function prefersSequences(): bool
+    {
+        throw new BadMethodCallException('Call to deprecated method.');
+    }
+
+    public function supportsSequences(): bool
+    {
+        return $this->supportsSequences;
     }
 
     /**
@@ -41,7 +44,7 @@ class DatabasePlatformMock extends AbstractPlatform
      */
     public function getSequenceNextValSQL($sequenceName)
     {
-        return $this->_sequenceNextValSql;
+        return '';
     }
 
     /**
@@ -95,19 +98,14 @@ class DatabasePlatformMock extends AbstractPlatform
 
     /* MOCK API */
 
-    public function setPrefersIdentityColumns(bool $bool): void
+    public function setSupportsIdentityColumns(bool $bool): void
     {
-        $this->_prefersIdentityColumns = $bool;
+        $this->supportsIdentityColumns = $bool;
     }
 
-    public function setPrefersSequences(bool $bool): void
+    public function setSupportsSequences(bool $bool): void
     {
-        $this->_prefersSequences = $bool;
-    }
-
-    public function setSequenceNextValSql(string $sql): void
-    {
-        $this->_sequenceNextValSql = $sql;
+        $this->supportsSequences = $bool;
     }
 
     /**
--- a/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
@@ -59,8 +59,8 @@ class ClassMetadataFactoryTest extends OrmTestCase
 
         $conn         = $entityManager->getConnection();
         $mockPlatform = $conn->getDatabasePlatform();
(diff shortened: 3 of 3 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -548,10 +549,12 @@ private function completeIdGeneratorMapping(ClassMetadataInfo $class): void
     {
         $idGenType = $class->generatorType;
         if ($idGenType === ClassMetadata::GENERATOR_TYPE_AUTO) {
-            if ($this->getTargetPlatform()->prefersSequences()) {
+            if (method_exists(AbstractPlatform::class, 'prefersSequences') && $this->getTargetPlatform()->prefersSequences()) {
```
greg0ire: :thinking: I think the logic is wrong… let's not call that method for as long as it exists, but call its replacement (`supportsSequence()` ? It's not clear to me, this method was first mentioned in the codebase in 2008) as soon as possible. If it turns out we can remove the calls entirely, let's do that instead?

derrabus: I was afraid I might change the behavior for DBAL drivers that implement `supportsSequences() === true`, but `prefersSequences() === false`. If you tell me that this is not an issue, I can remove that condition. Please advice.

greg0ire: @morozov @beberlei maybe you will know the answer? I don't.

morozov: There's no DBAL logic that would rely on `prefersSequences()`, it was just a bunch of `true`s and `false`s returned depending on the platform. I have no idea why the ORM would prefer one over the other.

Note, this whole thing may become irrelevant in light of the removal of sequence-based id generation (see https://github.com/doctrine/dbal/issues/4687, but it's only removed in `4.0.x`).

morozov: As far as I can tell, the only platform that supports sequences but doesn't prefer sequences is SQL Server because it supports identity columns as well. So in theory, a call to `prefersSequences()` could be replaced with `supportsSequences() && !supportsIdentityColumns()`. This may preserve the current behavior but I still don't know what problem this logic solves in the first place.

derrabus: Thank you. I've removed the first `if`. Neither `prefers*()` method is called now and if a driver supports both, identity columns and sequences, identity columns win.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -43,6 +43,7 @@
 use function explode;
 use function in_array;
 use function is_subclass_of;
+use function method_exists;
```
morozov: This looks no longer relevant but isn't caught by PHP_CodeSniffer for some reason.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
greg0ire: > identity columns and sequences, identity columns win.

As a consequence, some PostgreSQL-related tests fail now.

tests/Doctrine/Tests/Mocks/DatabasePlatformMock.php
```diff
@@ -20,18 +20,22 @@ class DatabasePlatformMock extends AbstractPlatform
     /** @var bool */
     private $_prefersSequences = false;
 
-    /**
-     * {@inheritdoc}
-     */
-    public function prefersIdentityColumns()
+    public function prefersIdentityColumns(): bool
```
morozov, line 22: While this approach is okay, it would be better if these mocks were reworked using the PHPUnit mocking API. See https://github.com/doctrine/dbal/pull/3555 and https://github.com/doctrine/dbal/pull/3474 for example. This could be done separately but before handling other API changes.

derrabus: See https://github.com/doctrine/orm/issues/8886#issuecomment-893391508

I'm planning to revise the mocks next weekend. This is a major pain point of the DBAL upgrade in this repository: too much mocking of classes that belong to a different package.

morozov: Awesome!

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
morozov: Gotta love the tests that make assertions against SQL. Especially the SQL generated by a different package. Should the ORM even care about SQL? I'd argue that even the DBAL shouldn't. There are multiple ways to do the same thing, not even speaking of the SQL formatting and case sensitivity.

morozov: > As a consequence, some PostgreSQL-related tests fail now.

This is probably why this "prefers" API existed. PostgreSQL also supports sequences and identity columns but for some reason, the ORM prefers sequences when working with PostgreSQL. Maybe this is because identity columns in PostgreSQL are emulated internally via sequences? But in any event, it doesn't seem the ORM and not even the DBAL concern.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
derrabus: Fixed. There are several other broken imports in that file, btw. :-/

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
derrabus: So how shall we proceed here? Even if I went back to my previous approach, the Postgres tests would fail again as soon as we upgraded to DBAL 3 where the `prefers*()` methods are gone. My best idea would be to duplicate DBAL's concept of the `*Platform` classes for all features that have been removed from DBAL (see also #8887).

Or… well… to re-add those methods to DBAL, of course. 😓

greg0ire: 11 years ago in the first commit (69158cd4f3074fb2a08c53375acd1c29087d0040), which is clearly an import from something written even before, it was documented that `IDENTITY` was not supported by PostgreSQL:

> `IDENTITY`: Tells Doctrine to use special identity columns in the database that usually generate a value on insertion of a row (i.e. MySql AUTO_INCREMENT). This strategy does currently not provide full portability. IDENTITY is supported by MySql, Sqlite and MsSql.

The current docs state this is not true anymore:

https://github.com/doctrine/orm/blob/95408cd8e41867c38ac9cc2395404e391b3df389/docs/en/reference/basic-mapping.rst#L377-L381

So IMO it's fine no longer to prefer sequences when working with a recent enough PostgreSQL version, not sure which, but maybe @morozov knows.

derrabus: > So IMO it's fine no longer to prefer sequences when working with a recent enough PostgreSQL version, not sure which, but maybe @morozov knows.

Assuming we simply changed the ID generation strategy that is picked if `@GeneratedValue(strategy="AUTO")` was used on Postgres: Would it be safe for existing applications with a Postgres database to upgrade ORM?

greg0ire: :thinking: hmmm… obviously not, the strategy is probably taken into account when creating a new record in the database right? If the ORM assumes `IDENTITY` is used, when a sequence actually is => :boom:  :disappointed: 

Doing a search leads to https://github.com/doctrine/orm/issues/6429  , which points out that such changes would require a migration.

Maybe the real problem is having `AUTO`: if in 2021 strategy X is optimal for Postgres, and then in 2022 strategy Y is optimal, but changing strategies causes a crash because Doctrine looks for database objects such as sequences and whatnot that do not exist, this will be an issue every time a new strategy is identified as optimal. Maybe we should deprecate `AUTO`.

greg0ire: To circle back to what possible solution we would have here, I think the best would be to move the "preference" concept to the ORM (and instantly deprecate it as well as `AUTO`): the ORM should not have a preference because it cannot change, since we cannot know what will be the best choice in the future.

derrabus: > 🤔 hmmm… obviously not, the strategy is probably taken into account when creating a new record in the database right? If the ORM assumes `SERIAL` is used, when a sequence actually is => 💥 😞

Yes, that's my expectation as well.

> Maybe we should deprecate `AUTO`.

That's probably a topic that should be discussed in a dedicated ticket. For this PR, my conclusion is: However we want to proceed with the `AUTO` strategy in the future, the behavior of that strategy for a specific database must not change.

derrabus:
I've pushed a new attempt, this time with special handling for Postgres and Oracle. Those were the only DBAL platforms that preferred sequences.

Theoretically, this would still be a breaking change, if someone used a custom DBAL driver implementation (is this a thing?) that supports sequences and identity columns _and_ prefers sequences.

greg0ire:
> Theoretically, this would still be a breaking change, if someone used a custom DBAL driver implementation (is this a thing?) that supports sequences and identity columns and prefers sequences.

Thanks for mentioning that, it sounds pretty unlikely.

greg0ire (approved):

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
greg0ire: I created https://github.com/doctrine/orm/issues/8893 for that separate discussion

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -658,6 +652,23 @@ private function completeIdGeneratorMapping(ClassMetadataInfo $class): void
         }
     }
 
+    private function determineIdGeneratorStrategy(AbstractPlatform $platform): int
+    {
+        if ($platform->getName() === 'oracle' || $platform->getName() === 'postgresql') {
```
morozov, line 657: Oracle doesn't support identity columns, so it seems like there's no need to list it here.

morozov:
Since there's an ongoing discussion in https://github.com/doctrine/orm/issues/8893, let's agree on something there first to make sure the changes here correspond to the agreement or explicitly agree that the discussion shouldn't hold the merge. Except for https://github.com/doctrine/orm/pull/8870#discussion_r685503430, looks good to me.

greg0ire:
I'd say doctrine/orm#8893 shouldn't hold the merge, these problems look independent to me.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
derrabus: > Oracle doesn't support identity columns

That is correct. My intention here was to prevent ORM from changing behavior if one fine day Oracle starts to support identity columns.

morozov (approved):

greg0ire:
Thanks @derrabus !

PR #8872, opened 2021-08-01
Fix CI SA failures

simPod:


greg0ire:
@simPod this looks like a bugfix. Shouldn't it target 2.9.x?

simPod:
Was not aware those issues disappeared on 2.9.x as well. Retargeted.

greg0ire (approved):

SenseException (approved):
Do we have any tests for this class?

greg0ire:
The test suite passes even after throwing right at the beginning of `execute`, so I don't think we do.

greg0ire:
Thanks @simPod !

PR #8894, opened 2021-08-07
Fix return types for PHP 8.1

derrabus:
This PR adds a `ReturnTypeWillChange` to iplementations of various core interfaces to indicate that we intend to add a return type declaration one fine day. This way, we avoid a deprecation warning in PHP 8.1.


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @derrabus !

PR #8895, opened 2021-08-07
Implement __serialize() and __unserialize()

derrabus:
The `Serializable` interface is deprecated in favor of the magic methods `__serialize()` and `__unserialize()`. PHP 8.1 will trigger a deprecation warning if a class implements `Serializable` but not those two magic methods.

This PR addresses this issue by adding `__serialize()` and `__unserialize()` to the `SequenceGenerator` class.

This does however change the serialized string on PHP 7.4 and above. I don't know if that is a problem.

### DIFF
--- a/lib/Doctrine/ORM/Id/SequenceGenerator.php
+++ b/lib/Doctrine/ORM/Id/SequenceGenerator.php
@@ -103,27 +103,43 @@ class SequenceGenerator extends AbstractIdGenerator implements Serializable
 
     /**
      * @return string
+     *
+     * @final
      */
     public function serialize()
     {
-        return serialize(
-            [
-                'allocationSize' => $this->_allocationSize,
-                'sequenceName'   => $this->_sequenceName,
-            ]
-        );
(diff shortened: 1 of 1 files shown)


greg0ire (approved):
I don't know either if the string change will be an issue.

SenseException (approved):
Will this string be somehow shared between versions for any reason in a project? Otherwise I don't see an issue with this.

beberlei:
This could be problematic imho in a patch release as we don't have the version number in the cache key of Metadata Cache. What would happen if on PHP 7.4 this was serialized before this change and then unserialized after this change? Will that work?

beberlei:
Tested it and it still works to unserialize an old string, so good to go from my POV.

beberlei (approved):

derrabus:
Unserializing old strings still works as long as we keep the `Serializable` implementation. But an object serialized with PHP 7.4 won't be unserializable for PHP 7.3 anymore.

greg0ire:
Thanks @derrabus !

PR #8896, opened 2021-08-07
Don't pass null as parameter

derrabus:
Passing `null` to various core functions triggers a deprecation warning in PHP 8.1. This PR attempts to fix this kind of error.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -511,7 +511,7 @@ class Parser
         $message .= $expected !== '' ? sprintf('Expected %s, got ', $expected) : 'Unexpected ';
         $message .= $this->lexer->lookahead === null ? 'end of string.' : sprintf("'%s'", $token['value']);
 
-        throw QueryException::syntaxError($message, QueryException::dqlError($this->query->getDQL()));
+        throw QueryException::syntaxError($message, QueryException::dqlError($this->query->getDQL() ?? ''));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -511,7 +511,7 @@ public function syntaxError($expected = '', $token = null)
         $message .= $expected !== '' ? sprintf('Expected %s, got ', $expected) : 'Unexpected ';
         $message .= $this->lexer->lookahead === null ? 'end of string.' : sprintf("'%s'", $token['value']);
 
-        throw QueryException::syntaxError($message, QueryException::dqlError($this->query->getDQL()));
+        throw QueryException::syntaxError($message, QueryException::dqlError($this->query->getDQL() ?? ''));
```
derrabus, line 514: 💡 `$message` of `Exception::__construct()` must not be `null`.

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -528,7 +528,7 @@ public function syntaxError($expected = '', $token = null)
     public function semanticalError($message = '', $token = null)
     {
         if ($token === null) {
-            $token = $this->lexer->lookahead ?? ['position' => null];
+            $token = $this->lexer->lookahead ?? ['position' => 0];
```
derrabus, line 531: 💡 `$token['position']` is later used as `$offset` for `substr()`.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @derrabus !

Issue #8886, opened 2021-08-05
Get PHPUnit to run with DBAL 3


greg0ire:
[Recently, a CI job was added on the 2.10.x branch to run the test suite this project after forcing the installation of `doctrine/dbal` v3](https://github.com/doctrine/orm/pull/8874).

In order not to disrupt contributors, we made sure it would not fail the build although it really does not pass at all:

https://github.com/doctrine/orm/blob/e09f126abfec0add53dc1ed8d4e3708117032fc1/.github/workflows/static-analysis.yml#L52

Right now, we get a crash without PHPUnit actually running, so we don't even know how many errors and failures there will be in the end. A good first step would be to fix that crash and get to a state where you can see the whole test suite running (with errors and failures)
If you want to help us fix it or get closer to fixing it, you can reproduce the issue locally by running (from the 2.10.x branch)

```shell
composer require doctrine/dbal ^3.0
vendor/bin/phpunit
```

When this issue is fixed, we should create a new one about making the tests pass.

derrabus:
The "crash" happens because the test suite contains mock versions of various DBAL classes that simply don't fit anymore. I have a half-baked solution that allowed me to run the tests to a certain degree. But we should rethink if mocking internal DBAL class structures is really the way to go. It's going to be hard to create a set of mocks that works with both DBAL 2 and 3.

greg0ire:
Getting rid of mocks of classes from an external package sounds good to me 👍
If it's not possible, then maybe the solution would be to have the mocks in question be provided by `doctrine/dbal` itself, I don't know if it makes sense yet.

derrabus:
I can look into reducing the mocks on the weekend, maybe. If nobody else wants to pick this ticket before that, of course. 😁

greg0ire:
To anyone reading this: consider starting with #8884 / #8885 first, they might be easier :slightly_smiling_face:

PR #8898, opened 2021-08-08
Throw exception NotSupported Exception for `UuidGenerator` with doctr…

scyzoryck:
Let's fix another issue:
```
[0;31mERROR[0m: UndefinedMethod - lib/Doctrine/ORM/Id/UuidGenerator.php:33:59 - Method Doctrine\DBAL\Platforms\AbstractPlatform::getGuidExpression does not exist (see https://psalm.dev/022)
        $sql  = 'SELECT ' . $conn->getDatabasePlatform()->[97;41mgetGuidExpression[0m();
```
Method has been removed with https://github.com/doctrine/dbal/pull/3211. 
Lets throw exception if method do not exists. Generator itself is already deprecated. 

Questions:
1. May it would be better to throw it in constructor?
2. What about unit tests? Should we cover also this case? 

It looks like there will be few similar issues, so let's start from this one to create a template for the next PRs :)

### DIFF
--- a/lib/Doctrine/ORM/Exception/NotSupported.php
+++ b/lib/Doctrine/ORM/Exception/NotSupported.php
@@ -10,4 +10,9 @@ final class NotSupported extends ORMException
     {
         return new self('This behaviour is (currently) not supported by Doctrine 2');
     }
+
+    public static function createForDbal3(): self
+    {
+        return new self('Feature was deprecated in doctrine/dbal 2.x and is not supported by installed doctrine/dbal:3.x, please see the doctrine/deprecations logs for new alternative approaches.');
+    }
 }
--- a/lib/Doctrine/ORM/Id/UuidGenerator.php
+++ b/lib/Doctrine/ORM/Id/UuidGenerator.php
@@ -4,8 +4,12 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Id;
 
+use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\Exception\NotSupported;
+
+use function method_exists;
 
 /**
  * Represents an ID generator that uses the database UUID expression
@@ -22,10 +26,16 @@ class UuidGenerator extends AbstractIdGenerator
             '%s is deprecated with no replacement, use an application-side generator instead',
             self::class
         );
+
+        if (! method_exists(AbstractPlatform::class, 'getGuidExpression')) {
+            throw NotSupported::createForDbal3();
+        }
     }
 
     /**
      * {@inheritDoc}
+     *
+     * @throws NotSupported
      */
     public function generate(EntityManager $em, $entity)
     {
--- a/phpstan-dbal3.neon
+++ b/phpstan-dbal3.neon
@@ -1,0 +1,8 @@
+includes:
+    - phpstan-baseline.neon
+    - phpstan-params.neon
+
+parameters:
+    ignoreErrors:
+        # deprecations from doctrine/dbal:3.x
+        - '/^Call to an undefined method Doctrine\\DBAL\\Platforms\\AbstractPlatform::getGuidExpression\(\).$/'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,21 +1,8 @@
 includes:
     - phpstan-baseline.neon
+    - phpstan-params.neon
 
 parameters:
-    level: 5
-    paths:
-        - lib
-        - tests/Doctrine/StaticAnalysis
-    excludePaths:
-        - lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php
-    earlyTerminatingMethodCalls:
-        Doctrine\ORM\Query\Parser:
-            - syntaxError
-    phpVersion: 70100
-
     ignoreErrors:
-        # The class was added in PHP 8.1
-        - '/^Attribute class ReturnTypeWillChange does not exist.$/'
-
         # https://github.com/doctrine/collections/pull/282
         - '/Variable \$offset in isset\(\) always exists and is not nullable\./'
--- a/psalm.xml
+++ b/psalm.xml
@@ -47,5 +47,11 @@
                 <referencedClass name="Doctrine\Common\Cache\XcacheCache"/>
             </errorLevel>
         </UndefinedClass>
+        <UndefinedMethod>
+            <errorLevel type="suppress">
+                <!-- See https://github.com/doctrine/orm/issues/8884 -->
+                <referencedMethod name="Doctrine\DBAL\Platforms\AbstractPlatform::getGuidExpression"/>
+            </errorLevel>
+        </UndefinedMethod>
     </issueHandlers>
 </psalm>
--- a/tests/Doctrine/Tests/ORM/Functional/UUIDGeneratorTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/UUIDGeneratorTest.php
@@ -4,13 +4,17 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\ORM\Functional;
 
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\Deprecations\PHPUnit\VerifyDeprecations;
+use Doctrine\ORM\Exception\NotSupported;
 use Doctrine\ORM\Mapping\Column;
 use Doctrine\ORM\Mapping\Entity;
 use Doctrine\ORM\Mapping\GeneratedValue;
 use Doctrine\ORM\Mapping\Id;
 use Doctrine\Tests\OrmFunctionalTestCase;
 
+use function method_exists;
 use function strlen;
 
 /**
@@ -22,6 +26,10 @@ class UUIDGeneratorTest extends OrmFunctionalTestCase
 
     public function testItIsDeprecated(): void
     {
+        if (! method_exists(AbstractPlatform::class, 'getGuidExpression')) {
+            self::markTestSkipped('Test valid for doctrine/dbal:2.x only.');
+        }
+
         $this->expectDeprecationWithIdentifier('https://github.com/doctrine/orm/issues/7312');
         $this->_em->getClassMetadata(UUIDEntity::class);
     }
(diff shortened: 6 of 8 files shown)


lib/Doctrine/ORM/Id/UuidGenerator.php
```diff
@@ -26,11 +29,17 @@ public function __construct()
...
     /**
      * {@inheritDoc}
+     *
+     * @throws NotSupported
      */
     public function generate(EntityManager $em, $entity)
     {
         $conn = $em->getConnection();
-        $sql  = 'SELECT ' . $conn->getDatabasePlatform()->getGuidExpression();
+        if (method_exists($conn, 'getGuidExpression')) {
```
greg0ire: This line looks wrong to me, for several reasons. If it is indeed wrong, I think that answers your question about unit tests :stuck_out_tongue:

greg0ire:
> May it would be better to throw it in constructor?

Throwing as early as possible is indeed a good idea 👍

scyzoryck:
Thanks @greg0ire! It was completly wrong. I've added Unit Tests to cover its logic. 
In PHPUnit I've defined 2 groups - dbal3 and dbal2 to differentiate tests that could be run with `doctrine/dbal` 2 and 3. I've adjusted also CI configuration.

psalm.xml
```diff
@@ -47,5 +47,11 @@
                 <referencedClass name="Doctrine\Common\Cache\XcacheCache"/>
             </errorLevel>
         </UndefinedClass>
+        <UndefinedMethod>
+            <errorLevel type="suppress">
+                <!-- See https://github.com/doctrine/orm/issues/8884 -->
+                <referencedMethod name="Doctrine\DBAL\Platforms\AbstractPlatform::getGuidExpression"/>
+            </errorLevel>
+        </UndefinedMethod>
```
greg0ire, line 55: Can you do the same for PHPStan? I can see an automatic comment above from that tool

lib/Doctrine/ORM/Exception/NotSupported.php
```diff
@@ -10,4 +10,9 @@ public static function create(): self
     {
         return new self('This behaviour is (currently) not supported by Doctrine 2');
     }
+
+    public static function createForDbal3(): self
+    {
+        return new self('This behaviour is not supported by doctrine/dbal:3.x, please use doctrine/dbal:2.x instead.');
```
beberlei: I think we should improve this exception message, because it gives no indiication on how to get to DBAL 3 making it sound DBAL 2 is better. Maybe:

```suggestion
        return new self('Feature was deprecated in doctrine/dbal 2.x and is not supported by installed doctrine/dbal:3.x, please see the doctrine/deprecations logs for new alternative approaches.');
```

beberlei (dismissed):
Small adjustment to the exception message, otherwise good to go for me.

psalm.xml
scyzoryck: @greg0ire Added.

lib/Doctrine/ORM/Exception/NotSupported.php
scyzoryck: Thanks @beberlei! Sounds much better :)

phpstan.neon
```diff
@@ -16,6 +16,8 @@ parameters:
     ignoreErrors:
         # The class was added in PHP 8.1
         - '/^Attribute class ReturnTypeWillChange does not exist.$/'
-
         # https://github.com/doctrine/collections/pull/282
         - '/Variable \$offset in isset\(\) always exists and is not nullable\./'
+        # method removed in doctrine/dbal3.x
+        - '/^Call to an undefined method Doctrine\\DBAL\\Platforms\\AbstractPlatform::getGuidExpression\(\).$/'
```
greg0ire: This is only useful when using DBAL 3, and thus makes the DBAL 2 phpstan job fail :(
Maybe we should have a separate file for DBAL 3? Like `phpstan-dbal3.neon`?

phpstan-dbal3.neon
```diff
@@ -0,0 +1,14 @@
+includes:
+    - phpstan.neon
+
+parameters:
+    ignoreErrors:
+        # The class was added in PHP 8.1
+        - '/^Attribute class ReturnTypeWillChange does not exist.$/'
+
+        # https://github.com/doctrine/collections/pull/282
+        - '/Variable \$offset in isset\(\) always exists and is not nullable\./'
```
greg0ire: Shouldn't the 2 errors above be only in `phpstan.neon`? They only appear with DBAL 3? Or is `ignoreErrors` being overwritten here?

scyzoryck: It looks like it might require a bit more work - as the ignored paths are merged from all files. May it would make sense to have 2 separate baselines and configs for dbal2 and dbal3?

greg0ire: Maybe yes, but I'd rather do that when absolutely necessary.

greg0ire (dismissed):

phpstan-dbal3.neon
scyzoryck: @greg0ire - splitted a little bit more. Common params moved to separate file.  It looks like the best what we can do to split them in a logic way.

greg0ire (dismissed):

.github/workflows/continuous-integration.yml
```diff
@@ -40,14 +40,14 @@ jobs:
 
       - name: "Run PHPUnit"
         continue-on-error: true
-        run: "vendor/bin/phpunit -c ci/github/phpunit/sqlite.xml --coverage-clover=coverage-no-cache.xml"
+        run: "vendor/bin/phpunit -c ci/github/phpunit/sqlite.xml --coverage-clover=coverage-no-cache.xml --exclude-group performance,locking_functional,dbal2"
```
SenseException: @greg0ire Are `performance` and `locking_functional` not needed for the regular "Run PHPUnit" step?

greg0ire: `performance` is run separately, that's for sure. As for `locking_functional`, it's about Gearman, so it might be great to run it someday. They are already excluded in `sqlite.xml`, but I think @scyzroyck had to respecify them in order to add `dbal2` to the list.

SenseException (dismissed):

.github/workflows/continuous-integration.yml
SenseException: We'll see how this will affects the code coverage in case some test won't be executed anymore.

lib/Doctrine/ORM/Id/UuidGenerator.php
```diff
@@ -22,10 +26,16 @@ public function __construct()
...
+        }
     }
 
     /**
      * {@inheritDoc}
+     *
+     * @throws NotSupported
      */
     public function generate(EntityManager $em, $entity)
     {
```
greg0ire, line 41: Apparently this method is no longer covered: https://app.codecov.io/gh/doctrine/orm/compare/8898/changes
Why is that?

.github/workflows/continuous-integration.yml
greg0ire: [looks like it does](https://github.com/doctrine/orm/pull/8898#discussion_r693210346)

lib/Doctrine/ORM/Id/UuidGenerator.php
scyzoryck: If the coverage is taken from tests with doctrine/dbal:3.0, this function will not be tested as the class itself will be never initialised.

scyzoryck: @greg0ire - it looks like the possible reason might be the same name `--coverage-clover=coverage-cache.xml` used for the tests with dbal2 and dbal3.

greg0ire: It's close to that, but that's not it, I think you can have the same name as long as it differs when it's uploaded (so the upload step needs to be modified)

greg0ire: Or maybe I'm mistaken, looks like I already took care of setting different names when introducing the DBAL 3 build

greg0ire: > If the coverage is taken from tests with doctrine/dbal:3.0, this function will not be tested as the class itself will be never initialised.

The coverage is supposed to be merged from multiple reports

greg0ire: The list can be seen here: https://github.com/doctrine/orm/pull/8898/checks?check_run_id=3343305825#step:3:21

I think some might be missing indeed.

greg0ire: Ah but the DBAL3 tests don't even run right now, so there will not be a report from them. The DBAL2 tests are running though, and are supposed to cover that path.

greg0ire: The job supposed to cover that path are the mysql ones, and they are present… not sure what to think :thinking:

tests/Doctrine/Tests/ORM/Functional/UUIDGeneratorTest.php
```diff
@@ -20,18 +23,28 @@ class UUIDGeneratorTest extends OrmFunctionalTestCase
 {
     use VerifyDeprecations;
 
+    /**
+     * @group dbal2
```
greg0ire: Why do we need these groups anyway? Can't the same be achieved with `markTestSkipped`?

greg0ire:
The mysql jobs are passing despite me pushing a very wrong commit, so it's not a coverage upload issue, it's an execution issue. Marking as draft until this is fixed by you or me :)

tests/Doctrine/Tests/ORM/Functional/UUIDGeneratorTest.php
```diff
@@ -20,18 +23,28 @@ class UUIDGeneratorTest extends OrmFunctionalTestCase
...
+     */
     public function testGenerateUUID(): void
     {
         if ($this->_em->getConnection()->getDatabasePlatform()->getName() !== 'mysql') {
             self::markTestSkipped('Currently restricted to MySQL platform.');
         }
 
+        if (! method_exists(Connection::class, 'getGuidExpression')) {
```
greg0ire: I don't think `Connection` is the right class for this, I think you meant `AbstractPlatform`. That's probably why the code was not executed.

greg0ire: Yup, that was it!

greg0ire:
Note: there is still a drop in coverage, because the DBAL 3 tests are not actually running yet (see #8886)

lib/Doctrine/ORM/Id/UuidGenerator.php
SenseException: Independent from this PR: Shouldn't this also be covered by the other platforms?

SenseException (approved):

lib/Doctrine/ORM/Id/UuidGenerator.php
greg0ire: :thinking: good question, might be worth removing that call to `self::markTestSkipped` and see on which other platforms it passes. The situation might have evolved since this test was written in 2012.

greg0ire:
Thanks @scyzoryck !

PR #8903, opened 2021-08-09
SchemaValidator: Fix mapped superclass missing in discriminator map

### LINKED ISSUE #8771: Doctrine ORM 2.9: MappedSuperclass in entity hierarchy
We have an entity hierarchy similar to:

```php
/**
 * @ORM\Entity
 * @ORM\Table(name="AbstractInvoiceItem")
 *
 * @ORM\DiscriminatorColumn(name="type")
 * @ORM\DiscriminatorMap({
 *     "SetupFee" = SetupFee::class
 * })
 */
abstract class AbstractInvoiceItem
{
}

/**
 * @MappedSuperclass
 */
abstract class AbstractProductInvoiceItem extends AbstractInvoiceItem
{
    /**
     * @ORM\ManyToOne(targetEntity=AbstractProduct::class)
     */
    protected ?AbstractProduct $product;
}

/**
 * @ORM\Entity
 * @ORM\Table(name=SetupFee")
 */
class SetupFee extends AbstractProductInvoiceItem
{
}
```

This was working fine with Doctrine 2.8.

Since we upgraded to Doctrine 2.9 however, `bin/console doctrine:schema:validate` started failing with:

> [FAIL] The entity-class App\(...)\AbstractProductInvoiceItem mapping is invalid:
> * Entity class 'App\(...)\AbstractProductInvoiceItem' is part of inheritance hierarchy, but is not mapped in the root entity 'App\(...)\AbstractInvoiceItem' discriminator map. All subclasses must be listed in the discriminator map.

I tried adding a "fake" entry to the discriminator map for the abstract class:

```php
/*
 * @ORM\DiscriminatorMap({
 *     "AbstractProductInvoiceItem" = AbstractProductInvoiceItem::class,
 *     "SetupFee" = SetupFee::class
 * })
 */
```

But now Doctrine attempts to `JOIN` a non-existing table when loading an entity in the hierarchy:

> SQLSTATE[42S02]: Base table or view not found: 1146 Table 'abstract_product_invoice_item' doesn't exist

**How can I fix this?**

A solution would be to make the intermediate abstract class an entity, but this would mean that `$product` would be moved from leaf entities to a new `AbstractProductInvoiceItem` table, which adds an extra `JOIN` to every query, and would mean a rather scary migration to do.

Is there a better solution? Or **are MappedSuperclasses not allowed in the middle of an entity hierarchy anymore?**

olsavmic:
Fixes https://github.com/doctrine/orm/issues/8771

I'm not sure if it's actually valid case for a MappedSuperclass to be a part of the DiscriminatorMap. 

It looks like the discussion continues in https://github.com/doctrine/orm/pull/8415, https://github.com/doctrine/orm/pull/7825.

This is a fix for the current behaviour so that it does not block people from upgrading to ORM 2.9

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -255,7 +255,7 @@ class SchemaValidator
             }
         }
 
-        if (! $class->isInheritanceTypeNone() && ! $class->isRootEntity() && array_search($class->name, $class->discriminatorMap) === false) {
+        if (! $class->isInheritanceTypeNone() && ! $class->isRootEntity() && ! $class->isMappedSuperclass && array_search($class->name, $class->discriminatorMap) === false) {
             $ce[] = "Entity class '" . $class->name . "' is part of inheritance hierarchy, but is " .
                 "not mapped in the root entity '" . $class->rootEntityName . "' discriminator map. " .
                 'All subclasses must be listed in the discriminator map.';
--- a/tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
+++ b/tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
@@ -199,6 +199,46 @@ class SchemaValidatorTest extends OrmTestCase
             $ce
         );
     }
+
+    /**
+     * @group 8771
+     */
+    public function testMappedSuperclassNotPresentInDiscriminator(): void
+    {
+        $class1 = $this->em->getClassMetadata(MappedSuperclass::class);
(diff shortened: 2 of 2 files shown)


olsavmic:
@greg0ire @beberlei @derrabus Could I get a review please?

It's a tiny change resolving an issue that's been open for a month already.

vojtabiberle:
Great fix @olsavmic 

This will help me too. Can someone please review it?

SenseException (approved):
It technically works, but I don't know if this is supposed to work that way. Additional reviewers are needed.

greg0ire:
Thanks @olsavmic !

PR #8905, opened 2021-08-10
Add explicit `@return` type next to `#[ReturnTypeWillChange]`

nicolas-grekas:
Same as doctrine/collections#280

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/CollectionRegionCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/CollectionRegionCommand.php
@@ -82,6 +82,8 @@ EOT
 
     /**
      * {@inheritdoc}
+     *
+     * @return int
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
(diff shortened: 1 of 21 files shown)


lib/Doctrine/ORM/Tools/Console/Command/ClearCache/CollectionRegionCommand.php
```diff
@@ -82,6 +82,8 @@ protected function configure()
 
     /**
      * {@inheritdoc}
+     *
+     * @return int
```
greg0ire, line 86: Why is this needed for Symfony commands? This return type is not inherited from a native type like `Countable` or `ArrayIterator`, I thought this was only about native types. Isn't it?

nicolas-grekas: When used with `DebugClassLoader`, this saves a deprecation: `@return` means the same as `#[ReturnTypeWillChange]` to it.

greg0ire: Ok, I thought `DebugClassLoader` wouldn't need it in the first place, since no stubs are involved, only normal PHP files. Apparently that's not the case.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @nicolas-grekas !

PR #8907, opened 2021-08-11
Update getAssociationTargetClass phpdoc

VincentLanglet:
Now the interface can return null, https://github.com/doctrine/persistence/pull/192, the phpdoc here should be precised because it never return null.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -3409,6 +3409,8 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * {@inheritDoc}
      *
+     * @param string $assocName
+     *
      * @return string
      * @psalm-return class-string
      *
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -3409,6 +3409,9 @@ public function getAssociationNames()
     /**
      * {@inheritDoc}
      *
+     * @return string
+     * @psalm-return class-string
+     *
      * @throws InvalidArgumentException
      */
     public function getAssociationTargetClass($assocName)
```
greg0ire, line 3419: Not sure why Psalm suddenly requires `$assocName` to be documented :thinking:

VincentLanglet: psalm has a "All or nothing" strategy. So indeed, since I add the return tag, I have to add the param too... edited.

greg0ire (approved):

greg0ire:
Thanks @VincentLanglet ! I squashed and used a commit message explaining why you are doing this: https://github.com/doctrine/orm/commit/b19a13f4edfaa5806109cd899f5912a7df1547b5

VincentLanglet:
Thanks ! Don't know if this is worth enough for a release, in order to avoid consumer with phpstan/psalm to not having too deal with null values (since the persistence repository was released recently)

greg0ire:
The last release is quite old, so while by itself it's probably not worth it, I think we might want to release anyway.

PR #8916, opened 2021-08-13
Check current class' discriminator map

### LINKED ISSUE #8914: DiscriminatorMap and MappedSuperclass giving problems in 2.9.4
### Bug Report

Introduced in #8402

|    Q        |   A
|------------ | ------
| BC Break    | **yes**/no
| Version     | 2.9.4

Giving this code:

```
#[MappedSuperclass]
abstract class BaseEntity {}

#[InheritanceType('SINGLE_TABLE')]
#[DiscriminatorColumn(name: 'discr', type: 'string')]
#[DiscriminatorMap(['organisation' => Organisation::class, 'customer' => Customer::class, 'supplier' => Supplier::class])]
class Organisation extends BaseEntity {}

#[Entity]
class Customer extends Organisation {}

#[Entity]
class Supplier extends Organisation {}
```
I get this error:

> Entity 'App\Entity\Organisation' has to be part of the discriminator map of 'App\Entity\Organisation' to be properly mapped in the inheritance hierarchy. Alternatively you can make 'App\Entity\Organisation' an abstract class to avoid this exception from occurring

greg0ire:
A class can be in its own discriminator map, as described in documentation example at
https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/inheritance-mapping.html#single-table-

Fixes #8914

# How can I test this?

```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/doctrine-orm
composer require doctrine/orm "dev-failing-test-8914 as 2.9.4"
```

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -285,7 +285,10 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
                 }
             } else {
                 assert($parent instanceof ClassMetadataInfo); // https://github.com/doctrine/orm/issues/8746
-                if ((! $class->reflClass || ! $class->reflClass->isAbstract()) && ! in_array($class->name, $parent->discriminatorMap)) {
+                if (
+                    (! $class->reflClass || ! $class->reflClass->isAbstract())
+                    && ! in_array($class->name, $class->discriminatorMap)
+                ) {
                     throw MappingException::mappedClassNotPartOfDiscriminatorMap($class->name, $class->rootEntityName);
                 }
             }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8914Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8914Test.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\OrmTestCase;
+
+use function assert;
+use function count;
+
+final class GH8914Test extends OrmTestCase
+{
+    /**
+     * @group GH-8914
+     * @doesNotPerformAssertions
+     */
+    public function testDiscriminatorMapWithSeveralLevelsIsSupported(): void
+    {
+        $entityManager = $this->getTestEntityManager();
+        $entityManager->getClassMetadata(GH8914Person::class);
+    }
+}
+
+/**
+ * @MappedSuperclass
+ */
+abstract class GH8914BaseEntity
+{
+}
+
+/**
+ * @Entity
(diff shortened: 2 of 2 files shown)


greg0ire:
@stephanvierkant @akinspe @ghdi @jexa please confirm this fixes your issue by applying the steps in the PR description.

stephanvierkant:
Yes, I can confirm that it fixes the issue!

greg0ire:
@mpdude please review

mpdude:
Will do ASAP, probably begin of next week

ostrolucky (dismissed):

akinspe:
I can also confirm it fixes our issue. Thank you!

SenseException (dismissed):

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -285,7 +285,11 @@ protected function validateRuntimeMetadata($class, $parent)
                 }
             } else {
                 assert($parent instanceof ClassMetadataInfo); // https://github.com/doctrine/orm/issues/8746
-                if ((! $class->reflClass || ! $class->reflClass->isAbstract()) && ! in_array($class->name, $parent->discriminatorMap)) {
```
beberlei, line 288: Do we need to check both? When `validateRuntimeMetadata` is called after the copy of discriminator map from parent to child, then the change becomes easier:

https://github.com/doctrine/orm/blob/2.9.x/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L129

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -285,7 +285,11 @@ protected function validateRuntimeMetadata($class, $parent)
...
+                if (
+                    (! $class->reflClass || ! $class->reflClass->isAbstract())
+                    && ! in_array($class->name, $parent->discriminatorMap)
```
beberlei: Do we need to check both? When validateRuntimeMetadata is called after the copy of discriminator map from parent to child, then the change becomes easier:

https://github.com/doctrine/orm/blob/b19a13f4edfaa5806109cd899f5912a7df1547b5/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L129

greg0ire: Indeed, looks like things still work when removing the parent class check :slightly_smiling_face:

SenseException (approved):

greg0ire:
ping @mpdude :slightly_smiling_face:

beberlei (approved):

beberlei:
@greg0ire lets merge and ship a bugfix release imho

mpdude:
Only 1.5 years later I stumble across this issue.

It's a bit late, but :+1: it's what I would suggest with today's knowledge as well.

PR #8919, opened 2021-08-15
feat(ScalarColumnHydrator): added ScalarColumnHydrator to get flat array results from query for single column

bhushan:
Description: 

- feat(AbstractQuery): added getFlatArrayResultByKey method to Abstract Query to get flat array results from query

### How it helps ?

- Imagine if We just want list of ids from database then we need to implement something like this in repository

```php
$transform = function($item) {
    return $item['id'];
};

$result = array_map($transform, $em->createQuery("SELECT a.id FROM Auction a")->getArrayResult());
```

With `getFlatArrayResultByKey`

```php
$em->createQuery("SELECT a.id FROM Auction a")->getFlatArrayResultByKey('id');
```

and if we don't pass any key then it works same as `getArrayResult`

### DIFF
--- a/docs/en/reference/dql-doctrine-query-language.rst
+++ b/docs/en/reference/dql-doctrine-query-language.rst
@@ -1184,6 +1184,7 @@ The constants for the different hydration modes are:
 -  ``Query::HYDRATE_ARRAY``
 -  ``Query::HYDRATE_SCALAR``
 -  ``Query::HYDRATE_SINGLE_SCALAR``
+-  ``Query::HYDRATE_SCALAR_COLUMN``
 
 Object Hydration
 ^^^^^^^^^^^^^^^^
@@ -1271,6 +1272,25 @@ You can use the ``getSingleScalarResult()`` shortcut as well:
 
     <?php
     $numArticles = $query->getSingleScalarResult();
+
+Scalar Column Hydration
+^^^^^^^^^^^^^^^^^^^^^^^
+
+If you have a query which returns a one-dimensional array of scalar values
+you can use scalar column hydration:
+
+.. code-block:: php
+
+    <?php
+    $query = $em->createQuery('SELECT a.id FROM CmsUser u');
+    $ids = $query->getResult(Query::HYDRATE_SCALAR_COLUMN);
+
+You can use the ``getSingleColumnResult()`` shortcut as well:
+
+.. code-block:: php
+
+    <?php
+    $ids = $query->getSingleColumnResult();
 
 Custom Hydration Modes
 ^^^^^^^^^^^^^^^^^^^^^^
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -71,6 +71,11 @@ abstract class AbstractQuery
     public const HYDRATE_SIMPLEOBJECT = 5;
 
     /**
+     * Hydrates scalar column value.
+     */
+    public const HYDRATE_SCALAR_COLUMN = 6;
+
+    /**
      * The parameter map of this query.
      *
      * @var ArrayCollection|Parameter[]
@@ -782,6 +787,18 @@ abstract class AbstractQuery
     public function getArrayResult()
     {
         return $this->execute(null, self::HYDRATE_ARRAY);
+    }
+
+    /**
+     * Gets one-dimensional array of results for the query.
+     *
+     * Alias for execute(null, HYDRATE_SCALAR_COLUMN).
+     *
+     * @return mixed[]
+     */
+    public function getSingleColumnResult()
+    {
+        return $this->execute(null, self::HYDRATE_SCALAR_COLUMN);
     }
 
     /**
--- a/lib/Doctrine/ORM/Exception/MultipleSelectorsFoundException.php
+++ b/lib/Doctrine/ORM/Exception/MultipleSelectorsFoundException.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Exception;
+
+use function implode;
+use function sprintf;
+
+final class MultipleSelectorsFoundException extends ORMException
+{
+    public const MULTIPLE_SELECTORS_FOUND_EXCEPTION = 'Multiple selectors found: %s. Please select only one.';
+
+    /**
+     * @param string[] $selectors
+     */
+    public static function create(array $selectors): self
+    {
+        return new self(
+            sprintf(
+                self::MULTIPLE_SELECTORS_FOUND_EXCEPTION,
+                implode(', ', $selectors)
+            )
+        );
+    }
+}
--- a/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Internal\Hydration;
+
+use Doctrine\DBAL\Driver\Exception;
+use Doctrine\ORM\Exception\MultipleSelectorsFoundException;
+
+use function count;
+
+/**
+ * Hydrator that produces one-dimensional array.
+ */
+final class ScalarColumnHydrator extends AbstractHydrator
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @throws MultipleSelectorsFoundException
+     * @throws Exception
+     */
+    protected function hydrateAllData(): array
+    {
+        if (count($this->resultSetMapping()->fieldMappings) > 1) {
+            throw MultipleSelectorsFoundException::create($this->resultSetMapping()->fieldMappings);
+        }
+
+        $result = [];
+
+        while ($row = $this->statement()->fetchOne()) {
+            $result[] = $row;
+        }
+
+        return $result;
+    }
+}
--- a/tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
@@ -21,10 +21,8 @@ use Doctrine\Tests\Models\CMS\CmsUser;
 use Doctrine\Tests\OrmFunctionalTestCase;
 use InvalidArgumentException;
 
-use function count;
 use function get_class;
-use function is_array;
-use function is_numeric;
+use function in_array;
 
 class BasicFunctionalTest extends OrmFunctionalTestCase
 {
@@ -256,6 +254,35 @@ class BasicFunctionalTest extends OrmFunctionalTestCase
         self::assertEquals('Guilherme', $usersScalar[0]['u_name']);
         self::assertEquals('gblanco', $usersScalar[0]['u_username']);
         self::assertEquals('developer', $usersScalar[0]['u_status']);
+    }
+
+    public function testSingleColumnQuery(): void
+    {
+        $gregoire           = new CmsUser();
+        $gregoire->name     = 'Gregoire';
+        $gregoire->username = 'greg0ire';
+        $gregoire->status   = 'developer';
+        $this->_em->persist($gregoire);
+
+        $bhushan           = new CmsUser();
+        $bhushan->name     = 'Bhushan';
+        $bhushan->username = 'bhushan';
+        $bhushan->status   = 'developer';
+        $this->_em->persist($bhushan);
+
+        $this->_em->flush();
+
+        $query = $this->_em->createQuery('select u.username from Doctrine\Tests\Models\CMS\CmsUser u order by u.username DESC');
+
+        $users = $query->getSingleColumnResult();
+
+        $expected = [
+            'greg0ire',
+            'bhushan',
+        ];
+
+        self::assertCount(2, $users);
+        self::assertSame($expected, $users);
     }
 
     public function testBasicOneToManyInnerJoin(): void
--- a/tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
+++ b/tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
@@ -1,0 +1,122 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Hydration;
+
+use Doctrine\ORM\Exception\MultipleSelectorsFoundException;
+use Doctrine\ORM\Internal\Hydration\ScalarColumnHydrator;
+use Doctrine\ORM\Query\ResultSetMapping;
+use Doctrine\Tests\Mocks\ArrayResultFactory;
+use Doctrine\Tests\Models\CMS\CmsUser;
+
+use function sprintf;
+
+class ScalarColumnHydratorTest extends HydrationTestCase
+{
+    /**
+     * Select u.id from CmsUser u
+     */
+    public function testEmptyResultTest(): void
+    {
+        $rsm = new ResultSetMapping();
+
+        $rsm->addEntityResult(CmsUser::class, 'u');
+        $rsm->addFieldResult('u', 'u__id', 'id');
+
+        $stmt     = ArrayResultFactory::createFromArray($emptyResultSet = []);
+        $hydrator = new ScalarColumnHydrator($this->entityManager);
+        $result   = $hydrator->hydrateAll($stmt, $rsm);
+
+        $this->assertIsArray($result);
+        $this->assertEmpty($result);
+    }
+
+    /**
+     * Select u.id from CmsUser u
+     */
+    public function testSingleColumnEntityQueryWithoutScalarMap(): void
+    {
+        $rsm = new ResultSetMapping();
+
+        $rsm->addEntityResult(CmsUser::class, 'u');
+        $rsm->addFieldResult('u', 'u__id', 'id');
+
+        $resultSet = [
+            ['u__id' => '1'],
+            ['u__id' => '2'],
+        ];
+
+        $stmt     = ArrayResultFactory::createFromArray($resultSet);
+        $hydrator = new ScalarColumnHydrator($this->entityManager);
+        $result   = $hydrator->hydrateAll($stmt, $rsm);
+
+        $this->assertIsArray($result);
+        $this->assertCount(2, $result);
+
+        $this->assertEquals(1, $result[0]);
+        $this->assertEquals(2, $result[1]);
+    }
+
+    /**
+     * Select u.id from CmsUser u
+     */
+    public function testSingleColumnEntityQueryWithScalarMap(): void
+    {
+        $rsm = new ResultSetMapping();
+
+        $rsm->addEntityResult(CmsUser::class, 'u');
+        $rsm->addFieldResult('u', 'u__id', 'id');
+        $rsm->addScalarResult('sclr0', 'id');
+        $rsm->addIndexByScalar('sclr0');
+
+        $resultSet = [
+            ['u__id' => '1'],
+            ['u__id' => '2'],
+        ];
+
+        $stmt     = ArrayResultFactory::createFromArray($resultSet);
+        $hydrator = new ScalarColumnHydrator($this->entityManager);
+        $result   = $hydrator->hydrateAll($stmt, $rsm);
+
+        $this->assertIsArray($result);
+        $this->assertCount(2, $result);
+
+        $this->assertEquals(1, $result[0]);
+        $this->assertEquals(2, $result[1]);
+    }
+
+    /**
+     * Select u.id, u.name from CmsUser u
+     */
+    public function testMultipleColumnEntityQueryThrowsException(): void
+    {
+        $rsm = new ResultSetMapping();
+
+        $rsm->addEntityResult(CmsUser::class, 'u');
+        $rsm->addFieldResult('u', 'u__id', 'id');
+        $rsm->addFieldResult('u', 'u__name', 'name');
+
+        $resultSet = [
+            [
+                'u__id'   => '1',
+                'u__name' => 'Gregoire',
+            ],
+            [
+                'u__id'   => '2',
+                'u__name' => 'Bhushan',
+            ],
+        ];
+
+        $stmt     = ArrayResultFactory::createFromArray($resultSet);
+        $hydrator = new ScalarColumnHydrator($this->entityManager);
+
+        $this->expectException(MultipleSelectorsFoundException::class);
+        $this->expectExceptionMessage(sprintf(
+            MultipleSelectorsFoundException::MULTIPLE_SELECTORS_FOUND_EXCEPTION,
+            'id, name'
+        ));
+
+        $hydrator->hydrateAll($stmt, $rsm);
+    }
+}
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -865,6 +865,9 @@ use function sprintf;
             case Query::HYDRATE_SIMPLEOBJECT:
                 return new Internal\Hydration\SimpleObjectHydrator($this);
 
+            case Query::HYDRATE_SCALAR_COLUMN:
+                return new Internal\Hydration\ScalarColumnHydrator($this);
+
             default:
                 $class = $this->config->getCustomHydrationMode($hydrationMode);
 
--- a/lib/Doctrine/ORM/Exception/NamedNativeQueryNotFound.php
+++ b/lib/Doctrine/ORM/Exception/NamedNativeQueryNotFound.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Exception;
 
-use LogicException;
-
 use function sprintf;
 
 final class NamedNativeQueryNotFound extends ORMException implements ConfigurationException


greg0ire:
Sounds like what you want is `array_column($em->createQuery("SELECT a.id FROM Auction a")->getArrayResult(), 'id');`

bhushan:
@greg0ire Yeah, There are multiple ways of doing it. But Idea is, If we have this built in in Doctrine then it will be helpful IMO.

`$key` could be replaced with `$column` and We could use `array_column` instead of `array_map` in PR.

Basically from this wrapper, we could eliminate many duplicate code in repositories where we have giant project and plenty of entities and most important developer convenience.

It will be addition to `getResults` and `getArrayResults` as helper

greg0ire:
You know what would be even more helpful? Not having to specify a key. `$em->createQuery("SELECT a.id FROM Auction a")->getSingleColumn();`, much like what exists in the DBAL: https://github.com/doctrine/dbal/blob/e974d4deca8c430d9eba0975f441538145a5a8f1/src/Connection.php#L534-L553

That could be done with a new hydrator that would look a lot like https://github.com/doctrine/orm/blob/d636d796868e90c43889e5510598686427e29dc7/lib/Doctrine/ORM/Internal/Hydration/ScalarHydrator.php#L30-L53
, except it would call `fetchOne()` instead of `fetch`.

bhushan:
Hey, Thats a nice idea 👌🏻 but there is one point, if we dont add key then we are forcing developer to use single column selector, though its a best practice to query only required columns but still your thoughts on it !! We could allow multiple columns or through specific exception if multiple columns are selected while using singleColumn method

greg0ire:
> Hey, Thats a nice idea 👌🏻 but there is one point, if we dont add key then we are forcing developer to use single column selector, though its a best practice to query only required columns but still your thoughts on it !!

Well yeah, it's a best practice indeed, so it's one more reason to do it. With your solution, you are also forcing the developer to do something (using an argument when calling the function).

>  We could allow multiple columns or through specific exception if multiple columns are selected while using singleColumn method

If you want multiple columns, wouldn't `getArrayResult` or `getScalarResult` do a better job?

bhushan:
Implementation can vary but i think we should have something from which we could get single column values as flattened array.. 

What do you suggest? I will do modifications in PR accordingly if you agree..

bhushan:
`getArrayResult` or `getScalarResult` gives

```php
[
0 => [ 'id' => 1 ],
1 => [ 'id' => 2 ],
2 => [ 'id' => 3 ],
3 => [ 'id' => 4 ],
]
```

where it will be nice to have if some method provides something like 

```php
[1,2,3,4]
```

this

greg0ire:
> What do you suggest?

[I suggest introducing a new hydrator](https://github.com/doctrine/orm/pull/8919#issuecomment-899106878) (`ScalarListHydrator`, `ScalarColumnHydrator`?), and a method to call it.

greg0ire:
I'm requesting review from more people before you spend too much time working on it.

ostrolucky:
Check out https://gist.github.com/ostrolucky/f9f1e0b271357573fde55b7a2ba91a32 for example how could the possible implementation look like. But yeah, ORM should have something like this inclulde.

SenseException:
Having tests to list and cover the use cases would help the feature development and coverage of the use cases for the hydrator. See `Doctrine\Tests\ORM\Hydration\SimpleObjectHydratorTest` as an example for the `SimpleObjectHydrator`.

bhushan:
> Having tests to list and cover the use cases would help the feature development and coverage of the use cases for the hydrator. See `Doctrine\Tests\ORM\Hydration\SimpleObjectHydratorTest` as an example for the `SimpleObjectHydrator`.

working on tests

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
```diff
@@ -0,0 +1,135 @@
...
+        $this->assertEquals(1, $result[0]);
+
+        $this->assertEquals(2, $result[1]);
+    }
+
+    /**
+     * SELECT PARTIAL u.{id, name}
+     *   FROM Doctrine\Tests\Models\CMS\CmsUser u
+     */
+    public function testMultipleColumnEntityQuery(): void
```
bhushan, line 105: > Need help with this test

Screenshot 2021-08-22 at 6 23 55 AM

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
```diff
@@ -0,0 +1,138 @@
...
+        $this->assertEquals(1, $result[0]);
+
+        $this->assertEquals(2, $result[1]);
+    }
+
+    /**
+     * SELECT PARTIAL u.{id}
+     *   FROM Doctrine\Tests\Models\CMS\CmsUser u
+     */
+    public function testSingleColumnEntityQueryWithScalarMap(): void
```
bhushan, line 64: > Not understanding with this test why It's giving nested results

Screenshot 2021-08-22 at 6 27 24 AM

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
```diff
@@ -0,0 +1,138 @@
...
+        $this->assertEquals(1, $result[0]);
+
+        $this->assertEquals(2, $result[1]);
+    }
+
+    /**
+     * SELECT PARTIAL u.{id, name}
+     *   FROM Doctrine\Tests\Models\CMS\CmsUser u
+     */
+    public function testMultipleColumnEntityQueryWithoutScalarMap(): void
```
bhushan: > Need help with this test

Screenshot 2021-08-22 at 6 23 55 AM

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,57 @@
...
+        if (empty($result)) {
+            return [];
+        }
+
+        $indexColumn = $this->_rsm->scalarMappings[$this->_rsm->indexByMap['scalars']];
+        $keys = array_keys(reset($result));
+
+        $column = isset($keys[1]) && $keys[0] === $indexColumn ? $keys[1] : $keys[0];
+
+        return array_column($result, $column, $indexColumn);
```
bhushan: This portion is still untested, once we fix last two tests, adding coverage for these line will be easy.

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
```diff
@@ -0,0 +1,170 @@
...
+
+        $stmt = new HydratorMockStatement($resultSet);
+        $hydrator = new ScalarColumnHydrator($this->entityManager);
+        $result = $hydrator->hydrateAll($stmt, $rsm);
+
+        $this->assertEquals(2, count($result));
+        $this->assertTrue(is_array($result));
+
+        $this->assertEquals('romanb', $result[0]);
+        $this->assertEquals('jwage', $result[1]);
```
bhushan: This test should return flat array of `$key[1]` ie `name`

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -799,6 +804,18 @@ public function getArrayResult()
         return $this->execute(null, self::HYDRATE_ARRAY);
     }
 
+    /**
+     * Gets the flat array of results for the query.
```
greg0ire: The documentation for `getScalarResult` already mentions "flat". I'm myself not sure what is meant by that. Let's use something more precise? See 

https://github.com/doctrine/orm/blob/1de28c2cab5cd7e3c9d781a8beff22e328190911/docs/en/reference/dql-doctrine-query-language.rst#L1049

Please add documentation for the new hydration method .

```suggestion
     * Gets the one-dimensional array of results for the query.
```

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,57 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
```
greg0ire: We no longer include the license header, at least not on 2.10.x which you should target, since this is a new feature.

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,57 @@
...
+class ScalarColumnHydrator extends ArrayHydrator
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function hydrateAllData(): array
+    {
+        if (!isset($this->_rsm->indexByMap['scalars'])) {
+            return $this->_stmt->fetchAll(PDO::FETCH_COLUMN);
```
greg0ire:
```suggestion
            return $this->_stmt->fetchOne();
```

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,57 @@
...
+
+use function reset;
+use function array_keys;
+use function array_column;
+
+/**
+ * The ScalarColumnHydrator is extended ArrayHydrator which produces a flat array single column "graph"
+ * that is often (not always) interchangeable with the corresponding object graph for read-only access.
+ */
+class ScalarColumnHydrator extends ArrayHydrator
```
greg0ire: @ostrolucky I don't get why we extend from `ArrayHydrator` here… the implementation I suggested should be more simple, shouldn't it?

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
greg0ire: That's one more reason not to extend from `ArrayHydrator` IMO

lib/Doctrine/ORM/AbstractQuery.php
bhushan: Updated

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
bhushan: Extending `AbstractHydrator` also simplifying implementation using ScalarHydrator and `fetchOne`

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
bhushan: Changing implementation of Hydrator

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: Target changed to 2.10.x and licence header as well removed

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: changed deprecated method `fetchAll` to `fetchOne`

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
ostrolucky: Decoupling from ArrayHydrator would require additional work, hence why I didn't do it before. I can't tell how much more work would it require.

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Internal\Hydration;
+
+/**
+ * Hydrator that produces one-dimensional array "graph" that is often (not always) interchangeable
+ * with the corresponding object graph for read-only access.
```
greg0ire: That comment is misleading, there is really no graph here, because you cannot select more than 1 scalar. In order to have something that would look like a graph, you would need to have at least 2 entities selected.

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
```diff
@@ -0,0 +1,166 @@
...
+
+        $this->assertEquals(1, $result[0]);
+        $this->assertEquals(2, $result[1]);
+    }
+
+    /**
+     * SELECT PARTIAL u.{id, name}
+     *   FROM Doctrine\Tests\Models\CMS\CmsUser u
+     */
+    public function testMultipleColumnEntityQueryWithoutScalarMap(): void
```
greg0ire: Maybe we should throw an exception when the user selects more than one field instead? What is the point of selecting more than one field, when in the end you only get one?

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @throws \Doctrine\DBAL\Driver\Exception
+     * @return array
+     */
+    protected function hydrateAssociativeData()
+    {
+        $result = [];
+
+        while ($data = $this->_stmt->fetchAssociative()) {
```
greg0ire: Shouldn't `fetchOne` be used here too?

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,51 @@
...
+            return $result;
+        }
+
+        $indexColumn = $this->_rsm->scalarMappings[$this->_rsm->indexByMap['scalars']];
+
+        $keys = array_keys(reset($result));
+
+        $column = isset($keys[1]) && $keys[0] === $indexColumn ? $keys[1] : $keys[0];
+
+        return array_column($result, $column, $indexColumn);
```
greg0ire: I don't understand why this needs to be that complicated…

The `ScalarHydrator` does more than this hydrator, but its code is more simple: https://github.com/doctrine/orm/blob/1de28c2cab5cd7e3c9d781a8beff22e328190911/lib/Doctrine/ORM/Internal/Hydration/ScalarHydrator.php#L32-L52


Why is all this needed?

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: I am trying to decouple things, so far things are good, after some time I will push latest changes, Which you guys can review.

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: https://gist.github.com/ostrolucky/f9f1e0b271357573fde55b7a2ba91a32

Based on this implementation, I am trying to cover situations with multiple columns selected and if Scalar Values provided.

```php
 $result = []; 
  
         while ($data = $this->_stmt->fetchOne()) { 
             $this->hydrateRowData($data, $result); 
         } 
  
         return $result; 
```

Is this enough ? @greg0ire 


```php
$result[] = $this->gatherScalarRowData($row); 
```

Also `fetchOne` gives non array values so `gatherScalarRowData` throws errors

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: https://github.com/doctrine/orm/pull/8919#discussion_r693490824

`fetchOne` throws error if we use `gatherScalarRowData` this to hydrate values

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan:
```php
 * Hydrator that produces one-dimensional array that is often (not always) interchangeable
 * with the corresponding object for read-only access.
```

Is this OK ?

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: It goes into Infinite loop

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: How `fetchOne` should behave ? 


```php
    public function fetchOne()
    {
        $current = current($this->resultSet);
        next($this->resultSet);

        return $current;
    }
```

is this correct ? @greg0ire

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: It introduces infinite loop problem

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
greg0ire: > that is often (not always) interchangeable with the corresponding object for read-only access.

I think you can remove this part of the comment, to understand why it does not make sense to have it here, you can read my recent answer on this SO question: https://stackoverflow.com/questions/30050886/doctrine-orm-hydrate-scalar-vs-hydrate-array/68795117#68795117

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
greg0ire: Yes that does look correct 👍

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
greg0ire: Not sure `gatherScalarRowData` will be useful here :thinking: I mean we don't care about getting the fieldname that corresponds to the column name we are fetching here, since in the end we will not return it.

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,36 @@
...
+        if (count($this->_rsm->fieldMappings) > 1) {
+            throw MultipleSelectorsFoundException::create($this->_rsm->fieldMappings);
+        }
+
+        $result = [];
+
+        $column = array_keys($this->_rsm->fieldMappings)[0];
+
+        while ($row = $this->_stmt->fetchOne()) {
+            $result[] = $row[$column];
```
greg0ire: Maybe we could simplify it further with `$result[] = current($row)`? That way you do not need to compute `$column`.

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: Simplified implementation accordingly

lib/Doctrine/ORM/Exception/MultipleSelectorsFoundException.php
```diff
@@ -0,0 +1,20 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Exception;
+
+class MultipleSelectorsFoundException extends ORMException
+{
+    public const MULTIPLE_SELECTORS_FOUND_EXCEPTION = 'Multiple selectors found: %s. Please select only one.';
+
+    public static function create(array $selectors): self
```
greg0ire, line 17:
```suggestion
    /**
     * @param string[] $selectors
     */
    public static function create(array $selectors): self
```

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: updated

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
bhushan: Updated

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: Updated Hydrator, now it's simplified.

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,33 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Internal\Hydration;
+
+use Doctrine\ORM\Exception\MultipleSelectorsFoundException;
+
+/**
+ * Hydrator that produces one-dimensional array.
+ */
+class ScalarColumnHydrator extends AbstractHydrator
```
greg0ire:
```suggestion
final class ScalarColumnHydrator extends AbstractHydrator
```

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * {@inheritdoc}
+     * @throws \Doctrine\ORM\Exception\MultipleSelectorsFoundException
+     * @throws \Doctrine\DBAL\Driver\Exception
+     */
+    protected function hydrateAllData()
```
greg0ire:
```suggestion
    protected function hydrateAllData(): array
```

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
```diff
@@ -0,0 +1,134 @@
...
+            ],
+        ];
+
+        $stmt = new HydratorMockResult($resultSet);
+        $hydrator = new ScalarColumnHydrator($this->entityManager);
+
+        $this->expectException(MultipleSelectorsFoundException::class);
+        $this->expectExceptionMessage(sprintf(
+            MultipleSelectorsFoundException::MULTIPLE_SELECTORS_FOUND_EXCEPTION,
+            implode(', ', $rsm->fieldMappings)
```
greg0ire: It's a good idea to reduce complexity in tests, otherwise you would need to test your tests :stuck_out_tongue: 
```suggestion
            'id, name'
```

bhushan: I will update 🤪, but I want to see github actions errors so everything can be fixed together.

greg0ire: It's starting to look good :slightly_smiling_face:

bhushan: I learned so many things in last two days 🤘🏻 thank you

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: made Hydrator and Exception `final`

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: Typehint added

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
bhushan: Updated

greg0ire:
Running `vendor/bin/phpcbf` should fix many of the coding standard issues.

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,44 @@
...
+use Doctrine\ORM\Exception\MultipleSelectorsFoundException;
+
+use function count;
+use function current;
+use function is_array;
+
+/**
+ * Hydrator that produces one-dimensional array.
+ */
+final class ScalarColumnHydrator extends AbstractHydrator
```
greg0ire, line 15: I'm going to make a separate PR that should help with these static analysis issues.

bhushan: Thank you 🙏🏻

greg0ire: Once https://github.com/doctrine/orm/pull/8945 is merged, it will be easy to fix the issues you have here

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
```diff
@@ -258,6 +256,31 @@ public function testBasicQuery(): void
...
+        $this->_em->persist($bhushan);
+
+        $this->_em->flush();
+
+        $query = $this->_em->createQuery('select u.username from Doctrine\Tests\Models\CMS\CmsUser u');
+
+        $users = $query->getSingleColumnResult();
+
+        self::assertCount(2, $users);
+        self::assertTrue(in_array('bhushan', $users));
```
bhushan: Only `postgres` has different order of these values, so changed it to use `in_array` for assertion.

cc @greg0ire

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
greg0ire: Please rebase and use `statement()` and `resultSetMapping()` instead of `_stmt` and `_rsm`.

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @throws MultipleSelectorsFoundException
+     * @throws Exception
+     */
+    protected function hydrateAllData(): array
+    {
+        if (count($this->_rsm->fieldMappings) > 1) {
```
greg0ire:
```suggestion
        if (count($this->resultSetMapping()->fieldMappings) > 1) {
```

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,44 @@
...
+            throw MultipleSelectorsFoundException::create($this->_rsm->fieldMappings);
+        }
+
+        $result = [];
+
+        while ($row = $this->_stmt->fetchOne()) {
```
greg0ire:
```suggestion
        while ($row = $this->statement()->fetchOne()) {
```

greg0ire (dismissed):

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: updated

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: updated

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: rebased the branch, and use these getters

bhushan:
@greg0ire PR is ready for final review, All linters and tests should be green at this moment, Please review when you have time :)

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
SenseException: What if you add an order in your query? The hydration format should still be the same, despite the order of the result.

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
```diff
@@ -0,0 +1,124 @@
...
+
+        $rsm->addEntityResult(CmsUser::class, 'u');
+        $rsm->addFieldResult('u', 'u__id', 'id');
+
+        $stmt     = new ResultMock($emptyResultSet = []);
+        $hydrator = new ScalarColumnHydrator($this->entityManager);
+        $result   = $hydrator->hydrateAll($stmt, $rsm);
+
+        $this->assertTrue(is_array($result));
+        $this->assertEquals(0, count($result));
```
SenseException:
```suggestion
        $this->assertCount(0, $result);
```

`assertEmpty()` should work too. The same with the other tests of ScalarColumnHydratorTest.

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
```diff
@@ -0,0 +1,124 @@
...
+        $rsm = new ResultSetMapping();
+
+        $rsm->addEntityResult(CmsUser::class, 'u');
+        $rsm->addFieldResult('u', 'u__id', 'id');
+
+        $stmt     = new ResultMock($emptyResultSet = []);
+        $hydrator = new ScalarColumnHydrator($this->entityManager);
+        $result   = $hydrator->hydrateAll($stmt, $rsm);
+
+        $this->assertTrue(is_array($result));
```
SenseException:
```suggestion
        $this->assertIsArray($result));
```

The same with the other tests of ScalarColumnHydratorTest.

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -0,0 +1,44 @@
...
+        if (count($this->resultSetMapping()->fieldMappings) > 1) {
+            throw MultipleSelectorsFoundException::create($this->resultSetMapping()->fieldMappings);
+        }
+
+        $result = [];
+
+        while ($row = $this->statement()->fetchOne()) {
+            if (is_array($row)) {
+                $result[] = current($row);
+                continue;
```
SenseException: Codecov sees uncovered code here.

SenseException (changes requested):

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
bhushan: Updated

tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
bhushan: Updated

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
bhushan: with new implementation of `fetchOne()`, these lines not required, removed it.

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
bhushan: @SenseException 

It is not actually related to this functionality,

```php
$gregoire           = new CmsUser();
        $gregoire->name     = 'Gregoire';
        $gregoire->username = 'greg0ire';
        $gregoire->status   = 'developer';
        $this->_em->persist($gregoire);

        $bhushan           = new CmsUser();
        $bhushan->name     = 'Bhushan';
        $bhushan->username = 'bhushan';
        $bhushan->status   = 'developer';
        $this->_em->persist($bhushan);
```

These lines create records `['bhushan', 'greg0ire']` when MySQLDriver and SqlDriver are used but when we use `PostgresDriver` records are stored `['greg0ire', 'bhushan']` like this which breaks tests.

May be its a bug or feature of PostgresDriver, I am not sure about this, You know better than me about this.

Regarding original question, (Please correct me if I am wrong) Hydrator is used only after query execution to fill up data, so order by should also work, Let me write test for it.

bhushan: Test added

bhushan: Tests are added and its working fine with order by clause. 

Do you think its a bug 🐛 in dbal which we should report ? 

Cc @greg0ire @SenseException

greg0ire: I think if you don't specify an order, you should not assume any. In fact, you should assume that an implementation could return one order or the other on subsequent calls. In practice, it's very unlikely that they would, but still. See for instance that Stack Overflow question: https://stackoverflow.com/a/10064571/353612

greg0ire (commented):
Please add docs for the new feature :slightly_smiling_face:

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
SenseException: So `testSingleColumnQuery()` still needs an order to compare it with an expected array.

bhushan: Should we mark this as resolved ?

bhushan:
> Please add docs for the new feature :slightly_smiling_face: 

@greg0ire where to find/add docs ?

greg0ire:
@bhushan in docs/en/reference/dql-doctrine-query-language.rst, there is a paragraph called "Hydration modes"

bhushan:
> @bhushan in docs/en/reference/dql-doctrine-query-language.rst, there is a paragraph called "Hydration modes"

@greg0ire updated doc to represent new feature

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -1272,6 +1273,25 @@ You can use the ``getSingleScalarResult()`` shortcut as well:
     <?php
     $numArticles = $query->getSingleScalarResult();
 
+Scalar Column Hydration
+^^^^^^^^^^^^^^^^^^^^^^^
+
+If you have a query which returns one-dimensional scalar values you can use
+scalar column hydration:
```
greg0ire:
```suggestion
If you have a query which returns a one-dimensional array of scalar values
you can use scalar column hydration:
```

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
SenseException: I thought about adding "order by" here too, so you can assert the result to an expected array.

greg0ire: :thinking: no… you did not add the ORDER, did you?

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
```diff
@@ -258,6 +256,64 @@ public function testBasicQuery(): void
...
+
+        $bhushan           = new CmsUser();
+        $bhushan->name     = 'Bhushan';
+        $bhushan->username = 'bhushan';
+        $bhushan->status   = 'developer';
+        $this->_em->persist($bhushan);
+
+        $this->_em->flush();
+
+        $query = $this->_em->createQuery('select u.username from Doctrine\Tests\Models\CMS\CmsUser u');
```
greg0ire:
```suggestion
        $query = $this->_em->createQuery('select u.username from Doctrine\Tests\Models\CMS\CmsUser u ORDER BY u.username');
```

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
bhushan: I wrote another test which asserts these things, in both `ASC` and `DESC` order, In this test we are checking only functionality with normal query

docs/en/reference/dql-doctrine-query-language.rst
bhushan: are you going to squash merge ? so I will commit this suggestions..

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
bhushan: There is already one more test to assert it with Order as `ASC` and `DESC` Orders

docs/en/reference/dql-doctrine-query-language.rst
bhushan: Pushed it, already

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
greg0ire: Ah right, might be worth testing. Works for me 👍 What do you think, @SenseException ?

greg0ire (dismissed):
LGTM

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
SenseException: The reason I asked for an order was for an easier assertion in the tests that fits for all platforms. So that something like

``` php
$expected = [
    'bhushan',
    'greg0ire',
];

self::assertSame($expected, $users);
```

Can be done. The order isn't relevant for the hydrator itself, but I want to prevent failing tests on different platforms. The hydrator is about the resulting data structure. Based on that information the tests only need either ASC or DESC and call the test `testSingleColumnQuery()`. The others are redundant.

SenseException (changes requested):

tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
bhushan: @SenseException with respect, I disagree on the approach you are suggesting. Just for the sake of easier assertions, we are polluting test with extra thing (order by clause). `testSingleColumnQuery()` is only responsible for testing if the data is one dimensional with base query (`select id from users`).

and based on above mentioned stack overflow answer, it is different databases which chooses different strategies to order data when specific order is not defined in query itself.

also it will give ambiguity for other devs if we are testing against `select id from users order by id DESC` this query, then what happens with `select id from users`.

Still, If you are not convinced, then I will do modifications in PR as suggested 🙂

bhushan: @SenseException please share your thoughts on this :)

SenseException: A hydrator is responsible for a resulting data structure. If it returns an array with assoc arrays as elements or an an array of objects or one array with elements of one column. This is what the tests should cover.

Usually a `select id from users` would be sufficient as test-query for its simplicity, but the mentioned different results between databases need to introduce a different query where the result is consistent. You also could use a much more complex query, but that would be unnecessary for a test case. That is why DESC and ASC (and neither of both) are actually testing the same case for a hydrator and its data structure. The order does not matter for the hydrator data structure. It was just added for multi platform compatibility.

From a documentation perspective of a test and what is expected from its result

``` php
$expected = [
    'bhushan',
    'greg0ire',
];
```

would be easier to read.

bhushan: removed order by tests and updated `testSingleColumnQuery` this test to use `DESC` orders for assertions

greg0ire (dismissed):

ostrolucky:
Does this do hydration? What happens if I select non-scalar field, eg. DateTime?

SenseException (dismissed):

bhushan:
> Does this do hydration? What happens if I select non-scalar field, eg. DateTime?

Sorry, I didn't understand question!!

greg0ire:
I think the question is: if you select a field hydrated into a `DateTimeImmutable`, do you actually with a `DateTimeImmutable` instance, or just with a string like `2021-08-31`?

bhushan:
> I think the question is: if you select a field hydrated into a `DateTimeImmutable`, do you actually with a `DateTimeImmutable` instance, or just with a string like `2021-08-31`?

In current implemented hydrator it gives string only, i am not sure before this hydrator any date-time hydration will take place or not.. 
Could you give me example scenario/code i will write test for it.. and if it fails i will write logic in code if *its needed * @ostrolucky   

Also do we have any other scenarios where non scalar values needs to be hydrated properly?

ostrolucky:
Well yes, that's difference between ORM and DBAL. ORM always hydrates, this is why in my example I extended from ArrayHydrator. As for test case, it's as simple as adding \DateTime typed $date property with default value in CmsUser and select `u.date` instead of `u.username`

greg0ire:
> ORM always hydrates

Correct me if I'm wrong, but I don't think that's true, see for instance https://github.com/doctrine/orm/blob/df5086196fafc47f1b3e548c8e6ce9ab48c832b6/lib/Doctrine/ORM/Internal/Hydration/ScalarHydrator.php#L26-L28

ostrolucky:
Hmm then I guess it's fine?

bhushan:
> Hmm then I guess it's fine?

What should we do in this case you suggest?

bhushan:
@ostrolucky could you please suggest what should be done here ? 😀

ostrolucky:
i think it's fine

bhushan:
I am checking from few days some checks are stuck at  `Expected — Waiting for status to be reported` What is that ?

greg0ire:
> I am checking from few days some checks are stuck at Expected — Waiting for status to be reported What is that ?

We changed the pipeline recently, a rebase should fix it.

greg0ire (approved):

greg0ire:
Thanks @bhushan ! And congrats!

PR #8920, opened 2021-08-16
Avoid returning null from `EntityManager::getReference()` and `getPartialReference()`

simPod:
_This bothers me for a long time, though I'm not sure the solution is viable._

Though I'd rather not assert for `notNull` every time I use the reference.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -746,6 +746,10 @@ following classes and methods:
 
 Use `toIterable()` instead.
 
+## BC BREAK: `Doctrine\ORM\EntityManagerInterface#getReference()` and `getPartialReference()` does not return `null` anymore
+
+Method `Doctrine\ORM\EntityManagerInterface#getReference()` and `getPartialReference()` throws `InstanceOfTheWrongTypeEncountered` in 3.0 when different entity type is found in inheritance hierachy.
+
 # Upgrade to 2.20
 
 ## Add `Doctrine\ORM\Query\OutputWalker` interface, deprecate `Doctrine\ORM\Query\SqlWalker::getExecutor()`
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -162,7 +162,7 @@
       <code><![CDATA[$entity]]></code>
       <code><![CDATA[$entity]]></code>
       <code><![CDATA[$entity]]></code>
-      <code><![CDATA[$entity instanceof $class->name ? $entity : null]]></code>
+      <code><![CDATA[$entity]]></code>
       <code><![CDATA[$persister->load($sortedId, null, null, [], $lockMode)]]></code>
       <code><![CDATA[$persister->loadById($sortedId)]]></code>
       <code><![CDATA[$this->metadataFactory]]></code>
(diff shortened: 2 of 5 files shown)


lib/Doctrine/ORM/EntityManagerInterface.php
```diff
@@ -170,8 +170,8 @@ public function createQueryBuilder();
      * @param mixed  $id         The entity identifier.
      * @psalm-param class-string<T> $entityName
      *
-     * @return object|null The entity reference.
-     * @psalm-return ?T
+     * @return object The entity reference.
+     * @psalm-return T
```
lcobucci: AFAIK this should be `T&Proxy` (applicable to `getPartialReference()` too).

simPod: You're right but I'd rather not mix things up cuz review than eventually drags for a long time.

I have opened https://github.com/doctrine/orm/pull/8922

beberlei:
This is inconsistent from my POV, `EntityManager::find` also returns null if not found and these two methods are related.

simPod:
@beberlei but this does not do any entity lookup. Currently, null is returned when there's issue with entity hierarchy when inheritance is used and that should IMO be an error.

lib/Doctrine/ORM/Exception/InstanceOfTheWrongTypeEncountered.php
```diff
@@ -0,0 +1,19 @@
...
+use LogicException;
+
+use function get_class;
+
+final class InstanceOfTheWrongTypeEncountered extends LogicException implements ORMException
+{
+    /** @param object $instance */
+    public static function forInstance($instance): self
+    {
+        return new self('Instance of the wrong type (' . get_class($instance) . ') was retrieved in inheritance hierachy.' .
```
derrabus:
```suggestion
        return new self('Instance of the wrong type (' . get_debug_type($instance) . ') was retrieved in inheritance hierachy.' .
```

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC1041Test.php
```diff
@@ -29,7 +30,11 @@ public function testGrabWrongSubtypeReturnsNull(): void
...
 
         self::assertNull($this->_em->find(Models\Company\CompanyFlexContract::class, $id));
-        self::assertNull($this->_em->getReference(Models\Company\CompanyFlexContract::class, $id));
-        self::assertNull($this->_em->getPartialReference(Models\Company\CompanyFlexContract::class, $id));
+
+        $this->expectException(InstanceOfTheWrongTypeEncountered::class);
+        $this->_em->getReference(Models\Company\CompanyFlexContract::class, $id);
+
+        $this->expectException(InstanceOfTheWrongTypeEncountered::class);
+        $this->_em->getPartialReference(Models\Company\CompanyFlexContract::class, $id);
```
derrabus: This part of the test will never be executed.

beberlei:
Before more work is done here, i believe the BC break here is too big to be reasonable and i am heavily 👎 on this change

simPod:
@beberlei what about adding new methods if BC break is the issue even for the major version?

PR #8922, opened 2021-08-16
Fix return type at `EntityManagerInterface::get(Partial)Reference()`

simPod:
@lcobucci mentioned this in https://github.com/doctrine/orm/pull/8920#discussion_r689589936


derrabus:
Looks reasonable. Can you make PHPStan and Psalm happy?

simPod:
Ok, it's happy now.

derrabus:
Looks like we need to do the same with `UnitOfWork` as well to get rid of the baselined errors. But anyway, your changes should improve the DX already, so let's merge.

derrabus (approved):

PR #8931, opened 2021-08-20
Deprecate reliance on non-optimal defaults

### LINKED ISSUE #5614: GeneratedValue strategy AUTO should use IDENTITY on PostgreSQL
As per http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/basic-mapping.html#identifier-generation-strategies - AUTO uses SEQUENCE by default on PostgreSQL. This seems sub-optimal as it forces INSERT queries to provide the id explicitly using `nextval('sequence_name')` which is not easy to do in ORM queries, and requires that you know the sequence name. On the other hand if it's set to IDENTITY it works better as it does it like MySQL's AUTO_INCREMENT, you can just omit the id field entirely.

As doing this would cause schema changes I can imagine it won't be fixed in a minor release, but should be considered for v3 I think.

### LINKED ISSUE #6429: GeneratedValue strategy AUTO should use IDENTITY on PostgreSQL

### LINKED ISSUE #8893: Deprecate `AUTO` identifier generator strategy
This strategy allows to pick the preferred strategy given the current platform. It allows full portability because you are guaranteed to have a working strategy no matter which RDBMS you use.

The issue is that a given strategy might be optimal at some point in time, then letter better strategies emerge.

For instance, for PostgreSQL, `AUTO` translates to using sequences, but, as mentioned in https://github.com/doctrine/dbal/issues/5614 , that might no longer be the best strategy.

Since we cannot really release a new major version of the ORM every time this happens, I suggest we deprecate the `AUTO` strategy and let people explicitly pick one.

The obvious drawback is that full portability is lost.

greg0ire:
What was optimal 10 years ago no longer is, and things might change in
the future. Using `AUTO` is still the best solution in most cases, and it
should be easy to make it mean something else when it is not.

Closes doctrine/orm#8893, and maybe we should close doctrine/dbal#5614 as well.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,33 @@
 # Upgrade to 2.17
 
+## Deprecated: reliance on the non-optimal defaults that come with the `AUTO` identifier generation strategy
+
+When the `AUTO` identifier generation strategy was introduced, the best
+strategy at the time was selected for each database platform.
+A lot of time has passed since then, and support for better strategies has been
+added.
+
+Because of that, it is now deprecated to rely on the historical defaults when
+they differ from what we recommend now.
+
+Instead, you should pick a strategy for each database platform you use, and it
+will be used when using `AUTO`. As of now, only PostgreSQL is affected by this.
+It is recommended that PostgreSQL user configure their new applications to use
+`IDENTITY`:
+
+```php
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
+use Doctrine\ORM\Configuration;
+
+assert($configuration instanceof Configuration);
+$configuration->setIdentityGenerationPreferences([
+    PostgreSQLPlatform::CLASS => ClassMetadata::GENERATOR_TYPE_IDENTITY,
+]);
+```
+
+If migrating an existing application is too costly, the deprecation can be
+addressed by configuring `SEQUENCE` as the default strategy.
+
 ## Deprecate `EntityManagerInterface::getPartialReference()`
 
 This method does not have a replacement and will be removed in 3.0.
@@ -14,7 +42,7 @@ Ensure `Doctrine\ORM\Configuration::setLazyGhostObjectEnabled(true)` is called t
 ## Deprecated accepting duplicate IDs in the identity map
 
 For any given entity class and ID value, there should be only one object instance
-representing the entity. 
+representing the entity.
 
 In https://github.com/doctrine/orm/pull/10785, a check was added that will guard this
 in the identity map. The most probable cause for violations of this rule are collisions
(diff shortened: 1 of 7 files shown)


greg0ire:
@derrabus I'm a bit unfamiliar with symfony recipes, would it be possible to add a configuration key iff the platform family is `postgresql`?

derrabus:
> @derrabus I'm a bit unfamiliar with symfony recipes, would it be possible to add a configuration key iff the platform family is `postgresql`?

No, but you could add a piece of configuration and comment it out. That would make it discoverable.

That being said, how do we want to leverage this setting in DoctrineBundle? I'd imagine it would be something like this:

```YAML
doctrine:
    orm:
        identity_generation_preferences:
            oracle: identity
            postgresql: identity
```

And that block is something we could add as-is to the recipe. If an application would use e.g. MySQL, those settings would just become no-ops.

greg0ire:
Yes that's how I imagine it should look like (except `oracle` does not support `identity` I think)

morozov:
I believe the current set of `GENERATOR_TYPE_*` constants mixes together a set of different concerns:
1. Which identifiers will be generated automatically? A sequence of integers (1, 2, 3, …) or UUIDs.
2. How the generation of numbers is implemented at the target platform (identity column or sequence).
3. Who decides on everything above (the user, the target platform, or the ORM).

It seems to me that each of the aspects should be represented by a separate abstraction at the ORM level. Mixing them together makes it challenging to reason about.

greg0ire:
I'm resuming this… indeed, there is a mix of concerns here. It's not clear to me however how to fix it or whether it should block this PR (which does introduce more usage of these constants).

morozov (approved):

DannyvdSluijs:
@greg0ire Thanks for all the work and I understand the reasoning. I do think the deprecation is a bit to opmisticly implemented. As for a `Postgres` platform it will always through the deprecation if no IdentityGenerationPreferences has been set in the configuration. Currently the`composer require doctrine/doctrine-bundle` doesn't have support for the  IdentityGenerationPreferences.

As a shortcut I would add the correct strategy `#[ORM\GeneratedValue(strategy: 'SEQUENCE')]` attribute but this leads to the deprecation being thrown as the predefined value on the Entity attributes isn't inspected. Is there another way around this?

greg0ire:
> Currently thecomposer require doctrine/doctrine-bundle doesn't have support for the IdentityGenerationPreferences.

You still probably have access to the configuration object though, right? Maybe in `src/Kernel.php`?

PR #8932, opened 2021-08-21
Drop support for generating json array fields

greg0ire:
`JSON_ARRAY` has been deprecated for a while, we should not encourage
people to generate new entities with it, since it will introduce
technical debt for them.

### DIFF
--- a/lib/Doctrine/ORM/Tools/EntityGenerator.php
+++ b/lib/Doctrine/ORM/Tools/EntityGenerator.php
@@ -190,7 +190,7 @@ class EntityGenerator
     /**
      * Hash-map for handle types.
      *
-     * @psalm-var array<Types::*|Type::*, string>
+     * @psalm-var array<Types::*|'json_array', string>
      */
     protected $typeAlias = [
         Types::DATETIMETZ_MUTABLE => '\DateTime',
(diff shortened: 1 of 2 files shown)


SenseException:
Shouldn't its removal happen in ORM 3?

greg0ire:
I don't think there is a big BC break here, since everything will work fine in production after updating to this commit. What will change I think is that developers will not be able to create more entities using `JSON_ARRAY` than already exist in their projects, and that's great because it is deprecated.

As far as I understand, this piece of code is used to generate php code from hand-written YAML or XML mapping files.

beberlei:
Json_array was replaced with Json type. Shouldnt we add that?

greg0ire:
Oh right, I didn't notice it wasn't in the list… will do!

SenseException:
If I'm not mistaken then there was no deprecation triggered for `json_array` in ORM. Shouldn't this happen before we remove a type from a release?

greg0ire:
@SenseException the context here is entity generation, so I don't think there should be a deprecation here. We could however add one for 7277afefb9799438ca61fc939773f9e10ddad697, which is called when the mapping is loaded from a database.

derrabus:
Note that the class we're patching here is already deprecated.

lib/Doctrine/ORM/Tools/EntityGenerator.php
```diff
@@ -205,7 +205,7 @@ class EntityGenerator
         Types::BLOB               => 'string',
         Types::DECIMAL            => 'string',
         Types::GUID               => 'string',
-        Type::JSON_ARRAY          => 'array',
+        Types::JSON               => 'array',
```
derrabus, line 209:
```suggestion
        'json_array'              => 'array',
        Types::JSON               => 'array',
```

This class is flagged for removal anyway. I would not introduce breaking changes at this point. Let's simply inline the deprecated constant so we get a clean build on DBAL 3.

derrabus (approved):

PR #8934, opened 2021-08-22
Remove calls to `AbstractPlatform::getSQLResultCasing()`

### LINKED ISSUE #8887: Move `getSQLResultCasing` code into ORM from DBAL
DBAL 2 deprecates and DBAL 3 removes the `getSQLResultCasing` API, which we use to work around Oracle, PostgreSQL and MySQL differences in the default result casing. 

To avoid a larger refactoring and be forwards compatible, we need move this functionality into ORM and provide a strategy pattern with different implementations based on DBAL 2 or 3.

```php
interface ResultCasingStrategy
{
    public function getSQLResultCasing(string $string) : string;
}
class DBAL2ResultCasingStrategy implements ResultCasingStrategy {
     public function getSQLResultCasing(string $string) : string {
         return $this->platform->getSQLResultCasing($string);
     }
}
class UpperCaseResultCasingStrategy implements ResultCasingStrategy {}
class LowerCaseResultCasingStrategy implements ResultCasingStrategy {}
class NullCasingStrategy implements ResultCasingStrategy {}
```
A factory method at some point in the ORM should select the right strategy based on DBAL version and database platform and make the service available to everyone who needs sql result casings.

derrabus:
`AbstractPlatform::getSQLResultCasing()` has been removed in DBAL 3. This PR is a pragmatic approach to move that logic into the ORM.

DBAL should be able to enforce a certain result casing for us. While the proposed change further unblocks the DBAL 3 migration, I think that in the long run we should try to leverage that feature instead.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Internal\Hydration;
 
+use Doctrine\ORM\Internal\SQLResultCasing;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Query;
 use Exception;
@@ -19,6 +20,8 @@ use function sprintf;
 
 class SimpleObjectHydrator extends AbstractHydrator
 {
+    use SQLResultCasing;
+
     /** @var ClassMetadata */
     private $class;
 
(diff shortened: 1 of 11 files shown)


greg0ire:
Psalm's baseline needs to be reset with `vendor/bin/psalm --set-baseline=psalm-baseline.xml`.

You might want to run `vendor/bin/psalm --update-baseline` on 2.10.x first, then commit the result, then apply your changes and reset the baseline, like I did in #8932

derrabus:
Okay, let's merge #8936 first then. 😃

greg0ire (approved):

SenseException (approved):
When was `getSQLResultCasing` removed from DBAL?

greg0ire:
One year ago, in https://github.com/doctrine/dbal/pull/4234 :slightly_smiling_face:

greg0ire:
Thanks @derrabus !

beberlei:
:+1: This is simple and easy to work off in the future. Thanks.

It solves #8887 because it allows custom platforms to just have `getSQLResultCasing` as an extension point.

PR #8939, opened 2021-08-22
Remove calls to `Driver::getName()`

derrabus:
`Driver::getName()` is gone in DBAL 3. Let's not call it.


greg0ire (approved):

greg0ire:
Thanks @derrabus !

PR #8940, opened 2021-08-22
Remove calls to `EchoSQLLogger`

derrabus:
`EchoSQLLogger` is gone in DBAL 3. Do we actually need it?


greg0ire (approved):
If it's just tests I think it's fine to remove them.

PR #8941, opened 2021-08-22
Remove calls to `fixSchemaElementName()`

derrabus:
Yet another method that has been removed in DBAL 3.


lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -669,6 +671,17 @@ private function determineIdGeneratorStrategy(AbstractPlatform $platform): int
...
+    private function truncateSequenceName(string $schemaElementName): string
+    {
+        $maxIdentifierLength = $this->getTargetPlatform()->getMaxIdentifierLength();
+
+        if (strlen($schemaElementName) > $maxIdentifierLength) {
+            return substr($schemaElementName, 0, $maxIdentifierLength);
+        }
+
+        return $schemaElementName;
+    }
```
greg0ire, line 683: It's weird because although all platforms have a max identifier length since they inherit it from `AbstractPlatform`, only SQLAnywhere and Oracle since to act on it… for other platforms, I suppose there is a crash and that's it.

Maybe we should trigger a deprecation on line 679 saying this will happen in the future? Or maybe even change the condition so that it only applies for `sqlanywhere` and `oracle`?

derrabus: In any case, those sequence names that we truncate here are generated by our own code, no user input. A deprecation would be confusing, I guess.

I've limited the truncation to those two platforms now.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -57,6 +59,11 @@
  */
 class ClassMetadataFactory extends AbstractClassMetadataFactory
 {
+    private const SEQUENCE_NAME_NEEDS_TRUNCATION = [
```
greg0ire: Why not use a list here, and `in_array` below?

derrabus: Me micro-optimizing stuff. A lookup by key is faster than searching the array for a value. Then again, the array only has two elements. 🙈 

What would be your preference?

greg0ire: Class metadata is cached , so you don't have to over-optimize here, that's for sure!

derrabus: `in_array()` it is

greg0ire (dismissed):

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -672,6 +666,22 @@ private function determineIdGeneratorStrategy(AbstractPlatform $platform): int
         return ClassMetadata::GENERATOR_TYPE_TABLE;
     }
 
+    private function truncateSequenceName(string $schemaElementName): string
```
SenseException, line 669: Codecov has multiple lines marked as uncovered for `truncateSequenceName()`. Are these lines relevant?

derrabus: Apparently, we're not testing sequences on Oracle. 🤔

greg0ire: We do test part of this method though, so there must be existing test that we might be enable for Oracle. Let's `debug_print_backtrace();` to figure this out :)

greg0ire: Apparently it's PostgreSQL (not suprised), but I couldn't get the name of the current test. Let's use `--debug`.

greg0ire: Ok so it's `Doctrine\Tests\ORM\Cache\DefaultCollectionHydratorTest::testImplementsCollectionEntryStructure` :thinking: It does use a `City` with a `GeneratedValue`, but the end goal does not really seem to be testing sequences.

greg0ire: Apparently, we're not testing ~sequences~ anything on Oracle

FTFY :stuck_out_tongue:

greg0ire (approved):
My attempt at running the tests on Oracle: https://github.com/doctrine/orm/pull/8958

It will require more work, in the meantime I think we can merge this.

PR #8942, opened 2021-08-22
Remove remaining call to `prefersSequences()`

derrabus:
This PR removes one call to `prefersSequences()` that slipped my eye. Here it is used to skip a test for platforms that don't "prefer" sequences.

While that test certainly was meant to reproduce a bug that occurred with sequences, I'm unsure if we really need to skip it in other environments.


greg0ire (approved):
Looks like it in fact works on those databases

PR #8945, opened 2021-08-22
Document possibly-null member variables

greg0ire:
Many of the variables in `AbstractHydrator` are not initialized in the
constructor, and should be documented as possibly null because of that.
Introducing accessors that perform null checks allows to to have to do
these null checks when using the accessors.
Making the member variables private would be a backwards-compatibility
break and could be considered for the next major version.

This makes Psalm's and PHPStan's baselines smaller, and should make
implementing new hydrators easier.

This should help with #8919

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -17,6 +17,7 @@ use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\ORM\Tools\Pagination\LimitSubqueryWalker;
 use Doctrine\ORM\UnitOfWork;
 use Generator;
+use LogicException;
 use ReflectionClass;
 use TypeError;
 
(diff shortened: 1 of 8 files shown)


lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -17,11 +17,13 @@
 use Doctrine\ORM\Tools\Pagination\LimitSubqueryWalker;
 use Doctrine\ORM\UnitOfWork;
 use Generator;
+use LogicException;
 use ReflectionClass;
 use TypeError;
 
 use function array_map;
 use function array_merge;
+use function assert;
```
derrabus: This is never used, is it?

greg0ire: Oh right, it's a remnant from a first attempt at fixing this, I'll remove it 👍

derrabus (approved):

PR #8946, opened 2021-08-22
Support for DBAL 3's platform classes

derrabus:
In DBAL 3, various platform classes have been removed or renamed. Furthermore, DBAL 3.2 will deprecate the `AbstractPlatform::getName()` method.

This PR makes the necessary adjustments in ORM.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -651,7 +651,11 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
 
     private function determineIdGeneratorStrategy(AbstractPlatform $platform): int
     {
-        if ($platform->getName() === 'oracle' || $platform->getName() === 'postgresql') {
+        if (
+            $platform instanceof Platforms\OraclePlatform
+            || $platform instanceof Platforms\PostgreSQL94Platform
+            || $platform instanceof Platforms\PostgreSQLPlatform
+        ) {
             return ClassMetadata::GENERATOR_TYPE_SEQUENCE;
         }
 
--- a/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryOutputWalker.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryOutputWalker.php
@@ -7,8 +7,10 @@ namespace Doctrine\ORM\Tools\Pagination;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Platforms\DB2Platform;
 use Doctrine\DBAL\Platforms\OraclePlatform;
-use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
+use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
 use Doctrine\DBAL\Platforms\SQLAnywherePlatform;
+use Doctrine\DBAL\Platforms\SQLServer2012Platform;
 use Doctrine\DBAL\Platforms\SQLServerPlatform;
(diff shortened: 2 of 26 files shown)


tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
```diff
@@ -31,9 +31,13 @@
 use Doctrine\Tests\OrmTestCase;
 use Exception;
 
+use function class_exists;
 use function get_class;
 use function sprintf;
 
+// DBAL 2 compatibility
+class_exists('Doctrine\DBAL\Platforms\MySqlPlatform');
```
derrabus, line 39: That's a tricky one. The class name was changed from `MySqlPlatform` to `MySQLPlatform`. Class names are case-insensitive in PHP, so just using the new name would theoretically work. However, on a case-sensitive filesystem, autoloading would fail if the case does not match. This is why I'm triggering the autoloader with old and new class name.

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryOutputWalker.php
```diff
@@ -50,6 +45,14 @@ class LimitSubqueryOutputWalker extends SqlWalker
 {
     private const ORDER_BY_PATH_EXPRESSION = '/(?<![a-z0-9_])%s\.%s(?![a-z0-9_])/i';
 
+    private const SUPPORTS_ROW_NUMBER = [
+        'db2' => true,
+        'mssql' => true,
+        'oracle' => true,
+        'postgresql' => true,
+        'sqlanywhere' => true,
+    ];
```
greg0ire: Let's use a list + `in_array` here too?

derrabus: I wasn't sure if `platformSupportsRowNumber()` might be on a hot execution path. A lookup by array-key is usually faster than a search.

greg0ire: I see. I don't think it's a hot path, we're in an output walker, the hot path in Doctrine would rather be in hydration-related features I think.

greg0ire: ping @derrabus :slightly_smiling_face:

derrabus: Now that `getName()` is deprecated, I need to solve this entirely different anyway. 🤷

greg0ire: Oh right, indeed 👍

derrabus:
PR is ready again.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -651,7 +651,11 @@ private function completeIdGeneratorMapping(ClassMetadataInfo $class): void
 
     private function determineIdGeneratorStrategy(AbstractPlatform $platform): int
     {
-        if ($platform->getName() === 'oracle' || $platform->getName() === 'postgresql') {
+        if (
+            $platform instanceof Platforms\OraclePlatform
+            || $platform instanceof Platforms\PostgreSQL94Platform
```
greg0ire, line 656: Maybe we should build against DBAL `3.2@dev`, that way you can check for `PostgreSQLPlatform`, since it was reinstated in https://github.com/doctrine/dbal/pull/4755 ?

derrabus: I don't know the ETA for DBAL 3.2.0. If DBAL 3.2 makes a stable release before ORM 2.10, I'll happily bump and remove the compat code again.

greg0ire (approved):

greg0ire:
Thanks @derrabus !

PR #8948, opened 2021-08-22
Remove ImportCommand from console

derrabus:
`ImportCommand` has been removed from DBAL.

This PR patches `ConsoleRunner` to only include `ImportCommand` if it's available.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
+++ b/lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
@@ -15,6 +15,8 @@ use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Command\Command as SymfonyCommand;
 use Symfony\Component\Console\Helper\HelperSet;
 
+use function class_exists;
+
 /**
  * Handles running the Console Tools inside Symfony Console context.
  */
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
```diff
@@ -73,41 +77,9 @@ public static function createApplication($helperSetOrProvider, array $commands =
...
-                new Command\GenerateEntitiesCommand($entityManagerProvider),
-                new Command\GenerateProxiesCommand($entityManagerProvider),
-                new Command\ConvertMappingCommand($entityManagerProvider),
-                new Command\RunDqlCommand($entityManagerProvider),
-                new Command\ValidateSchemaCommand($entityManagerProvider),
-                new Command\InfoCommand($entityManagerProvider),
-                new Command\MappingDescribeCommand($entityManagerProvider),
-            ]
-        );
+        $cli->addCommands(iterator_to_array(
```
greg0ire: :thinking: why not call `addCommands` twice, conditionally the first time? Having a generator only for it to be immediately converted into an array feels a bit clumsy

derrabus: Sure.

greg0ire (approved):

PR #8949, opened 2021-08-22
Make mock layer compatible with DBAL 3

### LINKED ISSUE #8886: Get PHPUnit to run with DBAL 3

derrabus:
😓 

Closes #8886


greg0ire:
Yay PHPUnit runs! :tada: :tada: :tada: 

> There were 1285 errors

:sweat_smile:

derrabus:
> > There were 1285 errors
> 
> 😅

If you merge my other open PRs, that number should go down to zero. 🤞🏻 😎

tests/Doctrine/Tests/Mocks/DriverResultMock.php
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchOne()
+    {
+        $row = $this->fetchNumeric();
+
+        return $row ? $row[0] : false;
```
greg0ire, line 62: `fetchOne` is about fetching one column, so it should look a bit more like `fetchAssociative` and not stop at 0 I think.

derrabus: > Returns the first value of the next row of the result

I think, my implementation is correct.

greg0ire: Oh right I misread, it is very much correct 👍

greg0ire (approved):

Issue #8884, opened 2021-08-05
Get Psalm job to pass with DBAL 3


greg0ire:
[Recently, a CI job was added on the 2.10.x branch to statically analyze this project after forcing the installation of `doctrine/dbal` v3](https://github.com/doctrine/orm/pull/8874).

In order not to disrupt contributors, we made sure it would not fail the build although it really does not pass at all:

https://github.com/doctrine/orm/blob/e09f126abfec0add53dc1ed8d4e3708117032fc1/.github/workflows/static-analysis.yml#L94

If you want to help us fix it or get closer to fixing it, you can reproduce the failures locally by running (from the 2.10.x branch)

```shell
composer require doctrine/dbal ^3.0
vendor/bin/psalm
```

When this issue is fixed, we should create a new one about making this job no longer experimental.

scyzoryck:
One of the loudest issue are changes introduced in https://github.com/doctrine/dbal/commit/5d03f6e443e8d444bf9917ca44b1b553ecaa4896 - it affects `Doctrine\ORM\Configuration` class. I wonder if it would be better to switch to the dedicated property per each config value, or stays with `$_attributes` property to keep backward compatibility with classes that might extend `Doctrine\ORM\Configuration`.

greg0ire:
Maybe re-declaring the property in the ORM class would solve the issue. Switching to a dedicated property per config value in the ORM class is a separate thing IMO, it can be done fully independently I think.

scyzoryck:
The last type of errors seems to be connected with things that has been deprecated in `doctrine/dbal` 2.x and removed in 3.x, for example: https://github.com/doctrine/dbal/pull/3211 . Do you think that it will be fine to thrown exception in that case if somebody is trying to use such functionalities with `doctrine/orm:2.10.x` and `doctrine/dbal:3.x`? It looks like all of such cases are marked as deprecated in `doctrine/orm:2.10.x`.

greg0ire:
Do you have an example?

scyzoryck:
Sure. 
`UuidGenerator` is using `getGuidExpression` function that has been removed in dbal 3.x doctrine/dbal#3211 
https://github.com/doctrine/orm/blob/f3e55fae9fdbdbc23897006bdbf016c20e11f6e9/lib/Doctrine/ORM/Id/UuidGenerator.php#L36
There is no internal replacement for this method - just suggested to use ramsey/uuid instead. 

[...]

PR #8953, opened 2021-08-24
Don't call deprecated `getSchemaManager()`

derrabus:
`Connection::getSchemaManager()` will be replaced by `Connection::createSchemaManager()`. We should call the new method if it's available.

Part of #8884


greg0ire (approved):

greg0ire:
Thanks @derrabus ! You're on fire :fire:

Issue #8885, opened 2021-08-05
Get Phpstan job to pass with DBAL 3


greg0ire:
[Recently, a CI job was added on the 2.10.x branch to statically analyze this project after forcing the installation of `doctrine/dbal` v3](https://github.com/doctrine/orm/pull/8874).

In order not to disrupt contributors, we made sure it would not fail the build although it really does not pass at all:

https://github.com/doctrine/orm/blob/e09f126abfec0add53dc1ed8d4e3708117032fc1/.github/workflows/static-analysis.yml#L52

If you want to help us fix it or get closer to fixing it, you can reproduce the failures locally by running (from the 2.10.x branch)

```shell
composer require doctrine/dbal ^3.0
vendor/bin/phpstan
```

When this issue is fixed, we should create a new one about making this job no longer experimental.

PR #8957, opened 2021-08-24
Only wire ConnectionHelper if it's available

derrabus:
DBAL deprecated its `ConnectionHelper` class, a helper for Symfony console applications. I propose to not use that class if it's not there.

Part of #8884, #8885.

I haven't found any tests nor usage of the `createHelperSet()` method I've changed here. The other helper that is wired here simply exposes a `getEntityManager()` method that… well… returns an entity manager. In modern Symfony applications, we would probably use DI to make the EM available to commands. I feel like we should follow the example of DBAL here and deprecate `createHelperSet()`, `EntityManagerHelper` and related classes and change our own commands to use DI instead. That could be done in a follow-up PR. WDYT?

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
+++ b/lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
@@ -27,12 +27,13 @@ final class ConsoleRunner
      */
     public static function createHelperSet(EntityManagerInterface $entityManager): HelperSet
     {
-        return new HelperSet(
-            [
-                'db' => new DBALConsole\Helper\ConnectionHelper($entityManager->getConnection()),
-                'em' => new EntityManagerHelper($entityManager),
-            ]
-        );
+        $helpers = ['em' => new EntityManagerHelper($entityManager)];
+
+        if (class_exists(DBALConsole\Helper\ConnectionHelper::class)) {
+            $helpers['db'] = new DBALConsole\Helper\ConnectionHelper($entityManager->getConnection());
+        }
+
+        return new HelperSet($helpers);
     }
 
     /**
--- a/phpstan-dbal3.neon
+++ b/phpstan-dbal3.neon
@@ -6,3 +6,11 @@ parameters:
     ignoreErrors:
         # deprecations from doctrine/dbal:3.x
         - '/^Call to an undefined method Doctrine\\DBAL\\Platforms\\AbstractPlatform::getGuidExpression\(\).$/'
+
+        # Fallback logic for DBAL 2
+        -
+            message: '/HelperSet constructor expects/'
+            path: lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
+        -
+            message: '/Application::add\(\) expects Symfony\\Component\\Console\\Command\\Command/'
+            path: lib/Doctrine/ORM/Tools/Console/ConsoleRunner.php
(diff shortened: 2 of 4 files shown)


greg0ire:
Maybe you figured out a lot of the following already, but in case it helps another reviewer understand this, let's go through it anyway.

This area of the code is unfamiliar to me so I did some digging, and found this page of the docs: https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/tools.html#tools

> In modern Symfony applications, we would probably use DI to make the EM available to commands. 

I think projects impacted by this are not Symfony applications, but non-Symfony applications. Users of these applications can still run doctrine commands by using `vendor/bin/doctrine`, provided they create a config file.

> I haven't found any tests nor usage of the createHelperSet() method I've changed here. 

That would be because it's supposed to be called in a file created by the users, not by Doctrine, as documented in the page above.

I think not wiring `db` anymore might be the right call though, because it's only useful for DBAL commands, and they no longer use it since https://github.com/doctrine/dbal/pull/3956/files#diff-48ddc2d792448b076086b4663a7aad88971526a9cc5ce6275964585258d3745eL124, so as long as we require DBAL 2.11 we should be fine IMO

greg0ire:
@PowerKiKi @dmaicher can you please advise about this? I think the BC break that might exist would be on commands manually registered by the user (in the config file). The docs say:

> If you want to use additional commands you have to register them yourself.

derrabus:
> > In modern Symfony applications, we would probably use DI to make the EM available to commands.
> 
> I think projects impacted by this are not Symfony applications, but non-Symfony applications. Users of these applications can still run doctrine commands by using `vendor/bin/doctrine`, provided they create a config file.

Fair enough, but the `bin/doctrine` script is code that we control. We can refactor it to use DI instead of that helper without changing the external behavior of the script.

> I think not wiring `db` anymore might be the right call though, because it's only useful for DBAL commands

We don't know if there is code out there that relies on `createHelperSet` returning a helper set that includes the `db` helper. Removing the helper entirely is a breaking change. This is why I'm only excluding it if it's not available anymore for now.

greg0ire (approved):
No reaction from the others, let's merge this 👍

greg0ire:
Thanks @derrabus !

PowerKiKi:
I assumed you asked for my inputs because I created #8327 a while ago.

My use-case is relatively simple. It is a non-Symfony app (Laminas Mezzio), and I strictly only use commands via `./vendor/bin/doctrine` and `./vendor/bin/doctrine-migrations`. So I have no direct use of those internal details. As long as the commands from both packages end up working (almost) out of the box with a shared configuration, I'm OK with the merge here.

greg0ire:
@PowerKiKi yes, that's why :slightly_smiling_face: I was thinking you would be a relevant reviewer since you are familiar with the binary (I don't use it at all, for instance)

dmaicher (commented):
Sorry I was busy the last days. This change looks good to me :+1:

PowerKiKi:
I just tested this PR and can confirm that the following commands still work as expected (after a few minor adjustments). So it's all good :tada:

- `./vendor/bin/doctrine-dbal dbal:run-sql 'SELECT 2+9'`
- `./vendor/bin/doctrine dbal:run-sql 'SELECT 2+9'`
- `./vendor/bin/doctrine orm:validate-schema`
- `./vendor/bin/doctrine-migrations migrations:status`

greg0ire:
Thanks for checking :bow:

PR #8959, opened 2021-08-25
Fix `getEntityChangeSet` return typehint

norkunas:
As discussed in #8955

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -590,7 +590,7 @@ class UnitOfWork implements PropertyChangedListener
      * @param object $entity
      *
      * @return mixed[][]
-     * @psalm-return array<string, array{mixed, mixed}>
+     * @psalm-return array<string, array{mixed, mixed}|PersistentCollection>
      */
     public function & getEntityChangeSet($entity)
     {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -590,7 +590,7 @@ private function executeExtraUpdates(): void
      * @param object $entity
      *
      * @return mixed[][]
-     * @psalm-return array<string, array{mixed, mixed}>
```
greg0ire, line 593: `array{mixed, mixed}` means an array with only one key (and value of course). Maybe that was wrong too, is that what you observe when you dump the return value?

norkunas: In all cases array{mixed, mixed} is okay, it's different only when replacing a collection - it gives a PersistentCollection directly instead of an array. I'd say it should return also an array, but that's out of scope

greg0ire (approved):

lib/Doctrine/ORM/UnitOfWork.php
SenseException: AFAIK It means two elements with the keys 0 and 1.

SenseException (approved):

greg0ire:
Thanks @norkunas !

norkunas:
Thanks for merging! :)

PR #8961, opened 2021-08-28
Deprecate / remove TABLE id generator strategy

greg0ire:
This strategy has been marked as TODO for more than 14 years. It should
be OK to remove some things related to it since they lead to an
exception being thrown.


derrabus (approved):

SenseException (approved):
As far as I can see  this was

* Untested code. You didn't have to remove code that was covering that todo-feature
* an unfinished feature (at least I assume that)

I guess this one is okay to be introduced in 2.10.

PR #8964, opened 2021-08-29
DBAL 3

### LINKED ISSUE #8884: Get Psalm job to pass with DBAL 3

derrabus:
Fixes #8884, #8885.


derrabus:
I don't understand the failure on Postgres. The test passes locally. 🤔

greg0ire:
I do reproduce the failure locally.

```
docker run -d -p 5432:5432 -e POSTGRES_PASSWORD=postgres postgres:9.6
vendor/bin/phpunit -c ci/github/phpunit/pdo_pgsql.xml   --stop-on-error
```

I use php 7.4

greg0ire:
This might be a new issue introduced in https://github.com/doctrine/dbal/pull/4082 … I don't understand why it only affects PostgreSQL though.
Previously, this worked because it resulted in

```php
    public function getDatabase(Connection $conn)
    {
        $params = $conn->getParams();

        return $params['dbname'] ?? $conn->query('SELECT CURRENT_DATABASE()')->fetchOne();
    }
```

Now it results in

`$this->connection = new PDO($dsn, (string) $user, (string) $password, (array) $options);`

greg0ire:
I'm currently working on a fix, we have the database name in `$testConnParams` in `TestUtil`

greg0ire:
Pushed the fix :)

derrabus:
We're green. The PR is ready, but because I've added a new parameter to various PHPUnit jobs, GitHub is waiting for jobs that don't exist anymore.

greg0ire:
Yes. We will have to tweak the settings, and it's going to be a bit complicated because we will have to do so for 2.10.x and 3.x but not for 2.9.x .

derrabus:
Shall I add `dbal-version: ["default"]` to 2.9.x as well? It wouldn't do anything there, but it would make the configration easier.

derrabus:
see #8965

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @derrabus ! That was impressive!

PR #8965, opened 2021-08-29
Add dbal-version dimension to CI matrix

derrabus:
This PR adds a `dbal-version` dimension to several CI jobs that doesn't do anything. This way we can have the same branch protection rules on `2.9.x` and `2.10.x` once we merge #8964.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -19,6 +19,8 @@ jobs:
           - "7.3"
           - "7.4"
           - "8.0"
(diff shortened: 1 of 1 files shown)


greg0ire:
I don't know, maybe we should do 2 patterns? The requirements will not be the same (for instance on 2.10.x, it will be required to have a 2.13 job passing)

derrabus:
You are right. My solution would only allow us to make the `default` builds mandatory for now. If that's not enough for us, we would need separate different configurations anyway.

PR #8979, opened 2021-09-05
Bump PHPStan and Psalm

derrabus:


SenseException:
I'd say that 2.9 should be the target branch because it is the default one.

derrabus:
> I'd say that 2.9 should be the target branch because it is the default one.

All right. I'm targeting 2.9.x now. This PR will also backport the changes of #8938 then.

SenseException (approved):

PR #8996, opened 2021-09-09
Support for PSR-6 result caches

derrabus:
Follows doctrine/dbal#4620

DBAL 3.2 will deprecate Doctrine Cache in favor of PSR-6 for result caches. This PR makes sure we call the non-deprecated methods.

I've also added a CI job that runs against a dev snapshot of DBAL, so we can actually test my changes.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -24,6 +24,8 @@ jobs:
         include:
           - php-version: "8.0"
             dbal-version: "2.13"
+          - php-version: "8.0"
+            dbal-version: "3.2@dev"
 
     steps:
       - name: "Checkout"
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Doctrine\ORM;
 
 use Countable;
+use Doctrine\Common\Cache\Psr6\CacheAdapter;
+use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
@@ -20,10 +22,12 @@ use Doctrine\ORM\Query\Parameter;
 use Doctrine\ORM\Query\QueryException;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\Persistence\Mapping\MappingException;
+use Psr\Cache\CacheItemPoolInterface;
 use Traversable;
 
 use function array_map;
 use function array_shift;
+use function assert;
 use function count;
 use function is_array;
 use function is_numeric;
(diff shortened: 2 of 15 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         include:
           - php-version: "8.0"
             dbal-version: "2.13"
+          - php-version: "8.0"
+            dbal-version: "3.2@dev"
```
greg0ire, line 28: What should happen to this when 3.2.0 is released? Should we change it to whatever the latest 3.1.x is then?

greg0ire:
I pushed an extra commit with tests for some of the uncovered paths, let's see if the build still passes.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -529,9 +534,25 @@ private function translateNamespaces(Query\ResultSetMapping $rsm): void
...
+            if (! $profile->getResultCacheDriver()) {
+                $defaultHydrationCacheImpl = $this->_em->getConfiguration()->getHydrationCacheImpl();
+                if ($defaultHydrationCacheImpl) {
+                    $profile = $profile->setResultCacheDriver($defaultHydrationCacheImpl);
+                }
+            }
+        } elseif (! $profile->getResultCache()) {
+            $defaultHydrationCacheImpl = $this->_em->getConfiguration()->getHydrationCacheImpl();
+            if ($defaultHydrationCacheImpl) {
+                $profile = $profile->setResultCache(CacheAdapter::wrap($defaultHydrationCacheImpl));
```
greg0ire, line 554: Weird, the new tests I introduced do cover that path according to `var_dump();die();`

greg0ire: Maybe https://github.com/doctrine/orm/pull/8964 did not take coverage perfectly into account?

greg0ire: Working on it at https://github.com/doctrine/orm/pull/8998

greg0ire: Looks better now.

greg0ire (commented):
@derrabus please review my commit and fix it up if it looks good to you

.github/workflows/continuous-integration.yml
derrabus: We heavily depend on DBAL. I think we should always have a CI job that checks against a dev version of DBAL. When 3.2.0 is released, this job will test against 3.3-dev automatically and I think that's fine.

derrabus:
> please review my commit and fix it up if it looks good to you

Done!

greg0ire (approved):

.github/workflows/continuous-integration.yml
greg0ire: Is there a syntax we could use to avoid this `2` being hardcoded here? It's a bit misleading. `3@dev` maybe?

greg0ire:
Thanks @derrabus !

PR #9001, opened 2021-09-11
Remove Proxy from EntityManagerInterface contract

derrabus:
After merging #8922 by @simPod, I had a deeper look that the implementation to see if we can fix some of the newly baselined errors. I'd like to partially revert the change now. Sorry for the confusion.

#8922 added an intersection with the `Proxy` interface to the return type of `EntityManagerInterface::getReference()` and `getPartialReference()`. Here's why I now think that this is problematic:

1. While our implementation `EntityManager` _might_ return such an object, we cannot guarantee that it will. If you look at the following code snippet, you can see that the actual entity is returned instead of a reference, if the entity has already been registered with the UoW. In that case, the returned entity might not be a proxy object.

   https://github.com/doctrine/orm/blob/04d28a93628157e7174f4322a1b1cbdbf60afe12/lib/Doctrine/ORM/EntityManager.php#L515-L520

   If we kept the `Proxy` interface in the return type, the correct type would be `(T&Proxy)|T|null`.

2. The `Proxy` interface is somewhat an implementation detail of our `EntityManager` implementation that should not be leaked to the interface. I believe, the ORM _should_ be able to move to a different proxy library without violating the contract of `EntityManagerInterface`.

### DIFF
--- a/lib/Doctrine/ORM/EntityManagerInterface.php
+++ b/lib/Doctrine/ORM/EntityManagerInterface.php
@@ -23,7 +23,6 @@ namespace Doctrine\ORM;
 use BadMethodCallException;
 use DateTimeInterface;
 use Doctrine\Common\EventManager;
-use Doctrine\Common\Proxy\Proxy;
 use Doctrine\DBAL\Connection;
 use Doctrine\ORM\Internal\Hydration\AbstractHydrator;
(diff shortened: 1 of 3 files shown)


greg0ire (approved):

greg0ire:
Thanks @derrabus !

PR #9004, opened 2021-09-11
Use PSR-6 for accessing the query cache

derrabus:


greg0ire:
Converting to draft since there is a `TODO`

derrabus:
> Converting to draft since there is a `TODO`

🙈 

Fixed.

lib/Doctrine/ORM/Query.php
```diff
@@ -489,16 +514,21 @@ public function useQueryCache($bool): self
...
+            'https://github.com/doctrine/orm/pull/9004',
+            '%s is deprecated and will be removed in Doctrine 3.0 without replacement.',
+            __METHOD__
+        );
 
-        $queryCache = $this->_em->getConfiguration()->getQueryCache();
+        $queryCache = $this->queryCache ?? $this->_em->getConfiguration()->getQueryCache();
 
         return $queryCache ? DoctrineProvider::wrap($queryCache) : null;
     }
```
greg0ire, line 534: Let's take this as an occasion to add tests for this method? The added lines result in a drop in coverage.

derrabus: #9007

derrabus:
I'm rebasing to make sure the new test is actually run.

greg0ire (approved):

VincentLanglet:
Hi @derrabus @greg0ire,

Priori to this PR I use
```
$query = $this->createQueryBuilder('foo')->(...)->getQuery();

if ($invalidateCache) {
     $query->getQueryCacheDriver()?->delete(self::CACHE_KEY);
}

return $query->enableResultCache(self::CACHE_LIFETIME, self::CACHE_KEY)->getOneOrNullResult();
```

What would be the new way to achieve the same behavior now `getQueryCacheDriver` is deprecated ?

I'd like to be able to delete a specific cache entry, and this PR doesn't introduce a `getQueryCache` method.

greg0ire:
`$configuration->getQueryCache()` would be the way I suppose.

VincentLanglet:
> `$configuration->getQueryCache()` would be the way I suppose.

Thanks, I can indeed solve it with
```
$queryCache = $this->getEntityManager()->getConfiguration()->getQueryCache();
$queryCache?->deleteItem(self::CACHE_KEY);
```

Issue #8963, opened 2021-08-28
PHP 8.1, SQLite: Decimal fields are hydrated as float instead of string


derrabus:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.x

#### Summary

Currently, Doctrine hydrates fields with the `decimal` type to a string. After upgrading to PHP 8.1, I receive a float instead. This change of behavior might break applications, for instance if a `string` return type declaration is used for a getter method of that field.

#### How to reproduce

Execute the test case `Doctrine\Tests\ORM\Functional\TypeTest::testDecimal` on PHP 8.1.

derrabus:
btw: In order to run the tests with PHP 8.1, I had to switch some dependencies to dev versions:

```JSON
{
    "doctrine/dbal": "~2.13.3@dev",
    "symfony/console": "~5.3.7@dev"
}
```

derrabus:
I've debugged the failing test and it appears that PDO has changed its behavior here.

<img width="1072" alt="Bildschirmfoto 2021-08-28 um 21 15 14" src="https://user-images.githubusercontent.com/1506493/131228516-2dd15066-c2f0-4146-8284-9cab3dfd24e9.png">

<img width="1072" alt="Bildschirmfoto 2021-08-28 um 21 14 46" src="https://user-images.githubusercontent.com/1506493/131228519-d0c0ce68-8021-4a0f-932f-a958339a603c.png">

derrabus:
Possibly realyted: php/php-src@438b025a28cda2935613af412fc13702883dd3a2

greg0ire:
:thinking: should we transfer this issue to the DBAL? That's where `$this->_stmt` lives, right?

derrabus:
Maybe. But honestly, I'm not sure PDO should behave that way. Casting a `DECIMAL` to a float might cause a loss of precision and rounding errors. Shall we report this to PHP instead? Maybe @morozov can advise here?

morozov:
> Casting a `DECIMAL` to a float might cause a loss of precision and rounding errors. Shall we report this to PHP instead?

Reporting upstream sounds right. I only noticed the change in behavior regarding integers when working on https://github.com/doctrine/dbal/pull/4682. But even then, using a large enough unsigned integer (if SQLite supports them) could lead to a loss of precision.

[...]

PR #9006, opened 2021-09-12
Add PHP 8.1 to CI

derrabus:
TODO:

- [x] #8963
- [x] `Doctrine\Tests\ORM\Functional\Ticket\GH7941Test::typesShouldBeConvertedForDQLFunctions`

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -19,6 +19,7 @@ jobs:
           - "7.3"
           - "7.4"
           - "8.0"
+          - "8.1"
 
     steps:
       - name: "Checkout"
@@ -33,6 +34,11 @@ jobs:
           extensions: "pdo, pdo_sqlite"
           coverage: "pcov"
           ini-values: "zend.assertions=1"
+
+      # To be removed as soon as https://github.com/doctrine/dbal/pull/4818 is released.
+      - name: "Switch to dev dependencies"
+        if: "${{ matrix.php-version == '8.1' }}"
+        run: "composer config minimum-stability dev"
 
       - name: "Install dependencies with Composer"
         uses: "ramsey/composer-install@v1"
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -34,6 +35,10 @@ jobs:
           coverage: "pcov"
           ini-values: "zend.assertions=1"
 
+      - name: "Switch to dev dependencies"
+        if: "${{ matrix.php-version == '8.1' }}"
+        run: "composer config minimum-stability dev"
+
```
derrabus, line 42: This can be reverted as soon as we release new DBAL versions with the fix for DECIMAL on SQLite.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7941Test.php
```diff
@@ -68,7 +70,7 @@ public function typesShouldBeConvertedForDQLFunctions(): void
         foreach ($query->getResult() as $i => $item) {
             $product = self::PRODUCTS[$i];
 
-            self::assertSame(ltrim($product['price'], '-'), $item['absolute']);
+            self::assertEquals(ltrim($product['price'], '-'), $item['absolute']);
```
SenseException, line 73: What was the reason to lower strictness for this assert?

derrabus: The type returned by the database driver is beyond our control. In PHP 8.1, various PDO drivers have become more type-aware and return integers and floats where we've received a string previously.

.github/workflows/continuous-integration.yml
greg0ire: Let's maybe add a link to the relevant DBAL issue and/or PRs then?

derrabus: Done.

greg0ire (approved):

PR #9009, opened 2021-09-13
Document fluent interfaces

derrabus:
I'd like to streamline how fluent interfaces are documented in the doc blocks. The annotation `@return $this` should be understood by Psalm, PHPStan and PhpStorm and is imho the most accurate way to represent such an interface.


greg0ire (approved):
Did you check if maybe this reduced the Psalm baseline? Phpstan would tell if it was the case for its baseline.

derrabus:
No, it didn't change the baseline.

PR #9019, opened 2021-09-16
validate schema command: allow to debug missing schema updates list

COil:
Hi,

This PR allows seeing why you get the message "The database schema is not in sync with the current mapping file" by displaying all the missing SQL so the database schema can be up-to-date. This debug is only activated when at least the verbose mode of the console is activated.

See you. COil :)

Questions: 
* Do I need to cache the update list, so there is only one call to `getUpdateSchemaList`? 
* Is it OK the tests the getVerbosity level? Or should it be displayed in every case (like for the mapping)?
* Need tests? There is currently no test for the `ValidateSchemaCommand` if I am right (except that is callable with `orm:validate-schema`)

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
@@ -10,6 +10,7 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
+use function count;
 use function sprintf;
 
 /**
@@ -76,6 +77,13 @@ class ValidateSchemaCommand extends AbstractEntityManagerCommand
             $ui->text('<comment>[SKIPPED] The database was not checked for synchronicity.</comment>');
         } elseif (! $validator->schemaInSyncWithMetadata()) {
             $ui->error('The database schema is not in sync with the current mapping file.');
+
+            if ($output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE) {
+                $sqls = $validator->getUpdateSchemaList();
+                $ui->comment(sprintf('<info>%d</info> schema diff(s) detected:', count($sqls)));
+                $ui->listing($sqls);
+            }
+
             $exit += 2;
         } else {
             $ui->success('The database schema is in sync with the mapping files.');
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -262,10 +262,20 @@ class SchemaValidator
      */
     public function schemaInSyncWithMetadata()
     {
+        return count($this->getUpdateSchemaList()) === 0;
+    }
+
+    /**
+     * Returns the list of missing Database Schema updates.
+     *
+     * @return array<string>
+     */
+    public function getUpdateSchemaList(): array
+    {
         $schemaTool = new SchemaTool($this->em);
 
         $allMetadata = $this->em->getMetadataFactory()->getAllMetadata();
 
-        return count($schemaTool->getUpdateSchemaSql($allMetadata, true)) === 0;
+        return $schemaTool->getUpdateSchemaSql($allMetadata, true);
     }
 }
--- a/docs/en/reference/tools.rst
+++ b/docs/en/reference/tools.rst
@@ -393,6 +393,11 @@ You can either use the Doctrine Command Line Tool:
 
     doctrine orm:validate-schema
 
+If the validation fails, you can change the verbosity level to
+check the detected errors:
+
+    doctrine orm:validate-schema -v
+
 Or you can trigger the validation manually:
 
 .. code-block:: php
--- a/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/UpdateCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/UpdateCommand.php
@@ -90,10 +90,7 @@ EOT
         if ($dumpSql) {
             $ui->text('The following SQL statements will be executed:');
             $ui->newLine();
-
-            foreach ($sqls as $sql) {
-                $ui->text(sprintf('    %s;', $sql));
-            }
+            $ui->listing($sqls);
         }
 
         if ($force) {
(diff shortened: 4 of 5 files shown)


derrabus:
Thank you for your pull request. I wonder, isn't the SQL that would be dumped here be the same as the `SchemaTool\UpdateCommand` would dump?

COil:
> Thank you for your pull request. I wonder, isn't the SQL that would be dumped here be the same as the `SchemaTool\UpdateCommand` would dump?

Hello Derrabus, yes of course, but you have to run a second command (yes, I am lazy 😁). I liked the idea of displaying it directly. Moreover, for the mapping, the errors are displayed.

derrabus:
I see. Nevertheless, we would need tests that cover your change. Also, since this is a new feature, the 2.10.x branch would be the appropriate target.

COil:
Hello, I can redo the PR to be on the 2.10.x. branch. Can you give some hint about the tests that should be done? Thanks.

greg0ire:
>  I can redo the PR to be on the 2.10.x. branch

No need! I re-targeted and rebased!

lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
```diff
@@ -76,6 +77,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+                $rows = [];
+                foreach ($updates as $idx => $sql) {
+                    $rows[] = [(int) $idx + 1, $sql];
+                }
+
+                $ui->comment(sprintf('<info>%d</info> schema diff(s) detected:', count($updates)));
+                $ui->table(['n°', 'SQL'], $rows);
+            }
+
```
greg0ire, line 86: > Can you give some hint about the tests that should be done?

I'd say here, let's not write a fragile test that would break every time Symfony changes the implementation of `$ui->table()`, but let's instead assert that the output contains the level of detail you expect (for instance, some table names are expected to be mentioned here, right?)

lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
```diff
@@ -76,6 +77,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $ui->text('<comment>[SKIPPED] The database was not checked for synchronicity.</comment>');
         } elseif (! $validator->schemaInSyncWithMetadata()) {
             $ui->error('The database schema is not in sync with the current mapping file.');
+
+            if ($output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE) {
+                $updates = $validator->getUpdateSchemaList();
+
+                $rows = [];
+                foreach ($updates as $idx => $sql) {
+                    $rows[] = [(int) $idx + 1, $sql];
```
greg0ire: In what case is `$idx` not an integer, and what will happen if it isn't? For instance if it's `'a'`, `'b'`, `'c'`, what will happen?

COil: Hello gregoire, I was thinking, the update command uses: 

            foreach ($sqls as $sql) {
                $ui->text(sprintf('    %s;', $sql));
            }

So, to be consistent we should do the same here, no ? Moreover, why not using `$ui->listing($sqls)`? Which has almost the same output and avoid doing a loop.

greg0ire: Maybe it did not exist when that piece of code was written? Having a consistent output would be great, yeah.

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -261,11 +261,21 @@ public function validateClass(ClassMetadataInfo $class)
...
+    {
+        return count($this->getUpdateSchemaList()) === 0;
+    }
+
+    /**
+     * Returns the list of missing Database Schema updates.
+     *
+     * @return string[]
+     */
+    public function getUpdateSchemaList()
```
greg0ire: Please add a return type here.

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -261,11 +261,21 @@ public function validateClass(ClassMetadataInfo $class)
...
      */
     public function schemaInSyncWithMetadata()
+    {
+        return count($this->getUpdateSchemaList()) === 0;
+    }
+
+    /**
+     * Returns the list of missing Database Schema updates.
+     *
+     * @return string[]
```
greg0ire: You will probably be interested in the outcome of https://github.com/doctrine/orm/issues/9025

lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
COil: Done.

lib/Doctrine/ORM/Tools/SchemaValidator.php
COil: Done.

lib/Doctrine/ORM/Tools/SchemaValidator.php
COil: `array<string>` then?

lib/Doctrine/ORM/Tools/Console/Command/ValidateSchemaCommand.php
COil: Hi @greg0ire, I pushed the new tests. It is OK for you?

greg0ire:
The tests fail with Postgresql, Mysql, MariaDB. Why?

COil:
> The tests fail with Postgresql, Mysql, MariaDB. Why?

The schema is in synch for Postgresql, Mysql, MariaDB so there is no error message. I guess I must force schema to be invalid.

COil:
I skipped the test for the other platforms, as we only want to test the error output, it's not related to the db platform. Maybe it can be cleaner? let me know.

greg0ire (dismissed):
The platform is not intervening at any point in the new code you introduced, so IMO it's good as is.

COil:
Hi @derrabus, do you think there is still work to do on this PR? 🙂

derrabus:
Looks good to me, I guess. Let's push it through the CI again.

derrabus (approved):

derrabus:
Thank you!

PR #9028, opened 2021-09-21
 Don't presume one-to-one lookup returned an entity 

### LINKED ISSUE #9027: PHP error thrown if a one-to-one relationship cannot be retrieved
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | Sorry, I don't know what this means
| Version     | 2.9.3

#### Summary

PHP error thrown if a one-to-one relationship cannot be retrieved.

I can't profess to understand this too much, but it has come up in live system due to a combination of Doctrine, JMSSerializer and the SoftDeleteable filter from doctrine-extensions. I've managed to narrow it down to the following:

#### Current behavior

When `UnitOfWork->createEntity` is called, it fetches related records to make them part of the object returned. For bidirectional one-to-one relationships, in certain circumstances, it uses `EntityManager->find` to do this, then calls `spl_object_hash()` and `ReflectionProperty->setValue()` with the result.

If `find` doesn't find an object then it will return `null`, which is an invalid argument to both of those functions. Hence one of the following errors occurs:

```
spl_object_hash(): Argument #1 ($object) must be of type object, null given
```

```
Exception: [ReflectionProperty::setValue(): Argument #1 ($objectOrValue) must be of type object, null given]
```

(The first error is what happens when I recreate locally, the second is what we saw on our live system. This has only started happening since upgrading to PHP 8.)

I believe the result of `Entitymanager->find` needs to be checked to ensure it is an object.

#### How to reproduce

See #9028 for a failing test case and suggested fix.

```php
/** @Entity */
class Customer
{
    /**
     * One Customer has One Cart.
     * @OneToOne(targetEntity="Cart", mappedBy="customer")
     */
    private $cart;
}

/** @Entity */
class Cart
{
    /**
     * One Cart has One Customer.
     * @OneToOne(targetEntity="Customer", inversedBy="cart")
     * @JoinColumn(name="customer_id", referencedColumnName="id")
     */
    private $customer;
}

$uow = new UnitOfWork($this->_em);
$hints = ['fetchMode' => [Cart::class => ['customer' => ClassMetadata::FETCH_EAGER]]];
$cart = $uow->createEntity(Cart::class, ['id' => 1, 'customer' => 24252], $hints);

$this->assertEquals(null, $cart->customer);
```

This throws:
> spl_object_hash(): Argument #1 ($object) must be of type object, null given

#### Expected behavior

The error shouldn't throw and `$cart->customer` should be null.

gregtyler:
If `$this->em->find()` returns null, don't treat it like an object. Instead, just set the field to null and back out of the switch statement.

Includes a was-failing-but-now-passing test case.

Fixes #9027

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
@@ -7,7 +7,6 @@ namespace Doctrine\Tests\ORM\Functional\Ticket;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\UnitOfWork;
 use Doctrine\Tests\OrmFunctionalTestCase;
-use Exception;
 
 class GH9027Test extends OrmFunctionalTestCase
 {
@@ -15,15 +14,20 @@ class GH9027Test extends OrmFunctionalTestCase
     {
         parent::setUp();
 
-        try {
-            $this->_schemaTool->createSchema(
-                [
-                    $this->_em->getClassMetadata(GH9027Customer::class),
-                    $this->_em->getClassMetadata(GH9027Cart::class),
-                ]
-            );
-        } catch (Exception $e) {
-        }
+        $this->_schemaTool->createSchema([
+            $this->_em->getClassMetadata(GH9027Cart::class),
+            $this->_em->getClassMetadata(GH9027Customer::class),
+        ]);
+    }
+
+    protected function tearDown(): void
+    {
+        $this->_schemaTool->dropSchema([
+            $this->_em->getClassMetadata(GH9027Cart::class),
+            $this->_em->getClassMetadata(GH9027Customer::class),
+        ]);
+
+        parent::tearDown();
     }
 
     /**
@@ -31,9 +35,14 @@ class GH9027Test extends OrmFunctionalTestCase
      */
     public function testUnitOfWorkHandlesNullRelations(): void
     {
-        $uow = new UnitOfWork($this->_em);
+        $uow   = new UnitOfWork($this->_em);
         $hints = ['fetchMode' => [GH9027Cart::class => ['customer' => ClassMetadata::FETCH_EAGER]]];
-        $cart = $uow->createEntity(GH9027Cart::class, ['id' => 1, 'customer' => 24252], $hints);
+
+        $cart = $uow->createEntity(
+            GH9027Cart::class,
+            ['id' => 1, 'customer' => 24252],
+            $hints
+        );
 
         $this->assertEquals(null, $cart->customer);
     }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
@@ -1,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\UnitOfWork;
+use Doctrine\Tests\OrmFunctionalTestCase;
+use Exception;
+
+class GH9027Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        try {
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
```diff
@@ -0,0 +1,77 @@
...
+
+        try {
+            $this->_schemaTool->createSchema(
+                [
+                    $this->_em->getClassMetadata(GH9027Customer::class),
+                    $this->_em->getClassMetadata(GH9027Cart::class),
+                ]
+            );
+        } catch (Exception $e) {
+        }
```
greg0ire: Can you please use something similar to what was done in https://github.com/doctrine/orm/pull/8962 ?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * @group GH-9027
+     */
+    public function testUnitOfWorkHandlesNullRelations(): void
+    {
+        $uow = new UnitOfWork($this->_em);
+        $hints = ['fetchMode' => [GH9027Cart::class => ['customer' => ClassMetadata::FETCH_EAGER]]];
+        $cart = $uow->createEntity(GH9027Cart::class, ['id' => 1, 'customer' => 24252], $hints);
```
greg0ire:
```suggestion
        $cart = $uow->createEntity(
                GH9027Cart::class,
                ['id' => 1, 'customer' => 24252],
                ['fetchMode' => [GH9027Cart::class => [
                        'customer' => ClassMetadata::FETCH_EAGER
                ]]];
        );
```

greg0ire:
`vendor/bin/phpcbf` can help with the cs issues

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
gregtyler: Done, sorry for using an old test as a template 🤦🏽

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
gregtyler: I would've preferred this too, but `$hints` needs to be passed by reference, not value.

gregtyler:
@greg0ire Thanks, I've fixed the issues I could :) Let me know if there's anything else I need to do!

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
greg0ire: No worries, really, I would have done the same :)

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9027Test.php
greg0ire: O_o OK

greg0ire (approved):

bendavies:
well this is due to
https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/SoftDeleteable/Filter/SoftDeleteableFilter.php filtering out the select query on the FK row.

and you are working around https://github.com/doctrine-extensions/DoctrineExtensions/pull/745

if the FK exists in your row, we know an fk row exists so it seems fine for the ORM to assume the data will never be null. 
i'm unsure if this fix is correct because to me, it's working around a design problem with the soft delete behavior.

gregtyler:
@bendavies Whilst the problem comes from SoftDeleteable in my instance, I figure we're _really_ working around a problem with Doctrine ORM Filters in general.

If I hydrate an object through `createEntity` and I have a filter which excludes an associated entity, Doctrine has to either:

1. Apply the filter and not add the associated entity
2. Ignore the filter and add the associated entity anyway
3. Throw an exception

As I understand it, Doctrine ORM's current behaviour is the first option, but it fails in PHP8 because of then mishandling the outcome if the entity is filtered out. I don't know if that's the *best* behaviour (I agree it's a silly situation to be in), but with this patch I'm just trying to keep it consistent across PHP versions.

SenseException (approved):

PR #9035, opened 2021-09-23
Fix PersistentCollection::get() returning entity not in the collection

### LINKED ISSUE #8607: Extra lazy PersistentCollection::get() with indexBy doesn't check mappedBy value
I want to make use of an `EXTRA_LAZY` one-to-many association and the fact that `Collection#get($key)` does not trigger a full load of the collection. Unfortunately, when `get()` is called accordingly, the association itself is not checked. Instead, only the key is used for querying the child entity in [PersistentCollection.php#L457](https://github.com/doctrine/orm/blob/fa588af3b1cb0edec22cae6526d5e20cbd658a8f/lib/Doctrine/ORM/PersistentCollection.php#L457).

I don't think that's intended - or is this meant by the following part in the [docs](https://github.com/doctrine/orm/blob/fa588af3b1cb0edec22cae6526d5e20cbd658a8f/docs/en/tutorials/working-with-indexed-associations.rst)?:
> _Fields that are used for the index by feature **HAVE** to be unique in the database. The behavior for multiple entities with the same index-by field value is undefined._

In that case, I'd suggest to make this more clear.

_(And one more note: The [extra lazy docs](https://github.com/doctrine/orm/blob/fa588af3b1cb0edec22cae6526d5e20cbd658a8f/docs/en/tutorials/extra-lazy-associations.rst) do not state that `indexBy` is necessary for `get()` to not trigger a full load.)_

---

Let me try to give a short example:
```php
class ParentEntity
{
    /**
     * @ORM\Column(name="id", type="integer", options={"unsigned":true})
     * @ORM\Id
     */
    private $id;

    /**
     * @ORM\OneToMany(targetEntity="AcmeBundle\Entity\ChildEntity", mappedBy="parent", indexBy="id", cascade={"persist"}, fetch="EXTRA_LAZY")
     */
    private $childs;

    public function getId(): int
    {
        return $this->id;
    }

    public function getChild(int $id): ?ChildEntity
    {
        return $this->childs->get($id);
    }
}
```
```php
class ChildEntity
{
    /**
     * @ORM\Column(name="id", type="integer", options={"unsigned":true})
     * @ORM\Id
     */
    private $id;

    /**
     * @ORM\ManyToOne(targetEntity="AcmeBundle\Entity\ParentEntity", inversedBy="childs")
     * @ORM\JoinColumns({
     *   @ORM\JoinColumn(name="parent_id", referencedColumnName="id", nullable=false, onDelete="CASCADE")
     * })
     */
    private $parent;

    public function getId(): int
    {
        return $this->id;
    }

    public function getParentId(): int
    {
        return $this->parent->getId();
    }
}
```

Now, calling `$parent->getChild($id)` with an id of a child that does *not* belong to this parent will still return the child with the given id. Thus, I need to check the following:
```php
    public function getChild(int $id): ?ChildEntity
    {
        $child = $this->childs->get($id);

        return null !== $child && $child->getParentId() == $this->id ? $child : null;
    }
```

sztyup:
Fixes https://github.com/doctrine/orm/issues/8607

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8607Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8607Test.php
@@ -1,0 +1,120 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\ManyToOne;
+use Doctrine\ORM\Mapping\OneToMany;
+use Doctrine\Tests\OrmFunctionalTestCase;
+use Exception;
+
+class GH8607Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        try {
+            $this->_schemaTool->createSchema([
+                    $this->_em->getClassMetadata(GH8607ParentEntity::class),
+                    $this->_em->getClassMetadata(GH8607ChildEntity::class),
+                ]);
+        } catch (Exception $e) {
+            //
+        }
+    }
+
+    public function testInvalidCollectionReturn(): void
+    {
+        $parent1 = new GH8607ParentEntity();
+        $parent2 = new GH8607ParentEntity();
+
+        $child = new GH8607ChildEntity();
+        $parent2->getChildren()->add($child);
+
+        $this->_em->persist($parent1);
+        $this->_em->persist($parent2);
+        $this->_em->persist($child);
+
+        $this->_em->flush();
+
+        $parentId = $parent1->getId();
+        $childId = $child->getId();
+
+        $this->_em->clear();
+
+        $parent = $this->_em->find(GH8607ParentEntity::class, $parentId);
+
+        $return = $parent->getChildren()->get($childId);
+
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH8607Test.php
```diff
@@ -0,0 +1,126 @@
...
+        parent::setUp();
+
+        try {
+            $this->_schemaTool->createSchema([
+                $this->_em->getClassMetadata(GH8607ParentEntity::class),
+                $this->_em->getClassMetadata(GH8607ChildEntity::class),
+            ]);
+        } catch (Exception $e) {
+            // skip errors
+        }
```
greg0ire: Please apply a change similar to what I did in https://github.com/doctrine/orm/pull/8962 :pray:

sztyup: done

beberlei:
Does this really require a new testcase with new entities? There is already a bunch of tests related to persistent collections and indexBy feature in general, from my pov it would make sense to just put it there.

derrabus:
Can you please rebase your PR onto 2.10.x? The 2.9.x branch has been closed for bugfixes.

Spilky:
@greg0ire would it be possible to merge this bugfix? It would be really helpful. I spent some time with debbuging because of this and fix seems easy.

greg0ire:
@Spilky it targets 2.10.x, which is not maintained, so no. Feel free to open a new PR based on it though. If you do, target 2.14.x

Spilky:
@greg0ire new PR opened here: https://github.com/doctrine/orm/pull/10439

derrabus:
Closing in favor of #10439.

PR #9043, opened 2021-09-29
Update documentation regarding caching

### LINKED ISSUE #9003: Documentation for PSR-6 caches
In 2.10, we will be able to configure all ORM caches through PSR-6 (#9002). We should update the documentation accordingly.

derrabus:
Fixes #9003

### DIFF
--- a/docs/en/reference/caching.rst
+++ b/docs/en/reference/caching.rst
@@ -1,266 +1,14 @@
 Caching
 =======
 
-Doctrine provides cache drivers in the ``doctrine/cache`` package for some
-of the most popular caching implementations such as APCu, Memcache
-and Xcache. We also provide an ``ArrayCache`` driver which stores
-the data in a PHP array. Obviously, when using ``ArrayCache``, the 
-cache does not persist between requests, but this is useful for 
-testing in a development environment.
+The Doctrine ORM package can leverage cache adapters implementing the PSR-6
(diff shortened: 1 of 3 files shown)


docs/en/reference/caching.rst
```diff
@@ -313,10 +61,12 @@ result cache driver.
...
 
     <?php
-    $cacheDriver = new \Doctrine\Common\Cache\PhpFileCache(
+    $cache = new \Symfony\Component\Cache\Adapter\PhpFilesAdapter(
+        'doctrine_results',
+        0,
         '/path/to/writable/directory'
     );
-    $query->setResultCacheDriver($cacheDriver);
+    $query->setResultCache($cache);
```
SenseException, line 69: Before it was `setResultCacheImpl()` and `setResultCacheDriver()`. Now both are `setResultCache` and duplicates.

SenseException (changes requested):

docs/en/reference/caching.rst
derrabus: The first example sets the result cache on the global configuration while the second example sets it on the query object.

Yes, the method names have been streamlined, but the two code snippets still show different approaches.

SenseException: I missed that $config line. My bad.

SenseException (approved):

PR #9072, opened 2021-10-03
Deprecate `AbstractHydrator::hydrateRow()`

derrabus:
While removing the `iterate()` methods, I noticed that the public method `AbstractHydrator::hydrateRow()` has become obsolete because we don't call it anymore. I'd like to remove it as well.


greg0ire:
Converting to draft for obvious reasons :stuck_out_tongue:

greg0ire (approved):

PR #9074, opened 2021-10-03
Deprecate `ensureProductionSettings()`

derrabus:
Doctrine does not provide its own cache implementation anymore and relies on the PSR-6 standard instead. As a consequence, we cannot determine anymore whether a given cache adapter is suitable for a production environment.

The corresponding functionality has always been error-prone. For instance if cache adapters are decorated, the `orm:ensure-production-settings` command would not complain anymore. By adopting PSR-6, this gets even harder because we would need to have internal knowledge of any PSR-6 compatible library a user might potentially use.

Moreover, I really don't think that we should offer this kind of inspection and therefore propose to remove it.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -5,6 +5,16 @@
 Following the deprecation of the method `AbstractHydrator::iterate()`, the
 method `hydrateRow()` has been deprecated as well.
 
+## Deprecate cache settings inspection
+
+Doctrine does not provide its own cache implementation anymore and relies on
+the PSR-6 standard instead. As a consequence, we cannot determine anymore
+whether a given cache adapter is suitable for a production environment.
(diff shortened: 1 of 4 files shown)


greg0ire (approved):

norkunas:
This also was checking for:
```
if ($this->getAutoGenerateProxyClasses()) {
    throw ProxyClassesAlwaysRegenerating::create();
}
```
and not only the cache :)

derrabus:
I see. Is this enough to keep that method, if we removed all cache checks? Do you use that method or the command?

norkunas:
Was in backlog sometime ago to use it :see_no_evil: not sure if it's needed because doctrine-bundle has a recipe which configures this for me :) Just wanted to point it out :)

PR #9079, opened 2021-10-04
Add integer cast in setMaxResults methods of Query and QueryBuilder

### LINKED ISSUE #9086: Type error in Query.php
After upgrading to version 2.10.0 we are facing with the following error:

TypeError in file: Query.php on line 692
Message: Return value of Doctrine\ORM\Query::getMaxResults() must be of the type int or null, string returned

mitelg:
Hello there :wave: 

This is my first PR here :grin: Hope, I have done it right.

With version 2.9.6 a native return type was added to the `Doctrine\ORM\Query::getMaxResults()` method.
To prevent type errors in userland due to wrong usage, an integer cast while setting the max result property was added.

Alternatively a native type to the parameter could be set. But I don't know, if this would be breaking. 
Psalm is failing due to a redundant cast, which is obvious :see_no_evil: 

Nevertheless I wanted to hear you opinions, whether this could be merged or if I need to fix our 50+ wrong usages :grimacing: 

Best regards :slightly_smiling_face:

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -675,6 +675,10 @@ final class Query extends AbstractQuery
      */
     public function setMaxResults($maxResults): self
     {
+        if ($maxResults !== null) {
+            $maxResults = (int) $maxResults;
+        }
+
         $this->maxResults = $maxResults;
         $this->_state     = self::STATE_DIRTY;
 
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -652,6 +652,10 @@ class QueryBuilder
      */
     public function setMaxResults($maxResults)
     {
+        if ($maxResults !== null) {
+            $maxResults = (int) $maxResults;
+        }
+
         $this->_maxResults = $maxResults;
 
         return $this;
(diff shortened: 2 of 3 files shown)


derrabus:
> Nevertheless I wanted to hear you opinions, whether this could be merged or if I need to fix our 50+ wrong usages 😬

You should do so either way. Even if we merge your PR, a future version will certainly add parameter types and break your code again.

derrabus:
I'm okay with the change, once Psalm does not complain anymore. 🙂

mitelg:
thanks for your feedback :+1: 

I am totally aware that we need to change our code sooner or later :see_no_evil: 
but due to time limitations I would like to postpone that and leave it for future-Michael :grin: 

I will add a change to the psalm config in a minute

SenseException (approved):

derrabus (approved):

mitelg:
thanks for merging @derrabus 🎉

YetiCGN:
@derrabus I think it would help a lot if this got released as 2.10.1 as soon as possible.

derrabus:
Please read your Slack messages, Jörn. 😉

derrabus:
https://github.com/doctrine/orm/releases/tag/2.10.1

PR #9085, opened 2021-10-04
Refactor ORMException into an interface

greg0ire:
Follow up of https://github.com/doctrine/orm/pull/8692


lib/Doctrine/ORM/Exception/ORMException.php
```diff
@@ -4,11 +4,11 @@
...
 namespace Doctrine\ORM\Exception;
 
-use Doctrine\ORM\ORMException as BaseORMException;
+use Throwable;
 
 /**
  * Should become an interface in 3.0
  */
-class ORMException extends BaseORMException
+interface ORMException extends Throwable
```
derrabus, line 12: Extending `Throwable` can have ugly side-effects on PHP 7.3 and below. I'd suggest that we bump to 7.4 at least, so we don't have to deal with them.

https://3v4l.org/rmkAc

derrabus (approved):

lib/Doctrine/ORM/Exception/ORMException.php
greg0ire: Fully agree, let's do that.

PR #9098, opened 2021-10-06
Honor INDEX BY construct in Query::toIterable

ajgarlag:
Fix the problem explained in #8927 as a bug fix.

Alternative implementation as a feature in #9099.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -142,7 +142,7 @@ abstract class AbstractHydrator
      * @param Result|ResultStatement $stmt
      * @psalm-param array<string, mixed> $hints
      *
-     * @return Generator<int, mixed>
+     * @return Generator<array-key, mixed>
      *
      * @final
      */
@@ -194,9 +194,12 @@ abstract class AbstractHydrator
             $this->hydrateRowData($row, $result);
 
             $this->cleanupAfterRowIteration();
-
             if (count($result) === 1) {
-                yield end($result);
+                if (count($resultSetMapping->indexByMap) === 0) {
+                    yield end($result);
+                } else {
+                    yield from $result;
+                }
             } else {
                 yield $result;
             }
--- a/tests/Doctrine/Tests/ORM/Functional/QueryIterableTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/QueryIterableTest.php
@@ -6,6 +6,7 @@ namespace Doctrine\Tests\ORM\Functional;
 
 use Doctrine\Tests\IterableTester;
 use Doctrine\Tests\Models\CMS\CmsAddress;
+use Doctrine\Tests\Models\CMS\CmsArticle;
 use Doctrine\Tests\Models\CMS\CmsUser;
 use Doctrine\Tests\OrmFunctionalTestCase;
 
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -196,7 +197,7 @@ public function toIterable($stmt, ResultSetMapping $resultSetMapping, array $hin
             $this->cleanupAfterRowIteration();
 
             if (count($result) === 1) {
-                yield end($result);
+                yield array_keys($result)[count($result) - 1] => end($result);
```
greg0ire: Can you explain why this works? Is the key for `INDEX BY` always the last one for some reason?

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -196,7 +197,7 @@ public function toIterable($stmt, ResultSetMapping $resultSetMapping, array $hin
...
             } else {
                 yield $result;
```
greg0ire, line 204: What is the purpose of this line and why doesn't it need to be changed in our case (sorry, I'm unfamiliar with that area of the code)?

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
ajgarlag: It was a polyfill for `array_last_key`. We have to return the last key, because current code is returning last position of the `$result` array.

But after reading the code again, if `count($result) === 1` it could be rewritten as a simple `yield from $result`.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
ajgarlag: It was an error. I've added a test to check the key when multiple entities are selected, and I've modified this line to pass the test.

ajgarlag:
@greg0ire After reviewing the code again, I think the problem exposed in #8927 is finally a bug.

I've found that the tests use `IterableTester` to ensure that the values returned by `Query::getResult` and `Query::toIterable` are equivalent. So I've written a small test that breaks it with the addition of the INDEX BY construct to a simple query.

https://github.com/doctrine/orm/blob/a7b4d9a1a955156d34a2e06bddaeb036d9f370ff/tests/Doctrine/Tests/ORM/Functional/QueryIterableTest.php#L82

After fixing it, I've found another bug: when selecting multiple entities, the results from `Query::getResult` and `Query::toIterable` are different, so I'll open another issue to discuss this problem that, once resolved, should be tested combining INDEX BY and multiple entities.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
greg0ire: The line appears untouched in the diff though :thinking:

greg0ire:

How to do that?

1. `git rebase -i origin/2.10.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

ajgarlag:
@greg0ire Squashed!!

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @ajgarlag !

PR #9099, opened 2021-10-06
Add new Query::toIndexedIterable to honor INDEX BY construct

ajgarlag:
Fix the problem explained in #8927 as a feature

Alternative implementation as a bugfix in #9098


ajgarlag:
After reviewing the code, I've found the `IterableTester` class that is used to ensure that results from `Query::getResult` and `Query::toIterable` must be equivalent. This does not occurs when using `INDEX BY` construct, so it's finally a bug.

Closing in favor of #9098.

PR #9100, opened 2021-10-06
ScalarColumnHydrator

calvera:
fix fail for 
`            ->select('o.id as _id, o as HIDDEN o_')->distinct(true);`

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
@@ -22,7 +22,7 @@ final class ScalarColumnHydrator extends AbstractHydrator
      */
     protected function hydrateAllData(): array
     {
-        if (count($this->resultSetMapping()->fieldMappings) > 1) {
+        if (count($this->resultSetMapping()->scalarMappings) > 1) {
             throw MultipleSelectorsFoundException::create($this->resultSetMapping()->fieldMappings);
         }
 


greg0ire:
@bhushan please review

@calvera can you please add a test case for this?

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
```diff
@@ -22,7 +22,7 @@ final class ScalarColumnHydrator extends AbstractHydrator
      */
     protected function hydrateAllData(): array
     {
-        if (count($this->resultSetMapping()->fieldMappings) > 1) {
+        if (count($this->resultSetMapping()->scalarMappings) > 1) {
```
bhushan, line 25: @calvera could you please write failing test for it.. 

This check is meant for selector count..

bhushan (changes requested):

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
calvera: yes, i can, but i'm not sure if the fix is right because ScalarColumnHydratorTest fails

bhushan: I think this fix is wrong, so only I asked to write regression test to produce bug, so once you write test, I understand exact scenario and we will be able to fix together.

bhushan (changes requested):
please write regression test (reproduce bug using test).

lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
calvera: the problem is when i do this DQL: 
`SELECT o.id as _id, o as HIDDEN o_ FROM Entity o` and then getSingleColumnResult... the o_ is HIDDEN so i should be able to get only column `_id` as one-dimensional array

bhushan: I think its expected behaviour to throw exception, if you are not going to use multiple columns in result then why include them in query? 

to use `getSingleColumnResult` you have to select only one column

bhushan: @calvera How this query works ?  `SELECT o.id as _id, o as HIDDEN o_ FROM Entity o` table is hidden, what is the use case when we do such a thing?

cc @greg0ire

calvera: for example when you select id from table but you need to order result by `COLLATE($root.name, cs_CZ) ` so you do 
```
                $qb->addSelect("COLLATE($root.name, cs_CZ) AS HIDDEN $field");
                $qb->addOrderBy($field, $order);
```

calvera: so you do not need the column hydrated, but you have to include it in the DQL as hidden

bhushan: Okay so that selector check needs to be extended..

PROPOSAL:
instead of checking count of selectors we should check count of non hidden selectors and if its count is greater than one then throw exception

check ScalarColumnHydrationTest class and its last test to get the idea and do the changes!!

calvera: it seems to me that it is not that easy... it might be bug in SqlWalker.... it calls `$this->rsm->addFieldResult` for fields of hidden entity... see https://github.com/doctrine/orm/blob/5c12d36be310616f7b1e84085627ac3a88a6381d/lib/Doctrine/ORM/Query/SqlWalker.php#L1460 
if i change it to
```
                    if (!$hidden) {
                        $this->rsm->addFieldResult($dqlAlias, $columnAlias, $fieldName, $class->name);
                    }
```
it works correctly but i do not know if there are any consequences...

bhushan: Lets do small step at a time.. you write failing test apply fix what you think is good push the code, i will check if it needs improvements and we fix one thing at a time.. right now we are just discussing on the fly.. better you push some code so i can pull it and see if it works .. 

Every problem is fixable 😀

bhushan: @calvera Hey, I tried it with new project, it working without any changes. check below repo..

https://github.com/bhushan/doctrine-functionality-test/blob/9d6f537c46543cd42a912da1f72eb8565843c438/src/Repository/BlogPostRepository.php#L14


also what exact errors you are getting that also provide please, also can you create separate public repo reproducing bug, as seems like its not a problem

calvera:
closing, i will create another PR

PR #9104, opened 2021-10-06
Changed EntityRepository count method $criteria parameter to optional

soltmar:
Currently to get count of all items there is need to provide empty array to count() method as $criteria parameter is required.
I believe there shouldn't be a need to provide it if I want to count all Entities without any criteria.

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -227,7 +227,7 @@ class EntityRepository implements ObjectRepository, Selectable
      *
      * @todo Add this method to `ObjectRepository` interface in the next major release
      */
-    public function count(array $criteria)
+    public function count(array $criteria = [])
     {
         return $this->_em->getUnitOfWork()->getEntityPersister($this->_entityName)->count($criteria);
     }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,10 @@
 # Upgrade to 3.0
 
+## BC BREAK: EntityRepository::count() signature change
+
+The argument `$criteria` of `Doctrine\ORM\EntityRepository::count()` is now
+optional. Overrides in child classes should be made compatible.
+
 ## BC BREAK: changes in exception hierarchy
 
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -227,7 +227,7 @@ public function findOneBy(array $criteria, ?array $orderBy = null)
      *
      * @todo Add this method to `ObjectRepository` interface in the next major release
      */
-    public function count(array $criteria)
+    public function count(?array $criteria = [])
```
greg0ire: This is a BC break for extending classes: https://3v4l.org/p4Mrb

soltmar: You are actually right. Maybe this should be included in the next major release then (3.0 possibly ? )

greg0ire: Yes, this cannot be merged into 2.10.x

greg0ire: Another remark: The `?` allows to pass `null`, but to me, it does not make sense to do `$entityRepository->count(null)`.

soltmar: Yeah, this was corrected in the commit below.
Thanks

greg0ire:

How to do that?

1. `git rebase -i origin/3.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

bhushan (dismissed):

soltmar:
All done. Thanks

greg0ire (dismissed):

SenseException (dismissed):

greg0ire:
Sorry, I forgot to tell you, but BC-breaks should be documented in UPGRADE.md

greg0ire (approved):

derrabus (approved):

PR #9106, opened 2021-10-07
`Query::getResult` and `Query::toIterable` results differ when selecting multiple entities.

ajgarlag:
While working on #9098 I've found this bug so I open this PR with a test for the exposed bug.

For the test in the PR with 4 articles (a), for 1 author (u), `Query::getResult` returns 5 results because it only returns one entity per position, while `Query::toIterable` returns 4 results because it combines the selected entities.

  - `Query::getResult` output is `[a1, u, a2, a3, a4]`
  - `Query::toIterable` output is `[[a1, u], [a2, u], [a3, u], [a4, u]]`

I prefer the `Query::toIterable` output, so my questions are:
  - must we fix `Query::toIterable` to mimic the `Query::getResult` output?
  - if we fix it, must we add another iterable funcion to `Query` with the same output currently produced by `Query::toIterable`?
  - can we assume that output from `Query::toIterable` and `Query::getResult` can differ when selecting multiple entities?

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/QueryIterableTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/QueryIterableTest.php
@@ -7,6 +7,7 @@ namespace Doctrine\Tests\ORM\Functional;
 use Doctrine\Tests\IterableTester;
 use Doctrine\Tests\Models\CMS\CmsAddress;
 use Doctrine\Tests\Models\CMS\CmsArticle;
+use Doctrine\Tests\Models\CMS\CmsTag;
 use Doctrine\Tests\Models\CMS\CmsUser;
 use Doctrine\Tests\OrmFunctionalTestCase;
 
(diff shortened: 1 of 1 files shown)


bhushan:
@greg0ire ‘Query::toIterable’ and ‘Query::getResult’ are they expected to have different  outputs? Seems like yes .. but if No then it seems valid problem!!!

greg0ire:
I don't know, but I've found this note in the docs:

> Iterating results is not possible with queries that fetch-join a collection-valued association. The nature of such SQL result sets is not suitable for incremental hydration.

https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/batch-processing.html

This use case does not seem to be supported at all. Maybe we should throw in that sort of case.
The DQL query does not make a lot of sense to me either: I'd use `'select a, a.user from ' . CmsArticle::class'`

bhushan:
@ajgarlag hi, by checking docs, Query::toIterable can be used for bulk operations and Query::getResult is for getting hydrated data.. 

Also Query::toIterable doesn't support joins .. 

IMHO functionality is working as expected for both functions.

bhushan:
May be you can PR something which throws exceptions when such case occurs !! @ajgarlag 

cc @greg0ire

greg0ire:
I think it might make sense. It would be a BC-break, and for that reason the PR should target 3.0.x.
A PR deprecating this could be done against 2.11.x

ajgarlag:
> Also Query::toIterable doesn't support joins ..

`Query::toIterable` does not support fetch-join from a `...toMany` associated entity.

There is an exception that is already thrown when the DQL expression includes a fetch-join using a mapped `...toMany` association without adding the `DISTINCT` hint.

https://github.com/doctrine/orm/blob/416aa1d2d71c8d1796781c3f57b17ba26debde26/lib/Doctrine/ORM/Query/SqlWalker.php#L983-L987

The problem exposed it the tests occur when selecting multiple entities with an arbitrary JOIN, or even without any JOIN. I'll update the test case to select completely unrelated entities.

PR #9108, opened 2021-10-07
[Psalm] always true/false conditions

orklah:
I used this repo to check for false positives in the next Psalm release. I found some real issues that seemed easy to fix so I may as well make a PR out of it :)


greg0ire (approved):

greg0ire:
Thanks @orklah !

PR #9111, opened 2021-10-08
fix scalar column hydrator when using hidden entity

calvera:
as discussed in #9100 @bhushan

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -1457,7 +1457,9 @@ class SqlWalker implements TreeWalker
 
                     $this->scalarResultAliasMap[$resultAlias][] = $columnAlias;
 
-                    $this->rsm->addFieldResult($dqlAlias, $columnAlias, $fieldName, $class->name);
+                    if (! $hidden) {
+                        $this->rsm->addFieldResult($dqlAlias, $columnAlias, $fieldName, $class->name);
+                    }
                 }
 
                 // Add any additional fields of subclasses (excluding inherited fields)
@@ -1488,7 +1490,9 @@ class SqlWalker implements TreeWalker
 
                             $this->scalarResultAliasMap[$resultAlias][] = $columnAlias;
 
-                            $this->rsm->addFieldResult($dqlAlias, $columnAlias, $fieldName, $subClassName);
+                            if (! $hidden) {
+                                $this->rsm->addFieldResult($dqlAlias, $columnAlias, $fieldName, $subClassName);
+                            }
                         }
                     }
                 }
--- a/tests/Doctrine/Tests/ORM/Functional/QueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/QueryTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Tests\ORM\Functional;
 
 use Doctrine\Common\Collections\ArrayCollection;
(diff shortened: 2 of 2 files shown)


bhushan:
https://github.com/doctrine/orm/pull/9100#discussion_r724717097

calvera:
> [#9100 (comment)](https://github.com/doctrine/orm/pull/9100#discussion_r724717097)

in your test you hide scalar column, try hide whole `bp`
```
        return $this->createQueryBuilder('bp')
            ->select('bp.title')
            ->addSelect('bp as HIDDEN i')
            ->orderBy('bp.id', 'DESC')
            ->getQuery()
            ->getSingleColumnResult();
```
ie. `bp.id` vs `bp`

bhushan:
> > [#9100 (comment)](https://github.com/doctrine/orm/pull/9100#discussion_r724717097)
> 
> in your test you hide scalar column, try hide whole `bp`
> 
> ```
>         return $this->createQueryBuilder('bp')
>             ->select('bp.title')
>             ->addSelect('bp as HIDDEN i')
>             ->orderBy('bp.id', 'DESC')
>             ->getQuery()
>             ->getSingleColumnResult();
> ```
> 
> ie. `bp.id` vs `bp`

I dont think it creates any problem, why do you want to hide `bp` ? what is this scenario ?

calvera:
> > > [#9100 (comment)](https://github.com/doctrine/orm/pull/9100#discussion_r724717097)
> > 
> > 
> > in your test you hide scalar column, try hide whole `bp`
> > ```
> >         return $this->createQueryBuilder('bp')
> >             ->select('bp.title')
> >             ->addSelect('bp as HIDDEN i')
> >             ->orderBy('bp.id', 'DESC')
> >             ->getQuery()
> >             ->getSingleColumnResult();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ie. `bp.id` vs `bp`
> 
> I dont think it creates any problem, why do you want to hide `bp` ? what is this scenario ?

if it did not create any problem i would not need a fix :-)
i need them in SQL because postgres error `for SELECT DISTINCT, ORDER BY expressions must appear in select list`

bhushan:
@greg0ire @derrabus could you please help? I don't understand scenario and problem, why we need to hide `bp` (table reference) from query.

greg0ire:
> for SELECT DISTINCT, ORDER BY expressions must appear in select list

I don't understand either… I mean… there is no distinct here, is there?

@calvera , please amend your commit message to include a detailed, crystal clear explanation of why this is needed (`git commit --amend && git push --force`). In an ideal world, every commit should be fully understandable even 6 months later without additional context.

calvera:
the point is not in the DQL or if distinct is needed or if it is there... the point is: when you select whole entity as hidden, it is not hidden for the new scalar column hydrator... i do not know how to explain it more clearly. i thought that it is clear when you see the test cases i've added

greg0ire:
> when you select whole entity as hidden, it is not hidden for the new scalar column hydrator

What happens then? Are you getting a crash? If yes, can you provide a stack trace? Because from the tests you added, it's not clear what happens. In a more general way, if you write a commit message with "fix", it's nice to indicate what is broken. "Trust me, it's broken" is not enough, sorry.

> the point is not in the DQL or if distinct is needed or if it is there

Agreed, this is not required but still, it would have been nice to know why you _need_ to hide (and select) `bp` in its entirety and why we need to support that use case.

PR #5728, opened 2016-03-21
Readonly column property for generated/virtual columns

Hikariii:
MySQL and MariaDB both support virtual/computed/generated columns.
This is a column in a table that has its value automatically calculated using a deterministic expression, in particular from the values of other fields in the table.
These columns are queried like any other column, but its values cannot be inserted or updated.
To use these columns in entities and DQL, Doctrine must find a way to define them, but not include them in updates and inserts. This PR accomplishes just that.

More info on these columns:
- MariaDB: https://mariadb.com/kb/en/mariadb/virtual-computed-columns/
- Mysql: http://mysqlserverteam.com/generated-columns-in-mysql-5-7-5/

PR #9118, opened 2021-10-12
Generated/Virtual Columns: Insertable / Updateable

mehldau:
Defines whether a column is included in an SQL INSERT and/or UPDATE statement.

* `insertable: false` - do not include this column as part of INSERT statement for entities of the type.
* `updatable: false` - do not include this column as part of UPDATE statement
* `generated: "always"` - after inserting or updating this entity, always fetch an updated value from the database.
* `generated: "insert"` - after inserting this entity, fetch an updated value from the database.

### Map a column several times.

```php
#[Entity] 
class User
{
     #[ManyToOne(targetEntity: Country:class), JoinColumn(name: "country_code", referencedColumnName: "country_code")]
     public $country;
     
     #[Column(type: "string", name: "country_code", insertable: false, updatable: false)]
     public $countryCode;
}
```

### Columns with database updates

```php
#[Entity]
class Article
{
    #[Column(type: "datetime", 
        columnDefinition: "TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP", 
        insertable: false,
        updatable: false,
        generated: "ALWAYS")]
    public $created;
}
```

#5728


greg0ire:
Please rebase on 2.11.x and force push, new features should go on that branch now. Also, please add documentation about this (in `docs`)

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -627,6 +628,19 @@ protected function prepareUpdateData($entity)
                 $fieldMapping = $this->class->fieldMappings[$field];
                 $columnName   = $fieldMapping['columnName'];
 
+                $isInsert = ! empty($this->queuedInserts[spl_object_id($entity)]);
```
beberlei: You can move this to line 609 already.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -627,6 +628,19 @@ protected function prepareUpdateData($entity)
...
+                if (! $isInsert && ! $this->class->isUpdateable($field)) {
```
beberlei: You don't need this, you already have `$fieldMapping` in 628 and can just test for `$fieldMapping['updateable']`.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -627,6 +628,19 @@ protected function prepareUpdateData($entity)
...
+                    if ($change[0] !== $change[1]) {
+                        throw NonUpdateableField::byName($field);
+                    }
+
+                    continue;
+                }
+
+                if ($isInsert && ! $this->class->isInsertable($field)) {
```
beberlei: Same here with `$fieldMapping` already available.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1282,6 +1318,8 @@ public function getColumnName($fieldName)
      *      columnName?: string,
      *      inherited?: class-string,
      *      nullable?: bool,
+     *      insertable?: bool,
```
beberlei: It would be better if we invert these here to `notInsertable` and `notUpdateable`. Then if a user sets `insertable=true` don't set this field at all. Then the code in `BasicEntityPersister` must only test for `isset($fieldMapping['notInsertable`])`

This optimizes the cache size of a serialized ClassMetadata by not storing anything in the 99% default case where insertable=true or updatetable=true.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
beberlei: Even better we can move a boolean flag nto `prepareUpdateData` for `isInsert` and directly pass it from `insert` or `update`.

beberlei (changes requested):
This is already amazing work and i am happy this looks like it be implemented with just a few lines in `BasicEntityPersister`.

I made a few suggestions about optimizations and towards more ORM idomatic style.

One thing I checked is weather we should even allow non updateable / insertable columns to show up in a changeset. But I believe this would cause more harm to performance since there is no access to `$fieldMapping` in changeset computation.

One thing this PR does not address here is auto updated columns in the database. In my opinion if there is a column that is not insertable / updateable, that means we need to fetch them after an insert / update and set them on the entity. The updatable = false should not be used to help with immutability, it means the database updates this value itself (via trigger for example).

mehldau:
> This is already amazing work and i am happy this looks like it be implemented with just a few lines in `BasicEntityPersister`.

Thank you @beberlei 

> I made a few suggestions about optimizations and towards more ORM idomatic style.
> 

I addressed this and your proposal of inverting 'insertable' and 'updateable' reg. optimization of the ClassMetadata cache in the current PR.

> One thing this PR does not address here is auto updated columns in the database. In my opinion if there is a column that is not insertable / updateable, that means we need to fetch them after an insert / update and set them on the entity. The updatable = false should not be used to help with immutability, it means the database updates this value itself (via trigger for example).

You're absolutely right. Maybe you could lent me your expertise to find the most suitable hook point to implement this. Maybe remember the necessity of reloading the Entity within _BasicEntityPersister::prepareUpdateData()_ and use this information somewhere within the update / insert - processing? Or could I use the event system either way?

beberlei:
@mehldau the event system should only be used by outside integrations. We hardcode our logic into the persisters somewhere.

There is already precedent for this in the `isVersioned` property and its code in `assignDefaultVersionValue` to fetch the database assigned version id. We can generalize this code. 

a.) Introduce a new property `requiresFetchAfterUpdate` that defaults to `false`. if a versioned field is set or any `notUpdateable` / `notInsertable` property exists set this property true.
b.) Refactor `assignDefaultVersionValue` to work over all not insertable / updateable / versioned columns.

mehldau:
> There is already precedent for this in the `isVersioned` property and its code in `assignDefaultVersionValue` to fetch the database assigned version id. We can generalize this code.
> 
> a.) Introduce a new property `requiresFetchAfterUpdate` that defaults to `false`. if a versioned field is set or any `notUpdateable` / `notInsertable` property exists set this property true. b.) Refactor `assignDefaultVersionValue` to work over all not insertable / updateable / versioned columns.

@beberlei All right. I implemented the changes as suggested and updated the pull request.

Nevertheless, I'm wondering whether we have to reload 'non insertable' columns in the edge case of 'updating' an entity which only has 'non insertable' columns defined.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -653,6 +661,14 @@ class ClassMetadataInfo implements ClassMetadata
      */
     public $changeTrackingPolicy = self::CHANGETRACKING_DEFERRED_IMPLICIT;
 
+    /**
+     * READ-ONLY: A Flag indicating whether or not one or more columns of this class have to be reloaded
+     * after insert / update operations.
+     *
+     * @var bool
+     */
+    public $requiresFetchAfterUpdate = false;
```
beberlei: You need to add support for this new property in serialization code here, should be something like:

```php
if ($this->requiresFetchAtfterUpdate) {
    $serialized[] = 'requiresFetchAtfterUpdate';
}
```

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -299,50 +302,67 @@ public function executeInserts()
...
+        }
+
+        $tableName  = $this->quoteStrategy->getTableName($versionedClass, $this->platform);
+        $identifier = $this->quoteStrategy->getIdentifierColumnNames($versionedClass, $this->platform);
 
         // FIXME: Order with composite keys might not be correct
-        $sql = 'SELECT ' . $columnName
-             . ' FROM ' . $tableName
-             . ' WHERE ' . implode(' = ? AND ', $identifier) . ' = ?';
+        $sql = 'SELECT ' . implode(', ', $columnNames)
```
beberlei, line 344: You need to use `$column as $fieldName` here otherwise the following access of `$this->class->fieldMappings[$field]` based on `$values` will not work when column name != field name.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1442,6 +1475,10 @@ protected function getInsertColumnList()
             }
 
             if (! $this->class->isIdGeneratorIdentity() || $this->class->identifier[0] !== $name) {
+                if (! $this->class->isInsertable($name)) {
```
beberlei: Can you use `$this->class->fieldMappings[$name]['notInsertable']` here please? property access is so much faster than dynamic methods.

beberlei (changes requested):
Getting very close here :)

A few questions:
- Did we miss the notUpdateable now? Not sure what condition/code prevents a field to be updated. Can you point me to it?
- Can you add a test for a joined table updateable as well?

beberlei:
@mehldau I took the day to improve this PR, catching a nasty edge case with Second Level Cache that we haven't thought of yet.

I removed the exception when updating a field that is not updatable, this value just gets overwritten with the correct one from the DB now. There are still these todos I want to tackle before merging this:

- Add a test with joined table inheritance
- Add tests for `AbstractMappingDriver` using `Insertable` and `Updatable` entities, will also test xml, attribute and other drivers.
- Fetch only the right columns on insert or update, not all of them

beberlei:
One point that i am still struggling with, there are cases where you want to disable `insertable` or `updateable` but don't want data to be fetched from the database after each write. For example if you want to use `updateable: false` for immutable properties, then  why fetch them from the database?

Hibernate (but not JPA) differentiates with another option `@Generated(time)` where time is an enum with the values `always` or `insert`. We should copy this behavior instead of magically adding this when just using insertable/updateable.

derrabus:
What is our BC policy regarding `BasicEntityPersister`? The class is not final and full of `protected` members. This PR renames some protected methods and changes the signature of others. Do we expect userland code to override that class and specific methods in it? If so, shall we document this PR as a minor BC break?

beberlei:
@derrabus its not possible to override the persisters, their factory method is guarded by final things in EM and UoW.

derrabus:
So we should probably finalize them as well, I guess. For 3.0 at least.🤔

derrabus:
Anyway, this would be out if scope here. Thanks for the explanation.

beberlei:
I believe I want to simplify this to use `virtual: true` instead of `insertable/updatable`, because to me the use-cases for disabling only one of insert/update for a column seems to be a more than edgy edge case that we can just choose not to support.

beberlei:
Hm ok, virtual makes no sense in naming either, because that would mean the column does not exist (??) in the schema? Maybe the naming is fine the way it is here.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1277,6 +1313,34 @@ public function isNullable($fieldName)
...
     }
 
+    /**
+     * Checks if the field is insertable.
+     *
+     * @param string $fieldName The field name.
+     *
+     * @return bool TRUE if the field is insertable, FALSE otherwise.
+     */
+    public function isInsertable($fieldName)
```
derrabus: Let's add proper types to brand-new methods.

```suggestion
    public function isInsertable(string $fieldName): bool
```

lib/Doctrine/ORM/Mapping/Column.php
```diff
@@ -53,6 +59,12 @@ final class Column implements Annotation
     /** @var string|null */
     public $columnDefinition;
 
+    /**
+     * @var string|null
+     * * @Enum({"NEVER", "INSERT", "ALWAYS"})
```
derrabus:
```suggestion
     * @psalm-var 'NEVER'|'INSERT'|'ALWAYS'|null
     * @Enum({"NEVER", "INSERT", "ALWAYS"})
```

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -718,6 +732,8 @@ private function joinColumnToArray(Mapping\JoinColumn $joinColumn): array
      *                   unique: bool,
      *                   nullable: bool,
      *                   precision: int,
+     *                   notInsertable?: bool,
+     *                   notUpdateble?: bool,
```
derrabus, line 738:
```suggestion
     *                   notUpdateble?: bool,
     *                   generated?: ClassMetadata::GENERATED_*,
```

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -633,6 +633,20 @@ private function getFetchMode(string $className, string $fetchMode): int
         return constant('Doctrine\ORM\Mapping\ClassMetadata::FETCH_' . $fetchMode);
     }
 
+    /**
+     * Attempts to resolve the generated mode.
+     *
```
derrabus, line 638:
```suggestion
     *
     * @psalm-return ClassMetadata::GENERATED_*
     *
```

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -659,13 +683,22 @@ class ClassMetadataInfo implements ClassMetadata
      */
     public $changeTrackingPolicy = self::CHANGETRACKING_DEFERRED_IMPLICIT;
 
+    /**
+     * A Flag indicating whether one or more columns of this class have to be reloaded
+     * after insert / update operations.
+     *
+     * @readonly
```
greg0ire: This means the property can only be written in the constructor, and cannot be used here for that reason.

beberlei: Ah, i was about to suppress them, so its readonly to all the other places in the code, I saw that you reverted it, fine by me.

docs/en/reference/attributes-reference.rst
```diff
@@ -178,6 +178,18 @@ Optional parameters:
...
+   If not specified, default value is true.
+
+-  **updatable**: Boolean value to determine if the column should be
+   included when updating the row of the underlying entities table.
+   If not specified, default value is true.
+
+-  **generated**: An enum with the possible values ALWAYS, INSERT, NEVER.  Is
+   used after an INSERT or UPDATE statement to determine if the database
+   generated this value and it needs to be fetched using a SELECT statement.
+
```
greg0ire, line 192: Can you please add an example for attributes as well? We are likely to remove annotations and their docs in the future.

lib/Doctrine/ORM/Mapping/Builder/FieldBuilder.php
```diff
@@ -110,6 +110,34 @@ public function precision($p)
         return $this;
     }
 
+    /**
+     * Sets insertable.
+     *
+     * @return $this
+     */
+    public function insertable(bool $flag = true)
```
greg0ire:
```suggestion
    public function insertable(bool $flag = true): self
```

lib/Doctrine/ORM/Mapping/Builder/FieldBuilder.php
```diff
@@ -110,6 +110,34 @@ public function precision($p)
...
+
+        return $this;
+    }
+
+    /**
+     * Sets updatable.
+     *
+     * @return $this
+     */
+    public function updatable(bool $flag = true)
```
greg0ire:
```suggestion
    public function updatable(bool $flag = true): self
```

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -439,6 +457,12 @@ class ClassMetadataInfo implements ClassMetadata
      * - <b>nullable</b> (boolean, optional)
      * Whether the column is nullable. Defaults to FALSE.
      *
+     * - <b>'notInsertable'</b> (boolean, optional)
+     * Whether the column is not insertable. Optional. Is only set if value is TRUE.
+     *
+     * - <b>'notUpdatable'</b> (boolean, optional)
+     * Whether the column is updatable. Optional. Is only set if value is TRUE.
+     *
```
greg0ire, line 465: I'd recommend against negative logic and do something like this:

```suggestion
     * - <b>'insertable'</b> (boolean, optional)
     * Whether the column is not insertable. Optional. Is only set if value is FALSE.
     *
     * - <b>'updatable'</b> (boolean, optional)
     * Whether the column is updatable. Optional. Is only set if value is FALSE.
     *
```

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -297,50 +299,71 @@ public function executeInserts()
     /**
      * Retrieves the default version value which was created
      * by the preceding INSERT statement and assigns it back in to the
-     * entities version field.
+     * entities version field if the given entity is versioned.
+     * Also retrieves values of columns marked as 'non insertable' and / or
+     * 'not updatetable' and assigns them back to the entities corresponding fields.
```
greg0ire:
```suggestion
     * 'not updatable' and assigns them back to the entities corresponding fields.
```

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -594,12 +617,13 @@ public function delete($entity)
...
      *
-     * @param object $entity The entity for which to prepare the data.
+     * @param object $entity   The entity for which to prepare the data.
+     * @param bool   $isInsert Whether the data to be prepared refers to an insert statement.
      *
      * @return mixed[][] The prepared data.
      * @psalm-return array<string, array<array-key, mixed|null>>
      */
-    protected function prepareUpdateData($entity)
+    protected function prepareUpdateData($entity, $isInsert = false)
```
greg0ire:
```suggestion
    protected function prepareUpdateData($entity, bool $isInsert = false)
```

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
beberlei: Then the values have to be stored in the cached metadata for all fields, making the serialized cache string sizes larger and deserialize performance slower. The way it is now also simplifies checks for just having to test for isset() of the key.

docs/en/reference/attributes-reference.rst
beberlei: good point, done!

greg0ire (approved):

derrabus (approved):

SenseException (approved):

beberlei (approved):

beberlei:
@mehldau Thank you very much for your work on this important feature!

PR #9120, opened 2021-10-13
#9119 Query\Parser resolve Lexer:lookahead null case

### LINKED ISSUE #9119: Query\Parser : Trying to access array offset on value of type null
### Bug Report
Executing a simple select query return this error :
`request.CRITICAL: Uncaught PHP Exception ErrorException: "Notice: Trying to access array offset on value of type null" at C:\Applications\wamp64\www\application\vendor\doctrine\orm\lib\Doctrine\ORM\Query\Parser.php line 1576 {"exception":"[object] (ErrorException(code: 0): Notice: Trying to access array offset on value of type null at C:\\Applications\\wamp64\\www\\application\\vendor\\doctrine\\orm\\lib\\Doctrine\\ORM\\Query\\Parser.php:1576)"} []`

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.5

#### Summary
Since the `Lexer::$lookahead` property is nullable, we should have a check on this line https://github.com/doctrine/orm/blob/2.10.x/lib/Doctrine/ORM/Query/Parser.php#L1562 (and maybe others !) (the same code remains in the last version)

An old merge request fix the same bug in another file : https://github.com/doctrine/orm/pull/7934/files

ybenhssaien:
Closes #9119

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -588,7 +588,7 @@ class Parser
      */
     private function isFunction(): bool
     {
-        $lookaheadType = $this->lexer->lookahead['type'];
+        $lookaheadType = $this->lexer->lookahead['type'] ?? null;
         $peek          = $this->lexer->peek();
 
         $this->lexer->resetPeek();
@@ -618,7 +618,7 @@ class Parser
     private function isNextAllAnySome(): bool
     {
         return in_array(
-            $this->lexer->lookahead['type'],
+            $this->lexer->lookahead['type'] ?? null,
             [Lexer::T_ALL, Lexer::T_ANY, Lexer::T_SOME],
             true
         );
@@ -1516,7 +1516,7 @@ class Parser
         }
 
         // Still need to decide between IdentificationVariable or ResultVariable
-        $lookaheadValue = $this->lexer->lookahead['value'];
+        $lookaheadValue = $this->lexer->lookahead['value'] ?? null;
 
         if (! isset($this->queryComponents[$lookaheadValue])) {
             $this->semanticalError('Cannot group by undefined identification or result variable.');
@@ -1559,7 +1559,7 @@ class Parser
                 $expr = $this->ScalarExpression();
                 break;
 
-            case $this->lexer->lookahead['type'] === Lexer::T_CASE:
+            case ($this->lexer->lookahead['type'] ?? null) === Lexer::T_CASE:
                 $expr = $this->CaseExpression();
                 break;
 
@@ -1956,7 +1956,7 @@ class Parser
         $token = $this->lexer->lookahead;
         $peek  = $this->lexer->glimpse();
 
-        if ($token['type'] === Lexer::T_OPEN_PARENTHESIS && $peek['type'] === Lexer::T_SELECT) {
+        if ($token && $token['type'] === Lexer::T_OPEN_PARENTHESIS && $peek['type'] === Lexer::T_SELECT) {
             $this->match(Lexer::T_OPEN_PARENTHESIS);
             $expression = $this->Subselect();
             $this->match(Lexer::T_CLOSE_PARENTHESIS);
@@ -2239,7 +2239,7 @@ class Parser
         $expression    = null;
         $identVariable = null;
         $peek          = $this->lexer->glimpse();
-        $lookaheadType = $this->lexer->lookahead['type'];
+        $lookaheadType = $this->lexer->lookahead['type'] ?? null;
 
         switch (true) {
             // ScalarExpression (u.name)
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -2367,56 +2367,58 @@ public function SimpleSelectExpression()
     {
         $peek = $this->lexer->glimpse();
 
-        switch ($this->lexer->lookahead['type']) {
-            case Lexer::T_IDENTIFIER:
-                switch (true) {
-                    case $peek['type'] === Lexer::T_DOT:
-                        $expression = $this->StateFieldPathExpression();
+        if ($this->lexer->lookahead !== null) {
+            switch ($this->lexer->lookahead['type']) {
```
derrabus:
```suggestion
        switch ($this->lexer->lookahead['type'] ?? null) {
```

Let's keep the diff small here.

ybenhssaien: Done, I had to add a check to the line 2390

derrabus:
Can we create a test that reproduces the issue you've encountered?

bhushan (changes requested):
We could have simple helper functions in lexer as getLookheadType and getLookheadValue and handle this null case there and then use those helpers to get type and value.. 

Also as Alexander suggested, having test would be great with reproducing problem

cc @derrabus

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -2583,7 +2583,7 @@ public function SimpleConditionalExpression()
             return $this->ExistsExpression();
         }
 
-        $token     = $this->lexer->lookahead;
+        $token     = $this->lexer->lookahead ?? ['type' => null, 'value' => null];
```
bhushan:
```suggestion
        $token     = $this->lexer->lookahead ?? ['type' => null, 'value' => ''];
```

ybenhssaien:
> We could have simple helper functions in lexer as getLookheadType and getLookheadValue and handle this null case there and then use those helpers to get type and value..
> 
> Also as Alexander suggested, having test would be great with reproducing problem
> 
> cc @derrabus

The bug can rarely happen, so almost the time the `$this->lexer->lookahead` is not null, so the cost of calling a function to return a not null value for maybe 99% of cases doesn't worth it in my opinion, I suggest to keep it simple.

The bug is caused by the [datatable-bundle](https://github.com/omines/datatables-bundle) when generates a select query without columns names in `ORDER BY` clause, I will try to write a test for this case.

bhushan:
I think it is complicated to remember in 10 different places that value could be empty string and type could be null. Also there is a chance to not have things consistent, as you also set value as null in one place.. 🙂 

IMHO adding these getters simplifies things more and keeps it consistent..

greg0ire:
> so the cost of calling a function to return a not null value for maybe 99% of cases doesn't worth it in my opinion

That part of the code is used for parsing DQL, right? If yes, the DQL => SQL transformation is usually cached, which means this is not going to be the hot path of your application.

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -3624,4 +3623,20 @@ public function CustomFunctionsReturningStrings()
 
         return $function;
     }
+
+    /**
+     * @return int
```
bhushan, line 3630:
```suggestion
     * Gets the lookahead type used by the parser.
     *
     * @return int
```

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -3624,4 +3623,20 @@ public function CustomFunctionsReturningStrings()
...
+     */
+    protected function getLookaheadType()
+    {
+        return $this->lexer->lookahead['type'] ?? -1;
+    }
+
+    /**
+     * @return string
```
bhushan, line 3640:
```suggestion
     * Gets the lookahead value used by the parser.
     *
     * @return string
```

bhushan (changes requested):
Could you also write test, so that without your changes it breaks as you shared in issue https://github.com/doctrine/orm/issues/9119

ybenhssaien:
I added a new test case for a query with an empty `ORDER BY` which was the origin for this bug and what should be an invalid query instead

bhushan:
Could you fix other errors.. also you could run it locally to make sure everything is green .. it will save lot of time.. !!

tests/Doctrine/Tests/ORM/Query/LanguageRecognitionTest.php
```diff
@@ -150,6 +150,9 @@ public function invalidDQL(): array
             ['SELECT \'foo\' AS foo: FROM Doctrine\Tests\Models\CMS\CmsUser u'],
             ['SELECT \'foo\' AS foo:bar FROM Doctrine\Tests\Models\CMS\CmsUser u'],
 
+            /* Checks for incomplete queries */
+            ['SELECT \'foo\' FROM Doctrine\Tests\Models\CMS\CmsUser u ORDER BY '],
```
bhushan, line 154: Is this test case looks good ? 

cc @greg0ire @derrabus

greg0ire: The `\'foo\'` part is a bit distracting. I would have used
```suggestion
            'incomplete query' => ['SELECT u FROM Doctrine\Tests\Models\CMS\CmsUser u ORDER BY '],
```

PR #9131, opened 2021-10-17
Overview table for events: Jump links

ThomasLandauer:
Well, the answer to what I said at https://github.com/doctrine/orm/pull/9039#issue-743906810

> I'd like to see if this does come out nicely on the web page...

... is clearly NO ;-)

Anyway, let's continue and care about the layout later...

Questions:

1:
Is it OK to link them like so? => Then I'll do the others.

2:
The order of the events differs between the table (=same as list) and the chapters further down. I think the chapters make more sense (starting with `prePersist`), so I'm going to change the table.

3:
Suggestion for the next table column: **Can change current entity**


SenseException:
1. Sure, links might help to get more infos on an event
2. Sounds reasonable :-)
3. Is this supposed to be a new column for the table? One that shows if it's meant to change an entity?

ThomasLandauer:
I'm done with the links, please merge.

Remarks/questions:
1: `onClear` doesn't have a section. So I'd create one, just with the short info from the list https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/events.html#lifecycle-events
2: `onClassMetadataNotFound` is dispatched when loading fails (right?), so I'll just add a short note to https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/events.html#load-classmetadata-event (i.e. don't create a dedicated section for this)
3: I'll do the changing of the order in a separate PR, otherwise this would be impossible to review ;-)
4: "Can change current entity": Yes, that would be a new column in the table. Answering the question: If I modify the current entity there (e.g. `$entity->setFoo(...)`), will this have an effect, or is it already too late?

SenseException:
> onClear doesn't have a section. So I'd create one, just with the short info from the list https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/events.html#lifecycle-events

> onClassMetadataNotFound is dispatched when loading fails (right?), so I'll just add a short note to https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/events.html#load-classmetadata-event (i.e. don't create a dedicated section for this)

Did you mean in this PR? Should I wait with further review or will you introduce it in another PR?

ThomasLandauer:
> Did you mean in this PR?

No! Please review and merge this as-is, then I'll continue in a new one.

SenseException (approved):

SenseException:
Thanks @ThomasLandauer :sunflower:

PR #9133, opened 2021-10-19
Adding a setup helper for attribute metadata config

judahnator:
I noticed there was a `\Doctrine\ORM\Tools\Setup::createAnnotationMetadataConfiguration` method, but a method for attributes was missing. This PR adds a  similar `\Doctrine\ORM\Tools\Setup::createAttributeMetadataConfiguration` method as well.

First time PR to Doctrine so go easy :)

### DIFF
--- a/lib/Doctrine/ORM/Tools/Setup.php
+++ b/lib/Doctrine/ORM/Tools/Setup.php
@@ -14,6 +14,7 @@ use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\Common\Cache\RedisCache;
 use Doctrine\Common\ClassLoader;
 use Doctrine\ORM\Configuration;
+use Doctrine\ORM\Mapping\Driver\AttributeDriver;
 use Doctrine\ORM\Mapping\Driver\XmlDriver;
 use Doctrine\ORM\Mapping\Driver\YamlDriver;
 use Memcached;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Tools/Setup.php
```diff
@@ -73,6 +74,23 @@ public static function createAnnotationMetadataConfiguration(array $paths, $isDe
...
+     *
+     * @param mixed[] $paths
+     * @param bool    $isDevMode
+     * @param string  $proxyDir
+     *
+     * @return Configuration
+     */
+    public static function createAttributeMetadataConfiguration(array $paths, $isDevMode = false, $proxyDir = null, ?Cache $cache = null)
+    {
+        $config = self::createConfiguration($isDevMode, $proxyDir, $cache);
```
```suggestion
     * Creates a configuration with an attribute metadata driver.
     *
     * @param mixed[] $paths
     */
    public static function createAttributeMetadataConfiguration(
        array $paths,
        bool $isDevMode = false,
        string $proxyDir = null,
        ?Cache $cache = null
    ): Configuration {
        $config = self::createConfiguration($isDevMode, $proxyDir, $cache);
```

greg0ire:
This is a new feature and as such, should target 2.11.x. You can retarget with the Edit button.
Note that I'm not 100% sure about this, since it could also be seen as a bugfix for something we forgot. WDYT @SenseException @derrabus

SenseException:
Technically 2.11, but `Setup` just exists to make things easier for users, so I guess 2.10 would be okay too.

judahnator:
Where are we at with this one? Any last changes needing made?

derrabus:
> Any last changes needing made?

Yes, the CI needs to be green.

derrabus:
Other than that, are we able to provide a test for your change?

judahnator:
I wrote a test and the checks have passed!

lib/Doctrine/ORM/Tools/Setup.php
greg0ire: ping @judahnator

greg0ire:
There are 7 commits, please get rid of the first one, and squash the others together

How to do that?

1. `git rebase -i origin/2,11,x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `drop` on the first line, and then `pick` with `fixup` on every line but the second one

judahnator:
Neat! I didn't know about those neat git tricks. Hopefully, I got everything right!

greg0ire (approved):

SenseException (approved):

greg0ire:
Congrats @judahnator !

PR #9135, opened 2021-10-20
Regenerate Psalm baseline

derrabus:
After the latest Doctrine Common release, a new error popped up in Psalm:

```
ERROR: PossiblyUndefinedMethod - lib/Doctrine/ORM/UnitOfWork.php:2875:44 - Method (Doctrine\Common\Proxy\Proxy<object>&Doctrine\Common\Proxy\Proxy<object>)::addPropertyChangedListener does not exist (see https://psalm.dev/108)
                                $newValue->addPropertyChangedListener($this);
```

This looks like a false positive to me, so I decided to regenerate the Psalm baseline to make the build green again.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -273,8 +273,7 @@
       <code>getCacheFactory</code>
       <code>getTimestampRegion</code>
     </PossiblyNullReference>
-    <RedundantConditionGivenDocblockType occurrences="2">
-      <code>$result</code>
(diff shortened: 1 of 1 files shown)


greg0ire (approved):
Definitely a false positive: the piece of code is preceded by this condition:

```
                            if (
                                $newValue instanceof NotifyPropertyChanged &&
                                ( ! $newValue instanceof Proxy || $newValue->__isInitialized())
                            ) {
```

PR #9136, opened 2021-10-20
Revert "Removing all the occurence of any"

greg0ire:
This reverts commit 84afd6c93739c1a112a00ecc9e88ebb42e8cfe42, because it
is a BC-break that seems to affect more people than we originally
thought it would.


derrabus (approved):
Merge #9135 if you want a green build. 🧐

SenseException (approved):
> Merge #9135 if you want a green build. :monocle_face: 

Ah, good. I was wondering about the broken builds.

PR #8837, opened 2021-07-10
Replace spl_object_hash() with spl_object_id()

greg0ire:
It is more efficient, and can be provided to 7.1 users thanks to
`symfony/polyfill-php72`.

Backports #6878 

@beberlei I know we wanted to target Doctrine 3 with this, but I found a way to target v2. WDYT?

PR #9143, opened 2021-10-21
Document BC break

### LINKED ISSUE #9141: `spl_object_hash()` vs `spl_object_id()`
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.10

#### Summary

After https://github.com/doctrine/orm/pull/8837, there are some UoW methods that the parameter has changed:

https://github.com/doctrine/orm/blob/81d472f6f96b8b571cafefe8d2fef89ed9446a62/lib/Doctrine/ORM/UnitOfWork.php#L3243-L3254

https://github.com/doctrine/orm/blob/81d472f6f96b8b571cafefe8d2fef89ed9446a62/lib/Doctrine/ORM/UnitOfWork.php#L3025-L3040

Because the OID is different, [at first I thought they were internal](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2271#issuecomment-948600889), but apparently it is some kind of [internal notes.](https://github.com/doctrine/orm/pull/1334) 

#### Previous behavior

Before this, for calling these method you should generate the oid with `spl_object_hash()`

#### Current behavior

Now that oid is not valid anymore.

#### How to reproduce

Tests started to fail in https://github.com/doctrine-extensions/DoctrineExtensions

greg0ire:
Closes #9141

I'm not great at assessing the severity of BC-breaks, so please do tell if you feel we should revert #8837 , or if you have an idea to somehow make both things work without affecting performance too much.

cc @franmomu

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,12 @@
 # Upgrade to 2.10
 
+## BC Break: `UnitOfWork` now relies on SPL object IDs, not hashes
+
+When calling the following methods, you are now supposed to use the result of
+`spl_object_id()`, and not `spl_object_hash()`:
+- `UnitOfWork::clearEntityChangeSet()`
+- `UnitOfWork::setOriginalEntityProperty()`
+
 ## BC Break: Removed `TABLE` id generator strategy
 
 The implementation was unfinished for 14 years.


franmomu (approved):
👍  I'll try to make it work with another approach

derrabus (approved):

stof:
@greg0ire a way to provide a BC layer would be to detect cases passing a string as argument instead of an int, and converting the object hash to an object id (and triggering a deprecation of course). the logic for that is available in the implementation of the `spl_object_id` polyfill (as that's exactly how the polyfill is implemented)

greg0ire:
Good to know! I'll reuse it if anyone else complains about this.

PR #9147, opened 2021-10-23
return 0 when there's no metadata to process

wickedOne:
fixes symfony command throwing a type error in the run method.

happens @ https://github.com/symfony/console/blob/5.3/Command/Command.php#L301


derrabus (dismissed):

derrabus:
Good catch, thank you. Let me update the PHPStan baseline for you.

derrabus:
Oh, I apparently cannot do that. Can you please run this on your branch and commit the `phpstan-baseline.neon` file?

```
./vendor/bin/phpstan --generate-baseline
```

wickedOne:
done...

derrabus (approved):

Issue #9021, opened 2021-09-17
Support for native PHP enums


derrabus:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

PHP 8.1 has introduced native enums to the language.

```php
enum Suit: string {
  case Hearts = 'H';
  case Diamonds = 'D';
  case Clubs = 'C';
  case Spades = 'S';
}

#[Entity]
class Card
{
    #[Column, Id]
    public int $id;

    #[Column]
    public Suit $suit;
}
```

I'd like to use an enum as type of a property. The ORM should store the enums in a `VARCHAR` field using the backed value. When hydrating such an entity, the enum should be hydrated properly.

The resulting table `card` would look like this:

| **id** | **suit** |
|---|---|
| 1 | H |
| 2 | H |
| 3 | S |
| 4 | C |

Nek-:
Using a VARCHAR looks weird because PostgresSQL, MariaDB, and MySQL support all enumerations. Is there any good reason (besides "it's work and we have nobody to do it") to not introduce the enumeration support in Doctrine?

_Btw thank you very much for maintaining Doctrine!_

derrabus:
> Using a VARCHAR looks weird

I disagree.

> because PostgresSQL, MariaDB, and MySQL support all enumerations.

Yes, but others don't and I don't see why we should forbid using PHP enums on e.g. SQL Server.

> Is there any good reason (besides "it's work and we have nobody to do it") to not introduce the enumeration support in Doctrine?

I personally stopped using native enums a long time ago because they rarely help. And from my point of view, they come with major drawbacks:

* There is (at least last time I checked) no elegant way to determine a list of enum members. This would make schema diffs for doctrine fairly complicated.
* Adding a new member to the enum means performing an `ALTER TABLE`. Depending on the size of your table, this can be a very expensive operation.
* MySQL enums behave weird for integer values.

Even if we added support for MySQL's ENUM type to Doctrine one fine day, it think it should not be our default solution for PHP enums.

sicet7:
@derrabus 

> There is (at least last time I checked) no elegant way to determine a list of enum members.

Correct me if im wrong, but you should be able to call the method "cases" on the Enum to get a list of the members.
Atleast from what i have read of the PHP 8.1 UnitEnum Interface.

greg0ire:
I think @Nek- was talking about database `ENUM` here, not PHP's Enum.
[...]

PR #9149, opened 2021-10-25
Detect enums and use type string (PHP 8.1 only)

Nek-:
This is a patch for #9021 .

Edit: test added.

See https://github.com/doctrine/dbal/pull/4930

### DIFF
--- a/tests/Doctrine/Tests/Models/TypedProperties/ArticleStateEnum.php
+++ b/tests/Doctrine/Tests/Models/TypedProperties/ArticleStateEnum.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\TypedProperties;
+
+enum ArticleStateEnum
+{
+case DRAFT;
+case PUBLISHED;
+    }
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1493,6 +1493,10 @@ class ClassMetadataInfo implements ClassMetadata
                     case 'string':
                         $mapping['type'] = Types::STRING;
                         break;
+                    default:
+                        if (PHP_VERSION_ID >= 80100 && enum_exists($type->getName())) {
+                            $mapping['type'] = Types::ENUM;
+                        }
                 }
             }
         }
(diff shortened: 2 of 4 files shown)


derrabus:
Converting to draft because tests are missing.

> About the tests I am sorry I didn't find out how to test something only for a specific version of PHP.

https://phpunit.readthedocs.io/en/9.5/annotations.html#requires

> This is a patch for #9021 .

I doubt that your change really solves the problem completely, but let's write some functional tests first an iterate from there.

Nek-:
@derrabus I think it's better like this indeed :) . And the rest should be inside the dbal! On it!

tests/Doctrine/Tests/Models/TypedProperties/ArticleStateEnum.php
```diff
@@ -0,0 +1,11 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\TypedProperties;
+
+enum ArticleStateEnum
+{
+case DRAFT;
+case PUBLISHED;
+    }
```
Nek-, line 11: Sorry for that, the current version of phpcs have issues with enums 😅 .

tests/Doctrine/Tests/Models/TypedProperties/ArticleStateEnum.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\TypedProperties;
+
+enum ArticleStateEnum
```
derrabus, line 7: We should support backed enums only.

Nek-: Why? Everything will work like this. I do not understand this limitation.

derrabus: Unit enums do not specify a way to convert them from or to a scalar value. This is the reason why backed enums exist in the first place.

Nek-: What about creating a Type for each enum type. A special serialization could be done for Unit enums making it easier to query. Maybe a json field.

I think all of this leads to more code but really seems fine to me. WDYT? (if you valid I go on this)

derrabus: No. If someone wants to hydrate/dehydrate enums from/to the database, they can back it with scalar values for that purpose. That's why backed enums exist.

For unbacked enums, we need a custom conversion logic and in that case, it would be easier if the developer wrote their own DBAL type for that.

derrabus:
Closing as explained in doctrine/dbal#4930.

Issue #9078, opened 2021-10-04
Schema Update not working from json_array to json with DBAL 3


alexander-schranz:
For all stumbling over this check: https://dunglas.dev/2022/01/json-columns-and-doctrine-dbal-3-upgrade/

### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.10.0

#### Summary

The deprecated json_array type is not longer supported but its actually not possible to upgrade in 2.10.0 the database from json_array to json.

#### Current behavior

After upgrade to doctrine/orm I did replace all `json_array` with `json` but when now running:

```
bin/console doctrine:schema:update --dump-sql
```

The the FieldRegistry fails as the database still has the comment `(DC2Type:json_array)` in it and is so mapped to the json_array type.

#### How to reproduce

Create a Entity with `json_array` update doctrine/orm to `2.10.0` change type from `json_array` to `json`  and run:

```
bin/console doctrine:schema:update --dump-sql
```

```
In Exception.php line 125:

  [Doctrine\DBAL\Exception]
  Unknown column type "json_array" requested. Any Doctrine type that you use has to be registered with \Doctrine\DBAL\Types\Type::addType(). You can get a list of all the known types with \Doct
  rine\DBAL\Types\Type::getTypesMap(). If this error occurs during database introspection then you might have forgotten to register all database types for a Doctrine Type. Use AbstractPlatform#
  registerDoctrineTypeMapping() or have your custom types implement Type#getMappedDatabaseTypes(). If the type name is empty you might have a problem with the cache or forgot some mapping infor
  mation.


Exception trace:
  at /Users/alexanderschranz/Documents/Sulu/sulu-develop.localhost/vendor/sulu/sulu/vendor/doctrine/dbal/src/Exception.php:125
 Doctrine\DBAL\Exception::unknownColumnType() at /Users/alexanderschranz/Documents/Sulu/sulu-develop.localhost/vendor/sulu/sulu/vendor/doctrine/dbal/src/Types/TypeRegistry.php:37
 Doctrine\DBAL\Types\TypeRegistry->get() at /Users/alexanderschranz/Documents/Sulu/sulu-develop.localhost/vendor/sulu/sulu/vendor/doctrine/dbal/src/Types/Type.php:143
[...]

PR #9152, opened 2021-10-26
Add doctrine/dbal to project composer.json

chasen:
As discussed in https://github.com/doctrine/orm/issues/9078 when entities utilize data mappings which are provided by the dbal lib it is expected behavior that users will explicitly define their dependency on the package.

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -114,10 +114,13 @@ Add the following directories:
     The YAML driver is deprecated and will be removed in version 3.0.
     It is strongly recommended to switch to one of the other mappings.
 .. note::
-    When using the entity data mapping it is best practice to explicity
-    specify the version of ``doctrine/dbal`` in your ``composer.json``.
-    This helps ensure that your entity data mappings remain supported
-    through package upgrades.
+    It is strongly recommended that you require doctrine/dbal in your
+    composer.json as well, because using the ORM means mapping objects
+    and their fields to database tables and their columns, and that
+    requires mentioning so-called types that are defined in doctrine/dbal
+    in your application. Having an explicit requirement means you control
+    when the upgrade to the next major version happens, so that you can
+    do the necessary changes in your application beforehand.
 
 Obtaining the EntityManager
 ---------------------------
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -114,7 +114,10 @@ Add the following directories:
     The YAML driver is deprecated and will be removed in version 3.0.
     It is strongly recommended to switch to one of the other mappings.
 .. note::
-    When using the entity data mapping it is best practice to explicity specify the version of doctrine/dbal in your composer.json. This helps ensure that your entity data mappings remain supported through package upgrades.
+    When using the entity data mapping it is best practice to explicity
+    specify the version of ``doctrine/dbal`` in your ``composer.json``.
+    This helps ensure that your entity data mappings remain supported
+    through package upgrades.
 
 Obtaining the EntityManager
 ---------------------------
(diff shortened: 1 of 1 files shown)


docs/en/tutorials/getting-started.rst
```diff
@@ -112,6 +113,8 @@ Add the following directories:
 .. note::
     The YAML driver is deprecated and will be removed in version 3.0.
     It is strongly recommended to switch to one of the other mappings.
+.. note::
+    When using the entity data mapping it is best practice to explicity specify the version of doctrine/dbal in your composer.json. This helps ensure that your entity data mappings remain supported through package upgrades.
```
greg0ire:
```suggestion
    When using the entity data mapping it is best practice to explicity
    specify the version of ``doctrine/dbal`` in your ``composer.json``.
    This helps ensure that your entity data mappings remain supported
    through package upgrades.
```

greg0ire: Are there usages of the ORM where you would not entity data mapping at all?

chasen: I cant think of any. Which is part of the reason why I was confused about needing to specify the dbal package.

greg0ire: Then I think we should rephrase this: beginners probably don't know what "entity data mapping" is and might think they should skip that requirement temporarily. How about this:

> It is strongly recommended that you require `doctrine/dbal` in your `composer.json` as well, because using the ORM means mapping objects and their fields to database tables and their columns, and that requires mentioning so-called types that are defined in `doctrine/dbal` in your application. Having an explicit requirement means you control when the upgrade to the next major version happens, so that you can do the necessary changes in your application beforehand.

docs/en/tutorials/getting-started.rst
```diff
@@ -112,6 +113,11 @@ Add the following directories:
 .. note::
     The YAML driver is deprecated and will be removed in version 3.0.
     It is strongly recommended to switch to one of the other mappings.
+.. note::
+    When using the entity data mapping it is best practice to explicity
+    specify the version of ``doctrine/dbal`` in your ``composer.json``.
+    This helps ensure that your entity data mappings remain supported
+    through package upgrades.
```
SenseException:
```suggestion
    When using the entity data mapping it is a best practice to explicity
    specify the version of ``doctrine/dbal`` in your ``composer.json``.
    This helps to ensure that your entity data mappings remain supported
    through package upgrades.
```

docs/en/tutorials/getting-started.rst
chasen: I like it! Ill update it

SenseException (dismissed):

docs/en/tutorials/getting-started.rst
```diff
@@ -112,6 +113,14 @@ Add the following directories:
 .. note::
     The YAML driver is deprecated and will be removed in version 3.0.
     It is strongly recommended to switch to one of the other mappings.
+.. note::
+    It is strongly recommended that you require doctrine/dbal in your
```
greg0ire:
```suggestion
    It is strongly recommended that you require ``doctrine/dbal`` in your
```

docs/en/tutorials/getting-started.rst
```diff
@@ -112,6 +113,14 @@ Add the following directories:
...
+    composer.json as well, because using the ORM means mapping objects
```
greg0ire:
```suggestion
    ``composer.json`` as well, because using the ORM means mapping objects
```

docs/en/tutorials/getting-started.rst
```diff
@@ -112,6 +113,14 @@ Add the following directories:
...
+    and their fields to database tables and their columns, and that
+    requires mentioning so-called types that are defined in doctrine/dbal
```
greg0ire:
```suggestion
    requires mentioning so-called types that are defined in ``doctrine/dbal``
```

docs/en/tutorials/getting-started.rst
chasen: Language was changed to be more clear, if there are more grammar issues in the new verbiage let me know happy to accept any improvements

greg0ire (approved):

greg0ire:
Thanks @chasen !

PR #9010, opened 2021-09-13
Fix ignoring custom types for PersistentCollection matching()

sztyup:
Fixes https://github.com/doctrine/orm/issues/8995, https://github.com/doctrine/orm/issues/8406, https://github.com/doctrine/orm/issues/7822

PR #9153, opened 2021-10-28
Infer type from field instead of column

### LINKED ISSUE #9109: ArrayCollection->matching() not working anymore with Criteria->eq('id', $number)
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.9.6

#### Summary

It seems that something changed with the `->matching()` method of an _ArrayCollection_. It persist in _v2.10.1_.

#### Current behavior

When I have an ArrayCollection I want to filter with `matching` and a Criteria, which has an expression like `->eq('id', $number)` my script tells me something like:

_"Could not resolve type of column "\`id\`" of class ..."_ and so on.

#### How to reproduce

Let's say you have an entity _"work"_ with an _ArrayCollection_ called _"connectedalbums"_, which stores other entities like _"album"_. I now want to access the _album_ with the id "3" and would do it like this:

```php
// $work is already fetched form the DB.

$criteria = new Criteria();
$criteria->where(Criteria::expr()->eq('id', 3));
$album = $work->getConnectedAlbums()->matching($criteria)->toArray();

if (!empty($album)) {
    $album = $album[0];
}
```

With this, PHP throws an _Uncaught RuntimeException_ like mentioned above.

#### Expected behavior

I expect an ArrayCollection with `->matching()` and the Criteria with the expression `->eq('id', $number)` to give me the correct item from the ArrayCollection, whichs column _id_ has the value, which is stored in _$number_. It worked like this in _v2.9.5_!

armenio:
getTypeOfColumn() relies on getTypeOfField(), and does not suffer from
mismatching issues caused by quoting, because you cannot quote a field.
Since a field can be composite, that method returns an array, hence why we
need to select the first element.

Fixes https://github.com/doctrine/orm/issues/9109 caused by https://github.com/doctrine/orm/pull/9010

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
@@ -1,0 +1,217 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\ManyToMany;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group GH-9109
+ */
+class GH9109Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema(
+            [
+                $this->_em->getClassMetadata(GH9109User::class),
+                $this->_em->getClassMetadata(GH9109Product::class),
+            ]
+        );
+    }
+
+    protected function tearDown(): void
+    {
+        $this->_schemaTool->dropSchema(
+            [
+                $this->_em->getClassMetadata(GH9109User::class),
+                $this->_em->getClassMetadata(GH9109Product::class),
+            ]
+        );
+
+        parent::tearDown();
+    }
+
+    public function testIssue(): void
+    {
+        $userFirstName = 'GH9109Test';
+        $userLastName  = 'UserGH9109';
+        $productTitle  = 'Test product';
+
+        $userRepository = $this->_em->getRepository(GH9109User::class);
+
+        $user = new GH9109User();
+        $user->setFirstName($userFirstName);
+        $user->setLastName($userLastName);
+
+        $product = new GH9109Product();
+        $product->setTitle($productTitle);
+
+        $this->_em->persist($user);
+        $this->_em->persist($product);
+        $this->_em->flush();
+
+        $product->addBuyer($user);
+
+        $this->_em->persist($product);
+        $this->_em->flush();
+
+        $this->_em->clear();
+
+        $persistedProduct = $this->_em->find(GH9109Product::class, $product->getId());
+
+        // assert Product was persisted
+        self::assertInstanceOf(GH9109Product::class, $persistedProduct);
+        self::assertEquals($productTitle, $persistedProduct->getTitle());
+
+        // assert Product has a Buyer
+        $count = $persistedProduct->getBuyers()->count();
+        self::assertEquals(1, $count);
+
+        // assert NOT QUOTED will WORK with findOneBy
+        $user = $userRepository->findOneBy(['lastName' => $userLastName]);
+        self::assertInstanceOf(GH9109User::class, $user);
+        self::assertEquals($userLastName, $user->getLastName());
+
+        // assert NOT QUOTED will WORK with Criteria
+        $criteria = Criteria::create();
+        $criteria->where($criteria->expr()->eq('lastName', $userLastName));
+        $user = $persistedProduct->getBuyers()->matching($criteria)->first();
+        self::assertInstanceOf(GH9109User::class, $user);
+        self::assertEquals($userLastName, $user->getLastName());
+
+        // assert QUOTED will WORK with findOneBy
+        $user = $userRepository->findOneBy(['firstName' => $userFirstName]);
+        self::assertInstanceOf(GH9109User::class, $user);
+        self::assertEquals($userFirstName, $user->getFirstName());
+
+        // assert QUOTED will WORK with Criteria
+        $criteria = Criteria::create();
+        $criteria->where($criteria->expr()->eq('firstName', $userFirstName));
+        $user = $persistedProduct->getBuyers()->matching($criteria)->first();
+        self::assertInstanceOf(GH9109User::class, $user);
+        self::assertEquals($userFirstName, $user->getFirstName());
+    }
+}
+
+/**
+ * @Entity
+ */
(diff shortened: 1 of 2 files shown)


armenio:
need labels: "Bug" and "Failing Test"

greg0ire:
>  In attempt to resolve doctrine#9109 

Specifically, the commit message should explain what did not work before exactly, and why it works now. Here is the commit message I would reverse-engineer after reading your PR:

```
Infer type from field instead of column

getTypeOfColumn() relies on getTypeOfField(), and does not suffer from
mismatching issues caused by quoting, because you cannot quote a field.
Since a field can be composite, that method returns an array, hence why we
need to select the first element.

Fixes #9109
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
```diff
@@ -0,0 +1,201 @@
...
+        $userFirstName = 'GH9109Test';
+        $userLastName  = 'UserGH9109';
+        $productTitle  = 'Test product';
+
+        $this->_schemaTool->createSchema(
+            [
+                $this->_em->getClassMetadata(GH9109User::class),
+                $this->_em->getClassMetadata(GH9109Product::class),
+            ]
+        );
```
greg0ire, line 31: Please call `$this->_schemaTool->dropSchema(…)` in `tearDown()`

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
```diff
@@ -0,0 +1,201 @@
...
+        $this->title = $title;
+    }
+
+    public function getTitle(): string
+    {
+        return $this->title;
+    }
+
+    /**
+     * @return Collection
```
greg0ire: Please use annotations similar to those line 112 - 113

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
```diff
@@ -0,0 +1,201 @@
...
+
+        $persistedProduct = $this->_em->find(GH9109Product::class, $product->getId());
+
+        // assert Product was persisted
+        self::assertInstanceOf(GH9109Product::class, $persistedProduct);
+        self::assertEquals($productTitle, $persistedProduct->getTitle());
+
+        // assert Product has a Buyer
+        $criteria = Criteria::create();
+        $count = $persistedProduct->getBuyers()->matching($criteria)->count();
```
greg0ire: What's the point of calling `matching` with an empty criteria?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
```diff
@@ -0,0 +1,201 @@
...
+
+        // assert NOT QUOTED WORK with findOneBy
+        $user = $userRepository->findOneBy(['lastName' => $userLastName]);
+        self::assertInstanceOf(GH9109User::class, $user);
+        self::assertEquals($userLastName, $user->getLastName());
+
+        // assert NOT QUOTED WORK with Criteria
+        $criteria = Criteria::create();
+        $criteria->where($criteria->expr()->eq('lastName', $userLastName));
+        $matchedUserLastName = $persistedProduct->getBuyers()->matching($criteria)->first();
```
greg0ire: This variable name is confusing: one might think it will be a name, when it really is a user. IMO you could just reuse `$user` here, like you did below

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
```diff
@@ -0,0 +1,201 @@
...
+
+        // assert QUOTED WORK with findOneBy
+        $user = $userRepository->findOneBy(['firstName' => $userFirstName]);
+        self::assertInstanceOf(GH9109User::class, $user);
+        self::assertEquals($userFirstName, $user->getFirstName());
+
+        // assert QUOTED DOES WORK with Criteria
+        $criteria = Criteria::create();
+        $criteria->where($criteria->expr()->eq('firstName', $userFirstName));
+        $matchedUserFirstName = $persistedProduct->getBuyers()->matching($criteria)->first();
```
greg0ire: Same here.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
armenio: done

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
armenio: done

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
armenio: fixed

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
armenio: fixed. Thanks for the suggestion

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9109Test.php
armenio: fixed. Thanks for the suggestion

greg0ire:

How to do that?

1. `git rebase -i origin/2.10.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
3. For the first line, use `r`, so that you get a chance to re-use the commit message in https://github.com/doctrine/orm/pull/9153#issuecomment-961661495
4. Close your editor, git should do its magic, and you should end up with one commit
5. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

armenio:
Commits are squashed

greg0ire:
Please use `git commit --amend` to drop the `In attempt to resolve #9109` part of your commit message.

armenio:
done.

greg0ire (dismissed):
As a related but separate issue from the issue at hand, I think we should either document that the `$columnName` argument is supposed to be quoted if that's indeed the intended usage, or improve that method to make it work regardless of whether it is quoted or not.

armenio:
Thank you very much for your effort @greg0ire

greg0ire:
@armenio you're welcome! I notice there is an issue with PostgreSQL though, can you please look into it?

armenio:
Fixed PgSQL error with changing strategy to "AUTO" in @GeneratedValue of PK's

greg0ire (approved):

beberlei:
@greg0ire This should maybe even target a new 2.9.7 or? It was merged to 2.9.6 as far as i see breaking a core function.

greg0ire:
@beberlei 2.9.x is no longer supported, and upgrading to 2.10.x does not seem like a huge deal so I would backport it iff someone complains they cannot upgrade. Otherwise, it's a good incentive to get people to upgrade IMO 🙂

SenseException (approved):

greg0ire:
Thanks @armenio !

beberlei:
@greg0ire I guess another suggestion could be to downgrade to 2.9.4

PR #9154, opened 2021-10-28
Fixing more links

ThomasLandauer:
The first two I missed in https://github.com/doctrine/orm/pull/9151
The third is probably older.
Shouldn't the chapter name be displayed as link text by default?? Are you sure that everything is set up correctly with the parser?

### DIFF
--- a/docs/en/reference/architecture.rst
+++ b/docs/en/reference/architecture.rst
@@ -184,6 +184,8 @@ in well defined units of work. Work with your objects and modify
 them as usual and when you're done call ``EntityManager#flush()``
 to make your changes persistent.
 
+.. _unit-of-work:
+
 The Unit of Work
 ~~~~~~~~~~~~~~~~
 
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -127,9 +127,10 @@ Registering Events
 There are two ways to register an event:
 
 * *All events* can be registered by calling ``$eventManager->addEventListener()``
-or ``eventManager->addEventSubscriber()``, see :ref:`listening-and-subscribing-to-lifecycle-events`
+or ``eventManager->addEventSubscriber()``, see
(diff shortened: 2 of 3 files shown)


ThomasLandauer:
@SenseException I just wanted to start changing the order in the table (what I announced in https://github.com/doctrine/orm/pull/9131), but the overall structure of the page seems flawed:
[Load ClassMetadata Event](https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/events.html#load-classmetadata-event) is a top (!) level heading and "SchemaTool Events" too. And only the former events are included in the table, the latter aren't. I see two ways to fix this:

1. Include *all* events in the table. Merge all text starting at [Implementing Event Listeners](https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/events.html#implementing-event-listeners) into *one* section (called something like "List of Events").
2. Remove those "special" events (`onClassMetadataNotFound`, `postGenerateSchemaTable`, etc.) from the table, and leave their text sections where they are. But then move the table closer to "Implementing Event Listeners".

I have no preference, since I still haven't fully understood how all those topics belong together - that's why I started the "overview/table project" ;-)

SenseException:
@ThomasLandauer Thanks again. Could you please combine your link-PRs into one PR instead of creating multiple ones? Maybe you could use this one and close the others.

ThomasLandauer:
Just double-checked: They all have specific comments/questions :-(
I usually don't mix unrelated stuff (learned that from symfony-docs ;-) cause when *one* is problematic, at least the other can be merged in the meantime...
But I'll try to improve in the future :-)

SenseException:
I don't want to have multiple commits with the same type of additions. In case of the links we can have them in one PR.

ThomasLandauer:
OK. I integrated two, and closed one.

SenseException (approved):

Issue #9020, opened 2021-09-17
Add documentation on how to use PHP attributes


andreas-aeschlimann:
Being new to Doctrine, I followed the guides and successfully integrated Doctrine into our project. As suggested in the docs, we used PHPDoc Annotation Metadata as follows:

```
/**
 * @ORM\Entity
 * @ORM\Table(name="user")
 */
class User { ... }
```

As PHP 8 is out now for a while, I decided that I would like to switch to PHP Attributes instead such that we may use:

```
#[ORM\Entity]
#[ORM\Table(name: "user")]
class User { ... }
```

After doing that, it did not work out of the box as inferred on the website – all PHP Attributes were completely ignored. I also did not find any information on how to make it properly work. Only after adding `$config->setMetadataDriverImpl(new AttributeDriver([]));` to the `EntityManager` config, PHP Attributes seemed to work. 

I do not know if this is the correct way to enable attributes, but it would be nice if it were documented on the website.

derrabus:
Do you want to submit a pull request with the documentation that would've helped you?

andreas-aeschlimann:
> Do you want to submit a pull request with the documentation that would've helped you?

I could do that, but I am probably not "good" enough yet to write about best practices in Doctrine. Did I even enable attributes in the appropriate way? Or are there other, maybe better ways?

derrabus:
> I could do that, but I am probably not "good" enough yet to write about best practices in Doctrine.

No worries, you're not alone in this. Someone will review your PR.

> Did I even enable attributes in the appropriate way?

Yes. 🙂

ThomasLandauer:
I'm currently overhauling the "Events" page, and I could add some attributes examples. But before, somebody needs to configure the RST parser to allow a new type of `.. code-block::`
(If you tell me in which file this is happening, I could even try myself.)

It's probably best to follow what Symfony did, see e.g. https://symfony.com/doc/current/doctrine/associations.html#mapping-the-manytoone-relationship
Four tabs above the code block:
Annotations | Attributes | YAML | XML
(I personally would even list Attributes (=future) first, and YAML (=past) last, but that's not so important right now)

Symfony's RST syntax:
```
.. code-block:: php-annotations
.. code-block:: php-attributes
.. code-block:: yaml
.. code-block:: xml
```
Doctrine's:
```
.. code-block:: php
.. code-block:: yaml
.. code-block:: xml
```
[...]

PR #9161, opened 2021-10-30
Adding Attributes code block

ThomasLandauer:
Just that there is some real-world example somewhere ;-) see https://github.com/doctrine/orm/issues/9020#issuecomment-955582801

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -825,7 +825,19 @@ An entity listener is a lifecycle listener class used for an entity.
 
 .. configuration-block::
 
-    .. code-block:: php
+    .. code-block:: attribute
+    
+        <?php
+        namespace MyProject\Entity;
+        use App\EventListener\UserListener;
+
+        #[Entity]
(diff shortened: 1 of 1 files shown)


docs/en/reference/events.rst
```diff
@@ -825,7 +825,19 @@ An entity listener is a lifecycle listener class used for an entity.
...
+        namespace MyProject\Entity;
+        use App\EventListener\UserListener;
+
+        #[Entity]
+        #[EntityListeners([UserListener::class])]
+        class User
+        {
+            // ....
+        }
+    .. code-block:: annotation
```
Nek-, line 840: This defines the code coloration and therefore will not work.

ThomasLandauer: Don't know what you mean :-( What I'm trying to create is basically the same as in https://github.com/doctrine/doctrine-website/pull/409

Nek-: mb it looks supported!

derrabus (approved):

PR #9168, opened 2021-11-01
Fix discriminatorColumn phpdoc

VincentLanglet:
Hi @greg0ire, since the property is not initialized, the phpdoc should allow `null`.
This avoid a phpstan error if we're doing an `isset` check on the property.

The other solution is to initialize the value to `[]` but tests are failing

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -12,7 +12,6 @@ use RuntimeException;
 
 use function array_keys;
 use function array_search;
-use function assert;
 use function count;
 use function in_array;
 use function key;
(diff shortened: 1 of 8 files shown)


VincentLanglet:
Psalm is failing because `null` check are not done in the ORM code.
Is the value always set ?

greg0ire:
Hi Vincent!

> Is the value always set ?

Maybe not always, but as soon as there is inheritance, it probably is:

https://github.com/doctrine/orm/blob/6f194eeabf8b16b898d54de70023a89a8d3cda13/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php#L200-L213

VincentLanglet:
I added some assert checks to fix the psalm build.
The others failing tests are not related, since it's already failing on the 2.10.x branch.

greg0ire:
Looks like the failure is due to https://github.com/doctrine/dbal/commit/317c05620 . We are asserting against the generated SQL :grimacing:

greg0ire:
Blocked by https://github.com/doctrine/orm/pull/9181

VincentLanglet:
Thanks, tests are fine now @greg0ire if you find some time to review :)

greg0ire (commented):
All these `assert` calls look a bit clumsy, repetitive, and are a code smell…
How about introducing a `final public function discriminatorColumn()` in `ClassMetadataInfo` that performs the assertion, or better, throws a `LogicException` saying that this class probably does not have inheritance mapping?

greg0ire (approved):

SenseException (approved):

PR #9176, opened 2021-11-04
Fine-tuning Licecycle Callbacks codeblock

ThomasLandauer:
* Deleting "Not needed for XML and YAML mapping" - this was stupid of me, since *all* annotations are obviously not needed in XML&YAML ;-)
* Shortening the `@Column` annotation, for consistency with the following event handlers
* Removing some blank lines from XML, for consistency with YAML
* Adding PHP Attributes


SenseException (approved):

SenseException:
Thank you.

PR #9180, opened 2021-11-06
[Documentation]: Adding example for Scalar Column Hydration

ThomasLandauer:


docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -1276,21 +1276,32 @@ You can use the ``getSingleScalarResult()`` shortcut as well:
 Scalar Column Hydration
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-If you have a query which returns a one-dimensional array of scalar values
-you can use scalar column hydration:
+If your query returns just **one column**, the result looks something like this:
```
SenseException:
```suggestion
If your query returns **one column**, the result may look like this:
```

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -1276,21 +1276,32 @@ You can use the ``getSingleScalarResult()`` shortcut as well:
...
-    $query = $em->createQuery('SELECT a.id FROM CmsUser u');
-    $ids = $query->getResult(Query::HYDRATE_SCALAR_COLUMN);
+    [
+        ['id' => 5],
+        ['id' => 12],
+        ['id' => 23],
+    ]
 
-You can use the ``getSingleColumnResult()`` shortcut as well:
+In this case, you can use scalar column hydration to "flatten" it into a
```
SenseException:
```suggestion
In this case, you can use the scalar column hydration to reduce the result into a
```

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -1276,21 +1276,32 @@ You can use the ``getSingleScalarResult()`` shortcut as well:
...
 
 .. code-block:: php
 
     <?php
+    use Doctrine\ORM\AbstractQuery;
+
+    $query = $entityManager->createQuery('SELECT u.id FROM User u');
     $ids = $query->getSingleColumnResult();
+    // Same as:
+    $ids = $query->getResult(AbstractQuery::HYDRATE_SCALAR_COLUMN);
```
SenseException:
```suggestion
    <?php

    $query = $entityManager->createQuery('SELECT u.id FROM User u');
    $ids = $query->getSingleColumnResult();

or

.. code-block:: php

    <?php
    use Doctrine\ORM\AbstractQuery;

    $query = $entityManager->createQuery('SELECT u.id FROM User u');
    $ids = $query->getResult(AbstractQuery::HYDRATE_SCALAR_COLUMN);
```

SenseException (changes requested):

greg0ire:
@bhushan please review

docs/en/reference/dql-doctrine-query-language.rst
ThomasLandauer: You mean because `use Doctrine\ORM\AbstractQuery;` is not needed in the first case? Is this really worth an entire code block?

bhushan: Yes, it makes sense to have separate block, to be more clear when it is needed.

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -1276,21 +1276,32 @@ You can use the ``getSingleScalarResult()`` shortcut as well:
...
-    $query = $em->createQuery('SELECT a.id FROM CmsUser u');
-    $ids = $query->getResult(Query::HYDRATE_SCALAR_COLUMN);
+    [
+        ['id' => 5],
+        ['id' => 12],
+        ['id' => 23],
+    ]
 
-You can use the ``getSingleColumnResult()`` shortcut as well:
+In this case, you can use the scalar column hydration to reduce the result into a
```
bhushan, line 1289:
```suggestion
In this case, you can use the single scalar column hydration to reduce the result into a
```

bhushan:
@ThomasLandauer could you please also add description, why do you think these changes are needed in docs? I see we have kind of pattern in explaining hydration modes https://github.com/doctrine/orm/blob/2.10.x/docs/en/reference/dql-doctrine-query-language.rst#hydration-modes here. It is already quite simple.

bhushan (changes requested):
@ThomasLandauer could you please also add description, why do you think these changes are needed in docs? I see we have kind of pattern in explaining hydration modes https://github.com/doctrine/orm/blob/2.10.x/docs/en/reference/dql-doctrine-query-language.rst#hydration-modes here. It is already quite simple.

bhushan:
only showing examples of the results makes sense to me.

ThomasLandauer:
Yeah, indeed there is a pattern - but I'm trying to improve it! ;-)
Main reason: If you don't already know *in advance* what this is about, then the term "Single Scalar Hydration" is so awkward that it tells you zero. (Before you ask: I don't have a better idea, cause the "thing" is complicated.) So it's best explained by showing a full example.
BTW: The same is true for all other hydration modes; so if you agree, I'd be happy to add a full example for them too!

bhushan:
Adding examples how to use as clearly as possible makes sense to me.. but having consistency in docs also makes sense to me.. so may be writing ✍️ it in more simple words is nice for all hydrators 👍🏻 

The only thing is i am not sure about the outcome yet, how it will look actually.. could you update PR with what you have in mind, so it is much more clear and it can be reviewed.

Cc @greg0ire @SenseException

ThomasLandauer:
Questions:

1. Scalar Hydration (=new for me): The only difference I'm seeing (to array hydration) is the prefix (`u_`) in the array keys?! So why is it described as "flat rectangular result set"?
2. What do you think of adding PHPStan-like docblocks: `/** var array<int, User> $users */`?

General:
With the help of @SenseException I'm currently overhauling https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/events.html Here are some suggestions for this page:

* Since queries is one of the most important topics, I would start this page (or even create a new page) with an overview about the 3 query methods (DQL, Query Builder, Native SQL), integrating https://github.com/doctrine/orm/blob/2.10.x/docs/en/reference/faq.rst#doctrine-query-language. Recommendation: Use DQL for simple stuff; Query Builder for dynamically creating a query; Native SQL for vendor-specific stuff.

* Those two sections are basically saying the same and should be merged:
  * https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/dql-doctrine-query-language.html#query-result-formats
  * https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/dql-doctrine-query-language.html#hydration-modes

* https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/dql-doctrine-query-language.html#dql-select-examples should me merged with https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/dql-doctrine-query-language.html#dql-functions, and what doesn't fit there should be integrated into some other "fitting" chapter (e.g. "Joins")

* FAQ: In the note at https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/query-builder.html#the-querybuilder there's a broken link to FAQs. I found https://github.com/doctrine/orm/blob/2.10.x/docs/en/reference/faq.rst, but is it actually displayed somewhere on the website?

PR #9186, opened 2021-11-11
Fixed array hydration

### LINKED ISSUE #9185: Array hydrator returns null values instead of actual elements in one-to-many relationships
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.2

#### Summary

The issue is associated with DBMS (MariaDB, in my case) returning same joined rows in different order for separate parent rows. When row ordering is 'random', it may cause `ArrayHydrator` mapping problems, which results in null values in nested arrays of one-to-many relationships. Check the **"Current behavior"** section, it will be easier to understand by example.

#### Current behavior

Assume that we have a result set of these rows:
| a__id | a__topic | u__id | u__username | p__phonenumber
| - | - | - | - | -
| 1 | Article 1 | 1 | LinasRam | 1234567890
| 1 | Article 1 | 1 | LinasRam | 9876543210
| 2 | Article 2 | 1 | LinasRam | 9876543210
| 2 | Article 2 | 1 | LinasRam | 1234567890

We fetch two articles, which belong to the same user and also join user's phone numbers. The key point is that for second article rows, phone numbers are in different order than they were for the first one.
Array hydrated results of such dataset will return:
```
[
    [
        'id' => 1,
        'topic' => 'Article 1',
        'user' => [
            'id' => 1,
            'username' => 'LinasRam',
            'phonenumbers' => [
                ['phonenumber' => '1234567890'],
                ['phonenumber' => '9876543210'],
            ],
        ],
    ],
    [
        'id' => 2,
        'topic' => 'Article 2',
        'user' => [
            'id' => 1,
            'username' => 'LinasRam',
            'phonenumbers' => [
                ['phonenumber' => '9876543210'],
                null, // <--- The problem is this null
            ],
        ],
    ],
];
```
As you can see, in the second article there is a `null` phone number and it is a bug.

#### How to reproduce

Added a pull request with a unit test and a possible solution to this bug: #9186 

#### Expected behavior

The expected result of my provided example should look like this:
```
[
    [
        'id' => 1,
        'topic' => 'Article 1',
        'user' => [
            'id' => 1,
            'username' => 'LinasRam',
            'phonenumbers' => [
                ['phonenumber' => '1234567890'],
                ['phonenumber' => '9876543210'],
            ],
        ],
    ],
    [
        'id' => 2,
        'topic' => 'Article 2',
        'user' => [
            'id' => 1,
            'username' => 'LinasRam',
            'phonenumbers' => [
                ['phonenumber' => '9876543210'],
                ['phonenumber' => '1234567890'],
            ],
        ],
    ],
];
```

LinasRam:
Resetting `$this->_identifierMap` of base element children each time the new to-many relation is added to base element.

Includes a was-failing-but-now-passing test case.

fixes #9185

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
@@ -1,0 +1,114 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Internal\Hydration\ArrayHydrator;
+use Doctrine\ORM\Query\ResultSetMapping;
+use Doctrine\Tests\Mocks\ArrayResultFactory;
+use Doctrine\Tests\Models\CMS\CmsArticle;
+use Doctrine\Tests\Models\CMS\CmsPhonenumber;
+use Doctrine\Tests\Models\CMS\CmsUser;
+use Doctrine\Tests\OrmTestCase;
+
+class GH9185Test extends OrmTestCase
+{
+    /**
+     * Sometimes, the way DBMS is ordering rows is kinda random.
+     * This 'randomness' caused ArrayHydrator to return null values in one-to-many relationships
+     * instead of correct elements.
+     *
+     * select a.id, a.topic, u.id, u.username, p.phonenumber
+     * form CmsArticle a
+     * join a.user u
+     * join u.phonenumbers p
+     */
+    public function testMultipleFetchJoinWithInconsistentRowOrder(): void
+    {
+        $rsm = new ResultSetMapping();
+
+        $rsm->addEntityResult(CmsArticle::class, 'a');
+        $rsm->addJoinedEntityResult(
+            CmsUser::class,
+            'u',
+            'a',
+            'user'
+        );
+        $rsm->addJoinedEntityResult(
+            CmsPhonenumber::class,
+            'p',
+            'u',
+            'phonenumbers'
+        );
+        $rsm->addFieldResult('a', 'a__id', 'id');
+        $rsm->addFieldResult('a', 'a__topic', 'topic');
+        $rsm->addFieldResult('u', 'u__id', 'id');
+        $rsm->addFieldResult('u', 'u__username', 'username');
+        $rsm->addFieldResult('p', 'p__phonenumber', 'phonenumber');
+
+        $resultSet = [
+            [
+                'a__id' => '1',
+                'a__topic' => 'Article 1',
+                'u__id' => '1',
+                'u__username' => 'LinasRam',
+                'p__phonenumber' => '1234567890',
+            ],
+            [
+                'a__id' => '1',
+                'a__topic' => 'Article 1',
+                'u__id' => '1',
+                'u__username' => 'LinasRam',
+                'p__phonenumber' => '9876543210',
+            ],
+            [
+                'a__id' => '2',
+                'a__topic' => 'Article 2',
+                'u__id' => '1',
+                'u__username' => 'LinasRam',
+                'p__phonenumber' => '9876543210', // Different order than in previous rows
+            ],
+            [
+                'a__id' => '2',
+                'a__topic' => 'Article 2',
+                'u__id' => '1',
+                'u__username' => 'LinasRam',
+                'p__phonenumber' => '1234567890', // Different order than in previous rows
+            ],
+        ];
+
+        $stmt     = ArrayResultFactory::createFromArray($resultSet);
+        $hydrator = new ArrayHydrator($this->getTestEntityManager());
+        $result   = $hydrator->hydrateAll($stmt, $rsm);
+
+        $expected = [
+            [
+                'id' => 1,
+                'topic' => 'Article 1',
+                'user' => [
+                    'id' => 1,
+                    'username' => 'LinasRam',
+                    'phonenumbers' => [
+                        ['phonenumber' => '1234567890'],
+                        ['phonenumber' => '9876543210'],
+                    ],
+                ],
+            ],
+            [
+                'id' => 2,
+                'topic' => 'Article 2',
+                'user' => [
+                    'id' => 1,
+                    'username' => 'LinasRam',
+                    'phonenumbers' => [
+                        ['phonenumber' => '9876543210'],
+                        ['phonenumber' => '1234567890'],
+                    ],
+                ],
+            ],
+        ];
+
+        self::assertSame($expected, $result);
+    }
+}
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
@@ -15,12 +15,12 @@ use Doctrine\Tests\OrmTestCase;
 class GH9185Test extends OrmTestCase
 {
     /**
-     * Sometimes, the way DBMS is ordering rows is kinda random.
+     * Sometimes, DBMS is ordering rows randomly.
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
```diff
@@ -0,0 +1,114 @@
...
+
+class GH9185Test extends OrmTestCase
+{
+    /**
+     * Sometimes, the way DBMS is ordering rows is kinda random.
+     * This 'randomness' caused ArrayHydrator to return null values in one-to-many relationships
+     * instead of correct elements.
+     *
+     * select a.id, a.topic, u.id, u.username, p.phonenumber
+     * form CmsArticle a
```
bhushan:
```suggestion
     * from CmsArticle a
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
```diff
@@ -0,0 +1,114 @@
...
+use Doctrine\Tests\Mocks\ArrayResultFactory;
+use Doctrine\Tests\Models\CMS\CmsArticle;
+use Doctrine\Tests\Models\CMS\CmsPhonenumber;
+use Doctrine\Tests\Models\CMS\CmsUser;
+use Doctrine\Tests\OrmTestCase;
+
+class GH9185Test extends OrmTestCase
+{
+    /**
+     * Sometimes, the way DBMS is ordering rows is kinda random.
```
bhushan:
```suggestion
     * Sometimes, the way DBMS is ordering rows randomly.
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
```diff
@@ -0,0 +1,114 @@
...
+     * This 'randomness' caused ArrayHydrator to return null values in one-to-many relationships
+     * instead of correct elements.
```
bhushan:
```suggestion
     * instead of correct values.
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
```diff
@@ -0,0 +1,114 @@
...
+                ],
+            ],
+            [
+                'id' => 2,
+                'topic' => 'Article 2',
+                'user' => [
+                    'id' => 1,
+                    'username' => 'LinasRam',
+                    'phonenumbers' => [
+                        ['phonenumber' => '9876543210'],
```
bhushan, line 105: JFI asking:

these values were returning null before right?

bhushan (changes requested):

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
```diff
@@ -0,0 +1,114 @@
...
+                'u__id' => '1',
+                'u__username' => 'LinasRam',
+                'p__phonenumber' => '9876543210',
+            ],
+            [
+                'a__id' => '2',
+                'a__topic' => 'Article 2',
+                'u__id' => '1',
+                'u__username' => 'LinasRam',
+                'p__phonenumber' => '9876543210', // Different order than in previous rows
```
bhushan:
```suggestion
                'p__phonenumber' => '9876543210', // Different order of phone number than in previous rows
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
```diff
@@ -0,0 +1,114 @@
...
+            ],
+            [
+                'a__id' => '2',
+                'a__topic' => 'Article 2',
+                'u__id' => '1',
+                'u__username' => 'LinasRam',
+                'p__phonenumber' => '1234567890', // Different order than in previous rows
```
bhushan:
```suggestion
                'p__phonenumber' => '1234567890', // Different order of phone number than in previous rows
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9185Test.php
```diff
@@ -0,0 +1,114 @@
...
+class GH9185Test extends OrmTestCase
+{
+    /**
+     * Sometimes, the way DBMS is ordering rows is kinda random.
+     * This 'randomness' caused ArrayHydrator to return null values in one-to-many relationships
+     * instead of correct elements.
+     *
+     * select a.id, a.topic, u.id, u.username, p.phonenumber
+     * form CmsArticle a
+     * join a.user u
```
bhushan, line 24:
```suggestion
     * join a.users u
```

bhushan (dismissed):
@greg0ire please review.. looks good to me

greg0ire (commented):
2 questions: 

1. Why does this fix the bug? I understand what the bug is (great explanation BTW), but I don't understand why it happens.
2. Is this going to have consequences in terms of performance? I think the hydration is in the hot path of `doctrine/orm`, and I think resetting an array means loosing data. We have a benchmark to test that kind of thing: https://github.com/doctrine/orm/blob/2.10.x/tests/Doctrine/Performance/Hydration/SimpleQueryArrayHydrationPerformanceBench.php

LinasRam:
1. It's difficult to explain... You could comment out my code change and run my written test with debugger. Pay attention at `ArrayHydrator:135` when the phone number mapping is being set. After setting 'Article 1' phone number elements, you will get `$this->_identifierMap` values (I'm only adding `u.p` relationship here):
```
[
    'u.p' => [
        '|1' => [
            '|1234567890' => '0',
            '|9876543210' => '1',
        ],
    ],
]
```
So far, everything works as expected, but then comes 'Article 2' phone numbers in different order than before. If `$this->_identifierMap['u.p']['|1']` values are not reset, algorithm gets array index of `9876543210` number from mapping. Index is `1`, but on `line 135` `|9876543210` value in `$this->_identifierMap` will be overridden to `0` and mapping will look like this:
```
[
    'u.p' => [
        '|1' => [
            '|1234567890' => '0',
            '|9876543210' => '0',
        ],
    ],
]
```
Two array elements with different keys, but same values. That will cause unexpected behavior in `updateResultPointer` method called on `line 154`. Then, on `line 268` if statement will be valid with `$index = 1`, but `$coll` will only have one element as the only the first phone number was set till this point (second phone number will be set on next iteration):
```
$coll = [
    0 => [
        'phonenumber' => '9876543210',
    ],
];
```
Because of that, on `line 269` `$coll[1]` is treated as `null` and this is the line where those null values are being set.
So, resetting `$this->_identifierMap` of base element children each time the new to-many relation is added to base element solves this problem, because new elements are not getting previous element indexes. It may not be the best solution, but it works. I am not deeply familiar with Doctrine internals.
Also, it may not be any easier to understand the bug after my explanation, but I tried my best.

2. I have run the benchmark, didn't notice performance difference between current and modified version.

LinasRam:
Also, while trying to explain this bug right now, I noticed another problem. If you changed the order of article rows in my test case (order by ids: 1, 2, 1, 2), results would be interesting, too.

greg0ire:
Thanks for your explanation! I can't say it helps me because I too am not very familiar with that part of the code, but I will try to get the right people to review this.

beberlei:
Does ObjectHydrator have similar code to handle this already? I find it strange this never came up before.

I am confused why its nested, since you dont fetch the object but properties. Can you turn this into a real DQL query as per your comment? I am wondering if the problem is just a wong ResultSetMapping

LinasRam:
Yeah, query didn't correspond to what I was trying to accomplish. I updated it, I want nested results.

LinasRam:
Is anybody looking at this issue? I don't want it to be forgotten.

Issue #4772, opened 2015-09-24
DDC-3914: [GH-1515] [DDC-3146] Remove event listener from abstract hydrator in cleanup function


doctrinebot:
Jira issue originally created by user @doctrinebot:

This issue is created automatically through a Github pull request on behalf of emiel-silverstreet:

Url: https://github.com/doctrine/doctrine2/pull/1515

Message:

When the iterate function is invoked an event listener is added, but never removed. This will keep a reference to the abstract hydrator in the event manager, which will prevent the php garbage collector from removing the instance. When used multiple times this will create a memory leak because the event manager will be filled with references to unused hydrators. This fix will remove the event listener in the cleanup function which is invoked when the hydrator is finished, allowing the class to be garbage collected.

PR #9188, opened 2021-11-15
Introduce 'literal-string' to QueryBuilder::where($predicates)

craigfrancis:
Both [Psalm 4.8](https://github.com/vimeo/psalm/releases/tag/4.8.0) and [PHPStan 0.12.97](https://github.com/phpstan/phpstan/releases/tag/0.12.97) support the `literal-string` type.

This is useful for method arguments that expect the input string to be defined by the developer (in the source code), allowing mistakes like these to be easily identified via Static Analsys:

```php
$result = $qb
  ->select('u')
  ->from('User', 'u')
  ->where('u.id = ' . $_GET['id']); // INSECURE
```

While `literal-string` can be used by other methods, I want to start with `where()`, to see if we get any feedback (which I'm happy to help with), and *if* it goes well, I can carefully/slowly update other methods (I don't want to cause issues for developers using Doctrine, but I do want to stop Injection Vulnerabilities being introduced by the thousands of developers who can easily use Doctrine incorrectly).

We're able to change the type from `mixed` to `literal-string|object|array`, because the `where()` method calls `$this->add()` and that already uses `string|object|array $dqlPart` (we can look at the object/array inputs at a future date).

The `literal-string` type does support string concatenation, to reduce false positives.

It will get developers to correctly use `->setParameter()` for values (e.g. not relying on the integer type for safety, as noted by Grégoire Paris) - because user values should always be parameterised, so they can be sent to the database separately (database/driver permitting).

For user defined identifiers, comparison operators, etc - while these will hopefully be rare, they should *already* be using an allow-list approach, like the following, and both of these work with `literal-string` (I'm using `$_GET` to clearly show the un-trusted/un-safe values):

```php
$fields = [
  'name' => 'u.full_name',
  'email' => 'u.email_address',
  'address' => 'u.postal_address',
];

$field = ($fields[$_GET['field']] ?? 'u.full_name');

// or

$fields = ['name', 'address', 'email'];

$field_id = array_search($_GET['field'], $fields);

$field = $fields[$field_id];

// Used with

$qb
  ->select('u')
  ->from('User', 'u')
  ->where($field . ' = :my_value')
  ->setParameter('my_value', $_GET['value']);
```

I would still consider providing a [setIdentifier() method](https://github.com/doctrine/orm/pull/9047), but I believe Sergei Morozov is already intending to add "support for handling identifiers as first-class citizens" via [#4357](https://github.com/doctrine/dbal/issues/4357) and [#4772](https://github.com/doctrine/dbal/issues/4772) (this isn't necessary, but it might help some edge cases).

Because this is being done by Static Analysis (how Sergei wanted to implement this check), you can still tell Psalm or PHPStan to ignore any errors - i.e. for those who have been using this method incorrectly, and don't have the time to fix their code.

Ref [Issue 8933](https://github.com/doctrine/orm/issues/8933).

### DIFF
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -1090,7 +1090,6 @@ class QueryBuilder
      * </code>
      *
      * @param                 string|object|array $predicates The restriction predicates.
-     * @psalm-param   literal-string|object|array $predicates
      * @phpstan-param literal-string|object|array $predicates
      *
      * @return $this
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -1089,7 +1089,9 @@ class QueryBuilder
      *         ->where($or);
      * </code>
      *
-     * @param mixed $predicates The restriction predicates.
+     * @param                 string|object|array $predicates The restriction predicates.
+     * @psalm-param   literal-string|object|array $predicates
+     * @phpstan-param literal-string|object|array $predicates
      *
      * @return $this
      */


lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -1076,7 +1076,9 @@ public function set($key, $value)
      *         ->where($or);
      * </code>
      *
-     * @param mixed $predicates The restriction predicates.
+     * @param                 string|object|array $predicates The restriction predicates.
+     * @psalm-param   literal-string|object|array $predicates
```
SenseException: Doesn't Psalm also read the phpstan annotations? You can probably omit the psalm one as it might be redundant.

SenseException (changes requested):

lib/Doctrine/ORM/QueryBuilder.php
craigfrancis: Yes, you're right, [psalm does read phpstan annotations](https://github.com/vimeo/psalm/blob/af8033adf037017c7088da773ec6b8bb7711d741/src/Psalm/Internal/Scanner/DocblockParser.php#L231) (I had no idea, thanks). While it is technically redundant at the moment, maybe it makes it clear that it's supported by both psalm and phpstan; it's also possible psalm could stop doing this (although, if that does happen, we can probably move to using `@psalm literal-string` for everyone, [when phan supports it](https://github.com/phan/phan/issues/4618)?). For reference, Mark Scherer preferred to list both for Propel2.

What would you prefer?

SenseException: For maintenance it's easier to have only one of them, which is the phpstan one.

craigfrancis: That's fair, @psalm-param has been removed, and a quick test confirms it still works.

SenseException (dismissed):

derrabus:
I still have mixed feelings about this because passing a concatenated string to this method is not inherently wrong from my PoV. But I won't block merging it if others think it's a good idea.

However, this is not a bugfix, so I'd suggest to target 2.12.x instead.

craigfrancis:
Thanks @derrabus, but just to confirm, concatenated literal-strings are fine (I specially asked the static analysis tools to allow that, to help adoption). It’s only when a non-literal string (e.g. a user value) is included does it complain (as it should), where developers should be using `setParameter()`, or something like the identifier examples above.

And I’m happy to target any version; you’re right it’s not a big fix, I’m mostly wanting to be sure it doesn’t cause any problems.

craigfrancis:
To test, [Doctrine extensions for PHPStan (1.2.10)](https://github.com/phpstan/phpstan-doctrine/releases/tag/1.2.10), when using [bleedingEdge](https://phpstan.org/blog/what-is-bleeding-edge), will check `QueryBuilder::where()` only accepts `literal-string|object|array<mixed>` for `$predicates` . Like this patch, it won't check object/array inputs have used a `literal-string`, but it should give us some feedback.

PR #9191, opened 2021-11-18
Add strict-sync option to ValidateSchemaCommand

alexndlm:
When removing ManyToMany relationship a command `orm:validate-schema` does not handle that need remove a table, which handling this relation, from DB.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -258,14 +258,16 @@ class SchemaValidator
     /**
      * Checks if the Database Schema is in sync with the current metadata state.
      *
+     * @param bool    $strictSync If TRUE, check that all tables in a database have a mapping files
+     *
      * @return bool
      */
-    public function schemaInSyncWithMetadata()
+    public function schemaInSyncWithMetadata($strictSync = false)
     {
         $schemaTool = new SchemaTool($this->em);
 
         $allMetadata = $this->em->getMetadataFactory()->getAllMetadata();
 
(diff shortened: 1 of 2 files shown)


derrabus:
Sorry, I don't understand the problem that you're trying to solve. Can you provide an example?

Also, please add tests that cover your change. This might also help understand what you're trying to achieve.

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -258,14 +258,16 @@ public function validateClass(ClassMetadataInfo $class)
     /**
      * Checks if the Database Schema is in sync with the current metadata state.
      *
+     * @param bool $strictSync If TRUE, check that all tables in a database have a mapping files
+     *
      * @return bool
      */
-    public function schemaInSyncWithMetadata()
+    public function schemaInSyncWithMetadata($strictSync = false)
```
derrabus, line 266: Adding a new parameter to a non-final method is a breaking change. We cannot do that.

alexndlm:
Command `\Doctrine\ORM\Tools\Console\Command\ValidateSchemaCommand` doesn't check if all tables in DB has mapped.

If I delete ManyToMany relationship and forgot to create a migration, this command still says that all is ok, but a table that handles this relation still exist in DB.

In this PR I add the option to strictly check that all tables in DB has mapped.

lib/Doctrine/ORM/Tools/SchemaValidator.php
alexndlm: Maybe you can do this in the next major release?

PR #9195, opened 2021-11-20
Use `equal to` instead of `equal of` in `assertSqlGeneration()`

simPod:


greg0ire:
Restarting the build after the merge up

greg0ire:
:thinking: 🤔 🤔 Why are you contributing this to 3.0.x and not 2.10.x?

simPod:
I had 3.0 checked out. changed.

derrabus (approved):

greg0ire:
I think the right order for this was switch branch, and then force push, Github CI looks confused…

PR #9198, opened 2021-11-22
Add a psalm type for field mapping

### LINKED ISSUE #9193: Field mapping definitions are different in ClassMetadataInfo
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | >2.7.X

#### Summary

Not sure that it is really a bug but psalm annotations are defined differently for the same element. So phpstan is considering it as an error.

#### Current behavior

In https://github.com/doctrine/orm/blob/2.10.x/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L451, `fieldMappings` property is defined with:
```
     * @psalm-var array<string, array{
     *      type: string,
     *      fieldName: string,
     *      columnName?: string,
     *      length?: int,
     *      id?: bool,
     *      nullable?: bool,
     *      columnDefinition?: string,
     *      precision?: int,
     *      scale?: int,
     *      unique?: string,
     *      inherited?: class-string,
     *      originalClass?: class-string,
     *      originalField?: string,
     *      quoted?: bool,
     *      requireSQLConversion?: bool,
     *      declared?: class-string,
     *      declaredField?: string,
     *      options: array<mixed>
     * }>
```
But in https://github.com/doctrine/orm/blob/2.10.x/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L1303, `getFieldMapping `method is defined to return:
```
     * @psalm-return array{
     *      type: string,
     *      fieldName: string,
     *      columnName?: string,
     *      inherited?: class-string,
     *      nullable?: bool,
     *      originalClass?: class-string,
     *      originalField?: string,
     *      scale?: int,
     *      precision?: int,
     *      length?: int
     * }
```

#### How to reproduce

We are currently working on fixing issues due to upgrading phpstan to 1.x on API Platform, those lines (https://github.com/api-platform/core/blob/main/src/Doctrine/Orm/Metadata/Property/DoctrineOrmPropertyMetadataFactory.php#L80):
```
$fieldMapping = $doctrineClassMetadata->getFieldMapping($property);
$propertyMetadata = $propertyMetadata->withDefault($fieldMapping['options']['default'] ?? $propertyMetadata->getDefault());
```
generate this error in phpstan: 
```
Offset 'options' on array{type: string, fieldName: string, columnName?: string, inherited?: class-string, nullable?: bool, originalClass?: class-string, originalField?: string, scale?: int, ...} on left side of ?? does not exist.   
```

#### Expected behavior

From my point of view but I may be wrong, @psalm-return annotation should be the same as in property definition.

I may provide a PR to change this if it is needed.

laryjulien:
Field mapping have different definitions
in property definition and method return.
As suggested in issue and to avoid further desynchronization,
a psalm type has been created.
Fixes #9193

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -70,6 +70,26 @@ use const PHP_VERSION_ID;
  *
  * @template-covariant T of object
  * @template-implements ClassMetadata<T>
+ * @psalm-type FieldMapping = array{
+ *      type: string,
+ *      fieldName: string,
+ *      columnName?: string,
+ *      length?: int,
+ *      id?: bool,
+ *      nullable?: bool,
+ *      columnDefinition?: string,
+ *      precision?: int,
+ *      scale?: int,
+ *      unique?: string,
+ *      inherited?: class-string,
+ *      originalClass?: class-string,
+ *      originalField?: string,
+ *      quoted?: bool,
+ *      requireSQLConversion?: bool,
+ *      declared?: class-string,
+ *      declaredField?: string,
+ *      options?: array<string, mixed>
+ * }
  */
 class ClassMetadataInfo implements ClassMetadata
 {
@@ -427,26 +447,7 @@ class ClassMetadataInfo implements ClassMetadata
      * Whether a unique constraint should be generated for the column.
      *
      * @var mixed[]
-     * @psalm-var array<string, array{
-     *      type: string,
-     *      fieldName: string,
-     *      columnName?: string,
-     *      length?: int,
-     *      id?: bool,
-     *      nullable?: bool,
-     *      columnDefinition?: string,
-     *      precision?: int,
-     *      scale?: int,
-     *      unique?: string,
-     *      inherited?: class-string,
-     *      originalClass?: class-string,
-     *      originalField?: string,
-     *      quoted?: bool,
-     *      requireSQLConversion?: bool,
-     *      declared?: class-string,
-     *      declaredField?: string,
-     *      options: array<mixed>
-     * }>
+     * @psalm-var array<string, FieldMapping>
      */
     public $fieldMappings = [];
 
@@ -1285,18 +1286,7 @@ class ClassMetadataInfo implements ClassMetadata
      * @param string $fieldName The field name.
      *
      * @return mixed[] The field mapping.
-     * @psalm-return array{
-     *      type: string,
-     *      fieldName: string,
-     *      columnName?: string,
-     *      inherited?: class-string,
-     *      nullable?: bool,
-     *      originalClass?: class-string,
-     *      originalField?: string,
-     *      scale?: int,
-     *      precision?: int,
-     *      length?: int
-     * }
+     * @psalm-return FieldMapping
      *
      * @throws MappingException
      */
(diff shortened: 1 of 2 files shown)


greg0ire:
@laryjulien thanks! Both Psalm and Phpstan now detect 7 issues with keys that are accessed without checking first if they might be null… for each of them, I see 3 possibilities:
- we `assert` that it is not null;
- we check that it is not null, throw a custom exception if it is;
- we ignore the error in the baseline of both tools.

I'm not sure what's best, but I think it mostly depends on whether the issue could actually happen or not.

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -712,18 +715,7 @@ private function joinColumnToArray(Mapping\JoinColumn $joinColumn): array
...
-     *                   scale: int,
-     *                   length: int,
-     *                   unique: bool,
-     *                   nullable: bool,
-     *                   precision: int,
-     *                   options?: mixed[],
-     *                   columnName?: string,
-     *                   columnDefinition?: string
-     *               }
+     * @psalm-return FieldMapping
```
azjezz: this is a BC break IMHO, previous type states that `fieldName`, `type`, `scale`, `length`, `unique`, `nullable`, and `precision` are always available, while the new type states that they might not exist in the returned array.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1534,7 +1532,7 @@ private function validateAndCompleteTypedAssociationMapping(array $mapping): arr
     /**
      * Validates & completes the given field mapping.
      *
-     * @psalm-param array<string, mixed> $mapping The field mapping to validate & complete.
+     * @psalm-param FieldMapping $mapping The field mapping to validate & complete.
```
azjezz: This is actually the BC break, as the other method is private.

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
greg0ire: Oh right, didn't catch that, why did you change this @laryjulien ?

laryjulien: I have changed my PR to only use FieldMapping type in the initial context of issue.
I suggest to open an other PR to use this type in a wider context. Indeed, there are small variations on field mapping definition which appear to be more complex than I thought at first glance.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -70,6 +70,26 @@
...
+ *      scale?: int,
+ *      unique?: string,
+ *      inherited?: class-string,
+ *      originalClass?: class-string,
+ *      originalField?: string,
+ *      quoted?: bool,
+ *      requireSQLConversion?: bool,
+ *      declared?: class-string,
+ *      declaredField?: string,
+ *      options: array<mixed>
```
```suggestion
 *      options?: array<string, mixed>
```

derrabus (commented):
Thank you very much for working on this issue. Just today I performed an ORM update on a codebase and I suddenly had lots of false-positive Psalm errors because of those inconsitently documented array shapes. Looking forward to seeing this merged. 👍🏻

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
laryjulien: Updated

greg0ire (approved):

greg0ire:
Thanks @laryjulien !

PR #9202, opened 2021-11-24
Add SchemaIgnoreClasses property

### LINKED ISSUE #8195: Schema exclude annotation

simPod:
Supersedes #8203
Fixes #8195

### DIFF
--- a/docs/en/reference/advanced-configuration.rst
+++ b/docs/en/reference/advanced-configuration.rst
@@ -440,6 +440,19 @@ That will be available for all entities without a custom repository class.
 The default value is ``Doctrine\ORM\EntityRepository``.
 Any repository class must be a subclass of EntityRepository otherwise you got an ORMException
 
+Ignoring entities (***OPTIONAL***)
+-----------------------------------
+
+Specifies the Entity FQCNs to ignore.
+SchemaTool will then skip these (e.g. when comparing schemas).
+
+.. code-block:: php
+
+    <?php
+    $config->setSchemaIgnoreClasses([$fqcn]);
+    $config->getSchemaIgnoreClasses();
+
+
 Setting up the Console
 ----------------------
 
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -1012,4 +1012,24 @@ class Configuration extends \Doctrine\DBAL\Configuration
     {
         $this->_attributes['defaultQueryHints'][$name] = $value;
     }
+
+    /**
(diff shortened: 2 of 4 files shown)


simPod:
@greg0ire rebased, diff looks the same, CI is green. I guess it's k

greg0ire:
@simPod doesn't this require any documentation? Also, is it really a bugfix? Looks more like a new feature to me, in which case you should retarget to 2.11.x

simPod:
It was already a feature in #8203 😄 I'll rebase and retarget then.

SenseException (changes requested):
https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/reference/advanced-configuration.html would need the additional config.

docs/en/reference/advanced-configuration.rst
```diff
@@ -440,6 +440,21 @@ That will be available for all entities without a custom repository class.
...
+SchemaTool will then skip these (e.g. when comparing schemas).
+
+.. code-block:: php
+
+    <?php
+    $config->setSchemaIgnoreClasses([$fqcn]);
+    $config->getSchemaIgnoreClasses();
+
+The default value is ``Doctrine\ORM\EntityRepository``.
+Any repository class must be a subclass of EntityRepository otherwise you got an ORMException
```
SenseException: I'm confused with this part of the explanation. It's talking about fqcn of entities, but here it references repositories.

simPod: Wrong copy paste, forgot to remove these lines. Fixed, thx

SenseException (approved):

beberlei (approved):

beberlei:
@simPod thanks!

Issue #8465, opened 2021-02-07
Deprecate YamlDriver, use Xml, Annotations, Attributes instead


beberlei:
The YAML Driver is deprecated and all users are encouraged to migrate to XML Driver instead for configuration that aims at separating code from mapping in different files.

YAML has no internal validation mechanism like XML has with XSD files and as such there is no validation in IDEs and only partial validation during the runtime of YAML driver. Especially typos in keys cannot be detected, so its possible to configure something that has no effect, because the config key does not actually exist.

You can migrate all your YAML to XML mappings by using the `orm:convert:mapping` command (for the Doctrine 2 CLI), `doctrine:mapping:convert` for the Symfony CLI, or directly using programmatic APIs:

```php
$metadata = $entityManager->getClassMetadataFactory()->getAllMetadata();
$factory = new ClassMetadataExporter();
$exporter = $factory->getExporter('xml', $destinationDirectory);
$exporter->setMetadata($metadata);
$exporter->export();
```
Please note that metadata exporter are also deprecated and removed in ORM 3 and they might not cover all available mappings 100%.

SerafimArts:
> YAML has no internal validation mechanism like XML has with XSD files and as such there is no validation in IDE

I understand that it is already late and the decision (about deprecation) has already been made... but.

Json Schema? All yaml files are validated (and auto-completed in the IDE) based on the Json Schema.

We get approximately the following situation (reasons: advantages and disadvantages):

**XSD**
- Formats: `*.xml`
- ✔️/❌ - PHP builtin support (expects `ext-xml`/`ext-dom` additional extension).
- ✔️/❌ - Partial IDE autocompletion.
- ✔️ - IDE validation.

**Json Schema**
- Formats: `*.yaml`, `*.toml`, `*.json`, `*.json5`, static `*.php`, probably `*.neon` (did not check for autocomplete support) and other...
- ❌ - External composer package required.
- ✔️ - IDE autocompletion (Including a description of each field, what it is used for and how it can be applied).
- ✔️ - IDE validation.

imho, a unified schema validation format will allow the implementation of arbitrary metadata drivers (except XML and annotations/attributes).

P.S. For example, this is what autocomplete and validation looks like in the IDE if you select a `composer.json` scheme for the yaml file.

![image](https://github.com/doctrine/orm/assets/2461257/330b3874-b9c1-4c57-bcf9-3daa1897c8fd)

PR #9206, opened 2021-11-28
Add missing deprecations for YAML metadata mapping

derrabus:
Follows #8465

I've deprecated:
* `SimplifiedYamlDriver`
* `setup::createYAMLMetadataConfiguration()`

The class and the method don't make much sense if we remove YAML metadata mapping.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Setup.php
+++ b/lib/Doctrine/ORM/Tools/Setup.php
@@ -13,6 +13,7 @@ use Doctrine\Common\Cache\Psr6\CacheAdapter;
 use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\Common\Cache\RedisCache;
 use Doctrine\Common\ClassLoader;
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Configuration;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Tools/Setup.php
```diff
@@ -121,6 +124,12 @@ public static function createXMLMetadataConfiguration(array $paths, $isDevMode =
      */
     public static function createYAMLMetadataConfiguration(array $paths, $isDevMode = false, $proxyDir = null, ?Cache $cache = null)
     {
+        Deprecation::trigger(
+            'doctrine/orm',
+            'https://github.com/doctrine/orm/issues/8465',
+            'YAML mapping driver is deprecated and will be removed in Doctrine ORM 3.0, please migrate to annotation or XML driver.'
```
```suggestion
            'YAML mapping driver is deprecated and will be removed in Doctrine ORM 3.0, please migrate to attribute or XML driver.'
```

derrabus: This message is C&P. Shall I adjust the other occurences as well?

greg0ire: That would be great 👍

greg0ire (approved):

PR #9214, opened 2021-12-01
Merge 2.7 into 2.10.x

### LINKED ISSUE #4257: DDC-3449: Single scalar Result and Hidden field

### LINKED ISSUE #7512: findBy broken for inherited entity property lookups

greg0ire:
2.7 should never have existed, things that have been merged into it have been lost (and were never released).

Fixes #4257 
Fixes #7512
Closes #7514

### DIFF
--- a/docs/en/reference/dql-doctrine-query-language.rst
+++ b/docs/en/reference/dql-doctrine-query-language.rst
@@ -718,7 +718,7 @@ clauses:
 -  ``SQRT(q)`` - Return the square-root of q.
 -  ``SUBSTRING(str, start [, length])`` - Return substring of given
    string.
--  ``TRIM([LEADING \| TRAILING \| BOTH] ['trchar' FROM] str)`` - Trim
+-  ``TRIM([LEADING | TRAILING | BOTH] ['trchar' FROM] str)`` - Trim
    the string by the given trim char, defaults to whitespaces.
 -  ``UPPER(str)`` - Return the upper-case of the given string.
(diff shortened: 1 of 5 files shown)


greg0ire:
I need to rebase this

greg0ire:
@Mediagone please review

greg0ire:
ping @Mediagone @beberlei

SenseException (approved):
I can recognize #8340, plus coding style and analysis fixes.
#7514 is also there, even though comparing `BasicEntityPersister` with the original one was more to read because of the early exit changes with the `return`.

PR #9233, opened 2021-12-10
Support MEMBER OF on entities with association IDs

Xesau:
Currently, SqlWalker assumes that entities in a MEMBER OF have a regular field as join column. When an association is used as join, an Undefined index is thrown because $class->fieldNames['association join column name'] does not exist. This commit fixes this by looking up the column name of the relevant join column instead.

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -2002,7 +2002,24 @@ class SqlWalker implements TreeWalker
             $sqlParts    = [];
 
             foreach ($joinColumns as $joinColumn) {
-                $targetColumn = $this->quoteStrategy->getColumnName($class->fieldNames[$joinColumn['referencedColumnName']], $class, $this->platform);
+                $referencedColumnName = $joinColumn['referencedColumnName'];
+
+                // when the referenced column is an association
+                if (!isset($class->fieldNames[$referencedColumnName])) {
+                    foreach ($class->associationMappings as $associationMapping) {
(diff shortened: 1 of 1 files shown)


SenseException:
I'm not sure if I understand the use case. Can you please cover this `MEMBER OF` logic with tests?

Xesau:
Assume we have the following entities:

- Person, with fields ID, name, and a ManyToMany "movies" to Movie
- Movie, with fields ID, name, but _no_ inverse ManyToMany "people" to Person

Now, on a query selecting `Person p`, you can do `LEFT JOIN Movie m WITH m MEMBER OF p.movies`. In [SqlWalker.php](https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/Query/SqlWalker.php#L2144), it is assumed that the primary key of Person is a regular field, so it works fine. However, let's say we have the following configuration of entities

- Person, with fields ID, name
- PersonOptions, with as ID a ManyToOne to Person, and a ManyToMany "movies" to Movie
- Movie, with fields ID, name, but _no_ inverse ManyToMany "people" to PersonOptions

If you select on `Person p`, `LEFT JOIN PersonOptions o WITH o.person = p`, you should be able to do `LEFT JOIN Movie m WITH m MEMBER OF o.movies`. However, because the `referencedColumnName` in the ManyToMany (person_id) is not a regular field but a field generated by the ManyToOne, this gives a Notice: Undefined index. This PR fixes this notice by checking if the referencedColumnName exists in an association if it is not a field.

PR #9236, opened 2021-12-11
Restore `query()` call for primary/replica connections

derrabus:
On DBAL 2, `PrimaryReadReplicaConnection::query()` executed the query on the primary connection. I broke the usage of this "hidden feature" while removing deprecated calls. This PR restores the functionality. At the same time, it removes mock logic that was supposed to catch my mistake but did not succeed doing so.

### DIFF
--- a/lib/Doctrine/ORM/Id/FetchOneFromPrimary.php
+++ b/lib/Doctrine/ORM/Id/FetchOneFromPrimary.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Id;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
+use Doctrine\DBAL\ForwardCompatibility\Result as ForwardCompatibilityResult;
+use Doctrine\DBAL\Result;
+
+/**
+ * @internal
+ */
+trait FetchOneFromPrimary
+{
+    /**
+     * @return string|int|false
+     */
+    private function fetchOneFromPrimary(Connection $connection, string $sql)
+    {
+        // Using `query` to force usage of the master server in PrimaryReadReplicaConnection
+        $result = $connection instanceof PrimaryReadReplicaConnection
+            ? $connection->query($sql)
+            : $connection->executeQuery($sql);
+        if (! $result instanceof Result) {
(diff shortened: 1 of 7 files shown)


stof:
@derrabus the (deprecated) UuidGenerator is also affected by the same issue (it was also using `query` for that reason)

derrabus:
Fixed as well!

lib/Doctrine/ORM/Id/FetchOneFromPrimary.php
```diff
@@ -0,0 +1,36 @@
...
+ * @internal
+ */
+trait FetchOneFromPrimary
+{
+    /**
+     * @return string|int|false
+     */
+    private function fetchOneFromPrimary(Connection $connection, string $sql)
+    {
+        // Using `query` to force usage of the master server in PrimaryReadReplicaConnection
```
greg0ire, line 22:
```suggestion
        // Using `query` to force usage of the primary server in PrimaryReadReplicaConnection
```

greg0ire (dismissed):

beberlei:
I  have an alternative here: https://github.com/doctrine/orm/pull/9239

lib/Doctrine/ORM/Id/FetchOneFromPrimary.php
```diff
@@ -0,0 +1,36 @@
...
+     * @return string|int|false
+     */
+    private function fetchOneFromPrimary(Connection $connection, string $sql)
+    {
+        // Using `query` to force usage of the primary server in PrimaryReadReplicaConnection
+        $result = $connection instanceof PrimaryReadReplicaConnection
+            ? $connection->query($sql)
+            : $connection->executeQuery($sql);
+        if (! $result instanceof Result) {
+            ForwardCompatibilityResult::ensure($result);
```
stof: this must use the return value here, otherwise it is useless

derrabus: Of course, good catch. 🙈

derrabus:
@beberlei Your alternative targets 2.11. Shall we merge this PR anyway to fix the regression on 2.10?

stof:
@derrabus the alternative version could be applied on 2.10 if that's the selected solution IMO.

derrabus:
If we allow that kind of change in a bugfix release, okay.

derrabus:
Closing in favor of #9239

PR #9237, opened 2021-12-11
Remove YAML metadata mapping

derrabus:


UPGRADE.md
```diff
@@ -1,5 +1,10 @@
 # Upgrade to 3.0
 
+## BC BREAK: Remove YAML mapping drivers.
+
+If your code relies on `YamlDriver` or `SimpleYamlDriver`, you **MUST** migrate to
```
beberlei, line 5: You could point to the `orm:convert-mapping` command here as well:

```
php doctrine orm:convert-mapping xml /path/to/mapping-path-converted-to-xml
```

UPGRADE.md
```diff
@@ -1,5 +1,10 @@
...
+attributem annotation or XML drivers instead.
```
norkunas:
```suggestion
attribute, annotation or XML drivers instead.
```

UPGRADE.md
derrabus: Good idea.

SenseException (dismissed):

greg0ire (approved):

PR #9239, opened 2021-12-11
Make PrimaryReadReplicaConnection enforcement explicit

beberlei:
Due a breaking API change with assumptions of `PrimaryReadReplicaConnection` in DBAL 3 without the distinction of `query` and `executeStatement` it is required to address the use of primary replica connection more explicitly in ORM.

This isn't the most pretty solution, but it detects the use of primary replica connection at the right points: 1. id generation 2. commit 3. DQL update/delete statements.

### DIFF
--- a/lib/Doctrine/ORM/Id/SequenceGenerator.php
+++ b/lib/Doctrine/ORM/Id/SequenceGenerator.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Id;
 
+use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\ORM\EntityManager;
 use Serializable;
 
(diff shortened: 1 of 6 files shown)


beberlei:
Just realized the primary also needs to be enforced when select statements are made with lock levels. This was a flaw previously too

beberlei:
Actually, `PrimaryReadReplicaConnection::beginTransaction` switches to the primary and `wrapInTransaction` (and `transactional`) as well as any call using a pessimistic lock start or require an open transaction. The only benefit would be to make this more explicit.

stof:
@beberlei should be done in 2.10.x IMO

lib/Doctrine/ORM/Id/SequenceGenerator.php
```diff
@@ -57,6 +58,10 @@ public function generate(EntityManager $em, $entity)
             $conn = $em->getConnection();
             $sql  = $conn->getDatabasePlatform()->getSequenceNextValSQL($this->_sequenceName);
 
+            if ($conn instanceof PrimaryReadReplicaConnection) {
+                $conn->ensureConnectedToPrimary();
+            }
+
             // Using `query` to force usage of the master server in MasterSlaveConnection
```
stof: this comment should be removed, as this is not accurate anymore

stof (approved):

PR #9241, opened 2021-12-12
Support for nesting attributes with PHP 8.1

### LINKED ISSUE #9240: Support for nesting attributes with PHP 8.1
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

With PHP 8.1 attributes can now be nested. This allows to add some of the missing functionality that we have missed on the Attribute Driver support for PHP 8.0:

- `#[AttributeOverrides([new AttributeOverride(name: "foo", column: new Column(type: "string"))])` and the same for `#[AssociationOverride]`.
- `#[JoinTable(joinColumns: [new JoinColumn], inverseJoinColumns: [new JoinColumn])`
- `#[Table(indexes: [new Index()], uniqueConstraints: [new UniqueConstraint()])]`

Follow up to #8266 

This ticket replaces #8772 and #9174

beberlei:
This adds a new mapping case that are possible because of nesting attributes in PHP 8.1

- Using `#[AssociationOverrides]` and `#[AttributeOverrides]`

We initially thought to also add nested supported for `#[JoinTable(joinColumns: ...)]` and `#[Table(indexes: ...)]` as is possible with annotaitons. However during documentation and test adjustment found that its getting super clumsy and complicated.

Fixes #9240


lib/Doctrine/ORM/Mapping/AssociationOverrides.php
```diff
@@ -4,18 +4,42 @@
...
     /**
      * Mapping overrides of relationship properties.
      *
      * @var array<\Doctrine\ORM\Mapping\AssociationOverride>
      */
-    public $value;
+    public $overrides = [];
+
+    /**
+     * @param array|AssociationOverride $overrides
```
greg0ire:
```suggestion
     * @param AssociationOverride[]|AssociationOverride $overrides
```

lib/Doctrine/ORM/Mapping/AttributeOverrides.php
```diff
@@ -4,18 +4,42 @@
...
     /**
      * One or more field or property mapping overrides.
      *
      * @var array<\Doctrine\ORM\Mapping\AttributeOverride>
      */
-    public $value;
+    public $overrides = [];
+
+    /**
+     * @param array|AttributeOverride $overrides
```
greg0ire:
```suggestion
     * @param AttributeOverride[]|AttributeOverride $overrides
```

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
```diff
@@ -397,12 +417,66 @@ public function loadMetadataForClass($className, ClassMetadata $metadata): void
             }
         }
 
+        // Evaluate AssociationOverrides annotation
```
greg0ire:
```suggestion
        // Evaluate AssociationOverrides attribute
```

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
```diff
@@ -397,12 +417,66 @@ public function loadMetadataForClass($className, ClassMetadata $metadata): void
...
+        if (isset($classAttributes[Mapping\AssociationOverrides::class])) {
+            $associationOverride = $classAttributes[Mapping\AssociationOverrides::class];
+
+            foreach ($associationOverride->overrides as $associationOverride) {
+                $override  = [];
+                $fieldName = $associationOverride->name;
+
+                // Check for JoinColumn/JoinColumns annotations
```
greg0ire: `#[JoinTable(inverseJoinColumns: [], joinColumns: [])]`

Here, there is no annotation, and `JoinTable` is an attribute. How should we name `inverseJoinColumns` and `joinColumns`? The docs seem to be referring to those as … attributes: https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/attributes-reference.html#column

Should we use "parameters" here instead?

There is the same issue in other comments below.

lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -941,4 +944,17 @@ public static function invalidUniqueConstraintConfiguration($className, $indexNa
...
+    /**
+     * @param mixed $givenValue
+     * @return self
+     */
+    public static function invalidOverrideType(string $expectdType, $givenValue)
+    {
+        return new self(sprintf(
+            'Expected %s, but %s was given.',
+            $expectdType,
+            is_object($givenValue) ? get_class($givenValue) : gettype($givenValue)
```
greg0ire:
```suggestion
            get_debug_type($givenValue)
```

Since we have the polyfill

lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -941,4 +944,17 @@ public static function invalidUniqueConstraintConfiguration($className, $indexNa
...
+
+    /**
+     * @param mixed $givenValue
+     * @return self
+     */
+    public static function invalidOverrideType(string $expectdType, $givenValue)
+    {
+        return new self(sprintf(
+            'Expected %s, but %s was given.',
+            $expectdType,
```
greg0ire:
```suggestion
     */
    public static function invalidOverrideType(string $expectedType, $givenValue): self
    {
        return new self(sprintf(
            'Expected %s, but %s was given.',
            $expectedType,
```

greg0ire:
Is this worth documenting anywhere?

SenseException:
That would be https://github.com/doctrine/orm/blob/2.11.x/docs/en/reference/attributes-reference.rst is this stays the target branch.

lib/Doctrine/ORM/Mapping/AssociationOverrides.php
derrabus: The code actually requires an array at the moment.

derrabus:
I must admit, I like the flattened version of the attribute mapping that has been built for PHP 8.0. Why do we want to nest attributes again? 🤔

lib/Doctrine/ORM/Mapping/AssociationOverrides.php
greg0ire: Oh right, because of line 33, indeed!

beberlei:
@derrabus you mean for `JoinTable` and `Table`? I wouldn't mind either just keeping the 8.0 way, just was thinking about consistency. We do need the overrides though, but the other too could stay the way they were.

beberlei:
@greg0ire added docs now

tests/Doctrine/Tests/ORM/Mapping/YamlMappingDriverTest.php
```diff
@@ -83,6 +84,22 @@ public function testSpacesShouldBeIgnoredWhenUseExplode(): void
         self::assertEquals(255, $nameField['length']);
         self::assertEquals(255, $valueField['length']);
     }
+
+    public function testCompositeKeyForJoinTableInManyToManyCreation(): void
```
beberlei, line 88: This is unrelated, but made tests fail due to PHP 8,0 and no nesting. Saw the original `DDC3711Test.php` where this came from re-executed all YamlMappingDriver tests *again*, so this skips about 50 tests here that were duplicated.

tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
```diff
@@ -762,6 +762,10 @@ public function testSqlResultSetMapping(): void
      */
     public function testAssociationOverridesMapping(): void
     {
+        if (PHP_VERSION_ID >= 80000 && PHP_VERSION_ID < 80100) {
```
beberlei, line 765: This kind of condition is annoying, but required for some models we interact with. This skips all assocation/attribute override tests on PHP 8, but since we also run against 7.4 and 8.1 this is not a problem.

docs/en/reference/attributes-reference.rst
```diff
@@ -59,12 +148,12 @@ inside the instance variables PHP DocBlock comment. Any value hold
 inside this variable will be saved to and loaded from the database
 as part of the lifecycle of the instance variables entity-class.
 
-Required attributes:
+Required parameters:
```
beberlei, line 151: This renames are unrelated, but seemed sane to do, since naming the fields attributes on Attribute classes is redundant and confusing as @greg0ire pointed out.

derrabus (approved):

derrabus:
Shall we backport the documentation fixes to 2.10?

greg0ire (approved):

beberlei:
@derrabus Yes I can do that.

PR #9243, opened 2021-12-12
Removing list "Lifecycle Events"

ThomasLandauer:
As announced in https://github.com/doctrine/orm/pull/9184#issuecomment-965837780

I'm committing the events one by one, to make it easier to follow.

Notes:
* postUpdate, postRemove, postPersist aren't similar enough to have them in one section, so it might be better to separate them.
* This warning I don't understand. I moved it to `postRemove`. Should it be repeated on all "events triggered after an entity removal" (which ones?):
    > The ``postRemove`` event or any events triggered after an entity removal
    can receive an uninitializable proxy in case you have configured an entity to
    cascade remove relations. In this case, you should load yourself the proxy in
    the associated ``pre*`` event.

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -197,10 +197,7 @@ events during the life-time of their registered entities.
 
 
 
--  ``prePersist`` - The ``prePersist`` event occurs for a given entity
-   before the respective ``EntityManager`` persist operation for that
-   entity is executed. It should be noted that this event is only triggered on
-   *initial* persist of an entity (i.e. it does not trigger on future updates).   
+ 
 -  ``preUpdate`` - The ``preUpdate`` event occurs before the database
    update operations to entity data. It is not called for a DQL
    ``UPDATE`` statement nor when the computed changeset is empty.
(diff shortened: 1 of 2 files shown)


docs/en/reference/events.rst
```diff
@@ -187,94 +189,6 @@ several reasons:
...
-The ``EventArgs`` instance received by the listener gives access to the entity,
-``EntityManager`` instance and other relevant data.
-
-.. note::
-
-    All Lifecycle events that happen during the ``flush()`` of
-    an ``EntityManager`` have very specific constraints on the allowed
-    operations that can be executed. Please read the
-    :ref:`reference-events-implementing-listeners` section very carefully
-    to understand which operations are allowed in which lifecycle event.
```
SenseException, line 276: I think this note has a history and should be kept. Maybe below the table?

SenseException (changes requested):

docs/en/reference/events.rst
ThomasLandauer: Well, this entire page is about documenting which actions are allowed where. So adding a note like "Not everything is allowed everywhere." doesn't add any value IMO. Was the history that some users didn't understand what is allowed where? Then the solution is not to add an *unspecific* "Watch out, guys!" warning, but rather to exactly list all the constraints for each event. => And that's what I'm trying ;-)
Bottom line: If you really insist, I'll bring it back :-)

SenseException: It's a note with a reference to another part of the docs, so I think you have a point.

SenseException:
Great work. Please change your commit messages into meaningful ones as described in https://www.doctrine-project.org/contribute/index.html#crafting-meaningful-commit-messages.

derrabus:
We can (and should) squash the commits on merge, @SenseException.

ThomasLandauer:
If you squash them, should I still change something?

derrabus (approved):
I think, we're good. Thank you.

PR #9244, opened 2021-12-12
Put actual value instead of index inside $originalEntityData

### LINKED ISSUE #8217: Redundant UPDATE queries from the second flush() in a row
### Bug Report

Simple bi-directional OneToMany relation with auto-increment primary key makes UnitOfWork detect non-existing changes.
When performing two `flush()` calls in a row, subsequent flush results in redundant UPDATE queries.

|    Q        |   A
|------------ | ------
| BC Break    | not sure
| Version     | 2.10.2

#### Summary

Let's say we have a OneToMany relation between TestCollection and TestItem entities. Additionally, TestCollection has a single-column auto-increment primary key (IDENTITY GeneratedValue). We also have cascade persist and remove operations for items.

In a simplest scenario of creating a new collection with a single item
```php
$collection = new TestCollection();
$item = new TestItem($collection, 100);
$item->value = 'TEST';
$collection->items->set(100, $item);
$entity_manager->persist($collection);
$entity_manager->flush();
```
all seems fine, but if we add an additional `flush()` just after the first one, it results in the additional UPDATE like
```sql
UPDATE test_item SET collection_nid = 5 WHERE collection_nid = 5 AND item_nid = 100
```

#### Current behavior

Subsequent `flush()` operation results in a redundant UPDATE query. Internally, the UnitOfWork seems to detect a difference between `$originalData['collection']` which consist of an int(5) **(!)** and `$actualData['collection']` which is a proper entity.

#### How to reproduce

TestItem:
```php
<?php

/**
 * @ORM\Entity
 * @ORM\Table(name="test_item")
 */
class TestItem
{
	/**
	 * @ORM\Id
	 * @ORM\ManyToOne(targetEntity="TestCollection", inversedBy="items")
	 * @ORM\JoinColumn(name="collection_nid", referencedColumnName="collection_nid")
	 * @var TestCollection
	 */
	public $collection;

	/**
	 * @ORM\Id
	 * @ORM\Column(type="integer", options={"unsigned": true})
	 * @var int
	 */
	public $item_nid;

	/**
	 * @ORM\Column(type="string", nullable=false)
	 * @var string
	 */
	public $value = '';

	public function __construct(TestCollection $collection, int $item_nid)
	{
		$this->collection = $collection;
		$this->item_nid = $item_nid;
	}
}
```

TestCollection:
```php
<?php

/**
 * @ORM\Table(name="test_collection")
 * @ORM\Entity
 */
class TestCollection
{
	/**
	 * @var int|null
	 *
	 * @ORM\Column(name="collection_nid", type="integer", nullable=false, options={"unsigned": true})
	 * @ORM\Id
	 */
	public $collection_nid;

	/**
	 * @ORM\OneToMany(targetEntity="TestItem", indexBy="item_nid",
	 *     mappedBy="collection", cascade={"persist", "remove"}, orphanRemoval=true)
	 * @var Collection
	 */
	public $items;

	public function __construct()
	{
		$this->items = new ArrayCollection();
	}
}
```

Database structure with an additional trigger to detect redundant UPDATE:
```sql
CREATE TABLE test_collection (
  collection_nid INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY
);

CREATE TABLE test_item (
  collection_nid INT UNSIGNED NOT NULL,
  item_nid INT UNSIGNED NOT NULL,
  value TEXT NOT NULL,
  PRIMARY KEY (collection_nid, item_nid)
);

DELIMITER :
CREATE TRIGGER test_item_before_update BEFORE UPDATE ON test_item FOR EACH ROW BEGIN
IF NEW.value = OLD.value THEN
signal sqlstate '45000' set message_text = 'redundant update';
END IF;
END:
DELIMITER ;
```

#### Expected behavior

Subsequent `flush()` operation should perform no UPDATE queries.

Feolius:
This fixes a bug with redundant UPDATE queries, that are executed when some entity uses foreign index of other entity as a primary key. This happens when after inserting related entities with $em->flush() call, you do the second $em->flush() without changing any data inside entities.
Fixes #8217.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1159,14 +1159,16 @@ class UnitOfWork implements PropertyChangedListener
         $identifier = [];
 
         foreach ($class->getIdentifierFieldNames() as $idField) {
-            $value = $class->getFieldValue($entity, $idField);
+            $origValue = $class->getFieldValue($entity, $idField);
 
+            $value = null;
             if (isset($class->associationMappings[$idField])) {
                 // NOTE: Single Columns as associated identifiers only allowed - this constraint it is enforced.
-                $value = $this->getSingleIdentifierValue($value);
+                $value = $this->getSingleIdentifierValue($origValue);
             }
 
-            $identifier[$idField] = $this->originalEntityData[$oid][$idField] = $value;
+            $identifier[$idField]                     = $value ?? $origValue;
+            $this->originalEntityData[$oid][$idField] = $origValue;
         }
 
         $this->entityStates[$oid]      = self::STATE_MANAGED;
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8217Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH8217Test.php
@@ -1,0 +1,108 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\JoinColumn;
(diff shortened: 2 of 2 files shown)


derrabus:
Can you have a look at the failing tests?

Feolius:
Yes, sorry. They failed because I used wrong php language level (7.4 instead of 7.1). Now should be fine.

derrabus (approved):

Issue #9042, opened 2021-09-29
Paginator does not replace WHERE clause and returned data may be incomplete


LuigiCardamone:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7.3

#### Summary

If I have a query like this (using pseudo code for brevity):
`SELECT * FROM Author a JOIN Book b WHERE b.price = 10 `

the Paginator creates a first query to get the list of id:
```
SELECT DISTINCT id_0
FROM (SELECT DISTINCT id_0
      FROM (
          SELECT * FROM Author a JOIN Book b WHERE b.price = 10 
       ) dctrn_result_inner
      ORDER BY id_0 ASC) dctrn_result
LIMIT 10
```

Then the Paginator creates a second query passing the list of id fetched from first query:
```
SELECT * FROM Author a JOIN Book b 
WHERE b.price = 10 
AND a.id IN (1,2,3,4,5,6,7,8,9,10)
```

The question is, why is the entire WHERE clause not replaced with the "IN condition"?
Which are the reason of keeping the original WHERE clause?

Keeping the original WHERE clause has the following problem:
the Author entity will not have the complete list of his books but only the books with price equal to 10.

I think that this may be a bug since the comment on the class WhereInWalker say that
"Replaces the whereClause of the AST with a WHERE id IN (:foo_1, :foo_2) equivalent."
but the code do not replace but "append".

#### Current behavior

IN clause is appended to the original WHERE clause

#### How to reproduce

- Create any query with a WHERE condition
- Create the Paginator: `$paginator = new Paginator($query, true);`
- Use `$paginator->getIterator()` to get data
- Inspect the executed queries

#### Expected behavior

IN clause should replace the original WHERE clause

stof:
> Keeping the original WHERE clause has the following problem:
> the Author entity will not have the complete list of his books but only the books with price equal to 10.

And this is exactly what happens as well when fetching all results without paginating, because that's what your query is asking for.

So what you suggest here is **introducing** a bug in the paginator, not fixing one.

LuigiCardamone:
Thank you for your feedback.
If you reason with tabular results in mind you are right. If you reason with object results in mind I would expect to get objects that are always complete (an author with all his books, as when I query the author for Id).
Does my point of view make sense for you?

stof:
@LuigiCardamone what does not make sense is expecting a different behavior when you run the **same** DQL query in a paginated way or an unpaginated way.

[...]

PR #9252, opened 2021-12-15
[doctrineGH-9042] Paginator: replace Where clause with IN condition to avoid results with incomplete data

LuigiCardamone:
This PR adds an optional feature to fully load child collections as described in this issue: https://github.com/doctrine/orm/issues/9042
This implementation has been developed and tested for doctrine v2.7.3 but it should be compatible also with latest versions.


derrabus:
Please resolve the merge conflicts and provide a test for your change.

stof:
See my comment on the issue. To me, this PR is introducing the bug, not fixing it.

LuigiCardamone:
I changed the implementation. The result is an optional feature that is disabled by default and can be enabled by using an additional parameter in the Paginator constructor.
We are currently using this feature and we believe that can be useful to other users. If you are willing to merge this PR I can add the proper test cases.

stof:
even as an optional feature, this is still wrong to me. Write the right DQL query instead of making the paginator change the behavior of DQL

Issue #9230, opened 2021-12-07
Single column hydration returns nothing if first item is false-like


TomaszGasior:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.3 

#### Summary

While working on my open source app I found out that scalar column hydration does not return any value, just empty array, when first element of the results is false-like (null, 0, ''…). 

#### How to reproduce

Do something like this but without `<> NULL` and `<> ''`: https://github.com/TomaszGasior/RadioLista-v3/blob/8a2b4fa0d6d9fb85d83a36a00bd61d2031a1a8b3/src/Repository/RadioStationRepository.php#L68

Simplified example:

```php
return $this->createQueryBuilder('radioStation')
    ->select('DISTINCT radioStation.'.$column)
    ->getQuery()
    ->getSingleColumnResult()
;
```

If first item from result set is false-like, empty array is returned.

#### Expected behavior

I would like to get all values from the column, even something like null, ''.

Adding `false !==` to https://github.com/doctrine/orm/blob/6414ad4cbb898e4ae027a1a0de48b31984002dea/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php#L31 kinda fixes the issue. I don't know whether Doctrine is able to return native `false` for DBMS-es supporting it and how to support that case if it's needed.

greg0ire:
@bhushan please take a look at this :pray:

bhushan:
@TomaszGasior are you planning to PR fix?

bhushan:
@TomaszGasior https://github.com/doctrine/orm/pull/9255 does this test case makes sense or you meant something else ?

TomaszGasior:
@bhushan Unfortunately I am not able to provide PR for now. I consider change in ScalarColumnHydrator.php as debugging hint, maybe not final solution. Yep, provided test looks good (except for lack of string with zero char about which I added comment in the PR).

MrMitch:
Hi @greg0ire @bhushan @TomaszGasior. Thank you for you work on this. I've opened PR #9663 to fix this issue. It includes a test and the fix. Let me know if there is anything else I can do to help.

PR #9255, opened 2021-12-16
fix: issue-9230 test case

bhushan:
Written test case for #9230.

Here first of all, in all hydration we cancel falsy value, I want to understand first, **do we really want this ?**

also if we dont have values then `$this->statement()->fetchOne()` returns false, then any how we will be ignoring false value.

@greg0ire @TomaszGasior

### DIFF
--- a/tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
+++ b/tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
@@ -119,4 +119,34 @@ class ScalarColumnHydratorTest extends HydrationTestCase
 
         $hydrator->hydrateAll($stmt, $rsm);
     }
+
+    /**
+     * Select u.id from CmsUser u
+     */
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Hydration/ScalarColumnHydratorTest.php
```diff
@@ -119,4 +119,34 @@ public function testMultipleColumnEntityQueryThrowsException(): void
...
+        $rsm = new ResultSetMapping();
+
+        $rsm->addEntityResult(CmsUser::class, 'u');
+        $rsm->addFieldResult('u', 'u__id', 'id');
+        $rsm->addScalarResult('sclr0', 'id');
+        $rsm->addIndexByScalar('sclr0');
+
+        $resultSet = [
+            ['u__id' => ''],
+            ['u__id' => 0],
```
TomaszGasior, line 137: Maybe you can add '0' (string with zero char) also.

bhushan:
already merged changes

Issue #9095, opened 2021-10-05
Regression: abstract entities in the middle of a hierarchy need to be part of discriminator map


BenMorel:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.9.x, 2.10.x

#### Summary

As mentioned in https://github.com/doctrine/orm/issues/8771#issuecomment-933454924, there is a regression that affects both `2.9` and `2.10`:

If an entity hierarchy has an abstract entity class in the middle of the hierarchy, Doctrine complains that it must be part of the discriminator map. Example:

- abstract class `A`
    - abstract class `B`
        - class `C`

`bin/console doctrine:schema:validate` reports:

>  [FAIL] The entity-class B mapping is invalid:
> * Entity class 'B' is part of inheritance hierarchy, but is not mapped in the root entity 'A' discriminator map. All subclasses must be listed in the discriminator map.

This doesn't make sense, as there cannot be an entry in the database for an abstract entity class.

**This was working fine up to Doctrine `2.8`.**

The quick fix for those affected is to add "fake" entries to the discriminator map for abstract child entities, that do not have an equivalent in the database `ENUM`; which is rather ugly.

derrabus:
Do you want to work on a fix?

BenMorel:
@derrabus Yes! I opened a PR in #9096, building on top of #8903.

I'm dubious as to whether it makes sense to have both `! $class->reflClass->isAbstract()` and `! $class->isMappedSuperclass` now though: does it make sense to have a non-abstract `@MappedSuperclass`? Isn't it enough to check if the class is abstract?

derrabus:
Looks good to me. 🙂

BenMorel:
Be careful that this has been reverted in #9262.

mpdude:
#10420 suggests to fix this again; it should now be possible due to #10411.

PR #9262, opened 2021-12-17
Fix SchemaValidator with abstract child class in discriminator map - Enforcing Abstract Entity in DiscriminatorMap is important for some operations

### LINKED ISSUE #9142: Enforcing Abstract Entity in DiscriminatorMap is important for some operations
#### Summary
Recent fix introduced in 2.10.2 [9096: Fix SchemaValidator with abstract child class in discriminator map](https://github.com/doctrine/orm/pull/9096) is a regression to the SchemaValidator as having all `@Entity` classes, including those that are non-instantiable (abstract) is important for some operation.

@beberlei already mentioned it here https://github.com/doctrine/orm/issues/8736#issuecomment-853044536 

I also mentioned this in the original issue (https://github.com/doctrine/orm/issues/8771#issuecomment-893266118) regarding this change.

I'll try to prepare a convincing test case. 

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.2

#### Current behaviour
Abstract class marked with `@Entity` annotation does not have to be part of DiscriminatorMap (according to SchemaValidator)

#### Expected behavior
Only `@MappedSuperclass` does not have to be part of the SchemaValidator, missing `@Entity` is dangerous in some cases

olsavmic:
Resolves https://github.com/doctrine/orm/issues/9142

> Recent fix introduced in 2.10.2 9096: Fix SchemaValidator with abstract child class in discriminator map is a regression to the SchemaValidator as having all @Entity classes, including those that are non-instantiable (abstract) is important for some operation.

This reverts commit bbb68d007213188dc6008ca04b7167b79433c1a5.

https://github.com/doctrine/orm/pull/9145 demonstrates that even abstract classes are required to be present in discriminator map.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -239,7 +239,7 @@ class SchemaValidator
             }
         }
 
-        if (! $class->isInheritanceTypeNone() && ! $class->isRootEntity() && ! $class->reflClass->isAbstract() && ! $class->isMappedSuperclass && array_search($class->name, $class->discriminatorMap, true) === false) {
+        if (! $class->isInheritanceTypeNone() && ! $class->isRootEntity() && ! $class->isMappedSuperclass && array_search($class->name, $class->discriminatorMap, true) === false) {
             $ce[] = "Entity class '" . $class->name . "' is part of inheritance hierarchy, but is " .
                 "not mapped in the root entity '" . $class->rootEntityName . "' discriminator map. " .
                 'All subclasses must be listed in the discriminator map.';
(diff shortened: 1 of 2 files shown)


beberlei (approved):

BenMorel:
@beberlei Shouldn't we work on making it work without the abstract classes in the discriminator map, rather than accepting this limitation?

olsavmic:
@BenMorel This PR did not change the existing behaviour in any way, it's just fixing the validator so that it now validates what's actually required. This issue is resolved now.

Making it work with abstract classes in the discriminator map is a different question and it's a feature/improvement which deserves its own issue & eventually PR. I'd love to make/see such contribution but I don't know how to overcome the limitation mentioned in https://github.com/doctrine/orm/issues/9142#issuecomment-955155622

Btw I actually wanted to integrate the test cases from https://github.com/doctrine/orm/pull/9145 to this PR so that the issue is not reintroduced in the future, that's why I marked it as Draft. Thanks for merging anyway, I may do this in separately

mpdude:
Let's discuss in #10389 whether having all abstract classes is really requirement and what the consequences of that are.

Apart from that, the problem shown in #9145 can possibly be solved without the requirement, see #10387.

mpdude:
Since #10411 has been merged, it should be safe to re-apply the changes from #9095.

PR #9264, opened 2021-12-17
Catch exception from Connection::commit()

morozov:
`Connection::commit()` will return `void` in DBAL 4. See https://github.com/doctrine/dbal/pull/3480.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -9,6 +9,7 @@ use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\EventManager;
 use Doctrine\Common\Proxy\Proxy;
+use Doctrine\DBAL;
 use Doctrine\DBAL\LockMode;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Cache\Persister\CachedPersister;
@@ -433,11 +434,19 @@ class UnitOfWork implements PropertyChangedListener
                 }
             }
 
-            // Commit failed silently
-            if ($conn->commit() === false) {
+            $commitFailed = false;
+            try {
+                if ($conn->commit() === false) {
+                    $commitFailed = true;
+                }
+            } catch (DBAL\Exception $e) {
+                $commitFailed = true;
+            }
+
+            if ($commitFailed) {
                 $object = is_object($entity) ? $entity : null;
 
-                throw new OptimisticLockException('Commit failed', $object);
+                throw new OptimisticLockException('Commit failed', $object, $e ?? null);
             }
         } catch (Throwable $e) {
             $this->em->close();
--- a/lib/Doctrine/ORM/OptimisticLockException.php
+++ b/lib/Doctrine/ORM/OptimisticLockException.php
@@ -8,6 +8,7 @@ use DateTimeInterface;
 use Doctrine\ORM\Exception\ORMException;
 use Exception;
 use RuntimeException;
+use Throwable;
 
 /**
  * An OptimisticLockException is thrown when a version check on an object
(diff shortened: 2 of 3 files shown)


derrabus:
DBAL 3's version of `commit()` might also throw exceptions already. Would it make sense to apply this change to the 2.11.x branch in that case?

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -433,8 +434,16 @@ public function commit($entity = null)
...
+                    $commitFailed = true;
+                }
+            } catch (DBAL\Exception $e) {
+                $commitFailed = true;
+            }
+
+            if ($commitFailed) {
                 $object = is_object($entity) ? $entity : null;
 
                 throw new OptimisticLockException('Commit failed', $object);
```
derrabus: If `$e` is defined at this point, we should attach it as previous exception.

morozov:
> DBAL 3's version of `commit()` might also throw exceptions already.

It can but those won't be the wrapper-level exceptions being caught in this patch. It could be a driver exception, a PDO exception, or anything else but I don't think that catching `\Exception` is a good idea since it won't necessarily identify a failed or even attempted commit.

I believe that since DBAL 3 doesn't provide an API for handling commit exceptions, the ORM shouldn't be using it.

derrabus:
> It can but those won't be the wrapper-level exceptions.

I was mainly thinking about `ConnectionException` which _is_ a wrapper-level exception.

morozov:
> I was mainly thinking about `ConnectionException` which _is_ a wrapper-level exception.

This is really out of scope. I'm trying to improve the ORM compatibility with DBAL 4, not address the API design issues of the earlier DBAL versions.

derrabus:
Okay, but right now, the 3.0.x branch is compatible with DBAL 3 and we don't want to change that, do we? You would introduce a piece of logic that would catch exceptions that can be raised by DBAL 3 already.

At the moment, the ORM behaves differently for those exceptions than it does for a `false` return value. After your change, that difference would be gone. Is that what we want?

morozov:
I definitely do not intend to change any existing behavior. What's the case with commit and `ConnectionException` thrown by `commit()` where this patch would change the behavior of the ORM? Would it be a change towards more consistent or less consistent error handling?

derrabus:
Looking at the code of `Connection::commit()`, that exception is raised if the connection is in a state that would be really unexpected at this point:

```php
        if ($this->transactionNestingLevel === 0) {
            throw ConnectionException::noActiveTransaction();
        }

        if ($this->isRollbackOnly) {
            throw ConnectionException::commitFailedRollbackOnly();
        }
```

Not catching that exception feels like the right thing to do. If DBAL 4 raises an exception for cases that resulted in a `false` return previously, maybe we should only catch that exception and let everything else bubble up?

morozov:
These two exceptions being `ConnectionException` is wrong IMO. Purely from the common-sense perspective, a connection exception is a runtime exception that should be caught and possibly handled by retrying the operation. But these two are obviously logical exceptions and aren't meant to be caught.

I'm not really familiar with the logic of `UnitOfWork`, so I don't know how to proceed here. If you have any specific solution in mind, I can implement it.

derrabus:
> But these two are obviously logical exceptions and aren't meant to be caught.

Is there a way for a client that consumes a DBAL connection to distinguish between logic exceptions and runtime exceptions?

> I'm not really familiar with the logic of `UnitOfWork`, so I don't know how to proceed here.

The code that we're about to change seems to assume that if a commit fails, it is because of a deadlock or similar situation. The exceptions raised by DBAL 4 should give us more insights on the reason for the failure. My idea would be to only catch exceptions related to locking and let everything else bubble up.

derrabus:
Another thing: Later in that method, we have a generic catch block that attempts to clean up after a failed transaction:

https://github.com/doctrine/orm/blob/ad97969e9309f7128c4810971c6b9d48097852d6/lib/Doctrine/ORM/UnitOfWork.php#L442-L452

Note that the return value of the `rollBack()` call is ignored. I think this is on purpose: We attempt a rollback and if that fails, we simply move on because the initial exception is more valuable than a potential rollback failure. If `rollBack()` also throws instead of returning `false`, I think we need to catch&ignore possible exceptions here.

morozov:
I believe it's better to wait until we can run the test suite. Otherwise, we're shooting in the dark.

The major blocker for running the suite is compile-time errors caused by the hard-coded mocks (https://github.com/doctrine/orm/issues/8886#issuecomment-893391508).

PR #9267, opened 2021-12-18
Enable UnusedUse sniff again

derrabus:
Keeping the UnusedUse sniff deactivated entirely does more harm than good, I figured. The sniff is still not perfect when it comes to annotations, but if we keep annotations multiline, it works. I think, we can live with that.


tests/Doctrine/Tests/Models/DDC117/DDC117Article.php
```diff
@@ -20,7 +20,8 @@ class DDC117Article
 {
     /**
      * @var int
-     * @Id @Column(type="integer", name="article_id") @GeneratedValue
+     * @Id @Column(type="integer", name="article_id")
```
greg0ire: Is it deliberate to have this one stay single line?

beberlei:
Couldn't we ignore tests/ for this sniff only if it causes these problems with annotations?

tests/Doctrine/Tests/Models/DDC117/DDC117Article.php
derrabus: No, adjusted. 👍🏻

derrabus:
> Couldn't we ignore tests/ for this sniff only if it causes these problems with annotations?

We could, but why should we? Our test suite does not need to cover all edge cases of the `doctrine/annotations` library, so adhering to a syntax that CodeSniffer understands seems okay to me. I think we gain more from a rule that removes unused imports from the tests, especially if at the same time we require code to explicitly import elements from the global scope.

greg0ire (approved):

PR #9269, opened 2021-12-19
Add DQL non-ASCII characters support

julienfalque:
This PR adds support for non-ASCII characters usage in DQL.

My use case was to use the actual domain experts language in the code (french, so with some diacritics) instead of translating everything to english. The only obstable I faced was non-ASCII characters not being parsed correctly in DQL, though I only tried a few simple scenarios.

I don't know if not supporting non-ASCII characters was on purpose. If so, feel free to close this PR.

The provided naming strategies also need an update but I will do that in a subsequent PR if this one is accepted.

### DIFF
--- a/lib/Doctrine/ORM/Query/Lexer.php
+++ b/lib/Doctrine/ORM/Query/Lexer.php
@@ -127,11 +127,11 @@ class Lexer extends AbstractLexer
     protected function getCatchablePatterns()
     {
         return [
-            '[a-z_][a-z0-9_]*\:[a-z_][a-z0-9_]*(?:\\\[a-z_][a-z0-9_]*)*', // aliased name
-            '[a-z_\\\][a-z0-9_]*(?:\\\[a-z_][a-z0-9_]*)*', // identifier or qualified name
+            '[[:alpha:]_][[:alpha:]0-9_]*\:[[:alpha:]_][[:alpha:]0-9_]*(?:\\\[[:alpha:]_][[:alpha:]0-9_]*)*', // aliased name
+            '[[:alpha:]_\\\][[:alpha:]0-9_]*(?:\\\[[:alpha:]_][[:alpha:]0-9_]*)*', // identifier or qualified name
             '(?:[0-9]+(?:[\.][0-9]+)*)(?:e[+-]?[0-9]+)?', // numbers
             "'(?:[^']|'')*'", // quoted strings
-            '\?[0-9]*|:[a-z_][a-z0-9_]*', // parameters
+            '\?[0-9]*|:[[:alpha:]_][[:alpha:]0-9_]*', // parameters
         ];
     }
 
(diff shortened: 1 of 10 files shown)


DGCarramona:
Hello, just saw this PR and wandering if all RDBS support non ASCII characters like in your test ? Just like not using non ASCII characters in code, I've been always told to avoid non ASCII characters in database

julienfalque:
I tested this with MySQL and had no issue so far. I don't know about other vendors. But DQL is about PHP classes and properties. Even if they use non-ASCII characters, you can still map them to ASCII tables and columns names.

SenseException:
You can use integration tests to test your changes in form of an SQL with the different databases.

julienfalque:
@SenseException I added a test but I don't know if it passes with all platforms, can you please run GitHub Actions workflows?

lib/Doctrine/ORM/Query/Lexer.php
```diff
@@ -117,17 +117,31 @@ public function __construct($input)
         $this->setInput($input);
     }
 
+    /**
+     * Overrides parent method in case it does not return the `u` modifier.
+     *
+     * This is required in case `doctrine/lexer` < 1.2.0 is installed and can be
+     * removed as soon as such version cannot be installed anymore, likely when
+     * dropping PHP 7.1 support.
```
derrabus: I think, we should just allow Lexer 1.2 to be installed on PHP 7.1: doctrine/lexer#58

derrabus:
> I added a test but I don't know if it passes with all platforms

Apparently, MySQL does not like what you did. 😕

julienfalque:
Yeah, I said I tested this with MySQL but it was MariaDB actually 😅️. I'm investigating the failures.

lib/Doctrine/ORM/Query/Lexer.php
derrabus: Lexer 1.2 is installable on PHP 7.1 now. You can bump the dependency if it helps you.

julienfalque: Done.

mpdude:
Are you still pursuing this @julienfalque ?

julienfalque:
I need to rebase and fix conflicts. Hopefully tests will still pass.

julienfalque:
Just rebased. Can a maintainer approve running workflows please?

julienfalque:
Sorry, didn't notice the PR still targeted unmaintained branch `2.12.x`. Rebased onto `2.14.x`.

julienfalque:
I don't actually need this anymore but I would still like to see it merged.

SenseException:
I assume the next target branch for this would be `2.21.x `. /cc @greg0ire

greg0ire:
We don't accept new features on 2.x unless they improve forward compatibility the proper target is 3.4

PR #9281, opened 2021-12-21
Better explain limitations of DQL "DELETE"

mpdude:
We think the current documentation does not stress these details enough, so that they are easily overlooked.


derrabus (approved):

SenseException (approved):

SenseException:
Thank you @mpdude

PR #9282, opened 2021-12-21
Improve documentation on `cascade: remove` behavior

mpdude:
When you shoot yourself in the foot, at least document the consequences so that other people are better informed when making their choices.

### DIFF
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -515,9 +515,23 @@ Thanks to ``cascade: remove``, you can easily delete a user and all linked comme
     overhead, especially when the cascaded collections are large. Make sure
     to weigh the benefits and downsides of each cascade operation that you define.
 
-    To rely on the database level cascade operations for the delete operation instead, you can
-    configure each join column with :doc:`the onDelete option <working-with-objects>`.
+    If you can go without lifecycle events, you can rely on database-level foreign keys
+    with cascade operations instead. To do so, configure each join column with 
+    :doc:`the onDelete option <working-with-objects>` and remove the ``cascade: remove`` setting.
+    
+    Contrary to in-memory cascade operations, those at the database level are also applied 
+    when using :doc:`DQL DELETE statements<dql-doctrine-query-language#delete-queries>`.
 
+.. warning::
+
+    For many-to-many associations, ``cascade: remove`` means that the associated entities
+    will be removed. 
+    
+    Consider an example where you have a many-to-many association between
+    a ``Book`` and an ``Author`` entity. When you use ``cascade: remove`` on the ``Book``
+    side and remove a ``Book`` entity, the corresponding ``Author`` will be removed as
+    well – even if there are other ``Book`` entities referring to it.
+    
 Even though automatic cascading is convenient, it should be used
 with care. Do not blindly apply ``cascade=all`` to all associations as
 it will unnecessarily degrade the performance of your application.


derrabus (dismissed):

docs/en/reference/working-with-associations.rst
```diff
@@ -511,9 +511,23 @@ Thanks to ``cascade: remove``, you can easily delete a user and all linked comme
...
+    If you can go without lifecycle events, you can rely on database-level foreign keys
+    with cascade operations instead. To do so, configure each join column with 
+    :doc:`the onDelete option <working-with-objects>` and remove the ``cascade: remove`` setting.
+    
+    Contrary to in-memory cascade operations, those at the database level are also applied 
+    when using :doc:`DQL DELETE statements<dql-doctrine-query-language#delete-queries>`.
+
+.. warning::
 
+    For many-to-many associations, ``cascade: remove`` means that the associated entities
```
beberlei, line 527: The same is true for one-to-many though, are we sure this warning here is necessary? It repeats what the previous docs stated in other words. Maybe the words in the beginning of the section could need more precision: "Thanks to cascade: remove, you can easily ...."

mpdude: In a one-to-many association (e.g. person -> address) I find it natural that removing the person also removes their addresses.

In many-to-many, I was somehow surprised (although I should have known better) that „remove“ not only affects the n:m-table, but the entity at the other side of the association.

Would Doctrine even go back across the association? E. g. remove book -> remove author(s) -> remove other books by these authors?

SenseException: Maybe a code example of the Book <-> Author relation can help some readers to understand what's happening in the example with the configuration.

mpdude: Not sure how to best proceed here... suggestions anyone?

mpdude:
Do we agree on the paragraphs before the warning, and would it help if I split those out into a separate PR?

docs/en/reference/working-with-associations.rst
```diff
@@ -511,9 +511,23 @@ Thanks to ``cascade: remove``, you can easily delete a user and all linked comme
...
+    will be removed. 
+    
+    Consider an example where you have a many-to-many association between
+    a ``Book`` and an ``Author`` entitiy. When you use ``cascade: remove`` on the ``Book``
```
greg0ire:
```suggestion
    a ``Book`` and an ``Author`` entity. When you use ``cascade: remove`` on the ``Book``
```

mpdude:
Rebased onto 2.14.x

docs/en/reference/working-with-associations.rst
greg0ire: I'd use an example that is more clearly a many to many. While books sometimes have several authors, that's rarely the case. How about an example with a relationship between developers and projects? I must admit I'm a bit surprised by this behavior as well, but in hindsight, it does make sense: if you don't cascade, you're still going to remove n-m table entries, so of course cascade does a bit more.

I think @beberlei's suggestion might make the first sentence a bit too big and distracting, I wouldn't move all your warning to there, but I would  maybe change the tone to a more startling one:

```rst
``cascade:remove`` is to ``EntityManager::remove()`` what ``-rf`` is to ``rm``, so please read carefully.
```

mpdude:
I have done some more experiments with this.

#### Model:

`Student *<--attends-->* Course`

... with the extra assumption that when you database-level delete a student, the "attends" entry will be cascade-deleted. The other way round, however, you may not database-level delete a course that has students attending. All "attends" must  be removed before.

(Example: Business process that students must be notified that the course is cancelled; then, have their subscription removed; then the course can be removed. I know that not all universities operate that way.)

```php
/**
 * @ORM\Entity()
 * @ORM\HasLifecycleCallbacks()
 */
class Course
{
    /**
     * @ORM\Column(type="integer")
     * @ORM\Id()
     * @ORM\GeneratedValue()
     */
    private int $id;

    /**
     * @ORM\ManyToMany(targetEntity="Student", mappedBy="courses")
     */
    private Collection $students;

    /**
     * @ORM\Column
     */
    private string $name;

    public function __construct(string $name)
    {
        $this->name = $name;
        $this->students = new ArrayCollection();
    }

    public function subscribed(Student $student): void
    {
        $this->students->add($student);
    }

    /**
     * @ORM\PreRemove
     */
    public function preRemoveHandler(PreRemoveEventArgs $event): void
    {
        print "Removing course {$this->name}".PHP_EOL;
    }
}


/**
 * @ORM\Entity()
 * @ORM\HasLifecycleCallbacks()
 */
class Student
{
    /**
     * @ORM\Column(type="integer")
     * @ORM\Id()
     * @ORM\GeneratedValue()
     */
    private int $id;

    /**
     * @ORM\ManyToMany(targetEntity="Course", inversedBy="students")
     * @ORM\JoinTable(name="student_course",
     *     joinColumns={@ORM\JoinColumn(onDelete="cascade")},
     *     inverseJoinColumns={@ORM\JoinColumn(onDelete="restrict")}
     * )
     */
    private Collection $courses;

    /**
     * @ORM\Column
     */
    private string $name;

    public function __construct(string $name)
    {
        $this->name = $name;
        $this->courses = new ArrayCollection();
    }

    public function subscribe(Course $course): void
    {
        $this->courses->add($course);
        $course->subscribed($this);
    }

    /**
     * @ORM\PreRemove
     */
    public function preRemoveHandler(PreRemoveEventArgs $event): void
    {
        print "Removing student {$this->name}".PHP_EOL;
    }
}
```

Generated schema (for MySQL):

```sql
CREATE TABLE student (id INT AUTO_INCREMENT NOT NULL, name VARCHAR(255) NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE `utf8_unicode_ci` ENGINE = InnoDB;
CREATE TABLE student_course (student_id INT NOT NULL, course_id INT NOT NULL, INDEX IDX_98A8B739CB944F1A (student_id), INDEX IDX_98A8B739591CC992 (course_id), PRIMARY KEY(student_id, course_id)) DEFAULT CHARACTER SET utf8 COLLATE `utf8_unicode_ci` ENGINE = InnoDB;
CREATE TABLE course (id INT AUTO_INCREMENT NOT NULL, name VARCHAR(255) NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE `utf8_unicode_ci` ENGINE = InnoDB;
ALTER TABLE student_course ADD CONSTRAINT FK_98A8B739CB944F1A FOREIGN KEY (student_id) REFERENCES student (id) ON DELETE CASCADE;
ALTER TABLE student_course ADD CONSTRAINT FK_98A8B739591CC992 FOREIGN KEY (course_id) REFERENCES course (id) ON DELETE RESTRICT;
```

With that, let's have two students `Joe` and `Jane`, both attending `Data modeling 101`.

#### Observations 1

* The SQL looks like what I was expecting.
* I have no doubt the database level foreign keys work as defined, consistent with my description above.

#### Deleting the course

When I `EntityManager::remove()` the course, the ORM will process the `PreRemove` event and then
* remove the course from both Joe's and Jane's many-to-many collections
* sync the database, i. e. delete the two `student_course` rows
* delete the course in the database.

👉🏻 A database-level `ON DELETE RESTRICT` _does not_ prevent deletion of the `Course` entity through the ORM, since the ORM is smart enough to delete the association table entries beforehand.

#### Deleting a student

Nothing to see here. Deletes the `student_course` row for the student, keeps the other one. I don't know if the deletion happened explicitly or through the foreign key constraint, but for the time being, I don't care.

#### ORM-level cascade

Add `cascade={"remove"}` to both entities' `@ManyToMany` annotations. Perform `EntityManager::remove()` for one student and `flush()` changes.

The ORM will process `PreRemove`
* for the "other" student (not the one being removed)
* for the course
* for the student initially removed.

👉🏻 `cascade="remove"` will literally "expand" through the associations. It fetches associated entities, processes removals for them, and may even "bounce back" to the originating entity class. Net effect: Both students and the course are removed.

#### Lessons learned

* `cascade` operates on the association level, the (n,m)-join table is transparent
* `@JoinColumn(onDelete=...)` defines the behaviour at the database level for the colunns in the (n,m)-join table, i.e. the technical details of having the table are relevant
* ORM updates collections in a way that sidesteps database level constraints.

greg0ire:
Nice experiment! I think we should definitely have the words "chain reaction" in the docs :P

PR #9285, opened 2021-12-23
Fix XML export for `change-tracking-policy`

Fedik:
The `change-tracking-policy` is an attribute, not an Entity child.

Even thought the xml exporter is deprecated it still should produce valid output for 2.x series ;)


derrabus (approved):

derrabus:
> Even thought the xml exporter is deprecated it still should produce valid output for 2.x series ;)

The exporter might be depreacted, but since we also deprecated the YAML mapping, the XML exporter will have one last mission before we remove it. 🦸🏻 

Thank you for your contribution!

PR #9286, opened 2021-12-23
Whitelist composer plugins used by this repository

derrabus:
Since Composer 2.2, plugins need to be whitelisted. Otherwise Composer will keep asking about them.


greg0ire (approved):
Reference: https://getcomposer.org/doc/06-config.md#allow-plugins

PR #9290, opened 2021-12-27
[RFC] Drop support for PHP 7.4

derrabus:
I'd like to propose to raise the minimum PHP version for the 3.0.x branch to 8.0.

We have taken this step on the DBAL 4.0.x branch already. Having the full power of the PHP 8 type system will make it easier for us to create a codebase that is compatible with DBAL 4. Moreover, bumping to PHP 8 would allow us to use native attributes as a default for old and new functional tests. Last but not least, developing features for a PHP-8-only project is a really delightful experience. 🙂 

According to Packagist's statistics, [more than 50% of ORM 2.10 installations](https://packagist.org/packages/doctrine/orm/php-stats#2.10) run PHP 8 already.

I don't know exactly when the 3.0.x  branch will be ready for a release, but even if we released right now, I would expect the 2.x series of the ORM to live on for some time, so that projects may upgrade to the new major (and PHP 8) at their own pace.

### DIFF
--- a/.github/workflows/coding-standard.yml
+++ b/.github/workflows/coding-standard.yml
@@ -10,6 +10,6 @@ on:
 
 jobs:
   coding-standards:
-    uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.1.1"
+    uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.4.1"
     with:
-      php-version: "7.4"
+      php-version: "8.0"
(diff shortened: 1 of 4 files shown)


morozov (approved):

greg0ire (approved):
I have updated the branch protection settings, they seemed wrong.

beberlei (approved):

beberlei:
In addition before 3.0 is final, it will still take a few month, so by then 8.0 is even more wide spread.

PR #1346, opened 2015-03-23
Fix: generated IDs are converted to integer

pepakriz:
When I try to use type `bigint` and `@GeneratedValue` together, generated value is converted to integer.

See PHP documentation http://php.net/manual/en/language.types.array.php

> Strings containing valid integers will be cast to the integer type. E.g. the key "8" will actually be stored under 8. On the other hand "08" will not be cast, as it isn't a valid decimal integer.

PR #9291, opened 2021-12-27
Replace mocks with middlewares

derrabus:
This PR improves a test by replacing a mock of DBAL's `Connection` class by a middleware. This should make the test resilient against future DBAL changes and ease the midgration path towards DBAL 4.


composer.json
```diff
@@ -27,7 +27,7 @@
         "doctrine/cache": "^1.12.1 || ^2.1.1",
         "doctrine/collections": "^1.5",
         "doctrine/common": "^3.0.3",
-        "doctrine/dbal": "^3.2",
+        "doctrine/dbal": "^3.3@dev",
```
derrabus, line 30: I need the abstract middleware classes for now. We can drop the `@dev` as soon as DBAL 3.3 is released.

greg0ire: This is not `require-dev` though, are you sure you want to do this? Or do you mean this should be a draft, and we drop the `@dev` before merging?

morozov: I believe it's fine to keep it this way before the release.

morozov: All static analysis issues seem related to the dependency update (the no longer existing baseline errors in PHPStan and the deprecation error in Psalm).

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3634Test.php
```diff
@@ -30,8 +27,17 @@
  */
 class DDC3634Test extends OrmFunctionalTestCase
 {
+    private LastInsertIdMocker $idMocker;
+
     protected function setUp(): void
     {
+        $this->idMocker = new LastInsertIdMocker();
+        $config         = new Configuration();
+        $config->setMiddlewares([new LastInsertIdMockMiddleware($this->idMocker)]);
```
morozov, line 36: This still looks like too much code to mock a single value returned by `Connection::lastInsertId()`. Would it make sense to use PHPUnit mocks for that? Or the test uses other Connection APIs for which we want to use the default implementation?

composer.json
derrabus: > This is not `require-dev` though, are you sure you want to do this?

I believe that DBAL 3.3 will be released way before ORM 3.0.0, so I think it's okay to do this temporarily.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3634Test.php
derrabus: Yes, the test persists entities in a real database. What we want to test is if the ORM handles very large IDs returned as string correctly. If you have an idea to simplify this partial mock setup, I'm all ears.

morozov: It is indeed a functional test. In this case, it shouldn't use mocks. It should reproduce a real case instead. BTW, what is the corresponding code change that it covers?

For instance, initialize the autoincrement field with `PHP_INT_MAX` and insert a new row.

Alternatively, since it's not a real functional test, it could be converted into a unit one and cover only the corresponding component w/o mocking the whole world.

composer.json
greg0ire: Oh right this is ORM 3, looks totally fine indeed!

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC3634Test.php
derrabus: > It is indeed a functional test. In this case, it shouldn't use mocks. It should reproduce a real case instead. BTW, what is the corresponding code change that it covers?

That would be this seven years old PR: #1346.

> For instance, initialize the autoincrement field with `PHP_INT_MAX` and insert a new row.

Does the schema manager give me a nice way to do that?

> Alternatively, since it's not a real functional test, it could be converted into a unit one and cover only the corresponding component w/o mocking the whole world.

Yes, a unit test would not be a bad idea, however since we need to make sure that UoW and persisters talk to each other correctly, I would not want to remove the functional test either way.

morozov: > Does the schema manager give me a nice way to do that?

I don't think so, since there's no portable API for that.

> Yes, a unit test would not be a bad idea, however since we need to make sure that UoW and persisters talk to each other correctly, I would not want to remove the functional test either way.

This test doesn't cover a significant part for the functionality. Specifically, the driver returning a bigint. As we learned recently, SQLite returns integers as integers. Mocking ~~this~~ anything in an integration test is a bit naive IMO.

FWIW, I don't mind these changes merged in any state. But IMO the test in question causes more trouble than produces value. If it was implemented properly, it wouldn't have to be updated in the first place.

morozov (approved):

PR #9292, opened 2021-12-27
Add types to entity manager

derrabus:
This PR adds native parameter, return and property types to our entity manager implementations.


tests/Doctrine/Tests/ORM/Decorator/EntityManagerDecoratorTest.php
```diff
@@ -1,112 +0,0 @@
...
-use Doctrine\ORM\Query\ResultSetMapping;
-use PHPUnit\Framework\MockObject\MockObject;
-use PHPUnit\Framework\TestCase;
-use ReflectionClass;
-use ReflectionMethod;
-
-use function array_fill;
-use function in_array;
-
-class EntityManagerDecoratorTest extends TestCase
```
derrabus, line 18: As soon as we add return types, this test is broken beyond repair because it violates the type contracts almost everywhere. 🙁

derrabus:
~~Waiting for #9293~~

SenseException:
How did you decide where you wanted to add a type and where not?

For example, why did `$className` get no type?
``` php
    public function getClassMetadata($className): ClassMetadata
    {
        return parent::getClassMetadata($className);
    }
```

derrabus:
> How did you decide where you wanted to add a type and where not?

I _want_ to add them everywhere. 🙂 

> For example, why did `$className` get no type?
> 
> ```
>     public function getClassMetadata($className): ClassMetadata
>     {
>         return parent::getClassMetadata($className);
>     }
> ```

The interface that is implemented here originates in the persitence repository:

https://github.com/doctrine/persistence/blob/9f226731048d8efbd42e3b0079c6fc353f3a6b96/lib/Doctrine/Persistence/ObjectManager.php#L124-L138

PHP does not allow me to add native parameter types downstream (except for `mixed` that is). So the type would need to be added to the `ObjectManager` interface first before we can proceed here. I _could_ declare the parameter as `mixed`, but I'd rather leave it untyped until the persistence has been patched with types in a stable release. Does that make sense?

SenseException (approved):
Yep, I forgot about the dependency to Persistence and a reminder to update it too.

PR #9293, opened 2021-12-27
Enable PHPCS rules for PHP 8.0

derrabus:
This will add some types, especially to the test suite. ~~Let's see which tests fail and fix the issues on lower branches.~~

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -9,7 +9,7 @@
     <!-- Ignore warnings, show progress of the run and show sniff names -->
     <arg value="nps"/>
 
-    <config name="php_version" value="70400"/>
+    <config name="php_version" value="80000"/>
 
     <file>lib</file>
     <file>tests</file>
@@ -26,6 +26,7 @@
         <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming"/>
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming"/>
+        <exclude name="SlevomatCodingStandard.Classes.ModernClassNameReference.ClassNameReferencedViaFunctionCall"/>
     </rule>
 
     <rule ref="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint">
--- a/tests/Doctrine/StaticAnalysis/get-metadata.php
+++ b/tests/Doctrine/StaticAnalysis/get-metadata.php
@@ -16,10 +16,9 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 abstract class GetMetadata
 {
     /**
-     * @param string|object $class
      * @psalm-param class-string|object $class
      */
-    abstract public function getEntityManager($class): EntityManagerInterface;
+    abstract public function getEntityManager(string|object $class): EntityManagerInterface;
 
     /**
      * @psalm-param class-string<TObject> $class
--- a/tests/Doctrine/Tests/DoctrineTestCase.php
+++ b/tests/Doctrine/Tests/DoctrineTestCase.php
@@ -25,10 +25,8 @@ abstract class DoctrineTestCase extends TestCase
 
     /**
      * @param array<mixed> $arguments
-     *
-     * @return mixed
      */
-    public static function __callStatic(string $method, array $arguments)
+    public static function __callStatic(string $method, array $arguments): mixed
     {
         if (isset(self::$phpunitMethodRenames[$method])) {
             $method = self::$phpunitMethodRenames[$method];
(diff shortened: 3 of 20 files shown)


lib/Doctrine/ORM/Cache/Persister/Entity/NonStrictReadWriteCachedEntityPersister.php
```diff
@@ -87,7 +85,7 @@ public function update($entity)
      */
     private function updateCache($entity, bool $isChanged): bool
     {
-        $class     = $this->metadataFactory->getMetadataFor(get_class($entity));
+        $class     = $this->metadataFactory->getMetadataFor($entity::class);
```
SenseException: Can `$entity` get an `object` type hint? If not, then `$entity::class` shouldn't be used.

lib/Doctrine/ORM/Id/AssignedGenerator.php
```diff
@@ -23,7 +21,7 @@ class AssignedGenerator extends AbstractIdGenerator
      */
     public function generate(EntityManager $em, $entity)
     {
-        $class      = $em->getClassMetadata(get_class($entity));
+        $class      = $em->getClassMetadata($entity::class);
```
SenseException: Can `$entity` be `null`? If not, then `$entity::class` shouldn't be used.

lib/Doctrine/ORM/Mapping/DefaultEntityListenerResolver.php
```diff
@@ -46,7 +45,7 @@ public function register($object)
             throw new InvalidArgumentException(sprintf('An object was expected, but got "%s".', gettype($object)));
         }
 
-        $this->instances[get_class($object)] = $object;
+        $this->instances[$object::class] = $object;
```
SenseException: Same as mentioned before. I won't comment the other lines with `$variable::class`

SenseException (commented):
This PR feels like it should be a follow up to a PR that introduces more type hints in all methods.

lib/Doctrine/ORM/Cache/Persister/Entity/NonStrictReadWriteCachedEntityPersister.php
derrabus: We shouldn't call `get_class()` either then.

I mean, we could deactivate that rule if we don't feel comfortable about it. But the `::class` syntax should be as (un)safe as a `get_class()` call.

derrabus:
Thank you for the review @SenseException. I think, I should've been more clear about how this PR was generated. All I did was raising the PHP language level for CodeSniffer. All other changes are automated with `phpcbf`. This is how our Doctrine coding standard wants a PHP 8 codebase to look like. 🙂 

> This PR feels like it should be a follow up to a PR that introduces more type hints in all methods.

Yes, and that's my plan. Applying types everywhere fails miserably (see #9270), mainly because our doc blocks are not as accurate as they need to be for automating that change.

Instead, I try to add types iteratively (see #9292) and if I find inaccurate typings while doing so, I backport the type fixes to 2.11.x (see #9273). Feel free to join that effort. ORM 3.0 is a good oppertunity to apply native types throughout the codebase.

lib/Doctrine/ORM/Id/AssignedGenerator.php
derrabus: Same as above, `get_class()` fails on `null` as well.

`AbstractIdGenerator::generate()` declares the parameter as nullable, but that doesn't feel right. Maybe we should evaluate tighten the parameter type here, but I'd consider this out of scope for this PR.

SenseException:
I like the change itself because a `null` value would consistently result in a fatal error in all supported PHP versions. I'm more comfortable when at least the involved methods can have a type hint, because if the phpdocs aren't accurate now, it might be more painful for those who need to migrate. Do I need to review one of the drafts in case of introducing types? Maybe partially introducing types where it works?

derrabus:
> I like the change itself because a `null` value would consistently result in a fatal error in all supported PHP versions.

"All supported PHP versions" on the 3.0.x branch are 8.0 and 8.1. Given `$foo = null`, both `get_class($foo)` and `$foo::class` would produce a `TypeError` on these versions. The error message changes, that's all.

* https://3v4l.org/5ZKZ1
* https://3v4l.org/vdLJD

Yes, we should get the types straight, but let's do this in follow-ups. This PR sets up our tooling for the new language level. Once it's merged we can work on adding native type declarations where it makes sense.

phpcs.xml.dist
```diff
@@ -26,6 +26,7 @@
         <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming"/>
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming"/>
+        <exclude name="SlevomatCodingStandard.Classes.ModernClassNameReference.ClassNameReferencedViaFunctionCall"/>
```
derrabus, line 29: I've deactivated the rule for now because the discussion around it seems to block the merge. I can activate it again in a follow-up.

SenseException:
Okay, once the build is green, I'll :+1:

SenseException (approved):

PR #9294, opened 2021-12-27
Fix union type on QueryExpressionVisitorTest::testWalkComparison()

derrabus:
Spotted while working on #9293

### DIFF
--- a/tests/Doctrine/Tests/ORM/Query/QueryExpressionVisitorTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/QueryExpressionVisitorTest.php
@@ -27,7 +27,7 @@ class QueryExpressionVisitorTest extends TestCase
     }
 
     /**
-     * @param QueryBuilder\Comparison|string $queryExpr
+     * @param QueryBuilder\Comparison|QueryBuilder\Func|string $queryExpr
      *
      * @dataProvider comparisonData
      */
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Query/QueryExpressionVisitorTest.php
```diff
@@ -39,7 +39,13 @@ public function testWalkComparison(CriteriaComparison $criteriaExpr, $queryExpr,
         }
     }
 
-    /** @psalm-return list<array{CriteriaComparison, QueryBuilder\Comparison|string, ?Parameter} */
+    /**
+     * @psalm-return list<array{
+     *                   0: CriteriaComparison,
+     *                   1: QueryBuilder\Comparison|QueryBuilder\Func|string,
+     *                   2?: Parameter,
```
greg0ire, line 46: Is this different than the previous specification for this parameter? Previously it meant always present but possibly null, whereas here it would mean never null, but optional? If there is no difference then are the numbers necessary?

greg0ire (approved):
Approving in case there is a difference

tests/Doctrine/Tests/ORM/Query/QueryExpressionVisitorTest.php
derrabus: Exactly. And if you look at the data provider you'll find that the declaration "never null, but optional" is more accurate.

Oh, and the closing `>` was missing, so the previous type declaration was invalid and thus ignored.

PR #9301, opened 2021-12-29
Fix UnitOfWork->originalEntityData is missing not-modified collections after computeChangeSet 

### LINKED ISSUE #9300: UnitOfWork->originalEntityData is missing not-modified collections after computeChangeSet
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.x and newer

#### Summary

`UnitOfWork->computeChangeSet(ClassMetadata $class, $entity)` overrides the `$this->originalEntityData[$oid]` by `$actualData` constructed inside this class.

Collections that were hydrated from the DB but not overridden (`$value instanceof PersistentCollection && $value->getOwner === $entity`) are skipped and not present in the `$actualData` variable.

This causes duplicate inserts and possibly other issues in case of subsequent flushes/computeChangeSet calls.

#### How to reproduce

See https://github.com/doctrine/orm/pull/9301

#### Expected behavior
All fields of entity should be present in $actualData (and therefore in originalEntityData).

olsavmic:
Resolves https://github.com/doctrine/orm/issues/9300

### DIFF
--- a/tests/Doctrine/Tests/Models/Issue9300/Issue9300Child.php
+++ b/tests/Doctrine/Tests/Models/Issue9300/Issue9300Child.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\Issue9300;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\ManyToMany;
+
+/**
+ * @Entity
+ */
+class Issue9300Child
+{
+    /**
+     * @var int
+     * @Id
+     * @Column(type="integer")
+     * @GeneratedValue
+     */
+    public $id;
+
+    /**
+     * @var Collection|Issue9300Parent
+     * @ManyToMany(targetEntity="Issue9300Parent")
+     */
+    public $parents;
+
+    /**
+     * @var string
+     * @Column(type="string")
+     */
+    public $name;
+
+    public function __construct()
+    {
+        $this->parents = new ArrayCollection();
+    }
+}
--- a/tests/Doctrine/Tests/Models/Issue9300/Issue9300Child.php
+++ b/tests/Doctrine/Tests/Models/Issue9300/Issue9300Child.php
@@ -26,7 +26,7 @@ class Issue9300Child
     public $id;
 
     /**
-     * @var Collection|Issue9300Parent
+     * @var Collection<int, Issue9300Parent>
      * @ManyToMany(targetEntity="Issue9300Parent")
      */
     public $parents;
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/Issue9300Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/Issue9300Test.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Tests\Models\Issue9300\Issue9300Child;
+use Doctrine\Tests\Models\Issue9300\Issue9300Parent;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group 9300
+ */
+class Issue9300Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        $this->useModelSet('issue9300');
+        parent::setUp();
+    }
+
+    /**
+     * @group #9300
+     */
+    public function testPersistedCollectionIsPresentInOriginalDataAfterFlush(): void
+    {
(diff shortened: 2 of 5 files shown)


beberlei:
The test is not sufficient, as it does not show that this happens when just using the EntityManager public API. What you are doing might not be a valid use of the internal API of UoW.

olsavmic:
> The test is not sufficient, as it does not show that this happens when just using the EntityManager public API. What you are doing might not be a valid use of the internal API of UoW.

I've rewritten the test to use EntityManager instead of UoW. The issue is with the internal UoW state so the assert does check the UoW directly.

I can add additional test case that tests the side effect of not having the persistent collection present in the original data that would throw DUPLICATE INSERT in case the fix is not present if you'd prefer that, let me know :)

olsavmic:
I added a second test case showing how dangerous this bug is. Both test cases fail without the fix.

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue9300Test.php
```diff
@@ -0,0 +1,79 @@
...
+        $this->_em->persist($parentOne);
+        $this->_em->persist($parentTwo);
+        $this->_em->persist($childOne);
+        $this->_em->flush();
+
+        // Recalculate change-set -> new original data
+        $childOne->name = 'abcd';
+        $this->_em->flush();
+
+        $childOne->parents = new ArrayCollection([$parentTwo]);
```
xificurk, line 71: Is this re-initialization of a collection a valid use-case? If you clear and modify the existing collection via its API, then the test passes even without the fix in UnitOfWork.

But the fact that an additional flush before this code results in different content being stored in database is confusing and quite dangerous.

olsavmic: I guess what's not forbidden is valid, it's not in the docs :)

olsavmic:
@derrabus @beberlei Any possibility of getting this reviewed? We mitigated the issue by removing all collection reassignments and using `->clear()` instead as proposed by @xificurk but it still remains dangerous.

olsavmic:
@beberlei @derrabus @mpdude

Can this please get a review or at least close this MR with a reason why it's not a welcomed change? Thanks!

tests/Doctrine/Tests/Models/Issue9300/Issue9300Child.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * @var int
+     * @Id
+     * @Column(type="integer")
+     * @GeneratedValue
+     */
+    public $id;
+
+    /**
+     * @var Collection|Issue9300Parent
```
derrabus:
```suggestion
     * @var Collection<int, Issue9300Parent>
```

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue9300Test.php
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Tests\Models\Issue9300\Issue9300Child;
+use Doctrine\Tests\Models\Issue9300\Issue9300Parent;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group 9300
```
derrabus:
```suggestion
 * @group GH-9300
```

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue9300Test.php
```diff
@@ -0,0 +1,79 @@
...
+class Issue9300Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        $this->useModelSet('issue9300');
+        parent::setUp();
+    }
+
+    /**
+     * @group #9300
```
derrabus:
```suggestion
     * @group GH-9300
```

tests/Doctrine/Tests/ORM/Functional/Ticket/Issue9300Test.php
```diff
@@ -0,0 +1,79 @@
...
+        $parent->name = 'abcd';
+        $child->name  = 'abcd';
+
+        $this->_em->flush();
+
+        self::assertArrayHasKey('parents', $this->_em->getUnitOfWork()->getOriginalEntityData($child));
+    }
+
+    /**
+     * @group #9300
```
derrabus:
```suggestion
     * @group GH-9300
```

derrabus (commented):
Sorry, I wasn't aware of this PR. Can you rebase onto 2.16.x please? We've changed a lot in the UoW recently and I'd like to see if the CI still passes.

olsavmic:
Thank you! It's rebased and review comments are fixed.

derrabus (approved):

derrabus:
Thank you!

VincentChalnot:
_The change made to UnitOfWork in this fix is breaking ManyToMany associations which are not persisted anymore: Collections are completely cleared on flush. (In my project at least)
I'm not sure what this is supposed to fix but the fact that it did not break any test is mind boggling.
Currently trying to find a way to reproduce this on a sample project._

**EDIT**: Nevermind this seems to be related to some custom code in my project, still I have a feeling that something is not right if a tiny change like this one can break a major feature.

PR #9304, opened 2021-12-30
Add support for PHP 8.1 enums.

### LINKED ISSUE #9021: Support for native PHP enums

beberlei:
ORM based support for enums.

```php
enum Suit: string {
    case Hearts = 'H';
    case Diamonds = 'D'; 
    case Clubs = 'C';
    case Spades = 'S';
}

#[Entity]
class Card
{
    /** ... */

    #[Column(type: 'string', enumType: Suit::class)]
    public $suit;
```

It works based on a new `ReflectionEnumProperty` that wraps around other reflection property types. This avoids the problem of other approaches that hook into DBAL Type abstraction, since that would require some kind of paramterization which is not supported at the moment. In addition it would at the moment require to register each enum explicitly, which this approach does not.

This includes support for typed property detection, the previous example could be written as:

```php
#[Column]
public Suit $suit;
```

As for error handling, when the database returns an invalid enum case, then the regular `ValueError` is not caught that `BackedEnum::from` throws.

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -199,6 +199,8 @@ list:
   unique key.
 - ``nullable``: (optional, default FALSE) Whether the database
   column is nullable.
+- ``enumType``: (optional, requires PHP 8.1 and ORM 2.11) The PHP enum type
+  name to convert the database value into.
 - ``precision``: (optional, default 0) The precision for a decimal
   (exact numeric) column (applies only for decimal column),
   which is the maximum number of digits that are stored for the values.
@@ -232,6 +234,9 @@ Additionally, Doctrine will map PHP types to ``type`` attribute as follows:
 - ``float``: ``float``
 - ``int``: ``integer``
 - ``string`` or any other type: ``string``
+
+As of version 2.11 Doctrine can also automatically map typed properties using a
+PHP 8.1 enum to set the right ``type`` and ``enumType``.
 
 .. _reference-mapping-types:
 
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Mapping;
 
+use BackedEnum;
 use BadMethodCallException;
 use DateInterval;
 use DateTime;
@@ -22,6 +23,7 @@ use Doctrine\Persistence\Mapping\ReflectionService;
 use InvalidArgumentException;
 use LogicException;
 use ReflectionClass;
+use ReflectionEnum;
 use ReflectionNamedType;
 use ReflectionProperty;
 use RuntimeException;
@@ -37,6 +39,7 @@ use function array_values;
 use function assert;
 use function class_exists;
 use function count;
+use function enum_exists;
 use function explode;
 use function gettype;
 use function in_array;
@@ -77,6 +80,7 @@ use const PHP_VERSION_ID;
  *      length?: int,
  *      id?: bool,
  *      nullable?: bool,
+ *      enumType?: class-string<BackedEnum>,
  *      columnDefinition?: string,
  *      precision?: int,
  *      scale?: int,
@@ -1025,6 +1029,13 @@ class ClassMetadataInfo implements ClassMetadata
             $this->reflFields[$field] = isset($mapping['declared'])
                 ? $reflService->getAccessibleProperty($mapping['declared'], $field)
                 : $reflService->getAccessibleProperty($this->name, $field);
+
+            if (isset($mapping['enumType']) && $this->reflFields[$field] !== null) {
+                $this->reflFields[$field] = new ReflectionEnumProperty(
+                    $this->reflFields[$field],
+                    $mapping['enumType']
+                );
+            }
         }
 
         foreach ($this->associationMappings as $field => $mapping) {
@@ -1468,6 +1479,15 @@ class ClassMetadataInfo implements ClassMetadata
                 ! isset($mapping['type'])
                 && ($type instanceof ReflectionNamedType)
             ) {
+                if (PHP_VERSION_ID >= 80100 && ! $type->isBuiltin() && enum_exists($type->getName(), false)) {
+                    $mapping['enumType'] = $type->getName();
+
+                    $reflection = new ReflectionEnum($type->getName());
+                    $type       = $reflection->getBackingType();
+
+                    assert($type instanceof ReflectionNamedType);
+                }
+
                 switch ($type->getName()) {
                     case DateInterval::class:
                         $mapping['type'] = Types::DATEINTERVAL;
@@ -1587,6 +1607,16 @@ class ClassMetadataInfo implements ClassMetadata
             }
 
             $mapping['requireSQLConversion'] = true;
+        }
+
+        if (isset($mapping['enumType'])) {
+            if (PHP_VERSION_ID < 80100) {
+                throw MappingException::enumsRequirePhp81($this->name, $mapping['fieldName']);
+            }
+
+            if (! enum_exists($mapping['enumType'])) {
+                throw MappingException::nonEnumTypeMapped($this->name, $mapping['fieldName'], $mapping['enumType']);
+            }
         }
 
         return $mapping;
--- a/lib/Doctrine/ORM/Mapping/Column.php
+++ b/lib/Doctrine/ORM/Mapping/Column.php
@@ -44,6 +44,9 @@ final class Column implements Annotation
     /** @var bool */
     public $nullable = false;
 
+    /** @var class-string<\BackedEnum>|null */
+    public $enumType = null;
+
     /** @var array<string,mixed> */
     public $options = [];
 
@@ -51,7 +54,8 @@ final class Column implements Annotation
     public $columnDefinition;
 
     /**
-     * @param array<string,mixed> $options
+     * @param class-string<\BackedEnum>|null $enumType
+     * @param array<string,mixed>            $options
      */
     public function __construct(
         ?string $name = null,
@@ -61,6 +65,7 @@ final class Column implements Annotation
         ?int $scale = null,
         bool $unique = false,
         bool $nullable = false,
+        ?string $enumType = null,
         array $options = [],
         ?string $columnDefinition = null
     ) {
@@ -71,6 +76,7 @@ final class Column implements Annotation
         $this->scale            = $scale;
         $this->unique           = $unique;
         $this->nullable         = $nullable;
+        $this->enumType         = $enumType;
         $this->options          = $options;
         $this->columnDefinition = $columnDefinition;
     }
--- a/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
@@ -718,6 +718,7 @@ class AnnotationDriver extends AbstractAnnotationDriver
      *                   unique: bool,
      *                   nullable: bool,
      *                   precision: int,
+     *                   enumType?: class-string,
      *                   options?: mixed[],
      *                   columnName?: string,
      *                   columnDefinition?: string
@@ -745,6 +746,10 @@ class AnnotationDriver extends AbstractAnnotationDriver
 
         if (isset($column->columnDefinition)) {
             $mapping['columnDefinition'] = $column->columnDefinition;
+        }
+
+        if ($column->enumType !== null) {
+            $mapping['enumType'] = $column->enumType;
         }
 
         return $mapping;
--- a/lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
+++ b/lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
@@ -1,0 +1,82 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping;
+
+use BackedEnum;
+use ReflectionProperty;
+use ReturnTypeWillChange;
+use ValueError;
+
+use function assert;
+use function get_class;
+use function is_int;
+use function is_string;
+
+class ReflectionEnumProperty extends ReflectionProperty
+{
+    /** @var ReflectionProperty */
+    private $originalReflectionProperty;
+
+    /** @var class-string<BackedEnum> */
+    private $enumType;
+
+    /**
+     * @param class-string<BackedEnum> $enumType
+     */
+    public function __construct(ReflectionProperty $originalReflectionProperty, string $enumType)
+    {
+        $this->originalReflectionProperty = $originalReflectionProperty;
+        $this->enumType                   = $enumType;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @param object|null $object
+     *
+     * @return int|string|null
+     */
+    #[ReturnTypeWillChange]
+    public function getValue($object = null)
+    {
+        if ($object === null) {
+            return null;
+        }
+
+        $enum = $this->originalReflectionProperty->getValue($object);
+
+        if ($enum === null) {
+            return null;
+        }
+
+        return $enum->value;
+    }
+
+    /**
+     * @param object $object
+     * @param mixed  $value
+     */
+    public function setValue($object, $value = null): void
+    {
+        if ($value !== null) {
+            $enumType = $this->enumType;
+            try {
+                $value = $enumType::from($value);
+            } catch (ValueError $e) {
+                assert(is_string($value) || is_int($value));
+
+                throw MappingException::invalidEnumValue(
+                    get_class($object),
+                    $this->originalReflectionProperty->getName(),
+                    (string) $value,
+                    $enumType,
+                    $e
+                );
+            }
+        }
+
+        $this->originalReflectionProperty->setValue($object, $value);
+    }
+}
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1746,7 +1746,7 @@ parameters:
 			path: lib/Doctrine/ORM/Tools/Export/Driver/XmlExporter.php
 
 		-
-			message: "#^Offset 'version' on array\\{type\\: string, fieldName\\: string, columnName\\?\\: string, length\\?\\: int, id\\?\\: bool, nullable\\?\\: bool, columnDefinition\\?\\: string, precision\\?\\: int, \\.\\.\\.\\} in isset\\(\\) does not exist\\.$#"
+			message: "#^Offset 'version' on array\\{type\\: string, fieldName\\: string, columnName\\?\\: string, length\\?\\: int, id\\?\\: bool, nullable\\?\\: bool, enumType\\?\\: class\\-string\\<BackedEnum\\>, columnDefinition\\?\\: string, \\.\\.\\.\\} in isset\\(\\) does not exist\\.$#"
 			count: 1
 			path: lib/Doctrine/ORM/Tools/Export/Driver/XmlExporter.php
 
--- a/tests/Doctrine/Tests/Models/Enums/Suit.php
+++ b/tests/Doctrine/Tests/Models/Enums/Suit.php
@@ -1,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\Enums;
+
+enum Suit: string
+{
+    case Hearts = 'H';
+    case Diamonds = 'D';
+    case Clubs = 'C';
+    case Spades = 'S';
+}
--- a/tests/Doctrine/Tests/ORM/Functional/EnumTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/EnumTest.php
@@ -1,0 +1,104 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use Doctrine\ORM\Mapping\Driver\AttributeDriver;
+use Doctrine\ORM\Mapping\MappingException;
+use Doctrine\ORM\Tools\SchemaTool;
+use Doctrine\Tests\Models\Enums\Card;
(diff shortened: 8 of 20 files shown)


lib/Doctrine/ORM/Mapping/Column.php
```diff
@@ -44,6 +44,9 @@ final class Column implements Annotation
     /** @var bool */
     public $nullable = false;
 
+    /** @var class-string<\Enum>|null */
```
derrabus:
```suggestion
    /** @var class-string<\BackedEnum>|null */
```

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -718,6 +718,7 @@ private function joinColumnToArray(Mapping\JoinColumn $joinColumn): array
      *                   unique: bool,
      *                   nullable: bool,
      *                   precision: int,
+     *                   enumType: class-string,
```
derrabus:
```suggestion
     *                   enumType: class-string<BackedEnum>,
```

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -0,0 +1,49 @@
...
+
+use ReflectionProperty;
+use UnitEnum;
+
+class ReflectionEnumProperty
+{
+    /** @var ReflectionProperty */
+    private $originalReflectionProperty;
+
+    /** @var class-string<UnitEnum> */
```
derrabus:
```suggestion
    /** @var class-string<BackedEnum> */
```

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -0,0 +1,49 @@
...
+    public function __construct($originalReflectionProperty, string $enumType)
+    {
+        $this->originalReflectionProperty = $originalReflectionProperty;
+        $this->enumType                   = $enumType;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return mixed
```
derrabus:
```suggestion
     * @return int|string
```

tests/Doctrine/Tests/Models/Enums/Suit.php
```diff
@@ -0,0 +1,12 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\Enums;
+
+enum Suit: string {
+case Hearts = 'H';
+case Diamonds = 'D';
+case Clubs = 'C';
+case Spades = 'S';
+    }
```
derrabus: CodeSniffer screws up the indentation for enums. We should deactivate the rule for files that contain enums until this has been fixed upstream.

tests/Doctrine/Tests/ORM/Functional/EnumTest.php
```diff
@@ -0,0 +1,42 @@
...
+
+use const PHP_VERSION_ID;
+
+class EnumTest extends OrmFunctionalTestCase
+{
+    public function setUp(): void
+    {
+        if (PHP_VERSION_ID < 80100) {
+            $this->markTestSkipped('Enum support requires PHP 8.1+');
+        }
```
derrabus: You could annotate the class with `@requires PHP 8.1` instead.

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -0,0 +1,49 @@
...
+
+class ReflectionEnumProperty
+{
+    /** @var ReflectionProperty */
+    private $originalReflectionProperty;
+
+    /** @var class-string<UnitEnum> */
+    private $enumType;
+
+    public function __construct($originalReflectionProperty, string $enumType)
```
derrabus:
```suggestion
    public function __construct(ReflectionProperty $originalReflectionProperty, string $enumType)
```

tests/Doctrine/Tests/Models/Enums/Suit.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\Enums;
+
+enum Suit: string {
```
beberlei: phpcs did that

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -0,0 +1,52 @@
...
+
+        if ($enum === null) {
+            return null;
+        }
+
+        return $enum->value;
+    }
+
+    #[ReturnTypeWillChange]
+    public function setValue($object, $value = null)
```
derrabus:
```suggestion
    public function setValue($object, $value = null): void
```

There's no need to use that attribute for a brand-new class (if we _can_ provide the native return type).

derrabus:
This looks like the most promising solution I've seen so far, thanks!

derrabus:
For phpcs.xml.dist:

```XML
<rule ref="Generic.WhiteSpace.ScopeIndent.Incorrect">
    <!-- see https://github.com/squizlabs/PHP_CodeSniffer/issues/3474 -->
    <exclude-pattern>your/path/here</exclude-pattern>
</rule>
<rule ref="Generic.WhiteSpace.ScopeIndent.IncorrectExact">
    <!-- see https://github.com/squizlabs/PHP_CodeSniffer/issues/3474 -->
    <exclude-pattern>your/path/here</exclude-pattern>
</rule>
```

If you add the paths to your enum files here, PHPCS should allow you to fix the indentation manually.

derrabus:
I think the following error scenarios should be covered by tests:

* `enumType` is set to a non-string or a string that is not a value enum FQCN.
* The database contains a non-null value that is not part of the enum.

Furthermore, I think we should be able to detect `type` and `enumType` fairly easy for a typed property.

lib/Doctrine/ORM/Mapping/Column.php
```diff
@@ -44,14 +44,18 @@ final class Column implements Annotation
...
+
     /** @var array<string,mixed> */
     public $options = [];
 
     /** @var string|null */
     public $columnDefinition;
 
     /**
-     * @param array<string,mixed> $options
+     * @param class-string<\BackedEnum> $enumType
```
derrabus:
```suggestion
     * @param class-string<\BackedEnum>|null $enumType
```

beberlei:
The `BackedEnum` usage is making Psalm crazy, a bit at loss how to workaround the class not existing in 8.0/7.4

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -77,6 +77,7 @@
  *      length?: int,
  *      id?: bool,
  *      nullable?: bool,
+ *      enumType?: class-string,
```
```suggestion
 *      enumType?: class-string<BackedEnum>,
```

derrabus:
> The `BackedEnum` usage is making Psalm crazy, a bit at loss how to workaround the class not existing in 8.0/7.4

```XML
<UndefinedDocblockClass>
    <errorLevel type="suppress">
        <referencedClass name="BackedEnum"/>
    </errorLevel>
</UndefinedDocblockClass>
```

greg0ire (commented):
Nice! This does look more straightforward than other approaches.

3 things:
- should there be checks regarding the type? What should happen if an enum backed by an int is used within a mapping that uses a stringy type? And vice versa?
- should we stop adding features to the YAML and Annotation drivers? They are deprecated.
- there should be docs

One more reason why it makes sense to implement this at the ORM level: what we obtain in our mapped property is a user-defined type (the enum), just like embedded classes, and those are not implemented in the DBAL.

beberlei:
@greg0ire about your bullets:

- The same issue already exists with typing `private int $id` and declaring it as `type: "string"` in mappings and so on. We don't validate that kind of missmatch at the moment. Maybe something for `SchemaValidator`?
- I was thinking the same, but since we target 2.11 and its only a few lines for each driver I did them anyways. Problem would be that the Exporter classes don't have support for these new mapping things, so you cannot fully convert between a yml with new features to xml for example.
- Docs will follow once i have everything else working :)

filiplikavcan:
> ```php
> #[Entity]
> class Card
> {
>     /** ... */
> 
>     #[Column(type: 'string', enumType: Suit::class)]
>     public $suit;
> ```

`type: 'string'` can be inferred from enum reflection. I would prefer either `#[Column(type: Suit::class)]` or `#[Enum(type: Suit::class)]`.

beberlei:
@filiplikavcan the tests usually go for the full information, you can just declare it as the type itself and Doctrine's automatic type completion will do that:

```php
#[Column]
public Suit $suit;
```

beberlei:
I have everything working now exception Psalm not detecting my ignoring of `ReflectionEnum`: https://github.com/doctrine/orm/runs/4673434142?check_suite_focus=true

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
derrabus: You've marked this comment as resolved without applying the change or explaining why you don't want to apply it. 🙃

phpstan-baseline.neon
```diff
@@ -1767,7 +1767,7 @@ parameters:
 			path: lib/Doctrine/ORM/Tools/Export/Driver/XmlExporter.php
 
 		-
-			message: "#^Offset 'version' on array\\{type\\: string, fieldName\\: string, columnName\\?\\: string, length\\?\\: int, id\\?\\: bool, nullable\\?\\: bool, columnDefinition\\?\\: string, precision\\?\\: int, \\.\\.\\.\\} in isset\\(\\) does not exist\\.$#"
+			message: "#^Offset 'version' on array#"
```
derrabus: Please don't edit the baseline manually, but let PHPStan regenerate it.

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
beberlei: @derrabus when you commented that it didn't extend `ReflectionProperty` yet, i thought doing that made the change impossible, but that was my mistake. Changed it now.

docs/en/reference/basic-mapping.rst
```diff
@@ -233,6 +235,9 @@ Additionally, Doctrine will map PHP types to ``type`` attribute as follows:
 - ``int``: ``integer``
 - ``string`` or any other type: ``string``
 
+As of version 2.11 Doctrine can also automatically map typed properties using a
+PHP 8.1 enums to set the right ``type`` and ``enumType``.
```
greg0ire:
```suggestion
PHP 8.1 enum to set the right ``type`` and ``enumType``.
```

greg0ire (commented):
I just created https://github.com/doctrine/orm/pull/9310 to make the baseline changes unnecessary (will require a merge up from 2.10 to 2.11)

derrabus:
I've modified `EnumTest` to cover the detection of the enum type property: beberlei/doctrine2#1

tests/Doctrine/Tests/ORM/Functional/EnumTest.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->_em->clear();
+
+        $metadata = $this->_em->getClassMetadata(Card::class);
+        $this->_em->getConnection()->update(
+            $metadata->table['name'],
+            [$metadata->fieldMappings['suit']['columnName'] => 'invalid'],
+            [$metadata->fieldMappings['id']['columnName'] => $card->id]
+        );
+
+        $this->expectException(ValueError::class);
```
derrabus: I think we should catch and upcast this error. There is something wrong with the user's database and the user should be made aware of that. DBAL has a `ConversionException` for errors like this one. Either we reuse that or we declare our own exception for that.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1468,6 +1479,15 @@ private function validateAndCompleteTypedFieldMapping(array $mapping): array
                 ! isset($mapping['type'])
                 && ($type instanceof ReflectionNamedType)
             ) {
+                if (PHP_VERSION_ID >= 80100 && ! $type->isBuiltin() && enum_exists($type->getName(), false)) {
```
greg0ire, line 1482: @beberlei  I pushed a commit that added the `$type->isBuiltin()` part, hopefully that will make Psalm green :slightly_smiling_face:

greg0ire:
It does make Psalm green, but now Phpstan is red… for reasons that have to do with the last release of the DBAL (2.10.x is also broken)

tests/Doctrine/Tests/ORM/Functional/EnumTest.php
greg0ire: I think we should have our own one, it would be less confusing. The message could look like this:

```
Context: trying to hydrate Property "MyClass::$foo"
Problem: That property is of enum type "My\Enum", which does not accept value "value"
Solution: migrate the database value to a supported case or add a new case to the enum
```

tests/Doctrine/Tests/ORM/Functional/EnumTest.php
```diff
@@ -0,0 +1,95 @@
...
+        $this->expectDeprecationMessage('"invalid" is not a valid backing value for enum "Doctrine\Tests\Models\Enums\Suit"');
```
greg0ire: :thinking: that's not supposed to be a deprecation is it? Weird that it does not fail the build…

```suggestion
        $this->expectExceptionMessage('"invalid" is not a valid backing value for enum "Doctrine\Tests\Models\Enums\Suit"');
```

psalm.xml
```diff
@@ -123,12 +123,20 @@
                 <!-- DBAL 3.2 forward compatibility -->
                 <referencedClass name="Doctrine\DBAL\Platforms\PostgreSQLPlatform"/>
                 <referencedClass name="Doctrine\DBAL\Platforms\SQLServerPlatform"/>
+
+                <!-- PHP 8.1 Enum support -->
+                <referencedClass name="BackedEnum"/>
+                <referencedClass name="ReflectionEnum"/>
```
derrabus: Is this still necessary now that we run Psalm with the PHP 8.1 language level?

derrabus (dismissed):

psalm.xml
greg0ire: Nope, fixed, thanks!

beberlei:
@greg0ire thank you for making the last changes. From my POV this is complete now and ready to merge

greg0ire (approved):

greg0ire:
Thanks for this great feature!

sxsws:
A much awaited feature! Well done everyone!

zmitic:
@beberlei 

Thank you **a lot** for this feature, I have already implemented it and works without a problem. 

Wish list: 
is it possible to have type guessing when XML is used?

```php
class BiomarkerRange
{
    public function __construct(
        private AgeCategoryEnum $ageCategory, 
    ) {}
```

If defined like this:

```xml
<field name="ageCategory" enum-type="AgeCategoryEnum"/>
```

everything works perfectly. But when I drop `enum-type`:

```
Cannot assign string to property BiomarkerRange::$ageCategory of type AgeCategoryEnum
```

---

It is  not a big deal if not possible, I am still very happy even if I have to put it. But it would make it more consistent with other guessers.

derrabus:
Type guessing is currently only implemented for attributes. Please open an issue if you want to extend the feature to other mapping types, but my gut feeling is that this is a bad idea.

ThomasLandauer:
@beberlei You didn't write the docs yet, did you? If you want, I could start a PR, but I'd need help from you, since I don't know all the details.

sxsws:
Just to be clear for this PR, are non-backed enums supported? Will the name value be stored as a string value?

beberlei:
Non backed enums are not supported, there is no value to store them with.

sxsws:
The 'value' is the name of the case, surely?

rkeet:
I would have to side with @sxsws on this. To quote [PHP manual](https://www.php.net/manual/en/language.enumerations.basics.php):

> This type of case, with no related data, is called a "Pure Case." An Enum that contains only Pure Cases is called a Pure Enum.
> 
> All Pure Cases are implemented as instances of their enum type. The enum type is represented internally as a class.
> 
> All Cases have a read-only property, name, that is the case-sensitive name of the case itself. That may sometimes be useful for debugging purposes.
> 
> ```php
> <?php
> print Suit::Spades->name;
> // prints "Spades"
> ?>
> ```

This would suggest to me that it's a valid, storable, value. The caveat is that the value matches the `case`. 

This is different from the [`BackedEnum`](https://www.php.net/manual/en/language.enumerations.backed.php) where the value (`int|string`) does not have to match the case. Example from the linked page: 

```php
<?php
enum Suit: string
{
    case Hearts = 'H';
    case Diamonds = 'D';
    case Clubs = 'C';
    case Spades = 'S';
}
?>

// other documentation

<?php
print Suit::Clubs->value;
// Prints "C"
?>
```

Another difference is, the value of a `Pure Enum` is accessed with `Enum::case->name`, where as a `Backed Enum` requires `Enum::case->value` usage. 

Differentiating between the can be done by using new [`UnitEnum` interface](https://www.php.net/manual/en/class.unitenum.php), which is applied to _*all enum*_ classes and the new [`BackedEnum`](https://www.php.net/manual/en/class.backedenum.php) interface, which, as the name implies, is only implied to `BackedEnum` instances. 

[This page](https://www.php.net/manual/en/language.enumerations.listing.php) also shows the existance of static method `cases()`, applied to all instances of `UnitEnum`, which returns all available cases, like so: 

```php
<?php
Suit::cases();
// Produces: [Suit::Hearts, Suit::Diamonds, Suit::Clubs, Suit:Spades]
?>
```

For mapping to a `BackedEnum`, [it's documentation page](https://www.php.net/manual/en/class.backedenum.php) shows methods [`from`](https://www.php.net/manual/en/backedenum.from.php) and [`tryFrom`](https://www.php.net/manual/en/backedenum.tryfrom.php) to map to a scalar instance, or a scalar instance or null respectively. 

The above scalar mapping for `BackedEnums` does not have a corresponding function mentioned for mapping a scalar to a [`UnitEnum`](https://www.php.net/manual/en/class.unitenum.php). 

---

Having said all of the above, I'm excited already with just BackedEnums :)

derrabus:
> > All Cases have a read-only property, name, that is the case-sensitive name of the case itself. That may sometimes be useful for debugging purposes.

It clearly says "useful for debugging purposes", not "a fallback if there is not backed value".

> This would suggest to me that it's a valid, storable, value. The caveat is that the value matches the `case`.

Probably, but it's not _meant_ for storing. I've heard alternative proposals that use the fully qualified name of the case or even the PHP-serialized representation or assigning a sequencial number to each case. Those are all valid proposals, but who are we to decide which of these options is the "right" one if the developer deliberately omits the backed value, the one piece of information that tells us, "Look, this is the way this enum is meant to be stored"?

The thing is, we can only guess how such a pure enum should be serialized. If you implement a custom DBAL type for your pure enum that takes the name as backed value, you implement a convention for your own enum and that absolutely fine. But we cannot infer a general rule from that. I think, for a library as Doctrine, it is fair to demand that you define your enum as backed if you want it to be stored properly. Besides, I fail to see the difficulty of implementing a backed enum instead of a pure one.

> Having said all of the above, I'm excited already with just BackedEnums :)

As am I. 🙂

bastien70:
Very good PR ! So, if we're good, a use case could be :

```php
enum Suit: string {
    case Hearts = 'H';
    case Diamonds = 'D'; 
    case Clubs = 'C';
    case Spades = 'S';
}

#[Entity]
class Card
{
    #[Column(type: 'string', enumType: Suit::class)]
    private ?Suit $suit = null;

    public function setSuit(Suit $suit): self
    {
        $this->suit = $suit;

        return $this;
    }

    public function getSuit(): ?Suit
    {
        return $this->suit;
    }
```

Is it okay ?

ThomasLandauer:
> It clearly says "useful for debugging purposes"

I just deleted that sentence from the PHP docs, see https://github.com/php/doc-en/pull/1331

ThomasLandauer:
When doing it like this, I'm getting this "TypeError":
> Cannot assign string to property App\Entity\Card::$suit of type ?App\Enum\Suit

After deleting the entire `var/cache/dev` directory, it works exactly *once* (i.e. after refreshing the page, the error is shown again)

Is this the problem that was fixed in https://github.com/symfony/symfony/pull/45012 ?

derrabus:
> Is this the problem that was fixed in https://github.com/symfony/symfony/pull/45012 ?

Well, there is an easy way to find out, isn't there? If you encounter a problem with enum columns, please open a bug report.

greg0ire:
> If you want, I could start a PR, but I'd need help from you, since I don't know all the details.

@ThomasLandauer did you end up starting something?

FrancescJR:
How would be the equivalent of 
```
 #[Column(type: 'string', enumType: Suit::class)]
 ```
 on xml? I can't find that anywhere on any documentation

**Comment by FrancescJR** [comment]
answering myself, (I had other mapping issues and I thought were related to that but not).


```
 #[Column(type: 'string', enumType: Suit::class)]
 ```
 
 equivalent is:
 
 ```
  <field name="[name-on-entity-attribute-forsuit" enumType="Namespace\To\Type\Suit::class"  column="your_suit_Column_name" type="string"/>
  ```

PR #9307, opened 2021-12-31
Add find to interface for extra arguments

malarzm:
Otherwise PHPStan is complaining about extra arguments even though we typehint against EntityManagerInterface instead of ObjectManager

### DIFF
--- a/tests/Doctrine/Performance/Mock/NonProxyLoadingEntityManager.php
+++ b/tests/Doctrine/Performance/Mock/NonProxyLoadingEntityManager.php
@@ -271,9 +271,9 @@ class NonProxyLoadingEntityManager implements EntityManagerInterface
     /**
      * {@inheritDoc}
      */
-    public function find($className, $id)
+    public function find($className, $id, $lockMode = null, $lockVersion = null)
     {
-        return $this->realEntityManager->find($className, $id);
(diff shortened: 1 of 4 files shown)


malarzm:
I'm a bit confused by psalm's failure :/

tests/Doctrine/Performance/Mock/NonProxyLoadingEntityManager.php
```diff
@@ -271,9 +271,9 @@ public function hasFilters()
     /**
      * {@inheritDoc}
      */
-    public function find($className, $id)
+    public function find($className, $id, $lockMode = null, $lockVersion = null)
```
malarzm, line 274: This probably means that I don't really want to introduce a BC break in a patch release :/ What would be best way forward? Add `find` as `@method` in the interface?

derrabus: Yes. You cannot add additional parameters to the interface before 3.0.

derrabus:
See #9292 for changing the `find()` signature on 3.0.x.

malarzm:
I've tried adding it in the same vein as `wrapInTransaction` but it looks like even PHPStan is not happy due to generics being used in `find`'s PHPDoc. Let's leave it as is.

PR #9308, opened 2022-01-01
[GH-9277] remove deprecated `PHPDriver`

### LINKED ISSUE #9277: Remove support for PHPDriver
|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes/

## Summary

We should deprecate and remove the PHPDriver, because using a magic variable `$metadata` in a file is confusing and the StaticPHPDriver offers a better API with a method on the entity. This will reduce the amount of test-code we have to maintain significantly.

Users can use the `orm:convert-mapping` command in the 2.* branches to move to another mapping driver without great effort.

In addition if they want to stay with programmatic metadata configuration, `StaticPHPDriver` could be used in a simple way to keep the old mapping:

```php
class City
{
    public static function loadMetadata(ClassMetadata $metadata)
    {
        include "path/to/mapping/City.php";
    }
}
```

beberlei:
Remove deprecated `PHPDriver`, use `StaticPHPDriver` instead.

Fixes #9277


derrabus (dismissed):

derrabus:
Don't forget to mention the BC break in UPGRADE.md.

derrabus (approved):

PR #9310, opened 2022-01-01
Run static analysis on PHP 8.1

greg0ire:
This will make it easier to add code that leverages features only
defined since that version of PHP.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
-          - "8.0"
+          - "8.1"
         dbal-version:
           - "default"
           - "2.13"
@@ -58,7 +58,7 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
-          - "8.0"
+          - "8.1"
         dbal-version:
           - "default"
           - "2.13"


derrabus:
I agree that we should analyze the codebase on PHP 8.1. While PHP 7.1 is the lowest supported  version, the repository contains features that require a more recent PHP version and we are currently hardly able to run static analysis on those parts.

Just raising the PHP interpreter version is not enough however. We need to toll both tools to apply PHP 8.1 rules. This is where you can do this for PHPStan:

https://github.com/doctrine/orm/blob/d30e748e645da3e2e57e33a49a22c6547c82070c/phpstan-params.neon#L11

For Psalm, you need to add the attribute `phpVersion="8.1"` to the top-level `<psalm>` element of psalm.xml. If you don't do that, Psalm will infer PHP 7.1 as language level from composer.json.

derrabus:
Moreover, rasining the language level should allow us to drop some hacks from Psalm's configuration:

https://github.com/doctrine/orm/blob/d30e748e645da3e2e57e33a49a22c6547c82070c/psalm.xml#L90-L96

https://github.com/doctrine/orm/blob/d30e748e645da3e2e57e33a49a22c6547c82070c/psalm.xml#L103-L108

greg0ire:
@derrabus I'm not aiming at testing PHP 8.1 features with this PR, I'm aiming at preventing them from making the baseline grow. If we want to test our usage of PHP 8.1 features, while still checking that our code is compatible with our lowest supported version (we want to do that, right?), we will had to rework the static analysis jobs.

Do we still want to have jobs with 7.1 as a target version?

derrabus:
> @derrabus I'm not aiming at testing PHP 8.1 features with this PR, I'm aiming at preventing them from making the baseline grow.

Does it really make a difference if we bump the PHP runtime only? Both tools will still assume PHP 7.1 as language level.

> Do we still want to have jobs with 7.1 as a target version?

Right now, letting the tools operate PHP 7.1 as language level causes more pain than it solves problems imho. Especially Psalm has a history of reporting funny errors if the dependencies are on a higher language level. I'd suggest to let the test suite cover PHP 7.1 support and let the SA tools operate on PHP 8.1.

derrabus:
If you like, we can merge this PR as-is and I work out a PR for bumping the language level for the SA tools.

derrabus (approved):

beberlei (approved):
Agree about PHPUnit covering 7.1 is good enough.

derrabus:
I've implemented my suggestion as #9314.

PR #9311, opened 2022-01-01
Accessing private properties and methods from the same class is forbidden

### LINKED ISSUE #934: Proxies - Access to private property from the same class scope (recursive) does not initialise Proxy
Recently I came across an issue that was hard to debug and quite unexpected to me.

Code like the one below, which is fully valid PHP code
```php
class Foo {
    /**
      * @ORM\ManyToOne()
      */
    private Foo $parent;

    ... // Always initialize $this->parent in constructor

    public function doSomething(): void {
         var_dump($parent->parent);
    }
}
```

will result in `Typed property accessed before initialization` error.

The situation is even worse with nullable properties which will simply return NULL which is pretty hard to spot.

----------------------------- 

I understand that it's an internal limitation of how currently proxy classes work BUT support for public properties has been added in 2.4 if I recall correctly.

I was not able to find this issue documented anywhere so I'd either:

1) Add visible warning in the documentation to never access private properties directly from the same class.
2) **Handle private/protected properties the same way as public properties are currently handled**

Also, since I've seen plans to move from custom implementation of Proxies to https://github.com/Ocramius/ProxyManager (https://github.com/doctrine/orm/issues/8518), is it possible that this transition will resolve this issue?

olsavmic:
Resolves https://github.com/doctrine/common/issues/934

### DIFF
--- a/docs/en/cookbook/accessing-private-properties-of-the-same-class-from-different-instance.rst
+++ b/docs/en/cookbook/accessing-private-properties-of-the-same-class-from-different-instance.rst
@@ -1,0 +1,74 @@
+Accessing private/protected properties/methods of the same class from different instance
+========================================================================================
+
+.. sectionauthor:: Michael Olsavsky (olsavmic)
+
+As explained in the :doc:`restrictions for entity classes in the manual <../reference/architecture>`,
+it is dangerous to access private/protected properties of different entity instance of the same class because of lazy loading.
+
+The proxy instance that's injected instead of the real entity may not be initialized yet
+and therefore not contain expected data which may result in unexpected behavior.
+That's a limitation of current proxy implementation - only public methods automatically initialize proxies.
+
+It is usually preferable to use a public interface to manipulate the object from outside the `$this`
+context but it may not be convenient in some cases. The following example shows how to do it safely.
+
+Safely accessing private properties from different instance of the same class
+-----------------------------------------------------------------------------
+
+To safely access private property of different instance of the same class, make sure to initialise
+the proxy before use manually as follows:
+
+.. code-block:: php
+
+    <?php
+
+    use Doctrine\Common\Proxy\Proxy;
+    use Doctrine\ORM\Mapping as ORM;
+
+    /**
+     * @ORM\Entity
+     */
+    class Entity
+    {
+        // ...
+
+        /**
+         * @ORM\ManyToOne(targetEntity="Entity")
+         * @ORM\JoinColumn(nullable=false)
+         */
+        private self $parent;
+
+        /**
+         * @ORM\Column(type="string", nullable=false)
+         */
+        private string $name;
+
+        // ...
+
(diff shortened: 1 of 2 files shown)


docs/en/cookbook/accessing-private-properties-of-the-same-class-from-different-instance.rst
```diff
@@ -0,0 +1,74 @@
+Accessing private/protected properties/methods of the same class from different instance
+============================
```
SenseException:
```suggestion
Accessing private/protected properties/methods of the same class from different instance
========================================================================================
```

docs/en/cookbook/accessing-private-properties-of-the-same-class-from-different-instance.rst
```diff
@@ -0,0 +1,74 @@
...
+
+The proxy instance that's injected instead of the real entity may not be initialized yet
+and therefore not contain expected data which may result in unexpected behavior.
+That's a limitation of current proxy implementation - only public methods automatically initialize proxies.
+
+It is usually preferable to use a public interface to manipulate the object from outside the `$this`
+context but it may not be convenient in some cases. The following example shows how to do it safely.
+
+Safely accessing private properties from different instance of the same class
+----------------------------
```
SenseException:
```suggestion
Safely accessing private properties from different instance of the same class
-----------------------------------------------------------------------------
```

docs/en/cookbook/accessing-private-properties-of-the-same-class-from-different-instance.rst
```diff
@@ -0,0 +1,74 @@
...
+
+To safely access private property of different instance of the same class, make sure to initialise
+the proxy before use manually as follows:
+
+.. code-block:: php
+
+    <?php
+
+    use Doctrine\Common\Proxy\Proxy;
```
SenseException, line 26: Please import the mapping for the ORM alias.

docs/en/cookbook/accessing-private-properties-of-the-same-class-from-different-instance.rst
```diff
@@ -0,0 +1,74 @@
+Accessing private/protected properties/methods of the same class from different instance
+============================
+
+.. sectionauthor:: Michael Olsavsky (olsavmic)
+
+As explained in the
+`restrictions for entity classes in the manual <https://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/architecture.html#entities>`_,
```
SenseException: You should use a target instead of using a link to the html file: https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#explicit-targets

SenseException (changes requested):

SenseException (approved):
I'm not sure about the content and need another pair of eyes.

derrabus (approved):

PR #9318, opened 2022-01-02
Document PHPUnit mocks with intersection types

derrabus:
PHPUnit's `createMock()` will return an object that extends the mocked type and implements `MockObject` at the same time. The accurate way to document this is an intersection type, but we often document a union type instead. This declaration is too wide because virtally any mock object would pass it.

This might look nit-picking now, but it will become more important later when we try to infer native types from the docblock types in an automated manner.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/SQLFilterTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/SQLFilterTest.php
@@ -7,7 +7,6 @@ namespace Doctrine\Tests\ORM\Functional;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Types\Types;
 use Doctrine\ORM\Configuration;
-use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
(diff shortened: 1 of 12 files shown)


tests/Doctrine/Tests/ORM/Functional/SQLFilterTest.php
```diff
@@ -221,31 +221,38 @@ public function testEntityManagerIsFilterEnabled(): void
...
+    private function getMockEntityManager(): EntityManagerInterface
     {
-        // Setup entity manager mock
-        return $this->createMock(EntityManager::class);
+        return $this->createMock(EntityManagerInterface::class);
     }
 
-    protected function addMockFilterCollection(EntityManagerInterface $em): FilterCollection
+    /**
+     * @psalm-param EntityManagerInterface&MockObject $em
```
derrabus, line 249: Note: I'm using `@psalm-param` instead of `@param` because CodeSniffer will do strange this otherwise, see squizlabs/PHP_CodeSniffer#2887.

greg0ire (approved):
I think it was done that way because of PHPStorm being unhappy with it at some point, but yes, let's use intersection types 👍

PR #9322, opened 2022-01-02
PSR-6 second level cache

derrabus:
I'd like to continue our effort to sunset Doctrine Cache. This PR enables `DefaultCacheFactory` and `DefaultRegion` to consume a PSR-6 cache item pool instead of a Doctrine Cache instance.

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
+++ b/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
@@ -4,9 +4,9 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Cache;
 
-use Doctrine\Common\Cache\Cache as CacheAdapter;
-use Doctrine\Common\Cache\CacheProvider;
-use Doctrine\Common\Cache\MultiGetCache;
+use Doctrine\Common\Cache\Cache as LegacyCache;
+use Doctrine\Common\Cache\Psr6\CacheAdapter;
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Cache;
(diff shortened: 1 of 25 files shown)


lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
```diff
@@ -47,9 +49,33 @@ class DefaultCacheFactory implements CacheFactory
     /** @var string|null */
     private $fileLockRegionDirectory;
 
-    public function __construct(RegionsConfiguration $cacheConfig, CacheAdapter $cache)
+    /**
+     * @param CacheItemPoolInterface $cacheItemPool
```
greg0ire, line 53: I'd also document the legacy types, so as not to confuse static analysis

derrabus: I intentionally documented the non-deprecated usage only. If downstream projects used a static analysis tool, that tool would start to emit errors which is a good thing imho. It makes this kind of deprecation discoverable.

greg0ire: I think we should stay away from it for the reason @orklah mentioned here: https://symfony-devs.slack.com/archives/C8SFXTD2M/p1641293007273300

derrabus: That discussion did not really convince me, I must admit.

greg0ire: Let's not block this PR for that then.

greg0ire (dismissed):

greg0ire (approved):

greg0ire (approved):

greg0ire:
Thanks @derrabus !

PR #9325, opened 2022-01-03
Use `EntityManagerInterface` in type declarations

derrabus:
The codebase inconsitently switched between `EntityManagerInterface` and `EntityManager` in type declarations. This will get us into trouble if we want to migrate to native type declarations on 3.0.

This PR switches all type declarations to `EntityManagerInterface` where possible.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,12 @@
 # Upgrade to 2.11
 
+## Rename `AbstractIdGenerator::generate()` to `generateId()`
+
+Implementations of `AbstractIdGenerator` have to override the method
+`generateId()` without calling the parent implementation. Not doing so is
+deprecated. Calling `generate()` on any `AbstractIdGenerator` implementation
+is deprecated.
+
 ## PSR-6-based second level cache
 
 The second level cache has been reworked to consume a PSR-6 cache. Using a
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -39,7 +39,7 @@ class EntityRepository implements ObjectRepository, Selectable
     /** @var string */
     protected $_entityName;
 
-    /** @var EntityManager */
+    /** @var EntityManagerInterface */
     protected $_em;
 
     /** @var ClassMetadata */
@@ -282,7 +282,7 @@ class EntityRepository implements ObjectRepository, Selectable
     }
 
     /**
-     * @return EntityManager
+     * @return EntityManagerInterface
      */
     protected function getEntityManager()
     {
--- a/lib/Doctrine/ORM/Event/LifecycleEventArgs.php
+++ b/lib/Doctrine/ORM/Event/LifecycleEventArgs.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Event;
 
-use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\Persistence\Event\LifecycleEventArgs as BaseLifecycleEventArgs;
 
 /**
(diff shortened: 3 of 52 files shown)


greg0ire:
I think @beberlei had a plan to kill that interface, not sure if this is the right move but I think he needs to review this.

derrabus:
@beberlei Do you have input on that topic?

beberlei:
We introduced the interface to help with the use of `EntityManagerDecorator`, since we didn't want people to extend the `EntityManager` and its marked as "soft final".

The problem is in a decoration, code like `new UnitOfWork($this)` or `$this->repositoryFactory->getRepository($this, $entityName);` does not pass the decorator, but the decoratee / original.

Its a hard problem, it seems some people want to extend/decorate the EM for whatever reason, but its not a good solution for us. We could just give up and recommend extending `EntityManager` instead to simplify the problem for us. A lot of time has been wasted into this interface already.

derrabus:
So your plan would be to get rid of the interface (and consequently the abstract decorator) completely? That would generate the harder migration path for applications. Especially in Symfony applications, it is common to type-hint `EntityManagerInterface` if write a service that depends on the entity manager. Also, the interface is quite handy if you want to write mock implementations of the EM.

I agree that decorating the entity manager is not really a good way of extending the ORM. But at this point, I'd keep the interface tbh.

lib/Doctrine/ORM/Id/LegacyIdGeneratorProxy.php
```diff
@@ -0,0 +1,35 @@
...
+
+use Doctrine\Deprecations\Deprecation;
+use Doctrine\ORM\EntityManager;
+
+use function get_debug_type;
+
+/**
+ * @internal This trait will be removed in 3.0.
+ */
+trait LegacyIdGeneratorProxy
```
beberlei: Wouldn't it make more sense to move this into `AbstractIdGenerator` instead?

beberlei (changes requested):
Probably a wishful thinking of me, but yeah it would have been nice. But you are right, we should just go with the interface now. :+1: from me then. One question, i would remove the trait and inline the code into `AbstractIdGenerator`, unless I am not seeing something.

lib/Doctrine/ORM/Id/LegacyIdGeneratorProxy.php
derrabus: That would generate an infinite recursion (`generateId()` and `generate()` would call each other) if you forget to override either method. That's why I moved our internal implementation into a trait.

beberlei: not sure thats something we need to account for, everyone will have a `generate` method already, not having both seems something only new code could achieve. Can't help everyone from shooting themselves in the foot.

derrabus: No, but we don't need to hand them the gun. 🙂

derrabus: I you see a way of moving this logic into the abstract class without having it default to an endless recursion, I'd happily implement it. Otherwise, I really believe we're better of this way.

beberlei: You hand them the gun anyways with anything code that you provide ;) for example you can easily get infinte recursions by doing flush in an UoW event for example. 

A simple workaround would be a private property that tracks you have already delegated `alreadyDelegatedToGenerate` and check that before calling the delegate function again, throwing if already true, otherwise setting it to true.

derrabus: Good idea. Let's do it that way.

beberlei (dismissed):

PR #9326, opened 2022-01-04
Add support for custom types with requireSQLConversion and ResultSetMappingBuilder::generateSelectClause()

kimhemsoe:


derrabus:
Can you elaborate what this change does?

kimhemsoe:
When using ResultSetMappingBuilder::generateSelectClause() with custom types with Type::canRequireSQLConversion() => true, we would not call Type::convertToPHPValueSQL() to inject the required sql. 

I am trying to fix that here when using ResultSetMappingBuilder::*EntityFromClassMetadata() methods. ResultSetMappingBuilder can help with some of the pain of building native queries. 

There is still the case when using ResultSetMappingBuilder::addFieldResult() and etc which is not covered. i am affraid that changing those could maybe break peoples hacks out there. 

Not sure if i can/should fix thee psalm issues. 

Also could be nice if we could do this one place one day. I stole some of this from https://github.com/doctrine/orm/blob/96bc214acd9ce7c103f6253c9c04ef87c6199a46/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1463

PS. Sorry for jumping the gun with review request with failing tests.

kimhemsoe:
Added test using addFieldResult(), works fine and i was wrong that *EntityFromClassMetadata() is a speciel case.

kimhemsoe:
@derrabus How do i solve the psalm issue?

greg0ire:
@kimhemsoe the field you are accessing is documented as possibly unset: https://github.com/doctrine/orm/blob/760397c429362114e18e8420a6eaab0a2d43aa1c/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L76 in #8409 , but looking at the codebase, there are several places where it is assumed to exist, and then the resulting Psalm/Phpstan errors are ignored in the baseline.

I think that after this piece of code is executed, it's always set: https://github.com/doctrine/orm/blob/4117ca349fb1c9f6a5290d22a77181be748d8f1e/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L1554-L1573

greg0ire:
@orklah , do you remember why you marked that key as possibly unset?

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
```diff
@@ -16,6 +17,7 @@
 use function sprintf;
 use function strpos;
 use function strtolower;
+use function var_dump;
```
beberlei: I believe this is unused, not sure why phpcs/psalm don't complain

beberlei (dismissed):
PR is good! This makes sense to add

You should update the baseline of psalm with these two errors included as @greg0ire said its done in other places already, we can then independently look into making it an always existing column in our psalm mapping.

You can update the baseline with `./vendor/bin/psalm --set-baseline=psalm-baseline.xml`

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
greg0ire: That would be because of this: https://github.com/doctrine/orm/blob/760397c429362114e18e8420a6eaab0a2d43aa1c/phpcs.xml.dist#L34-L35

beberlei: Ah forgot about having to add that myself :)

orklah:
> @orklah , do you remember why you marked that key as possibly unset?

Probably a lack of knowledge of Doctrine. What I can see is a property initialized to empty array and not initialized in constructor. But maybe Doctrine make sure to initialize some properties before anything happens with the property

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
greg0ire: Fixed in https://github.com/doctrine/orm/pull/9338

greg0ire:
> What I can see is a property initialized to empty array and not initialized in constructor. 

That's `fieldMappings` with an s, which contains an array of `FieldMapping`, which is true even if it's an empty array.

I think I'll just try removing that `?` and see how SA reacts.

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
```diff
@@ -432,19 +434,26 @@ public function generateSelectClause($tableAliases = [])
...
 
-            $sql .= $tableAlias . '.';
-
             if (isset($this->fieldMappings[$columnName])) {
-                $class = $this->em->getClassMetadata($this->declaringClasses[$columnName]);
-                $sql  .= $class->fieldMappings[$this->fieldMappings[$columnName]]['columnName'];
+                $class             = $this->em->getClassMetadata($this->declaringClasses[$columnName]);
+                $fieldName         = $this->fieldMappings[$columnName];
+                $classFieldMapping = $class->fieldMappings[$fieldName];
+                $columnSql         = $tableAlias . '.' . $classFieldMapping['columnName'];
```
kimhemsoe, line 444: @beberlei What we if i add a FieldMapping class with array access in another PR? How much will break? Or is there another better plan to get rid of some of the many arrays?

beberlei: that is a 3.0 task

greg0ire:
Addressed in https://github.com/doctrine/orm/pull/9339

derrabus:
This does not look like a bugfix. Let's target 2.11.x instead?

kimhemsoe:
"Use var_dump()" removed, rebased and retarget to 2.11.x. psalm issues should be solved been when 2.10.x is merged up to 2.11.x

greg0ire (approved):

greg0ire:
Thanks @kimhemsoe !

PR #9338, opened 2022-01-08
Restore UnusedUse rule

greg0ire:
Detection still seems broken when you have several annotations on one
line, so let us use linebreaks.


derrabus:
We have made that change already on 2.11. Is it really worth the effort backporting it to 2.10? I'd rather like to release 2.11 soon'ish because of the enum feature.

greg0ire:
Oh… I didn't remember, it's indeed not worth it then.

PR #9341, opened 2022-01-09
PHPStan 1.3.3, Psalm 4.18.1

derrabus:


psalm-baseline.xml
```diff
@@ -2192,10 +2191,6 @@
     </PropertyNotSetInConstructor>
   </file>
   <file src="lib/Doctrine/ORM/Query/AST/Functions/SubstringFunction.php">
-    <InvalidScalarArgument occurrences="2">
-      <code>$optionalSecondSimpleArithmeticExpression</code>
-      <code>$sqlWalker-&gt;walkSimpleArithmeticExpression($this-&gt;firstSimpleArithmeticExpression)</code>
-    </InvalidScalarArgument>
```
greg0ire: Some of these would need to be "moved" to a baseline specific to 2.13

greg0ire: See https://github.com/doctrine/orm/pull/9343

greg0ire:
Sorry, I meant to close my own PR…

psalm-baseline.xml
derrabus: I could also move them to psalm.xml.dist and add a comment that this applies to DBAL 2 only. That should stabilize the baseline as well.

greg0ire (approved):

PR #9343, opened 2022-01-09
Use a specific configuration file for DBAL 2

greg0ire:
Just like for PHPStan, we now have 2 configuration files with common
ignore rules. This should make it easier to update the baseline.

Follow-up of https://github.com/doctrine/orm/pull/9336


greg0ire:
Looks like this isn't working… no merge seems to be happening (I've experimenting doing the include after defining the rules, and everything defined in issue-handlers.xml was ignored.

greg0ire:
Because of [this discussion](https://symfony-devs.slack.com/archives/C8SFXTD2M/p1641728669154300), I now think our best bet is to just run Psalm once. PHPStan covering the DBAL 2 should be enough IMO.

PR #9356, opened 2022-01-10
Remove the `composer/package-versions-deprecated` package

derrabus:
see symfony/symfony#44726, doctrine/dbal#5078

With Composer's new plugin policy in mind, we should avoid dropping Composer plugins into an application's dependencies. I suggest to drop the `composer/package-versions-deprecated` dependency and rely on Composer's runtime API instead.

The consequence would be that people who are still on Composer 1 won't be able to upgrade the ORM anymore. Then again, at this point of time, there really should not be a reason anymore to use Composer 1.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -228,8 +228,7 @@ These methods have been deprecated:
 ## Deprecated `Doctrine\ORM\Version`
 
 The `Doctrine\ORM\Version` class is now deprecated and will be removed in Doctrine ORM 3.0:
-please refrain from checking the ORM version at runtime or use
-[ocramius/package-versions](https://github.com/Ocramius/PackageVersions/).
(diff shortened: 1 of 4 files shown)


greg0ire (dismissed):

UPGRADE.md
```diff
@@ -228,8 +228,7 @@ These methods have been deprecated:
 ## Deprecated `Doctrine\ORM\Version`
 
 The `Doctrine\ORM\Version` class is now deprecated and will be removed in Doctrine ORM 3.0:
-please refrain from checking the ORM version at runtime or use
-[ocramius/package-versions](https://github.com/Ocramius/PackageVersions/).
+please refrain from checking the ORM version at runtime or use Composer's runtime API.
```
Seldaek:
```suggestion
please refrain from checking the ORM version at runtime or use Composer's [runtime API](https://getcomposer.org/doc/07-runtime.md#knowing-whether-package-x-is-installed-in-version-y).
```

stof (approved):

greg0ire:
Thanks @derrabus !

PR #9370, opened 2022-01-12
Fix "Illegal mix of collations" in MultiTableDeleteExecutor

mlocati:
When database tables use a collation that's not the default database one, MultiTableDeleteExecutor fails with this error:

```
In AbstractMySQLDriver.php line 128:
                                                                                                                                                    
  [Doctrine\DBAL\Exception\DriverException]                                                                                                         
  An exception occurred while executing 'DELETE FROM MessengerTaskProcesses WHERE (id) IN (SELECT id FROM MessengerProcesses_id_tmp)':              
                                                                                                                                                    
  SQLSTATE[HY000]: General error: 1267 Illegal mix of collations (utf8mb4_unicode_ci,IMPLICIT) and (utf8mb4_general_ci,IMPLICIT) for operation '='
```

This issue should be fixed by using the configured charset/collation.

### DIFF
--- a/lib/Doctrine/ORM/Query/Exec/MultiTableDeleteExecutor.php
+++ b/lib/Doctrine/ORM/Query/Exec/MultiTableDeleteExecutor.php
@@ -49,6 +49,7 @@ class MultiTableDeleteExecutor extends AbstractSqlExecutor
         $conn          = $em->getConnection();
         $platform      = $conn->getDatabasePlatform();
         $quoteStrategy = $em->getConfiguration()->getQuoteStrategy();
+        $schemaConfig  = $conn->getSchemaManager()->createSchemaConfig();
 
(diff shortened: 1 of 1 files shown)


derrabus:
Thank you for your contribution.

* Please add a test that covers your change.
* Please run `./vendor/bin/phpcbf` to make the code style check happy

mlocati:
> Please add a test that covers your change.

I don't think that SQLite supports charsets/collations...

derrabus:
> I don't think that SQLite supports charsets/collations...

We also run tests on MySQL, MariaDB and Postgres. Your test may skip any platform on which your fix is not needed.

beberlei:
@mlocati See `MySqlSchemaToolTest` in the code which is already doing collation/charset related things as a template on how to do a MySQL only test.

mlocati:
Superseded by #9405

PR #9373, opened 2022-01-12
Support enum cases as parameters

### LINKED ISSUE #9372: Add type inferral for backed enums in `Parameter` as part of `QueryBuilder`
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

With the release of v2.11.0 just a few hours ago, support for PHP 8.1's enums was added. This works great, however, I would like to see that the `setParameter()` function (or rather the `Parameter` which is created in that function) has support for automatically inferring the type of backed enums.

Here an example:

```php
enum Suit: int {
    case Hearts = 0;
    case Diamonds = 1; 
    case Clubs = 2;
    case Spades = 3;
}

#[Entity]
class Card
{
    /** ... */

    #[Column(type: "integer", enumType: Suit::class)]
    private Suit $suit;
```

Then at some point you could want to do something along the of:

```php
$qb = $em->getRepository(Card::class)->createQueryBuilder('c');
$qb->where('c.suit = :suit')
    ->setParameter('suit', Suit::Hearts); // instead of Suit::Hearts->value

$qb->getQuery->getResult();
```

However, now you get the following error:

>  Error: Object of class Suit could not be converted to string.

Where I would have expected it to say `integer`. `string` is given because it is the default type (`\Doctrine\DBAL\ParameterType::STRING`) in `\Doctrine\ORM\Query\ParameterTypeInferer::inferType`. Which has no knowledge of what a backed enum is.

Clearly, this is also shows that DBAL is incomplete, as it has no idea what to do with the enum (which type of backed enum does not matter, you could force `\Doctrine\DBAL\Types\Types::INTEGER` in `setParameter` and DBAL would complain about being unable to convert the enum into an integer). However, that is an separate issue and not what this feature request is about (but it does depend on each other).

derrabus:
Fixes #9372

### DIFF
--- a/tests/Doctrine/Tests/ORM/Query/QueryTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/QueryTest.php
@@ -24,14 +24,20 @@ use Doctrine\Tests\Mocks\EntityManagerMock;
 use Doctrine\Tests\Models\CMS\CmsAddress;
 use Doctrine\Tests\Models\CMS\CmsGroup;
 use Doctrine\Tests\Models\CMS\CmsUser;
+use Doctrine\Tests\Models\Enums\AccessLevel;
+use Doctrine\Tests\Models\Enums\City;
+use Doctrine\Tests\Models\Enums\UserStatus;
 use Doctrine\Tests\Models\Generic\DateTimeModel;
 use Doctrine\Tests\OrmTestCase;
 use Generator;
 use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 
+use function array_map;
 use function assert;
 use function method_exists;
+
+use const PHP_VERSION_ID;
 
 class QueryTest extends OrmTestCase
 {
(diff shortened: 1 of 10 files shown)


tests/Doctrine/Tests/ORM/Query/QueryTest.php
```diff
@@ -251,6 +257,10 @@ public function provideProcessParameterValueIterable(): Generator
         yield 'simple_array' => [$baseArray];
         yield 'doctrine_collection' => [new ArrayCollection($baseArray)];
         yield 'generator' => [$gen()];
+
+        if (PHP_VERSION_ID >= 80100) {
+            yield [array_map([City::class, 'from'], $baseArray)];
```
greg0ire:
```suggestion
            yield 'enum' => [array_map([City::class, 'from'], $baseArray)];
```

Let's add phpdoc to the generator to enforce this?

derrabus: I've added a `@psalm-return` annotation.

derrabus: I've fixed `ParameterTypeInfererTest` as well.

greg0ire (approved):

PR #9377, opened 2022-01-13
Compute particular commit order for deletions

### LINKED ISSUE #9376: Doctrine tries to delete entities in incorrect order
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.0

#### Summary

In some cases, Doctrine tries to delete entities in incorrect order and it fails on foreign keys exception. It also happens in 2.10.4

#### Current behavior

I have 6 entities with circular reference, and I tried to delete 2 of them. But `UnitOfWork::commit()` fails because of `ForeignKeyConstraintViolationException`.

The reason is that it first tries to delete the entity set as a foreign key in the second entity, which hasn't been deleted yet.

I investigated it quite a bit and the order of these operations is determined by `$commitOrder` in `UnitOfWork::commit()`. It is calculated from all 6 entities, but for the delete itself, the order of those 2 entities is not correct within those 6 entities. It should be sufficient to compute the order for deletions only with those 2 entities set to be deleted. I prepared a test with described behavior and a possible fix. See linked pull request.

#### How to reproduce

Run the test in a linked pull request with a DB platform supporting foreign key constraints. (i.e., not SQLite)

#### Expected behavior

`UnitOfWork::commit()` should delete entities. I prepared a possible fix in the linked pull request. I created a new method called `getCommitOrderForDeletions()` for computing the correct order of entities set to be deleted. This function is similar to `getCommitOrder()`, but it operates only with entities set to be deleted.

jusephe:
Fixes: https://github.com/doctrine/orm/issues/9376

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -450,8 +450,9 @@ class UnitOfWork implements PropertyChangedListener
 
             // Entity deletions come last and need to be in reverse commit order
             if ($this->entityDeletions) {
-                for ($count = count($commitOrder), $i = $count - 1; $i >= 0 && $this->entityDeletions; --$i) {
-                    $this->executeDeletions($commitOrder[$i]);
+                $commitOrderForDeletions = $this->getCommitOrderForDeletions();
+                for ($count = count($commitOrderForDeletions), $i = $count - 1; $i >= 0 && $this->entityDeletions; --$i) {
+                    $this->executeDeletions($commitOrderForDeletions[$i]);
                 }
             }
 
@@ -1283,7 +1284,7 @@ class UnitOfWork implements PropertyChangedListener
         // are not yet available.
         $newNodes = [];
 
-        foreach (array_merge($this->entityInsertions, $this->entityUpdates, $this->entityDeletions) as $entity) {
+        foreach (array_merge($this->entityInsertions, $this->entityUpdates) as $entity) {
             $class = $this->em->getClassMetadata(get_class($entity));
 
             if ($calc->hasNode($class->name)) {
@@ -1329,6 +1330,49 @@ class UnitOfWork implements PropertyChangedListener
                     }
 
                     $calc->addDependency($targetSubClass->name, $class->name, 1);
+                }
+            }
+        }
+
+        return $calc->sort();
+    }
+
+    /**
+     * Gets the commit order for deletions.
+     *
+     * @return list<object>
+     */
+    private function getCommitOrderForDeletions(): array
+    {
+        $calc = $this->getCommitOrderCalculator();
+
+        $newNodes = [];
+
+        foreach ($this->entityDeletions as $entity) {
+            $class = $this->em->getClassMetadata(get_class($entity));
+
+            if ($calc->hasNode($class->name)) {
+                continue;
+            }
+
+            $calc->addNode($class->name, $class);
+
+            $newNodes[] = $class;
+        }
+
+        // Calculate dependencies for new nodes
+        while ($class = array_pop($newNodes)) {
+            foreach ($class->associationMappings as $assoc) {
+                if (! ($assoc['isOwningSide'] && $assoc['type'] & ClassMetadata::TO_ONE)) {
+                    continue;
+                }
+
+                $targetClass = $this->em->getClassMetadata($assoc['targetEntity']);
+
+                $joinColumns = reset($assoc['joinColumns']);
+
+                if ($calc->hasNode($targetClass->name)) {
+                    $calc->addDependency($targetClass->name, $class->name, (int) empty($joinColumns['nullable']));
                 }
             }
         }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9376Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9376Test.php
@@ -1,0 +1,208 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping as ORM;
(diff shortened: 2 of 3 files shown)


jusephe:
I just pushed new commit to fix "Static Analysis with Psalm".

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -444,8 +444,14 @@ public function commit($entity = null)
 
             // Entity deletions come last and need to be in reverse commit order
             if ($this->entityDeletions) {
-                for ($count = count($commitOrder), $i = $count - 1; $i >= 0 && $this->entityDeletions; --$i) {
-                    $this->executeDeletions($commitOrder[$i]);
+                // There are some cases in which the original $commitOrder is not correct for deletions,
+                // so we need to call a particular function to compute the right order.
+                // For example see tests/Doctrine/Tests/ORM/Functional/Ticket/GH9376Test.php.
+                // Without computing the special order for deletions, it tries to delete the entity set as a foreign key in another entity.
+                // getCommitOrderForDeletions() is similar to getCommitOrder(), but it operates only with entities set to be deleted.
```
SenseException: This should be part of the description of a commit message. Your method name seem to describe well what is happening.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1330,6 +1336,54 @@ private function getCommitOrder(): array
...
+     */
+    private function getCommitOrderForDeletions(): array
+    {
+        $calc = $this->getCommitOrderCalculator();
+
+        // See if there are any new classes in the changeset, that are not in the
+        // commit order graph yet (don't have a node).
+        // We have to inspect changeSet to be able to correctly build dependencies.
+        // It is not possible to use IdentityMap here because post inserted ids
+        // are not yet available.
```
SenseException:
```suggestion
```

SenseException (changes requested):

lib/Doctrine/ORM/UnitOfWork.php
jusephe: fixed

lib/Doctrine/ORM/UnitOfWork.php
jusephe: fixed

SenseException:
It looks like your test does not cover your fix, because when I undo your changes in UnitOfWork, the test doesn't fail. ORM seems to be able to handle the test case.

jusephe:
> 

@SenseException Have you tried with a DB platform supporting foreign key constraints. (i.e., not SQLite)? I tried it again and it still fails...

jusephe:
@SenseException Can I help you somehow with testing the issue? I rebased my branch onto current version and the test still fails if I undo my changes. The test has to be run with a DB platform supporting foreign key constraints. (i.e., NOT SQLite).

SenseException:
Sorry for the late answer and for forgetting about this PR. The commit order calculation isn't an easy topic and I'll have to take more time to review it. I also want to prevent possible impacts on performance.

SenseException:
Please change the base branch to 2.12.x and rebase.

jusephe:
@SenseException I rebased my branch onto 2.12.x. The test still fails if I undo my changes. (The test has to be run with a DB platform supporting foreign key constraints (i.e., NOT SQLite).

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1331,6 +1332,49 @@ private function getCommitOrder(): array
...
+     *
+     * @return list<object>
+     */
+    private function getCommitOrderForDeletions(): array
+    {
+        $calc = $this->getCommitOrderCalculator();
+
+        $newNodes = [];
+
+        foreach ($this->entityDeletions as $entity) {
```
SenseException, line 1351: With `$this->entityDeletions` in here, it wouldn't be needed in `$this->getCommitOrder()` anymore. Otherwise the commit order would calculate a second time using the deletions.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9376Test.php
```diff
@@ -0,0 +1,204 @@
...
+            $this->_em->getClassMetadata(GH9376Order::class),
+            $this->_em->getClassMetadata(GH9376ProductPartner::class),
+            $this->_em->getClassMetadata(GH9376Product::class),
+            $this->_em->getClassMetadata(GH9376Gift::class),
+        ]);
+
+        parent::tearDown();
+    }
+
+    public function testIssueRemove(): void
```
SenseException:
```suggestion
    public function testRemoveCircularRelatedEntities(): void
```

SenseException (commented):
@doctrine/doctrinecore

jusephe:
@SenseException Can I help you somehow with this PR? I rebased my branch onto 2.13.x and the test still fails if I undo my changes.

SenseException:
Thank you for your offer and your patience. There's currently no help required that I know of. We'll come back in case this PR needs changes or maybe new test cases to prevent breaking changes.

mpdude:
This will be fixed by #10547, make sure to track that one.

mpdude:
Also for me, the test case passes on the current 2.16.x branch without further modifications to the UoW. Yes, I am using a DBMS with foreign keys.

Could you please re-check that and/or provide only the failing test in a dedicated PR against 2.16 (and notify me)?

Thanks!

/cc @greg0ire

mpdude:
@jusephe are you still interested in working on this? Could you check if the problem is solved by #10913?

jusephe:
@mpdude Hi, sorry for late answer. The problem is fixed either by https://github.com/doctrine/orm/pull/10547 or https://github.com/doctrine/orm/pull/10913. Thank you very much! I am closing this.

PR #9379, opened 2022-01-13
Fixed #9378 / versioned entities in a class table inheritance cannot …

### LINKED ISSUE #9378: Versioned entities in a class table inheritance cannot be persisted because of thrown `LengthException`
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | I guess yes
| Version     | 2.11.0

#### Summary

Persisting versioned entities in a class table inheritance scenario fails with `LengthException`. Migration notes do not mention any changes regarding versioned entities.

#### Previous behavior

Persisting such entities worked fine in `2.10.3` (previous version I used).

#### Current behavior

Persisting versioned entities in a class table inheritance scenario fails with `LengthException`. I think this is related to commit e369cb6e73f7f6cd2f8fb6c977e1566ed273ca59 which introduced generated/virtual properties and requires `version` column to provide a value before initially flushing the newly created entity in `JoinedSubclassPersister.php`.

Simply moving `assignDefaultVersionAndUpsertableValues()` AFTER the insertion of the joined subclass fixes the issue.

I'll try to provide a PR asap.

#### How to reproduce

Try persisting a versioned entity in a class table inheritance scenario.

#### Example

~~~php
/**
 * @ORM\Entity()
 * @ORM\DiscriminatorColumn(name="type", type="string")
 * @ORM\DiscriminatorTable({
 *     "user_group_contract" = "UserGroupContract"
 * })
 */
abstract class SubjectOfElection {

    /**
     * @ORM\Column(type="guid")
     * @ORM\Id()
     */
    private string $id;

}

/**
 * @ORM\Entity()
 */
class UserGroupContract extends SubjectOfElection {
    /**
     * @ORM\Column(type="integer")
     * @ORM\Version()
     */
    private int $version;
}
~~~

kryptychon:
This is an attempt to fix #9378.


SenseException:
Thank you for your contribution. Please reduce the length of your commit message. You can find an example on our website.

https://www.doctrine-project.org/contribute/index.html#crafting-meaningful-commit-messages

kryptychon:
Hi,

sorry, I'm not familiar with this project's commit message style.
I followed the guide you referenced. If it's still not what you want, please let me know.

derrabus:
Thank you for your PR. My main problem is that you did not add a test that makes sure the bug you're fixing does not resurface on fine day. A test also helps a reviewer to understand what you want to achieve with your change.

kryptychon:
Okay, I get that. I'll have a look in the doctrine/orm testsuite and try my best. To be honest, I'm insecure about what the best approach for testing this would be. Persisting versioned entities in a class table inheritance scenario worked fine and introducing generated/virtual columns broke that. Should the test cover only my change or even this new (and apparently untested) feature as well?

Thanks for your advice.

derrabus:
The test must cover your change. That means, it should fail without your change and pass with your change applied.

You may however add additional tests for existing features if you found that the feature is not tested thoroughly. We're always thankful for improvements to the testsuite because they help us maintain this project in the long run.

PR #9381, opened 2022-01-14
[GH-9380] Bugfix: Delegate ReflectionEnumProperty::getAttributes().

### LINKED ISSUE #9380: ReflectionEnumProperty throws error on AttributeReader::getPropertyAnnotations()
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.0

#### Summary

An Error "Internal error: Failed to retrieve the reflection object" is thrown when calling `AttributeReader::getPropertyAnotations` because the inherited function `ReflectionProperty::getAttributes()` is not called on the original reflection property in the `ReflectionEnumProperty` class. 
In my specific case I 'm using custom annotations I have to read for custom behaviour. When trying to retrieve the custom annotations the error occurs. 

#### Current behavior
When trying to fetch all or a specific annotation of an entity property, the call fails with an error "Internal error: Failed to retrieve the reflection object".

#### How to reproduce

```
enum Suit: string {
    case Hearts = 'H';
    case Diamonds = 'D';
    case Clubs = 'C';
    case Spades = 'S';
}

#[Entity]
class Card
{
    /** ... */

    #[Column(type: 'string', enumType: Suit::class)]
    #[AnyCustomAnnotation(param: 'bla')]
    public $suit;
}

/** ... */

// @var \Doctrine\ORM\EntityManager
$metadata = $objectManager->getClassMetadata(Card::class);

foreach ($metadata->fieldMappings as $field => $mapping) {
    $property = $metadata->getReflectionProperty($field);
    // @var Doctrine\ORM\Mapping\Driver\AttributeReader
    $isCustom = $reader->getPropertyAnnotation($property, AnyCustomAnnotation::class);
    /** ... */
}

```

#### Expected behavior

The expected behaviour should return the property attributes like `ReflectionProperty::getAttributes()` is supposed to. A possible solution could be extending the `ReflectionEnumProperty` class with an inherited `getAttributes` method, that will return the attributes from the original reflection property.

```
class ReflectionEnumProperty extends ReflectionProperty
{
    ...
    /**
     * {@inheritDoc}
     */
     public function getAttributes(?string $name = null, int $flags = 0): array
     {
         return $this->originalReflectionProperty->getAttributes();
     }
}
```

beberlei:
Fixes #9380

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/EnumTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/EnumTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\ORM\Functional;
 
+use Doctrine\ORM\Mapping\Column;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
 use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\ORM\Tools\SchemaTool;
@@ -22,10 +23,10 @@ class EnumTest extends OrmFunctionalTestCase
 {
     public function setUp(): void
     {
+        parent::setUp();
+
         $this->_em         = $this->getEntityManager(null, new AttributeDriver([dirname(__DIR__, 2) . '/Models/Enums']));
         $this->_schemaTool = new SchemaTool($this->_em);
-
-        parent::setUp();
 
         if ($this->isSecondLevelCacheEnabled) {
             $this->markTestSkipped();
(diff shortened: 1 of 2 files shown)


ezkimo (dismissed):
Works out pretty well. Thank you.

derrabus:
Alternatively, you could add a call to the parent constructor to this class' constructor.

tests/Doctrine/Tests/ORM/Functional/EnumTest.php
```diff
@@ -101,4 +102,15 @@ public function provideCardClasses(): array
             TypedCard::class => [TypedCard::class],
         ];
     }
+
+    public function testGetAttributesDelegate(): void
```
greg0ire: There is no delegation anymore, is there? Let's focus on the desired behavior instead of the implementation.
```suggestion
    public function testItAllowsReadingAttributes(): void
```

greg0ire (approved):

PR #9383, opened 2022-01-15
Annotations docs examples actualization

AlexVasechkin:
Unique constraints annotations actualization @Table->@ORM\Table and @UniqueConstraint->@ORM\UniqueConstraint

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -1287,10 +1287,15 @@ Example:
 .. code-block:: php
 
     <?php
+
+    namespace App\Entity;
+
+    use Doctrine\ORM\Mapping as ORM;
+
     /**
      * @Entity
-     * @Table(name="user",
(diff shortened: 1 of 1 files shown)


greg0ire:
Why only these 2 specifically?

AlexVasechkin:
Becose i faced with the problem the example does`t works. I want to fix this problem.

SenseException:
It makes sense to update all namespaces on this page and make all the examples work. Could you please do that?

AlexVasechkin:
I will fix.

beberlei:
@SenseException @greg0ire This is not a small topic to tackle and requires some strategy. Current examples are based on the SimpleAnnotationReader everywhere in the docs at the moment. A change here must also update the docs on setup with Symfony based approach.

SenseException:
@beberlei I agree. But does this affect the current changes of this PR between `SimpleAnnotationReader` and `AnnotationReader`?

PR #9384, opened 2022-01-15
Allow using Enum from different namespace than Entity

javer:
Allows using Enums from different namespace than Entity, for example, `App\DBAL\Enum` and `App\Entity`.
Without this fix it works only when both Enum and Entity are placed in the same namespace like `App\Entity`, and nobody is responsible for autoloading these Enums.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1517,7 +1517,7 @@ class ClassMetadataInfo implements ClassMetadata
                 ! isset($mapping['type'])
                 && ($type instanceof ReflectionNamedType)
             ) {
-                if (PHP_VERSION_ID >= 80100 && ! $type->isBuiltin() && enum_exists($type->getName(), false)) {
+                if (PHP_VERSION_ID >= 80100 && ! $type->isBuiltin() && enum_exists($type->getName())) {
                     $mapping['enumType'] = $type->getName();
 
                     $reflection = new ReflectionEnum($type->getName());


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1517,7 +1517,7 @@ private function validateAndCompleteTypedFieldMapping(array $mapping): array
                 ! isset($mapping['type'])
                 && ($type instanceof ReflectionNamedType)
             ) {
-                if (PHP_VERSION_ID >= 80100 && ! $type->isBuiltin() && enum_exists($type->getName(), false)) {
+                if (PHP_VERSION_ID >= 80100 && ! $type->isBuiltin() && enum_exists($type->getName())) {
```
greg0ire, line 1520: Not sure why autoload was disabled in the first place. @beberlei , do you remember?

derrabus: Triggering the autoloader here is correct. We cannot be certain at this point that the property type has been autoloaded already.

However, this has ~~nothig~~ little to do with the namespace we place the enum in.

derrabus (approved):

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
javer: It does, I've checked this before writing it: with disabled autoloading Enum from the Entity namespace is somehow visible, that's why unit tests working with `Enums\TypedCard` are green. But moving Enum to another namespace break things.

javer:
So do I need to write some test for this or is the fix enough obvious? It's hard to write autoloading tests, because there is no guarantee that someone else doesn't loaded your class already.

beberlei:
Fix is obvious

beberlei (approved):

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
simPod: Can confirm we need removing `false`. Works for me now.

PR #9386, opened 2022-01-15
Corrected ORM version and added missing dependency

mahakaal:
Noticed that the version wasn't updated, pointing to 2.11.0 instead of 2.10.2. 
Also when following this tutotial ran into missing dependency for "doctrine/annotation" so added that too.

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -82,10 +82,10 @@ that directory with the following contents:
     {
         "require": {
             "doctrine/orm": "^2.11.0",
-            "doctrine/dbal": "^3.1.1",
+            "doctrine/dbal": "^3.2",
             "doctrine/annotations": "1.13.2",
-            "symfony/yaml": "2.*",
-            "symfony/cache": "^5.3"
+            "symfony/yaml": "^5.4",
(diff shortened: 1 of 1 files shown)


docs/en/tutorials/getting-started.rst
```diff
@@ -81,8 +81,9 @@ that directory with the following contents:
 
     {
         "require": {
-            "doctrine/orm": "^2.10.2",
+            "doctrine/orm": "^2.11.0",
             "doctrine/dbal": "^3.1.1",
+            "doctrine/annotations": "1.13.2",
             "symfony/yaml": "2.*",
             "symfony/cache": "^5.3"
```
derrabus: Thank you. The other dependencies were also pretty outdated.

```suggestion
            "doctrine/orm": "^2.11.0",
            "doctrine/dbal": "^3.2",
            "doctrine/annotations": "1.13.2",
            "symfony/yaml": "^5.4",
            "symfony/cache": "^5.4"
```

derrabus:
At some point, we should rewrite the tutorial to use attributes instead of annotations. But as long as it uses annotations, we should tell the reader to install the `doctrine/annotations` package to keep things simple.

derrabus (approved):

PR #9390, opened 2022-01-16
Introduce __unserialize behaviour in docs

SenseException:
I'm not sure about the detail in which this should be described in the docs, which is why the first change is another list entry.

This PR mentions the magic method `__unserialize()` in entities. `__wakeup` is already considered a problem when used in an entity and  `__unserialize()` introduces a new behaviour that should be mentioned in the docs: https://3v4l.org/EQ90P

It probably needs a link to the the new methods and maybe a more detailed description about how it should be handled.

### DIFF
--- a/docs/en/reference/architecture.rst
+++ b/docs/en/reference/architecture.rst
@@ -82,9 +82,9 @@ be any regular PHP class observing the following restrictions:
    :doc:`do so safely <../cookbook/implementing-wakeup-or-clone>`.
 -  An entity class must not implement ``__wakeup`` or
    :doc:`do so safely <../cookbook/implementing-wakeup-or-clone>`.
-   Also consider implementing
-   `Serializable <https://php.net/manual/en/class.serializable.php>`_
-   instead.
+   You can also consider implementing
+   `Serializable <https://php.net/manual/en/class.serializable.php>`_,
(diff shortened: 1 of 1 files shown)


docs/en/reference/architecture.rst
```diff
@@ -85,6 +85,9 @@ be any regular PHP class observing the following restrictions:
    Also consider implementing
    `Serializable <https://php.net/manual/en/class.serializable.php>`_
    instead.
+-  PHP 7.4 introduces the new magic method ``__unserialize``, which changes the execution
+   priority between ``__wakeup`` and itself when used. This can cause unexpected behaviour in
+   an Entity.
```
derrabus: I'd simply apply the same warning as for `__wakeup()` because effectively `__unserialize()` replaces `__wakeup()` if it's present.

Also, we should remove the recommendation to implement `Serializable` because that interface is considered deprecated. And `cookbook/implementing-wakeup-or-clone` might need adjustments as well.

SenseException: I agree. `__unserialize()` isn't something that can be added lightly to an entity, but should we discourage from using it as long as ORM can't use this method in entities? Currently it shouldn't be used there. I expect that a future version will deprecate `__wakeup()` in favour of `__unserialize()`.

derrabus (approved):

PR #9392, opened 2022-01-17
Remove old cache accessors and mutators from query classes

derrabus:


tests/Doctrine/Tests/ORM/Functional/HydrationCacheTest.php
```diff
@@ -35,8 +32,7 @@ protected function setUp(): void
 
     public function testHydrationCache(): void
     {
-        $arrayAdapter = new ArrayAdapter();
-        $cache        = method_exists(QueryCacheProfile::class, 'getResultCache') ? $arrayAdapter : DoctrineProvider::wrap($arrayAdapter);
```
derrabus, line 39: This is actually a leftover I cleaned up on the go: `getResultCache` is always defined in DBAL 3.2 and up.

greg0ire (approved):

PR #9400, opened 2022-01-18
Add native types to CacheLogger implementations

derrabus:


lib/Doctrine/ORM/Cache/Logging/CacheLogger.php
```diff
@@ -19,69 +19,69 @@ interface CacheLogger
      * @param string         $regionName The name of the cache region.
      * @param EntityCacheKey $key        The cache key of the entity.
```
greg0ire: This phpdoc is now quite tautological isn't it? Let's reconsider the comments?

derrabus: I've removed the descriptions.

greg0ire (approved):

PR #9401, opened 2022-01-18
Fix types on CacheLogger implementations

derrabus:
Some backports from #9400:

* More precise array types
* Simplified the ternary expressions in `StatisticsCacheLogger`


greg0ire (dismissed):

lib/Doctrine/ORM/Cache/Logging/CacheLoggerChain.php
```diff
@@ -34,9 +34,9 @@ public function getLogger($name)
     }
 
     /**
-     * @return array<CacheLogger>
+     * @return array<string, CacheLogger>
      */
-    public function getLoggers()
+    public function getLoggers(): array
```
derrabus: Oh, I should revert this. 🙈

PR #9402, opened 2022-01-18
Make test types compatible with Doctrine DBAL 4.0.x-dev

morozov:
This is a step towards making the ORM test suite compatible with DBAL 4.0.x-dev.

The test type classes may be reworked by adding return type declarations and remain compatible with DBAL 3.x. Other mock classes (e.g. `DatabasePlatformMock`) will have to be replaced with PHPUnit mocks since it's impossible to implement both the DBAL 3.x and 4.0-x SPIs statically (without code generation).

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC5684Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC5684Test.php
@@ -66,12 +66,12 @@ class DDC5684Test extends OrmFunctionalTestCase
     }
 }
 
-class DDC5684ObjectIdType extends DBALTypes\IntegerType
+class DDC5684ObjectIdType extends DBALTypes\BigIntType
 {
     /**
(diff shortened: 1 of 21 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/DDC5684Test.php
```diff
@@ -66,36 +66,30 @@ public function testFetchObjectWithAutoIncrementedCustomType(): void
     }
 }
 
-class DDC5684ObjectIdType extends DBALTypes\IntegerType
+class DDC5684ObjectIdType extends DBALTypes\BigIntType
```
morozov, line 69: This is a hack to be able to return a non-integer value from `convertToPHPValue()` and have the column autoincremented at the same time. Not sure how valid this test is in the first place.

derrabus (approved):

PR #9411, opened 2022-01-20
convertToDatabaseValue in \Doctrine\ORM\UnitOfWork::getSingleIdentifierValue()

e-kemal:
getSingleIdentifierValue() must return a scalar value, but it doesn't. Because of this, queries with entities in parameters do not work. This occurs on platforms where the string representation of the identifier differs from the representation in the database (for example, MySQL + symfony/uid).

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -429,7 +429,7 @@ abstract class AbstractQuery
         }
 
         try {
-            $value = $this->_em->getUnitOfWork()->getSingleIdentifierValue($value);
+            $value = $this->_em->getUnitOfWork()->getSingleIdentifierDbValue($value);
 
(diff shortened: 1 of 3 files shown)


derrabus:
Please provide a test that covers your change.

derrabus:
Looks like your change breaks quite some existing tests.

e-kemal:
I have added a test that reproduces my issue. My edit to UnitOfWork::getSingleIdentifierValue() fixes the issue. But, judging by the failed tests, I do not quite correctly understand the internal structure of the ORM. In particular, I was confused by the line
```       * @return mixed A scalar value.```
That's why I decided that getSingleIdentifierValue should return a DatabaseValue.
However, it will not work to do the conversion in the external code, since there is no information about the type of the field there (moreover, it is not known which field will be used). Maybe you need to add a separate method for this?

PR #9417, opened 2022-01-21
Fix accessing NULL variable as array in getSelectConditionStatementCo…

daften:
…lumnSQL

$association was initialized, while $assoc can still be NULL. The actual 
data used later is $assocation too.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -1707,7 +1707,7 @@ class BasicEntityPersister implements EntityPersister
                 }
 
                 $joinTableName = $this->quoteStrategy->getJoinTableName($association, $class, $this->platform);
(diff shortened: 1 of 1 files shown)


derrabus:
Thank you for your contribution. Please have a look at the tests that are failing now and please provide a new test that covers your change.

daften:
Hi @derrabus 

The main problem is the automatic support for many to many relations for UniqueEntity isn't there, see https://stackoverflow.com/questions/38655878/doctrine-uniqueentity-with-many-to-many-field. I think it's a combination of bug and this is one that I encountered. I however don't posses the knowledge about ORM or the time to fully fix that. This is just a small part of it.

If this doesn't help, I'd propose closing this, although it does seem like a fairly weird bug.

derrabus:
Let's close the PR then. I cannot merge a PR that breaks existing tests.

I you think you encounter a bug, please file an issue instead with precise steps on how to reproduce your problem.

PR #9418, opened 2022-01-23
Switch tests to the middleware logging system

derrabus:
This PR removes usages of the deprecated `DebugStack` class. If the middleware system is available (DBAL 3), it is used instead of `setSQLLogger()`.

### DIFF
--- a/tests/Doctrine/Tests/DbalExtensions/Connection.php
+++ b/tests/Doctrine/Tests/DbalExtensions/Connection.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DbalExtensions;
+
+use Doctrine\Common\EventManager;
+use Doctrine\DBAL\Configuration;
+use Doctrine\DBAL\Connection as BaseConnection;
+use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\Logging\Middleware as LoggingMiddleware;
+
+use function class_exists;
+
+class Connection extends BaseConnection
+{
+    /** @var QueryLog */
+    public $queryLog;
+
+    public function __construct(array $params, Driver $driver, ?Configuration $config = null, ?EventManager $eventManager = null)
+    {
+        $this->queryLog = new QueryLog();
+        if (class_exists(LoggingMiddleware::class)) {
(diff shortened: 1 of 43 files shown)


tests/Doctrine/Tests/DbalExtensions/Connection.php
```diff
@@ -0,0 +1,33 @@
...
+
+use Doctrine\Common\EventManager;
+use Doctrine\DBAL\Configuration;
+use Doctrine\DBAL\Connection as BaseConnection;
+use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\Logging\Middleware as LoggingMiddleware;
+
+use function class_exists;
+
+class Connection extends BaseConnection
```
morozov, line 15: It would be possible to avoid inheritance if there was a factory that would create and configure the connection, but that's minor.

morozov (dismissed):

tests/Doctrine/Tests/DbalExtensions/Connection.php
derrabus: We do have a factory for that. For now, the wrapper class makes it easier for us to access the logger of a connection. Since we only use that class in tests, I think it's just fine.

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -880,4 +869,59 @@ protected function setUpDBALTypes(): void
...
+        return $connection->queryLog;
+    }
+
+    /**
+     * Using the SQL Logger Stack this method retrieves the current query count executed in this test.
+     */
+    final protected function getCurrentQueryCount(): int
+    {
+        return count($this->getQueryLog()->queries);
+    }
```
greg0ire, line 908: It might be worth also introducing `assertQueryCount(int $expectedCount, string $message)`

greg0ire (approved):

tests/Doctrine/Tests/OrmFunctionalTestCase.php
derrabus: Yeah. 🤔

derrabus: I'll do this in a follow-up.

PR #9419, opened 2022-01-23
Fix support for PHP 8.1 enums in embedded classes

HypeMC:
Fixes using `enumType` that was added in #9304 with embedded classes.


derrabus (approved):

derrabus:
Good catch, thank you!

beberlei:
The code is wrong, getAccessibleProperty can actually be null when used with a `StaticReflectionService` during code generation or reverse engineering using `DisconnectedClassMetadataFactory`.

that is why in the case for a non declaredField there is an extra test for `&& $this->reflFields[$field] !== null`.

derrabus:
> getAccessibleProperty can actually be null

But wouldn't the previous logic also fail in that case? The return value of `getAccessibleProperty()` was previously passed directly to the constructor of `ReflectionEmbeddedProperty` which does not accept `null`.

beberlei:
Hm indeed, its likely embedded and entity generator have never been used in combination or the bug is somewhere deep in the list of 1k tickets ;) 

Since this gets all burned in a fire for 3.0 i am fine keeping it this way

PR #9423, opened 2022-01-24
Introduce `assertQueryCount()`

derrabus:
This PR introduces a new assertion `assertQueryCount()` to make it easier in tests to check how many queries have ben submitted to the database.


greg0ire (dismissed):
Needs a `vendor/bin/phpcbf` but looks good 👍

PR #9429, opened 2022-01-26
Add failing test for failing to update child-parent relation when the child has been prefetched in QB

simPod:
I have discovered this test case that reproduces bug when there's a child relation indexed

```
/**
 * @OneToMany(targetEntity=GH7212Child::class, mappedBy="parent", indexBy="id")
 * @var GH7212Child[]|Collection
 */
protected $children;
```

and then join in qb `->leftJoin('p.children', 'c')`

then the relationship is not getting updated. `$child->setParent(null)` produces no change.

### DIFF
--- a/tests/Tests/Models/GH7212/GH7212Child.php
+++ b/tests/Tests/Models/GH7212/GH7212Child.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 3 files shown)


simPod:
@greg0ire can you reopen please?

simPod:
@greg0ire can you reopen please?

PR #9430, opened 2022-01-26
Comparison when using false value has an error

gnat42:
We had a project where we had the following:

```
    public function setDuplicate(int $id, bool $value): void
    {
        $this->_em
            ->createQueryBuilder()
            ->update(PotentialClient::class, 'w')
            ->set('w.duplicate', $value)
            ->where('w.id = :wId')
            ->setParameter('wId', $id)
            ->getQuery()
            ->execute();
    }
```
When called and setting $value to true, everything worked. When setting it to false, we get an exception from the DB because the statement ends up being `w.duplicate = WHERE...` I traced this to the __toString on the Comparision.php class.

This test causes that to be detected and fail. No fix offered as I wasn't sure what way that should be fixed, but I don't mind also including the fix if anyone here has suggestions on how to implement it.

I should point out that we 'fixed' this in our code by doing `(int)$value` when passing it to the `->set('w.duplicate', $value)` call.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Query/ExprTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/ExprTest.php
@@ -445,4 +445,13 @@ class ExprTest extends \Doctrine\Tests\OrmTestCase
         
         $this->assertEquals(0, $andExpr->count());
     }
+
+    public function testComparisonBooleans()
+    {
+        $comparison = new Expr\Comparison('foo',Expr\Comparison::EQ, true);
+        $this->assertSame('foo = 1', (string)$comparison);
+
+        $comparison = new Expr\Comparison('foo',Expr\Comparison::EQ, false);
+        $this->assertSame('foo = 0', (string)$comparison);
+    }
 }


tests/Doctrine/Tests/ORM/Query/ExprTest.php
```diff
@@ -445,4 +445,13 @@ public function testAddNull() {
         
         $this->assertEquals(0, $andExpr->count());
     }
+
+    public function testComparisonBooleans()
+    {
+        $comparison = new Expr\Comparison('foo',Expr\Comparison::EQ, true);
+        $this->assertSame('foo = 1', (string)$comparison);
```
greg0ire, line 452: isn't `__toString()` supposed to return DQL? Because the EBNF shows `true` and `false` are `true and `false` in DQL: https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html#terminals

gnat42: I don't know the internals enough to say. All I can say is that I have the repo query as stated above, but if I pass false to the set method then I get a QueryException as follows:

```
        "class": "Doctrine\\ORM\\Query\\QueryException",
        "message": "UPDATE NS\\MidwiferyBundle\\Entity\\PotentialClient w SET w.duplicate =  WHERE w.id = :wId",
```

This test duplicates that output. and seems to mirror the issue reported in #1638.

PR #9431, opened 2022-01-27
Public readonly properties are unset in Proxy's constructor

solcik:
### Failing Test

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.0

- #9432
- https://github.com/doctrine/common/issues/954

#### Summary

Public readonly properties are unset in Proxy's constructor, which results in error.

#### Current behavior

`Cannot unset readonly property \Entity\SimpleBook::$title from scope Proxies\__CG__\Entity\SimpleBook`

#### How to reproduce

```php
#[Entity, Table]
class SimpleBook
{
    #[Column, Id, GeneratedValue]
    private readonly int $id;

    #[Column]
    public readonly string $title;

    #[ManyToOne, JoinColumn(nullable: false)]
    private readonly Author $author;

    public function getId(): int
    {
        return $this->id;
    }

    public function getTitle(): string
    {
        return $this->title;
    }

    public function getAuthor(): Author
    {
        return $this->author;
    }
}

class SimpleBook extends \Entity\SimpleBook implements \Doctrine\ORM\Proxy\Proxy
{
    // ...

    public function __construct(?\Closure $initializer = null, ?\Closure $cloner = null)
    {
        unset($this->title);

        $this->__initializer__ = $initializer;
        $this->__cloner__      = $cloner;
    }
}
```

#### Code that generates Proxy's constructor

I dug into ProxyGenerator and here is generation of these unsets:

- https://github.com/doctrine/common/blob/3.2.x/lib/Doctrine/Common/Proxy/ProxyGenerator.php#L979

`public readonly ` properties are not tested as I see in tests.

- https://github.com/doctrine/common/blob/3.2.x/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php#L481

### DIFF
--- a/tests/Doctrine/Tests/Models/ReadonlyProperties/SimpleBook.php
+++ b/tests/Doctrine/Tests/Models/ReadonlyProperties/SimpleBook.php
@@ -19,7 +19,7 @@ class SimpleBook
     private readonly int $id;
 
     #[Column]
-    private readonly string $title;
+    public readonly string $title;
 
     #[ManyToOne, JoinColumn(nullable: false)]
     private readonly Author $author;
--- a/tests/Doctrine/Tests/ORM/Functional/ReadonlyPropertiesTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/ReadonlyPropertiesTest.php
@@ -108,4 +108,22 @@ class ReadonlyPropertiesTest extends OrmFunctionalTestCase
         self::assertEquals($bookId, $book->getId());
         self::assertSame('Jane Austen', $book->getAuthors()[0]->getName());
     }
+
+    public function testEntityProxy(): void
+    {
+        $connection = $this->_em->getConnection();
+
+        $connection->insert('author', ['name' => 'Jane Austen']);
+        $authorId = $connection->lastInsertId();
+
+        $connection->insert('simple_book', ['title' => 'Pride and Prejudice', 'author_id' => $authorId]);
+        $bookId = $connection->lastInsertId();
+
+        $proxyBook = $this->_em->getReference(SimpleBook::class, $bookId);
+
+        self::assertSame('Pride and Prejudice', $proxyBook->title);
+        self::assertEquals($bookId, $proxyBook->getId());
+        self::assertSame('Jane Austen', $proxyBook->getAuthor()->getName());
+    }
+
 }


greg0ire:
The failing test does not seem to fail…

solcik:
Yeah, I saw, and did investigate it now on my local. Sorry, after investigation the problem is with public readonly properties, not nullable ones as I thought. Hopefully updated test case is going to show that. I also updated the description.

greg0ire:
Same player try again :P

solcik:
Sorry for the test not failing./ It is my first attempt for such a large project.) I am doing it blindly, based on Proxies generated in my local project. I do not have local setup to run tests. I am going to do that, but just now leaving for vacation. So you can expect it around 2022-02-08

I put more info into the description. Each public property is unset in Proxy's constructor.

I dig into ProxyGenerator and here is generation of these unsets:

- https://github.com/doctrine/common/blob/3.2.x/lib/Doctrine/Common/Proxy/ProxyGenerator.php#L979

`public readonly ` properties are not tested as I see... 

- https://github.com/doctrine/common/blob/3.2.x/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php#L481

I am going to do failing tests after vacation.

beberlei:
pretty sure the problem is that `SimpleBook` is not used as a proxy in the test code. You can modify the tests to use `$book = $entityManager->getReference(SimpleBook, $id);` and work on the proxy then it should trigger this error.

edipoReboucas:
Aparrently the issue is on `\Doctrine\Common\Proxy\ProxyGenerator::getLazyLoadedPublicPropertiesNames`.

The strategy of unset the property to made it _lazyloadable_ will not work on public readonly property and should be ignored.

solcik:
@greg0ire sorry for the delay, but eventually I got back to it.

@beberlei Thank you for the hint.

I added the test-case, which is failing now locally - hopefully it is going to fail also here .)

christian-kolb:
@greg0ire I'm using only public in entities and this is preventing me from using it entirely. Is there something specific I can do to support you on this ticket? If it's with some kind of implementation, can you point me to the relevant point in the codebase and documentation for the context? 🙂
Thanks a lot in any case.

greg0ire:
@christian-kolb I believe there is something you could do that would help: everyone here is pointing to `doctrine/common` as a culprit, yet this PR is on `doctrine/orm`. Maybe you could find a way to create another failing test that reproduces the bug solely with the `doctrine/common` APIs, and contribute it in a PR to `doctrine/common`? Something that would also help would be to find out why these calls to `unset` exist in the first place? What purpose do they serve?

beberlei:
I believe public readonly properties cannot be supported for entities that are used as proxy. At least i dont see a way we can implement that in the engine, because it requires unset and __get

christian-kolb:
@greg0ire That is what's happening here, isn't it? https://github.com/doctrine/common/pull/957
@beberlei Would that be an option?

Perhaps it makes sense to close this pull request here. At the moment there are 2 issues and 2 pull requests over both packages. I haven't even seen the other pull request until I dived into it further 🙂

beberlei:
It makes the error go away but breaks lazy loading, so it's not an option

christian-kolb:
@beberlei Too bad. So this means you don't think Doctrine will ever be able to offer public readonly properties?
I think this is something that should be added to this article https://www.doctrine-project.org/2022/01/11/orm-2.11.html. The example is done with private properties (I don't know if that was on purpose or just out of habit). But I wouldn't have guessed that it's only possible with private properties.

beberlei:
It depends if we find a way in the engine to hack around this

PR #9440, opened 2022-01-28
Added php 8.1 to CI

sir-kain:
Jobs updated:

- phpunit-postgres
- phpunit-mariadb
- phpunit-mysql

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -78,7 +78,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.4"
+          - "8.1"
         dbal-version:
           - "default"
         postgres-version:
@@ -139,7 +139,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.4"
+          - "8.1"
         dbal-version:
           - "default"
         mariadb-version:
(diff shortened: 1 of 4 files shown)


derrabus:
Modifications to the CI should target the lowest maintained branch instead (2.11.x currently). That being said, we already do test on PHP 8.1. Can you elaborate why you wanted to change the PHP version on those jobs specifically?

sir-kain:
> Modifications to the CI should target the lowest maintained branch instead (2.11.x currently). That being said, we already do test on PHP 8.1. Can you elaborate why you wanted to change the PHP version on those jobs specifically?

I saw that `phpstan`, `phpcs`, `psalm`, ... PHP version are all on `8.1` on the CI and I wanted to use the same version everywhere by default.

greg0ire (dismissed):
So the question is, what should we use for testing platform-specific code? The answer is probably: either the lowest or the highest PHP version. To me, it makes sense to use the highest here, as we are more likely to be using DBAL code we have never been exposed to that way, and catch new bugs.

greg0ire:
> we are more likely to be using DBAL code we have never been exposed to that way, and catch new bugs.

The build is failing, is that what is happening here? :sweat_smile: 

EDIT: the failing test is `Doctrine\Tests\ORM\Functional\ReadonlyPropertiesTest`, which is probably why this wasn't caught before (that test did not run at all for PostgreSQL)

greg0ire:
This is why https://github.com/doctrine/orm/pull/8931 is important

derrabus (approved):

greg0ire:
Thanks @sir-kain !

PR #9443, opened 2022-01-30
Introduce `DoctrineSetup` as a replacement for `Setup`

derrabus:
The `Setup` class is a collection of static methods for bootstrapping a ORM configuration with recommended settings. At least that's my understanding of that class. In applications where the ORM is configured through some kind of framework integration (like DoctrineBundle), that class is more or less irrelevant.

The `Setup` class as of today operates on Doctrine Cache instances. Since the whole ORM is now able to operate on PSR-6 we should do something about it. Unfortunately, since the `Setup` class is not final, altering all method signatures is quite difficult, so I decided to deprecate the whole class and add a new class named `DoctrineSetup` as replacement.

Key differences between the two setup classes:

* The new class is declared `final` which should make future changes to that class easier.
* It accepts PSR-6 caches instead of Doctrine Cache instances.
* If no cache is passed and Symfony Cache is installed, a suitible cache adapter is configured, just like the old class did for Doctrine Cache.
* The `useSimpleAnnotationReader` flag is not supported anymore.

Furthermore, I decided to move the `newDefaultAnnotationDriver()` method from the `Configuration` class to the new `DoctrineSetup` class. That method does not really mutate the configuration, it is a simple factory for the annotation driver. `DoctrineSetup` feels like the better place for it.

### DIFF
--- a/lib/Doctrine/ORM/Tools/DoctrineSetup.php
+++ b/lib/Doctrine/ORM/Tools/DoctrineSetup.php
@@ -1,0 +1,199 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Tools;
+
+use Doctrine\Common\Annotations\AnnotationReader;
+use Doctrine\Common\Annotations\PsrCachedReader;
+use Doctrine\Deprecations\Deprecation;
+use Doctrine\ORM\Configuration;
+use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
+use Doctrine\ORM\Mapping\Driver\AttributeDriver;
+use Doctrine\ORM\Mapping\Driver\XmlDriver;
+use Doctrine\ORM\Mapping\Driver\YamlDriver;
+use Memcached;
+use Psr\Cache\CacheItemPoolInterface;
+use Redis;
+use RuntimeException;
+use Symfony\Component\Cache\Adapter\ApcuAdapter;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
+use Symfony\Component\Cache\Adapter\MemcachedAdapter;
+use Symfony\Component\Cache\Adapter\RedisAdapter;
+
+use function class_exists;
+use function extension_loaded;
+use function md5;
+use function sys_get_temp_dir;
+
+final class DoctrineSetup
+{
+    /**
+     * Creates a configuration with an annotation metadata driver.
+     *
+     * @param string[] $paths
+     */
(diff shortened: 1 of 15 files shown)


lib/Doctrine/ORM/Tools/DoctrineSetup.php
```diff
@@ -0,0 +1,199 @@
...
+        $config = self::createConfiguration($isDevMode, $proxyDir, $cache);
+        $config->setMetadataDriverImpl(new XmlDriver($paths));
+
+        return $config;
+    }
+
+    /**
+     * Creates a configuration with a yaml metadata driver.
+     *
+     * @deprecated YAML metadata mapping is deprecated and will be removed in 3.0
```
derrabus, line 107: It felt a bit odd to introduce a method and instantly deprecate it. I decided to add the method to allow switching to a PSR-6 cache before migrating the whole YAML mapping configuration to XML or annotations.

lib/Doctrine/ORM/Tools/DoctrineSetup.php
```diff
@@ -0,0 +1,199 @@
...
+        ?CacheItemPoolInterface $cache = null
+    ): Configuration {
+        $config = self::createConfiguration($isDevMode, $proxyDir, $cache);
+        $config->setMetadataDriverImpl(new AttributeDriver($paths));
+
+        return $config;
+    }
+
+    /**
+     * Creates a configuration with a xml metadata driver.
```
greg0ire:
```suggestion
     * Creates a configuration with an xml metadata driver.
```

lib/Doctrine/ORM/Tools/DoctrineSetup.php
greg0ire: This reminds me that we should think about deprecating annotations as well… but maybe it's too early?

greg0ire (approved):

beberlei (commented):
Are there no uses of Setup in the docs folder? 

I am wondering if the name shouldbe ORMSetup and if in the spirit of DriverManager in DBAL we should move it to top lebel.

derrabus:
> Are there no uses of Setup in the docs folder?

I totally forgot about the docs, thanks for the hint. #9448

> I am wondering if the name shouldbe ORMSetup and if in the spirit of DriverManager in DBAL we should move it to top lebel.

Yeah, why not. 🤔 I'll look into it tomorrow. 😴

PR #9447, opened 2022-01-30
Remove deprecated Setup class

derrabus:
Follows #9443.


SenseException (dismissed):
Docs can be adapted after the upmerge of #9448.

PR #9448, opened 2022-01-30
Document the new `DoctrineSetup` class

derrabus:
Documentation for the changes of #9443.


SenseException (changes requested):
The docs also contain deprecation notes using `.. note::`. Please add one for `Setup` class.

derrabus:
> The docs also contain deprecation notes using `.. note::`. Please add one for `Setup` class.

I've added such a block to the `docs/en/reference/configuration.rst` file.

SenseException (approved):

PR #9416, opened 2022-01-21
annotations is required

remicollet:
See /usr/share/php/Doctrine/ORM/Configuration.php

PR #9452, opened 2022-01-31
Check requirements for metadata drivers

### LINKED ISSUE #9449: `doctrine/annotations` required for Doctrine Console Tools
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.11.x

#### Summary

`doctrine/annotations` is required when trying to create the schema with [Doctrine Console Tools](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/reference/tools.html)

This issue is a try to clarify what @remicollet tried to explain in report #9416 (/cc @derrabus)

Found myself the [same problem](https://github.com/llaville/php-compatinfo/issues/332) in my project `bartlett/php-compatinfo` 

#### Current behavior

`Fatal error: Uncaught Error: Class 'Doctrine\Common\Annotations\AnnotationRegistry' not found` was raised when running the Doctrine Console Tools `vendor/bin/doctrine`

#### How to reproduce

Begin with a very simple `composer.json` contents : 

```
{
    "minimum-stability": "stable",
    "require": {
        "doctrine/orm": "^2.11"
    }
}
```

On a PHP 7.4.27 platform
Today, when I install the dependencies (`composer update`), I got (`composer show`):

```
doctrine/cache                   2.1.1   PHP Doctrine Cache library is a popular cache implementation that supports many different drivers such as redi...
doctrine/collections             1.6.8   PHP Doctrine Collections library that adds additional functionality on top of PHP arrays.
doctrine/common                  3.2.1   PHP Doctrine Common project is a library that provides additional functionality that other Doctrine projects d...
doctrine/dbal                    3.3.1   Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection and manage...
doctrine/deprecations            v0.5.3  A small layer on top of trigger_error(E_USER_DEPRECATED) or PSR-3 logging with options to disable all deprecat...
doctrine/event-manager           1.1.1   The Doctrine Event Manager is a simple PHP event system that was built to be used with the various Doctrine pr...
doctrine/inflector               2.0.4   PHP Doctrine Inflector is a small library that can perform string manipulations with regard to upper/lowercase...
doctrine/instantiator            1.4.0   A small, lightweight utility to instantiate objects in PHP without invoking their constructors
doctrine/lexer                   1.2.2   PHP Doctrine Lexer parser library that can be used in Top-Down, Recursive Descent Parsers.
doctrine/orm                     2.11.1  Object-Relational-Mapper for PHP
doctrine/persistence             2.3.0   The Doctrine Persistence project is a set of shared interfaces and functionality that the different Doctrine o...
psr/cache                        1.0.1   Common interface for caching libraries
psr/container                    1.1.2   Common Container Interface (PHP FIG PSR-11)
psr/log                          1.1.4   Common interface for logging libraries
symfony/console                  v5.4.3  Eases the creation of beautiful and testable command line interfaces
symfony/deprecation-contracts    v2.5.0  A generic function and convention to trigger deprecation notices
symfony/polyfill-ctype           v1.24.0 Symfony polyfill for ctype functions
symfony/polyfill-intl-grapheme   v1.24.0 Symfony polyfill for intl's grapheme_* functions
symfony/polyfill-intl-normalizer v1.24.0 Symfony polyfill for intl's Normalizer class and related functions
symfony/polyfill-mbstring        v1.24.0 Symfony polyfill for the Mbstring extension
symfony/polyfill-php72           v1.24.0 Symfony polyfill backporting some PHP 7.2+ features to lower PHP versions
symfony/polyfill-php73           v1.24.0 Symfony polyfill backporting some PHP 7.3+ features to lower PHP versions
symfony/polyfill-php80           v1.24.0 Symfony polyfill backporting some PHP 8.0+ features to lower PHP versions
symfony/service-contracts        v2.5.0  Generic abstractions related to writing services
symfony/string                   v5.4.3  Provides an object-oriented API to strings and deals with bytes, UTF-8 code points and grapheme clusters in a ...
```

We may notice that `doctrine/annotations` package is not installed, but it's used by `doctrine/orm`: 
See https://github.com/doctrine/orm/blob/2.11.x/lib/Doctrine/ORM/Configuration.php#L10

Following documentation : https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/reference/tools.html

I create a `config/cli-config.php` file with following contents 

```php
<?php

// replace with file to your own project bootstrap
require_once 'bootstrap.php';

// replace with mechanism to retrieve EntityManager in your app
$entityManager = GetEntityManager();

return ConsoleRunner::createHelperSet($entityManager);
```
And I also create a `bootstrap.php` file with following contents

```php
<?php

use Doctrine\Common\Proxy\AbstractProxyFactory;
use Doctrine\ORM\EntityManager;
use Doctrine\ORM\Tools\Setup;

function GetEntityManager(): EntityManagerInterface
{
    $factory = null;

    $targetFile = 'database.sqlite';
    $dbUrl = sprintf('sqlite:///$PWD/%s', $targetFile);

    $connection = ['url' => $dbUrl];
    $isDevMode = true;
    $cache = null;

    $proxyDir = dirname(__DIR__) . DIRECTORY_SEPARATOR . 'cache';


    $paths = [implode(DIRECTORY_SEPARATOR, [__DIR__, 'Entity'])];
    $config = Setup::createAnnotationMetadataConfiguration($paths, $isDevMode, $proxyDir, $cache);
    if ($isDevMode) {
        // suggested for DEV mode: see Doctrine ORM documentation
        // at https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/advanced-configuration.html#auto-generating-proxy-classes-optional
        $config->setAutogenerateProxyClasses(AbstractProxyFactory::AUTOGENERATE_ALWAYS);
    } else {
        // lazy generation on PROD or TEST modes (i.e: CI)
        $config->setAutogenerateProxyClasses(AbstractProxyFactory::AUTOGENERATE_FILE_NOT_EXISTS);
    }

    return EntityManager::create($connection, $config);
}

```

The Use Case Project structure look like : 
```
.
├── cache
├── composer.json
├── composer.lock
├── config
│   ├── bootstrap.php
│   └── cli-config.php
├── Entity
└── vendor
    ├── autoload.php
    ├── bin
    ├── composer
    ├── doctrine
    ├── psr
    └── symfony
```

At first run attempt, we got following error : 
```
nfigure caches without doctrine/cache 1.11 or symfony/cache. Please add an explicit dependency to either library. in /shared/backups/github/my-sandboxes/vendor/doctrine/orm/lib/Doctrine/ORM/Tools/Setup.php:193
```

So I added `"symfony/cache": "^5.4"` constraint in my `composer.json` file. Update it and run once again to get final error

```
Fatal error: Uncaught Error: Class 'Doctrine\Common\Annotations\AnnotationRegistry' not found in /shared/backups/github/my-sandboxes/vendor/doctrine/orm/lib/Doctrine/ORM/Configuration.php:165
Stack trace:
#0 /shared/backups/github/my-sandboxes/vendor/doctrine/orm/lib/Doctrine/ORM/Tools/Setup.php(75): Doctrine\ORM\Configuration->newDefaultAnnotationDriver(Array, true)
#1 /shared/backups/github/my-sandboxes/config/bootstrap.php(25): Doctrine\ORM\Tools\Setup::createAnnotationMetadataConfiguration(Array, true, '/shared/backups...', NULL)
#2 /shared/backups/github/my-sandboxes/config/cli-config.php(9): GetEntityManager()
#3 /shared/backups/github/my-sandboxes/vendor/doctrine/orm/bin/doctrine.php(41): require('/shared/backups...')
#4 /shared/backups/github/my-sandboxes/vendor/doctrine/orm/bin/doctrine(4): include('/shared/backups...')
#5 {main}
  thrown in /shared/backups/github/my-sandboxes/vendor/doctrine/orm/lib/Doctrine/ORM/Configuration.php on line 165
```

Because I used the simple annotation reader (default behaviour). 
See https://github.com/doctrine/orm/blob/2.11.x/lib/Doctrine/ORM/Tools/Setup.php#L75

#### Expected behavior

Be able to run Doctrine Console Tools (create shema) with Entities that used default `Doctrine\Common\Annotations\SimpleAnnotationReader`

derrabus:
Fixes #9449, replaces #9416.

Our metadata drivers require additional dependencies. If those are not met, the user receives a cryptic error message instead of a helpful one. Especially regarding Doctrine Annotations, I consider this to be a bug because `doctrine/annotations` used to be a mandatory dependency.

Please review thoroughly. Except for the attribute driver, our current CI setup does not allow me to test code that reacts to the missing dependencies.

### DIFF
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -39,11 +39,13 @@ use Doctrine\ORM\Repository\DefaultRepositoryFactory;
 use Doctrine\ORM\Repository\RepositoryFactory;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Persistence\ObjectRepository;
+use LogicException;
 use Psr\Cache\CacheItemPoolInterface;
 use ReflectionClass;
 
 use function class_exists;
 use function method_exists;
+use function sprintf;
 use function strtolower;
 use function trim;
 
@@ -162,6 +164,14 @@ class Configuration extends \Doctrine\DBAL\Configuration
      */
     public function newDefaultAnnotationDriver($paths = [], $useSimpleAnnotationReader = true)
     {
+        if (! class_exists(AnnotationReader::class)) {
+            throw new LogicException(sprintf(
+                'The annotation metadata driver cannot be enabled because the "doctrine/annotations" library'
+                . ' is not installed. Please run "composer require doctrine/annotations" or choose a different'
+                . ' metadata driver.'
+            ));
+        }
+
         AnnotationRegistry::registerFile(__DIR__ . '/Mapping/Driver/DoctrineAnnotations.php');
 
         if ($useSimpleAnnotationReader) {
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -11,6 +11,7 @@ use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\FileDriver;
 use InvalidArgumentException;
+use LogicException;
 use SimpleXMLElement;
 
 use function assert;
@@ -18,6 +19,7 @@ use function constant;
 use function count;
 use function defined;
 use function explode;
+use function extension_loaded;
 use function file_get_contents;
 use function in_array;
 use function simplexml_load_string;
@@ -39,6 +41,13 @@ class XmlDriver extends FileDriver
      */
     public function __construct($locator, $fileExtension = self::DEFAULT_FILE_EXTENSION)
     {
+        if (! extension_loaded('simplexml')) {
+            throw new LogicException(sprintf(
+                'The XML metadata driver cannot be enabled because the SimpleXML PHP extension is missing.'
+                . ' Please configure PHP with SimpleXML or choose a different metadata driver.'
+            ));
+        }
+
         parent::__construct($locator, $fileExtension);
     }
 
--- a/lib/Doctrine/ORM/Mapping/Driver/YamlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/YamlDriver.php
@@ -11,9 +11,11 @@ use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\FileDriver;
 use InvalidArgumentException;
+use LogicException;
 use Symfony\Component\Yaml\Yaml;
 
 use function array_map;
+use function class_exists;
 use function constant;
 use function defined;
 use function explode;
@@ -44,6 +46,14 @@ class YamlDriver extends FileDriver
             'https://github.com/doctrine/orm/issues/8465',
             'YAML mapping driver is deprecated and will be removed in Doctrine ORM 3.0, please migrate to attribute or XML driver.'
         );
+
+        if (! class_exists(Yaml::class)) {
+            throw new LogicException(sprintf(
+                'The YAML metadata driver cannot be enabled because the "symfony/yaml" library'
+                . ' is not installed. Please run "composer require symfony/yaml" or choose a different'
+                . ' metadata driver.'
+            ));
+        }
 
         parent::__construct($locator, $fileExtension);
     }
--- a/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
@@ -11,6 +11,7 @@ use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\AnnotationDriver;
+use LogicException;
 use ReflectionClass;
 use ReflectionMethod;
 use ReflectionProperty;
@@ -20,6 +21,9 @@ use function class_exists;
 use function constant;
 use function defined;
 use function get_class;
+use function sprintf;
+
+use const PHP_VERSION_ID;
 
 class AttributeDriver extends AnnotationDriver
 {
@@ -35,6 +39,13 @@ class AttributeDriver extends AnnotationDriver
      */
     public function __construct(array $paths)
     {
+        if (PHP_VERSION_ID < 80000) {
+            throw new LogicException(sprintf(
+                'The attribute metadata driver cannot be enabled on PHP 7. Please upgrade to PHP 8 or choose a different'
+                . ' metadata driver.'
+            ));
+        }
+
         parent::__construct(new AttributeReader(), $paths);
     }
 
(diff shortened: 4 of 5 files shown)


llaville:
LGTM

lib/Doctrine/ORM/Configuration.php
```diff
@@ -162,6 +164,14 @@ public function setMetadataDriverImpl(MappingDriver $driverImpl)
      */
     public function newDefaultAnnotationDriver($paths = [], $useSimpleAnnotationReader = true)
     {
+        if (! class_exists(AnnotationReader::class)) {
+            throw new LogicException(sprintf(
+                'The annotation metadata driver cannot be enabled because the "doctrine/annotations" library'
+                . ' is not installed. Please run "composer require doctrine/annotations" or chose a different'
```
greg0ire: "choose, chose, chosen" :slightly_smiling_face: 

```suggestion
                . ' is not installed. Please run "composer require doctrine/annotations" or choose a different'
```

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -39,6 +41,13 @@ class XmlDriver extends FileDriver
      */
     public function __construct($locator, $fileExtension = self::DEFAULT_FILE_EXTENSION)
     {
+        if (! function_exists('simplexml_load_string')) {
```
greg0ire: Why not use https://www.php.net/manual/en/function.extension-loaded.php here?

lib/Doctrine/ORM/Mapping/Driver/YamlDriver.php
```diff
@@ -45,6 +47,14 @@ public function __construct($locator, $fileExtension = self::DEFAULT_FILE_EXTENS
             'YAML mapping driver is deprecated and will be removed in Doctrine ORM 3.0, please migrate to attribute or XML driver.'
         );
 
+        if (! class_exists(Yaml::class)) {
```
greg0ire, line 50: Let's use https://github.com/composer/composer/blob/3ae111140facdba8ae82adcd1085e4adfc7d715c/src/Composer/InstalledVersions.php#L95 ?

derrabus: Can you elaborate what we would gain from avoiding `class_exists()` here?

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
derrabus: No strong opinions here. 🤷🏻‍♂️

lib/Doctrine/ORM/Mapping/Driver/YamlDriver.php
greg0ire: It's very hypothetical of course, but in case there is a gigantic BC break and the `Yaml` class no longer exists in this package, it would be confusing for the user if they still have the package installed but get a message letting them know it is not installed. Another benefit apparently is that we can ignore dev requirements with that API. Not sure if we should use it here, but it might come in handy.

derrabus: The error that they would get if the `InstalledVersions` check would pass although that class is gone would be even more cryptic and confusing. Also, apart from `symfony/yaml`, this class could also be provided by `symfony/symfony` (although depending on the monorepo is discouraged).

I think, I'd stick to the `class_exists()` call if you don't mind. This is the class we're interacting with, so checking for its existence feels like the right thing to do.

greg0ire: > The error that they would get if the InstalledVersions check would pass although that class is gone would be even more cryptic and confusing.

The error they would get would be that `Yaml` doe snot exist… it's not cryptic nor confusing IMO, and gives them a better thing to google.

Good point about `symfony/symfony` though, let's stick with `class_exists`

greg0ire (approved):

PR #9466, opened 2022-02-03
Fix types on persisters

derrabus:
While working on adding native types to the persisters for 3.0, I stumbled accross several inaccuate typings that I'd like to correct on the 2.12 branch already.

### DIFF
--- a/lib/Doctrine/ORM/Cache/Persister/Collection/CachedCollectionPersister.php
+++ b/lib/Doctrine/ORM/Cache/Persister/Collection/CachedCollectionPersister.php
@@ -29,14 +29,14 @@ interface CachedCollectionPersister extends CachedPersister, CollectionPersister
     /**
      * Loads a collection from cache
      *
-     * @return PersistentCollection|null
+     * @return mixed[]|null
      */
     public function loadCollectionCache(PersistentCollection $collection, CollectionCacheKey $key);
 
     /**
      * Stores a collection into cache
      *
-     * @param array|mixed[]|Collection $elements
+     * @param mixed[]|Collection $elements
      *
      * @return void
      */
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -14,6 +14,7 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Result;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\MappingException;
@@ -164,7 +165,7 @@ class BasicEntityPersister implements EntityPersister
      * The INSERT SQL statement used for entities handled by this persister.
      * This SQL is only generated once per request, if at all.
      *
-     * @var string
+     * @var string|null
      */
     private $insertSql;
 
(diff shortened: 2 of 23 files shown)


lib/Doctrine/ORM/Cache/Persister/Collection/CachedCollectionPersister.php
```diff
@@ -29,14 +29,14 @@ public function getTargetEntityMetadata();
     /**
      * Loads a collection from cache
      *
-     * @return PersistentCollection|null
+     * @return mixed[]|null
```
derrabus, line 32: This type was just a plain lie. Implementations return arrays, never a `PersistentCollection`.

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -1580,11 +1581,22 @@ public function lock(array $criteria, $lockMode)
...
+        if ($lockMode === null) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/pull/9466',
+                'Passing null as argument to %s is deprecated, pass LockMode::NONE instead.',
+                __METHOD__
+            );
+
+            $lockMode = LockMode::NONE;
+        }
```
derrabus, line 1593: In this method, `$lockMode === null` actually means `LockMode::NONE`. Disallowing `null` here makes the calls more explicit and allows us to narrow down the type to `int`.

I don't know if that method is ever called from outside this repository, so I'm not sure if it's worth mentioning this deprecation in the upgrade log.

tests/Doctrine/Tests/ORM/Cache/Persister/Collection/AbstractCollectionPersisterTest.php
```diff
@@ -66,27 +41,25 @@ protected function setUp(): void
...
      */
     protected function createRegion(): Region
     {
-        return $this->getMockBuilder(Region::class)
-                    ->setMethods($this->regionMockMethods)
-                    ->getMock();
+        return $this->createMock(Region::class);
     }
 
-    protected function createCollection($owner, $assoc = null, $class = null, $elements = null): PersistentCollection
```
derrabus, line 84: The method is never called with more than one argument.

tests/Doctrine/Tests/ORM/Cache/Persister/Entity/AbstractEntityPersisterTest.php
```diff
@@ -258,23 +270,24 @@ public function testInvokeLoadById(): void
...
+        $owner     = (object) [];
 
         $this->entityPersister->expects(self::once())
             ->method('loadOneToOneEntity')
-            ->with(self::equalTo([]), self::equalTo('foo'), self::equalTo(['id' => 11]))
-            ->will(self::returnValue($entity));
+            ->with(self::identicalTo([]), self::identicalTo($owner), self::identicalTo(['id' => 11]))
+            ->willReturn($entity);
 
-        self::assertEquals($entity, $persister->loadOneToOneEntity([], 'foo', ['id' => 11]));
```
derrabus, line 277: We're passing a string `'foo'` here although the method demands an object. As soon as we add native types, this is going to crash.

tests/Doctrine/Tests/ORM/Cache/Persister/Entity/AbstractEntityPersisterTest.php
```diff
@@ -193,11 +199,12 @@ public function testInvokeDelete(): void
 
         $this->entityPersister->expects(self::once())
             ->method('delete')
-            ->with(self::equalTo($entity));
+            ->with(self::identicalTo($entity))
+            ->willReturn(true);
 
         $this->em->getUnitOfWork()->registerManaged($entity, ['id' => 1], ['id' => 1, 'name' => 'Foo']);
 
-        self::assertNull($persister->delete($entity));
```
derrabus, line 200: The method declares `bool` as return type but we assert `null`. We're basically testing a broken mock configuration.

derrabus (commented):
Some comments on why I changed what I've changed:

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
SenseException: I guess everything that is open for extension can result in a BC break.

phpstan-baseline.neon
```diff
@@ -45,6 +45,11 @@ parameters:
 			count: 1
 			path: lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php
 
+		-
+			message: "#^Else branch is unreachable because ternary operator condition is always true\\.$#"
+			count: 1
+			path: lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php
+
```
SenseException, line 52: Where does this happen?

SenseException (approved):
Making tests stricter is a nice addition too.

phpstan-baseline.neon
derrabus: In the method `storeCollectionCache()`, PHPStan believes `$elements instanceof Collection` to be always true, but the parameter `$elements` is declared as `Collection|mixed[]`. This seems to be a false positive.

PR #9471, opened 2022-02-05
Remove ability to clear the UoW partially

derrabus:
Clearing the UoW partially has been deprecated for quite a while. Let's remove the feature from the 3.0.x branch.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,16 @@
 # Upgrade to 3.0
 
+## BC BREAK: Removed ability to partially clear entity manager and unit of work
+
+* Passing an argument other than `null` to `EntityManager::clear()` will raise
+  an exception.
+* The unit of work cannot be cleared partially anymore. Passing an argument to
+  `UnitOfWork::clear()` does not have any effect anymore; the unit of work is
+  cleared completely.
+* The method `EntityRepository::clear()` has been removed.
+* The methods `getEntityClass()` and `clearsAllEntities()` have been removed
+  from `OnClearEventArgs`.
+
 ## BC BREAK: Remove support for Doctrine Cache
 
 The Doctrine Cache library is not supported anymore. The following methods
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -27,14 +27,12 @@ use Doctrine\ORM\Query\Expr;
 use Doctrine\ORM\Query\FilterCollection;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\ORM\Repository\RepositoryFactory;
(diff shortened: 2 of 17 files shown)


lib/Doctrine/ORM/EntityManager.php
```diff
@@ -448,32 +446,17 @@ public function getPartialReference(string $entityName, $identifier): ?object
...
-     * @param string|null $entityName if given, only entities of this type will get detached
+     * @param string|null $objectName The object name (not supported).
      *
-     * @throws ORMInvalidArgumentException If a non-null non-string value is given.
-     * @throws MappingException            If a $entityName is given, but that entity is not
-     *                                     found in the mappings.
+     * @throws ORMInvalidArgumentException If the caller attempted to clear the EM partially by passing an object name.
      */
-    public function clear($entityName = null): void
+    public function clear($objectName = null): void
```
derrabus, line 453: Note: We have to keep the parameter because the parent interface `ObjectManager` demands it. Because we don't use it anymore, It decided to rename it as declared by the parent interface.

UPGRADE.md
```diff
@@ -1,5 +1,16 @@
 # Upgrade to 3.0
 
+## BC BREAK Removed ability to partially clear entity manager and unit of work
```
```suggestion
## BC BREAK: Removed ability to partially clear entity manager and unit of work
```

lib/Doctrine/ORM/EntityManager.php
greg0ire: Looks like it is not deprecated in this other implementation: https://github.com/doctrine/mongodb-odm/blob/2.3.x/lib/Doctrine/ODM/MongoDB/DocumentManager.php#L696-L707

Not sure what is the right move for `doctrine/persistence`… dropping the parameter from the interface would work, but would make static analysis frown:
- https://3v4l.org/bIoRM
- https://phpstan.org/r/bdb4774d-7c20-408e-ac9d-fa5cab987742

Semantically, it feels like there should be an extended interface: https://phpstan.org/r/57d5847a-1f74-4598-8fe4-7ca319263d10

Aaaand I just noticed that we already picked a way: https://github.com/doctrine/persistence/pull/174

Maybe we should talk about the next major releases of persistence and ORM, and which one should happen before the other one? (I'd say persistence)

derrabus: Yes, we should have this discussion. However, as long as we support Persistence 2, this change in its current form is correct. As soon as we drop Persistence 2 (e.g. by moving to Persitence 3) we can revisit this method and drop the parameter.

greg0ire: Yes, should have said this wasn't a blocker, sorry.

greg0ire (approved):

PR #9474, opened 2022-02-05
Fix types on EntityRepository

derrabus:
This PR prepares the introduction of native types to the `EntityRepository` class on 3.0.

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ORM;
 
 use BadMethodCallException;
-use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\AbstractLazyCollection;
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\Common\Collections\Selectable;
 use Doctrine\DBAL\LockMode;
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -302,7 +310,7 @@ protected function getClassMetadata()
      * return a new collection containing these elements.
      *
      * @return LazyCriteriaCollection
-     * @psalm-return Collection<int, T>
+     * @psalm-return LazyCriteriaCollection<int, T>
```
SenseException: Did you chose to replace an interface with an implementation to drop a dependency to the Collections project? Or also because of `Selectable`?

derrabus: I wanted to match the existing `@return` annotation.

SenseException: I'd prefer to use an interface instead of an implementation detail, but I see that phpdoc already lists `LazyCriteriaCollection` above. I'm not sure if it can be set to an interface in 2.12.

derrabus: I'm fine either way. I could also change this to an intersection of `AbstractLazyCollection` and `Selectable` which is probably the contract that we want to expose here if we consider `LazyCriteriaCollection` to be an implementation detail.

SenseException: I think that's a compromise.

derrabus: Done.

SenseException (approved):

PR #9482, opened 2022-02-06
Run tests with stricter error handling

derrabus:


greg0ire:
The warning we are getting seems to happen before any test is executed… maybe this is why `failOnWarning` has no effect?

derrabus:
I guess so.

ci/github/phpunit/mysqli.xml
```diff
@@ -1,12 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="../../vendor/phpunit/phpunit/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="../../../vendor/phpunit/phpunit/phpunit.xsd"
          colors="true"
          beStrictAboutOutputDuringTests="true"
          beStrictAboutTodoAnnotatedTests="true"
          failOnRisky="true"
+         failOnWarning="true"
+         convertDeprecationsToExceptions="true"
```
greg0ire, line 8: According to the docs, this one is already `true` by default

ci/github/phpunit/mysqli.xml
```diff
@@ -1,12 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="../../vendor/phpunit/phpunit/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="../../../vendor/phpunit/phpunit/phpunit.xsd"
          colors="true"
          beStrictAboutOutputDuringTests="true"
          beStrictAboutTodoAnnotatedTests="true"
          failOnRisky="true"
+         failOnWarning="true"
```
greg0ire: Will this ever kick in? Since `convertWarningsToExceptions` is `true` by default?

ci/github/phpunit/mysqli.xml
derrabus: This has changed recently, see https://github.com/sebastianbergmann/phpunit/blob/9.5/ChangeLog-9.5.md#9510---2021-09-25

ci/github/phpunit/mysqli.xml
derrabus: Probably not. 🤔 Let's remove it?

ci/github/phpunit/mysqli.xml
greg0ire: Ok, let's keep it then.

greg0ire (approved):

PR #9485, opened 2022-02-07
Remove ability to flush the EM partially

derrabus:


lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
```diff
@@ -135,14 +135,6 @@ public function find($className, mixed $id, ?int $lockMode = null, ?int $lockVer
         return $this->wrapped->find($className, $id, $lockMode, $lockVersion);
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function flush($entity = null): void
-    {
-        $this->wrapped->flush($entity);
-    }
```
greg0ire, line 144: Why is this method removed from this class but not from `EntityManager`?

derrabus: The parent class `ObjectManagerDecorator` already has an implementation of this method without parameters. We don't need to override it anymore.

greg0ire (approved):

PR #9486, opened 2022-02-07
JoinedSubclassPersister fix nonInsertable columns with JOINED inheritance type + test

### LINKED ISSUE #9467: Using `nonInsertable/nonUpdateable` columns with `JOINED` inheritance type triggers SQL error
### Bug report

`JoinedSubclassPersister::getInsertColumnList()` does not take care of `notInsertable`/`notUpdateable` property of field mapping in parent class, therefore the column(s) appears in INSERT/UPDATE statement.

This wrong behaviour generates a `DriverException` with the following message, because `BasicEntityPersister::prepareUpdateData()` (correctly) does not provide the appropriate changeset:
>An exception occurred while executing a query: SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens

**ORM version: 2.11**

Follow up to https://github.com/doctrine/orm/pull/9118

gtgt:
Fixes #9467

There wasn't any check for fields with notInsertable mapping at JoinedSubclassPersister's getInsertColumnList. This causes more column in the resulting SQL Insert query than expected when insertable=false is set on a field of joined subclass.
A test is created for this case.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Persisters/JoinedSubclassPersisterTest.php
+++ b/tests/Doctrine/Tests/ORM/Persisters/JoinedSubclassPersisterTest.php
@@ -4,8 +4,13 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\ORM\Persisters;
 
+use Doctrine\DBAL\Logging\DebugStack;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Mapping\Builder\ClassMetadataBuilder;
 use Doctrine\ORM\Persisters\Entity\JoinedSubclassPersister;
+use Doctrine\Tests\Mocks\ConnectionMock;
+use Doctrine\Tests\Models\JoinedInheritanceType\AnotherChildClass;
+use Doctrine\Tests\Models\JoinedInheritanceType\ChildClassWithNonWritableFields;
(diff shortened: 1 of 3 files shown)


derrabus:
Can you please run `./vendor/bin/phpcbf` and commit the result? That should make the code style check happy.

gtgt:
> Can you please run `./vendor/bin/phpcbf` and commit the result? That should make the code style check happy.

Sure. Applied now.

tests/Doctrine/Tests/ORM/Persisters/JoinedSubclassPersisterTest.php
```diff
@@ -37,4 +43,21 @@ public function testExecuteInsertsWillReturnEmptySetWithNoQueuedInserts(): void
     {
         self::assertSame([], $this->persister->executeInserts());
     }
+
+    public function testNonInsertablePropertyInChildClass(): void
```
beberlei, line 47: I would like this to be a functional test instead. We want to slowly move away from these mock tests to functional tests.

The entity should also be in the `Upsertable` model space, best to add one InsertableChild and InsertableParent there, and then move the test to `InsertableUpdateableTest` instead.

mrVrAlex:
You can close this due to https://github.com/doctrine/orm/pull/10598

PR #9488, opened 2022-02-08
Remove ability to merge detached entities

derrabus:
The `EntityManager::merge()` method has been deprecated for a long time (#7901). This PR removes this feature from the 3.0 branch.


lib/Doctrine/ORM/EntityManager.php
```diff
@@ -542,35 +542,15 @@ public function detach($entity): void
...
-     * @param object $entity The detached entity to merge into the persistence context.
-     *
-     * @return object The managed copy of the entity.
-     *
-     * @throws ORMInvalidArgumentException
-     * @throws ORMException
+     * @psalm-return never
      */
-    public function merge($entity): object
+    public function merge($object): object
```
derrabus, line 551: We need to keep the method for compatibility with Persistence 2. We can drop it as soon as we move to Persistence >= 3.

greg0ire (approved):

SenseException (approved):

PR #9490, opened 2022-02-08
Add native types to QueryBuilder

derrabus:


derrabus:
~~Marking as draft until #9492 is merged up.~~

lib/Doctrine/ORM/Query/Parameter.php
```diff
@@ -15,50 +15,33 @@ class Parameter
 {
     /**
      * Returns the internal representation of a parameter name.
-     *
-     * @param string|int $name The parameter name or position.
-     *
-     * @return string The normalized parameter name.
      */
-    public static function normalizeName($name)
+    public static function normalizeName(int|string $name): string
```
SenseException, line 19: Technically, it's just `string` because it gets immediately casted. `$name` also suggests a string by its variable naming. It's a major release so maybe we should consider just to use `string`.

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -483,7 +428,7 @@ public function getRootAliases()
      * @return string[]
```
SenseException, line 428: Should the psalm-annotation be enough at this point and `@return` can be removed?

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -698,7 +602,7 @@ public function getMaxResults()
      *
      * @return $this
      */
-    public function add($dqlPartName, $dqlPart, $append = false)
+    public function add(string $dqlPartName, string|object|array $dqlPart, bool $append = false): static
```
SenseException, line 602: `@param` isn't needed anymore.

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -862,9 +760,9 @@ public function delete($delete = null, $alias = null)
      *
      * @return $this
      */
-    public function update($update = null, $alias = null)
+    public function update(?string $update = null, ?string $alias = null): static
```
SenseException, line 757: @param isn't needed anymore.

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -1063,7 +954,7 @@ public function leftJoin($join, $alias, $conditionType = null, $condition = null
      *
      * @return $this
      */
-    public function set($key, $value)
+    public function set(string $key, mixed $value): static
```
SenseException, line 948: @param isn't needed anymore.

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -1458,10 +1323,10 @@ private function getReducedDQLQueryPart(string $queryPartName, array $options =
      *
      * @return $this
      */
-    public function resetDQLParts($parts = null)
+    public function resetDQLParts(?array $parts = null): static
```
SenseException, line 1317: @param isn't needed anymore.

lib/Doctrine/ORM/Query/Parameter.php
derrabus: I believe the idea here is to allow integers for ordered parameters which is nice from a DX perspective. I would keep it as it doesn't really hurt us. Especially now that we have union types.

lib/Doctrine/ORM/QueryBuilder.php
derrabus: We haven't done so in other similar situations. But we could consider dropping the vendor-prefix for simple generics like this one. But let's do this in a separate PR.

lib/Doctrine/ORM/QueryBuilder.php
derrabus: It is because it's more specific than the native type.

SenseException (approved):

PR #9493, opened 2022-02-08
Added "false" value to $columnPrefix type declaration.

gordinskiy:
If i passing "false" value as third argument of ClassMetadataBuilder::addEmbedded method, I get an error from PHPStan:
`Parameter #3 $columnPrefix of method Doctrine\ORM\Mapping\Builder\ClassMetadataBuilder::addEmbedded() expects string|null, false given.`

It is a valid value according [documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/tutorials/embeddables.html#column-prefixing) and work well.


derrabus (approved):

derrabus:
Thank you!

PR #9497, opened 2022-02-09
Add support for array of enums

sanderdlm:
Allow the use of 'array' and 'simple_array' types in combination with the enumType attribute parameter.

This is my first PR to the project, any feedback or instructions are welcome.

I had to duplicate the try/catch which introduces a bit of redundancy. Should I make a new method for this?


derrabus:
Thank you for your PR. As this is a new feature, please target the 2.12.x branch. Also, we need tests that cover your change.

beberlei:
This is a great idea, i will comment on some code improvements later

derrabus:
By the way, you can run all those complaining tools locally and get faster feedback. You don't have to push your changes through our CI everytime.

Once you've installed the dependencies with `composer update`, the following tools should be available:

* `./vendor/bin/phpcs`: PHP CodeSniffer, the tool that performs our code-style check.
* `./vendor/bin/phpcbf`: PHP Code Beautifier, a tool that can fix most PHPCS errors automatically
* `./vendor/bin/phpstan`: PHPStan (static code analysis)
* `./vendor/bin/psalm`: Psalm (the other static code analyzer)

sanderdlm:
Cheers, think I've cleared them all now :slightly_smiling_face:

dreadnip:
I've noticed an issue while testing.

Using the `array` type, this solutions works, and enums are correctly serialized into the database:
```a:2:{i:0;E:23:"App\Enum\Weekday:Monday";i:1;E:23:"App\Enum\Weekday:Friday";}```

Using the `simple_array` type, the enum does not pass the `implode(',', $value)` step:
`Object of class App\Enum\Weekday could not be converted to string`

Maybe another type called `enum_array` could be introduced that would check for the content of the array and call `->value` on the members before imploding? Or this could be built into `simple_array`?

File: lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php [hunk 1]
```diff
@@ -67,18 +75,37 @@ public function setValue($object, $value = null): void
...
-
-                throw MappingException::invalidEnumValue(
-                    get_class($object),
-                    $this->originalReflectionProperty->getName(),
-                    (string) $value,
-                    $enumType,
-                    $e
-                );
+
+            if (is_array($value)) {
```
beberlei, line 74: What do you think of refactoring it to have less duplication?

Yo could introduce a private helper method `initializeEnumValue` that extracts what is currently in line 81-94 and 96-108. I believe its exactly the same code or did I miss something?

beberlei (changes requested):
Initially I was wondering if the generic `is_array` or else handling isn't a big "dangerous", because there is no connection to the `Type` instance, where we know if value is an array or not, essentially allowing to "use this wrong".

But since getValue/setValue only happen to be used on data coming from the database, not from user input, I haven't found a simple example where this is not programmers error.

I am inclined to ignore that for now and see what is coming out of this change.

Very happy about the change, its simple addition allowing arrays of enums makes sense.

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
sanderdlm: The code is exactly the same yeah. I've added the helper in [01a1b75](https://github.com/doctrine/orm/pull/9497/commits/01a1b75086982bed44c1be46c40aeca1f613b774).

The only thing I'm not sure about is the wording of '$item', but I couldn't come up with anything better.

derrabus:
How can we move forward here? The PR needs a rebase at least.

sanderdlm:
I can't replicate the phpcs failure locally.

derrabus:
> I can't replicate the phpcs failure locally.

I've pushed a commit that should fix it.

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -56,6 +58,12 @@ public function getValue($object = null)
             return null;
         }
 
+        if (is_array($enum)) {
+            return array_map(static function ($item): mixed {
```
derrabus:
```suggestion
            return array_map(static function (BackedEnum $item): mixed {
```

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -66,22 +74,38 @@ public function getValue($object = null)
...
+                $value = $this->initializeEnumValue($object, $value);
             }
         }
 
         $this->originalReflectionProperty->setValue($object, $value);
     }
+
+    /**
+     * @param object $object
+     * @param mixed  $value
```
derrabus:
```suggestion
     * @param int|string $value
```

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -66,22 +74,38 @@ public function getValue($object = null)
...
-
-                throw MappingException::invalidEnumValue(
-                    get_class($object),
-                    $this->originalReflectionProperty->getName(),
-                    (string) $value,
-                    $enumType,
-                    $e
-                );
+            if (is_array($value)) {
+                $value = array_map(function ($item) use ($object): mixed {
```
derrabus:
```suggestion
                $value = array_map(function ($item) use ($object): BackedEnum {
```

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
sanderdlm: Changed in [b6e7472](https://github.com/doctrine/orm/pull/9497/commits/b6e7472f91ffe29851eb08093f6f8c1e8eb817a8)

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
sanderdlm: Changed in [b6e7472](https://github.com/doctrine/orm/pull/9497/commits/b6e7472f91ffe29851eb08093f6f8c1e8eb817a8)

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
sanderdlm: Changed in [b6e7472](https://github.com/doctrine/orm/pull/9497/commits/b6e7472f91ffe29851eb08093f6f8c1e8eb817a8)

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
sanderdlm: Oh, the parameter should be `int|string|array`, no?

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -56,32 +58,54 @@ public function getValue($object = null)
...
+        }
+
         return $enum->value;
     }
 
     /**
-     * @param object $object
-     * @param mixed  $value
+     * @param object     $object
+     * @param int|string $value
```
derrabus: This should silence PHPStan:

```suggestion
     * @param object                    $object
     * @param int|string|int[]|string[] $value
```

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
derrabus: Yes, my bad.

derrabus: Actually no, `int|string` is correct here.

derrabus (dismissed):

derrabus:
Thank you @dreadnip!

derrabus (approved):

PR #9501, opened 2022-02-10
Constructor-Argument "options" has the same type as the associated property.

rejinka:
The constructor argument "options" of ``Doctrine\ORM\Mapping\Index`` wasn't properly alligned with it's property according to its type. Psalm then isn't happy about constructions like following:

```php
#[\Doctrine\ORM\Mapping\Index(
    fields: ['name'],
    name: 'name_idx',
    options: [
        'lengths' => [255],
    ]
)]
final class Entity
{
    // ...
}
```

I'm pretty sure that this change can backported to older versions.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Index.php
+++ b/lib/Doctrine/ORM/Mapping/Index.php
@@ -31,10 +31,10 @@ final class Index implements Annotation
     public $options;
 
     /**
-     * @param array<string>|null $columns
-     * @param array<string>|null $fields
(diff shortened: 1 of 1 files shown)


greg0ire (approved):

greg0ire:
> I'm pretty sure that this change can backported to older versions.

You picked the oldest supported version, so it's fine as is.

SenseException (approved):

PR #9502, opened 2022-02-10
Remove `Proxy` marker interface

derrabus:
Apparently, there are plans to switch to ProxyManager. I'm not sure if someone is actively working on that. Anyway, we can already remove the `Proxy` interface.


SenseException (dismissed):
Is lib/Doctrine/ORM/Proxy/Autoloader.php worth mentioning in the UPGRADE.md?

I'd like to have another reviewer who might know more about the ProxyManager topic.

beberlei:
I am not sure its a good idea to switch to proxy manager at this point, since it often takes way past release dates until its fully supported. We managed to support new php versions instantly with Proxy generator.

Removing the interface is tricky, because there is no good way to raise a deprecation if code outside doctrine instanceofs the orm interface.

derrabus:
> I am not sure its a good idea to switch to proxy manager at this point, since it often takes way past release dates until its fully supported. We managed to support new php versions instantly with Proxy generator.

True. Shall we revert the corresponding deprecations then?

> Removing the interface is tricky, because there is no good way to raise a deprecation if code outside doctrine instanceofs the orm interface.

That outside code would need to work with a static analysis tool. We cannot emit a runtime deprecation, that's true.

What would be the way forward? Undo the deprecation and keep this interface?

beberlei:
@derrabus I think its prudent to remove the deprecation. @greg0ire What do you think?

greg0ire:
A runtime deprecation in the file would kick in as soon as a proxy is instantiated I think, so yeah, it does not look like a good option. Let's revert the deprecation I suppose.

derrabus:
#9532

derrabus:
Closing in favor of #9532

PR #9504, opened 2022-02-11
Fix AbstractQuery::setParameter phpdoc

dmaicher:
This can be `Connection::PARAM_INT_ARRAY` or `Connection::PARAM_STR_ARRAY` for example and those are `int`

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -374,8 +374,8 @@ abstract class AbstractQuery
      * @param string|int      $key   The parameter position or name.
      * @param mixed           $value The parameter value.
      * @param string|int|null $type  The parameter type. If specified, the given value will be run through
-     *                           the type conversion of this type. This is usually not needed for
-     *                           strings and numeric types.
+     *                               the type conversion of this type. This is usually not needed for
+     *                               strings and numeric types.
      *
      * @return $this
      */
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -371,11 +371,11 @@ public function setParameters($parameters)
...
      * Sets a query parameter.
      *
-     * @param string|int  $key   The parameter position or name.
-     * @param mixed       $value The parameter value.
-     * @param string|null $type  The parameter type. If specified, the given value will be run through
-     *                           the type conversion of this type. This is usually not needed for
-     *                           strings and numeric types.
+     * @param string|int      $key   The parameter position or name.
+     * @param mixed           $value The parameter value.
+     * @param string|int|null $type  The parameter type. If specified, the given value will be run through
```
greg0ire, line 376: Should we make this more precise?
```suggestion
     * @param string|Connection::PARAM_*|ParameterType::*|null $type  The parameter type. If specified, the given value will be run through
```

dmaicher: Is that something IDE's and static analyzers understand? :thinking:

greg0ire: Static analyzers, yes. IDE, I don't know (I use neovim). Please test if you can.

derrabus: AFAIK, we use this kind of literal type syntax only in annotations with `psalm-` vendor prefix, e.g. `@psalm-param` in this case.

derrabus (approved):

lib/Doctrine/ORM/AbstractQuery.php
derrabus: Anyway, let's not go down that road in a bugfix PR.

PR #9508, opened 2022-02-13
Add native types to caches

derrabus:
This PR adds native types to classes in the `Cache` namespace.

~~Waiting for #9507 to be merged up.~~

### DIFF
--- a/lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php
+++ b/lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php
@@ -7,10 +7,8 @@ namespace Doctrine\ORM\Cache\Persister\Collection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\Common\Util\ClassUtils;
-use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Cache\CollectionCacheKey;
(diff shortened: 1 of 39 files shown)


lib/Doctrine/ORM/Cache/Persister/Collection/AbstractCollectionPersister.php
```diff
@@ -171,39 +170,4 @@ public function loadCriteria(PersistentCollection $collection, Criteria $criteri
...
-    {
-        $targetPersister = $this->uow->getEntityPersister($targetEntity);
-        assert($targetPersister instanceof CachedEntityPersister);
-        $targetRegion = $targetPersister->getCacheRegion();
-        $key          = new EntityCacheKey($targetEntity, $this->uow->getEntityIdentifier($element));
-
-        $targetRegion->evict($key);
-
-        $this->cacheLogger?->entityCachePut($targetRegion->getName(), $key);
-    }
```
SenseException: Their removal should be mentioned in UPGRADE.md. It doesn't seem like 2.12.x mention a deprecation either.

derrabus: The methods looked like internal code that is just not used anymore. But you are right, since this class is meant to be extended, I probably should document the deprecation properly. #9512

SenseException: Better safe than BC complaints :-)

SenseException (approved):

PR #9511, opened 2022-02-15
Replace `TreeWalkerChainIterator` with a generator

derrabus:
While investigating the type errors on the tree walker classes, I stumbled on `TreeWalkerChainIterator`. This class seems unnecessarily complex and can easily replaced with a three-line generator function. So I did just that.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -2700,54 +2700,6 @@
     <ParamNameMismatch occurrences="1">
       <code>$condPrimary</code>
     </ParamNameMismatch>
-    <PossiblyNullReference occurrences="46">
-      <code>walkAggregateExpression</code>
-      <code>walkArithmeticExpression</code>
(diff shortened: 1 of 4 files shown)


psalm-baseline.xml
```diff
@@ -2700,54 +2700,6 @@
...
-      <code>walkSimpleSelectExpression</code>
-      <code>walkStateFieldPathExpression</code>
-      <code>walkStringPrimary</code>
-      <code>walkSubselect</code>
-      <code>walkSubselectFromClause</code>
-      <code>walkUpdateClause</code>
-      <code>walkUpdateItem</code>
-      <code>walkUpdateStatement</code>
-      <code>walkWhereClause</code>
-    </PossiblyNullReference>
```
greg0ire, line 2750: Nice!

greg0ire (approved):

Issue #9516, opened 2022-02-15
Invalid Bidirectional Association Mapping with Mapped Superclass in Class Hierarchy


bobdercole:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.1

#### Summary

I receive a mapping exception after defining an inverse-side association on an abstract entity. A mapped superclass extends the abstract entity. A concrete entity extends the mapped superclass.

#### Current behavior

ClassMetadataFactory is currently throwing an exception if there is an inverse-side association on a mapped superclass.

https://github.com/doctrine/orm/blob/e8e61cbbd56566c27bc5f766ac017d7921eb5577/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L381

This is correct, as stated in the [documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/reference/inheritance-mapping.html#mapped-superclasses). However, I believe this limitation should be imposed only if the association is on the mapped superclass itself.

#### How to reproduce

- Define an abstract entity (A).
- Define a mapped superclass (B) extends A.
- Define a concrete entity (C) extends B.
- Define another entity (D) that has a many-to-one relationship with A.

I will follow-up with a failing test case.

#### Expected behavior

I think the above mapping should be valid since the association is not defined directly on the mapped superclass.

bobdercole:
Here is my failing test case: #9517.

My proposed fix is to add a condition to [ClassMetadataFactory](https://github.com/doctrine/orm/blob/e8e61cbbd56566c27bc5f766ac017d7921eb5577/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L381) to only throw the exception if the association is directly on the mapped superclass. Maybe something like this?

```php
if ($mapping['type'] & ClassMetadata::TO_MANY && ! $mapping['isOwningSide'] && $parentClass->name === $mapping['sourceEntity']) {
    throw MappingException::illegalToManyAssociationOnMappedSuperclass($parentClass->name, $field);
}
```

mpdude:
Related to #8415 (as per #9517).

mpdude:
Please check if this is already fixed in the current 2.14.x branch

mpdude:
You might also want to subscribe to #10398

bobdercole:
> Please check if this is already fixed in the current 2.14.x branch

Yes! It works now. Thank you for all the work you've done to resolve it!

[...]

PR #9517, opened 2022-02-15
Add Failing Test Case for Invalid Bidirectional Association Mapping with Mapped Superclass in Class Hierarchy

bobdercole:
See #9516.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9516Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9516Test.php
@@ -9,14 +9,12 @@ use Doctrine\ORM\Mapping\Entity;
 use Doctrine\ORM\Mapping\Id;
 use Doctrine\ORM\Mapping\ManyToOne;
 use Doctrine\ORM\Mapping\MappedSuperclass;
-use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\ORM\Mapping\OneToMany;
-use Doctrine\ORM\Tools\SchemaValidator;
 use Doctrine\Tests\OrmFunctionalTestCase;
 
 class GH9516Test extends OrmFunctionalTestCase
 {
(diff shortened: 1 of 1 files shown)


mpdude:
Is this related to #8415 somehow?

bobdercole:
Yes! The inheritance hierarchy you have defined is the same as mine.

```php
/**
 * @Entity
 */
class GH9516Passenger
{
    /**
     * @var int $id
     * @Id
     * @Column(type="integer")
     */
    private $id;

    /**
     * @var GH9516Vehicle $vehicle
     * @ManyToOne(targetEntity="GH9516Vehicle", inversedBy="passengers")
     */
    private $vehicle;
}

/**
 * @Entity
 */
abstract class GH9516Vehicle
{
    /**
     * @var int $id
     * @Id
     * @Column(type="integer")
     */
    private $id;

    /**
     * @var GH9516Passenger[] $passengers
     * @OneToMany(targetEntity="GH9516Passenger", mappedBy="vehicle")
     */
    private $passengers;
}

/**
 * @MappedSuperclass
 */
abstract class GH9516Car extends GH9516Vehicle
{
}

/**
 * @Entity
 */
class GH9516SportsCar extends GH9516Car
{
}
```

The big difference is the association. Mine defines a one-to-many bidirectional relationship on the base class. My issue occurs when validating the mapping; an exception is thrown. 

Your fix almost solves this issue. My test cases passes if I add your [extra condition](https://github.com/doctrine/orm/pull/8415/files#diff-3e426fe47b58d2d4b970b1b7a0cbf5999af950e4f84e48dce4bee5061b57ed31R385) to the illegal association on mapped superclass check, like so:

```php
if ($parentClass->isMappedSuperclass) {
    if ($mapping['type'] & ClassMetadata::TO_MANY && ! $mapping['isOwningSide'] && ! isset($mapping['inherited'])) {
    }

    if (! isset($mapping['inherited'])) {
    }
}
```

PR #7905, opened 2019-11-17
[Paginator] Fix type conversion during hydration of pagination limit subquery

lcobucci:
Continuation of #7890 and #7903, however, using an implementation that keeps DB->PHP conversion contained in the hydrator.

PR #9519, opened 2022-02-15
Trigger the desired code path

lcobucci:
Since v2.7.0 the ORM avoids using extra queries via the paginator component when the maximum results parameter isn't set on the original query. With that change, this test was not executing the code path that it was expected to run.

This makes sure we trigger the forcing of custom DBAL types when hydrating the identifiers, making sure we don't introduce bugs.

More info:
- Forcing DBAL type conversion: https://github.com/doctrine/orm/pull/7905
- Issue on optimisation: https://github.com/doctrine/orm/issues/7829
- PR on optimisation: https://github.com/doctrine/orm/pull/7863
- Minor BC-break docs: https://github.com/doctrine/orm/blob/2.12.x/UPGRADE.md#minor-bc-break-paginator-output-walkers-arent-be-called-anymore-on-sub-queries-for-queries-without-max-results


greg0ire (approved):

greg0ire:
Thanks @lcobucci !

PR #9521, opened 2022-02-17
Remove unused `FilterCollection::$filterHash` field

mpdude:
This field is never updated and only used as long as the collection is `clean`. This `clean`  state is flawed in its own way, deserving a separate PR.

Maybe #9523 is a better approach?


mpdude:
Oh, PHPStan knew it all the time.

Do I need to resolve the PHPStan failures somehow?

derrabus:
> Maybe #9523 is a better approach?

I think so.

derrabus:
Closing in favor of #9523

Issue #7980, opened 2020-01-13
SQLFilters are too locked down


oojacoboo:
I created a pull request for an issue with SQLFilters related to the EntityManager, which is a dependency of the `SQLFilter` class but not exposed to child classes.  However, I'm now faced with another issue related to SQLFilters that is more of the result of the same mentality that continues to haunt me with Doctrine.

I'm certain this won't be met with open arms, because it's honestly just an disagreement with approach.  But everything in Doctrine makes the assumption that developers using the library do not know what they're doing and do not need to extend it freely.

Situation:

I need to create an `SQLFilter` that reads annotations.  Due to performance implications, I need to be able to used the `CachedReader`, as the annotations are already cached, and being that it's a filter which is applied on every query, using the standard `AnnotationReader` without cache, isn't really sufficient.  The performance hit is real.

Problem:

Filters are locked down tight.  You do not have access to the constructor and the filters, after being constructed, are not easily accessible.  So, adding dependencies is a bit more of a challenge, and nasty, and prone to future BC breaks.

```php
$entityManager->getFilters()->getEnabledFilters()['multi-tenancy']->setAnnotationReader($cachedAnnotationReader);
```

That is really nasty.  It shouldn't be.  I also should be able to write filters and use this lib as needed for proper implementation.  Without SQLFitlers, this ORM simply wouldn't work.  We need proper multi-tenancy support, which needs to be applied via Filters for relational entity access, etc.  Annotations are the most logical way of handling this.

I'm requesting that SQLFilter design be re-visited.  I'd also encourage a bit of a change in mentality towards locking down the library, but I realize that's just a difference in opinion and most likely unwelcomed.

SenseException:
Hi @oojacoboo, can you please evaluate in which direction you want to see the filters go? Maybe you can provide a code-example for showing what you describe.
Also: Can you please explain in an code example what your custom filter is doing? For now I understand that you'e unhappy with the lack of using a constructor.

oojacoboo:
Hey @SenseException.  I've published the extension.  Maybe it'd be helpful to have a look over the Filter class there.

https://github.com/rentpost/doctrine-multi-tenancy

As you can see, there are a number of hacks.  Efforts to eliminate all those hacks would be my recommendation.

- Access to the constructor is definitely a start
- Access to the EntityManager
- A better way of managing state
- Ability to use the cached annotation reader

kgilden:
[...]

PR #9522, opened 2022-02-17
Remove `final` from `SQLFilter::__toString()`

mpdude:
This allows filter implementations to provide their own logic for computing cache keys.

#3955 describes a bug where caching the DQL -> SQL transformation result in the query cache does not correctly take into account that filters may generate different SQL depending on circumstances.

As long as filters only depend on their own parameters, the [results are probably correct](https://github.com/doctrine/orm/issues/3955#issuecomment-1042726216).

But, the [SoftDeleteable filter](https://github.com/doctrine-extensions/DoctrineExtensions/blob/2def628a9a8304c1e7f56847a0b21778a3a9ed61/src/SoftDeleteable/Filter/SoftDeleteableFilter.php) is an example for a filter that depends on "external" information, not primarily on filter parameters. Filters like that could benefit from being able to provide their own `__toString()` implementation.

(https://github.com/doctrine-extensions/DoctrineExtensions/pull/2112 adds a filter parameter to make sure `__toString()` changes, but I consider this to be a work-around)

### DIFF
--- a/lib/Doctrine/ORM/Query/Filter/SQLFilter.php
+++ b/lib/Doctrine/ORM/Query/Filter/SQLFilter.php
@@ -175,7 +175,7 @@ abstract class SQLFilter
      *
      * @return string String representation of the SQLFilter.
      */
-    final public function __toString()
+    public function __toString()
     {
         return serialize($this->parameters);
     }


derrabus (changes requested):
Thank your for your PR and even more for the analysis of the issue.

I find it astonishing that I'm supposed to override `__toString()` to influence the cachability of a filter. The fact that the string cast is used to compute the hash is an implementation detail that we should not turn into an extension point.

The parameters system is specifically build for influencing the hash computation, so populating the parameters as it has been done in doctrine-extensions/DoctrineExtensions#2112 seems like the right way.

mpdude:
Agree regarding that `__toString()` is not a good extension point.

What about adding a dedicated method for this purpose?

IMHO, the parameters system is built for passing parameters to filters, but not for busting the query cache. https://github.com/doctrine-extensions/DoctrineExtensions/pull/2112 is really a quirky hack, calling `setParameter()` only because it leads to a separate cache entry down the road. To me, any sane usage of `setParameter()` would have matching `getParameter()` call somewhere else.

derrabus:
> IMHO, the parameters system is built for passing parameters to filters, but not for busting the query cache.

What is a parameter in that context? The filter is supposed to emit pure SQL (without any parameter placeholders), and the parameters accessors are never even called by the ORM anywhere.

Yes, cache busting is all that those parameters do at the moment.

> To me, any sane usage of `setParameter()` would have matching `getParameter()` call somewhere else.

In your example that would mean that `SoftDeleteableFilter` would remove their own `disabled` property and use the parameters array instead.

This parameter system might not be the pinnacle of modern software architecture, but its intention is to remove some pitfalls from cache key computation. For instance, in your example, it makes sure that the order of classes for which `disableForEntity()` is called does not matter.

greg0ire:
This seems to relate to https://github.com/doctrine/orm/issues/7980

derrabus:
To be clear: I'm open to discuss how we can improve this parameter concept or if we can replace it with something better for 3.0. What I'd like to avoid is building a second way to achieve the same goal.

mpdude:
Yeah, filter parameters are a bit crude.

I've seen them as a way how you can configure a filter. For example, a `TenantFilter` would require that you (the user) provide it with a `tenantId` at some point. 

Since you cannot instantiate filters yourself (you only pass a FQCN to the configuration at some point), I considered setting filter parameters as the only way to pass data to filters.

The `SoftDeletableFilter` is an example for a filter that is very clever in reaching out and finding the data it needs. It does not need the parameter to be configured by users (haven't really checked... does it look at annotations or what?).

mpdude:
Maybe we should at least leave a note in the documentation that filter implementors need to use parameters to get the caching right?

derrabus:
Sure, a PR for the docs is more than welcome. We could also add a note to the doc blocks; who reads the docs anyway. 😅

beberlei:
Could we introduce a new method instead? Since this method is final, it should be no problem to just switch the code serializing the filters to use the method instead of __toString.

derrabus:
> Could we introduce a new method instead? Since this method is final, it should be no problem to just switch the code serializing the filters to use the method instead of __toString.

Sure we can. Should that new method complement the current parameter system or replace it? If a filter class implements its own cache busting, the parameters serve no purpose anymore. We might consider deprecating it or moving it e.g. to an optional trait.

mpdude:
> If a filter class implements its own cache busting, the parameters serve no purpose anymore.

I must be missing something…? Aren’t parameters the way to provide filters with values like e.g. a tenant id?

derrabus:
> Aren’t parameters the way to provide filters with values like e.g. a tenant id?

Or a list of classes for which a filter should be disabled? Yes, pretty much.

The thing is: the ORM does not need those parameters as a contract. It will only ever call the constructor and `addFilterConstraint()` on the filter instance. The rest of the `SQLFilter` class is convenience methods to make userland implementations fool-proof: cache-busting, parameter escaping, etc.

I understand that the `SoftDeleteableFilter` does not need the escaping for that class exclusion list since those values are not inserted into actualy SQL, but instead they prevent the generation of an actual filter constraint. Yet, that list is a parameter to the filter because it affects the filter's behavior.

So my point is: The author of the `SoftDeleteableFilter` class decided to opt-out of the one mechnism designed to solve their problem and we're discussion a way to make their code work regardless. So what is the problem we want to solve here?

derrabus:
I'm closing the PR for now because we've agreed that making `__toString()` non-final is not a good solution. If you have an idea how to improve the DX of the SQLFilter class, please open a new PR.

PR #9523, opened 2022-02-17
Bring `FilterCollection` to a "clean" state after hash computation

mpdude:
### Improvement

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | no

Effectively, the `FilterCollection` was never brought back into a "clean" state after filters were enabled or filter parameters were set.

As far as I can tell, the clean/dirty distinction only refers to the computed FilterCollection hash reflecting all enabled filters and their parameters.

The only place I could find where the distinction matters is here:

https://github.com/doctrine/orm/blob/152c04c03d68d39f485d367713dd69dec0f4106d/lib/Doctrine/ORM/Query.php#L241-L259

When the filter collection is dirty and the query cache is used, the call to `getQueryCacheId()` in the last line shown will reset the filter collection to a `clean` state.

When there is no query cache, however, `getQueryCacheId()` is not called and the `FilterCollection` will not revert to a clean state; this prevents re-using the query later on.

If desired, I could add a `FilterCollection::setFiltersStateClean()` method to improve this case as well.

Closes #9521

### DIFF
--- a/lib/Doctrine/ORM/Query/FilterCollection.php
+++ b/lib/Doctrine/ORM/Query/FilterCollection.php
@@ -192,7 +192,7 @@ class FilterCollection
             $filterHash .= $name . $filter;
         }
 
-        $this->filterHash = md5($filterHash);
+        $this->filterHash   = $filterHash;
         $this->filtersState = self::FILTERS_STATE_CLEAN;
 
         return $filterHash;
--- a/lib/Doctrine/ORM/Query/FilterCollection.php
+++ b/lib/Doctrine/ORM/Query/FilterCollection.php
@@ -97,8 +97,7 @@ class FilterCollection
             // Keep the enabled filters sorted for the hash
             ksort($this->enabledFilters);
 
-            // Now the filter collection is dirty
-            $this->filtersState = self::FILTERS_STATE_DIRTY;
+            $this->setFiltersStateDirty();
         }
 
         return $this->enabledFilters[$name];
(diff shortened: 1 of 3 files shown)


derrabus:
Thank you. Can you please provide a test that reproduces the issue that your attempt to fix?

lib/Doctrine/ORM/Query/FilterCollection.php
```diff
@@ -194,6 +192,9 @@ public function getHash()
             $filterHash .= $name . $filter;
         }
 
+        $this->filterHash = md5($filterHash);
```
derrabus: This change looks dangerous. The first call to the method will return the raw `$filterHash` while subsequent calls will return the MD5 sum. Why do you MD5 that value anyway?

mpdude: That was a knee-jerk reaction, to prevent building up a long concatenated string when all we need is change detection – similar to what `\Doctrine\ORM\Query::getQueryCacheId` does. 

I'll remove the `md5()` call.

mpdude:
> Can you please provide a test that reproduces the issue that your attempt to fix?

I don't know what I am trying to fix here 🐼 .

I came across this clean/dirty state distinction and figured it makes no sense the way it is implemented right now.

This PR is my guess what the original author might have intended and I'd be glad if this gets a bit more scrutiny.

derrabus (dismissed):

derrabus:
I've added a test.

derrabus (approved):

PR #9524, opened 2022-02-18
Add native types to SQLFilter

derrabus:


lib/Doctrine/ORM/Configuration.php
```diff
@@ -482,7 +483,7 @@ public function addFilter($name, $className)
      *
      * @return string|null The class name of the filter, or null if it is not
      *  defined.
-     * @psalm-return ?class-string
+     * @psalm-return class-string<SQLFilter>|null
```
SenseException: I assume this PR is about filter classes and not related methods?

SenseException (dismissed):

lib/Doctrine/ORM/Configuration.php
derrabus: This change was backported and dropped out of this PR after merging up and rebasing.

PR #9525, opened 2022-02-18
Fix more types on EntityRepository and FilterCollection

derrabus:
Backports from #9515 and #9524

### DIFF
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -34,6 +34,7 @@ use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
 use Doctrine\ORM\Mapping\EntityListenerResolver;
 use Doctrine\ORM\Mapping\NamingStrategy;
 use Doctrine\ORM\Mapping\QuoteStrategy;
+use Doctrine\ORM\Query\Filter\SQLFilter;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\ORM\Repository\DefaultRepositoryFactory;
 use Doctrine\ORM\Repository\RepositoryFactory;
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -3018,7 +3018,7 @@ protected function _storeAssociationMapping(array $assocMapping)
      * Registers a custom repository class for the entity class.
      *
      * @param string|null $repositoryClassName The class name of the custom mapper.
-     * @psalm-param class-string<EntityRepository>|null $repositoryClassName
+     * @psalm-param class-string<ObjectRepository>|null $repositoryClassName
```
greg0ire: :thinking: what makes you widen this type?

derrabus: Custom repositories don't have to extend `EntityRepository`. Is is sufficient if they implement `ObjectRepository`.

https://github.com/doctrine/orm/blob/193c3abf0e5f7a7e2699e7c7ac37c1e713d79ef8/lib/Doctrine/ORM/Exception/InvalidEntityRepository.php#L14

https://github.com/doctrine/orm/blob/193c3abf0e5f7a7e2699e7c7ac37c1e713d79ef8/lib/Doctrine/ORM/Configuration.php#L832

https://github.com/doctrine/orm/blob/193c3abf0e5f7a7e2699e7c7ac37c1e713d79ef8/lib/Doctrine/ORM/Repository/RepositoryFactory.php#L21

derrabus: Hmm, but this also means that the types on the entity manager are wrong. I'll remove the widening for now and try to straighten this out in a separate PR.

derrabus: Moved to #9531.

greg0ire (approved):

PR #9526, opened 2022-02-19
Make creating test models more straightforward and revert to swallowing exceptions

greg0ire:
In https://github.com/doctrine/orm/pull/8962, I established that
swallowing exceptions while creating model was bad because it could hide
other helpful exceptions.
As it turns out however, swallowing exceptions is really the way to go
here, because of the performance implication of calling dropSchema(). It
is possible to catch a more precise exception as well, which should
preserve the benefits of not swallowing them.

It looks like I based my grep on the comment inside the catch and today,
I found many other occurrences of this pattern, without the easy to grep
comment.

I decided to fix them as well, but in a lazier way: one no longer has to
remember to call dropSchema in tearDown() now, it is automated.

Here is the `grep` I used this time: `rg  --multiline  '\} catch \(Exception \$e\) \{\n\s+\}'`

## TODO
- [x] fix other 25 occurrences
- [x] `grep` for `dropSchema` and improve tests changed in #8962 as well.

### DIFF
--- a/tests/Doctrine/Tests/OrmFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/OrmFunctionalTestCase.php
@@ -19,8 +19,10 @@ use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Exception\ORMException;
+use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Tools\DebugUnitOfWorkListener;
 use Doctrine\ORM\Tools\SchemaTool;
+use Doctrine\ORM\Tools\ToolsException;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Tests\DbalTypes\Rot13Type;
 use Doctrine\Tests\EventListener\CacheMetadataListener;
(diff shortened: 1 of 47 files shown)


greg0ire:
@derrabus I'm requesting your review now because it's the most legible and interesting part.

EDIT: less so now that I have completed the first item of that todo list

beberlei:
DropSchema might be a bad idea, because schema creation is so expensive except for sqlite

greg0ire:
@beberlei I didn't check that in https://github.com/doctrine/orm/pull/8962

Once this is completed, I can rework `createSchemaForModels` to use exception swallowing and drop the `@after` method, that way we will be able to measure the difference.

EDIT: oh now it's failing because some tests create `cms_user` without dropping it :disappointed:  I will have to bisect this.

greg0ire:
@beberlei it looks like the performance loss is so big I can even "feel" it, even with SQLite. Let's revert to swallowing exceptions. I'll use a more specific exception, which should fix the issue I was aiming to fix in #8962 (I remember getting mad because the test was hiding an exception from me, which was probably from a different hierarchy that the one we want to catch).

greg0ire:
@beberlei now that I reverted what I did in #8962: there seems to be a big performance gain (or rather, I introduce a big performance loss before :disappointed: ) :
- https://github.com/doctrine/orm/actions/runs/1869689952 9m26s
- https://github.com/doctrine/orm/actions/runs/1857819286 15m20s

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -337,6 +339,22 @@ abstract class OrmFunctionalTestCase extends OrmTestCase
         ],
     ];
 
+    /**
+     * @param list<class-string> $models
```
derrabus:
```suggestion
     * @psalm-param class-string ...$models
```

derrabus:
LGTM, once the doc block is fixed. The code is much more readable this way, very nice.

PR #9527, opened 2022-02-20
Document QueryComponent array shape

derrabus:
I'm currently improving the types on SqlWalkers/TreeWalkers for 3.0. One issue I found was that the knowledge over the structure of the parser's internal `$queryComponents` array was spread/copied over multiple classes.

With this PR, I'm introducing a reusalbe Psalm type to streamline the type declarations.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -60,6 +60,7 @@ use Doctrine\ORM\Query\AST\UpdateItem;
 use Doctrine\ORM\Query\AST\UpdateStatement;
 use Doctrine\ORM\Query\AST\WhenClause;
 use Doctrine\ORM\Query\AST\WhereClause;
+use LogicException;
 use ReflectionClass;
 
 use function array_intersect;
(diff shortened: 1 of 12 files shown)


lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -3621,4 +3641,13 @@ public function CustomFunctionsReturningStrings()
 
         return $function;
     }
+
+    private function getMetadataForDqlAlias(string $dqlAlias): ClassMetadata
+    {
+        if (! isset($this->queryComponents[$dqlAlias]['metadata'])) {
+            throw new LogicException(sprintf('No metadata for DQL alias: %s', $dqlAlias));
```
SenseException, line 3648: Even though this is a LogicException, it can cause BC problems. Why introducing it in 2.x?

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -225,20 +222,22 @@ public function getEntityManager()
...
      */
     public function getQueryComponent($dqlAlias)
     {
         return $this->queryComponents[$dqlAlias];
     }
 
+    public function getMetadataForDqlAlias(string $dqlAlias): ClassMetadata
+    {
+        if (! isset($this->queryComponents[$dqlAlias]['metadata'])) {
+            throw new LogicException(sprintf('No metadata for DQL alias: %s', $dqlAlias));
```
SenseException, line 235: Same question here.

lib/Doctrine/ORM/Query/Parser.php
derrabus: The pieces of code that I've replaced with a call to this method did previously access the `metadata` key without prior check and afterwards assume that the result is a `ClassMetadata` instance.

**If** that `LogicException` is ever raised (I don't expect it to), it is raised on execution paths that triggered hard errors previously anyway. This is why I don't expect this change to be a BC break.

lib/Doctrine/ORM/Query/SqlWalker.php
derrabus: Same answer. 🙃

SenseException: Same resolve.

SenseException (approved):

PR #9531, opened 2022-02-20
Allow custom repository implementations to skip EntityRepository

derrabus:
This change might be controversial.

Currently, the ORM requires repositories to be implementations of the generic `ObjectRepository` interface only. The ORM's own `EntityRepository` class is the default and can be used as base class for custom implementations, but we don't have to use it.

On the other hand, `EntityManagerInterface::getRepository()` promises to always return a `EntityRepository`. This works fine now, but will fail if we migrate to native return types.

This PR widens all return types to hint at the more generic `ObjectRepository` interface. It also adds a functional test that makes sure, generic `ObjectRepository` implementations can be used.

However, I'm unsure if using generic `ObjectRepository` implementations is a case that we actually want to support. The alternative could be to deprecate not using an `EntityRepository` sub-class and trigger an error on 3.0.

### DIFF
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -824,6 +824,7 @@ class Configuration extends \Doctrine\DBAL\Configuration
      * Sets default repository class.
      *
      * @param string $className
+     * @psalm-param class-string<ObjectRepository> $className
      *
      * @return void
      *
@@ -844,7 +845,7 @@ class Configuration extends \Doctrine\DBAL\Configuration
      * Get default repository class.
      *
      * @return string
-     * @psalm-return class-string
+     * @psalm-return class-string<ObjectRepository>
      */
     public function getDefaultRepositoryClassName()
     {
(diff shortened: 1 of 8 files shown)


beberlei:
I think i am against this, EntityRepository adds a bunch more things that we show in the documentation for example, `createQuery` for example. Additionally, `EntityRepository` implements both `ObjectRepository` and `Selectable`, changing the return type will make the `matching` method invisible or?

derrabus:
> I think i am against this, EntityRepository adds a bunch more things that we show in the documentation for example, `createQuery` for example.

Yes, a pure `ObjectRepository` implementation would not have those. But those methods are helpers for userland code as far as I can tell. From the ORM's PoV, we don't need them.

> Additionally, `EntityRepository` implements both `ObjectRepository` and `Selectable`, changing the return type will make the `matching` method invisible or?

A static analyzer (like PhpStorm or Psalm) would need an additional plugin to know what the repository class for a given entity is, if that's what you mean. Those plugins already exist.

But given the current documentation, not receiving an `EntityRepository` when calling `$em->getRepository()` will probably generate some astonishment. Maybe it would be the better way to close that path instead. I'll work out a PR.

derrabus:
Alternative: #9533

derrabus:
Closing in favor of #9533. @beberlei if you have time, could you please have a look at the other PR?

PR #9533, opened 2022-02-21
Deprecate custom `ObjectRepository` implementations

derrabus:
Alternative approach to #9531.

Although the documentation never mentions it, it is currently possible to use an implementation of `ObjectRepository` as repository class that does not extend `EntityRepository`. A caller of `EntityManager::getRepository()` will probably not expect not getting an `EntityRepository` as return value.

This is why I'd like to deprecate that possibility. This will also allow us to safely use `EntityRepository` as a native return type in 3.0.

### DIFF
--- a/tests/Doctrine/Tests/ORM/ConfigurationTest.php
+++ b/tests/Doctrine/Tests/ORM/ConfigurationTest.php
@@ -16,6 +16,7 @@ use Doctrine\ORM\Cache\Exception\QueryCacheNotConfigured;
 use Doctrine\ORM\Cache\Exception\QueryCacheUsesNonPersistentCache;
 use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityRepository;
+use Doctrine\ORM\Exception\InvalidEntityRepository;
 use Doctrine\ORM\Exception\ORMException;
 use Doctrine\ORM\Exception\ProxyClassesAlwaysRegenerating;
 use Doctrine\ORM\Mapping as AnnotationNamespace;
@@ -25,10 +26,12 @@ use Doctrine\ORM\Mapping\PrePersist;
 use Doctrine\ORM\Mapping\QuoteStrategy;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
+use Doctrine\Persistence\ObjectRepository;
 use Doctrine\Tests\DoctrineTestCase;
 use Doctrine\Tests\Models\DDC753\DDC753CustomRepository;
 use Psr\Cache\CacheItemPoolInterface;
 use ReflectionClass;
+use stdClass;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 
 use function class_exists;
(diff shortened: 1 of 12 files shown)


tests/Doctrine/Tests/ORM/ConfigurationTest.php
```diff
@@ -388,10 +391,19 @@ public function setDefaultRepositoryClassName(): void
         self::assertSame(EntityRepository::class, $this->configuration->getDefaultRepositoryClassName());
         $this->configuration->setDefaultRepositoryClassName(DDC753CustomRepository::class);
         self::assertSame(DDC753CustomRepository::class, $this->configuration->getDefaultRepositoryClassName());
-        $this->expectException(ORMException::class);
+        $this->expectException(InvalidEntityRepository::class);
+        $this->expectExceptionMessage('Invalid repository class \'Doctrine\Tests\ORM\ConfigurationTest\'. It must be a Doctrine\ORM\EntityRepository.');
         $this->configuration->setDefaultRepositoryClassName(self::class);
     }
 
+    public function setDeprecatedDefaultRepositoryClassName(): void
```
SenseException: Where is this used?

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
```diff
@@ -655,17 +653,6 @@ public function testDefaultRepositoryClassName(): void
         self::assertEquals($this->_em->getConfiguration()->getDefaultRepositoryClassName(), EntityRepository::class);
     }
 
-    /**
-     * @group DDC-753
-     */
-    public function testSetDefaultRepositoryInvalidClassError(): void
```
SenseException, line 669: Does this test need to be removed?

tests/Doctrine/Tests/ORM/ConfigurationTest.php
derrabus: That's a typo actually. Fixed.

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
derrabus: It was redundant. `ConfigurationTest` literally tests the same thing.

SenseException (dismissed):

kbond:
I'd like to voice an objection to this. ~In my apps, I have my entity repositories implement `\Countable`. This is not possible when extending `EntityRepository` because of it's `count()` method.~ Edit: I was wrong about this - I can override as `count(array $criteria = [])`.

Still, I have a few concerns:
1. I'd like to keep the extra `EntityRepository` public methods out of my public API (ie `createQueryBuilder`).
2. I can have my custom `ObjectRepository`'s implement `ServiceEntityRepositoryInterface` in a way that keeps the constructor _clean_ for injecting additional services (using setter injection with the `Required` attribute).

Admittedly, these are both minor - my main concern was being able to implement `\Countable`.

derrabus:
Indeed, `Countable` can be implemented by repositories, so that should not hold you back.

Regarding the other concerns:

* I don't see a big problem with a repository having a `createQueryBuilder()` method. In fact, somone experienced with the ORM who joins your project might be surprised if your repositories _don't_ have that method.
* That might be a matter of taste, but imho there's nothing _clean_  about injecting a _mandatory_ dependency in a setter. Symfony's service entity repositories can alaredy have any dependencies you want.

Looking at the alternatives (#9531 mainly), this PR still seems to be the best option to me.

kbond:
For a bit of context, I've been working on a library (very wip) that provides an opinionated entity repository (iterable, countable, read-only vs writeable implementations). It's optimized for large data sets.

> I don't see a big problem with a repository having a createQueryBuilder() method.

Excluding was an attempt to lock user's out of "doing whatever they want" outside of the repo itself.

> ...there's nothing clean about injecting a mandatory dependency in a setter.

Clean was a poor word choice on my part. My library's "repository" classes are abstract and force you implement `ObjectRepository::getClassName()` and a protected `em(): EntityManagerInterface` method. How you do so is up to you but I provide an optional DX helper (that uses the setter injection) to make it easier to use with Symfony.

This change throws a bit of a wrench in my vision but isn't something I can't work around.

greg0ire (approved):

SenseException (approved):

PR #9535, opened 2022-02-21
Leverage the `Stringable` interface

derrabus:
In PHP 8, classes that implement the `__toString()` method automagically implement a new interface called `Stringable`. This means that we can rewrite `method_exists()` checks for `__toString()` to `instanceof` checks.

We could also consider announcing the `Stringable` interface explicitly on our own classes with a `__toString()` method. But that would be pure documentation because, as already mentioned, PHP will add the interface whether we announce it or not.


SenseException (approved):
Neither of `method_exists` or `instanceof` is nice, but the latter is easier on the eye.

PR #9537, opened 2022-02-22
Make error message suggestion accurate

### LINKED ISSUE #9536: Misleading error message for bi-dir O:O with no mappedBy
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.1

#### Summary
Wrong validation message is displayed when an incorrect bidirectional one-to-one mapping is set up. When the owning side is configured correctly and the target side is missing the back reference the ORM suggests adding `inverseBy` instead of `mappedBy`.

I may be wrong, but adding such `inversedBy` creates another error in the mapping, while using `mappedBy` results in mapping passing the validation.

#### Current behavior

```
Mapping
-------

 [FAIL] The entity-class App\Entity\Media mapping is invalid:
 * The field App\Entity\Media#attachment is on the owning side of a bi-directional relationship, but the specified mappedBy association on the target-entity App\Entity\Attachment# does not contain the required 'inversedBy' attribute.
```

#### How to reproduce
1. Configure two entities like so:

```php
#[ORM\Entity]
class Media
{
    //...
    #[ORM\OneToOne(inversedBy: 'media')]
    #[ORM\JoinColumn(name: 'ZTH1', referencedColumnName: 'ZPK')]
    private Attachment $attachment;
}

#[ORM\Entity]
class Attachment
{
    #[ORM\OneToOne]
    #[ORM\JoinColumn(name: 'ZMD0', referencedColumnName: 'ZPK')]
    private ?Media $media;
```

2. Run `bin/console doctrine:schema:validate`

#### Expected behavior

I believe that the message should suggest adding `mappedBy` and not `inversedBy`, as [according to the docs](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/reference/association-mapping.html#one-to-one-bidirectional), the owning side needs `inversedBy` while the target side uses `mappedBy`. In addition it seems like the target entity reference should contain a field which is missing (nothing after `#`).

With both fixed the message will look like the following:
```
 [FAIL] The entity-class App\Entity\Media mapping is invalid:
 * The field App\Entity\Media#attachment is on the owning side of a bi-directional relationship, but the specified mappedBy association on the target-entity App\Entity\Attachment#media does not contain the required 'mappedBy' attribute.
```

See PR https://github.com/doctrine/orm/pull/9537

kiler129:
Wrong validation message is displayed when an incorrect bidirectional
one-to-one mapping is set up. When the owning side is configured
correctly and the target side is missing the back reference the ORM
suggests adding inverseBy instead of mappedBy, with the field name
missing.

Fixes https://github.com/doctrine/orm/issues/9536

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -134,9 +134,9 @@ class SchemaValidator
                             'field ' . $assoc['targetEntity'] . '#' . $assoc['inversedBy'] . ' which does not exist.';
                 } elseif ($targetMetadata->associationMappings[$assoc['inversedBy']]['mappedBy'] === null) {
                     $ce[] = 'The field ' . $class->name . '#' . $fieldName . ' is on the owning side of a ' .
-                            'bi-directional relationship, but the specified mappedBy association on the target-entity ' .
-                            $assoc['targetEntity'] . '#' . $assoc['mappedBy'] . ' does not contain the required ' .
-                            "'inversedBy' attribute.";
+                            'bi-directional relationship, but the specified inversedBy association on the target-entity ' .
+                            $assoc['targetEntity'] . '#' . $assoc['inversedBy'] . ' does not contain the required ' .
+                            "'mappedBy=\"" . $fieldName . "\"' attribute.";
                 } elseif ($targetMetadata->associationMappings[$assoc['inversedBy']]['mappedBy'] !== $fieldName) {
                     $ce[] = 'The mappings ' . $class->name . '#' . $fieldName . ' and ' .
                             $assoc['targetEntity'] . '#' . $assoc['inversedBy'] . ' are ' .
--- a/tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
+++ b/tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
@@ -152,6 +152,25 @@ class SchemaValidatorTest extends OrmTestCase
     }
 
     /**
+     * @group 9536
+     */
+    public function testInvalidBiDirectionalRelationMappingMissingMappedByAttribute(): void
+    {
+        $class = $this->em->getClassMetadata(Issue9536Owner::class);
+        $ce    = $this->validator->validateClass($class);
+
+        self::assertEquals(
+            [
+                'The field Doctrine\Tests\ORM\Tools\Issue9536Owner#one is on the owning side of a bi-directional ' .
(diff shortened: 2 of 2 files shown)


lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -134,9 +134,9 @@ public function validateClass(ClassMetadataInfo $class)
                             'field ' . $assoc['targetEntity'] . '#' . $assoc['inversedBy'] . ' which does not exist.';
                 } elseif ($targetMetadata->associationMappings[$assoc['inversedBy']]['mappedBy'] === null) {
                     $ce[] = 'The field ' . $class->name . '#' . $fieldName . ' is on the owning side of a ' .
-                            'bi-directional relationship, but the specified mappedBy association on the target-entity ' .
-                            $assoc['targetEntity'] . '#' . $assoc['mappedBy'] . ' does not contain the required ' .
-                            "'inversedBy' attribute.";
+                            'bi-directional relationship, but the specified inversedBy association on the target-entity ' .
+                            $assoc['targetEntity'] . '#' . $assoc['inversedBy'] . ' does not contain the required ' .
+                            "'mappedBy' attribute.";
```
greg0ire: Another way to confirm that you are correct is to look at the block on line 114, which should be a mirror of this one.

greg0ire (commented):
Please write a unit test for this in `tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php`

lib/Doctrine/ORM/Tools/SchemaValidator.php
kiler129: That's why I was thinking. I think tripped over that message long time ago and now I did it again, but this time I decided to fix it :)

tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
```diff
@@ -435,6 +454,46 @@ class DDC3274Two
...
+    /**
+     * @var mixed
+     * @Id
+     * @Column
+     * @GeneratedValue
+     */
+    private $id;
+
+    /**
+     * @var ArrayCollection
```
greg0ire:
```suggestion
     * @var Issue9536Owner
```

kiler129: Thank you for your patience - fixed ;)

greg0ire (approved):

greg0ire:
Thanks @kiler129 !

PR #9539, opened 2022-02-22
Drop minor version number 

greg0ire:
We should make it explicit that we mean to test with whatever is the
latest 3.x

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -30,7 +30,7 @@ jobs:
           - php-version: "8.0"
             dbal-version: "2.13"
           - php-version: "8.1"
-            dbal-version: "3.2@dev"
+            dbal-version: "3@dev"
 
     steps:
       - name: "Checkout"


.github/workflows/continuous-integration.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           - php-version: "8.0"
             dbal-version: "2.13"
           - php-version: "8.1"
-            dbal-version: "3.2@dev"
+            dbal-version: "3.2"
```
derrabus: If we do this, this job would become redundant: We already do test with stable DBAL 3 on the other runs. Currently, this job pulls the 3.4.x branch.

I think we should keep a job that tests with a dev version of DBAL. The ORM is heavily coupled to DBAL, so if a commit to the DBAL dev branch breaks the ORM, I wanna know. :-)

If you want to emphasize that this job does not run with DBAL 3.2 anymore, you can bump here:

```suggestion
            dbal-version: "3.4@dev"
```

… or we could remove the minor:

```suggestion
            dbal-version: "3@dev"
```

derrabus (changes requested):

.github/workflows/continuous-integration.yml
greg0ire: Oh my god, yes :facepalm: 
Let's go with the second suggestion

derrabus (approved):

PR #9545, opened 2022-02-25
[RFC] add `HydrationFactory` as extension point for tracking hydrations

kbond:
This is a proposed extension point to help solve https://github.com/doctrine/DoctrineBundle/issues/109. I'm thinking a `HydratorInterface` that `AbstractHydrator` implements should be added but wanted to get some feedback/input before going further.

The doctrine bundle could use this to add hydration times to the timeline:
Symfony-Profiler

The doctrine bundle could add something similar to https://github.com/debesha/DoctrineProfileExtraBundle#screenshots to the profiler panel.

You can see an example of how this could be used in DoctrineBundle here: https://github.com/kbond/symfony-reproducer/tree/hydration-profiler-poc (specifically https://github.com/kbond/symfony-reproducer/commit/9a4e91b783bbc653457d6d22410068e00b679eed).

I don't think this causes any performance problems for hydration (when using the `DefaultHydratorFactory`).

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -181,8 +181,10 @@
     <PropertyTypeCoercion>
       <code><![CDATA[new $metadataFactoryClassName()]]></code>
     </PropertyTypeCoercion>
-    <RedundantCastGivenDocblockType>
-      <code><![CDATA[(string) $hydrationMode]]></code>
+  </file>
+  <file src="src/Internal/Hydration/DefaultHydratorFactory.php">
+    <RedundantCastGivenDocblockType occurrences="1">
+      <code>(string) $hydrationMode</code>
     </RedundantCastGivenDocblockType>
   </file>
   <file src="src/EntityRepository.php">
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -8,6 +8,8 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\ORM\Cache\CacheConfiguration;
 use Doctrine\ORM\Exception\InvalidEntityRepository;
 use Doctrine\ORM\Internal\Hydration\AbstractHydrator;
+use Doctrine\ORM\Internal\Hydration\DefaultHydratorFactory;
+use Doctrine\ORM\Internal\Hydration\HydratorFactory;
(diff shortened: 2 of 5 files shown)


beberlei:
I am torn, because it will encourage people to overwrite the existing hydrators, which are a bit too complex. Besides tracking the time there are not really many other use cases to use this extension point.

kbond:
Any other ideas how to achieve? I really think this is an important metric to make available.

Maybe a `HydrationLogger` - similar to `CacheLogger`? Maybe a `HydrationEvent`?

ostrolucky:
I would be fine with this. This solves one of the inconsistencies between ODM and ORM, as ODM has this option. And I'm not a fan of not having features just because you can abuse them. I think ORM learned this lesson as well, since multiple important features have been deprecated years ago, only to get comeback recently and de-deprecated.

kbond:
Rebased to 3.4 with the hope of continuing this discussion 😄 🙏🏻

ostrolucky:
This seems pretty simple and straightforward. Anything against merging it (after rebase and changing target branch ofc) @greg0ire ?

greg0ire:
I think we should take @beberlei 's concern into account. A `HydrationLogger` sounds good, although let's maybe not use a name that locks down too much what this is supposed to be used for. Maybe a name that includes "instrumentation", in reference to https://martinfowler.com/articles/domain-oriented-observability.html ?

ostrolucky:
I don't see how adding some instrumentation hydrator would work, since this hydrator needs to wrap all the other hydrators and needs to be flexible so that bundle can execute custom logic before/after hydration. It's not even clear how would user collect the hydration result if it's in some object they don't instantiate and don't normally have access to.

kbond:
> Maybe a HydrationLogger

I did not have a clear idea here and honestly, I feel like this would add overhead we don't want.

kbond:
I'd still like to see this implemented

ostrolucky:
@greg0ire why close without reply?

jdreesen:
I think it was auto-closed because the target branch was deleted.

ostrolucky:
Indeed, that's most likely. I guess we need to create a new PR then.

greg0ire:
Yes, sorry for this.

PR #9549, opened 2022-02-28
Leverage Lexer's Token type

derrabus:
Follows doctrine/lexer#64

The Lexer now exposes a Psalm type for the token array. Since that array structure is used by the parser internally, it makes sense to reuse it to make Psalm a little smarter.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Query;
 
+use Doctrine\Common\Lexer\AbstractLexer;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
@@ -86,6 +87,7 @@ use function substr;
  * An LL(*) recursive-descent parser for the context-free grammar of the Doctrine Query Language.
  * Parses a DQL query, reports any errors in it, and generates an AST.
  *
+ * @psalm-import-type Token from AbstractLexer
  * @psalm-type QueryComponent = array{
  *                 metadata?: ClassMetadata<object>,
  *                 parent?: string|null,
@@ -93,7 +95,7 @@ use function substr;
  *                 map?: string|null,
  *                 resultVariable?: AST\Node|string,
  *                 nestingLevel: int,
- *                 token: array
+ *                 token: Token,
  *             }
  */
 class Parser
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -86,14 +87,15 @@
  * An LL(*) recursive-descent parser for the context-free grammar of the Doctrine Query Language.
  * Parses a DQL query, reports any errors in it, and generates an AST.
  *
+ * @psalm-import-type Token from AbstractLexer
```
greg0ire, line 90: Should we bump `doctrine/lexer`'s version constraint?

psalm-baseline.xml
```diff
@@ -2402,11 +2402,14 @@
...
     </InvalidPropertyAssignmentValue>
-    <InvalidScalarArgument occurrences="4">
+    <InvalidScalarArgument occurrences="7">
       <code>$resultAlias</code>
       <code>$resultAlias</code>
       <code>$resultAlias</code>
       <code>$resultAlias</code>
+      <code>$this-&gt;queryComponents[$expr]['token']['value']</code>
+      <code>$this-&gt;queryComponents[$factor]['token']['value']</code>
+      <code>$this-&gt;queryComponents[$term]['token']['value']</code>
```
greg0ire, line 2412: Sweet :)

lib/Doctrine/ORM/Query/Parser.php
derrabus: Just for static analysis? Feels unnecessary.

greg0ire: Is it a big deal of bumping it? Also, isn't there a risk of breaking our users' SA pipeline? What will happen if they upgrade the ORM without upgrading `doctrine/lexer`?

greg0ire: Note that 1.2.3 supports PHP 7.1 so I wouldn't say it's a big deal.

derrabus: > Is it a big deal of bumping it?

No, we could do it. It's just that we don't really need to.

> Also, isn't there a risk of breaking our users' SA pipeline? What will happen if they upgrade the ORM without upgrading `doctrine/lexer`?

Most projects won't notice, unless they implement their own tree walkers maybe.

derrabus:
I've followed your suggestion and bumped Lexer. I guess it's better for our sanity if we keep the version range small here. After all, ORM and Lexer are very tightly coupled.

greg0ire (approved):

greg0ire:
Not sure why Codecov hangs, but let's merge.

PR #9553, opened 2022-03-01
Document tree walker class strings

derrabus:
Another backport from #9551

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -220,7 +220,7 @@ class Parser
     /**
      * The custom last tree walker, if any, that is responsible for producing the output.
      *
-     * @var class-string<TreeWalker>
+     * @var class-string<SqlWalker>|null
      */
     private $customOutputWalker;
 
@@ -245,6 +245,7 @@ class Parser
      * This tree walker will be run last over the AST, after any other walkers.
      *
      * @param string $className
+     * @psalm-param class-string<SqlWalker> $className
      *
      * @return void
      */
@@ -257,7 +258,7 @@ class Parser
      * Adds a custom tree walker for modifying the AST.
      *
      * @param string $className
-     * @psalm-param class-string $className
+     * @psalm-param class-string<TreeWalker> $className
      *
      * @return void
      */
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -172,7 +173,9 @@ class SqlWalker implements TreeWalker
     private $quoteStrategy;
 
     /**
-     * {@inheritDoc}
+     * @param Query        $query        The parsed Query.
+     * @param ParserResult $parserResult The result of the parsing process.
+     * @psalm-param array<string, QueryComponent> $queryComponents The query components (symbol table).
```
derrabus, line 178: Psalm seems to ignore `@inheritDoc` on constructors. 🤷🏻‍♂️

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -220,7 +220,7 @@ class Parser
     /**
      * The custom last tree walker, if any, that is responsible for producing the output.
      *
-     * @var class-string<TreeWalker>
+     * @var class-string<SqlWalker>|null
```
derrabus, line 223: As explained on #9551, output walkers cannot be arbitrary implementations of `TreeWalker`, we do expect an `SqlWalker` here.

lib/Doctrine/ORM/Query/SqlWalker.php
greg0ire: It is supposed to work though: https://github.com/vimeo/psalm/issues/4504#issuecomment-899725382

greg0ire: But I think you'd need to add `psalm-consistent-constructor` on the interface as well (or in fact, only there? not sure if it's necessary to have it in this class). Can you please try that?

derrabus: Doesn't really work. If I add this annotation to the interface, `@inheritdoc` on `SqlWalker`'s constructor is still ignored. If I remove the attotation from `SqlWalker`, `@inheritdoc` works again, but this also brings back the `UnsafeInstantiation` error for `new $outputWalkerClass`. 😕

greg0ire: Ok, I guess defining a constructor in an interface is quite the corner case.

greg0ire (approved):

PR #9554, opened 2022-03-01
Fixing typo

ThomasLandauer:
IMO "type" didn't make sense here.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/MappingException.php
+++ b/lib/Doctrine/ORM/Mapping/MappingException.php
@@ -991,7 +991,7 @@ class MappingException extends ORMException
             <<<'EXCEPTION'
 Context: Trying to hydrate enum property "%s::$%s"
 Problem: Case "%s" is not listed in enum "%s"
-Solution: Either add the case to the enum type or migrate the database column to use another case of the enum
+Solution: Either add the case to the enum or migrate the database column to use another case of the enum
 EXCEPTION
             ,
             $className,


derrabus:
Please have a look at the CI failures. We have tests that assert the exact wording of the exception message you've changed.

That being said, I'm not sure if this is really a typo. The enum is a PHP type and the message points out that this type declaration has to be changed.

ThomasLandauer:
"enum type" just sounds strange to me. An array is a type too, but you wouldn't say "Please add `foo` to your array type", would you?

derrabus:
I don't follow your argument. I'm not aware that PHP supports typed arrays.

ThomasLandauer:
I see what you mean. You're having the type hint in mind (right?):
```php
private FooEnum $foo;
```
Here, the enum is a type.

I'm having the enum itself in mind:
```php
enum FooEnum
{
    case Bar;
}
```
Here, the enum is something like a "list".

The message says
> Either add the case to the enum type

... which refers to the enum (=list), doesn't it? Cause that's where you need to add the case.

derrabus:
> I see what you mean. You're having the type hint in mind (right?):

No, a type.

> ```
> private FooEnum $foo;
> ```
> 
> Here, the enum is a type.

No, that's a property type declaration: `private` is the visibility modifier, `FooEnum` is the type and `$foo` is the name of the property.

> I'm having the enum itself in mind:
> 
> ```
> enum FooEnum
> {
>     case Bar;
> }
> ```
> 
> Here, the enum is something like a "list".

That is the declaration of the type `FooEnum`.

ThomasLandauer:
> That is the declaration of the type `FooEnum`

No, that's the declaration of the *enum* `FooEnum`.
An enum has characteristics of a type, but also of a class (example: `FooEnum::Bar`). So when saying "add it to the enum type", it would mean to just add it to the "type" character of the enum, but not to the "class" character - which doesn't make sense.

Other question: If you think that "enum type" is the right term in combination with "add the case", in which context would "enum" be the right term then? What would be a sentence where you would say "enum" and not "enum type?

BTW: IMO, the current wording is inconsistent, cause it just says "enum" at the end:

> Solution: Either add the case to the enum type or migrate the database column to use another case of the enum

derrabus:
> An enum has characteristics of a type, but also of a class (example: `FooEnum::Bar`).

Yes, that's because a class is a type as well.

---

I'm open to improving the exception message to make it less confusing, but simply deleting the word `type` is not correct imho.

ThomasLandauer:
10 From what you're saying I'm suspecting that you're preferring "enum type" over "enum" *everywhere*. That's why I asked for an example where you would prefer "enum".

20 Sure I could try to reword the entire message. But it would certainly still contain the term "enum".

30 goto 10

derrabus:
Let's close this PR. I don't see us getting anywhere. Anyway, thank you for your attempt at improving our exception messages.

PR #9557, opened 2022-03-02
Parser: SimpleArithmeticExpression should return ArithmeticTerm

janbarasek:
Bugfix of return type error.

The `SimpleArithmeticExpression()` method can also return an `ArithmeticTerm` type that is not defined in the annotation.

Example:

Snímek obrazovky 2022-03-02 v 14 13 59

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -823,9 +823,6 @@
       <code>joinColumnName</code>
       <code>joinColumnName</code>
     </TooManyArguments>
-    <UndefinedMethod occurrences="1">
-      <code>isReadOnly</code>
-    </UndefinedMethod>
   </file>
   <file src="lib/Doctrine/ORM/Mapping/ColumnResult.php">
     <MissingConstructor occurrences="1">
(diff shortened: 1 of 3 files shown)


derrabus:
Can you please update the baselines of Psalm and PHPStan?

janbarasek:
@derrabus I tried to fix the baseline. Please enable the tests. Thank you.

derrabus:
Both baselines are generated files. Don't edit them manually.

```sh
./vendor/bin/phpstan --generate-baseline
./vendor/bin/psalm --no-cache --set-baseline=psalm-baseline.xml
```

janbarasek:
I successfully updated the Psalm baseline and PhpStan did not need to make any changes:

Snímek obrazovky 2022-03-02 v 18 22 02

psalm-baseline.xml
```diff
@@ -823,6 +823,9 @@
       <code>joinColumnName</code>
       <code>joinColumnName</code>
     </TooManyArguments>
+    <UndefinedMethod occurrences="1">
+      <code>isReadOnly</code>
+    </UndefinedMethod>
```
derrabus: Looks like you ran Psalm with PHP < 8.1

derrabus:
Sorry for the trouble with our tooling. But fixing doc blocks in most cases means that static analysis will report ~~less~~ different errors, so we have to review and adjust the baselines.

Your fix is certainly correct. I've reviewed the return types on `Parser` and some `AST` classes recently and found a lot more inaccurate annotations. If you continue to work with the parser, please take all annotations with a grain of salt. 😕

psalm-baseline.xml
janbarasek: Just a moment, please. I need to reinstall PHP on my environment. :/

janbarasek: Updated.

derrabus (approved):

PR #9565, opened 2022-03-06
Fix type on `SqlWalker::walkPathExpression()`

derrabus:
The parameter of `SqlWalker::walkPathExpression()` is annotated as `mixed`, but passing anything but a `PathExpression` to it would result in an error because the method accesses properties on the passed value.

### DIFF
--- a/lib/Doctrine/ORM/Query/AST/JoinAssociationPathExpression.php
+++ b/lib/Doctrine/ORM/Query/AST/JoinAssociationPathExpression.php
@@ -26,12 +26,4 @@ class JoinAssociationPathExpression extends Node
         $this->identificationVariable = $identificationVariable;
         $this->associationField       = $associationField;
     }
-
-    /**
-     * {@inheritdoc}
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ORM/Query/AST/JoinAssociationPathExpression.php
```diff
@@ -26,12 +26,4 @@ public function __construct($identificationVariable, $associationField)
...
         $this->associationField       = $associationField;
     }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function dispatch($sqlWalker)
-    {
-        return $sqlWalker->walkPathExpression($this);
-    }
```
derrabus, line 36: I've removed this method because there's no way a call to this method could've returned anything. The parent class contains an implementation of that method that raises a proper exception, so it seems just right to fall back to that.

greg0ire (approved):

PR #9566, opened 2022-03-06
Fix types on `walkLiteral()` and `walkLikeExpression()`

derrabus:


lib/Doctrine/ORM/Query/AST/LikeExpression.php
```diff
@@ -12,21 +14,21 @@
...
     /** @var bool */
-    public $not;
+    public $not = false;
 
-    /** @var Node */
+    /** @var Node|string */
     public $stringExpression;
 
-    /** @var InputParameter */
+    /** @var InputParameter|FunctionNode|PathExpression|Literal */
```
derrabus, line 22: Those are the types we expect here, see:

https://github.com/doctrine/orm/blob/bc6c6c9f0c87b5b7830c16a85ce2ba09c98b8c01/lib/Doctrine/ORM/Query/SqlWalker.php#L2165-L2173

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -2156,9 +2156,12 @@ public function walkBetweenExpression($betweenExpr)
     public function walkLikeExpression($likeExpr)
     {
         $stringExpr = $likeExpr->stringExpression;
-        $leftExpr   = is_string($stringExpr) && isset($this->queryComponents[$stringExpr]['resultVariable'])
-            ? $this->walkResultVariable($stringExpr)
-            : $stringExpr->dispatch($this);
+        if (is_string($stringExpr)) {
+            assert(isset($this->queryComponents[$stringExpr]['resultVariable']));
```
derrabus: I've moved the `isset()` check from the condition into an `assert()`. In the current code, if that check would return `false`, we would call `dispatch()` on a string which would cause an error. This is why I believe that this check never fails. Alternatively, I could just remove the `isset()` check completely.

lib/Doctrine/ORM/Query/TreeWalker.php
```diff
@@ -335,7 +335,7 @@ public function walkInstanceOfExpression($instanceOfExpr);
     /**
      * Walks down a literal that represents an AST node, thereby generating the appropriate SQL.
      *
-     * @param mixed $literal
+     * @param AST\Literal $literal
```
derrabus, line 338: Passing anything else to the implementation in `SqlWalker` would cause an error because we access properties on `$literal`.

lib/Doctrine/ORM/Query/SqlWalker.php
SenseException: Does it mean we can now have an error case where `$this->queryComponents[$stringExpr]` does not exist? Will this run into an error too?

derrabus: The previous code assumed that this case would never happen. But we could be extra cautious here and throw an exception. After all, it's a public method, so validating arguments is a sane thing to do.

SenseException: I'd prefer an exception too. Having a second opinion would be advisable.

derrabus: I'm throwing an exception instead now.

SenseException (approved):
A second reviewer should be consulted. Despite that no one mentioned an error before and "smaller" BC breaks were applied in the past in 2.11, this PR can probably be merged after a second review.

greg0ire (approved):

PR #9568, opened 2022-03-07
Uncomment `wrapInTransaction()` method in `EntityManagerInterface`

pszymczyk:
Remove redundant method comment.


derrabus:
Sorry, this is a breaking change and we cannot do that in a bugfix release. The method has been added to the interface on the 3.0.x branch.

pszymczyk:
So I should not use `transactional()` cause it is deprecated and I cannot use `wrapInTransaction` (which is suggested in deprecated method comment) because it is not released, perfect 👌🏻

derrabus:
Sure you can use `wrapInTransaction()`. All of our implementations provide that method.

PR #9570, opened 2022-03-08
Creating Mapping Types overview table

ThomasLandauer:
I changed `json_array` to `json` - I guess this was a typo.

Also:
* I moved "PHP Types Mapping" below "Doctrine Mapping Types".
* "PHP Types Mapping": Only changed the last sentence.

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -48,10 +48,7 @@ mapping metadata:
 -  :doc:`Docblock Annotations <annotations-reference>`
 -  :doc:`XML <xml-mapping>`
 -  :doc:`PHP code <php-mapping>`
--  :doc:`YAML <yaml-mapping>` (deprecated and will be removed in ``doctrine/orm`` 3.0.)
-
-This manual will usually show mapping metadata via docblock annotations, though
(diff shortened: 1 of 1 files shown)


derrabus:
This mopping table already exists in the DBAL configuration. Shall we link to it rather than duplicate?

https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html

ThomasLandauer:
You mean the "Mapping Matrix" at https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#mapping-matrix ?
Is that table broken or is it rendered as source code cause it's just too wide? Anyway, right now it's unusable cause it's several pages long *and* requires horizontal scrolling *and* has footnotes. So while I usually say that a table is a good thing, I think in that case a list would be better...

derrabus:
> Is that table broken or is it rendered as source code cause it's just too wide?

IDK, I just wanted to avoid that we re-document the DBAL inside the ORM documentation.

> Anyway, right now it's unusable cause it's several pages long _and_ requires horizontal scrolling _and_ has footnotes. So while I usually say that a table is a good thing, I think in that case a list would be better...

That documentation is open source as well. If you can improve it, don't hesitate to do it. 🙂

Issue #9571, opened 2022-03-08
Memcached detection broken by Symfony/Cache 6.0.5


flack:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.1

#### Summary

With `symfony/cache` 6.0.5, calling

```php
\Doctrine\ORM\Tools\Setup::createConfiguration();
```

causes the following exception:

```
Symfony\Component\Cache\Exception\CacheException: Memcached > 3.1.5 is required. in vendor/symfony/cache/Adapter/MemcachedAdapter.php on line 59
Doctrine\ORM\Tools\Setup::createConfiguration()	.../connection.php:182
Doctrine\ORM\Tools\Setup::createCacheConfiguration()	.../Setup.php:153
Doctrine\ORM\Tools\Setup::createCacheInstance()	.../Setup.php:169
Symfony\Component\Cache\Adapter\MemcachedAdapter->__construct()
```

#### Current behavior

See exception above

#### How to reproduce

- have `memcached` extension installed
- don't have `apcu` extension installed

#### Expected behavior

This happens due to symfony/symfony#45601. I guess a workaround would be to call `MemcachedAdapter::isSupported` before instantiating

derrabus:
see symfony/symfony#45601

flack:
Are you sure about the "Can't fix" resolution? Wouldn't something like this be enough to fix it?

```diff
diff --git a/lib/Doctrine/ORM/Tools/Setup.php b/lib/Doctrine/ORM/Tools/Setup.php
index f6c09f608..274218843 100644
--- a/lib/Doctrine/ORM/Tools/Setup.php
+++ b/lib/Doctrine/ORM/Tools/Setup.php
@@ -193,11 +193,13 @@ class Setup
             throw new RuntimeException('Setup tool cannot configure caches without doctrine/cache 1.11 or symfony/cache. Please add an explicit dependency to either library.');
         }
 
+        $memcachedAvailable = extension_loaded('memcached') && (class_exists(MemcachedCache::class) || MemcachedAdapter::isSupported());
+
         if ($isDevMode === true) {
             $cache = class_exists(ArrayCache::class) ? new ArrayCache() : new ArrayAdapter();
         } elseif (extension_loaded('apcu')) {
             $cache = class_exists(ApcuCache::class) ? new ApcuCache() : new ApcuAdapter();
-        } elseif (extension_loaded('memcached')) {
[...]

PR #9574, opened 2022-03-09
Make sure MemcachedAdapter is supported before tring to use it

flack:
See discussion in #9571 

I don't think there is a way to add a unittest for this, but if anyone has an idea, just let me know!


derrabus (approved):

derrabus:
@flack The `Setup` class will be replaced by a new `ORMSetup` class in 2.12. Can you please test #9578 for me?

flack:
@derrabus I gave it a short try today, and everything looks good on my end

derrabus:
Thank you!

PR #9587, opened 2022-03-14
Implement colocated mapping driver

greg0ire:
This allows us to decouple further from `doctrine/annotations`, and to fix
some static analysis issues.

The assumption being made here is that the abstract class we are no
longer extending is not used in type declarations and `instanceof` checks.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
@@ -5,13 +5,16 @@ declare(strict_types=1);
 namespace Doctrine\ORM\Mapping\Driver;
 
 use Doctrine\Common\Annotations\AnnotationReader;
+use Doctrine\Common\Annotations\Reader;
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Events;
 use Doctrine\ORM\Mapping;
 use Doctrine\ORM\Mapping\Builder\EntityListenerBuilder;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\Persistence\Mapping\ClassMetadata;
-use Doctrine\Persistence\Mapping\Driver\AnnotationDriver as AbstractAnnotationDriver;
+use Doctrine\Persistence\Mapping\Driver\ColocatedMappingDriver;
+use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use ReflectionClass;
 use ReflectionMethod;
 use ReflectionProperty;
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -40,6 +51,20 @@ class AnnotationDriver extends AbstractAnnotationDriver
         Mapping\MappedSuperclass::class => 2,
     ];
 
+    /**
+     * Initializes a new AnnotationDriver that uses the given AnnotationReader for reading
+     * docblock annotations.
+     *
+     * @param Reader               $reader The AnnotationReader to use, duck-typed.
```
derrabus: We should move away from the `duck-typed` part, imho. It has caused us trouble already.

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -29,8 +31,17 @@
...
+class AnnotationDriver implements MappingDriver
 {
+    use ColocatedMappingDriver;
+
+    /**
+     * The annotation reader.
+     *
+     * @var Reader
+     */
+    protected $reader;
```
derrabus, line 46: Shall we flag the property as `@internal` and make it private on 3.0?

greg0ire: Shouldn't we deprecate `AnnotationDriver` and remove it entirely in 3.0?
Also, I'm wondering why `AttributeDriver` has a `getReader()` method, and is not final :thinking:

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
```diff
@@ -25,8 +26,10 @@
 
 use const PHP_VERSION_ID;
 
-class AttributeDriver extends AnnotationDriver
+class AttributeDriver implements MappingDriver
```
greg0ire, line 30: Ah, I should implement `getReader` as well, removing that method is a BC-break I guess :sweat_smile: … but since it's not breaking any tests, I think I should instantly deprecate it.

derrabus (approved):

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -40,6 +54,20 @@ class AnnotationDriver extends AbstractAnnotationDriver
         Mapping\MappedSuperclass::class => 2,
     ];
 
+    /**
+     * Initializes a new AnnotationDriver that uses the given AnnotationReader for reading
+     * docblock annotations.
+     *
+     * @param Reader               $reader The AnnotationReader to use
```
stof, line 61: why not using an actual parameter type ?

derrabus: That would be a BC break. The old constructor explicitly mentioned the possibility of duck-typed annotation readers.

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -29,8 +32,19 @@
 /**
  * The AnnotationDriver reads the mapping metadata from docblock annotations.
  */
-class AnnotationDriver extends AbstractAnnotationDriver
+class AnnotationDriver implements MappingDriver
```
Ocramius, line 35: This BC break needs to be reverted, and moved to the next major.

Ref: https://github.com/doctrine/orm/issues/9668

PR #9589, opened 2022-03-18
Fix type errors with XML export on PHP 8.1

vrestihnat:
Please see PHP interface SimpleXMLElement::addAttribute(string $name, string $value = null, string $namespace = null): void .... 
The $value must be string or null.


greg0ire:
Blocked by https://github.com/doctrine/orm/pull/9590

greg0ire (approved):

derrabus (approved):

PR #9590, opened 2022-03-18
Ignore deprecation from Persistence

greg0ire:
The deprecation is already addressed in the next minor branch.


greg0ire:
Philosophical question: where should deprecations be addressed (I'm deliberately avoiding "fixed" here)?
In #9587, I targeted 2.12.x because targeting 2.11.x would imply that we did something wrong and needed to fix it, which is not the case.

SenseException (approved):
I don't know about "when", but I guess at least with a release of a new minor version. 2.11 already got some releases, so I also would have targeted 2.12.

PR #9598, opened 2022-03-20
Use correct syntax for external links

greg0ire:
There is no leading underscore, and the trailing underscore should not
be forgotten.

Addresses https://github.com/doctrine/orm/pull/9184#issuecomment-1073303417


greg0ire:
Cc @ThomasLandauer

derrabus (approved):

PR #9602, opened 2022-03-22
provided passing test for mapping joins in native queries

grossmannmartin:
test to check whether https://github.com/doctrine/orm/issues/2633 is already fixed and can be closed


grossmannmartin:
closed and replaced with #9606

PR #9604, opened 2022-03-23
Indicate what feature is deprecated

greg0ire:
This is prompted by a question from somebody on the Symfony Slack: they could not figure out what was deprecated because they did not know they could get a stack trace by using `-vvv`


SenseException (approved):
I consider Exception classes outside of a BC promise, because they're not supposed to be used by other projects except for catching them. But I'm not sure if it's also handled that way in Doctrine projects.

beberlei (approved):

PR #9606, opened 2022-03-25
Fixed issue #2633

grossmannmartin:
Fixes issue #2633
Items in $rowData['data'] in ArrayHydrator is now sorted  by level obtained from ResultSetMapping::$parentAliasMap

ArrayHydrator now considers the set mapping to correctly provide data.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/ArrayHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ArrayHydrator.php
@@ -74,6 +74,18 @@ class ArrayHydrator extends AbstractHydrator
         $nonemptyComponents = [];
         $rowData            = $this->gatherRowData($row, $id, $nonemptyComponents);
 
+        // 1.1) Sort $rowData['data'] by level obtained from ResultSetMapping::$parentAliasMap
+        // It is necessary otherwise $this->_resultPointers can contain incorrect data (from previous row).
+        $dqlAliases = array_keys($rowData['data']);
+
+        $orderedDqlAliases = $this->getDqlAliasesOrderedByLevel($dqlAliases, $this->resultSetMapping()->parentAliasMap);
+
+        $rowDataData = $rowData['data'];
+        $rowData['data'] = [];
+        foreach ($orderedDqlAliases as $dqlAlias) {
+            $rowData['data'][$dqlAlias] = $rowDataData[$dqlAlias];
+        }
+
         // 2) Now hydrate the data found in the current row.
         foreach ($rowData['data'] as $dqlAlias => $data) {
             $index = false;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Internal/Hydration/DqlAliasTreeWalker.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Internal\Hydration;
+
+/**
+ * @internal
+ */
+class DqlAliasTreeWalker {
```
derrabus: This class is not really a tree walker. Let's merge those two methods into `ArrayHydrator`, shall we?

grossmannmartin: Thank you. you're right. I changed that :)

lib/Doctrine/ORM/Internal/Hydration/ArrayHydrator.php
```diff
@@ -277,4 +289,42 @@ private function updateResultPointer(
...
+        }
+
+        $dqlAliasesOrderedByLevel = [];
+        for ($level = 0; isset($dqlAliasesByLevel[$level]); $level++) {
+            foreach ($dqlAliasesByLevel[$level] as $dqlAliases2) {
+                $dqlAliasesOrderedByLevel[] = $dqlAliases2;
+            }
+        }
+
+        return $dqlAliasesOrderedByLevel;
```
derrabus:
```suggestion
        ksort($dqlAliasesByLevel,  SORT_NUMERIC);
        
        return array_merge(...array_values($dqlAliasesByLevel));
```

grossmannmartin: clever, thanks :)

beberlei:
All this code is executed on processing every row over and over again. This needs a performance evaluation. 

Is it possible to move some or all of this logic into the prepare state?

beberlei:
To summarize It requires two foreach, sorting and merging. Effectively adding 4O(N) to the inside of an existing loop, thats my concern

grossmannmartin:
I understand your concerns... I'm not sure about moving to the prepared state.
Maybe reorder parameters in the Native(Abstract)Query? TBH, I don't understand the internals well enough to be sure. 

It is worth mentioning that hydrating into objects works just fine, any idea how is that possible?

grossmannmartin:
Hi, 
Is anything I can do to help push this issue? 
In the current state it seems like possible performance degradation, so can you please hint me with another options to investigate to solve this issue better?

grossmannmartin:
well... it's probably not a problem anymore

PR #9607, opened 2022-03-25
PHPStan 1.5.0

derrabus:


composer.json
```diff
@@ -42,7 +42,7 @@
         "doctrine/annotations": "^1.13",
         "doctrine/coding-standard": "^9.0",
         "phpbench/phpbench": "^0.16.10 || ^1.0",
-        "phpstan/phpstan": "1.4.6",
+        "phpstan/phpstan": "~1.4.10 || 1.5.0",
```
derrabus, line 45: PHPStan 1.5 dropped support for PHP 7.1. This is not a problem for us because we run PHPStan on PHP 8.1 anyway.

By allowing 1.4.x to be installed, I make sure a `composer install` still works on PHP 7.1.

PR #9608, opened 2022-03-26
Remove Sphinx config

greg0ire:
I do not think this file is still useful, since AFAIK we are not using
Sphinx anymore. Besides, this is the only Doctrine project I could find
that still has that file. It was last updated 6 years ago.


SenseException (approved):
It probably is outdated even if we still used Sphinx. Are there any informations why we even wrote and maintain our own rst-parser?

PR #9613, opened 2022-03-28
Document that entity names have to be FQCNs now

derrabus:
Because we have removed support for short/aliased entity names, we can now boldly declare that an entity name is in fact a `class-string`.


beberlei:
Short and aliased names are two different festures, did we really remove short names as well?

derrabus:
You are right, short names are still supported. My bad. 😓

PR #9617, opened 2022-03-29
Allow multiple criterias on the same property.

### LINKED ISSUE #8702: "Too many parameters: ..." when adding multiple criteria having clauses on the same property
Hello,
First of all, thanks for all the work you contributors put in this library.
I'm trying to simplify my repository queries externalizing some logic to criteria.
Everything works pretty well and the repository is more readable but, on complex queries, where I have to add multiple criteria that are adding clauses on the same property, Doctrine is returning me an error saying that I have the wrong number of parameters defined.
Debugging the code I saw that, separate criteria (with a clause on the same property) are setting parameters with the same name and, as far as I saw, the parameter names univocity is ensured only within the same criteria.

The below code it's just an example of a possible scenario that leads to the `Too many parameters: ...` error (I know that this case can be dealt differently but I have other cases that are trickier).

```php
// Repository
function getArticles():
{
    $qb = $this->createQueryBuilder('article');
    $qb->addCriteria(self::createDateToCriteria($dateTo));
    $qb->addCriteria(self::createDateFromCriteria($dateFrom));

   return $qb->getQuery()->getResult();
}

static function createDateToCriteria($date)
{
    $criteria = Criteria::create();
    $criteria->andWhere(Criteria::expr()->lte( 'publishedAt', $date));

   return $criteria;
}

static function createDateFromCriteria($date)
{
    $criteria = Criteria::create();
    $criteria->andWhere(Criteria::expr()->gte( 'publishedAt', $date));

   return $criteria;
}
```

Is this a known issue?
Is there a solution to be able to set parameter names on criteria?
In case you are interested, I'm available to contribute to try to fix the issue.

n-e-m-a-nj-a:
When adding multiple criteria referencing the same field,
the query builder returnes an error.
The fix is passing existing parameters from previous parameters
to every new insteance of QueryExpressionVisitor.
Fixes [#8702](https://github.com/doctrine/orm/issues/8702)

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -19,17 +19,11 @@
     <InvalidNullableReturnType occurrences="1">
       <code>\Doctrine\Common\Cache\Cache</code>
     </InvalidNullableReturnType>
-    <LessSpecificReturnStatement occurrences="1">
-      <code>$queryCacheProfile-&gt;generateCacheKeys($sql, $parameters, $hints)</code>
-    </LessSpecificReturnStatement>
     <MissingClosureParamType occurrences="3">
       <code>$alias</code>
       <code>$data</code>
       <code>$data</code>
     </MissingClosureParamType>
-    <MoreSpecificReturnType occurrences="1">
-      <code>array{string, string}</code>
-    </MoreSpecificReturnType>
     <NullableReturnStatement occurrences="2">
       <code>$this-&gt;_em-&gt;getConfiguration()-&gt;getResultCacheImpl()</code>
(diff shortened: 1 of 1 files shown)


n-e-m-a-nj-a:
@derrabus I've addressed the formatting issues which caused the innitial commit to fail. 
Test are passing localy, but I still need an approval to run the workflow.

I'd really appriciate if you could take another look into this one.

Thanks in advance

derrabus:
There are commits in this PR that don't belong here.

PR #9619, opened 2022-03-31
Make Reflection available to ConvertMappingCommand

### LINKED ISSUE #9659: ConvertMappingCommand does not work from attributes to xml
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.2

#### Summary

Exporting attribute mappings to xml mappings does not work

#### Current behavior

When exporting attribute based mappings to xml mappings with the ConvertMappingCommand the following error occurs:

```
In AttributeReader.php line 27:
                                                                                                                                                                                                                                                                       
  Doctrine\ORM\Mapping\Driver\AttributeReader::getClassAnnotations(): Argument #1 ($class) must be of type ReflectionClass, null given, called in /home/bbos/projects/attributebug/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php on line 75  
                                                                                                                                                                                                                                                                       
```
The problem is that the AttributeDriver does not have a ReflectionClass available. This seems to be caused by the ConvertMappingCommand which instantiates a DisconnectedClassMetadataFactory. This metadataFactory only has a StaticReflectionService which does not create ReflectionClasses.

#### How to reproduce

https://github.com/bartholdbos/attribute_bug

run `bin/console doctrine:mapping:convert xml mappings`

#### Expected behavior

XML mappings files being generated into the given directory

bartholdbos:
Fixes #9659

This fixes an exception when exporting mapping information using the
ConvertMappingCommand from Attributes. The AttributeDriver needs a
ReflectionClass in the loadMetadataForClass method. The
ConvertMappingCommand sets up the ClassMetadataFactory as an instance of
DisconnectedClassMetadataFactory which does not return ReflectionClasses

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
@@ -108,7 +108,10 @@ class AttributeDriver extends CompatibilityAnnotationDriver
      */
     public function loadMetadataForClass($className, PersistenceClassMetadata $metadata): void
     {
-        $reflectionClass = $metadata->getReflectionClass();
+        $reflectionClass = $metadata->getReflectionClass()
+            // this happens when running annotation driver in combination with
+            // static reflection services. This is not the nicest fix
+            ?? new ReflectionClass($metadata->name);
 
         $classAttributes = $this->reader->getClassAnnotations($reflectionClass);
 
--- a/lib/Doctrine/ORM/Tools/Console/Command/ConvertMappingCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ConvertMappingCommand.php
@@ -5,9 +5,9 @@ declare(strict_types=1);
 namespace Doctrine\ORM\Tools\Console\Command;
 
 use Doctrine\DBAL\Connection;
-use Doctrine\ORM\Mapping\ClassMetadataFactory;
 use Doctrine\ORM\Mapping\Driver\DatabaseDriver;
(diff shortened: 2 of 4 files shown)


beberlei (commented):
The disconnected mf is required for xml/yml schemas. You can't just change it

bartholdbos:
Okay, could this be a fix? https://github.com/doctrine/orm/blob/692c3e1b4504f8f6881b49210a78b2a0c40beabe/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L52

derrabus:
Can you please open an issue that describes and reproduces your problem first? I'm having a hard time understanding this change.

bartholdbos:
I have opened an issue with how to reproduce the bug #9659

beberlei:
The AnnotationDriver has the same kind of code to allow mapping conversion. Its an unfortunate side effect of our abstraction being sub optimal

beberlei:
@bartholdbos yes this is the fix, can you duplicate it into AttributeDrivwr?

bartholdbos:
Yes I have duplicated the fix into AttributeDriver

bartholdbos:
I added the PHPStan warning to the baseline as was done for the AnnotationDriver

bartholdbos:
The pipeline is green, I think it's ready to be reviewed (and merged hopefully)

bartholdbos:
@beberlei Hi Benjamin, could you maybe review the fix please?

derrabus (approved):

derrabus:
Thank you @bartholdbos!

PR #9621, opened 2022-03-31
Fix types on ResultSetMapping

derrabus:
This PR prepares #9620 on the 2.12.x branch.

### DIFF
--- a/lib/Doctrine/ORM/Query/ResultSetMapping.php
+++ b/lib/Doctrine/ORM/Query/ResultSetMapping.php
@@ -73,7 +73,7 @@ class ResultSetMapping
      * Maps column names in the result set to the alias/field name to use in the mapped result.
      *
      * @ignore
-     * @psalm-var array<string, string>
+     * @psalm-var array<string, string|int>
      */
     public $scalarMappings = [];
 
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ORM/Query/ResultSetMapping.php
```diff
@@ -361,9 +364,9 @@ public function addJoinedEntityResult($class, $alias, $parentAlias, $relation)
     /**
      * Adds a scalar result mapping.
      *
-     * @param string $columnName The name of the column in the SQL result set.
-     * @param string $alias      The result alias with which the scalar result should be placed in the result structure.
-     * @param string $type       The column type
+     * @param string     $columnName The name of the column in the SQL result set.
+     * @param string|int $alias      The result alias with which the scalar result should be placed in the result structure.
```
derrabus, line 368: I'm widening the type here because we pass an integer in this line:

https://github.com/doctrine/orm/blob/18d6bc375711fd734f6d47b01079725ca7ace966/lib/Doctrine/ORM/Query/SqlWalker.php#L1637

The `$resultAlias` is definitely an int as we can see here:

https://github.com/doctrine/orm/blob/18d6bc375711fd734f6d47b01079725ca7ace966/lib/Doctrine/ORM/Query/SqlWalker.php#L1587

An alternative approach would be to cast the integer to a string.

greg0ire (approved):

PR #9624, opened 2022-04-02
Support for native ENUMs as identifiers

michnovka:
Right now using Native ENUMs works fine until one is used as an ID.

This code fixes it

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1604,6 +1604,10 @@ class UnitOfWork implements PropertyChangedListener
             return self::STATE_NEW;
         }
 
+        if ($class->containsForeignIdentifier) {
+            $id = $this->identifierFlattener->flattenIdentifier($class, $id);
+        }
+
         switch (true) {
             case $class->isIdentifierNatural():
                 // Check for a version field, if available, to avoid a db lookup.
@@ -1614,7 +1618,7 @@ class UnitOfWork implements PropertyChangedListener
                 }
 
                 // Last try before db lookup: check the identity map.
-                if ($this->tryGetById($id, $class)) {
+                if ($this->tryGetById($id, $class->rootEntityName)) {
                     return self::STATE_DETACHED;
                 }
 
@@ -1631,7 +1635,7 @@ class UnitOfWork implements PropertyChangedListener
                 // the last resort: a db lookup
 
                 // Last try before db lookup: check the identity map.
-                if ($this->tryGetById($id, $class)) {
+                if ($this->tryGetById($id, $class->rootEntityName)) {
                     return self::STATE_DETACHED;
                 }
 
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ORM/Utility/IdentifierFlattener.php
```diff
@@ -76,7 +76,11 @@ public function flattenIdentifier(ClassMetadata $class, array $id): array
 
                 $flatId[$field] = implode(' ', $associatedId);
             } else {
-                $flatId[$field] = $id[$field];
+                if (is_object($id[$field]) && $id[$field] instanceof \BackedEnum) {
```
derrabus:
```suggestion
                if ($id[$field] instanceof \BackedEnum) {
```

derrabus (changes requested):
We need a test for this change.

michnovka:
@derrabus where can I find similar test for ID flattening so that I can try to make a test for this as well? Thanks!

derrabus:
A functional test that persists an entity with an enum ID will do.

michnovka:
working on this now and want to ask whether to place it into IdentityFlattenerTest.php or into EnumTest.php

If I use IdentityFlattenerTest.php , can I put new Entity and Enum into `Doctrine\Tests\Models\Cache` ? I cannot use `Doctrine\Tests\Models\Enums` because these use PHP attribute syntax and not annotation, so it creates conflicts as the rest of IdentityFlattenerTest uses annotations.

michnovka:
@derrabus test added. However while doing that I came accross UnitOfWork::tryGetById(). Why dont we flatten ID here? It would allow to use `Repository->find(SomeEnum::XXX)` instead of `Repository->find(SomeEnum::XXX->value)`

michnovka:
@derrabus so I didnt like how I have to use `find()` with `SomeEnum::XXX->value`

I made some more changes that allow the use of `find(SomeEnum::XXX)`

Please run checks and review my code. In the test I made the `find()` already uses the new syntax

michnovka:
As a note I want to point out that the old syntax `find(SomeEnum::XXX->value)` still works, so this is NOT a breaking change, only improvement

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3114,18 +3106,21 @@ public function getSingleIdentifierValue($entity)
...
-     * @param string $rootClassName The name of the root class of the mapped entity hierarchy.
-     * @psalm-param class-string $rootClassName
+     * @param mixed                 $id        The entity identifier to look for.
+     * @param Mapping\ClassMetadata $rootClass The name of the root class of the mapped entity hierarchy.
      *
      * @return object|false Returns the entity with the specified identifier if it exists in
      *                      this UnitOfWork, FALSE otherwise.
      */
-    public function tryGetById($id, $rootClassName)
+    public function tryGetById($id, $rootClass)
```
derrabus: Changing the signature of a public method is a BC break. We cannot do that without a proper deprecation strategy.

michnovka: let me make a separate PR for that

michnovka:
@derrabus I have reverted the BC on this PR and opened a new PR here: https://github.com/doctrine/orm/pull/9628 . We shall continue discussion regarding the change of` UnitOfWork` there. Lets merge this PR as a fix to ENUMs not being able to be used as IDs at all. Thanks

michnovka:
Just for the sake of completeness, this PR was closed as it was rebased for 2.12.x branch and submitted at https://github.com/doctrine/orm/pull/9629

PR #9453, opened 2022-01-31
Support enums in findBy() calls

derrabus:
Fixes https://github.com/doctrine/orm/issues/9372#issuecomment-1025628689

This PR allows to use enums in `findBy()` calls:

```PHP
$cardRepository->findBy(['suit' => Suit::Clubs]);
```

PR #9628, opened 2022-04-04
Allow searching by PHP Enum ID

michnovka:
Follow-up on https://github.com/doctrine/orm/pull/9624 - this change makes it possible to use `find(SomeEnum::XXX)` as well as the originalway `find(SomeEnum::XXX->value)` for fields of PHP Enum type.

Moreover, all `UnitOfWork::tryGetById()` calls from `UnitOfWork` itself did ID flattening before calling `tryGetById()`, but other places (BasicEntityPersister, EntityManager) did not. That was an inconsistency which is fixed.

Given that we are changing public method signature, its a breaking change

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -459,7 +459,7 @@ use function sprintf;
 
         $unitOfWork = $this->getUnitOfWork();
 
-        $entity = $unitOfWork->tryGetById($sortedId, $class);
+        $entity = $unitOfWork->tryGetById($sortedId, $class->rootEntityName);
 
         // Check identity map first
         if ($entity !== false) {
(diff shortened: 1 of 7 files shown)


michnovka:
@derrabus we could instead of changing the method signature call `$this->em->getClassMetadata($rootClassName)` inside `tryGetById()`. I checked the EntityManager source and it seems that cache is used in `MetadataFactory` so the performance penalty would be negligible.  Thoughts?

derrabus:
As this is getting bigger, maybe we should shedule this as a feature for 2.12 instead. Note that we've also merged #9453 there which solves a similar problem.

michnovka:
@derrabus well this is obviously your call, I think that if I edit this PR to not change UoW method signature, then this is a minor fix which should be in the 2.11.x branch as its really a vital part of ENUM support which was brought in 2.11. Given this some more thought and I think that the latest commit is a good compromise. There is almost no performance penalty, no BC break and the missing functionality is achieved.

derrabus:
Still, the size of both of your PRs combined tells me that we're implementing a feature here. Also, given that we've already done more work with regards to enum support on 2.12, it should be easier to implement your changes there.

Please rebase on 2.12.x and please let's make it a single PR. Supporting enums as (part of) a PK and using an enum in a PK lookup are two aspects of the same feature, imho. Please also keep in mind, that the enum could also be part of a composite PK. We should at least add another test that makes sure that scenario works as well.

michnovka:
Just for the sake of completeness, this PR was closed as it was rebased for 2.12.x branch and submitted at https://github.com/doctrine/orm/pull/9629

PR #9629, opened 2022-04-04
Support Enum IDs and search by Enum fields

michnovka:
This is a 2.12.x branch PR from https://github.com/doctrine/orm/pull/9628 and https://github.com/doctrine/orm/pull/9624

It implements:

1. using Enum fields as ID columns
2. using both old `Repository->find(SomeEnum::XXX->value)` and `Repository->find(SomeEnum::XXX)` syntax
3. `UnitOfWork::tryGetById()` now supports Enum IDs as well as ID flattening when foreign identifiers are used (before this had to be handled by the caller, but it was on some and was not on other places)


beberlei:
This is becoming a larger rabbit hole than anticipated to support enum ids. I am unsure about proceeding with thiS as I currently have little tile to spare for review due to private life.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -639,6 +639,15 @@ class ClassMetadataInfo implements ClassMetadata
...
     public $containsForeignIdentifier = false;
 
+    /**
+     * READ-ONLY: Flag indicating whether the identifier/primary key contains at least one ENUM type.
+     *
+     * This flag is necessary because some code blocks require special treatment of this cases.
+     *
+     * @var bool
+     */
+    public $containsEnumIdentifier = false;
```
beberlei, line 649: Every new cm field needs to be handled in seriamization sleep code otherwise it gets lost in cache

michnovka:
@beberlei lol, tell me about it. its 2am here and what started as a simple fix turned out to 6 hour nightmare. I see some issues still, for some reason `ClassMetadataInfo::validateAndCompleteFieldMapping()` does not seem to be called in some instances, therefore `$containsEnumIdentifier` is not set to true. I will resume work on this tomorrow, see if I can find the cause and fix it.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1603,10 +1603,6 @@ public function getEntityState($entity, $assume = null)
             return self::STATE_NEW;
         }
 
-        if ($class->containsForeignIdentifier) {
```
beberlei, line 1606: Not sure about this, would need to review the side effects on a computer with IDE (on mobile app here)

beberlei (commented):
Note to fuether review: UoW additions potentially break second level cache

lib/Doctrine/ORM/UnitOfWork.php
michnovka: This was moved into UoW tryGetById

michnovka:
> Note to fuether review: UoW additions potentially break second level cache

Aha! Thanks, I have pushed new commit which fixes the issue thanks to your guidance

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
michnovka: Commit [98ecc21](https://github.com/doctrine/orm/pull/9629/commits/98ecc2111b00f0e5e06212d7160a5ee5563efdea)

michnovka:
@beberlei I have removed some of the more controversial changes which were causing issues with tryGetById(). That method is now unchanged. The latest commit passes all tests. There is no BC break. Please run workflows and review. Thanks!

lib/Doctrine/ORM/UnitOfWork.php
michnovka: This was indeed breaking stuff, so I reverted this change to original code

tests/Doctrine/Tests/ORM/Utility/IdentifierFlattenerEnumIdTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+        self::assertNotNull($findTypedCardEnumIdEntity, 'Search by Enum ID does not work');
+
+        $class = $this->_em->getClassMetadata(TypedCardEnumId::class);
+
+        $id = $class->getIdentifierValues($findTypedCardEnumIdEntity);
+
+        self::assertCount(1, $id, 'We should have 1 identifier');
+
+        self::assertEquals($id['suit'], Suit::Clubs->value);
```
derrabus:
```suggestion
        self::assertEquals($id['suit'], (Suit::Clubs)->value);
```

… to avoid parse errors on PHP 7.

derrabus: There are more places like this in this file.

michnovka: if I fix this we have then PHPCS error of "Useless parenthesis". I have put annotation of @requires PHP 8.1 on top of the whole class, why is the test even loaded and run by PHPUnit on older PHP versions?

michnovka: I have tested locally with PHP 7.4 and the test is indeed skipped. Do you want me to fix it with parenthesis and get that PHPCS issue instead?

derrabus: The test has to be parsed first before PHPUnit can read the annotation. That's why we need to make sure, a PHP 7.1 parser can read the file even if the test is skipped completely.

michnovka:
All tests passing now, can we have a merge please?

derrabus (approved):

derrabus:
Thank you @michnovka!

PR #9631, opened 2022-04-05
Add native types to reflection property classes

derrabus:


lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -6,44 +6,28 @@
...
 use ReflectionProperty;
-use ReturnTypeWillChange;
 use ValueError;
 
 use function array_map;
 use function get_class;
 use function is_array;
 
-class ReflectionEnumProperty extends ReflectionProperty
+final class ReflectionEnumProperty extends ReflectionProperty
```
derrabus, line 15: I've finalized the classes because they extend built-in classes. If the signatures change upstream, maintaining BC for child classes is going to be a nightmare.

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
```diff
@@ -68,7 +53,7 @@ public function getValue($object = null)
      * @param object                         $object
      * @param int|string|int[]|string[]|null $value
      */
-    public function setValue($object, $value = null): void
+    public function setValue(mixed $object, mixed $value = null): void
```
SenseException, line 56: Doesn't `mixed $value` break with the `getValue` method return type?

derrabus: `setValue()` is inherited from `ReflectionProperty` which declares `mixed` as parameter type already. I'm only allowed to make parameter types wider when overriding and it can't get wider than `mixed`. So `mixed` is actually the only possible native type here.

SenseException: Then I would expect `getValue()` to return mixed too. https://www.php.net/manual/en/reflectionproperty.getvalue.php

If any value is set that is not part of `int|string|array|null`, then we'll have a TypeError.

derrabus: > If any value is set that is not part of `int|string|array|null`, then we'll have a TypeError.

No, `$enum->value` would crash before that. Our implementation of `getValue()` can only return `int|string|array|null`, we can be sure about that.

Note, I'm declaring the parameter type as `mixed` because of a technical limitation which is PHP not supporting generics natively. The doc block is more precise in that regard.

lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
greg0ire: Should we at some point generate a list of finalized classes, then add an `@final` annotation on every such class on 2.x?

greg0ire (approved):

PR #9636, opened 2022-04-08
Use correct charset/collation from column/table defaults when creating relationships

### LINKED ISSUE #6823: Joined columns not created with matching custom schema options
I have a `many-to-one` join between two string fields with the MySQL collation set to `ASCII`. Unfortunately the join column generated by `SchemaTool` doesn't inherit this collation because the incorrect metadata key `options` is passed instead of 'customSchemaOptions`.

Changing [line 673 of SchemaTool.php](https://github.com/doctrine/doctrine2/blob/a82f6c5725cb31348eae0b4414e8992b1400ba30/lib/Doctrine/ORM/Tools/SchemaTool.php#L673) to:
```php
$columnOptions['customSchemaOptions'] = $fieldMapping['options'];
```

fixes the issue and the migration diff tool no longer tries to convert my ASCII fields to the default UTF-8 collation.

ruudk:
Fixes #6823

When you have entities and columns with different charset/collation the SchemaTool should be smart and use the correct charset/collection for the column that references the other table. 

In the tests I created the following entities:
| **User table** | Relation | _Default charset utf8mb4_ | _Default collation utf8mb4_bin_ |
|----------------|-|---------------------------|-------------------------------|
| id             | | _default_                 | _default_                     |
| group_id       | References Group table | ascii                     | ascii_general_ci              |
| status_id      | References Status table | latin1                    | latin1_bin                    |

👉 As you can see, the `group_id` column uses the charset/collation from the `group` table default.
For the `status_id`, it uses the charset/collation from the `id` column because that was set explicitly.

| **Group table** | _Default charset ascii | _Default collation ascii_general_ci |
|-----------------|------------------------|-----------------------------------|
| id              | _default_              | _default_                         |

👉 As you can see, the `id` column of `group` uses the table default.

| **Status table** | _Default charset koi8r | _Default collation koi8r_bin |
|------------------|------------------------|----------------------------|
| id               | latin1                 | latin1_bin                 |

👉 As you can see, the `id` column of `status` is different than the table default.

When I disable the fix in SchemaTool, the following errors would be produced:
```
// ALTER TABLE user ADD CONSTRAINT FK_8D93D649FE54D947 FOREIGN KEY (group_id) REFERENCES `group` (id)

An exception occurred while executing a query: SQLSTATE[HY000]: General error: 3780 Referencing column 'group_id' and referenced column 'id' in foreign key constraint 'FK_8D93D649FE54D947' are incompatible.
```

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -728,6 +728,27 @@ class SchemaTool
                     $columnOptions['precision'] = $fieldMapping['precision'];
                 }
 
+                // Only do this for DBAL v3 or higher.
+                if (! method_exists(AbstractPlatform::class, 'getGuidExpression')) {
+                    $theJoinTableCharset = $theJoinTable->hasOption('charset') ? $theJoinTable->getOption('charset') : null;
+                    if (
+                        ! isset($columnOptions['customSchemaOptions']['charset'])
+                        && isset($class->table['options']['charset'])
+                        && $theJoinTableCharset !== $class->table['options']['charset']
+                    ) {
+                        $columnOptions['customSchemaOptions']['charset'] = $class->table['options']['charset'];
+                    }
+
+                    $theJoinTableCollation = $theJoinTable->hasOption('collation') ? $theJoinTable->getOption('collation') : null;
+                    if (
+                        ! isset($columnOptions['customSchemaOptions']['collation'])
+                        && isset($class->table['options']['collation'])
+                        && $theJoinTableCollation !== $class->table['options']['collation']
+                    ) {
+                        $columnOptions['customSchemaOptions']['collation'] = $class->table['options']['collation'];
+                    }
+                }
+
                 $theJoinTable->addColumn($quotedColumnName, $fieldMapping['type'], $columnOptions);
             }
 
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php
@@ -1,0 +1,105 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\ManyToOne;
+use Doctrine\ORM\Mapping\Table;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+use function method_exists;
+
+class GH6823Test extends OrmFunctionalTestCase
+{
+    public function testCharsetCollationWhenCreatingForeignRelations(): void
+    {
+        if (! $this->_em->getConnection()->getDatabasePlatform() instanceof MySQLPlatform) {
+            self::markTestSkipped('This test is useful for all databases, but designed only for mysql.');
+        }
+
+        if (method_exists(AbstractPlatform::class, 'getGuidExpression')) {
+            self::markTestSkipped('Test valid for doctrine/dbal:3.x only.');
+        }
+
+        $this->createSchemaForModels(
+            GH6823User::class,
+            GH6823Group::class,
+            GH6823Status::class
+        );
+
+        self::assertSQLEquals('CREATE TABLE gh6823_user (id VARCHAR(255) NOT NULL, group_id VARCHAR(255) CHARACTER SET ascii DEFAULT NULL COLLATE `ascii_general_ci`, status_id VARCHAR(255) CHARACTER SET latin1 DEFAULT NULL COLLATE `latin1_bin`, INDEX idx_70dd1774fe54d947 (group_id), INDEX idx_70dd17746bf700bd (status_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_bin` ENGINE = InnoDB', $this->getLastLoggedQuery(4)['sql']);
+        self::assertSQLEquals('CREATE TABLE gh6823_group (id VARCHAR(255) NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET ascii COLLATE `ascii_general_ci` ENGINE = InnoDB', $this->getLastLoggedQuery(3)['sql']);
+        self::assertSQLEquals('CREATE TABLE gh6823_status (id VARCHAR(255) CHARACTER SET latin1 NOT NULL COLLATE `latin1_bin`, PRIMARY KEY(id)) DEFAULT CHARACTER SET koi8r COLLATE `koi8r_bin` ENGINE = InnoDB', $this->getLastLoggedQuery(2)['sql']);
+        self::assertSQLEquals('ALTER TABLE gh6823_user ADD CONSTRAINT fk_70dd1774fe54d947 FOREIGN KEY (group_id) REFERENCES gh6823_group (id)', $this->getLastLoggedQuery(1)['sql']);
+        self::assertSQLEquals('ALTER TABLE gh6823_user ADD CONSTRAINT fk_70dd17746bf700bd FOREIGN KEY (status_id) REFERENCES gh6823_status (id)', $this->getLastLoggedQuery(0)['sql']);
+    }
+}
+
+/**
+ * @Entity
+ * @Table(name="gh6823_user", options={
+ *     "charset"="utf8mb4",
+ *     "collation"="utf8mb4_bin"
+ * })
+ */
+class GH6823User
+{
+    /**
+     * @var string
+     * @Id
+     * @Column(type="string")
+     */
+    public $id;
+
+    /**
+     * @var GH6823Group
+     * @ManyToOne(targetEntity="GH6823Group")
+     */
+    public $group;
+
+    /**
+     * @var GH6823Status
+     * @ManyToOne(targetEntity="GH6823Status")
+     */
+    public $status;
+}
+
+/**
+ * @Entity
+ * @Table(name="gh6823_group", options={
+ *     "charset"="ascii",
+ *     "collation"="ascii_general_ci"
+ * })
+ */
+class GH6823Group
+{
+    /**
+     * @var string
+     * @Id
+     * @Column(type="string")
+     */
+    public $id;
+}
+
+/**
+ * @Entity
+ * @Table(name="gh6823_status", options={
+ *     "charset"="koi8r",
+ *     "collation"="koi8r_bin"
+ * })
+ */
+class GH6823Status
+{
+    /**
+     * @var string
+     * @Id
+     * @Column(type="string", options={"charset"="latin1", "collation"="latin1_bin"})
+     */
+    public $id;
+}


tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php
```diff
@@ -0,0 +1,97 @@
...
+            self::markTestSkipped('This test is useful for all databases, but designed only for mysql.');
+        }
+
+        $sql = $this->_schemaTool->getCreateSchemaSql(
+            [
+                $this->_em->getClassMetadata(GH6823User::class),
+                $this->_em->getClassMetadata(GH6823Group::class),
+                $this->_em->getClassMetadata(GH6823Status::class),
+            ]
+        );
```
greg0ire, line 34: Please use `createSchemaForModels` for this.

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -727,6 +727,24 @@ private function gatherRelationJoinColumns(
...
+                    && isset($class->table['options']['charset'])
+                    && $theJoinTableCharset !== $class->table['options']['charset']
+                ) {
+                    $columnOptions['customSchemaOptions']['charset'] = $class->table['options']['charset'];
+                }
+
+                $theJoinTableCollation = $theJoinTable->hasOption('collation') ? $theJoinTable->getOption('collation') : null;
+                if (
+                    ! isset($columnOptions['customSchemaOptions']['collation'])
+                    && isset($class->table['options']['collate'])
```
greg0ire: :thinking: why is it sometimes `collation`, sometimes `collate`? There is little to no mention of `collate` in the ORM, introducing it here seems weird…

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php
```diff
@@ -0,0 +1,97 @@
...
+        self::assertEquals('CREATE TABLE `group` (id VARCHAR(255) NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET ascii COLLATE `ascii_general_ci` ENGINE = InnoDB', $sql[1]);
+        self::assertEquals('CREATE TABLE status (id VARCHAR(255) CHARACTER SET latin1 NOT NULL COLLATE `latin1_bin`, PRIMARY KEY(id)) DEFAULT CHARACTER SET koi8r COLLATE `koi8r_bin` ENGINE = InnoDB', $sql[2]);
+        self::assertEquals('ALTER TABLE user ADD CONSTRAINT FK_8D93D649FE54D947 FOREIGN KEY (group_id) REFERENCES `group` (id)', $sql[3]);
+        self::assertEquals('ALTER TABLE user ADD CONSTRAINT FK_8D93D6496BF700BD FOREIGN KEY (status_id) REFERENCES status (id)', $sql[4]);
+    }
+}
+
+/**
+ * @Entity
+ * @Table(name="user", options={"charset"="utf8mb4", "collate"="utf8mb4_bin"})
```
greg0ire: Looks like this `options` attribute is not documented :grimacing: 
Shouldn't it be `collation` here? Where can I find a list of valid values?

lib/Doctrine/ORM/Tools/SchemaTool.php
ruudk: This is what I found about this:

For columns, it should be `collation`, see https://github.com/doctrine/dbal/blob/719663b15983278227669c8595151586a2ff3327/src/Platforms/AbstractPlatform.php#L2524-L2525 

For tables, `collation` is used, but it will also take `collate` when that one is provided instead, see https://github.com/doctrine/dbal/blob/719663b15983278227669c8595151586a2ff3327/src/Platforms/AbstractMySQLPlatform.php#L455-L464. I was using `collate` (without knowing why) in my project, so therefore I thought it needed to be called like that. I realize it's a legacy thing now.

So I updated the PR to only use `collation` everywhere. Hope it makes sense now.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php
ruudk: Done

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php
ruudk: Changed it to `collation`.

The "docs" are here: https://github.com/doctrine/dbal/blob/719663b15983278227669c8595151586a2ff3327/src/Platforms/AbstractMySQLPlatform.php#L455-L464

lib/Doctrine/ORM/Tools/SchemaTool.php
greg0ire: Ah yes, I'm the author of this last code block you pointed to: https://github.com/doctrine/dbal/commit/a18ffc1a1fde506b17c34e9523fce95f16838359

`collate` will be deprecated in the next minor, so this is the way to go.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php
```diff
@@ -0,0 +1,95 @@
...
+            self::markTestSkipped('This test is useful for all databases, but designed only for mysql.');
+        }
+
+        $this->createSchemaForModels(
+            GH6823User::class,
+            GH6823Group::class,
+            GH6823Status::class
+        );
+
+        $queries = $this->_sqlLoggerStack->queries;
```
greg0ire: This is going to conflict with 328f36846e8f194d02fb7abd3434a8de43859002 when merging up

ruudk: What do you suggest?

greg0ire: I see several solutions:
- implement something similar (not sure how feasable), might imply porting 328f368
- contribute this bugfix to 2.12.x. Since you describe this as "being smart", and not "being correct", it might make sense.
- wait for a release of 2.12.0. I don't know when it will happen, but you can take a look at the previous release dates of 2.11 and 2.10 to get an idea.

ruudk: I think in the end, this is really a bugfix. Even though I call it being smart, the produced SQL by DBAL is just invalid, as it doesn't work in MySQL (InnoDB). 

Is there an ETA for 2.12.0? If it will be released within 4 weeks I can live with that and can just target the 2.12.x branch.

greg0ire: Let me ask internally if I can release it, there are a lot of commits to release already.

ruudk: I updated the PR to target 2.12.x. Could you approve the workflow one more time so that we can see it goes green?

ruudk: Somehow the tests starts to fail now. It it possible to allow my workflows without approval? I might need to push more commits to fix this problem, that I cannot reproduce locally.

greg0ire: > Somehow the tests starts to fail now. It it possible to allow my workflows without approval? I might need to push more commits to fix this problem, that I cannot reproduce locally.

Yes, for that I would need to merge one of your PRs. Can you find something small to fix, maybe  in the docs or the readme?

ruudk: @greg0ire Done, https://github.com/doctrine/orm/pull/9639

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -343,15 +343,12 @@ abstract class OrmFunctionalTestCase extends OrmTestCase
...
     final protected function createSchemaForModels(string ...$models): void
     {
-        try {
-            $this->_schemaTool->createSchema(array_map(
-                function (string $className): ClassMetadata {
-                    return $this->_em->getClassMetadata($className);
-                },
-                $models
-            ));
-        } catch (ToolsException $e) {
```
ruudk: @greg0ire Do you know why these exceptions are caught and ignored? It seems that this discards [all kinds of errors](https://github.com/doctrine/orm/runs/5954950927?check_suite_focus=true) in the tests, which sounds terrible to be 🤯 ?

greg0ire: Yes I know, I even have written a lengthy commit message when writing that line, please check!

greg0ire: @ruudk in case you are having a hard time finding it: https://github.com/doctrine/orm/pull/9526/commits/26e85b8c8868cff5d26c5e1a420e0f4f7fde17b5

ruudk: Ok clear. I reverted my undo commit.

But I'm having a hard time trying to get it to work now.

On 2.11.x it just worked fine.

But somehow it fails on MySQL/MariaDB now. Any advice on how to debug this further?

ruudk: Oke, by doing this https://github.com/doctrine/orm/pull/9636/commits/cb6d10e99a1b734d8c631dceb0b0afe7f3f35ccb  I can see the actual problem:
```
Caused by
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'CREATE TABLE gh6823_group (id VARCHAR(255) NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET ascii COLLATE `ascii_unicode_ci` ENGINE = InnoDB':

SQLSTATE[HY000]: General error: 1273 Unknown collation: 'ascii_unicode_ci'
```

Somehow, the table is created with `ascii_unicode_ci` while I specify:
```
 * @Table(name="gh6823_group", options={"charset"="ascii", "collation"="ascii_general_ci"})
 ```

I think there is a bug somewhere in this logic: https://github.com/doctrine/dbal/blob/719663b15983278227669c8595151586a2ff3327/src/Platforms/AbstractMySQLPlatform.php#L459-L462

Will try to debug further

Comment (ruudk) [reply tests/Doctrine/Tests/OrmFunctionalTestCase.php]: Oke, so the problem is that dbal 2.13 doesn't work here... maybe I should just skip this test for dbal 2.13?

Comment (ruudk) [reply tests/Doctrine/Tests/OrmFunctionalTestCase.php]: DBAL 2.13 uses `collate` 🙈 
https://github.com/doctrine/dbal/blob/dc9b3c3c8592c935a6e590441f9abc0f9eba335b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L505-L507

Comment (greg0ire) [reply tests/Doctrine/Tests/OrmFunctionalTestCase.php]: Maybe make this fix only for DBAL 3 ? It will me one more incentive to upgrade. DBAL 2 is no longer maintained and will never switch to `collation`.

**Comment by ruudk** [comment]
@greg0ire Ok, the PR is green again 🎉

File: tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php [hunk 6]
```diff
@@ -0,0 +1,105 @@
...
+
+class GH6823Test extends OrmFunctionalTestCase
+{
+    public function testCharsetCollationWhenCreatingForeignRelations(): void
+    {
+        if (! $this->_em->getConnection()->getDatabasePlatform() instanceof MySQLPlatform) {
+            self::markTestSkipped('This test is useful for all databases, but designed only for mysql.');
+        }
+
+        if (method_exists(AbstractPlatform::class, 'getGuidExpression')) {
```
greg0ire, line 26: Shouldn't a similar condition be used in `lib/Doctrine/ORM/Tools/SchemaTool.php`?

ruudk: Done

greg0ire (approved):

derrabus (approved):

greg0ire:
Thanks @ruudk !

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php
```diff
@@ -0,0 +1,105 @@
...
+use Doctrine\ORM\Mapping\Table;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+use function method_exists;
+
+class GH6823Test extends OrmFunctionalTestCase
+{
+    public function testCharsetCollationWhenCreatingForeignRelations(): void
+    {
+        if (! $this->_em->getConnection()->getDatabasePlatform() instanceof MySQLPlatform) {
```
ruudk, line 22: @greg0ire one thing that crossed my mind, should we maybe change this test so that it runs on all platforms but only compares the sql for MySQL? That way we know it also works on Postgres

greg0ire: If you can do that it would be great, yes 👍

ruudk: I think it requires a new test because on Postgres (and other dbs) those collations don't exist.

Or shall I just set those options after reading the metadata from the entity? That way we can have multiple setups per database?

greg0ire: This sounds complex, and It's not a great idea to have complexity inside tests, so I'd say the first solution is best. I will let you judge if it is still worth it.

ruudk: I agree. Not gonna do it. I think it will work just fine.

ruudk:
@greg0ire @derrabus ⚠️ Heads up: I'm currently testing 2.12.x in my project and noticed that it produced invalid SQL. So I think we should not tag 2.12.0 yet, before we fix this.

The problem is as follows:
When the JoinColumn is an integer, it should never include `charset/collation` on the column, as integer columns don't require this.

For example, this SQL was produced:
```sql
CREATE TABLE table_name (
  // ..
  user_id INT CHARACTER SET utf8 NOT NULL COLLATE `utf8_unicode_ci`
  // ..
)
```
resulting in this error:
```
Query 1 ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CHARACTER SET utf8 NOT NULL COLLATE `utf8_unicode_ci`, type VARCHAR(255) NOT NUL' at line 1
```

ruudk:
I think the solution can be simple, never include `CHARACTER SET x` on the columns. Because this works:
```sql
CREATE TABLE table_name (
  // ..
  user_id INT NOT NULL COLLATE `utf8_unicode_ci`,
  string_user_id VARCHAR(255) NOT NULL COLLATE `utf8_unicode_ci`,
  // ..
)
```

ruudk:
This should fix the problem: https://github.com/doctrine/orm/pull/9652

PR #9638, opened 2022-04-09
Drop compatibility with DBAL 2, PHP < 7.3

greg0ire:


greg0ire:
@derrabus can you help me with the latest Psalm failures (if you're OK with this PR, obviously)? I'm not sure what to do here…

simonberger:
I am not strongly against that change, just wanting to give some insights of a single user still on Dbal 2.
I see you can throw away a bigger amount of code which is nice but is this step needed now and in a minor release? It reminds me a bit of the dbal and/or orm requirement of PHP 7.3 and step back to PHP 7.1 because of a still relevant user base that cannot receive the later versions that also add new deprecations.

Our reason to still be "stuck" on Dbal 2 is a bigger amount of `json_array` properties, a deprecation which is not that easy to fix.

derrabus:
Given that this PR dublicates a lot of work that has been done on the 3.0.x branch already, shouldn't we try to get 3.0 out soon instead?

greg0ire:
That works for me! The only downside as see is with the fact that we are probably not going to drop maintenance on 2.x right when 3.0.0 is out, but hopefully there won't be that much of it.

Issue #9433, opened 2022-01-27
Suggestion: `enumType`: Allow hydrating non-existing backed value as NULL (i.e. use `::tryFrom($value)` instead of `::try($value)`)


ThomasLandauer:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

TL;DR:
I'm suggesting an option for `enumType` so that when fetching a value from the database which is not present in the enum, the property is hydrated as `null`, instead of throwing a `ValueError`.

@derrabus: This is my answer to https://github.com/doctrine/orm/issues/9021#issuecomment-984153840, after gaining some real-world experience and some more thinking.

Real example: I'm having a `Product` entity which has an enum column `type`, holding some less-important information (just for internal use). Now I had to add a new case to `TypeEnum`. I did this only on the internal machine (where the admin backend is running), and didn't deploy it. Result: The product is not shown in production, even though - and this is the crucial part of the story - the `type` isn't even displayed there!

Lessons learnt: Whenever you need to reorganize the enum values, you need to have (1) the access rights, (2) the knowledge, and (3) the time to run the code deployment and the database migration simultaneously.

What I'd like to point out is that there are *unimportant* database columns (for statistics, internal use, whatever). So I think developers should have the choice to define enum columns as either "crucial" (i.e. exception when inconsistent) or "unimportant" (i.e. `NULL` when inconsistent).

I think it's happening at this line: https://github.com/doctrine/orm/blob/2.11.x/lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php#L71 So my suggestion would be to allow switching this to `::tryFrom()`.

derrabus:
As long as the feature is opt-in and does not generate too many edge cases, I'd be fine merging it. Do you want to provide a PR?

ThomasLandauer:
> Do you want to provide a PR?

That's the answer I was most afraid of ;-)

I will give it a try - some questions in advance:
1. Option or type? It's certainly better to add a new option to the existing `enumType`, rather than inventing a new type (e.g. `nullableEnumType`), right? So the mapping will look like this:
    ```php
    #[ORM\Column(enumType: FooEnum::class, newOption: 'someValue' )]
    ```
[...]

Issue #9622, opened 2022-04-01
Hydrating `enumType` as array gives *string* instead of Enum


janedbal:
### Bug Report

As discussed in original topic: https://github.com/doctrine/orm/discussions/9588, but I see this as a bug as it behaves differently than any custom type. So migrating from enums used as custom types to this native form is impossible. Tested version: 2.11.2

#### Summary

```php
$queryBuilder = $this->createQueryBuilder('f')
    ->select('f.id, f.suit')
    ->getQuery()
    ->getResult() // suit is string, not Suit enum
;
```

#### Current behavior

Column defined as enumType gets hydrated as string when array hydration occurs.

#### How to reproduce

Use entity as [described](https://www.doctrine-project.org/2022/01/11/orm-2.11.html) and fetch from query builder as array.

#### Expected behavior

To be hydrated as enum.

michnovka:
Ok, so given this some debugging and thought, this is a bit more difficult than it seemed. The reason why `find()`/`findAll()` methods hydrate the Enum type correctly is, that this is handled on the UoW level inside `createEntity()` through `ClassMetadata::reflFields[]->setValue()`. When using select of individual fields OR when using `getArrayResult()`, then hydration is done on individual fields and entity itself is not created.

michnovka:
@derrabus I think that the issue here is that `AbstractHydrator::hydrateAll()` gets `$resultSetMapping` with `scalarMappings` set to `string` for `suit` column. I dont think I can access the `enumType` of the column from there. What would be the proper way? Can we leverage some other "mapping", like `fieldMapping`?

michnovka:
So I see the `$resultSetMapping` is set inside `SqlWalker` here https://github.com/doctrine/orm/blob/f4d5283f705f6ac51e81c8299ace074a619b40d7/lib/Doctrine/ORM/Query/SqlWalker.php#L1335-L1352

Now the question is - do we introduce `enumMappings`? like

```php
if(!empty($fieldMapping['enumType'])) {
    $this->rsm->addEnumMapping($columnAlias, $fieldMapping['enumType']);
}
```

then when walking `$resultSetMapping->scalarMapping` check for `$resultSetMapping->enumMappings` and if present cast to enum.

Or do you have another idea? Can we leverage something already present in the `resultSetMapping`?

PR #9647, opened 2022-04-10
Add enumFallbackValue optional parameter for PHP enum types

michnovka:
As proposed in https://github.com/doctrine/orm/issues/9433 I added the option to specify default value in case DB value is not castable to native PHP Enum.

The new option is called `enumDefaultValue` and can be either null or Enum of the same type as the PHP property.

Multiple cases are covered, e.g. if DB column is nullable, but `enumDefaultValue` is set to `Suit::Spades` and not null, in such case the PHP property will still be null if DB has null value, but will be `Suit::Spades` for any invalid value.

In case DB value is NOT nullable and `enumDefaultValue` is set to `Suit::Spades`, then if DB has empty value/null/invalid value then `Suit::Spades` will be used.

I encourage you to check the test, it explains the behavior.

In case `enumDefaultValue` is not specified, then no behavior change occurs and the bundle works just like before.


michnovka:
@derrabus Hi, can you please have a look?

derrabus:
Thank you for working on this topic. I think, we should be extra careful about the wording. The new option defines a fallback value, for the case that the database contains broken values. If your database is consistent, you should not need this option. The word "default" sounds way too harmless for that.

derrabus:
What happens if I load such a broken entity from the database, change some property (no the enum) and flush it back to the database. Would the enum field be updated with the fallback value? Should it?

michnovka:
> What happens if I load such a broken entity from the database, change some property (no the enum) and flush it back to the database. Would the enum field be updated with the fallback value? Should it?

@derrabus this is a good point. So good that thinking about it, Id like to get my 4 hours of life back and remove this PR :D 

I agree with you. I think I am fixing a problem that should not be fixed. We should always throw an exception, when some unexpected broken value is loaded from DB. It should force programmers to keep their Enums in PHP code matching with DB values. But at the same time - this is EXACTLY why we need to introduce a proper ENUM support for MySQL-based databases at least. This native PHP Enum field support for a varchar, which can hold all kinds of data is wrong imo.

And further more, I dont like that we are adding more enum-specific clutter into the Column annotation/attribute.

As a side note, the issue that @ThomasLandauer has in https://github.com/doctrine/orm/issues/9433 is unfortunate, but its a problem with his configuration. Also he says that he has production vs. admin backend, but if the admin backend is touching live DB, it TOO is production. So of course there will be issues if they are not using the same synced code-base.

I think you question @derrabus illustrates this perfectly. If the field is nullable and we cast invalid values to null, then this wont be picked up as a change and we actually can not set the field to null. Same would be for any "default" or "fallback" value.

I am in favor of closing this PR and rejecting the proposed feature.

And would like to work on some actual ENUM support for at least MySQL platforms. The missing doctrine internal ENUM type (https://github.com/doctrine/dbal/blob/719663b15983278227669c8595151586a2ff3327/src/Platforms/AbstractMySQLPlatform.php#L1041 ) is causing me lot of issues in various places ( https://github.com/Elao/PhpEnums/pull/179 ) due to exception thrown at https://github.com/doctrine/dbal/blob/719663b15983278227669c8595151586a2ff3327/src/Platforms/AbstractPlatform.php#L387

I am fine with deciding that ENUM type has to be implemented separately, but the Platform class cannot act like it does not even exist. It should at least be aware of the type

beberlei:
Its always possible to map a value to just string and have a method in entity that does try to convert to enum.

michnovka:
> Its always possible to map a value to just string and have a method in entity that does try to convert to enum.

Thats a possible solution for the specific case of issue https://github.com/doctrine/orm/issues/9433 , but the idea behind making better ENUM support built-into Doctrine is that ENUMs provide more constraint. With the arrival of PHP Enums in 8.1 I think that the discussion about native Enum type in Doctrine for Platforms that support it should be restarted.

I feel like this is more for the 3.x milestone and asks for a total Type system re-design - one that can be parametrized. But the ultimate goal should be to provide ENUM() support on platforms which provide it, and fall back to the VARCHAR DB type for those that do not (the way it works now)

derrabus:
> I agree with you. I think I am fixing a problem that should not be fixed.

I didn't mean to imply that. It's just that this feature has a lot more edge cases than your tests cover. If we adress this issue in the ORM, we should make sure we get it right. And this especially applies to change tracking. Also, if we create new options for that, we need to make sure, it is super clear which effect that option has. Silently healing inconsistent data is a potential footgun.

> But at the same time - this is EXACTLY why we need to introduce a proper ENUM support for MySQL-based databases at least. This native PHP Enum field support for a varchar, which can hold all kinds of data is wrong imo.

No, MySQL's ENUMs have different drawbacks: just imaging running an `ALTER TABLE` on a 10M records table on production because your ENUM needs to have an additional case now. Yay. While I'm fine having support for MySQL's ENUM type in the ORM eventually, it's not the silver bullet you're looking for. In most cases, VARCHAR is just fine for storing PHP enumerations.

michnovka:
> No, MySQL's ENUMs have different drawbacks: just imaging running an ALTER TABLE on a 10M records table on production because your ENUM needs to have an additional case now. Yay. 

This is simply not true. Adding a new ENUM item does not require table rebuild (I think since around MySQL 5.7+, definitely not in MySQL 8 and/or MariaDB 10.4+). Ive been using ENUMs in DBs without Doctrine for ages, I have tables which have several gigabytes and quite recently I had to extend one enum column to support new case. The operation took under 1 second, which implies it was almost atomic.

But back to the original statement I made, I do stand by it. I am fixing a feature, which should NOT be fixed. We cannot populate value of PHP enum with something that is not supported. Imagine we have nullable enum field in entity (and nullable in DB) and some invalid value for it in DB is stored. When hydrating, the entity would be hydrated with NULL. Which is NOT what is in DB. Yet it is a valid DB value. So how do I set the value to be null and persist? There is no way.

If somebody wants to use PHP Enum type in entities, it comes at a price. The price should be that no unexpected values can be in DB, otherwise an Exception is thrown. Which is the current implementation and I believe it is correct.

michnovka:
To be more specific about full table copy on ENUM signature change, [some conditions have to be met](https://mariadb.com/kb/en/innodb-online-ddl-operations-with-the-inplace-alter-algorithm/#adding-a-new-enum-option):

1. It must be added to the end of the list.
2. The storage requirements must not change. 

So yes, I do agree it still does have its drawbacks, but this should not be the point of discussion about this feature. Fact is ENUMs are very commonly used data type. The decision whether to use them with their drawbacks should be left to the programmer. And I do believe that now, that PHP 8.1 has native PHP enum type it becomes more relevant to discuss the implementation of this feature officially into Doctrine.

And I would like to mention [this article](https://saveriomiroddi.github.io/Impact-of-ENUM-data-type-on-MySQL-data-storage/) in response to the ever-reappearing linking of the top 1 google search results for MySQL ENUMs [8 Reasons Why MySQL's ENUM Data Type Is Evil](http://komlenic.com/244/8-reasons-why-mysqls-enum-data-type-is-evil) from 2011, when lot of things were different in MySQL

ThomasLandauer:
Sorry for being late - my collected thoughts:

* Answer to https://github.com/doctrine/orm/issues/9433#issuecomment-1094379767

    > Making an exception for enums feels wrong

    For me, the reason is that enums enforce a much tighter binding between DB and PHP than any other type. The code and the DB need to be in sync *at every time*. How can you ever deploy the change of an enum value? Is this possible at all (without downtime)? Displaying "Vendor: unknown" for 5 minutes is still better than not displaying anything at all while the migration is running.

> I think, we should be extra careful about the wording. ... The word "default" sounds way too harmless for that.

What about "enumFallbackValue"?
"Default" indeed sounds like "you can keep it this way"; while "fallback" sounds like "something has gone wrong, and we're trying to fix it."

> What happens if I load such a broken entity from the database, change some property (no the enum) and flush it back to the database. Would the enum field be updated with the fallback value? Should it?

I would say yes. As soon as you load it, the fallback value is the value of the property. If you persist it, this will be written.
But if that's not possible (see below), then just don't change it and leave the (invalid) value as-is. I mean: When loading it the next time, it will be converted to NULL again anyway!

> When hydrating, the entity would be hydrated with NULL. Which is NOT what is in DB. Yet it is a valid DB value. So how do I set the value to be null and persist? There is no way.

This I don't understand. Why is there no way? Just set it to NULL!? (Or see below)

> If the field is nullable and we cast invalid values to null, then this wont be picked up as a change and we actually can not set the field to null.

You mean when Doctrine is computing the changeset, it doesn't notice the (silent) "change"? => Well, then don't persist the "change" - it doesn't make much of a difference (see above).

michnovka:
> You mean when Doctrine is computing the changeset, it doesn't notice the (silent) "change"? => Well, then don't persist the "change" - it doesn't make much of a difference (see above).

Yes, thats what I meant. The change wont be detected. So I am in a situation, where I need to set NULL on a field, but I cannot, because Doctrine thinks its null already. This is not ok DB abstraction imo.

I still dont understand why the solution is not to deploy the new Enum class to your production? Like i said, if you admin backend touches live server, then it is also production. So you are arguing for a fix that will enable you to have 2 different production versions, but will add logic inconsistencies (such as trying to persist the null value for an invalid one)

ThomasLandauer:
https://github.com/doctrine/orm/issues/9433 is just *one* possible problem that the tight coupling causes. Indeed, in this case, the solution is to deploy the enum class. However, running the entire deployment process whenever you just add a value to the enum, is cumbersome too!

Another problem is when you need to *change* an enum value, cause deploying the PHP class and running the migration *simultaneously* is hard (see above).

Third use-case: When `Maintainer::Alice` leaves the company, the value `NULL` (i.e. no maintainer) is the (domain-wise) correct value for all "her" articles. Here this feature could be used what I'd call a **lazy migration**: Leave "Alice" in the DB for now, and only migrate it to NULL when some other field is changed.

michnovka:
> However, running the entire deployment process whenever you just add a value to the enum, is cumbersome too!

I think that for your case you should store the field as type string and cast to/from enum class in your getters/setters. That way you can handle your exotic set-up.

Believe, I spent lot of time on this feature, and it is ready, so I am not arguing because its hard to make. I am arguing against because I realized that the issues with this outweigh the benefit of your specific use-case scenario.

As for your 3rd use-case: again, you are trying to bring user-land functionality to the bundle which should achieve DB <-> PHP mapping. Invalid enum values should not be ignored, they should be reported. So here as well as before, if you expect that invalid values can  be present in DB, then do use custom getters/setters.

This is why I believe we should work on actual ENUM DB type, because the whole point of enum is to have ONLY some CONSTRAINED values. This "fallback" solution feels wrong and imo should be left for users to implement in getter/setters, not provided as the default option. Primary goal of this package should be DB model consistency with PHP, and this PR breaks it.

Give me another example of Doctrine use, where invalid values from DB are silently fallen-back to some defaults/null. I dont think there is any.

derrabus:
> So yes, I do agree it still does have its drawbacks, but this should not be the point of discussion about this feature. Fact is ENUMs are very commonly used data type. The decision whether to use them with their drawbacks should be left to the programmer.

No objection here. All I'm saying is that MySQL's `ENUM` should not be our go-to DB type for PHP enums.

derrabus:
> > I think, we should be extra careful about the wording. ... The word "default" sounds way too harmless for that.
> 
> What about "enumFallbackValue"? "Default" indeed sounds like "you can keep it this way"; while "fallback" sounds like "something has gone wrong, and we're trying to fix it."

Yes, that's better.

> > What happens if I load such a broken entity from the database, change some property (no the enum) and flush it back to the database. Would the enum field be updated with the fallback value? Should it?
> 
> I would say yes. As soon as you load it, the fallback value is the value of the property. If you persist it, this will be written.

Okay, let's explore that path: Does that mean that the entity is flagged as changed immediately after loading, even if we perform only read-only operations? Or does it mean the enum property is regarded as unchanged until some other property on that same entity is changed? Either way, we would create interesting precedences.

> But if that's not possible (see below), then just don't change it and leave the (invalid) value as-is. I mean: When loading it the next time, it will be converted to NULL again anyway!

Okay, so the property is not regarded as changed. That again means that we don't have any way to deliberately write the fallback value (be it `null` or one of the enum cases) to the database through the ORM. This is probably what @michnovka meant here:

> > When hydrating, the entity would be hydrated with NULL. Which is NOT what is in DB. Yet it is a valid DB value. So how do I set the value to be null and persist? There is no way.
> 
> This I don't understand. Why is there no way? Just set it to NULL!?

If the fallback value is `null` and you set the property (which has been hydrated as `null` already) to `null`, Doctrine won't detect any change on it.

Another idea could be to use a value that is outside of the (nullable) enum range as a fallback. This could be `false` or a one-case enum like `ORMEnum::invalidDatabaseValue`. That would keep change tracking intact and moves the reposibility of "healing" the inconsistency to the userland. But it would of course create the precedence of a hydrated value that cannot be written back to the database, e.g. you cannot clone such an entity unless you deal with the broken enum.

> Third use-case: When `Maintainer::Alice` leaves the company

Sorry, but don't encode your employees into your codebase. That's no domain logic, that's data. Write it to your database, create a relation. I'm happy to discuss this feature and to find a good implementation, but don't make up rediculous use-cases.

michnovka:
> Another idea could be to use a value that is outside of the (nullable) enum range as a fallback. This could be `false` or a one-case enum like `ORMEnum::invalidDatabaseValue`

This is not really an option as strict typed Entity properties would be broken (in such case Id call it enumStrictCast default true)

```php
#[ORM\Column(enumType: FooEnum::class, enumStrictCast: false )]
private FooEnum $field;
```

Would throw an Exception when `ORMEnum::invalidDatabaseValue`/false is assigned to it.

derrabus:
> This is not really an option as strict typed Entity properties would be broken

No, you would need to include that value into the type declaration of course.

```php
private FooEnum|false $field;
private FooEnum|ORMEnum $field;
```

michnovka:
Ok, I am for either of the approaches:

1. use `enumStrictCast` boolean param with default true (current behavior) and in case its an invalid value, set `false` (false has sense as backed enums likely wont be `boolean`, but they might be nullable so `null` is wrong)
2. use `enumFallbackValue` which can be anything, BUT at the same time mark the entity field as modified. So when it is persisted, it should be pushed.

I do not know how (where) to mark the entity field as modified at the same time when hydration occurs. Need to check code. @derrabus if you know from the top of your head, could you point me in the right direction?

I would also like to remind this issue: https://github.com/doctrine/orm/issues/9622 . When this is addressed, it should follow the same logic when it comes to hydration imo.

michnovka:
Well this looks very promising. Seems our concern that 

> Okay, so the property is not regarded as changed. That again means that we don't have any way to deliberately write the fallback value (be it `null` or one of the enum cases) to the database through the ORM. This is probably what @michnovka meant

is incorrect. UoW keeps track of the actual DB values. So whataver we set, it always regarded as changed when computing changesets :)

I have confirmed this with a test.

Will push soon new version for review

derrabus:
> UoW keeps track of the actual DB values. So whataver we set, it always regarded as changed when computing changesets :)

So, this piece if code…

```php
$em->find(MyEntity::class, $someId);
$em->flush();
```

… would actually write something to the database sometimes? Is that desirable?

michnovka:
> … would actually write something to the database sometimes? Is that desirable?

Yes, I think it is. Considering that the `enumFallbackValue` is a shorthand to fetching a value and changing it to something else.

As I pointed out way above in discussion - I think its stupid to have separate values in ORM model and DB. So if you want to change value in your ORM, it SHOULD be considered a change and when persisted it SHOULD be pushed.

If you do not want this functionality, do not use this feature and implement custom getters/setters with native string type in the background

ThomasLandauer:
I think the `enumFallbackValue` way is better. To me, NULL is semantically correct in this case, having the meaning "no value". That's why I originally suggested to call the option `enumMethod` (with self-explanatory values `try` and `tryFrom`). However, if you allow other fallback-values as well, that's certainly fine too.

One drawback about something like `enumStrictCast` that hasn't been mentioned yet is that you'd have to constantly check if the property is `FooEnum` or the other (fallback) type, throughout the entire codebase.

Stupid question: Isn't a `$em->persist(...);` missing here?:
```php
$em->find(MyEntity::class, $someId);
$em->flush();
```

Anyway, I'm fine with writing the "fallbacked" value back to the DB. Pragmatic argument: After removing `Suit::Spades` from the enum, the database value "Spades" will (in the future) *always* get converted to the fallback value - so why not write this value right away?

michnovka:
> I think the `enumFallbackValue` way is better. To me, NULL is semantically correct in this case, having the meaning "no value". That's why I originally suggested to call the option `enumMethod` (with self-explanatory values `try` and `tryFrom`). However, if you allow other fallback-values as well, that's certainly fine too.

The way it is currently implemented is that `enumFallbackValue` can be either `null` or of `enumType` type. Any other value will throw Exception when attempting to fall back.

> Stupid question: Isn't a `$em->persist(...);` missing here?:
> 
> ```
> $em->find(MyEntity::class, $someId);
> $em->flush();
> ```

Thats the point, its not missing. At the time of hydration the entity is already different from the original data. So when flush() its called, the changeset is calculated and an `UPDATE` triggered

I have realized my `testFallbackEnumValueIsPersisted` was actually not testing things properly, hence the last commit which shows this behavior.

beberlei:
> but the idea behind making better ENUM support built-into Doctrine is that ENUMs provide more constraint

That argument makes no sense to me in combination with the fact that you want to fix a database value that is not defined on an enum entity. The enum object is a constraint, either the database fits it or the type should not be mapped to enum.

This has been true for all other types before, i dont see why we should introduce a fragile solution for enums now.

I have not seen convincing argument why we need this yet and am against merging this. 

Afaik We discussed using try in the original enum PR from what I remember and deliveraty decided against.

michnovka:
> That argument makes no sense to me in combination with the fact that you want to fix a database value that is not defined on an enum entity. The enum object is a constraint, either the database fits it or the type should not be mapped to enum.

Yes, I agree with you. This was my point too, after @derrabus commented 

> What happens if I load such a broken entity from the database, change some property (no the enum) and flush it back to the database. Would the enum field be updated with the fallback value? Should it?

But by then I have already had this implemented, so after seeing will from @derrabus and @ThomasLandauer to have this merged, I made the small changes of renaming `enumDefaultValue` to `enumFallbackValue`

I too am against merging this.

beberlei:
Sorry, i see that you have worked into this quite a bit and invested a lot of time. Thank you for doing this.

derrabus:
Thank you for brainstorming over this topic with me. We have explored multiple possible implementations and as far as I am concerned, each of them would produce additional complexity inside our already complex UoW. Moreover, each implementation produced unexpected behavior in specific situations.

I think it's the best solution if we kept things as they are right now.

drjele:
There are use cases (i currently am in this situation), where you are migrating an old app, and have many db instances deployed, that might not all be in perfect sync, and you would prefer \Doctrine\ORM\Internal\Hydration\AbstractHydrator::buildEnum to use tryFrom and have a null, for an unmapped value rather than an exception thrown.

Whould it complcate the sistem that mutch to have a flag on the column to enable/disable this behaviour?

PR #9652, opened 2022-04-11
Only put `collation` on join columns, never `charset`

ruudk:
The fix introduced in #9636 now breaks when working with integer relations.

Putting `charset` and `collation` on a text join column works fine, but when the
column is an integer, MySQL won't accept it:

```sql
CREATE TABLE table_name (
  // ..
  user_id INT CHARACTER SET utf8 NOT NULL COLLATE `utf8_unicode_ci`
  // ..
)
```

produces this error:
```
Query 1 ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CHARACTER SET utf8 NOT NULL COLLATE `utf8_unicode_ci`, type VARCHAR(255) NOT NUL' at line 1
```

Luckily the solution is simple. The charset is not needed as MySQL automatically derives that from
the `collation` when specified. It's also clever enough to ignore `collation` when the type is an integer.

Therefore the following works fine:
```sql
CREATE TABLE table_name (
  // ..
  user_id INT NOT NULL COLLATE `utf8_unicode_ci`,
  string_user_id VARCHAR(255) NOT NULL COLLATE `utf8_unicode_ci`,
  // ..
)
```

/cc @greg0ire

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -730,15 +730,6 @@ class SchemaTool
 
                 // Only do this for DBAL v3 or higher.
                 if (! method_exists(AbstractPlatform::class, 'getGuidExpression')) {
-                    $theJoinTableCharset = $theJoinTable->hasOption('charset') ? $theJoinTable->getOption('charset') : null;
-                    if (
-                        ! isset($columnOptions['customSchemaOptions']['charset'])
-                        && isset($class->table['options']['charset'])
-                        && $theJoinTableCharset !== $class->table['options']['charset']
-                    ) {
-                        $columnOptions['customSchemaOptions']['charset'] = $class->table['options']['charset'];
-                    }
-
                     $theJoinTableCollation = $theJoinTable->hasOption('collation') ? $theJoinTable->getOption('collation') : null;
                     if (
                         ! isset($columnOptions['customSchemaOptions']['collation'])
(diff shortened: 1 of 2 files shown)


ruudk:
Unfortunately, this still breaks when working with `binary` types 🤯 
```sql
contract_id BINARY(16) NOT NULL COLLATE `utf8_unicode_ci`
# Query 1 ERROR: COLLATION 'utf8_unicode_ci' is not valid for CHARACTER SET 'binary'
```

ruudk:
Is there a way to know which column types do support collation?

Maybe we can only set the collation when the type is `char`, `varchar` or `text`? See documentation: https://dev.mysql.com/doc/refman/8.0/en/charset-column.html

greg0ire:
@ruudk maybe we should revert this instead? It's already bad that the ORM has to know about charsets and collations, IMO it would be worse if it has to know about MySQL types. Maybe this should be somehow fixed in DBAL instead?

ruudk:
We can do that.

Maybe I can create a PR that adds support for `charset` and `collation` on `JoinColumn`? That would allow me to set up the correct collations for the joins without having to complicate the rest. I don't mind extra explicit configuration.

greg0ire:
PR for revert: https://github.com/doctrine/orm/pull/9654

greg0ire:
> Maybe I can create a PR that adds support for charset and collation on JoinColumn? 

That might make sense since it already exists on `Column`.

greg0ire:
Your future contribution might also fix https://github.com/doctrine/orm/issues/4286

ruudk:
Closing in favour of https://github.com/doctrine/orm/pull/9655

PR #9655, opened 2022-04-12
Allow setting column options like `charset` and `collation` on `JoinTable`, `JoinColumn` and `InverseJoinColumn`

ruudk:
Alternative approach for #9636 

This makes it possible to set custom options on the following:
* `JoinTable`
* `JoinColumn`
* `InverseJoinColumn`

/cc @greg0ire

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
@@ -407,6 +407,10 @@ class AttributeDriver implements MappingDriver
                         'name' => $joinTableAttribute->name,
                         'schema' => $joinTableAttribute->schema,
                     ];
+
+                    if ($joinTableAttribute->options) {
+                        $joinTable['options'] = $joinTableAttribute->options;
+                    }
                 }
 
                 foreach ($this->reader->getPropertyAnnotationCollection($property, Mapping\JoinColumn::class) as $joinColumn) {
@@ -645,12 +649,13 @@ class AttributeDriver implements MappingDriver
      *                   nullable: bool,
      *                   onDelete: mixed,
      *                   columnDefinition: string|null,
-     *                   referencedColumnName: string
+     *                   referencedColumnName: string,
+     *                   options?: array<string, mixed>
      *               }
      */
     private function joinColumnToArray($joinColumn): array
     {
-        return [
+        $mapping = [
             'name' => $joinColumn->name,
             'unique' => $joinColumn->unique,
             'nullable' => $joinColumn->nullable,
@@ -658,6 +663,12 @@ class AttributeDriver implements MappingDriver
             'columnDefinition' => $joinColumn->columnDefinition,
             'referencedColumnName' => $joinColumn->referencedColumnName,
         ];
+
+        if ($joinColumn->options) {
+            $mapping['options'] = $joinColumn->options;
+        }
+
+        return $mapping;
     }
 
     /**
--- a/lib/Doctrine/ORM/Mapping/InverseJoinColumn.php
+++ b/lib/Doctrine/ORM/Mapping/InverseJoinColumn.php
@@ -35,6 +35,12 @@ final class InverseJoinColumn implements Annotation
      */
     public $fieldName;
 
+    /** @var array<string, mixed> */
+    public $options = [];
+
+    /**
+     * @param array<string, mixed> $options
+     */
     public function __construct(
         ?string $name = null,
         string $referencedColumnName = 'id',
@@ -42,7 +48,8 @@ final class InverseJoinColumn implements Annotation
         bool $nullable = true,
         $onDelete = null,
         ?string $columnDefinition = null,
-        ?string $fieldName = null
+        ?string $fieldName = null,
+        array $options = []
     ) {
         $this->name                 = $name;
         $this->referencedColumnName = $referencedColumnName;
@@ -51,5 +58,6 @@ final class InverseJoinColumn implements Annotation
         $this->onDelete             = $onDelete;
         $this->columnDefinition     = $columnDefinition;
         $this->fieldName            = $fieldName;
+        $this->options              = $options;
     }
 }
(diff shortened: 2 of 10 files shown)


lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -728,6 +728,8 @@ private function gatherRelationJoinColumns(
                     $columnOptions['precision'] = $fieldMapping['precision'];
                 }
 
+                $columnOptions = $this->gatherColumnOptions($joinColumn) + $columnOptions;
```
ruudk, line 737: Should this be defined higher ?

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
```diff
@@ -645,19 +645,26 @@ private function getMethodCallbacks(ReflectionMethod $method): array
      *                   nullable: bool,
      *                   onDelete: mixed,
      *                   columnDefinition: string|null,
-     *                   referencedColumnName: string
+     *                   referencedColumnName: string,
+     *                   options?: mixed[]
```
greg0ire: Let's use something more precise, if we have an exhaustive list? `array{default: mixed, unsigned: bool…}`

lib/Doctrine/ORM/Mapping/InverseJoinColumn.php
```diff
@@ -35,14 +35,21 @@ final class InverseJoinColumn implements Annotation
      */
     public $fieldName;
 
+    /** @var array<string,mixed> */
```
greg0ire: If the list is the same as above, we could create a `psalm-type` (we have some in the codebase already), and reuse it where it makes sense to do so.

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
ruudk: This is the same as https://github.com/doctrine/orm/blob/7e49c703202fff2e3a9c753a496a2ac5bc342931/lib/Doctrine/ORM/Mapping/Column.php#L56-L57. I think it would be weird to start typing it here, and not do that for Column.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH6823Test.php
```diff
@@ -0,0 +1,140 @@
...
+    public $status;
+
+    /**
+     * @var Collection<int, GH6823Tag>
+     * @ManyToMany(targetEntity="GH6823Tag")
+     * @JoinTable(name="gh6823_user_tags", joinColumns={
+     *   @JoinColumn(name="user_id", referencedColumnName="id", options={"charset"="utf8mb4", "collation"="utf8mb4_bin"})
+     * }, inverseJoinColumns={
+     *   @JoinColumn(name="tag_id", referencedColumnName="id", options={"charset"="latin1", "collation"="latin1_bin"})
+     * })
```
ruudk: I want this to be tested for both Annotations and PHP 8 Attributes. 

I was thinking of the following:
* Add both Annotations and Attributes
* Create a `@dataProvider` that returns `annotations` and `attributes`
* When `annotations` configure the metadata factory to only look at annotations
* When `attributes` configure the metadata factory to only look at attributes

For both configurations, the output SQL should be the same.

Why do I want this? Because there are subtle differences with the example above. With PHP Attributes, you need to use `#[InverseJoinColumn]` which is an Attribute only. 

@greg0ire Do you have an idea on how to do this? Or is this already happening somewhere in the tests?

ruudk: Decided to just test this here:
tests/Doctrine/Tests/ORM/Mapping/AttributeReaderTest.php

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
greg0ire: I agree, that can be dealt with separately

beberlei:
This is the way, consistent with the rest. Thanks for the continued work on thi

ruudk:
Thanks!

I'm currently testing this thoroughly in our large monolith with 250+ entities and various collations. Will report back when I know 100% everything is working.

ruudk:
I finished my testing. I was finally able to escape the diff hell and now get:
Screenshot 2022-04-15 at 09 16 03@2x
🎉 

I tested the following: 
* import production table structure with mixed charset/collations and run `bin/console doctrine:schema:update --dump-sql` and assert no differences
* drop local database, create database, then run `bin/console doctrine:schema:update --force` to apply current metadata to database, then run `bin/console doctrine:schema:update --dump-sql` and assert no differences

@beberlei @greg0ire The PR is now ready for review. Please let me know if I need to add extra tests. Thanks!

ruudk:
@derrabus @beberlei @greg0ire could you have a look at this please? Too bad this missed the 2.12 release 😢

greg0ire:
Too bad indeed :disappointed:

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
derrabus: Okay, but it should be…

```suggestion
     *                   options?: array<string, mixed>
```

… then.

ruudk: @derrabus I changed it everywhere to your suggestion.

lib/Doctrine/ORM/Mapping/InverseJoinColumn.php
ruudk: Not possible, see above.

ruudk:
It's ready for review again.

greg0ire:
@ruudk the build is failing :see_no_evil:

ruudk:
@greg0ire Fixed now.

greg0ire (approved):

derrabus (approved):

ruudk:
Thanks

Should this be added to the https://github.com/doctrine/orm/milestone/110 milestone ?

derrabus:
Yes!

torbentschechne:
Hey there,
would be nice if this change would have been reflected on the documentation page:

https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/annotations-reference.html#annref_joincolumn

There is no information, that the `options` parameter can be used on the `@JoinColumn`. I had a lot of trouble because of this in the past days...

greg0ire:
Same goes for the XML driver I suppose… would you like to send a PR?

PR #9656, opened 2022-04-12
Indicate support for `doctrine/persistence` 3

derrabus:


composer.json
```diff
@@ -25,14 +25,14 @@
...
-        "doctrine/common": "^3.0.3",
+        "doctrine/common": "^3.0.3@dev",
         "doctrine/dbal": "^2.13.1 || ^3.2",
         "doctrine/deprecations": "^0.5.3",
         "doctrine/event-manager": "^1.1",
         "doctrine/inflector": "^1.4 || ^2.0",
         "doctrine/instantiator": "^1.3",
         "doctrine/lexer": "^1.2.3",
-        "doctrine/persistence": "^2.4",
+        "doctrine/persistence": "^2.4 || ^3@dev",
```
derrabus: If you are fine with this PR, I'd suggest that we release Common 3.3 and Persistence 3.0 before merging it.

lib/Doctrine/ORM/Exception/NotSupported.php
```diff
@@ -15,25 +15,23 @@ public static function create(): self
...
+        $message = <<<'EXCEPTION'
 Context: %s
 Problem: Feature was deprecated in doctrine/dbal 2.x and is not supported by installed doctrine/dbal:3.x
 Solution: See the doctrine/deprecations logs for new alternative approaches.
-EXCEPTION,
-            $context
-        ));
+EXCEPTION;
+
+        return new self(sprintf($message, $context));
```
derrabus: Apparently, the way we used HEREDOC syntax in this file was incompatible with PHP 7.2.

greg0ire: Yes, but you don't need to create a variable for this, having the comma on its own line should be enough, I'm going to amend this PR to fix that

greg0ire (dismissed):

composer.json
derrabus: Done.

PR #9657, opened 2022-04-12
Fix enum hydration when fetching partial results

michnovka:
This is the first attempt to fix the issue https://github.com/doctrine/orm/issues/9622

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/EnumTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/EnumTest.php
@@ -10,6 +10,7 @@ use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\ORM\Tools\SchemaTool;
 use Doctrine\Tests\Models\Enums\Card;
 use Doctrine\Tests\Models\Enums\CardWithDefault;
+use Doctrine\Tests\Models\Enums\CardWithNullable;
 use Doctrine\Tests\Models\Enums\Product;
 use Doctrine\Tests\Models\Enums\Quantity;
 use Doctrine\Tests\Models\Enums\Scale;
@@ -63,12 +64,16 @@ class EnumTest extends OrmFunctionalTestCase
 
     public function testEnumHydration(): void
     {
-        $this->setUpEntitySchema([Card::class]);
+        $this->setUpEntitySchema([Card::class, CardWithNullable::class]);
 
         $card       = new Card();
         $card->suit = Suit::Clubs;
 
+        $cardWithNullable       = new CardWithNullable();
+        $cardWithNullable->suit = null;
+
         $this->_em->persist($card);
+        $this->_em->persist($cardWithNullable);
         $this->_em->flush();
         $this->_em->clear();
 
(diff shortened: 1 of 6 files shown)


michnovka:
I suggest the following:

in AbstractHydrator use:

```php
case isset($cacheKeyInfo['isScalar']):
    $type = $cacheKeyInfo['type'];

    if (isset($cacheKeyInfo['enumType'])) {
        $value = $type->convertToPHPEnumValue($value, $this->_platform, $cacheKeyInfo['enumType']);
    } else {
        $value = $type->convertToPHPValue($value, $this->_platform);
    }

    $rowData['scalars'][$fieldName] = $value;

    break;
```

And in DBAL extend the `Type` class:

```php
    /**
     * Converts a value from its database representation to its PHP enum representation
     * of its type
     *
     * @param mixed $value The value to convert
     * @param AbstractPlatform $platform
     * @param string $enumType
     * @return BackedEnum
     * @throws ConversionException
     */
    public function convertToPHPEnumValue($value, AbstractPlatform $platform, $enumType)
    {
        try {
            return $enumType::from($value);
        }catch(ValueError $error){
            throw new ConversionException("Cannot convert value to PHP enum", 0, $error);
        }
    }
```
This will work well for most types. I will overload the method in those, where it would be problematic, such as `SimpleArrayType`.

If somebody wants to have PHP enum support in their type, they can overload this method too.

There is no BC break.

beberlei:
This is a good fix given our current data structures. 

It shows though that our reflection and type apis are not the best abstraction anymore.

 I was thinking about something new for ORM 3 but had no time sofar. For 2.x we can use this approach

michnovka:
@beberlei So can I make the DBAL PR? I need it merged there before I can finish this in ORM, otherwise tests will fail.

beberlei:
I am reviewing on mobile phone, the current patch looks like it can work without dbal changes? The fix must be orm only

michnovka:
The current fix has the limitations which I described, it will ONLY work with string type. If we use `SimpleArrayType` or any custom type with `enumType` (which is supported and even present in our tests) it will throw. I have described above a solution which requires `Type` change but allows fix for ALL types, including giving ability to custom types to make a fix the way they want.

beberlei:
Couldnt you switch over the scalar type inside the enumMappibngs case and reimplement the simple array case? We are already in workaround mode here so local patching is fine. Add a comment that it reimplements the ReflectionEnumPriperty code there

michnovka:
I can and it was my first approach, but we will fix only these 2 cases. I realized `enumType` can be used on any type, so that solution was allowing that. So workaround it is?

michnovka:
This is ready for merge. @beberlei please review

tests/Doctrine/Tests/ORM/Functional/EnumTest.php
```diff
@@ -61,6 +62,61 @@ public function testEnumMapping(string $cardClass): void
...
+
+        $result = $this->_em->createQueryBuilder()
+            ->from(Scale::class, 's')
+            ->select('s.id, s.supportedUnits')
+            ->getQuery()
+            ->getResult();
+
+        $this->assertIsArray($result[0]['supportedUnits']);
+        $this->assertContains(Unit::Gram, $result[0]['supportedUnits']);
+        $this->assertContains(Unit::Meter, $result[0]['supportedUnits']);
```
derrabus:
```suggestion
        self::assertEqualsCanonicalizing([Unit::Gram, Unit::Meter], $result[0]['supportedUnits']);
```

beberlei (dismissed):
Go from overall implementation pov, havent reviewed style and other non functional requirements.

derrabus (approved):

derrabus:
Thank you @michnovka!

PR #9661, opened 2022-04-15
Deprecate the doctrine binary

derrabus:
Follows doctrine/dbal#5084
Part of #7498

This PR deprecates the `vendor/bin/doctrine` binary.

Today, the ORM is usually configured by some kind of framework integration, like DoctrineBundle. In those cases, the existence of a `vendor/bin/doctrine` binary is highly confusing. When invoking it, the binary will simply not work and tell the user to create some configuration file. This is misleading because the integration package will most likely take care of boostrapping the entity manager and sets up the commands properly. In the context of a full-stack Symfony application, DoctrineBundle would merge the ORM's commands into the applications `bin/console` script.

Of course, there's still the possibility of using the ORM standalone. For this case, I have documented how to set up the a `bin/doctrine` binary. Setting up that binary is not really more complicated than setting up the `cli-config.php` script that would be required to make the `vendor/bin/doctrine` script work.

Moreover, it puts the developer in control over how the console is bootstrapped for their application. They can find out more easily how to set up the console for multiple entity managers or how to add more commands to it.

### DIFF
--- a/bin/doctrine-pear.php
+++ b/bin/doctrine-pear.php
@@ -1,5 +1,15 @@
 <?php
 
+fwrite(
+    STDERR,
+    '[Warning] The use of this script is discouraged. See'
+    . ' https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/tools.html#doctrine-console'
+    . ' for instructions on bootstrapping the console runner.'
+    . PHP_EOL
+);
+
+echo PHP_EOL . PHP_EOL;
+
 require_once 'Doctrine/Common/ClassLoader.php';
 
 $classLoader = new \Doctrine\Common\ClassLoader('Doctrine');
--- a/docs/en/reference/tools.rst
+++ b/docs/en/reference/tools.rst
@@ -7,24 +7,10 @@ Doctrine Console
 The Doctrine Console is a Command Line Interface tool for simplifying common
 administration tasks during the development of a project that uses ORM.
 
-Take a look at the :doc:`Installation and Configuration <configuration>`
-chapter for more information how to setup the console command.
+For the following examples, we will set up the CLI as ``bin/doctrine``.
 
-Display Help Information
-~~~~~~~~~~~~~~~~~~~~~~~~
-
-Type ``php vendor/bin/doctrine`` on the command line and you should see an
-overview of the available commands or use the --help flag to get
-information on the available commands. If you want to know more
-about the use of generate entities for example, you can call:
-
-.. code-block:: php
-
-    $> php vendor/bin/doctrine orm:generate-entities --help
-
-
-Configuration
-~~~~~~~~~~~~~
+Setting Up the Console
+~~~~~~~~~~~~~~~~~~~~~~
 
 Whenever the ``doctrine`` command line tool is invoked, it can
 access all Commands that were registered by a developer. There is no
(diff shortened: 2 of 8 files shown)


bin/doctrine-pear.php
```diff
@@ -1,5 +1,14 @@
 <?php
 
+fwrite(
+    STDERR,
+    '[Warning] The use of this script is discouraged.'
+    . ' You find instructions on how to boostrap the console runner in our documentation.'
```
greg0ire: You will find?

docs/en/reference/tools.rst
```diff
@@ -34,32 +20,52 @@ Doctrine DBAL and ORM. If you want to use additional commands you
...
 .. code-block:: php
 
+    #!/usr/bin/env php
     <?php
+
     use Doctrine\ORM\Tools\Console\ConsoleRunner;
+    use Doctrine\ORM\Tools\Console\EntityManagerProvider\SingleManagerProvider;
 
-    // replace this with the path to your own project bootstrap file.
+    // replace with file to your own project bootstrap
```
greg0ire: Replace with file to does not sound correct

docs/en/reference/tools.rst
```diff
@@ -34,32 +20,52 @@ Doctrine DBAL and ORM. If you want to use additional commands you
...
 
     You have to adjust this snippet for your specific application or framework
     and use their facilities to access the Doctrine EntityManager and
     Connection Resources.
 
+Display Help Information
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Type ``php bin/doctrine`` on the command line and you should see an
+overview of the available commands or use the --help flag to get
```
greg0ire: Please use backticks around the option

bin/doctrine-pear.php
derrabus: This is the same wording as we use in DBAL. I'm not a native speaker, but it sounds correct to me.

docs/en/reference/tools.rst
derrabus: You are right, but for the record: I only moved this block. 😓

docs/en/reference/tools.rst
derrabus: I think, it's better now.

bin/doctrine-pear.php
SenseException: I'd like to have the "You will find" correction here too. It sounds weird in the DBAL wording.

docs/en/reference/tools.rst
greg0ire: Sorry!

docs/en/reference/tools.rst
greg0ire: Yes :)

bin/doctrine-pear.php
greg0ire: @morozov WDYT? As someone who lives in the States?

morozov: I'd put it as "See \<url> for instructions on bootstrapping the console runner." (typo in "boostrap").

derrabus: Thanks, @morozov. I've updated the PR with your suggestion.

greg0ire (approved):

PR #9663, opened 2022-04-15
ScalarColumnHydrator: prevent early-bail on falsy values

### LINKED ISSUE #9230: Single column hydration returns nothing if first item is false-like

MrMitch:
Hi !

This is my first contribution to Doctrine, please let me know if there is anything I can do to facilitate the review process.

--

Fix #9230

Falsy values stop the hydration process in `$query->getSingleColumnResult()`. If the array of values fetched from the database contains a value that is "falsy", every value after that value and including that value is missing from the array that is returned.

This is because of the way the database values are retreived in ScalarColumnHydrator : https://github.com/doctrine/orm/blob/f4d5283f705f6ac51e81c8299ace074a619b40d7/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php#L31-L33

This PR fixes that by fetching all the values and returning them as is. As before, and to keep things consistent with the previous behavior of both ScalarHydrator and ScalarColumnHydrator, no type conversion is done between the database value and the PHP value. According to a comment in `AbstractHydrator`, this is the expected behavior for scalar hydrators that is a legacy behavior from 2.0 : https://github.com/doctrine/orm/blob/f4d5283f705f6ac51e81c8299ace074a619b40d7/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php#L498-L505

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
@@ -7,6 +7,7 @@ namespace Doctrine\ORM\Internal\Hydration;
 use Doctrine\DBAL\Driver\Exception;
 use Doctrine\ORM\Exception\MultipleSelectorsFoundException;
 
+use function array_column;
 use function count;
 
 /**
@@ -26,12 +27,8 @@ final class ScalarColumnHydrator extends AbstractHydrator
             throw MultipleSelectorsFoundException::create($this->resultSetMapping()->fieldMappings);
         }
 
-        $result = [];
+        $result = $this->statement()->fetchAllNumeric();
 
-        while ($row = $this->statement()->fetchOne()) {
-            $result[] = $row;
-        }
-
-        return $result;
+        return array_column($result, 0);
     }
 }
--- a/tests/Doctrine/Tests/Mocks/DriverResultMock.php
+++ b/tests/Doctrine/Tests/Mocks/DriverResultMock.php
@@ -64,7 +64,12 @@ class DriverResultMock implements Result, ResultStatement
 
     public function fetchAllNumeric(): array
     {
-        throw new BadMethodCallException('Not implemented');
+        $values = [];
+        while (($row = $this->fetchNumeric()) !== false) {
+            $values[] = $row;
+        }
+
+        return $values;
     }
 
     public function fetchAllAssociative(): array
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9230Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9230Test.php
@@ -1,0 +1,188 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group GH-9230
+ */
+class GH9230Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        // ensure entity table exists
+        $this->setUpEntitySchema([GH9230Entity::class]);
+    }
+
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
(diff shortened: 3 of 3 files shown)


tests/Doctrine/Tests/Models/ReadonlyProperties/Book.php
```diff
@@ -5,7 +5,6 @@
 namespace Doctrine\Tests\Models\ReadonlyProperties;
 
 use Doctrine\Common\Collections\ArrayCollection;
-use Doctrine\Common\Collections\Collection;
```
MrMitch: This as been automatically fixed by `vendor/bin/phpcbf` (and flagged by `vendor/bon/phpcs`). The original error was not related to this PR and was present before any changes were made.

```
# vendor/bin/phpcs 
.....................................E...................... 60 / 76 (79%)
................                                             76 / 76 (100%)



FILE: tests/Doctrine/Tests/Models/ReadonlyProperties/Book.php
--------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------------
 8 | ERROR | [x] Type Doctrine\Common\Collections\Collection is not used in this file.
   |       |     (SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse)
--------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------------

Time: 998ms; Memory: 174MB
```

```
# vendor/bin/phpcbf 
.................F.......................................... 60 / 76 (79%)
................                                             76 / 76 (100%)



PHPCBF RESULT SUMMARY
----------------------------------------------------------------------------
FILE                                                        FIXED  REMAINING
----------------------------------------------------------------------------
tests/Doctrine/Tests/Models/ReadonlyProperties/Book.php     1      0
----------------------------------------------------------------------------
A TOTAL OF 1 ERROR WERE FIXED IN 1 FILE
----------------------------------------------------------------------------

Time: 12,03 secs; Memory: 10MB
```



I'm happy to reverse if this is outside of the scope of this PR.

derrabus:
Your patch introduces some complexity to the codebase by duplicating logic from the DBAL. Naive question: Shouldn't your problem be fixed if you changed the loop this way?

```diff
-         while ($row = $this->statement()->fetchOne()) {
+         foreach ($this->statement()->iterateColumn() as $row) {
```

tests/Doctrine/Tests/Models/ReadonlyProperties/Book.php
derrabus: Yes, please revert. And please use PHP 8.1 to execute PHPCS.

MrMitch: @derrabus reverted, thanks !

MrMitch:
@derrabus Thank you for your quick reply. You're absolutely right, using a simple foreach loop produces the same result and reduces the complexity, I've amended the code accordingly.

We could even further reduce the length of the method by using `return iterator_to_array($this->statement()->iterateColumn())`, but I'm not sure that would add real value or maintainability.

MrMitch:
@derrabus I've amended the code further to fix support for databases that have native support for `boolean` values (like PostgreSQL).

derrabus (approved):

derrabus:
Thank you, @MrMitch!

MrMitch:
Happy to contribute, thank you for reviewing this with care.

PR #9665, opened 2022-04-19
Drop support for persistence 2

derrabus:


SenseException (dismissed):

.github/workflows/static-analysis.yml
```diff
@@ -23,10 +23,6 @@ jobs:
           - "default"
         persistence-version:
```
greg0ire: Should we remove this as well? I don't think we will test for Persistence 4 very soon.

derrabus: Good idea. Even if we do, we can still copy the config from the 2.5.x branch.

greg0ire (approved):

PR #9670, opened 2022-04-21
Revert "Implement colocated mapping driver"

greg0ire:
This reverts commit cd57768b0801ed380b372654a8db5edef5d38543.
The BC-break that was assumed to be harmless is in fact causing bugs in
the laminas integration.

A more involving but less BC-breaking upgrade path should be found later on.

https://github.com/doctrine/DoctrineModule/issues/774

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -41,16 +41,6 @@ This is now deprecated. Please extend `EntityRepository` instead.
 +$entityManager->getRepository(CmsUser::class);
 ```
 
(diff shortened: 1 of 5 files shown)


greg0ire:
Argh… this is not as easy as it looks, because now we support `doctrine/persistence` 3, and it no longer has this class. I might work on a conditional class definition later on, but this will be ugly :see_no_evil: 

It could look like this:

```
if (class_exists(PersistenceAnnotationDriver::class)) {
  previous code
} else {
  code with the BC break
}
```

derrabus:
See #9671 for an alternative.

derrabus:
Closing in favor of #9671.

PR #9671, opened 2022-04-22
Conditionally extend the old `AnnotationDriver` class

### LINKED ISSUE #9668: Annotation receives array of paths instead of reader
### Bug Report

2.12.0 doesn't work with `doctrine/doctrine-orm-module` : 5.1.0.

|    Q        |   A
|------------ | ------
| BC Break    | yes?
| Version     | 2.12.0

#### Summary

After upgrading from 2.11.1 to 2.12.0, I'm experiencing the following issue:

```
/var/www/html # vendor/bin/doctrine-module orm:generate-proxies
PHP Fatal error:  Uncaught Error: Call to a member function getClassAnnotations() on array in /var/www/html/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php:82
Stack trace:
#0 /var/www/html/vendor/doctrine/persistence/src/Persistence/Mapping/Driver/MappingDriverChain.php(79): Doctrine\ORM\Mapping\Driver\AnnotationDriver->loadMetadataForClass('tms\\V1\\Rest\\Sub...', Object(Doctrine\ORM\Mapping\ClassMetadata))
#1 /var/www/html/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php(133): Doctrine\Persistence\Mapping\Driver\MappingDriverChain->loadMetadataForClass('tms\\V1\\Rest\\Sub...', Object(Doctrine\ORM\Mapping\ClassMetadata))
```

#### Current behavior

Reading Metadata no longer works through the `doctrine-orm-module`.

In the following line:
https://github.com/doctrine/orm/blob/d5503644314ed825fe4b276a588458a033fa3e7c/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L82
`$this->reader` no longer contains the reader, but the array of paths:
```
[
    __DIR__ . '/../../module/xxx/src/V1/Rest',
    __DIR__ . '/../../module/xxx/src/V1/Rpc',
    __DIR__ . '/../../module/xxx/src',
]
```

Currently, my best guess is, that this class is instanciated with a missing first parameter. And the paths are then passed as first parameter in the readers place.

#### How to reproduce

`doctrine/doctrine-orm-module` : 5.1.0
`doctrine/orm`: 2.12.0
PHP : 8.0.16

derrabus:
Fixes #9668.
Replaces #9670.

This PR restores the old inheritance chain of `AnnotationDriver` and `AttributeDriver` if Persistence 2 is installed. Unfortunately, DoctrineModule uses a flawed logic to determine the constructor arguments of a mapping driver class, see doctrine/DoctrineModule#774. With this change, that logic should work again.

When merging this PR up to 3.0.x, I would simply revert it.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -41,11 +41,6 @@ This is now deprecated. Please extend `EntityRepository` instead.
 +$entityManager->getRepository(CmsUser::class);
 ```
 
-## BC Break: `AttributeDriver` and `AnnotationDriver` no longer extends parent class from `doctrine/persistence`
-
-Both these classes used to extend an abstract `AnnotationDriver` class defined
-in `doctrine/persistence`, and no longer do.
-
 ## Deprecate `AttributeDriver::getReader()` and `AnnotationDriver::getReader()`
 
 That method was inherited from the abstract `AnnotationDriver` class of
--- a/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
@@ -14,7 +14,6 @@ use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\ColocatedMappingDriver;
-use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use ReflectionClass;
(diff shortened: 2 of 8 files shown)


Ocramius (approved):
Approach LGTM :+1:

malarzm (approved):

greg0ire (approved):
Awesome! Because of my lack of experience with traits, I didn't expect the trait not to cause any issue since it redefines the same methods as the parent, but this looks really good 👍

PR #9675, opened 2022-04-23
Fix type on ClassMetadata discriminatorMap

franmomu:
The psalm issue is because of the `ltrim` call in https://github.com/doctrine/orm/blob/2e4a8722721b934149ff53b191522a6829b6d73b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L3245

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -171,12 +171,12 @@ parameters:
 			path: lib/Doctrine/ORM/Id/TableGenerator.php
 
 		-
-			message: "#^Parameter \\#2 \\$discrMap of static method Doctrine\\\\ORM\\\\Internal\\\\Hydration\\\\HydrationException\\:\\:invalidDiscriminatorValue\\(\\) expects array\\<string, string\\>, array\\<int, int\\|string\\> given\\.$#"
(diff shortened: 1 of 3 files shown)


phpstan-baseline.neon
```diff
@@ -171,12 +171,12 @@ parameters:
 			path: lib/Doctrine/ORM/Id/TableGenerator.php
 
 		-
-			message: "#^Parameter \\#2 \\$discrMap of static method Doctrine\\\\ORM\\\\Internal\\\\Hydration\\\\HydrationException\\:\\:invalidDiscriminatorValue\\(\\) expects array\\<string, string\\>, array\\<int, int\\|string\\> given\\.$#"
+			message: "#^Parameter \\#2 \\$discrMap of static method Doctrine\\\\ORM\\\\Internal\\\\Hydration\\\\HydrationException\\:\\:invalidDiscriminatorValue\\(\\) expects array\\<string, string\\>, array\\<int, string\\> given\\.$#"
```
derrabus, line 174: Those errors should be easy to fix as well. 🤔

derrabus (approved):

phpstan-baseline.neon
franmomu: I'll take a look

PR #9676, opened 2022-04-24
Fix HydrationException::invalidDiscriminatorValue parameter type

franmomu:
Ref https://github.com/doctrine/orm/pull/9675#discussion_r857144377

The calls to `HydrationException::invalidDiscriminatorValue` are using `array_keys`:

https://github.com/doctrine/orm/blob/1ac05f5e4e2819d092e7c9605d0758370f6acfc8/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php#L241

https://github.com/doctrine/orm/blob/1ac05f5e4e2819d092e7c9605d0758370f6acfc8/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php#L108

So what `invalidDiscriminatorValue` is receiving are the possible discriminator values.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/HydrationException.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/HydrationException.php
@@ -84,17 +84,17 @@ class HydrationException extends ORMException
 
     /**
      * @param string   $discrValue
-     * @param string[] $discrMap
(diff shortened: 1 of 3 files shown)


derrabus (approved):
Thanks! ❤️

PR #9677, opened 2022-04-24
Document ORM drivers only really load ORM metadata

greg0ire:
Note: I'm not sure I made a great choice in https://github.com/doctrine/persistence/pull/180 when I used templating in `MappingDriver` to bind `$className` to `$metadata`, as double templating does not seem possible, meaning we can't use it to get rid of the docblocks I added:
- https://phpstan.org/r/4e171edc-bd64-410d-a093-2105e0e10838
- https://psalm.dev/r/63c4eac059

Although this is clearly not a bugfix and touches `src`, I'm targeting 2.12.x because `assert` is not supposed to affect production, and touching baseline files is best done on the lowest-supported branch IMO, so that there are not too many conflicts.


lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -10,6 +10,7 @@
 use Doctrine\ORM\Events;
 use Doctrine\ORM\Mapping;
 use Doctrine\ORM\Mapping\Builder\EntityListenerBuilder;
+use Doctrine\ORM\Mapping\ClassMetadata as Metadata;
```
derrabus: I find this alias confusing. Let's use `OrmClassMetadata` or better: Alias `ClassMetadata` from Persistence as `PersistenceClassMetadata` and use `ClassMetadata` for our own implementation.

tests/Doctrine/Tests/ORM/Functional/DatabaseDriverTestCase.php
```diff
@@ -31,7 +31,7 @@ protected function convertToClassMetadata(array $entityTables, array $manyTables
 
         $metadatas = [];
         foreach ($driver->getAllClassNames() as $className) {
-            $class = new ClassMetadataInfo($className);
+            $class = new ClassMetadata($className);
```
greg0ire, line 34: I first implemented this with `assert`, then was forced to changed this test, which made me think this was a bit risky, hence the change towards phpdoc. It's better anyway, since it makes static analysis able to detect calls to these methods made with `ClassMetadataInfo` rather than `ClassMetadata`.

derrabus (approved):

derrabus:
> `assert` is not supposed to affect production

You're only removing `assert()` calls and not adding new ones. I think targeting 2.12.x is fine.

greg0ire:
> You're only removing assert() calls and not adding new ones. I think targeting 2.12.x is fine.

Sorry, when I wrote this I had added some calls to `assert`, then switched to phpdoc and forgot to edit that part out.

PR #9679, opened 2022-04-25
Drop SymfonyStyle[listing] for sqls

simPod:
This was super annoying as UpdateCommand printed sqls prefixed with `*` so it was not possible to copy statements anymore without manually removing those asterisks.

This removes prefixing sqls and makes it consistent with Create and Drop commands.

Also, semicolons at the end were missing.

image
image


derrabus (approved):
Looks okay to me. Not sure if this is really a bugfix though.

simPod:
It broke in 2.12 (https://github.com/doctrine/orm/pull/9019) 🤷🏾

derrabus:
Indeed.

PR #9680, opened 2022-04-25
Fill entity ids with foreign entities in uow::createEntity(). GH8926.

Feolius:
If entity uses identity through foreign key and contains some eager loading relationships, an error will be thrown on attempt to load parent entity (which key is used for identity). With this fix, I fill entity id on earlier stage before processing association mappings using identity map. 
Fixes GH8926

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2766,6 +2766,16 @@ class UnitOfWork implements PropertyChangedListener
         }
 
         foreach ($class->associationMappings as $field => $assoc) {
+            // If identity through foreign key is used, then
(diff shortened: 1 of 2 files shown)


beberlei:
Adding another loop over association mappings is a bit much in this critical path for performance. Since this is an edge use case, i am unsure and would prefer another fix

Feolius:
But this loop is super fast! you can see that there are basic checks only. If current association is not an id it will just pass. I could move this codeblock inside original loop. And I tried that initially. And it even works. All tests are passed. But it works ONLY because people usually put fields which are ids on the first place since the associations order is determined by fields order in entity description as far as I understand. So, it's not reliable.

guvkon (approved):

barzhomi (approved):

PR #9686, opened 2022-04-26
added missing exception type for deletions

ppaulis:
Signed-off-by: Pascal Paulis <ppaulis@gmail.com>

Deletions can potentially throw `Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException`. But in `UnitOfWork#commit()`, those are catched via `Throwable` and are rethrown after some closing operations.

So, currently, it's not evident for developers to see that they could get this type of exception.

This PR proposes to add the `ForeignKeyConstraintViolationException` in the catch():
```php
} catch (Throwable | ForeignKeyConstraintViolationException $e) {
```
(supported since PHP 7.1 : https://wiki.php.net/rfc/multiple-catch, so no BC break)

 and in the docBlock of `UnitOfWork#commit()`.

Like this, IDEs and quality tools can signal potential problems with this kind of exception.

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -12,6 +12,7 @@ use Doctrine\Common\Persistence\PersistentObject;
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException;
 use Doctrine\DBAL\LockMode;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Exception\EntityManagerClosed;
@@ -381,6 +382,8 @@ use function strpos;
      * @throws OptimisticLockException If a version check on an entity that
      * makes use of optimistic locking fails.
      * @throws ORMException
+     * @throws ForeignKeyConstraintViolationException
+     * @throws Throwable
      */
     public function flush($entity = null)
     {
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -10,6 +10,7 @@ use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\EventManager;
 use Doctrine\Common\Proxy\Proxy;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
+use Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException;
 use Doctrine\DBAL\LockMode;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Cache\Persister\CachedPersister;
@@ -39,7 +40,6 @@ use Doctrine\Persistence\Mapping\RuntimeReflectionService;
 use Doctrine\Persistence\NotifyPropertyChanged;
 use Doctrine\Persistence\ObjectManagerAware;
 use Doctrine\Persistence\PropertyChangedListener;
-use Exception;
 use InvalidArgumentException;
 use RuntimeException;
 use Throwable;
@@ -337,7 +337,7 @@ class UnitOfWork implements PropertyChangedListener
      *
      * @return void
      *
-     * @throws Exception
+     * @throws Throwable|ForeignKeyConstraintViolationException
      */
     public function commit($entity = null)
     {
@@ -456,7 +456,7 @@ class UnitOfWork implements PropertyChangedListener
 
                 throw new OptimisticLockException('Commit failed', $object);
             }
-        } catch (Throwable $e) {
+        } catch (Throwable | ForeignKeyConstraintViolationException $e) {
             $this->em->close();
 
             if ($conn->isTransactionActive()) {
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -337,7 +337,8 @@ class UnitOfWork implements PropertyChangedListener
      *
      * @return void
      *
-     * @throws Throwable|ForeignKeyConstraintViolationException
+     * @throws Throwable
+     * @throws ForeignKeyConstraintViolationException
      */
     public function commit($entity = null)
     {
@@ -456,7 +457,7 @@ class UnitOfWork implements PropertyChangedListener
 
                 throw new OptimisticLockException('Commit failed', $object);
             }
-        } catch (Throwable | ForeignKeyConstraintViolationException $e) {
+        } catch (Throwable $e) {
             $this->em->close();
 
             if ($conn->isTransactionActive()) {


greg0ire:
This looks weird to me… if you had to come up with a general rule that could be applied across all Doctrine projects that would determine when to do that and when not to do it, what would it look like?

Right now, I think I would expect SA tools and IDEs to point out that `Throwable | AnyException` is equivalent to `Throwable`. I approved the build and we can see that right now they don't, but I wouldn't be surprised if that became the case in the future.

ppaulis:
@greg0ire  The use case was that I showed a new colleague how to do error handling in PHP. So he handled in this particular case `\Throwable` only, because, following the docblock of this method, it's the only type that is thrown. The error was of course catched, but he had a hard time debugging the foreign key error afterwards.

So, to answer your question, I would at least expect the docblock to mention all the exception types via:
```php
/**
* @throws Exception1
* @throws Exception2
* ...
**/
```
Technically speaking, adding the Exception in the catch indeed isn't necessary, but imho, in the docblock it makes sense. So, when working with Doctrine for the first time, you can see what you need to take into consideration when dealing with errors. I cleaned it up a little bit, and added the correct throws-directives so that it gets mentionned in the IDE.

I had a similar case some time ago in the official Elasticsearch php client. There is a very handy exception called `Missing404Exception`. But the `get()` call doesn't document it. So I had to find out by trial & error how to handle missing documents.

derrabus:
`@throws Throwable` never makes sense to document, sorry. It's the widest possible exception type and virtually any method could raise it.

greg0ire:
@ppaulis let's say we find a tool to automate that… aren't you afraid this might result in bigger and bigger docblocks, the higher you are in the the call tree? I mean if `A()` calls `B()` and `C()`, and the 2 last functions throw 5 different exception types each, should be end up with 10 `@throws` lines on `A()`?

I've never given much thought to `@throws`, but when I think of it, I don't know what rule to come up with to get to a sensible result.

ppaulis:
@greg0ire I always prefer a list of more precise exceptions to handle. The user can still decide to handle them one by one or all together. I agree with @derrabus here, that it's not very practical to use `\Throwable`. I wasn't bold enough to rewrite the exception handling of the commit method more deeply.

For me it's a question of encapsulation. Each method is responsible for exposing a clear interface to it's callers. At the moment, `UnitOfWork#commit()` is hiding the foreign key exception behind a `\Throwable`. Instead, imho it should say that it throws this kind of exception and let the calling methods decide how to deal with it. `EntityMager#flush()` should then decide to handle it or to pass it through, but documented. By documenting the foreign key exception, we would make the current situation more easily understandable.

At the moment, the current docblock seems also incorrect to me, because the catch re-throws potentially a `\Throwable` but the docblock says `\Exception`.

ppaulis:
@greg0ire Let me put it that way perhaps: By not documenting the exact exception types in a method, we are simply obscuring a more fine-grained error handling that currently is already in place.

greg0ire:
Answering my own question: here is [a package that could allow us to automate this](https://github.com/pepakriz/phpstan-exception-rules). Without any configuration, it produces 1432 errors. 

Here is what it produces for `EntityManager`:

```
Note: Using configuration file /home/greg/dev/doctrine-orm/phpstan.neon.
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ---------------------------------------------------------------------------------
  Line   UnitOfWork.php
 ------ ---------------------------------------------------------------------------------
  468    Missing @throws Throwable annotation
  819    Missing @throws Doctrine\ORM\Exception\ORMException annotation
  819    Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  993    Missing @throws Doctrine\ORM\Mapping\MappingException annotation
  993    Missing @throws Doctrine\ORM\Mapping\MappingException annotation
  1007   Missing @throws InvalidArgumentException annotation
  1075   Missing @throws RuntimeException annotation
  1134   Missing @throws Doctrine\ORM\Mapping\MappingException annotation
  1135   Missing @throws Doctrine\ORM\Mapping\MappingException annotation
  1146   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1183   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1193   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1214   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1345   Useless @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1484   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1496   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1773   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1773   Missing @throws UnexpectedValueException annotation
  1858   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1858   Missing @throws UnexpectedValueException annotation
  1934   Missing @throws Doctrine\ORM\EntityNotFoundException annotation
  1934   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  1985   Missing @throws Doctrine\ORM\Exception\ORMException annotation
  2017   Missing @throws Doctrine\ORM\Exception\ORMException annotation
  2021   Missing @throws Doctrine\ORM\Exception\ORMException annotation
  2156   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2256   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2262   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2347   Missing @throws Doctrine\ORM\EntityNotFoundException annotation
  2347   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2347   Missing @throws Doctrine\ORM\OptimisticLockException annotation
  2350   Missing @throws Doctrine\ORM\EntityNotFoundException annotation
  2350   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2350   Missing @throws Doctrine\ORM\OptimisticLockException annotation
  2384   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2392   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2392   Missing @throws UnexpectedValueException annotation
  2399   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2406   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2406   Missing @throws UnexpectedValueException annotation
  2461   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2461   Missing @throws UnexpectedValueException annotation
  2546   Unused @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  2790   Missing @throws Doctrine\ORM\Mapping\MappingException annotation
  2811   Missing @throws Doctrine\ORM\Mapping\MappingException annotation
  2817   Missing @throws Doctrine\ORM\Mapping\MappingException annotation
  2866   Missing @throws Doctrine\ORM\Mapping\MappingException annotation
  2873   Missing @throws Doctrine\Common\Proxy\Exception\OutOfBoundsException annotation
  2881   Missing @throws Doctrine\Common\Proxy\Exception\OutOfBoundsException annotation
  3083   Missing @throws Doctrine\ORM\EntityNotFoundException annotation
  3198   Missing @throws RuntimeException annotation
  3264   Missing @throws Doctrine\ORM\ORMInvalidArgumentException annotation
  3535   Unused @throws Doctrine\ORM\Exception\ORMException annotation
  3535   Useless @throws Doctrine\ORM\OptimisticLockException annotation
  3535   Useless @throws Doctrine\ORM\TransactionRequiredException annotation
  3577   Missing @throws Doctrine\Common\Proxy\Exception\OutOfBoundsException annotation
  3681   Missing @throws Doctrine\DBAL\Exception annotation
 ------ ---------------------------------------------------------------------------------



 [ERROR] Found 57 errors
```

As you can see, it didn't spot the `ForeignKeyConstraintViolationException`, probably because we would need to do this on `doctrine/dbal ` first. Dealing with exceptions feels kind of hopeless to be honest, and keeping this up-to-date without something to check that the exception class still exists could be a challenge. I'm going to close this, as I don't see a realistic outcome, but feel free to research the topic and make proposals on how to deal with this.

PR #9708, opened 2022-05-01
Fix psalm annotation for Entity attribute

### LINKED ISSUE #9681: Entity Annotation & EntityRepository Template causes Psalm Covariance Issues
### Bug Report

When updating our `doctrine/orm` dependency to `2.12.1` we experienced Psalm issues of the following kind in our CI pipeline:

```
ERROR: InvalidArgument - src/Common/User/Entity/UserConfig.php:14:27 - Argument 1 of Doctrine\ORM\Mapping\Entity::__construct expects class-string<Doctrine\ORM\EntityRepository<object>>|null, IServ\Auth\Common\User\Repository\DoctrineUserConfigRepository::class provided (see https://psalm.dev/004)
#[Entity(repositoryClass: DoctrineUserConfigRepository::class, readOnly: true)]
```

|    Q        |   A
|------------ | ------
| BC Break    | yes?
| Version     | 2.12.1

#### Summary

This is reproducible for any entity that configures a repository class using the `Entity` Annotation / Class Attribute if the repository class uses Psalm generics:

```php
#[Entity(repositoryClass: DoctrineUserConfigRepository::class, readOnly: true)]
#[Table(name: "config")]
class UserConfig
{
}
```

```php
/**
 * @template-extends EntityRepository<UserConfig>
 */
final class DoctrineUserConfigRepository extends EntityRepository
{
}
```

The issue stems from the `Entity` Mapping annotation to require a `class-string<EntityRepository>` here: https://github.com/doctrine/orm/blob/a552df66a978fe91762e37f92662ec7900ec21ee/lib/Doctrine/ORM/Mapping/Entity.php#L28-L31

This breaks due to what I believe to be covariance issues:  https://psalm.dev/docs/annotating_code/templated_annotations/#template-covariance

#### How to reproduce

I have recreated a simplified reproduction case using the Psalm playground: https://psalm.dev/r/db2ba03745

#### Expected behavior

I would expect the `Entity` annotation to continue working as it did before the update.

#### Potential Solution 1

Adding `@template-covariant T` to the `EntityRepository` class DocBlock seems to resolve all Psalm issues for us. I am not 100% sure if that has any other implications, though.

Side Note: We use the `ServiceEntityRepository` class provided by `doctrine-bundle`, which transparently extends `EntityRepository` - any solution for this issue thus likely also applies to that class as well (e.g. adding the `@template-covariant T` there, too).

#### Potential Solution 2

Maybe the `Entity` Attribute / Mapping Annotation could also get a generic template parameter, something like this:

```patch
diff --git a/app/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/Entity.php b/app/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/Entity.php
--- a/app/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/Entity.php	
+++ b/app/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/Entity.php
@@ -12,13 +12,14 @@
  * @Annotation
  * @NamedArgumentConstructor()
  * @Target("CLASS")
+ * @template T of object
  */
 #[Attribute(Attribute::TARGET_CLASS)]
 final class Entity implements Annotation
 {
     /**
      * @var string|null
-     * @psalm-var class-string<EntityRepository>|null
+     * @psalm-var class-string<EntityRepository<T>>|null
      */
     public $repositoryClass;
 
@@ -26,7 +27,7 @@
     public $readOnly = false;
 
     /**
-     * @psalm-param class-string<EntityRepository>|null $repositoryClass
+     * @psalm-param class-string<EntityRepository<T>>|null $repositoryClass
      */
     public function __construct(?string $repositoryClass = null, bool $readOnly = false)
     {
```

This also solves the Psalm issues in our codebase and seems like a less intrusive, more "correct" fix.

michnovka:
Fixes https://github.com/doctrine/orm/issues/9681

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Entity.php
+++ b/lib/Doctrine/ORM/Mapping/Entity.php
@@ -12,13 +12,14 @@ use Doctrine\ORM\EntityRepository;
  * @Annotation
  * @NamedArgumentConstructor()
  * @Target("CLASS")
+ * @template T of object
  */
 #[Attribute(Attribute::TARGET_CLASS)]
 final class Entity implements Annotation
 {
     /**
      * @var string|null
-     * @psalm-var class-string<EntityRepository>|null
+     * @psalm-var class-string<EntityRepository<T>>|null
      */
     public $repositoryClass;
 
@@ -26,7 +27,7 @@ final class Entity implements Annotation
     public $readOnly = false;
 
     /**
-     * @psalm-param class-string<EntityRepository>|null $repositoryClass
+     * @psalm-param class-string<EntityRepository<T>>|null $repositoryClass
      */
     public function __construct(?string $repositoryClass = null, bool $readOnly = false)
     {


michnovka:
Since `EntityRepository` has already `@template` annotation declared, the only proper way is to always work with `EntityRepository<T>` in any psalm code. Not doing so is invalid psalm syntax, as can be seen here:
https://psalm.dev/r/d706b6cb58

greg0ire (approved):

SenseException (approved):

chalasr (approved):

greg0ire:
Thanks @michnovka !

michnovka:
np @greg0ire . When do you plan 2.12.2 patch release?

greg0ire:
@michnovka possibly this week

PR #9714, opened 2022-05-01
Undeprecated DisconnectedClassMetadataFactory

greg0ire:
It is used in the command that generates proxies.


greg0ire:
Wait no, it's used to test MetadataFilter, which is used in the command that generates proxies… but the disconnected cmf is only used in tests, so it can probably be removed somehow.

PR #9727, opened 2022-05-03
Run builds with DBAL 4.0.x-dev

morozov:
Relevant changes from DBAL 4.0.x:
1. https://github.com/doctrine/dbal/pull/3488
2. https://github.com/doctrine/dbal/pull/3569
3. https://github.com/doctrine/dbal/pull/3583

**TODO**:
- [x] Backport what's possible to 2.13.x and rebase: https://github.com/doctrine/orm/pull/9730
- [x] Address static analysis issues when testing against DBAL 3.4.x-dev.
- [x] ~~Why do discriminator maps have numeric keys despite their `<string,class-string>` type?~~ Known issue to be baselined.
- [x] Improve MySQL test suite compatibility with DBAL 4 (https://github.com/doctrine/orm/pull/9728).

**Improve DBAL upgrade path**:
- [x] Get https://github.com/doctrine/dbal/pull/5373 merged.
- [x] Get https://github.com/doctrine/dbal/pull/5377 merged.

**Next steps**:
1. https://github.com/doctrine/orm/pull/9735.
2. Test the build with `^4@dev` on PostgreSQL.
3. Run static analysis with `^4@dev`.
4. Bump DBAL version constraint to `^3.4` once it's released.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1063,6 +1063,9 @@
     <PossiblyUndefinedVariable occurrences="1">
       <code>$columnList</code>
     </PossiblyUndefinedVariable>
+    <RedundantCastGivenDocblockType occurrences="1">
+      <code>(string) $discrValues[$subclassName]</code>
+    </RedundantCastGivenDocblockType>
   </file>
   <file src="lib/Doctrine/ORM/Proxy/ProxyFactory.php">
     <ArgumentTypeCoercion occurrences="2">
--- a/tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterTypeValueSqlTest.php
+++ b/tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterTypeValueSqlTest.php
@@ -203,7 +203,7 @@ class BasicEntityPersisterTypeValueSqlTest extends OrmTestCase
 
         $connection->expects($this->atLeast(2))
             ->method('delete')
-            ->will($this->onConsecutiveCalls(
+            ->withConsecutive(
                 [
                     'customtype_parent_friends',
                     ['friend_customtypeparent_id' => 1],
@@ -214,7 +214,7 @@ class BasicEntityPersisterTypeValueSqlTest extends OrmTestCase
                     ['customtypeparent_id' => 1],
                     ['integer'],
                 ]
-            ));
+            );
 
         $persister->delete($parent);
     }
--- a/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
@@ -87,6 +87,7 @@ class AnnotationDriver implements MappingDriver
                 continue;
             }
 
+            unset($classAnnotations[$key]);
             $classAnnotations[get_class($annot)] = $annot;
         }
 
--- a/lib/Doctrine/ORM/OptimisticLockException.php
+++ b/lib/Doctrine/ORM/OptimisticLockException.php
@@ -7,6 +7,7 @@ namespace Doctrine\ORM;
 use DateTimeInterface;
 use Doctrine\ORM\Exception\ORMException;
 use Exception;
+use Throwable;
 
 /**
  * An OptimisticLockException is thrown when a version check on an object
@@ -21,9 +22,9 @@ class OptimisticLockException extends Exception implements ORMException
      * @param string      $msg
      * @param object|null $entity
      */
-    public function __construct($msg, $entity)
+    public function __construct($msg, $entity, ?Throwable $previous = null)
     {
-        parent::__construct($msg);
+        parent::__construct($msg, 0, $previous);
         $this->entity = $entity;
     }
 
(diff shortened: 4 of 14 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
         "ext-ctype": "*",
         "doctrine/collections": "^1.5",
         "doctrine/common": "^3.3",
-        "doctrine/dbal": "^3.3",
+        "doctrine/dbal": "^3.3 || 4.0.x-dev",
```
derrabus: We shouldn't declare compatibility here just yet. The new CI job should work without this change.

morozov: This is for testing (probably, needed only locally). I’ll revert it before the merge.

morozov:
@derrabus do you have a recommendation on addressing this static analysis issue?

> lib/Doctrine/ORM/Persisters/Entity/SingleTablePersister.php:144:44: RedundantCastGivenDocblockType: Redundant cast to string given docblock-provided type

It's related to this bullet point:

> Why do discriminator maps have numeric keys despite their `<string,class-string>` type?

One of the non-string occurrences is addressed by adding this: https://github.com/doctrine/orm/blob/bdf997afdcebd464e8139374f86bb8dabf512930/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L90

But there are more.

tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterTypeValueSqlTest.php
```diff
@@ -203,7 +203,7 @@ public function testDeleteManyToManyUsesTypeValuesSQL(): void
 
         $connection->expects($this->atLeast(2))
             ->method('delete')
-            ->will($this->onConsecutiveCalls(
+            ->withConsecutive(
```
morozov, line 206: Note, `withConsecutive()` will be deprecated in PHPUnit 10 without a replacement (https://github.com/sebastianbergmann/phpunit/issues/4564). The alternative looks quite clumsy:

```php
$expectedArguments = [
    [
        'customtype_parent_friends',
        ['friend_customtypeparent_id' => 1],
        ['integer'],
    ],
    [
        'customtype_parent_friends',
        ['customtypeparent_id' => 1],
        ['integer'],
    ],
];

$connection->expects($this->atLeast(2))
    ->method('delete')
    ->willReturnCallback(function (string $table, array $where, array $types) use (&$expectedArguments): int {
        $currentExpectedArguments = current($expectedArguments);
        next($expectedArguments);
        if ($currentExpectedArguments !== false) {
            self::assertEquals($currentExpectedArguments, [$table, $where, $types]);
        }

        return 0;
    });
```

Let's use the available API for now and rework it if/when it becomes necessary.

derrabus:
> @derrabus do you have a recommendation on addressing this static analysis issue?

This is a known issue with discriminator maps. Baseline it for now, fixing it is out of scope for this PR.

composer.json
```diff
@@ -25,7 +25,7 @@
         "ext-ctype": "*",
         "doctrine/collections": "^1.5",
         "doctrine/common": "^3.3",
-        "doctrine/dbal": "^3.3",
+        "doctrine/dbal": "^3.4.x-dev",
```
morozov: Would it be acceptable to bump this version constraint temporarily (we can pin it to a specific commit)?

I plan to add a couple more upgrade path improvements to the DBAL 3.4 (see the description). Since they are not yet released, if we don't bump the constraint, we will either have to force the release in order to avoid the dependency of the dev ORM on dev DBAL, or we will have to refrain from merging all the changes related to the DBAL 4 until they are implemented in full and the scope of the upgrade path improvements is fully defined.

I believe this change is acceptable since it's unlikely that ORM 3 will be released before DBAL 3.4. Even if this happens, we can release DBAL 3.4 right before that.

greg0ire: I think it's fine, as you say we can still force publish 3.4 in the unlikely event we want to publish ORM 3 before it is ready.

psalm-baseline.xml
```diff
@@ -1063,6 +1063,9 @@
     <PossiblyUndefinedVariable occurrences="1">
       <code>$columnList</code>
     </PossiblyUndefinedVariable>
+    <RedundantCastGivenDocblockType occurrences="1">
+      <code>(string) $discrValues[$subclassName]</code>
+    </RedundantCastGivenDocblockType>
```
greg0ire, line 1068: This should be fixed up into b291c4eda26014390f466cb03448e3545726089a

composer.json
derrabus: This is fine, but let's use this constraint instead:

```suggestion
        "doctrine/dbal": "^3.4@dev",
```

derrabus:
We want the 2.x series to run deprecation-free with DBAL 3.x. This PR seems to prepare the ORM for some deprecations coming with DBAL 3.4.0. Can we backport parts of this PR to 2.13.x?

morozov:
Yeah, about half of these commits could be backported to 2.13. Even though some of them wouldn't address runtime deprecations, they would set an example of writing code compatible with DBAL 4 (e.g. 9f0ff98, 96b1c37, 99aa247).

I can start with backporting and once they are merged back up, we can continue here. In the meantime, I can address any issues regarding the code changes.

composer.json
morozov: Yeah, we can use that. How is it better than the original constraint?

derrabus:
Sounds good.

composer.json
derrabus: It's consitent with what we usually use in those situations.

derrabus (approved):

PR #9730, opened 2022-05-05
Forward compatibility with DBAL 4

morozov:
This is a partial backport of https://github.com/doctrine/orm/pull/9727. The relevant changes from DBAL 4.0.x are:

1. doctrine/dbal#5208
2. doctrine/dbal#4933
3. doctrine/dbal#5043
4. doctrine/dbal#3586
5. doctrine/dbal#3574
6. doctrine/dbal#4764

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -146,9 +146,9 @@ final class Query extends AbstractQuery
     /**
      * The first result to return (the "offset").
      *
-     * @var int|null
+     * @var int
      */
-    private $firstResult = null;
+    private $firstResult = 0;
 
     /**
      * The maximum number of results to return (the "limit").
@@ -650,11 +650,7 @@ final class Query extends AbstractQuery
      */
     public function setFirstResult($firstResult): self
     {
-        if ($firstResult !== null) {
-            $firstResult = (int) $firstResult;
-        }
-
-        $this->firstResult = $firstResult;
+        $this->firstResult = (int) $firstResult;
         $this->_state      = self::STATE_DIRTY;
 
         return $this;
--- a/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
@@ -9,6 +9,7 @@ use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\ForeignKeyConstraint;
 use Doctrine\DBAL\Schema\SchemaException;
 use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\Inflector\Inflector;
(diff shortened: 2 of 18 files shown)


composer.json
```diff
@@ -26,7 +26,7 @@
         "doctrine/cache": "^1.12.1 || ^2.1.1",
         "doctrine/collections": "^1.5",
         "doctrine/common": "^3.0.3",
-        "doctrine/dbal": "^2.13.1 || ^3.2",
+        "doctrine/dbal": "^2.13.1 || ^3.4@dev",
```
morozov, line 29: This way, we test some builds twice: once as `^3.4@dev` here and once as overridden with `^3@dev` in the GitHub Actions configuration. I believe it's fine since it's a temporary state. Reworking the build matrix for the sake of purity isn't worth it.

lib/Doctrine/ORM/Query.php
```diff
@@ -146,9 +146,9 @@ final class Query extends AbstractQuery
     /**
      * The first result to return (the "offset").
      *
-     * @var int|null
+     * @var int
      */
-    private $firstResult = null;
+    private $firstResult = 0;
```
beberlei, line 151: This will require a deprecation of setFirstResult(null). The modifyLimitQuery related changes rhis enables are nice!

beberlei (approved):

lib/Doctrine/ORM/Query.php
morozov: I'd say it's a valid logical next step but not a requirement. From the standpoint of the end result, `0` and `null` here are and have always been identical, so this patch doesn't change anything about that.

derrabus: This is actually a breaking change. Symfony's test suite is now failing because Symfony's `ORMQueryBuilderLoader` checks if `getFirstResult()` returns `null`.

derrabus: Correction: The change on the `QueryBuilder` is actually the one that breaks Symfony's DoctrineBridge.

morozov: What part of the API contract does this change break?

derrabus: It breaks code that does not treat `0` and `null` as equal. One could argue though that such a piece of code has always been broken or fragile at least.

Here's the fix for Symfony: symfony/symfony#46282

morozov: No, I specifically mean the API contract. If passed null, the component will do this. If passed a zero, it will do that. What has changed? (Thanks for taking care of that, BTW).

Looking at the patch, I'd say you fixed an existing bug. It existed prior to the change in the ORM and now became reproducible via the existing test suite.

derrabus: The return value of `$em->createQueryBuilder()->getFirstResult()` has changed from `null` to `0`. Do we consider this to be part of the API contract? The doc block says we do:

https://github.com/doctrine/orm/blob/07ee5552793e618ac2e3ac9f3541b3749238b8e9/lib/Doctrine/ORM/QueryBuilder.php#L651

morozov: In this case, it's a minor BC break. The proper documentation should have said "Returns NULL or 0..." and the clients should have checked for null or zero. Then this change wouldn't have broken anything.

PR #9741, opened 2022-05-08
 Make UnderscoreNamingStrategy number-aware only

greg0ire:
See https://github.com/doctrine/orm/pull/7908


tests/Doctrine/Tests/ORM/Mapping/NamingStrategyTest.php
```diff
@@ -88,7 +70,7 @@ public function testClassToTableName(NamingStrategy $strategy, string $expected,
     /**
      * Data Provider for NamingStrategy#propertyToColumnName
      *
-     * @return array<NamingStrategy|string>
+     * @return array<array{NamingStrategy, string, string}>
```
derrabus: For 2.12?

derrabus (dismissed):

SenseException (dismissed):

tests/Doctrine/Tests/ORM/Mapping/NamingStrategyTest.php
greg0ire: I thought of that but elected not to because laziness, but now that I've given it a try, I've found more issue in that test that need to be fixed, so it feels a bit more worthwhile :slightly_smiling_face: . See https://github.com/doctrine/orm/pull/9746

derrabus:
Let's rebase and merge?

derrabus (approved):

PR #9743, opened 2022-05-08
[DDC-2917] Fix GroupBy SQL generation for entities with JOINED inheritance type

popov-a-e:
Resolves https://github.com/doctrine/orm/issues/3676

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -25,6 +25,7 @@ use function array_filter;
 use function array_keys;
 use function array_map;
 use function array_merge;
+use function array_reverse;
 use function assert;
 use function count;
 use function implode;
@@ -35,7 +36,6 @@ use function is_numeric;
 use function is_string;
 use function preg_match;
 use function reset;
-use function sort;
 use function sprintf;
 use function strtolower;
 use function strtoupper;
@@ -1739,7 +1739,7 @@ class SqlWalker implements TreeWalker
             return $this->walkPathExpression($pathExpr);
         }
 
-        $relatedClassesNames = array_merge([$class->name], $class->subClasses, $class->parentClasses);
+        $relatedClassesNames = array_merge(array_reverse($class->parentClasses), [$class->name], $class->subClasses);
         $relatedClasses      = array_map([$this->em, 'getClassMetadata'], $relatedClassesNames);
 
         $sqlParts = [];
@@ -1747,9 +1747,6 @@ class SqlWalker implements TreeWalker
         foreach ($relatedClasses as $subClassMetadata) {
             $sqlParts[] = $this->getSQLTableAlias($subClassMetadata->getTableName(), $dqlAlias) . '.' . $this->quoteStrategy->getColumnName($fieldName, $subClassMetadata, $this->platform);
         }
-
-        // for more predictability of group by order
-        sort($sqlParts);
 
         return implode(', ', $sqlParts);
     }
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -35,6 +35,7 @@ use function is_numeric;
 use function is_string;
 use function preg_match;
 use function reset;
+use function sort;
 use function sprintf;
 use function strtolower;
 use function strtoupper;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -1725,26 +1726,54 @@ public function walkGroupByClause($groupByClause)
         return ' GROUP BY ' . implode(', ', $sqlParts);
     }
 
+    public function walkGroupByPathExpression(Ast\PathExpression $pathExpr): string
```
beberlei: This should maybe be private? Since its an implementation detail of SqlWalker. Otherwise we need a strategy to put it on the interface and other walkers in the future.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC2917Test.php
```diff
@@ -0,0 +1,124 @@
...
+use Throwable;
+use function strtolower;
+
+class DDC2917Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+        try {
+                $this->_schemaTool->createSchema(
```
beberlei: I dont believe we need the tables created to test just the DQL generated.

DQL generation tests are also often in a spefic testcase.

I believe this still needs to scope the twst for Pgsql or do we want to add tis behavior to all vendors?

beberlei (changes requested):
Looks like a valid fix, but willbe important to clearly document in changelog when and how SQL group by generation changes since this could affect index selection in customer apps.

lib/Doctrine/ORM/Query/SqlWalker.php
derrabus: Either way, I wouldn't put this on the interface. Have you seen #9551?

popov-a-e: Yes, my bad, this should of course be private

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC2917Test.php
popov-a-e: > I believe this still needs to scope the twst for Pgsql or do we want to add tis behavior to all vendors?

This works for all vendors, so I didn't scope the test as well. 
The code on the other hand can be scoped, but this requires either a specific strategy which would be a little bit an overkill, either something like `if ($this->platform instanceof PostgreSqlPlatform)` which seem idiosyncratic.
I thought that since this behavior doesn't break anything in other databases, let it be for all vendors.
What do you think?

popov-a-e: > DQL generation tests are also often in a specific testcase.

Can you explain what do you mean here? I've found only 3 abstract test cases to use.

derrabus:
Can you please change the PR title to something you would like to read in a change log? If I read "DDC-2917: resolve issue" there, I wouldn't know what the change does.

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -1725,26 +1726,54 @@ public function walkGroupByClause($groupByClause)
...
+        $relatedClasses      = array_map([$this->em, 'getClassMetadata'], $relatedClassesNames);
+
+        $sqlParts = [];
+
+        foreach ($relatedClasses as $subClassMetadata) {
+            $sqlParts[] = $this->getSQLTableAlias($subClassMetadata->getTableName(), $dqlAlias) . '.' . $this->quoteStrategy->getColumnName($fieldName, $subClassMetadata, $this->platform);
+        }
+
+        // for more predictability of group by order
+        sort($sqlParts);
```
derrabus: This sort call looks a bit arbitrary to me. Are we just doing this to make writing tests for this code easier?

popov-a-e: well, it's not really about tests, there is no problem in putting any order to them, I was concerned that the order of those is not easily predictable and later changes in, for example, joins order logic, might break the tests in the future

popov-a-e: No, i'm stupid, my latest solution relies on a predictable order so it should be fine without sorting

popov-a-e: Forget that I've told  
Will remove that :D

popov-a-e: Thanks for the comment, I replaced that with correct $relatedClasses order

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC2917Test.php
popov-a-e: @beberlei can we close this one?

popov-a-e:
Hi guys, what else is need to be done for this to be merged?
I work with Postgres & Doctrine at my startup and this bug doesn't let us use the inheritance feature, because we know we won't be able to use any aggregation on those models.

PR #9747, opened 2022-05-09
Deprecate omitting second argument to joinColumnName() 

greg0ire:
I don't think there is a sensible way to actually enforce that deprecation.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,6 @@
 # Upgrade to 2.13
 
-# Deprecated methods related to named queries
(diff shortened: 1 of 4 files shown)


derrabus (approved):

stof:
if you add the `@param` in the interface, the DebugClassLoader of Symfony will be able to trigger a deprecation at autoloading time for projects using it.

greg0ire:
OK, I'll give that another try! IIRC it will require ignoring some errors from SA tools

PR #9752, opened 2022-05-11
Document unofficial parameter in phpdoc

greg0ire:
This is a follow-up for https://github.com/doctrine/orm/pull/9747#issuecomment-1122645899

Pros:
- The DebugClassLoader of Symfony can emit a deprecation about that
- The phpdoc no longer needs to be repeated in implementing classes

Cons:
- PHP_CodeSniffer is unhappy about this.
- PHPStan is unhappy about this.
- Psalm does not take into account the inherited phpdoc.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/DefaultNamingStrategy.php
+++ b/lib/Doctrine/ORM/Mapping/DefaultNamingStrategy.php
@@ -54,9 +54,6 @@ class DefaultNamingStrategy implements NamingStrategy
 
     /**
      * {@inheritdoc}
-     *
-     * @param string       $propertyName
-     * @param class-string $className
      */
     public function joinColumnName($propertyName, $className = null)
     {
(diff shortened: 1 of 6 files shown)


greg0ire:
@stof this does not look too great: we have to add ignore rules for 3 of our tools, to trigger a deprecation for a tool that isn't necessarily used by our users.

SenseException:
It seems like `@param class-string $className The fully-qualified class name.` had to be added to the classes instead of the interface. Maybe using the psalm- or phpstan annotations than the phpdoc one.

greg0ire:
@SenseException the goal of this is to trigger a deprecation at autoload time for implementing classes outside `doctrine/orm`, so the interface phpdoc has to be modified. That being said, given the cons I listed, I'm not convinced this is a great idea, so let's just close this.

PR #9754, opened 2022-05-11
Update transactions-and-concurrency.rst

AlexBDev:
Fix uppercase typo on method rollback from code example in section Approach 2: Explicitly


derrabus:
Sorry, but the 2.7.x branch is not maintained anymore. Please submit this kind of changes against the 2.12.x branch.

PR #9755, opened 2022-05-11
Require new argument to joinColumnName()

greg0ire:


UPGRADE.md
```diff
@@ -1,5 +1,42 @@
...
+the interface.
+
+```php
+<?php
+class MyStrategy implements NamingStrategy
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function joinColumnName(string $propertyName, string $className = null): string
```
derrabus: `$propertyName` does not (yet) have a parameter type on the `NamingStrategy` interface. Adding the type here would cause a fatal error. I suggest to add the type to the interface and the implementations.

Moreover, the second argument is not supposed to be optional. Why make it optional in this example?

greg0ire: > Moreover, the second argument is not supposed to be optional. Why make it optional in this example?

I copy/pasted what I did for the deprecation and forgot to remove that.

greg0ire: Regarding the type declarations, I think I will address this for the whole classes and interface separately, after https://github.com/doctrine/orm/pull/9756 is merged up. Let's pause this for now.

SenseException: Yeah, having all the changes at one place for 3.0 will make it more clear about the types.

greg0ire: It's done now :)

lib/Doctrine/ORM/Mapping/UnderscoreNamingStrategy.php
```diff
@@ -84,7 +84,7 @@ public function referenceColumnName(): string
      *
      * @param class-string $className
```
derrabus: This annotation can be removed because it is now inherited from the interface.

lib/Doctrine/ORM/Mapping/DefaultNamingStrategy.php
```diff
@@ -49,7 +49,7 @@ public function referenceColumnName(): string
      *
      * @param class-string $className
```
derrabus: This annotation can be removed because it is now inherited from the interface.

UPGRADE.md
derrabus:
```suggestion
    public function joinColumnName(string $propertyName, string $className): string
```

derrabus (changes requested):

derrabus (approved):

PR #9758, opened 2022-05-12
Use native type declaration for naming strategies

greg0ire:


lib/Doctrine/ORM/Mapping/NamingStrategy.php
```diff
@@ -15,64 +15,46 @@ interface NamingStrategy
...
 
     /**
      * Returns a join column name for a property.
-     *
-     * @param string $propertyName A property name.
-     *
-     * @return string A join column name.
      */
-    public function joinColumnName($propertyName/*, $className = null */);
+    public function joinColumnName(string $propertyName/*, $className = null */): string;
```
SenseException, line 49: Isn't $className supposed to  be added in 3.0?

docs/en/reference/namingstrategy.rst
```diff
@@ -110,28 +110,28 @@ You need to create a class which implements ``Doctrine\ORM\Mapping\NamingStrateg
...
         {
             return $propertyName;
         }
-        public function referenceColumnName()
+        public function referenceColumnName(): string
         {
             return 'id';
         }
-        public function joinColumnName($propertyName, $className = null)
+        public function joinColumnName(string $propertyName, string $className): string
```
SenseException:
```suggestion
        public function joinColumnName(string $propertyName, ?string $className = null): string
```

lib/Doctrine/ORM/Mapping/NamingStrategy.php
greg0ire: Yes, but I was gonna do that in a separate PR (#9755)

derrabus (approved):

PR #9760, opened 2022-05-12
[GH-3519] Deprecate passing the same class with different discriminator values.

beberlei:
Prepares the deprecation for the later removal of this mapping inconsistency.

Related: #3519

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,8 @@
+# Upgrade to 3.4
+
+Using the same class several times in a discriminator map is deprecated.
+In 4.0, this will be an error.
+
 # Upgrade to 3.3
 
 ## Deprecate `DatabaseDriver`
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -3225,6 +3225,14 @@ final public function getDiscriminatorColumn(): array
      */
     public function setDiscriminatorMap(array $map)
     {
+        if (count(array_flip($map)) !== count($map)) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/issues/3519',
+                'Mapping a class to multiple discriminator values is deprecated.'
```
greg0ire: Let's add the class that has multiple mappings and which those are with `$this->name` and `array_keys(array_filter(array_count_values($map), function ($value) { return $value > 1; }))`?

derrabus:
LGTM, once @greg0ire's suggestion is applied.

mpdude:
@beberlei This might be an easy thing to bring to an end, and it would add one more mapping consistency check to the codebase.

stof:
shouldn't this target the 3.4.x branch instead, as this is about a deprecation for a removal in 4.0 ?

stof:
And maybe the schema validator should already report an error for such case

greg0ire:
@stof you're probably right about the target branch, this was created in 2022, and 3.0 did not exist back then. Regarding the schema validator, maybe the initial intent was to contribute the validation in the next major?

stof:
AFAICT, the intent seems to be to turn the deprecation into a MappingException in the next major (which is stricter than something reported only by the SchemaValidator). But maybe the deprecation is indeed enough in that case, instead of turning it into an error reported only to people validating their schema.

SenseException (approved):
> And maybe the schema validator should already report an error for such case

> But maybe the deprecation is indeed enough in that case, instead of turning it into an error reported only to people validating their schema.

I assume the schema validator error report is off the table. (sql-pun intended)

PR #9761, opened 2022-05-14
Fix phpdoc and tests for NamingStrategy

greg0ire:
While working on https://github.com/doctrine/orm/pull/9758, I noticed 2 things that need to be fixed.

When computing a foreign key column name, the referenced column name
may be null in the case of a self referencing entity with join columns
defined in the mapping. I wrongly introduced that phpdoc in #9756.
Also, some tests were using the fact that some arguments of some methods of
the naming strategy interface are optional or nullable for now to avoid
providing some. In practice, these arguments are always provided, and
that should also be the case in tests.

For `propertyToColumnName`, I think `ClassMetadataInfo::$name` is always a non-nullable string
See https://github.com/doctrine/orm/blob/45e196eb57034a90d0e4a7afb9109d9537132c24/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L1635

For `joinTableName()` I think related mapping are always defined based on some property and cannot be null either. 
https://github.com/doctrine/orm/blob/45e196eb57034a90d0e4a7afb9109d9537132c24/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L2050

### DIFF
--- a/lib/Doctrine/ORM/Mapping/NamingStrategy.php
+++ b/lib/Doctrine/ORM/Mapping/NamingStrategy.php
@@ -78,7 +78,10 @@ interface NamingStrategy
      * Returns the foreign key column name for the given parameters.
      *
      * @param class-string $entityName           An entity.
-     * @param string       $referencedColumnName A property.
+     * @param string|null  $referencedColumnName A property name or null in
+     *                                           case of a self-referencing
+     *                                           entity with join columns
+     *                                           defined in the mapping
      *
      * @return string A join column name.
      */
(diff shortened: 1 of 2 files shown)


SenseException (approved):
> I wrongly introduced that phpdoc in https://github.com/doctrine/orm/pull/9756.

I didn't realized that either.

greg0ire:
> I didn't realized that either.

Yeah I think it's hard to realize until you actually conduct the actual breaking change on the next major branch :slightly_smiling_face:

PR #9762, opened 2022-05-16
Do not call AbstractSchemaManager::getDatabasePlatform()

morozov:
`AbstractSchemaManager::getDatabasePlatform()` was deprecated in https://github.com/doctrine/dbal/pull/5387.

The reason why the foreign keys are used conditionally is really unclear. The conditional logic was originally introduced in https://github.com/doctrine/orm/commit/556f8699ee5e622f0d2375b209cd2e59b3390b55 as a replacement for `try`/`catch`. FWIW, all supported platforms support foreign keys, although the SQLite platform declares that it doesn't.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
@@ -6,7 +6,6 @@ namespace Doctrine\ORM\Mapping\Driver;
 
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 use Doctrine\DBAL\Schema\Column;
-use Doctrine\DBAL\Schema\ForeignKeyConstraint;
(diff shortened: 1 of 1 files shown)


beberlei:
The reason is that we added working sqlite foreign key support in platforms much later than DBAL and ORM 2.0 releaSes, so at some point sqlite platform didnt support fks. Changing the default supports flag would have been a BC break at that point.

I remember working on a DBAl PR just 1-2 years ago regarding this which we also reverted or didnt merge due to bc breaks.

morozov:
It is clear that some platforms may not support foreign keys according to the API, but in this case, `AbstractSchemaManager::listTableDetails()` will populate the table object with an empty list of foreign keys.

Do you believe this condition is still necessary?

beberlei (approved):
No, it should work this way from the looks of it.

PR #9764, opened 2022-05-16
Merge 2.13.x up into 3.0.x

greg0ire:


greg0ire:
@morozov the tests fails in `tests/Doctrine/Tests/ORM/Functional/SequenceEmulatedIdentityStrategyTest.php`, in `tearDown`, where a sequence is being removed manually. 

https://github.com/doctrine/orm/blob/60f59b29edaf4a494453a383760693e6ee37ba24/tests/Doctrine/Tests/ORM/Functional/SequenceEmulatedIdentityStrategyTest.php#L38-L43

The comment refers to a dependency, and so does the error message:

```
PDOException: SQLSTATE[2BP01]: Dependent objects still exist: 7 ERROR:  cannot drop sequence seq_identity_id_seq because column id of table seq_identity requires it
HINT:  You can drop column id of table seq_identity instead.
```

I'm going to drop `tearDown` entirely and let `OrmFunctionalTestCase::createSchemaForModels()` do its thing instead. I don't really understand how this relates to #9762 though.

morozov:
> I don't really understand how this relates to #9762 though.

I don't think it's related to #9762. It's more likely related to https://github.com/doctrine/dbal/pull/5396.

morozov:
> I'm going to drop `tearDown` entirely and let `OrmFunctionalTestCase::createSchemaForModels()` do its thing instead.

The underlying sequence is now managed by Postgres, so there shouldn't be any additional handling in the ORM necessary.

greg0ire:
I updated my commit message 👍

morozov:
> I updated my commit message 👍

Should this change have been part of the merge commit? If done after it, it leaves a knowingly broken state behind and it couldn't have been done prior to the merge either.

greg0ire:
> it couldn't have been done prior to the merge either.

Why not? I think it could be cherry-picked on 3.0.x prior to the merge, couldn't it? The issue does not trigger because of the merge up, right?

morozov:
> I think it could be cherry-picked on 3.0.x prior to the merge, couldn't it?

You're actually right. Given that the dependency of dev ORM on dev DBAL isn't locked, this code could have been updated prior to the merge. If it was locked, it should have been part of the update of the lock.

morozov (approved):

PR #9765, opened 2022-05-17
Deprecate omitting the alias in QueryBuilder::update() and delete()

### LINKED ISSUE #9733: QueryBuilder Update Query doesn't work if no alias is used.
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 2.12.2

#### Summary

If no alias is used, the Query Builder seems to have problems to build the query, or the lexer that is checking the query thinks it is invalid when it is not.

#### Current behavior

Currently it is crashing with an Exception, see below.

#### How to reproduce

having this QueryBuilder:
```php
$qb = $em->createQueryBuilder();
$qb->update(User::class);
$qb->set('klassenstufe', 'klassenstufe+1');
$qb->where($qb->expr()->in('id', $users));
$qb->getQuery()->execute();
```
makes this Query:
`UPDATE App\Entity\User\User SET klassenstufe = klassenstufe+1 WHERE id IN(188)`

and causes this crash:
```
[Syntax Error] line 0, col 33: Error: Expected Doctrine\ORM\Query\Lexer::T_SET, got 'klassenstufe'
  at vendor/doctrine/orm/lib/Doctrine/ORM/Query/QueryException.php:32
  at Doctrine\ORM\Query\QueryException::syntaxError('line 0, col 33: Error: Expected Doctrine\\ORM\\Query\\Lexer::T_SET, got \'klassenstufe\'', object(QueryException))
     (vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:509)
  at Doctrine\ORM\Query\Parser->syntaxError('Doctrine\\ORM\\Query\\Lexer::T_SET')
     (vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:368)
  at Doctrine\ORM\Query\Parser->match(247)
     (vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:1315)
  at Doctrine\ORM\Query\Parser->UpdateClause()
     (vendor/doctrine/orm/lib/Doctrine/ORM/Query/Parser.php:946)
  at Doctrine\ORM\Query\Parser->UpdateStatement()
 
```

But having this QueryBuilder:
```php
$qb = $em->createQueryBuilder();
$qb->update(User::class, 'u');
$qb->set('u.klassenstufe', 'u.klassenstufe+1');
$qb->where($qb->expr()->in('u.id', $users));
$qb->getQuery()->execute();
```
makes this Query:
`UPDATE App\Entity\User\User u SET u.klassenstufe = u.klassenstufe+1 WHERE u.id IN(188)`

Causing it to work.

IF the alias is needed, then it shouldn't be an optional parameter for the update function.
Or it should be better documented.

i don't know if that is a BC or not, but i think it was working before.

#### Expected behavior

The Query to be executed.

Hanmac:
Closes #9733

Should the Trigger warning has the link to the Issue or to the PR?


lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -844,6 +845,14 @@ public function delete($delete = null, $alias = null)
             return $this;
         }
 
+        if (! $alias) [
```
derrabus:
```suggestion
        if (! $alias) {
```

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -871,6 +880,14 @@ public function update($update = null, $alias = null)
...
```
derrabus:
```suggestion
        if (! $alias) {
```

derrabus:
> Should the Trigger warning has the link to the Issue or to the PR?

The issue is just fine. Can you please add a small test for each method that checks if the exception is being raised?

derrabus:
Also, please run `./vendor/bin/phpcbf` and commit the result.

UPGRADE.md
```diff
@@ -1,5 +1,29 @@
 # Upgrade to 2.13
 
+## Deprecated omitting alias argument for `QueryBuilder::update` and `QueryBuilder::delete`
```
derrabus:
```suggestion
## Deprecated omitting only the alias argument for `QueryBuilder::update` and `QueryBuilder::delete`
```

UPGRADE.md
```diff
@@ -1,5 +1,29 @@
...
+
+When making an Update or Delete Query and when passing a class/type to the function, the alias argument should not be omitted.
```
derrabus:
```suggestion
When making an Update or Delete Query and when passing a class/type to the function, the alias argument must not be omitted.
```

UPGRADE.md
```diff
@@ -1,5 +1,29 @@
...
+$qb = $em->createQueryBuilder()
+    ->delete('User u')
+    ->where('u.id = :user_id')
+    ->setParameter('user_id', 1);
+```
+
+### After
+
+For backward-compatibility reasons, the parameter has to be optional and to make it possible to call `update()` and `delete()` without class/type argument.
+
```
derrabus:
```suggestion
```

I don't think that we need to explain that here.

Hanmac:
> Also, please run `./vendor/bin/phpcbf` and commit the result.

i'm only edited in github, i will try it later

any special php version under which i should do that?



> > Should the Trigger warning has the link to the Issue or to the PR?
> 
> The issue is just fine. Can you please add a small test for each method that checks if the exception is being raised?

currently there isn't any exception but trigger warning, but i will check on how such test would be done

Hanmac:
i will update the method description from update and delete later too so it is more clear that the two wanted use cased are:
* with no parameters
* with two parameters

derrabus:
> currently there isn't any exception but trigger warning, but i will check on how such test would be done

There is a `VerifyDeprecations` trait that gives you some tooling for that. Just look into other tests that use this trait.

derrabus:
The failing tests look related. Did you execute the tests locally?

UPGRADE.md
```diff
@@ -1,5 +1,27 @@
 # Upgrade to 2.13
 
+## Deprecated omitting only the alias argument for `QueryBuilder::update` and `QueryBuilder::delete`
+
+When making an Update or Delete Query and when passing a class/type to the function, the alias argument must not be omitted.
```
derrabus:
```suggestion
When building an UPDATE or DELETE query and when passing a class/type to the function, the alias argument must not be omitted.
```

derrabus (commented):
Looks good to me. Thank you for your PR.

The PR is still flagged as draft. Is there anything left to do from your side?

Hanmac:
> Looks good to me. Thank you for your PR.
> 
> The PR is still flagged as draft. Is there anything left to do from your side?

i was thinking about adding the info into the method description telling the user that they should only do either zero or two arguments. But i might not be good with the wording :p

does  the method description needs extra info that it might raise a deprecation or later an exception in the future?

derrabus:
> > Looks good to me. Thank you for your PR.
> > The PR is still flagged as draft. Is there anything left to do from your side?
> 
> i was thinking about adding the info into the method description telling the user that they should only do either zero or two arguments.

Feel free to do so, but I don't think it's necessary. Write the documentation that would've helped you. Don't worry too much about the wording.

> does the method description needs extra info that it might raise a deprecation or later an exception in the future?

No. I don't think we need to document the deprecation explicitly here. If you add extra documentation, just say that either none or both parameters have to be passed.

derrabus:
Closed via b931a59ebcdb8e79f78bbfd7fdd4e8979e9b8683. Thanks @Hanmac!

PR #9766, opened 2022-05-17
Document future argument better

greg0ire:
That argument is always provided, so the tests should provide it and the
commented out argument should reflect the future.

This should have been done as part of #9761 but I confused this method with another one and overlooked it.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/NamingStrategy.php
+++ b/lib/Doctrine/ORM/Mapping/NamingStrategy.php
@@ -61,7 +61,7 @@ interface NamingStrategy
      *
      * @return string A join column name.
      */
-    public function joinColumnName($propertyName/*, $className = null */);
+    public function joinColumnName($propertyName/*, string $className */);
 
     /**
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Mapping/NamingStrategy.php
```diff
@@ -61,7 +61,7 @@ public function referenceColumnName();
      *
      * @return string A join column name.
      */
-    public function joinColumnName($propertyName/*, $className = null */);
+    public function joinColumnName($propertyName/*, string $className */);
```
SenseException, line 64: Not that it matters in a comment, but 3.0 has a nullable type.

SenseException (approved):

lib/Doctrine/ORM/Mapping/NamingStrategy.php
greg0ire: Not after #9755 it won't :wink:

PR #9770, opened 2022-05-20
Add primary key on temp table

### LINKED ISSUE #9768: sql_require_primary_key on mysql 8
### Bug Report
|    Q        |   A
|------------ | ------
| Version     | 2.12.2

#### Summary

We are running a managed database on DigitalOcean.
We are using a class table inheritance. If we now try to update one entity from this inheritance the `Doctrine\ORM\Query\Exec\MultiTableUpdateExecutor::class` is  called and creates a temporary table but the mysql option **sql_require_primary_key** is default **true**. So we are getting an SQL Exception that no primary key exists on this table.

We work around this issue setting **sql_require_primary_key** to **false** but the recommendation from DigitalOcean is:
*"For clusters with more than one node, we strongly recommend keeping sql_require_primary_key as True. Setting it to False may severely impact replication and cause issues."*

#### Current behavior
With **sql_require_primary_key** set to **true** we are getting a SQL Exception.

#### How to reproduce

Use a mysql8 Database, create an entity with  class table inheritance and try update an entity.

#### Expected behavior

Entity update works with **sql_require_primary_key** set to **true**.

wiseguy1394:
… to prevent exception on sql_require_primary_key=ON

Fixes #9768

### DIFF
--- a/lib/Doctrine/ORM/Query/Exec/MultiTableDeleteExecutor.php
+++ b/lib/Doctrine/ORM/Query/Exec/MultiTableDeleteExecutor.php
@@ -98,7 +98,7 @@ class MultiTableDeleteExecutor extends AbstractSqlExecutor
         }
 
         $this->_createTempTableSql = $platform->getCreateTemporaryTableSnippetSQL() . ' ' . $tempTable . ' ('
-                . $platform->getColumnDeclarationListSQL($columnDefinitions) . ')';
+                . $platform->getColumnDeclarationListSQL($columnDefinitions) . ', PRIMARY KEY(' . implode(',', $idColumnNames) . '))';
         $this->_dropTempTableSql   = $platform->getDropTemporaryTableSQL($tempTable);
     }
 
--- a/lib/Doctrine/ORM/Query/Exec/MultiTableUpdateExecutor.php
+++ b/lib/Doctrine/ORM/Query/Exec/MultiTableUpdateExecutor.php
@@ -139,7 +139,7 @@ class MultiTableUpdateExecutor extends AbstractSqlExecutor
         }
 
         $this->_createTempTableSql = $platform->getCreateTemporaryTableSnippetSQL() . ' ' . $tempTable . ' ('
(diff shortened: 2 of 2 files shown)


derrabus:
This PR adds the PK no matter what database engine we're using, right? Given that we use that temp table in an `IN()` statement later on, it might be a good idea to have the PK anyway, even if the `sql_require_primary_key` restriction is not in place.

Opinions on this, @greg0ire @beberlei?

I think we should enable `sql_require_primary_key` for MySQL 8 in our CI to make sure we don't introduce further regressions with that setting.

@wiseguy1394 Can you please run `./vendor/bin/phpcbf` and commit the result?

greg0ire:
I don't have much to say on this, it seems fine, and I can't see how it would hurt.

greg0ire:
Please associate your email address with your Github account, or change the

wiseguy1394:
Ah ok.. sry..I tried to changed the author for my commit, hope it worked.

greg0ire:
It resulted in 3 commits, because I think you reset the author, then tried to push and noticed my commit, and you merged them both. I put it back to 1 commit. 

Please try again, but after doing: `git fetch --all && git reset --hard e06e215`. This will effectively use the commit I just pushed as a starting point.

beberlei (changes requested):
Wouldn't this need all columns of the primary key of the entity instead just the first one? Otherwise it will lead to primary key duplicates for entities with multiple columns.

Does the `MultiTableDeleteExecutor` require the same change?

This will probably increase performance of these commands.

wiseguy1394:
Is there anything I must/could do?

derrabus:
> Is there anything I must/could do?

You could answer @beberlei's questions. 😉

wiseguy1394:
> > Is there anything I must/could do?
> 
> You could answer @beberlei's questions. 😉

Oh ok, I answered the questions with yes in my commits ;)

derrabus (approved):

beberlei (approved):
Thank you!

PR #9774, opened 2022-05-21
Fix wrong types for AbstractQuery and child classes

greg0ire:
Refs #9772

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -36,6 +36,7 @@ use function is_array;
 use function is_numeric;
 use function is_object;
 use function is_scalar;
+use function is_string;
 use function iterator_count;
 use function iterator_to_array;
 use function ksort;
@@ -282,7 +283,7 @@ abstract class AbstractQuery
      * The returned SQL syntax depends on the connection driver that is used
      * by this query object at the time of this method call.
      *
-     * @return string SQL query
+     * @return list<string>|string SQL query
      */
     abstract public function getSQL();
 
@@ -400,8 +401,8 @@ abstract class AbstractQuery
      *
      * @param mixed $value
      *
-     * @return mixed[]|string|int|float|bool
-     * @psalm-return array|scalar
+     * @return mixed[]|string|int|float|bool|object|null
+     * @psalm-return array|scalar|object|null
      *
      * @throws ORMInvalidArgumentException
      */
@@ -1301,7 +1302,8 @@ abstract class AbstractQuery
             $parameters[$parameter->getName()] = $this->processParameterValue($parameter->getValue());
         }
 
-        $sql                    = $this->getSQL();
+        $sql = $this->getSQL();
+        assert(is_string($sql));
         $queryCacheProfile      = $this->getHydrationCacheProfile();
         $hints                  = $this->getHints();
         $hints['hydrationMode'] = $this->getHydrationMode();
@@ -1373,7 +1375,8 @@ abstract class AbstractQuery
      */
     protected function getHash()
     {
-        $query  = $this->getSQL();
+        $query = $this->getSQL();
+        assert(is_string($query));
         $hints  = $this->getHints();
         $params = array_map(function (Parameter $parameter) {
             $value = $parameter->getValue();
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -401,8 +401,8 @@ public function setParameter($key, $value, $type = null)
      *
      * @param mixed $value
      *
-     * @return mixed[]|string|int|float|bool
-     * @psalm-return array|scalar
+     * @return mixed[]|string|int|float|bool|object|null
+     * @psalm-return array|scalar|object|null
```
greg0ire, line 405: I've seen `DateTime`, `CustomIdObject` and `null` pass here when running unit tests with a native type declaration.

lib/Doctrine/ORM/NativeQuery.php
```diff
@@ -34,7 +34,7 @@ public function setSQL($sql): self
     /**
      * Gets the SQL query.
      *
-     * @return mixed The built SQL query or an array of all SQL queries.
+     * @return string The built SQL query
```
greg0ire: The class is final, and I don't see how we could have anything else than a string here.

psalm-baseline.xml
```diff
@@ -1970,6 +1964,9 @@
     <PropertyNotSetInConstructor occurrences="1">
       <code>MultiTableUpdateExecutor</code>
     </PropertyNotSetInConstructor>
+    <PropertyTypeCoercion occurrences="1">
+      <code>$this-&gt;_sqlStatements</code>
+    </PropertyTypeCoercion>
```
greg0ire, line 1969: The code is hard to analyse statically, but from my human point of view, what is intended here is a `list`

https://github.com/doctrine/orm/blob/8f7701279de84411020304f668cc8b49a5afbf5a/lib/Doctrine/ORM/Query/Exec/MultiTableUpdateExecutor.php#L88-L124

SenseException: Agree. `mixed[]` doesn't fit for _sqlStatements.

SenseException (commented):
The issue references 3.0.0. Is `2.12.x` the expected target branch?

greg0ire:
> The issue references 3.0.0. Is 2.12.x the expected target branch?

This PR is linked to 2 other PRs :)

lib/Doctrine/ORM/AbstractQuery.php
derrabus: That's basically `mixed`, isn't it?

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -282,7 +283,7 @@ public function setCacheMode($cacheMode)
      * The returned SQL syntax depends on the connection driver that is used
      * by this query object at the time of this method call.
      *
-     * @return string SQL query
+     * @return list<string>|string SQL query
```
derrabus, line 286: For 3.0, we should try to replace this method with one that always returns an array of strings.

lib/Doctrine/ORM/NativeQuery.php
derrabus: If you're sure about that, you can add a native return type. Doing that on a `final` class is not a BC break.

lib/Doctrine/ORM/Query.php
```diff
@@ -596,7 +596,7 @@ public function free(): void
     /**
      * Sets a DQL query string.
      *
-     * @param string $dqlQuery DQL Query.
+     * @param string|null $dqlQuery DQL Query.
```
derrabus, line 599: This method is a no-op when `null` is passed. Should we rather deprecate passing `null` so that we can use `string` as native parameter type in 3.0?

greg0ire: > Should we rather deprecate passing null so that we can use string as native parameter type in 3.0?

Yes, but not rather. We should do so on 2.13.x.

derrabus (approved):

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: Oh crap didn't see your answer… well I guess you can't have `resource` in there but yeah, mixed might be better here :sweat_smile:

lib/Doctrine/ORM/AbstractQuery.php
morrisonlevi: Doctrine noob here: I assume a single string is the SQL query, probably with placeholders and such, but what is an array of strings? And why would we prefer this for Doctrine 3?

PR #9776, opened 2022-05-22
Migrate to PHP 8: AbstractQuery and child classes

greg0ire:
Blocked by 
- https://github.com/doctrine/orm/pull/9777
- https://github.com/doctrine/orm/pull/9809

TODO:
- [x] merge the above
- [x] merge up the patch branch
- [x] merge up the minor branch
- [x] run rector

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -10,7 +10,6 @@ use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\Result;
-use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Cache\Logging\CacheLogger;
 use Doctrine\ORM\Cache\QueryCacheKey;
 use Doctrine\ORM\Cache\TimestampCacheKey;
@@ -28,7 +27,6 @@ use function array_map;
 use function array_shift;
 use function assert;
 use function count;
-use function in_array;
 use function is_array;
 use function is_countable;
 use function is_numeric;
@@ -655,16 +653,6 @@ abstract class AbstractQuery
      */
     public function setFetchMode(string $class, string $assocName, int $fetchMode): static
     {
-        if (! in_array($fetchMode, [ClassMetadata::FETCH_EAGER, ClassMetadata::FETCH_LAZY], true)) {
-            Deprecation::trigger(
-                'doctrine/orm',
-                'https://github.com/doctrine/orm/pull/9777',
-                'Calling %s() with something else than ClassMetadata::FETCH_EAGER or ClassMetadata::FETCH_LAZY is deprecated.',
-                __METHOD__
-            );
-            $fetchMode = ClassMetadata::FETCH_LAZY;
-        }
-
         $this->_hints['fetchMode'][$class][$assocName] = $fetchMode;
 
         return $this;
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -28,12 +28,10 @@ use function array_shift;
 use function assert;
 use function count;
 use function is_array;
-use function is_countable;
 use function is_numeric;
 use function is_object;
 use function is_scalar;
 use function is_string;
-use function iterator_count;
 use function iterator_to_array;
 use function ksort;
 use function reset;
@@ -861,10 +859,7 @@ abstract class AbstractQuery
             $this->setHydrationMode($hydrationMode);
         }
 
-        if (
-            (is_countable($parameters) && count($parameters) !== 0)
-            || ($parameters instanceof Traversable && iterator_count($parameters) !== 0)
-        ) {
+        if (count($parameters) !== 0) {
             $this->setParameters($parameters);
         }
 
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Doctrine\ORM;
 
 use BackedEnum;
-use Countable;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Util\ClassUtils;
@@ -15,6 +14,7 @@ use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Cache\Logging\CacheLogger;
 use Doctrine\ORM\Cache\QueryCacheKey;
 use Doctrine\ORM\Cache\TimestampCacheKey;
+use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\MappingException as ORMMappingException;
 use Doctrine\ORM\Query\Parameter;
 use Doctrine\ORM\Query\QueryException;
@@ -30,6 +30,7 @@ use function assert;
 use function count;
 use function in_array;
 use function is_array;
+use function is_countable;
 use function is_numeric;
 use function is_object;
 use function is_scalar;
@@ -94,11 +95,6 @@ abstract class AbstractQuery
     protected ?ResultSetMapping $_resultSetMapping = null;
 
     /**
-     * The entity manager used by this query object.
-     */
-    protected EntityManagerInterface $_em;
-
-    /**
      * The map of query hints.
      *
      * @psalm-var array<string, mixed>
@@ -147,12 +143,15 @@ abstract class AbstractQuery
     /**
      * Initializes a new instance of a class derived from <tt>AbstractQuery</tt>.
      */
-    public function __construct(EntityManagerInterface $em)
-    {
-        $this->_em        = $em;
+    public function __construct(
+        /**
+         * The entity manager used by this query object.
+         */
+        protected EntityManagerInterface $em
+    ) {
         $this->parameters = new ArrayCollection();
         $this->_hints     = $em->getConfiguration()->getDefaultQueryHints();
-        $this->hasCache   = $this->_em->getConfiguration()->isSecondLevelCacheEnabled();
+        $this->hasCache   = $this->em->getConfiguration()->isSecondLevelCacheEnabled();
 
         if ($this->hasCache) {
             $this->cacheLogger = $em->getConfiguration()
@@ -260,7 +259,7 @@ abstract class AbstractQuery
      */
     public function getEntityManager(): EntityManagerInterface
     {
-        return $this->_em;
+        return $this->em;
     }
 
     /**
@@ -272,7 +271,7 @@ abstract class AbstractQuery
     {
         $this->parameters = new ArrayCollection();
 
-        $this->_hints = $this->_em->getConfiguration()->getDefaultQueryHints();
+        $this->_hints = $this->em->getConfiguration()->getDefaultQueryHints();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -214,59 +193,52 @@ public function setCacheRegion($cacheRegion)
...
      * Sets the life-time for this query into second level cache.
      *
-     * @param int $lifetime
-     *
      * @return $this
      */
-    public function setLifetime($lifetime)
+    public function setLifetime(int $lifetime): static
     {
         $this->lifetime = (int) $lifetime;
```
derrabus:
```suggestion
        $this->lifetime = $lifetime;
```

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -310,7 +278,7 @@ public function free()
      * @return ArrayCollection The defined query parameters.
```
derrabus:
```suggestion
```

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -322,7 +290,7 @@ public function getParameters()
      *
      * @return Parameter|null The value of the bound parameter, or NULL if not available.
      */
-    public function getParameter($key)
+    public function getParameter($key): ?Parameter
```
derrabus:
```suggestion
    public function getParameter(mixed $key): ?Parameter
```

At least that's what the doc block says. Can we narrow that down to `int|string`?

greg0ire: That would make sense 👍

greg0ire: Incorporated that in #9777

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -217,59 +196,52 @@ public function setCacheRegion($cacheRegion)
...
      * @return string|null The cache region name; NULL indicates the default region.
      */
-    public function getCacheRegion()
+    public function getCacheRegion(): ?string
     {
         return $this->cacheRegion;
     }
 
     /**
      * @return bool TRUE if the query cache and second level cache are enabled, FALSE otherwise.
```
derrabus, line 200:
```suggestion
     * Checks if the query cache and second level cache are enabled.
```

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -217,59 +196,52 @@ public function setCacheRegion($cacheRegion)
      *
      * @return string|null The cache region name; NULL indicates the default region.
```
derrabus, line 192:
```suggestion
     * Returns the cache region name or NULL to indicate the default region.
```

lib/Doctrine/ORM/AbstractQuery.php
derrabus:
```suggestion
    public function getParameter(int|string $key): ?Parameter
```

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -348,7 +315,7 @@ static function (Query\Parameter $parameter) use ($key): bool {
      *
      * @return $this
      */
-    public function setParameters($parameters)
+    public function setParameters(ArrayCollection|array $parameters): static
```
derrabus, line 309: We should try to narrow this down in 2.13.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -395,14 +362,12 @@ public function setParameter($key, $value, $type = null)
     /**
      * Processes an individual parameter value.
      *
-     * @param mixed $value
-     *
      * @return mixed[]|string|int|float|bool|object|null
      * @psalm-return array|scalar|object|null
```
derrabus: Do these annotations provide any real value? They're hardly more specifiy than `mixed`.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -576,7 +532,7 @@ public function getHydrationCacheProfile()
      *
      * @return $this
      */
-    public function setResultCacheProfile(?QueryCacheProfile $profile = null)
+    public function setResultCacheProfile(?QueryCacheProfile $profile = null): static
```
derrabus: Can we remove the default here? That looks like a PHP5 hack to me.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -598,10 +554,8 @@ public function setResultCacheProfile(?QueryCacheProfile $profile = null)
 
     /**
      * Defines a cache driver to be used for caching result sets and implicitly enables caching.
-     *
-     * @return $this
      */
-    public function setResultCache(?CacheItemPoolInterface $resultCache = null)
+    public function setResultCache(?CacheItemPoolInterface $resultCache = null): static
```
derrabus: Same here.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -949,14 +870,12 @@ public function getHints()
      * Executes the query and returns an iterable that can be used to incrementally
      * iterate over the result.
      *
-     * @param ArrayCollection|array|mixed[] $parameters    The query parameters.
-     * @param string|int|null               $hydrationMode The hydration mode to use.
      * @psalm-param ArrayCollection<int, Parameter>|mixed[] $parameters
```
derrabus, line 849: It's a bit odd that we use `iterable` as native type, but a more specific `ArrayCollection` as annotated type. I think, we should fix the annotation.

derrabus (changes requested):

greg0ire:
@derrabus I'm AFK until Sunday, feel free to push to my branch, I agree with all of your comments.

lib/Doctrine/ORM/AbstractQuery.php
derrabus: I'm paving the way with #9791.

derrabus: #9798

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: :thinking: what do you mean? Drop the `ArrayCollection|` part? I know I have never used it that way.

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: I'll try widening to `mixed` on 2.13.x, see if that impacts static analysis.

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: I agree it's a problem, but I don't think fixing the annotation is the right thing to do: down the line, we call `setParameters`, which can deal with `mixed[]`, but not with `ArrayCollection<stdClass>`. It needs to be `ArrayCollection<Parameter>`.

SenseException (dismissed):

lib/Doctrine/ORM/AbstractQuery.php
derrabus: You're right. But then, we should have the same native parameter type as `setParameters()`, so `iterable` should be `ArrayCollection|array`.

lib/Doctrine/ORM/Query.php
```diff
@@ -479,16 +451,12 @@ public function useQueryCache($bool): self
     /**
      * Defines how long the query cache will be active before expire.
      *
-     * @param int $timeToLive How long the cache entry is valid.
+     * @param int|null $timeToLive How long the cache entry is valid.
```
derrabus: Should be fixed on 2.13.x as well.

lib/Doctrine/ORM/Query.php
```diff
@@ -509,7 +477,7 @@ public function getQueryCacheLifetime(): ?int
      *
      * @return $this
      */
-    public function expireQueryCache($expire = true): self
+    public function expireQueryCache(bool $expire = true): self
```
derrabus, line 479: We can delete the corresponding `@param`, I think.

lib/Doctrine/ORM/Query.php
```diff
@@ -537,7 +505,7 @@ public function free(): void
      *
      * @param string|null $dqlQuery DQL Query.
```
```suggestion
```

lib/Doctrine/ORM/Query.php
```diff
@@ -597,7 +565,7 @@ public function contains($dql): bool
      *
      * @return $this
      */
-    public function setFirstResult($firstResult): self
+    public function setFirstResult(?int $firstResult): self
```
derrabus, line 565: Shall we disallow `null`, to be consistent with the DBAL?

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: Yes, definitely, I can change that in this PR 👍

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: https://github.com/doctrine/orm/pull/9801

lib/Doctrine/ORM/Query.php
greg0ire: "fixed"? 2.12.x then.

lib/Doctrine/ORM/Query.php
greg0ire: Yes, that would make sense. That will need to be initiated on 2.13.x

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -712,24 +654,10 @@ public function getQueryCacheProfile()
...
-        if (! in_array($fetchMode, [Mapping\ClassMetadata::FETCH_EAGER, Mapping\ClassMetadata::FETCH_LAZY], true)) {
-            Deprecation::trigger(
-                'doctrine/orm',
-                'https://github.com/doctrine/orm/pull/9777',
-                'Calling %s() with something else than ClassMetadata::FETCH_EAGER or ClassMetadata::FETCH_LAZY is deprecated.',
-                __METHOD__
-            );
-            $fetchMode = Mapping\ClassMetadata::FETCH_LAZY;
-        }
-
```
greg0ire, line 731: From https://github.com/doctrine/orm/pull/9777#discussion_r879878197 : 

> On 3.0, we would throw if the caller passed anything but the two allowed values, right?

Do we really want to do this? If yes, maybe we should require an assertion library :thinking:

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -86,93 +83,74 @@ abstract class AbstractQuery
...
      * Initializes a new instance of a class derived from <tt>AbstractQuery</tt>.
      */
-    public function __construct(EntityManagerInterface $em)
-    {
-        $this->_em        = $em;
+    public function __construct(
+        /**
+         * The entity manager used by this query object.
+         */
+        protected EntityManagerInterface $_em
```
derrabus: I'd rather rename the property to `em` than adding that `_` prefix to the constructor parameter, tbh.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -325,12 +287,12 @@ public function getParameters()
...
      */
-    public function getParameter($key)
+    public function getParameter(int|string $key): ?Parameter
     {
-        $key = Query\Parameter::normalizeName($key);
+        $key = Parameter::normalizeName($key);
 
         $filteredParameters = $this->parameters->filter(
-            static function (Query\Parameter $parameter) use ($key): bool {
+            static function (Parameter $parameter) use ($key): bool {
```
derrabus:
```php
$filteredParameters = $this->parameters->filter(
    static fn (Parameter $parameter): bool => $parameter->getName() === $key
);
```

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: Me too, that was me not questioning Rector's work.

derrabus (dismissed):

PR #9777, opened 2022-05-22
Make phpdoc more precise for AbstractQuery

greg0ire:
Blocked by https://github.com/doctrine/orm/pull/9774

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -326,7 +326,7 @@ abstract class AbstractQuery
     /**
      * Gets a query parameter.
      *
-     * @param mixed $key The key (index or name) of the bound parameter.
+     * @param int|string $key The key (index or name) of the bound parameter.
      *
      * @return Parameter|null The value of the bound parameter, or NULL if not available.
      */
@@ -818,11 +818,10 @@ abstract class AbstractQuery
     /**
      * Change the default fetch mode of an association for this query.
      *
-     * $fetchMode can be one of ClassMetadata::FETCH_EAGER or ClassMetadata::FETCH_LAZY
-     *
-     * @param string $class
-     * @param string $assocName
-     * @param int    $fetchMode
+     * @param class-string $class
+     * @param string       $assocName
+     * @param int          $fetchMode
+     * @psalm-param Mapping\ClassMetadata::FETCH_EAGER|Mapping\ClassMetadata::FETCH_LAZY $fetchMode
      *
      * @return $this
      */
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -32,6 +32,7 @@ use function array_map;
 use function array_shift;
 use function assert;
 use function count;
+use function in_array;
 use function is_array;
 use function is_numeric;
 use function is_object;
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -817,11 +818,9 @@ public function getQueryCacheProfile()
...
      * Change the default fetch mode of an association for this query.
      *
-     * $fetchMode can be one of ClassMetadata::FETCH_EAGER or ClassMetadata::FETCH_LAZY
-     *
-     * @param string $class
-     * @param string $assocName
-     * @param int    $fetchMode
+     * @param class-string                                         $class
+     * @param string                                               $assocName
+     * @param ClassMetadata::FETCH_EAGER|ClassMetadata::FETCH_LAZY $fetchMode
```
derrabus:
```suggestion
     * @param int $fetchMode
     * @psalm-param ClassMetadata::FETCH_EAGER|ClassMetadata::FETCH_LAZY $fetchMode
```

I don't know how well literal types are supported outside of PHPStan and Psalm. So far, we have vendor-prefixed annotations that use them.

derrabus: Looking at the implementation, it seems to hide potential errors. Invalid fetch modes are healed right now, but I think, we should deprecate them, so we can throw in 3.0.

greg0ire: Ok, so let's keep `int` for now.

derrabus: Documenting literal types is a good idea, I think. It's just that I would vendor-prefix the `@param` here.

greg0ire: :thinking: sounds like you would with with @VincentLanglet on https://github.com/doctrine/orm/pull/9778 then
In this particular instance, I think it can't hurt, and I would be eager to know what your stance is on that PR. I was wondering if we should:
- only ever document the non-deprecated path
- always document both
- decide of that on a case-by-case basis.

greg0ire: > I think it can't hurt

It already makes the build fail though, not sure that it's a great idea.

lib/Doctrine/ORM/Query.php
```diff
@@ -118,7 +118,7 @@ final class Query extends AbstractQuery
     /**
      * The current state of this query.
      *
-     * @var int
+     * @var self::STATE_*
```
```suggestion
     * @var int
     * @psalm-var self::STATE_*
```

Let's keep literal types behind a vendor-prefix.

lib/Doctrine/ORM/Query.php
```diff
@@ -628,7 +628,7 @@ public function getDQL(): ?string
      * @see AbstractQuery::STATE_CLEAN
      * @see AbstractQuery::STATE_DIRTY
      *
-     * @return int The query state.
+     * @return self::STATE_* The query state.
```
```suggestion
     * @return int The query state.
     * @psalm-return self::STATE_* The query state.
```

lib/Doctrine/ORM/AbstractQuery.php
derrabus: `@param` annotations should document the public API, the intended usage of that method. In this case: It is not intended that anyone would pass `42`, `'hello world'` or even an array or a cURL resource to this method although it has always been possible.

> It already makes the build fail though

That's Psalm being Psalm. On 3.0, we would throw if the caller passed anything but the two allowed values, right? You would face the very same problem then. I'd argue that you cannot write that kind of input validation without Psalm complaining. PHPStan has a setting `treatPhpDocTypesAsCertain` for that very purpose.

derrabus (dismissed):

greg0ire:
Argh! a conflict!

PR #9779, opened 2022-05-23
Document callable as possible

greg0ire:
Custom string functions can either be a class string or a callable
returning the function.

https://github.com/doctrine/orm/blob/8f7701279de84411020304f668cc8b49a5afbf5a/lib/Doctrine/ORM/Query/Parser.php#L3636-L3654

As a follow up, I will make this more precise in 2.13.x (using `class-string<FunctionNode>`, and describing the callable more precisely.

### DIFF
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -608,7 +608,7 @@ class Configuration extends \Doctrine\DBAL\Configuration
      * @param string $name
      *
      * @return string|null
-     * @psalm-return ?class-string
+     * @psalm-return class-string|callable|null
      */
     public function getCustomStringFunction($name)
     {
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Configuration.php
```diff
@@ -608,7 +608,7 @@ public function addCustomStringFunction($name, $className)
      * @param string $name
      *
      * @return string|null
-     * @psalm-return ?class-string
+     * @psalm-return ?class-string|callable
```
derrabus:
```suggestion
     * @psalm-return class-string|callable|null
```

PHP does not allow `?` in unions and we should not use that syntax in doc blocks either.

derrabus:
I think, the doc block of `addCustomStringFunction()` can be made more precise as well (`class-string` instead of just `string`).

greg0ire:
@derrabus you're correct, and that's what I'm alluding to here:

>  I will make this more precise in 2.13.x

derrabus (dismissed):

derrabus:
PHPStan baseline needs an update.

derrabus (dismissed):

derrabus (approved):

PR #9784, opened 2022-05-23
Deprecate passing null to Query::setDQL()

greg0ire:
It is a no-op.

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -595,10 +595,19 @@ final class Query extends AbstractQuery
      */
     public function setDQL($dqlQuery): self
     {
-        if ($dqlQuery !== null) {
-            $this->dql    = $dqlQuery;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Query.php
```diff
@@ -595,11 +595,20 @@ public function free(): void
      */
     public function setDQL($dqlQuery): self
     {
-        if ($dqlQuery !== null) {
-            $this->dql    = $dqlQuery;
-            $this->_state = self::STATE_DIRTY;
+        if ($dqlQuery === null) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/pull/9779',
```
```suggestion
                'https://github.com/doctrine/orm/pull/9784',
```

🤓

greg0ire: Haha wrong prediction on my part!

derrabus (approved):

PR #9791, opened 2022-05-25
Deprecate calling setters without arguments

derrabus:
Calling a setter without anything to set does not really make sense semantically. However, declaring `null` as the default value for a parameter used to be a common workaround to get a nullable typed parameter. This workaround is obsolete since PHP 7.1 introduced actual nullable types.

This is why I'd like to get rid of those default values on setters.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -32,6 +32,7 @@ use function array_map;
 use function array_shift;
 use function assert;
 use function count;
+use function func_num_args;
 use function in_array;
 use function is_array;
 use function is_numeric;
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -548,6 +549,15 @@ private function translateNamespaces(Query\ResultSetMapping $rsm): void
     public function setHydrationCacheProfile(?QueryCacheProfile $profile = null)
     {
         if ($profile === null) {
+            if (func_num_args() < 1) {
+                Deprecation::trigger(
+                    'doctrine/orm',
+                    'TODO',
```
greg0ire: I'm the most forgetful person in the world and I want to thank you for making me feel better about that!

derrabus: 🤦🏻‍♂️

greg0ire (approved):

Issue #9335, opened 2022-01-07
doctrine orm Identity through foreign Entities Bug (when identifiers are of type RamseyUuid)


pytoccaz:
Hello doctrine community, 

### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | maybe ?
| Version     | 2.10.4

#### Summary

On a symfony api-platform project we tried to upgrade the following components :
symfony 5.2 --> symfony 5.4
php7.4 --> php8.1
mysql internal guid --> doctrine generated Ramsey uuid

All our entities verifying [Identity through foreign Entities](https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/tutorials/composite-primary-keys.html#identity-through-foreign-entities) design pattern lead to a break in code when "foreign entities" are refreshed.

#### Current buggy behavior

A buggy case with those 2 entities:

```php 
// Book.php
/**
 * @ORM\Entity 
 * @ORM\Table(name="book")
 */
class Book
{
    /**
     * @var \Ramsey\Uuid\UuidInterface
     *
     * @ORM\Id
     * @ORM\Column(type="uuid", unique=true)
     * @ORM\GeneratedValue(strategy="CUSTOM")
     * @ORM\CustomIdGenerator(class=UuidGenerator::class)
     */
    protected $id;
```

```php
// Author.php
/**
 * @ORM\Entity 
 * @ORM\Table(name="author")
 */
class Author
{  
 
    /**
     * @ORM\Id
     * @ORM\OneToOne(targetEntity="Book", inversedBy="author")
     * @ORM\JoinColumn(name="book")  
     */
    protected $book;

```

1. `Book` is identified by `$id` of type *Ramsey Uuid*
2.  `Author` is identified by `$book` of type `Book` (hence Book is the foreign entity). 

Note `@ORM\JoinColumn(name="book")` in `Author` class.
When `Author` `book` identifier is mapped with the same named `book` table column, doctrine hydrator breaks with the following error message when refreshing `Book` enties:

```
PHP Fatal error:  Uncaught ReflectionException: Given object is not an instance of the class this property was declared in in /home/olivier/work/projet/tests/doctrine/doctrine2-tutorial/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php:791
Stack trace:
[...]

PR #9793, opened 2022-05-26
Update IdentifierFlattener.php

olegsuvorkov:
Hello, I would like to make a small change.
The need arose when using \Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator in "symfony/doctrine-bridge" composite foreign keys
I'm sure these changes will not hurt performance
and allow other objects to be used as identifiers

### DIFF
--- a/lib/Doctrine/ORM/Utility/IdentifierFlattener.php
+++ b/lib/Doctrine/ORM/Utility/IdentifierFlattener.php
@@ -55,7 +55,7 @@ final class IdentifierFlattener
         $flatId = [];
 
         foreach ($class->identifier as $field) {
-            if (isset($class->associationMappings[$field]) && isset($id[$field]) && is_object($id[$field])) {
+            if (isset($class->associationMappings[$field]) && isset($id[$field]) && is_a($id[$field], $class->associationMappings[$field]['targetEntity'])) {
                 $targetClassMetadata = $this->metadataFactory->getMetadataFor(
                     $class->associationMappings[$field]['targetEntity']
                 );
--- a/lib/Doctrine/ORM/Utility/IdentifierFlattener.php
+++ b/lib/Doctrine/ORM/Utility/IdentifierFlattener.php
@@ -11,7 +11,7 @@ use Doctrine\Persistence\Mapping\ClassMetadataFactory;
 
 use function assert;
 use function implode;
-use function is_object;
+use function is_a;
 
 /**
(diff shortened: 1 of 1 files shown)


derrabus (changes requested):
Please describe the issue that you'd like to solve and add a test that covers your change.

beberlei:
I don't understand why a UUID is stored in the field instead of the associated entity when the ID is an association. That is the wrong result. I believe you are doing something wrong with your code and its not a Doctrine issue.

olegsuvorkov:
Thanks for the comments.
I am sure that in the object in the corresponding field of my code. an instance of the class specified in "targetEntity" is stored.
But "targetEntity" uses an instance of [\Symfony\Component\Uid\Uuid](https://github.com/symfony/uid/blob/5d1bba611fcd5d7ccc6435fd4ea8580f9ed75406/Uuid.php#L19) as an identifier.
And since the flattenIdentifier method is called recursively on the line
[\Ori\Eav\Utility\IdentifierFlattener:67](https://github.com/doctrine/orm/blob/48e4e333c7bf4757e9a4e19416d1a3e833ea28b0/lib/Doctrine/ORM/Utility/IdentifierFlattener.php#L67), on the next iteration, a ReflectionException occurs at the line 
[Doctrine/ORM/Mapping/ClassMetadataInfo:839](https://github.com/doctrine/orm/blob/48e4e333c7bf4757e9a4e19416d1a3e833ea28b0/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L839).
I will definitely write tests on this issue.

olegsuvorkov:
Details can be found here
[Storing UUIDs in Databases](https://symfony.com/doc/current/components/uid.html#storing-uuids-in-databases)

louisezetterlund:
Hi!
This PR solves issue #9335, which I'm also experiencing. The issue also contains a link to a repo with a simplified case where this bug exists. Do you want me to convert it to a test, so we can merge this PR? @derrabus @beberlei

derrabus:
Yes, the PR needs a test that proves the bug exists and makes sure it won't be introduced again by a future change.

louisezetterlund:
I've added a test in this commit https://github.com/rdey/orm/commit/58368e910589f58076197129d00d0d86bf3f5703.
Do you want me to create a new PR or can you add it to this PR? @derrabus

derrabus:
Please open a new PR that contains your test and the changes from this PR.

PR #9794, opened 2022-05-29
Add type for AssociationMapping

VincentLanglet:


VincentLanglet:
@greg0ire I tried to add all the key with some type to the AssociationMapping, by looking at the usage. I can't be sure if I have all the possible key, and if I didn't make any mistake about the optional ones `?`, but I think it can be a good start.

This is useful when using 
```
$this->em->getClassMetadata($associationMapping['targetEntity']);
```
because it avoid to write
```
/** @var class-string $targetEntity */
$targetEntity = $assoc['targetEntity'];
$targetClass = $this->em->getClassMetadata($targetEntity);
```
for phpstan.

I fixed two baseline errors, but I have a new one:
```
$associationMapping['type'] === 'manyToMany'
```
is throwing 
```
Strict comparison using === between int and 'manyToMany' will always evaluate to false.
```
Can the type be a string ? I thought it would be the const `public const MANY_TO_MANY = 8;`.

greg0ire:
I'm on my phone so it's hard to help, but I'm OK with the "good start, possibly incorrect" way of doing things when it's just about phpdoc. Regarding the new issue, maybe try throwing when it's a string and then run the test suite, to see if that case is covered?

VincentLanglet:
Seems like it was just a bug since `generateAssociationMappingPropertyDocBlock` is considering `$associationMapping['type']` to be an int.

The phpdoc added a lot of issues with psalm (most of them are possibly undefined offset); there are too much to solve so I updated the psalm baseline

derrabus:
We should try to reduce the set of optional keys in that array. Otherwise we'll find ourselves adding lots of pointless `assert()` calls to reduce the baseline again. 😕

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1952,7 +1955,7 @@ private function doMerge(
         $entity,
         array &$visited,
         $prevManagedCopy = null,
-        array $assoc = []
+        ?array $assoc = null
```
SenseException, line 1958: When does `$assoc` get a `null` value injected? The second argument of the `updateAssociationWithMergedEntity()` method used inside `doMerge()` doesn't allow `null`.

VincentLanglet:
> We should try to reduce the set of optional keys in that array. Otherwise we'll find ourselves adding lots of pointless `assert()` calls to reduce the baseline again. 😕

Sure, but do you see any key which are not optional and I made optional ?

I'll try to take a new look soon but I'm not sure that's possible since it depends on the type and the config of the relation.

lib/Doctrine/ORM/UnitOfWork.php
VincentLanglet: No args are passed here: https://github.com/doctrine/orm/blob/0f9cc194ae28881009db7c5a29b85c5abdef2f6d/lib/Doctrine/ORM/UnitOfWork.php#L1937

updateAssociationWithMergedEntity does not accept `null` but it doesn't accept an empty array either
https://github.com/doctrine/orm/blob/0f9cc194ae28881009db7c5a29b85c5abdef2f6d/lib/Doctrine/ORM/UnitOfWork.php#L2102

So I assume, that when it's called without args, the `updateAssociationWithMergedEntity` call is not done too.
IMHO it will be better to consider `null` as no value rather than `[]` as no value. And some null check can be added later for static analysis.

derrabus:
For instance, most of the boolean keys look like it would make sense to always define them. Not sure if that'll help us though.

VincentLanglet:
> For instance, most of the boolean keys look like it would make sense to always define them. Not sure if that'll help us though.

I thought you were talking about the phpdoc and the type I introduced.

Indeed it would make sens to aways define them. But I wouldn't risk those change in this PR because
- I know nothing about how this code work so I'm not even sure how/where all the association mappings are generated.
- This is much more risky to introduce a bug than my phpdoc update because there is sometimes checks like `isset($mapping['key'])` or `$mapping['key) === null`.
I would recommend to do in another (or mutiple others) PR.

Do you think it must be done with this PR ?

psalm-baseline.xml
```diff
@@ -1051,6 +1092,11 @@
       <code>$entity</code>
     </PossiblyInvalidArgument>
   </file>
+  <file src="lib/Doctrine/ORM/ORMSetup.php">
+    <UndefinedClass occurrences="1">
+      <code>MemcachedAdapter::createConnection('memcached://127.0.0.1')</code>
+    </UndefinedClass>
+  </file>
```
derrabus, line 1099: This error looks odd and unrelated to your changes.

derrabus (approved):

derrabus:
> Do you think it must be done with this PR ?

No. This PR is a great step towards getting our types under control. Let's iterate from there.

lib/Doctrine/ORM/UnitOfWork.php
SenseException: It's a weird situation that shouldn't be handled in this PR.

SenseException (approved):

greg0ire:
Thanks @VincentLanglet !

PR #9799, opened 2022-05-31
Prep work for enabling support for foreign keys on SQLite

morozov:


lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -857,12 +856,9 @@ public function dropDatabase()
      */
     public function getDropDatabaseSQL()
     {
-        $schema = $this->schemaManager->createSchema();
-
-        $visitor = new DropSchemaSqlCollector($this->platform);
```
morozov, line 862: `DropSchemaSqlCollector` will be deprecated in favor of another class. It's better if the ORM doesn't depend on such low-level implementation details at all and uses the schema manager for managing schema.

tests/Doctrine/Tests/ORM/Functional/SchemaTool/CompanySchemaTest.php
```diff
@@ -49,21 +48,4 @@ public function testSingleTableInheritance(Schema $schema): void
...
-        if (! $this->_em->getConnection()->getDatabasePlatform()->supportsForeignKeyConstraints()) {
-            self::markTestSkipped('Foreign Key test');
-        }
-
-        $sql = $this->_schemaTool->getDropSchemaSQL(
-            [
-                $this->_em->getClassMetadata(CompanyManager::class),
-            ]
-        );
-        self::assertCount(4, $sql);
```
morozov, line 67: The SQL that the DBAL uses for managing schema is an implementation detail that may and will change.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC832Test.php
```diff
@@ -41,6 +42,11 @@ public function tearDown(): void
         $sm->dropTable($platform->quoteIdentifier('TREE_INDEX'));
         $sm->dropTable($platform->quoteIdentifier('INDEX'));
         $sm->dropTable($platform->quoteIdentifier('LIKE'));
+
+        if ($platform instanceof PostgreSQLPlatform) {
```
morozov, line 46: The problem with this code is that the test suite ignores the exceptions thrown by the schema tool: https://github.com/doctrine/orm/blob/4f1072e1ac9480acaee37a6fff44f001247392e5/tests/Doctrine/Tests/OrmFunctionalTestCase.php#L348-L356

...and relies on the fact that `CreateSchemaSqlCollector` creates all tables before all sequences.

Without this change, if each sequence is created after its corresponding table, for each test except for the first, the `setUp()` will fail to create the first sequence (because it already exists) and thus will not create all three tables.

We could implement a `dropSchemaForModels()` method similar to `createSchemaForModels()` and use it instead but on the DBAL versions prior to 3.4 it would use `DropSchemaSqlCollector` internally which attempts to drop foreign keys before dropping the table, even if it's not necessary, but SQLite doesn't support that. The smallest of all evils, for now, seems to work around it.

tests/Doctrine/Tests/ORM/Tools/Console/Command/ValidateSchemaCommandTest.php
```diff
@@ -57,6 +57,11 @@ public function testNotInSync(): void
 
     public function testNotInSyncVerbose(): void
     {
+        $schemaManager = $this->createSchemaManager();
+        if ($schemaManager->tablesExist('cache_login')) {
```
morozov, line 61: This is also a mess. This test will fail if SQLite starts reporting that it supports foreign keys _and_ if some other tests are executed before it (e.g. `DDC2943Test`). The `cache_login` table will already exist at this point and won't be part of the diff.

The test currently passes because even by the time when it's executed, the `cache_login` table already exists, its foreign key is not introspected, so the schema diff always contains a missing foreign key.

The proper solution here would be instead of deleting data from all tables in the functional test teardown, to drop those tables but the hard-coded list of tables is inconsistent with the models, so reworking it requires additional effort.

derrabus (approved):

PR #9805, opened 2022-06-02
Remove empty test file

derrabus:
This file apparently should've been deleted in 770d00abe91220323680aa97ea82f9eae9479ef5, but was instead replaced with a zero-byte file. PHPUnit 10 will complain about the existence of that file, so let's remove it.


SenseException (approved):

greg0ire (approved):
It's year 12… they haven't noticed me yet… let's keep a low profile. — the file, probably

PR #9817, opened 2022-06-03
Deprecate passing ArrayCollection instances

greg0ire:
There does not seem to be a point to it, and it is very verbose.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Doctrine\ORM;
 
 use BackedEnum;
-use Countable;
 use Doctrine\Common\Cache\Psr6\CacheAdapter;
 use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\Common\Collections\ArrayCollection;
@@ -39,7 +38,6 @@ use function is_numeric;
 use function is_object;
 use function is_scalar;
 use function is_string;
-use function iterator_count;
 use function iterator_to_array;
 use function ksort;
 use function method_exists;
@@ -350,18 +348,24 @@ abstract class AbstractQuery
      */
     public function setParameters($parameters)
     {
-        if (is_array($parameters)) {
-            /** @psalm-var ArrayCollection<int, Parameter> $parameterCollection */
-            $parameterCollection = new ArrayCollection();
+        if (! is_array($parameters)) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/pull/9816',
+                'Passing an ArrayCollection to %s is deprecated.',
+                __METHOD__
+            );
+            $this->parameters = $parameters;
 
-            foreach ($parameters as $key => $value) {
-                $parameterCollection->add(new Parameter($key, $value));
-            }
-
-            $parameters = $parameterCollection;
+            return $this;
         }
 
-        $this->parameters = $parameters;
+        $this->parameters = new ArrayCollection();
+
+        foreach ($parameters as $key => $value) {
+            $parameter = $value instanceof Parameter ? $value : new Parameter($key, $value);
+            $this->parameters->add($parameter);
+        }
 
         return $this;
     }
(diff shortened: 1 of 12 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -356,18 +354,24 @@ static function (Query\Parameter $parameter) use ($key): bool {
...
-            $parameters = $parameterCollection;
+            return $this;
         }
 
-        $this->parameters = $parameters;
+        $this->parameters = new ArrayCollection();
+
+        foreach ($parameters as $key => $value) {
+            $parameter = $value instanceof Parameter ? $value : new Parameter($key, $value);
+            $this->parameters->add($parameter);
```
greg0ire, line 367: Before this PR, passing an array of `Parameter` instances does nothing special, so technically, this is a BC break.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -1131,14 +1135,21 @@ public function iterate($parameters = null, $hydrationMode = null)
...
+
         if ($hydrationMode !== null) {
             $this->setHydrationMode($hydrationMode);
         }
 
-        if (
-            ($this->isCountable($parameters) && count($parameters) !== 0)
-            || ($parameters instanceof Traversable && iterator_count($parameters) !== 0)
-        ) {
+        if (count($parameters) !== 0) {
```
greg0ire, line 1142: What do you think about that first commit? Too dangerous?

SenseException: It can definitely break something after an update to 2.13 because it was able to inject other iterators into `toIterable()`. Maybe this PR is more than just deprecating `ArrayCollection`.

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -356,18 +354,24 @@ static function (Query\Parameter $parameter) use ($key): bool {
...
+                __METHOD__
+            );
+            $this->parameters = $parameters;
 
-            $parameters = $parameterCollection;
+            return $this;
         }
 
-        $this->parameters = $parameters;
+        $this->parameters = new ArrayCollection();
```
SenseException, line 363: In other methods you're using `$parameters->toArray();` on an `ArrayCollection` to internally get rid of it. Here you explicitly set `$this->parameters` with an  `ArrayCollection`. Even though this PR is about deprecating  `ArrayCollection` as an injectable parameter type, shouldn't this type be prevented at all? Otherwise `getParameters()` would still return an  `ArrayCollection`, which is kind of unexpected for a deprecated type.

greg0ire: If `getParameters()` returns something else, we will have a BC-break though, won't we?

SenseException: Ah, right. The others aren't written into `$this->parameters`.

lib/Doctrine/ORM/AbstractQuery.php
greg0ire: OK, I went the safe road: I widened the phpdoc to `iterable`. This complex check can still go because I'm converting `Traversable` to array with `iterator_to_array`, so as not to trigger deprecations when calling other methods.

lib/Doctrine/ORM/AbstractQuery.php
derrabus: So we basically accept arrays that look like this?

```php
[
    'foo' => 'some value',
    new Parameter('bar', 'another value'),
]
```

Is this what we want?

Maybe we should split this method into two, one that accepts associative arrays and one that accepts a list of `Parameter` instances?

derrabus: Mmmh, I take back my proposal. It would work here, but it would complicate all other methods like `toIterable()`…

greg0ire: > Is this what we want?

Your example shows a side-effect of allowing this, but it's not really realistic, is it? Anyway, I don't think we should check and throw in case the array is a mix of a hash and a list.

derrabus: Yeah, probably. I can't put my finger on it, but it just feels like we haven't gotten to the good solution here. Sorry for being so indecisive. 😓

greg0ire: Take all the time you need :slightly_smiling_face:

lib/Doctrine/ORM/AbstractQuery.php
SenseException: `iterator_to_array` will remove all keys in its current usage. As far as I understand they need to be kept in the resulting array. There is still the chance that the values aren't `Parameter`.

SenseException (changes requested):

lib/Doctrine/ORM/AbstractQuery.php
derrabus: `iterator_to_array()` should actually preserve keys. https://3v4l.org/Xo4AI

SenseException: I mixed up the default value of the second parameter.

SenseException (dismissed):

SenseException (approved):

PR #9821, opened 2022-06-07
Keep schema null if its not defined #9820

### LINKED ISSUE #9820: dot prefixed TableName on AssociationOverrides via XmlDriver when not defining a schema
### Bug Report

When defining a AssociationOverride via XmlDriver, it tries to set an empty Schema that results in a table named with a dot prefix only.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.12.2

#### Summary

Tries to set empty string as schema on AssociationOverride on a join table .

#### Current behavior

Tries to set empty string as schema resulting in a Statement like that:

```
CREATE TABLE .folder_notes (folder_id UUID NOT NULL, note_id UUID NOT NULL, PRIMARY KEY(folder_id, note_id));
```

#### How to reproduce

Define a JoinTable association override without defining a schema, e.g.:

```
<association-overrides>
  <association-override name="notes">
    <join-table name="folder_notes">
      <join-columns>
        <join-column name="folder_id"/>
      </join-columns>
    </join-table>
  </association-override>
</association-overrides>
```

#### Expected behavior

Should not prefix when string is empty.

Issue seems to be the `(string)` here:

https://github.com/doctrine/orm/blob/67d82cdf722af84b41152553be8bfb35dd025796/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php#L668

develth:
Keep the schema on the JoinTable null if its not defined to prevent a dot only prefixed table.

Should fix #9820


derrabus:
Thank you for your PR. If this change fixes a bug, please add a test that reproduces the bug you're attempting to fix, to prevent possible future regressions.

PR #9827, opened 2022-06-09
bug: getSingleColumnResult with indexBy

gimler:
This change make it possible to use `indexBy` with `getSingleColumnResult`.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ScalarColumnHydrator.php
@@ -8,6 +8,7 @@ use Doctrine\DBAL\Driver\Exception;
 use Doctrine\ORM\Exception\MultipleSelectorsFoundException;
 
 use function array_column;
+use function array_combine;
 use function count;
 
 /**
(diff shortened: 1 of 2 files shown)


greg0ire:
> Document missing possible types

Why did you copy the title of #9818?

gimler:
@greg0ire sorry my fault

greg0ire:
Can you please add a new test to test the new behavior? Also, are there github issues that this should fix?

gimler:
@greg0ire 
I think the actual test `testSingleColumnEntityQueryWithScalarMap` is not correct thats why i have correct the result.

i did not create a bug issue, but i use the following code in my app
```
$this->metalRepository->createQueryBuilder('c', 'c.title')
    ->select('c.title')
    ->groupBy('c.title')
    ->orderBy('c.title', 'ASC')
    ->getQuery()
    ->getSingleColumnResult();
```

without the patch the result is the following:
```
[
 0 => 'alpha', 1 => 'beta'
]
```

with the patch the result is correct:
```
[
 'alpha' => 'alpha', 'beta' => 'beta'
]
```

greg0ire:
Ok, it's just that I was wondering if this was reported by somebody else, but I must be thinking of a similar bug.

greg0ire:
@bhushan please review :pray:

bhushan:
@gimler `getSingleColumnResult` is not intended to be used with group by, I mean why will you do that when you are only expecting single column?

In my opinion we should remove that test. If you also agree, please update PR to `remove test with scalarMap` and `keep without scalarmap test`.

cc @greg0ire

greg0ire:
:thinking: that makes sense, I mean, `['alpha' => 'alpha', 'beta' => 'beta']` has redundant information, right?

PR #9830, opened 2022-06-09
Run foreign key tests on all platforms

greg0ire:
SQLite now supports foreign keys.

2.13.x requires DBAL 3.4, so this should be fine.

https://github.com/doctrine/dbal/pull/5427


derrabus:
> 2.13.x requires DBAL 3.4

Right now, we still claim compatibility with DBAL 2.13 and 3.2.

greg0ire:
Oh right, I'm confusing with 3.0.x!

PR #9835, opened 2022-06-09
Migrate Configuration to PHP 8 syntax

greg0ire:


lib/Doctrine/ORM/Configuration.php
```diff
@@ -79,45 +73,35 @@ public function getAutoGenerateProxyClasses()
...
         $this->_attributes['proxyNamespace'] = $ns;
     }
 
     /**
      * Sets the cache driver implementation that is used for metadata caching.
      *
-     * @return void
-     *
      * @todo Force parameter to be a Closure to ensure lazy evaluation
      *       (as soon as a metadata cache is in effect, the driver never needs to initialize).
```
greg0ire, line 102: I don't understand this todo

derrabus: I think, the author wants that we can set a `Closure(): MappingDriver` instead of a `MappingDriver` instance here.

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -613,7 +613,7 @@ public function newHydrator(string|int $hydrationMode): AbstractHydrator
                 }
         }
 
-        throw InvalidHydrationMode::fromMode((string) $hydrationMode);
```
derrabus: I think, the string cast should stay.

greg0ire: I removed it because Psalm right infers that it is redundant according to this:

https://github.com/doctrine/orm/blob/3dc5581294ccc05e0150c8b6708ccefbea88849e/lib/Doctrine/ORM/EntityManagerInterface.php#L298

derrabus: I think, we've been here before. That annotation prevents nobody from passing an integer that is not in the `AbstractQuery::HYDRATE_*` range. Psalm is wrong complaining about this kind of "redundancy". "Fixing" this kind of error would basically mean removing input validation from all methods.

Note that this very piece of code is reached specifically for invalid input. So we have to assume that the caller passed something that does not match the doc block.

lib/Doctrine/ORM/Configuration.php
greg0ire: Ok… I think there is a number of todos we could challenge in this project.

lib/Doctrine/ORM/EntityManager.php
greg0ire: I've restored the string cast.

derrabus: You did not, apparently. 🤔

greg0ire: Ooooh… I was wondering why this didn't get approved :sweat_smile:

greg0ire: Restored for real this time!

SenseException (approved):

PR #9836, opened 2022-06-10
Merge 2.13.x up into 3.0.x

greg0ire:


greg0ire:
Failure caused by https://github.com/doctrine/dbal/pull/5441, let's see if we can address that on 2.12.x

PR #9839, opened 2022-06-11
Use proper API for introspection of tables

morozov:
The existing code would invalidate the optimization implemented in https://github.com/doctrine/dbal/pull/5268.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
@@ -280,15 +280,10 @@ class DatabaseDriver implements MappingDriver
             return;
         }
 
-        $tables = [];
-
-        foreach ($this->_sm->listTableNames() as $tableName) {
-            $tables[$tableName] = $this->_sm->listTableDetails($tableName);
-        }
-
         $this->tables = $this->manyToManyTables = $this->classToTableNames = [];
 
-        foreach ($tables as $tableName => $table) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -280,15 +280,12 @@ private function reverseEngineerMappingFromDatabase(): void
             return;
         }
 
-        $tables = [];
-
-        foreach ($this->_sm->listTableNames() as $tableName) {
-            $tables[$tableName] = $this->_sm->listTableDetails($tableName);
-        }
+        $tables = $this->_sm->listTables();
```
greg0ire: This variable seems to be used only once, and does not need to be defined before the assignments below.

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -280,15 +280,12 @@ private function reverseEngineerMappingFromDatabase(): void
...
 
         $this->tables = $this->manyToManyTables = $this->classToTableNames = [];
 
-        foreach ($tables as $tableName => $table) {
+        foreach ($tables as $table) {
```
greg0ire: Let's get rid of it and inline it here if that's indeed the case?

```suggestion
        foreach ($this->_sm->listTables() as $table) {
```

greg0ire (approved):

greg0ire:
Thanks @morozov !

PR #9840, opened 2022-06-12
Run tests on PHP 8.2

derrabus:
PHP 8.2.0-alpha1 has been released and I think we should start monitoring our compatibility with the next PHP version.


.github/workflows/continuous-integration.yml
```diff
@@ -31,6 +31,8 @@ jobs:
             dbal-version: "2.13"
           - php-version: "8.1"
             dbal-version: "3@dev"
+          - php-version: "8.2"
+            dbal-version: "3@dev"
```
derrabus, line 35: As long as doctrine/dbal#5381 is not released, the tests will fail on PHP 8.2 if we don't test against a dev snapshot.

.github/workflows/continuous-integration.yml
```diff
@@ -52,6 +54,8 @@ jobs:
 
       - name: "Install dependencies with Composer"
         uses: "ramsey/composer-install@v1"
+        with:
+          composer-options: "--ignore-platform-req=php+"
```
derrabus, line 58: We need to ignore the upper boundary of PHP version constraints. Otherwise, we cannot install Prophecy (as a dependency of PHPUnit) on PHP 8.2.

derrabus:
The test failures on Postgres are most likely unrelated to PHP 8.2.

derrabus:
@greg0ire @morozov Can you have a look that the failing Postgres tests in this PR? It looks like those tests have been deleted on higher branches (e.g. #9799). Can you check if we can backport that to 2.12.x as well? Otherwise, our CI on 2.12.x will break as soon as DBAL 3.3.7 is tagged.

tests/Doctrine/Tests/ORM/Functional/ExtraLazyCollectionTest.php
```diff
@@ -51,6 +51,9 @@ class ExtraLazyCollectionTest extends OrmFunctionalTestCase
     /** @var CmsPhonenumber */
     private $phonenumber;
 
+    /** @var array<string, mixed> */
+    private $previousCacheConfig = [];
```
derrabus: Cherry-picked to #9841

.github/workflows/continuous-integration.yml
greg0ire: Didn't know about that `+` syntax :bulb:

greg0ire:
> Can you check if we can backport that to 2.12.x as well?

I pushed the cherry-picked commit to your branch since it's the easiest way to test.

derrabus:
I have reverted all changes to production code. I think it's better to deliver those with 2.13.x. The CI is green now. Shall we merge?

morozov (approved):

PR #9841, opened 2022-06-12
Add missing property declaration

derrabus:
In `ExtraLazyCollectionTest`, we use a property `previousCacheConfig` without declaring it. In PHP 8.2, this will triggger a deprecation notice. I'm sure that the declaration is missing my accident, so this PR adds it.


greg0ire (approved):

greg0ire:
Thanks @derrabus !

PR #9845, opened 2022-06-13
Add hint how to generate proxy classes without database

meidlinga:


SenseException:
Thanks for the PR. What's the story behind this additional note? It sounds like it belongs to the DoctrineBundle.

meidlinga:
Hi,

While trying to get doctrine to production, I tried to run `vendor/bin/doctrine orm:generate-proxies` in a Dockerfile, where no database is available. It failed with:
```
HP Fatal error:  Uncaught Doctrine\DBAL\Driver\Mysqli\Exception\ConnectionFailed: php_network_getaddresses: getaddrinfo failed: Name or service not known in /app/vendor/doctrine/dbal/src/Driver/Mysqli/Exception/ConnectionFailed.php:26
Stack trace:
#0 /app/vendor/doctrine/dbal/src/Driver/Mysqli/Driver.php(74): Doctrine\DBAL\Driver\Mysqli\Exception\ConnectionFailed::new(Object(mysqli))
#1 /app/vendor/doctrine/dbal/src/Connection.php(340): Doctrine\DBAL\Driver\Mysqli\Driver->connect(Array)
#2 /app/vendor/doctrine/dbal/src/Connection.php(404): Doctrine\DBAL\Connection->connect()
#3 /app/vendor/doctrine/dbal/src/Connection.php(366): Doctrine\DBAL\Connection->getDatabasePlatformVersion()
#4 /app/vendor/doctrine/dbal/src/Connection.php(283): Doctrine\DBAL\Connection->detectDatabasePlatform()
#5 /app/vendor/doctrine/dbal/src/Connection.php(1565): Doctrine\DBAL\Connection->getDatabasePlatform()
#6 /app/bootstrap.php(43): Doctrine\DBAL\Connection->createSchemaManager()
``` 

After some searching I stumbled upon a comment that gave me a hint how to fix it.:
https://github.com/doctrine/dbal/issues/2409#issuecomment-224983765

Since it costed some time for me to figure this out and I guess other will struggle here to, because the platform is not specified in other examples in the documentation, I thought this hint would be a good addtion.

Maybe I get this wrong, but I think this is not related to the (Symfony) DoctrineBundle. Since this happens when would happen also when used in other frameworks or in plain PHP. At least it happened to me with Laminas. :)

docs/en/reference/advanced-configuration.rst
```diff
@@ -385,6 +385,13 @@ The other options are interesting in development environment:
 
 - ``AUTOGENERATE_EVAL`` will regenerate each proxy on each request,
   but without writing them to disk.
+  
+.. note::
+
+In some cases, for example when building a container, you would like
```
beberlei, line 391: This is not only relevant to proxy generation. Maybe settibg platform explicitly should be a more prominent section in the docs?

@derrabus @greg0ire should we require this in 3.0?

greg0ire: @beberlei Do you mean in (DBAL) 4.0? I don't think the ORM should have knowledge about the platform, or that it should inspect the DBAL connection (`Connection::getParams()` is marked as internal).

docs/en/reference/advanced-configuration.rst
```diff
@@ -385,6 +385,13 @@ The other options are interesting in development environment:
 
 - ``AUTOGENERATE_EVAL`` will regenerate each proxy on each request,
   but without writing them to disk.
+  
+.. note::
+
+In some cases, for example when building a container, you would like 
+to generate proxies while not having a database available.
+Then you should set ``platform`` in the connection ``params``. Otherwise 
+doctrine will try to make a connection to the database to determine it.
```
greg0ire, line 394: If the error message you got when attempting to generate proxies was unclear, maybe that's what needs to be improved?

meidlinga: @greg0ire Yes, transporting the hint to set the platform would also help in my opinion. Maybe this could be done of all the "db-less capable" commands, as @beberlei mentioned. The error could be catched in those commands I think.

meidlinga: @greg0ire , @beberlei Sorry, thinking about this again, it might be harder to achive to catch this error, since the connection is made in bootstrap.php where the actual command is not known. (If I understand correctly)

docs/en/reference/advanced-configuration.rst
beberlei: Ah, my mistake!

docs/en/reference/advanced-configuration.rst
greg0ire: `boostrap.php` is a file you have control over, and it seems weird to me that you would need to create a connection while bootstrapping your application… what if you run a command that does not have to do with Doctrine at all?

meidlinga: Yes, you are right. Sorry, I was distracted by the custom field mappings we also register.

So: yes, it would probably be possible (and as helpful as a hint in the docs) to look for this error in the generate proxies command (maybe others as well) and print an according hint.

PR #9848, opened 2022-06-15
Improve orphan removal documentation - recommend using cascade=persist 

### LINKED ISSUE #9847: OrphanRemoval does not work with ArrayCollection and manual persist which is not mentioned in the docs
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.x

#### Summary
I'd say this issue does not have a simple solution and it just requires documentation change.

Assume we have following entities:

```php
#[Entity]
class Foo
{

    #[OneToMany(mappedBy: "foo", targetEntity: Bar::class, orphanRemoval: true)]
    public Collection $collection;

    public function __construct()
    {
        $this->collection = new ArrayCollection();
    }

}

#[Entity]
class Bar
{

    #[JoinColumn]
    #[ManyToOne(targetEntity: Foo::class, inversedBy: "collection")]
    public Foo $foo;

}
```

#### Current behavior
Following code will create an entity even though the collection should be empty:
```php
$foo = new Foo();
$bar = new Bar();

$bar->foo = $foo;
$entityManager->persist($bar);
$foo->collection->add($bar);

$foo->collection->clear();

// Bar will be created in the DB and hydrated on refresh
$entityManager->flush();
```

The issue is caused by combination of `orphanRemoval` with manually calling `$entityManager->persist($entity)` which is hard to fix (as ArrayCollection does not have access to entity manager) yet undocumented so unexpected.

Following code will work as expected:
```php
// Working - Bar will be created & deleted
$foo = new Foo();
$bar = new Bar();

$bar->foo = $foo;
$entityManager->persist($bar);
$foo->collection->add($bar);

// Now the collection becomes PersistentCollection which does propagate changes to UnitOfWork
$entityManager->flush();

$foo->collection->clear();

$entityManager->flush();
```

#### Expected behavior
The only possible solution that comes to my mind is to document this behavior in `orphanRemoval` part and recommend using it alongside `cascade=["persist"]`

olsavmic:
Resolves https://github.com/doctrine/orm/issues/9847

### DIFF
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -37,7 +37,7 @@ information about its type and if it's the owning or inverse side.
     {
         /** @Id @GeneratedValue @Column(type="string") */
         private $id;
-    
+
         /**
          * Bidirectional - Many users have Many favorite comments (OWNING SIDE)
          *
@@ -45,7 +45,7 @@ information about its type and if it's the owning or inverse side.
          * @JoinTable(name="user_favorite_comments")
          */
         private $favorites;
-    
+
         /**
          * Unidirectional - Many users have marked many comments as read
          *
@@ -53,14 +53,14 @@ information about its type and if it's the owning or inverse side.
          * @JoinTable(name="user_read_comments")
          */
         private $commentsRead;
-    
+
         /**
          * Bidirectional - One-To-Many (INVERSE SIDE)
          *
          * @OneToMany(targetEntity="Comment", mappedBy="author")
          */
         private $commentsAuthored;
-    
+
         /**
          * Unidirectional - Many-To-One
          *
@@ -68,20 +68,20 @@ information about its type and if it's the owning or inverse side.
          */
         private $firstComment;
     }
-    
+
     /** @Entity */
     class Comment
     {
         /** @Id @GeneratedValue @Column(type="string") */
         private $id;
-    
+
         /**
          * Bidirectional - Many comments are favorited by many users (INVERSE SIDE)
          *
          * @ManyToMany(targetEntity="User", mappedBy="favorites")
          */
         private $userFavorites;
-    
+
         /**
          * Bidirectional - Many Comments are authored by one user (OWNING SIDE)
          *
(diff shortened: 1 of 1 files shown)


docs/en/reference/working-with-associations.rst
```diff
@@ -559,6 +559,13 @@ OrphanRemoval works with one-to-one, one-to-many and many-to-many associations.
     If you neglect this assumption your entities will get deleted by Doctrine even if
     you assigned the orphaned entity to another one.
 
+.. note::
+
+    ``orphanRemoval=true`` option should be used in combination with `cascade=["persist"]` option
+    as the child entity that is manually persisted will not be deleted automatically by Doctrine
+    when collection is still instance of ArrayCollection (before first flush / hydration).
+    This is a Doctrine limitation since ArrayCollection does not have access to UnitOfWork.
```
SenseException:
```suggestion
    ``orphanRemoval=true`` option should be used in combination with `cascade=["persist"]` option
    as the child entity, that is manually persisted, will not be deleted automatically by Doctrine
    when a collection is still an instance of ArrayCollection (before first flush / hydration).
    This is a Doctrine limitation since ArrayCollection does not have access to a UnitOfWork.
```

SenseException (changes requested):

SenseException:
I just saw that you're targeting 2.11 instead of 2.12. Please change and rebase to `2.12.x`

olsavmic:
Rebased onto 2.12.x, thanks!

docs/en/reference/working-with-associations.rst
```diff
@@ -559,6 +559,13 @@ OrphanRemoval works with one-to-one, one-to-many and many-to-many associations.
     If you neglect this assumption your entities will get deleted by Doctrine even if
     you assigned the orphaned entity to another one.
 
+.. note::
+
+    ``orphanRemoval=true`` option should be used in combination with `cascade=["persist"]` option
```
SenseException: I oversaw something in my last change request. Sorry. 
```suggestion
    ``orphanRemoval=true`` option should be used in combination with ``cascade=["persist"]`` option
```

SenseException (changes requested):
Maybe you can keep the original orphanRemoval example above your newly added note. It's still a valid example. You can create your own shorter code example below your note to separate both cases.

derrabus (approved):

SenseException (approved):

SenseException:
Thank you @olsavmic :rocket:

Issue #9849, opened 2022-06-15
Regression caused by e369cb6


asimonf:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.12.*

#### Summary

I have an inheritance tree using joined tables where a version field for optimistic locking exists for a subclass, but not in the parent. Whenever I try to insert, it fails after executing this query:

`SELECT versionLock FROM Car WHERE id = '...';`

The exception message is 'Unexpected empty result for database query.'

#### Current behavior

When using Version for optimistic locking in JOINED subclasses, the refresh fails on new inserts since the inserted version does not exist on the parent subclass, yet it tries to refresh the subclass.

The exception is triggered on the BasicEntityPersister.php file here:

```
        $values = $this->conn->fetchNumeric(
            $sql,
            array_values($flatId),
            $this->extractIdentifierTypes($id, $versionedClass)
        );

        if ($values === false) {
            throw new LengthException('Unexpected empty result for database query.');
        }
```

#### How to reproduce

Create an inheritance tree. The parent class must define a JOINED inheritance type and a child class must declare a field with the Version annotation. Afterwards, try to insert a child object into the database. This happens with numeric "version" that have a default value assigned by the database engine.

#### Expected behavior

That the child is inserted properly.

greg0ire:
Link: e369cb6

PR #9850, opened 2022-06-15
Fix for #9849

asimonf:
This fixed it for me on my project, and to me it looks logical that the ordering should be like proposed, but obviously I'd like to hear your thoughts on it.


derrabus:
Thank you. Please add a test that reproduces the bug you're attempting to fix.

mpdude:
@asimonf are you still interested in solving this, and if so, can you provide a test case that demonstrates the issue?

PR #9851, opened 2022-06-15
Migrate more code to PHP8

greg0ire:


lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -77,7 +69,7 @@ protected function skipClass(ClassMetadata $metadata)
     /**
      * {@inheritDoc}
      */
-    protected function createProxyDefinition($className)
+    protected function createProxyDefinition($className): ProxyDefinition
```
greg0ire: @derrabus I can't use a native type declaration here unless we update the parent class first, but as `doctrine/common` still supports 7.1, that would be a BC-break. Should we create branch 4.0.x, drop support for PHP 7 in it, and migrate `doctrine/common` to PHP 8 syntax?

ostrolucky: no need for major version when dropping 7.1 IMHO

greg0ire: Yes, but adding a parameter type hint is a BC-break for callers , and adding a return type declaration is a BC-break for extending classes, so there will be a need for a major release, won't there be?

lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -51,23 +43,23 @@ class ProxyFactory extends AbstractProxyFactory
      *                                             values are constants of {@see AbstractProxyFactory}.
      * @psalm-param bool|AutogenerateMode $autoGenerate
      */
-    public function __construct(EntityManagerInterface $em, $proxyDir, $proxyNs, $autoGenerate = AbstractProxyFactory::AUTOGENERATE_NEVER)
-    {
+    public function __construct(
+        /** @var EntityManagerInterface The EntityManager this factory is bound to. */
```
derrabus: The `@param` annotation should be enough. I'd remove the `@var` here.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
derrabus: You could add `mixed` as native parameter type though.

greg0ire: @derrabus I need to do that on the parent class first, otherwise there will be a crash.

greg0ire: I'll remove ProxyFactory from this PR so that we can move on, let's deal with it later.

greg0ire:
Argh, it seems the 3.0.x build is broken by https://github.com/doctrine/dbal/pull/5477

I'm working on that here: https://github.com/doctrine/orm/pull/9885

lib/Doctrine/ORM/Tools/DebugUnitOfWorkListener.php
```diff
@@ -40,26 +34,19 @@ class DebugUnitOfWorkListener
      * @param string $file
      * @param string $context
      */
-    public function __construct($file = 'php://output', $context = '')
+    public function __construct(private $file = 'php://output', private $context = '')
```
derrabus: Can we add native types to those parameters/properties?

derrabus (approved):

PR #9852, opened 2022-06-16
Migrate Internal namespace to PHP8

MarcBrillault:
Refs #9772

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -37,11 +37,6 @@ abstract class AbstractHydrator
     protected ?ResultSetMapping $_rsm = null;
 
     /**
-     * The EntityManager instance.
-     */
-    protected EntityManagerInterface $_em;
-
-    /**
      * The dbms Platform instance.
      */
     protected AbstractPlatform $_platform;
@@ -77,10 +72,12 @@ abstract class AbstractHydrator
      */
     protected array $_hints = [];
 
+    protected EntityManagerInterface $_em;
+
     /**
      * Initializes a new instance of a class derived from <tt>AbstractHydrator</tt>.
      */
-    public function __construct(EntityManagerInterface $em)
+    public function __construct(protected EntityManagerInterface $em)
     {
         $this->_em       = $em;
         $this->_platform = $em->getConnection()->getDatabasePlatform();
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ORM/Internal/Hydration/ArrayHydrator.php
```diff
@@ -124,9 +124,7 @@ protected function hydrateRowData(array $row, array &$result): void
                                 $baseElement[$relationAlias][] = $element;
                             }
 
-                            end($baseElement[$relationAlias]);
-
-                            $this->_identifierMap[$path][$id[$parent]][$id[$dqlAlias]] = key($baseElement[$relationAlias]);
+                            $this->_identifierMap[$path][$id[$parent]][$id[$dqlAlias]] = array_key_last($baseElement[$relationAlias]);
```
MarcBrillault, line 127: This causes a new psalm error, that we baselined, because we don't think it's legitimate. But please triple check.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -80,11 +75,14 @@ abstract class AbstractHydrator
...
      */
-    public function __construct(EntityManagerInterface $em)
-    {
-        $this->_em       = $em;
-        $this->_platform = $em->getConnection()->getDatabasePlatform();
-        $this->_uow      = $em->getUnitOfWork();
+    public function __construct(
+        /**
+         * The EntityManager instance.
+         */
```
greg0ire: Let's drop this comment

greg0ire (dismissed):

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -80,11 +75,10 @@ abstract class AbstractHydrator
     /**
      * Initializes a new instance of a class derived from <tt>AbstractHydrator</tt>.
      */
-    public function __construct(EntityManagerInterface $em)
+    public function __construct(protected EntityManagerInterface $_em)
```
derrabus: Let's not pollute the public API with those unfortunate underscores.

derrabus (dismissed):

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
greg0ire: Wasn't the public API already polluted by this (there was a protected property with that name). Unless we are taking named properties into account, the public API is not more polluted than it already was. But maybe we should stop polluted the public API with those, as an additional BC-break?

SenseException: The underscore can be removed in a separate PR to keep this PR small.

SenseException (dismissed):

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
derrabus: My point is that the underscore is being _added_ to the constructor parameter with this PR. We should avoid that. If that means, we cannot use CPP yet for this parameter, so be it.

SenseException (approved):

greg0ire:
Thanks @MarcBrillault !

PR #9859, opened 2022-06-23
Add parentheses to `EntityManager::contains()`

simPod:
Make order of operations more explicit.

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -836,8 +836,10 @@ use function strpos;
     public function contains($entity)
     {
         return $this->unitOfWork->isScheduledForInsert($entity)
-            || $this->unitOfWork->isInIdentityMap($entity)
-            && ! $this->unitOfWork->isScheduledForDelete($entity);
+            || (
+                $this->unitOfWork->isInIdentityMap($entity)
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/EntityManager.php
```diff
@@ -836,8 +836,10 @@ public function getRepository($entityName)
     public function contains($entity)
     {
         return $this->unitOfWork->isScheduledForInsert($entity)
-            || $this->unitOfWork->isInIdentityMap($entity)
-            && ! $this->unitOfWork->isScheduledForDelete($entity);
+            || (
+                $this->unitOfWork->isInIdentityMap($entity)
+                && ! $this->unitOfWork->isScheduledForDelete($entity)
+            );
```
greg0ire, line 842: This is questionable, and should be addressed at the coding standard level, unless for some reason what is not obvious here according to you would be obvious in other places where we combine several boolean operators.

derrabus:
Sorry, but changes like this one are a waste of everybody's time imho.

simPod:
It's possible it's just me, I had to lookup operator precedence to understand the condition. Feel free to close but it caused confusion.

greg0ire:
Sorry but I'm going to close this. Operator precedence between `&&` and `||` is something you should know, really.

simPod:
Keep forgetting what's first 😅

PR #8958, opened 2021-08-24
Try running the tests on Oracle

greg0ire:

PR #9862, opened 2022-06-24
CI for Oracle DB

FaritSlv:
Try running the tests on Oracle


.github/workflows/continuous-integration.yml
```diff
@@ -2,11 +2,7 @@ name: "Continuous Integration"
 
 on:
   pull_request:
-    branches:
-      - "*.x"
```
beberlei: Why is this removed, seems irrelevant?

FaritSlv: Not really. I took the [branch](https://github.com/doctrine/orm/pull/8958) as a basis and it was already deleted there. I'll return when I'm done

tests/Doctrine/Tests/TestUtil.php
```diff
@@ -88,7 +89,8 @@ private static function initializeDatabase(): void
 
         $platform = $privConn->getDatabasePlatform();
 
-        if ($platform->supportsCreateDropDatabase()) {
+        // skip re-create Database in Oracle DB XE
+        if ($platform->supportsCreateDropDatabase() && false === $platform instanceof OraclePlatform) {
```
derrabus:
```suggestion
        if ($platform->supportsCreateDropDatabase() && ! $platform instanceof OraclePlatform) {
```

.github/workflows/continuous-integration.yml
```diff
@@ -73,6 +69,64 @@ jobs:
...
+    #needs: "phpunit-smoke-check"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.4"
+        oracle-version:
+          - "11"
+        dbal-version:
+          - "2.12"
```
derrabus: We don't support DBAL 2.12 anymore. Let's try to make the tests work with DBAL 3 first and if we succeed, we can consider fixing the build for DBAL 2.13. But I'm not really sure if that's worth the effort tbh.

```suggestion
          - "default"
```

.github/workflows/continuous-integration.yml
derrabus: As long as this is WIP, I understand that this change can be helpful. Let's keep this comment here as a reminder that we need to revert it before merging.

.github/workflows/continuous-integration.yml
FaritSlv: > We don't support DBAL 2.12 anymore. Let's try to make the tests work with DBAL 3 first and if we succeed, we can consider fixing the build for DBAL 2.13. But I'm not really sure if that's worth the effort tbh.

I'm basically trying to run tests without errors. There are strange bugs due to Oracle features. For example: Doesn't delete entries in bulk, because it waits for an explicit commit until the session closes, which is why the test fails.

Lots of work here

derrabus: Right, but it doesn't get better if you force the CI to run on abandoned versions of DBAL only. Please apply my change and have the CI run again.

tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
```diff
@@ -722,7 +722,7 @@ public function testSqlResultSetMapping(): void
         self::assertEquals(['name' => 'id', 'column' => 'id'], $mapping['entities'][0]['fields'][0]);
         self::assertEquals(['name' => 'name', 'column' => 'name'], $mapping['entities'][0]['fields'][1]);
         self::assertEquals(['name' => 'status', 'column' => 'status'], $mapping['entities'][0]['fields'][2]);
-        self::assertEquals(['name' => 'phonenumbers.phonenumber', 'column' => 'number'], $mapping['entities'][0]['fields'][3]);
+        self::assertEquals(['name' => 'phonenumbers.phonenumber', 'column' => 'phone'], $mapping['entities'][0]['fields'][3]);
```
FaritSlv: It's embarrassing that I had to rename a column from a reserved word to pass the test. Most likely I will return and think about how to solve it more correctly

FaritSlv:
I tried several options. Running out of ideas

On the local, I found out that the Oracle DB at some point starts blocking requests or hangs up due to a large number of reads and writes. Tried to disable all possible logging. Autocommits.

The test passes on the github, because the process hangs for a long time, then the database comes to life and goes on. And so several times, because of this for so long

@greg0ire,  any ideas where else to go?

FaritSlv:
image
Stuck on the 1080th test, eventually restarted the database container

FaritSlv:
Individually, all tests pass quickly.

greg0ire:
> @greg0ire, any ideas where else to go?

No, the only experience with Oracle I have is with the PR you started from.

FaritSlv:
Hi
@derrabus, @morozov, @greg0ire 

After adding this fix https://github.com/doctrine/dbal/pull/768, there may have been a profit, but only if you run each test separately. But when all the tests are done, it eats up all the resources of Oracle, and this is very sensitive for the Express Edition. If you return it to joins, then the tests will pass even locally. There is something to think about, for example, tests pass more or less quickly, but one test freezes for 40 minutes "\Doctrine\Tests\ORM\Functional\SchemaTool\CompanySchemaTest::testGeneratedSchema", as I understand it, the bottleneck is getting metadata. I think Oracleplatform can be improved in dbal

morozov:
The optimization made in https://github.com/doctrine/dbal/pull/768 may be irrelevant as of https://github.com/doctrine/dbal/pull/5268. The Oracle's system views indeed take a while to initialize but this has been mitigated by querying them once for the entire schema introspection as opposed to querying them once per table.

We can consider reverting the previous optimization and using joins instead of sub-queries.

FaritSlv:
> The optimization made in [doctrine/dbal#768](https://github.com/doctrine/dbal/pull/768) may be irrelevant as of [doctrine/dbal#5268](https://github.com/doctrine/dbal/pull/5268). The Oracle's system views indeed take a while to initialize but this has been mitigated by querying them once for the entire schema introspection as opposed to querying them once per table.
> 
> We can consider reverting the previous optimization and using joins instead of sub-queries.

In principle, the same queries that I wrote for run locally. I'll try to use the 3.4.x branch now and I think the speed of the test will increase significantly.

FaritSlv:
@morozov With errors, but no longer 30 minutes
image

FaritSlv:
> @morozov With errors, but no longer 30 minutes
> image
image
 Resolve errors

tests/Doctrine/Tests/ORM/Mapping/AbstractMappingDriverTest.php
morozov: It's a known issue: https://github.com/doctrine/dbal/issues/4357.

FaritSlv: Revert use reserved words

ci/github/phpunit/oci8.xml
```diff
@@ -0,0 +1,53 @@
...
+    <php>
+        <ini name="error_reporting" value="-1" />
+
+        <var name="db_driver" value="oci8"/>
+        <var name="db_host" value="localhost"/>
+        <var name="db_user" value="doctrine"/>
+        <var name="db_password" value="oracle"/>
+        <var name="db_dname" value="XE"/>
+        <var name="db_servicename" value="XEPDB1"/>
+        <var name="db_service" value="yes"/>
```
morozov: The two parameters above aren't used in the DBAL test suite configuration. Are they necessary here?

FaritSlv:
@greg0ire for [php 7.4](https://github.com/FaritSlv/orm/runs/7273023807) and [php 8.0](https://github.com/FaritSlv/orm/runs/7273023840) is ready. Now 8.1 still needs to be completed. All tests are only for version 3.4.x so far

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -2939,9 +2939,13 @@ public function addSqlResultSetMapping(array $resultMapping)
                             if (str_contains($fieldName, '.')) {
                                 [, $fieldName] = explode('.', $fieldName);
                             }
-
                             $resultMapping['entities'][$key]['fields'][$k]['column'] = $fieldName;
                         }
+                        $column = $resultMapping['entities'][$key]['fields'][$k]['column'];
+                        if ($column[0] === '`') {
```
morozov, line 2990: This looks quite hacky. Is it the only place where it's necessary? I assume it's only added here because the corresponding test uses a keyword. Is that test supposed to test a keyword or it happens to be one on Oracle?

Note that there is `Doctrine\DBAL\Schema\Identifier` which implements some similar logic.

ci/github/phpunit/oci8.xml
```diff
@@ -0,0 +1,53 @@
...
+        </whitelist>
+    </filter>
+
+    <groups>
+        <exclude>
+            <group>performance</group>
+            <group>locking_functional</group>
+        </exclude>
+    </groups>
+</phpunit>
```
morozov: Please add a new line.

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
```diff
@@ -402,30 +410,37 @@ public function addNamedNativeQueryEntityResultMapping(ClassMetadataInfo $classM
                     [$relation, $fieldName] = explode('.', $fieldName);
                 }
 
+                $columnNameCasing = $field['column'];
+                $test = isset($field['quoted']) && !$field['quoted'];
```
morozov: The result seems to be unused.

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
```diff
@@ -402,30 +410,37 @@ public function addNamedNativeQueryEntityResultMapping(ClassMetadataInfo $classM
...
+                if (!isset($field['quoted']) || !$field['quoted']) {
```
morozov: The expression could be simplified to `!empty($field['quoted'])`.

tests/Doctrine/Tests/Models/CMS/CmsUser.php
```diff
@@ -454,7 +454,7 @@ public static function loadMetadata(ClassMetadata $metadata): void
                             ],
                             [
                                 'name'      => 'phonenumbers.phonenumber',
-                                'column'    => 'number',
+                                'column'    => '`number`',
```
morozov, line 453: Is this test about dealing with keywords? If not, would it make sense to use some other column name?

tests/Doctrine/Tests/Models/VersionedManyToOne/Article.php
```diff
@@ -14,7 +14,7 @@
 
 /**
  * @Entity
- * @Table(name="versioned_many_to_one_article")
+ * @Table(name="ver_m2o_article") // fix Max identifier length for OracleDB
```
morozov: This comment isn't necessary here. It could be applied to any other object name as well.

tests/Doctrine/Tests/Models/VersionedManyToOne/Category.php
```diff
@@ -13,7 +13,7 @@
 
 /**
  * @Entity
- * @Table(name="versioned_many_to_one_category")
+ * @Table(name="ver_m2o_category") // fix Max identifier length for OracleDB
```
morozov: This comment isn't necessary here.

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -946,4 +950,60 @@ final protected function getLastLoggedQuery(int $index = 0): array
...
+    {
+        if ($platform instanceof OraclePlatform || $platform instanceof DB2Platform) {
+            return strtoupper($key);
+        }
+
+        return $key;
+    }
+
+    /**
+     * Due to autocommit in oracle database, all data in temporary tables for the current session is dropped
```
morozov, line 996: What are the temporary tables? Why isn't this necessary in the DBAL test suite?

tests/Doctrine/Tests/ORM/Functional/DatabaseDriverTest.php
```diff
@@ -230,6 +232,7 @@ private static function supportsUnsignedInteger(AbstractPlatform $platform): boo
         // NOTE: PostgreSQL and SQL SERVER do not support UNSIGNED integer
```
morozov: This comment is no longer accurate. Please reword to eliminate the duplication with the actual code.

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
```diff
@@ -724,24 +725,27 @@ public function testFindByAssociationArray(): void
...
 
+        $loggedQuery        = $this->getLastLoggedQuery();
+        $loggedQuery['sql'] = strtolower($loggedQuery['sql']);
+
         if (! class_exists(LoggingMiddleware::class)) {
             // DBAL 2 logs queries before resolving parameter positions
             self::assertSame(
                 [
-                    'sql' => 'SELECT t0.id AS id_1, t0.country AS country_2, t0.zip AS zip_3, t0.city AS city_4, t0.user_id AS user_id_5 FROM cms_addresses t0 WHERE t0.user_id IN (?)',
+                    'sql' => strtolower('SELECT t0.id AS id_1, t0.country AS country_2, t0.zip AS zip_3, t0.city AS city_4, t0.user_id AS user_id_5 FROM cms_addresses t0 WHERE t0.user_id IN (?)'),
```
morozov: How come a functional test makes assertions against SQL? It's pretty pointless, I'd remove it instead.

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -946,4 +950,60 @@ final protected function getLastLoggedQuery(int $index = 0): array
...
+    {
+        $sql = ' LIMIT';
+        if ($platform instanceof OraclePlatform) {
+            $sql = ' WHERE ROWNUM <=';
+        }
+
+        return $sql . ' ' . $limit;
+    }
+
+    public function getDiscriminatorByPlatform($key, AbstractPlatform $platform): string
```
morozov, line 986: This looks like a hack. Aren't there non-test APIs for this? It shouldn't be public anyways.

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -946,4 +950,60 @@ final protected function getLastLoggedQuery(int $index = 0): array
 
         return $lastQuery;
     }
+
+    public function getLimitSQLByPlatform($limit, AbstractPlatform $platform): string
```
morozov, line 976: There's already an API for building a limit query. Why is another one needed?

tests/Doctrine/Tests/TestUtil.php
```diff
@@ -201,4 +208,61 @@ private static function mapConnectionParameters(array $configuration, string $pr
 
         return $parameters;
     }
+
+    public static function dropAllForOracle($connection) {
```
morozov: All platform-specific code should be implemented in the DBAL, not in the ORM test suite.

FaritSlv: This code was added only by for Express Edition, since version dbal 3.4 it can most likely be removed already.

tests/Doctrine/Tests/OrmFunctionalTestCase.php
FaritSlv: I didn't make an API. Yes, I know this functionality exists. This was necessary for the correct construction of native queries.

tests/Doctrine/Tests/OrmFunctionalTestCase.php
FaritSlv: Upper case problem

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
FaritSlv: Unfortunately, I did not write these tests. I can't say why they decided to add them. I just brought it to mind for comparison, to. Oracle all converts to uppercase unless explicitly specified.

tests/Doctrine/Tests/OrmFunctionalTestCase.php
FaritSlv: `CREATE GLOBAL TEMPORARY TABLE`

\Doctrine\DBAL\Platforms\OraclePlatform::getCreateTemporaryTableSnippetSQL

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
FaritSlv: remove

ci/github/phpunit/oci8.xml
FaritSlv: not sure

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
morozov: @greg0ire these are more or less the same queries that you're working on in https://github.com/doctrine/orm/pull/9890. Please consider the removal of the tests that cause issues instead of wasting time on maintaining them.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
FaritSlv: As a matter of fact only for the reserved words the decision. The `ClassMetadataInfo` class has this check in many places. We need to look at `Doctrine\DBAL\Schema\Identifier `and rewrite if this is what we need

tests/Doctrine/Tests/OrmFunctionalTestCase.php
morozov: > I didn't make an API.

The question isn't who made this API. The question is, is it worth butchering the already bad test suite just in order to have the build pass, and what value this new code provides.

FaritSlv: good question

FaritSlv: The new code only gives a well-formed SQL query

tests/Doctrine/Tests/Models/CMS/CmsUser.php
FaritSlv: I can't say. But if you use an alias for a reserved word column, there will be an error.

ci/github/phpunit/oci8.xml
FaritSlv: For some reason, without a servicename, it does not connect to the database. The dbal uses another container with version 11, I have a container with version 18

morozov: I can run DBAL tests locally against a `:18-slim` image as in this PR w/o these parameters. The DBAL tests against images the `:11` and `:21` image tags. It doesn't specify the `APP_USER` and `APP_USER_PASSWORD` variables in the container configuration.

Please make the configuration consistent with the DBAL.

tests/Doctrine/Tests/OrmFunctionalTestCase.php
morozov: The test is an example of a program that solves a problem using the API being tested. Does it mean that every user of the ORM now needs to implement such a function?

FaritSlv: Yes, that's right, since now the fields are case-sensitive

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
FaritSlv: `Doctrine\DBAL\Schema\Identifier` is not suitable for solving the problem

FaritSlv:
@greg0ire for OCI8, I checked all the tests. Added Oracle 11 and Oracle 21. How do you think for pdo_oci, too?
I tried to run it locally and got commit errors with transactions.

greg0ire:
I don't think it's worth testing with many drivers… if anything, I think we might want to reconsider having so many builds for existing platforms. Here, we're introducing 6 new jobs, I would drop the PHP 8.0 jobs for instance.

FaritSlv:
> I don't think it's worth testing with many drivers… if anything, I think we might want to reconsider having so many builds for existing platforms. Here, we're introducing 6 new jobs, I would drop the PHP 8.0 jobs for instance.

Well then it should be done for all jobs

greg0ire:
It would be great to have per-job statistics showing how much one job failed when the other ones didn't… right now I doubt we have much Oracle-specific code in the repository. Codecov seems to say we don't. On the other hand, we have tests that run only for MySQL, which whould justify having more jobs for that kind of platform.

.github/workflows/continuous-integration.yml
```diff
@@ -73,6 +69,78 @@ jobs:
...
+    needs: "phpunit-smoke-check"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.4"
+        oracle-version:
+          - "11"
+        dbal-version:
+          - "3@dev"
```
derrabus: Are we waiting for a release of doctrine/dbal#5466 here? Or is there anything else that prevents us from using a stable DBAL for the tests?

FaritSlv: I think we need to wait until https://github.com/doctrine/dbal/pull/5268 is released.

greg0ire: @FaritSlv it has been released :)

FaritSlv: Ok, need to update branch.

How do I change pr to 2.13?

greg0ire: Looks like you found by yourself :)

FaritSlv:
@greg0ire, I apologize, when resolve merge-conflicts I did not see the connection to oracle. Please run action

greg0ire:
Can you create a separate PR with a non-controversial commit? That way, once it is merged, you will not have to wait for us to click the "run action" button on this PR as you will be trusted.

FaritSlv:
> Can you create a separate PR with a non-controversial commit? That way, once it is merged, you will not have to wait for us to click the "run action" button on this PR as you will be trusted.

Ок

FaritSlv:
> Can you create a separate PR with a non-controversial commit? That way, once it is merged, you will not have to wait for us to click the "run action" button on this PR as you will be trusted.

Resolve 
[Static Analysis](https://github.com/FaritSlv/orm/actions/runs/3108654947), [Continuous Integration](https://github.com/FaritSlv/orm/actions/runs/3108654950), [Performance benchmark](https://github.com/FaritSlv/orm/actions/runs/3108654952), [Coding Standards](https://github.com/FaritSlv/orm/actions/runs/3108654991)

.github/workflows/continuous-integration.yml
FaritSlv: > As long as this is WIP, I understand that this change can be helpful. Let's keep this comment here as a reminder that we need to revert it before merging.

returned as it was

lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
```diff
@@ -756,11 +756,15 @@ private function getOrderingSql(Criteria $criteria, ClassMetadata $targetClass):
...
 
     /** @throws DBALException */
-    private function getLimitSql(Criteria $criteria): string
+    private function getLimitSql(string $sql, Criteria $criteria): string
     {
         $limit  = $criteria->getMaxResults();
         $offset = $criteria->getFirstResult();
 
-        return $this->platform->modifyLimitQuery('', $limit, $offset ?? 0);
+        if ($limit !== null || $offset !== null) {
```
derrabus, line 764: Why is this necessary? Also note that with regards to `$offset`, zero and `null` should be treated equally. In fact, `null` should be deprecated by now.

tests/Doctrine/Tests/Models/CMS/CmsUser.php
derrabus: I think we should rename the column.

tests/Doctrine/Tests/Models/OneToOneInverseSideLoad/InverseSide.php
```diff
@@ -13,7 +13,7 @@
 
 /**
  * @Entity()
- * @Table(name="one_to_one_inverse_side_load_inverse")
+ * @Table(name="o2o_side_inverse")
```
derrabus, line 16: Why do we need to rename these columns?

lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
FaritSlv: Is it better to check if it is greater than zero?

tests/Doctrine/Tests/Models/OneToOneInverseSideLoad/InverseSide.php
FaritSlv: [Because of the maximum identifier length](https://docs.oracle.com/en/database/oracle/oracle-database/21/odpnt/EFCoreIdentifier.html#GUID-FA43F1A1-EDA2-462F-8844-45D49EF67607) for OracleDB 11

FaritSlv:
@derrabus, what else is required of me?

lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
derrabus: No, it would be better to not change this line. I want to understand why you do.

derrabus: And *if* we need this change, make sure that the values `0` and `null` for `$offset` create the same query. I was like this before your change.

derrabus:
Sorry that this PR takes so long to get merged. The main issue is that the PR does not only add Oracle to the build matrix, it also changes a lot of code that at first glance is not specific to Oracle. This makes the changes hard to understand for us.

FaritSlv:
> Sorry that this PR takes so long to get merged. The main issue is that the PR does not only add Oracle to the build matrix, it also changes a lot of code that at first glance is not specific to Oracle. This makes the changes hard to understand for us.

@derrabus, ok, i'm all understanding. Won't be surprised if they don't merge this PR, because very big

derrabus:
One thing you can do is to isolate changes and submit them as individual PRs. Smaller, focused PRs are usually easier to review and get merged more quickly.

FaritSlv:
> One thing you can do is to isolate changes and submit them as individual PRs. Smaller, focused PRs are usually easier to review and get merged more quickly.

So the problem is that all the changes are related. I will think how to split

PR #9866, opened 2022-06-28
Migrate to Doctrine CS 10.0

derrabus:
I've opened this draft PR to monitor how Doctrine CS 10.0 at its current state would affect our codebase.

The biggest change that we can see currently is that a lot more multi-line doc blocks are collapsed to a single line.

cc @simPod @greg0ire

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
     },
     "require-dev": {
         "doctrine/annotations": "^1.13",
-        "doctrine/coding-standard": "^9.0",
+        "doctrine/coding-standard": "9.0.0 || ^10@dev",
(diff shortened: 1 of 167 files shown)


derrabus:
Apparently, PHPUnit 7 does not understand those collapsed doc blocks. The failing tests that we see on the PHP 7.1 run are annotated with

```php
/** @requires PHP 8.1 */
```

derrabus:
I've reverted the single-line annotations on the PHPUnit tests and configured the exclusion rules for the corresponding sniff accordingly.

derrabus:
Replaced by #10009

PR #9870, opened 2022-06-28
[GH-9807] Fix: initialize potentially empty collections at the hydration complete

### LINKED ISSUE #9807: Collection is hydrated as empty if first row contains nulls from a leftJoin
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.3

#### Summary

When the first row of a leftJoin result contains null values (because of a SqlFilter), the collection is hydrated as empty.

#### Current behavior

The collection is hydrated as empty if the first row is nullish.

#### How to reproduce

Details for reproduction is in https://github.com/freiondrej-lmc/doctrine_orm_bug_reproduction .

#### Expected behavior

The Collection should be hydrated with existing data.

popov-a-e:
Closes https://github.com/doctrine/orm/issues/9807

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
@@ -44,6 +44,9 @@ class ObjectHydrator extends AbstractHydrator
     /** @var mixed[] */
     private $initializedCollections = [];
 
+    /** @var array<string, PersistentCollection> */
+    private $uninitializedCollections = [];
+
     /** @var mixed[] */
     private $existingCollections = [];
 
@@ -112,10 +115,11 @@ class ObjectHydrator extends AbstractHydrator
 
         parent::cleanup();
 
-        $this->identifierMap          =
-        $this->initializedCollections =
-        $this->existingCollections    =
-        $this->resultPointers         = [];
+        $this->identifierMap            =
+        $this->initializedCollections   =
+        $this->uninitializedCollections =
+        $this->existingCollections      =
+        $this->resultPointers           = [];
 
         if ($eagerLoad) {
             $this->_uow->triggerEagerLoads();
@@ -126,10 +130,11 @@ class ObjectHydrator extends AbstractHydrator
 
     protected function cleanupAfterRowIteration(): void
     {
-        $this->identifierMap          =
-        $this->initializedCollections =
-        $this->existingCollections    =
-        $this->resultPointers         = [];
+        $this->identifierMap            =
+        $this->initializedCollections   =
+        $this->uninitializedCollections =
+        $this->existingCollections      =
+        $this->resultPointers           = [];
     }
 
     /**
@@ -146,6 +151,12 @@ class ObjectHydrator extends AbstractHydrator
         // Take snapshots from all newly initialized collections
         foreach ($this->initializedCollections as $coll) {
             $coll->takeSnapshot();
+        }
+
+        foreach ($this->uninitializedCollections as $coll) {
+            if (! $coll->isInitialized()) {
+                $coll->setInitialized(true);
+            }
         }
 
         return $result;
@@ -411,8 +422,8 @@ class ObjectHydrator extends AbstractHydrator
                         }
                     } elseif (! $reflFieldValue) {
                         $this->initRelatedCollection($parentObject, $parentClass, $relationField, $parentAlias);
-                    } elseif ($reflFieldValue instanceof PersistentCollection && $reflFieldValue->isInitialized() === false) {
-                        $reflFieldValue->setInitialized(true);
+                    } elseif ($reflFieldValue instanceof PersistentCollection && $reflFieldValue->isInitialized() === false && ! isset($this->uninitializedCollections[$oid . $relationField])) {
+                        $this->uninitializedCollections[$oid . $relationField] = $reflFieldValue;
                     }
                 } else {
                     // PATH B: Single-valued association
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9807Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9807Test.php
@@ -1,0 +1,129 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Internal\Hydration\ObjectHydrator;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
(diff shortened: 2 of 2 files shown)


popov-a-e:
Anybody alive here? 🙂

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9807Test.php
```diff
@@ -0,0 +1,134 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->_schemaTool->createSchema(
+            [
+                $this->_em->getClassMetadata(GH9807Main::class),
+                $this->_em->getClassMetadata(GH9807Join::class),
+            ]
+        );
```
greg0ire: Please use `OrmFunctionalTestCase::createSchemaForModels` instead.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9807Test.php
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * @ORM\ManyToMany(targetEntity="GH9807Join", inversedBy="starts")
+     *
+     * @var Collection<int, GH9807Join>
+     */
+    private $joins;
+
+    /**
+     * @return Collection
```
greg0ire:
```suggestion
     * @return Collection<int, GH9807Join>
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH9807Test.php
```diff
@@ -0,0 +1,134 @@
...
+ */
+class GH9807Main
+{
+    /**
+     * @var int
+     * @Column(type="integer")
+     * @Id
+     * @GeneratedValue
+     */
+    protected $id;
```
greg0ire: Why are ids protected in both entities?

greg0ire:

popov-a-e:
Thanks for the comments @greg0ire, I've fixed it

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
```diff
@@ -44,6 +44,9 @@ class ObjectHydrator extends AbstractHydrator
     /** @var mixed[] */
     private $initializedCollections = [];
 
+    /** @var mixed[] */
```
greg0ire: This is correct but I think you can be more precise.

```suggestion
    /** @var array<string, PersistentCollection> */
```

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
```diff
@@ -148,6 +151,12 @@ protected function hydrateAllData()
             $coll->takeSnapshot();
         }
 
+        foreach ($this->uninitializedCollections as $coll) {
+            if (! $coll->isInitialized()) {
```
greg0ire, line 157: > the collection will be considered "existed", and "existed" collections are only replaced if REFRESH hint is present

Did you mean to write "initialized"?

greg0ire (commented):
I forgot to mention that in the contribution guidelines, but can you wrap the body of your commit message (which is great, good job!) at 72 chars?

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
popov-a-e: I thought about this ofc but didn't want to differ from the same other collections. This file is to refactor anyway, aint' it?

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
popov-a-e: No, there is such a logic that if a Collection is PersistentCollection and initialized it is added to "existing"
```
else {
            // Is already PersistentCollection, and DON'T REFRESH or FETCH-JOIN!
            $this->existingCollections[$oid . $fieldName] = $value;
 ```

File: lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php (hunk 4 above)
Comment (greg0ire) [reply lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php]: Yes, but I'm likely to be the one doing it. Please don't give me more work :pray:

Comment (popov-a-e) [reply lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php]: Huh, OK, no problem

File: lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php (hunk 5 above)
Comment (greg0ire) [reply lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php:157]: I see 👍; please use "existing" then.

**Comment by popov-a-e** [comment]
> I forgot to mention that in the contribution guidelines, but can you wrap the body of your commit message (which is great, good job!) at 72 chars?

Hm, do you mean leaving only the headline? 72 chars is not enough to unfold the reasoning

**Comment by greg0ire** [comment]
> Hm, do you mean leaving only the headline? 

No, I mean wrapping as in adding a linebreak when a line is 72 chars long. With vim properly configured, you can do that with `gqq`

**Review by greg0ire — DISMISSED** [review]

**Comment by popov-a-e** [comment]
Got it, thanks for the hint. I misunderstood "wrapping" for "reducing"
I use PHPStorm however. Consider me casual developer for not using vim :D

**Review by SenseException — DISMISSED** [review]

File: lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php [hunk 6]
```diff
@@ -148,6 +151,12 @@ protected function hydrateAllData()
             $coll->takeSnapshot();
         }
 
+        foreach ($this->uninitializedCollections as $coll) {
```
beberlei, line 156: I believe you need to reset this in cleanup()

beberlei (commented):
Good from my first review

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
popov-a-e: Indeed, makes sense

SenseException: Shouldn't there be a test for the cleanup of the new property then?

popov-a-e: I believe that the details of the implementation are not something that needs to be tested by itself. There shouldn't be tests that break for on a correctly working class with an unchanged interface. Like there are no tests for other collections or private fields that are getting reset, for example `$existingCollections` or `$initializedCollections`.

popov-a-e: Tests for checking if collections are getting initialized or not in different circumstances are nice to have but those are not related to the original issue.

SenseException (approved):

greg0ire:
Thanks @popov-a-e !

Issue #9875, opened 2022-07-02
Change parent classes of some events


franmomu:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes and no

#### Summary

Ref: https://github.com/doctrine/orm/issues/4078 and https://github.com/doctrine/orm/pull/1144

`OnClearEventArgs`, `OnFlushEventArgs`, `PostFlushEventArgs` and `PreFlushEventArgs` can be updated keeping BC and `PreUpdateEventArgs` must be done in `3.x`.

- [x] [Update `OnClearEventArgs`, `OnFlushEventArgs`, `PostFlushEventArgs` and `PreFlushEventArgs`](https://github.com/doctrine/orm/pull/9876)
- [x] [Deprecate `LifecycleEventArgs ` in 2.x](https://github.com/doctrine/orm/pull/9906)
- [ ] Update `PreUpdateEventArgs` in 3.x

I'll create the PRs.

derrabus:
Do you want to continue your work on the 3.0.x branch?

franmomu:
Yeah sure, I tried, but bumped into https://github.com/doctrine/orm/pull/9939

PR #9877, opened 2022-07-02
Deprecate getEntityManager and getEntity from PreUpdateEventArgs

franmomu:
Ref: https://github.com/doctrine/orm/issues/9875

In 3.0, PreUpdateEventArgs will extend PreUpdateEventArgs from `doctrine/persistence` instead of `Doctrine\ORM\Event\LifecycleEventArgs` and those methods won't be available anymore.

This will ease the upgrade to 3.0


derrabus:
Shouldn't we rather deprecate `Doctrine\ORM\Event\LifecycleEventArgs` instead?

franmomu:
Closed in favor of https://github.com/doctrine/orm/pull/9906

Issue #9878, opened 2022-07-03
QueryBuilder function "createNamedParameter" like in doctrine/dbal


Digi92:
### Feature Request

Hello,
i noticed that the function "createNamedParameter" does not exist in the [QueryBuilder](https://github.com/doctrine/orm/blob/2.12.x/lib/Doctrine/ORM/QueryBuilder.php), unlike the [QueryBuilder](https://github.com/doctrine/dbal/blob/3.3.x/src/Query/QueryBuilder.php#L1463) from "doctrine/dbal".
Is there a reason why it is not available in "doctrine/orm"?
Otherwise I can create a pull request.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Documentation of the function "createNamedParameter": https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/query-builder.html#binding-parameters-to-placeholders

greg0ire:
I don't know why it wasn't added, but if we were to add `createNamedParameter`, I think it would make sense to add `createPositionalParameter` as well, wouldn't it?

Personally, I've never used `createNamedParameter`, I don't have an opinion for or against it right now.

PR #9883, opened 2022-07-07
Added the functions "createNamedParameter" and "createPositionalParameter"...

Digi92:
... to the class "QueryBuilder"

The two functions "createNamedParameter" and "createPositionalParameter", from the "QueryBuilder" in the package "doctrine/dbal", copied into the ORM "QueryBuilder" class and adapted to ORM.

See [#9878](https://github.com/doctrine/orm/issues/9878)

### DIFF
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -6,6 +6,8 @@ namespace Doctrine\ORM;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Criteria;
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Types\Type;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Query\Expr;
 use Doctrine\ORM\Query\Parameter;
@@ -148,6 +150,13 @@ class QueryBuilder
 
     /** @var int */
     protected $lifetime = 0;
+
+    /**
+     * The counter of bound parameters used with {@see createNamedParameter) and {@see createPositionalParameter).
+     *
+     * @var int
+     */
+    private $boundCounter = 0;
 
     /**
      * Initializes a new <tt>QueryBuilder</tt> that uses the given <tt>EntityManager</tt>.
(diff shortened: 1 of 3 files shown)


Digi92:
Hi,
i just looked at the errors from the failed psalm test.
It says that the $type of "setParameter" cannot contain "\Doctrine\DBAL\Types\Type" as @param type. In the annotation there is no "\Doctrine\DBAL\Types\Type" in the @param but in the comment next to it.
[Code Link](https://github.com/doctrine/orm/blob/f79ec43e70eff026af23adb0bcfdf4df13d54036/lib/Doctrine/ORM/QueryBuilder.php#L539)
```
...
* @param string|int|null $type  ParameterType::* or \Doctrine\DBAL\Types\Type::* constant
*
* @return $this
*/
...
```

I am now a little confused.
Should the type be added to the @param types or must the commentary be removed here and the annotation of "createNamedParameter" and "createPositionalParameter" be adjusted?

thanks for the clarification

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -1506,6 +1514,74 @@ public function __toString()
...
+     *
+     * @link http://www.zetacomponents.org
+     *
+     * @param mixed           $value
+     * @param int|string|null $type        ParameterType::* or \Doctrine\DBAL\Types\Type::* constant
+     * @param string          $placeHolder The name to bind with. The string must start with a colon ':'.
+     *
+     * @return string the placeholder name used.
+     */
+    public function createNamedParameter($value, $type = ParameterType::STRING, $placeHolder = null)
```
derrabus: Please use native parameter and return types when possible.

derrabus (changes requested):
Thank you for your PR. This is a new feature and thus needs to target the next minor release. Please rebase your PR to 2.13.x.

lib/Doctrine/ORM/QueryBuilder.php
Digi92: Hi,
i rebased to 2.13.x and changed the default parameter value.

DemianShtepa:
@derrabus, sorry for disturbing, are there any updates for this?

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -1519,6 +1526,74 @@ public function __toString()
...
+     * $value = 2;
+     * $q->eq( 'id', $q->createNamedParameter( $value ) );
+     * $stmt = $q->executeQuery(); // executed with 'id = 2'
+     * </code>
+     *
+     * @link http://www.zetacomponents.org
+     *
+     * @param mixed           $value
+     * @param int|string|null $type        \Doctrine\DBAL\ParameterType::* or \Doctrine\DBAL\Types\Type::* constant
+     * @param string          $placeHolder The name to bind with. The string must start with a colon ':'.
```
derrabus, line 1551: The type has to be `string|null`. Since it's the default, `null` is apparently a valid argument.

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -1519,6 +1526,74 @@ public function __toString()
...
+     *
+     * @return string The placeholder name used.
+     */
+    public function createNamedParameter($value, $type = 'string', $placeHolder = null)
```
derrabus, line 1555:
```suggestion
    public function createNamedParameter($value, $type = 'string', ?string $placeHolder = null)
```

lib/Doctrine/ORM/QueryBuilder.php
```diff
@@ -1519,6 +1526,74 @@ public function __toString()
...
+     * name ':dcValue1', ':dcValue2' etc.
+     *
+     * Example:
+     * <code>
+     * $value = 2;
+     * $q->eq( 'id', $q->createNamedParameter( $value ) );
+     * $stmt = $q->executeQuery(); // executed with 'id = 2'
+     * </code>
+     *
+     * @link http://www.zetacomponents.org
```
derrabus, line 1547: Why do we link to zetacomponents.org?

derrabus (changes requested):
I'm sorry nobody has picked up your PR yet. Let's try to get it ready for 2.14.

PR #9884, opened 2022-07-07
fix: attribute association override default values not set for joinColumns and inverseJoinColumns if joinTable is set for a many-to-many association

alanpoulain:
This PR fixes a bug when an attribute association override is used for a many-to-many association with only a join table set.
In this case, `joinColumns` and `inverseJoinColumns` will be set to an empty array:
https://github.com/doctrine/orm/blob/f79ec43e70eff026af23adb0bcfdf4df13d54036/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php#L487-L488
Later on, in the method `_validateAndCompleteManyToManyMapping`, default values will not be set and the empty arrays will be kept because of `isset`:
https://github.com/doctrine/orm/blob/f79ec43e70eff026af23adb0bcfdf4df13d54036/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L2056-L2064

The first commit adds a failing test. You can see that the configuration works for all the mappings (PHP, XML, annotations...) excepted the attribute mapping.
The second commit fixes it by using `null` instead of empty arrays.

Related PR (introducing the bug): https://github.com/doctrine/orm/pull/9241
@beberlei if you can have a look at this PR since you wrote the code :slightly_smiling_face:

### DIFF
--- a/tests/Doctrine/Tests/Models/DDC964/DDC964Admin.php
+++ b/tests/Doctrine/Tests/Models/DDC964/DDC964Admin.php
@@ -21,6 +21,11 @@ use Doctrine\ORM\Mapping\JoinTable;
  *              inverseJoinColumns=@JoinColumn(name="admingroup_id")
  *          )
  *      ),
+ *      @AssociationOverride(name="organizations",
+ *          joinTable=@JoinTable(
+ *              name="ddc964_users_adminorganizations"
+ *          )
+ *      ),
  *      @AssociationOverride(name="address",
  *          joinColumns=@JoinColumn(
  *              name="adminaddress_id", referencedColumnName="id"
@@ -29,7 +34,7 @@ use Doctrine\ORM\Mapping\JoinTable;
  * })
  */
 #[Entity]
-#[AssociationOverrides([new AssociationOverride(name: 'groups', joinTable: new JoinTable(name: 'ddc964_users_admingroups'), joinColumns: [new JoinColumn(name: 'adminuser_id')], inverseJoinColumns: [new JoinColumn(name: 'admingroup_id')]), new AssociationOverride(name: 'address', joinColumns: [new JoinColumn(name: 'adminaddress_id', referencedColumnName: 'id')])])]
+#[AssociationOverrides([new AssociationOverride(name: 'groups', joinTable: new JoinTable(name: 'ddc964_users_admingroups'), joinColumns: [new JoinColumn(name: 'adminuser_id')], inverseJoinColumns: [new JoinColumn(name: 'admingroup_id')]), new AssociationOverride(name: 'organizations', joinTable: new JoinTable(name: 'ddc964_users_adminorganizations')), new AssociationOverride(name: 'address', joinColumns: [new JoinColumn(name: 'adminaddress_id', referencedColumnName: 'id')])])]
 class DDC964Admin extends DDC964User
 {
     public static function loadMetadata(ClassMetadata $metadata): void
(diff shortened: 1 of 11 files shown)


tests/Doctrine/Tests/Models/DDC964/DDC964Admin.php
```diff
@@ -29,7 +34,11 @@
  * })
  */
 #[Entity]
-#[AssociationOverrides([new AssociationOverride(name: 'groups', joinTable: new JoinTable(name: 'ddc964_users_admingroups'), joinColumns: [new JoinColumn(name: 'adminuser_id')], inverseJoinColumns: [new JoinColumn(name: 'admingroup_id')]), new AssociationOverride(name: 'address', joinColumns: [new JoinColumn(name: 'adminaddress_id', referencedColumnName: 'id')])])]
+#[AssociationOverrides([
+    new AssociationOverride(name: 'groups', joinTable: new JoinTable(name: 'ddc964_users_admingroups'), joinColumns: [new JoinColumn(name: 'adminuser_id')], inverseJoinColumns: [new JoinColumn(name: 'admingroup_id')]),
+    new AssociationOverride(name: 'organizations', joinTable: new JoinTable(name: 'ddc964_users_adminorganizations')),
+    new AssociationOverride(name: 'address', joinColumns: [new JoinColumn(name: 'adminaddress_id', referencedColumnName: 'id')]),
+])]
```
greg0ire: Sadly, you cannot use a multiline attribute here, it will make the build fail with old versions of PHP

greg0ire:
You can use `vendor/bin/phpcbf` to fix many of the cs issues.

greg0ire:
Sorry didn't mean to close :sweat_smile:

tests/Doctrine/Tests/Models/DDC964/DDC964Admin.php
alanpoulain: Right, I forgot!

alanpoulain:
Thanks for the quick review @greg0ire! :slightly_smiling_face:

tests/Doctrine/Tests/Models/DDC964/DDC964Organization.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    private $id;
+
+    /**
+     * @var string|null
+     * @Column
+     */
+    private $name;
+
+    public function __construct($name = null)
```
greg0ire:
```suggestion
    public function __construct(string $name = null)
```

tests/Doctrine/Tests/Models/DDC964/DDC964User.php
```diff
@@ -97,6 +112,19 @@ public function getGroups(): ArrayCollection
...
 
+    public function addOrganization(DDC964Organization $organization): void
+    {
+        $this->organizations->add($organization);
+    }
+
+    /**
+     * @psalm-return Collection<int, DDC964Organization>
+     */
+    public function getOrganizations(): ArrayCollection
```
greg0ire: I think it could be a `PersistentCollection`

```suggestion
    public function getOrganizations(): Collection
```

tests/Doctrine/Tests/Models/DDC964/DDC964Organization.php
alanpoulain: Done it for `DDC964Group` too.

tests/Doctrine/Tests/Models/DDC964/DDC964User.php
alanpoulain: Done it for `getGroups` and for `DDC964Group` too.

dunglas (approved):

chalasr (approved):

PR #9885, opened 2022-07-07
Address API removals from upstream

greg0ire:
https://github.com/doctrine/dbal/pull/5477 has removed 2 things:
- array and object types
- custom schema options


greg0ire:
The removal of custom schema options seems to break the code introduced in https://github.com/doctrine/orm/pull/9655 for some reason…

greg0ire:
@ruudk @morozov maybe you can help with this failure: https://github.com/doctrine/orm/runs/7242015946?check_suite_focus=true#step:7:79

I don't understand what's happening, but I think it must be somehow related to custom schema options. The original PR of @ruudk relied on them: #9636 , but not the second one: #9655

tests/Doctrine/Tests/OrmFunctionalTestCase.php
```diff
@@ -1051,6 +1060,11 @@ final protected function getLastLoggedQuery(int $index = 0): array
         return $lastQuery;
     }
 
+    final protected function assertQueryLog(string ...$expectedQueries): void
+    {
+        self::assertSame($expectedQueries, array_column($this->getQueryLog()->queries, 'sql'));
+    }
```
greg0ire: Introducing that method allows to compare the 2 query logs rather than getting an exception because there is no seventh query.

greg0ire:
> but not the second one: https://github.com/doctrine/orm/pull/9655

Custom options are not mentioned in the code, but they are mentioned in the summary of that PR though :thinking: 

UPD: I think this line is throwing away the extra options: https://github.com/doctrine/orm/blob/291765e879e54fc569fd65e6262d8f5e4c00f04f/lib/Doctrine/ORM/Tools/SchemaTool.php#L799

Before my changes, there was  a separation into known options and custom ones: https://github.com/doctrine/orm/blob/2.12.x/lib/Doctrine/ORM/Tools/SchemaTool.php#L799-L800

Removing that breaks the test suite because of how enum types are implemented, but that might be the way to go, I'll have to investigate why that separation was needed in the first place.

The issue with the test suite is related to #9382 

Now it blows up with `The "enumType" column option is not supported.`, thrown here: https://github.com/doctrine/dbal/blob/2485d0a49dd17d05f54b68bb020eced4242388c8/src/Schema/Column.php#L83-L85

@morozov , what's the upgrade path for this?

UPD: nevermind, I think I've found what needed to be done.

greg0ire:
The build is green, but I'm not setting this as ready for review yet: there are things I can backport, and other things that might be incorrect.

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -786,8 +786,8 @@ private function gatherColumnOptions(array $mapping): array
             return [];
         }
 
-        $options                        = array_intersect_key($mappingOptions, array_flip(self::KNOWN_COLUMN_OPTIONS));
-        $options['customSchemaOptions'] = array_diff_key($mappingOptions, $options);
+        $options                    = array_intersect_key($mappingOptions, array_flip(self::KNOWN_COLUMN_OPTIONS));
+        $options['platformOptions'] = array_diff_key($mappingOptions, $options);
```
greg0ire, line 790: I don't know if I should backport this as well… `Column::getCustomSchemaOptions()` will not return the same thing as previously, not sure how acceptable that is.

morozov: I'd leave it as is in older branches. This API is supported up until DBAL 4.0, so there's no reason to break anything prior to supporting it.

greg0ire:
I've backported what was reasonable to backport; this is now the combination of a merge-up and fixes for the build (which are in separate commits)

morozov (approved):

PR #9886, opened 2022-07-08
Migrate to Doctrine CS 10 on 3.0.x

derrabus:
This is #9866 on 3.0.x. I'd like to keep track of what the new CS standard would do to our precious codebase. 😬


derrabus:
@simPod Apparently, the introduction of the `SlevomatCodingStandard.Functions.RequireTrailingCommaInCall.MissingTrailingComma` sniff causes some trouble for us. If I run PHPCBF, it fails to fix someof the reported errors. Do you have time to investigate that?

```
PHPCBF RESULT SUMMARY
--------------------------------------------------------------------------------------------------
FILE                                                                              FIXED  REMAINING
--------------------------------------------------------------------------------------------------
lib/Doctrine/ORM/Tools/Console/Command/ClearCache/ResultCommand.php               FAILED TO FIX
lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/UpdateCommand.php               FAILED TO FIX
lib/Doctrine/ORM/Tools/Console/Command/ClearCache/QueryRegionCommand.php          FAILED TO FIX
lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/CreateCommand.php               FAILED TO FIX
lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/DropCommand.php                 FAILED TO FIX
lib/Doctrine/ORM/Tools/Console/Command/InfoCommand.php                            FAILED TO FIX
lib/Doctrine/ORM/Tools/Console/Command/ClearCache/CollectionRegionCommand.php     FAILED TO FIX
lib/Doctrine/ORM/Tools/Console/Command/MappingDescribeCommand.php                 FAILED TO FIX
lib/Doctrine/ORM/Tools/Console/Command/ClearCache/MetadataCommand.php             FAILED TO FIX
lib/Doctrine/ORM/Tools/Console/Command/ClearCache/EntityRegionCommand.php         FAILED TO FIX
--------------------------------------------------------------------------------------------------
A TOTAL OF 19 ERRORS WERE FIXED IN 10 FILES
--------------------------------------------------------------------------------------------------
PHPCBF FAILED TO FIX 10 FILES
--------------------------------------------------------------------------------------------------
```

simPod:
@derrabus I think I've already did. Try disabling `Generic.Functions.FunctionCallArgumentSpacing.NoSpaceAfterComma`. It causes conflict for the fixer.

derrabus:
Indeed, that did the trick, thanks. Is this a bug we need to report to PHPCS or would we rather deactivate that rule in the Doctrine CS?

simPod:
I plan to investigate the conflict and then will see what to do about it. Either report origin lib or/and send PR to Doctrine CS.

tests/Doctrine/Tests/ORM/Functional/OneToOneEagerLoadingTest.php
```diff
@@ -402,11 +370,8 @@ class TrainOrder
     public $id;
 
     public function __construct(
-        /**
-         * @var Train
-         * @OneToOne(targetEntity="Train", fetch="EAGER")
-         */
-        public $train
```
derrabus, line 403: This change is a bit unfortunate. `SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint` is triggered here although it's not just a parameter type that is being added here, but also a property type.

simPod: Maybe v8.2 does the trick https://github.com/slevomat/coding-standard/releases/tag/8.2.0

simPod:
@derrabus the issue has been there for a long time so I don't think it's relevant for our CS v10.

https://github.com/squizlabs/PHP_CodeSniffer/issues/3477

There are two squizlabs sniffs conflicting because they have never understood trailing commas.

tests/Doctrine/Tests/ORM/Functional/OneToOneEagerLoadingTest.php
derrabus: No, these changes have been produced with version 8.2.0 already.

simPod: I'm looking at this again and I don't think I understand what the issue is 🤔.

The `@var` seems redundant so it's added as a native TH.

derrabus:
Okay, but by introducing more trailing comma rules, we create a footgun if we don't deactivate that conflicting rule in Doctrine CS, don't we?

simPod:
@derrabus both sniffs are useful I think, it will require a fix in squizlabs/PHP_CodeSniffer

lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -88,7 +88,7 @@ abstract class AbstractQuery
     /**
      * The user-specified ResultSetMapping to use.
      */
-    protected ?ResultSetMapping $_resultSetMapping = null;
+    protected ResultSetMapping|null $_resultSetMapping = null;
```
morozov, line 91: What's the rationale behind enforcing the usage of union types instead of nullable types?

simPod: https://github.com/doctrine/coding-standard/pull/266#discussion_r900753185

morozov: Personally, I believe arbitrary union types (e.g. `getValue(sting $key): int|string|false`) often reflect poor API design while nullable types are fine (e.g. `findUser(int $id): ?User`).

I think that enforcing a feature introduced primarily to support the _existing_ poor design of the PHP's own API and the type system as a one-size-fits-all solution is not the right move.

simPod: Union types are fine. I must disagree that union types often reflect poor API design. I think the premise is too simplified.

derrabus: We have two ways of expressing the same thing: a union of a type and `null`. Normalizing that is reasonable imho. Whether we use the old nullable syntax or the new union syntax does not matter that much to me, as long as we agree on a single way to do it.

And I don't follow that "poor API design" argument either.

morozov:
IMO, union types are often poor because they make the consumer deal with a set of types instead of one by adding conditionals. Without pattern matching, it's error-prone since there's no guarantee that every consumer will handle each type from the union.

Specifically, in the PHP's standard library, union types go hand in hand with the APIs that are known as extremely error-prone, e.g. `if (strpos($str, substr)) ...`) or `while ($value = $stmt->fetch(PDO::FETCH_COLUMN)) ...`. Union types are also needed to deal with array keys because there's no built-in support for maps in PHP and numeric array keys are converted to strings.

All the `int|string` types used in the code that fetches data from the database are because there's no support for unsigned integers and decimals in PHP. Otherwise, it would be possible to implement an API that returns a specific type.

Looking at the DBAL code, the only place I see where a union type is used deliberately is the expression builder where the methods accept strings or expressions but even there they could be split into two: one for stings, the other for expressions.

lib/Doctrine/ORM/AbstractQuery.php
morozov: IMO, union types are often poor because they make the consumer deal with a set of types instead of one by adding conditionals. Without pattern matching, it's error-prone since there's no guarantee that every consumer will handle each type from the union.

Specifically, in the PHP's standard library, union types go hand in hand with the APIs that are known as extremely error-prone, e.g. `if (strpos($str, substr)) ...`) or `while ($value = $stmt->fetch(PDO::FETCH_COLUMN)) ...`. Union types are also needed to deal with array keys because there's no built-in support for maps in PHP and numeric array keys are converted to strings.

All the `int|string` types used in the code that fetches data from the database are because there's no support for unsigned integers and decimals in PHP. Otherwise, it would be possible to implement an API that returns a specific type.

Looking at the DBAL code, the only place I see where a union type is used deliberately is the expression builder where the methods accept strings or expressions but even there they could be split into two: one for stings, the other for expressions.

derrabus: All right, so how do you suggest to proceed here?

morozov: We do not forbid nullable types in the coding standard: https://github.com/doctrine/coding-standard/pull/266#discussion_r900749599.

derrabus: > We do not forbid nullable types in the coding standard

That means, we would allow both ways of expressing nullable types? Or would we let the fixer change `Foo|null` to `?Foo`?

Do you want to open a PR on the coding-standards repo that makes the necessary config adjustments? I think, this change should be discussed there.

simPod: I think different topic is discussed here. The sniff is not about whether or not it is ok to design API with unions. It's not about API at all.

The sniff is about consistent usage of union types. If there are no unions, the sniff does not apply.

If there's union of `Type` and `null`, it should be written as union: `Type|null`. The archaic `?Type` was introduced because of absence of union types in the language. Since there's unions support now, there's no reason to use `?` because it's inconsistent. We already have `null` value and [will have `null` standalone type in 8.2](https://wiki.php.net/rfc/null-false-standalone-types).

morozov: > That means, we would allow both ways of expressing nullable types? Or would we let the fixer change Foo|null to ?Foo?

If the latter is possible, I would prefer that. Otherwise, the former is fine.

> Do you want to open a PR on the coding-standards repo that makes the necessary config adjustments?

No.

derrabus:
Replaced by #10011

PR #9892, opened 2022-07-10
Address array object type deprecation

greg0ire:
These types have been deprecated in https://github.com/doctrine/dbal/pull/5470

This PR should help making #9885 smaller.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Mapping/yaml/Doctrine.Tests.Models.Generic.BooleanModel.dcm.yml
+++ b/tests/Doctrine/Tests/ORM/Mapping/yaml/Doctrine.Tests.Models.Generic.BooleanModel.dcm.yml
@@ -7,7 +7,5 @@
       generator:
         strategy: AUTO
   fields:
-    array:
-      type: array
-    object:
-      type: object
+    booleanField:
+      type: boolean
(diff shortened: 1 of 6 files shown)


tests/Doctrine/Tests/ORM/Mapping/yaml/Doctrine.Tests.Models.Generic.BooleanModel.dcm.yml
```diff
@@ -7,7 +7,5 @@
       generator:
         strategy: AUTO
   fields:
-    array:
-      type: array
-    object:
-      type: object
```
morozov, line 13: As I understand from the other patch, this test doesn't test the array and object types specifically, it tests serialization which can be tested with any type. Correct?

morozov (approved):

tests/Doctrine/Tests/ORM/Mapping/yaml/Doctrine.Tests.Models.Generic.BooleanModel.dcm.yml
greg0ire: From what I understood, it does not test the serialization either, but rather that we have a good error message. I tried to reflect that by changing the name of the method in the third commit.

I came to that conclusion after reading the issue that test is tagged with: https://github.com/doctrine/orm/issues/2098

Also, it does not appear in the diff, but what is wrong with that mapping is the name of the class:

https://github.com/doctrine/orm/blob/291765e879e54fc569fd65e6262d8f5e4c00f04f/tests/Doctrine/Tests/ORM/Mapping/xml/Doctrine.Tests.Models.Generic.SerializationModel.dcm.xml#L7

PR #9895, opened 2022-07-11
Avoid SQL assertions

greg0ire:
doctrine/dbal is the component responsible for generating the queries.
Let us make the test suite more robust by asserting that things work
from a functional point of view.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
@@ -8,10 +8,7 @@ use BadMethodCallException;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\Common\Persistence\PersistentObject;
-use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\LockMode;
-use Doctrine\DBAL\Logging\Middleware as LoggingMiddleware;
-use Doctrine\DBAL\ParameterType;
 use Doctrine\Deprecations\PHPUnit\VerifyDeprecations;
 use Doctrine\ORM\EntityRepository;
 use Doctrine\ORM\Exception\NotSupported;
@@ -34,7 +31,6 @@ use Doctrine\Tests\Models\DDC753\DDC753EntityWithCustomRepository;
 use Doctrine\Tests\Models\DDC753\DDC753EntityWithDefaultCustomRepository;
 use Doctrine\Tests\OrmFunctionalTestCase;
 
-use function array_fill;
 use function array_values;
 use function class_exists;
 use function reset;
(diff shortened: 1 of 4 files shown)


greg0ire:
I didn't rid the entire codebase from calls to `getLastLoggedQuery()`, but did what I could when I understood clearly what to do. Feel free to checkout the branch and propose/push additional changes.

derrabus (dismissed):

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
```diff
@@ -721,29 +717,45 @@ public function testInvalidOrientation(): void
...
-            self::assertSame(
-                [
-                    'sql' => 'SELECT t0.id AS id_1, t0.country AS country_2, t0.zip AS zip_3, t0.city AS city_4, t0.user_id AS user_id_5 FROM cms_addresses t0 WHERE t0.user_id IN (?, ?, ?)',
-                    'params' => [1 => 1, 2 => 2, 3 => 3],
-                    'types' => array_fill(1, 3, ParameterType::INTEGER),
-                ],
-                $this->getLastLoggedQuery()
-            );
-        }
+        $pigs = $repo->findBy(['user' => $ids]);
```
morozov: As far as I understand, this is `$addresses`, not `$pigs`.

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
```diff
@@ -721,29 +717,45 @@ public function testInvalidOrientation(): void
...
+
+        self::assertCount(3, $pigs);
```
morozov: Given that the previous version of the test makes an assertion regarding the query parameters, here, we need to check if the same three addresses have been returned, not just three arbitrary addresses.

tests/Doctrine/Tests/ORM/Functional/OneToOneEagerLoadingTest.php
```diff
@@ -146,24 +146,18 @@ public function testEagerLoadWithNullableColumnsGeneratesLeftJoinOnBothSides():
...
+        $this->_em->persist($driver);
         $this->_em->flush();
         $this->_em->clear();
 
         $train = $this->_em->find(get_class($train), $train->id);
-        $this->assertSQLEquals(
-            'SELECT t0.id AS id_1, t0.driver_id AS driver_id_2, t3.id AS id_4, t3.name AS name_5, t0.owner_id AS owner_id_6, t7.id AS id_8, t7.name AS name_9 FROM Train t0 LEFT JOIN TrainDriver t3 ON t0.driver_id = t3.id INNER JOIN TrainOwner t7 ON t0.owner_id = t7.id WHERE t0.id = ?',
-            $this->getLastLoggedQuery()['sql']
-        );
+        self::assertNotNull($train, 'It should be possible to find the train even though it has no driver');
```
morozov, line 158: Should we check for a specific train, not for an arbitrary one?

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
greg0ire: Ok, I'll add a wolf with a 4th address in there.

derrabus (approved):

PR #9900, opened 2022-07-15
Migrate part of the tools namespace to PHP8 syntax

greg0ire:


lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -72,48 +66,39 @@ public function getQuery()
...
 
         return $this;
     }
 
-    /**
-     * {@inheritdoc}
-     *
-     * @return int
-     */
     #[ReturnTypeWillChange]
```
derrabus: The attribute can be removed now.

greg0ire: Oh right, same below.

greg0ire: I have to look at the files one by one because for instance here, the constructor was not affected by Rector.

lib/Doctrine/ORM/Query/TreeWalker.php
```diff
@@ -16,12 +16,10 @@ interface TreeWalker
...
      * Initializes TreeWalker with important information about the ASTs to be walked.
      *
-     * @param AbstractQuery $query           The parsed Query.
-     * @param ParserResult  $parserResult    The result of the parsing process.
-     * @param mixed[]       $queryComponents The query components (symbol table).
+     * @param mixed[] $queryComponents The query components (symbol table).
      * @psalm-param array<string, QueryComponent> $queryComponents The query components (symbol table).
      */
-    public function __construct($query, $parserResult, array $queryComponents);
+    public function __construct(AbstractQuery $query, ParserResult $parserResult, array $queryComponents);
```
greg0ire: This forces us to use `AbstractQuery` in classes that implement the interface. We can still annotate the parameter with `Query`, but this is kind of weird.

greg0ire:
I think it will be better to touch the walkers after https://github.com/doctrine/orm/pull/9551 is finished. I will remove change to them from this PR

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -124,11 +100,9 @@ public function count()
     /**
      * {@inheritdoc}
      *
-     * @return ArrayIterator
      * @psalm-return ArrayIterator<array-key, T>
      */
-    #[ReturnTypeWillChange]
-    public function getIterator()
+    public function getIterator(): ArrayIterator
```
SenseException: It was already an ArrayIterator before, but should this method rely on an `Iterator` implementation in the return type instead of the interface itself? Even the name of the method doesn't refer to a specific Iterator implementation.

It's not really a blocker for a merge, but food for thoughts for 3.0.

SenseException (dismissed):

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
greg0ire: No, no, I totally see what you mean: if the consumer don't leverage it, why should we provide that level of detail. Here, they shouldn't be calling `getIterator()` directly but use a `foreach`, so that's one more reason to do this. Let me change that.

derrabus (approved):

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
SenseException: Yes, `IteratorAggregate` allows to access `getIterator()` directly. Technically it should be `Traversable` as the docs suggests, but I wasn't sure if this step was a too large for this PR.

PR #9903, opened 2022-07-16
schema tool: remove useless text from --dump-sql output

### LINKED ISSUE #7186: [request] option to only use SQL in dumped schema information
Currently the dumped information of the [UpdateCommand](https://github.com/doctrine/doctrine2/blob/c4c59df1b4251f5030a08bd1253a0e472e4bdb7e/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/UpdateCommand.php#L87) also includes non-SQL text:

_"The following SQL statements will be executed:"_

Is there a specific reason to do this? We currently store the output in a .sql file, which can then be easily executed, for example when spinning up a Docker instance. But the generated SQL is now "invalid" due to the help text. 

Of course this text can be stripped after generation, but still, it feels awkward.

### LINKED ISSUE #8263: `./bin/console doctrine:schema:create --dump-sql` generates invalid SQL
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.3

#### Summary

Running `./bin/console doctrine:schema:create --dump-sql` creates invalid SQL, the results are prepended with the following line `The following SQL statements will be executed:`

It's probably better to insert a comment sign before this line or omit it at all, since the user probably knows what to expect when running with `--dump-sql:` Not some redundant message, but proper SQL.

I'm using DoctrineBundle 2.1.2

Taken from [DoctrineBundle doctrine/dbal#1208] (https://github.com/doctrine/DoctrineBundle/issues/1208)

#### Current behaviour

Some status message is uncommented part of the SQL

#### How to reproduce

`./bin/console doctrine:schema:create --dump-sql`

#### Expected behaviour

Valid SQL, that can be used by `mysql` for example.

glaszig:
the description and semantics of the `--dump-sql` switch
indicate an sql dump. without that useless line

    The following SQL statements will be executed:

the output can actually be used to generate a plain
dump.sql file that can be fed into sql command-
consuming programs.

resolves #8263.
resolves #7186.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/CreateCommandTest.php
+++ b/tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/CreateCommandTest.php
@@ -8,10 +8,14 @@ use Doctrine\ORM\Tools\Console\Command\SchemaTool\CreateCommand;
 
 class CreateCommandTest extends AbstractCommandTest
 {
+    /**
+     * @doesNotPerformAssertions
+     */
     public function testItPrintsTheSql(): void
     {
         $tester = $this->getCommandTester(CreateCommand::class);
         $tester->execute(['--dump-sql' => true]);
-        self::assertStringContainsString('CREATE TABLE keyboard', $tester->getDisplay());
+
+        self::$sharedConn->executeStatement($tester->getDisplay());
     }
 }
--- a/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/CreateCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/CreateCommand.php
@@ -49,11 +49,9 @@ EOT
 
         if ($dumpSql) {
             $sqls = $schemaTool->getCreateSchemaSql($metadatas);
-            $ui->text('The following SQL statements will be executed:');
-            $ui->newLine();
 
             foreach ($sqls as $sql) {
-                $ui->text(sprintf('    %s;', $sql));
+                $ui->writeln(sprintf('%s;', $sql));
             }
 
             return 0;
--- a/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/DropCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/DropCommand.php
@@ -60,11 +60,8 @@ EOT
                 $sqls = $schemaTool->getDropSchemaSQL($metadatas);
             }
 
-            $ui->text('The following SQL statements will be executed:');
-            $ui->newLine();
-
             foreach ($sqls as $sql) {
-                $ui->text(sprintf('    %s;', $sql));
+                $ui->writeln(sprintf('%s;', $sql));
             }
 
             return 0;
--- a/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/UpdateCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/UpdateCommand.php
@@ -88,10 +88,8 @@ EOT
         $force   = $input->getOption('force') === true;
 
         if ($dumpSql) {
-            $ui->text('The following SQL statements will be executed:');
-            $ui->newLine();
             foreach ($sqls as $sql) {
-                $ui->text(sprintf('    %s;', $sql));
+                $ui->writeln(sprintf('%s;', $sql));
             }
         }
 
--- a/tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/DropCommandTest.php
+++ b/tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/DropCommandTest.php
@@ -9,11 +9,15 @@ use Doctrine\Tests\ORM\Tools\Console\Command\SchemaTool\Models\Keyboard;
 
 final class DropCommandTest extends AbstractCommandTest
 {
+    /**
+     * @doesNotPerformAssertions
+     */
     public function testItPrintsTheSql(): void
     {
         $this->createSchemaForModels(Keyboard::class);
         $tester = $this->getCommandTester(DropCommand::class);
         $tester->execute(['--dump-sql' => true]);
-        self::assertStringContainsString('DROP TABLE keyboard', $tester->getDisplay());
+
+        self::$sharedConn->executeStatement($tester->getDisplay());
     }
 }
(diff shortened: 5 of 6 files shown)


greg0ire (dismissed):

derrabus:
Thank you for bringing up this topic.

I don't think that this qualifies a bugfix. Sure, nobody will miss this line, but until now, it was not a requirement that the whole output can be piped to an SQL client. This is why I'd rather ship this as a feature for 2.13 instead.

That being said, *if* we want the output of those commands to be executable, we should add functional tests that make sure that it actually is.

greg0ire:
I'm working on making it easy to add tests at ~https://github.com/doctrine/orm/pull/9903~ #9907

greg0ire:
@glaszig please modify the tests I introduced in ~#9903~ #9907 to assert that the display is valid SQL that can be executed with `Connection::executeStatement()`

glaszig:
i found your additions in #9907 ;)

opted to test the exact output to prevent unwanted things to appear in the future.

> assert that the display is valid SQL

testing the schema dumpers validity should happen in the proper location, not in the command test imo.

i find my changes in the tests a little ugly. would have liked to use heredocs but php heredoc are inferior to ruby's. please advise.

in case you're wondering about the leading space in each line: symfony/symfony@64e7b6f7afc13437105ba51534e1ea61488b6d5d

tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/CreateCommandTest.php
```diff
@@ -12,6 +12,9 @@ public function testItPrintsTheSql(): void
     {
         $tester = $this->getCommandTester(CreateCommand::class);
         $tester->execute(['--dump-sql' => true]);
```
greg0ire, line 17: > we should add functional tests that make sure that it actually is.

I think @derrabus meant asserting that this returns an integer (instead of crashing) : `self::$sharedConn->executeStatement($tester->getDisplay());`

Asserting the SQL generated by another package leads to fragile tests.

greg0ire:
> i find my changes in the tests a little ugly. would have liked to use heredocs but php heredoc are inferior to ruby's. please advise.

I'd use nowdoc anyway, it's better than concatenation IMO.

SenseException:
> That being said, if we want the output of those commands to be executable, we should add functional tests that make sure that it actually is.

I assume it's now settled that `--dump-sql` is now supposed to return executable SQL queries?

glaszig:
> I think @derrabus meant asserting that this returns an integer (instead of crashing): self::$sharedConn->executeStatement($tester->getDisplay());

oh nice. thanks for the pointer. implemented.

greg0ire:

How to do that?

1. `git rebase -i origin/2.13.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/CreateCommandTest.php
```diff
@@ -12,6 +12,9 @@ public function testItPrintsTheSql(): void
...
-        self::assertStringContainsString('CREATE TABLE keyboard', $tester->getDisplay());
+
+        $statements = $tester->getDisplay();
+        self::$sharedConn->executeStatement($statements);
+        self::assertTrue(true);
```
greg0ire: Please use `@doesNotPerformAssertions` instead, here and below, it's better than asserting it returns an integer as I previously suggested, and is certainly better than asserting that true is true.

tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/CreateCommandTest.php
```diff
@@ -8,10 +8,15 @@
...
+     * @doesNotPerformAssertions
+     */
     public function testItPrintsTheSql(): void
     {
         $tester = $this->getCommandTester(CreateCommand::class);
         $tester->execute(['--dump-sql' => true]);
-        self::assertStringContainsString('CREATE TABLE keyboard', $tester->getDisplay());
+
+        $statements = $tester->getDisplay();
+        self::$sharedConn->executeStatement($statements);
```
greg0ire: Will this really work with several statements? Or should you use a foreach?

glaszig: it does not throw any exception with multiple statements. but since it's a bunch of raw queries `exec()` might be more appropriate?

greg0ire: Didn't know this `method()`, but given its phpdoc comment, I'd advise against using it, it seems bound to be removed. Also, take a look at the implementation: you gain nothing by using it. To know if it works with several statement, try running a second statement that is syntactically valid but will break, like deleting a table that does not exist.

glaszig: aha. `exec()` is deprecated and just delegates to `executeStatement()` 👀 

https://github.com/doctrine/dbal/blob/39a05f63ffb95baf13716a6c11e268a720bf2552/src/Connection.php#L1846-L1854

so we're good i guess.

greg0ire (dismissed):
> To know if it works with several statement, try running a second statement that is syntactically valid but will break, like deleting a table that does not exist.

tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/CreateCommandTest.php
glaszig: > To know if it works with several statement, try running a second statement that is syntactically valid but will break, like deleting a table that does not exist.

works.

greg0ire: Now that we've establish that this works, let's get rid of the extra variable, here and below.

```suggestion
        self::$sharedConn->executeStatement($tester->getDisplay());
```

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks for contributing this @glaszig . It was a very welcome change in attitude after your comment on https://github.com/doctrine/orm/issues/8263#issuecomment-1186263119

I appreciate that you decided to take the matter into your own hands, and look forward to your next contributions.

glaszig:
thanks for guiding me through this.

php has not been my daily driver for 10 years. so i didn't even have a proper dev setup. there were quite a few people being bothered by this and this one-line-change, bascically, has not been swept in for years. so for me it was actually even more work. given the contexts, everybody is entitled to their "attitude".

greg0ire:
Yeah it's always going to be more work for an external contributor, and I understand why you might wonder why something hasn't been fixed for years… for maintainers though, most of us spend a lot of time working on the project, and if we were to take upon ourselves responsibility for fixing all the issues, we would have no time left for things like preparing the next major release. It just does not scale. Getting people to make their first contribution is a good way to removing many barriers for the next time, and transforming programming work into just code reviews, which takes a bit less time.

PR #9906, opened 2022-07-17
Deprecate `LifecycleEventArgs`

franmomu:
Ref: https://github.com/doctrine/orm/pull/9877#issuecomment-1178858059

`PreUpdateEventArgs` extends from `LifecycleEventArgs` and in `3.0` will extend from `Doctrine\Persistence\Event\PreUpdateEventArgs`, so there is no need for this class anymore and its usage can be replace with `Doctrine\Persistence\Event\PreUpdateEventArgs`.

I've added deprecations to `LifecycleEventArgs::getEntity()` and `LifecycleEventArgs::getEntityManager()` for someone using `PreUpdateEventArgs` and calling those methods.

While deprecating this, I've seen that we can also deprecate `LoadClassMetadataEventArgs` I guess in favor of the one in `doctrine/persistence`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -59,6 +59,10 @@ This method has been deprecated in:
 
 It will be removed in 3.0. Use `getObjectManager()` instead.
 
+## Deprecated `Doctrine\ORM\Event\LifecycleEventArgs` class
+
+It will be removed in 3.0. Use `Doctrine\Persistence\Event\LifecycleEventArgs` instead.
+
 ## Prepare split of output walkers and tree walkers
 
 In 3.0, `SqlWalker` and its child classes won't implement the `TreeWalker`
--- a/docs/en/cookbook/strategy-cookbook-introduction.rst
+++ b/docs/en/cookbook/strategy-cookbook-introduction.rst
@@ -87,12 +87,12 @@ Such an interface could look like this:
          * @return \Zend_View_Helper_Interface
          */
         public function setView(\Zend_View_Interface $view);
-   
+
         /**
          * @return \Zend_View_Interface
          */
         public function getView();
-   
+
         /**
          * Renders this strategy. This method will be called when the user
          * displays the site.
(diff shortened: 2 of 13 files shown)


derrabus (approved):

greg0ire (approved):

greg0ire:
Thanks @franmomu !

franmomu:
Now I'm testing this in a real project, I'm not sure from a DX perspective (using SA tools) if this was a good move because if you have an event subscriber like:

```php
use use Doctrine\ORM\Event\LifecycleEventArgs;

public function postLoad(LifecycleEventArgs $args) {
    $blockItem = $args->getObject();
    // ...
}
```

now to make SA tools happy you have to write:

```php
use Doctrine\Persistence\Event\LifecycleEventArgs;

/**
 * @param LifecycleEventArgs<EntityManagerInterface> $args
 */
public function postLoad(LifecycleEventArgs $args) {
    $blockItem = $args->getObject();
    // ...
}
```

Maybe should we create specific event classes for each one? like [`PostFlushEventArgs`](https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/Event/PostFlushEventArgs.php) or [`PreFlushEventArgs`](https://github.com/doctrine/orm/blob/e3e7f3c2090a2503734eb58b0c44e2309f204c15/lib/Doctrine/ORM/Event/PreFlushEventArgs.php)? or undeprecate it?

derrabus:
Having a dedicated class per event makes sense to me. This would also allow us to migrate to the PSR event dispatcher contract eventually.

franmomu:
👍 I'll create specific event classes, but since that would target `2.14`, should we undeprecate `LifecycleEventArgs` (keeping the deprecation to `getEntity` and `getEntityManager`) in the meantime? to release a patch version so we don't make users change their code yet

derrabus:
Yeah, that's probably a good idea. Let's give this topic a bit more thought and postpone to 2.14.

PR #8319, opened 2020-10-26
Fix variadic args in Expr

simPod:
Closes #8537

PR #9909, opened 2022-07-18
Show Expr::countDistinct() and Expr::concat() use variable-length argument lists

craigfrancis:
The functions `Expr::countDistinct()` and `Expr::concat()` both use `func_get_args()` to accept a variable number of arguments.

They should explicitly show this via the "`...`" operator, which has been supported since PHP 5.6.

### DIFF
--- a/lib/Doctrine/ORM/Query/Expr.php
+++ b/lib/Doctrine/ORM/Query/Expr.php
@@ -262,7 +262,7 @@ class Expr
      */
     public function countDistinct(...$x)
     {
-        return 'COUNT(DISTINCT ' . implode(', ', func_get_args()) . ')';
+        return 'COUNT(DISTINCT ' . implode(', ', $x) . ')';
     }
 
     /**
--- a/lib/Doctrine/ORM/Query/Expr.php
+++ b/lib/Doctrine/ORM/Query/Expr.php
@@ -260,7 +260,7 @@ class Expr
      *
      * @return string
      */
-    public function countDistinct(...$x)
+    public function countDistinct(mixed ...$x)
     {
         return 'COUNT(DISTINCT ' . implode(', ', $x) . ')';
     }
--- a/lib/Doctrine/ORM/Query/Expr.php
+++ b/lib/Doctrine/ORM/Query/Expr.php
@@ -544,7 +544,7 @@ class Expr
      *
      * @return Expr\Func
      */
-    public function concat(...$x)
+    public function concat(mixed ...$x)
     {
         return new Expr\Func('CONCAT', $x);
     }
--- a/lib/Doctrine/ORM/Query/Expr.php
+++ b/lib/Doctrine/ORM/Query/Expr.php
@@ -6,7 +6,6 @@ namespace Doctrine\ORM\Query;
 
 use Traversable;
 
-use function func_get_args;
 use function implode;
 use function is_bool;
 use function is_iterable;
--- a/lib/Doctrine/ORM/Query/Expr.php
+++ b/lib/Doctrine/ORM/Query/Expr.php
@@ -546,7 +546,7 @@ class Expr
      */
     public function concat(...$x)
     {
-        return new Expr\Func('CONCAT', func_get_args());
+        return new Expr\Func('CONCAT', $x);
     }
 
     /**
--- a/lib/Doctrine/ORM/Query/Expr.php
+++ b/lib/Doctrine/ORM/Query/Expr.php
@@ -256,11 +256,11 @@ class Expr
     /**
      * Creates an instance of COUNT(DISTINCT) function, with the given argument.
      *
-     * @param mixed $x Argument to be used in COUNT(DISTINCT) function.
+     * @param mixed ...$x Argument to be used in COUNT(DISTINCT) function.
      *
      * @return string
      */
-    public function countDistinct($x)
+    public function countDistinct(...$x)
     {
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -256,11 +256,11 @@ public function count($x)
...
      *
-     * @param mixed $x Argument to be used in COUNT(DISTINCT) function.
+     * @param mixed ...$x Argument to be used in COUNT(DISTINCT) function.
      *
      * @return string
      */
-    public function countDistinct($x)
+    public function countDistinct(...$x)
     {
         return 'COUNT(DISTINCT ' . implode(', ', func_get_args()) . ')';
```
greg0ire:
```suggestion
        return 'COUNT(DISTINCT ' . implode(', ', $x) . ')';
```

lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -540,12 +540,11 @@ public function notLike($x, $y)
...
-     * @param mixed $x     First argument to be used in CONCAT() function.
-     * @param mixed $y,... Other arguments to be used in CONCAT() function.
+     * @param mixed ...$x Arguments to be used in CONCAT() function.
      *
      * @return Expr\Func
      */
-    public function concat($x, $y)
+    public function concat(...$x)
     {
         return new Expr\Func('CONCAT', func_get_args());
```
greg0ire:
```suggestion
        return new Expr\Func('CONCAT', $x);
```

greg0ire:
Not sure why this wasn't addressed in https://github.com/doctrine/orm/pull/8319

greg0ire:
There is a (small) BC-break for extending classes: https://3v4l.org/qQUlA

craigfrancis:
Thanks Grégoire.

I wonder if a few other instances of the `func_get_args()` functions could be removed as well, but I should probably keep this PR simple.

craigfrancis:
As to the BC-break for extending Expr... that's a good point... is this something that happens with Expr?

greg0ire:
> is this something that happens with Expr?

I'm really not sure how to find out :sweat_smile: … all I know is that it is not `final` nor `internal`.

craigfrancis:
> > is this something that happens with Expr?
> 
> I'm really not sure how to find out 😅 … all I know is that it is not `final` nor `internal`.

Good point, I suppose anyone could want a better version of these methods... where they could prefer a lower-case `count()` :-)

greg0ire:
Maybe retarget to 3.0.x then, and make sure that there are no other occurrences of `func_get_args()` in the codebase?

craigfrancis:
I've changed the target for this PR.

And a quick search of `3.0.x` suggests the only other instance of `func_get_args()` is in `ReflectionReadonlyProperty`, and I'm not sure I should be touching that (I'm not too familiar with what it's doing).

greg0ire:
I think even if you tried it wouldn't work, as this class extends another one defined by PHP itself. You would probably end up with an incompatible signature.

greg0ire (dismissed):

lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -256,13 +256,13 @@ public function count($x)
     /**
      * Creates an instance of COUNT(DISTINCT) function, with the given argument.
      *
-     * @param mixed $x Argument to be used in COUNT(DISTINCT) function.
+     * @param mixed ...$x Argument to be used in COUNT(DISTINCT) function.
      *
      * @return string
      */
-    public function countDistinct($x)
+    public function countDistinct(...$x)
```
derrabus:
```suggestion
    public function countDistinct(mixed ...$x)
```

lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -540,14 +540,13 @@ public function notLike($x, $y)
...
      * Creates a CONCAT() function expression with the given arguments.
      *
-     * @param mixed $x     First argument to be used in CONCAT() function.
-     * @param mixed $y,... Other arguments to be used in CONCAT() function.
+     * @param mixed ...$x Arguments to be used in CONCAT() function.
      *
      * @return Expr\Func
      */
-    public function concat($x, $y)
+    public function concat(...$x)
```
derrabus:
```suggestion
    public function concat(mixed ...$x)
```

derrabus (commented):
Good catch, just some nit-picking.

craigfrancis:
Thanks Alexander

greg0ire (dismissed):

greg0ire:
Please rebase interactively, get rid of 58df2e137e9f66cd5195d67ba1b2582719e4ae62 and squash all other commits into one.

craigfrancis:
Sorry, made even more of a mess of the commit history (I'm putting it down to the heat)... I've opened a new one, which uses `3.0.x` as the base, and has a single commit.

greg0ire:
> Sorry, made even more of a mess of the commit history (I'm putting it down to the heat)... I've opened a new one, which uses 3.0.x as the base, and has a single commit.

Why don't you reopen and force push instead?

craigfrancis:
> Why don't you reopen and force push instead?

While I created a branch for this change, I wasn’t paying attention, and committed to the numbered branch… and when it came to the rebase, I was rushing, as I’m cooking food atm and nearly burnt the rice :-)

greg0ire:
Slow down lol

PR #9915, opened 2022-07-19
Make EntityManager `@final` and its constructor public

### LINKED ISSUE #9896: Make EntityManager::__construct() public
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

Right now, one is forced to call `EntityManager::create()` in order to instantiate an entity manager.

This makes the ORM hardly compatible with lazy ghost objects. The reason is that lazy ghost objects need to first be created with e.g. `ReflectionClass::newInstanceWithoutConstructor()` and later on be populated on demand.

Right now, the object graph contains e.g: EM -> UOW -> EM. This circular reference is defined in the constructor of EM.

Because of the factory, the current process when creating a lazy ghost object involves the following steps:
- create empty EM1
- on demand: trigger EM1 initialization
- initialization calls the factory, which returns a populated EM2 (which contains UOW -> EM2)
- copy all properties of EM2 into EM1 to initialize EM1 as needed

The resulting object graph is EM1 -> UOW -> EM2. It's broken of course.

If the constructor were made public, one could call the constructor in the initialization step and this concern would disappear.

The alternative right now would be to deep-search-n-replace EM2 by EM1 in EM1, but this would be super hacky code involving complex reflection/rebinding logic (and incompatible with `readonly` even if EM doesn't use it for now.)

nicolas-grekas:
Fix #9896

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -70,8 +70,10 @@ use function strpos;
  * is not a valid extension point for the EntityManager. Instead you
  * should take a look at the {@see \Doctrine\ORM\Decorator\EntityManagerDecorator}
  * and wrap your entity manager in a decorator.
+ *
+ * @final
  */
-/* final */class EntityManager implements EntityManagerInterface
+class EntityManager implements EntityManagerInterface
 {
     /**
      * The used Configuration.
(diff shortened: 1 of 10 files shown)


derrabus:
Thanks. Can you have a look at the failing tests?

nicolas-grekas:
Likely not before a week but otherwise yes!
What about removing the last argument of the constructor btw?

derrabus:
> What about removing the last argument of the constructor btw?

That's probably a good idea.

nicolas-grekas:
Tests updated, PR should be ready.

derrabus (approved):

greg0ire (approved):

greg0ire:
There was a cancelled job, I'm running it again, hopefully it will pass. I don't know why it was cancelled.

greg0ire:
The performance of that job is suspiciously bad: https://github.com/doctrine/orm/runs/7559763390?check_suite_focus=true, even when compared with this other random PR: https://github.com/doctrine/orm/runs/7558748372?check_suite_focus=true
Let me re-run that last one.

greg0ire:
It's slower than usual (1m13), but not as slow, let me restart the first one then.

greg0ire:
53s :slightly_smiling_face:

greg0ire:
Thanks @nicolas-grekas !

nicolas-grekas:
Thanks for merging \o/

PR #9918, opened 2022-07-20
Enrich types

greg0ire:
Yesterday, I tried to migrate a new namespace to PHP8 syntax, and it turns out it has wrong PHPDoc. I suspect the same holds true for numerous types under the AST namespace. Here are types widenings I applied based on the unit tests, but I think it would be better to use a common ancestor in many cases.

### DIFF
--- a/lib/Doctrine/ORM/Query/AST/Functions/AbsFunction.php
+++ b/lib/Doctrine/ORM/Query/AST/Functions/AbsFunction.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Query\AST\Functions;
 
+use Doctrine\ORM\Query\AST\ArithmeticTerm;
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ORM/Query/AST/Functions/AbsFunction.php
```diff
@@ -16,7 +16,7 @@
  */
 class AbsFunction extends FunctionNode
 {
-    /** @var SimpleArithmeticExpression */
+    /** @var SimpleArithmeticExpression|ArithmeticTerm|PathExpression */
```
greg0ire, line 21: The common ancestor would be `Node`… what should happen to the comment on line 13 then, in both situations?

derrabus:
Yes, the types in the AST namespace are wild. I've tried to fix them myself numerous times, but always ended up in a massive rabbit hole. 😓

PR #9929, opened 2022-07-22
Switch to setUpEntitySchema

greg0ire:
It might be more performant since we are keeping track of tables that
were already created.

This can also result in the metadata for some classes not being loaded,
which makes a test fail, and I think it reveals the bug covered by this
test was not actually fixed.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC832Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC832Test.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Doctrine\Tests\ORM\Functional\Ticket;
 
 use Doctrine\DBAL\Platforms\OraclePlatform;
-use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/DDC832Test.php
```diff
@@ -34,21 +33,6 @@ protected function setUp(): void
...
-        $sm = $this->createSchemaManager();
-        $sm->dropTable($platform->quoteIdentifier('TREE_INDEX'));
-        $sm->dropTable($platform->quoteIdentifier('INDEX'));
-        $sm->dropTable($platform->quoteIdentifier('LIKE'));
-
-        if ($platform instanceof PostgreSQLPlatform) {
-            $sm->dropSequence($platform->quoteIdentifier('INDEX_id_seq'));
-            $sm->dropSequence($platform->quoteIdentifier('LIKE_id_seq'));
-        }
-    }
```
greg0ire, line 50: I don't fully understand why I had to do this yet.

PR #9935, opened 2022-07-26
Removed orderBy clause in walk select statement

Yolo-plop:
I've encountered an issue with an IBM Informix database
system would suffer heavily from having an order by in the
count select statement. Currently, this file contains a condition
where it will remove the order by statement for Sql Server.
Why are we keeping for other database systems? Even if it should
only make a minor difference, i can't see a reason to keep this
in the query since we are simply doing a count.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Pagination/CountOutputWalker.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/CountOutputWalker.php
@@ -71,9 +71,7 @@ class CountOutputWalker extends SqlWalker
      */
     public function walkSelectStatement(SelectStatement $AST)
     {
-        if ($this->platform instanceof SQLServerPlatform) {
-            $AST->orderByClause = null;
-        }
+        $AST->orderByClause = null;
 
         $sql = parent::walkSelectStatement($AST);
 


greg0ire:
> Why are we keeping for other database systems?

Please find the commit that introduced that condition, hopefully it has the answer to your question.

derrabus:
For reference: #9934

Yolo-plop:
> > Why are we keeping for other database systems?
> 
> Please find the commit that introduced that condition, hopefully it has the answer to your question.

Here it is, it just says that removing the order by fixes pagination for sql server.

https://github.com/doctrine/orm/commit/6a69b4700ca0d30ac6ae6f793e46f53b242c6daf

greg0ire:
Ok, I guess we'll never know why it was done like that in the first place :sweat: 
Please amend your commit to address the remaining issues with this patch.

lib/Doctrine/ORM/Tools/Pagination/CountOutputWalker.php
```diff
@@ -71,9 +71,7 @@ public function __construct($query, $parserResult, array $queryComponents)
      */
     public function walkSelectStatement(SelectStatement $AST)
     {
-        if ($this->platform instanceof SQLServerPlatform) {
```
derrabus, line 74: Now that `$this->platform` is not used anymore, we need to remove the property from the class and the assignment from the constructor, to make PHPStan and Psalm happy.

mpdude:
@Yolo-plop this should be an easy thing to finish and get merged, not?

Yolo-plop:
Sorry, i've been on sick leave those past months and had forgotten about this.

Turns out i'm simply an idiot, the explanation is in the class doc:

_Note that the ORDER BY clause is not removed. Many SQL implementations (e.g. MySQL) are able to cache subqueries. By keeping the ORDER BY clause intact, the limitSubQuery that will most likely be executed next can be read from the native SQL cache._

Now, i have no idea whether it is true that it actually works for the given database systems. Clearly, it doesn't work with mine, but mine really sucks to be honest.

mpdude:
So shall we close this, or how could we proceed?

Yolo-plop:
Actually i thought this was closed because i received a closed notification at some point but it was for another Doctrine related issue. Sorry.

We can close this, or it might be worth creating some tests cases with the more usual database systems to check whether this really has a positive impact on performances. I can add this to my todo list if people are interested but right now i don't have extra time to give to this.

PR #9937, opened 2022-07-27
Use a more precise phpdoc for ClassMetadataInfo::versionField than mixed

VincentLanglet:
The phpdoc for $versionField is mixed but this should be `string|null` instead
- Everytime it's not null, it's used as an array key.
- `public function setVersionField($versionField)` were only accepting string.

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultEntityHydrator.php
+++ b/lib/Doctrine/ORM/Cache/DefaultEntityHydrator.php
@@ -12,6 +12,7 @@ use Doctrine\ORM\UnitOfWork;
 use Doctrine\ORM\Utility\IdentifierFlattener;
 
 use function array_merge;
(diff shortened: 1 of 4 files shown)


derrabus:
Hey, thanks for the PR. The PR description should give me a reason why I would want to merge your change. Oh, and a test would be nice. 🙃

VincentLanglet:
> Hey, thanks for the PR. The PR description should give me a reason why I would want to merge your change. Oh, and a test would be nice. 🙃

Hi, what test do you need for a phpdoc update ?

derrabus:
I just saw that you changed code as well, but I realize now that you've only added a couple of asserts. This should be fine without a test. And thanks for adding a description.

derrabus (approved):

PR #9939, opened 2022-07-28
Store PersistentCollection changes in entity change set as array

franmomu:
The property `$entityChangeSets` of UnitOfWork stores entity changes as an array of two items for each property changed. 

https://github.com/doctrine/orm/blob/a3a8caae516d8b88c025692159ce44f7ee7f1844/lib/Doctrine/ORM/UnitOfWork.php#L143-L149

But for `Collection` it has another behaviour as pointed out in https://github.com/doctrine/orm/discussions/8955 and changed the phpdoc in https://github.com/doctrine/orm/pull/8959.

That was added in https://github.com/doctrine/orm/commit/65bbdc30de891340b685c483046b535be4282718.

This PR modifies that case storing an `array` instead of the `PersistentCollection`, this will allow `PreUpdateEventArgs` to extend from `PreUpdateEventArgs` of `doctrine/persistence`.

I'm not sure if this is the right fix, but apparently the test added in the original commit passes with this change.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -141,6 +141,12 @@ Use one of the dedicated event classes instead:
 * `Doctrine\ORM\Event\PostRemoveEventArgs`
 * `Doctrine\ORM\Event\PostLoadEventArgs`
 
+## BC BREAK: Changed entity changeset when a `Collection` is modified
+
+When working with an entity with a `Collection` property, when that property
+is changed, the changeset stored in `UnitOfWork` includes the old and the new
+value of this property as an array of two items.
+
 ## BC BREAK: Removed `AttributeDriver::$entityAnnotationClasses` and `AttributeDriver::getReader()`
 
 * If you need to change the behavior of `AttributeDriver::isTransient()`,
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -514,7 +514,7 @@ class UnitOfWork implements PropertyChangedListener
      * Gets the changeset for an entity.
      *
      * @return mixed[][]
-     * @psalm-return array<string, array{mixed, mixed}|PersistentCollection>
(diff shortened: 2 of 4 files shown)


UPGRADE.md
```diff
@@ -1,5 +1,11 @@
 # Upgrade to 3.0
 
+## BC BREAK: Changed entity change set when a `Collection` is modified
```
```suggestion
## BC BREAK: Changed entity changeset when a `Collection` is modified
```

UPGRADE.md
```diff
@@ -1,5 +1,11 @@
...
+
+When working with an entity with a `Collection` property, when that property
+is changed, the changeset stored in `UnitOfWork` includes the old and the new
+value of this property as array of two items.
```
```suggestion
value of this property as an array of two items.
```

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -683,7 +683,7 @@ public function computeChangeSet(ClassMetadata $class, object $entity): void
                     }
 
                     $this->collectionDeletions[$coid] = $orgValue;
-                    $changeSet[$propName]             = $orgValue; // Signal changeset, to-many assocs will be ignored.
+                    $changeSet[$propName]             = [$orgValue, $actualData];
```
SenseException: I wonder if this change will leave dead code somewhere that handled the non-array value of `$changeSet[$propName] = $orgValue;`, but I haven't found any.

SenseException (dismissed):

greg0ire (dismissed):

derrabus:
Can we prepare this change on 2.x somehow, e.g. by allowing apps to opt-in to it?

SenseException (dismissed):

franmomu:
I forgot about this PR.

> Can we prepare this change on 2.x somehow, e.g. by allowing apps to opt-in to it?

I cannot think a way to do that 😞

PR #9950, opened 2022-08-02
Fix DBAL 4 compatibility

derrabus:
DBAL 4's switch to enums made some changes necessary. Maybe it would be wise at this point to introduce our own enums for `LockMode` and `ParameterType`, to be more resilient to this kind of changes?

### DIFF
--- a/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
+++ b/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
@@ -6,6 +6,7 @@ namespace Doctrine\ORM\Cache\Persister\Entity;
 
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\Common\Util\ClassUtils;
+use Doctrine\DBAL\LockMode;
 use Doctrine\ORM\Cache;
 use Doctrine\ORM\Cache\CollectionCacheKey;
 use Doctrine\ORM\Cache\EntityCacheKey;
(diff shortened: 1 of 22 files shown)


morozov:
> Maybe it would be wise at this point to introduce our own enums for `LockMode` and `ParameterType`, to be more resilient to this kind of changes?

Eventually, similar to the replacement of the DBAL's `FetchMode` with a set of `fetch*()` methods (https://github.com/doctrine/dbal/pull/4019), I'd like to make the parameter type enum internal and replace it with the methods like `bindInteger()`, `bindString()`, etc in the userland.

On the one hand, it will improve the type safety of the code, on the other, there will be no longer the need to keep the `ParameterType::STRING` binding type as the default.

morozov (dismissed):

lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
```diff
@@ -83,7 +84,7 @@ public function getInserts(): array
         return $this->persister->getInserts();
     }
 
-    public function getSelectSQL(array|Criteria $criteria, ?array $assoc = null, ?int $lockMode = null, ?int $limit = null, ?int $offset = null, ?array $orderBy = null): string
+    public function getSelectSQL(array|Criteria $criteria, ?array $assoc = null, LockMode|int|null $lockMode = null, ?int $limit = null, ?int $offset = null, ?array $orderBy = null): string
```
greg0ire: I think we should start using linebreaks.

greg0ire (approved):

PR #9955, opened 2022-08-03
fix: disabling versioning

writ3it:
When ClassMetadataInfo has the version field information (versioning), `->setVersioned(false)` will set the `requiresFetchAfterChange` flag to invalid if "generated" fields are not mapped. This PR fixes setting the flag state.

The consequence of the invalid state of `requiresFetchAfterChange` is the generation of invalid queries by the BasicEntityPersister::fetchVersionAndNotUpsertableValues method during the INSERT and UPDATE operations.

In this PR, I added two tests that verify the state after the `setVersioned(false)` operation is performed.

**Why**
`setVersioned(...)` is public and not restricted (by comments or doc) method. It is possible to call `setVersioned(false)` after `setVersionMapping(...)` which will produce invalid state. In some cases we need to overwrite state to achieve business feature f.e. to disable version verification during flush() in app with annotated entities with @Version.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -3509,9 +3509,8 @@ class ClassMetadataInfo implements ClassMetadata
     {
         $this->isVersioned = $bool;
 
-        if ($bool) {
-            $this->requiresFetchAfterChange = true;
-        }
+        $this->requiresFetchAfterChange = $bool 
+                                          || count(array_column($this->fieldMappings, 'generated')) > 0;
     }
 
     /**
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -28,6 +28,7 @@ use ReflectionNamedType;
 use ReflectionProperty;
 use RuntimeException;
 
+use function array_column;
 use function array_diff;
 use function array_flip;
 use function array_intersect;
@@ -3509,7 +3510,7 @@ class ClassMetadataInfo implements ClassMetadata
     {
         $this->isVersioned = $bool;
 
-        $this->requiresFetchAfterChange = $bool 
+        $this->requiresFetchAfterChange = $bool
                                           || count(array_column($this->fieldMappings, 'generated')) > 0;
     }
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -3509,9 +3509,8 @@ public function setVersioned($bool)
     {
         $this->isVersioned = $bool;
 
-        if ($bool) {
-            $this->requiresFetchAfterChange = true;
-        }
+        $this->requiresFetchAfterChange = $bool 
+                                          || count(array_column($this->fieldMappings, 'generated')) > 0;
```
Janczur, line 3514: Does $this->fieldMappings always have the same structure?

I would transfer this condition to a properly named method that describes what's going on here

Janczur (dismissed):
Looks ok

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
writ3it: Hi!

`fieldMappings` always has type `array<string, FieldMapping>` and FieldMapping = `array<string, mixed>`. FieldMapping type describes fields of entities and contains some optional keys, but array_column is processing this cases correctly.

Check the example code:
```php
$fieldMappings = [['fieldName'=>'col1'], ['fieldName'=>'col2', 'generated'=>0]];
var_dump(array_column($fieldMappings, 'generated')); // should equals to [0=>0]

$fieldMappings = [['fieldName'=>'col1'], ['fieldName'=>'col2']];
var_dump(array_column($fieldMappings, 'generated')); // should equals to []
```

Janczur: Ok looks good, but you still did not transfer this conditional to a properly named method :>

Janczur (dismissed):

derrabus:
So, to understand the problem that you're trying to fix: In your case, the class metadata is changed at runtime? Can you provide a use-case for that? I'm not really sure if we should "fix" this kind of problem. The class metadata is supposed to be generated by the metadata driver and should be considered immutable afterwards.

writ3it:
In my case I am adding to an app optimistic locking through app instance configuration. I have multiple app instances with different requirement of optimistic blocking (f.e. uat and production). The easiest way which I see is preparing entities with (f.e.) @Version annotation and ignoring them in driver when instance configuration requires no optimistic locking. To avoid duplicating a code of existing drivers I am decorating AnnotationDriver but I need to "disable" versioning in produced ClassMetada objects. Now I need to make correct state of requiresFetchAfterChange on my own which is weird because imho this field is internal detail of the class. The state isn't trivial so imo setVersioned should set correct state in both possible calls.

PR #9961, opened 2022-08-04
Deprecate `EntityManager::create()`

derrabus:
Now that the entity manager's constructor is public, why not make it the only way to construct an EM?

The only additional feature that `create()` has is that you can pass the connection parameters instead of a real DBAL connection. However, I believe that bootstrapping the DBAL connection is a concern that belongs into the DBAL library and the DBAL's `DriverManager` class already does a fine job in that regard.

This is why I believe that we shouldn't need the `create()` method anymore.

### DIFF
--- a/docs/en/cookbook/resolve-target-entity-listener.rst
+++ b/docs/en/cookbook/resolve-target-entity-listener.rst
@@ -127,7 +127,8 @@ the targetEntity resolution will occur reliably:
     // Add the ResolveTargetEntityListener
     $evm->addEventListener(Doctrine\ORM\Events::loadClassMetadata, $rtel);
 
-    $em = \Doctrine\ORM\EntityManager::create($connectionOptions, $config, $evm);
+    $connection = \Doctrine\DBAL\DriverManager::createConnection($connectionOptions, $config, $evm);
+    $em = new \Doctrine\ORM\EntityManager($connection, $config, $evm);
 
 Final Thoughts
 --------------
@@ -136,5 +137,3 @@ With the ``ResolveTargetEntityListener``, we are able to decouple our
 bundles, keeping them usable by themselves, but still being able to
 define relationships between different objects. By using this method,
 I've found my bundles end up being easier to maintain independently.
-
-
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -60,8 +60,8 @@ use function strpos;
  *     $paths = ['/path/to/entity/mapping/files'];
  *
  *     $config = ORMSetup::createAttributeMetadataConfiguration($paths);
- *     $dbParams = ['driver' => 'pdo_sqlite', 'memory' => true];
- *     $entityManager = EntityManager::create($dbParams, $config);
+ *     $connection = DriverManager::getConnection(['driver' => 'pdo_sqlite', 'memory' => true], $config);
+ *     $entityManager = new EntityManager($connection, $config);
  *
  * For more information see
  * {@link http://docs.doctrine-project.org/projects/doctrine-orm/en/stable/reference/configuration.html}
@@ -156,7 +156,7 @@ class EntityManager implements EntityManagerInterface
      * Creates a new EntityManager that operates on the given database connection
      * and uses the given Configuration and EventManager implementations.
      */
-    public function __construct(Connection $conn, Configuration $config)
+    public function __construct(Connection $conn, Configuration $config, ?EventManager $eventManager = null)
     {
         if (! $config->getMetadataDriverImpl()) {
             throw MissingMappingDriverImplementation::create();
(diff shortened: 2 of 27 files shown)


beberlei:
Can you also check the docs for necesarry changes? 

I am unsure about this. While your argument rings true, if we are on a greenfield project, we are not snd this method is not expensive to maintain, but will avoid breaking not only a lot of vode, but also tutorials, blog posts, books, ...

derrabus:
Marking as draft for now. The documentation indeed deserves an update.

> breaking not only a lot of code,

I don't believe that this is the case, actually. Usually, the EM is initialized once at a central point in your application. This is the one piece of code that you need to apply a trivial fix to.

Even more though, in applications that use some kind of framework integration like DoctrineBundle. In that case, this integration code will be patched and people won't even notice.

> but also tutorials, blog posts, books, ...

Those will all still work for ORM 2. Deprecated code is not broken code. ORM 3 will be a little different than ORM 2 and I think it's acceptable that blogposts and tutorials need to be adjusted for a new major version.

My rationale here is that the ORM and the DBAL are not as inseparable anymore as they maybe used to be. The `::create()` method might not be hard to maintain right now, but it also means that the DBAL can't change its own bootstrap too much without breaking the ORM.

lib/Doctrine/ORM/EntityManager.php
```diff
@@ -156,15 +156,15 @@ class EntityManager implements EntityManagerInterface
      * Creates a new EntityManager that operates on the given database connection
      * and uses the given Configuration and EventManager implementations.
      */
-    public function __construct(Connection $conn, Configuration $config)
+    public function __construct(Connection $conn, Configuration $config, ?EventManager $eventManager = null)
```
greg0ire, line 159: The class-level phpdoc comment refers to the factory method your deprecated, please update it.

SenseException:
Is this still a draft or can it be changed into a PR?

derrabus:
Still a draft because I'm updating the docs atm.

derrabus:
Ready.

I chose to resurrect this PR because the latest deprecation in DBAL (removal of the EventManager integration) force us to change the `create()` method again. Let's remove the `create()` method, it's a burden.

greg0ire (approved):

frsmoray:
Please update docs as [official documentation on creating an EntityManager](https://www.doctrine-project.org/projects/doctrine-orm/en/2.13/reference/advanced-configuration.html#advanced-configuration) actually tells us to use the `::create()` method. Not a single deprecation advice on that.

Thanks

greg0ire:
@frsmoray You are linking to the 2.13.x docs. This was perfectly valid in 2.13.x

Screenshot 2023-01-13 at 13-51-47 Advanced Configuration - Doctrine Object Relational Mapper (ORM)

> Not a single deprecation advice on that.

Use a nicer tone, this is not OK.

frsmoray:
Sorry, mistakenly got pulled 2.14 onto my project. Expeting to be 2.13 as usual.

>Use a nicer tone, this is not OK.

It's written text. Tone is a perception on spoken words. Apart from my mistake searching wrong docs, and 2.13 docs having nothing to do with any deprecation advice on that purpose, there is no deprecation notice at all there. If you think it twice, you may "find another tone" in these words

> Not a single deprecation advice on that.

Thanks for the help :)

greg0ire:
> Sorry, mistakenly got pulled 2.14 onto my project. Expeting to be 2.13 as usual.

2.13 is no longer maintained, so if you upgrade, that's not a bad things, really.

> It's written text. Tone is a perception on spoken words. 

Wanna nit-pick? From https://www.merriam-webster.com/dictionary/tone :

>  style or manner of expression in speaking or _writing_

(emphasis mine)

Note that I didn't have to discard several web pages until finding one that fits my argument, it's literally the first one I clicked.

> there is no deprecation notice at all there

You're right, but what's your point? Our policy is to document those in `UPGRADE.md`, not in the docs themselves.

> If you think it twice, you may "find another tone" in these words

I can do that, and on your end, you can be more careful when addressing people that provide you with software updates for free. I guess you meant well, but having to do that assumption for everyone can be a bit exhausting, don't you think?

docs/en/cookbook/resolve-target-entity-listener.rst
```diff
@@ -127,7 +127,8 @@ the targetEntity resolution will occur reliably:
     // Add the ResolveTargetEntityListener
     $evm->addEventListener(Doctrine\ORM\Events::loadClassMetadata, $rtel);
 
-    $em = \Doctrine\ORM\EntityManager::create($connectionOptions, $config, $evm);
+    $connection = \Doctrine\DBAL\DriverManager::createConnection($connectionOptions, $config, $evm);
```
flack, line 130: is this supposed to be `getConnection`? There doesn't seem to be a `createConnection` function in this class

derrabus: Yes.

SenseException: Yes, it became getConnection by now. It's totally okay if you want to create a pull request with changes for that.

If you haven't done a pull request for Doctrine, you might find this guide helpful:
https://www.doctrine-project.org/contribute/index.html

flack: See #11104

PR #9962, opened 2022-08-05
Add types to ProxyFactory

derrabus:
Part of #9772


lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -175,7 +164,6 @@ private function createCloner(ClassMetadata $classMetadata, EntityPersister $ent
                     continue;
                 }
 
-                $property->setAccessible(true);
```
SenseException, line 178: Is this unnecessary code or a decision that only public properties should be used?

derrabus: This method is a no-op since PHP 8.1.

greg0ire (approved):

PR #9966, opened 2022-08-06
fix: class normalisation test

Tofandel:
As per https://github.com/doctrine/persistence/pull/301

This test was not added to fix a bug, but to make sure that classes are not normalized at the time for "[performance reasons](https://github.com/doctrine/orm/issues/1750#issuecomment-162355601)" this check makes no sense now and in fact causes bugs without the normalization, this was solved in the linked PR and now causes this test to fail

I changed the test to check for the exact opposite as per when it was first introduced https://github.com/doctrine/orm/blame/705a477067e7a68c834ed65605d8d230677b5eaf/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php#L62

### DIFF
--- a/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
@@ -151,7 +151,7 @@ class ClassMetadataFactoryTest extends OrmTestCase
         $actual = $cmf->getMetadataFor($cm1->name);
     }
 
-    public function testHasGetMetadataNamespaceSeparatorIsNotNormalized(): void
+    public function testHasGetMetadataClassIsNormalized(): void
     {
         require_once __DIR__ . '/../../Models/Global/GlobalNamespaceModel.php';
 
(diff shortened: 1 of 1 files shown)


writ3it (dismissed):

Tofandel:
Might need to skip this test on php7.1 as the version of doctrine/persistence with this fix is limited to php 7.2

writ3it:
Tests for 7.1 are using doctrine/persistence:2.5.x which is not using normalization. Class which inherits Abstract ClassMetadataFactory from doctrine/persistence could work in different way depends on php version. This is a little bit weird because doctrine/orm is supporting 7.1-8.1. Alternative solution is introducing normalization in doctrine/persistence:2.5.x.

Tofandel:
Yes that's what I figured, unless we backport the fix to 2.5 we need to skip the test

So the question is do we backport or leave it as is?

For now this PR should at least get the build back on track for other PRs

Tofandel:
Backport is there https://github.com/doctrine/persistence/pull/304

derrabus (dismissed):

derrabus:
The low-deps test ist still failing. Let's delete that test. I don't think that we should test the behavior of the persistence library here.

Tofandel:
Sure, I agree we shouldn't test class behavior of another package here that leads to this kind of CI pains 😬, the test is in persistence anyways already 

But if its failing it just means it still hasn't picked the new version, does it need to be bumped manually or something?

I'm thinking it's being cached by the github action

Tofandel:
Yep it's cached based on the commit, so it got cached before the release and it's not trying to upgrade anymore
```txt
Cache restored successfully
Cache restored from key: linux-php-7.1.33-47+ubuntu20.04.1+deb.sury.org+1-lowest-b97b176328ff755e64de67a70fffbd8b35ac93958fe87d295c581b2046a136ce
```

derrabus (dismissed):

derrabus:
PHPCS fails, can you run PHPCBF please? You can ignore the Psalm and PHPStan errors. I'm going to take care of those.

derrabus:
Thanks

PR #9976, opened 2022-08-07
Add types to EntityListenerResolver

derrabus:
Part of #9772

### DIFF
--- a/lib/Doctrine/ORM/Mapping/EntityListenerResolver.php
+++ b/lib/Doctrine/ORM/Mapping/EntityListenerResolver.php
@@ -13,26 +13,18 @@ interface EntityListenerResolver
      * Clear all instances from the set, or a specific instance when given its identifier.
      *
      * @param string|null $className May be any arbitrary string. Name kept for BC only.
-     *
-     * @return void
      */
-    public function clear($className = null);
+    public function clear(?string $className = null): void;
 
     /**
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ORM/Mapping/EntityListenerResolver.php
```diff
@@ -25,12 +23,12 @@ public function clear($className = null);
...
-    public function resolve($className);
+    public function resolve(string $className): object;
 
     /**
      * Register a entity listener instance.
      *
      * @param object $object An entity listener
      */
-    public function register($object);
+    public function register(object $object): void;
```
SenseException, line 29: Do you want to keep the arguments `@param` docblock?

SenseException (dismissed):

lib/Doctrine/ORM/Mapping/EntityListenerResolver.php
derrabus: Not necessarily. I was unsure if it still provides value. What do you think? Remove?

SenseException: Yes, it can be removed.

derrabus: Done.

SenseException (dismissed):

PR #9978, opened 2022-08-07
Add types to `Node::dispatch()`

derrabus:
Part of #9772

This PR adds native types to all implementations of `Node::dispatch()` and makes sure the only parameter of that method is consistently named `$walker`.


SenseException (approved):
Also thank you for the argument name consistency.

SenseException (approved):
Also thank you for the argument name consistency.

PR #9980, opened 2022-08-08
Undeprecate LifecycleEventArgs

franmomu:
In https://github.com/doctrine/orm/pull/9906 I didn't think about what that meant when using event subscribers.

Deprecating this class means that users using SA tools have to update their code specifying the ObjectManager implementation.

From

```php
use use Doctrine\ORM\Event\LifecycleEventArgs;

public function postLoad(LifecycleEventArgs $args) {
    $blockItem = $args->getObject();
    // ...
}
```

to:

```php
use Doctrine\Persistence\Event\LifecycleEventArgs;

/**
 * @param LifecycleEventArgs<EntityManagerInterface> $args
 */
public function postLoad(LifecycleEventArgs $args) {
    $blockItem = $args->getObject();
    // ...
}
```

That's a bad DX, instead of this, we can create specific classes for each event and deprecate this class again when doing so and we don't make users to change their code twice (if they want to use the dedicate classes).

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -73,10 +73,6 @@ This method has been deprecated in:
 
 It will be removed in 3.0. Use `getObjectManager()` instead.
 
-## Deprecated `Doctrine\ORM\Event\LifecycleEventArgs` class
-
-It will be removed in 3.0. Use `Doctrine\Persistence\Event\LifecycleEventArgs` instead.
-
 ## Prepare split of output walkers and tree walkers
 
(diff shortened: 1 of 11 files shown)


derrabus (approved):

guillaumesmo:
Question: the class has been undeprecated, yet the deprecation "The %s class is deprecated and will be removed in ORM 3.0. Use %s instead." in the constructor hasn't been reverted; is that intentional?

guillaumesmo:
answering myself for future reference: it was deprecated again in ORM version 2.14 #10086

Issue #9960, opened 2022-08-04
No preUpdate event if removing elements from PersistentCollection results in empty collection


s-bernstein:
Q | A
-- | --
Version | 2.12.3

**Summary**
Removing elements from a PersistentCollection will trigger a preUpdate event, except if the resulting collection becomes empty.

**Current behavior**
Updating an entity through changing a PersistentCollection triggers a preUpdate event, if the resulting collection is not empty.

**Expected behavior**
Updating an entity through changing a PersistentCollection should always result in a preUpdate event being triggert, even if the resulting collection becomes empty.

**How to reproduce**
Have an entity with a many-to-many association represented by a PersistentCollection. Remove all elements from said collection and flush.

**Possible problem**
During debugging I realized, that removing all elements from a PersistentCollection doesn't set the dirty flag, and therefore the UnitOfWork::computeChangeSet() does not enter the last if-Statement (beginning line 821). Manually setting the dirty flag in the debugger will trigger the preUpdate event.
Some more debugging revealed, that SonataAdmin is calling PersistentCollection::clear() to remove all elements, which will set the dirty flag until the subsequent call of PersistentCollection::takeSnapshot() in PersistenCollection:clear() line 553 resets the dirty flag.

derrabus:
Would you be able to prepare a test case that reproduces your problem? And since you've already debugged the issue, would you be able to work on a fix?

s-bernstein:
@derrabus Regarding your questions: First, I might be able to create a test but haven't looked into it. Also I'm not quite sure if this can be tested with a unit test or if this needs a functional test.

For the second question: Well, I can fix it - I actually have identified two places, where this could be fixed - but I have no understanding of all the implications this fixes will have. As described before, calling PersistentCollection::clear() will reset the dirty flag on the collection by calling PersistentCollection::takeSnapshot(). I suppose, whoever written this part had a good reason for it. I don't know what side effects will arise, if this behaviour is changed - although this seems to me the logical place because resetting the dirty flag looks like a logical error for me.

The other place would be in UnitOfWork::computeChangeSet(), but this involves changing an already long-ish if statement and feels a little forced.

s-bernstein:
OK, I have a PR (#9985) to fix this problem - hopefully...

s-bernstein:
Any chance this one will proceed in the near future?

s-bernstein:
Anything new?

SherinBloemendaal:
Any update? We have the same problem

[...]

PR #9985, opened 2022-08-09
No preUpdate event if removing elements from PersistentCollection results in empty collection

s-bernstein:
Fix for #9960


s-bernstein:
bump

PR #9988, opened 2022-08-10
Import @Enum annotation

cstffx:
Fix: 
[Semantical Error] The annotation "@Enum" in property Doctrine\ORM\Mapping\ManyToOne::$fetch was never imported. Did you maybe forget to add a "use" statement for this annotation?


derrabus:
Please provide steps to reproduce the bug you're attempting to fix.

cstffx:
Seems to be resolved in the latest version. To hard to reproduce.

PR #9992, opened 2022-08-13
Fix reflField phpdoc

VincentLanglet:


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -793,7 +793,7 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * The ReflectionProperty instances of the mapped class.
      *
-     * @var array<string, ReflectionProperty|null>
+     * @var array<string, ReflectionProperty>
```
VincentLanglet, line 796: Every time in the code, `$reflFields` is used this way
```
$reflFields['foo']->setValue()
```
so it would crash if null was in the array.

And looking at how the array is populated, we only set ReflectionProperty.

Issue #9505, opened 2022-02-12
Bug: UnitOfWork attempts to update readonly property (object IDs don't match)


zanbaldwin:
### Bug Report

| Q           |   A
|-------------| ------
| BC Break    | no
| ORM Version | 2.11.x
| PHP Version | 8.1+

#### Summary

Found a bug where UnitOfWork attempts to update a readonly property of a type that is an object not managed by Doctrine (eg, value object). a `\LogicException` is thrown.

#### Current behavior

Even though the data has not changed, the value object inside the managed entity (held inside `UnitOfWork::$identityMap`) is a different instance to the value object constructed from the data/row fetched from the database. When `ReflectionReadonlyProperty::setValue()` does an identical comparison `===` the object IDs don't match and a `LogicException` is thrown.

#### How to reproduce

Set an entity column property to be both `readonly` and type-hinted to be an object that is not an entity managed by the object manager (eg, `Symfony\Component\Uid\Ulid`, `App\ValueObject\Email`, `\DateTimeImmutable`, etc).

```php
#[ORM\Entity]
class Test1
{
    #[ORM\Id]
    #[ORM\Column(name: 'id', type: 'integer', nullable: false)]
    #[ORM\GeneratedValue(strategy: 'AUTO')]
    public readonly int $id;

    public function __construct(
        #[ORM\Column(name: 'date', type: 'datetime', nullable: false)]
        public readonly \DateTimeImmutable $date = new \DateTime,
    ) {}
}
```

```php
$entityManager->persist($entity = new TestEntity);
$entityManager->flush();
$entityManager->refresh($entity);
````

#### Expected behavior

Read-only properties should never be written to after initialization.

However, it's logical to assume that someone is calling `ObjectManager::refresh()` because they expect the data to have changed in the database. In this case, the only way I can think to fetch that updated data would be to detach the managed entity first before loading again.

#### Workaround

I've temporarily avoided this issue by not resetting readonly properties that have already been initialized, but it's not a solution since it purposefully avoids performing that value comparison inside `ReflectionReadonlyProperty`.

```diff
index cd5d72716..aadc82d32 100644
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2740,7 +2740,11 @@ class UnitOfWork implements PropertyChangedListener
         }
 
         foreach ($data as $field => $value) {
-            if (isset($class->fieldMappings[$field])) {
[...]

PR #9998, opened 2022-08-17
Fix change detection for non-scalar readonly properties

nCrazed:


beberlei:
This has no relation with readonly properties. Other types doont allow object compare via equals() either.

composer.json
```diff
@@ -47,6 +47,7 @@
         "psr/log": "^1 || ^2 || ^3",
         "squizlabs/php_codesniffer": "3.7.1",
         "symfony/cache": "^4.4 || ^5.4 || ^6.0",
+        "symfony/uid": "^4.4 || ^5.4 || ^6.0",
```
derrabus, line 50: I'm not keen on adding that dependency, tbh. We should be able to reproduce the problem without it.

lib/Doctrine/ORM/Mapping/ReflectionReadonlyProperty.php
```diff
@@ -44,8 +45,19 @@ public function setValue(mixed $objectOrValue, mixed $value = null): void
...
+        if ($this->isValueDifferent($objectOrValue, $value)) {
             throw new LogicException(sprintf('Attempting to change readonly property %s::$%s.', $this->class, $this->name));
         }
     }
+
+    private function isValueDifferent(object $object, mixed $value): bool
+    {
+        $objectValue = parent::getValue($object);
+
+        if (is_object($objectValue) && method_exists($objectValue, 'equals')) {
```
derrabus, line 57: I'm not comfortable with applying duck typing here. I understand the problem that you're trying to solve, but I'd rather not compare incomparable values than to guess how to compare them.

derrabus (changes requested):

composer.json
nCrazed: That's fair. I only added it because it's technically supposed to be supported out of the box (but that might be a symfony concern, so it makes sense to not add it in doctrine)

lib/Doctrine/ORM/Mapping/ReflectionReadonlyProperty.php
nCrazed: I am not fan of duck typing here either, but other than adding explicit type checks (and thus requiring a non-dev dependency for every supported type), I am not sure how else the problem could be fixed

nCrazed:
@beberlei the failure I am attempting to fix is specific to readonly property support. The one of the added test cases (same) fails without the change and the error that comes out of it is blatantly incorrect

nCrazed:
Not sure how I missed this related issue: #9505

lib/Doctrine/ORM/Mapping/ReflectionReadonlyProperty.php
derrabus: I mean, we could just assume equality if we cannot compare the values. I would prefer that solution over calling a method that just looks like it does the right thing…

nCrazed: Would exposing a configuration hook for providing custom equality checks be acceptable here?

nCrazed: The above question still stands. I am happy to work towards another fix, but would like a confirmation that the approach is acceptable before I do so

icanhazstring:
This has been stale for quite some time now.
As mentioned by @nCrazed, one solution could be a configurable map or general method that should be supported.

```yaml
doctrine:
    orm:
        type_comparison:
            # Map of methods that check for equality
            default: 'equals' # default for every other object type or even just fallback to that !== comarison
            Symfony\Uid\Uuid: 'compare'
```

This way, the `UnitOfWork` can also use this to compute changes. There are no changes when I generate a Uuid with the same string twice except having another object hash.

If that sounds like a good plan, please let us know. So we can get this feature integrated into doctrine.

derrabus:
I think, we can close this PR because of implementation issues. I'll happily review a new attempt that works without additional dependencies or duck typing. Some kind of discovery mechanism for comparator functions as suggested could be the solution, but I'm not sure about that.

nCrazed:
An alternative was suggested, but never approved/declined :man_shrugging:

icanhazstring:
Closing it is good and all. But we gave some suggestions and got no feedback for it.

The current behavior is clearly a bug and should be fixed. I think @nCrazed and I both would be happy to contribute, but we would kindly appreciate any kind of feedback to the proposed suggestion(s).

PR #10000, opened 2022-08-18
made sure that target entity is cacheable

meiyasan:
Target entity is not always related to a CachedEntityPersister. 
e.g Two Entities inheriting from each other and only **child class** is cached.


meiyasan:
Not sure that I properly fixed the linter issues. Any help would be very welcome... xD

greg0ire:
`vendor/bin/phpcbf` can help you with that

meiyasan:
Ok, I didn't manage to fix this issue. Any input would be very welcome

PR #10002, opened 2022-08-22
Make paginator covariant

VincentLanglet:
see https://phpstan.org/blog/whats-up-with-template-covariant
see https://psalm.dev/docs/annotating_code/templated_annotations/

A method which accepts `Paginator<Animal>` should be able to accept `Paginator<Cat>`.

I received the error
```
Template type T is declared as covariant, but occurs in invariant position in return type of method Doctrine\ORM\Tools\Pagination\Paginator::getIterator().
```

It's because the template of ArrayIterator is not covariant.
But the template for Traversable is covariant.

It makes sens to use Traversable instead of ArrayIterator as return type because
- IteratorAggregate::getIterator() use Traversable as return type
- it doesn't make sens to allow adding elements to this Iterator

ArrayIterator implements SeekableIterator, ArrayAccess, Serializable, Countable
the issue is the ArrayAccess.

If you preferred I can use somethink like
```
@psalm-return Traversable<array-key, T>&Countable
@psalm-return Traversable<array-key, T>&Serializable&Countable
@psalm-return SeekableIterator<array-key, T>&Serializable&Countable
...
```

Any suggestion @greg0ire ?

### DIFF
--- a/tests/Doctrine/StaticAnalysis/Tools/Pagination/paginator-covariant.php
+++ b/tests/Doctrine/StaticAnalysis/Tools/Pagination/paginator-covariant.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\StaticAnalysis\Tools\Pagination;
+
+use Doctrine\ORM\Tools\Pagination\Paginator;
+
+/**
+ * @template-covariant T of object
+ */
+abstract class PaginatorFactory
+{
+    /** @var class-string<T> */
+    private $class;
+
+    /**
+     * @param class-string<T> $class
+     */
+    final public function __construct(string $class)
+    {
+        $this->class = $class;
+    }
+
+    /**
+     * @return class-string<T>
+     */
+    public function getClass(): string
+    {
+        return $this->class;
+    }
+
+    /**
+     * @psalm-return Paginator<T>
+     */
+    abstract public function createPaginator(): Paginator;
+}
+
+interface Animal
+{
+}
+
+class Cat implements Animal
+{
+}
+
+/**
+ * @param Paginator<Animal> $paginator
+ */
+function getFirstAnimal(Paginator $paginator): ?Animal
+{
+    foreach ($paginator as $result) {
+        return $result;
+    }
+
+    return null;
+}
+
+/**
+ * @param PaginatorFactory<Cat> $catPaginatorFactory
+ */
+function test(PaginatorFactory $catPaginatorFactory): ?Animal
+{
+    return getFirstAnimal($catPaginatorFactory->createPaginator());
+}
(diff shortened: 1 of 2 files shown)


greg0ire:
No suggestion yet, but can you add a piece of code under `tests/Doctrine/StaticAnalysis` to demonstrate the benefit and avoid regressions? It will also help understanding the PR.

VincentLanglet:
> No suggestion yet, but can you add a piece of code under `tests/Doctrine/StaticAnalysis` to demonstrate the benefit and avoid regressions? It will also help understanding the PR.

I added as much tests as possible ^^

tests/Doctrine/StaticAnalysis/Tools/Pagination/paginator-covariant.php
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * @return class-string<T>
+     */
+    public function getClass(): string
+    {
+        return $this->class;
+    }
+
+    /**
+     * @psalm-return Paginator<T>
```
VincentLanglet, line 34: This wouldn't be possible if Paginator wasn't covariant.

tests/Doctrine/StaticAnalysis/Tools/Pagination/paginator-covariant.php
```diff
@@ -0,0 +1,65 @@
...
+
+    return null;
+}
+
+/**
+ * @param PaginatorFactory<Cat> $catPaginatorFactory
+ */
+function test(PaginatorFactory $catPaginatorFactory): ?Animal
+{
+    return getFirstAnimal($catPaginatorFactory->createPaginator());
```
VincentLanglet, line 64: Passing a Paginator<Cat> to getFirstAnimal wouldn't be possible if Paginator wasn't covariant

greg0ire:
Please squash your commits, and retarget to 2.14.x: this is an improvement and there is a small BC-break.

greg0ire:
> If you preferred I can use something like…

actually, `Traversable` is fine, because it's exactly what we are supposed to return to fulfill the contract. Since we do not intend people to use `getIterator()` directly, but rather indirectly through `foreach`, it makes sense to give as little detail as possible on the type. That way, we can change the implementation and something else if we want to.

VincentLanglet:
> Please squash your commits, and retarget to 2.14.x: this is an improvement and there is a small BC-break.

Done

greg0ire (approved):

derrabus (approved):

mitelg:
hey @VincentLanglet 

sorry for coming back after this MR is merged a long time ago :grimacing: 

we want to update to the recent 2.14.1 version. our problem is that we are using `$paginator->getIterator()->getArrayCopy()` a lot :sweat_smile: 

Just to be sure: the migration path would simply be exchanging this with `iterator_to_array($paginator)` ? the output array is the same in my first tests, just wanted to be doubled checked :smile: 

thanks in advance :v:

VincentLanglet:
> `$paginator->getIterator()->getArrayCopy()` a lot 😅
> 
> Just to be sure: the migration path would simply be exchanging this with `iterator_to_array($paginator)` ?

Sure.

PR #10009, opened 2022-08-27
Upgrade to doctrine/coding-standard 10.0.0

greg0ire:


greg0ire:
On PHP 7.1, with the dependencies that are pulled, the `@requires` annotation is not taken into when single-line.

SenseException (approved):

PR #10011, opened 2022-08-27
Upgrade to doctrine/coding-standard 10.0 on 3.0.x

greg0ire:
The plan is to do a merge up and ignore incoming changes.

Closes #9886


greg0ire:
I have baselined the following issues:

```
ERROR: InvalidReturnType - lib/Doctrine/ORM/EntityManager.php:279:116 - The declared return type '(T:fn-doctrine\orm\entitymanagerinterface::find as object)|null' for Doctrine\ORM\EntityManager::find is incorrect, got 'null|object' (see https://psalm.dev/011)
    public function find($className, mixed $id, LockMode|int|null $lockMode = null, int|null $lockVersion = null): object|null


ERROR: InvalidReturnType - lib/Doctrine/ORM/EntityManager.php:378:60 - The declared return type '(T:fn-doctrine\orm\entitymanagerinterface::getreference as object)|null' for Doctrine\ORM\EntityManager::getReference is incorrect, got 'Doctrine\Common\Proxy\Proxy<object>|(T:fn-doctrine\orm\entitymanagerinterface::getreference as object)|null|object' (see https://psalm.dev/011)
    public function getReference(string $entityName, $id): object|null


ERROR: InvalidReturnType - lib/Doctrine/ORM/EntityManager.php:422:75 - The declared return type '(T:fn-doctrine\orm\entitymanagerinterface::getpartialreference as object)|null' for Doctrine\ORM\EntityManager::getPartialReference is incorrect, got 'null|object' (see https://psalm.dev/011)
    public function getPartialReference(string $entityName, $identifier): object|null


ERROR: PossiblyNullArgument - lib/Doctrine/ORM/Query/SqlWalker.php:1183:66 - Argument 1 of Doctrine\ORM\Query\SqlWalker::walkSimpleArithmeticExpression cannot be null, possibly null value provided (see https://psalm.dev/078)
        $sql .= ' ELSE ' . $this->walkSimpleArithmeticExpression($generalCaseExpression->elseScalarExpression) . ' END';


ERROR: PossiblyNullArgument - lib/Doctrine/ORM/Query/SqlWalker.php:1193:62 - Argument 1 of Doctrine\ORM\Query\SqlWalker::walkStateFieldPathExpression cannot be null, possibly null value provided (see https://psalm.dev/078)
        $sql = 'CASE ' . $this->walkStateFieldPathExpression($simpleCaseExpression->caseOperand);


ERROR: PossiblyNullArgument - lib/Doctrine/ORM/Query/SqlWalker.php:1200:66 - Argument 1 of Doctrine\ORM\Query\SqlWalker::walkSimpleArithmeticExpression cannot be null, possibly null value provided (see https://psalm.dev/078)
        $sql .= ' ELSE ' . $this->walkSimpleArithmeticExpression($simpleCaseExpression->elseScalarExpression) . ' END';


ERROR: InvalidFalsableReturnType - lib/Doctrine/ORM/QueryBuilder.php:489:52 - The declared return type 'Doctrine\ORM\Query\Parameter|null' for Doctrine\ORM\QueryBuilder::getParameter does not allow false, but 'Doctrine\ORM\Query\Parameter|false|null' contains false (see https://psalm.dev/143)
    public function getParameter(string|int $key): Parameter|null
```

lib/Doctrine/ORM/Query/AST/WhereClause.php
```diff
@@ -13,15 +13,9 @@
...
 class WhereClause extends Node
 {
-    /** @var ConditionalExpression|ConditionalTerm */
-    public $conditionalExpression;
-
-    /**
-     * @param ConditionalExpression $conditionalExpression
-     */
-    public function __construct($conditionalExpression)
+    /** @param ConditionalExpression|ConditionalTerm $conditionalExpression */
```
greg0ire, line 16: Allowing A|B in the property but just A with the constructor causes many issues, I don't think it's worth the hassle to keep it.

SenseException: At least not in this PR. 882 files are quite tiring to read. :sweat_smile:. New type hint PRs will follow for sure for 3.0 in the future.

composer.json
```diff
@@ -37,7 +37,7 @@
     },
     "require-dev": {
         "doctrine/annotations": "^1.13",
-        "doctrine/coding-standard": "^9.0.2",
+        "doctrine/coding-standard": "^9.0.2 || ^10.0",
```
SenseException: We can drop 9.0 for ORM 3.0, right?

SenseException (changes requested):

composer.json
greg0ire: No we can't, otherwise it's not possible to install the project on PHP 7.1 (one more reason to drop it)

SenseException: ORM 3.0 doesn't support PHP 7 anymore

greg0ire: Oh right, sorry, I forgot what PR it was 👍

SenseException (dismissed):

SenseException (dismissed):

greg0ire:
Sorry @SenseException , the DBAL methods are removed faster than I'm making changes on this PR!

SenseException:
@greg0ire Aren't you doing stuff that's out of scope of a coding standard upgrade?

greg0ire:
I'm doing the strict required minimum to get a green build :sweat_smile: , and yes, that involves stuff that does not have to do with CS. We have several jobs that use DBAL 4, so such breakage is frequent and needs to be addressed before it piles up. I've regrouped all of that in the last commit.

SenseException:
A separate PR would be preferable than a PR with > 800 changed files :grin: but you already released the ghost from the lamp. :diya_lamp:

SenseException (approved):

PR #10021, opened 2022-08-30
Merge 2.14.x into 3.0.x

derrabus:


SenseException:
Are the coding styles supposed to be fixed in another PR or in this PR for the upmerge?

greg0ire:
I think PRs have been merged that passed the build before the CS PR was merged.

UPD: actually no, all 3 branches pass the CS build. This means the failures happen because of the difference in config between branch (3.0.x has PHP-8.1-specific rules)

SenseException:
> Error: Usage of short nullable type hint in "?Animal" is disallowed.

derrabus:
I've fixed the CS issue while merging.

PR #10030, opened 2022-09-05
DDC-1680

tgauges:
Create PostRemoveEventArgs to be able to get entity identifier in postRemove listeners.

Fixes #2326

- [ ] update documentation
- [ ] implement real test case checking if id is available
  - [ ] What happens with multiple identifier fields? I'm not sure because only the first identifer seems to be set to null anyway.

### DIFF
--- a/lib/Doctrine/ORM/Event/PostRemoveEventArgs.php
+++ b/lib/Doctrine/ORM/Event/PostRemoveEventArgs.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Event;
+
+use Doctrine\ORM\EntityManagerInterface;
+
+class PostRemoveEventArgs extends LifecycleEventArgs
+{
+    /** @param object $entity */
+    public function __construct($entity, EntityManagerInterface $objectManager, private mixed $identifier)
+    {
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Event/PostRemoveEventArgs.php
```diff
@@ -0,0 +1,21 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Event;
+
+use Doctrine\ORM\EntityManagerInterface;
+
+class PostRemoveEventArgs extends LifecycleEventArgs
+{
+    /** @param object $entity */
+    public function __construct($entity, EntityManagerInterface $objectManager, private mixed $identifier)
```
derrabus, line 12: Our codebase still maintains compatibility with PHP 7 and because of that we cannot use the `mixed` type yet.

derrabus (changes requested):
Sorry we haven't picked up your PR yet. The `PostRemoveEventArgs` class has been added in #10086 already. But I see that you've added an identifier property, so your PR might still be interesting.

Can you please rebase onto 2.14.x and resolve the conflicts?

tgauges:
The new pull request on 2.14.x is #10188

I will close this one.

PR #10034, opened 2022-09-09
Don't look for metadata on standard classes

duzenko:
...and don't throw exceptions when it's not found

Discussed here: https://github.com/doctrine/orm/discussions/10019#discussioncomment-3528300

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -434,7 +434,7 @@ abstract class AbstractQuery
         if (! $reflection->isUserDefined()) {
             return $this->potentiallyProcessIterable($value);
         }
-        
+
         try {
             $class = ClassUtils::getClass($value);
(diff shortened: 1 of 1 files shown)


derrabus:
For the record: I don't think that raising and catching an exception is a problem that deserves a fix.

duzenko:
> For the record: I don't think that raising and catching an exception is a problem that deserves a fix.

Can you suggest any workaround for debugger stopping on internal ORM exceptions during debugging as discussed in the link above?

mpdude:
I don’t think we should merge this. 

It’s a problem related to how you use the debugger, not the code itself.

for sure there’s a way how to configure your debugger to ignore this exception, not stop for it or similar.

greg0ire:
Let's close this then.

duzenko:
Typical PHP-world hair-brained "solution"

derrabus:
I guess, we PHP guys are just too stupid. Better move to go. 🤷🏻‍♂️

Issue #8390, opened 2020-12-13
The hydration cost can easily & significantly be decreased in case of repeated row data + costly type conversion


NoiseByNorthwest:
The well known performance issue with fetch joins on collection-like relations (*-to-many) is that the result set will contain repeated data. The simple case being an entity with one of its collection relations fetched which will then cause the (root) entity data to be repeated as many times as there are items in the collection.

It get obviously exponentially worse when nested or sibling collections are also fetched. This issue could BTW be avoided with a feature like this one (not merged yet) https://github.com/doctrine/orm/pull/1569 .

That being said, I've discovered another performance issue last week, which does not need thousands of rows to be noticeable. It occurs when the repeated data are mapped to a type with a costly conversion. In my case this is a `json_document` column, this type is provided by https://github.com/dunglas/doctrine-json-odm .

More concretly my query targets the table A, and fetch a collection of table B records. A contains a `json_document` column and for each A record there are ~66 B records.

The result set then contains 526 rows representing 526 distinct B records and only 8 distinct A records.

The query was to slow to meet my requirements and after having profiled that (in tracing mode to get actual call counts & sampling mode to get accurate timings) I've discovered that 70% of the hydration time is taken by `JsonDocumentType::convertToPHPValue()` which is called 526 times (instead of 8 as  logically required).

And this slowness has obviously more to do with the fact that `JsonDocumentType::convertToPHPValue()` is called 526 times where it could be called only 8 times than with the own slowness of `JsonDocumentType::convertToPHPValue()` which is a bit expected and with probably not so much room for optimizations.

The problem lies in the fact that `AbstractHydrator::gatherRowData()` which is, among other things, responsible to call the proper type conversion function for each column of the given row, does not do any caching logic to avoid doing this process multiple times for the same logical set of column values (representing the root or a fetched relation AKA DQL alias).

So I've written my own hydrator, which extends `ObjectHydrator` and after having copy/pasted the `gatherRowData()` function I've added this caching logic.

When used, I've observed a 78% decrease of the `hydrateAllData()` cost. Here is a table below with more details:

| Metric | Original hydrator | Custom hydrator | Decrease (%) |
|-----|-----|-----|-----|
| `hydrateAllData()` wall time (ms) | 232  | 52 | -78% |
| `gatherRowData()` call count | 526  | 526 | - |
| `gatherRowData()` wall time (ms) | 200  | 26 | -87% |
[...]

PR #10035, opened 2022-09-09
Optimize Object Hydration performance on large result sets

tucksaun:
Hi folks!

While assessing a client's project performance I found some optimizations regarding the ORM hydration process that improves speed when working with large result sets and collections.

<details>
 <summary>Expand to get some context on how we reached this.</summary>
The context is about online carts on a B2B e-commerce platform with a high number of items and complex personalization logic.
We reached the point in the optimization process where most of the time was spent in the ORM (after serialization) when fetching a cart.

This makes sense because of the volume and the logic at stake.
Denormalization or caching does not make sense for us because of the low read/write ratio and the business logic involved so we first went with the 2-step hydration technic.
</details>

While analyzing profiling results we more or less came to the same conclusion as #8390: the hydration process converts data over and over again even if the entity is already hydrated>

But after trying to work around performance for specific DBAL types that were costly to convert (namely Symfony's UUID and Dunglas's JSON document) and digging deeper into profiles I concluded something could be done directly in the Hydrator for the benefit of everyone instead of writing our specific optimized hydrator:
Screenshot 2022-09-08 at 17 51 38

This PR acts according to one consideration: most parts of the hydration process are not slow per se but are repeated for every row and column. By applying some small optimizations, moving some code before the first iteration, or some code at the very last moment, we can improve the overall process.
The more rows and columns, the more the process is accelerated.

For examples:
- type conversion can be costly but only identifiers are required to determine if the entity is already hydrated or not: the actual data can be converted later on in the process only if needed using only the columns required for the current entity (not the data for other entities in relations);
- we can prepare the mapping information in a way that prevents repeating some part of the process thousands of times;
- and we can avoid using the reflection to retrieve relations properties if they are available in another mapping.

With this patch, I manage to reach the following results:

| Metric | Before | After | Notes |
|-----------------------------|--------|-------|----------------|
|`->getQuery()->getResult()`| 726.39 ms | 358.67 ms | -51%, best of 20 iterations, query result cache enabled, 11 826 entities created
|`Doctrine\ORM\PersistentCollection::*`| 164 634 | 77 345
|`TypedNoDefaultReflectionProperty::getValue` | 162 770  | 125 669
|`Doctrine\ORM\Internal\Hydration\AbstractHydrator::resultSetMapping` | 441 742 | 19 745
|`Doctrine\ORM\Internal\Hydration\AbstractHydrator::hydrateColumnInfo`| 946 970 | 337
|`Doctrine\DBAL\Types\IntegerType::convertToPHPValue`| 406 982 | 279 366
|`Doctrine\DBAL\Types\StringType::convertToPHPValue`| 181 207 | 63 473
|`Doctrine\DBAL\Types\BooleanType::convertToPHPValue`| 75 163 | 7 865
|`Doctrine\DBAL\Types\DateTimeImmutableType::convertToPHPValue`| 25 118 |  789
|`App\DBAL\Types\UuidType::convertToPHPValue`| 13 056 | 413

I hope we can manage to get this work so that everyone enjoys better performance 🤗
If you are willing to move in this direction, we can probably adapt ArrayHydrator the same way.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -414,7 +414,7 @@ abstract class AbstractHydrator
      *                   data: array<array-key, array>,
      *                   newObjects?: array<array-key, array{
      *                       class: mixed,
-     *                       args?: array
+     *                       args: array
      *                   }>,
      *                   scalars?: array
      *               }
@@ -548,8 +548,10 @@ abstract class AbstractHydrator
      *
      * @psalm-return array<array-key, array{
      *                       class: mixed,
-     *                       args?: array
+     *                       args: array
      *                   }>
+     *
+     * @psalm-suppress MoreSpecificReturnType Psalm does not detect that args keys is always defined
      */
     protected function gatherRowDataFromNewObjectsMapping(array $data): array
     {
@@ -569,6 +571,7 @@ abstract class AbstractHydrator
             $newObjectsData[$objIndex]['args'][$argIndex] = $value;
         }
 
+        /** @psalm-suppress LessSpecificReturnStatement Psalm does not detect that args keys is always defined */
         return $newObjectsData;
     }
 
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -22,6 +22,9 @@ use LogicException;
 use ReflectionClass;
 use TypeError;
 
+use function array_flip;
+use function array_key_exists;
+use function array_keys;
 use function array_map;
 use function array_merge;
 use function count;
@@ -94,6 +97,13 @@ abstract class AbstractHydrator
     protected $_hints = [];
 
     /**
+     * Precomputed information about how result's rows should be mapped to object or scalars.
+     *
+     * @var array<string, mixed>
+     */
+    protected $_rowMappingInfo = [];
+
+    /**
      * Initializes a new instance of a class derived from <tt>AbstractHydrator</tt>.
      *
      * @param EntityManagerInterface $em The EntityManager to use.
@@ -134,6 +144,7 @@ abstract class AbstractHydrator
         $evm->addEventListener([Events::onClear], $this);
 
         $this->prepare();
+        $this->prepareRowMappingInfo();
 
         return new IterableResult($this);
     }
@@ -181,6 +192,7 @@ abstract class AbstractHydrator
         $evm->addEventListener([Events::onClear], $this);
 
         $this->prepare();
+        $this->prepareRowMappingInfo();
 
         while (true) {
             $row = $this->statement()->fetchAssociative();
@@ -265,6 +277,7 @@ abstract class AbstractHydrator
 
         $this->_em->getEventManager()->addEventListener([Events::onClear], $this);
         $this->prepare();
+        $this->prepareRowMappingInfo();
 
         try {
             $result = $this->hydrateAllData();
@@ -339,10 +352,11 @@ abstract class AbstractHydrator
     {
         $this->statement()->free();
 
-        $this->_stmt          = null;
-        $this->_rsm           = null;
-        $this->_cache         = [];
-        $this->_metadataCache = [];
+        $this->_stmt           = null;
+        $this->_rsm            = null;
+        $this->_rowMappingInfo = [];
+        $this->_cache          = [];
+        $this->_metadataCache  = [];
 
         $this
             ->_em
@@ -396,6 +410,7 @@ abstract class AbstractHydrator
      *                                             row, grouped by their
      *                                             component alias.
      * @psalm-return array{
+     *                   ids: array<array-key, array<array-key, mixed>>,
      *                   data: array<array-key, array>,
      *                   newObjects?: array<array-key, array{
      *                       class: mixed,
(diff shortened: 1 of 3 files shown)


beberlei:
Hey @tucksaun, this looks like fantastic work! Thank you! It will take some time to review the PR through as its touching essentially everything about hydrators, so I am not sure when we will get to it at this time.

sips-richard:
I found this PR in relation to an issue we've been having when hydrating a graph with a couple left joins. Specifically that Doctrine calls convertToPHPValue() on all the NULL values for each column in any empty left join records, prior to hydrating the entities and then discarding those empty records. We noticed this because we have a custom data type that throws an exception if it encounters a NULL (the fix for now was to eliminate the check).

But I was wondering whether it is necessary for the calls to convertToPHPValue() up front when ultimately the record ID might be NULL and the empty data set gets thrown in the trash, or could the code could be reworked to only convert the values for columns belonging to entities that will actually be hydrated?

tucksaun:
> Hey @tucksaun, this looks like fantastic work! Thank you! It will take some time to review the PR through as its touching essentially everything about hydrators, so I am not sure when we will get to it at this time.

@beberlei sure, I understand as this is quite a change.
If it can help you: I tried to keep a "logical" order in commits so that maybe it is easier to follow.
Also, feel free to ping me if you want to chat about it if you want.

I will try to maintain this PR as up-to-date as possible in the meantime.

tucksaun:
> I found this PR in relation to an issue we've been having when hydrating a graph with a couple left joins. Specifically that Doctrine calls convertToPHPValue() on all the NULL values for each column in any empty left join records, prior to hydrating the entities and then discarding those empty records. We noticed this because we have a custom data type that throws an exception if it encounters a NULL (the fix for now was to eliminate the check).
> 
> But I was wondering whether it is necessary for the calls to convertToPHPValue() up front when ultimately the record ID might be NULL and the empty data set gets thrown in the trash, or could the code could be reworked to only convert the values for columns belonging to entities that will actually be hydrated?

@sips-richard I would say that the hydrator has to maintain the call to `convertToPHPValue` because the logic of what is done with a NULL value in the database might be reflected differently in PHP depending on the needs and the type implementation.
But if I got your problem correctly, the number of calls might be lower with this patch.

greg0ire:
Hi! `git rebase --exec vendor/bin/phpcbf` should fix some (but not all) cs issues :)

tucksaun:
@greg0ire I though I already fixed them all and only casually rebased this PR today so thank you for the heads up about he CS issues.
They should be fixed now.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+
+            if ($value !== null && isset($cacheKeyInfo['enumType'])) {
+                $value = $this->buildEnum($value, $cacheKeyInfo['enumType']);
+            }
+
+            $newObjectsData[$objIndex]['class']           = $cacheKeyInfo['class'];
+            $newObjectsData[$objIndex]['args'][$argIndex] = $value;
+        }
+
+        /** @psalm-suppress LessSpecificReturnStatement Psalm does not detect that args keys is always defined */
```
derrabus, line 574: Don't suppress Psalm errors via comments. We usually configure this in psalm.xml.dist.

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
```diff
@@ -538,8 +558,8 @@ protected function hydrateRowData(array $row, array &$result)
         // Append scalar values to mixed result sets
         if (isset($rowData['scalars'])) {
             if (! isset($resultKey)) {
-                $resultKey = isset($this->resultSetMapping()->indexByMap['scalars'])
-                    ? $row[$this->resultSetMapping()->indexByMap['scalars']]
+                $resultKey = isset($resultSetMapping->indexByMap['scalars'])
+                    ? $row[$resultSetMapping->indexByMap['scalars']]
                     : $this->resultCounter - 1;
```
derrabus, line 563:
```suggestion
                $resultKey = $resultSetMapping->indexByMap['scalars'] ?? $this->resultCounter - 1;
```

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC6303Test.php
```diff
@@ -80,7 +80,7 @@ private function assertHydratedEntitiesSameToPersistedOnes(array $persistedEntit
         self::assertCount(count($persistedEntities), $entities);
 
         foreach ($entities as $entity) {
-            self::assertEquals($entity, $persistedEntities[$entity->id]);
+            self::assertEquals($persistedEntities[$entity->id], $entity);
```
derrabus: This change looks unrelated (and also not correct).

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+            $rowData['newObjects'] = $this->gatherRowDataFromNewObjectsMapping($data);
+        }
+
+        return $rowData;
+    }
+
+    /**
+     * @param mixed[] $data SQL Result Row.
+     * @psalm-param array<array-key, string> $id                 Dql-Alias => ID-Hash.
+     * @psalm-param array<array-key, bool>   $nonemptyComponents Does this DQL-Alias has at least one non NULL value?
```
derrabus:
```suggestion
     * @param string[] $id                 Dql-Alias => ID-Hash.
     * @param bool[]   $nonemptyComponents Does this DQL-Alias has at least one non NULL value?
```

No need to vendor-prefix those.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+     *
+     * @psalm-return array<array-key, array<array-key, mixed>> An array with all the ids (name => value) of the row.
```
derrabus:
```suggestion
     * @return mixed[][] An array with all the ids (name => value) of the row.
```

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+                $nonemptyComponents[$dqlAlias] = true;
             }
+        }
+
+        return $ids;
+    }
+
+    /**
+     * @param mixed[] $data SQL Result Row.
+     * @psalm-param string $dqlAlias                 Dql-Alias
```
derrabus:
```suggestion
```

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+     *
+     * @psalm-return array<array-key, array> An array with all the fields
```
derrabus:
```suggestion
     * @return mixed[][] An array with all the fields
```

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+     *                                             (name => value) of the data
+     *                                             row, grouped by their
+     *                                             component alias.
+     */
+    protected function gatherRowDataFromDQLMapping(array $data, string $dqlAlias)
```
derrabus: Does this method have to be protected? And please use native return types on new methods if you can.

```suggestion
    private function gatherRowDataFromDQLMapping(array $data, string $dqlAlias): array
```

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+    }
+
+    /**
+     * @param mixed[] $data SQL Result Row.
+     * @psalm-param array<array-key, string> $id                 Dql-Alias => ID-Hash.
+     * @psalm-param array<array-key, bool>   $nonemptyComponents Does this DQL-Alias has at least one non NULL value?
+     *
+     * @psalm-return array<array-key, array<array-key, mixed>> An array with all the ids (name => value) of the row.
+     */
+    protected function gatherRowIDs(array $data, array &$id, array &$nonemptyComponents)
```
derrabus:
```suggestion
    private function gatherRowIDs(array $data, array &$id, array &$nonemptyComponents): array
```

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+            $rowData[$fieldName] = $value;
         }
 
         return $rowData;
     }
 
+    /**
+     * @param mixed[] $data SQL Result Row.
+     *
+     * @psalm-return array<array-key, mixed>
```
derrabus:
```suggestion
     * @return mixed[]
```

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+     */
+    protected function gatherRowDataFromScalarsMapping(array $data)
```
derrabus:
```suggestion
    private function gatherRowDataFromScalarsMapping(array $data): array
```

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -396,94 +410,173 @@ abstract protected function hydrateAllData();
...
+     * @param mixed[] $data SQL Result Row.
+     *
+     * @psalm-return array<array-key, array{
+     *                       class: mixed,
+     *                       args: array
+     *                   }>
+     *
+     * @psalm-suppress MoreSpecificReturnType Psalm does not detect that args keys is always defined
+     */
+    protected function gatherRowDataFromNewObjectsMapping(array $data)
```
derrabus:
```suggestion
    private function gatherRowDataFromNewObjectsMapping(array $data): array
```

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -525,6 +618,51 @@ protected function gatherScalarRowData(&$data)
         return $rowData;
     }
 
+    /** @return void */
+    protected function prepareRowMappingInfo()
```
derrabus:
```suggestion
    private function prepareRowMappingInfo(): void
```

derrabus (changes requested):

sips-richard:
> > I found this PR in relation to an issue we've been having when hydrating a graph with a couple left joins. Specifically that Doctrine calls convertToPHPValue() on all the NULL values for each column in any empty left join records, prior to hydrating the entities and then discarding those empty records. We noticed this because we have a custom data type that throws an exception if it encounters a NULL (the fix for now was to eliminate the check).
> > But I was wondering whether it is necessary for the calls to convertToPHPValue() up front when ultimately the record ID might be NULL and the empty data set gets thrown in the trash, or could the code could be reworked to only convert the values for columns belonging to entities that will actually be hydrated?
> 
> @sips-richard I would say that the hydrator has to maintain the call to `convertToPHPValue` because the logic of what is done with a NULL value in the database might be reflected differently in PHP depending on the needs and the type implementation. But if I got your problem correctly, the number of calls might be lower with this patch.

@tucksaun Absolutely this will speed up our code, but I will try to clarify on our problem scenario because I think a change in the internals might be beneficial here.

Our ORM query included several left joins. Internally Doctrine was passing all the null values for any non-existent left join entities to the convertToPHPValue() method and only determining afterward that there was no relationship and then throwing these values away.

The `convertToPHPValue` method on the type for one of our columns did not permit null values and would throw an exception (because this would never happen in normal operation), but because Doctrine is hydrating null values for the empty left joins it was.

**My question therefore is:** can the logic be reworked in such a way that Doctrine could determine whether or not the ID columns represent a joined entity or an empty left join *up front*, prior to any hydration.

It would save a lot of function calls where there are empty left joins for a start.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
tucksaun: @derrabus I don't manage to suppress it for this specific method inside `psalm.xml`.
And to suppress it for the whole file looks a bit too wide to me.

Do you have any recommendations?
Or should I add this to the baseline?
Or does it mean this is a legitimate use case to you?

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
tucksaun: hum does not look correct to me as we are accessing `$row`, not only `$resultSetMapping->indexByMap` ? 🤔

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
tucksaun: this method (and the other ones introduced here) are made protected to be callable from the Object Hydrator

tucksaun: (return types added)

NoiseByNorthwest:
It looks very close to the issue I've described here https://github.com/doctrine/orm/issues/8390
And I've shared a custom hydrator which fixes the useless "repeated rows hydration" with a simple caching logic.

Does my workaround make sense ? Should it be implemented into ORM ?

Issue #10039, opened 2022-09-11
Allow to lock entity with EntityManager#refresh()


michnovka:
### Feature Request

Recently I stumbled upon how [useless](https://stackoverflow.com/questions/73665780/when-does-entitymanagerlock-even-have-sense) the `EntityManager#lock()` method is in scenarios where you need up-to-date data during transaction.

Now the way to go is to use `$entity = $em->find($entity->id, LockMode::PESSIMISTIC_WRITE);`

It would be nice and more readable if we would allow to call `EntityManager#refresh()` and specify lockMode to acquire a lock on already loaded entity whilst making sure we have the latest version of it at runtime.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

I would like to extend the `EntityManager#refresh()` to:

```php
public function refresh($entity, $lockMode = null)
```

This will require similar change to `UnitOfWork#refresh()` signature, but otherwise should be a simple feature to implement.

What do you think?

PR #10040, opened 2022-09-11
Add lockMode to EntityManager#refresh()

michnovka:
Attempt to implement https://github.com/doctrine/orm/issues/10039

### DIFF
--- a/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
+++ b/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
@@ -216,7 +216,7 @@ abstract class EntityManagerDecorator extends ObjectManagerDecorator implements 
      */
     public function refresh($object, $lockMode = null)
     {
-        return $this->wrapped->refresh($object, $lockMode);
+        $this->wrapped->refresh($object, $lockMode);
     }
 
     /**
--- a/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
+++ b/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
@@ -9,6 +9,8 @@ use Doctrine\ORM\EntityRepository;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\Persistence\ObjectManagerDecorator;
 
+use function func_get_arg;
+use function func_num_args;
 use function get_debug_type;
 use function method_exists;
 use function sprintf;
@@ -214,8 +216,14 @@ abstract class EntityManagerDecorator extends ObjectManagerDecorator implements 
     /**
      * {@inheritdoc}
      */
-    public function refresh($object, $lockMode = null)
+    public function refresh($object)
     {
+        $lockMode = null;
+
+        if (func_num_args() === 2) {
+            $lockMode = func_get_arg(1);
+        }
+
         $this->wrapped->refresh($object, $lockMode);
     }
 
--- a/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
+++ b/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
@@ -220,7 +220,7 @@ abstract class EntityManagerDecorator extends ObjectManagerDecorator implements 
     {
         $lockMode = null;
 
-        if (func_num_args() === 2) {
+        if (func_num_args() > 1) {
             $lockMode = func_get_arg(1);
         }
 
--- a/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
+++ b/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
@@ -214,6 +214,14 @@ abstract class EntityManagerDecorator extends ObjectManagerDecorator implements 
     /**
      * {@inheritdoc}
      */
+    public function refresh($object, $lockMode = null)
+    {
+        return $this->wrapped->refresh($object, $lockMode);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getEventManager()
     {
         return $this->wrapped->getEventManager();
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2197,7 +2197,7 @@ class UnitOfWork implements PropertyChangedListener
 
         $lockMode = null;
 
-        if (func_num_args() === 2) {
+        if (func_num_args() > 1) {
             $lockMode = func_get_arg(1);
         }
 
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -58,6 +58,8 @@ use function array_values;
 use function assert;
 use function count;
 use function current;
+use function func_get_arg;
+use function func_num_args;
 use function get_class;
 use function get_debug_type;
 use function implode;
@@ -2183,16 +2185,21 @@ class UnitOfWork implements PropertyChangedListener
      * any local, unpersisted changes.
      *
      * @param object $entity The entity to refresh
-     * @psalm-param LockMode::*|null $lockMode
      *
      * @return void
      *
      * @throws InvalidArgumentException If the entity is not MANAGED.
      * @throws TransactionRequiredException
      */
-    public function refresh($entity, $lockMode = null)
+    public function refresh($entity)
     {
         $visited = [];
+
+        $lockMode = null;
+
+        if (func_num_args() === 2) {
+            $lockMode = func_get_arg(1);
+        }
 
         $this->doRefresh($entity, $visited, $lockMode);
     }
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2182,17 +2182,18 @@ class UnitOfWork implements PropertyChangedListener
      * Refreshes the state of the given entity from the database, overwriting
      * any local, unpersisted changes.
      *
-     * @param object $entity The entity to refresh.
+     * @param object $entity The entity to refresh
+     * @psalm-param LockMode::*|null $lockMode
      *
      * @return void
      *
      * @throws InvalidArgumentException If the entity is not MANAGED.
      */
-    public function refresh($entity)
+    public function refresh($entity, $lockMode = null)
     {
         $visited = [];
 
-        $this->doRefresh($entity, $visited);
+        $this->doRefresh($entity, $visited, $lockMode);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


michnovka:
TODO: if accepted, update https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/transactions-and-concurrency.html to reflect another case to specify LockMode

michnovka:
Are there any cases against adding the lock mode into `EntityManager#refresh()` ?

michnovka:
@derrabus is there anything I can do to have this merged into 2.14? Anything missing from this PR?

derrabus:
I really wish I had more time to review changes like this one, sorry. 😞 I'm adding it to the 2.14.0 milestone, so we don't forget it this time.

One thing though: adding parameters to public methods that are neither final nor internal is a breaking change because that would break code that overrides those methods. However, you could use `func_get_arg()` to access extra parameters passed to a method.

michnovka:
@derrabus well using `func_get_args` is such an ugly hack that I would prefer to postpone this to 3.0 rather than try to force this feature thru such obscure parameter hiding.

derrabus:
Not an option: we don't accept features for 3.0.x.

michnovka:
you mean 3.0 is under feature freeze already? or it is too soon yet?

michnovka:
@derrabus I would then suggest we add `EntityManager#refreshAndLock()` and `UnitOfWork#refreshAndLock()`. The `func_get_args` solution is really sth Id like to avoid.

derrabus:
> you mean 3.0 is under feature freeze already?

No, 3.0 is not a feature release. It won't ship a single new feature.

> The `func_get_args` solution is really sth Id like to avoid.

I don't share that sentiment, but I'm open to discussing your solution.

michnovka:
> I don't share that sentiment, but I'm open to discussing your solution.

This is your project, so I am going to implement it with your standards in mind. I just prefer to make BC break in some bigger version release. What versions would be allowed to do that? 2.14?

I checked doctrine code and I don't see such hack being used elsewhere, or do you have some example? If this is not used elsewhere to avoid BC breaks, then I really dont want to be the one blamed for starting it in the future. We can ask others also for their opinion, @beberlei @greg0ire ?

If the version which would allow changing signature of the 2 public methods `EntityManager#refreshAndLock()` and `UnitOfWork#refreshAndLock()` were to be too far in the future, I still propose we do now `refreshAndLock()`, mark it deprecated from the start and when the time comes, add the param to the `refresh()` function itself.

greg0ire:
Funny you mention that kind of hack, we got asked about it today: https://github.com/doctrine/dbal/commit/16efd099220155cdda9c1522d208205c94da9fcb#r87189625

Personally, I'm fine with it. In any case, please don't contribute BC-breaks to ORM 3 without also providing an upgrade path in ORM 2.

michnovka:
@greg0ire what is the proper upgrade path for public method signature change? It seems to me the only way to do this is to just do it and then mention it in UPGRADE.md ... There is nothing like `#[SignatureWillChange]` of the likes of `#[ReturnTypeWillChange]` or `#[Deprecated]`

greg0ire:
One proper way I can think of is using `func_get_args()`. The other way is to add another method with a different name and deprecate the first one in favor of the second one, but then you loose the name.

derrabus:
> I just prefer to make BC break in some bigger version release. What versions would be allowed to do that? 2.14?

Do you want to contribute a feature or do you want to break things? Contribute your feature in a backwards-compatible manner to 2.14. If the feature is accepted, you can clean up BC hacks in 3.0 **after** your feature has been merged.

michnovka:
I have pushed commit which doesnt break BC.

> If the feature is accepted, you can clean up BC hacks in 3.0 **after** your feature has been merged.
Ok, this is what I didnt get, I thought we wont be allowed to change func signatures even in 3.0

When this is merged and changes synced into 3.x, I will commit to the 3.0.x branch to change the signature and remove `func_get_args`, is that OK?

derrabus:
> When this is merged and changes synced into 3.x, I will commit to the 3.0.x branch to change the signature and remove `func_get_args`, is that OK?

This is the way

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2182,29 +2184,46 @@ private function doDetach(
...
      *
      * @param object $entity The entity to refresh.
      * @psalm-param array<int, object>  $visited The already visited entities during cascades.
+     * @psalm-param LockMode::*|null $lockMode
      *
      * @throws ORMInvalidArgumentException If the entity is not MANAGED.
+     * @throws TransactionRequiredException
      */
-    private function doRefresh($entity, array &$visited): void
+    private function doRefresh($entity, array &$visited, $lockMode = null): void
```
derrabus: Can we add a native parameter type for `$lockMode`?

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2221,19 +2240,21 @@ private function doRefresh($entity, array &$visited): void
...
 
     /**
      * Cascades a refresh operation to associated entities.
      *
      * @param object $entity
      * @psalm-param array<int, object> $visited
+     * @psalm-param LockMode::*|null $lockMode
      */
-    private function cascadeRefresh($entity, array &$visited): void
+    private function cascadeRefresh($entity, array &$visited, $lockMode = null): void
```
derrabus: Same here.

derrabus (changes requested):
Thank you. Please update `EntityManagerDecorator` as well. We should also update `EntityManagerInterface`, indicating the upcoming change of the method signatures.

Also, we consider the `EntityManager` class itself as final, so updating the signatures of public methods would be acceptable there.

michnovka:
I am on vacation till 7th Nov, can u make the changes yourself or it requires my commits?

derrabus:
We're not in a hurry here. Do it whenever you feel like you have time. 🙂

michnovka:
> Also, we consider the EntityManager class itself as final, so updating the signatures of public methods would be acceptable there.

@derrabus so can I get rid of the `func_get_args` code and change the signature as in the original commit? Keeping this workaround for `UnitOfWork` though as its not `final`.

michnovka:
> Thank you. Please update `EntityManagerDecorator` as well. We should also update `EntityManagerInterface`, indicating the upcoming change of the method signatures.

Neither `EntityManagerInterface` nor `EntityManagerDecorator` have the `refresh` method. Should I add it?

derrabus:
> so can I get rid of the `func_get_args` code

Yes, but only on the entity manager.

> Neither `EntityManagerInterface` nor `EntityManagerDecorator` have the `refresh` method. Should I add it?

They do, inherited from `ObjectManager` / `ObjectManagerDecorator`!

michnovka:
> They do, inherited from `ObjectManager` / `ObjectManagerDecorator`!

These are part of https://github.com/doctrine/persistence. Should I make PR there?

derrabus:
No, the Persistence lib does not care about locking.

michnovka:
@derrabus I have added the `refresh()` method to `EntityManagerDecorator` but not `EntityManagerInterface`. The `find()` method which has `lockMode` also is present only in the decorator but not in the interface.

If we want to add the `lockMode` to the interface, we would need to edit the `ObjectManager` interface from persistence, as otherwise there will always be signature mismatch, since all `EntityManagerDecorator`, `EntityManagerInterface` and `ObjectManagerDecorator` extend or implement `ObjectManager` with `refresh` function signature without `lockMode`.

lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
```diff
@@ -211,6 +211,14 @@ public function flush($entity = null)
         $this->wrapped->flush($entity);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function refresh($object, $lockMode = null)
```
derrabus: `EntityManagerDecorator` is a class that is meant to be extended, its methods are meant to be overridden. Because of that, we again cannot add new parameters here without breaking downstream implementations. `func_get_arg()` it is again, sorry.

derrabus (changes requested):

derrabus:
> @derrabus I have added the `refresh()` method to `EntityManagerDecorator` but not `EntityManagerInterface`.

You can add a `@method` annotation to the interface that documents the new signature. You should already find other examples there.

> If we want to add the `lockMode` to the interface, we would need to edit the `ObjectManager` interface from persistence

No. We simply add the method including the new parameter in 3.0. No need to change Persistence.

michnovka:
> No. We simply add the method including the new parameter in 3.0. No need to change Persistence.

The interface `EntityManagerInterface` extends `Doctrine\Persistence\ObjectManager`. The `ObjectManager` defines

```php
    /**
     * Refreshes the persistent state of an object from the database,
     * overriding any local changes that have not yet been persisted.
     *
     * @param object $object The object to refresh.
     *
     * @return void
     */
    public function refresh(object $object);
```
How can we change method signature in `EntityManagerInterface` without changing the persistence? If the signature of a method of a child interface does not match the parent, it wont run. Correct me if I'm wrong.

derrabus:
> Correct me if I'm wrong.

Sure: https://3v4l.org/9EuXT

michnovka:
@derrabus aha, I didnt realize that by using default parameter value it can work. Thanks.

I have implemented all feedback, this should be ready for a merge.

lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php
```diff
@@ -211,6 +213,20 @@ public function flush($entity = null)
...
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function refresh($object)
+    {
+        $lockMode = null;
+
+        if (func_num_args() === 2) {
```
derrabus:
```suggestion
        if (func_num_args() > 1) {
```

This might look like nit-picking, but if we want to emulate how PHP behaves if we actually had that parameter, we need to allow the caller to pass three arguments. Or fourty-two arguments.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2182,29 +2184,46 @@ private function doDetach(
...
+     * @throws TransactionRequiredException
      */
     public function refresh($entity)
     {
         $visited = [];
 
-        $this->doRefresh($entity, $visited);
+        $lockMode = null;
+
+        if (func_num_args() === 2) {
```
derrabus:
```suggestion
        if (func_num_args() > 1) {
```

derrabus:
Looks good to me. Can you please check if the documentation needs an update?

michnovka:
@derrabus docs updated

derrabus (approved):

derrabus:
Thank you.

michnovka:
@derrabus let me know when you merge this into 3.0.x, Ill work to update the signatures on that branch then

derrabus:
FTR, I've pushed 474f76fc8b3e4f7587c514ef100a023d5136548d. We usually don't baseline Psalm errors from BC layers.

derrabus:
Merged up via #10195

michnovka:
@derrabus should I update stubs at https://github.com/psalm/psalm-plugin-doctrine/tree/2.x/stubs ?

derrabus:
You should direct this question at the authors of that plugin, I guess. I won't stop you. 🙂

Issue #10031, opened 2022-09-06
Hydrating  with `enumType` as array gives *string* instead of Enum (DQL with join)


mzk:
### Bug Report

When I have a query:
```
return $this->entityManager->createQuery('SELECT o.orderStatus FROM Entity\Order o')
            ->getResult(AbstractQuery::HYDRATE_ARRAY);
```
Then `$result[0]['orderStatus']` is instance of BackendEnum OrderStatus.

But when I have a query:
```
return $this->entityManager->createQuery('SELECT o FROM Entity\Order o')
            ->getResult(AbstractQuery::HYDRATE_ARRAY);
```
Then `$result[0]['orderStatus']` is simple string.

Our Order entity field:
```
    #[Column(type: Types::STRING, nullable: false, enumType: OrderStatus::class, options: ['charset' => 'utf8', 'collation' => 'utf8_unicode_ci'])]
    private OrderStatus $orderStatus;
```

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.13.1

#### Summary

It should be the same result for both cases. 

#### Current behavior

`SELECT o` retuns string, `SELECT o.orderStatus` returns the BackendEnum

#### How to reproduce

Use entity as [described](https://www.doctrine-project.org/2022/01/11/orm-2.11.html) and fetch from query builder as an array.

#### Expected behavior
`SELECT o` returns an array that contains BackendEnum.

for ref: https://github.com/doctrine/orm/issues/9622

michnovka:
I generally try to avoid Array Hydration, but I agree that this should work nonetheless. If you use standard object hydration (default) is there any issue?

mzk:
The default object hydration works fine.

janedbal:
The biggest issue I see here is the same reasoning I wrote in #9622:

> **Migrating from enums used as custom types to this native form is impossible as it behaves differently**

derrabus:
Duplicate of #9947

PR #9995, opened 2022-08-16
Fix EnumType not being hydrated with HYDRATE_ARRAY

kvush:
Fixes #9947

add new processEnum method in ArrayHydrator

PR #10041, opened 2022-09-12
Fix ArrayHydration of enums

michnovka:
Fix for https://github.com/doctrine/orm/issues/10031

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -464,6 +464,10 @@ abstract class AbstractHydrator
                         break;
                     }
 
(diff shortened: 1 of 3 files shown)


derrabus:
This had been fixed by #9995 already, but I must admit that I find your solution more elegant.

@greg0ire @beberlei WDYT, shall we revert #9995 and go with this solution instead?

greg0ire:
I think we should. @michnovka can you add a revert commit to this PR? After that we should release because 2 fixes for the same bug means it must be quite painful.

derrabus (approved):

derrabus:
I've pushed a revert commit 5f29fcdea238a23b4fb28805940041ee1752d1f8 before merging this one. Thanks @michnovka and @kvush!

PR #10062, opened 2022-09-24
Update Index.php

lxregistry:
fix named arguments order like in UniqueConstraint and like in docs to avoid phpstorm codestyle warnings


greg0ire:
What warnings are you talking about?

derrabus:
Sorry, but that PhpStorm inspection is pointless. There's nothing to fix here.

PR #10065, opened 2022-09-26
Lazy and eager collection refresh inconsistency

goetas:
When there is a `cascade={"refresh"}`,  `EAGER` collections are refreshed, `LAZY` one are not refreshed, even if initialized...

The issue lies in the fact that `\Doctrine\ORM\UnitOfWork::cascadeRefresh` is called after the entity holding that collection has been refreshed, thus the lazy collections in there that have been edited are not refreshed (because the initial entity refresh has marked them as not initialized). 

Moving the `cascadeRefresh` before the entity is refreshed.. but that might break other things... i'm waiting for github actions to tell.

I fear that the same should be done for `\Doctrine\ORM\UnitOfWork::cascadeDetach`

### DIFF
--- a/src/UnitOfWork.php
+++ b/src/UnitOfWork.php
@@ -2473,13 +2473,13 @@ EXCEPTION
             throw ORMInvalidArgumentException::entityNotManaged($entity);
         }
 
+        $this->cascadeRefresh($entity, $visited, $lockMode);
+
         $this->getEntityPersister($class->name)->refresh(
             array_combine($class->getIdentifierFieldNames(), $this->entityIdentifiers[$oid]),
             $entity,
             $lockMode
         );
-
-        $this->cascadeRefresh($entity, $visited, $lockMode);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


goetas:
first commit add the test, the second provides a fix for it

SenseException (dismissed):
Builds are green. Do you still expect BC breaks despite that?

goetas:
i do not think that this would cause bc breaks. 

one of my questions here is: should i do the same fore the detach cascade operation? i'm pretty sure it has the same buggy behavior

goetas:
hi, are there any comment on this? is there an interest in merging it?

goetas:
I think that this is still relevant for 2.x...

greg0ire:
> one of my questions here is: should i do the same fore the detach cascade operation? I'm pretty sure it has the same buggy behavior

Well then yes. Also, please rebase on 2.20.x :pray: 

> hi, are there any comment on this? is there an interest in merging it?

The fact that you did not break any test by changing the code makes me think it was not intentionally like this, and the patch looks sensible, so I'm OK with this on principle, will merge after my comments are addressed. Thanks for working on this.

goetas:
@greg0ire thanks for looking again into this! I have rebased it and targeted 2.20.x as base branch. 
Please let me know if anything else is needed. If this gets merged, I would like to do the same with `cascade detach`

goetas:
The psalm failure seems unrelated to my change (it is failing in the base branch as well)

greg0ire:
Thanks @goetas , please go ahead with `detach()`

goetas:
Created https://github.com/doctrine/orm/pull/11717 for the detach feature

Issue #10071, opened 2022-09-27
Fetched Entity with native Enum field with non empty changeSet (Enum <=> string)


mzk:
### Bug Report

```php
// Order.php
    #[Column(type: Types::STRING, nullable: false, enumType: OrderStatus::class, options: ['charset' => 'utf8', 'collation' => 'utf8_unicode_ci'])]
    private OrderStatus $orderStatus;


// code.
$order = new Order();
$em->flush();
$em->clear();

$freshOrder = $em->createQueryBuilder()->select('o')->from(Order::class, 'o')->setMaxResults(1)->getQuery()->getSingleResult();

$uow->computeChangeSets();
$em->getUnitOfWork()->isScheduledForDelete($order); // returns true.
```

(The freshEneity is scheduled for the update because in the changeSet is 'strinngValue' of Enum::Foo.

But when I get the Entity like this

```php
$freshEntity = $this->getEntityManager()->find(Order::class, $orderId);
$uow->computeChangeSets();
$em->getUnitOfWork()->isScheduledForUpdate($order); // returns false, there is no update on fresh Entity.
```

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.13.x-dev

#### Summary

The $freshEntity with native Enum field shouldn't scheduled for updated

the failing test case is here https://github.com/doctrine/orm/pull/10073

PR #10073, opened 2022-09-27
add failing testcase test enum Change sets

mzk:
testcase for reported issue https://github.com/doctrine/orm/issues/10071


greg0ire:
Used in #10074

PR #10074, opened 2022-09-27
Fix change set computation with enums

### LINKED ISSUE #10071: Fetched Entity with native Enum field with non empty changeSet (Enum <=> string)

HypeMC:
Fixes #10071 , uses test cases from #10073 by @mzk

My attempt at solving the issue. The `ReflectionEnumProperty::getValue()` method returns the value of the enum, while some hydrators pass an enum object in the data set to the `UnitOfWork::$originalEntityData` array.

The `SimpleObjectHydrator` doesn't convert strings into enums, while the `ObjectHydrator` (or to be more precise the `AbstractHydrator::gatherRowData()` method) does. Even though this doesn't seem to be causing any issues, the inconsistency seems unintentional to me, so if you think it should be fixed please let me know and I'll open a PR.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM;
 
+use BackedEnum;
 use DateTimeInterface;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
@@ -742,6 +743,10 @@ class UnitOfWork implements PropertyChangedListener
                 }
 
                 $orgValue = $originalData[$propName];
+
+                if ($orgValue instanceof BackedEnum) {
+                    $orgValue = $orgValue->value;
+                }
 
                 // skip if value haven't changed
                 if ($orgValue === $actualValue) {
--- a/tests/Doctrine/Tests/ORM/Functional/EnumTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/EnumTest.php
@@ -260,6 +260,46 @@ class EnumTest extends OrmFunctionalTestCase
         self::assertEqualsCanonicalizing([Unit::Gram, Unit::Meter], $result[0]->supportedUnits);
     }
 
+    public function testEnumChangeSetsSimpleObjectHydrator(): void
+    {
+        $this->setUpEntitySchema([Card::class]);
+
+        $card       = new Card();
+        $card->suit = Suit::Clubs;
+
+        $this->_em->persist($card);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $result = $this->_em->find(Card::class, $card->id);
+
+        $this->_em->getUnitOfWork()->computeChangeSets();
+
+        self::assertFalse($this->_em->getUnitOfWork()->isScheduledForUpdate($result));
+    }
+
+    public function testEnumChangeSetsObjectHydrator(): void
+    {
+        $this->setUpEntitySchema([Card::class]);
+
+        $card       = new Card();
+        $card->suit = Suit::Clubs;
+
+        $this->_em->persist($card);
+        $this->_em->flush();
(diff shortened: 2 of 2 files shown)


michnovka:
This is a good fix.

I think `SimpleObjectHydrator` should also convert strings into enums.

HypeMC:
> This is a good fix.
> 
> I think `SimpleObjectHydrator` should also convert strings into enums.

See #10088

michnovka:
@derrabus this is an important fix for enums, can we have it merged for the next 2.13 point release?

derrabus (approved):

derrabus:
Thank you @HypeMC.

mroeling:
Is there an e.t.a. on a new release with this fix please?

greg0ire:
@mroeling we are waiting on #10088 to release this.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -743,6 +744,10 @@ public function computeChangeSet(ClassMetadata $class, $entity)
 
                 $orgValue = $originalData[$propName];
 
+                if ($orgValue instanceof BackedEnum) {
+                    $orgValue = $orgValue->value;
+                }
+
                 // skip if value haven't changed
                 if ($orgValue === $actualValue) {
```
shadowhand, line 752: This PR is broken, and I'm pretty sure this is why. Based on the changes in line 747, we probably also need to have:

```php
if ($actualValue instanceof BackedEnum) {
    $actualValue = $actualValue->value;
}
```

shadowhand: https://github.com/doctrine/orm/issues/10249

michnovka: I don think so. Check the code a bit above:

https://github.com/HypeMC/orm/blob/a1ed32f697f66d33de42fe96b1ea90e873ba2d5f/lib/Doctrine/ORM/UnitOfWork.php#L670

then 

https://github.com/HypeMC/orm/blob/a1ed32f697f66d33de42fe96b1ea90e873ba2d5f/lib/Doctrine/ORM/UnitOfWork.php#L699

So actual value is fetched through reflection. And if you check the `ReflectionEnumProperty::getValue()`

https://github.com/HypeMC/orm/blob/a1ed32f697f66d33de42fe96b1ea90e873ba2d5f/lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php#L36-L63

It returns the `$enum->value` already. Therefore your proposed 

```php
if ($actualValue instanceof BackedEnum) {
    $actualValue = $actualValue->value;
}
```

seems redundant.

shadowhand: I could be wrong. All I know is that this patch broke several of our entities that use enums, and that seemed like a possibility. I haven't tested the theory.

PR #10075, opened 2022-09-28
Infer types in `OneToManyPersister::deleteEntityCollection`

norkunas:
Similar to #8401

In our app we are using Symfony Uuid and when collection needs to be cleared the deletion query is executed without types therefore Uuid casts to string instead of to binary and no relations are deleted.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Persisters/OneToManyPersisterTest.php
+++ b/tests/Doctrine/Tests/ORM/Persisters/OneToManyPersisterTest.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Persisters;
+
+use Doctrine\DBAL\Types\Type as DBALType;
+use Doctrine\ORM\PersistentCollection;
+use Doctrine\ORM\Persisters\Collection\OneToManyPersister;
+use Doctrine\Tests\DbalTypes\CustomIdObject;
+use Doctrine\Tests\DbalTypes\CustomIdObjectType;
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/ORM/Persisters/OneToManyPersisterTest.php
```diff
@@ -0,0 +1,65 @@
...
+        $persister = new OneToManyPersister($em);
+        $persister->delete($parent->children);
+
+        $conn = $em->getConnection();
+        assert($conn instanceof ConnectionMock);
+
+        $updates       = $conn->getExecuteStatements();
+        $lastStatement = array_pop($updates);
+
+        self::assertSame('DELETE FROM onetomanypersister_child WHERE parent_id = ?', $lastStatement['sql']);
```
derrabus, line 61:
```suggestion
        self::assertSame('DELETE FROM onetomanypersister_child WHERE parent_id = ?', $lastStatement['sql']);
```

Running assertions on generated SQL is always brittle. We should run the test against a real database and assert that the given record has been deleted properly.

derrabus (changes requested):
Thank you. Sorry that we're picking up your PR that late.

tests/Doctrine/Tests/ORM/Persisters/OneToManyPersisterTest.php
norkunas: There was no tests at all for this persister,so I've just copied from another one and adapted it.. will see If I could manage to add a better one.. Do you have any suggestions where I should add it?

PR #10076, opened 2022-09-28
Widen LazyCriteriaCollection::contains() param type

greg0ire:
That method is supposed to be able to deal with any mixed.


greg0ire:
Will make sense with `doctrine/collections` 2

PR #10084, opened 2022-09-30
Migrate collections to PHP 8 syntax

greg0ire:


lib/Doctrine/ORM/LazyCriteriaCollection.php
```diff
@@ -87,21 +79,15 @@ public function contains($element)
         return $this->entityPersister->exists($element, $this->criteria);
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    public function matching(Criteria $criteria)
+    public function matching(Criteria $criteria): ReadableCollection&Selectable
```
greg0ire: [Maybe we should switch to Psalm v5 because of this](https://github.com/vimeo/psalm/issues/8229)

lib/Doctrine/ORM/LazyCriteriaCollection.php
```diff
@@ -87,21 +79,16 @@ public function contains($element)
         return $this->entityPersister->exists($element, $this->criteria);
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    public function matching(Criteria $criteria)
+    /** @return ReadableCollection<TKey, TValue>&Selectable<TKey, TValue> */
+    public function matching(Criteria $criteria): ReadableCollection
```
derrabus, line 83: Can we add `Selectable` to the native return type?

```suggestion
    public function matching(Criteria $criteria): ReadableCollection&Selectable
```

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -44,65 +43,47 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
...
      */
-    private $association;
-
-    /**
-     * The EntityManager that manages the persistence of the collection.
-     *
-     * @var EntityManagerInterface|null
-     */
-    private $em;
+    private array|null $association;
```
derrabus:
```suggestion
    private array|null $association = null;
```

lib/Doctrine/ORM/LazyCriteriaCollection.php
greg0ire: I haven't found how to do so, see https://github.com/doctrine/orm/pull/10084#discussion_r984973270

derrabus: So we're not adding the correct native return type because it'll make Psalm sad? Shall we rather ignore the error?

greg0ire: It haven't found how to ignore it: it's a crash, not a simple error.

derrabus: 😭

greg0ire: @derrabus I've discussed this here: https://symfony-devs.slack.com/archives/C8SFXTD2M/p1665220038738889

I see 4 options:
1. wait for Psalm 5 before merging this;
2. merge without the intersection type, wait for a merge up between Psalm `4.x` and `master` and switch all branches to Psalm `master` branch, and add the type;
3. merge without the intersection type, wait for Psalm 5 to be released and add the type then
4. switch to Psalm `master` branch now and add the type.

I don't see switching only 3.0.x to another version of Psalm as a reasonable option.

2 and 3 have my preference, what about you?

derrabus: 2 sounds good.

greg0ire:
Blocked by https://github.com/doctrine/orm/pull/10116 and a merge up

lib/Doctrine/ORM/LazyCriteriaCollection.php
greg0ire: Extracted that into a new issue: https://github.com/doctrine/orm/issues/10118

derrabus (approved):

PR #10058, opened 2022-09-22
Fix using enums with the QueryBuilder

HypeMC:
Fixes #10057.

After upgrading to version `2.13.2` I started getting the following error:

```
App\Enum\TestEnum::from(): Argument #1 ($value) must be of type string, App\Enum\TestEnum given
```

It happens when an enum is used with the `QueryBuilder`, eg:

```php
$testRepository
    ->createQueryBuilder('t')
    ->where('t.enum = :type')
    ->setParameter('type', TestEnum::Foo)
    ->getQuery()->getResult()
;
```

Caused by #10041.

Since my knowledge of the ORM code base is limited I'm not sure if this is the best place to fix the bug, so any suggestions are appreciated.

PR #10088, opened 2022-10-01
Ensure consistent original data with enums

HypeMC:
As mentioned in #10074, the `SimpleObjectHydrator` doesn't convert strings into enums, while the `ObjectHydrator` does, causing inconsistent values in the `UnitOfWork::$originalEntityData` array. This PR aims to fix that.

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
@@ -697,7 +697,7 @@ abstract class AbstractHydrator
      *
      * @return BackedEnum|array<BackedEnum>
      */
-    private function buildEnum($value, string $enumType)
+    final protected function buildEnum($value, string $enumType)
     {
         if (is_array($value)) {
             return array_map(static function ($value) use ($enumType): BackedEnum {
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -6,9 +6,11 @@ namespace Doctrine\ORM\Internal\Hydration;
 
 use Doctrine\ORM\Internal\SQLResultCasing;
 use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\ORM\Query;
 use Exception;
 use RuntimeException;
+use ValueError;
 
 use function array_keys;
 use function array_search;
@@ -138,6 +140,21 @@ class SimpleObjectHydrator extends AbstractHydrator
             if (isset($cacheKeyInfo['type'])) {
                 $type  = $cacheKeyInfo['type'];
                 $value = $type->convertToPHPValue($value, $this->_platform);
+            }
+
+            if ($value !== null && isset($cacheKeyInfo['enumType'])) {
+                $originalValue = $value;
+                try {
+                    $value = $this->buildEnum($originalValue, $cacheKeyInfo['enumType']);
+                } catch (ValueError $e) {
+                    throw MappingException::invalidEnumValue(
+                        $entityName,
+                        $cacheKeyInfo['fieldName'],
+                        (string) $originalValue,
+                        $cacheKeyInfo['enumType'],
+                        $e
+                    );
+                }
             }
 
             $fieldName = $cacheKeyInfo['fieldName'];
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -1506,6 +1506,9 @@ class BasicEntityPersister implements EntityPersister
         $columnAlias  = $this->getSQLColumnAlias($fieldMapping['columnName']);
 
         $this->currentPersisterContext->rsm->addFieldResult($alias, $columnAlias, $field);
+        if (! empty($fieldMapping['enumType'])) {
+            $this->currentPersisterContext->rsm->addEnumResult($columnAlias, $fieldMapping['enumType']);
+        }
 
         if (isset($fieldMapping['requireSQLConversion'])) {
             $type = Type::getType($fieldMapping['type']);


michnovka:
I think this is the last piece of the puzzle to have consistent treating of enums inside the internals. The changes are compatible with the last PR https://github.com/doctrine/orm/pull/10058/files merged into 2.13.3, Id like to merge this into the next 2.13 point release together with https://github.com/doctrine/orm/pull/10074 @derrabus

derrabus:
Can you please rebase on 2.13.x? I'd like to see that this PR does not break with the changes of #10074.

michnovka:
@derrabus https://github.com/doctrine/orm/pull/10119 check this, all checks pass, so you can merge this one and close my PR. Thanks

HypeMC:
@derrabus Done

EDIT:
Oh looks like @michnovka beat me to it :smile:

tests/Doctrine/Tests/ORM/Functional/EnumTest.php
```diff
@@ -347,15 +349,10 @@ public function testEnumWithNonMatchingDatabaseValueThrowsException(string $card
             [$metadata->fieldMappings['id']['columnName'] => $card->id]
         );
 
-        $this->expectException(MappingException::class);
```
derrabus: Why do we need to change the expectation here? This could qualify as a breaking change if calling code expects a `MappingException` in this scenario.

HypeMC: @derrabus Before the change the string was converted to an enum in the `ReflectionEnumProperty` class https://github.com/doctrine/orm/blob/1ed0057276b3846ba6f1c3e77f926bb69467a033/lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php#L88-L107
Now it's converted earlier in the `AbstractHydrator` class instead
https://github.com/doctrine/orm/blob/1ed0057276b3846ba6f1c3e77f926bb69467a033/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php#L700-L709

Since the `AbstractHydrator` class throws a different exception the test fails. I'm not sure if this inconsistency in exceptions is a bug itself.

michnovka: @HypeMC we should wrap with the same try-catch block and throw MappingException I think

michnovka: @HypeMC can you please incorporate this change?

HypeMC: Yes, I've already started, unfortunately it's not as trivial as I thought. The current exception message contains the class name, which is currently not always available. I'll try to finish the PR in the following days.

michnovka: replace 
https://github.com/doctrine/orm/blob/b41f5ec60530c87b48fc5ccd2c93729bc8060f64/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php#L143-L145 
with
```php
            if ($value !== null && isset($cacheKeyInfo['enumType'])) {
                try {
                    $value = $this->buildEnum($value, $cacheKeyInfo['enumType']);
                } catch (ValueError $e) {
                    throw MappingException::invalidEnumValue(
                        $entityName,
                        $cacheKeyInfo['fieldName'],
                        (string) $value,
                        $cacheKeyInfo['enumType'],
                        $e
                    );
                }
            }
```

One issue though - the `AbstractHydrator::hydrateRowData()` throws only `HydrationException`, so throwing `MappingException` from here will result in psalm errors. @derrabus what do you suggest, should we update the `@throws` PHPDoc or should we wrap the Mapping exception into HydrationException (this by the same logic could be considered BC break)

HypeMC: @michnovka Good point, I wanted to change the `buildEnum` method to throw a `MappingException`, but that would actually also be a BC break, so this makes the most sense.

stephanvierkant:
What's the status of this PR? It's not clear to me what should be done to get this merged.

michnovka (dismissed):
I think its OK now

stephanvierkant:
@derrabus WDYT?

stephanvierkant (dismissed):

michnovka:
@derrabus can we add this to the 2.13.4 milestone? Thanks

derrabus:
Can we write a functional test that reproduces the problem we're attempting to fix here?

michnovka:
@derrabus Well this really does not fix any actual functional bug, its just inconsistency in how originalEntityData is populated. Some hydrators store enums, some store strings (enum values). 

See original post - https://github.com/doctrine/orm/pull/10074#issue-1388454813

> The SimpleObjectHydrator doesn't convert strings into enums, while the ObjectHydrator (or to be more precise the AbstractHydrator::gatherRowData() method) does. Even though this doesn't seem to be causing any issues, the inconsistency seems unintentional to me, so if you think it should be fixed please let me know and I'll open a PR.

With this fix, all hydrators store originalEntityData enum fields as native enums.

michnovka:
@derrabus have a look here - https://github.com/doctrine/orm/pull/10198

derrabus:
So it's really just about the internal state of the persisters? Why's that so important to you that you want to push it into a bugfix release?

michnovka:
well we introduced this inconsistency with other enum-related PRs that were part of 2.13.x ( https://github.com/doctrine/orm/pull/10041 and https://github.com/doctrine/orm/pull/10058 ) and this issue did have functional impact prior to https://github.com/doctrine/orm/pull/10074 . So I wanted to close this last piece of related issues. TBH I did think it does have functional implication as I didnt debug myself the actual effects of https://github.com/doctrine/orm/pull/10074 on this issue. Its clear now it doesnt, but still worth a merge into 2.13.x for the above mentioned reasons

NoiseByNorthwest:
It has functionnal implications, our application does not work with 2.13.2 & 2.13.3 due do enum related bugs. For instance this issue will cause false positive changeset (same logical enum value but different type).

derrabus:
@NoiseByNorthwest Would you be able to provide a functional test that reproduces the issue your application has?

michnovka:
@NoiseByNorthwest changeset computations are fixed by https://github.com/doctrine/orm/pull/10074 which is planned for 2.13.4 release

NoiseByNorthwest:
> @NoiseByNorthwest Would you be able to provide a functional test that reproduces the issue your application has?

I have a FooEntity type with a (many-to-one) relationship to a BarEntity type. The BarEntity type has an enum property.

When I create a FooEntity from an existing BarEntity, the resolved changeset will consider the BarEntity as updated since its enum property will be wrongly considered as changed.

That causes, among other annoyances, the BarEntity::updatedAt property to be updated.

greg0ire:
People are starting to ask me an ETA for the next release. Should I release 2.13.4 now, or should we wait for somebody to transform @NoiseByNorthwest 's summary above into a test?

HypeMC:
> People are starting to ask me an ETA for the next release. Should I release 2.13.4 now, or should we wait for somebody to transform @NoiseByNorthwest 's summary above into a test?

@greg0ire I can take a look at it today. I'll let you know how it goes.

Warxcell:
I have (eventually) a test-case for you https://github.com/doctrine/orm/pull/10245 :)

michnovka:
@greg0ire I think it has been confirmed now in https://github.com/doctrine/orm/pull/10245 that there is no more error. You can make the release, and ideally merge this PR into 2.13.4 too

tests/Doctrine/Tests/ORM/Hydration/SimpleObjectHydratorTest.php
```diff
@@ -155,4 +157,29 @@ public function testWrongValuesShouldNotBeConvertedToPhpValue(): void
...
+                'id_1' => 1,
+                'suit_2' => 'C',
+            ],
+        ];
+
+        $stmt     = ArrayResultFactory::createFromArray($resultSet);
+        $hydrator = new SimpleObjectHydrator($this->entityManager);
+        $result   = $hydrator->hydrateAll($stmt, $rsm)[0];
+
+        self::assertEquals(Suit::Clubs, $result->suit);
```
greg0ire: This assertion passes without your changes.

tests/Doctrine/Tests/ORM/Hydration/SimpleObjectHydratorTest.php
```diff
@@ -155,4 +157,29 @@ public function testWrongValuesShouldNotBeConvertedToPhpValue(): void
...
+        self::assertEquals(Suit::Clubs, $this->entityManager->getUnitOfWork()->getOriginalEntityData($result)['suit']);
```
greg0ire: This one doesn't, but it uses the API of UnitOfWork, which users are not really suppose to deal with. This makes it hard for me to understand where you're driving at with this test.

tests/Doctrine/Tests/ORM/Hydration/SimpleObjectHydratorTest.php
```diff
@@ -155,4 +157,29 @@ public function testWrongValuesShouldNotBeConvertedToPhpValue(): void
         $result   = $hydrator->hydrateAll($stmt, $rsm);
         self::assertEquals($result[0], $expectedEntity);
     }
+
+    /**
+     * @requires PHP 8.1
+     */
+    public function testEnumsAreBuilt(): void
```
greg0ire: I don't understand what this method name means.

tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterResultMappingTest.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * @requires PHP 8.1
+     */
+    public function testEnumTypeIsAddedToResultMapping(): void
+    {
+        $statement = $this->persister->getSelectSQL([]);
+        self::assertEquals('SELECT t0.id AS id_1, t0.suit AS suit_2 FROM Card t0', $statement);
+        self::assertEquals(['suit_2' => Suit::class], $this->persister->getResultSetMapping()->enumMappings);
```
greg0ire: Same here, who uses that API directly? I don't understand what you are fixing here…

greg0ire (commented):
The commit message consists in a title.

This is fine for trivial changes, but here, to understand this change, one has to have prior knowledge of whatever was discussed in https://github.com/doctrine/orm/pull/10074, which itself mentions https://github.com/doctrine/orm/issues/10071, which is unclear to me… https://github.com/doctrine/orm/pull/10073 is a bit more clear though.

Long story short, I don't want to have to read 3 issues and 4 PRs before deciding whether to merge this so if you want this merged, you'll have to:
- create a crystal clear commit message explaining what is broken, and why your changes fix the current situation
- write a test that shows the current bug using high level APIs, not only plumbing APIs

I will release without this for now.

michnovka:
@greg0ire this is really not a "fix" for any functional problem. I explained this here: https://github.com/doctrine/orm/pull/10088#issuecomment-1300856964

Its internal inconsistency. If we do not merge this, it will come bite us back when we least expect it, as sometimes data is stored in one way, sometimes in different. In the places where this might cause an issue, there is always a check in place which handles both cases. But its logical to store all `originalEntityData` identically for individual fields regardless of which hydrator is used.

I think we do NOT even have to include tests. As you pointed out in your review, we are testing UnitOfWork API which is not supposed to be used by users. I am fine with removing all the tests and just merging the actual fix


> This is fine for trivial changes, but here, to understand this change, one has to have prior knowledge of whatever was discussed in #10074, which itself mentions #10071, which is unclear to me… #10073 is a bit more clear though.

All these references are irrelevant, they were brought up by people who thought this PR affects them, but it does not. The issue is self-isolated in this PR and is best explained by the above linked comment on this PR - https://github.com/doctrine/orm/pull/10088#issuecomment-1300856964

greg0ire:
Ok so remove the tests, and target 2.14.x, since this is more likely to break stuff than to fix it.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
```diff
@@ -697,7 +697,7 @@ protected function registerManaged(ClassMetadata $class, $entity, array $data)
      *
      * @return BackedEnum|array<BackedEnum>
      */
-    private function buildEnum($value, string $enumType)
+    protected function buildEnum($value, string $enumType)
```
greg0ire: If we are going to expose this to other classes, let's at least make it final
```suggestion
    final protected function buildEnum($value, string $enumType)
```

greg0ire:
> is best explained by the above linked comment on this PR

Please reuse whatever you see fit as a commit message body. I will read that.

michnovka:
Ensure consistent `UnitOfWork::$originalEntityData` regardless of hydrator type for enum fields.

michnovka:
I would really get rid of the tests @greg0ire , you are right it makes no sense since only "plumbing" APIs are affected

greg0ire:
If there was a test to be added, it would be in `AbstractHydratorTest` => tests for things that should apply to all hydrators. It makes no sense not to have tests for high level API if we know what we are fixing, but here, we don't.

michnovka:
whats the decision here, do we remove the tests?

greg0ire:
I'd say either remove them or promote them to `AbstractHydratorTest` if doable. What the current tests are doing is only ensuring a behavior that happens to be the same as other hydrators now.

michnovka:
@HypeMC can you please do either of what @greg0ire suggested? Thanks

HypeMC:
I'll take a look to see what can be done.

lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php
HypeMC: Done

HypeMC:
@greg0ire I've removed the tests as I didn't see a way to promote them to the `AbstractHydratorTest`. The changes I made are in the `SimpleObjectHydrator` & the `AbstractHydratorTest` only mocks the `AbstractHydrator`.

I've also switched the target branch to 2.14.x & updated the commit message.

greg0ire (approved):

michnovka (approved):

SenseException:
@greg0ire @HypeMC I'm confused about the removal of the tests and why there are no test cases for the newly added code. If someone removes the newly added code at some time in the future there would be no test to catch that and reintroduce this inconsistency again.

greg0ire:
If we keep the test and somebody changes other hydrators, we would also reintroduce an inconsistency, that's why I recommended we either promote this test as an abstract test or remove it. It gives a false sense of security on top of being hard to understand.

michnovka:
This was not merged into 2.14.0. However I really dont think this is a new feature, as it is only changing internal states. Do we need to retarget 2.15.x or can we merge into 2.14.1?

greg0ire:
Seems fair, let's merge as is.

PR #10089, opened 2022-10-02
Document properties as possibly null

greg0ire:
`__sleep()` does not list these properties as suposed to be serialized.

This means we have to assert that it is not in the many scenarios where
we resort to these properties. Introducing a private method allows us to
centralize the assertions.


greg0ire:
Blocked by https://github.com/doctrine/orm/pull/10092

SenseException (approved):

Issue #9917, opened 2022-07-19
OPCache wasted memory increases on each request


icedevelopment:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.12.3

#### Summary

The OPCache wasted memory increases on each request if a paginator is used.

#### Current behavior

Whenever you make a request on a controller with a `Doctrine\ORM\Query` that uses [Doctrine\ORM\Tools\Pagination\Paginator#getIterator()](https://github.com/doctrine/orm/blob/c05e1709e9ffb9abe8d37260a78975cc816ee385/lib/Doctrine/ORM/Tools/Pagination/Paginator.php#L131), the DQL to SQL parser cache is redone each time.
Seems that this is bad a side effect introduced in commit https://github.com/doctrine/orm/pull/7865/commits/023e94661a1956215a6ddf17cb1b8908274b0d01 which was merged in [2.6.5](https://github.com/doctrine/orm/releases/tag/v2.6.5) so every version between this and [2.12.3](https://github.com/doctrine/orm/releases/tag/2.12.3) should be impacted.

#### How to reproduce

1. Install a Symfony app with Doctrine with the following standard cache configuration:

```yaml
# config/packages/prod/doctrine.yaml
doctrine:
    orm:
        auto_generate_proxy_classes: false
        metadata_cache_driver:
            type: pool
            pool: doctrine.system_cache_pool
        query_cache_driver:
            type: pool
            pool: doctrine.system_cache_pool
        result_cache_driver:
            type: pool
            pool: doctrine.result_cache_pool

framework:
    cache:
        pools:
            doctrine.result_cache_pool:
                adapter: cache.app
            doctrine.system_cache_pool:
                adapter: cache.system

```
2. Create a controller that uses a `Doctrine\ORM\Query` with a paginator (i.e. [knplabs/knp-paginator-bundle](https://packagist.org/packages/knplabs/knp-paginator-bundle)).

3. Set the environment to prod in .env (APP_ENV=prod and APP_DEBUG=0).

4. Make requests on the controller and watch `opcache_get_status()['memory_usage']['wasted_memory']` increase each time until OPCache is full and resets itself.

The expireQueryCache flag is always set to true here:

https://github.com/doctrine/orm/blob/c05e1709e9ffb9abe8d37260a78975cc816ee385/lib/Doctrine/ORM/Tools/Pagination/Paginator.php#L163

This makes Doctrine reparse the DQL each time:

[...]

PR #10095, opened 2022-10-04
Use NullCacheItemPool when using a Paginator.

icedevelopment:
This fixes a bug where Doctrine would recreate the DQL to SQL parsing
cache on every request which would fill up OPCache with duplicates.
We now use a "Null" cache implementation instead of calling
Query::expireQueryCache().
@see https://github.com/doctrine/orm/issues/9917

### DIFF
--- a/lib/Doctrine/ORM/Cache/NullCacheItemPool.php
+++ b/lib/Doctrine/ORM/Cache/NullCacheItemPool.php
@@ -1,0 +1,106 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Cache;
+
+use Closure;
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+use Traversable;
+
+/**
+ * @internal this class is used as a workaround for a cache issue.
+ *
+ * @see https://github.com/doctrine/orm/pull/10095
+ */
+final class NullCacheItemPool implements CacheItemPoolInterface
+{
+    /** @var Closure|null */
+    private static $createCacheItem;
+
+    public function __construct()
+    {
+        self::$createCacheItem ?? self::$createCacheItem = Closure::bind(
+            static function (string $key) {
+                return new NullCacheItem();
+            },
+            null,
+            NullCacheItem::class
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItem($key): CacheItemInterface
+    {
+        return (self::$createCacheItem)($key);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItems(array $keys = []): Traversable
+    {
+        return $this->generateItems($keys);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function hasItem($key): bool
+    {
+        return false;
+    }
+
+    public function clear(string $prefix = ''): bool
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function deleteItem($key): bool
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function deleteItems(array $keys): bool
+    {
+        return true;
+    }
+
+    public function save(CacheItemInterface $item): bool
+    {
+        return true;
+    }
+
+    public function saveDeferred(CacheItemInterface $item): bool
+    {
+        return true;
+    }
+
+    public function commit(): bool
+    {
+        return true;
+    }
+
+    /**
+     * @param array<string> $keys
+     *
+     * @return Traversable<NullCacheItem>
+     */
+    private function generateItems(array $keys): Traversable
+    {
+        $f = self::$createCacheItem;
+
+        foreach ($keys as $key) {
+            yield $key => $f($key);
+        }
+    }
+}
--- a/lib/Doctrine/ORM/Cache/NullCacheItem.php
+++ b/lib/Doctrine/ORM/Cache/NullCacheItem.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Cache;
+
+use Psr\Cache\CacheItemInterface;
+
+/**
+ * @internal this class is used as a workaround for a cache issue.
+ *
+ * @see https://github.com/doctrine/orm/pull/10095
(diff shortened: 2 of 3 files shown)


lib/Doctrine/ORM/Cache/NullCacheItemPool.php
```diff
@@ -0,0 +1,124 @@
...
+
+namespace Doctrine\ORM\Cache;
+
+use Closure;
+use Doctrine\Common\Cache\Psr6\CacheItem;
+use Generator;
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+
+class NullCacheItemPool implements CacheItemPoolInterface
```
greg0ire:
```suggestion
/**@internal this class is used as a workaround for a cache issue */
final class NullCacheItemPool implements CacheItemPoolInterface
```

lib/Doctrine/ORM/Cache/NullCacheItemPool.php
```diff
@@ -0,0 +1,124 @@
...
+            },
+            null,
+            CacheItem::class
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItem($key)
```
greg0ire: You can use as many native type declarations as possible in this class, please do so.

lib/Doctrine/ORM/Cache/NullCacheItemPool.php
```diff
@@ -0,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Cache;
+
+use Closure;
+use Doctrine\Common\Cache\Psr6\CacheItem;
```
greg0ire: This package is deprecated, you shouldn't need to use it, I think you will need to introduce your own dummy implementation.

lib/Doctrine/ORM/Cache/NullCacheItemPool.php
```diff
@@ -0,0 +1,103 @@
...
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+use Traversable;
+
+/** @internal this class is used as a workaround for a cache issue @see https://github.com/doctrine/orm/pull/10095 */
```
greg0ire: Since you added another annotation, this needs to become multiline.

lib/Doctrine/ORM/Cache/NullCacheItemPool.php
icedevelopment: I'm not sure what I should use instead as I still need an implementation of `Psr\Cache\CacheItemInterface` since getItem must always return one.

greg0ire: Maybe you can write a dummy implementation?

icedevelopment:
I think most tests are failing on php 8+ because of some return type hints that don't exist in php < 8. How do you want to handle that?

greg0ire:
> as many native type declarations as possible

"as possible" was a keyword here :wink: 

When not possible, use phpdoc instead please :)

greg0ire:
:thinking: it does not seem you can fix it unless you only fix it for PHP 8 only :sweat:

derrabus:
> 🤔 it does not seem you can fix it unless you only fix it for PHP 8 only 😓

All PSR Cache implementations should be flagged as `final` and `@internal`. We can ship two implementations and pick the right one depending on the PHP version.

That being said, I'd really think we should try to fix that issue without shipping our own PSR Cache implementation.

icedevelopment:
That's what I tried to do but PhpStan PHPStan is still yelling at me for example `Error: Return type mixed of method Doctrine\ORM\Cache\NullCacheItem::get() is not covariant with return type mixed of method Psr\Cache\CacheItemInterface::get().`

> 😓

Yeah that's what I thought.

icedevelopment:
> 
> All PSR Cache implementations should be flagged as `final` and `@internal`. We can ship two implementations and pick the right one depending on the PHP version.
> 
> That being said, I'd really think we should try to fix that issue without shipping our own PSR Cache implementation.



> 🤔 it does not seem you can fix it unless you only fix it for PHP 8 only 😓

Yeah that's what I thought.

icedevelopment:
> > 🤔 it does not seem you can fix it unless you only fix it for PHP 8 only 😓
> 
> All PSR Cache implementations should be flagged as `final` and `@internal`. We can ship two implementations and pick the right one depending on the PHP version.
> 
> That being said, I'd really think we should try to fix that issue without shipping our own PSR Cache implementation.

Do you think we should require symfony/cache so that we could use `Symfony\Component\Cache\Adapter\NullAdapter` instead?

derrabus:
> That's what I tried to do but PhpStan PHPStan is still yelling at me

Yes well, we know the cause and we're smarter than PHPStan in that case. There's no shame in ignoring false positives.

> Do you think we should require symfony/cache so that we could use `Symfony\Component\Cache\Adapter\NullAdapter` instead?

That's not an option.

icedevelopment:
> > That's what I tried to do but PhpStan PHPStan is still yelling at me
> 
> Yes well, we know the cause and we're smarter than PHPStan in that case. There's no shame in ignoring false positives.
> 
> > Do you think we should require symfony/cache so that we could use `Symfony\Component\Cache\Adapter\NullAdapter` instead?
> 
> That's not an option.

Do you have a solution for fixing this without making our own PSR cache implementation?

derrabus:
I don't and I haven't investigated the issue tbh. But if you're rolling your own cache implementation to fix a paginator bug, I'm confident that you're using a sledgehammer to crack a nut. There has to be a better way.

icedevelopment:
> I don't and I haven't investigated the issue tbh. But if you're rolling your own cache implementation to fix a paginator bug, I'm confident that you're using a sledgehammer to crack a nut. There has to be a better way.

Well I agree in a sense. The original call to `expireQueryCache()`  (see https://github.com/doctrine/orm/issues/7837) already looked like a hack and is what created this nasty side effect in OPCache. I'm afraid I don't know the internals of Doctrine well enough to be able to find another solution but I'd really appreciate if you had a little bit of time to investigate.

mpdude:
Could you please check if #8797 solves your issue?

mpdude:
A suggested fix is in #10434.

PR #10097, opened 2022-10-04
fix: readonly property's Entity deletion

### LINKED ISSUE #10032: Exception when deleting entity with readonly ID property in model (PHP 8.1 feature)
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.13.1
| PHP |  8.1.6 
| Database | MariaDB 10.4.24

#### Summary

When I try to remove an entity with a model which as a readonly ID property through the entity manager, an exception is thrown. I am using PHP 8.1 which added support for readonly properties, and this is supported since ORM 2.11 according to the [release notes](https://www.doctrine-project.org/2022/01/11/orm-2.11.html).

I am using Slim 4 as a framework.

#### Current behavior

The following exception is thrown when trying to call the delete method on my entity repository class:

Type: LogicException
Code: 0
Message: Attempting to change readonly property App\Models\Text::$id.
C:\...\vendor\doctrine\orm\lib\Doctrine\ORM\Mapping\ReflectionReadonlyProperty.php

Trace:

```
#0 C:\...\vendor\doctrine\orm\lib\Doctrine\ORM\UnitOfWork.php(1260): Doctrine\ORM\Mapping\ReflectionReadonlyProperty->setValue(Object(App\Models\Text), NULL)
#1 C:\...\vendor\doctrine\orm\lib\Doctrine\ORM\UnitOfWork.php(453): Doctrine\ORM\UnitOfWork->executeDeletions(Object(Doctrine\ORM\Mapping\ClassMetadata))
#2 C:\...\vendor\doctrine\orm\lib\Doctrine\ORM\EntityManager.php(403): Doctrine\ORM\UnitOfWork->commit(NULL)
#3 C:\...\src\Repository\TextRepository.php(67): Doctrine\ORM\EntityManager->flush()
#4 C:\...\src\Controllers\Backend\TextController.php(33): App\Repository\TextRepository->delete(15)
#5 [internal function]: App\Controllers\Backend\TextController->index(Object(Slim\Psr7\Request), Object(Slim\Psr7\Response), Object(App\Repository\TextRepository))
#6 C:\...\vendor\php-di\invoker\src\Invoker.php(74): call_user_func_array(Array, Array)
#7 C:\...\vendor\php-di\slim-bridge\src\ControllerInvoker.php(47): Invoker\Invoker->call(Array, Array)
#8 C:\...\vendor\slim\slim\Slim\Routing\Route.php(384): DI\Bridge\Slim\ControllerInvoker->__invoke(Array, Object(Slim\Psr7\Request), Object(Slim\Psr7\Response), Array)
#9 C:\...\vendor\slim\slim\Slim\MiddlewareDispatcher.php(81): Slim\Routing\Route->handle(Object(Slim\Psr7\Request))
#10 C:\...\vendor\tuupola\slim-basic-auth\src\HttpBasicAuthentication.php(187): Slim\MiddlewareDispatcher->handle(Object(Slim\Psr7\Request))
#11 C:\...\vendor\slim\slim\Slim\MiddlewareDispatcher.php(147): Tuupola\Middleware\HttpBasicAuthentication->process(Object(Slim\Psr7\Request), Object(Slim\MiddlewareDispatcher))
#12 C:\...\vendor\slim\slim\Slim\MiddlewareDispatcher.php(81): Psr\Http\Server\RequestHandlerInterface@anonymous->handle(Object(Slim\Psr7\Request))
#13 C:\...\vendor\slim\slim\Slim\Routing\Route.php(341): Slim\MiddlewareDispatcher->handle(Object(Slim\Psr7\Request))
#14 C:\...\vendor\slim\slim\Slim\Routing\RouteRunner.php(84): Slim\Routing\Route->run(Object(Slim\Psr7\Request))
#15 C:\...\vendor\slim\slim\Slim\Middleware\ErrorMiddleware.php(107): Slim\Routing\RouteRunner->handle(Object(Slim\Psr7\Request))
#16 C:\...\vendor\slim\slim\Slim\MiddlewareDispatcher.php(147): Slim\Middleware\ErrorMiddleware->process(Object(Slim\Psr7\Request), Object(Slim\Routing\RouteRunner))
#17 C:\...\vendor\slim\twig-view\src\TwigMiddleware.php(115): Psr\Http\Server\RequestHandlerInterface@anonymous->handle(Object(Slim\Psr7\Request))
#18 C:\...\vendor\slim\slim\Slim\MiddlewareDispatcher.php(147): Slim\Views\TwigMiddleware->process(Object(Slim\Psr7\Request), Object(Psr\Http\Server\RequestHandlerInterface@anonymous))
#19 C:\...\vendor\bryanjhv\slim-session\src\Slim\Middleware\Session.php(82): Psr\Http\Server\RequestHandlerInterface@anonymous->handle(Object(Slim\Psr7\Request))
#20 C:\...\vendor\slim\slim\Slim\MiddlewareDispatcher.php(209): Slim\Middleware\Session->__invoke(Object(Slim\Psr7\Request), Object(Psr\Http\Server\RequestHandlerInterface@anonymous))
#21 C:\...\vendor\slim\slim\Slim\MiddlewareDispatcher.php(81): Psr\Http\Server\RequestHandlerInterface@anonymous->handle(Object(Slim\Psr7\Request))
#22 C:\...\vendor\slim\slim\Slim\App.php(215): Slim\MiddlewareDispatcher->handle(Object(Slim\Psr7\Request))
#23 C:\...\vendor\slim\slim\Slim\App.php(199): Slim\App->handle(Object(Slim\Psr7\Request))
#24 C:\...\src\bootstrap.php(92): Slim\App->run()
#25 C:\...\public\index.php(3): require('C:\\devel\\php\\et...')
#26 {main}
```

#### How to reproduce

My model:

```
<?php

namespace App\Models;

use Doctrine\ORM\Mapping\Column;
use Doctrine\ORM\Mapping\Entity;
use Doctrine\ORM\Mapping\Id;
use Doctrine\ORM\Mapping\Table;

#[Entity, Table(name: 'texts')]
class Text
{
    #[Column(name: 'text_id'), Id, GeneratedValue]
    public readonly int $id;

    #[Column(name: 'text_keyword', unique: true)]
    public string $keyword;

    #[Column(name: 'text_text')]
    public string $content;
}
```

My Repository:

```
<?php

namespace App\Repository;

use App\Models\Text;
use Doctrine\ORM\EntityManager;
use Doctrine\ORM\EntityRepository;

final class TextRepository
{
    private EntityRepository $repo;

    public function __construct(private EntityManager $em)
    {
        $this->repo = $em->getRepository(Text::class);
    }

    // ... some more methods

    public function delete(int $id): void
    {
        $text = $this->em->getReference(Text::class, $id);
        $this->em->remove($text);
        $this->em->flush();
    }
}
```

SQL used to create the relevant database table:

```
CREATE TABLE `texts` (
  `text_id` int(11) NOT NULL,
  `text_keyword` varchar(255) NOT NULL,
  `text_text` text NOT NULL
) ENGINE=MyISAM DEFAULT CHARSET=utf8;

ALTER TABLE `texts`
  ADD PRIMARY KEY (`text_id`),
  ADD UNIQUE KEY `text_keyword` (`text_keyword`);
```

#### Expected behavior

No exception should be thrown.

Nayte91:
fixes #10032

- What ultimately happens if an entity has its identifier [not nulled](https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/UnitOfWork.php#L1252)? What happens in a world where every Entities in a project have readonly identifiers and therefore aren't nulled? What about non nullable but writable identifiers?
- On a more general design, should we treat an Entity with a readonly identifier [as we do for a readonly Entity](https://github.com/Nayte91/orm/blob/cc6c38b8b0f084a6fff49396ef96ec1c03b57c5e/lib/Doctrine/ORM/UnitOfWork.php#L853)?
- What is the minimal PHP version for orm? 7.4? How will behave the [isReadOnly()](https://github.com/Nayte91/orm/blob/cc6c38b8b0f084a6fff49396ef96ec1c03b57c5e/lib/Doctrine/ORM/UnitOfWork.php#L3423) method on PHP < 8.1?

### DIFF
--- a/tests/Doctrine/Tests/Models/GH10097/EntityWithReadonlyIdentifier.php
+++ b/tests/Doctrine/Tests/Models/GH10097/EntityWithReadonlyIdentifier.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\GH10097;
+
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\Mapping\Entity;
+
+/** @Entity */
+class EntityWithReadonlyIdentifier
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column
+     * @ORM\GeneratedValue
+     */
+    public readonly int $id;
+}
--- a/tests/Doctrine/Tests/ORM/EntityManagerTest.php
+++ b/tests/Doctrine/Tests/ORM/EntityManagerTest.php
@@ -23,6 +23,7 @@ use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Persistence\Mapping\MappingException;
 use Doctrine\Tests\Models\CMS\CmsUser;
 use Doctrine\Tests\Models\GeoNames\Country;
+use Doctrine\Tests\Models\GH10097\EntityWithReadonlyIdentifier;
 use Doctrine\Tests\OrmTestCase;
 use Generator;
 use InvalidArgumentException;
@@ -365,5 +366,26 @@ class EntityManagerTest extends OrmTestCase
         $this->expectDeprecationWithIdentifier('https://github.com/doctrine/orm/issues/8459');
 
         $this->entityManager->flush($entity);
+    }
+
+    /**
+     * @requires PHP 8.1
+     *
+     * Note: before the fix, removing and Entity with a readonly identifier
+     * threw a LogicException from Doctrine\ORM\Mapping\ReflectionReadonlyProperty::setValue().
+     */
+    public function testDeleteEntityWithReadonlyIdentifierWorks(): void
+    {
+        $entity = new EntityWithReadonlyIdentifier();
+
+        $this->entityManager->persist($entity);
+        $this->entityManager->flush();
+
+        self::assertTrue($this->entityManager->contains($entity), 'The entity is not stored');
+
+        $this->entityManager->remove($entity);
+        $this->entityManager->flush();
+
+        self::assertFalse($this->entityManager->contains($entity), 'The entity is still stored');
     }
 }


derrabus:
Please add a test that reproduces the issue you're attempting to fix.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1252,7 +1252,7 @@ private function executeDeletions(ClassMetadata $class): void
             // Entity with this $oid after deletion treated as NEW, even if the $oid
             // is obtained by a new entity because the old one went out of scope.
             //$this->entityStates[$oid] = self::STATE_NEW;
-            if (! $class->isIdentifierNatural()) {
+            if (! $class->isIdentifierNatural() && !$class->reflFields[$class->identifier[0]]->isReadOnly()) {
```
beberlei: Would it make more sense maybe to modify the readonly relfection object instead and remove that following block and the exception entirely? https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/Mapping/ReflectionReadonlyProperty.php#L45

That way it stays within the reflection abstraction and does not pollute this code here with an implementation detail.

gnanette: That's a way to explore, I stopped it because of 2 questions: 
- Is it OK to ask setValue(), return nothing and don't set the value ? Should we at least return a boolean to tell "false" when the value is not setted?
- What happens in a doctrine world where every Entities in a project have readonly identifiers and therefore aren't nulled when we ask with setValue()? Why do we set ids to null in the first place?

Nayte91: That's a way to explore, I [went away of it](https://github.com/doctrine/orm/issues/10032#issuecomment-1266959818) because of 3 questions:

- Is it OK to ask setValue(), return nothing and don't set the value ? Should we at least return a boolean to tell "false" when the value is not setted? --> It needs to change contract within parent ReflectionProperty, and potentially breaks a lot of things. 👎 
- What happens in a doctrine world where every project's Entities have readonly identifiers and therefore aren't nulled when asked with setValue()? Why do we set ids to null in the first place?

Because UnitOfWork only aim to modify ID, but the setValue() is meant to change any property, the exception seems relevant. So I feel the question is more: how do you plan to manage the 8.2 readonly classes? Then we should handle a writable Entity with a readonly identifier as a readonly Entity. I feel like it's the cleaner path?

Nayte91:
removed unused "uses" in ReflectionReadonlyProperty

Nayte91:
I moved the test from UnitOfWorkTest to EntityManagerTest, as the bug can be only reproduced from a EntityManager functional test.
Let me know if you put functional tests elsewhere

michnovka:
While I think the feature is needed as readonly IDs make totally sense to me and we need a way to delete such entities, I do not like your implementation one-bit. It seems that you just removed a safety check without even checking for the type of the field. So with your fix, not only readonly IDs can be set to nothing/null, ANY field can. And that is not ok

Nayte91:
Hello @michnovka, thank you for your feedback!

> It seems that you just removed a safety check without even checking for the type of the field.

I feel your concern, but we are in the ReflectionReadonlyProperty class, so whatever you can check or ask to set as a value won't happens anyway.
So I feel more concerned that a method 'setValue()' is available for a readonly property.. Except if this one is not initialized, and that's checked [here](https://github.com/doctrine/orm/pull/10097/files#diff-86642c8ac964b931a087a7650e11d185036d522bd5698876ecb340d977b5b2feL37).

> So with your fix, not only readonly IDs can be set to nothing/null, ANY field can. And that is not ok
Hmm I didn't changed anything beyond the ReadOnlyProperty::setValue() method, where I just ensured that anything is done, so I don't get where I allow some behavior on writable properties?

What would you see as a fix or strategy here, to help me? I followed @beberlei advice [here](https://github.com/doctrine/orm/pull/10097#discussion_r988400068)and sticked on the setValue() implementation as a strategy. How do you think it should be done?

michnovka:
@Nayte91 one main issue I see with your change is that after the `flush()` call the `$entity` still has ID set to the original value.

What we need to do IMO is treat differently readonly ID columns from the rest of the readonly properties. Also, we should ONLY allow change of the readonly ID (set to uninitialized) for deletion, it would be incorrect to allow changing of readonly ID property in any other context. If your code requires that, do not mark it as readonly.

i think this is very similar debate as to how to treat non-nullable typed IDs during deletion. As far as I know it was decided to set them to uninitialized state.

@beberlei what do you think, I think we cannot just work inside ReflectionReadonlyProperty class but the logic deciding whether to allow "de-initializing" readonly properties has to be made byt he caller, i.e. the `UnitOfWork::executeDeletions()`

michnovka:
Actually this turns out to be impossible to implement due to PHP limitations:

From https://wiki.php.net/rfc/readonly_properties_v2#reflection :
> ReflectionProperty::setValue() can bypass the requirement that initialization occurs from the scope where the property has been declared. However, reflection cannot modify a readonly property that has already been initialized. 

So there is no way even with reflection to modify (unset) the ID of the entity. This is bad, as if the entity is deleted and then persisted again, the ID is no longer auto generated.

Nayte91:
@michnovka thank you for sharing your thoughts, it seems that you know well how it works (or at least I didn't figured out by reading the code by myself 😶‍), so I guess you're able to answer the few questions I had at beginning, that it seems to be relevant with your informations:

> @Nayte91 one main issue I see with your change is that after the flush() call the $entity still has ID set to the original value.

- Why the doctrine's engine needs to set the ID to null/unset/0/whatever instead of keeping it or unsetting the whole entity? What ultimately happens if an entity has its identifier [not nulled](https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/UnitOfWork.php#L1252)? What happens in a world where every Entities in a project have readonly identifiers and therefore aren't nulled? What about non nullable but writable identifiers?

> So there is no way even with reflection to modify (unset) the ID of the entity. This is bad, as if the entity is deleted and then persisted again, the ID is no longer auto generated.

- I also followed the internals discussion about reflection limitations, but I was wondering how would you/should we handle 8.2's readonly entities? Do you agree that, on a more general design, we should manage an Entity with a readonly ID [as we do for a readonly Entity](https://github.com/Nayte91/orm/blob/cc6c38b8b0f084a6fff49396ef96ec1c03b57c5e/lib/Doctrine/ORM/UnitOfWork.php#L853)?

I'm still interested in fixing this, but I lack some knowledge and you seems to have some :)

michnovka:
@Nayte91 I am not a core developer here, so take my comments as an opinion why and how things work.

When new entity is created, before being persisted it has no ID. During persisting, the ID is set. When deleting the entity, the ID is set to null. Doctrine works with reflection when modifying the object values and when the field is typed and it is defined as non-nullable, and attempted to be set as null its is simply set to "uninitialized".

If you set ID before persisting, I believe then it is not auto-generated.

But regardless of this, the main issue is that you are trying to force some language construct where is doesnt belong. If entity is read-only, it by definition means it should be ONLY READ. It should not be modified, and by extension, not even deleted. This makes sense on Entity level, as you might want to use this to protect entities from being modified/deleted. But applying read-only on individual fields creates logic clashes with current implementation. The ID set to null/uninitialized being one possibility, another is e.g. having `readonly DateTimeInterface $createdAt;` which is set in the constructor.

So I dont think there is actually any good approach how to treat readonly entity properties which have to be modified at some point by doctrine lifecycle, being it  auto-generated IDs, IDs, or other fields meant to be set once during persistance of the entity.

Nayte91:
This is not "missing test" anymore 😉

derrabus:
No, but the CI is deep red.

Nayte91:
How can I ensure that tests for a specific PHP8.1 aren't breaking PHP7.2? Is there a comment, or something to skip by a way or another ?

michnovka:
@Nayte91 move the Entity data into separate file (`tests/Doctrine/Tests/Models/GH10097` e.g.) -> this will get rid of the PHP7 issue with readonly keyword.

Then for the test function add 

```php
/**
 * @requires PHP 8.1
 */
public function testDeleteEntityWithReadonlyIdentifierWorks(): void
```

also move the comment into the `/**` block (now its using `/*` only)

Nayte91:
@michnovka changes applied with your directives, sir

michnovka:
@Nayte91 I can see that `tests/Doctrine/Tests/Models/GH10097/EntityWithReadonlyIdentifier.php` has missing end line, add it. You should check your files with phpcs, just run `vendor/bin/phpcs` and see what it reports.

Nayte91:
@michnovka OK for PHPCS.
It gives me an error in every file, the same: 
 1 | ERROR | [x] End of line character is invalid; expected "\n" but found "\r\n" (Generic.Files.LineEndings.InvalidEOLChar)

beside this, I amended sir

Nayte91:
What can I do more to help passing the other tests?

michnovka:
@Nayte91 well read the messages of the tests why they failed, you still have even PHPCS issues. I suggest you read thorougly https://github.com/doctrine/orm/blob/2.13.x/CONTRIBUTING.md

Nayte91:
I think this PR is becoming anachronistic, as in 8.2 we have to deal with entirely readonly classes; Therefore I guess this specific use case will be handled by a larger feature that deals with readonly classes.

So what I suggest is to remove the readonly support affirmation from the [ORM 2.11 blogpost](https://www.doctrine-project.org/2022/01/11/orm-2.11.html), as it clearly missdirects users.
 
WDYT?

PR #10098, opened 2022-10-05
Deprecate annotation driver

greg0ire:
No new project should use that driver.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -417,7 +417,7 @@ function foo(EntityManagerInterface $entityManager, callable $func) {
     if (method_exists($entityManager, 'wrapInTransaction')) {
         return $entityManager->wrapInTransaction($func);
     }
-    
+
     return $entityManager->transactional($func);
 }
 ```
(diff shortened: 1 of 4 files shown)


docs/en/reference/annotations-reference.rst
```diff
@@ -1,6 +1,11 @@
 Annotations Reference
 =====================
 
+.. warning::
+    The annotation driver is deprecated and will be removed in version
+   3.0. It is strongly recommended to switch to one of the other
+   mappings.
```
stof: should be `one of the other mapping drivers` (or maybe `mapping formats`).

derrabus:
Please add an entry to UPGRADE.md.

docs/en/reference/annotations-reference.rst
```diff
@@ -1,6 +1,11 @@
 Annotations Reference
 =====================
 
+.. warning::
+    The annotation driver is deprecated and will be removed in version
+   3.0. It is strongly recommended to switch to one of the other
+   mapping drivers.
+
```
derrabus:
```suggestion
.. warning::
    The annotation driver is deprecated and will be removed in version
    3.0. It is strongly recommended to switch to one of the other
    mapping drivers.

```

UPGRADE.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade to 2.14
 
+## Deprecated annotation mapping driver.
+
+You should switch to one of the other mapping drivers.
```
derrabus: This deprecation will affect a lot of projects out there. Let's try to be as helpful as possible.

```suggestion
Please switch to one of the other mapping drivers. Native attributes which PHP
supports since version 8.0 are probably your best option.
```

derrabus (commented):
We should also work on documentation on how to migrate existing large codebases to attributes, maybe as a blog article.

derrabus (approved):

PR #10102, opened 2022-10-06
Address method rename

greg0ire:
See https://github.com/doctrine/dbal/pull/5724


derrabus:
LGTM, but we could could do this in 2.14.x already.

derrabus (approved):

greg0ire:
@derrabus I thought of this, and it cannot be tested properly unless we use an unstable version of DBAL on 2.14. My current plan is to do a single backporting PR with all the other relevant changes when the next minor of the DBAL is released, but I'm unsure about this, let me know how you think we should handle this.

derrabus:
But we DO test 2.14. with DBAL `3@dev`, don't we?

greg0ire:
Oh right, I missed that. I'll do that backporting PR soon then 👍

PR #10104, opened 2022-10-06
Add isMemberOf and isInstanceOf to Expr helper list

ker0x:
These 2 methods were missing from the list of helper methods!

### DIFF
--- a/docs/en/reference/query-builder.rst
+++ b/docs/en/reference/query-builder.rst
@@ -435,10 +435,10 @@ complete list of supported helper methods available:
         public function isNotNull($x); // Returns string
 
         // Example - $qb->expr()->isMemberOf('?1', 'u.groups') => ?1 MEMBER OF u.groups
-        public function isMemberOf($x, $y); // Return Expr\Comparison instance
+        public function isMemberOf($x, $y); // Returns Expr\Comparison instance
 
         // Example - $qb->expr()->isInstanceOf('u', Employee::class) => u INSTANCE OF Employee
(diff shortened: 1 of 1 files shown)


docs/en/reference/query-builder.rst
```diff
@@ -434,6 +434,12 @@ complete list of supported helper methods available:
         // Example - $qb->expr()->isNotNull('u.id') => u.id IS NOT NULL
         public function isNotNull($x); // Returns string
 
+        // Example - $qb->expr()->isMemberOf('?1', 'u.groups') => ?1 MEMBER OF u.groups
+        public function isMemberOf($x, $y); // Return Expr\Comparison instance
```
derrabus:
```suggestion
        public function isMemberOf($x, $y); // Returns Expr\Comparison instance
```

docs/en/reference/query-builder.rst
```diff
@@ -434,6 +434,12 @@ complete list of supported helper methods available:
...
+
+        // Example - $qb->expr()->isInstanceOf('u', Employee::class) => u INSTANCE OF Employee
+        public function isInstanceOf($x, $y); // Return Expr\Comparison instance
```
derrabus:
```suggestion
        public function isInstanceOf($x, $y); // Returns Expr\Comparison instance
```

derrabus (changes requested):
Thank you, let's make the comments consistent with the ones above your change.

derrabus (approved):

PR #10107, opened 2022-10-06
Update getting-started.rst

pipanni:
Some code was missing from file Product.php and the following script execution calling the file create_product.php would fail.

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -531,7 +531,20 @@ but you only need to choose one.
              */
             private $name;
 
-            // .. (other code)
+            public function getId()
+            {
+                return $this->id;
+            }
+
+            public function getName()
+            {
+                return $this->name;
+            }
+
+            public function setName($name)
+            {
+                $this->name = $name;
+            }
         }
 
     .. code-block:: xml


derrabus:
The code example shows how metadata mapping with annotations looks like. The accessors have been left out here to keep the example brief. I wouldn't change this part.

pipanni:
Ok, removing this code does make the example brief, but it also produces an error when the instruction to execute the code "php create-product.php" arrives soon after. Any newbie to Doctrine (like me) who's following this tutorial will be stuck.
Cheers,
Pipanni
    On Thursday, October 6, 2022 at 04:43:11 PM GMT+1, Alexander M. Turek ***@***.***> wrote:  
 
 


The code example shows how metadata mapping with annotations looks like. The accessors have been left out here to keep the example brief. I wouldn't change this part.

SenseException:
It depends what "getting-started" was supposed to be: A documentation or a how-to with step-by-step code examples. The section also offers this tutorial code in a [Git repository](https://github.com/doctrine/doctrine2-orm-tutorial), but it doesn't look like it can be changed anymore if we need to now that it's an archive.

If those getters and setters are the only changes to make the "getting-started" code runnable I'd suggest to keep the changes of this PR and to remove the `The code of this tutorial is available on Github.` line. Reading the first few paragraphs does give the impression that this is a how-to.

derrabus:
> Ok, removing this code does make the example brief, but it also produces an error when the instruction to execute the code "php create-product.php" arrives soon after.

Yes, and if you decide to use XML to configure the mapping, the script will fail because the whole Product class is missing. I understand that listing the entire class somewhere might be helpful, but this is the wrong place.

derrabus:
Let's close here because we could not come to an agreement about the proposed change. Thank you for your proposal and if you want to pick up this topic again, feel free to open a new PR.

Issue #10063, opened 2022-09-24
2.13.2 breaks *array* of enums


ThomasLandauer:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.13.2

#### Summary

When having an *array* of enums like this:

```php
#[ORM\Column(type: Types::SIMPLE_ARRAY, length: 255, nullable: true, enumType: Color::class)]
private array $colors = [];
```
I'm getting this error:

> "Green,Red,Blue" is not a valid backing value for enum "App\Enum\Color"

I'm guessing that https://github.com/doctrine/orm/pull/10041 broke it, since @michnovka was referring to a *string* column:

```php
#[Column(type: Types::STRING, nullable: false, enumType: OrderStatus::class, options: ['charset' => 'utf8', 'collation' => 'utf8_unicode_ci'])]
private OrderStatus $orderStatus;
```
#### Workaround

For anybody else observing this, a workaround for now is to add this to your `composer.json` to continue using 2.13.1:
```json
"conflict": {
    "doctrine/orm": "2.13.2"
},
```

#### Next Step

Well, I think it's about time to properly document arrays of enums at https://www.doctrine-project.org/projects/doctrine-orm/en/2.13/reference/basic-mapping.html - I'll come up with a PR after getting some feedback here.

greg0ire:
Can you provide a test case for this? Also, please note that the code has already changed since 2.13.2: #10058

michnovka:
@ThomasLandauer I believe that https://github.com/doctrine/orm/pull/10058 fixes this issue, but please do provide a test case, we should push it for CI sake

ThomasLandauer:
There you go - my first attempt for a test here ;-) https://github.com/doctrine/orm/pull/10111
Please take a look, I probably need some help there.

Anyway, the issue itself does indeed look fixed in v2.13.3, so I'm closing here.

PR #10111, opened 2022-10-07
Test for array of enums

ThomasLandauer:
As requested in https://github.com/doctrine/orm/issues/10063

I'm not sure if this is really testing anything, since (I think) it passes in 2.13.2 too. Could you please double-check?

Are PHP attributes (mapping information) not available in tests?

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10063Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10063Test.php
@@ -1,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH10063Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->createSchemaForModels(GH10063Entity::class);
+    }
+
+    public function testArrayOfEnums(): void
+    {
+        $entity = (new GH10063Entity())->setColors([GH10063Enum::Red, GH10063Enum::Green]);
+
+        $this->_em->persist($entity);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $entity = $this->_em->find(GH10063Entity::class, $entity->id);
+        assert($entity instanceof GH10063Entity);
+        self::assertEquals([GH10063Enum::Red, GH10063Enum::Green], $entity->getColors());
+    }
+}
+
+/** @Entity */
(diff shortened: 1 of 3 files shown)


derrabus:
You need to make sure the test is not executed on PHP 8.0 and below. Especially the enum must not be parsed on older PHP versions.

ThomasLandauer:
Sorry, where can I configure this?

greg0ire:
You can use `@requires`, like so: https://github.com/doctrine/orm/blob/b2a4fac40b74276dfc13fea0734d45b8fec89e98/tests/Doctrine/Tests/ORM/ORMSetupTest.php#L53-L55

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10063Test.php
```diff
@@ -0,0 +1,82 @@
...
+
+        $entity = $this->_em->find(GH10063Entity::class, $entity->id);
+        assert($entity instanceof GH10063Entity);
+        self::assertEquals([GH10063Enum::Red, GH10063Enum::Green], $entity->getColors());
+    }
+}
+
+/**
+ * @Entity
+ * @requires PHP 8.1
```
greg0ire: This has no effect

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10063Test.php
```diff
@@ -0,0 +1,82 @@
...
+ */
+class GH10063Entity
+{
+    /**
+     * @Column
+     * @Id
+     * @GeneratedValue
+     */
+    public int $id;
```
greg0ire: Code using recent syntax should be moved to a separate file, otherwise PHP 7.1 will crash when parsing the file prior to realizing it should not execute the test.

ThomasLandauer: Well, the entire test is about enums (i.e. recent syntax). So how can I tell PHP 7 to ignore it?

greg0ire: 1. Make sure `GH10063Test` can be _parsed_ (parsing and executing are 2 different things) by PHP 7
2. Move `GH10063Entity` and `GH10063Enum` to a separate file.

greg0ire:
Looks better, doesn't it? :slightly_smiling_face: 

Do you need help with the remaining issues?

derrabus:
The test looks okay, but does it really cover the issue that was introduced with 2.13.2? I would've expected the use of a query builder to reproduce the problem.

michnovka:
I think these tests are redundant now, they are part of https://github.com/doctrine/orm/issues/10058 . see:
https://github.com/HypeMC/orm/blob/d69a0fa2cfd3b883200a6324829945509f4ecc07/tests/Doctrine/Tests/ORM/Functional/EnumTest.php#L100

ThomasLandauer:
Sorry guys, I wasn't sure from the beginning what I'm supposed to do in this test; and I know even less now ;-) So please somebody take over, or just close...

derrabus:
Okay, let's close this PR then.

PR #10113, opened 2022-10-08
Migrate to attributes

greg0ire:


tests/Doctrine/Tests/Models/CMS/CmsAddress.php
```diff
@@ -21,74 +21,25 @@
 use Doctrine\ORM\Mapping\SqlResultSetMappings;
 use Doctrine\ORM\Mapping\Table;
 
-/**
- * CmsAddress
- *
- * @Entity
- * @Table(name="cms_addresses")
- * @SqlResultSetMappings({
```
greg0ire: Todo: figure out what's up with these annotations: why are they removed, why are other similar annotations in other files not removed, and why removing them does not cause any test to fail… maybe the corresponding feature was removed?

beberlei: Yes it was removed

greg0ire: Ah thanks! I'll remove these annotations in another PR then.

greg0ire: :thinking: I think it was removed in https://github.com/doctrine/orm/pull/7095, but that PR was not backported, and the feature is not deprecated in 2.x

greg0ire: That's weird because it is featured in https://github.com/doctrine/orm/projects/4#card-63894409 , and marked as done. I'll finish the job.

greg0ire: See https://github.com/doctrine/orm/pull/10114

docs/en/reference/advanced-configuration.rst
```diff
@@ -28,7 +29,7 @@ steps of configuration.
 
     $config = new Configuration;
     $config->setMetadataCache($metadataCache);
-    $driverImpl = ORMSetup::createDefaultAnnotationDriver('/path/to/lib/MyProject/Entities');
+    $driverImpl = new AttributeDriver('/path/to/lib/MyProject/Entities');
```
derrabus: We should recommend using the attribute driver on 2.x already, maybe with a little note that it requires PHP 8.

tests/Doctrine/Tests/Models/Cache/Attraction.php
```diff
@@ -18,43 +18,27 @@
...
-        /** @Column(unique=true) */
-        protected string $name,
-        /**
-         * @Cache
-         * @ManyToOne(targetEntity="City", inversedBy="attractions")
-         * @JoinColumn(name="city_id", referencedColumnName="id")
-         */
-        protected City $city,
+        #[Column(unique: true)] protected string $name,
+        #[Cache] #[ManyToOne(targetEntity: 'City', inversedBy: 'attractions')] #[JoinColumn(name: 'city_id', referencedColumnName: 'id')] protected City $city,
```
derrabus: My attempt at making this more readable. Are there already PHPCS rules for that?

```suggestion
        #[Column(unique: true)]
        protected string $name,

        #[Cache]
        #[ManyToOne(targetEntity: 'City', inversedBy: 'attractions')]
        #[JoinColumn(name: 'city_id', referencedColumnName: 'id')]
        protected City $city,
```

docs/en/reference/advanced-configuration.rst
greg0ire: Yes, while looking into this I noticed there are a lot more things to do on 2.x regarding the attribute and yaml driver deprecations.

tests/Doctrine/Tests/Models/Cache/Attraction.php
greg0ire: Ugh yeah indeed the output of Rector is not very easy on the eye here :sweat: I'll try to see what phpcs rules currently exist if any.

michnovka: I did some digging and didnt find anything. Opened a feature request here - https://github.com/slevomat/coding-standard/issues/1442 . Will see if we can find some resources in our company to work on this if it turns out that this is not available in some other standard already. No promises tho ;]

greg0ire: Awesome! Thanks!

michnovka: @greg0ire @derrabus https://github.com/slevomat/coding-standard/releases/tag/8.6.0 have a look

```xml
    <rule ref="SlevomatCodingStandard.Attributes.DisallowAttributesJoining"/>
    <rule ref="SlevomatCodingStandard.Attributes.DisallowMultipleAttributesPerLine"/>
    <rule ref="SlevomatCodingStandard.Attributes.AttributeAndTargetSpacing"/>
```

greg0ire: Next step should be to include those in https://github.com/doctrine/coding-standard/blob/10.0.x/lib/Doctrine/ruleset.xml I guess 👍

greg0ire: For now, let's stick with the ugly version, and then when we use doctrine/coding-standard 11, it will be pretty again?

derrabus (approved):
I did not look into each and every file, there's way too many of them. The files I've picked looked good. Since this affects mainly tests, I think the PR is good to merge.

TODO for a follow-up:

* Adjust UPGRADE.md because you've removed the annotation driver.
* Purge annotations from the docs.

greg0ire:
Will do, thanks!

PR #10122, opened 2022-10-11
Automap events in AttachEntityListenersListener

HypeMC:
Required for https://github.com/doctrine/DoctrineBundle/issues/1560

Currently, the `AttachEntityListenersListener::addEntityListener()` method requires an `eventName` to be specified when registering a listener. With this change `null` can be passed allowing the event names to be read from the listener class by iterating over the method names.


derrabus (approved):

derrabus:
Thanks!

Issue #10049, opened 2022-09-21
PHP readonly properties and inheritance cause error on hydration


chapterjason:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.13.1

#### Summary

Using php readonly properties and inheritance (changing scope) causes an error while trying to set the property in the hydration process.

#### Current behavior

<details>
<summary>Error from the doctrine reproducer</summary>

```
PHP Fatal error:  Uncaught Error: Cannot initialize readonly property Chapterjason\DoctrineReadonlyReproducer\ValueObject\AggregatedRootId::$value from scope Chapterjason\DoctrineReadonlyReproducer\ValueObject\BookId in /[...]/doctrine-readonly-reproducer/vendor/doctrine/persistence/src/Persistence/Reflection/RuntimePublicReflectionProperty.php:59
Stack trace:
#0 /[...]/doctrine-readonly-reproducer/vendor/doctrine/persistence/src/Persistence/Reflection/RuntimePublicReflectionProperty.php(59): ReflectionProperty->setValue(Object(Chapterjason\DoctrineReadonlyReproducer\ValueObject\BookId), 'agri-632b6cb768...')
#1 /[...]/doctrine-readonly-reproducer/vendor/doctrine/persistence/src/Persistence/Reflection/TypedNoDefaultReflectionPropertyBase.php(64): Doctrine\Persistence\Reflection\RuntimePublicReflectionProperty->setValue(Object(Chapterjason\DoctrineReadonlyReproducer\ValueObject\BookId), 'agri-632b6cb768...')
#2 /[...]/doctrine-readonly-reproducer/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ReflectionReadonlyProperty.php(40): Doctrine\Persistence\Reflection\TypedNoDefaultRuntimePublicReflectionProperty->setValue(Object(Chapterjason\DoctrineReadonlyReproducer\ValueObject\BookId), 'agri-632b6cb768...')
#3 /[...]/doctrine-readonly-reproducer/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ReflectionEmbeddedProperty.php(81): Doctrine\ORM\Mapping\ReflectionReadonlyProperty->setValue(Object(Chapterjason\DoctrineReadonlyReproducer\ValueObject\BookId), 'agri-632b6cb768...')
[...]

PR #10126, opened 2022-10-12
Modernize documentation code

greg0ire:
- migrate to attributes;
- add helpful phpdoc annotations;
- use typed properties;
- add type declarations.

### DIFF
--- a/docs/en/cookbook/decorator-pattern.rst
+++ b/docs/en/cookbook/decorator-pattern.rst
@@ -23,48 +23,32 @@ concrete subclasses, ``ConcreteComponent`` and ``ConcreteDecorator``.
 
     namespace Test;
 
-    /**
-     * @Entity
-     * @InheritanceType("SINGLE_TABLE")
-     * @DiscriminatorColumn(name="discr", type="string")
-     * @DiscriminatorMap({"cc" = "Test\Component\ConcreteComponent",
-        "cd" = "Test\Decorator\ConcreteDecorator"})
-     */
+    #[Entity]
+    #[InheritanceType('SINGLE_TABLE')]
+    #[DiscriminatorColumn(name: 'discr', type: 'string')]
+    #[DiscriminatorMap(['cc' => Component\ConcreteComponent::class,
+        'cd' => Decorator\ConcreteDecorator::class])]
     abstract class Component
     {
 
-        /**
-         * @Id @Column(type="integer")
-         * @GeneratedValue(strategy="AUTO")
-         */
-        protected $id;
+        #[Id, Column]
+        #[GeneratedValue(strategy: 'AUTO')]
+        protected int|null $id = null;
 
-        /** @Column(type="string", nullable=true) */
+        #[Column(type: 'string', nullable: true)]
         protected $name;
 
-        /**
-         * Get id
-         * @return integer $id
-         */
-        public function getId()
+        public function getId(): int|null
         {
             return $this->id;
         }
 
-        /**
-         * Set name
-         * @param string $name
-         */
-        public function setName($name)
+        public function setName(string $name): void
         {
             $this->name = $name;
         }
 
-        /**
-         * Get name
-         * @return string $name
-         */
-        public function getName()
+        public function getName(): string
         {
             return $this->name;
         }
@@ -86,7 +70,7 @@ purpose of keeping this example simple).
 
     use Test\Component;
 
-    /** @Entity */
+    #[Entity]
     class ConcreteComponent extends Component
     {}
 
@@ -103,14 +87,11 @@ use a ``MappedSuperclass`` for this.
 
     namespace Test;
 
-    /** @MappedSuperclass */
+    #[MappedSuperclass]
     abstract class Decorator extends Component
     {
-
-        /**
-         * @OneToOne(targetEntity="Test\Component", cascade={"all"})
-         * @JoinColumn(name="decorates", referencedColumnName="id")
-         */
+        #[OneToOne(targetEntity: Component::class, cascade: ['all'])]
+        #[JoinColumn(name: 'decorates', referencedColumnName: 'id')]
         protected $decorates;
 
         /**
@@ -126,25 +107,19 @@ use a ``MappedSuperclass`` for this.
          * (non-PHPdoc)
          * @see Test.Component::getName()
          */
-        public function getName()
+        public function getName(): string
         {
     	    return 'Decorated ' . $this->getDecorates()->getName();
         }
 
-        /**
-         * the component being decorated
-         * @return Component
-         */
-        protected function getDecorates()
+        /** the component being decorated */
+        protected function getDecorates(): Component
         {
     	    return $this->decorates;
         }
 
-        /**
-         * sets the component being decorated
-         * @param Component $c
-         */
-        protected function setDecorates(Component $c)
+        /** sets the component being decorated */
+        protected function setDecorates(Component $c): void
         {
     	    $this->decorates = $c;
         }
@@ -187,27 +162,19 @@ of the getSpecial() method to its return value.
 
     use Test\Decorator;
 
-    /** @Entity */
+    #[Entity]
     class ConcreteDecorator extends Decorator
     {
 
-        /** @Column(type="string", nullable=true) */
-        protected $special;
+        #[Column(type: 'string', nullable: true)]
+        protected string|null $special = null;
 
-        /**
-         * Set special
-         * @param string $special
-         */
-        public function setSpecial($special)
+        public function setSpecial(string|null $special): void
         {
             $this->special = $special;
         }
 
-        /**
-         * Get special
-         * @return string $special
-         */
-        public function getSpecial()
+        public function getSpecial(): string|null
         {
             return $this->special;
         }
@@ -216,7 +183,7 @@ of the getSpecial() method to its return value.
          * (non-PHPdoc)
          * @see Test.Component::getName()
          */
-        public function getName()
+        public function getName(): string
         {
             return '[' . $this->getSpecial()
                 . '] ' . parent::getName();
@@ -270,4 +237,3 @@ objects
 
     echo $d->getName();
     // prints: [Really] Decorated Test Component 2
-
--- a/docs/en/cookbook/working-with-datetime.rst
+++ b/docs/en/cookbook/working-with-datetime.rst
@@ -15,13 +15,16 @@ these comparisons are always made **BY REFERENCE**. That means the following cha
 .. code-block:: php
 
     <?php
-    /** @Entity */
+
+    use DateTime;
+
+    #[Entity]
     class Article
     {
-        /** @Column(type="datetime") */
-        private $updated;
+        #[Column(type='datetime')]
+        private DateTime $updated;
 
-        public function setUpdated()
+        public function setUpdated(): void
         {
             // will NOT be saved in the database
             $this->updated->modify("now");
@@ -33,12 +36,14 @@ The way to go would be:
 .. code-block:: php
 
     <?php
+    use DateTime;
+
     class Article
     {
-        public function setUpdated()
+        public function setUpdated(): void
         {
             // WILL be saved in the database
-            $this->updated = new \DateTime("now");
+            $this->updated = new DateTime("now");
         }
     }
 
@@ -84,16 +89,14 @@ the UTC time at the time of the booking and the timezone the event happened in.
 
     namespace DoctrineExtensions\DBAL\Types;
 
+    use DateTimeZone;
     use Doctrine\DBAL\Platforms\AbstractPlatform;
     use Doctrine\DBAL\Types\ConversionException;
     use Doctrine\DBAL\Types\DateTimeType;
 
     class UTCDateTimeType extends DateTimeType
     {
-        /**
-         * @var \DateTimeZone
-         */
-        private static $utc;
+        private static DateTimeZone $utc;
 
         public function convertToDatabaseValue($value, AbstractPlatform $platform)
         {
@@ -126,10 +129,10 @@ the UTC time at the time of the booking and the timezone the event happened in.
 
             return $converted;
         }
-        
-        private static function getUtc(): \DateTimeZone
+
+        private static function getUtc(): DateTimeZone
         {
-            return self::$utc ?: self::$utc = new \DateTimeZone('UTC');
+            return self::$utc ??= new DateTimeZone('UTC');
         }
     }
 
--- a/docs/en/reference/association-mapping.rst
+++ b/docs/en/reference/association-mapping.rst
@@ -40,19 +40,17 @@ A many-to-one association is the most common association between objects. Exampl
     .. code-block:: php
 
         <?php
-        /** @Entity */
+        #[Entity]
         class User
         {
             // ...
 
-            /**
-             * @ManyToOne(targetEntity="Address")
-             * @JoinColumn(name="address_id", referencedColumnName="id")
-             */
-            private $address;
+            #[ManyToOne(targetEntity: Address::class)]
+            #[JoinColumn(name: 'address_id', referencedColumnName: 'id')]
+            private Address|null $address = null;
         }
 
-        /** @Entity */
+        #[Entity]
         class Address
         {
             // ...
@@ -82,9 +80,11 @@ A many-to-one association is the most common association between objects. Exampl
 
 .. note::
 
-    The above ``@JoinColumn`` is optional as it would default
+    The above ``#[JoinColumn]`` is optional as it would default
     to ``address_id`` and ``id`` anyways. You can omit it and let it
     use the defaults.
+    Likewise, inside the ``#[ManyToOne]`` annotation you can omit the
+    ``targetEntity`` attribute and it will default to ``Address``.
 
 Generated MySQL Schema:
 
@@ -114,22 +114,20 @@ references one ``Shipment`` entity.
     .. code-block:: php
 
         <?php
-        /** @Entity */
+        #[Entity]
         class Product
         {
             // ...
 
-            /**
-             * One Product has One Shipment.
-             * @OneToOne(targetEntity="Shipment")
-             * @JoinColumn(name="shipment_id", referencedColumnName="id")
-             */
-            private $shipment;
+            /** One Product has One Shipment. */
+            #[OneToOne(targetEntity: Shipment::class)]
+            #[JoinColumn(name: 'shipment_id', referencedColumnName: 'id')]
+            private Shipment|null $shipment = null;
 
             // ...
         }
 
-        /** @Entity */
+        #[Entity]
         class Shipment
         {
             // ...
@@ -156,7 +154,7 @@ references one ``Shipment`` entity.
                 name: shipment_id
                 referencedColumnName: id
 
-Note that the @JoinColumn is not really necessary in this example,
+Note that the ``#[JoinColumn]`` is not really necessary in this example,
 as the defaults would be the same.
 
 Generated MySQL Schema:
@@ -191,31 +189,27 @@ object.
     .. code-block:: php
 
         <?php
-        /** @Entity */
+        #[Entity]
         class Customer
         {
             // ...
 
-            /**
-             * One Customer has One Cart.
-             * @OneToOne(targetEntity="Cart", mappedBy="customer")
-             */
-            private $cart;
+            /** One Customer has One Cart. */
+            #[OneToOne(targetEntity: Cart::class, mappedBy: 'customer')]
+            private Cart|null $cart = null;
 
             // ...
         }
 
-        /** @Entity */
+        #[Entity]
         class Cart
         {
             // ...
 
-            /**
-             * One Cart has One Customer.
-             * @OneToOne(targetEntity="Customer", inversedBy="cart")
-             * @JoinColumn(name="customer_id", referencedColumnName="id")
-             */
-            private $customer;
+            /** One Cart has One Customer. */
+            #[OneToOne(targetEntity: Customer::class, inversedBy: 'cart')]
+            #[JoinColumn(name: 'customer_id', referencedColumnName: 'id')]
+            private Customer|null $customer = null;
 
             // ...
         }
@@ -281,17 +275,15 @@ below.
 .. code-block:: php
 
     <?php
-    /** @Entity */
+    #[Entity]
     class Student
     {
         // ...
 
-        /**
-         * One Student has One Mentor.
-         * @OneToOne(targetEntity="Student")
-         * @JoinColumn(name="mentor_id", referencedColumnName="id")
-         */
-        private $mentor;
+        /** One Student has One Mentor. */
+        #[OneToOne(targetEntity: Student::class)]
+        #[JoinColumn(name: 'mentor_id', referencedColumnName: 'id')]
+        private Student|null $mentor = null;
 
         // ...
     }
@@ -331,15 +323,16 @@ bidirectional many-to-one.
         <?php
         use Doctrine\Common\Collections\ArrayCollection;
 
-        /** @Entity */
+        #[Entity]
         class Product
         {
             // ...
             /**
              * One product has many features. This is the inverse side.
-             * @OneToMany(targetEntity="Feature", mappedBy="product")
+             * @var Collection<int, Feature>
              */
-            private $features;
+            #[OneToMany(targetEntity: Feature::class, mappedBy: 'product')]
+            private Collection $features;
             // ...
 
             public function __construct() {
@@ -347,16 +340,14 @@ bidirectional many-to-one.
             }
         }
 
-        /** @Entity */
+        #[Entity]
         class Feature
         {
             // ...
-            /**
-             * Many features have one product. This is the owning side.
-             * @ManyToOne(targetEntity="Product", inversedBy="features")
-             * @JoinColumn(name="product_id", referencedColumnName="id")
-             */
-            private $product;
+            /** Many features have one product. This is the owning side. */
+            #[ManyToOne(targetEntity: Product::class, inversedBy: 'features')]
+            #[JoinColumn(name: 'product_id', referencedColumnName: 'id')]
+            private Product|null $product = null;
             // ...
         }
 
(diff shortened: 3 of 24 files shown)


docs/en/reference/working-with-associations.rst
```diff
@@ -32,62 +32,66 @@ information about its type and if it's the owning or inverse side.
 .. code-block:: php
```
greg0ire, line 32: Should this stay as is or should I use
```suggestion
.. code-block:: attribute
```

docs/en/reference/working-with-associations.rst
```diff
@@ -32,62 +32,66 @@ information about its type and if it's the owning or inverse side.
...
     <?php
-    /** @Entity */
+    #[Entity]
     class User
     {
-        /** @Id @GeneratedValue @Column(type="string") */
-        private $id;
+        #[Id]
+        #[GeneratedValue]
+        #[Column(type: 'string')]
```
greg0ire: Should we recommend one attribute per line, or `#[Id, GeneratedValue, Column(type: 'string')`?

docs/en/reference/working-with-associations.rst
```diff
@@ -32,62 +32,66 @@ information about its type and if it's the owning or inverse side.
...
+        private string $id;
```
greg0ire: Not sure why `string` was used here in the first place (instead of int). Also, should I be using `string|null` here? Or is it better that the user gets an error when they attempt to get an id that hasn't been initialized yet?

docs/en/reference/working-with-associations.rst
```diff
@@ -32,62 +32,66 @@ information about its type and if it's the owning or inverse side.
...
 
         /**
          * Bidirectional - Many users have Many favorite comments (OWNING SIDE)
          *
-         * @ManyToMany(targetEntity="Comment", inversedBy="userFavorites")
-         * @JoinTable(name="user_favorite_comments")
+         * @var Collection<int, Comment>
```
greg0ire, line 44: The key doesn't really matter here, and I'm not sure it should be `int` either. I don't know if we should recommend omitting it, or using `array-key`, etc.

greg0ire:
:thinking: should I be targeting 2.13.x? After all, there is not much reason to wait for this.

docs/en/reference/working-with-associations.rst
derrabus: I think, we can leave it as is? 🤔

docs/en/reference/working-with-associations.rst
derrabus: Both are fine for me, but in this case I'd lean towards the one-liner.

docs/en/reference/working-with-associations.rst
derrabus: String looks wrong indeed. This should be an integer.

```suggestion
        private int|null $id = null;
```

`null` is explicitly a valid state for the entity (≠ the table) because the ORM will generate the ID on flush.

docs/en/reference/working-with-associations.rst
derrabus: The key does matter if you use the `indexedBy` option. If you don't, `int` is correct.

docs/en/reference/working-with-associations.rst
```diff
@@ -32,62 +32,66 @@ information about its type and if it's the owning or inverse side.
...
 
-        /**
-         * Unidirectional - Many-To-One
-         *
-         * @ManyToOne(targetEntity="Comment")
-         */
-        private $firstComment;
+        /** Unidirectional - Many-To-One */
+        #[ManyToOne(targetEntity: Comment::class)]
+        private Comment $firstComment;
```
derrabus:
```suggestion
        private Comment|null $firstComment = null;
```

`ManyToOne` relations are nullable by default.

docs/en/reference/working-with-associations.rst
```diff
@@ -32,62 +32,66 @@ information about its type and if it's the owning or inverse side.
...
+        private Collection $userFavorites;
 
         /**
          * Bidirectional - Many Comments are authored by one user (OWNING SIDE)
-         *
-         * @ManyToOne(targetEntity="User", inversedBy="commentsAuthored")
          */
-         private $author;
+        #[ManyToOne(targetEntity: User::class, inversedBy: 'commentsAuthored')]
+        private User $author;
```
derrabus:
```suggestion
        private User|null $author = null;
```

derrabus:
> 🤔 should I be targeting 2.13.x? After all, there is not much reason to wait for this.

I wouldn't wait with the 2.14.0 release too long either, so targeting 2.14.x is just fine imho.

docs/en/reference/working-with-associations.rst
greg0ire: I don't know, unitialized (no state yet) sound valid too. And I don't know in what scenario it would make sense to call `getId()` before flush.

derrabus: > I don't know, unitialized (no state yet) sound valid too.

The uninitialized state is a booby trap. Only use it if you're certain that you will never access the property before initialization. If however you need to deal with that case that the property does not have a value (yet), better make it explicit and allow `null`. This also helps static analysis tools to help you.

> And I don't know in what scenario it would make sense to call `getId()` before flush.

A possible scenario would be a piece of code that needs to check if a given entity has been persisted yet.

greg0ire: Since there is no clean way to check that a property is initialized, I think you are right. People that prefer getting a clear exception can perform a null check in `getId()`.

greg0ire: If the entity gets deleted though, the code my get a wrong impression that it is persisted : https://github.com/doctrine/orm/issues/10049#issuecomment-1274213094

docs/en/reference/working-with-associations.rst
```diff
@@ -32,62 +32,64 @@ information about its type and if it's the owning or inverse side.
...
 
     <?php
-    /** @Entity */
+    #[Entity]
     class User
     {
-        /** @Id @GeneratedValue @Column(type="string") */
-        private $id;
+        #[Id, GeneratedValue, Column(type: 'integer')]
+        private int|null $id;
```
derrabus:
```suggestion
        private int|null $id = null;
```

We should initialize nullable properties with null to avoid the uninitialized state.

greg0ire:
> I wouldn't wait with the 2.14.0 release too long either, so targeting 2.14.x is just fine imho.

I retargeted to 2.13.x so as not to have things to convert when merging up.

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -33,12 +36,14 @@ The way to go would be:
 .. code-block:: php
 
     <?php
+    use DateTime;
+
     class Article
     {
-        public function setUpdated()
+        public function setUpdated(): DateTime
```
derrabus:
```suggestion
        public function setUpdated(): void
```

docs/en/cookbook/working-with-datetime.rst
```diff
@@ -126,10 +129,10 @@ the UTC time at the time of the booking and the timezone the event happened in.
 
             return $converted;
         }
-        
-        private static function getUtc(): \DateTimeZone
+
+        private static function getUtc(): DateTimeZone
         {
-            return self::$utc ?: self::$utc = new \DateTimeZone('UTC');
+            return self::$utc ?: self::$utc = new DateTimeZone('UTC');
```
derrabus: This won't work because you're accessing an uninitialized property to see if it's initialized. Only the `??`/`??=` operators will work here.

```suggestion
            return self::$utc ??= new DateTimeZone('UTC');
```

docs/en/reference/association-mapping.rst
```diff
@@ -82,7 +80,7 @@ A many-to-one association is the most common association between objects. Exampl
 
 .. note::
 
-    The above ``@JoinColumn`` is optional as it would default
+    The above ``#[JoinColumn]`` is optional as it would default
     to ``address_id`` and ``id`` anyways. You can omit it and let it
     use the defaults.
```
derrabus, line 85: When using attributes and typed properties, we can also omit the `targetEntity`. Shall we add another note for that?

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -672,18 +672,18 @@ The same restrictions apply for the reference of related entities.
 
     DQL DELETE statements are ported directly into an SQL DELETE statement.
     Therefore, some limitations apply:
-    
+
     - Lifecycle events for the affected entities are not executed.
-    - A cascading ``remove`` operation (as indicated e. g. by ``cascade={"remove"}`` 
-      or ``cascade={"all"}`` in the mapping configuration) is not being performed 
+    - A cascading ``remove`` operation (as indicated e. g. by ``cascade={"remove"}``
+      or ``cascade={"all"}`` in the mapping configuration) is not being performed
```
derrabus: This is annotations syntax and should be changed to attributes syntax as well. We should scan the docs for more occurrences.

```suggestion
    - A cascading ``remove`` operation (as indicated e. g. by ``cascade: ['remove']``
      or ``cascade: ['all']`` in the mapping configuration) is not being performed
```

docs/en/reference/inheritance-mapping.rst
```diff
@@ -335,26 +328,20 @@ Example:
...
-             * @JoinTable(name="users_groups",
-             *  joinColumns={@JoinColumn(name="user_id", referencedColumnName="id")},
-             *  inverseJoinColumns={@JoinColumn(name="group_id", referencedColumnName="id")}
-             * )
-             */
+            #[JoinTable(name: 'users_groups')]
+            #[JoinColumn(name: 'user_id', referencedColumnName: 'id')]
+            #[InverseJoinColumn(name: 'group_id', referencedColumnName: 'id')]
+            #[ManyToMany(targetEntity: 'Group', inversedBy: 'users')]
             protected $groups;
```
derrabus: Property types?

docs/en/reference/inheritance-mapping.rst
```diff
@@ -377,6 +364,21 @@ Example:
          *      )
          * })
          */
+        #[Entity]
+        #[AssociationOverrides([
```
derrabus, line 353: Would you like to delete the annotations that those attributes are supposed to replace?

docs/en/reference/inheritance-mapping.rst
```diff
@@ -495,42 +497,41 @@ Could be used by an entity that extends a mapped superclass to override a field
...
-        /**
-         * @MappedSuperclass
-         */
+
+        #[MappedSuperclass]
         class User
         {
-            /** @Id @GeneratedValue @Column(type="integer", name="user_id", length=150) */
+            #[Id, GeneratedValue, Column(type: 'integer', name: 'user_id', length: 150)]
             protected $id;
```
derrabus: Types?

docs/en/reference/second-level-cache.rst
```diff
@@ -280,23 +280,17 @@ level cache region.
...
-             */
-            protected $id;
-
-            /**
-             * @Column(unique=true)
-             */
-            protected $name;
+            #[Id]
+            #[GeneratedValue]
+            #[Column(type: 'integer')]
```
derrabus:
```suggestion
            #[Column]
```

We omit the type for the string property below and there's no particular reason we need to be more explicit here.

docs/en/reference/second-level-cache.rst
```diff
@@ -343,36 +337,27 @@ It caches the primary keys of association and cache each element will be cached
...
-            protected $country;
-
-            /**
-             * @Cache("NONSTRICT_READ_WRITE")
-             * @OneToMany(targetEntity="City", mappedBy="state")
-             */
-            protected $cities;
+            #[Id]
+            #[GeneratedValue]
+            #[Column(type: 'integer')]
```
derrabus:
```suggestion
            #[Column]
```

docs/en/reference/security.rst
```diff
@@ -98,27 +98,28 @@ entity might look like this:
...
     {
-        /** @Id @Column(type="integer") @GeneratedValue */
-        private $id;
-        /** @Column */
-        private $email;
-        /** @Column(type="boolean") */
-        private $isAdmin;
-
-        public function fromArray(array $userInput)
+        #[Id, Column(type: 'integer'), GeneratedValue]
```
derrabus:
```suggestion
        #[Id, Column, GeneratedValue]
```

docs/en/reference/security.rst
```diff
@@ -98,27 +98,28 @@ entity might look like this:
...
+        private int|null $id = null;
+
+        #[Column]
+        private string $email;
+
+        #[Column(type: 'boolean')]
```
derrabus:
```suggestion
        #[Column]
```

docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -195,7 +195,7 @@ example we'll use an integer.
         class User
         {
             // ...
-            /** @Version @Column(type="integer") */
+            #[Version, Column(type: "integer")]
             private $version;
```
derrabus: Type?

docs/en/reference/working-with-associations.rst
```diff
@@ -32,62 +32,64 @@ information about its type and if it's the owning or inverse side.
 .. code-block:: php
 
     <?php
-    /** @Entity */
+    #[Entity]
     class User
     {
-        /** @Id @GeneratedValue @Column(type="string") */
-        private $id;
+        #[Id, GeneratedValue, Column(type: 'integer')]
```
derrabus:
```suggestion
        #[Id, GeneratedValue, Column]
```

docs/en/reference/working-with-associations.rst
```diff
@@ -32,62 +32,64 @@ information about its type and if it's the owning or inverse side.
...
-    /** @Entity */
+    #[Entity]
     class Comment
     {
-        /** @Id @GeneratedValue @Column(type="string") */
-        private $id;
+        #[Id]
+        #[GeneratedValue]
+        #[Column(type: 'string')]
+        private string $id;
```
derrabus:
```suggestion
        #[Id, GeneratedValue, Column)]
        private int|null $id = null;
```

docs/en/reference/working-with-objects.rst
```diff
@@ -95,19 +95,19 @@ from newly opened EntityManager.
 .. code-block:: php
 
     <?php
-    /** @Entity */
+    #[Entity]
     class Article
     {
-        /** @Id @Column(type="integer") @GeneratedValue */
+        #[Id, Column(type: 'integer'), GeneratedValue]
         private $id;
```
derrabus: Types?

docs/en/tutorials/composite-primary-keys.rst
```diff
@@ -149,42 +144,37 @@ We keep up the example of an Article with arbitrary attributes, the mapping look
...
-         * @Entity
-         */
+        #[Entity]
         class Article
         {
-            /** @Id @Column(type="integer") @GeneratedValue */
-            private $id;
-            /** @Column(type="string") */
-            private $title;
+            #[Id, Column(type='integer'), GeneratedValue]
```
derrabus: Syntax! There's more errors like this in this file.

```suggestion
            #[Id, Column(type: 'integer'), GeneratedValue]
```

derrabus (changes requested):

greg0ire:
@derrabus please take another look :pray:

docs/en/cookbook/decorator-pattern.rst
```diff
@@ -187,27 +162,19 @@ of the getSpecial() method to its return value.
...
     use Test\Decorator;
 
-    /** @Entity */
+    #[Entity]
     class ConcreteDecorator extends Decorator
     {
 
-        /** @Column(type="string", nullable=true) */
+        #[Column(type: 'string', nullable: true)]
         protected $special;
```
derrabus:
```suggestion
        protected string|null $special = null;
```

docs/en/cookbook/decorator-pattern.rst
```diff
@@ -187,27 +162,19 @@ of the getSpecial() method to its return value.
...
 
-        /**
-         * Set special
-         * @param string $special
-         */
-        public function setSpecial($special)
+        public function setSpecial(string $special): void
```
derrabus:
```suggestion
        public function setSpecial(string|null $special): void
```

The property is annotated as `nullable: true`.

docs/en/cookbook/decorator-pattern.rst
```diff
@@ -187,27 +162,19 @@ of the getSpecial() method to its return value.
...
         {
             $this->special = $special;
         }
 
-        /**
-         * Get special
-         * @return string $special
-         */
-        public function getSpecial()
+        public function getSpecial(): string
```
derrabus:
```suggestion
        public function getSpecial(): string|null
```

docs/en/tutorials/composite-primary-keys.rst
```diff
@@ -23,33 +23,28 @@ and year of production as primary keys:
...
-
-            public function __construct($name, $year)
-            {
-                $this->name = $name;
-                $this->year = $year;
+            public function __construct(
+                #[Id, Column(type: 'string')]
+                string $name,
+                #[Id, Column(type: 'integer')]
+                int $year,
```
derrabus:
```suggestion
                #[Id, Column(type: 'string')]
                private string $name,
                #[Id, Column(type: 'integer')]
                private int $year,
```

docs/en/tutorials/composite-primary-keys.rst
```diff
@@ -240,21 +230,18 @@ One good example for this is a user-address relationship:
...
-        /**
-         * @Entity
-         */
+
+        #[Entity]
         class User
         {
-            /** @Id @Column(type="integer") @GeneratedValue */
+            #[Id, Column(type: 'integer'), GeneratedValue]
             private $id;
```
derrabus:
```suggestion
            private int|null $id = null;
```

docs/en/tutorials/composite-primary-keys.rst
```diff
@@ -240,21 +230,18 @@ One good example for this is a user-address relationship:
...
 
-        /**
-         * @Entity
-         */
+        #[Entity]
         class Address
         {
-            /** @Id @OneToOne(targetEntity="User") */
+            #[Id, OneToOne(targetEntity: User::class)]
             private $user;
```
derrabus:
```suggestion
            private User|null $user = null;
```

docs/en/tutorials/composite-primary-keys.rst
```diff
@@ -288,68 +275,72 @@ of products purchased and maybe even the current price.
...
+        #[Column(type: 'boolean')]
+        private bool $paid = false;
+        #[Column(type: 'boolean')]
+        private bool $shipped = false;
+        #[Column(type: 'datetime')]
+        private DateTime $created;
 
         public function __construct(Customer $customer)
         {
             $this->customer = $customer;
```
derrabus: Shall we propagate CPP here?

```suggestion
        public function __construct(
            #[ManyToOne(targetEntity: Customer::class)]
            private Customer $customer,
        ) {
```

docs/en/tutorials/ordered-associations.rst
```diff
@@ -14,18 +14,16 @@ can specify the ``@OrderBy`` in the following way:
...
         class User
         {
             // ...
-        
-            /**
-             * @ManyToMany(targetEntity="Group")
-             * @OrderBy({"name" = "ASC"})
-             **/
+
+            #[ManyToMany(targetEntity: "Group")]
```
derrabus:
```suggestion
            #[ManyToMany(targetEntity: Group::class)]
```

docs/en/tutorials/ordered-associations.rst
```diff
@@ -14,18 +14,16 @@ can specify the ``@OrderBy`` in the following way:
...
+            #[OrderBy(["name" => "ASC"])]
             private $groups;
```
derrabus:
```suggestion
            private Collection $groups;
```

derrabus (changes requested):

greg0ire:
> Shall we propagate CPP here?

I applied your suggestions and removed the property that was outside the constructor for that particular comment.

derrabus (approved):

SenseException (approved):
PHP 7.4 is nearly gone, but it still feels weird to have attributes as the primary configuration when PHP 7 is still supported in composer :-D

docs/en/reference/basic-mapping.rst
```diff
@@ -451,17 +449,15 @@ besides specifying the sequence's name:
 
 .. configuration-block::
 
-    .. code-block:: php
+    .. code-block:: attribute
```
stof, line 452: As we have several formats here, shouldn't we keep the annotation code-block too ?

docs/en/cookbook/decorator-pattern.rst
```diff
@@ -23,48 +23,32 @@ concrete subclasses, ``ConcreteComponent`` and ``ConcreteDecorator``.
...
     namespace Test;
 
-    /**
-     * @Entity
-     * @InheritanceType("SINGLE_TABLE")
-     * @DiscriminatorColumn(name="discr", type="string")
-     * @DiscriminatorMap({"cc" = "Test\Component\ConcreteComponent",
-        "cd" = "Test\Decorator\ConcreteDecorator"})
-     */
+    #[Entity]
```
oleg-andreyev, line 26: Despite the fact that branch is already merged: I personally prefer multiple attributes in single '#[...]' looks a bit cleaner ;)

docs/en/reference/change-tracking-policies.rst
```diff
@@ -78,18 +77,16 @@ follows:
...
+    #[ChangeTrackingPolicy('NOTIFY')]
     class MyEntity implements NotifyPropertyChanged
     {
         // ...
-    
-        private $_listeners = array();
-    
-        public function addPropertyChangedListener(PropertyChangedListener $listener)
+
+        private array $_listeners = array();
```
oleg-andreyev, line 87: Short array?

docs/en/reference/faq.rst
```diff
@@ -32,11 +32,12 @@ upon insert:
 
     class User
     {
-        const STATUS_DISABLED = 0;
-        const STATUS_ENABLED = 1;
+        private const STATUS_DISABLED = 0;
```
oleg-andreyev, line 35: If referring to 8.1, enum example can be used here :)

docs/en/reference/inheritance-mapping.rst
```diff
@@ -38,38 +38,36 @@ Example:
...
     {
-        /** @Id @Column(type="integer") */
-        private $id;
-        /** @Column(type="string") */
-        private $name;
-    
+        #[Id, Column(type: 'integer')]
+        private int|null $id = null;
+        #[Column(type: 'string')]
+        private string $name;
```
oleg-andreyev, line 61: Probably should provide a default value, otherwise: creating new entity and calling getName will return fatal - if getter is strict too.

docs/en/reference/second-level-cache.rst
```diff
@@ -343,36 +337,27 @@ It caches the primary keys of association and cache each element will be cached
     .. code-block:: php
 
         <?php
-        /**
-         * @Entity
-         * @Cache("NONSTRICT_READ_WRITE")
-         */
+        #[Entity]
+        #[Cache(usage: 'NONSTRICT_READ_WRITE')]
```
oleg-andreyev, line 341: This string can be a real constant

docs/en/reference/working-with-associations.rst
```diff
@@ -132,11 +132,13 @@ relations of the ``User``:
...
     {
         // ...
-        public function getReadComments() {
+        /** @return Collection<int, Comment> */
+        public function getReadComments(): Collection {
              return $this->commentsRead;
         }
 
-        public function setFirstComment(Comment $c) {
+        /** @param Collection<int, Comment> $c */
```
oleg-andreyev, line 140: Wrong comment?

docs/en/reference/working-with-associations.rst
```diff
@@ -437,8 +443,10 @@ only accessing it through the User entity:
     // User entity
     class User
     {
-        private $id;
-        private $comments;
+        private int $id;
+
+        /** @var Collection<int, Comment> */
```
oleg-andreyev, line 448: Here: key can be mixed as we don't know how people are adding to collection via add or set :)

docs/en/reference/basic-mapping.rst
greg0ire: It would be better yeah, I'll try to find some time to do that.

docs/en/cookbook/decorator-pattern.rst
greg0ire: It was like that before :shrug:

docs/en/reference/change-tracking-policies.rst
greg0ire: Please consider sending a PR

greg0ire:
To anyone finding more stuff to say after the merge: please send PRs instead :pray:

docs/en/cookbook/decorator-pattern.rst
```diff
@@ -23,48 +23,32 @@ concrete subclasses, ``ConcreteComponent`` and ``ConcreteDecorator``.
...
+    #[InheritanceType('SINGLE_TABLE')]
+    #[DiscriminatorColumn(name: 'discr', type: 'string')]
+    #[DiscriminatorMap(['cc' => Component\ConcreteComponent::class,
+        'cd' => Decorator\ConcreteDecorator::class])]
```
michnovka, line 30: See discussion here https://github.com/slevomat/coding-standard/issues/1456

docs/en/reference/inheritance-mapping.rst
greg0ire: On the other hand, you might have a constructor preventing this entirely.

docs/en/reference/second-level-cache.rst
greg0ire: What do you mean?

docs/en/reference/working-with-associations.rst
greg0ire: Good catch 👍

docs/en/reference/working-with-associations.rst
greg0ire: And if they do, they should get a warning from PHPStan/Psalm thanks to this annotation, so I stand by it.

docs/en/reference/basic-mapping.rst
greg0ire: See https://github.com/doctrine/orm/pull/10157

Issue #10125, opened 2022-10-12
Unit of Work calculates unnecessary UPDATE queries for enum properties in cascaded collections


mbabker:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.13.3

#### Summary

While troubleshooting performance issues in one of our Symfony 6.1 applications, we found a request where a large number of UPDATE queries were being made which were generally unrelated to the actual request, screenshot is from New Relic:

![enum-updates-gone-awry](https://user-images.githubusercontent.com/368545/195388534-78428c1c-785b-4411-8e46-d32f11180189.png)

Except for the `time_sold` column (unrelated to this bug report, it's something I still need to debug), all of the changes in the UPDATE statements are touching columns that use backed enums in their mapping.

After debugging, I've found that this happens when persisting an entity that includes a "cascade persist" on a collection.  When the unit of work calculates the change set with one of these entities loaded, it's computing the following change for one of these enum fields:

```
  [338] =>
  array(1) {
    'status' =>
    array(2) {
      [0] =>
      enum App\LotStatus::AVAILABLE : string("available");
      [1] =>
      string(9) "available"
    }
  }
```

#### Current behavior

Unit of Work calculates incorrect changes, causing unnecessary UPDATE statements during a transaction

#### How to reproduce

https://github.com/mbabker/enum-uow-bug includes the minimal reproducer I was able to build to represent this issue.

Ensuring you have `sqlite` available on your system, clone the repo and run the following commands:

```sh
php app.php orm:schema-tool:create
php app.php orm:schema-tool:update --force
php app.php app:seed-database
php app.php app:update-auction-name --auction-id=1 --name="Changed Name"
```

When the last command runs, it'll dump out the computed change set from the Unit of Work as well as the log records from the middleware:

```
array(3) {
  [342] =>
  array(3) {
    'name' =>
    array(2) {
      [0] =>
      string(13) "Testing Event"
      [1] =>
      string(12) "Changed Name"
    }
    'type' =>
    array(2) {
      [0] =>
      enum App\AuctionType::LIVE : string("live");
      [1] =>
      string(4) "live"
    }
    'status' =>
    array(2) {
      [0] =>
      enum App\AuctionStatus::ACTIVE : string("active");
      [1] =>
      string(6) "active"
    }
  }
  [338] =>
  array(1) {
    'status' =>
    array(2) {
      [0] =>
      enum App\LotStatus::AVAILABLE : string("available");
[...]

PR #10131, opened 2022-10-13
Make `ReflectionEnumProperty::getValue()` return the enum

### LINKED ISSUE #10124: UnitOfWork calculates change set for enum value when the value stays the same
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.13.2

#### Summary

Ever since the enum bug fixes introduced in 2.13.2 https://github.com/doctrine/orm/pull/10041/files and subsequently 2.13.3 https://github.com/doctrine/orm/pull/10058/files the UnitOfWork will always calculate a change set for the entity because it compares the BackedEnum vs a string.
I've managed to narrow the issue to https://github.com/doctrine/orm/blob/1ed0057276b3846ba6f1c3e77f926bb69467a033/lib/Doctrine/ORM/UnitOfWork.php#L752

Let's assume we have a backed enum `Test` with the value `'test'`
`$orgValue` is a BackedEnum `Test`
`$actualValue` is a string `test`

This results in the entity being force updated even though no *real* changes were present.

This has major implications in our codeception test suite, since, for example, we call the endpoint to create an entity, get that entity in the codeception test (which uses doctrine orm) and it will have the enum value of `foo`, call another endpoint which will change the enum to `bar` (but in our codeception test the entity is still `foo`) so when we call flush to update another entity (unrelated to that one) it will override `bar` with `foo` in the database resulting in the test breaking since the enum changed value unexpectedly.

But this is a general thing, lets say you grab an entity X (with enum value `foo`) (without updating anything on it) somehow the enum changes from another call (`bar`) (but remember we havent updated it in this call) and we call flush to update/persist an unrelated entity Y, X would get updated in the database with the enum value `foo`. (when it was `bar`) thus resulting in bugs

#### Current behavior

We grab an entity that has an enum type value
A flush happens without changing anything on it, and it gets updated.

#### How to reproduce

To test might be a bit hard...
Lets assume the Entity has the enum value of `Foo` (`'foo'`) and is flushed to the database (first time, stored).

Need to clear the entity manager and re-load it from the database (tried to replicate it when first persisted, but it wont replicate then)

One way might be to manually update the value in the database (query `UPDATE table SET enum_type = 'bar'`)
Do a flush on the entity without changing anything, and check that the database should still be `bar` (if it is `foo` the test should fail since nothing should have been changed so no flush for that entity)

Another way could be using a post update event which would update the column every time the entity gets updated, ie: updatedAt. After the first flush where it's inserted, the updatedAt would be created, and after the second flush it should not call the post update, and not update the updatedAt and we could check this value that it didnt change.

#### Expected behavior

The expected behaviour is that if an enum does not change value, it should not trigger an update on the entity in the database.

simPod:
UoW compares orig and actual value. Orig value is the enum. But without this change, actual value is the inner value of the BackedEnum. Therefore, Entities with enum field are always considered a change.

So I though about it like this:

1. the orig and actual values MUST be enums so they're comparable – implies that property must return enum as a value and not the backed value.
2. I've found out the enumType is not filled when hydrating entity.

Resolves #10124, #10125.

### DIFF
--- a/lib/Doctrine/ORM/Id/AssignedGenerator.php
+++ b/lib/Doctrine/ORM/Id/AssignedGenerator.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Id;
 
+use BackedEnum;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Exception\EntityMissingAssignedId;
 
@@ -37,6 +38,10 @@ class AssignedGenerator extends AbstractIdGenerator
             if (isset($class->associationMappings[$idField])) {
                 // NOTE: Single Columns as associated identifiers only allowed - this constraint it is enforced.
                 $value = $em->getUnitOfWork()->getSingleIdentifierValue($value);
+            }
+
+            if ($value instanceof BackedEnum) {
+                $value = $value->value;
             }
 
             $identifier[$idField] = $value;
--- a/lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
+++ b/lib/Doctrine/ORM/Mapping/ReflectionEnumProperty.php
@@ -38,7 +38,7 @@ class ReflectionEnumProperty extends ReflectionProperty
      *
      * @param object|null $object
      *
-     * @return int|string|int[]|string[]|null
+     * @return BackedEnum|null
      */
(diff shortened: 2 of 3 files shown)


simPod:
Maybe I'm fighting against 10074

PR #10133, opened 2022-10-14
Clarify difference between transactional() methods of Connection and EntityManager

sbusch:
One could interpret the old description as if `Connection#transactional()` would not rollback the transaction. Also, the fact that the `EntityManager` gets closed in case of an exception was not mentioned.

### DIFF
--- a/docs/en/reference/transactions-and-concurrency.rst
+++ b/docs/en/reference/transactions-and-concurrency.rst
@@ -114,8 +114,8 @@ functionally equivalent to the previously shown code looks as follows:
 The difference between ``Connection#transactional($func)`` and
 ``EntityManager#transactional($func)`` is that the latter
 abstraction flushes the ``EntityManager`` prior to transaction
-commit and rolls back the transaction when an
-exception occurs.
+commit and in case of an exception the ``EntityManager`` gets closed
+in addition to the transaction rollback.
 
 .. _transactions-and-concurrency_exception-handling:
 


docs/en/reference/transactions-and-concurrency.rst
```diff
@@ -114,8 +114,8 @@ functionally equivalent to the previously shown code looks as follows:
 The difference between ``Connection#transactional($func)`` and
 ``EntityManager#transactional($func)`` is that the latter
 abstraction flushes the ``EntityManager`` prior to transaction
-commit and rolls back the transaction when an
-exception occurs.
+commit and in case of an exception the ``EntityManager`` gets closed
+in addition to the transaction rollback.
```
SenseException, line 118: Maybe this formulation is more clear. What do you think about it?
```suggestion
after the transaction rollback.
```

SenseException (changes requested):

derrabus:
@sbusch Do you want to reply to @SenseException's review?

docs/en/reference/transactions-and-concurrency.rst
sbusch: Sorry for late reply.

IMO there's no benefit to imply an order (EM gets closed **after** transaction rollback; is that even the correct order?) And for the user, it makes no difference for understanding the behavior. That's why I decided to write "in addition to", in contrast to the first part of the changed text where order matters.

But in the end it's a matter of taste 😃

sbusch:
@derrabus I posted an answer. Its a matter of taste, up to you maintainers. Shouldnt block the merge

SenseException (approved):

PR #10134, opened 2022-10-15
Run only relevant workflows

greg0ire:
The downside of this is that we will have to tweak the settings so that no job is required anymore.
The upside is that builds should be faster, and less resource-intensive. Note that this particular pull request is going to run all the workflows, as I'm touching the workflow files.

Inspired by #4919, prompted by https://afup.org/talks/4016-papa-et-maman-nettoient-l-internet

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -1,0 +1,27 @@
+name: "Coding Standards"
+
+on:
+  pull_request:
+    branches:
+      - "*.x"
+    paths:
+      - .github/workflows/coding-standards.yml
+      - bin/**
+      - composer.*
+      - lib/**
+      - phpcs.xml.dist
+      - tests/**
+  push:
+    branches:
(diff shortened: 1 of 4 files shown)


derrabus (dismissed):

greg0ire:
Woops: https://github.com/doctrine/orm/pull/10135

mabar:
You can have jobs required even when they are cancelled or skipped. And you can also manage them via workflow file instead of UI
Feel free to copy my reusable workflow from here
https://github.com/orisai/github-workflows/blob/v1.x/docs/README.md#status-check-for-branch-protection

greg0ire:
@mabar I don't understand; what would that extra workflow do in the case of https://github.com/doctrine/orm/pull/10126, where there are no checks: https://github.com/doctrine/orm/pull/10126

Does the list inside `needs` need to be dynamic? Feel free to open a PR, it might be easier to discuss that way.

PR #10136, opened 2022-10-15
Stop triggering static analysis workflows on tests

greg0ire:
In my opinion it is not great that we do not run static analysis tools on tests, but since we do not, let us stop triggering extra jobs for no reason.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -11,7 +11,7 @@ on:
       - lib/**
       - phpstan*
       - psalm*
-      - tests/**
+      - tests/Doctrine/StaticAnalysis/**
   push:
     branches:
       - "*.x"
@@ -21,7 +21,7 @@ on:
       - lib/**
       - phpstan*
       - psalm*
-      - tests/**
+      - tests/Doctrine/StaticAnalysis/**
 
 jobs:
   static-analysis-phpstan:


SenseException:
I don't understand what's the reason for this PR.

beberlei:
👍 Static analysis on tests is really not necessary

greg0ire:
> I don't understand what's the reason for this PR.

@SenseException this should help:

https://github.com/doctrine/orm/blob/f08b67f0cc71076c90753709ad78a0a66e1343aa/phpstan-params.neon#L3-L5

https://github.com/doctrine/orm/blob/f08b67f0cc71076c90753709ad78a0a66e1343aa/psalm.xml#L12-L13

Right now, only `lib` and `tests/Doctrine/StaticAnalysis` (which I originally forgot about) are analysed by both tools.

greg0ire:
ping @SenseException @derrabus :slightly_smiling_face:

derrabus (approved):

SenseException:
I'm too much of a fan for SA for tests because it can improve strictness and might make tests fail where they should fail. Are we going to drop tests-directories in all repositories to be consistent?

greg0ire:
> Are we going to drop tests-directories in all repositories to be consistent?

I have no such plan, and I think other repositories already have a good level of SA, so I'd say let's keep the ORM an exception to this rule for now.

Issue #5542, opened 2015-12-10
Allow to not Compare Object Types by reference 


beberlei:
If we have a Doctrine DBAL type with an object (DateTime, any custom object), then its compared by reference and we cannot use mutable objects.

We should consider allowing a way to have mutable objects here and delegate the changeset compuutation to a service or do something clever.

Ocramius:
I proposed comparing the DB-side representation in the past, but it may be
expensive...

beberlei:
@Ocramius maybe we can enable this comparison based on a setting, so its only done for fields where its necessary.

DHager:
I expect that if someone has something like a `DateTime` or `ColorCode`, they will *always* want it to be compared via the same mechanism, regardless of which entity it appears on. So then it's the type of the field which implicitly determines the comparison-rules.

What about a system where the `EntityManager` can be given an (optional) `EntityScalarComparator` (or whatever you want to call it), and as part of the startup process it queries the comparator to say: "Hey, what datatypes should I be watching out for?"

Would it be possible to weave that information into the class-metadata as it is built? ("Oh, that field is one of the ones the Comparator said it was interested in mark it as `useComparator=true`)

Ocramius:
> @Ocramius maybe we can enable this comparison based on a setting, so its only done for fields where its necessary.

Possibly, but I'd still look at it after @guilhermeblanco is done with digging through JPA's field mappings.

 > Then `EntityScalarComparator` is where you might , say, have an option to change how `DateTime` is compared, or a mechanism to chain/overlay another comparison service to support your own custom objects.

The idea is to generate specific comparators per entity (you pretty much always compare entity to entity, not just field-to-field.

DHager:
> The idea is to generate specific comparators per entity (you pretty much always compare entity to entity, not just single fields).

I think @beberlei is describing a different scenario, where someone has used a [custom DBAL mapping type](http://doctrine-orm.readthedocs.org/projects/doctrine-dbal/en/latest/reference/types.html#custom-mapping-types) for a specific field on an entity. The issue is with how the ORM does dirty-checking: Since it always uses `===` only immutable DBAL types are possible.

JakeFalcor:
Probably affects objects serialization: http://stackoverflow.com/questions/30193351/how-to-update-doctrine-object-type-field

Fedik:
I have couple custom types, and it a bit annoying 😄 , so I got 2 idea:

[...]

PR #10137, opened 2022-10-15
Allow different strategy of Property change detection

arno14:
Proposal to implement different strategy concerning the property change computation.

https://github.com/doctrine/orm/issues/5542

todo:
- [  ] allow determining ChangeDetector from custom DBAL\type
- [  ] more functional test with differents type (json, etc..)
- [  ] unit test for the differents ChangeDetector
- [  ] allow using Yaml mapping
- [  ] allow using Xml mapping


Toflar:
I actually like the idea of this PR. Instead of using `===` to compare whether values are still the same, just give the developers a way to hook into that process and decide for themselves. It's a pity it never got any feedback 😢

PR #10138, opened 2022-10-15
Add custom highly optimized metadata cache

beberlei:
There have been a few performance features in PHP engine since we originally built the Metadata cache layer for PHP 5.3 and ORM 2.0 in 2010. Mainly, Opcache, Interned Strings, Packed Arrays and other optimizations of parsing literal code vs unserializing an object.

This custom metadata driver has been in production at Tideways for roughly 5 years now and we haven't gotten around to open-source it.

It creates a PHP file that is assigning properties on `ClassMetadata`, the complex ones using `var_export`. See an example generated cache file for the widely used `CmsUser` model in the testsuite: https://gist.github.com/beberlei/a10fea49b2140b029fe13d97b66f159e

These cache files can be stored in Opcache and are quicker to load than any other metadata cache mechanism.

Using the debug flag, an invalidation of the metadata when compared to the filemtime of the actual entity is implemented.

Loading the metadata with the wrong version of doctrine is guarded against with Composer Runtime API serializing the version of orm into the cache.

For Doctrine 3.0 it might be sensible to completly switch to this driver only and make sure metadata are generated at the same time as proxys for production.

Added this as a first solution proposal for https://twitter.com/benjamincremer/status/1481571522259083266 - also need to do some benchmarks and profiling, as well as fixing all the CI issues before undrafting.

One potentially problematic thing will be the PSR Cache version differences, maybe using two classes.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Cache/PhpMetadataCache.php
+++ b/lib/Doctrine/ORM/Mapping/Cache/PhpMetadataCache.php
@@ -6,6 +6,7 @@ namespace Doctrine\ORM\Mapping\Cache;
 
 use Composer\InstalledVersions;
 use Doctrine\Common\Cache\Psr6\CacheItem;
+use Doctrine\Common\Cache\Psr6\TypedCacheItem;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Psr\Cache\CacheItemInterface;
 use Psr\Cache\CacheItemPoolInterface;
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Mapping/Cache/PhpMetadataCache.php
```diff
@@ -0,0 +1,197 @@
...
+        return true;
+    }
+
+    /** @param string[] $keys */
+    public function deleteItems(array $keys): bool
+    {
+        return true;
+    }
+
+    public function saveDeferred(CacheItemInterface $item): bool
```
bcremer, line 176: Recent versions of `AbstractClassMetadataFactory` are calling `saveDeferred`  and `commit` instead of `save`.

See:  https://github.com/doctrine/persistence/blob/05612da375f8a3931161f435f91d6704926e6ec5/src/Persistence/Mapping/AbstractClassMetadataFactory.php#L219

bcremer:
As written in the twitter thread  the `Symfony\Component\Cache\Adapter\PhpFilesAdapter` in append only mode provides approximately the same performance as the deprecated `Doctrine\Common\Cache\PhpFileCache` and your custom `PhpMetadataCache.php`.

Nevertheless I very much support your idea of providing a default metadata-cache implementation in `doctrine/orm`.

> For Doctrine 3.0 it might be sensible to completly switch to this driver only and make sure metadata are generated at the same time as proxys for production.

PR #10141, opened 2022-10-17
Add CI jobs for SQLite3 driver

derrabus:


.github/workflows/continuous-integration.yml
```diff
@@ -38,15 +38,24 @@ jobs:
...
           - php-version: "8.1"
             dbal-version: "3@dev"
-          - php-version: "8.2"
+            extension: "pdo_sqlite"
+          - php-version: "7.4"
             dbal-version: "3@dev"
+            extension: "sqlite3"
+          - php-version: "8.1"
+            dbal-version: "3@dev"
+            extension: "sqlite3"
```
greg0ire, line 55: I think one extra job for that driver should be enough. Similarly, we only have one job for mysqli (well 2: one for MySQL, one for MariaDB). Do you expect to catch failures with one job that you would not catch with the other one?

derrabus: Okay, let's remove the PHP 7.4 job then.

greg0ire (approved):

PR #10145, opened 2022-10-17
Fix type doc blocks in annotation classes

derrabus:
In 3.0, I'd like to flag all public properties of our annotation classes as `readonly`. Since we always use them as is they were immutable, this is  the reasonable thing to do imho.

This PR prepares this change by adding a `@readonly` annotation to all dock blocks. I've also fixed a few other inconsistencies.


greg0ire (approved):
Potential improvement: define new `psalm-type` for string unions such as `'LAZY'|'EAGER'|"EXTRA_LAZY'`

derrabus:
Yeah, or make that an enum. 🤔

PR #10148, opened 2022-10-17
Stop testing intermediary versions

greg0ire:
In my opinion, testing with the lowest-supported version, the latest version of each major version, and optionally the next minor version should be enough.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -33,15 +33,12 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.2"
-          - "7.3"
           - "7.4"
-          - "8.0"
           - "8.1"
         dbal-version:
           - "default"
         include:
-          - php-version: "8.0"
+          - php-version: "8.1"
             dbal-version: "2.13"
           - php-version: "8.1"
             dbal-version: "3@dev"
@@ -103,7 +100,7 @@ jobs:
           - "9.6"
           - "14"
         include:
-          - php-version: "8.0"
+          - php-version: "8.1"
             dbal-version: "2.13"
             postgres-version: "14"
           - php-version: "8.2"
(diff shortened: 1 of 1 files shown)


SenseException:
I'm in favour of reducing the PHP versions for the ORM build, but I'd like to keep the lowest supported PHP version as long as it's allowed in the Composer constraints.

derrabus:
I think, having at least a smoke test for each supported PHP branch in reasonable. I don't think that we should drop all jobs for PHP 7.3 for instance as long as we support PHP 7.3.

greg0ire:
> I'm in favour of reducing the PHP versions for the ORM build, but I'd like to keep the lowest supported PHP version as long as it's allowed in the Composer constraints.

It's still tested: https://github.com/doctrine/orm/actions/runs/3268036772/jobs/5373969736

That's because of this configuration block: https://github.com/doctrine/orm/blob/90bc6dc300be0987cd75f2027e0612f05fb6f8ff/.github/workflows/continuous-integration.yml#L316-L346

> I think, having at least a smoke test for each supported PHP branch in reasonable. I don't think that we should drop all jobs for PHP 7.3 for instance as long as we support PHP 7.3.

Ok, I'll trust your judgement. Still, I'd be curious to know if we are likely to ever end up in a situation where we would have code that fails on one of the versions I removed, and not on the other ones.

derrabus:
> Ok, I'll trust your judgement. Still, I'd be curious to know if we are likely to ever end up in a situation where we would have code that fails on one of the versions I removed, and not on the other ones.

Note that it's not only about PHP versions. We're also testing against different sets of dependencies because the minimum PHP versions supported by our dependencies. But I also believe that the cases where a PHP 7.3 job would catch an error which would remain undetected by the PHP 7.1 or 7.4 job are rare. On the other hand, the SQLite tests are really fast, so keeping one of them per supported PHP version should not hurt too much.

PR #10153, opened 2022-10-18
Address dbal deprecations

greg0ire:


greg0ire:
@morozov the failing test calls the update command, then feeds its output to the DBAL. It seems that it is choking on `DROP TABLE quote-address;`, and the weird thing is that I got that information from the DBAL 2 job. The DBAL 3 jobs don't display that information, is that expected?
Regarding that `-`, did this test just discover a latent bug?

morozov:
> Regarding that `-`, did this test just discover a latent bug?

There are two issues with this potentially:
1. The `UpdateCommandTest` isn't isolated. It seems to drop whatever schema that the previous tests created. If you run it alone, it will pass. Prior to this change, the DBAL would omit all `DROP` statements, so the test would pass.
2. When accepting the `$table` parameter as a string `AbstractPlatform::getDropTableSQL()` expects it to be an SQL fragment (not just a value representing name), so the caller is supposed to quote the name. For some reason the `$table` with the name `quote-address` has `$quoted = false` in its properties which leads to the unquoted name being passed to the method. UPD: the problem seems to be in the DBAL which is what introspects the schema. The object names are automatically marked as quoted only by the Oracle and DB2 schema manager while certain quoting rules apply to all platforms.

> The DBAL 3 jobs don't display that information, is that expected?

No idea.

morozov:
I filed https://github.com/doctrine/dbal/issues/5776 for documentation purposes but it's one of the side effects of https://github.com/doctrine/dbal/issues/4357, so it won't be fixed in the near future.

greg0ire:
Thanks for looking into this! Do you think it will be fixed in the far future? Because I pushed something that should fix the issue until then.

@derrabus the fix implies using `stderr`, if you think it is right, I can review other Symfony commands to make them [unix filters](https://en.wikipedia.org/wiki/Filter_(software)#Unix) as well

morozov:
> Do you think it will be fixed in the far future?

It can be fixed even in the near future but I don't plan for it at the moment. If anybody is interested in fixing it themselves sooner, I can facilitate the process.

greg0ire:
Blocked by #10162

lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -956,10 +977,18 @@ public function getUpdateSchemaSql(array $classes, $saveMode = false)
         $schemaDiff = $comparator->compareSchemas($fromSchema, $toSchema);
 
         if ($saveMode) {
+            if (! method_exists($schemaDiff, 'toSaveSql')) {
+                throw GlobalNotSupported::createForDbal4('Using save mode when getting update schema SQL');
+            }
```
derrabus: This looks like dead code to me. We don't support DBAL 4 on this branch and on the 3.0.x branch, we're going to remove the `$saveMode` parameter anyway.

derrabus (approved):

PR #10155, opened 2022-10-19
feat: allow STI child entities to have non-nullable relationships

simPod:
Currently, child entities cannot have relationships marked as non-nullable because SchemaTool generates the relationship join column as `NOT NULL`.

It has to be marked as nullable, the same as is done for non-nullable columns that are present only in child entity.

### DIFF
--- a/src/Tools/SchemaTool.php
+++ b/src/Tools/SchemaTool.php
@@ -536,6 +536,7 @@ class SchemaTool
                 $this->gatherRelationJoinColumns(
                     $mapping->joinColumns,
                     $table,
+                    $class,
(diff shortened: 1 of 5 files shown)


mpdude:
Do you mean

Currently, child entities cannot have relationships marked as **nullable**…?

simPod:
as `non-nullable` as stated in the title.

https://github.com/doctrine/orm/pull/10155/files#diff-fe519274c2d9ecab59f442c05f190586e7320f74719505288502bdb0ca56bc9aR15

simPod:
Unstale

Issue #10132, opened 2022-10-14
Enum fields can't be an id field.


johannes85:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.13.3

#### Summary

Id Field with enumType is not working.

#### Current behavior

I have a entity called "display" with a primary key over multiple columns:
[id] int feedback_id (relation to feedback)
[id] string with enumType identity_type
string identity_id

There also is a parent entity called "feedback" with the following columns:
[id] int id
collection<display> displays  (relation to display)

When I query for a entry from feedback with a join to displays, this call failes when done a second time, with:
Object of class App\Entity\IdentityType could not be converted to string
in vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php:286

It is caused by this code:
```
foreach ($class->identifier as $fieldName) {
                $idHash .= ' ' . (isset($class->associationMappings[$fieldName])
                    ? $data[$class->associationMappings[$fieldName]['joinColumns'][0]['name']]
                    : $data[$fieldName]);
            }
```
not expecting $data[$fieldName]) to be an enum.

#### Expected behavior

Support for enum fields as index.

beberlei:
This is the expected behavior at the monent, should work with toString implemented

johannes85:
There is no way to implement toString for an enum.

beberlei:
I did not know / forgot this, hmmm

michnovka:
Why isnt identifier flattener used in this case?

Check https://github.com/doctrine/orm/pull/9629 this feature should have been already implemented and I am using enum IDs in my project without issue.

Please provide test which shows how its broken.

flight643:
I am also facing this problem. In my case enum field is a part of the composite primary key.  Latest working version is 2.13.1.
I checked #10158 would have fixed the problem if the https://github.com/doctrine/orm/pull/10158#discussion_r999745713 was taken into account

michnovka:
@flight643 please provide broken test case

tomudding:
I am experiencing the same problem on 2.13.4. While our actual implementation is a bit more complex I have been able to abstract it to [a failing test case](https://github.com/tomudding/orm/blob/45d5239b1434cec3ca4051f40316ba6f8eef0701/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10132Test.php).

PHPunit gives the following information:
```
[...]

PR #10158, opened 2022-10-19
Implement BC compatibel BackedEnum switch

SerethiX:


SerethiX:
Fixes  #10132

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
```diff
@@ -281,9 +281,13 @@ private function getEntityFromIdentityMap(string $className, array $data)
             $idHash = '';
 
             foreach ($class->identifier as $fieldName) {
-                $idHash .= ' ' . (isset($class->associationMappings[$fieldName])
-                    ? $data[$class->associationMappings[$fieldName]['joinColumns'][0]['name']]
-                    : $data[$fieldName]);
+                if (isset($class->associationMappings[$fieldName])) {
+                    $idHash .= ' ' . $data[$class->associationMappings[$fieldName]['joinColumns'][0]['name']];
+                } elseif (class_exists('\BackedEnum') && is_a($data[$fieldName], '\BackedEnum')) {
```
derrabus, line 286: The `class_exists()` feels unnecessary and also wrong, given that `BackedEnum` is actually an interface.

derrabus (changes requested):
Thank you, but as always: we need a test that covers your change.

derrabus:
Closing due to lack of feedback and given that the patch is apparently not correct. Feel free to open a new PR if you want to pick up the work.

PR #10162, opened 2022-10-22
Use error style for notifications

greg0ire:
`stderr` is not a great name for something that is not meant to be processed (piped into) a program, but to be read by humans. Most commands should use stderr, and some of them should partially use `stdout`, typically when dumping SQL requests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "doctrine/lexer": "^1.2.3",
         "doctrine/persistence": "^2.4 || ^3",
         "psr/cache": "^1 || ^2 || ^3",
-        "symfony/console": "^3.0 || ^4.0 || ^5.0 || ^6.0",
+        "symfony/console": "^3.3 || ^4.0 || ^5.0 || ^6.0",
         "symfony/polyfill-php72": "^1.23",
         "symfony/polyfill-php80": "^1.16"
     },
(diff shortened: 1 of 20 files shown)


greg0ire:
@derrabus doctrine/dbal 3.5.x has been released, which causes Psalm to fail. What should we do about this?
1. backport fixes to 2.13.x?
2. release 2.14.x and address the remaining issues?
3. add the issues to the Psalm baseline on 2.13.x?

derrabus:
> doctrine/dbal 3.5.x has been released, which causes Psalm to fail.

#10163

SenseException:
I don't know why, but 2.14 is still supporting symfony/console 3.0 and `SymfonyStyle::getErrorStyle()` was introduced in 3.3. As  long as we support such old versions the introduced getErrorStyle-calls are breaking changes.

greg0ire:
I'll try to understand why this wasn't caught by the prefer-lowest build, which apparently installs v3.4.17

UPDATE:

```
composer why-not symfony/console 3.3.0
Info from https://repo.packagist.org: #StandWithUkraine
phpbench/phpbench 1.2.7  requires  symfony/console (^4.2 || ^5.0  || ^6.0)
symfony/yaml      v6.1.6 conflicts symfony/console (<5.4)
vimeo/psalm       4.29.0 requires  symfony/console (^3.4.17 || ^4.1.6 || ^5.0 || ^6.0)
```

greg0ire:
@SenseException I bumped the constraint to 3.3 minimum, that shouldn't be an issue since Symfony v3 is no longer maintained anyway (we could drop support for v3 entirely IMO).

derrabus:
Indeed. We still declare compatibility with Symfony Console 3 mainly because we haven't yet used any features that were introduced later and not the other way round. Since Symfony 3 is EOL, we don't need to care about it anymore. So if bumping to 4.4 gets us any benefit, let's do it.

SenseException (approved):

PR #10169, opened 2022-10-24
Discriminator column generated value

vansari:
Situation:

We have a Table which have a discriminator column with a generated value. 

Expected:
If we persist a new Record of the mapped Entity, the Insert should not write the discriminator value because it is auto generated.

ISSUE:
The Insert tries to insert the default value of the entity. This will fail because the value can not be inserted if it is a generated column.

Solution:
This MR allows the dev to use generated Column as a discriminator column. If a new Entity should be persisted and the discriminator column has the generated Value set, the Insert will ignore the Value of the mapped column.

---------------------------------------
I need some help to write the correct Tests. 
Please can you tell me if the solution is done in the right way?

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -3243,6 +3243,13 @@ class ClassMetadataInfo implements ClassMetadata
             if (in_array($columnDef['type'], ['boolean', 'array', 'object', 'datetime', 'time', 'date'], true)) {
                 throw MappingException::invalidDiscriminatorColumnType($this->name, $columnDef['type']);
             }
+            if (! isset($columnDef['generated'])) {
+                $columnDef['generated'] = null;
+            }
+
+            if (!in_array($columnDef['generated'], ['NEVER', 'INSERT', 'ALWAYS', null], true)) {
+                throw MappingException::invalidGeneratedMode($columnDef['generated']);
+            }
 
             $this->discriminatorColumn = $columnDef;
         }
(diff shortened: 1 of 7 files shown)


vansari:
Another Idea is to use a bool Value to identify if the discriminator column has a generated value.

```
#[DiscriminatorColumn(name: 'discr', type: 'string', generated: true)]
#[DiscriminatorMap([
    0 => SubOne::class,
    1 => SubTwo::class,
])]
```

What do you think is the best way?

mpdude:
What exactly do you mean with "generated column"?

There has to be a column where Doctrine may put the information which class an entity has, so that it can pick the right class when hydrating the entity. Doctrine wants to write the discriminator value itself whenever an entity is written to that table.

vansari:
You have the possibility to create a column where the value for the discriminator column is generated by the DB. This will be realized with an Insert trigger. By this way the DB Admin disallow a wrong input for this column.

The column identifies which art of entry the record is.

Yes, Doctrine *wants* but doesnt *must write the value by itself. The user should have possibility to choose the way how to use the discriminator column.

My solution allows this. And my question is: Have I done it in the correct way? :)

mpdude:
> If a new Entity should be persisted and the discriminator column has the generated Value set, the Insert will ignore the Value of the mapped column.

How can the value be set when a new row (for the new entity) shall be inserted?

I have the impression that you want to use Doctrine update parts of rows that already exist in the database?

vansari:
Hi @mpdude 

I don't want to update Parts.
I have created a repo for an Example: https://github.com/vansari/orm-discriminator-example

I will generate an README as soon as possible.

But at the moment this Repo will show you how it should work. It will not break anything but it will allow the users to use generated columns as discriminator columns.

PR #10170, opened 2022-10-24
Merge 2.14.x into 3.0.x

derrabus:


derrabus:
The last commit contains some quick fixes to get the build green again with DBAL 4. We need dome cleanup afterwards, but that should be done in separate PRs.

greg0ire (approved):

PR #10175, opened 2022-10-25
Remove DriverChain

derrabus:


greg0ire (approved):
Looks good, but there is a mention [here](https://github.com/doctrine/orm/blob/2.13.x/docs/en/reference/advanced-configuration.rst). It needs to be changed on a lower branch.

derrabus:
Good catch, see #10179.

PR #10177, opened 2022-10-26
Purge annotations

greg0ire:
I refrained form touching `lib` too much, I noticed there is an `Annotation` interface, and that the `AttributeDriver` has quite a few methods containing the word `Annotation`. We'll have to decide what to do about this. Also, as a follow-up, I will have to purge the yaml driver even more than it already is.


derrabus (approved):

derrabus:
The `Annotation` interface is a marker interface. We should be able to rename it easily.

derrabus:
see #10178

PR #10178, opened 2022-10-26
Deprecate the `Annotation` interface

derrabus:
Once we remove the support for annotations, having a marker interface named `Annotation` for our attribute classes seems weird. Let's rename it.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/MappingAttribute.php
+++ b/lib/Doctrine/ORM/Mapping/MappingAttribute.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping;
+
+/** A marker interface for mapping attributes. */
+interface MappingAttribute extends Annotation
+{
+}
--- a/psalm.xml
+++ b/psalm.xml
@@ -39,6 +39,8 @@
                 <referencedClass name="Doctrine\ORM\Tools\Console\Command\GenerateRepositoriesCommand"/>
                 <referencedClass name="Doctrine\ORM\Tools\Console\Helper\EntityManagerHelper"/>
                 <referencedClass name="Doctrine\ORM\Tools\Console\EntityManagerProvider\HelperSetManagerProvider"/>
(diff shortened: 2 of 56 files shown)


psalm.xml
```diff
@@ -39,6 +39,8 @@
                 <referencedClass name="Doctrine\ORM\Tools\Console\Command\GenerateRepositoriesCommand"/>
                 <referencedClass name="Doctrine\ORM\Tools\Console\Helper\EntityManagerHelper"/>
                 <referencedClass name="Doctrine\ORM\Tools\Console\EntityManagerProvider\HelperSetManagerProvider"/>
+                <!-- https://github.com/vimeo/psalm/issues/8617 -->
+                <referencedClass name="Doctrine\ORM\Mapping\Annotation"/>
```
derrabus, line 43: The interface is reported as a class when used in generics. I've created vimeo/psalm#8617 to track this issue.

lib/Doctrine/ORM/Mapping/MappingAttribute.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping;
+
+/** A marker interface for mapping attributes. */
+interface MappingAttribute extends Annotation
```
greg0ire, line 8: Why did you name it `MappingAttribute` and not just `Attribute`? It is already inside a `Mapping` namespace.

derrabus: Because literally all classes that implement this interface also have an attribute attached to them named `Attribute` and I didn't want to screw up the imports everywhere. I also wanted the name of the marker interface to be a bit more specific. After all, it's not just an arbitrary attribute, it's one that we use to configure mapping information.

greg0ire (approved):

lib/Doctrine/ORM/Mapping/MappingAttribute.php
greg0ire: Makes sense. Thanks!

PR #10187, opened 2022-10-28
Leverage LazyGhostTrait when possible

nicolas-grekas:
This PR replaces Doctrine Common's proxies by [VarExporter's `LazyGhostTrait`](https://symfony.com/blog/revisiting-lazy-loading-proxies-in-php).

It requires running `composer require symfony/var-exporter:^6.2@dev doctrine/persistence:^3.1@dev` for now until stable versions of both packages are released (will happen in one month for var-exporter.)

As you can see from the changes on the documentation, the benefit is quite clear: this removes many limitations of the current proxy implementation.

From what I've understood through many interactions around `doctrine/common`, deprecating everything in the `Doctrine\Common\Proxy` namespace is highly desired and this PR provides a critical step in this direction. Thanks to the design design + maintenance policies of `LazyGhostTrait`, this should also make it way simpler to support newer versions of PHP in the future (this is already the case for PHP 8.1 which is fully supported by the trait.).

The target version on ORM is 2.14. This should help move away from Common's proxies quite quickly, and will allow deprecating them ASAP. In turn, Common's Proxies wouldn't need to add support for newer versions of PHP (which is non trivial, as highlighted by the work still happening to support PHP 8.1.), making maintenance easier for the Doctrine project. Targeting ORM v3 wouldn't allow this.

Another reason to target ORM v2 is that the change is mostly backward compatible, so it doesn't need to target v3. In my experience, this can help adoption of ORM v3 when it will be released, by reducing the number of dependencies that one would have to update in the process of upgrading.

Still, enabling the new proxies is opt-in for a few reasons:
- ORM v2 supports PHP >= 7.1 while symfony/var-exporter requires PHP >= 8.1, so we cannot make it a hard dep yet;
- generated proxies do not implement the (deprecated) `Doctrine\ORM\Proxy\Proxy` interface anymore so ppl need to replace their `instanceof` checks to target `Doctrine\Persistence\Proxy` instead (which is implemented by both new and old proxy implems.)
- `__wakeup()` is not called anymore when initializing entities that implement it (this should be transparent if ppl implemented it as advised in the doc - aka not doing anything when the identifier is not yet known.)

One needs to call `Configuration::setLazyGhostObjectEnabled(true)` to turn on the new proxies.

For cross-ref, this takes some inspiration+changes from #6719.

As I see for now, the steps after merging this could be to:
1. deprecate everything in `Doctrine\Common\Proxy`
2. deprecate NOT enabling the new proxies.
3. make `Configuration::setLazyGhostObjectEnabled()` a no-op in ORM v3 (throwing when called with `false` as argument)
4. deprecate `Configuration::setLazyGhostObjectEnabled()` but not before ORM v3.1 to ease the transition (we cannot remove the method in v3.0 since we cannot first deprecate the method in v2).

For inspiration for other projects, the plan on Symfony should be to add a new config option to DoctrineBundle to allow opting-in, to deprecate not setting it to `true`, and to update the default recipe to make new projects opt-in by default.


lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -180,4 +291,76 @@ private function createCloner(ClassMetadata $classMetadata, EntityPersister $ent
...
+        return $code;
+    }
+
+    private function generateSerializeImpl(ClassMetadata $class): string
+    {
+        $reflector  = $class->getReflectionClass();
+        $properties = $reflector->hasMethod('__serialize') ? 'parent::__serialize()' : '(array) $this';
+
+        $code = '$properties = ' . $properties . ';
+        unset($properties["\0" . self::class . "\0lazyObjectId"]);
```
stof: shouldn't this also using `__isCloning` ?

stof: And shouldn't this `unset` be done only for the array cast and not for `parent::__serialize()` ?

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3691,4 +3685,43 @@ private function convertSingleFieldIdentifierToPHPValue(ClassMetadata $class, $i
...
     }
+
+    /**
+     * Given a flat identifier, this method will produce another flat identifier, but with all
+     * association fields that are mapped as identifiers replaced by entity references, recursively.
+     *
+     * @param mixed[] $flatIdentifier
+     *
+     * @return mixed[]
+     * @psalm-return array<string, mixed>
```
stof: This could be put as the normal `@return` directly. Tooling support for the `array<TKey, TValue>` notation is good.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
nicolas-grekas: updated, this won't change much in practice but that's still better to unset indeed.
About your second question: no, the property should never be serialized, even if the parent returns it.
FYI, it is legit to implement __serialize without implementing __unserialize. When doing to, __serialize shall return an array structure modeled after `(array) $this`. This is when the property can leak and should be unset.

lib/Doctrine/ORM/UnitOfWork.php
nicolas-grekas: works for me, updated

beberlei (dismissed):

beberlei:
Great work, ready to merge from my POV. There are some changes in UoW that are maybe a bit risky, but nothing stands out as buggy from my side. The normalizeIdentifier function is unfortunate to have, as it adds even more looping inside looping, but we can always look to improve that later.

derrabus:
Thank you very much! I haven't had time to test this yes, but the changes look promising. Since the PR does not deprecate the old proxy from the Common package (yet), we could ship this change as an experimental feature in 2.14. For 3.0 however, our goal should be to have only one proxy implementation.

beberlei:
Yes since 3.0 will bump requirement to php 8.1 we could go with the new implementation only.

nicolas-grekas:
This PR is now ready, using only stable dependencies! :rocket:

.github/workflows/static-analysis.yml
```diff
@@ -100,6 +99,9 @@ jobs:
           coverage: "none"
           php-version: "${{ matrix.php-version }}"
 
+      - name: "Require specific persistence version"
+        run: "composer require symfony/var-exporter ^6.2@dev doctrine/persistence ^3.1@dev --no-update"
```
greg0ire: > using only stable dependencies

I think you missed this workflow

.github/workflows/continuous-integration.yml
```diff
@@ -53,6 +55,10 @@ jobs:
           - php-version: "8.2"
             dbal-version: "default"
             extension: "sqlite3"
+          - php-version: "8.1"
+            dbal-version: "3@dev"
```
greg0ire: Why not use the `default` version here?

nicolas-grekas: updated

.github/workflows/static-analysis.yml
nicolas-grekas: right, updated

greg0ire (approved):

derrabus (approved):

greg0ire:
Thanks @nicolas-grekas !

PR #10188, opened 2022-10-28
Implement entity identifier in PostRemoveEventArgs

tgauges:
Implement setting and getting of entity identifier in PostRemoveEventArgs.

Fixes #2326

## TODO

- [ ] update documentation
- [ ] implement real test case checking if id is available
  - [ ] What happens with multiple identifier fields? I'm not sure because only the first identifer seems to be set to null anyway.

### DIFF
--- a/lib/Doctrine/ORM/Event/PostRemoveEventArgs.php
+++ b/lib/Doctrine/ORM/Event/PostRemoveEventArgs.php
@@ -4,6 +4,31 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Event;
 
+use Doctrine\ORM\EntityManagerInterface;
+
 final class PostRemoveEventArgs extends LifecycleEventArgs
 {
+    /**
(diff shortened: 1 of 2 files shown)


derrabus:
Can you have a look at the CI failures? Also, please pick a PR title that tells us what the PR does. The title will appear in the release notes eventually.

derrabus:
I'm converting to draft because the PR has open TODOs and is lacking tests.

tgauges:
Before continuing I'd like to answer the question why only the first identifer seems to be set to null in `lib/Doctrine/ORM/UnitOfWork.php:1267`. Is this correct?

If yes: I'd probably rename `PostRemoveEventArgs::getIdentifier` to `PostRemoveEventArgs::getFirstIdentifier`.

If no: I'd create a new pull request to merge first which sets all identifiers to `null`.

felipyamorim:
any news ?

PR #10189, opened 2022-10-28
Added parameter to getSelectJoinColumnSQL

### LINKED ISSUE #9822: Collection is empty for one-to-many association using class table inheritance
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.12.2

#### Summary

The `Collection` of a `One-To-Many Association` is empty when using a junction entity having `Class Table Inheritance` to connect 2 entities.

I have 2 entities (`Field` & `Mapping`) which I want to "connect" using junction entities which inherit of a `Assignment` entity which uses a composite key to connect them both.
Something like this:
- FieldA <-> Assignment_FieldA <-> MappingA
- FieldA <-> Assignment_FieldB <-> MappingA
- FieldA <-> Assignment_FieldC <-> MappingA

I'm using a junction entity because I want to store additional data in the assignment/connection.

Also, I want to retrieve all the mappings a field is assigned/connected to, and also in the other direction retrieve all the fields a mapping has connected/assigned.
Therefore, I use a `One-To-Many Association` inside the `Field` & `Mapping` entity which should return a list of `Assignment*` entities but when calling, the `Collection` is empty.

I checked all association's and the stored database rows, also the fetching query, and executed them manually and all are correct.

#### How to reproduce

``` PHP
#[Entity, Table("Fields")]
class Field {

	#[Id]
	#[Column, GeneratedValue]
	public int $ID;

	#[OneToMany(targetEntity: Assignment::class, mappedBy: "Field", cascade: ["persist"], orphanRemoval: true)]
	public Collection $AssignedMappings;

	public function __construct() {
		$this->AssignedMappings = new ArrayCollection();
	}

}

#[Entity, Table("Mappings")]
class Mapping {

	#[Id]
	#[Column, GeneratedValue]
	public int $ID;

	#[OneToMany(targetEntity: Assignment::class, mappedBy: "Mapping", cascade: ["persist"], orphanRemoval: true)]
	public Collection $AssignedFields;

	public function __construct() {
		$this->AssignedFields = new ArrayCollection();
	}

}

#[Entity, Table("Assignments")]
#[
	InheritanceType("JOINED"),
	DiscriminatorColumn("Type"),
	DiscriminatorMap([
		"A" => Assignment_FieldA::class,
		"B" => Assignment_FieldB::class,
	])
]
abstract class Assignment {

	#[Id]
	#[ManyToOne(inversedBy: "AssignedFields")]
	public Mapping $Mapping;

	#[Id]
	#[ManyToOne(inversedBy: "AssignedMappings")]
	public Field $Field;

	public function __construct(Mapping $mapping, Field $field) {
		$this->Mapping = $mapping;
		$this->Field = $field;
	}

}

#[Entity, Table("Assignments_FieldA")]
class Assignment_FieldA extends Assignment {}

#[Entity, Table("Assignments_FieldB")]
class Assignment_FieldB extends Assignment {}

// PHPUnit Test
public function testAssignments() {
	// Create
	$fieldA = new Field();
	$fieldB = new Field();
	$mapping = new Mapping();

	$this->em->persist($fieldA);
	$this->em->persist($fieldB);
	$this->em->persist($mapping);
	$this->em->flush();

	$mapping->AssignedFields->add(new Assignment_FieldA($mapping, $fieldA));
	$mapping->AssignedFields->add(new Assignment_FieldB($mapping, $fieldB));

	$this->em->persist($mapping);
	$this->em->flush();
	$this->em->clear();

	// Fetch
	$fieldA = $this->em->find(Field::class, 1);
	$fieldB = $this->em->find(Field::class, 2);
	$mapping = $this->em->find(Mapping::class, 1);

	// assertCount will internally use `count()` to retrieve the count of the `Collection` which will also initialize the `Collection`.
	$this->assertCount(1, $fieldA->AssignedMappings);
	$this->assertCount(1, $fieldB->AssignedMappings);
	$this->assertCount(2, $mapping->AssignedFields);
}
```

#### My investigation and possible bug discovery/solution

I tried to investigate the problem myself and I noticed the query is correct, and the row data is received in the `ObjectHydrator` but the identifier is not detected, and the row is therefore discarded.

I noticed in `ObjectHydrator.php:467` that the row got discared because the `$nonemptyComponents` was empty.

I also noticed when the `Assignment` is a single entity (without inheritance) everything works as expected and `$nonemptyComponents` was not empty and also the `Collection` had all entities correctly.

Therefore, I checked how the `$nonemptyComponents` gets filled, which happens in `AbstractHydrator.php:477` if the identifier for the row data is found and is not `null`. For the single entity that worked, but not for my inherited `Assignment*` entity.

Therefore, I checked how the identifier is detected and this happens in the select column query-building part where the column is checked if it is an id, which is different for single and inherited entities.

The single entity will use the `BasicEntityPersister` and at line `1348` if the column is an id, the `ResultSetMapping` will get informed by that, which will result in the `$nonemptyComponents` not being empty, and therefore the `Collection` not empty.

The inherited entity will use the `JoinedSubclassPersister` which inherits the `AbstractEntityInheritancePersister` class and at line `77` the column will be selected but the `isIdentifier` parameter which informs the `ResultSetMapping` for an found id column is hardcoded to `false`, which will result in an always empty `$nonemptyComponents` and therefore an always empty `Collection`.
I found that very strange, why is that hardcoded to `false`?

I confirmed that by adding the same id detection code for the `JoinedSubclassPersister` as it is in the `BaseEntityPersister` and the `Collection` was indeed correctly filled with desired entities.
But this is no permanent solution because I modified the vendor files.
I almost wanted to do an pull request with this fix, but because I don't fully understand the inner workings of doctrine, I was not sure if that is correct what I did. Because there must be a reason it is hardcoded to `false` right? Maybe the id column is detected somewhere else for joined rows?

NordFox7:
The Collection of a One-To-Many Association is empty when using a junction entity having Class Table Inheritance to connect 2 entities.
The problem arises because the Collection will only be hydrated if the association mappings have an identifier, but the identifier is not detected because $isIdentifierColumn is hardcoded to false.
Fixes #9822

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/AbstractEntityInheritancePersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/AbstractEntityInheritancePersister.php
@@ -66,15 +66,16 @@ abstract class AbstractEntityInheritancePersister extends BasicEntityPersister
      * @param string $tableAlias
      * @param string $joinColumnName
      * @param string $quotedColumnName
+     * @param bool   $isIdentifierColumn
      * @param string $type
      *
      * @return string
      */
-    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $quotedColumnName, $type)
+    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $quotedColumnName, $isIdentifierColumn, $type)
     {
         $columnAlias = $this->getSQLColumnAlias($joinColumnName);
 
-        $this->currentPersisterContext->rsm->addMetaResult('r', $columnAlias, $joinColumnName, false, $type);
+        $this->currentPersisterContext->rsm->addMetaResult('r', $columnAlias, $joinColumnName, $isIdentifierColumn, $type);
 
         return $tableAlias . '.' . $quotedColumnName . ' AS ' . $columnAlias;
     }
--- a/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
@@ -436,13 +436,15 @@ class JoinedSubclassPersister extends AbstractEntityInheritancePersister
                 ? $this->getSQLTableAlias($mapping['inherited'])
                 : $baseTableAlias;
 
-            $targetClass = $this->em->getClassMetadata($mapping['targetEntity']);
+            $targetClass  = $this->em->getClassMetadata($mapping['targetEntity']);
+            $isIdentifier = isset($mapping['id']) && $mapping['id'] === true;
 
             foreach ($mapping['joinColumns'] as $joinColumn) {
                 $columnList[] = $this->getSelectJoinColumnSQL(
                     $tableAlias,
                     $joinColumn['name'],
                     $this->quoteStrategy->getJoinColumnName($joinColumn, $this->class, $this->platform),
+                    $isIdentifier,
                     PersisterHelper::getTypeOfColumn($joinColumn['referencedColumnName'], $targetClass, $this->em)
                 );
             }
@@ -479,13 +481,15 @@ class JoinedSubclassPersister extends AbstractEntityInheritancePersister
                     continue;
                 }
 
-                $targetClass = $this->em->getClassMetadata($mapping['targetEntity']);
+                $targetClass  = $this->em->getClassMetadata($mapping['targetEntity']);
+                $isIdentifier = isset($mapping['id']) && $mapping['id'] === true;
 
                 foreach ($mapping['joinColumns'] as $joinColumn) {
                     $columnList[] = $this->getSelectJoinColumnSQL(
                         $tableAlias,
                         $joinColumn['name'],
                         $this->quoteStrategy->getJoinColumnName($joinColumn, $subClass, $this->platform),
+                        $isIdentifier,
                         PersisterHelper::getTypeOfColumn($joinColumn['referencedColumnName'], $targetClass, $this->em)
                     );
                 }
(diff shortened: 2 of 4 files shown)


lib/Doctrine/ORM/Persisters/Entity/AbstractEntityInheritancePersister.php
```diff
@@ -66,15 +66,16 @@ protected function getSelectColumnSQL($field, ClassMetadata $class, $alias = 'r'
      * @param string $tableAlias
      * @param string $joinColumnName
      * @param string $quotedColumnName
+     * @param bool   $isIdentifierColumn
      * @param string $type
      *
      * @return string
      */
-    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $quotedColumnName, $type)
+    protected function getSelectJoinColumnSQL($tableAlias, $joinColumnName, $quotedColumnName, $isIdentifierColumn, $type)
```
derrabus, line 74: This change breaks custom implementations of `AbstractEntityInheritancePersister`. We cannot do this in a bugfix or minor release.

NordFox7: I see. It seemed like the obvious change for me. I'm not fully custom with the internal workings of doctrine. Maybe there is a non-breaking solution, which I'm not aware of. As I also stated in the Issue post, that I'm not sure if my change is the correct way of tackling this problem.

derrabus: I've checked back with the core team. The API you're about to change is internal to the ORM, so changing it is fine. Sorry for the confusion.

PR #10192, opened 2022-10-31
Remove the Annotation interface

derrabus:


greg0ire (approved):
Are there things that should be ported to lower branches?

derrabus:
I don't think so.

PR #10199, opened 2022-11-03
refactor: use list type in SchemaTool

simPod:
phpstan 1.9.0 has some additional support for lists. It discovered some new complains on my code because it is not typed as list here.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -110,7 +110,7 @@ class SchemaTool
      *
      * @psalm-param list<ClassMetadata> $classes
      *
-     * @return string[] The SQL statements needed to create the schema for the classes.
+     * @return list<string> The SQL statements needed to create the schema for the classes.
      */
     public function getCreateSchemaSql(array $classes)
     {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Tools/SchemaTool.php
```diff
@@ -947,11 +947,11 @@ public function updateSchema(array $classes, $saveMode = false)
      * Gets the sequence of SQL statements that need to be performed in order
      * to bring the given class mappings in-synch with the relational schema.
      *
-     * @param mixed[] $classes  The classes to consider.
-     * @param bool    $saveMode If TRUE, only generates SQL for a partial update
-     *                           that does not include SQL for dropping assets which are scheduled for deletion.
+     * @param bool $saveMode If TRUE, only generates SQL for a partial update
+     *                       that does not include SQL for dropping assets which are scheduled for deletion.
+     * @psalm-param list<ClassMetadata> $classes  The classes to consider.
```
derrabus:
```suggestion
     * @param list<ClassMetadata> $classes  The classes to consider.
```

No need to vendor-prefix this annotation.

derrabus (approved):

Issue #8151, opened 2020-05-19
[PHPDoc] Parameter #4 of method Doctrine\ORM\QueryBuilder::innerJoin() expects st ring|null, Doctrine\ORM\Query\Expr\Func given.


soullivaneuh:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | v2.7.2

#### Summary

The [`QueryBuilder::innerJoin`](https://github.com/doctrine/orm/blob/9a8e1fb47776fc5ff49f995f3fef93a5ff8469b0/lib/Doctrine/ORM/QueryBuilder.php#L963) does not expect Doctrine\ORM\Query\Expr\Func according to the PHPDoc.

This typing should be added a stronger typing and to make tools like PHPStan happier.

#### Current behavior

PHPStan error:

```
Parameter #4 $condition of method Doctrine\ORM\QueryBuilder::innerJoin() expects st
ring|null, Doctrine\ORM\Query\Expr\Func given.
```

#### How to reproduce

Related code producing the error (currently working properly):

```php
$qb = $this->getRepository(PowerDNSRecord::class)->createQueryBuilder('r');
$qb->innerJoin('r.domain', 'd', Expr\Join::WITH, $qb->expr()->in('d.name', ':domains'))
    ->select('d.name, r.content, r.type')
    ->where($qb->expr()->in('r.type', ':types'))
    ->andWhere('r.name = d.name')
    ->setParameters(['domains' => $hostings, 'types' => ['A', 'AAAA']])
;
```

#### Expected behavior

Everything should be ok as the method works with a func.

najmi9:
The same problem, any updates?

PR #10202, opened 2022-11-04
Allow "Expr\Func" as condition in join in PHPDoc

willemverspyck:
I use "Expr\Func" as 4th parameter of innerJoin, but according to the PHPDoc it's not allowed:

Solves #8151

Current parameter is https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/QueryBuilder.php#L1025

For example when I use the "in" expression in the join condition:

```mysql
$qb = $this->entityManager->createQueryBuilder();
$qb
    ->select('u', 'a')
    ->from(CmsUser::class, 'u')
    ->innerJoin('u.articles', 'a', Join::WITH, $qb->expr()->in(
        'u.id',
        [1, 2, 3]
    ));
```

it generates the query .... WITH u.id IN(1, 2, 3), but currently it's not allowed according to the PHPDoc.

### DIFF
--- a/lib/Doctrine/ORM/Query/Expr/Join.php
+++ b/lib/Doctrine/ORM/Query/Expr/Join.php
@@ -37,19 +37,19 @@ class Join
      */
     protected $conditionType;
 
-    /** @var string|Comparison|Composite|null */
+    /** @var string|Comparison|Composite|Func|null */
     protected $condition;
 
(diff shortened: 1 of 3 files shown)


derrabus (approved):

derrabus:
Thank you!

PR #10203, opened 2022-11-05
Purge annotations from /lib

greg0ire:
There are still mentions of annotations in the attribute driver and the attribute reader, as method names. I suppose we'll have to deprecate those.


lib/Doctrine/ORM/Mapping/AssociationOverride.php
```diff
@@ -4,13 +4,7 @@
...
 
-/**
- * This annotation is used to override association mapping of property for an entity relationship.
- *
- * @Annotation
- * @NamedArgumentConstructor
- * @Target("ANNOTATION")
- */
+/** This attribute is used to override association mapping of property for an entity relationship. */
 final class AssociationOverride implements MappingAttribute
```
derrabus, line 8: This class does not have an `#[Attribute]` attribute? Is it still needed?

lib/Doctrine/ORM/Mapping/AssociationOverrides.php
```diff
@@ -9,13 +9,7 @@
...
 use function is_array;
 
-/**
- * This annotation is used to override association mappings of relationship properties.
- *
- * @Annotation
- * @NamedArgumentConstructor()
- * @Target("CLASS")
- */
+/** This attribute is used to override association mappings of relationship properties. */
```
derrabus, line 12: You've rephrased the documentation on each attribute class to use the term "attribute" instead of "annotation". Would it make sense to do this on the 2.14.x branch already?

lib/Doctrine/ORM/Mapping/AttributeOverride.php
```diff
@@ -4,13 +4,7 @@
...
 
-/**
- * This annotation is used to override the mapping of a entity property.
- *
- * @Annotation
- * @NamedArgumentConstructor
- * @Target("ANNOTATION")
- */
+/** This attribute is used to override the mapping of a entity property. */
 final class AttributeOverride implements MappingAttribute
```
derrabus, line 8: This isn't an attribute (yet?) either.

lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -762,20 +762,15 @@ public static function duplicateEntityListener($listenerName, $methodName, $clas
...
 
-    /**
-     * @param string $className
-     * @param string $annotation
-     *
-     * @return MappingException
-     */
-    public static function invalidFetchMode($className, $annotation)
+    /** @param class-string $className */
+    public static function invalidFetchMode(string $className, string $fetchMode): self
```
derrabus: Backport the parameter name change to 2.14?

lib/Doctrine/ORM/Mapping/AttributeOverride.php
SenseException: Just like `JoinColumns`.

lib/Doctrine/ORM/Mapping/AssociationOverride.php
greg0ire: Removing it breaks the test suite. I think it's because it's only supposed to ever be used nested in an attribute, and not standalone.

derrabus: Ah, that makes sense, thanks.

derrabus (approved):

PR #10204, opened 2022-11-05
Rename internal methods

greg0ire:


lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
```diff
@@ -30,15 +31,24 @@ class AttributeDriver extends CompatibilityAnnotationDriver
 {
     use ColocatedMappingDriver;
 
-    /** @var array<string,int> */
-    // @phpcs:ignore
+    /** @var array<class-string<MappingAttribute>, int> */
+    protected $entityAttributeClasses = [
```
greg0ire: Let me know if you think this should be a private property, I don't know why `$entityAnnotationClasses` is exposed or why the class isn't final.

SenseException: Probably from the dark ages of PHP.

SenseException (dismissed):

SenseException:
I'm not sure though if this does break something when someone extended it.

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
derrabus: `$entityAnnotationClasses` is an extension point of the `AnnotationDriver` from Persistence 2. However, we already override all functionality that uses the property.

Theoretically, you could invent your own attributes and extend `AttributeDriver` to make use of them. Either way, a protected property is a bad extension point. The property is used in the `isTransient()` method only.

My suggestion: Move the array into a private constant, deprecate the property and tell people to override `isTransient()` instead of overriding or changing the property. For the time being, our implementation of `isTransient()` has to use the property, but we can switch it to the new constant on 3.0.x.

derrabus (changes requested):

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
```diff
@@ -30,15 +31,23 @@ class AttributeDriver extends CompatibilityAnnotationDriver
...
+
+    /**
+     * @deprecated override isTransient() instead of overriding it
+     *
+     * @var array<class-string<MappingAttribute>, int>
+     */
     protected $entityAnnotationClasses = [
         Mapping\Entity::class => 1,
         Mapping\MappedSuperclass::class => 2,
     ];
```
derrabus:
```suggestion
    protected $entityAnnotationClasses = self::ENTITY_ATTRIBUTE_CLASSES;
```

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
```diff
@@ -30,15 +31,23 @@ class AttributeDriver extends CompatibilityAnnotationDriver
...
 
-    /** @var array<string,int> */
-    // @phpcs:ignore
+    private const ENTITY_ATTRIBUTE_CLASSES = [
+        Mapping\Entity::class => 1,
+        Mapping\MappedSuperclass::class => 2,
+    ];
+
+    /**
+     * @deprecated override isTransient() instead of overriding it
```
derrabus:
```suggestion
     * @deprecated override isTransient() instead of overriding this property
```

derrabus (dismissed):
Minor nit-picking, feel free to merge after my comments have been addressed.

PR #10210, opened 2022-11-08
Migrate Cache namespace to php 8

greg0ire:
This one comes with a controversial change that I did not have the courage to extract to a separate commit PR/commit (feel free to insist I do so): switching `@readonly` properties to native `readonly`. I had to tweak some tests to achieve that, and I've extracted that part in a separate commit, which I can backport if you think it should happen.

### DIFF
--- a/lib/Doctrine/ORM/Cache/AssociationCacheEntry.php
+++ b/lib/Doctrine/ORM/Cache/AssociationCacheEntry.php
@@ -4,35 +4,16 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Cache;
 
-/**
- * Association cache entry
- */
 class AssociationCacheEntry implements CacheEntry
 {
     /**
-     * The entity identifier
-     *
-     * @readonly Public only for performance reasons, it should be considered immutable.
-     * @var array<string, mixed>
-     */
-    public array $identifier;
-
-    /**
-     * The entity class name
-     *
-     * @readonly Public only for performance reasons, it should be considered immutable.
-     * @psalm-var class-string
-     */
-    public string $class;
-
-    /**
      * @param array<string, mixed> $identifier The entity identifier.
-     * @psalm-param class-string $class
+     * @param class-string         $class      The entity class name
      */
-    public function __construct(string $class, array $identifier)
-    {
-        $this->class      = $class;
-        $this->identifier = $identifier;
+    public function __construct(
+        public readonly string $class,
+        public readonly array $identifier,
+    ) {
     }
 
     /**
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -15,6 +15,7 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\PersistentCollection;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\Query\ResultSetMapping;
+use Doctrine\ORM\Query\SqlWalker;
 use Doctrine\ORM\UnitOfWork;
 use Doctrine\Persistence\Proxy;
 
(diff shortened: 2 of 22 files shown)


psalm-baseline.xml
```diff
@@ -43,6 +43,9 @@
     </NullableReturnStatement>
   </file>
   <file src="lib/Doctrine/ORM/Cache/DefaultEntityHydrator.php">
+    <InvalidScalarArgument occurrences="1">
+      <code>$data</code>
+    </InvalidScalarArgument>
```
greg0ire, line 48: This doesn't seem to be an issue with Psalm v5: https://psalm.dev/r/6cd352a83c

lib/Doctrine/ORM/Cache/Region/DefaultRegion.php
```diff
@@ -58,7 +58,7 @@ public function get(CacheKey $key): CacheEntry|null
     public function getMultiple(CollectionCacheEntry $collection): array|null
     {
         $keys = array_map(
-            Closure::fromCallable([$this, 'getCacheEntryKey']),
+            Closure::fromCallable($this->getCacheEntryKey(...)),
```
derrabus:
```suggestion
            $this->getCacheEntryKey(...),
```

derrabus (commented):
> This one comes with a controversial change […]: switching `@readonly` properties to native `readonly`.

That was then plan after all. Good job. 🙂 

> I had to tweak some tests to achieve that, and I've extracted that part in a separate commit, which I can backport if you think it should happen.

I think that would be a good idea.

I think, `Doctrine\ORM\Cache\CacheKey::$hash` can also be declared as `readonly`. Can you have a look?

lib/Doctrine/ORM/Cache/DefaultQueryCache.php
```diff
@@ -139,9 +138,7 @@ public function get(QueryCacheKey $key, ResultSetMapping $rsm, array $hints = []
                     continue;
                 }
 
-                $generateKeys = static function ($id) use ($assocMetadata): EntityCacheKey {
-                    return new EntityCacheKey($assocMetadata->rootEntityName, $id);
-                };
+                $generateKeys = static fn ($id): EntityCacheKey => new EntityCacheKey($assocMetadata->rootEntityName, $id);
```
SenseException: The second argument of `EntityCacheKey` is an array.

```suggestion
                $generateKeys = static fn (array $id): EntityCacheKey => new EntityCacheKey($assocMetadata->rootEntityName, $id);
```

greg0ire:
> I think, Doctrine\ORM\Cache\CacheKey::$hash can also be declared as readonly. Can you have a look?

There are a few more I missed, probably got distracted during my `grep` :sweat_smile:

SenseException (changes requested):
> I think, Doctrine\ORM\Cache\CacheKey::$hash can also be declared as readonly. Can you have a look?

I haven't done a deeper look into it, but does it make sense to introduce a constructor for the abstract class?

greg0ire:
>  think, Doctrine\ORM\Cache\CacheKey::$hash can also be declared as readonly.

This causes ~300 errors and failures in the test suite, ~I don't think it's possible, and if it is,~ it should happen in another PR IMO.

greg0ire:
> I haven't done a deeper look into it, but does it make sense to introduce a constructor for the abstract class?

Ah maybe that's an easy way to fix it, yes :thinking:

greg0ire:
@SenseException that was it, thanks! I left out a few more instances of `@readonly` inside Parser.php, which is in another namespace, and looks a bit more involving.

derrabus:
> > think, Doctrine\ORM\Cache\CacheKey::$hash can also be declared as readonly.
> 
> This causes ~300 errors and failures in the test suite

Awesome. 🙈 Let's do this in a follow-up then.

greg0ire:
> Awesome. see_no_evil Let's do this in a follow-up then.

They all stemmed from the same 5 or so lines, after following @SenseException recommendation, they're gone :slightly_smiling_face:

lib/Doctrine/ORM/Cache/AssociationCacheEntry.php
```diff
@@ -9,30 +9,14 @@
  */
 class AssociationCacheEntry implements CacheEntry
```
derrabus, line 7: You've removed the DocBlock of `CollectionCacheEntry`, I think you can remove this one too.

SenseException (dismissed):

derrabus (dismissed):

derrabus (approved):

PR #10212, opened 2022-11-10
Add a constructor to `CacheKey`

derrabus:
Prepares #10210 on 2.14.x.

### DIFF
--- a/tests/Doctrine/Tests/Mocks/CacheKeyMock.php
+++ b/tests/Doctrine/Tests/Mocks/CacheKeyMock.php
@@ -13,9 +13,9 @@ use Doctrine\ORM\Cache\CacheKey;
  */
 class CacheKeyMock extends CacheKey
 {
-    /** @param string $hash The string hash that represend this cache key */
+    /** @param string $hash The string hash that represents this cache key */
     public function __construct(string $hash)
     {
-        $this->hash = $hash;
+        parent::__construct($hash);
     }
 }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,10 @@
 # Upgrade to 2.14
 
+## Deprecated constructing a `CacheKey` without `$hash`
+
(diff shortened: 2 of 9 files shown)


tests/Doctrine/Tests/Mocks/CacheKeyMock.php
```diff
@@ -13,9 +13,9 @@
  */
 class CacheKeyMock extends CacheKey
 {
-    /** @param string $hash The string hash that represend this cache key */
+    /** @param string $hash The string hash that represents this cache key */
     public function __construct(string $hash)
     {
-        $this->hash = $hash;
+        parent::__construct($hash);
     }
```
greg0ire, line 20: :thinking: Can't you remove the constructor entirely instead? Or maybe even the whole class is useless?

SenseException: Actually we would need a test with constructor and without, because despite the deprecation both ways need to work.

derrabus: > 🤔 Can't you remove the constructor entirely instead?

I can, but I was reluctant to do so because it's stricter than `CacheKey`'s.

> Or maybe even the whole class is useless?

`CacheKey` is abstract and I don't really want to change that. So if we dropeed `CacheKeyMock`, I would need to replace it with an anonymous class in every test that uses it. I'm not sure that's an improvement really.

derrabus: > Actually we would need a test with constructor and without, because despite the deprecation both ways need to work.

I've pushed two new tests.

greg0ire: Didn't notice the extra strictness 👍

greg0ire (approved):

PR #10219, opened 2022-11-11
Migrate AST namespace to PHP 8 syntax

greg0ire:


lib/Doctrine/ORM/Query/AST/AggregateExpression.php
```diff
@@ -8,21 +8,16 @@
 
 class AggregateExpression extends Node
 {
-    /**
-     * Some aggregate expressions support distinct, eg COUNT.
```
derrabus, line 12: We should keep this piece of documentation by moving it to the `@param` annotation.

lib/Doctrine/ORM/Query/AST/WhenClause.php
```diff
@@ -13,11 +13,8 @@
  */
 class WhenClause extends Node
 {
-    /**
-     * @param ConditionalExpression $caseConditionExpression
-     * @param mixed                 $thenScalarExpression
-     */
-    public function __construct(public $caseConditionExpression = null, public $thenScalarExpression = null)
+    /** @param ConditionalExpression $caseConditionExpression */
```
derrabus, line 16: Apparently, we allow this parameter to be `null`.

```suggestion
    /** @param ConditionalExpression|null $caseConditionExpression */
```

Should be backported to 2.13.x or 2.14.x probably.

derrabus (dismissed):

lib/Doctrine/ORM/Query/AST/WhenClause.php
greg0ire: it's incomplete. 2.13.x is for wrong stuff, 2.14.x for imprecise stuff. I think incomplete here is wrong: it doesn't have to always be a `ConditionalExpression`

derrabus: Looking at #10224: The `$caseConditionExpression` parameter was not optional on 2.x. Why is it here? Let's remove the default and document it as not nullable instead?

greg0ire: :thinking: indeed. I think that must be a choice rector makes between the doc of the property, and the doc of the parameter. I'll amend #10224 after confirming it's safe to do so

greg0ire: Using `object` as a type hint passes the test suite (and `ConditionalExpression` does not, of course lol)

greg0ire:
Blocked by https://github.com/doctrine/orm/pull/10224

lib/Doctrine/ORM/Query/AST/BetweenExpression.php
```diff
@@ -8,28 +8,18 @@
...
-
-    /** @var ArithmeticExpression */
-    public $leftBetweenExpression;
-
-    /** @var ArithmeticExpression */
-    public $rightBetweenExpression;
-
-    /** @var bool */
-    public $not;
+    public bool $not;
```
derrabus:
```suggestion
    public bool $not = false;
```

We should not allow this property to be uninitialized. In fact, I'd suggest to add it as promoted property to the constructor. `Parser::BetweenExpression()` currently assigns it right after construction anyway.

See #10267.

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -1007,7 +1007,7 @@ public function AliasIdentificationVariable()
     /**
      * AbstractSchemaName ::= fully_qualified_name | identifier
      *
-     * @return string
+     * @return class-string
```
derrabus: The implementation looks like we don't always get a FQCN here. I'd revert this change.

derrabus (changes requested):
Rebase needed.

greg0ire:
Converting to draft until your changes on this namespace are merged up :slightly_smiling_face:

derrabus (approved):

PR #10224, opened 2022-11-11
Document property as non-nullable

greg0ire:
The constructor forbids it.


greg0ire:
@derrabus I retargeted to 2.14.x because the `|null` part is technically true.

SenseException (approved):
To what `|null` are you referring to?

greg0ire:
@SenseException it was something added by Rector on https://github.com/doctrine/orm/pull/10219/files#r1020406559 to the type based on this `= null` I just removed. I was persuaded it was already on lower branches but as it turns out, only the `= null` was there :sweat_smile: Anyway, the PR is still valid 👍

PR #10225, opened 2022-11-11
Remove deprecated event classes and functions

derrabus:
Attempting to clean up the deprecations inside the `Event` namespace. WIP because we need some additional work on 2.14.x.


greg0ire:
@derrabus I think this has been handled in 84e78b8200ed39c6a820a5f6f78f07defa947659. Is there anything left?

PR #10227, opened 2022-11-12
Migrate Expr and Lexer to PHP 8 syntax

greg0ire:


lib/Doctrine/ORM/Query/Expr/Comparison.php
```diff
@@ -4,51 +4,47 @@
...
 
     /**
      * Creates a comparison expression with the given arguments.
      *
-     * @param mixed  $leftExpr
      * @param string $operator
-     * @param mixed  $rightExpr
      */
-    public function __construct(protected $leftExpr, protected $operator, protected $rightExpr)
+    public function __construct(protected mixed $leftExpr, protected $operator, protected mixed $rightExpr)
```
SenseException: If `$operator` is string, why no type hint here?

lib/Doctrine/ORM/Query/Expr/Composite.php
```diff
@@ -31,8 +30,7 @@ public function __toString()
         return implode($this->separator, $components);
     }
 
-    /** @param string|object $part */
-    private function processQueryPart($part): string
+    private function processQueryPart(string|object $part): string
```
SenseException: I wonder if `string|Stringable` would be more fitting here.

lib/Doctrine/ORM/Query/Expr/From.php
```diff
@@ -4,42 +4,44 @@
...
  *
  * @link    www.doctrine-project.org
  */
-class From
+class From implements Stringable
 {
     /**
      * @param class-string $from    The class name.
      * @param string       $alias   The alias of the class.
      * @param string       $indexBy The index for the from.
```
SenseException: Can be removed now.

lib/Doctrine/ORM/Query/Expr/Select.php
```diff
@@ -11,20 +11,17 @@
...
-    protected $preSeparator = '';
-
-    /** @var string */
-    protected $postSeparator = '';
+    protected string $preSeparator  = '';
+    protected string $postSeparator = '';
 
     /** @var string[] */
-    protected $allowedClasses = [Func::class];
+    protected array $allowedClasses = [Func::class];
```
SenseException, line 18: I don't know if that belongs into the scope (or if it fits here): `class-string[]` would be a nice addition.

SenseException (changes requested):

lib/Doctrine/ORM/Query/Expr/Select.php
greg0ire: That's something that should be done on 2.14.x

greg0ire: And IIRC, it's a rabbit hole if you want to specify something inside the `<>` of `class-string<>` :stuck_out_tongue:

lib/Doctrine/ORM/Query/Expr/Comparison.php
greg0ire: Not sure why I left that one out, let's use a type hint.

lib/Doctrine/ORM/Query/Expr/Select.php
SenseException: Let's revisit this in 2.14 again.

SenseException (approved):

PR #10229, opened 2022-11-13
Widen parameter type

greg0ire:
This exception is used in two places, one where $generatedMode is clearly a string, and another where typing it as a string will cause a type error, because $generatedMode is supposed to be an int there.

Supposed to be an int: https://github.com/doctrine/orm/blob/465c02fe68d683efd32a314417940629c54374c4/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L1720-L1728

Supposed to be a string: https://github.com/doctrine/orm/blob/465c02fe68d683efd32a314417940629c54374c4/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L680-L682

@mehldau this was introduced in https://github.com/doctrine/orm/pull/9118, please review

### DIFF
--- a/lib/Doctrine/ORM/Mapping/MappingException.php
+++ b/lib/Doctrine/ORM/Mapping/MappingException.php
@@ -824,7 +824,8 @@ class MappingException extends ORMException
         return new self("Entity '" . $className . "' has a mapping with invalid fetch mode '" . $annotation . "'");
     }
 
-    public static function invalidGeneratedMode(string $annotation): MappingException
+    /** @param int|string $annotation */
+    public static function invalidGeneratedMode($annotation): MappingException
     {
         return new self("Invalid generated mode '" . $annotation . "'");
     }


SenseException:
https://github.com/doctrine/orm/blob/465c02fe68d683efd32a314417940629c54374c4/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L1656

Isn't this supposed to be `array<string, FieldMapping>` or did I miss something? Maybe that's why the `int` could be removed without PHPStan complaining in the PR?

greg0ire:
There is [this comment](https://github.com/doctrine/orm/pull/9198#discussion_r754619633) that suggests `FieldMapping` should be used more widely, so maybe yeah

EDIT: also, you mean `FieldMapping`, not `array<string, FieldMapping>` :wink:

greg0ire:
@SenseException I tried your suggestion, and no, that's not right, because then you get 

```
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  1669   Offset 'fieldName' on array{type: string, fieldName: string, columnName: string, length?: int, id?: bool, nullable?: bool, notInsertable?: bool, notUpdatable?: bool, ...} in isset() always exists and is not nullable.
```

At that point, the mapping is supposed to be incomplete.

SenseException (approved):
Then let's keep it the way it is.

PR #10231, opened 2022-11-14
Make the code easier to statically analyse

greg0ire:
While working on migrating the codebase to PHP 8, I noticed that a rule (`NullToStrictStringFuncCallArgRector`) often triggered that consists in adding casts to string, because internal functions have become more strict in PHP 8.1 (they no longer accept anything than a string).

I tried finding why those arguments were not considered already strings, and tried to fix some of that with this PR.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1574,9 +1574,9 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * Validates & completes the given field mapping based on typed property.
      *
-     * @param  mixed[] $mapping The field mapping to validate & complete.
+     * @param  array{fieldName: string, type?: mixed} $mapping The field mapping to validate & complete.
      *
-     * @return mixed[] The updated mapping.
+     * @return array{fieldName: string, enumType?: string, type?: mixed} The updated mapping.
      */
     private function validateAndCompleteTypedFieldMapping(array $mapping): array
     {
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1631,7 +1631,7 @@ private function validateAndCompleteTypedFieldMapping(array $mapping): array
     /**
      * Validates & completes the basic mapping information based on typed property.
      *
-     * @param mixed[] $mapping The mapping.
+     * @param array{type: self::ONE_TO_ONE|self::MANY_TO_ONE|self::ONE_TO_MANY|self::MANY_TO_MANY, fieldName: string, targetEntity?: class-string} $mapping The mapping.
```
greg0ire, line 1634: I couldn't find a way to make this multiline and have phpcs be happy with it… it's probably the pipes.

lib/Doctrine/ORM/Query/Expr.php
```diff
@@ -620,13 +620,15 @@ public function literal($literal)
     /**
      * Quotes a literal value, if necessary, according to the DQL syntax.
      *
-     * @param mixed $literal The literal value.
+     * @param scalar $literal The literal value.
      */
     private function quoteLiteral($literal): string
     {
-        if (is_numeric($literal) && ! is_string($literal)) {
+        if (is_int($literal) || is_float($literal)) {
```
greg0ire, line 627: About this, see https://github.com/vimeo/psalm/issues/8703

SenseException (approved):

PR #10234, opened 2022-11-15
Support `NOT` expression

remi-san:
This PR aims at providing support for the `NOT` expression which might to be a part of [`doctrine/collections`](https://github.com/doctrine/collections) in the future.

~It is not mergeable as it references a branch that has not been merged yet into `doctrine/collections` (see [PR #348 on `doctrine/collections`](https://github.com/doctrine/collections/pull/348)) but will be updated as soon as it is.~ `doctrine/collections` `2.1.0` having been released, this PR is now complete.

Given the fact that the `not` expression is already supported by [`Doctrine\ORM\Query\Expr`](https://github.com/doctrine/orm/blob/827cb0c10bd0440b5e1b8ec7f5dc1fc1810f1821/lib/Doctrine/ORM/Query/Expr.php#L323), it is pretty straightforward.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/SqlExpressionVisitor.php
+++ b/lib/Doctrine/ORM/Persisters/SqlExpressionVisitor.php
@@ -12,6 +12,7 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Persisters\Entity\BasicEntityPersister;
 use RuntimeException;
 
+use function defined;
 use function implode;
 use function in_array;
 use function is_object;
@@ -81,6 +82,11 @@ class SqlExpressionVisitor extends ExpressionVisitor
                 return '(' . implode(' OR ', $expressionList) . ')';
 
             default:
+                // Multiversion support for `doctrine/collections` before and after v2.1.0
+                if (defined(CompositeExpression::class . '::TYPE_NOT') && $expr->getType() === CompositeExpression::TYPE_NOT) {
+                    return 'NOT (' . $expressionList[0] . ')';
+                }
+
                 throw new RuntimeException('Unknown composite ' . $expr->getType());
         }
     }
--- a/tests/Doctrine/Tests/ORM/Query/SqlExpressionVisitorTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/SqlExpressionVisitorTest.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Query;
+
+use Doctrine\Common\Collections\ExpressionBuilder as CriteriaBuilder;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Persisters\Entity\BasicEntityPersister;
+use Doctrine\ORM\Persisters\SqlExpressionVisitor;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+use function method_exists;
+
+class SqlExpressionVisitorTest extends TestCase
+{
+    /** @var SqlExpressionVisitor */
+    private $visitor;
+
+    /** @var BasicEntityPersister&MockObject */
+    private $persister;
+    /** @var ClassMetadata */
+    private $classMetadata;
+
+    protected function setUp(): void
+    {
+        $this->persister     = $this->createMock(BasicEntityPersister::class);
+        $this->classMetadata = new ClassMetadata('Dummy');
+        $this->visitor       = new SqlExpressionVisitor($this->persister, $this->classMetadata);
+    }
+
+    public function testWalkNotCompositeExpression(): void
+    {
+        if (! method_exists(CriteriaBuilder::class, 'not')) {
+            self::markTestSkipped('doctrine/collections in version ^2.1 is required for this test to run.');
+        }
+
+        $cb = new CriteriaBuilder();
+
+        $this->persister
+            ->expects(self::once())
+            ->method('getSelectConditionStatementSQL')
+            ->will(self::returnValue('dummy expression'));
+
+        $expr = $this->visitor->walkCompositeExpression(
+            $cb->not(
+                $cb->eq('foo', 1)
+            )
+        );
+
+        self::assertEquals('NOT (dummy expression)', $expr);
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "composer-runtime-api": "^2",
         "ext-ctype": "*",
         "doctrine/cache": "^1.12.1 || ^2.1.1",
-        "doctrine/collections": "^1.5 || ^2.0",
+        "doctrine/collections": "^1.5 || ^2.1",
         "doctrine/common": "^3.0.3",
         "doctrine/dbal": "^2.13.1 || ^3.2",
         "doctrine/deprecations": "^0.5.3 || ^1",
--- a/docs/en/reference/working-with-associations.rst
+++ b/docs/en/reference/working-with-associations.rst
@@ -718,6 +718,7 @@ methods:
 
 * ``andX($arg1, $arg2, ...)``
 * ``orX($arg1, $arg2, ...)``
+* ``not($expression)``
 * ``eq($field, $value)``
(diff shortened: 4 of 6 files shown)


greg0ire:
> It is not mergeable as it references a branch that has not been merged yet into doctrine/collections (see https://github.com/doctrine/collections/pull/348) but will be updated as soon as it is.

Converting to draft until then 👍

greg0ire:
This should target 2.14.x since it's a new feature. Also, should docs be updated?

remi-san:
The reason I put it under 3.0.x was that I assumed forcing an update to `doctrine/collection` version ^2.1 was maybe a bit too much for a minor version.

But, if you tell me that's not a problem, I'll be glad to rebase my branch on `2.14.x`.

greg0ire:
:thinking: you're right, that won't be OK… but maybe you can make the new ORM feature optional? Meaning that if `doctrine/collection` is < 2.1, the feature is not enabled?

remi-san:
@greg0ire I have no idea on how to do that 😅 . If you can point me to some resource on the matter, I'll be glad to try something.

And I will update the doc, of course.

greg0ire:
A way to detect that your feature is present would be to use `method_exists(CompositeExpression::class, 'not')`. Based on that, you can decide to do something or throw an exception defined in https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/Exception/NotSupported.php

remi-san:
Ok, so I've done that. For my use case I don't think I need to throw any exception as if it's a previous version of `doctrine/collections`, the code won't be called as the cases can't exist.

Still waiting for the `doctrine/collections` v2.1.0 to be tagged before it can stop being a draft, but I think everything is ready otherwise.

I just found one relevant reference to the `collections` in the doc and updated it,  but might have missed something.

greg0ire:
> Still waiting for the doctrine/collections v2.1.0 to be tagged before it can stop being a draft, but I think everything is ready otherwise.

Let's wait for a green build before doing that… it would be a shame to release v2.1.0 only to have to release v2.1.1 immediately after in case we are missing something.

tests/Doctrine/Tests/ORM/Query/SqlExpressionVisitorTest.php
```diff
@@ -0,0 +1,64 @@
...
+        $expr = $this->visitor->walkCompositeExpression(
+            $cb->not(
+                $cb->eq('foo', 1),
+            ),
+        );
+
+        self::assertEquals('NOT (dummy expression)', $expr);
+    }
+
+    private function getBasicEntityPersister(): BasicEntityPersister&MockObject
```
greg0ire: Intersection types are not supported in PHP 7

remi-san: Completely forgot I had to make it PHP7 compatible, sorry.

remi-san:
I just updated `composer.json` to use the correct version of `doctrine/collections`. This is not a draft anymore.

tests/Doctrine/Tests/ORM/Query/SqlExpressionVisitorTest.php
```diff
@@ -0,0 +1,64 @@
...
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Persisters\Entity\BasicEntityPersister;
+use Doctrine\ORM\Persisters\SqlExpressionVisitor;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+use function method_exists;
+
+/**
+ * Test for QueryExpressionVisitor
```
greg0ire: Looks wrong, and anyway, I don't think this is useful, please remove it.

tests/Doctrine/Tests/ORM/Query/SqlExpressionVisitorTest.php
```diff
@@ -0,0 +1,64 @@
...
+        );
+
+        self::assertEquals('NOT (dummy expression)', $expr);
+    }
+
+    private function getBasicEntityPersister(): BasicEntityPersister
+    {
+        return $this->getMockBuilder(BasicEntityPersister::class)
+            ->disableOriginalConstructor()
+            ->getMock();
```
greg0ire: Why not use `createMock()` here?

tests/Doctrine/Tests/ORM/Query/SqlExpressionVisitorTest.php
remi-san: This is a correction from php-cbf, otherwise it's a php-cs violation... and I need it to know which version of `doctrine/collections` I'm using.

greg0ire: Not having a comment is a phpcs violation? that would be a first… I don't see the link with needing to know the version of `doctrine/collections` you are using, so maybe you think I'm talking about something else than the comment on this test class?

remi-san: sorry, I thought you were talking about `use function method_exists;`

remi-san: It's fixed

tests/Doctrine/Tests/ORM/Query/SqlExpressionVisitorTest.php
```diff
@@ -0,0 +1,59 @@
...
+            )
+        );
+
+        self::assertEquals('NOT (dummy expression)', $expr);
+    }
+
+    private function getBasicEntityPersister(): BasicEntityPersister
+    {
+        return $this->createMock(BasicEntityPersister::class);
+    }
```
greg0ire: Let's inline this in `setUp`? I already didn't think it was worth extract this in a private method when it was more complex.

greg0ire:
Let's wait on  https://github.com/doctrine/orm/pull/10239

derrabus:
What's the status of this PR?

remi-san:
> What's the status of this PR?

I was waiting on the pipeline to be launched after rebasing my branch because something else broke it. But it seems in the mean time, some other commits broke it, so I've rebased it again and this should be good now.

lib/Doctrine/ORM/Persisters/SqlExpressionVisitor.php
```diff
@@ -80,6 +81,10 @@ public function walkCompositeExpression(CompositeExpression $expr)
             case CompositeExpression::TYPE_OR:
                 return '(' . implode(' OR ', $expressionList) . ')';
 
+            // Multiversion support for `doctrine/collections` before and after v2.1.0
+            case defined(CompositeExpression::class . '::TYPE_NOT') ? CompositeExpression::TYPE_NOT : 'NOT':
+                return 'NOT (' . $expressionList[0] . ')';
```
derrabus: I don't fully understand this change. Why are we falling back to `'NOT'` here? If the constant is not defined, doesn't this mean we cannot support that expression type?

remi-san: You're right. But because I'm in a `switch` structure, I couldn't define a `case` that should only be available if the constant exists, unless modifying it a bit more. Would you suggest I change its structure to something like this ?
```
$type = $expr->getType();
switch (true) {
    case ($type === CompositeExpression::TYPE_AND):
        return 'NOT (' . $expressionList[0] . ')';
    case ($type === CompositeExpression::TYPE_OR):
        return '(' . implode(' OR ', $expressionList) . ')';
    case (defined(CompositeExpression::class . '::TYPE_NOT') && $type === CompositeExpression::TYPE_NOT):
        return return 'NOT (' . $expressionList[0] . ')';
    default:
        throw new RuntimeException('Unknown composite ' . $type);
}
```

derrabus: I'd keep the `switch` as it was before and handle your new case inside the default block.

remi-san: I moved it to the default case. Tell me if there's anything else I can do.

derrabus (approved):

Issue #7786, opened 2019-07-31
Interfaces mis `@throw` annotations


hansbogert:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.3

#### Summary

Interfaces of, e.g., `EntityManagerInterface` do not include `@throw` annotations. 

#### Current behavior

This causes IDE's to not notice that for example calls to `EntityManagerInterface::flush` may cause exceptions.

#### How to reproduce

Reference `EntityManagerInterface` in an IDE like PhpStorm or other, and  have the following snippet

```php
function foo(EntityManagerInterface $em) {
  $em->flush();
}
```

#### Expected behavior

I expect the IDE to be able to tell, based on docblocks of the Doctrine library, that we should catch, (or re-throw) the exceptions which can occur.

Ocramius:
Documenting all `@throws` is not possible here, since there are multiple un-checked exception types upstream from that interface (in implementations).

You can gladly suggest those that are directly thrown in the `EntityManager` (implementation).

Related: https://github.com/doctrine/orm/issues/7780

hansbogert:
Well if you mean by unchecked exceptions, runtime exceptions, then I totally agree.

So basically we'd want to add 

```php
    /**
     * @throws \Doctrine\ORM\OptimisticLockException If a version check on an entity that
     *         makes use of optimistic locking fails.
     * @throws ORMException
     * @throws UniqueConstraintViolationException
     */
```

right?

Ocramius:
Something like that, but please look at the current 3.x work.

Ocramius:
See https://github.com/doctrine/orm/pull/6743

derrabus:
We're doing some housekeeping on the 3.0.0 milestone. Since nobody seems to be actively working on this topic I'm removing this PR from the 3.0.0 milestone.

That being said, the issue does not really look actionable at the moment. I'd be happy to discuss and merge any PR adding reasonable `@throws` annotations though. Also, if we take this topic seriously, we should think about some kind of tooling that tells us about correctness and completenes of such annotations.

greg0ire:
@morozov mentioned https://phpstan.org/blog/bring-your-exceptions-under-control in https://github.com/doctrine/dbal/pull/5777, we might want to look into it.

PR #10235, opened 2022-11-16
Add missing `throws` annotations

drupol:
This PR is from a discussion we had on Slack.

Basically, the `Repositoru::find()` method might throw an exception. However, it is not in the interface, this PR fix that.

Here's the missing `throws` annotations.

A following PR is incoming for `doctrine/persistence`.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/EntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/EntityPersister.php
@@ -5,7 +5,9 @@ declare(strict_types=1);
 namespace Doctrine\ORM\Persisters\Entity;
 
 use Doctrine\Common\Collections\Criteria;
+use Doctrine\DBAL\Exception;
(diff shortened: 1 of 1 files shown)


greg0ire:
> As discussed yesterday on Slack,

Other maintainers don't have access to my DMs, please don't refer to that discussion and share the reason you are doing this publicly.

greg0ire:
Relates to https://github.com/doctrine/orm/issues/7786

lib/Doctrine/ORM/Persisters/Entity/EntityPersister.php
```diff
@@ -190,6 +192,9 @@ public function getOwningTable($fieldName);
      *
      * @return object|null The loaded and managed entity instance or NULL if the entity can not be found.
      *
+     * @throws Exception
+     * @throws ORMException
```
derrabus, line 196: In which situations do we throw these exceptions and why should a caller handle them?

PR #10237, opened 2022-11-18
Fix entity being scheduled for update although DateTime values are logically equal

XCame:
Stumbled across this because I'm using the PreUpdate event and it got triggered after submitting an unchanged "entity edit" form containing a date input field.

It happens because:

```
$a = new \DateTime('2011-08-01 00:00:00.0');
$b = new \DateTime('2011-08-01 00:00:00.0');
var_dump($a === $b);  //<-- this returns bool(false)
```

and the UnitOfWork checks '"fully" MANAGED entities' for changes by

```
if ($orgValue === $actualValue) {
continue;
}
```

I hope the supplied unit-test is sufficient and in the right place.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -752,6 +752,13 @@ class UnitOfWork implements PropertyChangedListener
                 if ($orgValue === $actualValue) {
                     continue;
                 }
+                
+                //skip if values are DateTime objects and their value is equal
+                if($orgValue instanceof \DateTime && $actualValue instanceof \DateTime){
+                    if($orgValue->format('U') == $actualValue->format('U')){
+                        continue;
+                    }
+                }
 
                 // if regular field
                 if (! isset($class->associationMappings[$propName])) {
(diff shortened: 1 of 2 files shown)


beberlei:
This is currently the way its intended, and I believe its also documented as such

greg0ire:
[Indeed](https://www.doctrine-project.org/projects/doctrine-orm/en/2.13/reference/basic-mapping.html#doctrine-mapping-types):

> DateTime and Object types are compared by reference, not by value. Doctrine updates this values if the reference changes and therefore behaves as if these objects are immutable value objects.

derrabus:
I'm closing this PR now since the current behavior is documented and changing it might break existing applications. If we wanted to change the change detection on `DateTime` instances (not sure that we do), we cannot do this in a bugfix release.

Thank you for working on this PR though.

XCame:
No problem. You're absolutely right. I didn't see the big yellow box in the documentation.

I "needed" this behavior because I reset a "verified" status of an entity if a property is changed.
Then I'll find another way around my problem.

Thank you for your excellent work in this beauty of a project!

PR #10239, opened 2022-11-19
Adjust expectation

greg0ire:
`RuntimePublicReflectionProperty` has been deprecated in favor of `RuntimeReflectionProperty` in https://github.com/doctrine/persistence/pull/307.


composer.json
```diff
@@ -32,7 +32,7 @@
         "doctrine/inflector": "^1.4 || ^2.0",
         "doctrine/instantiator": "^1.3",
         "doctrine/lexer": "^1.2.3",
-        "doctrine/persistence": "^2.4 || ^3",
+        "doctrine/persistence": "^2.4 || ^3.1",
```
SenseException: Shouldn't this 3.0 version skip at least happen in a minor release instead of a patch one?

greg0ire: Sorry, I'll revert this, it's no longer necessary now that the test handles both cases.

derrabus (approved):

PR #10242, opened 2022-11-20
Solve some PHPStan baseline errors

VincentLanglet:
This is some things I saw in the baseline which could be easily solve I think.

### DIFF
--- a/lib/Doctrine/ORM/PersistentCollection.php
+++ b/lib/Doctrine/ORM/PersistentCollection.php
@@ -97,7 +97,7 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
      *
      * @param EntityManagerInterface $em    The EntityManager the collection will be associated with.
      * @param ClassMetadata          $class The class descriptor of the entity type of this collection.
-     * @psalm-param Collection<TKey, T> $collection The collection elements.
+     * @psalm-param Collection<TKey, T>&Selectable<TKey, T> $collection The collection elements.
      */
     public function __construct(EntityManagerInterface $em, $class, Collection $collection)
     {
@@ -675,11 +675,12 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
     /**
      * Retrieves the wrapped Collection instance.
      *
-     * @return Collection<TKey, T>
+     * @return Collection<TKey, T>&Selectable<TKey, T>
      */
     public function unwrap(): Collection
     {
-        assert($this->collection !== null);
+        assert($this->collection instanceof Collection);
+        assert($this->collection instanceof Selectable);
 
         return $this->collection;
     }
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -222,14 +222,14 @@ class UnitOfWork implements PropertyChangedListener
     /**
      * All pending collection deletions.
      *
-     * @psalm-var array<int, Collection<array-key, object>>
+     * @psalm-var array<int, PersistentCollection<array-key, object>>
      */
     private $collectionDeletions = [];
 
     /**
      * All pending collection updates.
      *
-     * @psalm-var array<int, Collection<array-key, object>>
+     * @psalm-var array<int, PersistentCollection<array-key, object>>
      */
     private $collectionUpdates = [];
 
@@ -238,7 +238,7 @@ class UnitOfWork implements PropertyChangedListener
      * At the end of the UnitOfWork all these collections will make new snapshots
      * of their data.
      *
-     * @psalm-var array<int, Collection<array-key, object>>
+     * @psalm-var array<int, PersistentCollection<array-key, object>>
      */
     private $visitedCollections = [];
 
@@ -419,12 +419,6 @@ class UnitOfWork implements PropertyChangedListener
         try {
             // Collection deletions (deletions of complete collections)
             foreach ($this->collectionDeletions as $collectionToDelete) {
-                if (! $collectionToDelete instanceof PersistentCollection) {
-                    $this->getCollectionPersister($collectionToDelete->getMapping())->delete($collectionToDelete);
-
-                    continue;
-                }
-
                 // Deferred explicit tracked collections can be removed only when owning relation was persisted
                 $owner = $collectionToDelete->getOwner();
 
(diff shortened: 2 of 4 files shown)


lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -675,11 +675,12 @@ public function matching(Criteria $criteria): Collection
     /**
      * Retrieves the wrapped Collection instance.
      *
-     * @return Collection<TKey, T>
+     * @return Collection<TKey, T>&Selectable<TKey, T>
      */
     public function unwrap(): Collection
```
VincentLanglet, line 680: In the matching method, there is the code
```
public function matching(Criteria $criteria): Collection
    {
        if ($this->isDirty) {
            $this->initialize();
        }

        if ($this->initialized) {
            return $this->unwrap()->matching($criteria);
        }
```
So, unwrap need to return a Selectable if we want to call `matching` on it.

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -97,7 +97,7 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
      *
      * @param EntityManagerInterface $em    The EntityManager the collection will be associated with.
      * @param ClassMetadata          $class The class descriptor of the entity type of this collection.
-     * @psalm-param Collection<TKey, T> $collection The collection elements.
+     * @psalm-param Collection<TKey, T>&Selectable<TKey, T> $collection The collection elements.
      */
     public function __construct(EntityManagerInterface $em, $class, Collection $collection)
```
VincentLanglet, line 102: Since `unwrap` need to return a Selectable (cf next comment), we need to only accepts selectable.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -419,12 +419,6 @@ public function commit($entity = null)
         try {
             // Collection deletions (deletions of complete collections)
             foreach ($this->collectionDeletions as $collectionToDelete) {
-                if (! $collectionToDelete instanceof PersistentCollection) {
```
VincentLanglet, line 422: This is not possible since `collectionDeletions` is always containing PersistentCollection.
- when looking at the added value they are always persistent collection
- it it was possible to go in this if, we would have called `$collectionToDelete->getMapping()` but `getMapping` only exists for PersistentCollection.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -222,14 +222,14 @@ class UnitOfWork implements PropertyChangedListener
     /**
      * All pending collection deletions.
      *
-     * @psalm-var array<int, Collection<array-key, object>>
+     * @psalm-var array<int, PersistentCollection<array-key, object>>
```
VincentLanglet, line 225: We only add PersistentCollection to this array
```
if ($orgValue instanceof PersistentCollection) {
                    // A PersistentCollection was de-referenced, so delete it.
                    $coid = spl_object_id($orgValue);

                    if (isset($this->collectionDeletions[$coid])) {
                        continue;
                    }

                    $this->collectionDeletions[$coid] = $orgValue;
                    $changeSet[$propName]             = $orgValue; // Signal changeset, to-many assocs will be ignored.

                    continue;
                }
```

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -222,14 +222,14 @@ class UnitOfWork implements PropertyChangedListener
...
      */
     private $collectionDeletions = [];
 
     /**
      * All pending collection updates.
      *
-     * @psalm-var array<int, Collection<array-key, object>>
+     * @psalm-var array<int, PersistentCollection<array-key, object>>
```
VincentLanglet, line 232: We only add PersistentCollection to this array
```
if ($value instanceof PersistentCollection && $value->isDirty()) {
            $coid = spl_object_id($value);

            $this->collectionUpdates[$coid]  = $value;
            $this->visitedCollections[$coid] = $value;
        }
```

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -238,7 +238,7 @@ class UnitOfWork implements PropertyChangedListener
      * At the end of the UnitOfWork all these collections will make new snapshots
      * of their data.
      *
-     * @psalm-var array<int, Collection<array-key, object>>
+     * @psalm-var array<int, PersistentCollection<array-key, object>>
```
VincentLanglet, line 241: We only add PersistentCollection to this array
```
if ($value instanceof PersistentCollection && $value->isDirty()) {
            $coid = spl_object_id($value);

            $this->collectionUpdates[$coid]  = $value;
            $this->visitedCollections[$coid] = $value;
        }
```

VincentLanglet:
I just have one error to solve for phpstan https://github.com/doctrine/orm/actions/runs/3509913825/jobs/5879333229
What do you recommend @greg0ire ?

It's related to the code
```
            $value = $refProp->getValue($entity);

            if ($class->isCollectionValuedAssociation($name) && $value !== null) {
                if ($value instanceof PersistentCollection) {
                    if ($value->getOwner() === $entity) {
                        continue;
                    }

                    $value = new ArrayCollection($value->getValues());
                }

                // If $value is not a Collection then use an ArrayCollection.
                if (! $value instanceof Collection) {
                    $value = new ArrayCollection($value);
                }

                $assoc = $class->associationMappings[$name];

                // Inject PersistentCollection
                $value = new PersistentCollection(
                    $this->em,
                    $this->em->getClassMetadata($assoc['targetEntity']),
                    $value
                );
```
we're only checking value is an instanceof Collection before passing the value to the PersistentCollection, but as explained in the comment of the PR, Persistent require we pass a Selectable&Collection or the `matching` method won't work.

Should I just add this error to the baseline ? (So it would be 5 less, for 1 more in the baseline)

greg0ire:
> Should I just add this error to the baseline ? (So it would be 5 less, for 1 more in the baseline)

Yes, it sounds like a fair tradeoff 👍

lib/Doctrine/ORM/UnitOfWork.php
greg0ire: This code was deliberately added in https://github.com/doctrine/orm/pull/7861, which is a bit worrying, but also, the test that was added in that PR still passes, so I guess it's fine?

VincentLanglet: Yes. I would say that the developer wanted to be BC
- If not PersistentCollection do like before
- If PersistentCollection add the bugfix

But it's not possible to have something else than PersistentCollection ^^
(You can look at the place where `collectionDeletions` is updated)

greg0ire (approved):

greg0ire:
Thanks @VincentLanglet !

Issue #10236, opened 2022-11-18
Deprecate: Load *-to-one entity with subClasses 


dmitryuk:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.0.0

#### Summary

UnitOfWork [has a way ](https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/UnitOfWork.php#L2871)to hydrate *-to-one relation entity, that has subClasses (inherited):

```
 case $targetClass->subClasses:
          // If it might be a subtype, it can not be lazy. There isn't even
          // a way to solve this with deferred eager loading, which means putting
          // an entity with subclasses at a *-to-one location is really bad! (performance-wise)
          $newValue = $this->getEntityPersister($assoc['targetEntity'])->loadOneToOneEntity($assoc, $entity, $associatedId);
          break;

```

But in my case I received a large performance overhead because my entity has hierarchic structure like `parent->parent->parent`.
So querying one Entity will execute and hydrate all `parent`s and it was really hard to find reason.
 

#### Previous behavior

Large overhead without notice

#### Current behavior proposal (2.0)
`trigger_deprecation` if this going on.

#### Doctrine 3 proposal
Throw an exception

PR #10244, opened 2022-11-22
[BC Break] Validate *-to-one relations with subclasses

dmitryuk:
Another way to fix an issue #10236 is to validate these *-to-one with inheritance classes
Target: 3.0


derrabus:
We don't merge new features into the 3.0.x branch. You need to target a lower branch, please.

PR #10245, opened 2022-11-22
Enums are considered as changes when Entity is fetched through query.

Warxcell:
Test case that reproduces the behaviour.


kbond:
I thought https://github.com/doctrine/orm/pull/10074 fixed but maybe not?

Warxcell:
> I thought #10074 fixed but maybe not?

Ah, could be, but its not released yet.

michnovka:
I will have a look into this today in few hours

michnovka:
Hmm, I dont get why

```php
$card = $this->_em->find(CardWithDefault::class, $cardId);
```
returns Proxy class for `CardWithDefault`. `suit` is not initialized it seems. And I am not sure whether passing proxy class to `computeChangeSet` is supported. @HypeMC @greg0ire can you have a look please?

Warxcell:
> Hmm, I dont get why
> 
> ```
> $card = $this->_em->find(CardWithDefault::class, $cardId);
> ```
> 
> returns Proxy class for `CardWithDefault`. `suit` is not initialized it seems. And I am not sure whether passing proxy class to `computeChangeSet` is supported. @HypeMC @greg0ire can you have a look please?

Ok, I will create exactly our use case then, for accuracy.

Warxcell:
Ok, seems its indeed resolved with https://github.com/doctrine/orm/pull/10074 , closing this.

PR #10247, opened 2022-11-23
Fix types for Paginator

derrabus:


greg0ire:
If this is a fix, why target 2.14.x?

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -27,6 +27,7 @@
  * The paginator can handle various complex scenarios with DQL.
  *
  * @template-covariant T
+ * @implements IteratorAggregate<array-key, T>
```
greg0ire, line 29: What does this fix?

greg0ire: Ah I think I know: it fixes a warning about implementing IteratorAggregate w/o specifying the types. I have that same fix in a branch that fixes part of the issues detected by Psalm v5, I will share it soon.

derrabus: Yes. I found it as well while playing with Psalm 5.

derrabus:
> If this is a fix, why target 2.14.x?

No strong opinion here, I can rebase onto 2.13.x.

derrabus:
Done.

greg0ire (approved):

PR #10256, opened 2022-11-26
Make use statements redundant

greg0ire:
In 68bc00b6c672349925c247951f13c7c90eb801e6, while fixing coding style, I introduced many, many use statements in that file. Using half-qualified names sometimes, and unqualified names some other times makes no sense.

If `AST\` is used at least once, use it always.


SenseException (approved):
I wouldn't mind the many use statements.

greg0ire:
I find it's a bit much, but also, since this was the original style, I elected to pick that one to eliminate the inconsistency.

PR #10257, opened 2022-11-26
Leverage Lexer's Token type (follow up)

greg0ire:
Follow-up of f82db6a8946ed43bfc9294f6fd90ee81ed8264f6. The previous value introduced in
dc37c2cd2fd22fe8b57ddb1743aa629db98b76e3 was too restrictive, hence the changes to the Psalm baseline.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -551,7 +551,7 @@ class Parser
      * @param bool $resetPeek Reset peek after finding the closing parenthesis.
      *
      * @return mixed[]
-     * @psalm-return array{value: string, type: int|null|string, position: int}|null
+     * @psalm-return Token|null
      */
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -551,7 +551,7 @@ public function semanticalError($message = '', $token = null)
      * @param bool $resetPeek Reset peek after finding the closing parenthesis.
      *
      * @return mixed[]
-     * @psalm-return array{value: string, type: int|null|string, position: int}|null
+     * @psalm-return Token|null
```
SenseException, line 554: If anything is forcing to install Lexer 1.2.0 this won't work. `Token` wasn't in `AbstractLexer` yet.

SenseException: Forget about it. 1.2.3 is set in Composer.

SenseException (approved):

PR #10261, opened 2022-11-30
Bump Psalm to 5.0.0 and fix errors for Symfony 6.2

derrabus:
The Symfony 6.2 release breaks our static analysis once again. This PR should fix the issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         "squizlabs/php_codesniffer": "3.7.1",
         "symfony/cache": "^4.4 || ^5.4 || ^6.0",
         "symfony/yaml": "^3.4 || ^4.0 || ^5.0 || ^6.0",
-        "vimeo/psalm": "4.30.0 || 5.0.0-rc1"
+        "vimeo/psalm": "4.30.0 || 5.0.0"
     },
     "conflict": {
         "doctrine/annotations": "<1.13 || >= 2.0"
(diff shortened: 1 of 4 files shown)


greg0ire:
The breaking test seems to have to do with `$COLUMNS`  / a `symfony/console` upgrade

greg0ire:
It looks as if https://github.com/doctrine/orm/blob/65da1fe8cb5612c018dd38192f8dd26faad567a5/phpunit.xml.dist#L69 was previously ignored and `100` was used instead.

greg0ire:
This is the commit that breaks the test according to git bisect: https://github.com/symfony/console/commit/5edfa27fd51e1af4d47f66e6efa3845999aa4fa7

Full PR at https://github.com/symfony/symfony/pull/45318

derrabus:
> It looks as if
> 
> https://github.com/doctrine/orm/blob/65da1fe8cb5612c018dd38192f8dd26faad567a5/phpunit.xml.dist#L69
> was previously ignored and `100` was used instead.

That setting still is ignored: If I set it to `100`, the test keeps failing.

greg0ire:
Must be a red herring, let's try the solution mentioned here: https://github.com/symfony/symfony-docs/pull/16476/files

derrabus:
> Full PR at [symfony/symfony#45318](https://github.com/symfony/symfony/pull/45318)

That PR was closed in favor of symfony/symfony#47062

greg0ire:
I don't think we can access the output wrapper form the test, so maybe let's just drop this `// ` from the assertion.

derrabus:
Yes, that's what I did now.

greg0ire (approved):

PR #10270, opened 2022-12-05
Fix deprecation message

derrabus:
Follow-up to #10267. 😓


greg0ire (approved):
Woops didn't spot that one

PR #10273, opened 2022-12-06
Added MultiManagerProvider

jordan-schnur:
A SingleManagerProvider was not enough for a more complex application using more than one entity manager in the cli.


derrabus:
Please add tests.

lib/Doctrine/ORM/Tools/Console/EntityManagerProvider/MultiManagerProvider.php
```diff
@@ -0,0 +1,40 @@
...
+namespace Doctrine\ORM\Tools\Console\EntityManagerProvider;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Tools\Console\EntityManagerProvider;
+
+use function array_keys;
+
+final class MultiManagerProvider implements EntityManagerProvider
+{
+    /** @var EntityManagerInterface[] */
```
derrabus, line 14: To make Psalm happy:

```suggestion
    /** @var array<string, EntityManagerInterface> */
```

derrabus, line 14: To make Psalm happy:

```suggestion
    /** @var array<string, EntityManagerInterface> */
```

tests/Doctrine/Tests/ORM/Tools/Console/EntityManagerProvider/MultiManagerProviderTest.php
```diff
@@ -0,0 +1,65 @@
...
+{
+    /** @var MultiManagerProvider */
+    private $multiManagerProvider;
+
+    /** @var EntityManagerInterface[] */
+    private $entityManagers;
+
+    protected function setUp(): void
+    {
+        parent::setUp(); // TODO: Change the autogenerated stub
```
derrabus: TODO: Remove the comment.

derrabus (changes requested):
Thank you for the test. However, that test does not really tell me that an instance of your provider can be used in a console application. Since that is the main purpose of such a provider, we should have a test that covers this scenario.

jordan-schnur:
> Thank you for the test. However, that test does not really tell me that an instance of your provider can be used in a console application. Since that is the main purpose of such a provider, we should have a test that covers this scenario.

Sorry for taking awhile. The test are updated.

derrabus (approved):

derrabus (changes requested):
Your test seems to be failing in our CI. Can you please check?

jordan-schnur:
> Your test seems to be failing in our CI. Can you please check?

@derrabus Finally back, sorry for taking so long. Can we rerun the CI pipeline so I can get the logs back? Everything passes locally so I am confused as to where the error is occurring.

PR #10274, opened 2022-12-06
Fix enum IDs in association mappings

### LINKED ISSUE #10132: Enum fields can't be an id field.

michnovka:
Attempt to fix https://github.com/doctrine/orm/issues/10132

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2812,6 +2812,10 @@ class UnitOfWork implements PropertyChangedListener
                         $joinColumnValue = $data[$srcColumn] ?? null;
 
                         if ($joinColumnValue !== null) {
+                            if ($joinColumnValue instanceof BackedEnum) {
+                                $joinColumnValue = $joinColumnValue->value;
+                            }
+
                             if ($targetClass->containsForeignIdentifier) {
                                 $associatedId[$targetClass->getFieldForColumn($targetColumn)] = $joinColumnValue;
                             } else {
--- a/tests/Doctrine/Tests/Models/GH10132/Complex.php
+++ b/tests/Doctrine/Tests/Models/GH10132/Complex.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\GH10132;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Id;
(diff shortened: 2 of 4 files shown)


michnovka:
@greg0ire please review, this fixes a very specific case where we are using complex associations to entities with one or more ID columns, one of which is an enum.

My original thought was to use identifierFlattener like this:

```php
$associatedIdFlattened = $this->identifierFlattener->flattenIdentifier($targetClass, $associatedId);
$relatedIdHash         = implode(' ', $associatedIdFlattened);
```

here - https://github.com/doctrine/orm/blob/20d6c53f299fc97997939671dd5ae8544c9e7e11/lib/Doctrine/ORM/UnitOfWork.php#L2850

But when running tests, this created some issues and BC breaks. I this this fix is sufficient. Other cases (like a single enum ID not part of association) already use checks like this

greg0ire:
Please squash your commits together and use a commit message as good as this one: 9d5ab4ce763810f42d2079bf42b8bf48264ffa5c
You can read [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches) to get guidelines rather than an example.

greg0ire:
> a very specific case where we are using complex associations to entities with one or more ID columns, one of which is an enum

In your test, all 3 entities have the enum appear in the primary key, is that required to reproduce the issue?

michnovka:
@greg0ire 

> In your test, all 3 entities have the enum appear in the primary key, is that required to reproduce the issue?

It is not required, but I did like the complex structure of the example as it works through 2 levels of association instead of just one.

michnovka:
I have reduced the example entity to a minimum and provided a nice commit message.

@greg0ire please review

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @michnovka !

PR #10277, opened 2022-12-08
Fix changeset computation for enum arrays

### LINKED ISSUE #10251: UnitOfWork still not right - probably Enum-related
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.13.2 and .3 and .4

#### Summary

I think I'm observing the same issue as https://github.com/doctrine/orm/issues/10124, but it wasn't fixed by 2.13.4, so I'm posting it separately.

I have a table with some thousand entries, and when I change a single one (through a Symfony form), *all* are getting "updated" (i.e. overwritten with their current values). 

The last version, which really works for me, is 2.13.1.

2.13.2 introduced the error:

> "Hearts,Clubs" is not a valid backing value for enum "App\Suit"

This was fixed in 2.13.3, but since then, persisting a single changed entity takes several minutes.

#### How to reproduce

I tried `dd($this->getEntityManager()->getUnitOfWork());`, but couldn't find any differences between the working and non-working versions - where should I look at to provide more information?

michnovka:
Fix https://github.com/doctrine/orm/issues/10251

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -744,8 +744,18 @@ class UnitOfWork implements PropertyChangedListener
 
                 $orgValue = $originalData[$propName];
 
-                if ($orgValue instanceof BackedEnum) {
-                    $orgValue = $orgValue->value;
+                if (! empty($class->fieldMappings[$propName]['enumType'])) {
+                    if (is_array($orgValue)) {
+                        foreach ($orgValue as $id => $val) {
+                            if ($val instanceof BackedEnum) {
+                                $orgValue[$id] = $val->value;
+                            }
+                        }
+                    } else {
+                        if ($orgValue instanceof BackedEnum) {
+                            $orgValue = $orgValue->value;
+                        }
+                    }
                 }
 
                 // skip if value haven't changed
@@ -804,6 +814,7 @@ class UnitOfWork implements PropertyChangedListener
                     }
 
                     if ($orgValue !== null && $assoc['orphanRemoval']) {
+                        assert(is_object($orgValue));
                         $this->scheduleOrphanRemoval($orgValue);
                     }
                 }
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -744,8 +744,18 @@ public function computeChangeSet(ClassMetadata $class, $entity)
 
                 $orgValue = $originalData[$propName];
 
-                if ($orgValue instanceof BackedEnum) {
-                    $orgValue = $orgValue->value;
+                if (! empty($class->fieldMappings[$propName]['enumType'])) {
```
michnovka, line 747: if somebody is interested why I added this check, its because I want to avoid the case where we have an array and go over the whole array checking if its `BackedEnum` for no reason. This way, unless there are enums in the play for the given field, no checks whatsoever are performed. Likely 0 performance gains with simple enums, but large performance gain with arrays of enums.

michnovka:
@greg0ire please review

greg0ire:
> Fix changeset computation for enum arrays

Please explain how it is broken, and what solution you applied in the commit message. I will review when that's done.

michnovka:
@greg0ire done

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -804,6 +814,7 @@ public function computeChangeSet(ClassMetadata $class, $entity)
                     }
 
                     if ($orgValue !== null && $assoc['orphanRemoval']) {
+                        /** @psalm-var object $orgValue */
```
greg0ire: We usually use `assert` instead of `@var`.

```suggestion
                        assert(is_object($orgValue));
```

greg0ire:
@ThomasLandauer can you please test this?

greg0ire (approved):

michnovka:
> @ThomasLandauer can you please test this?

I have tested it on https://github.com/ThomasLandauer/doctrine-issue-10251 and it did fix the issue

michnovka:
can we add this on 2.13.5 milestone?

greg0ire:
We usually do this right after merging.

SenseException (approved):

derrabus (approved):

PR #10278, opened 2022-12-09
Check if the intermediate table already exists before trying to generate it using associated mappings

0xf1a:
Doctrine already supports using customized intermediate entities in ManyToMany joins on multiple sides. However there is a problem when trying to migrate with this setup because SchemaTool will attempt to generate the table from both entity classes which contain ManyToMany joins and finally from the intermediate entity class. This makes sense because SchemaTool is not "aware" of the intermediate entity class' existence when parsing association mappings.

[EntityTest.zip](https://github.com/doctrine/orm/files/10195340/EntityTest.zip)


derrabus:
Please add tests.

PR #10282, opened 2022-12-10
Control proxy implementation via env

derrabus:
Our test suite is currently run with the lazy ghost implementation if the var-exporter component is installed. In our CI, we install the component if we want to test the lazy ghost implementation.

This is a bit brittle given that Symfony Cache depends on VarExporter, so the component will be installed no matter what. Also, if I want to run tests locally, it's a bit annoying that I have to install the component and live with local changes in my composer.json.

With this PR, I'm adding VarExporter to the dev requirements. I'm introducing an environment variable `ORM_PROXY_IMPLEMENTATION` which allows us to control the proxy implementation that should be used by the tests.

### DIFF
--- a/tests/Doctrine/Tests/TestUtil.php
+++ b/tests/Doctrine/Tests/TestUtil.php
@@ -10,17 +10,22 @@ use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Exception\DriverException;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
+use Doctrine\ORM\Configuration;
+use LogicException;
+use Symfony\Component\VarExporter\LazyGhostTrait;
 use UnexpectedValueException;
 
 use function assert;
 use function explode;
 use function fwrite;
 use function get_debug_type;
+use function getenv;
 use function method_exists;
 use function sprintf;
 use function str_starts_with;
 use function strlen;
 use function substr;
+use function trait_exists;
 
 use const STDERR;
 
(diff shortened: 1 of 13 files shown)


derrabus:
cc @nicolas-grekas

tests/Doctrine/Tests/TestUtil.php
```diff
@@ -74,6 +78,34 @@ public static function getPrivilegedConnection(): DbalExtensions\Connection
         return $connection;
     }
 
+    public static function configureProxies(Configuration $configuration): void
+    {
+        $configuration->setProxyDir(__DIR__ . '/Proxies');
+        $configuration->setProxyNamespace('Doctrine\Tests\Proxies');
+
+        $proxyImplementation = $_ENV['ORM_PROXY_IMPLEMENTATION']
+            ?? $_SERVER['ORM_PROXY_IMPLEMENTATION']
```
greg0ire: Why is this fallback to `$_SERVER` needed?

tests/Doctrine/Tests/TestUtil.php
```diff
@@ -74,6 +78,34 @@ public static function getPrivilegedConnection(): DbalExtensions\Connection
...
+            ?? null;
+
+        if (! $proxyImplementation) {
+            $proxyImplementation = trait_exists(LazyGhostTrait::class) ? 'lazy-ghost' : 'common';
```
greg0ire: Why not add an new `??` line with that ternary?

derrabus: Because I also want the fallback to kick in if the environment variable has been set to an empty string.

tests/Doctrine/Tests/TestUtil.php
derrabus: Some SAPIs store environment variables in `$_ENV`, some in `$_SERVER`. But since we're not dealing with any SAPI other than CLI, I can probably remove it.

tests/Doctrine/Tests/TestUtil.php
greg0ire: Why would we do that? :thinking:

derrabus: Because unsetting environment variables is hard. If you're dealing with enviroment variables, you should always treat the empty string as if the variable wasn't set.

Fun fact: I did not take the empty string into account at first which made the CI fail: https://github.com/doctrine/orm/actions/runs/3664887468/jobs/6195591631

greg0ire: I'm sorry, where are we unsetting that environment variable? And is that not as simple as `putenv('MYVAR')`?

tests/Doctrine/Tests/TestUtil.php
derrabus: In fact, on CLI, environment variables are stored in `$_SERVER`. I think I might just switch to `getenv()`. 🤦🏻

greg0ire: I remember stuggling with that myself in the past, and it seems to be quite a mess: https://www.php.net/manual/en/ini.core.php#ini.variables-order

`getenv()` seems like a good idea 👍

greg0ire (approved):

tests/Doctrine/Tests/TestUtil.php
derrabus: > I'm sorry, where are we unsetting that environment variable?

We aren't, but e.g. Github Actions needs to do that.

PR #6141, opened 2016-11-20
Allow using objects as discriminators

lcobucci:
@Ocramius this is the small change to make it work on `master`.

The test passes without any change on `2.5-x` so it looks like we have a small break on `master` which is fixed with this changeset.

PR #10288, opened 2022-12-11
Allow enum discriminator columns

michnovka:
In all hydrators, discriminator columns are for all purposes converted to string using explicit (string) conversion. This is not allowed with PHP enums. Therefore, I added code which checks whether the variable is a BackedEnum instance and if so, instead its ->value is used (which can be cast to string without issues as its a scalar)

The test builds upon https://github.com/doctrine/orm/pull/6141 which allowed objects to be discriminator column values. This fix further extends the use-case to allow PHP enums also.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -785,7 +785,7 @@
     </UndefinedInterfaceMethod>
   </file>
   <file src="lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php">
-    <InvalidArgument occurrences="1"/>
+    <InvalidArgument occurrences="2"/>
     <InvalidArrayAccess occurrences="4">
       <code>$value[0]</code>
       <code>$value[0]</code>
@@ -865,7 +865,7 @@
       <code>addNamedNativeQuery</code>
       <code>addNamedQuery</code>
     </DeprecatedMethod>
-    <InvalidArgument occurrences="4">
+    <InvalidArgument occurrences="5">
       <code>$this-&gt;cacheToArray($manyToManyElement-&gt;cache)</code>
       <code>$this-&gt;cacheToArray($manyToOneElement-&gt;cache)</code>
       <code>$this-&gt;cacheToArray($oneToManyElement-&gt;cache)</code>
@@ -931,6 +931,7 @@
       <code>addNamedNativeQuery</code>
       <code>addNamedQuery</code>
     </DeprecatedMethod>
+    <InvalidArgument occurrences="1"/>
     <InvalidDocblock occurrences="1">
       <code>private function cacheToArray(array $cacheMapping): array</code>
     </InvalidDocblock>
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10288Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10288Test.php
@@ -1,0 +1,240 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\StringType;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\ORM\AbstractQuery;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\DiscriminatorColumn;
+use Doctrine\ORM\Mapping\DiscriminatorMap;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\InheritanceType;
+use Doctrine\Tests\Models\GH10288\GH10288People;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * This test makes sure that Discriminator columns can use both custom types using PHP enums as well as
+ * enumType definition of enums.
+ *
+ * @requires PHP 8.1
+ */
+class GH10288Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        if (Type::hasType(GH10288PeopleType::NAME)) {
+            Type::overrideType(GH10288PeopleType::NAME, GH10288PeopleType::class);
+        } else {
+            Type::addType(GH10288PeopleType::NAME, GH10288PeopleType::class);
+        }
+
+        $this->createSchemaForModels(
+            GH10288PersonWithEnumType::class,
+            GH10288BossWithEnumType::class,
+            GH10288EmployeeWithEnumType::class,
+            GH10288PersonCustomEnumType::class,
+            GH10288BossCustomEnumType::class,
+            GH10288EmployeeCustomEnumType::class
+        );
+    }
+
+    /**
+     * @param class-string $personType
+     * @psalm-param GH10288BossWithEnumType|GH10288BossCustomEnumType $boss
+     * @psalm-param GH10288EmployeeWithEnumType|GH10288EmployeeCustomEnumType $employee
+     */
+    private function performEnumDiscriminatorTest($boss, $employee, string $personType): void
+    {
+        $boss->bossId = 1;
+
+        $this->_em->persist($boss);
+        $this->_em->persist($employee);
+        $this->_em->flush();
+        $bossId = $boss->id;
+        $this->_em->clear();
+
+        // Using DQL here to make sure that we'll use ObjectHydrator instead of SimpleObjectHydrator
+        $query = $this->_em->createQueryBuilder()
+            ->select('person')
+            ->from($personType, 'person')
+            ->where('person.name = :name')
+            ->setMaxResults(1)
+            ->getQuery();
+
+        $query->setParameter('name', 'John');
+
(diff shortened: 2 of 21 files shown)


michnovka:
@greg0ire please review

greg0ire (commented):
Kudos on the great commit message, it makes it really easier to understand what's happening here.

Any way you could test the scenarios Codecov is pointing at?

derrabus:
I'm a bit hesitant to treat this as a bugfix. Let's target 2.14, please.

michnovka:
@derrabus I have rebased to 2.14.x
@greg0ire I have cleaned up the code. Turns out the only problematic place was `ObjectHydrator`, the other Hydrators do not convert discriminator into enum at the time when they work with it. Ive added test for `SimpleHydrator` as well to prove this point and throw an exception should this behavior change in the future

greg0ire (dismissed):

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10288Test.php
```diff
@@ -0,0 +1,230 @@
...
+    public function getName()
+    {
+        return self::NAME;
+    }
+}
+
+/**
+ * @Entity
+ * @InheritanceType("JOINED")
+ * @DiscriminatorColumn(name="discr", type="GH10288people")
```
derrabus: Why do I still need a custom type for my enum?

michnovka: `DiscriminatorColumn` is not a `Column`, you cannot specify `enumType`. So the only way to emulate the case where discriminator column gets a PHP enum assigned to it during hydration is to use a custom type which returns `BackedEnum` for its `convertToPhpValue`

derrabus: > `DiscriminatorColumn` is not a `Column`, you cannot specify `enumType`

I know, but I would kinda expect that a PR that bears the title "Allow enum discriminator columns" would change exactly that. 🤔 

As long as this custom type is needed, the whole feature feels like a hack to me.

derrabus (changes requested):
Blocking the merge until https://github.com/doctrine/orm/pull/10288#discussion_r1046254084 is resolved.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10288Test.php
michnovka: but there is absolutely no point to change this. discriminator column is special, it is not accessible from user-end code. The column is defined only at the top of the Entity, and there is no PHP field that can be accessed when working with such entity.
The issue arises, when you try to use custom type, which works with enums. This is because PHP does not support `__toString()` overload on enums. Objects are supported as discriminator columns exactly by leveraging `__toString()` override. Since this is not possible with enums, the only way is to check with `instanceof` and use enum's value.
There is absolutely no reason nor use-case to allow adding `enumType` into the discriminator column. There is reason however to allow custom types using enums, for the exact same reason you want to allow objects to be used as discriminator columns (implemented with https://github.com/doctrine/orm/pull/6141)

derrabus: > discriminator column is special, it is not accessible from user-end code

And why would I need a custom type for an inaccessible column?

> The column is defined only at the top of the Entity, and there is no PHP field that can be accessed when working with such entity.

You can still select and hydrate that column, like you did in your test, right?

michnovka: > And why would I need a custom type for an inaccessible column?

Well e.g. in my case, I use custom type that creates native MySQL ENUM. When converting to PHP, it converts to PHP enum value thats specified by the given type. So if I want to have in mysql DB the field as ENUM, I have to use a custom type.

> You can still select and hydrate that column, like you did in your test, right?

ok, let me try to implement this

michnovka:
@derrabus I have implemented what you asked for.

`enumType` is now added to `DiscriminatorColumn`, so it can (and is) properly hydrated.

I have added tests to `MappingDriverTestCase` for the new option as well as specific Functional tests which tests both custom type using enum and "native doctrine" `enumType`.

please review

michnovka:
@derrabus not sure why the checks are not running for my last commit, can you please force run them?

michnovka:
I am happy with the PR now, I think it covers everything:
- add support to all mapping drivers
- add support to `ClassMetadataBuilder`
- add support into hydrators
- add tests for everything newly implemented
- add and fix documentation
- squash commits and add a single descriptive commit message

please review @greg0ire @derrabus

psalm-baseline.xml
```diff
@@ -785,7 +785,7 @@
     </UndefinedInterfaceMethod>
   </file>
   <file src="lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php">
-    <InvalidArgument occurrences="1"/>
+    <InvalidArgument occurrences="2"/>
```
derrabus, line 788: Which new errors did you encounter here? Is it possible to fix them?

michnovka: Its because of
https://github.com/doctrine/orm/blob/497ee166bdffe45dd85494fe1eb31d454fad363b/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php#L268-L276

Since the `ClassMetadataInfo::setDiscriminatorColumn($columnDef)` expects for `$columnDef` an array with `enumType?: class-string<BackedEnum>|null` but the `(string)` conversion confuses psalm into treating it as `string`

```
::error file=lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php,line=269,col=25,title=InvalidArgument::lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php:269:25: InvalidArgument: Argument 1 of Doctrine\ORM\Mapping\ClassMetadata::setDiscriminatorColumn expects array{columnDefinition?: null|string, enumType?: class-string<BackedEnum>|null, fieldName?: string, length?: int, name: null|string, type?: string}|null, but array{columnDefinition: null|string, enumType: null|string, length: int, name: null|string, type: string} provided (see https://psalm.dev/004)
```

derrabus: All right.

michnovka: Btw, I do have some more ideas:

```php
$discrColumnAttribute = $classAttributes[Mapping\DiscriminatorColumn::class];
assert($discrColumnAttribute instanceof Mapping\DiscriminatorColumn);
```

This does resolve the issue, but it creates new one:

```php
::error file=lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php,line=271,col=88,title=RedundantCastGivenDocblockType::lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php:271:88: RedundantCastGivenDocblockType: Redundant cast to string given docblock-provided type (see https://psalm.dev/263)
::error file=lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php,line=272,col=88,title=RedundantCastGivenDocblockType::lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php:272:88: RedundantCastGivenDocblockType: Redundant cast to string given docblock-provided type (see https://psalm.dev/263)
::error file=lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php,line=273,col=90,title=RedundantCastGivenDocblockType::lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php:273:90: RedundantCastGivenDocblockType: Redundant cast to int given docblock-provided type (see https://psalm.dev/263)
::error file=lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php,line=274,col=100,title=RedundantCastGivenDocblockType::lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php:274:100: RedundantCastGivenDocblockType: Redundant cast to string given docblock-provided type (see https://psalm.dev/263)
::error file=lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php,line=275,col=92,title=RedundantCastGivenDocblockType::lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php:275:92: RedundantCastGivenDocblockType: Redundant cast to class-string<BackedEnum> given docblock-provided type (see https://psalm.dev/263)
```

However I think it should be safe to remove the casting and strong-type the `DiscriminatorColumn` fields. It will require some fixing in internal code, as sometimes I saw we assign `$length` as `string` instead of int, and we need to treat this as BC break, but since this targets 2.14 and the ultimate goal is to have strong typed everything by 3.0, I think we should do this. WDYT?

michnovka: Actually, let me fix this in 3.0 after merge, I will replace

https://github.com/doctrine/orm/blob/2bfcb6e5ab3d4f66090179a2f4e936511e64d175/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php#L258-L265

with 

```php
assert($discrColumnAttribute instanceof Mapping\DiscriminatorColumn);

$metadata->setDiscriminatorColumn(
    [
        'name'             => $discrColumnAttribute->name ?? null,
        'type'             => $discrColumnAttribute->type ?? 'string',
        'length'           => $discrColumnAttribute->length ?? 255,
        'columnDefinition' => $discrColumnAttribute->columnDefinition ?? null,
        'enumType'         => $discrColumnAttribute->enumType ?? null,
    ]
);
```

derrabus (approved):

greg0ire:
Thanks @michnovka !

michnovka:
please merge into 3.x so that I can do the cleanup. thanks!

Issue #9561, opened 2022-03-03
Allow custom DBAL types autoconfiguration from PHP types


kiler129:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Current problem
Starting with Doctrine 2.9 mapping `type` can be autoconfigured based on the class property type (per [*"Typed Property Defaults"*](https://www.doctrine-project.org/2021/05/24/orm2.9.html)). As of now only certain built-in types can be configured in such a way. Practically, looking at the [current code](https://github.com/doctrine/orm/blob/6d5da83c68c1e1dcb408290cc7bc6f485c0a156f/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L1521), only scalars, arrays and DateTime can be configured in such a way.

#### Suggested feature
I suggest that [custom types](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/cookbook/custom-mapping-types.html) should also be part of that autoconfiguration effort to make the transition more complete.

#### Rationale
Current implementation feels slightly underbaked from the user's perspective, and make sense only internally. Users see that they can automagically get their PHP types configured (strings, arrays, etc), they can also count on enums being configured (afaik: custom type), but when they define/install a totally custom type typehinting the object suddenly produces a confusing exception:

```
Cannot assign string to property Entity::$property of type ?CustomType
```

#### Solution
Custom types could get a new method, similar to `getBindingType()` or `getMappedDatabaseTypes()`, which returns a list of FQCNs which the types is "claiming". Then `validateAndCompleteTypedFieldMapping()` can get the type based on the FQCN registered from `TypeRegistry`.

This will allow a configuration like this:
```php
    #[ORM\Column]
    public ?string $title;

    #[ORM\Column]
    public \DateTimeInterface $createdAt;

    #[ORM\Column(type: 'custom_type')]
    public CustomType $data;
```

to be squashed into:
```php
    #[ORM\Column]
    public ?string $title;

    #[ORM\Column]
    public \DateTimeInterface $createdAt;

    #[ORM\Column]
    public CustomType $data;
```

WDYT?

michnovka:
@derrabus I like this proposal, if it gets some support from you and @beberlei I am willing to work on this

michnovka:
How would we solve the case where 2 or more types want to claim ownership of the same FQCN? Should this be a fatal error, or we assume last one is always valid?

[...]

PR #10290, opened 2022-12-11
Allow custom DBAL types autoconfiguration from PHP types

michnovka:
Work in progress to implement https://github.com/doctrine/orm/issues/9561

I wanted to use this syntax in configuration:

```yaml
doctrine:
    orm:
        typed_field_mapping:
            App\Objects\CustomObject: App\DBAL\Types\CustomTypeFQCN
            App\Objects\CustomObject2: custom_type2_using_name_from_dbal
```

The problem I am facing is that `ClassMetadataInfo` does not seem to have access to the `Configuration` object. I wonder what would be the proper strategy here?

I've explored some ideas, like adding another param to the `ClassMetadataInfo::mapField($mapping)` function, but even its callers do not have access to the `Configuration` object.

@beberlei @derrabus @greg0ire please, any ideas?

### DIFF
--- a/docs/en/reference/basic-mapping.rst
+++ b/docs/en/reference/basic-mapping.rst
@@ -289,6 +289,77 @@ These are the "automatic" mapping rules:
 As of version 2.11 Doctrine can also automatically map typed properties using a
 PHP 8.1 enum to set the right ``type`` and ``enumType``.
 
+.. versionadded:: 2.14
+
+Since version 2.14 you can specify custom typed field mapping between PHP type and DBAL type using ``Configuration``
+
+.. code-block:: php
+
+    <?php
+    use App\CustomIds\CustomIdObject;
+    use App\DBAL\Type\CustomIdObjectType;
+
+    $configuration->addTypedFieldMapping(CustomIdObject::class, CustomIdType::class);
+
+Then, an entity using the ``CustomIdObject`` typed field will be correctly assigned its DBAL type
+(``CustomIdObjectType``) without the need of explicit declaration.
+
+.. configuration-block::
+
+    .. code-block:: attribute
+
+        <?php
+        #[ORM\Entity]
+        #[ORM\Table(name: 'cms_users_typed_with_custom_typed_field')]
+        class UserTypedWithCustomTypedField
+        {
+            #[ORM\Column]
+            public CustomIdObject $customId;
+
+            // ...
+        }
+
+    .. code-block:: annotation
+
+        <?php
+        /**
+         * @Entity
+         * @Table(name="cms_users_typed_with_custom_typed_field")
+         */
+        class UserTypedWithCustomTypedField
+        {
+            /** @Column */
+            public CustomIdObject $customId;
+
+            // ...
+        }
+
+    .. code-block:: xml
+
+        <doctrine-mapping>
+          <entity name="UserTypedWithCustomTypedField">
+            <field name="customId"/>
+            <!-- -->
+          </entity>
+        </doctrine-mapping>
+
+    .. code-block:: yaml
+
+        UserTypedWithCustomTypedField:
+          type: entity
+          fields:
+            customId: ~
+
+It is perfectly valid to override even the "automatic" mapping rules mentioned above:
+
+.. code-block:: php
+
+    <?php
+    use App\CustomIds\CustomIdObject;
+    use App\DBAL\Type\CustomIdObjectType;
+
+    $configuration->addTypedFieldMapping('int', CustomIntType::class);
+
 .. _reference-mapping-types:
 
 Doctrine Mapping Types
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -13,6 +13,7 @@ use Doctrine\Common\Cache\Cache as CacheDriver;
 use Doctrine\Common\Cache\Psr6\CacheAdapter;
 use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\Common\Persistence\PersistentObject;
+use Doctrine\DBAL\Types\Type;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Cache\CacheConfiguration;
 use Doctrine\ORM\Cache\Exception\CacheException;
@@ -28,6 +29,7 @@ use Doctrine\ORM\Exception\ProxyClassesAlwaysRegenerating;
 use Doctrine\ORM\Exception\UnknownEntityNamespace;
 use Doctrine\ORM\Internal\Hydration\AbstractHydrator;
 use Doctrine\ORM\Mapping\ClassMetadataFactory;
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Mapping\DefaultEntityListenerResolver;
 use Doctrine\ORM\Mapping\DefaultNamingStrategy;
 use Doctrine\ORM\Mapping\DefaultQuoteStrategy;
@@ -63,6 +65,7 @@ use function trim;
  * Internal note: When adding a new configuration option just write a getter/setter pair.
  *
  * @psalm-import-type AutogenerateMode from ProxyFactory
+ * @psalm-import-type ScalarName from ClassMetadataInfo
  */
 class Configuration extends \Doctrine\DBAL\Configuration
 {
@@ -719,7 +722,7 @@ class Configuration extends \Doctrine\DBAL\Configuration
      * @param string $name
      *
      * @return string|callable|null
-     * @psalm-return class-string|callable|null $name
+     * @psalm-return class-string|callable|null
      */
     public function getCustomDatetimeFunction($name)
     {
@@ -746,6 +749,57 @@ class Configuration extends \Doctrine\DBAL\Configuration
         foreach ($functions as $name => $className) {
             $this->addCustomDatetimeFunction($name, $className);
         }
+    }
+
+    /**
+     * Sets a map of php typed fields to DBAL types for auto-completion.
+     *
+     * Keys must be FQCNs of the PHP types
+     * Values are either FQCNs of DBAL Type or their name
+     *
+     * @param array $mappings The map of custom DQL date/time functions.
+     * @psalm-param array<class-string|ScalarName, class-string<Type>|string> $mappings
+     */
+    public function setTypedFieldMappings(array $mappings): void
+    {
+        foreach ($mappings as $phpType => $dbalType) {
+            $this->addTypedFieldMapping($phpType, $dbalType);
+        }
+    }
+
+    /**
+     * Registers a custom PHP type => DBAL Type mapping for typed Entity fields.
+     * With this configuration option used there is no need to specify the Column::type
+     * for given explicitly typed field.
+     *
+     * @psalm-param class-string|ScalarName   $phpType  the FQCN of PHP type to be mapped
+     * @psalm-param class-string<Type>|string $dbalType either FQCN of DBAL Type or its name
+     */
+    public function addTypedFieldMapping(string $phpType, string $dbalType): void
+    {
+        $this->_attributes['typedFieldMappings'][$phpType] = $dbalType;
+    }
+
+    /**
+     * Gets the mapped DBAL Type for given PHP type
+     *
+     * @psalm-param class-string|ScalarName $phpType  the FQCN of PHP type to be mapped
+     *
+     * @psalm-return class-string<Type>|string|null either FQCN of DBAL Type or its name
+     */
+    public function getTypedFieldMapping($phpType): ?string
+    {
+        return $this->_attributes['typedFieldMappings'][$phpType] ?? null;
+    }
+
+    /**
+     * Gets the array of all the PHP type => DBAL Type mappings
+     *
+     * @psalm-return array<class-string|ScalarName,class-string<Type>|string>
+     */
+    public function getTypedFieldMappings(): array
+    {
+        return $this->_attributes['typedFieldMappings'] ?? [];
     }
 
     /**
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -142,6 +142,7 @@ use const PHP_VERSION_ID;
  *     type: int,
  *     unique?: bool,
  * }
+ * @psalm-type ScalarName = 'array'|'bool'|'float'|'int'|'string'
  */
 class ClassMetadataInfo implements ClassMetadata
 {
@@ -800,19 +801,24 @@ class ClassMetadataInfo implements ClassMetadata
     /** @var InstantiatorInterface|null */
     private $instantiator;
 
+    /** @var array<class-string|ScalarName,class-string<Type>|string> $typedFieldMappings */
+    private $typedFieldMappings;
+
     /**
      * Initializes a new ClassMetadata instance that will hold the object-relational mapping
      * metadata of the class with the given name.
      *
-     * @param string $entityName The name of the entity class the new instance is used for.
+     * @param string                                                    $entityName         The name of the entity class the new instance is used for.
+     * @param array<class-string|ScalarName, class-string<Type>|string> $typedFieldMappings
      * @psalm-param class-string<T> $entityName
      */
-    public function __construct($entityName, ?NamingStrategy $namingStrategy = null)
+    public function __construct($entityName, ?NamingStrategy $namingStrategy = null, array $typedFieldMappings = [])
     {
-        $this->name           = $entityName;
-        $this->rootEntityName = $entityName;
-        $this->namingStrategy = $namingStrategy ?: new DefaultNamingStrategy();
-        $this->instantiator   = new Instantiator();
+        $this->name               = $entityName;
+        $this->rootEntityName     = $entityName;
+        $this->namingStrategy     = $namingStrategy ?: new DefaultNamingStrategy();
+        $this->instantiator       = new Instantiator();
+        $this->typedFieldMappings = $typedFieldMappings;
     }
 
     /**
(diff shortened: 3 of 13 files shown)


michnovka:
I propose this solution: 
- new `Configuration` option added with helper functions to get/set/add `typedFieldMappings`
- this mapping is inserted into `ClassMetadata` during construction (using `ClassMetadataFactory`)
- only if the field is not in the list of built-in types with automatic mapping (DateInterval, DateTime, DateTimeImmutable,
array, bool, float, int) it is checked whether we have configured custom PHP type => DBAL Type mapping

I have also updated docs under Basic Mapping section.

TODO: DoctrineBundle needs to be updated to support config syntax

```yaml
doctrine:
    orm:
        typed_field_mapping:
            App\Objects\CustomObject: App\DBAL\Types\CustomTypeFQCN
            App\Objects\CustomObject2: custom_type2_using_name_from_dbal
```

@derrabus @greg0ire please review

docs/en/reference/basic-mapping.rst
```diff
@@ -289,6 +289,68 @@ These are the "automatic" mapping rules:
...
+.. code-block:: php
+
+    <?php
+    use App\CustomIds\CustomIdObject;
+    use App\DBAL\Type\CustomIdObjectType;
+
+    $configuration->addTypedFieldMapping(CustomIdObject::class, CustomIdType::class);
+
+then, an entity using the ``CustomIdObject`` typed field will be correctly assigned its DBAL type
+(``CustomIdObjectType``) without the need of explicit declaration
```
greg0ire:
```suggestion
Then, an entity using the ``CustomIdObject`` typed field will be correctly assigned its DBAL type
(``CustomIdObjectType``) without the need of explicit declaration.
```

lib/Doctrine/ORM/Configuration.php
```diff
@@ -748,6 +749,60 @@ public function setCustomDatetimeFunctions(array $functions)
...
+     * Values are either FQCNs of DBAL Type or their name
+     *
+     * Any previously added mappings are discarded.
+     *
+     * @param array $mappings The map of custom DQL date/time functions.
+     * @psalm-param array<class-string, class-string<Type>|string> $mappings
+     *
+     * @return void
+     */
+    public function setTypedFieldMappings(array $mappings)
```
greg0ire: Please use native type declarations here and below.

```suggestion
     */
    public function setTypedFieldMappings(array $mappings): void
```

lib/Doctrine/ORM/Configuration.php
```diff
@@ -748,6 +749,60 @@ public function setCustomDatetimeFunctions(array $functions)
         }
     }
 
+    /**
+     * Sets a map of php typed fields to DBAL types for auto-completion.
+     *
+     * Keys must be FQCNs of the PHP types
+     * Values are either FQCNs of DBAL Type or their name
+     *
+     * Any previously added mappings are discarded.
```
greg0ire: That does not seem to be what the method does.

lib/Doctrine/ORM/Configuration.php
```diff
@@ -748,6 +749,60 @@ public function setCustomDatetimeFunctions(array $functions)
...
+     *
+     * @psalm-param class-string $phpType  the FQCN of PHP type to be mapped
+     *
+     * @return string|null
+     * @psalm-return class-string<Type>|string|null either FQCN of DBAL Type or its name
+     */
+    public function getTypedFieldMapping($phpType)
+    {
+        return $this->_attributes['typedFieldMappings'][$phpType] ?? null;
+    }
```
greg0ire, line 793: This method is never used.

michnovka: it is not at the moment, but I think its useful method and it follows the pattern of other other functions like `set/get/addCustomHydrationMode(s)`  or `set/get/addCustomDatetimeFunction(s)`. Id like to keep it.

lib/Doctrine/ORM/Configuration.php
michnovka: Hmm, what do you mean? The part that

> Any previously added mappings are discarded.

is indeed confusing, I believe it means to say that previous mapping for the same PHP type is overriden. I copy pasted this from other places, but it seems that this comment is wrong on every single place in the configuration. check
`setCustomStringFunctions`, `setCustomNumericFunctions` and `setCustomDatetimeFunctions`

I would say this is also wrong there, the previous mappings are never discarded, it always add the new ones on top of the existing stack and in case key already exists, it overwrites it with new value.

Should I fix all these comments?

greg0ire: > Should I fix all these comments?

If you have the time to do that, that would be great yes! Not a blocker for this PR though. In any case, I don't think consistency should be used to reproduce past errors :wink:

lib/Doctrine/ORM/Configuration.php
greg0ire: Ok, I'll trust your judgement on this, although it's still unclear to me where/why/how those other functions are used.

lib/Doctrine/ORM/Configuration.php
michnovka: I will do this in a separate PR to maintain issue-per-PR isolation

michnovka:
@greg0ire all feedback is incorporated, please review

michnovka:
btw, 2.14 release is very close, id like to include this in that release please. Can you add it to the milestone @derrabus please?

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1621,6 +1626,10 @@ private function validateAndCompleteTypedFieldMapping(array $mapping): array
                     case 'string':
                         $mapping['type'] = Types::STRING;
                         break;
+                    default:
+                        if (isset($this->typedFieldMappings[$type->getName()])) {
```
greg0ire: I think `getName()` is deprecated. I don't think we should use it in a new feature, right?

michnovka: this is `ReflectionNamedType` https://www.php.net/manual/en/reflectionnamedtype.getname.php

I think you confused this with `Doctrine\DBAL\Types\Type`

greg0ire: Ah yes indeed, sorry for the misunderstanding!

greg0ire (dismissed):

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1621,6 +1626,10 @@ private function validateAndCompleteTypedFieldMapping(array $mapping): array
                     case 'string':
                         $mapping['type'] = Types::STRING;
                         break;
+                    default:
```
beberlei: Is the way we want it? You couldn't for example override string globally to map to a different Type with this implementation. I would think we can improve this by moving all the default type mappings into `typedFieldMappings` as default value.

beberlei (commented):
So far so good, I have one improvement idea, but generally this is great, thank you for working on it

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
michnovka: ok, good idea, Ill implement this

michnovka: can we wait with 2.14.x for this? id hate to have it postponed until 2.15.x whe its almost done

michnovka:
@beberlei I took your suggestion and extended the implementation to allow overriding of the default types as well. Added test for this as well as documentation

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -800,19 +800,24 @@ class ClassMetadataInfo implements ClassMetadata
...
+    /** @var array<class-string|string,class-string<Type>|string> $typedFieldMappings */
+    private $typedFieldMappings;
+
     /**
      * Initializes a new ClassMetadata instance that will hold the object-relational mapping
      * metadata of the class with the given name.
      *
-     * @param string $entityName The name of the entity class the new instance is used for.
+     * @param string                                                $entityName         The name of the entity class the new instance is used for.
+     * @param array<class-string|string, class-string<Type>|string> $typedFieldMappings
```
michnovka: is there a way to make some custom psalm-type which would only allow subset of strings? What Id like is to make e.g. `scalar-name` which would only allow 'int','string','float','bool','array'. This way we could replace with more precise
```php
@param array<class-string|scalar-name, class-string<Type>|string> $typedFieldMappings
```

greg0ire: Yes I think you can use `'int'|'string'|'float'` etc. in your type alias

stof: We cannot restrict the type names there. Custom types are registered with a name too so the list of available strings is infinite.

michnovka: @stof this is PHP types, not DBAL types

michnovka:
I have added better psalm type for the `typedFieldMappings` and associated functions for the PHP types which can be mapped. This PR is finished imo and ready for merge.

michnovka:
I got one more idea, for another config option, but it should be part of this commit I think:

```yaml
doctrine:
    orm:
        typed_field_mapping:
            App\Objects\CustomObject: App\DBAL\Types\CustomTypeFQCN
            App\Objects\CustomObject2: custom_type2_using_name_from_dbal
        automatic_enum_field_mapping: true #default false
```

If this is set, it would consider

```php
#[Column]
public MyEnum $enumField;
```
as
```php
#[Column(type: 'string', enumType: MyEnum::class)]
public MyEnum $enumField;
```

(of course `string` or `int` based on enum backing type)

WDYT?

michnovka:
well actually this is already implemented, and it is always used, i.e.

```php
#[Column]
public MyEnum $enumField;
```

will work just fine. However I got another idea in the process. Imagine you have a custom DBAL type for enums. All your enums in your code are supposed to use this DBAL type. With the feature as is now in this PR, the only way is to list all your enums like

```yaml
doctrine:
    orm:
        typed_field_mapping:
            App\Enum\Enum1: App\DBAL\Types\CustomEnumType
            App\Enum\Enum2: App\DBAL\Types\CustomEnumType
            # ...
```

### Proposal:

Lets move the `ClassMetadataInfo::validateAndCompleteTypedFieldMapping()` function into a newly created class:

```php
class DefaultTypedFieldMapper implements TypedFieldMapper
```
where
```php
interface TypedFieldMapper
{
    /**
     * Validates & completes the given field mapping based on typed property.
     *
     * @param  array{fieldName: string, type?: mixed} $mapping The field mapping to validate & complete.
     *
     * @return array{fieldName: string, enumType?: string, type?: mixed} The updated mapping.
     */
    public function validateAndCompleteTypedFieldMapping(array $mapping, ReflectionProperty $field): array;

    /**
     * Registers a custom PHP type => DBAL Type mapping for typed Entity fields.
     * With this configuration option used there is no need to specify the Column::type
     * for given explicitly typed field.
     *
     * @psalm-param class-string|ScalarName   $phpType  the FQCN of PHP type to be mapped
     * @psalm-param class-string<Type>|string $dbalType either FQCN of DBAL Type or its name
     */
    public function addTypedFieldMapping(string $phpType, string $dbalType): void;
}
```

we would then allow this config as well:

```yaml
doctrine:
    orm:
        typed_field_mapper: App\ORM\CustomTypedFieldMapper
        typed_field_mapping:
            App\Objects\CustomObject: App\DBAL\Types\CustomTypeFQCN
            App\Objects\CustomObject2: custom_type2_using_name_from_dbal
```

such custom typed field mapper would be able to check e.g. for inheritance using instanceof, compare type names or whatever flexibility is desired. Notice that we can still add custom typed field mapping in the config, this would be passed to the `TypedFieldMapper` using its `addTypedFieldMapping` function. The mapper then decides whatever to do with this information.

The `DefaultTypedFieldMapper` would behave in the exact same way as it is implemented now in this PR. But say you want to assign all classes from your `App\CusomtObjects\Stringable` namespace to the string DBAL type. With this feature, all you need is a custom `TypedFieldMapper`  and no need to list all the types individually.

WDYT @greg0ire @derrabus @beberlei ?

stof:
> Notice that we can still add custom typed field mapping in the config, this would be passed to the `TypedFieldMapper` using its `addTypedFieldMapping` function. The mapper then decides whatever to do with this information.

this requires the TypedFieldMapper to be a mutable object, preventing sharing the same instance between multiple configuration objects (as that would merge the mappings of each configuration). This looks like a bad design to me.

I would rather keep this simple mapping as a constructor argument of the DefaultTypedFieldMapper instead (and custom mappers are free to implement their config as they want)

michnovka:
> this requires the TypedFieldMapper to be a mutable object, preventing sharing the same instance between multiple configuration objects (as that would merge the mappings of each configuration). This looks like a bad design to me.

The main reason I wanted to implement is this way is that I still want to allow config like this:
```yaml
doctrine:
    orm:
        typed_field_mapping:
            App\Objects\CustomObject: App\DBAL\Types\CustomTypeFQCN
            App\Objects\CustomObject2: custom_type2_using_name_from_dbal
```
where the default type mapper is used, but a list is provided for some additional mappings.

I dont understand why mutability of the TypedFieldMapper is important, one instance will be created and passed into `Configuration` object, then this will be passed to `ClassMetadataInfo` from the `ClassMetadataFactory` in most cases. I do not expect multiple TypedFieldMappers to be used in a project, unless you are working with `ClassMetadataInfo` object directly. Most people will just use the DoctrineBundle and specify one TypedFieldMapper if they want some advanced type mapping logic, but even more of the people will likely just use the `DefaultTypedFieldMapper` and only specify few additional types explicitly.

michnovka:
@stof ok, I think I get your point now, so how about this:

```php
<?php

declare(strict_types=1);

namespace ORM\Mapping;

use Doctrine\DBAL\Types\Type;
use ReflectionProperty;

/** @psalm-import-type ScalarName from ClassMetadataInfo */
interface TypedFieldMapper
{
    /** @param array<class-string|ScalarName, class-string<Type>|string> $typedFieldMappings */
    public function __construct(array $typedFieldMappings = []);

    /**
     * Validates & completes the given field mapping based on typed property.
     *
     * @param  array{fieldName: string, type?: mixed} $mapping The field mapping to validate & complete.
     *
     * @return array{fieldName: string, enumType?: string, type?: mixed} The updated mapping.
     */
    public function validateAndCompleteTypedFieldMapping(array $mapping, ReflectionProperty $field): array;
}
```

stof:
@michnovka I would not enforce the signature of the constructor. Custom mapper will legitimately have different configuration needs.
And the Configuration class should take an instance anyway so it won't care about the constructor signature.

stof:
@michnovka don't only think about how things look like in DoctrineBundle. The doctrine/orm project is usable in other contexts as well. So the design of the API must make sense on its own.

michnovka:
@stof ok, so can you please suggest what you meant with the mutability? I dont see how you want to share the instance between multiple configurations AND at the same time have different behavior. Nothing prevents you to make more instances of your CustomTypedFieldMapper and use them in different configuration scenarios with different default typedFieldMapping settings. So if you see some issue in this, can you please give me a concrete example so that I can understand the problem and think of a solution?

michnovka:
@stof the problem is, that I want to keep an option to only specify the additional `typed_field_mappings`, whilst using the `DefaultTypedFieldMapper`. (or any other `TypedFieldMapper`). It has a real-use case where you want some logic to be implemented by the `CustomTypedFieldMapper`, but at the same time maybe explicitly type some mappings. I can either pass this in constructor, or in some other method. 

If I pass it in constructor, the object CAN be immutable (but no way to force this with PHP, somebody can make their own `TypedFieldMapper` implementation which will ignore this requirement and be mutable). or

If I pass it in some additonal method, I can do this:
```php
    /**
     * @param array<class-string|ScalarName, class-string<Type>|string> $typedFieldMappings
     */
    public function setTypedFieldMappings(array $typedFieldMappings = []): self;
```
Which would return a new instance always, thus imitate immutability (like `DateTimeImmutable`).

But I think we should not attempt to force immutability. For all we know, mutability can be desired under some scenario. If somebody wants to have 2 different configs for one `TypedFieldMapper`, then they should create 2 separate instances, not share them.

Like I said, we cannot enforce that any implementation of `TypedFieldMapper` interface wont break the requirement of immutability (by adding some modifying setter), there is no `immutable` keyword for the interface which might enforce such desired behavior.

For that and the fact that some scenario might benefit from mutability, I dont want to satisfy this requirement.

Thoughts?

stof:
@michnovka to me, this shows a need for a ChainTypedFieldMapper allowing to combine multiple mappers to use them together.

> But I think we should not attempt to force immutability.

I don't want to enforce immutability. I want to avoid **requiring mutability**. that's not the same.

michnovka:
@stof and what about this:

```php
<?php

declare(strict_types=1);

namespace ORM\Mapping;

use ReflectionProperty;

interface TypedFieldMapper
{
    /**
     * Validates & completes the given field mapping based on typed property.
     *
     * @param  array{fieldName: string, type?: mixed} $mapping The field mapping to validate & complete.
     *
     * @return array{fieldName: string, enumType?: string, type?: mixed} The updated mapping.
     */
    public function validateAndCompleteTypedFieldMapping(array $mapping, ReflectionProperty $field): array;
}
```

this does not require mutability

then the yaml config can look like

```yaml
doctrine:
    orm:
        typed_field_mapper: 'default'
            arguments:
                $typedFieldMappings:
                    - App\Objects\CustomObject: App\DBAL\Types\CustomTypeFQCN
                    - App\Objects\CustomObject2: custom_type2_using_name_from_dbal
```
or
```yaml
doctrine:
    orm:
        typed_field_mapper: App\ORM\CustomTypedFieldMapper
            arguments:
                $typedFieldMappings:
                    - App\Objects\CustomObject: App\DBAL\Types\CustomTypeFQCN
                    - App\Objects\CustomObject2: custom_type2_using_name_from_dbal
                $someSecondArgumentOfConstructor: true
```
or
```yaml
doctrine:
    orm:
        typed_field_mapper: App\ORM\CustomTypedFieldMapper2
            arguments:
                $totallyDifferentConstructorArgument: false
```

The doctrine-bundle syntax is not as nice as originally proposed, but it does allow for greater flexibility. Not all `TypedFieldMappers` will require the `$typedFieldMappings` argument, so it makes no sense to force it on all (Ive realized this thanks to you @stof ).

And I can also make a class like

```php
class ChainTypedFieldMapper implements TypedFieldMapper
{

    public function __construct(TypedFieldMapper ...$typedFieldMappers){
        // ... store them in some collection and process them all accordingly with validateAndCompleteTypedFieldMapping
    }

    // ...
```
which could have this support in doctrine bundle:
```yaml
doctrine:
    orm:
        typed_field_mapper: 'chain'
            arguments:
                $typedFieldMappers:
                    - 'default':
                        arguments:
                            $typedFieldMappings:
                                - App\Objects\CustomObject: App\DBAL\Types\CustomTypeFQCN
                                - App\Objects\CustomObject2: custom_type2_using_name_from_dbal
                    - App\ORM\CustomTypedFieldMapper2:
                        arguments:
                            $totallyDifferentConstructorArgument: false
```

stof:
you are over-thinking the DoctrineBundle configuration syntax. A custom mapper will be defined as a service seperately and will only be referenced in the config (or registered through a tag applied on it depending on the final solution). But anyway, that's a discussion to have in DoctrineBundle after we know what the ORM API is.

michnovka:
@stof regardless of the doctrinebundle config, the other things I proposed in the above comment address your feedback completely, so do you agree with that proposal?

michnovka:
I can no longer vote to merge this PR, since a far superior version has implemented here - https://github.com/doctrine/orm/pull/10313. We will continue discussion there

PR #10296, opened 2022-12-12
Bump coding standard to v11 (3.0 edition)

derrabus:
#10295 on 3.0.x

### DIFF
--- a/tests/Doctrine/Tests/Models/Cache/City.php
+++ b/tests/Doctrine/Tests/Models/Cache/City.php
@@ -9,11 +9,15 @@ use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
 use Doctrine\ORM\Mapping\ClassMetadata;
 
-#[ORM\Entity, ORM\Table(name: 'cache_city'), ORM\Cache]
+#[ORM\Entity]
+#[ORM\Table(name: 'cache_city')]
+#[ORM\Cache]
 class City
 {
     /** @var int */
-    #[ORM\Id, ORM\GeneratedValue, ORM\Column(type: 'integer')]
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: 'integer')]
     protected $id;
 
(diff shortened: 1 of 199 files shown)


tests/Doctrine/Tests/Models/Cache/City.php
```diff
@@ -9,25 +9,37 @@
...
-        #[ORM\Column(unique: true)] protected string $name,
-        #[ORM\Cache] #[ORM\ManyToOne(targetEntity: 'State', inversedBy: 'cities')] #[ORM\JoinColumn(name: 'state_id', referencedColumnName: 'id')] protected State|null $state = null,
+        #[ORM\Column(unique: true)]
+        protected string $name,
+        #[ORM\Cache]
+        #[ORM\ManyToOne(targetEntity: 'State', inversedBy: 'cities')]
+        #[ORM\JoinColumn(name: 'state_id', referencedColumnName: 'id')]
+        #[ORM\Cache]
+        #[ORM\ManyToOne(targetEntity: 'State', inversedBy: 'cities')]
+        #[ORM\Cache]
```
greg0ire: Something weird happened here.

derrabus: Indeed!

simPod: I think it might be a bug in Slevomat's sniff fixer where multiple sniffs collided somehow (cc @kukulich). Is it blocking for us?

kukulich: 👀

derrabus: > Is it blocking for us?

I think, I'd like to see this resolved before we release the new version. Otherwise I would need to review quite a lot of files and watch out for this kind of error.

kukulich: I've just released Slevomat CS 8.7.0 - it should be resolved there.

derrabus: Looks good, thanks a lot!

PR #10305, opened 2022-12-14
Clean-up code after PR10288 merge

michnovka:
Leverage typed system better in `AttributeDriver`, add psalm-type for `DiscriminatorColumn::enumType` and remove psalm-baseline ignored `InvalidArgument`

### DIFF
--- a/lib/Doctrine/ORM/Mapping/DiscriminatorColumn.php
+++ b/lib/Doctrine/ORM/Mapping/DiscriminatorColumn.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\ORM\Mapping;
 
 use Attribute;
+use BackedEnum;
 
 #[Attribute(Attribute::TARGET_CLASS)]
 final class DiscriminatorColumn implements MappingAttribute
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Mapping/DiscriminatorColumn.php
```diff
@@ -14,6 +14,7 @@ public function __construct(
         public readonly string|null $type = null,
         public readonly int|null $length = null,
         public readonly string|null $columnDefinition = null,
+        /** @var class-string<\BackedEnum>|null */
```
greg0ire: Can you add a backed statement for `BackedEnum` instead? I think the cs should catch that, and might at some point in the future.

michnovka: sorry, I dont understand what you mean

greg0ire: I meant a use statement sorry

greg0ire (approved):

derrabus (approved):

Issue #10310, opened 2022-12-15
Apcu cache is used even if it is not enabled


aleksejs1:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.13.x

#### Summary

In case of apcu extension exists, but not enabled Doctrine/ORM/Tools::createCacheInstance still creates ApcuCache or ApcuAdapter.
When someone try to write the cache "Failed to save key xxx of type Doctrine\ORM\Mapping\ClassMetadata." error appears.

#### Current behavior

When someone try to write the cache "Failed to save key xxx of type Doctrine\ORM\Mapping\ClassMetadata." error appears.

#### How to reproduce

Just an example:
Make the environment with existed disabled apcu extension.
Create doctrine migration with entityManager dependency.
Run doctrine:migration:status in cli.
Got "Failed to save key xxx of type Doctrine\ORM\Mapping\ClassMetadata." error.

#### Solution options

Change if statement in https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/Tools/Setup.php#L239
from `} elseif (extension_loaded('apcu')) {`
to `} elseif (extension_loaded('apcu') && apcu_enabled()) {`
or `} elseif (extension_loaded('apcu') && ini_get('apc.enabled')) {`

derrabus:
The proposed solution looks good to me, PR welcome.

mpdude:
Fixed by #10311, right?

PR #10311, opened 2022-12-15
add apcu enabled check if apcu extension loaded (#10310)

aleksejs1:


derrabus:
Can you take a look at the CI failures?

aleksejs1:
I have no idea why is CI fails... apcu enabled in setup, but looks like apcu_enabled() return false.
If ini_get will return the same, then need some more time for investigation

lib/Doctrine/ORM/ORMSetup.php
```diff
@@ -181,7 +182,7 @@ private static function createCacheInstance(
 
         $namespace = 'dc2_' . md5($proxyDir);
 
-        if (extension_loaded('apcu')) {
+        if (extension_loaded('apcu') && ini_get('apc.enabled')) {
```
greg0ire:
```suggestion
        if (extension_loaded('apcu') && apcu_enabled()) {
```

lib/Doctrine/ORM/Tools/Setup.php
```diff
@@ -236,7 +237,7 @@ private static function createCacheInstance(bool $isDevMode, ?Cache $cache): Cac
 
         if ($isDevMode === true) {
             $cache = class_exists(ArrayCache::class) ? new ArrayCache() : new ArrayAdapter();
-        } elseif (extension_loaded('apcu')) {
+        } elseif (extension_loaded('apcu') && ini_get('apc.enabled')) {
```
```suggestion
        } elseif (extension_loaded('apcu') && apcu_enabled()) {
```

greg0ire:
It makes little sense to enable APCu in the CLI. Because of that, [there is another setting](https://www.php.net/manual/en/apcu.configuration.php#ini.apcu.enable-cli) if you want to do it anyway.

I think you should set that setting in the failing tests, because it does not make much sense to use `AcpuAdapter` if that cache system is not actually available with the CLI.

aleksejs1:
> I think you should set that setting in the failing tests

Sorry, but how to enable apcu-cli in failing tests? Should it done in ci-configs, or in test annotation?

greg0ire:
That setting is `PHP_INI_SYSTEM`, so I'm not sure it can be done in test annotations. In fact, I must confess I didn't know this was doable with test annotations, can you link to the relevant PHPUnit docs?

Anyway, I think I would use https://github.com/shivammathur/setup-php#ini-values-optional in this file: https://github.com/doctrine/orm/blob/2.13.x/.github/workflows/continuous-integration.yml

.github/workflows/continuous-integration.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           php-version: "${{ matrix.php-version }}"
           extensions: "apcu, pdo, ${{ matrix.extension }}"
           coverage: "pcov"
-          ini-values: "zend.assertions=1"
+          ini-values: "zend.assertions=1, apc.enable_cli=1"
```
greg0ire, line 69: Look for more occurrences of `zend.assertions=1` in the file to see where you might need to add this.

derrabus (approved):

PR #10313, opened 2022-12-16
Add TypedFieldMapper for automatic mapping of typed PHP fields to DBAL types

michnovka:
This is an alternative implementation of https://github.com/doctrine/orm/pull/10290, fixes https://github.com/doctrine/orm/issues/9561

Previously, only a predefined set of automatic mappings was allowed for as `array`, `bool`, `int`, `float`, `string`, `DateTime`, `DateTimeImmutable` and `DateInterval`.

With this extension, it is possible to supply custom `TypedFieldMapper` implementation. Its `validateAndComplete` takes as parameter the ReflectionProperty of a given field and decides the appropriate mapping.

A new configuration option was added to set and get the `TypedFieldMapper`.

The old logic was moved into a class `DefaultTypedFieldMapper` which is used by default when no mapper is supplied.

The selected `TypedFieldMapper` is passed into `ClassMetadataInfo`'s constructor. If empty, the `DefaultTypedFieldMapper` is used.

There is also `ChainTypedFieldMapper` class which allows chaining multiple `TypedFieldMappers` and apply them in a cascade (always if a field gets type assigned by the earlier mapper in the list, it will not be changed later).

Documentation has been updated to explain the new features. Tests were added to cover the new classes as well as `ClassMetadataInfo` and all mapping drivers.

### DIFF
--- a/docs/en/reference/typedfieldmapper.rst
+++ b/docs/en/reference/typedfieldmapper.rst
@@ -1,0 +1,176 @@
+Implementing a TypedFieldMapper
+===============================
+
+.. versionadded:: 2.14
+
+You can specify custom typed field mapping between PHP type and DBAL type using ``Configuration``
+and a custom ``Doctrine\ORM\Mapping\TypedFieldMapper`` implementation.
+
+.. code-block:: php
+
+    <?php
+
+    $configuration->setTypedFieldMapper(new CustomTypedFieldMapper());
+
+
+DefaultTypedFieldMapper
+-----------------------
+
+By default the ``Doctrine\ORM\Mapping\DefaultTypedFieldMapper`` is used, and you can pass an array of
+PHP type => DBAL type mappings into its constructor to override the default behavior or add new mappings.
+
+.. code-block:: php
+
+    <?php
+    use App\CustomIds\CustomIdObject;
+    use App\DBAL\Type\CustomIdObjectType;
+
+    $configuration->setTypedFieldMapper(new DefaultTypedFieldMapper([
+        CustomIdObject::class => CustomIdObjectType::class,
+    ]));
+
+Then, an entity using the ``CustomIdObject`` typed field will be correctly assigned its DBAL type
+(``CustomIdObjectType``) without the need of explicit declaration.
+
+.. configuration-block::
+
+    .. code-block:: attribute
+
+        <?php
+        #[ORM\Entity]
+        #[ORM\Table(name: 'cms_users_typed_with_custom_typed_field')]
+        class UserTypedWithCustomTypedField
+        {
+            #[ORM\Column]
+            public CustomIdObject $customId;
+
+            // ...
+        }
+
+    .. code-block:: annotation
+
+        <?php
+        /**
+         * @Entity
+         * @Table(name="cms_users_typed_with_custom_typed_field")
+         */
+        class UserTypedWithCustomTypedField
+        {
+            /** @Column */
+            public CustomIdObject $customId;
+
+            // ...
+        }
+
+    .. code-block:: xml
+
+        <doctrine-mapping>
+          <entity name="UserTypedWithCustomTypedField">
+            <field name="customId"/>
+            <!-- -->
+          </entity>
+        </doctrine-mapping>
+
+    .. code-block:: yaml
+
+        UserTypedWithCustomTypedField:
+          type: entity
+          fields:
+            customId: ~
+
+It is perfectly valid to override even the "automatic" mapping rules mentioned above:
+
+.. code-block:: php
+
+    <?php
+    use App\DBAL\Type\CustomIntType;
+
+    $configuration->setTypedFieldMapper(new DefaultTypedFieldMapper([
+        'int' => CustomIntType::class,
+    ]));
+
+.. note::
+
+    If chained, once the first ``TypedFieldMapper`` assigns a type to a field, the ``DefaultTypedFieldMapper`` will
+    ignore its mapping and not override it anymore (if it is later in the chain). See below for chaining type mappers.
+
+
+TypedFieldMapper interface
+-------------------------
+The interface ``Doctrine\ORM\Mapping\TypedFieldMapper`` allows you to implement your own
+typed field mapping logic. It consists of just one function
+
+
+.. code-block:: php
+
+    <?php
+    /**
+     * Validates & completes the given field mapping based on typed property.
+     *
+     * @param array{fieldName: string, enumType?: string, type?: mixed}  $mapping The field mapping to validate & complete.
+     * @param \ReflectionProperty                                        $field
+     *
+     * @return array{fieldName: string, enumType?: string, type?: mixed} The updated mapping.
+     */
+    public function validateAndComplete(array $mapping, ReflectionProperty $field): array;
+
+
+ChainTypedFieldMapper
+---------------------
+
+The class ``Doctrine\ORM\Mapping\ChainTypedFieldMapper`` allows you to chain multiple ``TypedFieldMapper`` instances.
+When being evaluated, the ``TypedFieldMapper::validateAndComplete`` is called in the order in which
+the instances were supplied to the ``ChainTypedFieldMapper`` constructor.
+
+.. code-block:: php
+
+    <?php
+    use App\DBAL\Type\CustomIntType;
+
+    $configuration->setTypedFieldMapper(
+        new ChainTypedFieldMapper(
+            DefaultTypedFieldMapper(['int' => CustomIntType::class,]),
+            new CustomTypedFieldMapper()
+        )
+    );
+
+
+Implementing a TypedFieldMapper
+-------------------------------
+
+If you want to assign all ``BackedEnum`` fields to your custom ``BackedEnumDBALType`` or you want to use different
+DBAL types based on whether the entity field is nullable or not, you can achieve this by implementing your own
+typed field mapper.
+
+You need to create a class which implements ``Doctrine\ORM\Mapping\TypedFieldMapper``.
+
+.. code-block:: php
+
+    <?php
+    final class CustomEnumTypedFieldMapper implements TypedFieldMapper
+    {
+        /**
+         * {@inheritdoc}
+         */
+        public function validateAndComplete(array $mapping, ReflectionProperty $field): array
+        {
+            $type = $field->getType();
+
+            if (
+                ! isset($mapping['type'])
+                && ($type instanceof ReflectionNamedType)
+            ) {
+                if (! $type->isBuiltin() && enum_exists($type->getName())) {
+                    $mapping['type'] = BackedEnumDBALType::class;
+                }
+            }
+
+            return $mapping;
+        }
+    }
+
+.. note::
+
+    Note that this case checks whether the mapping is already assigned, and if yes, it skips it. This is up to your
+    implementation. You can make a "greedy" mapper which will always override the mapping with its own type, or one
+    that behaves like ``DefaultTypedFieldMapper`` and does not modify the type once its set prior in the chain.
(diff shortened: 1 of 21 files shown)


michnovka:
@stof Please have a look and let me know your opinion. @beberlei @greg0ire and @derrabus also welcome your feedback

docs/en/reference/typedfieldmapper.rst
```diff
@@ -0,0 +1,176 @@
...
+    use App\DBAL\Type\CustomIntType;
+
+    $configuration->setTypedFieldMapper(new DefaultTypedFieldMapper([
+        'int' => CustomIntType::class,
+    ]));
+
+.. note::
+
+    If chained, once the first ``TypedFieldMapper`` assigns a type to a field, the ``DefaultTypedFieldMapper`` will
+    ignore its mapping and not override it anymore (if its later in the chain). See below for chaining type mappers.
```
greg0ire:
```suggestion
    ignore its mapping and not override it anymore (if it is later in the chain). See below for chaining type mappers.
```

docs/en/reference/typedfieldmapper.rst
```diff
@@ -0,0 +1,176 @@
...
+    <?php
+    /**
+     * Validates & completes the given field mapping based on typed property.
+     *
+     * @param array{fieldName: string, enumType?: string, type?: mixed}  $mapping The field mapping to validate & complete.
+     * @param \ReflectionProperty                                        $field
+     *
+     * @return array{fieldName: string, enumType?: string, type?: mixed} The updated mapping.
+     */
+    public function validateAndCompleteTypedFieldMapping($mapping, $field): array;
```
greg0ire: Why not use native type declarations here?

lib/Doctrine/ORM/Configuration.php
```diff
@@ -63,6 +65,7 @@
  * Internal note: When adding a new configuration option just write a getter/setter pair.
  *
  * @psalm-import-type AutogenerateMode from ProxyFactory
+ * @psalm-import-type ScalarName from ClassMetadataInfo
```
greg0ire: You're not using this are you? Same below.

docs/en/reference/typedfieldmapper.rst
greg0ire: Also, maybe this could be shortened to `validateAndComplete()`? The classname already contains the rest.

tests/Doctrine/Tests/ORM/Mapping/TypedFieldMapperTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+    private static function chainTypedFieldMapper(): ChainTypedFieldMapper
+    {
+        return new ChainTypedFieldMapper(self::customTypedFieldMapper(), self::defaultTypedFieldMapper());
+    }
+
+    /**
+     * Data Provider for NamingStrategy#classToTableName
+     *
+     * @return array<array{TypedFieldMapper, ReflectionClass, array{fieldName: string, enumType?: string, type?: mixed}, array{fieldName: string, enumType?: string, type?: mixed}}>
```
greg0ire: Please use linebreaks here.

lib/Doctrine/ORM/Mapping/ChainTypedFieldMapper.php
```diff
@@ -0,0 +1,31 @@
...
+     */
+    public function validateAndCompleteTypedFieldMapping($mapping, $field): array
+    {
+        foreach ($this->typedFieldMappers as $typedFieldMapper) {
+            $mapping = $typedFieldMapper->validateAndCompleteTypedFieldMapping($mapping, $field);
+        }
+
+        return $mapping;
+    }
+}
```
greg0ire, line 33: Please add a test for this class.

greg0ire (dismissed):

lib/Doctrine/ORM/Mapping/ChainTypedFieldMapper.php
michnovka: There is a test for this class already in `tests/Doctrine/Tests/ORM/Mapping/TypedFieldMapperTest.php` I am not sure why codecov is picking it up as not tested...

michnovka: ah, is it possible that codecov uses PHP < 7.4? As the tests are for PHP 7.4+ (doesnt have sense to test this with PHP version not supporting typed properties)

michnovka: Ive confirmed in debugger the test is being run on PHP7.4+

greg0ire: Codecov is supposed to be used on every job. As you can see, it's no longer complaining for some reason :thinking:

docs/en/reference/typedfieldmapper.rst
```diff
@@ -0,0 +1,176 @@
...
+    <?php
+    /**
+     * Validates & completes the given field mapping based on typed property.
+     *
+     * @param array{fieldName: string, enumType?: string, type?: mixed}  $mapping The field mapping to validate & complete.
+     * @param \ReflectionProperty                                        $field
+     *
+     * @return array{fieldName: string, enumType?: string, type?: mixed} The updated mapping.
+     */
+    public function validateAndComplete($mapping, $field): array;
```
greg0ire: Why not use native parameter type declarations?

lib/Doctrine/ORM/Mapping/ChainTypedFieldMapper.php
michnovka: it is still

docs/en/reference/typedfieldmapper.rst
michnovka: I was getting tons of failed tests when I was, due to 7.1,7.2 and 7.3 CI

michnovka: the reason is likely that no matter what PHP version, the `DefaultTypedFieldMapper` needs to be initialized, and if this one uses PHP7.4 features (typed params) it will cause syntax errors on PHP <7.4.
The workaround would be to NOT initialize the DefaultTypedFieldMapper on PHP < 7.4 inside` ClassMetadataInfo` and instead assign `null`.
I am fine with keeping it not-typed tho for now and I will fix it in 3.0.x (unless its fixed automatically)

michnovka: Ive pushed an attempt to fix this, let me see what i can do

greg0ire: typed parameters are not a PHP 7.4 feature, typed properties are…

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -800,19 +795,26 @@ class ClassMetadataInfo implements ClassMetadata
...
-    public function __construct($entityName, ?NamingStrategy $namingStrategy = null)
+    public function __construct($entityName, ?NamingStrategy $namingStrategy = null, ?TypedFieldMapper $typedFieldMapper = null)
     {
         $this->name           = $entityName;
         $this->rootEntityName = $entityName;
         $this->namingStrategy = $namingStrategy ?: new DefaultNamingStrategy();
         $this->instantiator   = new Instantiator();
+
+        if (PHP_VERSION_ID >= 70400) {
+            $this->typedFieldMapper = $typedFieldMapper ?: new DefaultTypedFieldMapper();
```
```suggestion
            $this->typedFieldMapper = $typedFieldMapper ?? new DefaultTypedFieldMapper();
```

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
```diff
@@ -0,0 +1,74 @@
...
+            $defaultTypedFieldMappings = [
+                DateInterval::class => Types::DATEINTERVAL,
+                DateTime::class => Types::DATETIME_MUTABLE,
+                DateTimeImmutable::class => Types::DATETIME_IMMUTABLE,
+                'array' => Types::JSON,
+                'bool' => Types::BOOLEAN,
+                'float' => Types::FLOAT,
+                'int' => Types::INTEGER,
+                'string' => Types::STRING,
+            ];
```
derrabus: `private const` ?

docs/en/reference/typedfieldmapper.rst
michnovka: my bad. fixed

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
```diff
@@ -0,0 +1,74 @@
...
+
+            $typedFieldMappings = array_merge($defaultTypedFieldMappings, $this->typedFieldMappings);
```
derrabus: `array_merge()` is an expensive operation and `validateAndComplete()` will be called quite a few times. Given that we will always merge the same two arrays here, wouldn't it be better to perform this operation once and store the result?

tests/Doctrine/Tests/ORM/Mapping/MappingDriverTestCase.php
```diff
@@ -293,6 +305,28 @@ public function testFieldTypeFromReflection(): void
         self::assertEquals(Contact::class, $class->embeddedClasses['contact']['class']);
     }
 
+    /** @group GH10313 */
+    public function testCustomFieldTypeFromReflection(): void
+    {
+        if (PHP_VERSION_ID < 70400) {
+            self::markTestSkipped('requires PHP 7.4');
+        }
```
derrabus: Please use an annotation for that.

derrabus (changes requested):

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
michnovka: thisis used a bit above also

```php
        $this->namingStrategy = $namingStrategy ?: new DefaultNamingStrategy();
```

Do you want me to change that too?

derrabus: Yes please.

michnovka: done already, all feedback incorporated

lib/Doctrine/ORM/Mapping/ChainTypedFieldMapper.php
michnovka: @derrabus any idea why codecov is complaining here? The test is present, it is executed, the code had to run

michnovka: well now its complaining only about lines with brackets.. I dont get it. If there is anything I can do to improve this, let me know. But the ChainTypedFieldMapper is tested from 2 places - TypedFieldMapperTest and ClassMetadataTest.

greg0ire: Indeed, it's just the brackets, so no big deal.

greg0ire (approved):

beberlei (approved):

derrabus (approved):
LGTM

docs/en/reference/typedfieldmapper.rst
```diff
@@ -0,0 +1,176 @@
...
+the instances were supplied to the ``ChainTypedFieldMapper`` constructor.
+
+.. code-block:: php
+
+    <?php
+    use App\DBAL\Type\CustomIntType;
+
+    $configuration->setTypedFieldMapper(
+        new ChainTypedFieldMapper(
+            DefaultTypedFieldMapper(['int' => CustomIntType::class,]),
```
andersonamuller, line 132: missing 
```php
new DefaultTypedFieldMapper(['int' => CustomIntType::class,]),
```

michnovka: Ive already submitted PR to fix this https://github.com/doctrine/orm/pull/10327

PR #10315, opened 2022-12-18
Document stdClass structures used by `CommitOrderCalculator`

derrabus:
This PR fixes some static analysis errors that popped up while I attempted to upgrade to Psalm 5.3. `CommitOrderCalculator` internally uses undocumented object structures which I've replaced with actual classes.

### DIFF
--- a/lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
+++ b/lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
@@ -4,7 +4,10 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Internal;
 
-use stdClass;
+use Doctrine\ORM\Internal\CommitOrder\Edge;
+use Doctrine\ORM\Internal\CommitOrder\Vertex;
+use Doctrine\ORM\Internal\CommitOrder\VertexState;
+use Doctrine\ORM\Mapping\ClassMetadata;
 
 use function array_reverse;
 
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
```diff
@@ -17,33 +20,28 @@
...
-    public const IN_PROGRESS = 1;
-    public const VISITED     = 2;
+    /** @deprecated */
+    public const NOT_VISITED = VertexState::NOT_VISITED;
+
+    /** @deprecated */
+    public const IN_PROGRESS = VertexState::IN_PROGRESS;
+
+    /** @deprecated */
+    public const VISITED = VertexState::VISITED;
```
greg0ire, line 30: These should be mentioned in `UPGRADE.md`

derrabus: I wasn't sure if I even need a deprecation layer for them since this class is located in a namespace that literally says `Internal`.

derrabus: I've added a note to UPGRADE.md.

greg0ire: Ah sorry, I didn't spot this. Wasn't required indeed.

greg0ire (dismissed):

PR #10320, opened 2022-12-19
Support doctrine/annotations 2

derrabus:
WIP until doctrine/annotations 2 has been released.

This PR allows the installation of Annotations 2 and fixes all failing tests.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -78,6 +78,9 @@ jobs:
         run: "composer require doctrine/dbal ^${{ matrix.dbal-version }} --no-update"
         if: "${{ matrix.dbal-version != 'default' }}"
 
+      - name: "Uninstall PHPBench"
+        run: "composer remove --dev --no-update phpbench/phpbench"
+
       - name: "Install dependencies with Composer"
         uses: "ramsey/composer-install@v2"
         with:
@@ -153,6 +156,9 @@ jobs:
         run: "composer require doctrine/dbal ^${{ matrix.dbal-version }} --no-update"
         if: "${{ matrix.dbal-version != 'default' }}"
 
+      - name: "Uninstall PHPBench"
+        run: "composer remove --dev --no-update phpbench/phpbench"
+
       - name: "Install dependencies with Composer"
         uses: "ramsey/composer-install@v2"
         with:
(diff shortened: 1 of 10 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -78,6 +78,9 @@ jobs:
         run: "composer require doctrine/dbal ^${{ matrix.dbal-version }} --no-update"
         if: "${{ matrix.dbal-version != 'default' }}"
 
+      - name: "Uninstall PHPBench"
+        run: "composer remove --dev --no-update phpbench/phpbench"
```
greg0ire, line 82: I'm looking into adding support for v2 to phpbench

greg0ire: https://github.com/phpbench/phpbench/pull/1013

I ran the tests locally, and apart from an unrelated failure, all is well.

lib/Doctrine/ORM/Configuration.php
```diff
@@ -182,17 +183,26 @@ public function newDefaultAnnotationDriver($paths = [], $useSimpleAnnotationRead
             );
         }
 
-        AnnotationRegistry::registerFile(__DIR__ . '/Mapping/Driver/DoctrineAnnotations.php');
+        if (method_exists(AnnotationRegistry::class, 'registerFile')) {
+            AnnotationRegistry::registerFile(__DIR__ . '/Mapping/Driver/DoctrineAnnotations.php');
+        }
```
greg0ire: This line could be removed instead, couldn't it? https://github.com/doctrine/annotations/pull/271 ?

.github/workflows/continuous-integration.yml
derrabus: I think, we can live with this hack in our CI until PHPBench has merged and released your PR.

lib/Doctrine/ORM/Configuration.php
derrabus: I guess we could've removed it earlier already. 🤔

greg0ire: Did a separate PR for that: https://github.com/doctrine/orm/pull/10321

lib/Doctrine/ORM/Configuration.php
```diff
@@ -182,14 +183,21 @@ public function newDefaultAnnotationDriver($paths = [], $useSimpleAnnotationRead
         }
 
         if ($useSimpleAnnotationReader) {
```
greg0ire, line 185: To be deprecated later on I suppose.

greg0ire (approved):

lib/Doctrine/ORM/Configuration.php
derrabus: The whole method is deprecated already and its replacement does not have that parameter anymore.

PR #10328, opened 2022-12-20
Require dev version of phpbench

greg0ire:
It is important to have the same version of all dependencies in dev and in the CI, otherwise it makes it hard to have the right static analysis baseline for every environment.


greg0ire:
@dantleech we are looking forward to a release. No pressure though, as you can see, we're fine without it :slightly_smiling_face:

derrabus (approved):

PR #10331, opened 2022-12-20
Merge 2.15.x up into 3.0.x

greg0ire:


.github/workflows/static-analysis.yml
```diff
@@ -84,11 +82,18 @@ jobs:
         run: "composer require doctrine/dbal ^${{ matrix.dbal-version }} --no-update"
         if: "${{ matrix.dbal-version != 'default' }}"
 
+<<<<<<< HEAD
```
derrabus: Unresolved conflict?

derrabus (changes requested):

.github/workflows/static-analysis.yml
greg0ire: Good catch! :sweat_smile:

derrabus (approved):

PR #10329, opened 2022-12-20
Drop doctrine/lexer 1

greg0ire:
This is a backport of #10309 , except I'm not migrating to an enum for obvious reasons.

PR #10332, opened 2022-12-20
Drop doctrine/lexer 2

greg0ire:
This allows us to have a totally typed lexer

### DIFF
--- a/lib/Doctrine/ORM/Query/Lexer.php
+++ b/lib/Doctrine/ORM/Query/Lexer.php
@@ -56,8 +56,7 @@ class Lexer extends AbstractLexer
         return ['\s+', '--.*', '(.)'];
     }
 
-    /** @param string $value */
-    protected function getType(&$value): TokenType
+    protected function getType(string &$value): TokenType
     {
         $type = TokenType::T_NONE;
 
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "doctrine/event-manager": "^1.2 || ^2",
         "doctrine/inflector": "^1.4 || ^2.0",
         "doctrine/instantiator": "^1.3 || ^2",
-        "doctrine/lexer": "^2.1 || ^3",
+        "doctrine/lexer": "^3",
         "doctrine/persistence": "^3.1.1",
         "psr/cache": "^1 || ^2 || ^3",
         "symfony/console": "^5.4 || ^6.0"


derrabus:
Do we _have to_ drop Lexer 2 for this change or could we do the same with `^2 || ^3`?

greg0ire:
I thought like you and I've backported a lot of it to #10329

lib/Doctrine/ORM/Query/Lexer.php
```diff
@@ -56,10 +56,7 @@ protected function getNonCatchablePatterns(): array
         return ['\s+', '--.*', '(.)'];
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    protected function getType(&$value): TokenType
+    protected function getType(string &$value): TokenType
```
greg0ire, line 59: This is the change that we can't do without altering the version constraint

derrabus: Oh, because we didn't add parameter types in v2. That's unfortunate.

greg0ire: If we did that then we would have to bump to 7.2 in order to drop lexer 1 on ORM 2, so all in all, I think things are better like this. I won't pretend I considered that before though 😅

derrabus: Can we remove the `string` type for now, so we can keep `^2 || ^3` as constraint? I'm afraid that strictly requiring v3 might block the adoption of ORM 3 in the beginning. We can still drop v2 at a later time.

greg0ire: I'm AFK for 2 weeks, feel free to open another PR with just the ctor type hint, the rest can be type hinting, and we can decide what to do with this PR at the last minute

SenseException (dismissed):

lib/Doctrine/ORM/Query/Lexer.php
SenseException: I'm in favour of `^3`. Version 2, as far as I understood, had the job to be a transition for such changes and is supposed to be a short lived release.

derrabus: Yes, but Lexer is not only used in the ORM. And I don't want to block the adoption of ORM 3 because of a single parameter type in a single method.

greg0ire: Let's monitor the install stats of v3 once 2.15 is released, and of v2 starting now, it should give some insight about whether this is going to be a big blocker or not.

derrabus:
I've created #10347 which contains all the changes we could agree on. Let's merge #10347 and rebase this PR afterwards. That should give us a very small diff (one parameter type + composer.json). We can leave this PR open until we feel the time is right to drop Lexer 2.

lib/Doctrine/ORM/Query/Lexer.php
greg0ire: 2.15 is not even released yet, and the install stats of v3 are already rising, quite close to the install stats of v2: https://packagist.org/packages/doctrine/lexer/stats

I think this is fine to merge, although we might want to do it right before the release of ORM v3, so that we get more data points.

derrabus (approved):

SenseException (approved):

Issue #10336, opened 2022-12-22
Clearing the em before accessing an uninitialized relation results in "typed property must not be accessed before initialization..." error


kbond:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.14.0

#### Summary

Consider the following code:

```php
$results = $repo->createQueryBuilder('e')
    ->innerJoin('e.relation', 'r')
    ->getQuery()
    ->execute()
;

$em->clear();

echo $results[0]->relation->value;
```

#### Previous behavior

In 2.13.5, would echo the relation's value property.

#### Current behavior

In 2.14.0:

> Error: Typed property Relation::$value must not be accessed before initialization

@nicolas-grekas, could this be related to #10187?

Reproducer: #10337

darthf1:
I have the same exception, but thrown in my tests when a method is called on a proxied object. The proxied object is returned via the [referenceRepository](https://github.com/liip/LiipTestFixturesBundle/blob/2.x/doc/database.md#referencing-fixtures-in-tests-) of [LiipTestFixturesBundle](https://github.com/liip/LiipTestFixturesBundle)

```
1) Tests\Functional\UserInterface\DataCube\Rest\DataCubeProjectionTest::testGetDataCubes
Error: Typed property Domain\DataCube\DataCube::$analysisType must not be accessed before initialization
/builds/xxx/api/src/Domain/DataCube/DataCube.php:101
/builds/xxx/api/var/cache/test/doctrine/orm/Proxies/__CG__DomainDataCubeDataCube.php:210
/builds/xxx/tests/Functional/UserInterface/DataCube/Rest/DataCubeProjectionTest.php:59
```

brilsergei:
I have probably similar problem in tests using the [referenceRepository](https://github.com/liip/LiipTestFixturesBundle/blob/2.x/doc/database.md#referencing-fixtures-in-tests-) of [LiipTestFixturesBundle](https://github.com/liip/LiipTestFixturesBundle). I call method which should return a string (field of an entity) but it returns null. While debugging found that in older version of this lib in method \Doctrine\ORM\UnitOfWork::createEntity the condition `if (isset($this->identityMap[$class->rootEntityName][$idHash])) {` gets true while in 2.14 this condition gets false. identityMap is empty in UnitOfWork object.

mathieudz:
[...]

PR #10337, opened 2022-12-22
Add reproducer for "typed property must not be accessed before initialization..." bug

kbond:
This is a reproducer for #10336.

_The test passes on 2.13.x_

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10336Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10336Test.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\Models\GH10336\GH10336Entity;
+use Doctrine\Tests\Models\GH10336\GH10336Relation;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH10336Test extends OrmFunctionalTestCase
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH10336Test.php
```diff
@@ -0,0 +1,44 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\Models\GH10336\GH10336Entity;
+use Doctrine\Tests\Models\GH10336\GH10336Relation;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @requires PHP 8.0
```
simonberger: PHP 8 is not needed for the test.

nicolas-grekas:
Can you please add back the lines mentioned in https://github.com/doctrine/orm/issues/10336#issuecomment-1368578980 to see the effect on tests?

kbond:
Re-adding the lines fixed this specific test but breaks `MappedSuperclassTest`.

nicolas-grekas:
Great thanks. And the break appears only with the new proxies. Looks like they brought some behavior change that we need to understand.

simonberger:
There is another problem in the `ProxyFactory::generateSkippedProperties` ignoring private properties of parent classes which the id property of the test classes is. I can implement a solution or at least prototype later today.

nicolas-grekas:
Fixed in #10385. This PR can be closed. Thanks for the reproducer!

kbond:
Great! Thanks @nicolas-grekas!

PR #10342, opened 2022-12-26
Add Fully-Qualified class name in UnrecognizedField exception

Kern046:
Currently if we use a field that is not properly registered in our mapping, we end up with the following exception from the `BasicEntityPersister`:

```
Unrecognized field: {field}
```

As with a complete stack trace we may be able to dig enough to find the right mapping, it can take more time to find the concerned entity in a context with less informations, such as logs.

This PR adds the entity FQCN to the exception message to ease debugging in such cases, using the class metadata available in `BasicEntityPersister::class` property. We end up with:

```
Unrecognized field: {fqcn}::{field}
```

As I didn't encountered any tests for the existing `UnrecognizedException` class behavior, I didn't add some but I can if needed.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 2.14
 
+## Deprecated `Doctrine\ORM\Persisters\Exception\UnrecognizedField::byName($field)` method.
+
+It will be removed in 3.0. Use `Doctrine\ORM\Persisters\Exception\UnrecognizedField::byFullyQualifiedName($className, $field)` instead.
+
 ## Deprecated constants of `Doctrine\ORM\Internal\CommitOrderCalculator`
 
 The following public constants have been deprecated:
--- a/lib/Doctrine/ORM/Persisters/Exception/UnrecognizedField.php
+++ b/lib/Doctrine/ORM/Persisters/Exception/UnrecognizedField.php
@@ -10,8 +10,15 @@ use function sprintf;
 
 final class UnrecognizedField extends PersisterException
 {
+    /** @deprecated This method is deprecated and will be removed in Doctrine ORM 3.0. Use {@see byFullyQualifiedName} instead */
     public static function byName(string $field): self
     {
         return new self(sprintf('Unrecognized field: %s', $field));
+    }
+
+    /** @param class-string $className */
+    public static function byFullyQualifiedName(string $className, string $field): self
+    {
+        return new self(sprintf('Unrecognized field: %s::$%s', $className, $field));
     }
 }
--- a/tests/Doctrine/Tests/ORM/Persisters/Exception/UnrecognizedFieldTest.php
+++ b/tests/Doctrine/Tests/ORM/Persisters/Exception/UnrecognizedFieldTest.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Persisters\Exception;
+
+use Doctrine\ORM\Persisters\Exception\UnrecognizedField;
+use Doctrine\Tests\Models\Taxi\Car;
+use PHPUnit\Framework\TestCase;
+
+class UnrecognizedFieldTest extends TestCase
+{
+    public function testByFullyQualifiedName(): void
+    {
+        static::expectException(UnrecognizedField::class);
+        static::expectExceptionMessage('Unrecognized field: Doctrine\Tests\Models\Taxi\Car::$color');
+
+        throw UnrecognizedField::byFullyQualifiedName(Car::class, 'color');
+    }
+}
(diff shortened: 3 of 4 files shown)


lib/Doctrine/ORM/Persisters/Exception/UnrecognizedField.php
```diff
@@ -10,8 +10,8 @@
 
 final class UnrecognizedField extends PersisterException
 {
-    public static function byName(string $field): self
+    public static function byName(string $class, string $field): self
```
greg0ire: While I think this BC break is OK, the name no longer fits. Let's create another method called `byEntityAndName`? Or `byFullyQualifiedName`? Or something else? We would remove the old method in 3.0. Also, let's document the first parameter as `class-string` please.

greg0ire (commented):
Great idea, I think I remember losing time over this a few years ago.

lib/Doctrine/ORM/Persisters/Exception/UnrecognizedField.php
Kern046: done ! Thanks for the feedback :)

greg0ire (dismissed):

SenseException (commented):
As far as I can see were these the only occurrences of `UnrecognizedField::byName()` and this method is now unused. Exceptions' named constructors are for internal usage, but can be used elsewhere in userland. Should it be removed now or set as deprecated first?

greg0ire:
I'm OK with it being removed now.

Kern046:
> I'm OK with it being removed now.

I thought it would be a BC to remove it, some users may use this method. Maybe we should deprecate it instead ?

derrabus (changes requested):
Sorry if I'm nit-picking too much here, but if the exception message is that important to you, I believe we should have a test that makes sure the exception message is as you expect it to be.

derrabus:
> I thought it would be a BC to remove it, some users may use this method. Maybe we should deprecate it instead ?

This PR is filed as a bugfix, so I'd really vote against removing any methods. Yes, I believe we should deprecate the method, but we cannot introduce deprecations in a bugfix release either. So let's do that in a follow-up on 2.15.x.

greg0ire:
Regarding what is a BC and what is not : https://xkcd.com/1172/

I always fear I'm going to be too strict, only to find my fellow maintainers think otherwise. The BC is small, but this isn't just any library, so let's avoid unnecessary risks.

Kern046:
> Sorry if I'm nit-picking too much here, but if the exception message is that important to you, I believe we should have a test that makes sure the exception message is as you expect it to be.

Done :)

> Regarding what is a BC and what is not : https://xkcd.com/1172/
> 
> I always fear I'm going to be too strict, only to find my fellow maintainers think otherwise. The BC is small, but this isn't just any library, so let's avoid unnecessary risks.

I added a deprecation notice with a link to the new method.

lib/Doctrine/ORM/Persisters/Exception/UnrecognizedField.php
```diff
@@ -10,8 +10,15 @@
 
 final class UnrecognizedField extends PersisterException
 {
+    /** @deprecated This method is deprecated and will be removed in Doctrine ORM 3.0. Use {@see byFullyQualifiedName} instead */
```
greg0ire, line 13: Please add a little something to UPGRADE.md

tests/Doctrine/Tests/ORM/Persisters/Exception/UnrecognizedFieldTest.php
```diff
@@ -0,0 +1,20 @@
...
+use Doctrine\ORM\Persisters\Exception\UnrecognizedField;
+use Doctrine\Tests\Models\Taxi\Car;
+use PHPUnit\Framework\TestCase;
+
+class UnrecognizedFieldTest extends TestCase
+{
+    public function testByFullyQualifiedName(): void
+    {
+        static::expectException(UnrecognizedField::class);
+        static::expectExceptionMessage('Unrecognized field: Doctrine\Tests\Models\Taxi\Car::color');
```
derrabus:
```suggestion
        static::expectExceptionMessage('Unrecognized field: Doctrine\Tests\Models\Taxi\Car::$color');
```

I would expect a `$` here since we're talking about a property and not a constant.

derrabus (changes requested):

UPGRADE.md
```diff
@@ -66,6 +66,10 @@ It will be removed in 3.0. Use one of the dedicated event classes instead:
 * `Doctrine\ORM\Event\PostRemoveEventArgs`
 * `Doctrine\ORM\Event\PostLoadEventArgs`
 
+## Deprecated `Doctrine\ORM\Persisters\Exception\UnrecognizedField::byName($field)` method.
+
+It will be removed in 3.0. Use `Doctrine\ORM\Persisters\Exception\UnrecognizedField::byFullyQualifiedName($className, $field)` instead.
+
```
derrabus: We usually add those at the top.

derrabus (approved):

PR #10351, opened 2022-12-30
Allow `doctrine/instantiator` 2

derrabus:
`doctrine/instantiator` 2 adds some type declarations that should not affect us.


composer.json
```diff
@@ -43,7 +43,7 @@
         "doctrine/coding-standard": "^9.0.2 || ^11.0",
         "phpbench/phpbench": "^0.16.10 || ^1.0 || dev-master",
         "phpstan/phpstan": "~1.4.10 || 1.9.4",
-        "phpunit/phpunit": "^7.5 || ^8.5 || ^9.5",
+        "phpunit/phpunit": "^7.5 || ^8.5 || ^9.5.28@dev",
```
derrabus, line 46: PHPUnit uses `doctrine/instantiator` as well and would prevent this installation of v2. We can revert back to stable as soon as sebastianbergmann/phpunit#5126 has been released.

greg0ire (approved):

PR #10353, opened 2022-12-30
Merge 2.15.x into 3.0.x

derrabus:


derrabus:
Looks like I need to fix the build for DBAL 4 first. That's unrelated to the merge-up though.

PR #10354, opened 2022-12-30
Fix build for DBAL 4

derrabus:
This PR fixes the build after the changes of doctrine/dbal#5839.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -1030,9 +1030,11 @@ abstract class AbstractQuery
     protected function getHydrationCacheId(): array
     {
         $parameters = [];
+        $types      = [];
 
         foreach ($this->getParameters() as $parameter) {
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ORM/AbstractQuery.php
```diff
@@ -1044,7 +1046,7 @@ protected function getHydrationCacheId(): array
         ksort($hints);
         assert($queryCacheProfile !== null);
 
-        return $queryCacheProfile->generateCacheKeys($sql, $parameters, $hints);
+        return $queryCacheProfile->generateCacheKeys($sql, $parameters, $types, $hints);
```
derrabus, line 1049: This is actually an interesting find. We've called this method the wrong way the whole time. 🫣

derrabus: Backported as #10355.

greg0ire (approved):

PR #10355, opened 2022-12-30
Include parameter types in hydration cache key generation

derrabus:
https://github.com/doctrine/orm/pull/10354#discussion_r1059527048

We were calling this DBAL method wrong and passed the connection hints as parameter types. I'm not sure if this has really caused issues, but let's call the method as it's meant to be called.


greg0ire (approved):
Oh wow nice catch!

PR #10357, opened 2022-12-31
Shorter deprecation message

derrabus:
Sorry for the follow-up to #10342. 😓 

Everything we deprecate is supposed to be removed in 3.0, so let's keep this kind of message brief.

cc @Kern046


Kern046:
@derrabus I used the same pattern from UPGRADE.md::l.63 if you also want to update this message :)

derrabus:
Yeah, maybe we should do revise the upgrade log before we release 3.0. 😅

PR #9555, opened 2022-03-02
Adding PHP attributes

ThomasLandauer:
@greg0ire This is the first (small) step towards documenting enums - see https://github.com/doctrine/orm/pull/9304#issuecomment-1047950899
I started reading the page from the top, and did some minor changes.
Biggest change: If you agree, I'd continue adding php attributes to all code blocks, and always mention it as first choice.

PR #10364, opened 2023-01-03
Docs: Moving *attributes* mapping to first position

ThomasLandauer:
Question: There are various occurrences of `#[Column(type: 'integer')]` on this page. I'd like to replace them with *either*:
1. `#[Column()]` since in recent versions the type is derived from the PHP typehint anyway
2. `#[Column(type: Types::INTEGER)]`
Which one is better?

### DIFF
--- a/docs/en/tutorials/composite-primary-keys.rst
+++ b/docs/en/tutorials/composite-primary-keys.rst
@@ -190,7 +190,52 @@ We keep up the example of an Article with arbitrary attributes, the mapping look
 
 .. configuration-block::
 
-    .. code-block:: php
+    .. code-block:: attribute
+
+        <?php
+        namespace Application\Model;
+
+        use Doctrine\Common\Collections\ArrayCollection;
+
+        #[Entity]
+        class Article
+        {
+            #[Id, Column(type: 'integer'), GeneratedValue]
+            private int|null $id = null;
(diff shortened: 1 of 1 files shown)


derrabus:
> Which one is better?

Option 3: `#[Column]` 🤓

derrabus:
Is this the only place where would need to swap annotations and attributes?

Note to my future self: This is going to be hell to merge up. 🙈

ThomasLandauer:
> Is this the only place where would need to swap annotations and attributes?

AFAIK, the agreed best-practice is to always show attributes first: https://github.com/doctrine/orm/pull/9555#issue-1157146746
But it looks like there was no systematic approach to change them all. So whenever I stumble over one, I fix it ;-) So I don't know how many old versions are still out there...

derrabus:
> AFAIK, the agreed best-practice is to always show attributes first: [#9555 (comment)](https://github.com/doctrine/orm/pull/9555#issue-1157146746)

Absolutely.

> But it looks like there was no systematic approach to change them all. So whenever I stumble over one, I fix it ;-) So I don't know how many old versions are still out there...

The reason for my asking is that I'd rather merge one big painful PR that fixes this inconsistency once and for all that to merge hundreds of micro-PRs for the same problem. 😓

derrabus (approved):

PR #10367, opened 2023-01-03
Migrate files with the least changes to PHP 8

greg0ire:
These are the lowest hanging fruits I could find after running Rector: I looked for files with a diff of 2 lines.
I did not include some changes that I find controversial, such as marking some constants as final when we should maybe consider making classes themselves final.

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
+++ b/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
@@ -47,7 +47,7 @@ class DefaultCacheFactory implements CacheFactory
         $this->fileLockRegionDirectory = $fileLockRegionDirectory;
     }
 
-    public function getFileLockRegionDirectory(): string
+    public function getFileLockRegionDirectory(): string|null
     {
         return $this->fileLockRegionDirectory;
     }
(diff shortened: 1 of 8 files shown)


lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -102,9 +102,9 @@ public function count(): int
     /**
      * {@inheritdoc}
      *
-     * @psalm-return Traversable<array-key, T>
+     * @psalm-return Traversable<array-key, T>&ArrayIterator
```
greg0ire: Here, we can't just use `ArrayIterator<array-key, T>` because that would make `T` appear in an invariant position, while it's declared as covariant.

derrabus: Do we want `ArrayIterator` to be part of the contract? I don't see a reason to change the return type.

greg0ire: :thinking: hum you're right, it makes no sense to leak that info.

SenseException (approved):

PR #10368, opened 2023-01-03
Docs: Removing `type: 'integer'` from mappings

ThomasLandauer:
Yet another micro-PR ;-) - as requested at https://github.com/doctrine/orm/pull/10364#issuecomment-1370155521

I also changed `$currentPrice` from `float` to `int`, since IMO it's better to store prices as `int` (=cents).

Question: Is there a reason why most typehints are `int|null`, instead of `?int`? Should I change them?

### DIFF
--- a/docs/en/tutorials/composite-primary-keys.rst
+++ b/docs/en/tutorials/composite-primary-keys.rst
@@ -32,9 +32,9 @@ and year of production as primary keys:
         class Car
         {
             public function __construct(
-                #[Id, Column(type: 'string')]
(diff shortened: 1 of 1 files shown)


SenseException (approved):

SenseException:
Thank you @ThomasLandauer 

> Question: Is there a reason why most typehints are int|null, instead of ?int? Should I change them?

It's part of the coding-standard, but as far as I know there's no deeper meaning into this decision. It was introduced in doctrine/coding-standard#266

PR #10369, opened 2023-01-03
[Documentation] Changing `int|null` to `?int`

ThomasLandauer:
Answer to https://github.com/doctrine/orm/pull/10368#issuecomment-1370320653 :

PSR-12 https://www.php-fig.org/psr/psr-12/ has this example, so it looks like it's preferring `?string`:
```php
public function functionName(?string $arg1, ?int &$arg2): ?string
```
But changing it in `ruleset.xml` would lead to errors allover the Doctrine code itself, right?


derrabus (changes requested):
> But changing it in `ruleset.xml` would lead to errors allover the Doctrine code itself, right?

Yes! And I'm afraid, our code style applies to the docs as well.

PR #10374, opened 2023-01-04
Migrate classes with the fewest changes to PHP 8

greg0ire:
Another round of PHP-8 changes, with slightly more changes per file

### DIFF
--- a/lib/Doctrine/ORM/EntityNotFoundException.php
+++ b/lib/Doctrine/ORM/EntityNotFoundException.php
@@ -18,12 +18,9 @@ class EntityNotFoundException extends RuntimeException implements ORMException
     /**
      * Static constructor.
      *
-     * @param string   $className
      * @param string[] $id
-     *
-     * @return self
      */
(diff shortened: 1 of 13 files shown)


lib/Doctrine/ORM/Event/OnFlushEventArgs.php
```diff
@@ -21,10 +21,8 @@ class OnFlushEventArgs extends ManagerEventArgs
      * Retrieve associated EntityManager.
      *
      * @deprecated 2.13. Use {@see getObjectManager} instead.
-     *
-     * @return EntityManagerInterface
      */
-    public function getEntityManager()
+    public function getEntityManager(): EntityManagerInterface
```
greg0ire: Actually, let me remove these methods in another PR.

greg0ire: See #10375

derrabus (approved):

PR #10385, opened 2023-01-09
Fix initializing lazy objects and get rid of "Typed property must not be accessed before initialization" errors

### LINKED ISSUE #10336: Clearing the em before accessing an uninitialized relation results in "typed property must not be accessed before initialization..." error

nicolas-grekas:
Fix #10336

### DIFF
--- a/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
+++ b/lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
@@ -165,6 +165,10 @@ class SimpleObjectHydrator extends AbstractHydrator
             }
         }
 
+        if (isset($this->_hints[Query::HINT_REFRESH_ENTITY])) {
+            $this->registerManaged($this->class, $this->_hints[Query::HINT_REFRESH_ENTITY], $data);
+        }
+
         $uow    = $this->_em->getUnitOfWork();
         $entity = $uow->createEntity($entityName, $data, $this->_hints);
 
--- a/lib/Doctrine/ORM/Proxy/ProxyFactory.php
+++ b/lib/Doctrine/ORM/Proxy/ProxyFactory.php
@@ -314,11 +314,10 @@ EOPHP;
     {
         $skippedProperties = ['__isCloning' => true];
         $identifiers       = array_flip($class->getIdentifierFieldNames());
+        $filter            = ReflectionProperty::IS_PUBLIC | ReflectionProperty::IS_PROTECTED | ReflectionProperty::IS_PRIVATE;
+        $reflector         = $class->getReflectionClass();
 
-        $filter = ReflectionProperty::IS_PUBLIC | ReflectionProperty::IS_PROTECTED | ReflectionProperty::IS_PRIVATE;
-        $reflector = $class->getReflectionClass();
-
-        do {
+        while ($reflector) {
             foreach ($reflector->getProperties($filter) as $property) {
                 $name = $property->getName();
 
@@ -330,8 +329,10 @@ EOPHP;
 
                 $skippedProperties[$prefix . $name] = true;
             }
-            $filter = ReflectionProperty::IS_PRIVATE;
-        } while ($reflector = $reflector->getParentClass());
+
+            $filter    = ReflectionProperty::IS_PRIVATE;
+            $reflector = $reflector->getParentClass();
+        }
 
         uksort($skippedProperties, 'strnatcmp');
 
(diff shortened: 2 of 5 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH10336Test.php
```diff
@@ -0,0 +1,44 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\Models\GH10336\GH10336Entity;
+use Doctrine\Tests\Models\GH10336\GH10336Relation;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @requires PHP 8.0
```
kbond:
```suggestion
 * @requires PHP 7.4
```

simonberger:
Where is the point in taking my fix and implement it in just a different coding style?
https://github.com/doctrine/orm/pull/10362

nicolas-grekas:
No specific point, I just missed your PR sorry, thanks for point at it.
Now that I compared both, I find the code style easier to follow in this very PR. WDYT?

pableu:
@simonberger closed #10362, so let's get this one merged :-)

bendavies:
LGTM, fixes my similar issue. thanks NG.

lib/Doctrine/ORM/Internal/Hydration/SimpleObjectHydrator.php
```diff
@@ -165,6 +165,10 @@ protected function hydrateRowData(array $row, array &$result)
             }
         }
 
+        if (isset($this->_hints[Query::HINT_REFRESH_ENTITY])) {
+            $this->registerManaged($this->class, $this->_hints[Query::HINT_REFRESH_ENTITY], $data);
+        }
+
```
greg0ire, line 171: These lines were removed in https://github.com/doctrine/orm/pull/10187 as pointed out in https://github.com/doctrine/orm/issues/10336#issuecomment-1368578980

lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -313,17 +314,24 @@ private function generateSkippedProperties(ClassMetadata $class): string
...
 
-            $prefix = $property->isPrivate() ? "\0" . $property->getDeclaringClass()->getName() . "\0" : ($property->isProtected() ? "\0*\0" : '');
+                $prefix = $property->isPrivate() ? "\0" . $property->getDeclaringClass()->getName() . "\0" : ($property->isProtected() ? "\0*\0" : '');
+
+                $skippedProperties[$prefix . $name] = true;
+            }
 
-            $skippedProperties[$prefix . $name] = true;
+            $filter    = ReflectionProperty::IS_PRIVATE;
+            $reflector = $reflector->getParentClass();
```
greg0ire, line 334: This seems to do the same as before, but for does it also for ancestors in the class hierarchy, retrieving only private properties for those (I guess public and protected are already retrieved at the first iteration because they are inherited).

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10336Test.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->_em->persist($relation);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $entity = $this->_em->find(GH10336Entity::class, 1);
+
+        $this->_em->clear();
+
+        $this->assertSame('foo', $entity->relation->value);
+    }
```
greg0ire, line 43: Although this test 100% covers this change, the `while` loop only ever does one iteration. Yet when I use `git restore --source=origin/2.14.x lib/Doctrine/ORM/Proxy/ProxyFactory.php`, a preexisting test fails `Doctrine\Tests\ORM\Functional\MappedSuperclassTest::testCRUD`, which means the scenario where there is more than 1 iteration is already covered 👍

It gets broken by the changes in `SimpleObjectHydrator`, and fixed by the changes in `ProxyFactory`

greg0ire (approved):
Looks good because there are more tests, but don't ask me to explain this change :sweat_smile: If somebody can help me figure it out, I might feel confident enough to merge.

stof (approved):

PR #10387, opened 2023-01-12
Make sure the schema tool does not miss columns from abstract entity classes in a STI tree

mpdude:
This PR fixes `SchemaTool` so it does not miss fields inherited from abstract `@Entity` classes in the middle of a single table inheritance (STI) tree, when these classes are not listed in the Discriminator Map (DM).

Abstract entitiy classes not listed in the DM will not show up in the root entity's ClassMetadata `subClasses` property. Fields inherited from them will show up in the field mapping metadata for subclasses, with `inherited => ...` being set.

The Schema Tool would skip these inherited fields, probably based on the assumption that they will be taken care of when the declaring class is processed. This, however, did not happen for abstract classes not listed in the DM.

**Update**  #10389 discusses whether it should in fact be a requirement to list also `abstract` classes in the DM, which might make this issue go away (masking it).

**Update 2** #10411 fixes this by making sure the missing `subClasses` are filled in

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -30,6 +30,7 @@ use function array_diff_key;
 use function array_filter;
 use function array_flip;
 use function array_intersect_key;
+use function array_merge;
 use function assert;
 use function count;
 use function current;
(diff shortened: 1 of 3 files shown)


mpdude:
Inviting @olsavmic as the OP for #9142 as well as @BenMorel as the OP for #9095 to the discussion.

PR #10394, opened 2023-01-13
Write down what "transient" means

mpdude:
It feels as if every time I try to debug mapping-related code, I stumble across the term "transient" and have to figure out what it means.

I hope that with this change, Google will bring up a meaningful definition next time.

### DIFF
--- a/docs/en/reference/architecture.rst
+++ b/docs/en/reference/architecture.rst
@@ -94,6 +94,19 @@ classes, and non-entity classes may extend entity classes.
     never calls entity constructors, thus you are free to use them as
     you wish and even have it require arguments of any type.
 
+Mapped Superclasses
+~~~~~~~~~~~~~~~~~~~
+
+A mapped superclass is an abstract or concrete class that provides persistent entity state and mapping information for its subclasses, but which is not itself an entity.
+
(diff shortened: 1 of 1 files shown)


docs/en/reference/architecture.rst
```diff
@@ -94,6 +94,19 @@ classes, and non-entity classes may extend entity classes.
...
+A mapped superclass is an abstract or concrete class that provides persistent entity state and mapping information for its subclasses, but which is not itself an entity.
+
+Mapped superclasses are explained in greater detail in the chapter on :doc:`inheritance mapping <reference/inheritance-mapping>`.
+
+Transient Classes
+~~~~~~~~~~~~~~~~~
+
+The term "transient class" appears in some places in the mapping drivers as well as the code dealing with metadata handling.
+
+A transient class is a class that is neither an entity nor a mapped superclass. From the ORMs point of view, these classes can be completely ignored, and no class metadata is loaded for them at all.
```
greg0ire:
```suggestion
A transient class is a class that is neither an entity nor a mapped superclass. From the ORM's point of view, these classes can be completely ignored, and no class metadata is loaded for them at all.
```

greg0ire (commented):
Please use linebreaks as in the rest of this document

mpdude:
Fixed

greg0ire (approved):

greg0ire:
Thanks @mpdude !

PR #10396, opened 2023-01-13
Document the meanings of 'inherited' and 'declared' in field mapping information

mpdude:
While trying to understand and debug inheritance-related issues, I figured I have no clear idea what `inherited` and `declared` actually means. 

This PR tries to document my findings, hopefully in a place where others might look for it.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -656,7 +656,7 @@ class ClassMetadataInfo implements ClassMetadata
      * The name of the field in the entity the association is mapped to.
      *
      * - <b>sourceEntity</b> (string)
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -625,6 +655,11 @@ class ClassMetadataInfo implements ClassMetadata
      * - <b>fieldName</b> (string)
      * The name of the field in the entity the association is mapped to.
      *
+     * - <b>sourceEntity</b> (string)
+     * The class name of the source entity. In the case of to-many-associations initially
```
greg0ire:
```suggestion
     * The class name of the source entity. In the case of to-many associations initially
```

greg0ire (approved):

SenseException (approved):

PR #10397, opened 2023-01-13
Factor out logic that tracks mapping inheritance

mpdude:
While working on #10396, I noticed the three kinds of mapping information track `inherited` and `declared` in the same way.

This PR suggests move the common code into a dedicated method, making it easier to see and understand where the information comes from.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -355,19 +355,29 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
     }
 
     /**
+     * Puts the `inherited` and `declared` values into mapping information for fields, associations
+     * and embedded classes.
+     *
+     * @param mixed[] $mapping
+     */
+    private function addMappingInheritanceInformation(array &$mapping, ClassMetadata $parentClass): void
+    {
+        if (! isset($mapping['inherited']) && ! $parentClass->isMappedSuperclass) {
+            $mapping['inherited'] = $parentClass->name;
+        }
+
+        if (! isset($mapping['declared'])) {
+            $mapping['declared'] = $parentClass->name;
+        }
+    }
+
+    /**
      * Adds inherited fields to the subclass mapping.
      */
     private function addInheritedFields(ClassMetadata $subClass, ClassMetadata $parentClass): void
     {
         foreach ($parentClass->fieldMappings as $mapping) {
-            if (! isset($mapping['inherited']) && ! $parentClass->isMappedSuperclass) {
-                $mapping['inherited'] = $parentClass->name;
-            }
-
-            if (! isset($mapping['declared'])) {
-                $mapping['declared'] = $parentClass->name;
-            }
-
+            $this->addMappingInheritanceInformation($mapping, $parentClass);
             $subClass->addInheritedFieldMapping($mapping);
         }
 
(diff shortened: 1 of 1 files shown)


mpdude:
Would appreciate any help how to fix static analysis type hinting here

greg0ire:
> Would appreciate any help how to fix static analysis type hinting here

I've tried looking into it, and it's a nightmare. I think the right course of action would be to experiment with DTOs on 3.0.x. That would allow us to determine where to use which type hint, and whether any fields are missing in the type aliases defined at the beginning of `ClassMetadataInfo`… would you be willing to work on this? I think in any case, you can reset to `HEAD^` and add the error you get to the baseline.

mpdude:
> I've tried looking into it, and it's a nightmare.

Probably thats because introducing a tool like Psalm into a huge legacy codebase is not fun, eh?

mpdude:
How do I add a phpcs error to the baseline?

greg0ire:
> Probably thats because introducing a tool like Psalm into a huge legacy codebase is not fun, eh?

Yeah it's kind of a pain, thankfully there is a baseline.

greg0ire:
> How do I add a phpcs error to the baseline?

I'd rather we add Psalm issues to the baseline, so I'd make phpcs happy, then ignore Psalm.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -47,6 +47,9 @@
  * to a relational database.
  *
  * @extends AbstractClassMetadataFactory<ClassMetadata>
+ *
+ * @psalm-import-type AssociationMapping from ClassMetadataInfo
+ * @psalm-import-type FieldMapping from ClassMetadataInfo
```
greg0ire: Run `vendor/bin/phpcbf` to get rid of this one.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -354,20 +357,30 @@ private function getShortName(string $className): string
         return strtolower(end($parts));
     }
 
+    /**
+     * Puts the `inherited` and `declared` values into mapping information for fields, associations
+     * and embedded classes.
+     *
+     * @param FieldMapping|AssociationMapping|array<string,mixed[]> $mapping
+     */
+    private function addMappingInheritanceInformation(array &$mapping, ClassMetadata $parentClass): void
```
greg0ire, line 363: I'd ignore both errors with `vendor/bin/psalm --set-baseline=psalm-baseline.xml`

greg0ire (commented):
Please squash everything into one commit that passes the build.

mpdude:
🟢 , without updating the baseline file 🥳

greg0ire (dismissed):

greg0ire:
Nice job!

greg0ire:
I think this should target 2.15.x (it's not a bugfix)

mpdude:
Rebased onto 2.15.x

greg0ire (dismissed):

SenseException (dismissed):
:+1:

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -354,20 +356,30 @@ private function getShortName(string $className): string
         return strtolower(end($parts));
     }
 
+    /**
+     * Puts the `inherited` and `declared` values into mapping information for fields, associations
+     * and embedded classes.
+     *
+     * @param FieldMapping|AssociationMapping|mixed[] $mapping
```
derrabus: That union is basically the same as `mixed[]`. How does this help us?

derrabus (changes requested):

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
mpdude: It was the only way I found that on the one hand satisfies CS requirements and on the other works for Psalm 🤷🏻‍♂️

derrabus: Maybe I missed what exactly the problem with PHPCS was. But unless we find a way to remove `mixed[]` from the union, it doesn't make sense to declare such a complex composite type.

mpdude: It works with `mixed[]` alone... is that ok?

derrabus: Sure, it's not worse than what we already have. 😅

derrabus (approved):

derrabus:
Thank you!

PR #10398, opened 2023-01-13
Move runtime check for to-many associations in mapped superclasses to a more suitable place

mpdude:
**Closing this** since #10473 was merged; it removed the check altogether.

<hr>

This PR moves the check that a mapped superclass does not declare to-many associations to a more suitable (IMHO) place.

It is no longer checked when another class inherits the association from the mapped superclass, but at the final runtime validation stage when loading the metadata for the mapped superclass itself.

In addition to that, can we review whether the check is correct for the case where we have 

* a base entity class with a to-many (non-owning) association
* a mapped superclass below it
* possibly another entity class below that

In other words, only the mapped superclass may not define to-many associations, but it is fine if an entity class above it does and when an entity class below inherits all that? My feeling is that should be fine.


mpdude:
Converted to draft for the time being. Not sure if it is working correctly – test coverage is unclear, and weird side effects come into play when trying to write a test (#10417, #10449, #10450)

mpdude:
X-Ref #9702

PR #10402, opened 2023-01-14
Migrate remaining exceptions to PHP 8 syntax

greg0ire:


lib/Doctrine/ORM/ORMInvalidArgumentException.php
```diff
@@ -104,96 +76,54 @@ static function (array $newEntityWithAssociation): string {
...
 
-    /**
-     * @param object $entry
-     * @psalm-param array<string, string> $associationMapping
-     *
-     * @return ORMInvalidArgumentException
-     */
-    public static function newEntityFoundThroughRelationship(array $associationMapping, $entry)
+    /** @psalm-param array<string, string> $associationMapping */
+    public static function newEntityFoundThroughRelationship(array $associationMapping, object $entry): ORMInvalidArgumentException
```
derrabus:
```suggestion
    public static function newEntityFoundThroughRelationship(array $associationMapping, object $entry): self
```

Can we enforce this via PHPCS, btw?

lib/Doctrine/ORM/ORMInvalidArgumentException.php
```diff
@@ -104,96 +76,54 @@ static function (array $newEntityWithAssociation): string {
...
     }
 
-    /**
-     * @param object $entry
-     * @psalm-param array<string, string> $assoc
-     *
-     * @return ORMInvalidArgumentException
-     */
-    public static function detachedEntityFoundThroughRelationship(array $assoc, $entry)
+    /** @psalm-param array<string, string> $assoc */
```
derrabus, line 87: Probably something we should fix on a lower branch: The method has a pretty strong opinion on which keys should be defined on that array. It might be worth either documenting the shape or splitting `$assoc` into three (?) parameters.

derrabus (dismissed):

lib/Doctrine/ORM/ORMInvalidArgumentException.php
greg0ire: @kukulich hi :wave: ! Would you maybe know about this?

kukulich: I think it's solved by `Squiz.Classes.SelfMemberReference`

kukulich: Hmm, probably not, because the sniff is used in Doctrine CS :(

kukulich: It looks I'll have to write the sniff :)

derrabus:
Rebase needed.

greg0ire:
:facepalm: sorry

derrabus (approved):

PR #10403, opened 2023-01-14
Reuse association mapping array shape

greg0ire:


mpdude (approved):

mpdude:
May I ask why such changes are made on 2.15, not on 2.14?

I understand that 2.14 is the current bugfix branch, and this is not a bugfix.

The idea of separating bugfix from feature branches (as I get it) is to avoid as many changes as possible on the bugfix one, to avoid the risk of unintentional breakage.

A change like this one here, however, does not change the running code, does it? It only improves static analysis capabilities and thus might be helpful for people fixing things on 2.14?

So why not merge such changes there?

greg0ire:
It might break static analysis on a downstream project. I use 2.14.x for phpdoc that is wrong. The current annotations are vague but not wrong

derrabus (approved):

PR #10405, opened 2023-01-14
Extract mappings to their  own DTOs

greg0ire:
An experiment, not sure if I will manage to do it, but at least it will allow us to port some changes to 2.x:
- for instance, everywhere `FieldMapping` the DTO is used, `FieldMapping` the array-shape should be used
- we can add missing fields to `FieldMapping` (for instance, `version`, or `default`)

## Todo

- [x] field mapping
- [x] association mapping
- [x] embedded classes mapping
- [x] discriminator column mapping
- [x] avoid repetition (array access implementation,  common fields)
- [x] use different DTOs for different association types
- [x] use different DTOs for different sides of the association
- [x] make classes that should never be instantiated abstract
- [x] Fix Psalm issues
- [x] Fix PHPStan issues
- [x] Implement `__sleep()`
- [x] Avoid redundant `isCascade*` properties

### DIFF
--- a/lib/Doctrine/ORM/Cache/DefaultCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultCache.php
@@ -9,6 +9,7 @@ use Doctrine\ORM\Cache;
 use Doctrine\ORM\Cache\Persister\CachedPersister;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Mapping\ToManyAssociationMapping;
 use Doctrine\ORM\ORMInvalidArgumentException;
 use Doctrine\ORM\UnitOfWork;
 
@@ -157,7 +158,7 @@ class DefaultCache implements Cache
 
         foreach ($metadatas as $metadata) {
             foreach ($metadata->associationMappings as $association) {
-                if (! $association['type'] & ClassMetadata::TO_MANY) {
+                if (! $association instanceof ToManyAssociationMapping) {
                     continue;
                 }
 
--- a/lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
+++ b/lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping;
+
+use InvalidArgumentException;
+
+use function property_exists;
+
+/** @internal */
+trait ArrayAccessImplementation
+{
+    /** @param string $offset */
+    public function offsetExists(mixed $offset): bool
+    {
+        return isset($this->$offset);
+    }
+
+    /** @param string $offset */
+    public function offsetGet(mixed $offset): mixed
+    {
+        if (! property_exists($this, $offset)) {
+            throw new InvalidArgumentException('Undefined property: ' . $offset);
+        }
+
+        return $this->$offset;
+    }
+
+    /** @param string $offset */
+    public function offsetSet(mixed $offset, mixed $value): void
+    {
+        $this->$offset = $value;
+    }
+
+    /** @param string $offset */
+    public function offsetUnset(mixed $offset): void
+    {
+        $this->$offset = null;
+    }
+}
--- a/lib/Doctrine/ORM/Mapping/AssociationMapping.php
+++ b/lib/Doctrine/ORM/Mapping/AssociationMapping.php
@@ -81,9 +81,6 @@ class AssociationMapping implements ArrayAccess
 
     public bool $isOwningSide = true;
 
-    /** @var list<JoinColumnData> */
-    public array|null $joinColumns = null;
-
     /** @var array<string, string> */
     public array|null $joinColumnFieldNames = null;
 
@@ -228,17 +225,6 @@ class AssociationMapping implements ArrayAccess
     /** @return mixed[] */
     public function toArray(): array
     {
-        $array = (array) $this;
-
-        if ($array['joinColumns'] !== null) {
-            $joinColumns = [];
-            foreach ($array['joinColumns'] as $column) {
-                $joinColumns[] = (array) $column;
-            }
-
-            $array['joinColumns'] = $joinColumns;
-        }
-
-        return $array;
+        return (array) $this;
     }
 }
--- a/lib/Doctrine/ORM/Mapping/AssociationMapping.php
+++ b/lib/Doctrine/ORM/Mapping/AssociationMapping.php
@@ -12,7 +12,7 @@ use function in_array;
 use function property_exists;
 
 /** @template-implements ArrayAccess<string, mixed> */
-class AssociationMapping implements ArrayAccess
+abstract class AssociationMapping implements ArrayAccess
 {
     /**
      * required for bidirectional associations
@@ -113,7 +113,6 @@ class AssociationMapping implements ArrayAccess
      *                                   source entity.
      */
     final public function __construct(
-        public readonly int $type,
         public string $fieldName,
         public string $sourceEntity,
         public string $targetEntity,
@@ -125,15 +124,13 @@ class AssociationMapping implements ArrayAccess
      *     fieldName: string,
      *     sourceEntity: class-string,
      *     targetEntity: class-string,
-     *     type: int,
      *     joinColumns?: mixed[]|null,
      *     joinTable?: mixed[]|null, ...} $mappingArray
      */
-    public static function fromMappingArray(array $mappingArray): self
+    public static function fromMappingArray(array $mappingArray): static
     {
         unset($mappingArray['isOwningSide']);
         $mapping = new static(
-            $mappingArray['type'],
             $mappingArray['fieldName'],
             $mappingArray['sourceEntity'],
             $mappingArray['targetEntity'],
@@ -181,6 +178,10 @@ class AssociationMapping implements ArrayAccess
             return $this->isOwningSide();
         }
 
+        if ($offset === 'type') {
+            return $this->type();
+        }
+
         return $this->$offset;
     }
 
@@ -211,6 +212,17 @@ class AssociationMapping implements ArrayAccess
         return $this instanceof ManyToManyOwningSideMapping;
     }
 
+    public function type(): int
+    {
+        return match (true) {
+            $this instanceof OneToOneAssociationMapping => ClassMetadata::ONE_TO_ONE,
+            $this instanceof OneToManyAssociationMapping => ClassMetadata::ONE_TO_MANY,
+            $this instanceof ManyToOneAssociationMapping => ClassMetadata::MANY_TO_ONE,
+            $this instanceof ManyToManyAssociationMapping => ClassMetadata::MANY_TO_MANY,
+            default => throw new Exception('Cannot determine type for ' . $this::class),
+        };
+    }
+
     /**
      * {@inheritDoc}
      */
@@ -234,6 +246,11 @@ class AssociationMapping implements ArrayAccess
     /** @return mixed[] */
     public function toArray(): array
     {
-        return (array) $this;
+        $array = (array) $this;
+
+        $array['isOwningSide'] = $this->isOwningSide();
+        $array['type']         = $this->type();
+
+        return $array;
     }
 }
--- a/lib/Doctrine/ORM/Mapping/AssociationMapping.php
+++ b/lib/Doctrine/ORM/Mapping/AssociationMapping.php
@@ -1,0 +1,278 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping;
+
+use ArrayAccess;
+
+use function in_array;
+use function property_exists;
+
+/** @template-implements ArrayAccess<string, mixed> */
+final class AssociationMapping implements ArrayAccess
+{
+    /**
+     * required for bidirectional associations
+     * The name of the field that completes the bidirectional association on
+     * the owning side. This key must be specified on the inverse side of a
+     * bidirectional association.
+     */
+    public string|null $mappedBy = null;
+
+    /**
+     * required for bidirectional associations
+     * The name of the field that completes the bidirectional association on
+     * the inverse side. This key must be specified on the owning side of a
+     * bidirectional association.
+     */
+    public string|null $inversedBy = null;
+
+    /**
+     * The names of persistence operations to cascade on the association.
+     *
+     * @var list<'persist'|'remove'|'detach'|'merge'|'refresh'|'all'>
+     */
+    public array|null $cascade = null;
+
+    /**
+     * one-to-many/many-to-many only
+     * A map of field names (of the target entity) to sorting directions
+     *
+     * @var array<string, 'asc'|'desc'>
+     */
+    public array|null $orderBy = null;
+
+    /**
+     * The fetching strategy to use for the association, usually defaults to FETCH_LAZY.
+     *
+     * @var ClassMetadata::FETCH_EAGER|ClassMetadata::FETCH_LAZY
+     */
+    public int|null $fetch = null;
+
+    /**
+     * many-to-many only
+     * Specification of the join table and its join columns (foreign keys).
+     * Only valid for many-to-many mappings. Note that one-to-many associations
+     * can be mapped through a join table by simply mapping the association as
+     * many-to-many with a unique constraint on the join table.
+     */
+    public JoinTableMapping|null $joinTable = null;
+
+    /**
+     * to-many only
+     * Specification of a field on target-entity that is used to index the
+     * collection by. This field HAS to be either the primary key or a unique
+     * column. Otherwise the collection does not contain all the entities that
+     * are actually related.
+     */
+    public string|null $indexBy = null;
+
+    /**
+     * This is set when the association is inherited by this class from another
+     * (inheritance) parent <em>entity</em> class. The value is the FQCN of the
+     * topmost entity class that contains this association. (If there are
+     * transient classes in the class hierarchy, these are ignored, so the
+     * class property may in fact come from a class further up in the PHP class
+     * hierarchy.) To-many associations initially declared in mapped
+     * superclasses are <em>not</em> considered 'inherited' in the nearest
+     * entity subclasses.
+     *
+     * @var class-string
+     */
+    public string|null $inherited = null;
+
+    /**
+     * This is set when the association does not appear in the current class
+     * for the first time, but is initially declared in another parent
+     * <em>entity or mapped superclass</em>. The value is the FQCN of the
+     * topmost non-transient class that contains association information for
+     * this relationship.
+     */
+    public string|null $declared = null;
+
+    public array|null $cache = null;
+
+    public bool|null $id = null;
+
+    public bool $isCascadeRemove  = false;
+    public bool $isCascadePersist = false;
+    public bool $isCascadeRefresh = false;
+    public bool $isCascadeMerge   = false;
+    public bool $isCascadeDetach  = false;
+
+    public bool|null $isOnDeleteCascade = null;
+
+    public bool $isOwningSide = true;
+
+    /** @var list<JoinColumnData> */
+    public array|null $joinColumns = null;
+
+    /** @var array<string, string> */
+    public array|null $joinColumnFieldNames = null;
+
+    /** @var list<mixed> */
+    public array|null $joinTableColumns = null;
+
+    /** @var class-string */
+    public string|null $originalClass = null;
+
+    /** @var string */
+    public string|null $originalField = null;
+
+    public bool|null $orphanRemoval = null;
+
+    public array|null $relationToSourceKeyColumns = null;
+    public array|null $relationToTargetKeyColumns = null;
+
+    /** @var array<string, string> */
+    public array|null $sourceToTargetKeyColumns = null;
+
+    /** @var array<string, string> */
+    public array|null $targetToSourceKeyColumns = null;
+
+    public bool|null $unique = null;
+
+    /**
+     * @param string       $fieldName    The name of the field in the entity
+     *                                   the association is mapped to.
+     * @param class-string $sourceEntity The class name of the source entity.
+     *                                   In the case of to-many associations
+     *                                   initially present in mapped
+     *                                   superclasses, the nearest
+     *                                   <em>entity</em> subclasses will be
+     *                                   considered the respective source
+     *                                   entities.
+     * @param class-string $targetEntity The class name of the target entity.
+     *                                   If it is fully-qualified it is used as
+     *                                   is. If it is a simple, unqualified
+     *                                   class name the namespace is assumed to
+     *                                   be the same as the namespace of the
+     *                                   source entity.
+     */
+    public function __construct(
+        public int $type,
+        public string $fieldName,
+        public string $sourceEntity,
+        public string $targetEntity,
+    ) {
+    }
+
+    /**
+     * @psalm-param array{
+     *     fieldName: string,
+     *     sourceEntity: class-string,
+     *     targetEntity: class-string,
+     *     type: int,
+     *     joinColumns?: mixed[]|null,
+     *     joinTable?: mixed[]|null, ...} $mappingArray
+     */
+    public static function fromMappingArray(array $mappingArray): self
+    {
+        $mapping = new self(
+            $mappingArray['type'],
+            $mappingArray['fieldName'],
+            $mappingArray['sourceEntity'],
+            $mappingArray['targetEntity'],
+        );
+        foreach ($mappingArray as $key => $value) {
+            if (in_array($key, ['type', 'fieldName', 'sourceEntity', 'targetEntity'])) {
+                continue;
+            }
+
+            if ($key === 'joinColumns') {
+                if ($value === null) {
+                    continue;
+                }
+
+                $joinColumns = [];
+                foreach ($value as $column) {
+                    $joinColumns[] = JoinColumnData::fromMappingArray($column);
+                }
+
+                $mapping->joinColumns = $joinColumns;
+
+                continue;
+            }
+
+            if ($key === 'joinTable') {
+                if ($value === [] || $value === null) {
+                    continue;
+                }
+
+                $mapping->joinTable = JoinTableMapping::fromMappingArray($value);
+
+                continue;
+            }
+
+            if (property_exists($mapping, $key)) {
+                $mapping->$key = $value;
+            }
+        }
+
+        return $mapping;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function offsetExists($offset): bool
+    {
+        return isset($this->$offset);
+    }
+
+    public function offsetGet($offset): mixed
+    {
+        return $this->$offset;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function offsetSet($offset, $value): void
+    {
+        if ($offset === 'joinColumns') {
+            $joinColumns = [];
+            foreach ($value as $column) {
+                $joinColumns[] = JoinColumnData::fromMappingArray($column);
+            }
+
+            $value = $joinColumns;
+        }
+
+        if ($offset === 'joinTable') {
+            $value = JoinTableMapping::fromMappingArray($value);
+        }
+
+        $this->$offset = $value;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function offsetUnset($offset): void
+    {
+        $this->$offset = null;
+    }
+
+    /** @return mixed[] */
+    public function toArray(): array
+    {
+        $array = (array) $this;
+
+        if ($array['joinColumns'] !== null) {
+            $joinColumns = [];
+            foreach ($array['joinColumns'] as $column) {
+                $joinColumns[] = (array) $column;
+            }
+
+            $array['joinColumns'] = $joinColumns;
+        }
+
+        if ($array['joinTable'] !== null) {
+            $array['joinTable'] = $array['joinTable']->toArray();
+        }
+
+        return $array;
+    }
+}
--- a/lib/Doctrine/ORM/Mapping/AssociationMapping.php
+++ b/lib/Doctrine/ORM/Mapping/AssociationMapping.php
@@ -165,30 +165,30 @@ abstract class AssociationMapping implements ArrayAccess
     }
 
     /** @psalm-assert-if-true AssociationOwningSideMapping $this */
-    public function isOwningSide(): bool
+    final public function isOwningSide(): bool
     {
         return $this instanceof AssociationOwningSideMapping;
     }
 
     /** @psalm-assert-if-true ToOneAssociationMapping $this */
-    public function isToOne(): bool
+    final public function isToOne(): bool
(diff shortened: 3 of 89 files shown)


mpdude:
What is the supposed migration path for this? 

In 2.x, the `ClassMetadataInfo::$fieldMappings` is a public property, marked as `READ-ONLY`, but not internal. To my understanding, that means it is allowed to read and work with them.

I don't know about the deprecation/transition rules for this project. But changing this on the 3.0 branch means things will break during the upgrade, and there is no "smooth" transition (e. g. by getting deprecation warnings in 2.x, which can be fixed, then the 3.0 upgrade can be made).

I generally support the idea and think it's a good thing design-wise.

Can we make that more user-friendly, by
* adding the DTO in 2.x already, 
* implementing `ArrayAccess`, make things look like before
* trigger a deprecation when the array access is actually used
* drop the array access in 3.0?

greg0ire:
I'd actually add the array access on 3.0,much like what I did recently for the Token class in doctrine/lexer, that way the type does not change from array to DTO in a minor. What we might do to ease the transition could be to introduce a factory that would produce arrays with nullable properties instead of optional properties, to make the array closer to the final dto

lib/Doctrine/ORM/Mapping/FieldMapping.php
```diff
@@ -0,0 +1,117 @@
...
+use function in_array;
+use function property_exists;
+
+/** @template-implements ArrayAccess<string, mixed> */
+final class FieldMapping implements ArrayAccess
+{
+    /** @var int|null The database length of the column. Optional. Default value taken from the type. */
+    public int|null $length = null;
+    /**
+     * @var bool|null Marks the field as the primary key of the entity. Multiple
```
greg0ire, line 21: While trying to fix the build, I reverted to `bool|null` here, to be closer to the previous situation. Now that the build is green, I can do a separate commit to make boolean field not-nullable, and default to `false`.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -328,15 +328,16 @@ private function getShortName(string $className): string
     private function addInheritedFields(ClassMetadata $subClass, ClassMetadata $parentClass): void
     {
         foreach ($parentClass->fieldMappings as $mapping) {
-            if (! isset($mapping['inherited']) && ! $parentClass->isMappedSuperclass) {
-                $mapping['inherited'] = $parentClass->name;
+            $subClassMapping = clone $mapping;
```
greg0ire, line 416: Lost a huge amount of time on this, but yeah, we have to clone here now.

mpdude: #10397 changed that area, did you include/merge that already?

why can’t you just update the value?

greg0ire: I need to update the child value without updating the parent value. When using an array, a copy is made. When using objects, we're altering the parent objects.

We do need to do a merge up.

mpdude: Would it be better to do the clone up in `doLoadMetadata` or so, in a more central location?

greg0ire: Merge up done in https://github.com/doctrine/orm/pull/10421

About `doLoadMetadata`, I'm not sure… I mean there's one clone per mapping to do, so it's not like you could pass clones to `addInheritedFields` or something :thinking:

greg0ire:
Did a new push, comes with static analysis issues that I should be able to solve once I have converted every argument to `addMappingInheritanceInformation` to a DTO.

greg0ire:
Pushed a new version with an association mapping DTO, hopefully that was the hardest part of the job.

greg0ire:
Just found out that I'm doing the opposite of what's been done 13 years ago in 8d3e0e61eaeaaf8753838a8580b2ee36dbd8f34f :sweat_smile: 
Why oh why?

mpdude:
Maybe performance? Arrays might be faster to access than methods to call?

greg0ire:
Discussed it internally, and you are right… well, you were. It was true in PHP 5.3, but no longer is. I think I can proceed with this.

mpdude:
Before you put too much work into it, are you sure there is a smooth migration path?

greg0ire:
I think the `ArrayAccess` implementation is enough. All the arrays I'm changing are object-like arrays, so they are very unlikely to be used with anything else than the `[]` notation. Also, I experimented with this on `doctrine/lexer` and got no complaints.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
```diff
@@ -0,0 +1,226 @@
...
+    public function isToOneOwningSide(): bool
+    {
+        return $this instanceof ToOneAssociationMapping && $this->isOwningSide();
+    }
+
+    /** @psalm-assert-if-true ManyToManyOwningSideMapping $this */
+    public function isManyToManyOwningSide(): bool
+    {
+        return $this instanceof ManyToManyOwningSideMapping;
+    }
```
greg0ire, line 183: @mpdude doing `instanceof` checks got old real fast, so I introduced the 3 methods above, and created https://github.com/phpstan/phpstan/issues/8904

greg0ire:
@phansys @franmomu hi! In this PR, I change arrays into DTOs, and I've been told there might be places in `gedmo/doctrine-extensions` where extra properties are written to these arrays. You seem to be active on those repos, so if this rings a bell, please point me to where this happens, so that I can come up with a plan to support this.

franmomu:
> @phansys @franmomu hi! In this PR, I change arrays into DTOs, and I've been told there might be places in `gedmo/doctrine-extensions` where extra properties are written to these arrays. You seem to be active on those repos, so if this rings a bell, please point me to where this happens, so that I can come up with a plan to support this.

hey @greg0ire, off the top of my head I don't remember modifying these arrays directly, there are usually calls through the `UoW`, but I have to take a look. What I think we might have are calls to `array_key_exists` to these arrays [which will fail](https://3v4l.org/rGDm9), but we can change that beforehand.

greg0ire:
Thanks for looking into this. We can always add support for extra properties later if they are a must anyway.

greg0ire:
I removed the item "implement `__sleep()`", I think it would be best done later with a performance benchmark to determine whether it's worth it or not. Let's mark this as ready for review then.

derrabus:
Wow, amazing work. I hope, I'll find some time to review it. 🤞🏻

lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping;
+
+trait ArrayAccessImplementation
+{
+    /** @param array-key $offset */
+    public function offsetExists($offset): bool
```
derrabus:
```suggestion
    public function offsetExists(mixed $offset): bool
```

lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        return isset($this->$offset);
+    }
+
+    /** @param array-key $offset */
+    public function offsetGet($offset): mixed
```
derrabus:
```suggestion
    public function offsetGet(mixed $offset): mixed
```

lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        return $this->$offset;
+    }
+
+    /**
+     * @param array-key $offset
+     * @param mixed     $value
+     */
+    public function offsetSet($offset, $value): void
```
derrabus:
```suggestion
    public function offsetSet(mixed $offset, mixed $value): void
```

lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $this->$offset = $value;
+    }
+
+    /** @param array-key $offset */
+    public function offsetUnset($offset): void
```
derrabus:
```suggestion
    public function offsetUnset(mixed $offset): void
```

lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping;
+
+trait ArrayAccessImplementation
```
derrabus, line 12: `@internal` please!!! 🙀

lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /** @param array-key $offset */
```
derrabus: All of the consumers of this trait use string offsets, so we might as well…

```suggestion
    /** @param string $offset */
```

Apart from that, PHPStan and Psalm should support generic traits.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        return $mapping;
+    }
+
+    /** @psalm-assert-if-true AssociationOwningSideMapping $this */
+    public function isOwningSide(): bool
+    {
+        return $this instanceof AssociationOwningSideMapping;
+    }
```
derrabus: Maybe I'm getting old, but wouldn't the saner approach be…

```suggestion
    public function isOwningSide(): bool
    {
        return false;
    }
```

… and overriding this method in `AssociationOwningSideMapping`?

lib/Doctrine/ORM/Mapping/AssociationMapping.php
```diff
@@ -0,0 +1,240 @@
...
+    public function type(): int
+    {
+        return match (true) {
+            $this instanceof OneToOneAssociationMapping => ClassMetadata::ONE_TO_ONE,
+            $this instanceof OneToManyAssociationMapping => ClassMetadata::ONE_TO_MANY,
+            $this instanceof ManyToOneAssociationMapping => ClassMetadata::MANY_TO_ONE,
+            $this instanceof ManyToManyAssociationMapping => ClassMetadata::MANY_TO_MANY,
+            default => throw new Exception('Cannot determine type for ' . $this::class),
+        };
+    }
```
derrabus: We could make a `public readonly int $type` that is initialized by the constructor using this `match` statement. This would remove an edge-case from your `offsetGet`/`offsetSet` and `toArray` implementations.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
```diff
@@ -0,0 +1,240 @@
...
+
+    public function offsetGet($offset): mixed
```
derrabus: You're overriding methods that are provided by the `ArrayAccessImplementation` trait already. You do this to cover two edge cases. The methods `offsetExists()` and `offsetUnset()` are currently broken because those edge cases are not taken into account.

Suggestion:

* Either don't use the trait in this class and implement all `ArrayAccess` methods in a consistent manner,
* or eliminate the edge-cases, or at least make them work without having to override the trait.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
```diff
@@ -0,0 +1,240 @@
...
+    {
+        if ($offset === 'isOwningSide') {
+            return $this->isOwningSide();
+        }
+
+        if ($offset === 'type') {
+            return $this->type();
+        }
+
+        return $this->$offset;
```
derrabus: Y U NO `match($offset)`?

lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
```diff
@@ -0,0 +1,35 @@
...
+    public function offsetExists($offset): bool
+    {
+        return isset($this->$offset);
+    }
+
+    /** @param array-key $offset */
+    public function offsetGet($offset): mixed
+    {
+        return $this->$offset;
```
derrabus, line 27: Should we check if the property exists? Or do we want to allow arbitrary properties? In the latter case, we should put `#[AllowDynamicProperties]` on all implementations.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function offsetSet($offset, $value): void
+    {
+        if ($offset === 'joinTable') {
+            assert($this instanceof ManyToManyAssociationMapping);
```
derrabus: We should raise a proper exception instead.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
```diff
@@ -0,0 +1,240 @@
...
+     *                                   If it is fully-qualified it is used as
+     *                                   is. If it is a simple, unqualified
+     *                                   class name the namespace is assumed to
+     *                                   be the same as the namespace of the
+     *                                   source entity.
+     */
+    final public function __construct(
+        public string $fieldName,
+        public string $sourceEntity,
+        public string $targetEntity,
```
derrabus: Can we make those `readonly`? I don't think we want to mutate them.

lib/Doctrine/ORM/Mapping/DiscriminatorColumnMapping.php
```diff
@@ -0,0 +1,57 @@
...
+            $mappingArray['name'],
+        );
+        foreach ($mappingArray as $key => $value) {
+            if (in_array($key, ['type', 'fieldName', 'name'])) {
+                continue;
+            }
+
+            if (property_exists($mapping, $key)) {
+                $mapping->$key = $value;
+            }
```
derrabus: In other classes, you throw for invalid properties while you silently ignore them here. Is that on purpose?

lib/Doctrine/ORM/Mapping/AssociationMapping.php
```diff
@@ -0,0 +1,240 @@
...
+
+use ArrayAccess;
+use Exception;
+
+use function assert;
+use function in_array;
+use function property_exists;
+
+/** @template-implements ArrayAccess<string, mixed> */
+abstract class AssociationMapping implements ArrayAccess
```
derrabus, line 16: Can we `final`ize all public methods in this class that are not overridden by child classes?

lib/Doctrine/ORM/Mapping/JoinColumnData.php
```diff
@@ -0,0 +1,46 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping;
+
+use ArrayAccess;
+
+use function property_exists;
+
+/** @template-implements ArrayAccess<string, mixed> */
+final class JoinColumnData implements ArrayAccess
```
derrabus: Data is very generic and doesn't seem like a good fit. `JoinColumnMapping`? `JoinColumnDefinition`? `JoinColumn`?

lib/Doctrine/ORM/Mapping/JoinColumnData.php
```diff
@@ -0,0 +1,46 @@
...
+    public string|null $onDelete             = null;
+    public string|null $columnDefinition     = null;
+    public bool|null $nullable               = null;
+    public string|null $referencedColumnName = null;
+    /** @var array<string, mixed> */
+    public array|null $options = null;
+
+    public function __construct()
+    {
+    }
```
derrabus: Can be removed.

lib/Doctrine/ORM/Mapping/ManyToManyOwningSideMapping.php
```diff
@@ -0,0 +1,131 @@
...
+    /** @param mixed[] $mappingArray */
+    public static function fromMappingArrayAndNamingStrategy(array $mappingArray, NamingStrategy $namingStrategy): self
+    {
+        $mapping = parent::fromMappingArray($mappingArray);
+
+        // owning side MUST have a join table
+        if (! isset($mapping['joinTable']['name'])) {
+            if (! isset($mapping->joinTable)) {
+                $mapping->joinTable = new JoinTableMapping();
+            }
```
derrabus:
```suggestion
            $mapping->joinTable ??= new JoinTableMapping();
```

lib/Doctrine/ORM/Mapping/JoinTableMapping.php
```diff
@@ -0,0 +1,87 @@
...
+
+/** @template-implements ArrayAccess<string, mixed> */
+final class JoinTableMapping implements ArrayAccess
+{
+    use ArrayAccessImplementation;
+
+    public bool|null $quoted = null;
+
+    /** @var list<JoinColumnData> */
+    public array|null $joinColumns = null;
```
derrabus: The native type allows `null` while the `@var` type doesn't. This should be fixed. Are `null` and `[]` actually different cases? If not, I suggest to remove the nullability here.

lib/Doctrine/ORM/Mapping/JoinTableMapping.php
```diff
@@ -0,0 +1,87 @@
...
+
+    /** @var list<JoinColumnData> */
+    public array|null $inverseJoinColumns = null;
```
derrabus: Same.

lib/Doctrine/ORM/Mapping/ManyToManyAssociationMapping.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    public array|null $relationToSourceKeyColumns = null;
+    public array|null $relationToTargetKeyColumns = null;
+
+    /** @param mixed[] $mappingArray */
+    public static function fromMappingArray(array $mappingArray): static
+    {
+        $mapping = parent::fromMappingArray($mappingArray);
+
+        $mapping['orphanRemoval'] = isset($mapping['orphanRemoval']) && $mapping['orphanRemoval'];
```
derrabus: This line confuses me. Why are you mapping this property to itself?

lib/Doctrine/ORM/Mapping/ManyToManyOwningSideMapping.php
```diff
@@ -0,0 +1,131 @@
...
+
+            $mapping->joinTable['name'] = $namingStrategy->joinTableName(
```
derrabus: I noticed that you're switching between array access and properties within the new DTOs. I understand that keeping array access keeps the diff small for old code, but inside the new classes, I believe using properties consistently would improve the readability. Especially in methods where we map from arrays to DTOs and vice versa, using properties would make it clearer what is the array and what the DTO.

In this very line, you're even using both methods in one expression. 😵

derrabus (changes requested):
🤯

lib/Doctrine/ORM/Mapping/AssociationMapping.php
greg0ire: It doesn't look saner to me. When stumbling upon the code in your proposition, one has might think "what's going on here? Ah, maybe that method is overridden somewhere (but where?). With the approach I picked, you have everything you need to understand this in the same place.

lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
greg0ire: I added the check, and at some point it didn't work, so I removed it. I just tried adding it back, and it the test suite passes. I'll push to see if it works with no skipped tests.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
greg0ire: I feel uneasy duplicating information contained in the type itself as a property.

lib/Doctrine/ORM/Mapping/ArrayAccessImplementation.php
greg0ire: All checks pass!

lib/Doctrine/ORM/Mapping/AssociationMapping.php
greg0ire: Not `$sourceEntity` according to the test suite, because of
https://github.com/doctrine/orm/blob/13d86957808f3fab3dfb9c681c59984a2bb4e4dd/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L444

It works for the other ones though.

lib/Doctrine/ORM/Mapping/DiscriminatorColumnMapping.php
greg0ire: IIRC I started added that check in other classes after this one, I must have forgotten to update it.

lib/Doctrine/ORM/Mapping/JoinColumnData.php
greg0ire: `JoinColumnMapping` 👍

lib/Doctrine/ORM/Mapping/JoinTableMapping.php
greg0ire: `[]` causes an exception from DBAL on this piece of code:

```
        if (count($foreignKey->getLocalColumns()) === 0) {
            throw new InvalidArgumentException("Incomplete definition. 'local' required.");
        }
```

So I guess they are different cases, and `[]` is not supposed to ever happen.

lib/Doctrine/ORM/Mapping/ManyToManyAssociationMapping.php
greg0ire: I'm just moving code around here:

https://github.com/doctrine/orm/blob/13d86957808f3fab3dfb9c681c59984a2bb4e4dd/lib/Doctrine/ORM/Mapping/ClassMetadata.php#L1685

This ensures the property ends up being a boolean, treating `null` as `false`. I will remove this line and default to `false` instead.

greg0ire:
Sudden issue with `cs2pr`: `chmod: cannot access '/usr/local/bin/cs2pr'`

Opened https://github.com/shivammathur/setup-php/issues/715

lib/Doctrine/ORM/Mapping/DiscriminatorColumnMapping.php
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @template-implements ArrayAccess<string, mixed>
+ * @psalm-type DiscriminatorColumnMappingType = 'boolean'|'array'|'object'|'datetime'|'time'|'date'
+ */
+final class DiscriminatorColumnMapping implements ArrayAccess
+{
+    use ArrayAccessImplementation;
+
+    /** @var int|null The database length of the column. Optional. Default value taken from the type. */
```
derrabus:
```suggestion
    /** The database length of the column. Optional. Default value taken from the type. */
```

lib/Doctrine/ORM/Mapping/EmbeddedClassMapping.php
```diff
@@ -0,0 +1,66 @@
...
+     * This is set when this embedded-class field is inherited by this class
+     * from another (inheritance) parent <em>entity</em> class. The value is
+     * the FQCN of the topmost entity class that contains mapping information
+     * for this field. (If there are transient classes in the class hierarchy,
+     * these are ignored, so the class property may in fact come from a class
+     * further up in the PHP class hierarchy.) Fields initially declared in
+     * mapped superclasses are <em>not</em> considered 'inherited' in the
+     * nearest entity subclasses.
+     *
+     * @var class-string
```
derrabus:
```suggestion
     * @var class-string|null
```

lib/Doctrine/ORM/Mapping/EmbeddedClassMapping.php
```diff
@@ -0,0 +1,66 @@
...
+    public string|null $inherited = null;
+
+    /**
+     * This is set when the embedded-class field does not appear for the first
+     * time in this class, but is originally declared in another parent
+     * <em>entity or mapped superclass</em>. The value is the FQCN of the
+     * topmost non-transient class that contains mapping information for this
+     * field.
+     *
+     * @var class-string
```
derrabus:
```suggestion
     * @var class-string|null
```

lib/Doctrine/ORM/Mapping/AssociationMapping.php
greg0ire: Why? I think we should avoid new exceptions for temporary code such as this. It has no equivalent for properties.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
derrabus: > Ah, maybe that method is overridden somewhere

Yeah, that's the general idea of abstract classes. 😅 

It's a bit weird that the abstraction knows details about its implementations. But all right, let's roll with it for now.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
derrabus: Understood. I was trying to eliminate that special treatment from those three methods, but maybe I was trying to hard. A method is probably the better approach.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
derrabus: So the plan is to remove the `ArrayAccess` implementations before the release? In that case, I'm fine with the assert. Otherwise we need to assume that the array access is used downstream and this assert is basically input validation which is not what assert should be used for.

lib/Doctrine/ORM/Mapping/JoinTableMapping.php
derrabus: But is `null` supposed to happen? A join table without columns doesn't sound right to me either.

lib/Doctrine/ORM/Mapping/AssociationMapping.php
greg0ire: No, the plan is to stop using array access internally before the release, to deprecate it in 3.1, and to remove it in 4.0.

I'll remove this assert, is was just a way for me to make sure I was doing this refactoring properly (checking my assumptions).

lib/Doctrine/ORM/Mapping/JoinTableMapping.php
greg0ire: Disallowing `null` seems OK according to the test suite, defaulting to `[]` isn't, though… so I guess it's not `null`, but uninitialized? I'm confused now…

lib/Doctrine/ORM/Mapping/AssociationMapping.php
greg0ire: > It's a bit weird that the abstraction knows details about its implementations

I'd like to make this a bit more like a sealed class (https://wiki.php.net/rfc/sealed_classes), but I didn't know how. But yeah, that's indeed the concept here.

lib/Doctrine/ORM/Mapping/JoinTableMapping.php
derrabus: > so I guess it's not `null`, but uninitialized? I'm confused now…

What we're looking for is a representation for an incomplete `JoinTableMapping`. You construct that object and then you add columns to it. For that matter, the empty array seems like a good initial value. `null` is just an unnecessary edge case that static analysis will always force us to cover.

greg0ire: I took a closer look at the failing test, it fails because of what's in `ManyToManyOwningSideMapping`:

```
        if (! isset($mapping->joinTable->joinColumns)) {
            $mapping->joinTable->joinColumns = [
                JoinColumnMapping::fromMappingArray([
                    'name' => $namingStrategy->joinKeyColumnName($mapping->sourceEntity, $selfReferencingEntityWithoutJoinColumns ? 'source' : null),
                    'referencedColumnName' => $namingStrategy->referenceColumnName(),
                    'onDelete' => 'CASCADE',
                ]),
            ];
        }
```

Using `=== []` instead of `isset()` fixes this particular issue, I'm looking at the next one.

greg0ire: fixed :)

beberlei (commented):
I didnt do a full review, but I like the parts i saw. If you are confident about this change you have my +1.

The only thing insm wondering about are  some of the duplications, for example the isCascade... - specifically I am wondering if all these new mapping claasea shoulf have code that omits the default values to end up in the serialized payloads for metadata cache, like its done in ClassMetadata

greg0ire:
You mean `__sleep()` methods? I see there were such methods in 8d3e0e61eaeaaf8753838a8580b2ee36dbd8f34f, before switching to arrays, so I can definitely look into doing something similar.

greg0ire:
> some of the duplications, for example the isCascade

Oh right I didn't spot that, but it looks like the properties `AssociationMapping::$isCascade*` could become methods, based on `AssociationMapping::$cascade`.

greg0ire:
@beberlei my 2 last commits should address your concerns :)

lib/Doctrine/ORM/Mapping/DiscriminatorColumnMapping.php
```diff
@@ -0,0 +1,74 @@
...
+use ArrayAccess;
+use BackedEnum;
+use Exception;
+
+use function in_array;
+use function property_exists;
+
+/**
+ * @template-implements ArrayAccess<string, mixed>
+ * @psalm-type DiscriminatorColumnMappingType = 'boolean'|'array'|'object'|'datetime'|'time'|'date'|'string'
```
greg0ire: This is wrong. These are values that `$type` can _not_ take.

greg0ire: Fixed

lib/Doctrine/ORM/Cache/DefaultCache.php
```diff
@@ -157,7 +158,7 @@ public function evictCollectionRegions(): void
 
         foreach ($metadatas as $metadata) {
             foreach ($metadata->associationMappings as $association) {
-                if (! $association['type'] & ClassMetadata::TO_MANY) {
+                if (! $association instanceof ToManyAssociationMapping) {
```
SenseException, line 161: If those are all `AssociationMapping`, wouldn't the use of a method be an option instead of an `instanceof` check? You've used methods at other places too.

SenseException (commented):
I haven't done the full review yet. The amount of changes slows my progress, so maybe I'll do it in parts.

lib/Doctrine/ORM/Cache/DefaultCache.php
greg0ire: You're right, I will add such a method.

greg0ire:
Closing in favor of smaller PRs, the first of which will be https://github.com/doctrine/orm/pull/10607

PR #10408, opened 2023-01-15
Field mapping improvements

greg0ire:
It will be hard to finish https://github.com/doctrine/orm/pull/10405, but this should get us closer to it.
Also, I learnt that DTOs shouldn't be used in mapping drivers, because drivers produce incomplete field mappings (they might not have a type), and also produce mapping overrides that might have even fewer fields. DTOs should be used in the class metadata and its consumers.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1582,9 +1582,9 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * Validates & completes the given field mapping based on typed property.
      *
-     * @param  array{fieldName: string, type?: mixed} $mapping The field mapping to validate & complete.
+     * @param  array{fieldName: string, type?: string} $mapping The field mapping to validate & complete.
      *
-     * @return array{fieldName: string, enumType?: string, type?: mixed} The updated mapping.
+     * @return array{fieldName: string, enumType?: class-string<BackedEnum>, type?: string} The updated mapping.
      */
     private function validateAndCompleteTypedFieldMapping(array $mapping): array
     {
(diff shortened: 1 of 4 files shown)


greg0ire:
@michnovka please review

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1584,7 +1586,7 @@ private function isTypedProperty(string $name): bool
      *
      * @param  array{fieldName: string, type?: mixed} $mapping The field mapping to validate & complete.
```
michnovka: if we update return type, we should update the `type?: string` here in param too

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1628,7 +1630,7 @@ private function validateAndCompleteTypedAssociationMapping(array $mapping): arr
      *     enumType?: class-string,
      * } $mapping The field mapping to validate & complete.
      *
-     * @return mixed[] The updated mapping.
+     * @return FieldMapping The updated mapping.
```
michnovka, line 1633: FieldMapping type has a compulsory array-key `columnName`. I am not sure that this is satisfied here, as the function does not set it and @param $mapping does not specify it.

michnovka (changes requested):
have a look at my review

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
greg0ire: It's satisfied right below the comment "Complete fieldName and columnName mapping"

michnovka: Thats a different function tho `validateAndCompleteFieldMapping`. The change you made is for `validateAndCompleteTypedAssociationMapping` and that one does NOT return `columnName` in the array

michnovka: sorry, my bad I was working on some badly lined version in my local repo.

michnovka (approved):
looks good now

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
greg0ire: No worries, and thanks for reviewing 👍

SenseException (approved):

PR #10409, opened 2023-01-15
Require that inheritance be declared when it is used

mpdude:
It seems to be possible to have two `@Entity` classes inherit from each other, but omit the `@InheritanceType` declaration at the root entity. This will lead to mapping information being loaded without complaints and the inheritance type being set to `NONE` for both classes.

At the same time, the subclass will have its field and association mappings set with the `inherited` and `declared` metadata, and also the `rootEntity` in the metadata will correctly point to the parent class.

So, the question is: Is it valid to have two entities inherit from each other without declaring an inheritance type?

The documentation does not explicitly address this.

This PR here adds a runtime validation failure for this case. Let’s see what the testsuites have to say.

Thoughts?


mpdude:
Replaced by #10431.

PR #10411, opened 2023-01-16
Fill in missing subclasses when loading ClassMetadata

### LINKED ISSUE #6558: Invalid discriminator maps should be detected during schema/mapping validation

### LINKED ISSUE #8127: Three level table class inheritance not loading second level data due to wrong query generation
I have a inheritance of three levels using class table inheritance like this:

Class Test
    
```php

    namespace App\Entities\Test;
    
    /**
     * Class Test
     * @package App\Entities\Test
     * @ORM\Entity
     * @ORM\DiscriminatorColumn(name="discr", type="string")
     * @ORM\DiscriminatorMap({"TestA" = "TestA", "TestB" = "TestB"})
     * @ORM\Table(name="test")
     */
    abstract class Test {
    
      /**
       * @var integer
       * @ORM\Column(type="integer")
       * @ORM\Id
       */
      protected $id;
    
      /**
       * @var string
       * @ORM\Column(type="string")
       */
      protected $columnTest;
    }
```
Class TestM extends Test

```php
   
    namespace App\Entities\Test;
    
    /**
     * Class TestM
     * @package App\Entities\Test
     * @ORM\Entity
     */
    abstract class TestM extends Test{
    
      /**
       * @var string
       * @ORM\Column(type="string")
       */
      protected $columnTestM;
    
    }
```

Class TestA extends TestM
```php
   
    namespace App\Entities\Test;
    
    /**
     * Class TestA
     * @package App\Entities\Test
     * @ORM\Entity
     */
    class TestA extends TestM{
    
      /**
       * @var string
       * @ORM\Column(type="string")
       */
      private $columnTestA;
    
      public function __construct(string $columnTest, string $columnTestM, string $columnTestA) {
        $this->columnTest = $columnTest;
        $this->columnTestM = $columnTestM;
        $this->columnTestA = $columnTestA;
      }
    
      /**
       * @return string
       */
      public function getColumnTest(): string {
        return $this->columnTest;
      }
    
      /**
       * @param string $columnTest
       */
      public function setColumnTest(string $columnTest): void {
        $this->columnTest = $columnTest;
      }
    
      /**
       * @return string
       */
      public function getColumnTestM(): string {
        return $this->columnTestM;
      }
    
      /**
       * @param string $columnTestM
       */
      public function setColumnTestM(string $columnTestM): void {
        $this->columnTestM = $columnTestM;
      }
    
      /**
       * @return string
       */
      public function getColumnTestA(): string {
        return $this->columnTestA;
      }
    
      /**
       * @param string $columnTestA
       */
      public function setColumnTestA(string $columnTestA): void {
        $this->columnTestA = $columnTestA;
      }
    }
```
I'm having problem because, when I'm going to retrieve the entity from my DB, it comes with the second level with no data, only the first and last levels comes with all the data. Notice that columnTestM is blank. Is it a bug or am I missing something? It's persisting all three levels with data, the problem is only when I have to get it. As an example I put the column's content with it's own name

    >>> print_r(\EntityManager::getRepository('App\Entities\Test\Test')->find(1));
    App\Entities\Test\TestA Object
    (
        [columnTestA:App\Entities\Test\TestA:private] => columnTestA
        [columnTestM:protected] => 
        [id:protected] => 1
        [columnTest:protected] => columnTest
    )

**EDIT:**

I put my MariaDB to log all queries to check what query was being generated and I noticed that, instead of generating a query that involves TestM, it tries to left join with TestB. Below the generated query:

```sql
SELECT 
    t0.id AS id_3,
    t0.column_test AS column_test_4,
    t0.discr,
    t1.column_test_a AS column_test_a_5,
    t2.column_test_b AS column_test_b_6
FROM
    test t0
        LEFT JOIN
    test_as t1 ON t0.id = t1.id
        LEFT JOIN
    test_bs t2 ON t0.id = t2.id
WHERE
    t0.id = 1
```

mpdude:
This PR suggests that for inheritance hierarchies, `abstract` `@Entity` classes shall be identified automatically, so that they do not have to be declared by users in their `@DiscriminatorMap` configuration.

#### Background

For entities and mapped superclasses that are part of an STI or JTI inheritance hierarchy, `ClassMetadata::$subClasses` contains a list of all entities that are a subclass of the current class. This field is of crucial importance since it is used to enumerate all entity classes and thus tables that need to be dealt with when loading or persisting data, creating the schema etc.

When this list is incomplete, possibly subtle or weird bugs like #9145 or #8127 may occur.

It is not possible in PHP to obtain a list of all existing subclasses for a given class. Also, our mapping drivers cannot provide that list, at least not in a general way and at acceptable cost. For that reason, the discriminator map (DM) at the root entity is used to "discover" all the entity (sub)classes that may need to be persisted in an inheritance tree.

#### Situation before/without this PR

Runtime validation of the discriminator map allows `abstract` `@Entity` classes to be omitted from it. Also, the error message given when a class missing from the DM is detected refers to `abstract` as an alternative:

https://github.com/doctrine/orm/blob/c5e4e41e0517887b7ce5ffd2d825f9c337bc5cd4/lib/Doctrine/ORM/Mapping/MappingException.php#L761-L768

From the users point of view, there is no point in listing abstract entity classes and assigning a discriminator value for them, when no such entities can be created and persisted in practice. When defining the mapping, the primary concern is to specify discriminator values for the entity classes being used, not to declare a full class hierarchy.

I suspect there may be other bugs lurking in the context of the feature from #1257, where an event listener was added to re-map interfaces to classes in the DM at runtime. When the entity class provided at runtime inherits from another abstract entity, this feature falls short as this extra class cannot be provided.

#### Suggested change

After `ClassMetadata` has been loaded for the root entity class in a STI/JTI hierarchy, fill in missing `subClasses` automatically. 

This does not amend, change or autogenerate the DM in any way. It only takes the entity classes given in the DM as starting pointers, works upwards through the class hierarchy from there on and identifies and adds `abstract` `@Entity` classes to the `subClasses` field.

#### Implementation remarks

There's a challenge since we need to identify all classes that are `abstract` and `@Entity` (especially not `@MappedSuperclass`). That would be easy if we had all the class metadata already loaded.

However, metadata loading happens through `doctrine/persistence` and goes from root classes towards child classes. So, we will always need to process "higher up" classes first and do not have child class mapping information available at that time.

It also seems unsafe to update parent class metadata at a later time when child class information has been loaded, given that the `ClassMetadata` is distributed with the `loadClassMetadata` event and is put into the metadata cache.

So, my choice was to put the initialization at the end of `doLoadMetadata`: That way, we can run it at the end of loading metadata for the root entity. 

* No need to update/"fixup" the root entity metadata later on
* Child classes can inherit the `subClass` field data, keeping only the classes relevant for them (→ we need just one "discovery" run for the root class)

The initialization happens _after_ the `loadClassMetadata` event (at the risk of having a possibly incomplete subclass list at the time it is run), to allow the mechanism from #1257 to kick in and provide classes _before_ we do the resolution.

Since we need to exclude mapped superclasses from the `subClass` list, I saw no other option than to make the driver do an extra load request. This, however, only has to happen for classes that we know are `abstract` (remember, non-abstract ones must be specified in the DM), and the resulting information will be cached with the root entity's metadata.

The driver thing maybe can be improved if drivers get a "shortcut" method to identify mapped superclasses, without doing all the other extra loading. Might be possible to add this through an add-on interface implemented by drivers.

#### Tests

For a start, this PR includes test cases that demonstrate #6558, #9145 and #8127 would be fixed _without_ having to provide complete (including `abstract` entities) DMs.

Ideas for other reasonable test cases are welcome!

#### Related issues

Closes #10389, closes #10387, closes #10388, fixes #8127, includes #9145, includes #6578, fixes #6558.

A follow-up fix for an edge case is in #10557.

A warning is given in [this comment](https://github.com/doctrine/orm/issues/10552#issuecomment-1451606300) that there may still be issues when event listeners are used to provide or change mapping information, which is currently not supported in the new `peekIfIsMappedSuperclass()` method.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -115,6 +115,7 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
             $class->setVersioned($parent->isVersioned);
             $class->setVersionField($parent->versionField);
             $class->setDiscriminatorMap($parent->discriminatorMap);
+            $class->addSubClasses($parent->subClasses);
             $class->setLifecycleCallbacks($parent->lifecycleCallbacks);
             $class->setChangeTrackingPolicy($parent->changeTrackingPolicy);
 
@@ -219,10 +220,15 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
             $this->addDefaultDiscriminatorMap($class);
         }
 
+        // During the following event, there may also be updates to the discriminator map as per GH-1257/GH-8402.
+        // So, we must not discover the missing subclasses before that.
+
         if ($this->evm->hasListeners(Events::loadClassMetadata)) {
             $eventArgs = new LoadClassMetadataEventArgs($class, $this->em);
             $this->evm->dispatchEvent(Events::loadClassMetadata, $eventArgs);
         }
+
+        $this->findAbstractEntityClassesNotListedInDiscriminatorMap($class);
 
         if ($class->changeTrackingPolicy === ClassMetadata::CHANGETRACKING_NOTIFY) {
             Deprecation::trigger(
@@ -336,6 +342,58 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
         }
 
         $class->setDiscriminatorMap($map);
+    }
+
+    private function findAbstractEntityClassesNotListedInDiscriminatorMap(ClassMetadata $rootEntityClass): void
+    {
+        // Only root classes in inheritance hierarchies need contain a discriminator map,
+        // so skip for other classes.
+        if (! $rootEntityClass->isRootEntity() || $rootEntityClass->isInheritanceTypeNone()) {
+            return;
+        }
+
+        $processedClasses = [$rootEntityClass->name => true];
+        foreach ($rootEntityClass->subClasses as $knownSubClass) {
+            $processedClasses[$knownSubClass] = true;
+        }
+
+        foreach ($rootEntityClass->discriminatorMap as $declaredClassName) {
+            // This fetches non-transient parent classes only
+            $parentClasses = $this->getParentClasses($declaredClassName);
+
+            foreach ($parentClasses as $parentClass) {
+                if (isset($processedClasses[$parentClass])) {
+                    continue;
+                }
+
+                $processedClasses[$parentClass] = true;
+
+                // All non-abstract entity classes must be listed in the discriminator map, and
+                // this will be validated/enforced at runtime (possibly at a later time, when the
+                // subclass is loaded, but anyways). Also, subclasses is about entity classes only.
+                // That means we can ignore non-abstract classes here. The (expensive) driver
+                // check for mapped superclasses need only be run for abstract candidate classes.
+                if (! (new ReflectionClass($parentClass))->isAbstract() || $this->peekIfIsMappedSuperclass($parentClass)) {
+                    continue;
+                }
+
+                // We have found a non-transient, non-mapped-superclass = an entity class (possibly abstract, but that does not matter)
+                $rootEntityClass->addSubClass($parentClass);
+            }
+        }
+    }
+
+    /** @param class-string $className */
+    private function peekIfIsMappedSuperclass(string $className): bool
+    {
+        // TODO can we shortcut this for classes that have already been loaded? can that be the case at all?
+        $reflService = $this->getReflectionService();
+        $class       = $this->newClassMetadataInstance($className);
+        $this->initializeReflection($class, $reflService);
+
+        $this->driver->loadMetadataForClass($className, $class);
+
+        return $class->isMappedSuperclass;
     }
 
     /**
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -386,7 +386,6 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
     /** @param class-string $className */
     private function peekIfIsMappedSuperclass(string $className): bool
     {
-        // TODO can we shortcut this for classes that have already been loaded? can that be the case at all?
         $reflService = $this->getReflectionService();
         $class       = $this->newClassMetadataInstance($className);
         $this->initializeReflection($class, $reflService);
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -397,7 +397,27 @@ class ClassMetadataInfo implements ClassMetadata
     public $parentClasses = [];
 
     /**
-     * READ-ONLY: The names of all subclasses (descendants).
+     * READ-ONLY: For classes in inheritance mapping hierarchies, this field contains the names of all
+     * <em>entity</em> subclasses of this class. These may also be abstract classes.
+     *
+     * This list is used, for example, to enumerate all necessary tables in JTI when querying for root
+     * or subclass entities, or to gather all fields comprised in an entity inheritance tree.
+     *
+     * For classes that do not use STI/JTI, this list is empty.
+     *
+     * Implementation note:
+     *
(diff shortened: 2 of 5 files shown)


greg0ire (dismissed):
Looks really good, my only qualm is about the remaining TODO, hopefully somebody has a solution for this.

beberlei (commented):
This is very clever.

Just to clarify, the `peekIfIsMappedSuperclass` method is only necessary if the discriminator map is "incomplete"?

Should we also add a validation of discriminator map completeness to the schema validator?

mpdude:
@beberlei

> Just to clarify, the peekIfIsMappedSuperclass method is only necessary if the discriminator map is "incomplete"?

Yes. 

After a root entity class metadata has been loaded, take all the entity classes declared in the DM as starting points. Work upwards from there on, using the driver's `isTransient` method to find only non-transient classes. For those, use reflection to check whether the class is abstract. Only for such candidates, have an extra driver request to figure out if it's a MappedSuperclass or not.

Looking at the AnnotationDriver, `isTransient` has to inspect annotations already, so a dedicated method might bring the final (conclusive) result at the same cost; that's an optimization I'd like to keep for a follow-up PR.

> Should we also add a validation of discriminator map completeness to the schema validator?

It's there already, and it is stricter (`abstract` `@Entity` must be declared) than the runtime validation. #9096 and #9142 have the discussion.

IMHO, if we get this PR here merged, we should relax the validator by re-applying #9096.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -338,6 +344,58 @@ private function addDefaultDiscriminatorMap(ClassMetadata $class): void
...
+                // We have found a non-transient, non-mapped-superclass = an entity class (possibly abstract, but that does not matter)
+                $rootEntityClass->addSubClass($parentClass);
+            }
+        }
+    }
+
+    /** @param class-string $className */
+    private function peekIfIsMappedSuperclass(string $className): bool
+    {
+        // TODO can we shortcut this for classes that have already been loaded? can that be the case at all?
```
mpdude: Thinking about this a bit more... Class metadata is loaded from root classes downwards, and so we find root entities/DMs before the other classes have their metadata loaded. 

So, I'd say there is no better/cheaper place where we might look for the information at that time...?

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -338,6 +344,58 @@ private function addDefaultDiscriminatorMap(ClassMetadata $class): void
...
+        $reflService = $this->getReflectionService();
+        $class       = $this->newClassMetadataInstance($className);
+        $this->initializeReflection($class, $reflService);
+
+        $this->driver->loadMetadataForClass($className, $class);
```
mpdude, line 393: I don't know if all that is really necessary for what we're trying to achieve here... But it's probably a safe default to start with, and may go away with a bit more sophisticated driver support.

mpdude:
@greg0ire Thought about the TODO a little, maybe there's nothing that needs to be done. See comment above.

mpdude:
Questions:

* Does the root entity have to be in the DM (under which conditions)? If so, why? For sure it need not go into the subclasses list.
* May there be entity classes (abstract or not) above the entity with the DM? Or do you need to declare an inheritance type as soon as one entity class inherits from another?

beberlei (approved):

greg0ire:
Thanks @mpdude !

Issue #10217, opened 2022-11-11
TABLE_PER_CLASS InheritanceType


Kenan525:
Why is there inheritanceType = "TABLE_PER_CLASS" and Persister for that kind of inheritance is missing ?

PR #10414, opened 2023-01-17
Remove concrete table inheritance

### LINKED ISSUE #7371: Documentation of Concrete Inheritance
The documentation of Concrete Inheritance when you use one table per class seems to be missing.

I only figured out you could use InheritanceType TABLE_PER_CLASS by looking at the source code.

mpdude:
This seems to be a leftover from early days, but never actually implemented or documented.

#10423 adds the deprecation notices in 2.14.

Closes #7371, references #10217, closes #10220.

### DIFF
--- a/docs/en/reference/php-mapping.rst
+++ b/docs/en/reference/php-mapping.rst
@@ -194,7 +194,6 @@ Inheritance Getters
 -  ``isInheritanceTypeNone()``
 -  ``isInheritanceTypeJoined()``
 -  ``isInheritanceTypeSingleTable()``
--  ``isInheritanceTypeTablePerClass()``
 -  ``isInheritedField($fieldName)``
 -  ``isInheritedAssociation($fieldName)``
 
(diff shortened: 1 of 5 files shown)


greg0ire (approved):
Not implemented in a decade. Let's indeed clean this up 👍

mpdude:
Since it technically a BC break, do we require deprecation warnings or similar in 2.x?

greg0ire:
It can't hurt to do so. If you do, you'll have to add something in `UPGRADE.md` about this, and also, you will have to edit `UPGRADE.md` in this PR with `s/deprecated/removed/`

mpdude:
I don't know how access to a `const` could be deprecated, and where can I find the pattern that this project uses to announce deprecations (e. g. a `@trigger_error(...)` or so?)

greg0ire:
For once, we have a clearly documented policy on this: https://www.doctrine-project.org/policies/deprecation.html :slightly_smiling_face: 

TL;DR: use `Deprecation::trigger()`, and when you can't, use `@deprecated`

SenseException (approved):

derrabus (approved):

PR #10416, opened 2023-01-17
2.9.x  inverse side orphan removal

markuswoessner:


greg0ire:
2.9.x is no longer maintained. If you have something to fix, target 2.14.x

mpdude:
@markuswoessner if I get it right, this adds missing tests for an existing feature?

Could you please rebase this to 2.14.x and leave a short comment at the top, so reviewers have a bit more context?

PR #10420, opened 2023-01-17
Fix #9095 by re-applying #9096

### LINKED ISSUE #9095: Regression: abstract entities in the middle of a hierarchy need to be part of discriminator map

mpdude:
Since #10411 has been merged, no more need to specify all intermediate abstract entity classes.

Thus, we can relax the schema validator check as requested in #9095. The reasons given in #9142 to keep it no longer apply.

Fixes #9095.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -253,7 +253,12 @@ class SchemaValidator
             }
         }
 
-        if (! $class->isInheritanceTypeNone() && ! $class->isRootEntity() && ! $class->isMappedSuperclass && array_search($class->name, $class->discriminatorMap, true) === false) {
+        if (
+            ! $class->isInheritanceTypeNone()
+            && ! $class->isRootEntity()
+            && ! $class->isMappedSuperclass
+            && array_search($class->name, $class->discriminatorMap, true) === false
+        ) {
             $ce[] = "Entity class '" . $class->name . "' is part of inheritance hierarchy, but is " .
                 "not mapped in the root entity '" . $class->rootEntityName . "' discriminator map. " .
                 'All subclasses must be listed in the discriminator map.';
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -253,7 +253,7 @@ public function validateClass(ClassMetadataInfo $class)
             }
         }
 
-        if (! $class->isInheritanceTypeNone() && ! $class->isRootEntity() && ! $class->isMappedSuperclass && array_search($class->name, $class->discriminatorMap, true) === false) {
+        if (! $class->isInheritanceTypeNone() && ! $class->isRootEntity() && ($class->reflClass !== null && ! $class->reflClass->isAbstract()) && ! $class->isMappedSuperclass && array_search($class->name, $class->discriminatorMap, true) === false) {
```
greg0ire: At what point should we use linebreaks :joy: ?

greg0ire (approved):

SenseException (approved):

greg0ire:
thanks @mpdude!

PR #10424, opened 2023-01-18
Fixup GH8127 test case

mpdude:
This removes the unnecessary "middle2" class and puts back in the effect of the data provider.

Both changes were accidentally committed when I was working on #10411 and just meant as experiments during debugging.


derrabus:
👍🏻  …once PHPCS is happy.

derrabus (approved):

PR #10426, opened 2023-01-18
Slight docblock improvements for `CM::parentClasses`

mpdude:
This explains a bit more precisely that `parentClasses` is about _entities_ only, and in what order to expect the classes to be listed.

I fear the trouble that comes with a full re-name to a more suitable name.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -394,7 +394,8 @@ class ClassMetadataInfo implements ClassMetadata
     public $isEmbeddedClass = false;
 
     /**
-     * READ-ONLY: The names of the parent classes (ancestors).
+     * READ-ONLY: The names of the parent <em>entity</em> classes (ancestors), starting with the
+     * nearest one and ending with the root entity class.
      *
      * @psalm-var list<class-string>
      */
@@ -2543,7 +2544,8 @@ class ClassMetadataInfo implements ClassMetadata
     }
 
     /**
-     * Sets the parent class names.
+     * Sets the parent class names. Only <em>entity</em> classes may be given.
+     *
      * Assumes that the class names in the passed array are in the order:
      * directParent -> directParentParent -> directParentParentParent ... -> root.
      *


lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -100,7 +100,11 @@ protected function onNotFoundMetadata($className)
     }
 
     /**
-     * {@inheritDoc}
+     * @param ClassMetadata      $class
+     * @param ?ClassMetadata     $parent
+     * @param bool               $rootEntityFound      True when there is another entity (non-mapped superclass) class above the current class in the PHP class hierarchy.
+     * @param list<class-string> $nonSuperclassParents All parent class names that are not marked as mapped superclasses, with the direct parent class being the first and the root entity class the last element.
+     * @psalm-suppress MoreSpecificImplementedParamType
```
greg0ire: We're discussing `psalm-suppress` internally but haven't reached an agreement yet. Until then, let's stick to using `psalm.xml` for this please.

mpdude: I've tried that, but was unable to use `referencedMethod` in `psalm.xm` to pin it down to this particular method (only worked for the entire file). So I chose the annotation to be more precise... 

I don't mind, just tell me what you prefer.

greg0ire: I have used the argument you are making in the internal discussion, but so far, not much feedback. I'd rather not unilaterally decide to switch to `@psalm-suppress` though, I need to get more maintainers to agree on it. Also see: https://github.com/vimeo/psalm/issues/9127

mpdude: Yeah, definetly agree about not making unilateral changes and will change it.

Just one more argument: The execption is much more visible when it is in the code at the place where it is needed. Much higher probability it will be cleaned up/removed once it is no longer necessary.

For `psalm.xml` entries, nobody is ever going to review them and clean up.

greg0ire: > Much higher probability it will be cleaned up/removed once it is no longer necessary

In the internal thread, I mentioned another argument: the existence of `find-unused-psalm-suppress`. But issue handlers are catching up it seems: https://github.com/vimeo/psalm/issues/9102

greg0ire: >  nobody is ever going to review them and clean up.

I guess this means I'm not a nobody :)

https://github.com/doctrine/orm/pull/10415

psalm.xml
```diff
@@ -248,5 +248,10 @@
                 <file name="lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php"/>
             </errorLevel>
         </ArgumentTypeCoercion>
+        <MoreSpecificImplementedParamType>
+            <errorLevel type="suppress">
+                <file name="lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php"/>
+            </errorLevel>
+        </MoreSpecificImplementedParamType>
```
greg0ire: Hey by the way, why do we have this ignore in the first place?

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
mpdude: Few people are as well organized and painstaking as you 🙂

psalm.xml
mpdude: I guess because not all versions of doctrine/persistence (where the interface is defined) have precise docblocks.

mpdude: It’s about the doLoadMetadata method

greg0ire: :thinking: but we are only running Psalm with the latest version of `doctrine/persistence`, right?

mpdude: ... which is not precise enough: 

https://github.com/doctrine/persistence/blob/7ee1ce5e5e0d462921d1eb7171f5637916e37663/src/Persistence/Mapping/AbstractClassMetadataFactory.php#L379

https://github.com/doctrine/persistence/pull/327

greg0ire (dismissed):

psalm.xml
greg0ire: Okay let me do a release.

mpdude: I see what you did there.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -100,7 +100,10 @@ protected function onNotFoundMetadata($className)
     }
 
     /**
-     * {@inheritDoc}
```
mpdude: What is the purpose of `inheritdoc`, by the way? Does it add any value when it is the only thing in the docblock?

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -100,7 +100,10 @@ protected function onNotFoundMetadata($className)
...
+     * @param ClassMetadata      $class
+     * @param ?ClassMetadata     $parent
+     * @param bool               $rootEntityFound      True when there is another entity (non-mapped superclass) class above the current class in the PHP class hierarchy.
+     * @param list<class-string> $nonSuperclassParents All parent class names that are not marked as mapped superclasses, with the direct parent class being the first and the root entity class the last element.
```
mpdude: We can now remove all this, right?

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
greg0ire: I don't know exactly, but I think `phpcs` lets you know when it is useless, so there's that. I think for instance if you have an `array` parameter, `phpcs` will whine that there is no documentation for it unless you have `inheritDoc`

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
greg0ire: Yes

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -100,7 +100,7 @@ protected function onNotFoundMetadata($className)
     }
 
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
```
greg0ire: Why change the case?

mpdude: by accident, fixed now

greg0ire (approved):

greg0ire:
Thanks @mpdude!

PR #10429, opened 2023-01-19
Review the documentation regarding entity inheritance

mpdude:
This PR gives the documentation chapter on inheritance a brush-up.

* It explains concepts that apply to STI and JTI in a common section at the beginning. 
* It more precisely (IMHO) states when inheritance has to be used, where the mapping configuration has to be placed and what it needs to include.
* Small orthographic fixups and typography fixes, no need to **SHOUT AT PEOPLE**.
* ~Add new typos~

### DIFF
--- a/docs/en/reference/inheritance-mapping.rst
+++ b/docs/en/reference/inheritance-mapping.rst
@@ -133,7 +133,7 @@ lowercase short name of each class as key.
 
 .. note::
 
-    Automatically generating the descriminator map is very expensive
+    Automatically generating the discriminator map is very expensive
     computation-wise. The mapping driver has to provide all classes
     for which mapping configuration exists, and those have to be
     loaded and checked whether they are part of the current
--- a/docs/en/reference/inheritance-mapping.rst
+++ b/docs/en/reference/inheritance-mapping.rst
@@ -254,7 +254,7 @@ Class Table Inheritance
 
 `Class Table Inheritance <https://martinfowler.com/eaaCatalog/classTableInheritance.html>`_
 is an inheritance mapping strategy where each class in a hierarchy
-is mapped to several tables: Its own table and the tables of all
+is mapped to several tables: its own table and the tables of all
 parent classes. The table of a child class is linked to the table
 of a parent class through a foreign key constraint.
 
--- a/docs/en/reference/inheritance-mapping.rst
+++ b/docs/en/reference/inheritance-mapping.rst
@@ -136,8 +136,8 @@ lowercase short name of each class as key.
     Automatically generating the discriminator map is very expensive
     computation-wise. The mapping driver has to provide all classes
     for which mapping configuration exists, and those have to be
-    loaded and checked whether they are part of the current
-    inheritance hierarchy. The resulting map, however, can be kept
+    loaded and checked against  the current inheritance hierarchy
+    to see if they are part of it. The resulting map, however, can be kept
     in the metadata cache.
 
 Single Table Inheritance
--- a/docs/en/reference/inheritance-mapping.rst
+++ b/docs/en/reference/inheritance-mapping.rst
@@ -345,6 +345,11 @@ relationship defined in that mapped superclass or trait.
 
 It is not supported to use overrides in entity inheritance scenarios.
 
+.. note::
+
+    When using traits, make sure not to miss the warnings given in the
+    :doc:`Limitations and Known Issues<reference/limitations-and-known-issues>` chapter.
+
 
 Association Override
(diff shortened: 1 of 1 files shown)


docs/en/reference/inheritance-mapping.rst
```diff
@@ -90,14 +97,46 @@ for the entity subclass. All the mappings from the mapped
...
+inheritance hierarchy. That includes the root as well as any intermediate
+entity classes, given they are not abstract.
+
+The names of the classes in the discriminator map do not need to be fully
+qualified if the classes are contained in the same namespace as the entity
+class on which the discriminator map is applied.
+
+If no discriminator map is provided, then the map is generated
+automatically. The automatically generated discriminator map contains the
+lowercase short name of each class as key.
```
mpdude, line 132: I have kept that, but in fact, I am not sure if it is even true?

mpdude: It fact it is.

docs/en/reference/inheritance-mapping.rst
```diff
@@ -14,6 +17,10 @@ Mapped superclasses, just as regular, non-mapped classes, can
 appear in the middle of an otherwise mapped inheritance hierarchy
 (through Single Table Inheritance or Class Table Inheritance).
 
+No database table will be created for a mapped superclasse itself,
```
greg0ire:
```suggestion
No database table will be created for a mapped superclass itself,
```

docs/en/reference/inheritance-mapping.rst
```diff
@@ -90,14 +97,55 @@ for the entity subclass. All the mappings from the mapped
...
+descendants of this root entity.
+
+On the root entity class, ``#[InheritanceType]``,
+``#[DiscriminatorColumn]`` and ``#[DiscriminatorMap]`` must be specified.
+
+``#[InheritanceType]`` specifies one of the two available inheritance
+mapping strategies that are explained in the following sections.
+
+``#[DiscriminatorColumn]`` designates the so-called discriminator column.
+This is an extra column in the database that keeps information about which
```
greg0ire: Let's talk about the table for the root entity class instead of talking about the database? One may think there is only one such column in the whole database.

```suggestion
This is an extra column in the table that keeps information about which
```

docs/en/reference/inheritance-mapping.rst
```diff
@@ -90,14 +97,55 @@ for the entity subclass. All the mappings from the mapped
...
+qualified if the classes are contained in the same namespace as the entity
+class on which the discriminator map is applied.
+
+If no discriminator map is provided, then the map is generated
+automatically. The automatically generated discriminator map contains the
+lowercase short name of each class as key.
+
+.. note::
+
+    Automatically generating the descriminator map is very expensive
```
greg0ire:
```suggestion
    Automatically generating the discriminator map is very expensive
```

docs/en/reference/inheritance-mapping.rst
```diff
@@ -90,14 +97,55 @@ for the entity subclass. All the mappings from the mapped
...
+    computation-wise. The mapping driver has to provide all classes
+    for which mapping configuration exists, and those have to be
+    loaded and checked whether they are part of the current
+    inheritance hierarchy. The resulting map, however, can be kept
```
greg0ire:
```suggestion
    loaded and checked against  the current inheritance hierarchy
    to see if they are part of it. The resulting map, however, can be kept
```

docs/en/reference/inheritance-mapping.rst
```diff
@@ -216,20 +246,21 @@ For Single-Table-Inheritance to work in scenarios where you are
...
+null values. Columns that have ``NOT NULL`` constraints have to be on
 the root entity of the single-table inheritance hierarchy.
 
 Class Table Inheritance
 -----------------------
 
 `Class Table Inheritance <https://martinfowler.com/eaaCatalog/classTableInheritance.html>`_
 is an inheritance mapping strategy where each class in a hierarchy
-is mapped to several tables: its own table and the tables of all
+is mapped to several tables: Its own table and the tables of all
```
greg0ire: I don't think that fix is correct: the second part isn't even a valid, standalone sentence (it has no verb). Even if it did, I would keep this lowercase.

mpdude: okay, will fix

mpdude:
All feedback addressed

greg0ire (approved):

greg0ire:
Thanks @mpdude !

PR #10431, opened 2023-01-19
Deprecate undeclared entity inheritance

mpdude:
Inheritance has to be declared as soon as one entity class extends (directly or through middle classes) another one.

This is also pointed out in the opening comment for #8348: 

> Entities are not allowed to extend from entities without an inheritence mapping relationship (Single Table or Joined Table inheritance). [...] While Doctrine so far allowed these things, they are fragile and will break on certain scenarios.

Throwing an exception in case of this misconfiguration is nothing we should do light-heartedly, given that it may surprise users in a bugfix or feature release. So, we should start with a deprecation notice  and make this an exception in 3.0. The documentation is updated accordingly at #10429.

Catching missing inheritance declarations early on is important to avoid weird errors further down the road, giving users a clear indication of the root cause.

In case you are affected by this, please understand that although things "previously worked" for you, you have been using the ORM outside of what it was designed to do. That may have worked in simple cases, but may also have caused invalid results (wrong or missing data after hydration?) that possibly went unnoticed in subtle cases.

Update: For one example of strange things that may happen down the road, see https://github.com/doctrine/orm/issues/3037#issuecomment-1605900686.


derrabus:
As always, we don't introduce deprecations in bugfix releases. Please target 2.15.

mpdude:
Rebased onto 2.15.x, and tied a knot in my handkerchief.

tests/Doctrine/Tests/Models/DDC5934/DDC5934BaseContract.php
```diff
@@ -8,13 +8,13 @@
...
 use Doctrine\ORM\Mapping\Column;
-use Doctrine\ORM\Mapping\Entity;
 use Doctrine\ORM\Mapping\GeneratedValue;
 use Doctrine\ORM\Mapping\Id;
 use Doctrine\ORM\Mapping\ManyToMany;
+use Doctrine\ORM\Mapping\MappedSuperclass;
 
-/** @Entity */
-#[Entity]
+/** @MappedSuperclass() */
```
derrabus:
```suggestion
/** @MappedSuperclass */
```

UPGRADE.md
```diff
@@ -10,6 +10,11 @@ now deprecated and will be removed in 3.0.
 * Using `TABLE_PER_CLASS` as the value for the `InheritanceType` attribute or annotation
   or in XML configuration files.
 
+## Deprecated undeclared entity inheritance
+
+As soon as an entity class inherits from another entity class, inheritance has to 
+be declared by adding the appropriate configuration for the root entity.
```
derrabus: We usually add new messages at the top.

derrabus (changes requested):

mpdude:
Feedback addressed

derrabus (approved):

derrabus:
Thanks!

mpdude:
💪🏻

Let’s hope it will not cause too much confusion and problems for users in 2.15.

Shall I create a dedicated PR to make the switch in 3.0? Requires changes to be merged up before, though.

PR #10433, opened 2023-01-19
Make sure tests from #7837 are actually run

mpdude:
I noticed that when I remove this line

https://github.com/Ocramius/doctrine2/blob/98e557b68eeecd7f266f1e4628007aed2714026c/lib/Doctrine/ORM/Tools/Pagination/Paginator.php#L169

added in #7837 the corresponding tests still pass.

This is because the relevant branch in `\Doctrine\ORM\Tools\Pagination\Paginator::getIterator()` is not reached. A new necessary condition was added briefly afterwards in #7863, and that caused the tests to miss the relevant section.

With the change suggested here, removing the line will make the specific test fail again.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7820Test.php
@@ -71,7 +71,8 @@ class GH7820Test extends OrmFunctionalTestCase
     {
         $query = $this->_em->getRepository(GH7820Line::class)
             ->createQueryBuilder('l')
-            ->orderBy('l.lineNumber', Criteria::ASC);
+            ->orderBy('l.lineNumber', Criteria::ASC)
+            ->setMaxResults(100);
 
         self::assertSame(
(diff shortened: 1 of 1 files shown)


greg0ire:
:thinking: codecov says the Coverage is not affected

mpdude:
For a start, can you confirm my findings?

What exactly does codecov measure?

mpdude:
I think this comment anticipated what actually happened, we’re testing the walker thing but did not add the limit.

https://github.com/doctrine/orm/pull/7863#issuecomment-554578313

mpdude:
#9519 already fixed this problem for another test, but missed the one I found

mpdude:
Reviewed all other tests that deal with the Paginator and they seem fine.

greg0ire (approved):
The lines are executed in other tests, but removing the line with `expireQueryCache` does indeed not make the test in `GH7820` fail prior to this PR.

greg0ire:
Thanks @mpdude !

PR #10438, opened 2023-01-20
Merge 2.14.x up into 2.15.x

greg0ire:


lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -394,15 +451,21 @@ private function addInheritedFields(ClassMetadata $subClass, ClassMetadata $pare
...
+            // of mapped superclasses) above, then declare this current entity class as the source of
+            // the relationship.
+            // According to the definitions given in https://github.com/doctrine/orm/pull/10396/,
+            // this is the case <=> ! isset($mapping['inherited']).
+            if (! isset($mapping['inherited'])) {
                 if ($mapping['type'] & ClassMetadata::TO_MANY && ! $mapping['isOwningSide']) {
                     throw MappingException::illegalToManyAssociationOnMappedSuperclass($parentClass->name, $field);
                 }
 
                 $mapping['sourceEntity'] = $subClass->name;
```
greg0ire, line 466: @mpdude this line seems to be causing the error below, but I don't understand why: both `$mapping['sourceEntity']` and `$subClass->name` are reported as `class-string` by `@psalm-trace`, yet the error message mentions `class-string|mixed`. Also, `inherited` is reported as mixed…

Help appreciated, but first, I'd like to know if I fixed the conflict properly according to you.

greg0ire: Ah it looks like removing the call to `addMappingInheritanceInformation` also fixes this. It uses `mixed[]` so I think that might explain a lot.

mpdude: Conflict resolution looks good :+1:

mpdude: https://github.com/doctrine/orm/pull/10397/files#diff-3e426fe47b58d2d4b970b1b7a0cbf5999af950e4f84e48dce4bee5061b57ed31R361 causes the Psalm error IMHO, `mixed[]` makes Psalm assume the type may change (array shape is lost/amended?)

greg0ire: Here is the output of `@psalm-trace` before and after:

```
ERROR: Trace - lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:455:13 - $mapping: array<array-key, mixed> (see https://psalm.dev/224)
            /** @psalm-trace $mapping */
            $this->addMappingInheritanceInformation($mapping, $parentClass);


ERROR: Trace - lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:463:13 - $mapping: array{inherited?: mixed, sourceEntity?: class-string|mixed, ...<array-key, mixed>} (see https://psalm.dev/224)
```

I'd expect something more precise on the first call.

greg0ire: I think it might have to do with the `foreach`, because when I copy `$mapping` to a separate variable, then `psalm-trace` shows what I expect.

greg0ire: Copying `$mapping` to some random variable then using that random variable fixes the issue.

greg0ire: What's funny is that this is the very method that made me lose a lot of time the other day in https://github.com/doctrine/orm/pull/10405/files#r1072806010

greg0ire: I wonder if this should be considered a bug: https://psalm.dev/r/a1767b6886
At the position where I put `psalm-trace`, there is no way there's a `test` key in that array.

greg0ire: Not a bug apparently, that's just how `psalm-trace` works… it will describe the variable after the assignment

mpdude: Here's a more exact snippet: https://psalm.dev/r/ca012a1298

greg0ire: Here is one that is even closer I think: https://psalm.dev/r/84ffd92a4c

greg0ire: And something closerer: https://psalm.dev/r/ebf182b2e6

greg0ire: I'll open a bug from this and reference it in `psalm.xml`

mpdude (approved):
Looks good from my point of view.

Thank you!

PR #10439, opened 2023-01-20
Fix PersistentCollection::get() returning entity not in the collection

### LINKED ISSUE #8607: Extra lazy PersistentCollection::get() with indexBy doesn't check mappedBy value

Spilky:
Fixes https://github.com/doctrine/orm/issues/8607
Closes https://github.com/doctrine/orm/pull/9035

### DIFF
--- a/lib/Doctrine/ORM/PersistentCollection.php
+++ b/lib/Doctrine/ORM/PersistentCollection.php
@@ -432,7 +432,13 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
             assert($this->em !== null);
             assert($this->typeClass !== null);
             if (! $this->typeClass->isIdentifierComposite && $this->typeClass->isIdentifier($this->association['indexBy'])) {
-                return $this->em->find($this->typeClass->name, $key);
+                $entity = $this->em->find($this->typeClass->name, $key);
+
+                if ($entity === null || ! $this->isOwned($entity)) {
+                    return null;
+                }
+
+                return $entity;
             }
 
             return $this->getUnitOfWork()->getCollectionPersister($this->association)->get($this, $key);
@@ -701,6 +707,19 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
         if ($newlyAddedDirtyObjects) {
             $this->restoreNewObjectsInDirtyCollection($newlyAddedDirtyObjects);
         }
+    }
+
+    /** @param object $entity */
+    private function isOwned($entity): bool
+    {
+        if (! isset($this->association['mappedBy'])) {
+            return false;
+        }
+
+        assert($this->typeClass !== null);
+        $owner = $this->typeClass->getFieldValue($entity, $this->association['mappedBy']);
+
+        return $owner === $this->getOwner();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


greg0ire:
Converting to draft until the build is green.

derrabus:
Thank you for the PR. The test case reflects the buggy behavior very well. What I don't like about the implementation is that we still fetch and hydrate the child entity just to throw it away after the check. It should be possible to combine PK and FK into a query, so that the child entity is not found if the parent entity does not own it.

mpdude:
Or you can get the Unit of Work, check the identity map and use the entity if it is in there.

If it is not, I don’t know if it is better to build and parse a “smart” DQL query (that’s what was suggested?) or just fetch by ID and check?

Might depend on the anticipated hit ratio? If we’re going to fetch/hydrate most of the time anyways…?

PR #10440, opened 2023-01-21
Migrate entity manager to PHP 8 syntax

greg0ire:


lib/Doctrine/ORM/EntityManager.php
```diff
@@ -119,14 +114,18 @@ class EntityManager implements EntityManagerInterface
...
      * Creates a new EntityManager that operates on the given database connection
      * and uses the given Configuration and EventManager implementations.
+     *
+     * @param Connection $conn The database connection used by the EntityManager.
      */
-    public function __construct(Connection $conn, Configuration $config, EventManager|null $eventManager = null)
-    {
+    public function __construct(
+        private readonly Connection $conn,
+        Configuration $config,
```
derrabus: Why don't we use CPP for `$config` as well?

SenseException (approved):

PR #10441, opened 2023-01-21
Use the right property

greg0ire:
association mappings do not have a `fetchMode` property, they have a `fetch` property.

Found while working on https://github.com/doctrine/orm/pull/10405


mpdude:
Whatever the test tries to do there, since this is also the default mode. It wrote to the wrong array key, but the right key contained the same mode as default value.

derrabus (approved):

PR #10442, opened 2023-01-22
Add embedded class mapping array shape

greg0ire:
This should be replaced with a DTO in the next major. To be able to have something usable, I had to move the validation of the DTO (checking that it has a "class" attribute) to mapEmbedded, which happens earlier than doLoadMetadata(). It is unclear to me why it was not put here in the first place in https://github.com/doctrine/orm/pull/1133

Cc @mpdude

### DIFF
--- a/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
@@ -464,29 +464,6 @@ class ClassMetadataFactoryTest extends OrmTestCase
         self::assertSame($entityManager, $property->getValue($classMetadataFactory));
     }
 
-    /** @group DDC-3305 */
-    public function testRejectsEmbeddableWithoutValidClassName(): void
-    {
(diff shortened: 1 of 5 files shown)


tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -469,22 +469,13 @@ public function testRejectsEmbeddableWithoutValidClassName(): void
...
-
         $this->expectException(MappingException::class);
         $this->expectExceptionMessage('The embed mapping \'embedded\' misses the \'class\' attribute.');
-
-        $cmf->getMetadataFor($metadata->name);
+        $metadata->mapEmbedded([
+            'fieldName'    => 'embedded',
+            'class'        => '',
+            'columnPrefix' => false,
+        ]);
```
SenseException: This technically isn't a ClassMetadataFactory test anymore and belongs to the ClassMetadata tests

SenseException (changes requested):

SenseException (approved):

PR #10444, opened 2023-01-23
Make Paginator-internal query cacheable in the query cache

mpdude:
Make the `Paginator`-internal query (`... WHERE ... IN (id, id2, id3...)`) cacheable in the query cache again.

#### Background

When the Paginator creates the internal subquery that does the actual result limiting, it has to take DBAL type conversions for the identifier column of the paginated root entity into account (#7820, fixed in #7821).

In order to perform this type conversion, we need to know the DBAL type class for the root entity's `#[Id]`, and we have to figure it out based on a given (arbitrary) DQL query. This requires DQL parsing and inspecting the AST, so #7821 placed the conversion code in the `WhereInWalker` where all the necessary information is available.

The problem is that type conversion has to happen every time the paginator is run, but the query that results from running `WhereInWalker` would be kept in the query cache. This was reported in #7837 and fixed by #7865, by making this particular query expire every time. The query must not be cached, since the necessary ID type conversion happens as a side-effect of running the `WhereInWalker`. 

#### Current status

The Paginator internal query that uses `WhereInWalker` has its DQL re-parsed and transformed in every request.

#### Suggested solution

This PR moves the code that determines the DBAL type out of `WhereInWalker` into a dedicated SQL walker class, `RootTypeWalker`. 

`RootTypeWalker` uses a ~hack~  clever trick to report the type back: It sets the type as the resulting "SQL" string. The benefit is that `RootTypeWalker` results can be cached in the query cache themselves. Only the first time a given DQL query has to be paginated, we need to run this walker to find out the root entity's ID type. After that, the type will be returned from the query cache.

With the type information being provided, `Paginator` can take care of the necessary conversions by itself. This happens every time the Paginator is used.

The internal query that uses `WhereInWalker` can be cached again since it no longer has side effects.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
@@ -21,7 +21,9 @@ use Traversable;
 use function array_key_exists;
 use function array_map;
 use function array_sum;
+use function assert;
 use function count;
+use function is_string;
 
 /**
  * The paginator can handle various complex scenarios with DQL.
@@ -160,9 +162,10 @@ class Paginator implements Countable, IteratorAggregate
             $this->appendTreeWalker($whereInQuery, WhereInWalker::class);
             $whereInQuery->setHint(WhereInWalker::HINT_PAGINATOR_ID_COUNT, count($ids));
             $whereInQuery->setFirstResult(0)->setMaxResults(null);
-            $whereInQuery->setParameter(WhereInWalker::PAGINATOR_ID_ALIAS, $ids);
             $whereInQuery->setCacheable($this->query->isCacheable());
-            $whereInQuery->useQueryCache(false);
+
+            $databaseIds = $this->convertWhereInIdentifiersToDatabaseValues($ids);
+            $whereInQuery->setParameter(WhereInWalker::PAGINATOR_ID_ALIAS, $databaseIds);
 
             $result = $whereInQuery->getResult($this->query->getHydrationMode());
         } else {
@@ -264,5 +267,24 @@ class Paginator implements Countable, IteratorAggregate
         }
 
         $query->setParameters($parameters);
+    }
+
+    /**
+     * @param mixed[] $identifiers
+     *
+     * @return mixed[]
+     */
+    private function convertWhereInIdentifiersToDatabaseValues(array $identifiers): array
+    {
+        $query = $this->cloneQuery($this->query);
+        $query->setHint(Query::HINT_CUSTOM_OUTPUT_WALKER, RootTypeWalker::class);
+
+        $connection = $this->query->getEntityManager()->getConnection();
+        $type       = $query->getSQL();
+        assert(is_string($type));
+
+        return array_map(static function ($id) use ($connection, $type) {
+            return $connection->convertToDatabaseValue($id, $type);
+        }, $identifiers);
     }
 }
--- a/lib/Doctrine/ORM/Tools/Pagination/RootTypeWalker.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/RootTypeWalker.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Tools\Pagination;
+
+use Doctrine\ORM\Query\AST;
+use Doctrine\ORM\Query\SqlWalker;
+use Doctrine\ORM\Utility\PersisterHelper;
+use RuntimeException;
+
+use function count;
+use function reset;
+
+/**
+ * Infers the DBAL type of the #Id (identifier) column of the given query's root entity, and
+ * returns it in place of a real SQL statement.
+ *
+ * Obtaining this type is a necessary intermediate step for \Doctrine\ORM\Tools\Pagination\Paginator.
+ * We can best do this from a tree walker because it gives us access to the AST.
+ *
+ * Returning the type instead of a "real" SQL statement is a slight hack. However, it has the
+ * benefit that the DQL -> root entity id type resolution can be cached in the query cache.
+ */
+class RootTypeWalker extends SqlWalker
+{
+    public function walkSelectStatement(AST\SelectStatement $AST): string
+    {
+        // Get the root entity and alias from the AST fromClause
+        $from = $AST->fromClause->identificationVariableDeclarations;
+
+        if (count($from) > 1) {
+            throw new RuntimeException('Can only process queries that select only one FROM component');
+        }
+
+        $fromRoot            = reset($from);
+        $rootAlias           = $fromRoot->rangeVariableDeclaration->aliasIdentificationVariable;
+        $rootClass           = $this->getMetadataForDqlAlias($rootAlias);
+        $identifierFieldName = $rootClass->getSingleIdentifierFieldName();
+
+        return PersisterHelper::getTypeOfField(
+            $identifierFieldName,
+            $rootClass,
+            $this->getQuery()
+                ->getEntityManager()
+        )[0];
+    }
+}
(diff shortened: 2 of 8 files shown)


derrabus:
> (to be written)

Marking as draft for now. 🙂

mpdude:
> Marking as draft for now. 🙂

@derrabus will you ignore PRs as long as they are drafts, and should I switch back to "ready for review" when I would like to get feedback (even if it is not :100: ready yet)?

derrabus:
> @derrabus will you ignore PRs as long as they are drafts

I can ignore non-draft PRs just as good as drafts, no worries. Don't hack my backlog. ✌🏻

tests/Doctrine/Tests/ORM/Tools/Pagination/WhereInWalkerTest.php
```diff
@@ -4,274 +4,79 @@
...
-            $whereInQuery->getSQL()
-        );
-
-        $this->assertPaginatorWhereInParameterToBe(
-            $whereInQuery,
-            [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
-        );
-    }
-
-    public function testWillReplaceBoundQueryIdentifiersWithConvertedTypesAsPerIdentifierMapping(): void
```
mpdude, line 237: This is no longer the `WhereInWalker`'s responsibility

tests/Doctrine/Tests/ORM/Tools/Pagination/WhereInWalkerTest.php
```diff
@@ -4,274 +4,79 @@
...
-        $whereInQuery->setHint(WhereInWalker::HINT_PAGINATOR_ID_COUNT, 3);
-        $whereInQuery->setParameter(WhereInWalker::PAGINATOR_ID_ALIAS, ['foo', 'bar', 'baz']);
-
-        $this->assertPaginatorWhereInParameterToBe(
-            $whereInQuery,
-            ['sbb', 'one', 'onm']
-        );
-    }
-
-    public function testWillReplaceBoundQueryIdentifiersWithConvertedTypesAsPerAssociatedEntityIdentifierMapping(): void
```
mpdude, line 252: This is no longer the `WhereInWalker`'s responsibility

mpdude:
Ready for review. Please advise how to resolve the Psalm issue.

greg0ire:
Rebasing on ed56f42cd should fix the issue.

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -265,4 +268,23 @@ private function unbindUnusedQueryParams(Query $query): void
 
         $query->setParameters($parameters);
     }
+
+    /**
+     * @param mixed[] $identifiers
+     *
+     * @return mixed[]
+     */
+    private function convertWhereInIdentifiersToDatabaseValue(array $identifiers): array
```
greg0ire: Unless I'm missing something this should be plural.

```suggestion
    private function convertWhereInIdentifiersToDatabaseValues(array $identifiers): array
```

mpdude:
Computer says no.

> Error: lib/Doctrine/ORM/Tools/Pagination/WhereInWalker.php:0:0: UnusedBaselineEntry: Baseline for issue "MissingClosureReturnType" has 1 extra entry. (see https://psalm.dev/316)

Shall I update the baseline in/for this PR? That's probably gonna cause a merge conflict down the road.

lib/Doctrine/ORM/Tools/Pagination/RootTypeWalker.php
```diff
@@ -0,0 +1,48 @@
...
+ */
+class RootTypeWalker extends SqlWalker
+{
+    public function walkSelectStatement(AST\SelectStatement $AST): string
+    {
+        // Get the root entity and alias from the AST fromClause
+        $from = $AST->fromClause->identificationVariableDeclarations;
+
+        if (count($from) > 1) {
+            throw new RuntimeException('Can only process queries that select only one FROM component');
```
greg0ire, line 33: Probably a dumb question, but what guarantees that it is the case when using the paginator?

greg0ire:
> Shall I update the baseline in/for this PR? That's probably gonna cause a merge conflict down the road.

Oh right, didn't realised but yes, there is one less entry. I wouldn't worry about the merge conflicts.

tests/Doctrine/Tests/ORM/Tools/Pagination/RootTypeWalkerTest.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function testResolveTypeMapping(string $dqlQuery, string $expectedType): void
+    {
+        $query = $this->entityManager->createQuery($dqlQuery);
+        $query->setHint(Query::HINT_CUSTOM_OUTPUT_WALKER, RootTypeWalker::class);
+
+        self::assertSame($expectedType, $query->getSQL());
+    }
+
+    public function exampleQueries(): Generator
```
greg0ire, line 38:
```suggestion
    /** @return iterable<array{string, string}> */
    public function exampleQueries(): Generator
```

lib/Doctrine/ORM/Tools/Pagination/RootTypeWalker.php
mpdude: I don't know. The condition/limitation was there before, only in another class.

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
mpdude: fixed

lib/Doctrine/ORM/Tools/Pagination/RootTypeWalker.php
greg0ire: Oh ok that's probably fine then.

lib/Doctrine/ORM/Tools/Pagination/RootTypeWalker.php
```diff
@@ -0,0 +1,48 @@
...
+ * Infers the DBAL type of the #Id (identifier) column of the given query's root entity, and
+ * returns it in place of a real SQL statement.
+ *
+ * Obtaining this type is a necessary intermediate step for \Doctrine\ORM\Tools\Pagination\Paginator.
+ * We can best do this from a tree walker because it gives us access to the AST.
+ *
+ * Returning the type instead of a "real" SQL statement is a slight hack. However, it has the
+ * benefit that the DQL -> root entity id type resolution can be cached in the query cache.
+ */
+class RootTypeWalker extends SqlWalker
```
greg0ire: Given the hacky nature of this class, I wonder if it should maybe be final or internal, since we probably don't want people to build more on this.

greg0ire (dismissed):

lib/Doctrine/ORM/Tools/Pagination/RootTypeWalker.php
mpdude: Good point, will do

mpdude:
@greg0ire Class is final now, please re-approve

greg0ire (approved):

SenseException (approved):

mpdude:
🎉

greg0ire:
Thanks @mpdude ! Right after clicking merged, I noticed the target branch. Not sure it should really have been 2.14.x but oh well…

mpdude:
I think that’s fine. I see no risk for breakage and the same argument as in https://github.com/doctrine/orm/pull/8797#issuecomment-877588988 applies

mpdude:
Maybe we can also bring #8797 to an end? It’s closely related, and I’ll resolve merge conflicts shortly.

PR #10445, opened 2023-01-23
Psalm 5.5.0

derrabus:
With this release, Psalm stops writing `occurrences="X"` into the baseline which is why I had to regenerate the whole thing. This is going to be fun when merging up! 🥲

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -3,6 +3,8 @@
     errorLevel="2"
     phpVersion="8.2"
     resolveFromConfigFile="true"
+    findUnusedBaselineEntry="true"
+    findUnusedCode="false"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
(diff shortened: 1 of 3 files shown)


psalm.xml
```diff
@@ -3,6 +3,8 @@
     errorLevel="2"
     phpVersion="8.2"
     resolveFromConfigFile="true"
+    findUnusedBaselineEntry="true"
```
derrabus, line 6: I need to set this flag to avoid a deprecation warning. It defaults to `false`, but I think enabling it will help us keep the baseline clean. 🤞🏻

psalm.xml
```diff
@@ -3,6 +3,8 @@
...
+    findUnusedCode="false"
```
derrabus, line 7: Another flag that needs to be set now. Setting it to `true` causes a lot of new errors, which is why I'd like to do this in a later PR (if we want to do it at all).

psalm.xml
greg0ire: Yeah, we are the ones (well I) [asking for that feature](https://github.com/vimeo/psalm/issues/9102) so better use it at some point :stuck_out_tongue:

greg0ire (approved):

PR #10448, opened 2023-01-23
Migrate ParserResult to PHP 8 syntax

greg0ire:


SenseException (approved):
My inner Monk would prefer to remove the underscore prefixes in the property names before merging.

greg0ire:
@SenseException I considered it, but these are private properties, so nothing prevents from doing this on 2.15.x. Let's try to find them throughout the project?

Issue #10417, opened 2023-01-17
Weird conditional skipping of mapping information in attribute/annotations driver


mpdude:
I am looking into an issue where an invalid association definition on a mapped superclass is not rejected as it should be.

It turns out the assocation is not even reported by the mapping driver, due to the following lines of code.

https://github.com/doctrine/orm/blob/69c7791ba256d947ddb1aafe5f2439ab31704937/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L345-L357

This is present in the attribute driver as well. To add insult to injury, the thing I am trying to chase depends on wheter an association is mapped in a `public` or `private` property.

I have seen this exact piece of code has been questioned in https://github.com/doctrine/orm/issues/5744 before, but neither that nor the reference to #4198 help.

Can anybody help reasoning about this – what's the purpose of these continuations, and what does it have to do with `private` in mapped superclasses?

Maybe it has to do with that the Reflection API also reports non-private properties inherited from base classes, but I don't get it. 

How do we want to treat such inherited properties, and/or can we tell whether a property was just inherited or inhertied + redeclared?

mpdude:
X-Ref #9702

mpdude:
#10449, #10450 and #10454 show different ways how the conditions from the OP fail.

mpdude:
#10455 tries to fix the mapping driver.

mpdude:
Closing, since #10455 has been merged and fixes this.

PR #10449, opened 2023-01-24
Failing test: Illegal to-many association on MappedSuperclass not rejected

mpdude:
**Closing** since this is no longer an error condition we need to check for.

#10473 relaxed the runtime check that mapped superclasses must not use one-to-many associations. It turned the check the other way round, so that no mapped superclass may be used as the targetEntity of an association. (In addition, #10554 also made it a "lazy" (non-runtime) check in the schema validator.)

<hr>

This test case demonstrates a situation where an invalid to-many association on a mapped superclass is not being detected.

Since at least the annotations and attribute drivers currently skip non-private properties from mapped superclasses (see #10417), the `ClassMetadataFactory` is not even aware of the association on the mapped superclass and so cannot reject it.

IMHO, we should bail out as soon as we detect this on the mapped superclass. I don't see how we could proceed at that time and try to rescue/fixup the situation afterwards.

In addition to that, maybe an association with a `target` that is a mapped superclass should be rejected in its own right?

More details on what happens down the road, just for the sake of completeness:

When the mapping driver processes the entity class with the inherited association, PHP reflection will report the association property. Since the association was not reported for the mapped superclass, the driver thinks (see #10417) this is a "new" (non-inherited) field and reports to the `ClassMetadataFactory`.

The association will show up as being from `GH10449Entity` to `GH10449ToManyAssociationTarget`. On `GH10449ToManyAssociationTarget`, it will refer back to `GH10449MappedSuperclass`.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10449Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10449Test.php
@@ -1,0 +1,92 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\Mapping\MappingException;
+use Doctrine\Tests\OrmTestCase;
+
+use function get_parent_class;
+use function method_exists;
+
+class GH10449Test extends OrmTestCase
+{
+    public function testToManyAssociationOnMappedSuperclassShallBeRejected(): void
+    {
+        $em      = $this->getTestEntityManager();
+        $classes = [GH10449MappedSuperclass::class, GH10449Entity::class, GH10449ToManyAssociationTarget::class];
+
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessageMatches('/illegal to put an inverse side one-to-many or many-to-many association on mapped superclass/');
+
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH10449Test.php
```diff
@@ -0,0 +1,92 @@
...
+     * @ORM\Column(type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue
+     *
+     * @var int
+     */
+    public $id;
+
+    /**
+     * @ORM\ManyToOne(targetEntity="GH10449MappedSuperclass", inversedBy="targets")
```
mpdude, line 58: A mapped superclass cannot be a target entity at all, can it?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10449Test.php
```diff
@@ -0,0 +1,92 @@
...
+     * @ORM\Column(type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue
+     *
+     * @var int
+     */
+    public $id;
+
+    /**
+     * @ORM\OneToMany(targetEntity="GH10449ToManyAssociationTarget", mappedBy="base")
```
mpdude, line 80: Not allowed on a mapped superclass, since the mapped superclass has no table that the foreign key from the other side could refer to.

PR #10453, opened 2023-01-24
Add regression test for a to-many relationship on a base class & mapped superclass in the hierarchy

### LINKED ISSUE #9516: Invalid Bidirectional Association Mapping with Mapped Superclass in Class Hierarchy

mpdude:
This picks the test case from #9517 and rebases it onto 2.14.x.

The problem has been covered in #8415, so this PR closes #9517 and fixes #9516.

Co-authored-by: Robert D'Ercole <bobdercole@gmail.com>

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9516Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH9516Test.php
@@ -1,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\ManyToOne;
+use Doctrine\ORM\Mapping\MappedSuperclass;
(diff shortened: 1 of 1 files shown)


greg0ire (approved):

greg0ire:
Thank you all!

Issue #5744, opened 2016-03-29
AnnotationDriver handles field mappings differently


dylanschoenmakers:
A problem with a mapped superclass led me to this piece of code that conditionally maps fields in the `loadMetadataForClass` method in the AnnotationDriver. See commit 09a48a896eb3adfb1569beafab8566b69167b01e

``` php
    public function loadMetadataForClass($className, ClassMetadata $metadata)
    {
…
        // Evaluate annotations on properties/fields
        /* @var $property \ReflectionProperty */
        foreach ($class->getProperties() as $property) {
            if ($metadata->isMappedSuperclass && ! $property->isPrivate()
                ||
                $metadata->isInheritedField($property->name)
                ||
                $metadata->isInheritedAssociation($property->name)
                ||
                $metadata->isInheritedEmbeddedClass($property->name)) {
                continue;
            }
```

Other drivers like yaml or xml don't have these conditions, or not that I can see. Does anyone have more information about this or has encountered the same problems with overriding field mappings with annotations?

ostrolucky:
See referenced issue DDC-34 = https://github.com/doctrine/doctrine2/issues/4198

mpdude:
👀 needed regarding this in #10455

PR #9528, opened 2022-02-20
Implement forward compatibility with Persistence 3

greg0ire:
This PR aims at getting more compatible with Persistence 3 / see how much work is needed.

PR #10455, opened 2023-01-24
Make Annotations/Attribute mapping drivers report fields for the classes where they are declared

### LINKED ISSUE #10417: Weird conditional skipping of mapping information in attribute/annotations driver

mpdude:
This PR will make the annotations and attribute mapping drivers report mapping configuration for the classes where it is declared, instead of omitting it and reporting it for subclasses only. This is necessary to be able to catch mis-configurations in `ClassMetadataFactory`.

Fixes #10417, closes #10450, closes #10454.

**DoctrineBundle** users: A new config setting to opt-in to the new mode will be implemented in https://github.com/doctrine/DoctrineBundle/pull/1661.

#### ⚠️ Summary for users getting `MappingExceptions` with the new mode

When you set the `$reportFieldsWhereDeclared` constructor parameters to `true` for the AnnotationDriver and/or AttributesDriver and get `MappingExceptions`, you may be doing one of the following:

* Using `private` fields with the same name in different classes of an entity inheritance hierarchy (see #10450)
* Redeclaring/overwriting mapped properties inherited from mapped superclasses and/or other entities (see #10454)

As explained in these two PRs, the ORM cannot (or at least, was not designed to) support such configurations. Unfortunately, due to the old – now deprecated – driver behaviour, the misconfigurations could not be detected, and due to previously missing tests, this in turn was not noticed.

#### Current situation

The annotations mapping driver has the following condition to skip properties that are reported by the PHP reflection API:

https://github.com/doctrine/orm/blob/69c7791ba256d947ddb1aafe5f2439ab31704937/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L345-L357

This code has been there basically unchanged since the initial 2.0 release. The same condition can be found in the attribute driver, probably it has been copied when attributes were added.

I _think_ what the driver tries to do here is to deal with the fact that Reflection will also report `public`/`protected` properties inherited from parent classes. This is supported by the observation (see #5744) that e. g. YAML and XML drivers do not contain this logic.

The conditions are not precise enough for edge cases. They lead to some fields and configuration values not even being reported by the driver. 

Only since the fields would be "discovered" again when reflecting on subclasses, they eventually end up in class metadata structures for the subclasses. In one case of inherited ID generator mappings, the `ClassMetadataFactory` would also rely on this behaviour.

Two potential bugs that can result from this are demonstrated in #10450 and #10454.

#### Suggested solution

In order to find a more reliable way of separating properties that are merely reported again in subclasses from those that are actual re-declarations, use the information already available in `ClassMetadata`. In particular, `declared` tells us in which non-transient class a "field" was first seen.

Make the mapping driver skip only those properties for which we already know that they have been declared in parent classes, and skip them only when the observed declaring class matches the expectation. 

For all other properties, report them to `ClassMetadataFactory` and let that deal with consistency checking/error handling.

#10450 and #10454 are merged into this PR to show that they pass now.

#### Soft deprecation strategy

To avoid throwing new/surprising exceptions (even for misconfigurations) during a minor version upgrade, the new driver mode is opt-in.

Users will have to set the `$reportFieldsWhereDeclared` constructor parameters to `true` for the `AnnotationDriver` and/or `AttributesDriver`. Unless they do so, a deprecation warning will be raised.

In 3.0, the "new" mode will become the default. The constructor parameter can be deprecated (as of ORM 3.1, probably) and is a no-op.

We need to follow up in other places (DoctrineBundle, ... – what else?) to make this driver parameter an easy-to-change configuration setting.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,18 @@
+# Upgrade to 2.16
+
+## Changing the way how reflection-based mapping drivers report fields, deprecated the "old" mode
+
+In ORM 3.0, a change will be made regarding how the `AttributeDriver` reports field mappings.
+This change is necessary to be able to detect (and reject) some invalid mapping configurations.
+
+To avoid surprises during 2.x upgrades, the new mode is opt-in. It can be activated on the 
+`AttributeDriver` and `AnnotationDriver` by setting the `$reportFieldsWhereDeclared`
+constructor parameter to `true`. It will cause `MappingException`s to be thrown when invalid
+configurations are detected.
+
+Not enabling the new mode will cause a deprecation notice to be raised. In ORM 3.0, 
+only the new mode will be available.
+
 # Upgrade to 2.15
 
 ## Deprecated configuring `JoinColumn` on the inverse side of one-to-one associations
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -164,7 +164,7 @@ class Configuration extends \Doctrine\DBAL\Configuration
      *
      * @return AnnotationDriver
      */
-    public function newDefaultAnnotationDriver($paths = [], $useSimpleAnnotationReader = true)
+    public function newDefaultAnnotationDriver($paths = [], $useSimpleAnnotationReader = true, bool $reportFieldsWhereDeclared = false)
     {
         Deprecation::trigger(
             'doctrine/orm',
@@ -203,7 +203,8 @@ class Configuration extends \Doctrine\DBAL\Configuration
 
         return new AnnotationDriver(
             $reader,
-            (array) $paths
+            (array) $paths,
+            $reportFieldsWhereDeclared
         );
     }
 
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -110,7 +110,7 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
         if ($parent) {
             $class->setInheritanceType($parent->inheritanceType);
             $class->setDiscriminatorColumn($parent->discriminatorColumn);
-            $class->setIdGeneratorType($parent->generatorType);
+            $this->inheritIdGeneratorMapping($class, $parent);
             $this->addInheritedFields($class, $parent);
             $this->addInheritedRelations($class, $parent);
             $this->addInheritedEmbeddedClasses($class, $parent);
@@ -138,12 +138,8 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
             throw MappingException::reflectionFailure($class->getName(), $e);
         }
 
-        // If this class has a parent the id generator strategy is inherited.
-        // However this is only true if the hierarchy of parents contains the root entity,
-        // if it consists of mapped superclasses these don't necessarily include the id field.
-        if ($parent && $rootEntityFound) {
-            $this->inheritIdGeneratorMapping($class, $parent);
-        } else {
+        // Complete id generator mapping when the generator was declared/added in this class
+        if ($class->identifier && (! $parent || ! $parent->identifier)) {
             $this->completeIdGeneratorMapping($class);
         }
 
--- a/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
@@ -34,6 +34,7 @@ use function is_numeric;
 class AnnotationDriver extends CompatibilityAnnotationDriver
 {
     use ColocatedMappingDriver;
+    use ReflectionBasedDriver;
 
     /**
      * The annotation reader.
@@ -60,7 +61,7 @@ class AnnotationDriver extends CompatibilityAnnotationDriver
      * @param Reader               $reader The AnnotationReader to use
      * @param string|string[]|null $paths  One or multiple paths where mapping classes can be found.
      */
-    public function __construct($reader, $paths = null)
+    public function __construct($reader, $paths = null, bool $reportFieldsWhereDeclared = false)
     {
         Deprecation::trigger(
             'doctrine/orm',
@@ -70,6 +71,17 @@ class AnnotationDriver extends CompatibilityAnnotationDriver
         $this->reader = $reader;
 
         $this->addPaths((array) $paths);
+
+        if (! $reportFieldsWhereDeclared) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/pull/10455',
+                'In ORM 3.0, the AttributeDriver will report fields for the classes where they are declared. This may uncover invalid mapping configurations. To opt into the new mode also with the AnnotationDriver today, set the "reportFieldsWhereDeclared" constructor parameter to true.',
+                self::class
+            );
+        }
+
+        $this->reportFieldsWhereDeclared = $reportFieldsWhereDeclared;
     }
 
     /**
@@ -348,15 +360,7 @@ class AnnotationDriver extends CompatibilityAnnotationDriver
 
         // Evaluate annotations on properties/fields
         foreach ($class->getProperties() as $property) {
-            if (
-                $metadata->isMappedSuperclass && ! $property->isPrivate()
-                ||
-                $metadata->isInheritedField($property->name)
-                ||
-                $metadata->isInheritedAssociation($property->name)
-                ||
-                $metadata->isInheritedEmbeddedClass($property->name)
-            ) {
+            if ($this->isRepeatedPropertyDeclaration($property, $metadata)) {
                 continue;
             }
 
--- a/lib/Doctrine/ORM/Mapping/Driver/ReflectionBasedDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/ReflectionBasedDriver.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping\Driver;
+
+use Doctrine\ORM\Mapping\ClassMetadata;
+use ReflectionProperty;
+
+/** @internal */
+trait ReflectionBasedDriver
+{
+    /** @var bool */
+    private $reportFieldsWhereDeclared = false;
+
+    /**
+     * Helps to deal with the case that reflection may report properties inherited from parent classes.
+     * When we know about the fields already (inheritance has been anticipated in ClassMetadataFactory),
+     * the driver must skip them.
+     *
+     * The declaring classes may mismatch when there are private properties: The same property name may be
+     * reported multiple times, but since it is private, it is in fact multiple (different) properties in
+     * different classes. In that case, report the property as an individual field. (ClassMetadataFactory will
+     * probably fail in that case, though.)
+     */
+    private function isRepeatedPropertyDeclaration(ReflectionProperty $property, ClassMetadata $metadata): bool
+    {
+        if (! $this->reportFieldsWhereDeclared) {
+            return $metadata->isMappedSuperclass && ! $property->isPrivate()
+                || $metadata->isInheritedField($property->name)
+                || $metadata->isInheritedAssociation($property->name)
+                || $metadata->isInheritedEmbeddedClass($property->name);
+        }
+
+        $declaringClass = $property->getDeclaringClass()->getName();
+
+        if (
+            isset($metadata->fieldMappings[$property->name]['declared'])
+            && $metadata->fieldMappings[$property->name]['declared'] === $declaringClass
+        ) {
+            return true;
+        }
+
+        if (
+            isset($metadata->associationMappings[$property->name]['declared'])
+            && $metadata->associationMappings[$property->name]['declared'] === $declaringClass
+        ) {
+            return true;
+        }
+
+        return isset($metadata->embeddedClasses[$property->name]['declared'])
+            && $metadata->embeddedClasses[$property->name]['declared'] === $declaringClass;
+    }
+}
--- a/tests/Doctrine/Tests/Models/Company/CompanyFlexContract.php
+++ b/tests/Doctrine/Tests/Models/Company/CompanyFlexContract.php
@@ -11,8 +11,6 @@ use Doctrine\ORM\Mapping\Column;
 use Doctrine\ORM\Mapping\Entity;
 use Doctrine\ORM\Mapping\EntityResult;
 use Doctrine\ORM\Mapping\FieldResult;
-use Doctrine\ORM\Mapping\GeneratedValue;
-use Doctrine\ORM\Mapping\Id;
 use Doctrine\ORM\Mapping\JoinColumn;
 use Doctrine\ORM\Mapping\JoinTable;
 use Doctrine\ORM\Mapping\ManyToMany;
@@ -67,14 +65,6 @@ use Doctrine\ORM\Mapping\SqlResultSetMappings;
 #[ORM\Entity]
 class CompanyFlexContract extends CompanyContract
 {
-    /**
-     * @Id
-     * @GeneratedValue
-     * @Column(type="integer")
-     * @var int
-     */
-    public $id;
-
     /**
      * @Column(type="integer")
      * @var int
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC719Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC719Test.php
@@ -31,7 +31,7 @@ class DDC719Test extends OrmFunctionalTestCase
     {
         $q = $this->_em->createQuery('SELECT g, c FROM Doctrine\Tests\ORM\Functional\Ticket\DDC719Group g LEFT JOIN g.children c  WHERE g.parents IS EMPTY');
 
-        $referenceSQL = 'SELECT g0_.name AS name_0, g0_.description AS description_1, g0_.id AS id_2, g1_.name AS name_3, g1_.description AS description_4, g1_.id AS id_5 FROM groups g0_ LEFT JOIN groups_groups g2_ ON g0_.id = g2_.parent_id LEFT JOIN groups g1_ ON g1_.id = g2_.child_id WHERE (SELECT COUNT(*) FROM groups_groups g3_ WHERE g3_.child_id = g0_.id) = 0';
+        $referenceSQL = 'SELECT g0_.id AS id_0, g0_.name AS name_1, g0_.description AS description_2, g1_.id as id_3, g1_.name AS name_4, g1_.description AS description_5 FROM groups g0_ LEFT JOIN groups_groups g2_ ON g0_.id = g2_.parent_id LEFT JOIN groups g1_ ON g1_.id = g2_.child_id WHERE (SELECT COUNT(*) FROM groups_groups g3_ WHERE g3_.child_id = g0_.id) = 0';
 
         self::assertEquals(
             strtolower($referenceSQL),
--- a/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
@@ -1164,7 +1164,7 @@ class SelectSqlGenerationTest extends OrmTestCase
     {
         $this->assertSqlGeneration(
             'SELECT f FROM Doctrine\Tests\Models\DirectoryTree\File f JOIN f.parentDirectory d WHERE f.id = ?1',
-            'SELECT f0_.id AS id_0, f0_.extension AS extension_1, f0_.name AS name_2 FROM "file" f0_ INNER JOIN Directory d1_ ON f0_.parentDirectory_id = d1_.id WHERE f0_.id = ?'
+            'SELECT f0_.id AS id_0, f0_.name AS name_1, f0_.extension AS extension_2 FROM "file" f0_ INNER JOIN Directory d1_ ON f0_.parentDirectory_id = d1_.id WHERE f0_.id = ?'
         );
     }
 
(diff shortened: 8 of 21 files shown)


mpdude:
@greg0ire Do you still remember what you intended when you made this change?

https://github.com/doctrine/orm/commit/2c769acf8cf5838c65b4dd90b1966418679177cb#diff-22f2b83fc9f9b4f9e167ff860549dd20c8d2985a82bfaaa3e04e56f04ad0f4c9R70-R76

That now (with the changes in this PR) causes failures, since the ID column is now a field that is declared twice.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -135,14 +135,7 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
...
         }
 
-        // If this class has a parent the id generator strategy is inherited.
-        // However this is only true if the hierarchy of parents contains the root entity,
-        // if it consists of mapped superclasses these don't necessarily include the id field.
-        if ($parent && $rootEntityFound) {
-            $this->inheritIdGeneratorMapping($class, $parent);
-        } else {
-            $this->completeIdGeneratorMapping($class);
-        }
```
mpdude: The code was based on the assumption that we can only inherit the ID generator mapping when we have already seen an entity class above in the hierarchy, because at that time the ID must have been defined.

As long as we're working downwards the hierarchy and have only seen mapped superclasses so far, it would not inherit the generator mapping, but instead fill in the defaults.

This did not make a difference, since the "id" field would be reported again by the mapping driver as soon as we find an entity class, this time initializing the settings correctly.

With the mapping driver now reporting the ID field only once at the time where it occurs in the class hierarchy, we need to inherit the generator settings right away, as all other inherited configuration.

tests/Doctrine/Tests/Models/Company/CompanyFlexContract.php
```diff
@@ -67,14 +67,6 @@
 #[ORM\Entity]
 class CompanyFlexContract extends CompanyContract
 {
-    /**
-     * @Id
-     * @GeneratedValue
-     * @Column(type="integer")
-     * @var int
-     */
-    public $id;
```
mpdude, line 76: @greg0ire Do you still remember what you intended when you made this change?

https://github.com/doctrine/orm/commit/2c769acf8cf5838c65b4dd90b1966418679177cb#diff-22f2b83fc9f9b4f9e167ff860549dd20c8d2985a82bfaaa3e04e56f04ad0f4c9R70-R76

That now (with the changes in this PR) causes some of the test failures, since the ID column is now a field that is declared twice.

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
```diff
@@ -1164,7 +1164,7 @@ public function testMappedSuperclassAssociationJoin(): void
     {
         $this->assertSqlGeneration(
             'SELECT f FROM Doctrine\Tests\Models\DirectoryTree\File f JOIN f.parentDirectory d WHERE f.id = ?1',
-            'SELECT f0_.id AS id_0, f0_.extension AS extension_1, f0_.name AS name_2 FROM "file" f0_ INNER JOIN Directory d1_ ON f0_.parentDirectory_id = d1_.id WHERE f0_.id = ?'
+            'SELECT f0_.id AS id_0, f0_.name AS name_1, f0_.extension AS extension_2 FROM "file" f0_ INNER JOIN Directory d1_ ON f0_.parentDirectory_id = d1_.id WHERE f0_.id = ?'
```
mpdude, line 1167: In this test, there are some fields in an abstract mapped superclass, and an additional field in the entity subclass.

Previously, the mapping driver would not report the fields for the superclass (they were not contained in the class metadata!). Only on the subclass, PHP reflection would provide the properties again, and this time they end up in the subclass metadata.

With the change suggested in this PR, the fields are now (IMHO correctly) associated with the mapped superclass and then inherited as metadata to the child class, whereare the duplicate reports by PHP reflection (for the subclass) are ignored.

That causes the order of fields in the generated SQL statement to change.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC719Test.php
```diff
@@ -31,7 +31,7 @@ public function testIsEmptySqlGeneration(): void
     {
         $q = $this->_em->createQuery('SELECT g, c FROM Doctrine\Tests\ORM\Functional\Ticket\DDC719Group g LEFT JOIN g.children c  WHERE g.parents IS EMPTY');
 
-        $referenceSQL = 'SELECT g0_.name AS name_0, g0_.description AS description_1, g0_.id AS id_2, g1_.name AS name_3, g1_.description AS description_4, g1_.id AS id_5 FROM groups g0_ LEFT JOIN groups_groups g2_ ON g0_.id = g2_.parent_id LEFT JOIN groups g1_ ON g1_.id = g2_.child_id WHERE (SELECT COUNT(*) FROM groups_groups g3_ WHERE g3_.child_id = g0_.id) = 0';
+        $referenceSQL = 'SELECT g0_.id AS id_0, g0_.name AS name_1, g0_.description AS description_2, g1_.name AS name_3, g1_.description AS description_4, g1_.id AS id_5 FROM groups g0_ LEFT JOIN groups_groups g2_ ON g0_.id = g2_.parent_id LEFT JOIN groups g1_ ON g1_.id = g2_.child_id WHERE (SELECT COUNT(*) FROM groups_groups g3_ WHERE g3_.child_id = g0_.id) = 0';
```
mpdude: With the mapping driver changes, the ID column is seen as coming from the mapped superclass at the root of the hierarchy, not as a field of the child entity.

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -135,14 +135,7 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
...
+        $this->completeIdGeneratorMapping($class);
```
mpdude, line 138: This only fills in missing defaults; probably not necessary once we've got all this in the base class and inherit it. 

I don't know if it is worth coming up with fancy rules here when we may skip this step, since in the end, all metadata is cached, right?

mpdude:
😭 Who can help me understand what the single issue is in Postgres 8.1 with `Doctrine\Tests\ORM\Functional\Ticket\GH5998Test::testIssue`? I don't have a PostgresSQL DB at hand.

tests/Doctrine/Tests/Models/Company/CompanyFlexContract.php
greg0ire: Sorry, I wish I had written a better commit message, I don't remember.

greg0ire:
stab in the dark worked 👍

tests/Doctrine/Tests/Models/Company/CompanyFlexContract.php
mpdude: https://github.com/doctrine/orm/pull/9528#issuecomment-1046209939

This comment shows you had issues with the id missing, so maybe you added it to see if it changes anything? (My guess would be that change was not necessary and you fixed the issue in another way.)

greg0ire: I might have forgotten to remove it, yes… I suppose the best way to find out is to remove it.

lib/Doctrine/ORM/Mapping/Driver/ReflectionBasedDriver.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping\Driver;
+
+use Doctrine\ORM\Mapping\ClassMetadata;
+use ReflectionProperty;
+
+trait ReflectionBasedDriver
```
derrabus, line 11: Please flag this trait as `/** @internal */`. We probably want to inline it into the attribute driver in 3.0.x.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10449Test.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        $em      = $this->getTestEntityManager();
+        $classes = [GH10449MappedSuperclass::class, GH10449Entity::class, GH10449ToManyAssociationTarget::class];
+
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessageMatches('/illegal to put an inverse side one-to-many or many-to-many association on mapped superclass/');
+
+        foreach ($classes as $class) {
+            $cm = $em->getClassMetadata($class);
+        }
```
derrabus: For which of the classes do we expect the exception? This test will pass if `getClassMetadata()` throws for only one of them, no matter which one.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10449Test.php
```diff
@@ -0,0 +1,92 @@
...
+     * Override for BC with PHPUnit <8
+     */
+    public function expectExceptionMessageMatches(string $regularExpression): void
+    {
+        if (method_exists(get_parent_class($this), 'expectExceptionMessageMatches')) {
+            parent::expectExceptionMessageMatches($regularExpression);
+        } else {
+            parent::expectExceptionMessageRegExp($regularExpression);
+        }
+    }
```
derrabus: Would it make sense to move this into a trait? We have a `Doctrine\Tests\PHPUnitCompatibility` namespace for that kind of stuff.

derrabus (changes requested):

lib/Doctrine/ORM/Mapping/Driver/ReflectionBasedDriver.php
mpdude: will do

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10449Test.php
mpdude: good point, will fix

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10449Test.php
mpdude: ~Agree, but I'd like to do this as a clean-up in another PR if you don't mind.~

~TBH I've copied this method from other tests, so we should have a dedicated PR to clean up, not clutter this PR here.~

In fact, that was from `doctrine/annotations`, so no other tests to clean up. Will fix here.

mpdude:
Feedback addressed

greg0ire:
> Potential breakage

Were the bugs you're fixing here introduced in 2.14.x? If not, maybe targeting 2.15.x would be wiser, because otherwise, people with an application that works despite wrong configuration cannot update without a change to their application. If you feel this "bugfix" is going to generate a lot of support, then please target 2.15.x. patch releases should make the software more stable IMO.

mpdude:
You’re right, I will rebase this onto 2.15. It‘s a bug that has been sitting there for a long time, so no need to hurry and surprise users in a bugfix release.

mpdude:
All feedback taken care of, also rebased onto 2.15.x and squashed most of the commits. 

I kept the tests merged from other PRs (where they failed, here they pass) as separate commits.

tests/Doctrine/Tests/PHPUnitCompatibility/ExceptionMatching.php
```diff
@@ -0,0 +1,23 @@
...
+     */
+    public function expectExceptionMessageMatches(string $regularExpression): void
+    {
+        if (method_exists(get_parent_class($this), 'expectExceptionMessageMatches')) {
+            parent::expectExceptionMessageMatches($regularExpression);
+        } else {
+            parent::expectExceptionMessageRegExp($regularExpression);
+        }
+    }
+}
```
greg0ire: I think this should go in there: https://github.com/doctrine/orm/blob/3843d7e0cc71c78c359af4f962b5ec2efc0312cd/tests/Doctrine/Tests/DoctrineTestCase.php#L20-L24

mpdude: Wasn't aware of that – updated.

greg0ire (dismissed):

SenseException (dismissed):

mpdude:
I have an idea how to mitigate the potential breakage for users:

* Add a switch (on the `Configuration` class maybe?) to turn the new behaviour on
* Have a new internal interface for the two drivers that know about the new mode
* In CMF, when the driver implements that interface, read the setting from Config and configure mapping driver accordingly
* Trigger a deprecation notice when the new mode/behaviour is not activated

That way, users can opt-in to the new feature. We probably should follow up in DoctrineBundle etc. (don't know what else?) to make it easy to flip that switch, tough.

In 3.0, 
* throw an exception when the new setting is not turned on
* make the Mapping drivers always use the new mode, revert the interface created for it
* the setting itself is no longer used.

In 3.1, deprecate the switch on the `Configuration` class (users can simply remove the call, it's a no-op).

In 4.0 (did I just say "ORM 4.0"?), remove the method.

WDYT?

mpdude:
The MappingDriverChain prevents this, since we cannot be sure that we will be able to see/configure all drivers. So, I'll have to make this a setting in the Attributes/Annotations driver directly.

mpdude:
@greg0ire and @SenseException I think I've found a way how users can opt-in to the new mode in 2.x. So, things are not going to break in a surprising way during a 2.x upgrade, even when users have a mapping configuration that is invalid and will be rejected in 3.0.

To make this possible, I added a new constructor parameter with a default value as the last argument in `AttributeDriver` and `AnnotationsDriver`.

I hope that's OK and would be happy to get your re-approvals.

UPGRADE.md
```diff
@@ -1,5 +1,18 @@
 # Upgrade to 2.15
 
+## Changing the way how reflection-based mapping drivers report fields, deprecated the "old" mode
+
+In ORM 3.0, a change will be made how the `AttributeDriver` reports field mappings.
```
greg0ire: Isn't this sentence missing a word? (might just be me not knowing all the subtleties of the english language, in which case please let me know).
```suggestion
In ORM 3.0, a change will be made regarding how the `AttributeDriver` reports field mappings.
```

UPGRADE.md
```diff
@@ -1,5 +1,18 @@
...
+This change is necessary to be able to detect (and reject) some invalid mapping configurations.
+
+To avoid surprises during 2.x upgrades, the new mode is opt-in. It can be activated on the 
+`AttributesDriver` and `AnnotationsDriver` by setting the `$reportFieldsWhereDeclared`
```
```suggestion
`AttributeDriver` and `AnnotationDriver` by setting the `$reportFieldsWhereDeclared`
```

lib/Doctrine/ORM/Configuration.php
```diff
@@ -164,7 +164,7 @@ public function setMetadataDriverImpl(MappingDriver $driverImpl)
      *
      * @return AnnotationDriver
      */
-    public function newDefaultAnnotationDriver($paths = [], $useSimpleAnnotationReader = true)
+    public function newDefaultAnnotationDriver($paths = [], $useSimpleAnnotationReader = true, bool $reportFieldsWhereDeclared = false)
```
greg0ire, line 167: Just to clarify, the plan is to deprecate this new argument in ORM 3.1?

lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
```diff
@@ -70,6 +71,17 @@ public function __construct($reader, $paths = null)
         $this->reader = $reader;
 
         $this->addPaths((array) $paths);
+
+        if (! $reportFieldsWhereDeclared) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/pull/10455',
+                'In ORM 3.0, the AttributeDriver will report fields for the classes where they are declared. This may uncover invalid mapping configurations. To opt into the new mode also with the AnnotationsDriver today, set the "reportFieldsWhereDeclared" constructor parameter to true.',
```
greg0ire:
```suggestion
                'In ORM 3.0, the AttributeDriver will report fields for the classes where they are declared. This may uncover invalid mapping configurations. To opt into the new mode also with the AnnotationDriver today, set the "reportFieldsWhereDeclared" constructor parameter to true.',
```

lib/Doctrine/ORM/Configuration.php
mpdude: Yes, or in 3.0? Not so sure what the release policy mandates/allows.

To my understanding, in Symfony the „next .0“ equals „the last .x minus the compat layer“, so there are no additional features or deprecations. So, there is no room for new deprecations which will go into the next .1 release.

But I don’t mind, if you think 3.0 makes more sense that’s fine for me.

greg0ire: 3.1 sounds good to me 👍

mpdude:
@SenseException and @greg0ire – do you think you could give this a review? What kind of additional information and/or starting pointers could help you?

lib/Doctrine/ORM/Mapping/Driver/ReflectionBasedDriver.php
```diff
@@ -0,0 +1,57 @@
...
+            return $metadata->isMappedSuperclass && ! $property->isPrivate()
+                || $metadata->isInheritedField($property->name)
+                || $metadata->isInheritedAssociation($property->name)
+                || $metadata->isInheritedEmbeddedClass($property->name);
+        }
+
+        $declaringClass = $property->getDeclaringClass()->getName();
+
+        if (
+            isset($metadata->fieldMappings[$property->name])
```
SenseException: AFAIR can this `isset` call be removed. The one in the next line should also cover when a key with the name of `$property->name]` doesn't exist. Strange isset-behaviours on multidimensional arrays shouldn't be an issue with the supported PHP versions.

SenseException (dismissed):

lib/Doctrine/ORM/Mapping/Driver/ReflectionBasedDriver.php
mpdude: Feedback addressed – this can be simplified indeed. Thanks!

UPGRADE.md
```diff
@@ -1,5 +1,18 @@
 # Upgrade to 2.15
 
+## Changing the way how reflection-based mapping drivers report fields, deprecated the "old" mode
```
derrabus, line 3: Now that 2.15 has been released, we need to target 2.16.

derrabus (changes requested):
Looks good, but we need to target 2.16 now.

mpdude:
@derrabus Updated to target 2.16.

UPGRADE.md
mpdude: Done

derrabus (approved):

greg0ire (approved):

greg0ire:
Thanks @mpdude !

mpdude:
🤞🏻 let's hope its not going to cause people too much headache

mpdude:
Attempt to add the new config setting in https://github.com/doctrine/DoctrineBundle/pull/1661

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
DemoniacDeath: Please see BC break https://github.com/doctrine/orm/issues/10927

janklan:
It's worth noting that the fix for the incorrect mapping is fairly straightforward, don't panic!

 In a scenario of `Something extends Base` where `Base` defines a property that `Something` re-defines, all you have to do is removing the property from `Something` and [override the association](https://www.doctrine-project.org/projects/doctrine-orm/en/current/tutorials/override-field-association-mappings-in-subclasses.html) instead.

In my case, I had a base class defining uni-directional ManyToOne association and some classes adding an inverse side to it - which is the invalid mapping this PR is about.

It means that this code:

```php

#[ORM\MappedSuperClass]
class BaseClass {
    #[ORM\ManyToOne]
    #[ORM\JoinColumn(name: 'organisation_id', nullable: false, onDelete: 'CASCADE')]
    protected Organisation $organisation;
}

#[ORM\Entity]
class ExtendingClass extends BaseClass {
    #[ORM\ManyToOne(inversedBy: 'labels')]
    #[ORM\JoinColumn(name: 'organisation_id', nullable: false, onDelete: 'CASCADE')]
    protected Organisation $organisation;
}
```

simply had to become

```php

// only ExtendingClass changed

#[ORM\Entity]
#[ORM\AssociationOverrides([
    new ORM\AssociationOverride(name: 'organisation', inversedBy: 'labels'),
])]
class ExtendingClass extends BaseClass {
    // no more $organisation
}
```

I've been putting this off for almost a year, thinking it's a big deal, but it's not.

GabyCrolage:
That is the biggest pain in my **Project** since a lot of time.
I can't get a way of fixing that *messy* deprecation.

PR #10457, opened 2023-01-25
Merge 2.15.x up into 3.0.x

greg0ire:
As you predicted @derrabus, that one was quite painful. I used https://github.com/mhagger/git-imerge, but still.

Note that I'm not even merging the tip of 2.15.x, but the commit right before we bump Psalm to 5.5.0, so there will be a need for another merge up after this one. Just thought it would be best to deal with the Psalm bump separately.

Cc @mpdude

### DIFF
--- a/tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/UpdateCommandTest.php
+++ b/tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/UpdateCommandTest.php
@@ -19,4 +19,31 @@ class UpdateCommandTest extends CommandTestCase
 
         self::$sharedConn->executeStatement($tester->getDisplay());
     }
+
+    /**
+     * @dataProvider getCasesForWarningMessageFromCompleteOption
+     */
+    public function testWarningMessageFromCompleteOption(?string $name, string $expectedMessage): void
+    {
+        $tester = $this->getCommandTester(UpdateCommand::class, $name);
+        $tester->execute(
+            [],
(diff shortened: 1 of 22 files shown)


greg0ire:
Seems it fails with DBAL 4 :disappointed: 
Somebody must have introduced some deprecated calls.

tests/Doctrine/Tests/ORM/Tools/Console/Command/SchemaTool/UpdateCommandTest.php
```diff
@@ -26,4 +26,29 @@ public function testItPrintsTheSql(): void
...
+        yield 'default_name' => [
+            null,
+            '[WARNING] Not passing the "--complete" option to "orm:schema-tool:update" is deprecated',
+        ];
+
+        yield 'custom_name' => [
+            'doctrine:schema:update',
+            '[WARNING] Not passing the "--complete" option to "doctrine:schema:update" is deprecated',
+        ];
+    }
```
greg0ire, line 57: Ah that must be the culprit code :mag:

greg0ire: Introduced in #10443

mpdude:
Looks good from my POV, at least from what I can tell off the top of my head.

Keep up the good work!

Resolving merge conflicts like these is the hard work maintainers do in the background, without hardly anyone noticing.

derrabus (approved):

PR #10460, opened 2023-01-26
Merge 2.15.x into 3.0.x

derrabus:
Fun with Psalm Baselines. 🎉


lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -344,6 +344,19 @@ public static function duplicateDiscriminatorEntry(string $className, array $ent
...
+    // * @param class-string $childEntityClass
+    // */
+    //public static function missingInheritanceTypeDeclaration(string $rootEntityClass, string $childEntityClass): self
+    //{
+    //    return new self(sprintf(
+    //        "Entity class '%s' is a subclass of the root entity class '%s', but no inheritance mapping type was declared.",
+    //        $childEntityClass,
+    //        $rootEntityClass
+    //    ));
+    //}
```
greg0ire, line 358: ?

greg0ire: Let's do the switch to the new code suggested by @mpdude since it's causing issues?

greg0ire (approved):
Or part of the switch, also works 👍

derrabus:
Yeah, I only switch the piece of code that caused trouble, so I can leave all the fun to @mpdude.

greg0ire:
@mpdude I think the right way to do this would be not to commit these commented out lines of code and instead create a draft PR for 3.0 and link to it… you don't even know whether the code you're committing is actually going to work unless you try it on 3.0.

derrabus:
Let's remove the commented-out code from 2.15.x again?

mpdude:
#10465 removes comments

PR #10464, opened 2023-01-26
Fix some tests that were missed in #10431

mpdude:
In #10431, some invalid inheritance declarations in our test base were fixed. However, the change missed to update XML, PHP and static PHP mapping configurations as well.

Apparently, this did not raise any flags because the misconfiguration only caused a deprecation notice in 2.15.x. Running the tests against 3.0 (where the misconfiguration will be an error) unveiled the mistake.

Reviewers: Compare to https://github.com/doctrine/orm/pull/10431/files#diff-3acd7a05f18efb0e3cb41d042f0699ff8f97ce91c3964f9ee588cb876304449b, this PR here does the same change for the other mapping types.

### DIFF
--- a/tests/Doctrine/Tests/Models/DDC5934/DDC5934BaseContract.php
+++ b/tests/Doctrine/Tests/Models/DDC5934/DDC5934BaseContract.php
@@ -42,6 +42,8 @@ class DDC5934BaseContract
 
     public static function loadMetadata(ClassMetadata $metadata): void
     {
+        $metadata->isMappedSuperclass = true;
+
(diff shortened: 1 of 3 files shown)


mpdude:
Should deprecation messages triggered by the test base be an error, or how could I have spotted this on 2.15 already?

derrabus (approved):

derrabus:
> Should deprecation messages triggered by the test base be an error, or how could I have spotted this on 2.15 already?

This is currently not the case. Maybe it should.

PR #10465, opened 2023-01-26
Remove commented-out code sections

mpdude:
This removes comments added in #10431 on the 2.15.x branch, as suggested in https://github.com/doctrine/orm/pull/10460#issuecomment-1404889903.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -156,8 +156,6 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
                     $class->name,
                     end($nonSuperclassParents)
                 );
-                // enable this in 3.0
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -156,8 +156,6 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
                     $class->name,
                     end($nonSuperclassParents)
                 );
-                // enable this in 3.0
-                // throw MappingException::missingInheritanceTypeDeclaration(end($nonSuperclassParents), $class->name);
```
derrabus, line 160: `MappingException::missingInheritanceTypeDeclaration()` is also commented out, isn't it?

mpdude: Right you are, fixed

derrabus (approved):

PR #10466, opened 2023-01-26
Enabled lazy-loading for subclasses.

SanderBoelaars:
This enables lazy-loading of classes that have subclasses by using 
the discriminator column to get the right subtype for the Proxy.

This also changes a few tests since lazy-loading subclasses 
would now be possible and fetched classes are expected to be 
an instance of the Proxy class.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2893,19 +2893,52 @@ class UnitOfWork implements PropertyChangedListener
 
                             break;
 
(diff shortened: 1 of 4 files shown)


derrabus:
The test failures look related.

Also, can you please elaborate which problem your change solves? A _new_ test that covers your change would also help reviewing the PR.

mpdude:
Maybe it has to do with the limitations given in the last paragraph of https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/inheritance-mapping.html#performance-impact?

Issue #10289, opened 2022-12-11
Disallow override of inherited property


michnovka:
I came across https://github.com/doctrine/orm/pull/8348 and I found in current code we still have 

https://github.com/doctrine/orm/blob/b25561ad9623e1436f23241b180b6958d1a7e09c/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L2632-L2635

I think that this exception should be thrown here, don't you think? @beberlei

mpdude:
I think pointing out configurations that are not officially supported and only work by chance is a good thing to do. Hopefully it works towards bringing down the number of bug reports for dubious use cases.

What about creating a PR for this? Are there tests already?

mpdude:
#10470 solved this, please close

PR #10470, opened 2023-01-26
Deprecate overriding fields/associations inherited from other entities

mpdude:
This was brought up in https://github.com/doctrine/orm/pull/8348, but seemingly forgotten to be implemented in later versions.

Exact reference:
https://github.com/doctrine/orm/pull/8348/files#diff-732e324167dd49e48b221477c3e0f6d7934f3eb5ec0970dbf1c06f6c7df15398R2261-R2264

Probably never really worked, because a subclass entity cannot have its fields mapped to other columns than the parent entity (just think about what that would mean for STI/JTI, for example).

Closes https://github.com/doctrine/orm/issues/10289.

**Update** #10519 adds the check for associations.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -2693,10 +2693,17 @@ class ClassMetadataInfo implements ClassMetadata
 
         $mapping = $this->fieldMappings[$fieldName];
 
-        //if (isset($mapping['inherited'])) {
-            // TODO: Enable this exception in 2.8
+        if (isset($mapping['inherited'])) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/pull/10470',
+                'Overrides are only allowed for fields or associations declared in mapped superclasses or traits. This is not the case for %s::%s, which was inherited from %s. This is a misconfiguration and will be an error in Doctrine ORM 3.0.',
+                $this->name,
+                $fieldName,
+                $mapping['inherited']
+            );
             //throw MappingException::illegalOverrideOfInheritedProperty($this->name, $fieldName);
-        //}
+        }
 
         if (isset($mapping['id'])) {
             $overrideMapping['id'] = $mapping['id'];
--- a/tests/Doctrine/Tests/ORM/Mapping/BasicInheritanceMappingTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/BasicInheritanceMappingTest.php
@@ -25,6 +25,7 @@ use Doctrine\ORM\Mapping\SequenceGenerator;
 use Doctrine\ORM\Mapping\Table;
 use Doctrine\ORM\Mapping\UniqueConstraint;
 use Doctrine\Persistence\Mapping\RuntimeReflectionService;
+use Doctrine\Tests\Models\Company\CompanyFixContract;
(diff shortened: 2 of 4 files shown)


mpdude:
I am open for suggestions regarding more precise wording.

We don’t care where the field comes from, be it a trait, a mapped superclass or maybe even a transient class.

Key point is that it may not be present in another entity of the same inheritance tree and be overridden for a subclass only.

tests/Doctrine/Tests/ORM/Mapping/BasicInheritanceMappingTest.php
```diff
@@ -250,6 +251,15 @@ public function invalidHierarchyDeclarationClasses(): Generator
         yield 'complex example (Entity Root -> Mapped Superclass -> transient class -> Entity)'
             => [InvalidComplexRoot::class, InvalidComplexEntity::class];
     }
+
+    /** @group DDC-964 */
+    public function testInvalidOverrideFieldInheritedFromEntity(): void
+    {
+        $this->expectDeprecationWithIdentifier('https://github.com/doctrine/orm/pull/10470');
```
greg0ire, line 260: If this behaves like `expectException`, it should be moved between the 2 lines below, so that triggering the deprecation in `getMetadataFor` doesn't make this test pass.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -2693,10 +2693,16 @@ public function setAttributeOverride($fieldName, array $overrideMapping)
 
         $mapping = $this->fieldMappings[$fieldName];
 
-        //if (isset($mapping['inherited'])) {
-            // TODO: Enable this exception in 2.8
+        if (isset($mapping['inherited'])) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/pull/10470',
+                'Overrides are only allowed for fields or associations declared in mapped superclasses or traits, which is not the case for %s::%s. This is a misconfiguration and will be an error in Doctrine ORM 3.0.',
```
greg0ire: Maybe mention where this was inherited from in the deprecation message. That information is available in `$mapping['inherited']`, correct?

tests/Doctrine/Tests/ORM/Mapping/BasicInheritanceMappingTest.php
mpdude: Fixed that.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
mpdude: Yes, improved that.

greg0ire (approved):

SenseException (approved):

mpdude:
How can we make sure not to forget enabling the exception after this is been merged up into 3.0?

greg0ire:
We have no systematic way of doing that, sadly. We will perform a `grep` before releasing 3.0.x, but I hope we won't stumble on this contributed by then-unreachable contributors.

greg0ire:
I'm working on the merge up. While doing so, I noticed a comment similar to the one you changed, that one is in `setAssociationOverride`, the one you modified in `setAttributeOverride`.

mpdude:
Then we should do the same there, x-ref #8348 and /cc @beberlei

mpdude:
#10520 turns the deprecation notice into an exception in 3.0.x

mpdude:
#10519 adds the check for associations.

PR #10473, opened 2023-01-27
Allow to-many associations on mapped superclasses w/ ResolveTargetEntityListener

mpdude:
Allow to-many associations to be used on mapped superclasses when the owning (inverse) side does not refer back to the mapped superclass, thanks to `ResolveTargetEntityListener`.

#### Current situation

The [documentation states](https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/inheritance-mapping.html):

> No database table will be created for a mapped superclass itself

> [...] persistent relationships defined by a mapped superclass must be unidirectional (with an owning side only). This means that One-To-Many associations are not possible on a mapped superclass at all.

That's a though limitation. 

~Obviously~ ~apparently~ Probably the limitation comes from the fact that in a to-many association the "many" side has to hold a foreign key. Since the mapped superclass does not have a database table (it's not an entity), no such backreference can be established.

Currently, to-many associations trigger an exception as soon as they are seen on a mapped superclass:

https://github.com/doctrine/orm/blob/d6c0031d44f04e04bbc0cd57a3ed7e05c7ea8b40/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L459-L461

#### `ResolveTargetEntityListener`

The `ResolveTargetEntityListener` can be used to substitute interface or class names in mapping configuration at runtime, during the metadata load phase.

When this gimmick is used to replace _all_ references to the mapped superclass with an entity class in time, it should be possible to have to-many associations on the inheriting entity classes.

#### Suggested solution

Instead of rejecting to-many associations on mapped superclasses right away, validate that at the end of the day (after the `loadClassMetadata` event has been processed) no association may target at a non-entity class. That includes mapped superclasses as well as transient classes.

#### Motivating example

Consider a library that comes with a `User` base class. This class is `abstract` and has to be subclassed/filled when the library is used.

By making this a mapped superclass, library users have the freedom to either have a simple user entity class or a user class hierarchy, but we do not impose any requirements on them. (NB we also don't want to have a root entity in the library, because that would have to declare the entire class hierarchy, including library users' classes.)

The actual user class to be used will be configured through the `ResolveTargetEntityListener`.

The library also includes a `SocialMediaAccount` entity. A `User` can have multiple of these accounts, and we want to be able to navigate the accounts from the user side.

To make the example even more fancy, there is a self-referencing association on the `User`: A `User` has been created by another user, and holds a collection of all other `User`s it created.

The test case contained in this PR contains this example and validates that all association mappings look just as if the final user class had been written as an entity directly, without the superclass.

#### Potential review talking points

- Am I missing other reasons why to-many is not feasible?
- We now reject association mappings with `targetEntity`s that are not entities; relevant BC break? (IMHO: no.)

#### Review tip

Review commit by commit, not all files at once. The last commit adds a lot of entity declarations that were previously missed in tests and now raised exceptions; that's a lot of clutter in the PR.

#### Relationship to #10455 

This PR here does not make any changes to the inner workings of the ORM (!).

In fact, the example given can be run with current versions/releases of the ORM with no issues. That is possible because the one config validation rule that would stop it is in fact not effective, since the annotations/attribute mapping drivers do not report the potentially problematic association mapping in the first place.

This would change with #10455. Once that is merged, the to-many association would be dismissed right away and break the example. 

So we need this PR here to keep examples like the one given working with #10455, by finding a more precise rule why/when to reject invalid configurations.

Other: This closes #10398, since the check no longer exists.

### DIFF
--- a/docs/en/reference/inheritance-mapping.rst
+++ b/docs/en/reference/inheritance-mapping.rst
@@ -39,7 +39,7 @@ have to be used.
     to replace references to a mapped superclass with an entity class at runtime.
     As long as there is only one entity subclass inheriting from the mapped
     superclass and all references to the mapped superclass are resolved to that
-    entity class at runtime, the mapped superclass _can_ use One-To-Many associations
+    entity class at runtime, the mapped superclass *can* use One-To-Many associations
     and be named as the ``targetEntity`` on the owning sides.
 
 .. note::
--- a/docs/en/reference/inheritance-mapping.rst
+++ b/docs/en/reference/inheritance-mapping.rst
@@ -15,22 +15,32 @@ is common to multiple entity classes.
 
 Mapped superclasses, just as regular, non-mapped classes, can
 appear in the middle of an otherwise mapped inheritance hierarchy
-(through Single Table Inheritance or Class Table Inheritance).
+(through Single Table Inheritance or Class Table Inheritance). They
+are not query-able, and need not have an ``#[Id]`` property.
 
 No database table will be created for a mapped superclass itself,
-only for entity classes inheriting from it. Also, a mapped superclass
-need not have an ``#[Id]`` property.
+only for entity classes inheriting from it. That  implies that a
+mapped superclass cannot be the ``targetEntity`` in associations.
+
+In other words, a mapped superclass can use unidirectional One-To-One
+and Many-To-One associations where it is the owning side.
+Many-To-Many associations are only possible if the mapped
+superclass is only used in exactly one entity at the moment. For further
+support of inheritance, the single or joined table inheritance features
+have to be used.
 
 .. note::
 
-    A mapped superclass cannot be an entity, it is not query-able and
-    persistent relationships defined by a mapped superclass must be
-    unidirectional (with an owning side only). This means that One-To-Many
-    associations are not possible on a mapped superclass at all.
-    Furthermore Many-To-Many associations are only possible if the
-    mapped superclass is only used in exactly one entity at the moment.
-    For further support of inheritance, the single or
-    joined table inheritance features have to be used.
+    One-To-Many associations are not generally possible on a mapped
+    superclass, since they require the "many" side to hold the foreign
+    key.
+
+    It is, however, possible to use the :doc:```ResolveTargetEntityListener`` <cookbook/resolve-target-entity-listener>`
+    to replace references to a mapped superclass with an entity class at runtime.
+    As long as there is only one entity subclass inheriting from the mapped
+    superclass and all references to the mapped superclass are resolved to that
+    entity class at runtime, the mapped superclass _can_ use One-To-Many associations
+    and be named as the ``targetEntity`` on the owning sides.
 
 .. note::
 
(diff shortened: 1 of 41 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -303,6 +304,16 @@ protected function validateRuntimeMetadata($class, $parent)
         }
     }
 
+    private function validateAssociationTargets(ClassMetadata $class): void
+    {
+        foreach ($class->getAssociationMappings() as $mapping) {
+            $targetEntity = $mapping['targetEntity'];
+            if ($this->driver->isTransient($targetEntity) || $this->peekIfIsMappedSuperclass($targetEntity)) {
```
mpdude, line 311: Is this correct, or am I missing a case where different drivers might be used for entities from different modules or similar? 

There is just one ClassMetadataFactory for all entities (within the boundaries of a particular entity manager), and it has a single driver, right?

mpdude:
@greg0ire Any ideas where the Psalm errors might come from?

greg0ire:
DBAL 3.6.0, published 20 hours ago?

greg0ire (dismissed):
This example reminds me of FOSUserBundle, so I think it's fair to say you're unlocking a valid scenario there.

docs/en/reference/inheritance-mapping.rst
```diff
@@ -15,22 +15,32 @@ is common to multiple entity classes.
...
-    joined table inheritance features have to be used.
+    One-To-Many associations are not generally possible on a mapped
+    superclass, since they require the "many" side to hold the foreign
+    key.
+
+    It is, however, possible to use the :doc:```ResolveTargetEntityListener`` <cookbook/resolve-target-entity-listener>`
+    to replace references to a mapped superclass with an entity class at runtime.
+    As long as there is only one entity subclass inheriting from the mapped
+    superclass and all references to the mapped superclass are resolved to that
+    entity class at runtime, the mapped superclass _can_ use One-To-Many associations
```
SenseException: Isn't `_can_` markdown?

SenseException (dismissed):

mpdude:
Super happy to have two approvals already 🤩 and a bit afraid to lose them again when I push a markdown/rst fix 🤪

mpdude:
@greg0ire / @SenseException Fixed the .rst glitch, please re-approve.

SenseException (approved):

greg0ire (approved):

mpdude:
🎉 🍾  Awesome, thanks!

mpdude:
#10455 is the logical next step ;-)

BreyndotEchse:
Causes an exception when using overrides:

`MappingException.php line 138: Invalid field override named 'fieldToOverride' for class 'Foo'.`

`ClassMetadataFactory::peekIfIsMappedSuperclass` calls `MappingDriver::loadMetadataForClass` with a new `ClassMetadataInfo` instance that does not contain information about its parent(s) - unlike an instance created by `ClassMetadataFactory::doLoadMetadata`. If the MappingDriver tries to load metadata for an entity with "AssociationOverrides" it cannot override an inherited association it is not aware of. `ClassMetadataInfo::setAssociationOverride` then throws the exception mentioned above.

mpdude:
@BreyndotEchse could you please create a new issue for that? It might easily get lost here on the merged PR. Also, do you think you could provide a functional test for it?

mpdude:
@BreyndotEchse  and @alexander-schranz Could you please try and report back in #10554 if that solves the problems this caused for you?

PR #10474, opened 2023-01-27
Exclude traits from runtime association mappings check

mpdude:
When checking at runtime whether an association mapping is valid/correct, do not check for traits.

I don't see how it would make sense to have an association targeting at a trait. Traits never exist on their own, they are only used in classes. So you cannot have an association aim at a trait in any meaningful way?

(I may be wrong and would be happy to learn, though!)

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1297,12 +1297,10 @@ class ClassMetadataInfo implements ClassMetadata
     public function validateAssociations()
     {
         foreach ($this->associationMappings as $mapping) {
-            if (
-                ! class_exists($mapping['targetEntity'])
-                && ! interface_exists($mapping['targetEntity'])
-                && ! trait_exists($mapping['targetEntity'])
-            ) {
-                throw MappingException::invalidTargetEntityClass($mapping['targetEntity'], $this->name, $mapping['fieldName']);
+            /** @var string */
+            $targetEntity = $mapping['targetEntity'];
+            if (! class_exists($targetEntity)) {
+                throw MappingException::invalidTargetEntityClass($targetEntity, $this->name, $mapping['fieldName']);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


mpdude:
How do I tell Psalm that we don't really trust `targetEntity` to be a `class-string` here?

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1297,11 +1297,7 @@ public function validateIdentifier()
     public function validateAssociations()
     {
         foreach ($this->associationMappings as $mapping) {
-            if (
-                ! class_exists($mapping['targetEntity'])
-                && ! interface_exists($mapping['targetEntity'])
-                && ! trait_exists($mapping['targetEntity'])
-            ) {
+            if (! class_exists($mapping['targetEntity']) && ! interface_exists($mapping['targetEntity'])) {
```
mpdude: Thinking a bit more about this, I don't think we want to allow interfaces either?

This method is called at a very late stage of loading metadata, and nothing is going to fixup/fill/amend metadata after that; and the result is meant to be read-only.

So, how can we deal with an interface name as the association target?

If `ResolveTargetEntityListener` comes to mind, that should have taken place before we run this check.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1297,12 +1297,10 @@ public function validateIdentifier()
     public function validateAssociations()
     {
         foreach ($this->associationMappings as $mapping) {
-            if (
-                ! class_exists($mapping['targetEntity'])
-                && ! interface_exists($mapping['targetEntity'])
-                && ! trait_exists($mapping['targetEntity'])
-            ) {
-                throw MappingException::invalidTargetEntityClass($mapping['targetEntity'], $this->name, $mapping['fieldName']);
+            /** @var string */
```
mpdude, line 1300: Without this, I'd have to suppress `TypeDoesNotContainType` for the entire `ClassMetadataInfo`. 

Given this class benefits a lot form static analysis, I'd really like to avoid having to do that. I'm open for alternative suggestions how that could be achieved.

Issue #8177, opened 2020-06-09
Support unique yet stable identifier generation in "Specification"-style designs


mpdude:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

I've seen quite a few applications of the "Specification" pattern. In essence, a Specification is a
"business" or "domain logic" class that encapsulates a particular business condition, filter expression or the like. There are various interpretations and approaches of this pattern, but [here](https://martinfowler.com/apsupp/spec.pdf), [this](https://beberlei.de/2013/03/04/doctrine_repositories.html) or [that](https://github.com/Happyr/Doctrine-Specification) might be good starting points if you don't know about it.

Among other things, a Specification needs to be able to "apply" itself to a Doctrine `QueryBuilder`. While doing so, it might need to add conditions and parameters to a query or perform joins. In both of these cases, the Specification needs to come up with unique names for parameters or join aliases. The challenge with these unique names is that the Specification itself cannot know which other Specifications are applied to the same `QueryBuilder`, and even several instances of the same Specification class may be applied at once.

In some cases I've seen, Specifications even need to make these aliases publicly available and have to do so _even before_ they got hold of the `QueryBuilder` that they'll be applied to. (For example, a Specification might express "eager load a particular entity", and another Specification might build on top of it and need the alias.) This defeats an approach where you could use a central `\SplObjectStorage` to keep track of alias names per `QueryBuilder` instance, because you don't even have that yet.

One likely solution is to start using `uniqid()` for the parameter and alias names. But this has a serious drawback: It defeats the Query Cache, since repeated requests for even the very same page will result in different DQL every time. Depending on cache implementation, your cache will just become evicted again and again (APCu behavior, to my knowledge) or gradually fill up disk space and ultimately use up all available OPcache memory for cached, but ever ever reused PHP files (e. g. default Symfony Cache pools).

Another solution would be to use something like `static` variables to generate identifiers based on a counter that re-starts from 0 upon every request. That way, chances are that "the very same request" will generate identical queries again; but of course, conditional code paths that execute additional Specifications - even in completely unrelated queries - break things again.

So, what I would like to discuss here is if and how we could support this from the ORM side.

[...]

PR #10475, opened 2023-01-27
Allow multiple criterias on the same property.

### LINKED ISSUE #8702: "Too many parameters: ..." when adding multiple criteria having clauses on the same property

n-e-m-a-nj-a:
When adding multiple criteria referencing the same field, the query builder returnes an error.
The fix is passing existing parameters from previous parameters to every new insteance of QueryExpressionVisitor.
Fixes [#8702](https://github.com/doctrine/orm/issues/8702)

Ensure unique test entity name, and sync criteria field name with entity property.

### DIFF
--- a/lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
+++ b/lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
@@ -11,6 +11,7 @@ use Doctrine\Common\Collections\Expr\ExpressionVisitor;
 use Doctrine\Common\Collections\Expr\Value;
 use RuntimeException;
 
+use function array_merge;
 use function count;
 use function str_replace;
 use function str_starts_with;
@@ -37,11 +38,18 @@ class QueryExpressionVisitor extends ExpressionVisitor
     /** @var list<mixed> */
     private $parameters = [];
 
-    /** @param mixed[] $queryAliases */
-    public function __construct($queryAliases)
+    /** @var list<mixed> */
+    private array $existingParameters;
+
+    /**
+     * @param mixed[] $queryAliases
+     * @param mixed[] $existingParameters
+     */
+    public function __construct($queryAliases, array $existingParameters = [])
     {
-        $this->queryAliases = $queryAliases;
-        $this->expr         = new Expr();
+        $this->queryAliases       = $queryAliases;
+        $this->expr               = new Expr();
+        $this->existingParameters = $existingParameters;
     }
 
     /**
@@ -120,9 +128,11 @@ class QueryExpressionVisitor extends ExpressionVisitor
 
         $parameterName = str_replace('.', '_', $comparison->getField());
 
-        foreach ($this->parameters as $parameter) {
+        $parameters = array_merge($this->parameters, $this->existingParameters);
+
+        foreach ($parameters as $parameter) {
             if ($parameter->getName() === $parameterName) {
-                $parameterName .= '_' . count($this->parameters);
+                $parameterName .= '_' . count($parameters);
                 break;
             }
         }
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -1360,7 +1360,7 @@ class QueryBuilder
             throw new Query\QueryException('No aliases are set before invoking addCriteria().');
         }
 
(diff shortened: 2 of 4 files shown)


mpdude:
Congrats on your first time contribution! And also thank you for providing the test case, that is really helpful.

As I get it, the `QueryBuilder` creates a new `QueryExpressionVisitor` every time a `Criteria` is added. This `QueryExpressionVisitor` can walk the `Criteria` and will then provide the `QueryBuilder` with the necessary WHERE condition and also the parameters that need to be bound.

Internally, the `QueryExpressionVisitor` creates `Parameter` instances every time it needs to compare against a value. These `Parameter`s are named according to the field that the comparison is against. `QueryExpressionVisitor` is able to detect collisions within its own parameter names and resolves those by appending an incrementing number (`_1`, `_2`).

In the end, the `QueryBuilder` takes all parameters from the `QueryExpressionVisitor` and adds them for the query that is being built.

My first thought was that maybe the `QueryBuilder` should not trust the `QueryExpressionVisitor` with regard to parameter names and do its own naming/sanitization when copying all the parameters. But then I realized that since the `QueryBuilder` incorporates the given WHERE clause, we're dealing with parameter names placed within that DQL – so there is no way we could change the names, we will have to accept what the `QueryExpressionVisitor` came up with.

Your solution is to give the `QueryExpressionVisitor` _all_ parameters that are defined on the `QueryBuilder` at the time the visitor is created. That enables the visitor to come up with new names that not only avoid collisions with other criteria based on the same field, but also avoid parameter names set by the user (with `QueryBuilder::setParameter()`).

This, however, means the `QueryExpressionVisitor` will have to carry around all of the `QueryBuilders` current parameters, just to be able to do this name evasion. 

Also – but that's unrelated to your suggested change – the user of the `QueryBuilder` may be surprised that after calling `addCriteria()`, new parameters with semi-unpredictable names may show up on the query builder. That might be another source of parameter name collisions, depending on how queries are being built.

So, let me make an alternative proposal:

In DBAL, the SQL query builder has a `createNamedParameter()` method that can be used to create new parameter names, useful in case you don't care about the exact name ([docs](https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/query-builder.html#binding-parameters-to-placeholders)). 

What if we had a similar feature on the DQL query builder?

We could even say that "reserved" parameter names (like `:dcValue1`, `:dcValue...`?) must not be added by the user with `setParameter()` or similar, but may only be created and bound by that new method, but that's a detail.

That way, not the `QueryExpressionVisitor` would be responsible for finding unique names, but it could use the `QueryBuilder` for that. At the same time, the `QueryBuilder` would not have to expose its own parameter list that much.

What do you think?

mpdude:
By the way, I'm interested in this because it seems to be a sub-problem of an issue I raised a while ago (#8177). My issue is a bit broader since I also need unique (collision-free) alias names for JOINs, and I (think I) need to be able to have those identifiers before the QueryBuilder is even created.

mpdude:
#9883 is about the new methods as well

PR #10476, opened 2023-01-28
Hunt down invalid doc blocks

greg0ire:
It is OK to ignore some of the errors we get, but not this one.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1353,7 +1353,7 @@ class ClassMetadataInfo implements ClassMetadata
     /**
      * @param string $fieldName
      * @param array  $cache
-     * @psalm-param array{usage?: int, region?: string|null} $cache
+     * @psalm-param array{usage?: int|null, region?: string|null} $cache
      *
      * @return int[]|string[]
      * @psalm-return array{usage: int, region: string|null}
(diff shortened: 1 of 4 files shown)


greg0ire:
:thinking: there are some more occurrences of `InvalidDocBlock` in the baseline. I'm going to try to remove them.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1793,25 +1793,7 @@ protected function validateAndCompleteFieldMapping(array $mapping): array
...
-     *                   isCascadePersist: bool,
-     *                   isCascadeRefresh: bool,
-     *                   isCascadeMerge: bool,
-     *                   isCascadeDetach: bool,
-     *                   type: int,
-     *                   originalField: string,
-     *                   originalClass: class-string,
-     *                   ?orphanRemoval: bool
-     *               }
+     * @psalm-return AssociationMapping
```
greg0ire, line 1796: Slightly out of scope, but doing this change allows me not to pollute the phpstan baseline too much.

greg0ire:
ping @derrabus :slightly_smiling_face:

lib/Doctrine/ORM/Mapping/Driver/YamlDriver.php
```diff
@@ -922,6 +922,8 @@ private function cacheToArray(array $cacheMapping): array
             $usage = (int) constant('Doctrine\ORM\Mapping\ClassMetadata::CACHE_USAGE_' . $usage);
         }
 
+        assert($usage !== '0' && $usage !== '');
```
derrabus: Can we be sure? I think we should either map all falsy values to `null` or make sure falsy values other than null also trigger an `InvalidArgumentException`.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
derrabus: Looks reasonable to me.

lib/Doctrine/ORM/Mapping/Driver/YamlDriver.php
greg0ire: I think you're right, but let's maybe not do this in a patch release?

derrabus: Is a suddenly failing `assert()` better in a patch release? 😬

greg0ire: Provided assertions are not enabled in production, I'd say yes.

SenseException: Not a doc block (as the title suggests) though, but in comparison it screams at dev runtime when something's wrong. :-)

greg0ire: OK, I'll remove that line and ignore whatever SA errors come up then.

derrabus (approved):

PR #10480, opened 2023-01-29
Run tests with ext-pgsql

derrabus:
A new CI job that runs our test suite with the new pgsql driver introduced in doctrine/dbal#5880.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -116,13 +116,18 @@ jobs:
           - "3@dev"
         postgres-version:
           - "15"
+        extension:
+          - pdo_pgsql
+          - pgsql
         include:
           - php-version: "8.0"
             dbal-version: "2.13"
             postgres-version: "14"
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -116,13 +116,21 @@ jobs:
...
           - php-version: "8.0"
             dbal-version: "2.13"
             postgres-version: "14"
+            extension: pdo_pgsql
           - php-version: "8.2"
             dbal-version: "default"
             postgres-version: "9.6"
+            extension: pdo_pgsql
+          - php-version: "8.2"
+            dbal-version: "3@dev"
```
greg0ire: Let's wait for a release of DBAL and use a stable version? That way you're unlikely to break the ORM build for nothing, and there is no need to make another PR to alter this.

derrabus: All right.

greg0ire (approved):

Issue #10483, opened 2023-01-30
Unable to find "App\Entity\Role" entity identifier associated with the UnitOfWork


specdrum-agc:
### Bug Report

Unable to find "App\Entity\Role" entity identifier associated with the UnitOfWork

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.14.1

#### Summary

When I try to flush changes in one Entity, that related to deleted one, exception is thrown.

#### Current behavior

Somewhat works wrong when deleting ManyToMany relation and saving related entity after flushing these changes.

I investigated the problem and found out that removing entity inside of method `computeAssociationChanges` does not detect such changes and collection stays dirty.

#### How to reproduce

https://github.com/specdrum-agc/doctrine-bug (See Readme)

The main part of test case
```
        $role = $this->doctrine->getRepository(Role::class)->find(1);
        $user = $role->getUsers()->current();
        $user->setName('test');

        // Need for collection initialization
        $roles = $user->getRoles()->toArray();
        $entityManager->remove($role);
        $entityManager->flush();

        // Second flush fails
        $entityManager->flush();
```
The test case might be strange, but I have similar behavior in real application. I have second flush in different part of application than first.

#### Expected behavior

Second `flush` call do not throw exception.

mpdude:
So you change the user that it has to be flushed, but the `roles` association still contains the role that has been deleted?

specdrum-agc:
No, it doesn't. Only in collection's snapshot.
This is the point that state of collection comes into invalid state. When second flush in progress, UnitOfWork sees that and try to persist that change, but it was persisted in first flush.

mpdude:
Did this work in previous ORM versions?

specdrum-agc:
I had 2.11.3 version installed, there was the same issue.

mpdude:
I am not sure what the expected behaviour would be, but #10485 provides a failing test for the time being.

specdrum-agc:
@mpdude  Thank you very much for adding test case for this issue! It reproduces the mentioned behavior. But looks like you are left excess comma on line 20. Test log says about it.

mpdude:
#10486 might be a fix

mpdude:
Duplicate: #10060

PR #10485, opened 2023-01-31
Removing an entity that is contained in a m2m association fails when `flush()` is called twice

mpdude:
Here is a failing test for #10483.

When an entity is removed that is part of a many-to-many collection, that collection will be updated during the `flush()` operation to no longer contain the entity.

However, afterwards the collection will be in a dirty state and the internal snapshot still contains the removed entity.

That causes the collection to be processed again when `flush()` is called another time. This time, the ORM assumes the entity has been removed from the collection. It will try to build a `DELETE` operation but fails since the entity (from the snapshot) is no longer in the identity map.

The error message is `Doctrine\ORM\EntityNotFoundException: Unable to find "..." entity identifier associated with the UnitOfWork`.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10483Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10483Test.php
@@ -1,0 +1,114 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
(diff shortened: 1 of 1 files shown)


mpdude:
Can anyone advise what would be the expected state of the collection after the UoW removed the entity from it?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10483Test.php
```diff
@@ -0,0 +1,114 @@
...
+
+class GH10483Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->createSchemaForModels(
+            GH10483Role::class,
+            GH10483User::class,
```
specdrum-agc: Syntax error that fails the test.

mpdude: Indeed, for PHP < 7.3. Fixed.

PR #10486, opened 2023-01-31
Fix to-many collections left in dirty state after entities are removed by the UoW

### LINKED ISSUE #10060: EntityNotFoundException after second flush when deleting item of ManyToMany relationship
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.13.2

#### Summary

Given two entities `$a` and `$b` of type `A` and `B`, referenced via `ManyToMany` relationship. If one of them is delete via `$entityManager->remove($a)`, a second call to `$entityManager->flush()` will result in an `EntityNotFoundException`.

#### Current behavior

During the first `flush()` `$a` is deleted from the database. Plus, it is removed from `$b->items`. This makes `$b->items` dirty. Next call to `flush()` will try to sync empty `$b->items` with the database by trying to delete `$a`, which is already removed, leading to the EntityNotFoundException.

#### How to reproduce

```
$a = new A();
$b = new B();
$b->items->add($a);
$a->items->add($b);

$this->em->persist($b);
$this->em->persist($a);
$this->em->flush();

$this->em->remove($a);
$this->em->flush();
$this->em->flush(); < throws EntityNotFoundException 
``` 

#### Expected behavior

I think `flush()` should sync the memory state to the database leaving all entites in a clean state. Multiple calls to `flush()` should always be possible without sideeffects.

### LINKED ISSUE #10483: Unable to find "App\Entity\Role" entity identifier associated with the UnitOfWork

mpdude:
#### Current situation

When an entity is removed from the EM, the UoW updates collections that contain that entity, removing it from the collections as well.

The collections are left in a dirty state afterwards with outdated snapshots, which leads to errors e. g. when the Entity Manager is flushed again (see #10485).

#### Suggested fix

This PR fixes the problem by moving a check further down, so that also collections updated by the UoW itself (by removing the entity in question) are captured as "processed" by the current commit phase.

#### Tests added

It seems tests for this feature were completely missing, so I added them.

Also added a second test case for the feature added in #256 that the event for many-to-many collections is not only generated for added, but also for removed entities.

Closes #10485, fixes #10483, fixes #10060, closes #10061.

### DIFF
--- a/tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
+++ b/tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
@@ -31,6 +31,7 @@ use Doctrine\Tests\Mocks\ConnectionMock;
 use Doctrine\Tests\Mocks\EntityManagerMock;
 use Doctrine\Tests\Mocks\EntityPersisterMock;
 use Doctrine\Tests\Mocks\UnitOfWorkMock;
+use Doctrine\Tests\Models\CMS\CmsGroup;
 use Doctrine\Tests\Models\CMS\CmsPhonenumber;
 use Doctrine\Tests\Models\CMS\CmsUser;
 use Doctrine\Tests\Models\Forum\ForumAvatar;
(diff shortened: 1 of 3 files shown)


specdrum-agc (dismissed):
Looks OK.

mpdude:
@SenseException maybe you have an opinion on this?

mpdude:
@lcobucci Do you think you could review this?

mpdude:
@lcobucci maybe you could review this? It's not that much of a change...

FelixBehr:
Hi Guys,

this would be the solution we currently have in a project. Is there any chance one could review this asap?
Thanks for helping out.

tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
```diff
@@ -866,6 +867,59 @@ public function testItThrowsWhenLookingUpIdentifierForUnknownEntity(): void
...
+
+        $this->_unitOfWork->commit();
+
+        self::assertFalse($user->groups->isDirty());
+
+        $this->_unitOfWork->remove($group);
+        $this->_unitOfWork->commit();
+
+        // Test that the removed entity has been removed from the many to many collection
+        self::assertEmpty($user->groups);
```
greg0ire: Your comments are great, why not use them as a second argument for the `assert*` calls?

```suggestion
        // Test that the removed entity has been removed from the many to many collection
        self::assertEmpty(
                $user->groups,
                'the removed entity should have been removed from the many to many collection'
        );
```

I myself don't do it that often, I think they are best if formulated as 'XXX should have happened', but you might want to check the final result.

mpdude:
@SenseException  maybe you could review this?

SenseException (dismissed):
Tests fail with the old code, so your fix seems to be what's needed. Once the build is green (CS fails) this is :+1:. Please check the Coding Style.

greg0ire (approved):

greg0ire:
Thanks @mpdude !

mpdude:
Thank you for fixing the remaining issues

mpdude:
@greg0ire could you close the fixed issues I referenced in the initial post? Thanks.

greg0ire:
:thinking:  that's been done automatically, hasn't it? Everything in the first message is either merged or closed.

mpdude:
Sorry, you’re right. I had the impression that did not work in other cases when a PR opened by me claims to close issues raised by others.

greg0ire:
Yes, it has happened to me as well.

PR #10489, opened 2023-02-03
Fix forgotten exception throws

janedbal:
I believe those exceptions are meant to be thrown, right? Otherwise, this code is useles.

Detected by https://github.com/shipmonk-rnd/phpstan-rules#forbidunusedexception

### DIFF
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -227,7 +227,7 @@ class Configuration extends \Doctrine\DBAL\Configuration
                 $alias
             );
         } else {
-            NotSupported::createForPersistence3(sprintf(
+            throw NotSupported::createForPersistence3(sprintf(
                 'Using short namespace alias "%s" by calling %s',
                 $alias,
                 __METHOD__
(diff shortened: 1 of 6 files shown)


greg0ire (dismissed):
Oh wow good catch! I'm pretty careless sometimes :facepalm:

greg0ire:
Argh it breaks the test suite… somehow we are relying on this feature, which does not work with Persistence 3, and somehow that did not break the tests before :sweat: 

I should have changed them in https://github.com/doctrine/orm/pull/9592

greg0ire:
I'm looking into it.

greg0ire (approved):

janedbal:
@greg0ire Do you want me to send PR with [this rule](https://github.com/shipmonk-rnd/phpstan-rules#forbidunusedexception) enabled to make sure this error wont happen again? It is pretty common error actually :)

derrabus (approved):

greg0ire:
@janedbal I think adding a dependency for one rule isn't a great idea. Maybe there are more rules worth implementing?

janedbal:
There might be some, but the package is aiming to more strict codebases than this one is, so many of them could not be enabled. Also, it is tested [only on PHP 7.4+](https://github.com/shipmonk-rnd/phpstan-rules/blob/master/composer.json#L13). So probably not worth the effort.

PhilETaylor:
this created this b/c break in a minor release :( https://github.com/doctrine/orm/issues/10621

PR #10497, opened 2023-02-05
Migrate more classes to PHP 8 syntax

greg0ire:


lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -83,18 +78,17 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
...
-     * @param ClassMetadata          $class The class descriptor of the entity type of this collection.
+     * @param EntityManagerInterface $em        The EntityManager the collection will be associated with.
+     * @param ClassMetadata          $typeClass The class descriptor of the entity type of this collection.
      * @psalm-param Collection<TKey, T>&Selectable<TKey, T> $collection The collection elements.
      */
     public function __construct(
         EntityManagerInterface $em,
-        ClassMetadata $class,
+        private readonly ClassMetadata|null $typeClass,
         Collection $collection,
```
SenseException, line 88: `$em` and `$collection` could also use constructor promotion.

lib/Doctrine/ORM/Query/Printer.php
```diff
@@ -13,28 +13,16 @@
...
+    protected int $_indent = 0;
 
     /**
      * Constructs a new parse tree printer.
      *
-     * @param bool $silent Parse tree will not be printed if true.
+     * @param bool $_silent Parse tree will not be printed if true.
      */
-    public function __construct($silent = false)
+    public function __construct(protected $_silent = false)
```
SenseException:
```suggestion
    public function __construct(protected bool $_silent = false)
```

The PR is already labeled as a BC break. Isn't this also a good time to remove the underscore prefix?

SenseException (dismissed):

lib/Doctrine/ORM/PersistentCollection.php
greg0ire: I don't think so: `$em` is not supposed to be passed as `null`, but can become `null` on unserialization. `$collection` is a property inherited from the parent.

lib/Doctrine/ORM/Query/Printer.php
greg0ire: I will do a separate, global PR if you don't mind.

greg0ire: See https://github.com/doctrine/orm/pull/10500

derrabus: I think we've discussed similar changes already: I wouldn't switch parameter names to the discouraged underscore notation just for the sake of using CPP.

derrabus (changes requested):

lib/Doctrine/ORM/Query/Printer.php
greg0ire: Then approve #10500. Or approve this one and let me rebase #10500, the end result will be the same: CPP and no underscore.

greg0ire: Actually I just pushed a version with no underscores so that we can deal with both PRs separately. More simple for everyone.

derrabus (approved):

PR #10506, opened 2023-02-07
Migrate Parser to PHP 8

greg0ire:
In some cases, I had to widen the documented type, running the test suite showed me it was wrong.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -45,13 +45,13 @@ use function substr;
  *                 token: DqlToken,
  *             }
  */
-class Parser
+final class Parser
 {
     /**
      * @readonly Maps BUILT-IN string function names to AST class names.
      * @psalm-var array<string, class-string<Functions\FunctionNode>>
      */
-    private static $stringFunctions = [
+    private static array $stringFunctions = [
         'concat'    => Functions\ConcatFunction::class,
         'substring' => Functions\SubstringFunction::class,
         'trim'      => Functions\TrimFunction::class,
@@ -64,7 +64,7 @@ class Parser
      * @readonly Maps BUILT-IN numeric function names to AST class names.
      * @psalm-var array<string, class-string<Functions\FunctionNode>>
      */
-    private static $numericFunctions = [
+    private static array $numericFunctions = [
         'length'    => Functions\LengthFunction::class,
         'locate'    => Functions\LocateFunction::class,
         'abs'       => Functions\AbsFunction::class,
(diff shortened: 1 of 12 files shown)


lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -2541,18 +2354,8 @@ public function ConditionalPrimary()
...
-     *         AST\EmptyCollectionComparisonExpression|
-     *         AST\ExistsExpression|
-     *         AST\InExpression|
-     *         AST\InstanceOfExpression|
-     *         AST\LikeExpression|
-     *         AST\NullComparisonExpression
-     */
-    public function SimpleConditionalExpression()
+     */
+    public function SimpleConditionalExpression(): AST\ExistsExpression|AST\BetweenExpression|AST\LikeExpression|AST\InListExpression|AST\InSubselectExpression|AST\InstanceOfExpression|AST\CollectionMemberExpression|AST\NullComparisonExpression|AST\EmptyCollectionComparisonExpression|AST\ComparisonExpression
```
greg0ire, line 2358: Should I simplify this one to `AST\Node`?

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -370,19 +334,11 @@ public function parse()
...
-                    break;
-
-                case $AST instanceof AST\SelectStatement:
-                default:
-                    $treeWalkerChain->walkSelectStatement($AST);
-            }
+            match (true) {
+                $AST instanceof AST\UpdateStatement => $treeWalkerChain->walkUpdateStatement($AST),
+                $AST instanceof AST\DeleteStatement => $treeWalkerChain->walkDeleteStatement($AST),
+                default => $treeWalkerChain->walkSelectStatement($AST),
```
derrabus: Let's be explicit about the remaining case that we expect here. Static analysis should recognize that the `match` statement covers all cases without a `default` and will notify us if that will ever change in the future. 💪🏻 

```suggestion
                $AST instanceof AST\SelectStatement => $treeWalkerChain->walkSelectStatement($AST),
```

derrabus (changes requested):
Shall we `final`ize the `Parser` class? With all those native union types, it'll otherwise hard to change that method in the future.

lib/Doctrine/ORM/Query/Parser.php
derrabus: Depends on how we answer my question regarding `final`. 🙂

greg0ire:
Let's try `final`. If it turns out this was a wrong move, we can always remove it afterwards, after adjusting the return types.

lib/Doctrine/ORM/Query/Parser.php
greg0ire: Looks nicer indeed 👍

derrabus (approved):

PR #10508, opened 2023-02-08
Fix id hash of entity with enum as identifier

### LINKED ISSUE #10334: ObjectHydrator::getEntityFromIdentityMap() does not work with backed enum
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.14.0

#### Summary

I have an entity with a backed enum as primary key. When I fetch a fresh entity from the database, everything works fine. But if I try to fetch an entity that is already known in the identity map, it fails.

#### Current behavior

It cannot load an entity with backed enum as primary key from the ObjectHydrator identity map.

> Error "Object of class MyEnum could not be converted to string"

From ObjectHydrator on line 292

#### How to reproduce

We need 2 entities, with ManyToOne relation, so we can use a join in our request.
```php
enum MyEnum: string
{
    case Yellow = 'yellow';
    case Blue = 'blue';
}
```

```php
#[ORM\Entity, ORM\Table(name: 'foo')]
class Foo
{
    #[ORM\Id, ORM\ManyToOne(inversedBy: 'foos'), ORM\JoinColumn(nullable: false)]
    private readonly FooCollection $collection;

    #[ORM\Id, ORM\Column]
    private readonly MyEnum $myEnum;
}
```

```php
#[ORM\Entity, ORM\Table(name: 'foo_collection')]
class FooCollection
{
    #[ORM\Id, ORM\Column]
    private int $id;

    /** @var Collection<int, Foo> */
    #[ORM\OneToMany(targetEntity: Foo::class, mappedBy: 'collection')]
    private Collection $foos;
}
```

```mysql
create table foo_collection
(
    id int not null primary key
);

create table foo
(
    collectionId int          not null,
    myEnum       varchar(255) not null,
    primary key (collectionId, myEnum),
    constraint foo_collection_id_fk
        foreign key (collectionId) references foo_collection (id)
);

INSERT INTO foo_collection (id) VALUES (1);
INSERT INTO foo (collectionId, myEnum) VALUES (1, 'yellow'), (1, 'blue');
```

And then, just run 2 queries to trigger the loading from the identity map
```php
    public function test(EntityManagerInterface $entityManager): void
    {
        $entityManager
            ->getRepository(FooCollection::class)
            ->createQueryBuilder('collection')
            ->leftJoin('collection.foos', 'foo')->addSelect('foo')
            ->getQuery()
            ->getResult();

        $entityManager
            ->getRepository(FooCollection::class)
            ->createQueryBuilder('collection')
            ->leftJoin('collection.foos', 'foo')->addSelect('foo')
            ->getQuery()
            ->getResult();
    }
```

#### Expected behavior

No error

### LINKED ISSUE #10471: Fail to load entity that has enum as primary key via ManyToOne association
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.14.1

#### Summary

I have an entity "Field" that has a ManyToOne relationship to "FieldType". The ID column of FieldType is backed by a native PHP enum FieldTypeID.

I can load a FieldType entity just fine, but if I load a Field and try to access the FieldType from it, I get an error.

#### Previous behavior

This was working fine in 2.14.0

#### Current behavior

Exception is thrown:

Object of class App\\Entity\\Enum\\FieldTypeID could not be converted to string
vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php
Line: 1575

#### How to reproduce

```
class Field
{
    #[ORM\Id]
    #[ORM\GeneratedValue(strategy: 'IDENTITY')]
    #[ORM\Column(name: 'id', type: Types::INTEGER, nullable: false)]
    private ?int $id = null;

    #[ORM\ManyToOne(targetEntity: FieldType::class)]
    #[ORM\JoinColumn(name: 'type_id', referencedColumnName: 'id', nullable: false)]
    private FieldType $type;

    public function getId(): ?int
    {
        return $this->id;
    }

    public function getType(): ?FieldType
    {
        return $this->type;
    }
}

class FieldType
{
    #[ORM\Id]
    #[ORM\GeneratedValue(strategy: 'NONE')]
    #[ORM\Column(name: 'id', type: Types::INTEGER, nullable: false, enumType: FieldTypeID::class)]
    private readonly FieldTypeID $id;

    public function getId(): int
    {
        return $this->id->value;
    }
}

enum FieldTypeID: int
{
    case FOO = 1;
    case BAR = 2;
    case BAZ = 3;
}

// accessing FieldType directly works OK
$fieldType = $doctrine->getRepository(FieldType::class)->find(1);
$id = $fieldType->getId();

// accessing via association fails
$field = $doctrine->getRepository(Field::class)->find(1);
$fieldType = $field->getType();
$id = $fieldType->getId(); // exception thrown here
```

Gwemox:
**Current situation**
When an entity have a backed enum as identifier, Unit Of Work tries to cast to string when generating the hash of the id.

`Error "Object of class MyEnum could not be converted to string"`

**Suggested fix**
This PR fixes the problem by adding function to get id of entity with backed enum check.

**Tests added**
It seems tests for this feature were completely missing, so I added them.

Fixes https://github.com/doctrine/orm/issues/10471 
Fixes https://github.com/doctrine/orm/issues/10334

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1566,14 +1566,8 @@ class UnitOfWork implements PropertyChangedListener
     public function addToIdentityMap($entity)
     {
         $classMetadata = $this->em->getClassMetadata(get_class($entity));
-        $identifier    = $this->entityIdentifiers[spl_object_id($entity)];
-
-        if (empty($identifier) || in_array(null, $identifier, true)) {
-            throw ORMInvalidArgumentException::entityWithoutIdentity($classMetadata->name, $entity);
-        }
-
-        $idHash    = implode(' ', $identifier);
-        $className = $classMetadata->rootEntityName;
+        $idHash        = $this->getIdHashByEntity($entity);
+        $className     = $classMetadata->rootEntityName;
 
         if (isset($this->identityMap[$className][$idHash])) {
             return false;
@@ -1582,6 +1576,50 @@ class UnitOfWork implements PropertyChangedListener
         $this->identityMap[$className][$idHash] = $entity;
 
         return true;
+    }
+
+    /**
+     * Gets the id hash of an entity by its identifier.
+     *
+     * @param array<string|int, mixed> $identifier The identifier of an entity
+     *
+     * @return string The entity id hash.
+     */
+    final public static function getIdHashByIdentifier(array $identifier): string
+    {
+        return implode(
+            ' ',
+            array_map(
+                static function ($value) {
+                    if ($value instanceof BackedEnum) {
+                        return $value->value;
+                    }
+
+                    return $value;
+                },
+                $identifier
+            )
+        );
+    }
+
+    /**
+     * Gets the id hash of an entity.
+     *
+     * @param object $entity The entity managed by Unit Of Work
+     *
+     * @return string The entity id hash.
+     */
+    public function getIdHashByEntity($entity): string
+    {
+        $identifier = $this->entityIdentifiers[spl_object_id($entity)];
+
+        if (empty($identifier) || in_array(null, $identifier, true)) {
+            $classMetadata = $this->em->getClassMetadata(get_class($entity));
+
+            throw ORMInvalidArgumentException::entityWithoutIdentity($classMetadata->name, $entity);
+        }
+
+        return self::getIdHashByIdentifier($identifier);
     }
 
     /**
@@ -1686,7 +1724,7 @@ class UnitOfWork implements PropertyChangedListener
     {
         $oid           = spl_object_id($entity);
         $classMetadata = $this->em->getClassMetadata(get_class($entity));
-        $idHash        = implode(' ', $this->entityIdentifiers[$oid]);
+        $idHash        = self::getIdHashByIdentifier($this->entityIdentifiers[$oid]);
 
         if ($idHash === '') {
             throw ORMInvalidArgumentException::entityHasNoIdentity($entity, 'remove from identity map');
@@ -1756,7 +1794,7 @@ class UnitOfWork implements PropertyChangedListener
         }
 
         $classMetadata = $this->em->getClassMetadata(get_class($entity));
-        $idHash        = implode(' ', $this->entityIdentifiers[$oid]);
+        $idHash        = self::getIdHashByIdentifier($this->entityIdentifiers[$oid]);
 
         return isset($this->identityMap[$classMetadata->rootEntityName][$idHash]);
     }
@@ -2713,7 +2751,7 @@ class UnitOfWork implements PropertyChangedListener
         $class = $this->em->getClassMetadata($className);
 
         $id     = $this->identifierFlattener->flattenIdentifier($class, $data);
-        $idHash = implode(' ', $id);
+        $idHash = self::getIdHashByIdentifier($id);
 
         if (isset($this->identityMap[$class->rootEntityName][$idHash])) {
             $entity = $this->identityMap[$class->rootEntityName][$idHash];
@@ -2872,7 +2910,7 @@ class UnitOfWork implements PropertyChangedListener
                     // Check identity map first
                     // FIXME: Can break easily with composite keys if join column values are in
                     //        wrong order. The correct order is the one in ClassMetadata#identifier.
-                    $relatedIdHash = implode(' ', $associatedId);
+                    $relatedIdHash = self::getIdHashByIdentifier($associatedId);
 
                     switch (true) {
                         case isset($this->identityMap[$targetClass->rootEntityName][$relatedIdHash]):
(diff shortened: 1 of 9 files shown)


Gwemox:
Can someone validate the workflow?

derrabus:
Right away, sorry for the delay. 😓

Gwemox:
Thank's @derrabus !
I fixed a problem

Gwemox:
@derrabus Does this seem correct to you?

pierrelemee:
Hey 👋 @Gwemox I also faced this issue today (I'm glad to be part of the _composite key / enum_ enthousiasts group ✊ ) so when I was about to open a pull request I saw yours.

Good job and perfect timing! Obviously 👍  for this

Gwemox:
@greg0ire can you review this ? Do you expect anything from me?

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1584,6 +1578,50 @@ public function addToIdentityMap($entity)
         return true;
     }
 
+    /**
+     * Gets the id hash of an entity by its identifier.
+     *
+     * @param array<mixed, mixed> $identifier The identifier of an entity
```
greg0ire: What's the key type for this array? It should be at least `array-key`, but maybe you can narrow it down even more. Maybe this is even a `list<mixed>`

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1584,6 +1578,50 @@ public function addToIdentityMap($entity)
...
+     *
+     * @return string The entity id hash.
```
greg0ire, line 1586: You can use a return type

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1584,6 +1578,50 @@ public function addToIdentityMap($entity)
...
+     */
+    public static function getIdHashByIdentifer($identifier)
```
greg0ire: Let's make this method `final` please.

greg0ire: Also, it should be "Identifier" with an i

tests/Doctrine/Tests/Models/GH10334/GH10334FooCollection.php
```diff
@@ -0,0 +1,48 @@
...
+     * @var int
+     * @Id
+     * @Column(type="integer")
+     * @GeneratedValue
+     */
+    protected $id;
+
+    /**
+     * @OneToMany(targetEntity="GH10334Foo", mappedBy="collection", cascade={"persist", "remove"})
+     * @var Collection $foos
```
greg0ire:
```suggestion
     * @var Collection<GH10334Foo> $foos
```

tests/Doctrine/Tests/Models/GH10334/GH10334FooCollection.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $this->foos = new ArrayCollection();
+    }
+
+    public function getId(): int
+    {
+        return $this->id;
+    }
+
+    public function getFoos(): Collection
```
greg0ire, line 42:
```suggestion
    /**
     * @return Collection<GH10334Foo>
     */
    public function getFoos(): Collection
```

tests/Doctrine/Tests/Models/GH10334/GH10334Product.php
```diff
@@ -0,0 +1,70 @@
...
+        $this->name        = $name;
+        $this->productType = $productType;
+    }
+
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    public function setName(string $name): void
```
greg0ire: Please remove any unused getter/setter

lib/Doctrine/ORM/UnitOfWork.php
Gwemox: PHPStan tells me it's called with an int or string key.

greg0ire: Ok, then use `array-key` I guess.

Gwemox:
Thank you for this quick review, I made the requested changes.

greg0ire (dismissed):

BartVanGestel:
@Gwemox @greg0ire  could you merge this PR, we currently have this exact problem and this will fix alot of unneeded persists.

Gwemox:
> @Gwemox @greg0ire could you merge this PR, we currently have this exact problem and this will fix alot of unneeded persists.

I don't have merge rights

greg0ire (approved):

greg0ire:
Ping @SenseException @derrabus

SenseException (approved):

wmouwen:
@greg0ire Could you please tag a release containing this bugfix? It is cause for major locking issues on our systems and blocking releases of our software.

2023-05-31-090211_771x294_scrot

PR #10509, opened 2023-02-08
Migrate the rest of the source code to PHP 8 syntax

### LINKED ISSUE #9772: Migrate to PHP 8.1
This is something that I think should go in milestone 3.0.0

A big difference between branches 2 and 3 is the `php` requirement:

```diff
- "php": "^7.1 || ^8.0",
+ "php": "^8.1",
```

Now that support for PHP 7 has been dropped, we can use features that are specific to PHP 8.

@derrabus has already done a lot about that. When this is over, we should be able to get rid of lines mentioning native type declarations in `phpcs.xml.dist`. It can also be used to figure out what is left to do.

We should rely on constructor property promotion when possible.

phpdoc that is found to be wrong on the way should be fixed on the next patch branch.
phpdoc that is found to be imprecise on the way should be improved on the next minor branch. 

Here is an example for `AbstractQuery` and related classes:
- patch: https://github.com/doctrine/orm/pull/9774
- minor: https://github.com/doctrine/orm/pull/9777
- major: https://github.com/doctrine/orm/pull/9776

greg0ire:
Closes #9772 (yes, finally!)

None of the remaining changes look legit.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -11,6 +11,7 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\Persistence\Mapping\ClassMetadata as PersistenceClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\FileDriver;
+use Doctrine\Persistence\Mapping\Driver\FileLocator;
 use DOMDocument;
 use InvalidArgumentException;
 use LogicException;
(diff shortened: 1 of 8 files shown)


lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -41,14 +41,14 @@ class XmlDriver extends FileDriver
...
-    private $isXsdValidationEnabled;
-
     /**
      * {@inheritDoc}
      */
-    public function __construct($locator, $fileExtension = self::DEFAULT_FILE_EXTENSION, bool $isXsdValidationEnabled = false)
-    {
+    public function __construct(
+        $locator,
+        $fileExtension = self::DEFAULT_FILE_EXTENSION,
```
derrabus: Types? 🤔

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -41,14 +41,14 @@ class XmlDriver extends FileDriver
 {
     public const DEFAULT_FILE_EXTENSION = '.dcm.xml';
```
derrabus, line 43: `final`?

greg0ire: `SimplifiedXmlDriver` overrides it.

derrabus: 😭

derrabus (approved):

PR #10510, opened 2023-02-08
Suppress three Psalm errors caused by DBAL 3.6.0

mpdude:
On CI, Psalm will run with the just-released DBAL 3.6.0, which causes three errors. To keep builds here green until that is fixed and released, suppress the errors for the time being.


mpdude:
#10511 brings #10507 to 2.15, so it’s a better fix

PR #10513, opened 2023-02-09
Use array shapes where appropriate

greg0ire:
Working on converting these array shapes to DTO allowed me to find every signature where they are supposed to be used.

The Psalm baseline gets worse because it considers accessing an array key differently depending on whether it is defined vaguely, as `array<string, mixed>`, or precisely, as `array{my-key?: string}`.

See https://psalm.dev/r/5940ba233e

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
@@ -24,6 +24,8 @@ use function sprintf;
 
 /**
  * Persister for many-to-many collections.
+ *
+ * @psalm-import-type AssociationMapping from ClassMetadata
  */
 class ManyToManyPersister extends AbstractCollectionPersister
 {
@@ -286,7 +288,7 @@ class ManyToManyPersister extends AbstractCollectionPersister
      * JOIN.
      *
      * @param mixed[] $mapping Array containing mapping information.
-     * @psalm-param array<string, mixed> $mapping
+     * @psalm-param AssociationMapping $mapping
      *
      * @return string[] ordered tuple:
      *                   - JOIN condition to add to the SQL
@@ -339,7 +341,7 @@ class ManyToManyPersister extends AbstractCollectionPersister
      * Generate ON condition
      *
      * @param mixed[] $mapping
-     * @psalm-param array<string, mixed> $mapping
+     * @psalm-param AssociationMapping $mapping
      *
      * @return string[]
      * @psalm-return list<string>
--- a/lib/Doctrine/ORM/Cache/CacheFactory.php
+++ b/lib/Doctrine/ORM/Cache/CacheFactory.php
@@ -14,6 +14,8 @@ use Doctrine\ORM\Persisters\Entity\EntityPersister;
 
 /**
  * Contract for building second level cache regions components.
+ *
+ * @psalm-import-type AssociationMapping from ClassMetadata
  */
 interface CacheFactory
 {
@@ -31,7 +33,7 @@ interface CacheFactory
     /**
      * Build a collection persister for the given relation mapping.
      *
-     * @param mixed[] $mapping The association mapping.
+     * @param AssociationMapping $mapping The association mapping.
      *
      * @return CachedCollectionPersister
      */
--- a/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
+++ b/lib/Doctrine/ORM/Cache/DefaultCacheFactory.php
@@ -138,6 +138,7 @@ class DefaultCacheFactory implements CacheFactory
      */
     public function buildCachedCollectionPersister(EntityManagerInterface $em, CollectionPersister $persister, array $mapping)
     {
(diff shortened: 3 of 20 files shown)


lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
```diff
@@ -24,6 +24,8 @@
 
 /**
  * Persister for many-to-many collections.
+ *
+ * @psalm-import-type AssociationMapping from ClassMetadata
```
greg0ire, line 28: The most Psalm errors occur in this file, because Psalm does not know that `joinTable` is always set for mappings that represent the owning side of a many to many relationship.

As it stands, https://github.com/doctrine/orm/pull/10405 won't fix that either, because I did not create a separate type for owning sides, but maybe I should?

## Before

```mermaid
classDiagram
    AssociationMapping <|-- OneToOneAssociationMapping
    AssociationMapping <|-- OneToManyAssociationMapping
    AssociationMapping <|-- ManyToOneAssociationMapping
    AssociationMapping <|-- ManyToManyAssociationMapping
```

## After

For `ManyToOne` and for `OneToMany`, `isOwningSide` is redundant according to [the docs](https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/unitofwork-associations.html#association-updates-owning-side-and-inverse-side). I think this could be modeled with an interface (not implementing it would mean the relationship is owned).

```mermaid
classDiagram
    AssociationMapping <|-- OneToOneAssociationMapping
    AssociationMapping <|-- OneToManyAssociationMapping
    AssociationMapping <|-- ManyToOneAssociationMapping
    AssociationMapping <|-- ManyToManyAssociationMapping
    ManyToManyAssociationMapping <|-- OwningManyToManyAssociationMapping
    ManyToManyAssociationMapping <|-- OwnedManyToManyAssociationMapping
    OneToOneAssociationMapping <|-- OwningOneToOneAssociationMapping
    OneToOneAssociationMapping <|-- OwnedOneToOneAssociationMapping
```

Cc @mpdude

mpdude: How could/would Psalm know when/where you're dealing with an `AssociationMapping` that represents the owning side? Would have have to make `instanceof` checks?

greg0ire: Yes, exactly! We would replace the guard check here, for instance:

https://github.com/doctrine/orm/blob/31ff969628946ef2df87d0b6d49dabb1cc40119d/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php#L37-L46

Looking at this, you can see how the owning and owned sides are very different in the case of the many to many relationship: https://github.com/doctrine/orm/blob/0f77181a2cd09eb124c8eedcf5a6d9f48ad21f28/lib/Doctrine/ORM/Mapping/ClassMetadata.php#L1802-L1886

mpdude: Then let's see that we define the class hierarchy not only by technical criteria (i.e. which fields are common in several classes and extract those into a base class), but from a semantical point of view. What kind of checks do we have in the codebase?

For example `if ($mapping['isOwningSide']) ...` would become `if ($mapping instanceof AssociationOwningSide) ...` or so.

greg0ire: I think we have two kind of checks: you just mentioned the first one, and the other one is against `$mapping['type']`, sometimes with the `&` operator, for instance to detect if it's a `to-many`, for instance here:
https://github.com/doctrine/orm/blob/31ff969628946ef2df87d0b6d49dabb1cc40119d/lib/Doctrine/ORM/PersistentCollection.php#L354

We might need to introduce `ToOneAssociationMapping` and `ToManyAssociationMapping` interfaces. ATM I used those names for traits but I can change those to `To{One,Many}AssociationMappingTrait` I guess.

The thing is, interfaces can't have properties, so it's only a viable option if we switch to methods instead.

greg0ire: Ah but we only ever test for `TO_MANY` and `TO_ONE`, never for `ONE_TO` or `MANY_TO` (those constants don't even exist), so it should be possible to just include these as classes in the hierarchy. I will ditch the traits and do that instead.

greg0ire: I pushed a new commit on the other PR, switching from traits to classes.

> The thing is, interfaces can't have properties, so it's only a viable option if we switch to methods instead.

Just realised that this might not be an issue because we should be able to change the phpdoc on `ClassMetadata::AssociationMapping` from `array<string, AssociationMapping>` to `array<string, list-of-concrete-classes>`. An `instanceof` check should allow SA to rule out some of the classes and to notice that some of them have common properties :crossed_fingers:

SenseException (approved):

PR #10516, opened 2023-02-11
fix: use executeStatement in SchemaTool

### LINKED ISSUE #9556: Permission denied when creating schema for primary/replicas
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11

#### Summary

Creating the schema from the command line fails if a primary/replica configuration is used.

    php bin/console doctrine:schema:create

SQL commands are sent to the replica instead of the primary db.

The culprit is [SchemaTool](https://github.com/doctrine/orm/blob/2.11.x/lib/Doctrine/ORM/Tools/SchemaTool.php#L98)

It executes CREATE statements with `executeQuery($sql)` instead of `executeStatement($sql)`.

#### Current behavior

Creating the schema fails due to CREATE statement being denied on replica. The replica is read only. The schema can't be created on the replica.

#### How to reproduce

Unless you have two db servers configured as master/slave, the setup can be replicated on a single db server. Create a R/W user to simulate the primary and a R/O user with SELECT permission to simulate the replica.

~~~~~
doctrine:
    dbal:
        driver: 'pdo_mysql'
        server_version: 'mariadb-10.3.11'
        charset: utf8mb4
        url: 'mysql://live_fred:password@127.0.0.1:3306/live_fred'
        replicas:
            replica1:
                url: 'mysql://live_fred_ro:password@127.0.0.1:3306/live_fred'
~~~~~

Run the command:

    php bin/console doctrine:schema:create

~~~~~
In ToolsException.php line 19:
                                                                                                                                                                                                                                            
  Schema-Tool failed with Error 'An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1142 CREATE command denied to user 'live_fred_ro'@'localhost' for table 'core_print_ticket_translation'  
  ' while executing DDL: CREATE TABLE core_print_ticket_translation … 
~~~~~


#### Expected behavior

SQL CREATE statements must be sent to the primary.

CREATE statements are not queries. SchemaTool must call `executeStatement($sql)`.

simPod:
When using master replica connection the query is routed to replica. We need to use `executeStatement` so mutation queries are routed to master.

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaTool.php
+++ b/lib/Doctrine/ORM/Tools/SchemaTool.php
@@ -97,7 +97,7 @@ class SchemaTool
 
         foreach ($createSchemaSql as $sql) {
             try {
-                $conn->executeQuery($sql);
+                $conn->executeStatement($sql);
             } catch (Throwable $e) {
                 throw ToolsException::schemaToolFailure($sql, $e);
             }
@@ -831,7 +831,7 @@ class SchemaTool
 
         foreach ($dropSchemaSql as $sql) {
             try {
-                $conn->executeQuery($sql);
+                $conn->executeStatement($sql);
             } catch (Throwable $e) {
                 // ignored
             }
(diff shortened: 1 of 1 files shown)


mpdude (approved):

derrabus (approved):

derrabus:
Thank you @simPod!

PR #10518, opened 2023-02-13
Turn deprecation from #10470 into an exception in 3.0.x

mpdude:


mpdude:
Closing – wrong head branch, cannot fix this.

PR #10519, opened 2023-02-13
Deprecate overriding associations not inherited from a mapped superclass

mpdude:
This follows up on #10470, this time taking care of the associations.

I use the same UPGRADE notice, deprecation text etc. since it already covered this use case as well.


SenseException (approved):

greg0ire (approved):

greg0ire:
Thanks @mpdude!

Issue #10514, opened 2023-02-09
2.14.0 BC Break: UOW::normalizeIdentifier() bug


darckking:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.14.0

#### Current behavior

After doctrine/orm upgrade from 2.13.4 to 2.14.1 I get "Uncaught Error: Object of class **{PROXY_CLASS}** could not be converted to string in **{PROJECT_PATH}**/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:3160"

#### Summary
I looked into UOW internals a little bit and I traced down the issue to normalizeIdentifier() function. When ORM creates the entity B  (UOW::createEntity()) and processes the association to entity C, the UOW::normalizeIdentifier() in recursive calls, it passes the A proxy object as an id to EM::getReference() and it crashes in UOW::tryGetById() as implode expects array of strings.

#### How to reproduce

To reproduce the problem I created a test project from scratch.

##### Dependencies in composer.json
```json
    "require": {
        "doctrine/orm": "2.14.1",
        "symfony/cache": "^6.2",
        "doctrine/annotations": "^1.13",
        "doctrine/collections": "^1.5",
        "doctrine/event-manager": "^1.1",
        "doctrine/lexer": "^1.2.3"
    }
```
**Note**: I don't use latest versions of doctrine/annotations, doctrine/collections, doctrine/event-manager, doctrine/lexer because they are not compatible with doctrine/orm:2.13.5 and for the BC Break demonstration purpose I swap doctrine/orm 2.14.1 with 2.13.5

##### Entities

```php
<?php

namespace DoctrineORMTest;

use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity()
 * @ORM\Table(name="A")
 */
class A
{
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    protected int $id;
}
``` 

```php
<?php

namespace DoctrineORMTest;

use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity()
 * @ORM\Table(name="B")
 */
class B
{
    /**
     * @ORM\Id
     * @ORM\OneToOne(targetEntity="DoctrineORMTest\A")
     * @ORM\JoinColumn(name="id", referencedColumnName="id", nullable=false, onDelete="CASCADE")
     */
    protected $a;

[...]

PR #10521, opened 2023-02-13
Add a failing test for #10514

mpdude:
Here is a failing test for #10514. The commit that causes breakage is e5e674c68630c38d7da5241ffc316c5fb18edfec, the parent commit still works.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10514Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10514Test.php
@@ -1,0 +1,84 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH10514Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->createSchemaForModels(
+            GH10514A::class,
+            GH10514B::class,
+            GH10514C::class
+        );
+    }
+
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
The fix might be something like this, but if I apply it the test fails with a schema error.
Can you have a look?

```diff
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -3157,7 +3157,11 @@ class UnitOfWork implements PropertyChangedListener
      */
     public function tryGetById($id, $rootClassName)
     {
-        $idHash = implode(' ', (array) $id);
+        $class  = $this->em->getClassMetadata($rootClassName);
+        $idHash = implode(' ', $class->containsForeignIdentifier || $class->containsEnumIdentifier
+            ? $this->identifierFlattener->flattenIdentifier($class, (array) $id)
+            : (array) $id
+        );
 
         return $this->identityMap[$rootClassName][$idHash] ?? false;
     }
```

mpdude:
Will do but can happen next wednesday in the best case

nicolas-grekas:
Friendly ping @mpdude

mpdude:
Ouch, lost sight of this, sorry. I'm about to leave for summer vacation and not sure how much I can help in the next three to four weeks.

I have the (yet to be proven) impression that there are several places dealing with "id hash" computation, and that a few of them were touched in the last months when either backed enums and/or objects-as-id-values came into play.

Maybe it might be worthfile to go out looking for usages/patterns and see if we can consolidate this.

The place Nicolas is referring to also looks to be related to it, and the patch suggested does not apply since the changes from #10508.

Wilt:
Another friendly ping @mpdude
(We are unable to migrate to later versions of Doctrine because of this bug. It is a major breaking issue for our whole application).

PR #10528, opened 2023-02-19
Add QueryBuilder::when method

jdecool:
Add a `when` helper method that allow to limit `if-else` condition in our code.

For exemple:

```php
$qb = $this->createQueryBuilder();
$qb->when($isEnabled === true, static fn (QueryBuilder $q) => $q->andWhere('c.enabled = true'));
```

If the condition is `false`, it's possible to add a default query expression, otherwise it will do nothing.

### DIFF
--- a/lib/Doctrine/ORM/QueryBuilder.php
+++ b/lib/Doctrine/ORM/QueryBuilder.php
@@ -1209,6 +1209,27 @@ class QueryBuilder
     }
 
     /**
+     * Adds restriction defined by the callback to the query results if a condition is true.
+     * Otherwise it apply the default if defined.
+     *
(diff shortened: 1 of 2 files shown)


derrabus:
Okay, and why exactly wouldn't I use an `if` block here?

jdecool:
This is syntactic sugar that we can use to approach functional programming.

It is a way to reduce the indentation levels of the code.

It's a bit like using an `array_map` when you can do a `foreach` loop.

derrabus:
Thank you for the explanation. I'm afraid, that's not enough added value for me to merge and maintain this feature.

Thank you for your proposal though!

PR #10529, opened 2023-02-20
Correct use of PHP attribute

joshpme:
Incorrect syntax used in php 8 attribute. This should be key: value, rather than key=value


greg0ire (approved):

greg0ire:
Thanks @joshpme

PR #10530, opened 2023-02-21
Add handling of 'not' composite expression in QueryExpressionVisitor

oscmarb:
Add handling of ```NOT``` composite expression in ```QueryExpressionVisitor::walkCompositeExpression``` like in ```Expr::not```.

Before, if you ran the following criteria it gave the error ```Unknown composite NOT```:

```php
new Criteria(
    new CompositeExpression(
        CompositeExpression::TYPE_NOT,
        [new Comparison('id', Comparison::EQ, '1')]
    )
);
```

With the introduced changes, a handler for the ```NOT``` expression is found and the query is constructed correctly:

```sql
SELECT * FROM Entity u0_ WHERE NOT (u0_.id = ?)
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,11 +20,11 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^8.1",
         "composer-runtime-api": "^2",
         "ext-ctype": "*",
         "doctrine/cache": "^1.12.1 || ^2.1.1",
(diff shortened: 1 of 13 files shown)


mpdude:
If this is a new feature, it should be based on 2.15.x.

And, are there tests for the other expression types and could you add a test for the new one following the same pattern?

Thank you!

mpdude:
@greg0ire could you approve workflow runs here please?

PR #10531, opened 2023-02-21
Add test: Commit order calculation must happen on the entity level

mpdude:
Add tests for entity insertion and deletion that require the commit order calculation to happen on the entity level. This demonstrates the necessity for the changes in #10547.

This PR contains two tests with carefully constructed entity relationships, where we have a non-nullable `parent` foreign key relationships between entities stored in the same table. 

Class diagram:

```mermaid
classDiagram
    direction LR
    class A
    class B
    A --> B : parent
    B --|> A
```

Object graph:

```mermaid
graph LR;
    b1 --> b2;
    b2 --> a;
    b3 --> b2;
```

#### Situation before #10547 

The commit order is computed by looking at the associations at the _table_ (= _class_) level. Once the ordering of _tables_ has been found, entities being mapped to the same table will be processed in the order they were given to `persist()` or `remove()`.

That means only a particular ordering of `persist()` or `remove()` calls (see comment in the test) works: 

For inserts, the order must be `$a, $b2, $b1, $b3` (or `... $b3, $b1`), for deletions `$b1, $b3, $b2, $a`.

#### Situation with entity-level commit order computation (as in #10547)

The ORM computes the commit order by considering associations at the _entity_ level. It will be able to find a working order by itself.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10531Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10531Test.php
@@ -1,0 +1,151 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH10531Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->createSchemaForModels(
+            GH10531A::class,
+            GH10531B::class
+        );
+    }
+
+    public function tearDown(): void
+    {
+        $conn = static::$sharedConn;
+        $conn->executeStatement('DELETE FROM gh10531_b');
+        $conn->executeStatement('DELETE FROM gh10531_a');
+    }
+
+    public function testInserts(): void
+    {
+        $a  = new GH10531A();
+        $b1 = new GH10531B();
+        $b2 = new GH10531B();
+        $b3 = new GH10531B();
+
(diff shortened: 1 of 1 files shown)


mpdude:
I rebased this onto the branch where we implement a fix, so it’s no longer a “failing test” (remove tag?).

mpdude:
@SenseException and @greg0ire: I've rewritten the OP to make it more concise. Hopefully that makes it easier for you to understand what this test tries to show?

It's failing on the current `2.16.x` branch, but passes against `entity-level-commit-order`.

greg0ire:
Just read the OP, fully understood it I think, thanks! Let me review the diff now.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10531Test.php
```diff
@@ -0,0 +1,151 @@
...
+        $this->_em->flush();
+    }
+}
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="gh10531_a")
+ * @ORM\DiscriminatorColumn(name="discr", type="string")
+ * @ORM\DiscriminatorMap({ "A": "GH10531A", "B": "GH10531B" })
+ * @ORM\InheritanceType("JOINED", )
```
greg0ire: :raised_eyebrow: 

```suggestion
 * @ORM\InheritanceType("JOINED")
```

greg0ire:
:thinking: I'm going to merge 2.15.x into entity-level-commit order

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @mpdude !

PR #10532, opened 2023-02-21
Add test: Entity insertions must not happen table-wise

mpdude:
Add tests for entity insertion and deletion that require writes to different tables in an interleaved fashion, and that have to re-visit a particular table. This is part of the larger story in #10547.

#### Background 

Before #10547, the UoW computes the commit order on the table level. All entity insertions for a particular table are passed through `EntityPersister::addInsert()` and finally written through `EntityPersister::executeInserts()`. This is also noted here:

https://github.com/doctrine/orm/blob/dba90c1a91283f1252c8380fec9b35bf7b1d28ec/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L73-L74

#### Suggested change

The test in this PR contains a carefully constructed set of four entities. Two of them are of the same class (are written to the same table), but require other entities to be processed first. 

In order to be able to insert this set of entities, the ORM must be able to perform inserts for a given table repeatedly, interleaved with writing other entities to their respective tables. The UnitOfWork can no longer batch inserts per table, but will have to call `EntityPersister::addInsert()` and `EntityPersister::executeInserts()` multiple times in succession – this is implemented in #10547.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10532Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10532Test.php
@@ -1,0 +1,186 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH10532Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->createSchemaForModels(
+            GH10532A::class,
+            GH10532B::class,
+            GH10532C::class,
+            GH10532X::class
+        );
+    }
+
+    public function tearDown(): void
+    {
+        $conn = static::$sharedConn;
+        $conn->executeStatement('DELETE FROM gh10532_c');
+        $conn->executeStatement('DELETE FROM gh10532_b');
+        $conn->executeStatement('DELETE FROM gh10532_a');
+        $conn->executeStatement('DELETE FROM gh10532_x');
+    }
+
+    public function testInserts(): void
+    {
+        // Dependencies are $a1 -> $b -> $a2 -> $c
+
+        $a1 = new GH10532A();
+        $b  = new GH10532B();
+        $a2 = new GH10532A();
+        $c  = new GH10532C();
+
+        $a1->x = $b;
+        $b->a  = $a2;
+        $a2->x = $c;
+
+        /*
+         * The following would force a working commit order, but that's not what
+         * we want (the ORM shall sort this out internally).
+         *
+         * $this->_em->persist($c);
+         * $this->_em->persist($a2);
+         * $this->_em->flush();
+         * $this->_em->persist($b);
+         * $this->_em->persist($a1);
+         * $this->_em->flush();
+         */
+
+        $this->_em->persist($a1);
+        $this->_em->persist($a2);
+        $this->_em->persist($b);
+        $this->_em->persist($c);
+        $this->_em->flush();
+
+        self::assertNotNull($a1->id);
+        self::assertNotNull($b->id);
+        self::assertNotNull($a2->id);
+        self::assertNotNull($c->id);
+    }
+
+    public function testDeletes(): void
+    {
+        // Dependencies are $a1 -> $b -> $a2 -> $c
+
+        $this->expectNotToPerformAssertions();
+        $con = $this->_em->getConnection();
+
+        // The "c" entity
+        $con->insert('gh10532_x', ['id' => 1, 'discr' => 'C']);
+        $con->insert('gh10532_c', ['id' => 1]);
+        $c = $this->_em->find(GH10532C::class, 1);
+
+        // The "a2" entity
+        $con->insert('gh10532_a', ['id' => 2, 'gh10532x_id' => 1]);
+        $a2 = $this->_em->find(GH10532A::class, 2);
+
+        // The "b" entity
+        $con->insert('gh10532_x', ['id' => 3, 'discr' => 'B']);
+        $con->insert('gh10532_b', ['id' => 3, 'gh10532a_id' => 2]);
+        $b = $this->_em->find(GH10532B::class, 3);
+
+        // The "a1" entity
+        $con->insert('gh10532_a', ['id' => 4, 'gh10532x_id' => 3]);
+        $a1 = $this->_em->find(GH10532A::class, 4);
+
+        /*
+         * The following would make the deletions happen in an order
+         * where the not-nullable foreign key constraints would not be
+         * violated. But, we want the ORM to be able to sort this out
+         * internally.
+         *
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH10532Test.php
```diff
@@ -0,0 +1,186 @@
...
+
+        $this->createSchemaForModels(
+            GH10532A::class,
+            GH10532B::class,
+            GH10532C::class,
+            GH10532X::class
+        );
+    }
+
+    public function tearDown(): void
```
mpdude, line 24: @greg0ire you recently mentioned `tearDown` is no good/rarely necessary?

In this case, it seems the tests fail without it, possibly because we need a particular sequence to drop tables?

greg0ire: Without `tearDown()` , the tests fail because something else drops table, in a wrong order?

greg0ire: I tried removing `tearDown()`, and tests fail because of the insert on line 78, which violates a unique constraint.

mpdude: Yes… https://github.com/doctrine/orm/actions/runs/5158576597/jobs/9292414242

Though I cannot fully reproduce it locally.

Maybe somehow the automatic tear down does not work and so there are leftovers in the DB when the next test runs?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10532Test.php
```diff
@@ -0,0 +1,186 @@
...
+    {
+        // Dependencies are $a1 -> $b -> $a2 -> $c
+
+        $this->expectNotToPerformAssertions();
+        $con = $this->_em->getConnection();
+
+        // The "c" entity
+        $con->insert('gh10532_x', ['id' => 1, 'discr' => 'C']);
+        $con->insert('gh10532_c', ['id' => 1]);
+        $c = $this->_em->find(GH10532C::class, 1);
```
greg0ire, line 80: I think it could be possible to remove ids by using `$con->lastInsertId()`

mpdude: Is it a problem to use tearDown and have fixed IDs here?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10532Test.php
```diff
@@ -0,0 +1,186 @@
...
+        $a2 = $this->_em->find(GH10532A::class, 2);
+
+        // The "b" entity
+        $con->insert('gh10532_x', ['id' => 3, 'discr' => 'B']);
+        $con->insert('gh10532_b', ['id' => 3, 'gh10532a_id' => 2]);
+        $b = $this->_em->find(GH10532B::class, 3);
+
+        // The "a1" entity
+        $con->insert('gh10532_a', ['id' => 4, 'gh10532x_id' => 3]);
+        $a1 = $this->_em->find(GH10532A::class, 4);
```
greg0ire, line 93: This allows to remove the `tearDown()`

```suggestion
        $con->insert('gh10532_x', ['discr' => 'C']);
        $cId = $con->lastInsertId();
        $con->insert('gh10532_c', ['id' => $cId]);
        $c = $this->_em->find(GH10532C::class, $cId);

        // The "a2" entity
        $con->insert('gh10532_a', ['gh10532x_id' => $cId]);
        $a2Id = $con->lastInsertId();
        $a2   = $this->_em->find(GH10532A::class, $a2Id);

        // The "b" entity
        $con->insert('gh10532_x', ['discr' => 'B']);
        $bId = $con->lastInsertId();
        $con->insert('gh10532_b', ['id' => $bId, 'gh10532a_id' => $a2Id]);
        $b = $this->_em->find(GH10532B::class, $bId);

        // The "a1" entity
        $con->insert('gh10532_a', ['gh10532x_id' => $bId]);
        $a1Id = $con->lastInsertId();
        $a1   = $this->_em->find(GH10532A::class, $a1Id);

```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10532Test.php
greg0ire: I don't think there is any cleanup being done aside from a few "well-known" models, so yeah the DB must be full of clutter, and that should be OK.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10532Test.php
greg0ire: Aside from reduced performance due to unnecessary deletes, not AFAIK.

mpdude: Resolved this as you suggested.

greg0ire (approved):

mpdude:
@greg0ire do you know how to avoid the not-NULL problem in PostgreSQL when inserting the row that needs an auto-increment id?

greg0ire:
Not sure what you are talking about. When inserting without an ID and using postgres, there's an issue, is that it? Is that why you reverted to the previous situation?

mpdude:
Exactly, see https://github.com/doctrine/orm/actions/runs/5163227611/jobs/9301443233

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10532Test.php
```diff
@@ -0,0 +1,186 @@
...
+ * @ORM\InheritanceType("JOINED")
+ *
+ * We are using JTI here, since STI would relax the not-nullable constraint for the "parent"
+ * column. Causes another error, but not the constraint violation I'd like to point out.
+ */
+abstract class GH10532X
+{
+    /**
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
```
greg0ire, line 133: I guess I should resume work on https://github.com/doctrine/orm/pull/8931

See the value of `LEGACY_DEFAULTS_FOR_ID_GENERATION` vs `RECOMMENDED_STRATEGY`. The only RDBMS that shouldn't be using `IDENTITY` is Oracle, and we are not running the test suite with it. I'm really not sure if that will fix the issue, but if you could try with `IDENTITY` for science, that would be great.

mpdude:
Asking @SenseException for 👀  as well 🙏🏻

mpdude:
Maybe the last PR pending for the #10547 feature 🎉 ?

greg0ire:
ping @derrabus @SenseException :pray:

derrabus (approved):

greg0ire:
@mpdude time to merge `entity-level-commit-order`?

mpdude:
I'll try to prepare that via #10547 – there are merge conflicts with 2.16 that I need to resolve.

PR #10533, opened 2023-02-22
Add handling of 'not' composite expression in QueryExpressionVisitor

oscmarb:
Add handling of ```NOT``` composite expression in ```QueryExpressionVisitor::walkCompositeExpression``` like in ```Expr::not```.

Before, if you ran the following criteria it gave the error ```Unknown composite NOT```:

```php
new Criteria(
    new CompositeExpression(
        CompositeExpression::TYPE_NOT,
        [new Comparison('id', Comparison::EQ, '1')]
    )
);
```

With the introduced changes, a handler for the ```NOT``` expression is found and the query is constructed correctly:

```sql
SELECT * FROM Entity u0_ WHERE NOT (u0_.id = ?)
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,11 +20,11 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^8.1",
         "composer-runtime-api": "^2",
         "ext-ctype": "*",
         "doctrine/cache": "^1.12.1 || ^2.1.1",
-        "doctrine/collections": "^1.5 || ^2.1",
+        "doctrine/collections": "^2.1",
         "doctrine/common": "^3.0.3",
         "doctrine/dbal": "^2.13.1 || ^3.2",
         "doctrine/deprecations": "^0.5.3 || ^1",
--- a/lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
+++ b/lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
@@ -96,6 +96,9 @@ class QueryExpressionVisitor extends ExpressionVisitor
             case CompositeExpression::TYPE_OR:
                 return new Expr\Orx($expressionList);
 
+            case CompositeExpression::TYPE_NOT:
+                return new Expr\Func('NOT', $expressionList);
+
             default:
                 // Multiversion support for `doctrine/collections` before and after v2.1.0
                 if (defined(CompositeExpression::class . '::TYPE_NOT') && $expr->getType() === CompositeExpression::TYPE_NOT) {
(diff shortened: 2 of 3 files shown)


oscmarb:
@mpdude I created a new branch with the changes and another PR. I added the tests and modified the ```composer.json``` because version ```2.1``` of ```doctrine/collections``` is needed.

@greg0ire from @mpdude:

```@greg0ire could you approve workflow runs here please?```

**Comment by oscmarb** [comment]
I only added the test ```ExprTest::testAndxNotxExpr``` because some tests on ```not``` have already been implemented in the ```2.15.x``` branch, for example ```QueryExpressionVisitorTest::testWalkNotCompositeExpression```.

File: composer.json [hunk 1]
```diff
@@ -20,11 +20,11 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^8.1",
         "composer-runtime-api": "^2",
         "ext-ctype": "*",
         "doctrine/cache": "^1.12.1 || ^2.1.1",
-        "doctrine/collections": "^1.5 || ^2.1",
+        "doctrine/collections": "^2.1",
```
derrabus, line 27: Please don't drop support for Collections 1 just yet.

derrabus (changes requested):

composer.json
```diff
@@ -20,11 +20,11 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^8.1",
```
mpdude, line 23: How's that change related to the new feature?

oscmarb: Collections 2.1 only accepts php ^8.1

derrabus: This is why you cannot drop Collections 1. Sorry, this change is a show-stopper.

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
```diff
@@ -96,6 +96,9 @@ public function walkCompositeExpression(CompositeExpression $expr)
             case CompositeExpression::TYPE_OR:
                 return new Expr\Orx($expressionList);
 
+            case CompositeExpression::TYPE_NOT:
```
mpdude, line 99: Now I get it… this is a doctrine/collections ^2.1 thing?

lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
```diff
@@ -96,6 +96,9 @@ public function walkCompositeExpression(CompositeExpression $expr)
...
+                return new Expr\Func('NOT', $expressionList);
+
             default:
                 // Multiversion support for `doctrine/collections` before and after v2.1.0
```
mpdude, line 103: What is this? Does it help with collections < 2.1?

composer.json
mpdude: Anyways, the PHP Version requirement of doctrine/collections is not relevant here.

mpdude:
Maybe you can make this change possible without requiring doctrine/collections 2.x? 

In a way that the expression can be used as long as the appropriate version is installed…?

PR #10538, opened 2023-02-24
Enabled lazy-loading for subclasses to improve performance when loading entities with a to-one association

SanderBoelaars:
This PR is a continuation of the closed PR [#10466](https://github.com/doctrine/orm/pull/10466) (which cannot be reopend).

We encountered a performance problem regarding entity instances that point to parent entity instances, when the entity class was involved in entity inheritance. When the chain of parent instances was long (42) it would cause one of the following:

* When XDebug was enabled, error: `Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '256' frames`
* When XDebug was not enabled, error: `Allowed memory size of .... bytes exhausted`

The reason for this was because when loading a related entity instance of a class that has subclasses (when the association was many-to-one or one-to-one), it would fully load the related entity, regardless of whether that entity was needed. It would then also fully load that entity's parent, all the way until there is no parent anymore.

Before this change, the relevant code in UnitOfWork was:

```
case ($targetClass->subClasses):
    // If it might be a subtype, it can not be lazy. There isn't even
    // a way to solve this with deferred eager loading, which means putting
    // an entity with subclasses at a *-to-one location is really bad! (performance-wise)
    $newValue = $this->getEntityPersister($assoc['targetEntity'])->loadOneToOneEntity($assoc, $entity, $associatedId);
    break;
```

This PR changes the behavior so that the related entity will not be fully loaded, but instead will be a Proxy class that will only load extra data when it is required.
Doing this causes the loading to stop at 1 parent instead of loading every single parent up until there is no parent anymore.

It does this by doing the following:

1. Find the identifier column for this entity.
2. Find the discriminator column for this entity that is used to determine the subclass of the entity.
3. Find the discriminator column's value for the entity using the identifier by executing a query on the database.
4. Create a Proxy class for the correct subclass of the entity using the discriminatorMap.

The documentation about the limitation [linked](https://github.com/doctrine/orm/pull/10466#issuecomment-1405418619) by @mpdude is indeed relevant to the PR, but the limitation not only affects single table inheritance, because in our case we are using joined inheritance.

I created a test which simulates the scenario for single table inheritance. It defines 2 entities, and one of the entities inherits the other. We create 2 instances of the superclass, and let instance 2 point to instance 1 via a parent field. When we load instance 2, the test asserts that the parent of instance 2 is a Proxy instead of the fully loaded instance 1.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2893,19 +2893,52 @@ class UnitOfWork implements PropertyChangedListener
 
                             break;
 
-                        case $targetClass->subClasses:
-                            // If it might be a subtype, it can not be lazy. There isn't even
-                            // a way to solve this with deferred eager loading, which means putting
-                            // an entity with subclasses at a *-to-one location is really bad! (performance-wise)
-                            $newValue = $this->getEntityPersister($assoc['targetEntity'])->loadOneToOneEntity($assoc, $entity, $associatedId);
-                            break;
-
                         default:
                             $normalizedAssociatedId = $this->normalizeIdentifier($targetClass, $associatedId);
 
                             switch (true) {
                                 // We are negating the condition here. Other cases will assume it is valid!
                                 case $hints['fetchMode'][$class->name][$field] !== ClassMetadata::FETCH_EAGER:
+                                    if ($targetClass->subClasses) {
+                                        if (! $targetClass->isInheritanceTypeNone() && isset($targetClass->discriminatorColumn)) {
+                                            $connection = $this->em->getConnection();
+
+                                            $discriminatorColumnName = $targetClass->discriminatorColumn['name'];
+                                            $selectClause            = $connection->quoteIdentifier($discriminatorColumnName);
+                                            $fromClause              = $connection->quoteIdentifier($this->em->getClassMetadata(
+                                                $targetClass->rootEntityName
+                                            )->getTableName());
+
+                                            $whereClauses = [];
+                                            $whereValues  = [];
+                                            foreach ($targetClass->getIdentifierColumnNames() as $pkName) {
+                                                $whereClauses[] = $connection->quoteIdentifier($pkName) . ' = ?';
+                                                $whereValues[]  = $associatedId[$targetClass->fieldNames[$pkName]];
+                                            }
+
+                                            $whereClause = implode(' AND ', $whereClauses);
+
+                                            $query = 'SELECT ' . $selectClause . ' FROM ' . $fromClause . ' WHERE ' . $whereClause;
+                                            $stmt  = $this->em->getConnection()->prepare($query);
+
+                                            $index = 1;
+                                            foreach ($whereValues as $whereValue) {
+                                                $stmt->bindValue($index, $whereValue);
+                                                $index++;
+                                            }
+
+                                            $result     = $stmt->executeQuery();
+                                            $fieldValue = $result->fetchFirstColumn()[0];
+
+                                            $targetSubClass = $targetClass->discriminatorMap[$fieldValue];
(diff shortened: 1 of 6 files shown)


mpdude:
At first, I need to understand the problem.

Obviously, the documentation (https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/inheritance-mapping.html) uses almost the same wording in two sections, so maybe in general we can say:

> When the target entity of a to-one association is part of an entity inheritance hierarchy, it is preferable for performance reasons that it be a leaf entity. Otherwise Doctrine cannot create proxy instances of this entity and will always load the entity eagerly.

So where does this limitation come from or why is it mentioned in the docs?

My guess is that when we don't know the exact type of entity that will be found at the other end of the association, we cannot create a proxy matching the class. 

Examle: We have `BaseType` with `SubType1` and `SubType2` as entity subclasses. An association has `targetEntity=BaseType`. A proxy (or lazy ghost as well?) cannot be of that type or could at least not change its type at runtime later on. So, we have to load the "real" entity immediately.

When the association has `targetEntity=SubType1` (for example), we do for sure know what to expect, no other subclasses possible. → A proxy can be used.

Is that correct so far?

mpdude:
I appreciate that you're trying to fix a performance issue in this case! I'd be glad to see this resolved, since I have use cases myself that would benefit from the improvement. I must also say that I believe you have picked a rather complicated problem.

With a brief look at the UnitOfWork and it seems to me that it does not currently deal with database queries in any way. It seems to be concerned with connection-level properties and transaction control at best.

So, from my point of view it seems the change you are suggesting here probably works for your particular use case. But it does not seem to fit the general architecture/design of the ORM internal components.

Maybe you can find the place where the queries are constructed that provide the data that will then be passed to `\Doctrine\ORM\UnitOfWork::createEntity`. Would it be possible to take care of the associations in that place and make the query include the discriminator value for referred-to entities, so that this information is already available (and does also not take one extra query per row) in `createEntity`?

SanderBoelaars:
> Maybe you can find the place where the queries are constructed that provide the data that will then be passed to \Doctrine\ORM\UnitOfWork::createEntity. 

We found that the queries which supply the data for the createEntity function in the UnitOfWork are constructed in BasicEntityPersister->getSelectSQL(). 

> Would it be possible to take care of the associations in that place and make the query include the discriminator value for referred-to entities, so that this information is already available (and does also not take one extra query per row) in createEntity?"

This would be possible by adding the discriminator column name to the query and adding a JOIN with the associated entity.
However, we do not know how to integrate our functionality with the already existing code at BasicEntityPersister->getSelectColumnsSQL() because when not using FETCH_EAGER the foreach continues before any JOIN happens, this means we cannot add a new JOIN and so we cannot select our discriminator column name in the target entity's table.

beberlei:
Fetching the subclass via an extra query is a clever solution that trades an extra query for not having to load the entity in its whole. Good idea!

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2893,19 +2893,52 @@ public function createEntity($className, array $data, &$hints = [])
...
-
                         default:
                             $normalizedAssociatedId = $this->normalizeIdentifier($targetClass, $associatedId);
 
                             switch (true) {
                                 // We are negating the condition here. Other cases will assume it is valid!
                                 case $hints['fetchMode'][$class->name][$field] !== ClassMetadata::FETCH_EAGER:
+                                    if ($targetClass->subClasses) {
+                                        if (! $targetClass->isInheritanceTypeNone() && isset($targetClass->discriminatorColumn)) {
+                                            $connection = $this->em->getConnection();
```
beberlei, line 2904: Could you extract this block into a new method? It feels safer and maintains more readability. Given we execute a query here inlining this block has no performance value anyways.

PR #10539, opened 2023-02-24
More precisely document the performance impact of to-one associations towards inheritance hierarchies

mpdude:
This puts the remarks that apply to both JTI/STI in a common section at the beginning, and tries to explain a bit more in detail why it is that way.

It was sparked off by the discussion in #10538.

### DIFF
--- a/docs/en/reference/inheritance-mapping.rst
+++ b/docs/en/reference/inheritance-mapping.rst
@@ -153,6 +153,20 @@ lowercase short name of each class as key.
     to see if they are part of it. The resulting map, however, can be kept
     in the metadata cache.
 
+Performance impact on to-one associations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There is a general performance consideration when using entity inheritance:
+If the target-entity of a many-to-one or one-to-one association is part of
+an inheritance hierarchy, it is preferable for performance reasons that it
+be a leaf entity (ie. have no subclasses).
+
+Otherwise, the ORM is currently unable to tell beforehand which entity class
+will have to be used, and so no appropriate proxy instance can be created.
+That means the referred-to entities will *always* be loaded eagerly, which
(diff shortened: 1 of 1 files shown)


mpdude:
@SanderBoelaars does this currently take one extra database query for every single row that is hydrated?

mpdude:
@greg0ire A small documentation update, maybe something you could review?

docs/en/reference/inheritance-mapping.rst
```diff
@@ -153,6 +153,20 @@ lowercase short name of each class as key.
...
+There is a general performance consideration when using entity inheritance:
+If the target-entity of a many-to-one or one-to-one association is part of
+an inheritance hierarchy, it is preferable for performance reasons that it
+be a leaf entity (ie. have no subclasses).
+
+Otherwise, the ORM is currently unable to tell beforehand which entity class
+will have to be used, and so no appropriate proxy instance can be created.
+That means the referred-to entities will *always* be loaded eagerly, which
+might even include other entities as well (in the case of self-referencing
+associations).
```
greg0ire: > which might even include other entities as well (in the case of self-referencing
associations).



I don't understand what that last bit is about… "other entities" than what?

mpdude: When the referred-to entity that is loaded eagerly has other associations of the same kind, it may trigger another round of eager loads 🔁

greg0ire: By "of the same kind", you mean "self-referencing", right?

If yes, I'd go with "That means the referred-to entities will always be loaded eagerly, which might even propagate to relationships of these entities (in the case of self-referencing associations).

mpdude: :+1:, changed that.

greg0ire (approved):

greg0ire:
Thanks @mpdude !

PR #10540, opened 2023-02-24
Mark SqlWalker methods as not deprecated

stof:
phpstan treats implementations of deprecated methods of an interface as being deprecated themselves by default. However, SqlWalker does not intend to deprecate all those methods that are deprecated in TreeWalker, as they are moved down. Marking them as not deprecated will avoid reporting usages of deprecated APIs when implementing custom DQL functions for instance.

See https://github.com/phpstan/phpstan/discussions/7422 for the discussion about that kind of cases in the phpstan repository, which contains links justifying the current behavior of phpstan.
Without this `@not-deprecated` (or with it but without the dev version of phpstan as support for it is merged but not released yet), I get such error in my project when implementing custom DQL functions (which work with the SqlWalker API to generate the SQL):

```
 ------ --------------------------------------------------------------------------------------- 
  Line   src/Incenteev/WebBundle/Doctrine/Query/AnyUuid.php                                     
 ------ --------------------------------------------------------------------------------------- 
  32     Call to deprecated method walkInputParameter() of class Doctrine\ORM\Query\SqlWalker:  
         This method will be removed from the interface in 3.0.                                 
 ------ --------------------------------------------------------------------------------------- 
```

But while it will indeed be removed from the TreeWalker interface in 3.0, it won't be removed from SqlWalker.

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -258,6 +258,8 @@ class SqlWalker implements TreeWalker
      * @psalm-param QueryComponent $queryComponent
      *
      * @return void
+     *
+     * @not-deprecated
      */
     public function setQueryComponent($dqlAlias, array $queryComponent)
     {
@@ -276,6 +278,8 @@ class SqlWalker implements TreeWalker
      * @param AST\DeleteStatement|AST\UpdateStatement|AST\SelectStatement $AST
      *
      * @return Exec\AbstractSqlExecutor
+     *
+     * @not-deprecated
      */
     public function getExecutor($AST)
     {
@@ -628,6 +632,8 @@ class SqlWalker implements TreeWalker
      * @param string $identVariable
      *
      * @return string
+     *
+     * @not-deprecated
      */
     public function walkEntityIdentificationVariable($identVariable)
     {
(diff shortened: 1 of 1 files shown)


mpdude:
Would it make sense to have a dedicated sub-interface for the concept that SqlWalker implements, which seems to be more specific than being a tree walker?

If so, would that solve the issue as well?

stof:
As long as that new interface extends TreeWalker (which is needed for BC) which contains the deprecated methods, the issue of having a non-deprecated method overriding/implementing a deprecated method would still be there.

And introducing an interface for the SqlWalker would be a bad idea. If AST nodes were written against a SqlWalkerInterface instead of the SqlWalker for their conversion to SQL, adding new features in DQL would be a BC break any time it would require adding a new method in the SqlWalker to handle some new kind of AST node.
If you look at SqlWalker vs TreeWalker deprecated methods, you will see that SqlWalker has additional `walk*` methods (the ones added later that doctrine/orm 2.0)

derrabus:
Interesting, I didn't know about that annotation.

@stof Can you make PHPCS happy please?

---

> Would it make sense to have a dedicated sub-interface for the concept that SqlWalker implements, which seems to be more specific than being a tree walker?

No, we have no use for a `SqlWalker` interface.

mpdude:
@stof thank you for the explanations, makes sense

stof:
@derrabus this annotation is quite new (the solution was suggested by Ondrej a while ago but the implementation was done today)

stof:
@derrabus I fixed the order of phpdoc annotations to make phpcs happy.

derrabus (approved):

PR #8703, opened 2021-05-20
Change persistence ordering for OneToOne to allow making JoinColumns non-nullable

rvanlaak:
closes #7006
closes #6499
closes #5538
closes #4230

The issue as discussed earlier in #7006, #6499, #5538 and #4230 makes it harder to encapsulate behavior inside domain objects, when following some basic principles of DDD aggregate roots. In other words, the aggregate should not leak it's child nor have a dependency on a em/repo for the child to get persisted.

To be clear on the workaround, setting the `@JoinColumn` to nullable solves this, as Doctrine does take care of setting the `id` on the object at a later moment.

In response to discussing a fix for #6499 we borrowed the test as @Frikkle proposed in #6533 and slightly adjusted it to demonstrate the issue on persisting cascading relationships.

## Expected

The exception we should get on running the test, that would confirm that the persistence ordering is wrong:

> Exception : [Doctrine\ORM\ORMException] The identifier id is missing for a query of Doctrine\Tests\ORM\Functional\Ticket\DDC6499A

## Actual

There is pending discussion in https://github.com/doctrine/orm/pull/6533 on a proposed fix.

As the tests in this PR pass, that proves that the commit ordering works correctly.

PR #10547, opened 2023-02-27
Compute the commit order (inserts/deletes) on the entity level

### LINKED ISSUE #5538: Commit order issue
I have a very annoying issue regarding the commit order calculated by the UnitOfWork & CommitOrderCalculator components.

I have two entities, the important parts in this case looks like this:

``` php
/**
 * @ORM\Entity()
 */
class Hotel
{
    /**
     * @var HotelImage
     *
     * @ORM\ManyToOne(targetEntity="HotelImage", inversedBy="thumbnailHotels")
     * @ORM\JoinColumn(name="thumbnail_id", referencedColumnName="id", nullable=true, onDelete="SET NULL")
     */
    private $thumbnail;

    /**
     * @var ArrayCollection
     *
     * @ORM\OneToMany(targetEntity="HotelImage", mappedBy="hotel")
     */
    private $images;
}
```

``` php
/**
 * @ORM\Entity()
 */
class HotelImage
{
    /**
     * @var Hotel
     *
     * @Assert\NotBlank()
     * @ORM\ManyToOne(targetEntity="Hotel", inversedBy="images")
     * @ORM\JoinColumn(name="hotel_id", referencedColumnName="id", onDelete="CASCADE", nullable=false)
     */
    private $hotel;

    /**
     * @var ArrayCollection
     *
     * @ORM\OneToMany(targetEntity="Hotel", mappedBy="thumbnail")
     */
    private $thumbnailHotels;
}
```

The code snippet I have problem when looks like this:

``` php
$country = $countryRepo->findByName('Sweden');
$city = $cityRepo->findByCountryAndName($country, 'Stockholm');

$hotel = $hotelRepo->findByCityAndName($city, 'Radisson Blu Waterfront Stockholm');

$image = new HotelImage();
$image->setHotel($hotel);
$hotel->addImage($image);
$em->persist($image);

$hotel2 = new Hotel();
$hotel2->setCity($city);
$hotel2->setName('xxxxxxx');
$em->persist($hotel2);

$image = new HotelImage();
$image->setHotel($hotel2);
$hotel2->addImage($image);
$em->persist($image);

$em->flush();
```

Then the following exception will be thrown:

```
  [Doctrine\DBAL\Driver\PDOException]
  SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'hotel_id' cannot be null
```

When looking at the sql statements executed I see these in the logs:

```
[2015-12-08 21:13:57] doctrine.DEBUG: INSERT INTO hotel_images (file_name, mime_type, url, code, created, updated, supplier_id, hotel_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?) {"1":null,"2":null,"3":null,"4":null,"5":"2015-12-08 21:13:56","6":"2015-12-08 21:13:56","7":null,"8":628550}
[2015-12-08 21:13:57] doctrine.DEBUG: INSERT INTO hotel_images (file_name, mime_type, url, code, created, updated, supplier_id, hotel_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?) {"1":null,"2":null,"3":null,"4":null,"5":"2015-12-08 21:13:56","6":"2015-12-08 21:13:56","7":null,"8":null}
```

If I instead run the following snippet:

``` php
$country = $countryRepo->findByName('Sweden');
$city = $cityRepo->findByCountryAndName($country, 'Stockholm');

$hotel2 = new Hotel();
$hotel2->setCity($city);
$hotel2->setName('xxxxxxx');
$em->persist($hotel2);

$image = new HotelImage();
$image->setHotel($hotel2);
$hotel2->addImage($image);
$em->persist($image);

$em->flush();
```

```
[2015-12-08 21:18:00] doctrine.DEBUG: INSERT INTO hotels (type, latitude, longitude, rating, user_score, number_of_reviews, phone_number, fax, url, check_in_time, check_out_time, number_of_rooms, street, postal_code, certified, certification_rating, center_distance, name, created, updated, thumbnail_id, hotel_chain_id, address_id, coordinates_id, city_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) {"1":null,"2":null,"3":null,"4":null,"5":0,"6":0,"7":null,"8":null,"9":null,"10":null,"11":null,"12":null,"13":null,"14":null,"15":false,"16":null,"17":null,"18":"xxxxxxx","19":"2015-12-08 21:18:00","20":"2015-12-08 21:18:00","21":null,"22":null,"23":null,"24":null,"25":2535}
[2015-12-08 21:18:00] doctrine.DEBUG: INSERT INTO hotel_images (file_name, mime_type, url, code, created, updated, supplier_id, hotel_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?) {"1":null,"2":null,"3":null,"4":null,"5":"2015-12-08 21:18:00","6":"2015-12-08 21:18:00","7":null,"8":642815}
```

The error occurs becouse depending on what entity gets persisted first, the \Doctrine\ORM\UnitOfWork::getCommitOrder will build commit order differently because the two entities are dependent of each other, but since the thumbnail_id is nullable this should still work..

One solution to this would of course be a different approach to the thumbnails but I'd stil expect the above test cases not to throw errors and the commit order to be consistent.

Please suggest a way around this without adding extra flushes, or change the order I persist the objects. I have a number of use cases in the project where this is used where I need to persist the objects in the order as in the test cases and can't add flushes because of performance reasons.

Any help is appreciated, thanks.

### LINKED ISSUE #6499: Persisting entities with OneToOne and OneToMany/ManyToOne relationship between them.

### LINKED ISSUE #7006: Wrong commit order in some relation cases

### LINKED ISSUE #7866: Incorrect commit order when inserting multiple entities

### LINKED ISSUE #9192: [2.10.2] Ignore dependencies having onDelete="SET NULL" when computing commit order for removal operations
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | ?
| BC Break    | no

#### Summary

**TLDR**: Doctrine could find a valid delete commit order in some circular references cases if it were ignoring dependencies with `onDelete="SET NULL"`.

I have 3 entities with the following relationships:

- A contains a `OneToMany` to B with `cascade={"remove"}`.
- B contains a `OneToMany` to C with `cascade={"remove"}`.
- A also contains a `OneToOne` to C with no inverse side, `nullable=true` and `onDelete="SET NULL"`.

When I remove A via the entity manager, I get a foreign key violation error because the commit order Doctrine computes is B, C, A. Indeed B can't be deleted first since it is referenced by C.

I expected the order to be C, B, A instead. That wouldn't violate FK constraints. I eventually realized I have a circular reference (A references C and C references A via B) and that [Doctrine doesn't know how to deal with it](https://github.com/doctrine/orm/issues/5665#issuecomment-183994915).

The thing is, the circular reference is not really one when deleting, because of `onDelete="SET NULL"`. If Doctrine were to ignore such dependencies on delete, the commit order would end up correct.

Is there any chance this could be implemented?

Disclaimer: I'm not very familiar with Doctrine and DB engines in general, there could be downsides I'm not seeing.

#### More precisely what happens

[`UnitOfWork::getCommitOrder`](https://github.com/doctrine/orm/blob/9a74ae6280714aa4fe30960bce9c715535939fd3/lib/Doctrine/ORM/UnitOfWork.php#L1252) correctly lists the dependencies for each nodes:

- Node C with dep A,
- Node B with dep C,
- Node A with deps B and C (in that order, because B is the first entity encountered which is the owning side of A).

The order in which nodes are listed (C, B, A), comes from [`UnitOfWork::doRemove`](https://github.com/doctrine/orm/blob/9a74ae6280714aa4fe30960bce9c715535939fd3/lib/Doctrine/ORM/UnitOfWork.php#L1853)'s recursion putting the most deeply nested child first in the deletion list, and working back towards the parents.

Note in particular the dep A in C. It comes from the `OneToOne` relationship between A and C (the one with `onDelete="SET NULL"`).

A bit later in [`CommitOrderCalculator::visit`](https://github.com/doctrine/orm/blob/9a74ae6280714aa4fe30960bce9c715535939fd3/lib/Doctrine/ORM/Internal/CommitOrderCalculator.php#L134) (which computes the actual commit order from the dependency list):
- C is processed first.
- Then A (being a dependency of C).
- Then B (being the first dependency of A).
- C, the dependency of B, is in progress already so there's no new recursion on it.
- **B is added to the sorted list**.
- Being done with B we bubble back to A but its next dependency C is in progress already so we bubble back to C.
- **C is added to the sorted list** (it has no other dependency).
- Finally **A**, of which all dependencies have been visited already, **is added to the sorted list**.

Thus the order of `CommitOrderCalculator::sortedNodeList` ends up being: B, C, A. That order is [reversed](https://github.com/doctrine/orm/blob/9a74ae6280714aa4fe30960bce9c715535939fd3/lib/Doctrine/ORM/Internal/CommitOrderCalculator.php#L126) then again is used in  [reversed](https://github.com/doctrine/orm/blob/9a74ae6280714aa4fe30960bce9c715535939fd3/lib/Doctrine/ORM/UnitOfWork.php#L430) (i.e. original order).

When dependencies are first listed, Doctrine could ignore the dep A in C because that dep has `onDelete="SET NULL"`. The whole process would then compute a valid commit order C, B, A.

### LINKED ISSUE #10348: Order of referenced child deletions not correct when referencing to itself
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.14.0

#### Summary

The order in which the UoW wants to delete children is somehow missing a dependency when there is an association to the same class even if you specified `cascade: ['remove']`.
Given you have

- Parent -> has an association to children with `cascade: ['remove']`
- Child 1
- Child 2 -> has an association to Child 1 with `cascade: ['remove']`.

Now you delete the parent.
Doctrine tries to remove Child 1 first instead of Child 2 causing a foreign constraint violation: 

> Uncaught PDOException: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`test`.`child_entities`, CONSTRAINT `FK_DAEB194156A273CC` FOREIGN KEY (`origin_id`) REFERENCES `child_entities` (`id`))

#### How to reproduce

I've created a test setup because I'm not familiar with the internals of ORM at all. So here we go:

1. Clone https://github.com/Toflar/doctrine-orm-bug-reproducer
2. Run `composer install`
3. Adjust dummy mysql connection example in `bootstrap.php`.
4. Run `bin/doctrine orm:schema-tool:drop --force && bin/doctrine orm:schema-tool:create`
5. Run `php test.php`.

#### Expected behavior

It should not throw an exception. Doctrine should remove Child 2 first.

mpdude:
Fixes some long-standing issues that have to do with commit order computation and cyclic associations.

When performing inserts or deletes of entities, the ORM tries to find a sequence of operations so that all foreign key constraints can be satisfied. Basically, before inserting an entity, all other entities it refers to have to be inserted. And before deleting an entity, all other entities that refer to it have to be deleted, or database-level `SET NULL` has to be used.

The current implementation for this works at the _entity class_ level. It checks which entity class refers to which other class(es), and based on this information, computes an ordering of entity classes. Then, all entities of the first class will be processed, then those of the second class and so on. The same ordering is used for deletions.

In the PRs #10531 and #10532, two examples are given why this is not sufficient: There may be references between entities _of the same class_, so that we can only insert those in the database by looking at their associations _at the entity level_. Or, we may have references between entities of two different classes that require us to go back and forth – insert one entity of the first class, before we can insert one for the second class, before we can go back to the first class for a third entity.

Then, there is a shortcoming in the current `CommitOrderCalculator`: It does not detect cycles at all. This refers to situations where we cannot find a commit order because _every_ single entity to be processed requires another one to come first. Cycles can be broken at NULLable foreign keys, where a NULL value can temporarily be used and the value updated at a later stage. This is called "extra updates". Places where such extra updates can be used to save the day can be found by backtracking in the commit order computation, in case cycles are found.

Last, the commit order for INSERTs and DELETEs needs to be based on different criteria: INSERTs can make use of extra updates and require NULLable foreign key definitions for this to work. DELETEs, however, currently do not make use of such extra updates. For them, whether the database can use "SET NULL" on foreign key deletions is the important factor.

To make it possible to review the changes, they were broken down into a series of independent PRs against a temporary branch. This PR here does a final merge of the result into `2.16.x`.

The individual PRs are:

* [x] #10592 
* [x] #10651
* [x] #10689
* [x] #10566
* [x] #10732
* [x] #10749
* [x] #10750
* [x] #10531
* [x] #10532
* [x] #10751 
* [x] #10755

(Potentially) fixed issues/closed PRs:
* [X] fixes #5538, although no complete test was provided 
* [X] fixes #6499, tests were picked from #7180
* [X] closes #6533, tests were picked from #7180
* [X] closes #8703, it was using the tests from #6499 as well
* [X] fixes #7866, using the same tests as #8703
* [X] fixes #7006, tests added via #10750
* [X] closes #7180, picked tests from their PR and extra tests from discussion comments via #10751
* [X] fixes #9192, added a test via #10755
* [X] closes #10348, adding tests via #10732
* [X] closes #8349, their tests being included in #10592

_Not_ fixed issues/PRs:
* #5109 – this is about extra unique constraints and/or performing DELETE before INSERT, beyond what the ORM requires itself to make foreign keys work; similar in #5742, #6776
* #5665, #10548 – those are about scheduling extra updates before DELETE, to null out foreign keys and not require a database-level `SET NULL`

(Probably) related issues/PR with unclear state and/or missing reproducers:

* #9376
* #9377

Definitely too big for this PR:
* Performance: Persisters no longer being able to batch inserts
* Performance: Optimizing the DFS to find a minimal feedback arc set/minimize the number of scheduled extra updates


rvanlaak:
I am a bit rusty on the topic regarding the debugging of the commit order, but based on looking at the `CommitOrderCalculator` your suggested changes look extremely promising in solving the long standing commit order bug. 👏🏻

lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
```diff
@@ -118,47 +98,45 @@ public function sort()
      */
     private function visit(Vertex $vertex): void
     {
+        if ($vertex->state === VertexState::IN_PROGRESS) {
+            // This node is already on the current DFS stack. We've found a cycle!
+            throw new CycleDetectedException();
```
derrabus: An exception should have a message.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2571,9 +2547,16 @@ public function lock($entity, int $lockMode, $lockVersion = null): void
     /**
      * Gets the CommitOrderCalculator used by the UnitOfWork to order commits.
      *
+     * @deprecated use createCommitOrderCalculator() instead
```
derrabus: What's the reason for this deprecation? We usually don't introduce deprecations in bugfix releases and this one does not seem to be necessary.

derrabus (changes requested):
I think I'll need to review this with more time than I have right now. Awesome work. 🤯

lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
mpdude: Yes, will need to think about this a bit more. Probably the message should convey the exact cycle, ie which chain of entity associations is causing a foreign key loop

mpdude:
@derrabus Yes, this PR is much bigger than what I'd like to see to make reviews easy.

I did not find a good way to break it down into chunks that would make sense on their own and could be merged. Computing the order on the entity level with a broken CommitOrderCalculator does not work. Fixing the COC will break use cases that currently work as long as we don't consider individual entities.

What I could offer is that I try to break this down into pieces and guide you and everyone interested through reviews in dedicated PRs, with smaller scope each.

We'd need to merge all those PRs into a dedicated branch when they are approved. Tests for individual features probably can be working, but the full test suite won't pass until we're done.

In the end, the temporary branch could be merged and should roughly be equivalent to this PR here.

WDYT – would that help, or do you have other suggestions?

derrabus:
I could also offer to setup a Google Meet call where you can guide me and maybe @greg0ire and @beberlei through the changes. That might be more effective.

jhony1104 (dismissed):
This is no complete review, but since @mpdude [asked](https://github.com/doctrine/orm/pull/8349#issuecomment-1446836044) if I could look at it:

Only looking at the CommitOrderCalculator, unlike #8349, #7866, #8067, #7180, #9377 and others this PR correctly deals with circles using backtracing and if needed ignores nullable relations. Unlike some of the listet PRs it does not only work reliably for (simple) testcases, but I have also testet it in our product with ~130 tables (where for example #8349 didn't work).

Merging this would allow us to start using unmodified doctrine.

lib/Doctrine/ORM/Internal/CommitOrderCalculator.php
mpdude: Fixed

lib/Doctrine/ORM/UnitOfWork.php
mpdude: Reverted

mpdude:
Rebased onto 2.15.x – given the nature of these changes, due to potential collateral damage this is nothing we should roll out in a bugfix release.

My plan is to break this down into a series of individual PRs to make reviews easier, as outlined in https://github.com/doctrine/orm/pull/10547#issuecomment-1449702387.

So, could somebody from the @doctrine group please create a dedicated branch (say, `entity-level-commit-order`?) off `2.15.x` for me? 

That way, all the individual PRs can happen here in doctrine/orm, but we collect the results on that branch until and can eventually merge the branch with results as a whole.

lib/Doctrine/ORM/Internal/TopologicalSort.php
```diff
@@ -0,0 +1,146 @@
...
+use function array_keys;
+use function array_reverse;
+use function array_unshift;
+use function spl_object_id;
+
+/**
+ * TopologicalSort implements topological sorting, which is an ordering
+ * algorithm for directed graphs (DG) using a depth-first searching (DFS)
+ * to traverse the graph built in memory.
+ * This algorithm have a linear running time based on nodes (V) and edges
```
vctls:
```suggestion
 * This algorithm has a linear running time based on nodes (V) and edges
```

greg0ire:
> So, could somebody from the https://github.com/doctrine group please create a dedicated branch (say, entity-level-commit-order?) off 2.15.x for me?

Done :)

mpdude:
Here we go: #10592 adds the new topological sort impl. Looking forward to reviews!

mpdude:
@greg0ire could you update the CI config so that it will run on the new branch as well?

greg0ire:
Done, I think.

mpdude:
> Done, I think.

Could you have a look at #10593 please? (No checks running in #10592.)

greg0ire (commented):
There are conflicts

mpdude:
Force-pushed this branch to be the same as `entity-level-commit-order` where all the intermediate work took place. Also, merged 2.16.x and resolved conflicts 🤞🏻

greg0ire:
y u no use git hooks

.github/workflows/coding-standards.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - "*.x"
+      - entity-level-commit-order
```
greg0ire: Isn't the plan to merge this PR and then remove that branch?

mpdude: Yes. I had those things in `.github/workflows` in mind, but thought we cannot safely remove them unless this PR is merged?

greg0ire: Yes exactly… so now you should push a new commit to remove those, and the PR should be ready to merge, right?

mpdude: Did that. Surprised to see that tests etc. still being run for that new commit.

mpdude: D'oh. I get it. They run for the PR branch, no longer need to run for the other one.

mpdude:
@derrabus Maybe you need to approve before we can merge this :-)

greg0ire (dismissed):

lib/Doctrine/ORM/Internal/TopologicalSort.php
```diff
@@ -0,0 +1,165 @@
...
+     * boolean value indicates whether the edge is optional or not.
+     *
+     * @var array<int, array<int, bool>>
+     */
+    private $edges = [];
+
+    /**
+     * Builds up the result during the DFS.
+     *
+     * @psalm-var list<object>
```
derrabus: I think, we can un-prefix simple generics like `list<>` and `array<>` nowadays.

```suggestion
     * @var list<object>
```

lib/Doctrine/ORM/Internal/TopologicalSort.php
```diff
@@ -0,0 +1,165 @@
...
+        $this->edges[$fromId][$toId] = $optional;
+    }
+
+    /**
+     * Returns a topological sort of all nodes. When we have an edge A->B between two nodes
+     * A and B, then A will be listed before B in the result.
+     *
+     * @psalm-return list<object>
+     */
+    public function sort()
```
derrabus:
```suggestion
    public function sort(): array
```

derrabus (dismissed):
LGTM

lib/Doctrine/ORM/Internal/TopologicalSort.php
mpdude: Updated, and found one more case

lib/Doctrine/ORM/Internal/TopologicalSort.php
mpdude: done

mpdude:
My push dismissed your reviews.

mpdude:
Disclosure: I've added a commit bb21865cba0475d74ada1f487e4c2fef59256ee4 that did not go through the review process on the dedicated `entity-level-commit-order` branch. It deprecates the old classes around `CommitOrderCalculator`.

greg0ire:
The extra commit looks correct to me.

lib/Doctrine/ORM/Internal/CommitOrder/Edge.php
```diff
@@ -27,6 +29,13 @@ final class Edge
 
     public function __construct(string $from, string $to, int $weight)
     {
+        Deprecation::triggerIfCalledFromOutside(
+            'doctrine/orm',
+            'https://github.com/doctrine/orm/pull/10547',
+            'The %s class is deprecated and will be removed in ORM 3.0',
+            self::class
+        );
```
derrabus, line 40: If the class is deprecated, it should be annotated with `@deprecated`. That being said, it's already `@internal`, so we might as well remove it right away?

mpdude: I'd love to see it being removed right away, but the `CommitOrderCalculator` depends on it and was itself not marked as `@internal` (it only lived in a `...\Internal\...` namespace).

So, I'll keep the classes for now but added `@deprecated` annotations.

sylfabre:
Hello @mpdude 

Thank you for addressing this long-standing issue 👍 

I tried your branch as per your comments on https://github.com/doctrine/orm/pull/10714 & https://github.com/doctrine/orm/pull/10716
Unfortunately, I get an exception in `Statement.php` with this message: 
```
 [PDOException (42000)] SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '?, ?, ?, ?, ?)' at line 1
```

This comes from the insert statement `INSERT INTO payment_settings (updated_at, retribution_method, id, created_at, organization_id) VALUES (?, ?, ?, ?, ?)` which is done with an empty params array 😬 

The same code works using `doctrine/orm` v2.15.3

So I cannot confirm if your PR solves my issues https://github.com/doctrine/orm/issues/10713 & https://github.com/doctrine/orm/issues/10715

Let me know how I may help you.
Our database contains over 250 tables so it can be a strong real-world test!

mpdude:
@sylfabre Can you provide a test (as a PR against the `entity-level-commit-order` branch) that demonstrates the issue? 

The test outlined in #10713 works directly with the `CommitOrderCalculator`, a class replaced by this PR here.

sylfabre:
Hello @mpdude 

I'm sorry my last message was a false positive 👌 

After checking my codebase, I discovered a postPersist listener triggering a flush operation. It leads to trying to insert two times the same entity: the first time works, the second fails because the entity changeset is now empty so the insert query is missing a lot of params. I guess flushing within a postPersist listener is a bad practice and we are just lucky it works today!

Thank you again for your work

mpdude:
Added f01d107edc5bebbd3d628a7c6cb1934339aea435 to address a new edge case supported in 2.16 since #10735.

greg0ire (approved):

greg0ire:
Congrats on this huge piece of work, and welcome back!

rvanlaak:
Congrats! This is an awesome step in the reliability of the commit order.

mpdude:
Thanks to everyone involved and supporting this over the last months with many reviews in a very constructive and responsive manner – especially @greg0ire, @SenseException and @derrabus!

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -437,32 +434,37 @@ public function commit($entity = null)
...
+                // in the database.
+                $this->executeInserts();
             }
 
             if ($this->entityUpdates) {
-                foreach ($commitOrder as $class) {
-                    $this->executeUpdates($class);
-                }
+                // Updates do not need to follow a particular order
+                $this->executeUpdates();
```
polaeskandar, line 447: https://github.com/doctrine/orm/issues/10864

PR #10551, opened 2023-03-01
Remove `Doctrine\ORM\Proxy\Proxy` interface

derrabus:


SenseException:
Seems like the psalm baseline needs to be adapted too.

SenseException (approved):

Issue #10552, opened 2023-03-01
Resolve Target Entity Listener not longer working like expected


alexander-schranz:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 2.15.x-dev 4fad7a1 (stable 2.14.1 and 2.14@dev works)

#### Summary

Our tests against `@dev` versions are currently failing with the dev version of doctrine/orm and it ssems like the EntityTargetResolver is not longer working like expected in current `@dev` version.

We have some special handling for resolved targets and they are not configured via the doctrine bundle instead we are adding the resolved targets inside an own compilerpass not sure if that could cause an issue now: https://github.com/sulu/sulu/blob/57bf8c4dcc59361033ef6d7acc5898b850f15b5c/src/Sulu/Bundle/PersistenceBundle/DependencyInjection/Compiler/ResolveTargetEntitiesPass.php#L50-L51. There is some special handling behind which not only @sulu but also @sylius use to make things easier overwritable.

The last commit where the CI did run was `979b3dcb8df6815caef022d78262bcfa232238fc` so it must be some of this changes:

https://github.com/doctrine/orm/compare/979b3dcb8df6815caef022d78262bcfa232238fc...2.15.x

The changes from @mpdude looks releated to the EntityResolveListener: https://github.com/doctrine/orm/pull/10473

#### Current behavior

```
1) Sulu\Bundle\ActivityBundle\Tests\Functional\Infrastructure\Doctrine\Repository\ActivityRepositoryTest::testCreateForDomainEvent
Doctrine\ORM\Mapping\MappingException: The target entity class Sulu\Bundle\SecurityBundle\Entity\User specified for Sulu\Bundle\TrashBundle\Domain\Model\TrashItem::$user is not an entity class.
```

<details>

<summary>Stack Trace:</summary>

```bash
/Users/alexanderschranz/Documents/Projects/sulu-develop.localhost/vendor/sulu/sulu/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/MappingException.php:856
/Users/alexanderschranz/Documents/Projects/sulu-develop.localhost/vendor/sulu/sulu/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:312
[...]

PR #10554, opened 2023-03-01
Make "targetEntity must not be a mapped superclass" a lazy check

mpdude:
#10473 relaxed the runtime check that mapped superclasses must not use one-to-many associations. It turned the check the other way round, so that no mapped superclass may be used as the `targetEntity` of an association.

To find out whether a class is a mapped superclass, the `\Doctrine\ORM\Mapping\ClassMetadataFactory::peekIfIsMappedSuperclass` method was used, which is directly using the mapping driver.

#10552 reports a case where mapping configuration is not collected through the driver alone, but also a metadata listener is contributing relevant information. Updating metadata with a listener is currently not supported by the check in that method.

Also, https://github.com/doctrine/orm/pull/10473#issuecomment-1450048660 has a mapping configuration that cannot be loaded in the improvised way that `\Doctrine\ORM\Mapping\ClassMetadataFactory::peekIfIsMappedSuperclass` employs.

As I explained in https://github.com/doctrine/orm/issues/10552#issuecomment-1450794332, we cannot use the full ClassMetadataFactory itself to obtain information about the `targetEntity` while we're performing runtime validation for another class currently being loaded.

So, I don't see any other way than to make this check a non-runtime (offline?) check in the `SchemaValidator`. There, we can first completely load metadata for all classes and then inspect it.

Many fixes to the test models that were made in #10473 can be reverted, since these files are no longer scrutinized at runtime, and we probably don't care about the extra validation checks that much.


alexander-schranz (approved):
I did not have a look at the code as I'm not unfamiliar with internals here but this PR fixes our problems inside @sulu: https://github.com/sulu/sulu/pull/7026

Thank you @mpdude 👍

BreyndotEchse (approved):
MappingException gone and all tests passed!
ty @mpdude

mpdude:
Doctrine core team reviewers:

Please confirm that using a `SchemaValidator` based check only instead of a runtime check is a valid approach. That's what the schema validator is made for, right? Not all problems can easily be spotted at runtime.

mpdude:
@derrabus, @greg0ire or @SenseException – can we try to get this merged in the near future? I have the impression it would unblock people who are kind enough to try 2.15-dev in their CI pipelines.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @mpdude !

PR #10557, opened 2023-03-02
Fix regression with an edge case for the `CMF::peekIfIsMappedSuperclass()` method

mpdude:
[This comment](https://github.com/doctrine/orm/pull/10473#issuecomment-1450048660) hints to a case where the `ClassMetadataFactory::peekIfIsMappedSuperclass()` method introduced in #10411 causes a failure.

`CMF::peekIfIsMappedSuperclass()` has to perform improvised metadata loading in a situation where the CMF is currently loading a class. So, we cannot use the full/real `ClassMetadataFactory` mechanisms, since it would require a re-entry for a subclass of the current class, causing an infinite loop (loads parent classes first, and that's what we're currently doing).

The problem is that the improvised call to `$driver->loadMetadataForClass()` cannot provide a pre-filled `ClassMetadata` instance populated with all parent class fields and associations. But, when attribute or association overrides are used, a check is made to see if the overridden field/association actually exists, and this information is missing in that situation.

This PR suggests to override the methods to get around this. In fact, we do not care about all these details, we only want to ask the driver if the class is a mapped superclass or not.

A much better fix would be to have a dedicated method on the driver to ask it just that particular question (also better performance-wise). But I do not see how we could get that done in a BC way – ideas? 💡

A few things that need to come together to make the bug surface:

* Load an entity declaring an inheritance tree
* There must be a mapped superclass in the inheritance tree to provide the field that shall be overriden
* An entity class must inherit from the mapped superclass and override the field
* That entity class must be an abstract, intermediate class not be declared in the discriminator map so we can "discover" it
* The overriden property must be private so the mapping drivers (using reflection) do not see it when looking at the overriding entity class.

### DIFF
--- a/lib/Doctrine/ORM/Internal/OverrideDisregardingClassMetadata.php
+++ b/lib/Doctrine/ORM/Internal/OverrideDisregardingClassMetadata.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Internal;
+
+use Doctrine\ORM\Mapping\ClassMetadata;
+
+/**
+ * @internal
+ *
+ * @psalm-internal Doctrine\ORM\Mapping
+ * @template       T of object
+ * @template-extends ClassMetadata<T>
+ */
+class OverrideDisregardingClassMetadata extends ClassMetadata
+{
+    /**
+     * @param string $fieldName
+     * @psalm-param array<string, mixed> $overrideMapping
+     */
+    public function setAttributeOverride($fieldName, array $overrideMapping): void
+    {
+        // override to ignore consistency checks not relevant in this use case
+    }
+
+    /**
+     * @param string $fieldName
+     * @psalm-param array<string, mixed> $overrideMapping
+     */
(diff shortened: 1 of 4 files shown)


mpdude:
@greg0ire I am afraid I don't know how to fix the Psalm issues. Can you help?

mpdude:
@derrabus Can you tell me how to proceed with Psalm errors here?

derrabus:
Your class extends `ClassMetadata` which is generic. You need to add a `@extends` annotation that specifies the template parameters for `ClassMetadata`.

mpdude:
Thank you, will 👀

mpdude:
~@derrabus The remaining Psalm error seems to be unrelated to my changes...?~

Ah, I get it!

I am consuming one of the baseline entries with the change I made, so we're out of baseline entries at the end of the file and the error pops up there.

greg0ire:
> But I do not see how we could get that done in a BC way – ideas?

Add the method to all existing drivers, and deprecate not adding it to a driver (and fallback to the old behavior when that's the case).

You can also add an `@method` annotation to the interface of drivers so that static analysis lets people know that they should implement the new method.

PR #10559, opened 2023-03-05
Convert PHPUnit annotations to attributes

chr-hertel:
Rector did the work and I cleaned up. Not sure if you're open for this kind of PRs, but was my sunday meditation anyways :)

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/SQLFilterTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/SQLFilterTest.php
@@ -27,6 +27,7 @@ use Doctrine\Tests\Models\Company\CompanyOrganization;
 use Doctrine\Tests\Models\Company\CompanyPerson;
 use Doctrine\Tests\OrmFunctionalTestCase;
 use InvalidArgumentException;
+use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\MockObject\MockObject;
 use ReflectionMethod;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
@@ -38,9 +39,8 @@ use function serialize;
 
 /**
  * Tests SQLFilter functionality.
- *
- * @group non-cacheable
  */
+#[Group('non-cacheable')]
 class SQLFilterTest extends OrmFunctionalTestCase
 {
     private int $userId;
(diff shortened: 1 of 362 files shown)


greg0ire:
By all means, we're open to such PRs. Can you take a look at phpcs?

derrabus:
Looks good so far. One thing: Doesn't PHPUnit 10 have an attribute equivalent for `@requires` or did Rector just not catch those?

> Can you take a look at phpcs?

PHPCS fails because a test has been commented out and because the following test does not have a PHPDoc block anymore, it assumes that the commented-put test is a badly formatted PHPDoc block.

Here's my take at fixing the issue on 2.14.x by skipping the test instead of commenting it out: #10560. Alternatively, we could just delete that test.

chr-hertel:
> Doesn't PHPUnit 10 have an attribute equivalent for @requires or did Rector just not catch those?

It does, Rector missed it or wasn't able. I will have another look 👍

derrabus:
The PHPCS failure should be gone after a rebase.

greg0ire:
> The PHPCS failure should be gone after a rebase.

You mean one failure out of three I think. @chr-hertel please do check it works before force-pushing.

tests/Doctrine/Tests/ORM/Functional/SQLFilterTest.php
```diff
@@ -25,8 +25,10 @@
 use Doctrine\Tests\Models\Company\CompanyManager;
 use Doctrine\Tests\Models\Company\CompanyOrganization;
 use Doctrine\Tests\Models\Company\CompanyPerson;
+use Doctrine\Tests\ORM\Functional\Ticket\MyEntity;
```
greg0ire: This needs to be reverted

greg0ire:
This should help: https://github.com/doctrine/orm/pull/10563

derrabus:
Ah, I missed the other test. Thanks, @greg0ire.

derrabus:
Browsing through all of the `RequiresPhp` attributes, I realized that they could actually all be removed because we don't run the test suite on PHP < 8.1 on this branch. I did remove a lot of those `@requires` annotations previously, but merge-ups keep reintroducing them. 😓 

But we can remove those in a follow-up, no need to block a merge because of that.

Great work. 👍🏻

greg0ire (approved):

derrabus (approved):

Issue #5665, opened 2016-02-14
Cascade deletion complicate-referenced entities does not work [ForeignKeyConstraintViolationException]


mtal:
```
[Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException]             
  An exception occurred while executing 'DELETE FROM subitem WHERE id = ?' wi  
  th params [4]:                                                               
  SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or upda  
  te a parent row: a foreign key constraint fails (`shop`.`item`, CONSTRAINT   
  `FK_1F1B251EB3AE4213` FOREIGN KEY (`featured_item_id`) REFERENCES `subitem`  
   (`id`))                                                                     
```

is caused by

```PHP
        $item = new Item();
        $sub1 = new Subitem();
        $sub2 = new Subitem();
        $item->addItem($sub1);
        $item->addItem($sub2);
        $item->setFeaturedItem($sub2);
        $em->persist($item);
        $em->flush();
        $em->remove($item);
        $em->flush();
```

Entity mappings:

```PHP
use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity
 */
class Item
{

    /**
     * @var integer
     *
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    protected $id;

    /**
     * @var ArrayCollection
     *
     * @ORM\OneToMany(targetEntity="Subitem", mappedBy="item", cascade={"all"}, orphanRemoval=true)
     */
    protected $items;

    /**
     * @var Subitem
     *
     * @ORM\ManyToOne(targetEntity="Subitem")
     */
    protected $featuredItem;

    public function __construct()
    {
        $this->items = new ArrayCollection();
    }

    /**
     * @return int
     */
    public function getId()
    {
        return $this->id;
    }

    /**
     * @return ArrayCollection
     */
    public function getItems()
    {
        return $this->items;
    }

    public function addItem(Subitem $item)
    {
        $this->items[] = $item;
        $item->setItem($this);
    }

    public function removeItem(Subitem $item  = null) {
        if ($this->featuredItem === $item) {
            $this->featuredItem = null;
        }
        $this->items->removeElement($item);
    }

    /**
     * @return Subitem
     */
    public function getFeaturedItem()
    {
        return $this->featuredItem;
    }

    /**
[...]

PR #10566, opened 2023-03-06
Commit order for removals has to consider `SET NULL`, not `nullable`

mpdude:
When computing the commit order for entity removals, we have to look out for `@ORM\JoinColumn(onDelete="SET NULL")` to find places where cyclic associations can be broken.

This is part of the efforts in #10547 to solve various commit-order related problems.

#### Background

The UoW computes a "commit order" to find the sequence in which tables shall be processed when inserting entities into the database or performing delete operations.

For the insert case, the ORM is able to schedule _extra updates_ that will be performed after all entities have been inserted. Associations which are configured as `@ORM\JoinColumn(nullable=true, ...)` can be left as `NULL` in the database when performing the initial `INSERT` statements, and will be updated once all new entities have been written to the database. This can be used to break cyclic associations between entity instances.

For removals, the ORM does not currently implement up-front `UPDATE` statements to `NULL` out associations before `DELETE` statements are executed. That means when associations form a cycle, users have to configure `@ORM\JoinColumn(onDelete="SET NULL", ...)` on one of the associations involved. This transfers responsibility to the DBMS to break the cycle at that place. 

_But_, we still have to perform the delete statements in an order that makes this happen early enough. This may be a _different_ order than the one required for the insert case. We can find it _only_ by looking at the `onDelete` behaviour. We must ignore the `nullable` property, which is irrelevant, since we do not even try to `NULL` anything.

#### Example

Assume three entity classes `A`, `B`, `C`. There are unidirectional one-to-one associations `A -> B`, `B -> C`, `C -> A`. All those associations are `nullable= true`. 

Three entities `$a`, `$b`, `$c` are created from these respective classes and associations are set up.

All operations `cascade` at the ORM level. So we can test what happens when we start the operations at the three individual entities, but in the end, they will always involve all three of them.

_Any_ insert order will work, so the improvements necessary to solve #10531 or #10532 are not needed here. Since all associations are between different tables, the current table-level computation is good enough.

For the removal case, only the `A -> B` association has `onDelete="SET NULL"`. So, the only possible execution order is `$b`, `$c`, `$a`. We have to find that regardless of where we start the cascade operation.

The DBMS will set the `A -> B` association on `$a` to `NULL` when we remove `$b`. We can then remove `$c` since it is no longer being referred to, then `$a`.

#### Related cases

These cases ask for the ORM to perform the extra update before the delete by itself, without DBMS-level support: 
* #5665
* #10548

#### Extra bonus

This is what the DALL·E AI thinks it looks like when the UnitOfWork is sequencing entity deletions with cascade.

DALL·E 2023-05-09 13 09 58

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -28,6 +28,7 @@ use Doctrine\ORM\Exception\UnexpectedAssociationValue;
 use Doctrine\ORM\Id\AssignedGenerator;
 use Doctrine\ORM\Internal\CommitOrderCalculator;
 use Doctrine\ORM\Internal\HydrationCompleteHandler;
+use Doctrine\ORM\Internal\TopologicalSort;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\MappingException;
 use Doctrine\ORM\Mapping\Reflection\ReflectionPropertiesGetter;
@@ -57,7 +58,6 @@ use function array_key_exists;
 use function array_map;
 use function array_merge;
 use function array_pop;
-use function array_reverse;
 use function array_sum;
 use function array_values;
 use function assert;
@@ -74,6 +74,7 @@ use function method_exists;
 use function reset;
 use function spl_object_id;
 use function sprintf;
+use function strtolower;
 
 /**
  * The UnitOfWork is responsible for tracking changes to objects during an
@@ -408,9 +409,6 @@ class UnitOfWork implements PropertyChangedListener
 
         $this->dispatchOnFlushEvent();
 
-        // Now we need a commit order to maintain referential integrity
-        $commitOrder = $this->getCommitOrder();
-
         $conn = $this->em->getConnection();
         $conn->beginTransaction();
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1287,26 +1282,58 @@ private function computeInsertExecutionOrder(array $commitOrder): array
...
+                assert(isset($assoc['joinColumns']));
+                $joinColumns = reset($assoc['joinColumns']);
+                if (isset($joinColumns['onDelete'])) {
+                    $onDeleteOption = strtolower($joinColumns['onDelete']);
+                    if ('cascade' === $onDeleteOption || 'set null' === $onDeleteOption) {
+                        continue;
+                    }
+                }
+
+                $targetEntity = $class->getFieldValue($entity, $assoc['fieldName']);
```
mpdude, line 1321: Discussed with ppl on Slack that we need not worry about objects being proxies here

mpdude:
@greg0ire @SenseException and @derrabus  what do you think about this? 

We need to agree on this one to make progress for the #10547 epic.

greg0ire (approved):

SenseException (approved):
Seems okay, but shouldn't the tests explicitly check that the entities were all 3 deleted?

mpdude:
Thank you @SenseException! I’ll amend the tests to cover that as well.

mpdude:
Tests now contain assertions that insertions/deletions did in fact happen.

mpdude:
@SenseException Can we (could you?) merge this, since I addressed your remarks?

#10689 probably will have merge conflicts afterwards, so I could address those.

SenseException:
Yeah, seems like no other review will be done.

mpdude:
Thank you

PR #10567, opened 2023-03-06
Forbid PHPUnit annotations

derrabus:
Following up on #10559, this PR introduces a PHPCS rule that will fail on PHPUnit annotations. This should serve as a constant reminder to switch them to attributes when merging up tests.

I've tested the rule locally by running PHPCS on the commits before and after @chr-hertel's changes.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -116,6 +116,36 @@
         <exclude-pattern>lib/Doctrine/ORM/Cache/DefaultQueryCache.php</exclude-pattern>
     </rule>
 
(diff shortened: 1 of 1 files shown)


greg0ire (approved):
I don't know of a way to avoid this repetition, when exactly should we revert this?

derrabus:
> I don't know of a way to avoid this repetition

Neither do I. 😕 

> when exactly should we revert this?

Maybe we can move this whole block upstream once all of our projects use PHPUnit 10?

PR #10572, opened 2023-03-09
docs: Update Logger removal in Batch Processing documentation

cicoub13:
Current documented way is using Doctrine DBAL2.
Since DBAL3, logging is done with Middlewares, so we updated the recommendation in documentation.

It's low level criticality

### DIFF
--- a/docs/en/reference/batch-processing.rst
+++ b/docs/en/reference/batch-processing.rst
@@ -19,11 +19,13 @@ especially what the strategies presented here provide help with.
 .. note::
 
     Having an SQL logger enabled when processing batches can have a serious impact on performance and resource usage.
-    To avoid that you should disable it in the DBAL configuration:
+    To avoid that you should remove the corresponding middleware.
+    To remove all middlewares, you can use this line:
 .. code-block:: php
 
     <?php
-    $em->getConnection()->getConfiguration()->setSQLLogger(null);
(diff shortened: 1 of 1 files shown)


docs/en/reference/batch-processing.rst
```diff
@@ -23,7 +23,7 @@ especially what the strategies presented here provide help with.
 .. code-block:: php
```
greg0ire, line 24: We should probably update this paragraph to let the user know that some middlewares can be memory-consuming, that they should consider keeping or removing each middleware, and that if they don't want to think too much they can just remove them all.

docs/en/reference/batch-processing.rst
```diff
@@ -23,7 +23,7 @@ especially what the strategies presented here provide help with.
...
 
     <?php
-    $em->getConnection()->getConfiguration()->setSQLLogger(null);
+    $em->getConnection()->getConfiguration()->setMiddlewares([]);
```
greg0ire: Maybe it could be cool to show both ways?
```suggestion
    $em->getConnection()->getConfiguration()->setMiddlewares([]); // DBAL 3
    $em->getConnection()->getConfiguration()->setSQLLogger(null); //DBAL 2
```

greg0ire (approved):

SenseException (approved):

SenseException:
Thank you @cicoub13

PR #10581, opened 2023-03-15
Make "EntityChangeSet" available in PostUpdateEventArgs

willemverspyck:
Not a final PR yet, but why not make "EntityChangeSet" available in PostUpdateEventArgs?

I see a lot of Stack Overflow solutions were they temporary cache the Changeset in "PreUpdate", use it in "PostUpdate" and then remove it, but that not a steady solution.

With this PR you can do some magic after the entity has been updated (ok, it can still fail if the transaction fails) and check which fields has changed. For example:

```php
    public function postUpdate(ProjectInterface $project, PostUpdateEventArgs $event): void
    {
        if ($event->hasChangedField('dateStart')) {
            $this->projectService->sendEmailAboutNewStartDate($project);
        }
    }
```

Todo:

- Function "setNewValue" must only available in PreUpdateEventArgs, as you can't set the new value in PostUpdate.


ttk:
This is a useful feature.  I often require EntityChangeSet in the PostUpdateEvent.  There are workarounds, but having it supported natively just like the PreUpdateEvent would make the DX better.

The latest 3.2.x branch is still missing this feature.

PR #10582, opened 2023-03-15
Subclass lazy loading

SanderBoelaars:
This PR is a continuation of the closed PR https://github.com/doctrine/orm/pull/10538 (which cannot be reopened, my apologies, I am new to making Pull Requests)

I have extracted the functionality for retrieving the discriminator column value into a separate function, like suggested by @beberlei.

I also noticed a grammatical error while reading https://github.com/doctrine/orm/blob/2.14.x/CONTRIBUTING.md and included a fix, I hope it is okay to do this in the same PR.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -2710,6 +2710,14 @@ class UnitOfWork implements PropertyChangedListener
      */
     public function createEntity($className, array $data, &$hints = [])
     {
+        $extra = $data['extra'] ?? null;
+
+        if (count($data) === 2) {
+            if (array_key_exists('data', $data) && array_key_exists('extra', $data)) {
+                $data = $data['data'];
+            }
+        }
+
         $class = $this->em->getClassMetadata($className);
 
         $id     = $this->identifierFlattener->flattenIdentifier($class, $data);
@@ -2901,15 +2909,24 @@ class UnitOfWork implements PropertyChangedListener
                                 case $hints['fetchMode'][$class->name][$field] !== ClassMetadata::FETCH_EAGER:
                                     if ($targetClass->subClasses) {
                                         if (! $targetClass->isInheritanceTypeNone() && isset($targetClass->discriminatorColumn)) {
-                                            // Attention! This function executes an extra query on the database.
-                                            $targetSubClass = $this->getTargetSubClass($targetClass, $associatedId);
+                                            foreach ($assoc['joinColumns'] as $joinColumn) {
+                                                $key = $joinColumn['name'];
+                                                if (isset($extra) && array_key_exists($key, $extra)) {
+                                                    $subTypeValue = $extra[$key][$targetClass->discriminatorColumn['name']];
+                                                }
+                                            }
 
-                                            $newValue = $this->em->getProxyFactory()->getProxy($targetSubClass, $associatedId);
+                                            if (! isset($subTypeValue)) {
+                                                $newValue = $this->getEntityPersister($assoc['targetEntity'])->loadOneToOneEntity($assoc, $entity, $associatedId);
+                                                break;
+                                            }
+
+                                            $targetSubClass = $targetClass->discriminatorMap[$subTypeValue];
+
+                                            $normalizedSubAssociatedId = $this->normalizeIdentifier($this->em->getClassMetadata($targetSubClass), $associatedId);
+                                            $newValue                  = $this->em->getProxyFactory()->getProxy($targetSubClass, $normalizedSubAssociatedId);
                                             break;
                                         }
-
-                                        $newValue = $this->em->getProxyFactory()->getProxy($assoc['targetEntity'], $associatedId);
-                                        break;
                                     }
 
                                     $newValue = $this->em->getProxyFactory()->getProxy($assoc['targetEntity'], $normalizedAssociatedId);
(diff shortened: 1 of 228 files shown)


mpdude:
Maybe someone from the Doctrine team can reopen the old PR? There are some comments that give explanations and background, and it would be great if we could continue with these instead of starting with a blank slate here.

SanderBoelaars:
Can a maintainer please take a look at this: https://github.com/doctrine/orm/actions/runs/4436405563
It needs to be approved before it can start. 

Sadly I don't think it would be possible to reopen the closed PR's like @mpdude suggested because the relevant branches have been deleted.

lib/Doctrine/ORM/UnitOfWork.php
mever, line 1: loadOneToOneEntity performs a query using BasicEntityPersister::load(). And $this->getTargetSubClass() is also fetching data from the database. Both data can reasonably be assumed to be reached with an index. So, all in all, this seems like a very good feature.

mever (approved):
This feature looks very nice. It adorns Doctrine to include this PR and make the project yet another bit more awesome :)

greg0ire:
> Maybe someone from the Doctrine team can reopen the old PR? There are some comments that give explanations and background, and it would be great if we could continue with these instead of starting with a blank slate here.

For some reason I cannot.

mever:
@beberlei , can you spare some time to approve/disapprove this pull request? We would like to upgrade to a version of Doctrine that has this feature.

mpdude:
As I've written in https://github.com/doctrine/orm/pull/10538#issuecomment-1445337437, to me it seems a bit strange that the `UnitOfWork` until now does not do any SQL querying/processing. So to me it seems your new code is in the wrong place, if it all of a sudden makes the UoW build SQL queries.

Imagine we'd need some more abstraction, like for SQL variants. Where would you place this code then? Can you find similar SQL code there already, which would be an indication of being in the right place architecture-wise?

mever:
Hi @mpdude , first I was thinking about placing it in the EntityPersister because it executes queries. But then I realized it does not deal with proxies, which is exactly what we want. That said, maybe the ProxyFactory would be a better place.

One method there looks like a good candidate: ProxyFactory::getProxy(), it can return a subclass of the first parameter and still be compliant with Liskov Substitutability. The downside of that approach, however, is that then the proxy factory is going to execute queries. Or at least for proxies that do not have a fixed type (i.e.: in the case of subclasses that aren't proxies themselves). To communicate intent, we could also add ProxyFactory::getSubclassProxy() which only deals with inherited entities but is functionally equivalent to ProxyFactory::getProxy(). That would be more transparent, and adding a comment about it executing a query to fetch subclass information from the database would be really clear for people that care about performance, as we do ;)

What do you think about splitting the methods or should we combine them to increase abstraction? Or were you thinking in a different direction?

SanderBoelaars:
Hello @gregOire @mpdude @beberlei, Like @mever, I myself was also looking at placing the code in the EntityPersister, but after reading @mever's message I do agree that ProxyFactory is probably a better class for this. Like @beberlei suggested earlier (https://github.com/doctrine/orm/pull/10538) I have extracted the code which executes the extra query into a separate function and I have now moved this function to the ProxyFactory. Would this be acceptable? If not, maybe EntityPersister would be a good place after all?

mever:
@beberlei What do you think? Should we put some effort into this in order to close this PR?

mpdude:
The test case you added is valueable: I used it to step through the code and to me it seems `\Doctrine\ORM\Persisters\Entity\BasicEntityPersister::getSelectColumnsSQL` would be a good place for (some) of these changes.

That method already contains some checks to see if we're associated with other entities, to consider the type of association and if inheritance is used. I haven't thought through all the possible combinations/association types, but to me it seems this method rather `continue`s early in these challenging situations.

What you could do there is to JOIN the table of the referred-to entity. But, instead of adding all the fields and potential associations, select just the discriminator column. Most of the code should already  be there.

Also, make sure to add the relevant information to the ResultSetMapping, but make sure to call `addMetaResult` only (indicating names/aliases used for information like a discriminator column).

By looking at `BasicEntityPersister::getSelectColumnsSQL()`, you also see things you potentially need to deal with when building SQL selects yourself – namely the `quoteStrategy` or the `filterSql` mechanism.

Having the changes in `BasicEntityPersister` puts them in a class that is already involved in building SQL queries, instead of adding this as a new concern to the `UnitOfWork`.

Also, it should be possible to load at least the discriminator values _for all rows at once_ instead of having to look them up one by one for every entity being hydrated.

SanderBoelaars:
Hello @mpdude, @beberlei, @gregoire, @mever

We have come up with a solution. 

Like suggested by @mpdude, we started by looking at the \Doctrine\ORM\Persisters\Entity\BasicEntityPersister::getSelectColumnsSQL function, in here the getSelectColumnAssociationSQL function gets called for all associations of the object we are trying to find. In this function we have added a check to see if the association object has a discriminator column, which would mean it uses inheritance. If so, we adjust the query to also include the discriminator column value of this association (which we are going to need later to initialize a Proxy). 

We also map the found discriminator column name to the association column name it belongs to, by using a new list created in the ResultSetMapping (named associationDiscriminatorMapping). Since we need the discriminator type of the association we also add a JOIN to the query in this function. With this we have altered the query to include the types of associations which are using inheritance, and have the necessary data in the ResultSetMapping to later find out which of those newly included types belong to which associations.

We then added a check inside the \Doctrine\ORM\Internal\Hydration\AbstractHydrator::hydrateColumnInfo function, which gets executed when we are hydrating the row data from our executed query. This check detects whether a key is a discriminator column name belonging to one of the associations of the object we are trying to hydrate. We do this by checking the new associationDiscriminatorMapping for this key. If we find a result, the return array will include: 
1) the field name belonging to the key 
2) the associated field on the object we are trying to find indicating the relation with the object to which the key belongs to 
3) the alias to which the associated field belongs too

Important here is that even though technically the discriminator field of an association would belong to that association, and therefore would be included under the alias of that association's class in the rowData we get back from the gatherRowData function inside the hydrateRowData function, this is not what we want. This is because for each alias in the rowData, the getEntity function gets called and the fields belonging to that alias are passed as arguments, therefore when creating the Proxy we still wouldn't have the type of the association because this field would not be included (since it would belong to a different alias). 

We tried adding the discriminator field to the alias of the class which has the reference to the associated object and needs to know its type to create a Proxy, but this is troublesome because this alias will also include its own discriminator field and these two will overwrite each other, we also tried giving the association discriminator fields unique names but couldn't find a way to make sure these names wouldn't already exist as fields of the class. We finally decided to add an 'extra' array to the data sent to the getEntity class, which includes the discriminator columns types of the associations using inheritance. 

In the \Doctrine\ORM\UnitOfWork::createEntity function, when trying to make a Proxy for an association we check the 'extra' data for the discriminator types of the associations. If there is no 'extra' data, the object will get loaded like it would be, before these code changes.

mever:
First of all, sorry for the long wait. I've already reviewed a part of this PR when it had a different approach, but I like the new direction it took. It looks very well written. I do not see any blocks or other issues. Therefore, I hope it can be merged now.
@mpdude , @beberlei  Can anyone with the power to approve this take a look?

mever:
Hi @mpdude , can you take a look at this? It would be nice if this can be a default feature of Doctrine ORM.

PR #10583, opened 2023-03-15
docs: Remove incorrect @SequenceGenerator info

Mika56:
### Improvement

docs: Remove incorrect @SequenceGenerator info

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | no

#### Summary

The default allocationSize for @SequenceGenerator has actually been set to 1 ever since 434325e (back in 2010!)
This was done to remove confusion, but the docs were never updated to reflect the change


derrabus (approved):

derrabus:
Good catch, thanks.

Issue #5827, opened 2016-05-16
Criteria in Many-To-Many relations broken if not manually accessing data before or EAGER used


enekochan:
I updated `doctrine/orm` and `doctrine/dbal` to `dev-master` some time ago to be able to use Criteria in Many-To-Many relations (with https://github.com/doctrine/doctrine2/pull/885). Suddenly today (now I'm not sure if it ever worked but I would say it did when using the same function in other places) I wasn't getting any objects in the collections of any of the sides. The only way it worked was If I added a bogus loop before the function with Criteria code was called (I suppose because the objects were being loaded beforehand and then they were available) or if I added `fetch="EAGER"` in the ManyToMany relation.

I'll try to resume my code bellow as best as I can. The function with the Criteria is `getActiveServices` inside the `Resource` class. It's used to get the Services from the Resource that are available depending on their start/end date (both can have a value or be null at the same time: no dates, only start date, only end date or both dates specified).

The action in the controller:

```
    public function servicesLoadAction(Request $request, Business $business)
    {
        $params = $request->getMethod() === 'POST' ? $request->request->all() : $request->query->all();
        $resource = array_key_exists('resource', $params) ? $params['resource'] : null;

        if (!is_null($resource)) {
            $resource = $this->getDoctrine()->getRepository('AppBundle:Resource')->find($resource);
        }
/*
        // The bogus loop that make this work without using fetch="EAGER"
        foreach ($business->getResources() as $resource) {
            $i = $resource->getServices()->count();
        }
*/
        $scheduleManager = $this->get('app.model.schedule');
        if (!is_null($resource)) {
            $services = $scheduleManager->getAvailableResourceServices($business, $resource);
        } else {
            $services = $scheduleManager->getAvailableServices($business);
        }

        return new JsonResponse($this->json_decode_group($services, 'service_minimum'));
    }
```

The model:

```
    public function getAvailableResourceServices(Business $business, \AppBundle\Entity\Resource $resource)
    {
        if ($business->getEnabledHumanResources()->contains($resource)) {
            return $resource->getActiveServices(true); // THIS IS THE FUNCTION WITH Criteria
        }

        return array();
    }
```

Resource class:

```
/**
 * @ORM\Entity(repositoryClass="AppBundle\Entity\ResourceRepository")
 * @ORM\Table(name="resources")
 */
class Resource
{
    /**
[...]

PR #10587, opened 2023-03-21
Handle null comparisons in `ManyToManyPersister`

### LINKED ISSUE #5587: @ManyToMany relation matching criteria ignores isNull(column)
I have an entity class where ManyMany relation described.

Then Im trying to do something like 

``` php
public function getActiveRel()
{
   $criteria = Criteria::create()
   ->where(Criteria::expr()->isNull('column'));

   return $this->rel->matching($criteria);
}
```

IsNull condition is ignored in SqlValueVisitor.php
https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Persisters/SqlValueVisitor.php#L57

And no any IS NULL in final SQL statement.

### LINKED ISSUE #7717: isNull criteria doesn't work

MatTheCat:
Fixes #5587, #5827 and #7717

The reason for https://github.com/doctrine/orm/blob/a50a611bee96f5c23c6e8bc0b46c8baba599d328/lib/Doctrine/ORM/Persisters/SqlValueVisitor.php#L34-L38 was that [`BasicEntityPersister` replaces null comparisons with `IS [NOT] NULL`](https://github.com/doctrine/orm/blob/a50a611bee96f5c23c6e8bc0b46c8baba599d328/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1650-L1655) so it does not need the corresponding parameter.

Problem is, `ManyToManyPersister` has then no longer access to the comparison.

My fix is to remove the condition in `SqlValueVisitor` and make each persister wary of null comparisons. It feels super weird but tests pass.

### DIFF
--- a/tests/Doctrine/Tests/Models/GH7717/GH7717Parent.php
+++ b/tests/Doctrine/Tests/Models/GH7717/GH7717Parent.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\GH7717;
+
+use Doctrine\Common\Collections\Selectable;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="gh7717_parents")
+ */
+class GH7717Parent
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue
+     */
+    public ?int $id = null;
+
+    /**
+     * @ORM\ManyToMany(targetEntity="GH7717Child", cascade={"persist"})
+     */
+    public Selectable $children;
+}
--- a/tests/Doctrine/Tests/Models/GH7717/GH7717Parent.php
+++ b/tests/Doctrine/Tests/Models/GH7717/GH7717Parent.php
@@ -22,6 +22,8 @@ class GH7717Parent
 
     /**
      * @ORM\ManyToMany(targetEntity="GH7717Child", cascade={"persist"})
+     *
+     * @var Selectable<int, GH7717Child>
      */
     public Selectable $children;
 }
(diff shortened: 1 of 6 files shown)


tests/Doctrine/Tests/Models/GH7717/GH7717Parent.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue
+     */
+    public ?int $id = null;
+
+    /**
+     * @ORM\ManyToMany(targetEntity="GH7717Child", cascade={"persist"})
```
greg0ire, line 24:
```suggestion
     * @ORM\ManyToMany(targetEntity="GH7717Child", cascade={"persist"})
     * @var Selectable<int, GH7717Child>
```

MatTheCat: (I added a line break between the annotations to make PHP_CodeSniffer happy.)

greg0ire (approved):

derrabus (approved):

MatTheCat:
Thanks! #5827 and #7717 can be closed; I thought GitHub would do it automatically.

greg0ire:
I think you have to repeat `Fixes` every time for that to happen :slightly_smiling_face:

MatTheCat:
Indeed, just read https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword 😅

PR #10588, opened 2023-03-21
add Automatic left join

AlexGouget:
QueryBuilder function that creates and adds multiple left join base on entity's relations.


beberlei:
This is not something we would be adding to the query builder, I recommend that you add a method to your code base that does this along the lines of:

```php
QueryBuilderHelper::addAutomaticLeftJoin($queryBuilder);
```

PR #1570, opened 2015-11-25
[RFC] Tests around reported cases over DDC-2524

guilhermeblanco:
Fixes #707

PR #10592, opened 2023-03-22
Add a new topological sort implementation

mpdude:
This is the first chunk to break #10547 into smaller PRs suitable for reviewing. It adds a new topological sort implementation.

 #### Background

Topological sort is an algorithm that sorts the vertices of a directed acyclic graph (DAG) in a linear order such that for every directed edge from vertex A to vertex B, vertex A comes before vertex B in the ordering. This ordering is called a topological order.

Ultimately (beyond the scope of this PR), in the ORM we'll need this to find an order in which we can insert new entities into the database. When one entity needs to refer to another one by means of a foreign key, the referred-to entity must be inserted before the referring entity. Deleting entities is similar.

A topological sorting can be obtained by running a depth first search (DFS) on the graph. The order in which the DFS finishes on the vertices is a topological order. The DFS is possible iif there are no cycles in the graph. When there are cycles, the DFS will find them.

For more information about topological sorting, as well as a description of an DFS-based topological sorting algorithm, see https://en.wikipedia.org/wiki/Topological_sorting.

 #### Current situation

There is a DFS-based topological sorting implemented in the `CommitOrderCalculator`. This implementation has two kinks:

1. It does not detect cycles

When there is a cycle in the graph, we need to know about it. Ultimately, this means we will not be able to insert entities into the database in any order that allows all foreign keys constraints to be satisfied.

If you look at `CommitOrderCalculator`, you'll see that there is no code dealing with this situation.

2. It has an obscure concept of edge "weights"

To me, it is not clear how those are supposed to work. The weights are related to whether a foreign key is nullable or not, but can (could) be arbitrary integers. An edge will be ignored if it has a higher (lower) weight than another, already processed edge... 🤷🏻?

 #### Suggested change

In fact, when inserting entities into the database, we have two kinds of foreign key relationships: Those that are `nullable`, and those that are not.

Non-nullable foreign keys are hard requirements: Referred-to entities must be inserted first, no matter what. These are "non-optional" edges in the dependency graph.

Nullable foreign keys can be set to `NULL` when first inserting an entity, and then coming back and updating the foreign key value after the referred-to (related) entity has been inserted into the database. This is already implemented in `\Doctrine\ORM\UnitOfWork::scheduleExtraUpdate`, at the expense of performing one extra `UPDATE` query after all the `INSERT`s have been processed. These edges are "optional".

When finding a cycle that consists of non-optional edges only, treat it as a failure. We won't be able to insert entities with a circular dependency when all foreign keys are non-NULLable.

When a cycle contains at least one optional edge, we can use it to break the cycle: Use backtracking to go back to the point before traversing the last _optional_ edge. This omits the edge from the topological sort order, but will cost one extra UPDATE later on.

To make the transition easier, the new implementation is provided as a separate class, which is marked as `@internal`.

 #### Outlook

Backtracking to the last optional edge is the simplest solution for now. In general, it might be better to find _another_ (optional) edge that would also break the cycle, if that edge is also part of _other_ cycles.

Remember, every optional edge skipped costs an extra UPDATE query later on. The underlying problem is known as the "minimum feedback arc set" problem, and is not easily/efficiently solvable. Thus, for the time being, picking the nearest optional edge seems to be reasonable.

#### Extra bonus

This is what the DALL·E AI thinks it looks like when you have a directed graph with cycles, and some wizardry finds a solution to break those.

DALL·E 2023-03-22 22 53 55

### DIFF
--- a/lib/Doctrine/ORM/Internal/TopologicalSort.php
+++ b/lib/Doctrine/ORM/Internal/TopologicalSort.php
@@ -1,0 +1,165 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Internal;
+
+use Doctrine\ORM\Internal\TopologicalSort\CycleDetectedException;
+
+use function array_keys;
+use function array_reverse;
+use function array_unshift;
+use function spl_object_id;
+
+/**
+ * TopologicalSort implements topological sorting, which is an ordering
+ * algorithm for directed graphs (DG) using a depth-first searching (DFS)
+ * to traverse the graph built in memory.
+ * This algorithm has a linear running time based on nodes (V) and edges
+ * between the nodes (E), resulting in a computational complexity of O(V + E).
+ *
+ * @internal
+ */
+final class TopologicalSort
+{
+    private const NOT_VISITED = 1;
+    private const IN_PROGRESS = 2;
+    private const VISITED     = 3;
+
+    /**
+     * Array of all nodes, indexed by object ids.
+     *
+     * @var array<int, object>
+     */
+    private $nodes = [];
+
+    /**
+     * DFS state for the different nodes, indexed by node object id and using one of
+     * this class' constants as value.
+     *
+     * @var array<int, self::*>
+     */
+    private $states = [];
+
+    /**
+     * Edges between the nodes. The first-level key is the object id of the outgoing
+     * node; the second array maps the destination node by object id as key. The final
+     * boolean value indicates whether the edge is optional or not.
+     *
+     * @var array<int, array<int, bool>>
+     */
+    private $edges = [];
+
+    /**
+     * Builds up the result during the DFS.
+     *
+     * @psalm-var list<object>
+     */
+    private $sortResult = [];
+
+    /** @param object $node */
+    public function addNode($node): void
+    {
+        $id                = spl_object_id($node);
+        $this->nodes[$id]  = $node;
+        $this->states[$id] = self::NOT_VISITED;
+        $this->edges[$id]  = [];
+    }
+
+    /** @param object $node */
+    public function hasNode($node): bool
+    {
+        return isset($this->nodes[spl_object_id($node)]);
+    }
+
+    /**
+     * Adds a new edge between two nodes to the graph
+     *
+     * @param object $from
+     * @param object $to
+     * @param bool   $optional This indicates whether the edge may be ignored during the topological sort if it is necessary to break cycles.
+     */
+    public function addEdge($from, $to, bool $optional): void
+    {
+        $fromId = spl_object_id($from);
+        $toId   = spl_object_id($to);
+
+        if (isset($this->edges[$fromId][$toId]) && $this->edges[$fromId][$toId] === false) {
+            return; // we already know about this dependency, and it is not optional
+        }
+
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ORM/Internal/TopologicalSort.php
```diff
@@ -0,0 +1,146 @@
...
+use function array_keys;
+use function array_reverse;
+use function array_unshift;
+use function spl_object_id;
+
+/**
+ * TopologicalSort implements topological sorting, which is an ordering
+ * algorithm for directed graphs (DG) using a depth-first searching (DFS)
+ * to traverse the graph built in memory.
+ * This algorithm have a linear running time based on nodes (V) and edges
```
greg0ire:
```suggestion
 * This algorithm has a linear running time based on nodes (V) and edges
```

lib/Doctrine/ORM/Internal/TopologicalSort.php
```diff
@@ -0,0 +1,146 @@
...
+ * between the nodes (E), resulting in a computational complexity of O(V + E).
+ */
+class TopologicalSort
```
greg0ire: It's internal, but I think I'd still make it `final`, for us.

lib/Doctrine/ORM/Internal/TopologicalSort.php
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * @var array<int, object>
+     */
+    private $nodes = [];
+
+    /** @var array<int, int> */
+    private $states = [];
+
+    /** @var array<int, array<int, bool>> */
+    private $edges = [];
```
greg0ire, line 51: Please also document `states` and `edges`, like you did `nodes`

lib/Doctrine/ORM/Internal/TopologicalSort.php
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @var array<int, object>
+     */
+    private $nodes = [];
+
+    /**
+     * DFS state for the different nodes, indexed by node object id and using one of
+     * this class' constants as value.
+     *
+     * @var array<int, int>
```
greg0ire:
```suggestion
     * @var array<int, self::*>
```

mpdude: Oh 🤗

greg0ire:
Git says the concept of "weight" was introduced in https://github.com/doctrine/orm/pull/1570
@guilhermeblanco hi! Could you maybe shed some light on how they work?

mpdude:
@stof you fixed something related to circular dependencies in #7260 a while ago, so maybe you’re familiar enough with this topic to give a review?

Thanks!

mpdude:
@stof you‘re one of the persons who last worked on or fixed the old `CommitOrderCalculator`.

Do you think you could give this implementation a review?

mpdude:
@derrabus I've added the `@internal` docbock and a comment explaining the use of `array_reverse`.

mpdude:
@derrabus Do you :+1: this? I need to get this merged into the temporary branch so I can continue working on #10547.

derrabus (approved):

PR #10599, opened 2023-03-28
fix(persistent-collection): check association is not nullable before using it as an array

amina-seraoui:
Hi,

I've faced the same issue explained here : https://github.com/doctrine/orm/pull/8356 so I decided to retry his commit.

Warning: Trying to access array offset on value of type null

I only check if `$this->association !== null` like on the linked PR.

Thank you !

### DIFF
--- a/lib/Doctrine/ORM/PersistentCollection.php
+++ b/lib/Doctrine/ORM/PersistentCollection.php
@@ -12,6 +12,7 @@ use Doctrine\Common\Collections\Selectable;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use ReturnTypeWillChange;
 use RuntimeException;
+use UnexpectedValueException;
 
 use function array_combine;
 use function array_diff_key;
@@ -162,7 +163,7 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
 
         // If _backRefFieldName is set and its a one-to-many association,
         // we need to set the back reference.
-        if ($this->backRefFieldName && $this->association['type'] === ClassMetadata::ONE_TO_MANY) {
+        if ($this->backRefFieldName && $this->getMapping()['type'] === ClassMetadata::ONE_TO_MANY) {
             assert($this->typeClass !== null);
             // Set back reference to owner
             $this->typeClass->reflFields[$this->backRefFieldName]->setValue(
@@ -191,7 +192,7 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
 
         // If _backRefFieldName is set, then the association is bidirectional
         // and we need to set the back reference.
-        if ($this->backRefFieldName && $this->association['type'] === ClassMetadata::ONE_TO_MANY) {
+        if ($this->backRefFieldName && $this->getMapping()['type'] === ClassMetadata::ONE_TO_MANY) {
             assert($this->typeClass !== null);
             // Set back reference to owner
             $this->typeClass->reflFields[$this->backRefFieldName]->setValue(
@@ -272,10 +273,14 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
     /**
      * INTERNAL: Gets the association mapping of the collection.
      *
-     * @psalm-return AssociationMapping|null
+     * @psalm-return AssociationMapping
      */
-    public function getMapping(): ?array
+    public function getMapping(): array
     {
+        if ($this->association === null) {
+            throw new UnexpectedValueException('The underlying association mapping is null although it should not be');
+        }
+
         return $this->association;
     }
 
@@ -292,8 +297,8 @@ final class PersistentCollection extends AbstractLazyCollection implements Selec
 
         if (
             $this->association !== null &&
-            $this->association['isOwningSide'] &&
-            $this->association['type'] === ClassMetadata::MANY_TO_MANY &&
+            $this->getMapping()['isOwningSide'] &&
+            $this->getMapping()['type'] === ClassMetadata::MANY_TO_MANY &&
             $this->owner &&
             $this->em !== null &&
             $this->em->getClassMetadata(get_class($this->owner))->isChangeTrackingNotify()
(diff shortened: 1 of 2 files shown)


greg0ire:
Thanks for your contribution! 

In what situation does this arise?

I'm not sure this is the right way to handle this though: it feels like sweeping an issue under the rug. Why is `association` empty in the first place?

Instead, I would do something like this:

```php
use UnexpectedValueException();

/** @return AssociationMapping */
private function association(): array
{
    if ($this->association === null) {
        throw new UnexpectedValueException('The underlying association mapping is null although it should not be');
    }
   
    return $this->association;
}
```

And then use `->association()` where you performed the check.

Cc @Kocal

amina-seraoui:
Hello,

Yes indeed, the problem was coming from my object which has been wrongly initialized, I'm not sure. In any case, the problem was solved without modify the ORM package. Even if it was an easy way out to just check if is null at first sight.

Do you want me to implement your suggestion on this PR ?

greg0ire:
Yes please 👍

amina-seraoui:
We already have a function that return association :

```php
/**
 * INTERNAL: Gets the association mapping of the collection.
 *
 * @psalm-return array<string, mixed>|null
 */
public function getMapping(): ?array
{
    return $this->association;
}
```

Should I throw the exception from here ?

greg0ire:
Yes please, that's even better 👍

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -271,10 +272,14 @@ public function getInsertDiff(): array
...
-     * @psalm-return array<string, mixed>|null
+     * @psalm-return array<string, mixed>
      */
-    public function getMapping(): ?array
+    public function getMapping(): array
     {
+        if ($this->association === null) {
+            throw new UnexpectedValueException('The underlying association mapping is null although it should not be');
+        }
+
```
greg0ire, line 283: Nice! Now you should use `getMapping()` where you previously added the null checks.

greg0ire:
Ok, now I expect Psalm to be failing because there are issues in the baseline that we have ignored, and that no longer need to be ignored.

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -426,8 +431,8 @@ public function get($key)
     {
         if (
             ! $this->initialized
-            && $this->association['fetch'] === ClassMetadata::FETCH_EXTRA_LAZY
-            && isset($this->association['indexBy'])
+            && $this->getMapping()['fetch'] === ClassMetadata::FETCH_EXTRA_LAZY
+            && isset($this->getMapping()['indexBy'])
```
greg0ire, line 436: Here you could store the result in a variable called `$association`, and use it also on line 434/439

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -598,7 +603,7 @@ public function __sleep(): array
      */
     public function slice($offset, $length = null): array
     {
-        if (! $this->initialized && ! $this->isDirty && $this->association['fetch'] === ClassMetadata::FETCH_EXTRA_LAZY) {
+        if (! $this->initialized && ! $this->isDirty && $this->getMapping()['fetch'] === ClassMetadata::FETCH_EXTRA_LAZY) {
```
greg0ire, line 608: Same here and there are more occurrences below

amina-seraoui: Do I replace occurences when null is checked before like :

```php
    private function changed(): void
    {
        if ($this->isDirty) {
            return;
        }

        $this->isDirty = true;

        if (
            $this->association !== null &&
            $this->getMapping()['isOwningSide'] &&
            $this->getMapping()['type'] === ClassMetadata::MANY_TO_MANY &&
            $this->owner &&
            $this->em !== null &&
            $this->em->getClassMetadata(get_class($this->owner))->isChangeTrackingNotify()
        ) {
            $this->getUnitOfWork()->scheduleForDirtyCheck($this->owner);
        }
    }
```

It can maybe result side effects.

greg0ire: No you should not.

lib/Doctrine/ORM/PersistentCollection.php
```diff
@@ -424,26 +431,27 @@ public function contains($element): bool
      */
     public function get($key)
     {
+        $association = $this->getMapping();
```
greg0ire: In restrospect, this was not a good idea… the association might be `null` here, and that is a legit case when `$this->initialized` is `true`. You should probably inline `getMapping()` everywhere.

amina-seraoui: Do you mean by doing an assignement like this ?
```php
if (! $this->initialized && ($association = $this->getMapping())['fetch'] === ClassMetadata::FETCH_EXTRA_LAZY)
```

(assignement in if condition is  not allowed by a phpcs rule)

greg0ire: No I mean use `if (! $this->initialized && $this->getMapping()['fetch'] === ClassMetadata::FETCH_EXTRA_LAZY)` (inline `$this->getMapping()`, not the whole assignement

greg0ire:
I think all you need to do now is run `vendor/bin/psalm --update-baseline` to get rid of the issues you fixed :)

greg0ire:
Rebased on 2.15.x since this is more an improvement than it is a fix. Also, I took care of the baseline.

greg0ire (approved):

SenseException (approved):
A lot of removed lines in the psalm-baseline :+1:

greg0ire:
Thanks @amina-seraoui!

amina-seraoui:
Thanks to you ! 🎉

Kocal:
Thanks :)

PR #10601, opened 2023-03-29
Support options like charset and collation on DiscriminatedColumn

### LINKED ISSUE #10462: Add options (charset, collation) to DiscriminatorColumn to align it with Column options
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

As stated in [related dbal issue](https://github.com/doctrine/dbal/pull/5869#issue-1556699509), when trying to declare charset and collation for discriminator column, the only option to use is `columnDefinition` (which causes the migration issue described in the dbal issue). I believe that [DiscriminatorColumn](https://github.com/doctrine/orm/blob/2.14.1/lib/Doctrine/ORM/Mapping/DiscriminatorColumn.php) is not that much different from regular [Column](https://github.com/doctrine/orm/blob/2.14.1/lib/Doctrine/ORM/Mapping/Column.php) and it should support the same features.

JanTvrdik:
Closes: https://github.com/doctrine/orm/issues/10462

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Builder/ClassMetadataBuilder.php
+++ b/lib/Doctrine/ORM/Mapping/Builder/ClassMetadataBuilder.php
@@ -220,10 +220,11 @@ class ClassMetadataBuilder
      * @param string $type
      * @param int    $length
      * @psalm-param class-string<BackedEnum>|null $enumType
+     * @psalm-param array<string, mixed> $options
      *
      * @return $this
      */
-    public function setDiscriminatorColumn($name, $type = 'string', $length = 255, ?string $columnDefinition = null, ?string $enumType = null)
+    public function setDiscriminatorColumn($name, $type = 'string', $length = 255, ?string $columnDefinition = null, ?string $enumType = null, array $options = [])
     {
         $this->cm->setDiscriminatorColumn(
             [
@@ -232,6 +233,7 @@ class ClassMetadataBuilder
                 'length' => $length,
                 'columnDefinition' => $columnDefinition,
                 'enumType' => $enumType,
+                'options' => $options,
             ]
         );
 
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -3131,7 +3131,7 @@ class ClassMetadataInfo implements ClassMetadata
      * @see getDiscriminatorColumn()
      *
      * @param mixed[]|null $columnDef
-     * @psalm-param array{name: string|null, fieldName?: string, type?: string, length?: int, columnDefinition?: string|null, enumType?: class-string<BackedEnum>|null}|null $columnDef
+     * @psalm-param array{name: string|null, fieldName?: string, type?: string, length?: int, columnDefinition?: string|null, enumType?: class-string<BackedEnum>|null, options?: array<string, mixed>}|null $columnDef
      *
      * @return void
      *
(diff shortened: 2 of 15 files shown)


lib/Doctrine/ORM/Mapping/Builder/ClassMetadataBuilder.php
```diff
@@ -220,10 +220,11 @@ public function setSingleTableInheritance()
      * @param string $type
      * @param int    $length
      * @psalm-param class-string<BackedEnum>|null $enumType
+     * @psalm-param array<string, mixed> $options
      *
      * @return $this
      */
-    public function setDiscriminatorColumn($name, $type = 'string', $length = 255, ?string $columnDefinition = null, ?string $enumType = null)
+    public function setDiscriminatorColumn($name, $type = 'string', $length = 255, ?string $columnDefinition = null, ?string $enumType = null, array $options = [])
```
greg0ire, line 227: Technically a breaking change for extending classes, but we didn't frown upon a similar change in https://github.com/doctrine/orm/pull/10288 , so maybe we should just make the class as final on 3.0.x.

greg0ire (commented):
This looks like a new feature, and as such should target 2.15.x

lib/Doctrine/ORM/Mapping/Builder/ClassMetadataBuilder.php
greg0ire: BTW, that PR modified docs/en/reference/php-mapping.rst , and this one should too I think. You might want to look at that PR for other changes that need to be done.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -3131,7 +3131,7 @@ public function addEntityListener($eventName, $class, $method)
      * @see getDiscriminatorColumn()
      *
      * @param mixed[]|null $columnDef
-     * @psalm-param DiscriminatorColumnMapping|array{name: string|null, fieldName?: string, type?: string, length?: int, columnDefinition?: string|null, enumType?: class-string<BackedEnum>|null}|null $columnDef
+     * @psalm-param DiscriminatorColumnMapping|array{name: string|null, fieldName?: string, type?: string, length?: int, columnDefinition?: string|null, enumType?: class-string<BackedEnum>|null, options?: array<string, mixed>}|null $columnDef
```
greg0ire: :thinking: I don't know why I left this `|array{…}` part in https://github.com/doctrine/orm/pull/10513 but it looks wrong now.

greg0ire: https://github.com/doctrine/orm/pull/10602

greg0ire: Merged. please address the conflict.

tests/Doctrine/Tests/ORM/Mapping/AttributeReaderTest.php
```diff
@@ -101,3 +122,21 @@ class TestTag
...
+#[Entity]
+#[InheritanceType('SINGLE_TABLE')]
+#[DiscriminatorColumn(name: 'discr', options: ['charset' => 'ascii', 'collation' => 'ascii_general_ci'])]
+#[DiscriminatorMap(['person' => TestPerson::class, 'employee' => TestEmployee::class])]
+class TestPerson
+{
+    #[Id]
+    #[Column(type: 'integer')]
+    #[GeneratedValue]
+    public int|null $id = null;
```
greg0ire: While PHP 7.4 and below understands attributes as just comments, it cannot understand this union type.

greg0ire (approved):

derrabus (approved):

greg0ire:
Thanks @JanTvrdik !

PR #10602, opened 2023-03-29
Remove duplicate array shape

greg0ire:
`DiscriminatorColumnMapping` is just a specialization of the array shape
that is right of the pipe: it has the same fields, except fewer fields
are nullable. The union of that is the same thing as the array shape.


SenseException (approved):
There were no failing checks before?

greg0ire:
If you mean before I force-pushed, there were, because I wrongly thought the right thing was to eliminate the right-hand side of the pipe.

PR #10607, opened 2023-03-31
Field mapping DTO

greg0ire:
In the past, it has been decided to use arrays for this out of
legitimate performance concerns. But PHP has evolved, and now, it is
more performant and memory efficient to use objects.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/FieldMapping.php
+++ b/lib/Doctrine/ORM/Mapping/FieldMapping.php
@@ -1,0 +1,142 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Mapping;
+
+use ArrayAccess;
+use BackedEnum;
+
+use function in_array;
+use function property_exists;
+
+/** @template-implements ArrayAccess<string, mixed> */
+final class FieldMapping implements ArrayAccess
+{
+    use ArrayAccessImplementation;
+
+    /** The database length of the column. Optional. Default value taken from the type. */
+    public int|null $length = null;
+    /**
+     * Marks the field as the primary key of the entity. Multiple
+     * fields of an entity can have the id attribute, forming a composite key.
+     */
+    public bool|null $id                 = null;
+    public bool|null $nullable           = null;
+    public bool|null $notInsertable      = null;
+    public bool|null $notUpdatable       = null;
+    public string|null $columnDefinition = null;
+    /** @psalm-var ClassMetadata::GENERATED_* */
+    public int|null $generated = null;
+    /** @var class-string<BackedEnum>|null */
+    public string|null $enumType = null;
+    /**
+     * The precision of a decimal column.
+     * Only valid if the column type is decimal
+     */
+    public int|null $precision = null;
+    /**
+     * The scale of a decimal column.
+     * Only valid if the column type is decimal
+     */
+    public int|null $scale = null;
+    /** Whether a unique constraint should be generated for the column. */
+    public bool|null $unique = null;
+    /**
+     * @var class-string|null This is set when the field is inherited by this
+     * class from another (inheritance) parent <em>entity</em> class. The value
+     * is the FQCN of the topmost entity class that contains mapping information
+     * for this field. (If there are transient classes in the class hierarchy,
+     * these are ignored, so the class property may in fact come from a class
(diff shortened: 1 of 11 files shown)


lib/Doctrine/ORM/Mapping/FieldMapping.php
```diff
@@ -0,0 +1,142 @@
...
+
+use function in_array;
+use function property_exists;
+
+/** @template-implements ArrayAccess<string, mixed> */
+final class FieldMapping implements ArrayAccess
+{
+    use ArrayAccessImplementation;
+
+    /** @var int|null The database length of the column. Optional. Default value taken from the type. */
```
derrabus:
```suggestion
    /** The database length of the column. Optional. Default value taken from the type. */
```

lib/Doctrine/ORM/Mapping/FieldMapping.php
```diff
@@ -0,0 +1,142 @@
...
+    public int|null $length = null;
+    /**
+     * @var bool|null Marks the field as the primary key of the entity. Multiple
```
derrabus:
```suggestion
     * Marks the field as the primary key of the entity. Multiple
```

lib/Doctrine/ORM/Mapping/FieldMapping.php
```diff
@@ -0,0 +1,142 @@
...
+     * fields of an entity can have the id attribute, forming a composite key.
+     */
+    public bool|null $id                 = null;
+    public bool|null $nullable           = null;
+    public bool|null $notInsertable      = null;
+    public bool|null $notUpdatable       = null;
+    public string|null $columnDefinition = null;
+    /** ClassMetadata::GENERATED_* */
```
derrabus:
```suggestion
    /** @psalm-var ClassMetadata::GENERATED_* */
```

lib/Doctrine/ORM/Mapping/FieldMapping.php
```diff
@@ -0,0 +1,142 @@
...
+    public int|null $generated = null;
+    /** @var class-string<BackedEnum>|null */
+    public string|null $enumType = null;
+    /**
+     * @var int|null The precision of a decimal column.
```
derrabus:
```suggestion
     * The precision of a decimal column.
```

lib/Doctrine/ORM/Mapping/FieldMapping.php
```diff
@@ -0,0 +1,142 @@
...
+     * Only valid if the column type is decimal
+     */
+    public int|null $precision = null;
+    /**
+     * @var int|null The scale of a decimal column.
```
derrabus:
```suggestion
     * The scale of a decimal column.
```

lib/Doctrine/ORM/Mapping/FieldMapping.php
```diff
@@ -0,0 +1,142 @@
...
+     * Only valid if the column type is decimal
+     */
+    public int|null $scale = null;
+    /** @var bool|null Whether a unique constraint should be generated for the column. */
```
derrabus:
```suggestion
    /** Whether a unique constraint should be generated for the column. */
```

derrabus (dismissed):

lib/Doctrine/ORM/Tools/Console/Command/MappingDescribeCommand.php
```diff
@@ -257,7 +257,7 @@ private function formatMappings(array $propertyMappings): array
         foreach ($propertyMappings as $propertyName => $mapping) {
             $output[] = $this->formatField(sprintf('  %s', $propertyName), '');
 
-            foreach ($mapping as $field => $value) {
+            foreach ((array) $mapping as $field => $value) {
```
SenseException, line 260: I assume having an IteratorAggregate would be too much and is supposed to be kept simple.

SenseException (approved):

SenseException:
Thank you for the effort of splitting up the former PR. :heart:

lib/Doctrine/ORM/Tools/Console/Command/MappingDescribeCommand.php
greg0ire: Yes, and it's not like this is a collection or something.

PR #10609, opened 2023-03-31
Extract discriminator column mapping into its own DTO

greg0ire:


psalm-baseline.xml
```diff
@@ -1566,6 +1566,10 @@
     <ArgumentTypeCoercion>
       <code>$classes</code>
     </ArgumentTypeCoercion>
+    <InvalidArrayOffset>
+      <code><![CDATA[$mapping['enumType']]]></code>
+      <code><![CDATA[$mapping['options']]]></code>
+    </InvalidArrayOffset>
```
greg0ire, line 1572: I think these 2 issues will disappear once `JoinColumnData` is migrated as well.

lib/Doctrine/ORM/Persisters/Entity/JoinedSubclassPersister.php
```diff
@@ -487,7 +487,7 @@ protected function getInsertColumnList(): array
 
         // Add discriminator column if it is the topmost class.
         if ($this->class->name === $this->class->rootEntityName) {
-            $columns[] = $this->class->getDiscriminatorColumn()['name'];
+            $columns[] = $this->class->getDiscriminatorColumn()->name;
```
greg0ire, line 490: This change and the one below make static analysis issues go away, that's why I decided to add them. There will be PRs dedicated to migrating to the object interface later.

derrabus (approved):

PR #1238, opened 2015-01-07
Avoid prefixing columns when `false` is assigned to `column-prefix`

andreyvital:
http://www.doctrine-project.org/jira/browse/DDC-3293

As the issue on JIRA wasn't updated since September/2014, here's a PR.

@Ocramius suggested on [issue](http://www.doctrine-project.org/jira/browse/DDC-3293) about introducing a new attribute called `use-column-prefix`, I'd say that keeping the same attribute as in others drivers would be better.

That's it.

PR #10612, opened 2023-04-02
Extract embedded class mapping into its own DTO

greg0ire:


lib/Doctrine/ORM/Mapping/EmbeddedClassMapping.php
```diff
@@ -0,0 +1,84 @@
...
+use ArrayAccess;
+
+use function property_exists;
+
+/** @template-implements ArrayAccess<string, mixed> */
+final class EmbeddedClassMapping implements ArrayAccess
+{
+    use ArrayAccessImplementation;
+
+    public string|bool|null $columnPrefix = null;
```
SenseException: Where does `bool` come from? From a PHP function that can return `false` on error and wasn't fixed yet?

SenseException (dismissed):

lib/Doctrine/ORM/Mapping/EmbeddedClassMapping.php
```diff
@@ -0,0 +1,84 @@
...
+     * @var class-string|null
+     */
+    public string|null $declared = null;
+
+    /** @param class-string $class */
+    public function __construct(public string $class)
+    {
+    }
+
+    /** @param array{class: class-string, columnPrefix: string|null, declaredField: string|null, originalField: string|null} $mappingArray */
```
derrabus:
```suggestion
    /** @param array{class: class-string, columnPrefix?: string|null, declaredField?: string|null, originalField?: string|null} $mappingArray */
```

tests/Doctrine/Tests/ORM/Mapping/EmbeddedClassMappingTest.php
```diff
@@ -0,0 +1,34 @@
...
+        $mapping->declared      = self::class; // sense
+
+        $resurrectedMapping = unserialize(serialize($mapping));
+        assert($resurrectedMapping instanceof EmbeddedClassMapping);
+
+        self::assertSame($resurrectedMapping->columnPrefix, 'these');
+        self::assertSame($resurrectedMapping->declaredField, 'values');
+        self::assertSame($resurrectedMapping->originalField, 'make');
+        self::assertSame($resurrectedMapping->inherited, self::class);
+        self::assertSame($resurrectedMapping->declared, self::class);
```
derrabus: Parameter order: expected value first, then the actual value.

derrabus (changes requested):

lib/Doctrine/ORM/Mapping/EmbeddedClassMapping.php
greg0ire: I must admit I didn't look into it until now. Apparently, it's the way to specify there should not be any column prefix: https://github.com/doctrine/orm/pull/1238. If you specify `null`, I think the column will be prefixed with the name of the entity the embedded class is embedded into.

greg0ire: Only `false` should be allowed, BTW.

greg0ire: Oh and it's legal in union types with php 8.1, so I can use it!

psalm-baseline.xml
```diff
@@ -553,6 +554,7 @@
     </ArgumentTypeCoercion>
     <InvalidArgument>
       <code>$columnDef</code>
+      <code>$mapping</code>
```
greg0ire, line 557: The new entries correspond to drivers (XMLDriver and AttributeDriver), when calling `mapEmbedded()`. An improvement could be to add some checks in the drivers regarding the types.

derrabus (approved):

PR #10617, opened 2023-04-11
Use Rector to migrate to the object API of FieldMapping

greg0ire:
This makes the array access implementation of FieldMapping useful only to consumers, it is no longer useful internally, and should be deprecated as of Doctrine 3.1.0

### DIFF
--- a/tests/Doctrine/Tests/ORM/Mapping/MappingDriverTestCase.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/MappingDriverTestCase.php
@@ -242,15 +242,15 @@ abstract class MappingDriverTestCase extends OrmTestCase
         self::assertTrue($class->isVersioned);
         self::assertEquals('version', $class->versionField);
 
-        self::assertFalse(isset($class->fieldMappings['version']['version']));
+        self::assertFalse(isset($class->fieldMappings['version']->version));
     }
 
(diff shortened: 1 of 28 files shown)


tests/Doctrine/Tests/ORM/Mapping/MappingDriverTestCase.php
```diff
@@ -504,23 +504,23 @@ public function testDefaultFieldType(): void
         self::assertArrayHasKey('id', $class->fieldMappings);
         self::assertArrayHasKey('name', $class->fieldMappings);
 
-        self::assertArrayHasKey('type', $class->fieldMappings['id']);
-        self::assertArrayHasKey('type', $class->fieldMappings['name']);
+        self::assertNotNull($class->fieldMappings['id']->type);
+        self::assertNotNull($class->fieldMappings['name']->type);
```
SenseException: Do we need those at this point? It's not an array anymore and if `type` isn't initialized, we would get a "must not be accessed before initialization" (but afaik the props are initialized by default). Nevertheless, the actual value is checked in the next lines. Same with the other assertNotNull methods below.

SenseException (changes requested):

derrabus (approved):

SenseException (approved):

PR #10636, opened 2023-04-17
Move joinTableColumns down

greg0ire:
This field only makes sense for the owning side of many-to-many mappings. Moving it down allows us to make it non-nullable.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -30,6 +30,10 @@ parameters:
             path: lib/Doctrine/ORM/Mapping/ToOneAssociationMapping.php
 
         -
+            message: "#^Access to an undefined property Doctrine\\\\ORM\\\\Mapping\\\\AssociationMapping\\:\\:\\$joinTableColumns\\.$#"
+            path: lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
+
+        -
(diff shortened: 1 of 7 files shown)


greg0ire:
At this point you're probably wondering why I'm not pushing the switch to the object API of the association mapping… well, it comes with lots of errors. In this cases, it's revealing a flaw in my design, so that's good I guess, and we shouldn't add the errors to the baseline too quickly. I'm going to do small PRs that should prevent some of them from ever showing up.

phpstan.neon
```diff
@@ -29,6 +29,10 @@ parameters:
             message: "#^Access to an undefined property Doctrine\\\\ORM\\\\Mapping\\\\ToOneAssociationMapping\\:\\:\\$joinColumns\\.$#"
             path: lib/Doctrine/ORM/Mapping/ToOneAssociationMapping.php
 
+        -
+            message: "#^Access to an undefined property Doctrine\\\\ORM\\\\Mapping\\\\AssociationMapping\\:\\:\\$joinTableColumns\\.$#"
+            path: lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
+
```
greg0ire, line 35: This is caused by the issue already linked line 27

SenseException (approved):

PR #10643, opened 2023-04-20
Fix single table inheritance with intermediate abstract class(es)

### LINKED ISSUE #10625: 2.14.2: Entity Inheritance (at least STI) with intermediate abstract class is broken
### BC Break Report

The Single Table Inheritance Entity Inheritance is defined in your documentation with

> #[DiscriminatorMap] declares the possible values for the discriminator column and maps them to class names in the hierarchy. This discriminator map has to declare all non-abstract entity classes that exist in that particular inheritance hierarchy. That includes the root as well as any intermediate entity classes, given they are not abstract.

but will fail on EntityManager::find() with an intermediate abstract entity class:
```
Undefined index: "DoctrineOrmTest\Intermediate"

vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/SingleTablePersister.php:157
vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/SingleTablePersister.php:128
vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php:1092
vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php:748
vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php:768
vendor/doctrine/orm/lib/Doctrine/ORM/EntityManager.php:521
vendor/doctrine/orm/lib/Doctrine/ORM/Decorator/EntityManagerDecorator.php:205
```
as the code iterates over all subclasses (abstract and non-abstract) and expects them in the discriminator map.

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.14.2

#### Summary

Apparently the intermediate abstract class seems to be expected in the discriminator map, even though the documentation states otherwise.

Not quite sure if the `ClassMetadataInfo::subClasses` iteration should directly access the discriminator values `discrValues` in `SingleTablePersister::getSelectConditionDiscriminatorValueSQL()`
```
        $discrValues = array_flip($this->class->discriminatorMap);

        foreach ($this->class->subClasses as $subclassName) {
            $values[] = $this->conn->quote($discrValues[$subclassName]);
        }
```
as `subClasses` is documented with having also abstract classes
```
     * READ-ONLY: For classes in inheritance mapping hierarchies, this field contains the names of all                  
     * <em>entity</em> subclasses of this class. These may also be abstract classes. 
```
where those are not to be expected in the discriminator map.

#### Previous behavior

It worked :tm: 

#### Current behavior

Access violation on runtime, but not on validation.

#### How to reproduce

Assume the following example of entities:
```php
namespace DoctrineOrmTest;

/**
 * @ORM\DiscriminatorMap({"1" = "DoctrineOrmTest\Concrete"})
 * @ORM\Entity()
 */
abstract class Base
{
    /**
     * @ORM\Column
     * @ORM\Id
     */
    private int $id;
}
```
```php
namespace DoctrineOrmTest;

/**
 * @ORM\Entity()
 */
abstract class Intermediate extends Base
{
}
```
```php
namespace DoctrineOrmTest;

/**
 * @ORM\Entity()
 */
class Concrete extends Intermediate
{
}
```
plus some boilerplate:

composer.json
```json
{
    "require": {
        "doctrine/annotations": "^2.0",
        "doctrine/orm": "^2.14.1",
        "symfony/cache": "^5.4"
    },
    "autoload": {
        "psr-4": {
            "DoctrineOrmTest\\": "src"
        }
    }
}
```
bootstrap.php
```php
require_once(__DIR__ . '/vendor/autoload.php');

$config = Doctrine\ORM\ORMSetup::createAnnotationMetadataConfiguration([__DIR__ . '/src'], true);
$connection = Doctrine\DBAL\DriverManager::getConnection(['driver' => 'pdo_sqlite', 'memory' => true,], $config);
return new Doctrine\ORM\EntityManager($connection, $config);
```
cli-config.php
```php
return Doctrine\ORM\Tools\Console\ConsoleRunner::createHelperSet(require_once(__DIR__ . '/bootstrap.php'));
```
and test.php
```php
set_error_handler(fn () => die(var_export(func_get_args(), true)));

$em = require_once(__DIR__ . '/bootstrap.php');
(new Doctrine\ORM\Tools\SchemaTool($em))->createSchema([$em->getClassMetadata(DoctrineOrmTest\Base::class)]);
$em->clear();

$em->find(DoctrineOrmTest\Concrete::class, 1);
$em->find(DoctrineOrmTest\Intermediate::class, 1);
$em->find(DoctrineOrmTest\Base::class, 1);
```
will result in
```
 $ php test.php 
array (
  0 => 8,
  1 => 'Undefined index: DoctrineOrmTest\\Intermediate',
  2 => 'vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/SingleTablePersister.php',
  3 => 157,
  4 => 
  array (
    'values' => 
    array (
      0 => '\'1\'',
    ),
    'discrValues' => 
    array (
      'DoctrineOrmTest\\Concrete' => 1,
    ),
    'subclassName' => 'DoctrineOrmTest\\Intermediate',
  ),

```
while
```
$ vendor/bin/doctrine orm:info
 Found 3 mapped entities:

 [OK]   DoctrineOrmTest\Base
 [OK]   DoctrineOrmTest\Concrete
 [OK]   DoctrineOrmTest\Intermediate
```

ghost:
We need to make sure to query the discriminator map only with contained classes, otherwise we end up trying to access it with e.g. abstract sub-classes of the current class, which are not to be contained in the discriminator map. This is caused by the introduction of auto-discovering of missing sub-classes in https://github.com/doctrine/orm/pull/10411 which also added abstract classes to ClassMetadata::subClasses.

Fixes #10625
Test based on PR https://github.com/doctrine/orm/pull/10411

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/SingleTablePersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/SingleTablePersister.php
@@ -10,6 +10,9 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Utility\PersisterHelper;
 
 use function array_flip;
+use function array_intersect;
+use function array_map;
+use function array_unshift;
 use function implode;
 
 /**
@@ -145,16 +148,13 @@ class SingleTablePersister extends AbstractEntityInheritancePersister
     /** @return string */
     protected function getSelectConditionDiscriminatorValueSQL()
     {
-        $values = [];
+        $values = array_map(
+            [$this->conn, 'quote'],
+            array_flip(array_intersect($this->class->discriminatorMap, $this->class->subClasses))
+        );
 
         if ($this->class->discriminatorValue !== null) { // discriminators can be 0
-            $values[] = $this->conn->quote($this->class->discriminatorValue);
-        }
-
-        $discrValues = array_flip($this->class->discriminatorMap);
-
-        foreach ($this->class->subClasses as $subclassName) {
-            $values[] = $this->conn->quote($discrValues[$subclassName]);
+            array_unshift($values, $this->conn->quote($this->class->discriminatorValue));
         }
 
         $discColumnName = $this->class->getDiscriminatorColumn()['name'];
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10625Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10625Test.php
@@ -1,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+/**
+ * @group GH-10625
+ */
+class GH10625Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->createSchemaForModels(
+            GH10625Root::class,
+            GH10625Middle::class,
+            GH10625Leaf::class
+        );
+    }
+
+    /**
+     * @dataProvider queryClasses
+     */
+    public function testLoadFieldsFromAllClassesInHierarchy(string $queryClass): void
+    {
+        $entity = new GH10625Leaf();
+
+        $this->_em->persist($entity);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $loadedEntity = $this->_em->find($queryClass, $entity->id);
+
+        self::assertNotNull($loadedEntity);
+        self::assertInstanceOf(GH10625Leaf::class, $loadedEntity);
+    }
+
+    public static function queryClasses(): array
+    {
+        return [
+            'query via root entity' => [GH10625Root::class],
+            'query via intermediate entity' => [GH10625Middle::class],
+            'query via leaf entity' => [GH10625Leaf::class],
+        ];
+    }
+}
+
+/**
+ * @ORM\Entity
+ * @ORM\InheritanceType("SINGLE_TABLE")
+ * @ORM\DiscriminatorMap({ "1": "GH10625Leaf"})
+ * ^- This DiscriminatorMap contains the single non-abstract Entity class only
+ */
+abstract class GH10625Root
+{
+    /**
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
+     * @ORM\Column(type="integer")
+     *
+     * @var int
+     */
+    public $id;
+}
+
+/**
+ * @ORM\Entity
+ */
+abstract class GH10625Middle extends GH10625Root
+{
+}
+
+/**
+ * @ORM\Entity
+ */
+class GH10625Leaf extends GH10625Middle
+{
+}


lib/Doctrine/ORM/Persisters/Entity/SingleTablePersister.php
```diff
@@ -145,16 +148,13 @@ protected function getSelectConditionCriteriaSQL(Criteria $criteria)
...
         if ($this->class->discriminatorValue !== null) { // discriminators can be 0
-            $values[] = $this->conn->quote($this->class->discriminatorValue);
-        }
-
-        $discrValues = array_flip($this->class->discriminatorMap);
-
-        foreach ($this->class->subClasses as $subclassName) {
-            $values[] = $this->conn->quote($discrValues[$subclassName]);
+            array_unshift($values, $this->conn->quote($this->class->discriminatorValue));
         }
```
ghost, line 158: A different approach would be extending of ClassMetadata(Info) with something in the lines of `discriminatorSubClasses`. This would cause more work and refactoring, though.

mpdude (approved):
As the author of #10411 I can approve this.

Before #10411, _all_ subclasses had to be declared in the discriminator map, or you'd get strange errors elsewhere.

Since #10411, abstract entity classes may be omitted in the discriminator map, and they'd be filled into `\Doctrine\ORM\Mapping\ClassMetadataInfo::$subClasses` automatically. That, in turn, means there may be `subClasses` without corresponding discriminator map entries.

We cannot look at the discriminator map only, since it contains all classes from the inheritance tree. In this situation here, we are only looking for those at or below the current class, so we need to intersect both `subClasses` and the `discriminatorMap`.

ghost:
I'm not quite sure, that _all_ subclasses had to be in the discriminator map before https://github.com/doctrine/orm/pull/10411 when the documentation only included concrete implementations itself, e.g. in [1]. On the other hand the annotation documentation is a bit vague [2]
>  Its only argument is an array which defines which class should be saved under which name in the database.

Though this doesn't address abstract classes, it only refers to database backed entities, which IMHO shouldn't be abstract anyway.

[1] https://www.doctrine-project.org/projects/doctrine-orm/en/2.12/cookbook/decorator-pattern.html
[2] https://www.doctrine-project.org/projects/doctrine-orm/en/2.12/reference/annotations-reference.html#discriminatormap

mpdude:
Having all classes in the DM was a recommendation given (eg at https://github.com/doctrine/orm/issues/8736#issuecomment-853044536) to work around some quirks when not all subclasses were in the `$subClasses` list. Documentation and some exception messages in the codebase were contradictory on this.

#10411 was a change to fill in the missing subclasses automatically, so that only non-abstract entity classes need to be listed in the DM.

ghost:
Ah, thanks for the explanation and clarification. We probably just had luck for all these years then :see_no_evil:

ghost:
@mpdude Hey @mpdude, since the PR sits around for four weeks by now, do you by chance have any idea if I need to mention/add someone to the PR and if so, who?

mpdude:
@greg0ire can you help us to get this along?

Changes are reasonable from my POV. 

Do you see anything that should be addressed?

greg0ire (approved):

greg0ire:
Thanks @htto! And thanks @mpdude for the review!

greg0ire:
Argh! The target branch was wrong. Let me create a merge up PR

ghost:
@mpdude @greg0ire Thanks for caring, much appreciated.

PR #10645, opened 2023-04-21
Restore document proxy state to uninitialized on load exception

notrix:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary
I had an issue in mongodb-odm bundle with proxy left in initialized state on load exception and thought if same could be applied in orm bundle. Test shows same issue is here.

#### Related
https://github.com/doctrine/mongodb-odm/pull/2521

### DIFF
--- a/tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
@@ -21,6 +21,7 @@ use Doctrine\Tests\Models\Company\CompanyPerson;
 use Doctrine\Tests\Models\ECommerce\ECommerceFeature;
 use Doctrine\Tests\OrmTestCase;
 use Doctrine\Tests\PHPUnitCompatibility\MockBuilderCompatibilityTools;
+use Exception;
 use ReflectionProperty;
 use stdClass;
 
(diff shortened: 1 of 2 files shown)


zilionis (approved):

Pionell (approved):

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
```diff
@@ -145,6 +146,38 @@ public function testFailedProxyLoadingDoesNotMarkTheProxyAsInitialized(): void
...
+            ->getMock();
+        $this->uowMock->setEntityPersister(ECommerceFeature::class, $persister);
+
+        $proxy = $this->proxyFactory->getProxy(ECommerceFeature::class, ['id' => 42]);
+        assert($proxy instanceof Proxy);
+
+        $exception = $this
+            ->getMockBuilder(ConnectionException::class)
+            ->disableOriginalConstructor()
+            ->getMock();
```
derrabus: Do we really need to mock an exception class?

derrabus (commented):
Sorry that this PR hasn't been picked up yet. You change looks reasonable.

Can you give a real-world example of what kind of failure could happen here? I wonder if we could build a functional test for this scenario. All this mocking in your test makes the scenario look artificial/hypothetical.

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
notrix: Well `ConnectionException` was the closest exception to the real-world scenario. And as it has a nasty constructor I have decided to mock it. It could be less complicated 👍 
And for the example: in my experience we have never encountered this with ORM, but with ODM and native `Client-Side Field Level Encryption` we had this issue. As our supervisor consumers catches all exceptions and keep working on messages. On `em->flush()` proxies that were not initialized has been flushed to db removing all fields that had no default values.
After I made PR to ODM lib decided to check if the same is possible with ORM also. And theoretically it is possible.

derrabus (approved):

PR #10649, opened 2023-04-23
Use valid relationships

greg0ire:
During a recent refactoring, I had to pick a relationship type for this piece of code, and I chose wrong, because a many-to-one cannot have a `mappedBy` field.


SenseException (approved):
Maybe the psalm/phpstan array annotation of `fromMappingArray` can be improved. In `AssociationMapping` the keys are defined like

```
    /**
     * @psalm-param array{
     *     fieldName: string,
     *     sourceEntity: class-string,
     *     targetEntity: class-string,
     *     joinTable?: mixed[]|null,
     *     type?: int,
     *     isOwningSide: bool, ...} $mappingArray
     */
```

and in `OneToManyAssociationMapping` it's `/** @param mixed[] $mappingArray */`.

PR #10651, opened 2023-04-24
Prepare entity-level commit order computation in the `UnitOfWork`

mpdude:
This is the second chunk to break #10547 into smaller PRs suitable for reviewing. It prepares the `UnitOfWork` to work with a commit order computed on the entity level, as opposed to a class-based ordering as before.

#### Background

#10531 and #10532 show that it is not always possible to run `UnitOfWork::commit()` with a commit order given in terms of  entity _classes_. Instead it is necessary to process entities in an order computed on the _object_ level. That may include

* a particular order for multiple entities of the _same_ class
* a particular order for entities of _different_ classes, possibly even going back and forth (entity `$a1` of class `A`, then `$b` of class `B`, then `$a2` of class `A` – revisiting that class).

This PR is about preparing the `UnitOfWork` so that its methods will be able to perform inserts and deletions on that level of granularity. Subsequent PRs will deal with implementing that particular order computation.

#### Suggested change

Change the private `executeInserts` and `executeDeletions` methods so that they do not take a `ClassMetadata` identifying the class of entities that shall be processed, but pass them the list of entities directly.

The lists of entities are built in two dedicated methods. That happens basically as previously, by iterating over `$this->entityInsertions` or `$this->entityDeletions` and filtering those by class.

#### Potential (BC breaking?) changes that need review scrutiny

* `\Doctrine\ORM\Persisters\Entity\EntityPersister::addInsert()` was previously called multiple times, before the insert would be performed by a call to `\Doctrine\ORM\Persisters\Entity\EntityPersister::executeInserts()`. With the changes here, this batching effectively no longer takes place. `executeInserts()` will always see one entity/insert only, and there may be multiple `executeInserts()` calls during a single `UoW::commit()` phase.
* The caching in `\Doctrine\ORM\Cache\Persister\Entity\AbstractEntityPersister` previously would cache entities from the last executed insert batch only. Now it will collect entities across multiple batches. I don't know if that poses a problem.
* Overhead in `\Doctrine\ORM\Persisters\Entity\BasicEntityPersister::executeInserts` is incurred multiple times; that may, however, only be about SQL statement preparation and might be salvageable.
* The `postPersist` event previously was not dispatched before _all_ instances of a particular entity class had been written to the database. Now, it will be dispatched immediately after every single entity that has been inserted.

#### `postPersist` event changes – meaningful?

Documentation on the `postPersist` event is here: https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/events.html#postupdate-postremove-postpersist

> The postPersist event occurs for an entity after the entity has been made persistent. It will be invoked after the database insert operations. Generated primary key values are available in the postPersist event.

"It will be invoked after the database insert operations" is a bit imprecise. It previously was "after the database insert operations for that entity class, but before operations for other classes". With this PR, it would be "after the database insert operations for this entity instance".

"Generated primary key values are available in the postPersist event" should better be written as "a generated primary key value for the entity is available in the postPersist event".

Alternatively, we could defer `postPersist` events until _all_ new entities have been inserted. Either way, it's a slight change from the previous implementation where the event would run after all entities of a particular class had been processed.

#### Extra bonus

This is what the DALL·E AI thinks it looks like when an entity persister will have to process entities one-by-one and is no longer able to batch the inserts.

DALL·E 2023-04-24 14 07 14

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -707,8 +707,8 @@ not directly mapped by Doctrine.
    ``UPDATE`` statement.
 -  The ``postPersist`` event occurs for an entity after
    the entity has been made persistent. It will be invoked after the
-   database insert operations. Generated primary key values are
-   available in the postPersist event.
+   database insert operation for that entity. A generated primary key value for
+   the entity will be available in the postPersist event.
 -  The ``postRemove`` event occurs for an entity after the
    entity has been deleted. It will be invoked after the database
    delete operations. It is not called for a DQL ``DELETE`` statement.
--- a/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
+++ b/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
@@ -23,6 +23,7 @@ use Doctrine\ORM\PersistentCollection;
 use Doctrine\ORM\Persisters\Entity\EntityPersister;
 use Doctrine\ORM\UnitOfWork;
 
+use function array_merge;
 use function assert;
 use function serialize;
 use function sha1;
@@ -314,7 +315,13 @@ abstract class AbstractEntityPersister implements CachedEntityPersister
      */
     public function executeInserts()
     {
-        $this->queuedCache['insert'] = $this->persister->getInserts();
+        // The commit order/foreign key relationships may make it necessary that multiple calls to executeInsert()
+        // are performed, so collect all the new entities.
+        $newInserts = $this->persister->getInserts();
+
+        if ($newInserts) {
+            $this->queuedCache['insert'] = array_merge($this->queuedCache['insert'] ?? [], $newInserts);
+        }
 
         return $this->persister->executeInserts();
     }
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -57,10 +57,10 @@ use function array_key_exists;
 use function array_map;
 use function array_merge;
 use function array_pop;
+use function array_reverse;
 use function array_sum;
 use function array_values;
 use function assert;
-use function count;
 use function current;
 use function func_get_arg;
 use function func_num_args;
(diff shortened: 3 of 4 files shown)


mpdude:
Documentation on the `postPersist` event is here: https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/events.html#postupdate-postremove-postpersist

What we will not be able to do anymore is to run `postPersist` events for _all_ entities of a particular class _before_ we start inserting entities of another class (after all, this ordering is not always possible).

We could:

* dispatch the event immediately after the particular entity has been inserted
* defer all events until all entities have been inserted.

WDYT?

mpdude:
@FabioBatSilva you were the last one to make relevant API changes to the persister classes in #808.

Maybe you still remember those days and have any thoughts on this?

FabioBatSilva:
> @FabioBatSilva you were the last one to make relevant API changes to the persister classes in #808.
> 
> Maybe you still remember those days and have any thoughts on this?

Wow.. Can't believe that i'ts been almost 10y since that PR :scream: 

I'ts been a while since i looked into that code,
But Looks good as far as i can tell.

mpdude:
@FabioBatSilva thanks a lot for your feedback. Yes, time flies...

In particular, with regard to 2nd level caching, do you see any problem with calling `\Doctrine\ORM\Persisters\Entity\EntityPersister::executeInserts` multiple times before `\Doctrine\ORM\Cache\Persister\CachedPersister::afterTransactionComplete` is being called?

FabioBatSilva:
Is `getInserts` still return the entire set ?

This bit here might be problematic :
https://github.com/doctrine/orm/blob/1142a396d36e0c9b33ff13875f2833d8386d8f66/lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php#L317-L318
https://github.com/doctrine/orm/blob/1142a396d36e0c9b33ff13875f2833d8386d8f66/lib/Doctrine/ORM/Cache/Persister/Entity/NonStrictReadWriteCachedEntityPersister.php#L23

IIRC it assumes the all inserts are present when `afterTransactionComplete` is called.
If `executeInserts` is called multiple times it might be resetting the queue and `afterTransactionComplete` will only see the last set of entries.

mpdude:
I think I’ve got this covered with the changes to lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php in this PR.

greg0ire (approved):

SenseException (approved):
Is `entity-level-commit-order` going to be merged into the 3.0.x branch? If the impact of the changes isn't graspable, do we miss tests for the commit order?

mpdude:
@SenseException Thank your for taking the time to review this!

`entity-level-commit-order` is a temporary branch to break #10547 down into manageable stories. Privately, I was hoping that we could merge it into 2.15 (or whatever the next 2.x release is). Since it might include deprecations and/or there is always a risk associated with such changes, no need to rush it and surprise users in a bugfix release.

Reasons (in my opinion) _not_ to aim at 3.0:

* The entire goal of #10547 is to fix a long list of bugs (see the list over there), _without_ requiring changes on the users' side.
* In case we realize that there are releant changes (causing breakage), we should not include them in 3.0 in a "like it or lump it" style, but instead find ways so that users can opt-in as simple and as soon as possible, i. e. in 2.x. 

That being said, yes - there always remains a risk that changes break things for users, and I try hard to come up with creative scenarios to imagine how that might happen.

But, on the other hand, we also need to make clear where users overstressed our promises given. One example is the semantics of the `postPersist` event I mentioned in the initial comment of this PR. Clearly, it is supposed to be emitted after a particular entity has been persisted – but we did not mention anything about the order in which these events happen, or if one may assume _all_ persist operations have happened at the time the _first_ event is being emitted.

In these cases, we'll need to discuss/decide what the reasonable assumptions are that one might have made, and where users (unknowingly?) relied on ~accidental~ implementation details only. I hope that, in the long run, we can improve specifications, tests, documentation and the like when we find such places. Hopefully, this makes it more clear for users what they may expect, and makes it more clear for us what may be changed.

Hm, did that answer your question 🤔 ?

Regarding test coverage:

The commit order is very central for everything that writes entities to the DB, so almost every functional test we have depends on it in one way or another. 

For the tricky edge cases that currently do not work right (reported by many users over the years), I have created test cases in various other PRs, but all of them are consolidated in #10547.

SenseException:
> One example is the semantics of the postPersist event I mentioned in the initial comment of this PR. Clearly, it is supposed to be emitted after a particular entity has been persisted – but we did not mention anything about the order in which these events happen, or if one may assume all persist operations have happened at the time the first event is being emitted.

Okay, I was afraid that this would introduce changes that would only fit into a major release. Sounds like it is not. 

> Hm, did that answer your question thinking ?

Yes, it did. Thank you.

greg0ire:
Thanks @mpdude !

PR #10652, opened 2023-04-24
Remove JoinColumn from inverse side

greg0ire:
This makes no sense because it describes a one-to-one where each table references the other one. I do not think this is deliberately supported, and it probably will not be supported at all in 3.0.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
+++ b/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
@@ -1929,7 +1929,7 @@ class SelectSqlGenerationTest extends OrmTestCase
     {
         $this->assertSqlGeneration(
             'SELECT u, a FROM Doctrine\Tests\Models\Quote\User u JOIN u.address a',
-            'SELECT q0_."user-id" AS userid_0, q0_."user-name" AS username_1, q1_."address-id" AS addressid_2, q1_."address-zip" AS addresszip_3, q1_.type AS type_4 FROM "quote-user" q0_ INNER JOIN "quote-address" q1_ ON q0_."address-id" = q1_."address-id" AND q1_.type IN (\'simple\', \'full\')'
+            'SELECT q0_."user-id" AS userid_0, q0_."user-name" AS username_1, q1_."address-id" AS addressid_2, q1_."address-zip" AS addresszip_3, q1_.type AS type_4 FROM "quote-user" q0_ INNER JOIN "quote-address" q1_ ON q0_."user-id" = q1_."user-id" AND q1_.type IN (\'simple\', \'full\')'
         );
 
         $this->assertSqlGeneration(
--- a/tests/Doctrine/Tests/Models/Quote/User.php
+++ b/tests/Doctrine/Tests/Models/Quote/User.php
@@ -45,7 +45,6 @@ class User
 
     /**
      * @var Address
-     * @JoinColumn(name="`address-id`", referencedColumnName="`address-id`")
      * @OneToOne(targetEntity="Address", mappedBy="user", cascade={"persist"}, fetch="EAGER")
      */
     public $address;
--- a/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Mapping/ClassMetadataFactoryTest.php
@@ -385,11 +385,6 @@ class ClassMetadataFactoryTest extends OrmTestCase
         self::assertEquals('user-id', $userMetadata->fieldMappings['id']['columnName']);
         self::assertEquals('user-name', $userMetadata->fieldMappings['name']['columnName']);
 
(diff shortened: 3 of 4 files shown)


greg0ire:
@mpdude please review :pray:

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
```diff
@@ -1929,7 +1929,7 @@ public function testQuotedWalkJoinVariableDeclaration(): void
     {
         $this->assertSqlGeneration(
             'SELECT u, a FROM Doctrine\Tests\Models\Quote\User u JOIN u.address a',
-            'SELECT q0_."user-id" AS userid_0, q0_."user-name" AS username_1, q1_."address-id" AS addressid_2, q1_."address-zip" AS addresszip_3, q1_.type AS type_4 FROM "quote-user" q0_ INNER JOIN "quote-address" q1_ ON q0_."address-id" = q1_."address-id" AND q1_.type IN (\'simple\', \'full\')'
+            'SELECT q0_."user-id" AS userid_0, q0_."user-name" AS username_1, q1_."address-id" AS addressid_2, q1_."address-zip" AS addresszip_3, q1_.type AS type_4 FROM "quote-user" q0_ INNER JOIN "quote-address" q1_ ON q0_."user-id" = q1_."user-id" AND q1_.type IN (\'simple\', \'full\')'
```
mpdude, line 1932: Why does that change? Are we using an association from the non-owning side in this test?

Can we change this test so that it still covers configuration read from a `@JoinColumn` annotation, maybe using the other side of the association?

greg0ire: > Are we using an association from the non-owning side in this test?

Yes I think we are. Here, we are establishing a join with `address-id`, which should not even be allowed to exist in the first place IMO.

> Can we change this test so that it still covers configuration read from a @JoinColumn annotation, maybe using the other side of the association?

I'm not sure what exactly you suggest, or if it should even be allowed. Is it valid to have such an annotation when not on the owning side?

greg0ire: BTW, if the goal of this test is to ensure the columns used to perform the join are quoted, that's the case, so…

SenseException (approved):

derrabus (approved):

mpdude (approved):
:+1:, and I'll take a look at the one test independently

greg0ire:
@derrabus regarding the milestone, I was hoping to release 2.15.0 soon. Can you talk with @beberlei to resolve the issue with the PR remaining in that milestone 🙏?

tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php
mpdude: When you remove the `@JoinColumn` from `User::$address`, which is the inverse side of the association (since it uses `mappedBy`), and we all think that the inverse side does not matter at all... _why_ does the SQL change then?

mpdude (changes requested):
Wait – we do think that `@JoinColumn` does not matter on the inverse side of an association, which for `OneToOne` is the side using `mappedBy`, right?

greg0ire:
That my opinion, and I think at least @beberlei shares it.

mpdude:
That way my opintion and knowledge so far as well, but it seems the implementation does not behave that way.

Hold your horses :-)

mpdude:
https://github.com/doctrine/orm/blob/1ea8424e07463c6cc8a563542139ed808204cd76/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L1958-L1960

When there is a `@JoinColumn` given, it will treat this side as the owning side?

greg0ire:
Yes, which means here we have 2 owning sides, but there is nothing (yet) that throws if you have a `mappedBy` attribute on the owning side, so really, this does not make sense to me.

Also, the docs don't mention this as a possibility: https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/association-mapping.html#one-to-one-bidirectional

mpdude (approved):
Go ahead, merge this one. It seems we all agree that `JoinColumn` makes no sense on the inverse side of a `OneToOne` association.

I have created #10654 where we can deprecate this (bogus?) behaviour, then drop it in 3.0

derrabus:
> @derrabus regarding the milestone, I was hoping to release 2.15.0 soon.

Me too, but you're targeting 2.14, so the milestone is the next 2.14.x release. If we decide that it won't happen, we can still reassign all PRs of that milestone to 2.15.0.

greg0ire:
> Me too, but you're targeting 2.14, so the milestone is the next 2.14.x release. If we decide that it won't happen, we can still reassign all PRs of that milestone to 2.15.0.

Yes, I forgot about that when opening the MR. Since this is just about the tests, it does not absolutely need to be in a milestone anyway.

PR #10654, opened 2023-04-25
Deprecate usage of `@JoinColumn` on the inverse side of one-to-one associations

mpdude:
Following up on #10652:

#### Current situation

The implementation of `\Doctrine\ORM\Mapping\ClassMetadataInfo::_validateAndCompleteOneToOneMapping` will consider a field with a one-to-one association to be the owning side also when it configures `@JoinColumn` settings.

#### Suggested change

For a one to one association, a field should be the inverse side when it uses the `mappedBy` attribute, and be the owning side otherwise. The `JoinColumn` may be configured on the owning side only.

This PR adds a deprecation notice when `@JoinColumn` is used on the side of a one-to-one association where `mappedBy` occurs.

In 3.0, this will throw a `MappingException`.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
@@ -1864,6 +1864,14 @@ class ClassMetadataInfo implements ClassMetadata
     {
         $mapping = $this->_validateAndCompleteAssociationMapping($mapping);
 
+        if (isset($mapping['joinColumns']) && $mapping['joinColumns'] && ! $mapping['isOwningSide']) {
+            Deprecation::trigger(
+                'doctrine/orm',
+                'https://github.com/doctrine/orm/pull/10654',
+                'JoinColumn configuration is not allowed on the inverse side of one-to-one associations, and will throw a MappingException in Doctrine ORM 3.0'
+            );
+        }
+
         if (isset($mapping['joinColumns']) && $mapping['joinColumns']) {
             $mapping['isOwningSide'] = true;
         }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,12 @@
 # Upgrade to 2.15
 
+## Deprecated configuring `JoinColumn` on the inverse side of one-to-one associations
+
+For one-to-one associations, the side using the `mappedBy` attribute is the inverse side.
+The owning side is the entity with the table containing the foreign key. Using `JoinColumn`
+configuration on the _inverse_ side now triggers a deprecation notice and will be an error
+in 3.0.
+
 ## Deprecated overriding fields or associations not declared in mapped superclasses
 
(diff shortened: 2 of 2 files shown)


mpdude:
Tests will pass with the changes from #10652.

lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -345,6 +345,11 @@ public static function joinColumnMustPointToMappedField($className, $joinColumn)
         ));
     }
 
+    public static function joinColumnNotAllowedOnInverseSide(string $className, string $fieldName): self
+    {
+        return new self(sprintf('The %s#%s field is not the owning side of the association, so it may not configure join column(s)', $className, $fieldName));
```
derrabus:
```suggestion
        return new self(sprintf('The %s#%s field is not the owning side of the association, so it must not configure join column(s)', $className, $fieldName));
```

lib/Doctrine/ORM/Mapping/MappingException.php
```diff
@@ -345,6 +345,11 @@ public static function joinColumnMustPointToMappedField($className, $joinColumn)
         ));
     }
 
+    public static function joinColumnNotAllowedOnInverseSide(string $className, string $fieldName): self
```
derrabus: This method won't be used in 2.x, will it? Let's not introduce it before 3.0, then.

mpdude: fixed

derrabus:
Something went wrong when rebasing this PR, apparently. The diff contains changes that have already been merged to 2.15.x.

mpdude:
Fixed the rebase & trying to resolve Psalm errors – hope removing baseline entries is the right thing to do.

lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php
```diff
@@ -1864,8 +1864,12 @@ protected function _validateAndCompleteOneToOneMapping(array $mapping)
     {
         $mapping = $this->_validateAndCompleteAssociationMapping($mapping);
 
-        if (isset($mapping['joinColumns']) && $mapping['joinColumns']) {
-            $mapping['isOwningSide'] = true;
+        if (isset($mapping['joinColumns']) && $mapping['joinColumns'] && ! $mapping['isOwningSide']) {
```
greg0ire, line 1867: I think this condition should be used on 3.0 to throw. On 2.x though, I wouldn't touch the condition and keep setting `isOwningSide`, but I would check the value of `isOwningSide` before setting it, and trigger a deprecation then.

mpdude: Totally agree, did not intend to change logic in any way. Thanks for spotting this!

greg0ire (approved):

greg0ire:
Thanks @mpdude !

PR #10662, opened 2023-04-27
Adds metadata field type validation against Entity property type

### LINKED ISSUE #10661: Metadata field type validation against Entity property type
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    |  no

#### Summary

We found out that if the metadata field type differs from the entity property type Doctrine marks it as "to be updated" right after it being loaded from the database and hydrated, even if no real change happened to the entity. If there's a flush an update is performed for every entity marked as "to be updated", as it should happen.

For example if this is the metadata (Xml in this case)
MyEntity.xml:
```xml
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping https://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
  <entity name="MyEntity" table="my_table">
    <id name="id" .../>
    <field name="a" type="boolean" column="a_column" nullable="false"/>
  </entity>
</doctrine-mapping>
```
MyEntity.php:
```php
class MyEntity
{
    public function __construct(private int $a)
    {
    }
}
```

This code updates to the database
```php
/** @var EntityManagerInterface $em */
$myEntity = $em->getRepository(MyEntity::class)->find($myEntityId);
$em->flush();
```

Please note that $a is `int` in MyEntity.php and `boolean` in MyEntity.xml

It would be great to have some strictness when setting the Entity value during Hydration or some field type validation in `Doctrine\ORM\Tools\SchemaValidator`.

DavideBicego:
This adds some properties type validation to `SchemaValidator`.

I think some strictness when setting the Entity value during Hydration would be better but `ReflectionProperty::setValue` skips strictness validation.

Resolve https://github.com/doctrine/orm/issues/10661


Ocramius:
Note: potentially worth going to `2.15.x`, but unsure what the maintainers are planning for the 2.x branch, especially in the direction of adding features.

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
```diff
@@ -36,12 +57,47 @@ final class DefaultTypedFieldMapper implements TypedFieldMapper
...
+        GuidType::class => 'string',
+        IntegerType::class => 'int',
+        JsonType::class => 'array',
+        SimpleArrayType::class => 'array',
+        SmallIntType::class => 'int',
+        StringType::class => 'string',
+        TextType::class => 'string',
+        TimeType::class => DateTime::class,
+        TimeImmutableType::class => DateTimeImmutable::class,
+    ];
```
Ocramius: Discussed this with @DavideBicego in a preparation PR: the idea here is that the **exact** DBAL type must be used (no subclasses), since userland consumers of `doctrine/orm` may have their own customization around field types.

It is worth documenting such a detail in the docblock.

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
```diff
@@ -36,12 +57,47 @@ final class DefaultTypedFieldMapper implements TypedFieldMapper
...
 
+    /**
+     * @internal
+     *
+     * @param class-string<Type> $typeName
+     */
+    public function getBuiltInType(string $typeName): string
+    {
+        return self::BUILTIN_TYPES_MAP[$typeName] ?? $typeName;
+    }
```
Ocramius: Perhaps, instead of exposing the type, we could ask a boolean question `typeConformsType()` or such?

Hard to say in this scope, so the naming is a bit wonky, but just an idea.

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
```diff
@@ -36,12 +57,47 @@ final class DefaultTypedFieldMapper implements TypedFieldMapper
...
     /** @param array<class-string|ScalarName, class-string<Type>|string> $typedFieldMappings */
     public function __construct(array $typedFieldMappings = [])
     {
         $this->typedFieldMappings = array_merge(self::DEFAULT_TYPED_FIELD_MAPPINGS, $typedFieldMappings);
     }
 
+    /**
+     * @internal
+     *
+     * @param class-string<Type> $typeName
```
Ocramius: Perhaps just `Type`, rather than `class-string<Type>`?

tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
```diff
@@ -675,3 +700,31 @@ abstract class Issue9095AbstractChild extends Issue9095Parent
...
+     * @Column(type="boolean")
+     */
+    protected int $property;
+}
+
+/** @Entity */
+class InvalidChildEntity extends InvalidEntity3
+{
+    /** @Column(type="string") */
+    protected int $property;
```
Ocramius: Note: `doctrine/orm` does not (AFAIK) allow parent/child properties with the same name. The suggestion I sent you before to add an example with inheritance was mostly about having parent class properties that are **not** visible to the child class

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -298,4 +308,47 @@ public function getUpdateSchemaList(): array
...
+                            "The field '%s#%s' has the property type '%s' that differs from the metadata field type '%s'.",
+                            $class->name,
+                            $fieldName,
+                            $propertyType->getName(),
+                            $metadataFieldType,
+                        );
+                    },
+                    $class->fieldMappings,
+                ),
+                static fn (?string $issue) => $issue !== null,
```
Ocramius: This closure can be omitted, since the default behavior of `array_filter()` is to remove empty-ish values, like `null`, in this case

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -298,4 +308,47 @@ public function getUpdateSchemaList(): array
...
+    /** @return list<string> containing the found issues */
+    private function validatePropertiesTypes(ClassMetadataInfo $class): array
+    {
+        return array_values(
+            array_filter(
+                array_map(
+                    /** @param FieldMapping $fieldMapping */
+                    static function (array $fieldMapping) use ($class): string|null {
+                        $fieldName    = $fieldMapping['fieldName'];
+                        $propertyType = $class->reflFields[$fieldName]?->getType();
```
Ocramius: Is the `?` needed here? I'd expect a field to always exist :thinking:

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -298,4 +308,47 @@ public function getUpdateSchemaList(): array
...
+                        if (! Type::hasType($fieldMapping['type'])) {
+                            return null;
+                        }
+
+                        // If the property type is not a named type, we cannot check it
+                        if (! ($propertyType instanceof ReflectionNamedType)) {
+                            return null;
+                        }
+
+                        $metadataFieldType = (new DefaultTypedFieldMapper())->getBuiltInType(get_class(Type::getType($fieldMapping['type'])));
```
Ocramius: Does the metadata not have a reference to the `Type` instance? Do we have to go through `Type::getType()`?

Ocramius: Could the `DefaultTypedFieldMapper` be kept in memory as a `private` property? A bit wasteful to instantiate it all the time, but also no big deal.

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
```diff
@@ -36,12 +57,47 @@ final class DefaultTypedFieldMapper implements TypedFieldMapper
...
+        FloatType::class => 'float',
+        GuidType::class => 'string',
+        IntegerType::class => 'int',
+        JsonType::class => 'array',
+        SimpleArrayType::class => 'array',
+        SmallIntType::class => 'int',
+        StringType::class => 'string',
+        TextType::class => 'string',
+        TimeType::class => DateTime::class,
+        TimeImmutableType::class => DateTimeImmutable::class,
```
Ocramius: Careful: in these cases, `DateTimeInterface` is perfectly valid!

I suggest skipping on any types where objects are de-serialized

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
```diff
@@ -36,12 +57,47 @@ final class DefaultTypedFieldMapper implements TypedFieldMapper
...
+        AsciiStringType::class => 'string',
+        BigIntType::class => 'string',
+        BooleanType::class => 'bool',
+        DateType::class => DateTime::class,
+        DateImmutableType::class => DateTimeImmutable::class,
+        DateIntervalType::class => DateInterval::class,
+        DateTimeType::class => DateTime::class,
+        DateTimeImmutableType::class => DateTimeImmutable::class,
+        DateTimeTzType::class => DateTime::class,
+        DateTimeTzImmutableType::class => DateTimeImmutable::class,
```
Ocramius: Careful: in these cases, `DateTimeInterface` is perfectly valid!

I suggest skipping on any types where objects are de-serialized

tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
```diff
@@ -675,3 +700,31 @@ abstract class Issue9095AbstractChild extends Issue9095Parent
...
+     * @var int
+     * @Id
+     * @Column
+     */
+    protected $key;
+
+    /**
+     * @Column(type="boolean")
+     */
+    protected int $property;
```
Ocramius: Ooooops, the test fails here because Doctrine didn't yet drop PHP 7.3 support (REALLY?!)

I think the best way to handle this would be to:

 1. split this out to a separate file
 2. mark the tests with `@requires php>=7.4`
 3. exclude the separated file from analysis by tooling like phpcs/phpstan

Ocramius (changes requested):

greg0ire:
> Note: potentially worth going to 2.15.x, but unsure what the maintainers are planning for the 2.x branch, especially in the direction of adding features.

Features welcome, and indeed, this should target 2.15.x (2.15.0 should be released quite soon BTW)

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
Ocramius: I think this is a detail for which we should write a small regression test too, to prevent misuse in future refactoring attempts.

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
Ocramius: Perhaps the best approach remains to keep this `private` in the validator itself, for now, rather than having it in `DefaultTypedFieldMapper`

tests/Doctrine/Tests/ORM/Tools/SchemaValidatorTest.php
DavideBicego: How would you prevent the file from being read with php < 7.4?
The test should not be executed with a `@requires php 70400` but the file contains some classes definition too...

Ocramius: I don't remember how the test suite works precisely, but can check it tomorrow.

I'd say something like `require_once __DIR__ . '/some-php-7.4-only-file.php` and then reference the classes in that file that way

greg0ire: Or use autoload, or team up with @MatTheCat who is having the same issue on https://github.com/doctrine/orm/pull/10666/files#r1186698321

We could have a subdirectories with tests that require a specific version of PHP, and then future contributors would not need to bother with `@requires` or with splitting test files and companion classes.

DavideBicego:
For some reason the [CI step "PHPUnit with SQLite (7.1)"](https://github.com/KanbanBOX/orm/actions/runs/4948681585/jobs/8849759918) ignores the PHP version requirement...

Ocramius:
CI is also stuck due to https://www.githubstatus.com/

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -4,33 +4,73 @@
...
+        DecimalType::class => 'string',
+        FloatType::class => 'float',
+        GuidType::class => 'string',
+        IntegerType::class => 'int',
+        JsonType::class => 'array',
+        SimpleArrayType::class => 'array',
+        SmallIntType::class => 'int',
+        StringType::class => 'string',
+        TextType::class => 'string',
+    ];
```
greg0ire, line 72: What I don't like about this PR is this harcoded map, but for now there is nothing better I'm afraid. Once `doctrine/dbal` 4 is published and supported, we can start using reflection to get the return type.

greg0ire:
Triggered the CI again

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10661/GH10661Test.php
```diff
@@ -0,0 +1,51 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket\GH10661;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Tools\SchemaValidator;
+use Doctrine\Tests\OrmTestCase;
+
+/** @requires PHP >= 7.4 */
```
greg0ire, line 11: Other occurrences of this in the project are multiline. Maybe old versions of PHPUnit ignore single line stuff. The next suspect will be `>=`

greg0ire: Ah I can't push… push from maintainers is disabled. I'll open another PR I guess.

greg0ire: Looks like that was the reason: https://github.com/doctrine/orm/pull/10946 :)

greg0ire:
Closing in favor of #10946

PR #10665, opened 2023-05-01
Rename AbstractSqlExecutor class to AbstractSQLExecutor

### LINKED ISSUE #10664: Rename AbstractSqlExecutor to AbstractSQLExecutor
For maintenance purposes I've been checking in the todos and found that this: https://github.com/doctrine/orm/blob/f82485e651763fbd1b34879726f4d3b91c358bd9/lib/Doctrine/ORM/Query/Exec/AbstractSqlExecutor.php#L17
I believe this could be an improvement of naming convention for this class.

abdelazizmirasoft:
Resolve https://github.com/doctrine/orm/issues/10664
This task was in the todo list of the and can be considered as improvement/maintenance PR:
https://github.com/doctrine/orm/blob/f82485e651763fbd1b34879726f4d3b91c358bd9/lib/Doctrine/ORM/Query/Exec/AbstractSqlExecutor.php#L17

### DIFF
--- a/lib/Doctrine/ORM/Query.php
+++ b/lib/Doctrine/ORM/Query.php
@@ -17,7 +17,7 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Query\AST\DeleteStatement;
 use Doctrine\ORM\Query\AST\SelectStatement;
 use Doctrine\ORM\Query\AST\UpdateStatement;
-use Doctrine\ORM\Query\Exec\AbstractSqlExecutor;
+use Doctrine\ORM\Query\Exec\AbstractSQLExecutor;
 use Doctrine\ORM\Query\Parameter;
 use Doctrine\ORM\Query\ParameterTypeInferer;
 use Doctrine\ORM\Query\Parser;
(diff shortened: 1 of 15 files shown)


derrabus (changes requested):
Thank you for your attempt at improving the codebase, but renaming a class is a breaking change. We cannot do this in a bugfix release.

abdelazizmirasoft:
> We cannot do this in a bugfix release.

@derrabus Please when this could be done?

SenseException:
That would be in a major version, but I don't see anything wrong with using pascalcase `AbstractSqlExecutor`. Yes, we aren't consistent with that, having an all uppercase "ORM" in some class names, but we also have names like e.g. `SqlResultSetMapping` where it's "Sql" and not "SQL".

abdelazizmirasoft:
> That would be in a major version, but I don't see anything wrong with using pascalcase `AbstractSqlExecutor`.

Thank you for your feedback @SenseException I do agree, my goal was to resolve some task from the todo list which was set by one of the authors,..., I'm gonna find something more interesting anyway :)

PR #10668, opened 2023-05-03
Added doc for the fields parameter of the UniqueConstraint attribute

stollr:
The description for the *fields* parameter of the `UniqueConstraint` attribute is missing at the moment (see [here](https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/attributes-reference.html#uniqueconstraint)).

### DIFF
--- a/docs/en/reference/attributes-reference.rst
+++ b/docs/en/reference/attributes-reference.rst
@@ -1104,7 +1104,8 @@ context.
 Required parameters:
 
 -  **name**: Name of the Index
--  **columns**: Array of columns.
+-  **fields**: Array of fields (the names of the properties, used in the entity class).
+-  **columns**: Array of columns (the names of the columns, used in the schema).
 
 Optional parameters:
 


SenseException (approved):

SenseException:
It doesn't seem to be in use in the ORM code/tests.

docs/en/reference/attributes-reference.rst
```diff
@@ -1104,7 +1104,8 @@ context.
 Required parameters:
 
 -  **name**: Name of the Index
--  **columns**: Array of columns.
+-  **fields**: Array of fields (the names of the properties, used in the entity class).
```
greg0ire, line 1107: If it's not used in tests, how can it be a required parameter?

greg0ire:
This was introduced in https://github.com/doctrine/orm/pull/8629. This is, in fact, covered by tests.

docs/en/reference/attributes-reference.rst
greg0ire: :bulb: nevermind, it's required to provide either `columns` or `fields`.

greg0ire:
Thanks @stollr

PR #10681, opened 2023-05-05
Rework association mapping hierarchy

greg0ire:
Please review  https://github.com/doctrine/orm/pull/10682 before this one.

- Each type is now either final, abstract or an interface. Inverse mappings are no longer represented by intermediary types, which means you no longer have this weird inheritance between owning and inverse. 
- The mappedBy attribute is no longer nullable and moved down the hierarchy, because it makes sense only for inverse side mappings.
- The inversedBy attribute is still nullable and also moved down the hierarchy, because it makes sense only for owning side mappings.
- Code common to ManyToOneAssociationMapping and OneToOneOwningSideMapping is de-duplicated and moved up the hierarchy
- Code inside ToManyInverseSideMapping and ToManyOwningSideMapping comes from a trait (:nauseated_face:) to avoid duplication.

# Before

## Classes

```mermaid
classDiagram
    AssociationMapping <|-- ToManyAssociationMapping
    ToManyAssociationMapping <|-- ManyToManyAssociationMapping
    ManyToManyAssociationMapping <|-- ManyToManyOwningSideMapping
    AssociationMapping <|-- ToOneAssociationMapping
    ToOneAssociationMapping <|-- ManyToOneAssociationMapping
    ToManyAssociationMapping <|-- OneToManyAssociationMapping
    ToOneAssociationMapping <|-- OneToOneAssociationMapping
    OneToOneAssociationMapping <|-- OneToOneOwningSideMapping
```

## Interfaces

```mermaid
classDiagram
    AssociationOwningSideMapping <|.. OneToOneOwningSideMapping
    AssociationOwningSideMapping <|.. ManyToManyOwningSideMapping
    AssociationOwningSideMapping <|.. ManyToOneAssociationMapping
```

# After

## Classes

```mermaid
classDiagram
    AssociationMapping <|-- InverseSideMapping
    AssociationMapping <|-- OwningSideMapping
    InverseSideMapping <|-- ToManyInverseSideMapping
    OwningSideMapping <|-- ToManyOwningSideMapping
    ToManyInverseSideMapping <|-- ManyToManyInverseSideMapping
    ToManyOwningSideMapping <|-- ManyToManyOwningSideMapping
    OwningSideMapping <|-- ToOneOwningSideMapping
    ToOneOwningSideMapping <|-- ManyToOneOwningSideMapping
    ToManyInverseSideMapping <|-- OneToManyAssociationMapping
    InverseSideMapping <|-- ToOneInverseSideMapping
[...]

PR #10682, opened 2023-05-05
Make sure unneeded fields are not present

greg0ire:
Instead of ensuring every mapping array has a mappedBy and an inversedBy field, let us do the opposite, and remove them when they are null. Likewise if there is a joinColumns field, it is useless if null or empty.

Found while working on #10681 
These unneeded fields make the work unnecessarily hard.


SenseException (approved):
Why removing the array elements in `ClassMetadataBuilderTest` when they don't have an impact in `_validateAndCompleteAssociationMapping` anymore?

greg0ire:
~The array elements in that tests are compare used to build DTOs that we compare to other DTO built thanks to this class, which no longer includes them now.~ Wait no that's not it… I don't remember why I did this :thinking: 

Maybe that's something I did before extracting this to a separate commit :thinking: 

Ah yes that's it! When I use `git restore --source=origin/3.0.x tests/Doctrine/Tests/ORM/Mapping/ClassMetadataBuilderTest.php` on the branch from #10681 , then the tests fail. The new, reworked hierarchy understands this is not normal:

> OutOfRangeException: Unknown property mappedBy on class Doctrine\ORM\Mapping\ManyToOneAssociationMapping

That's what forced me to do this change, but this is the correct thing to do IMO, regardless of whether the test suite passes or not.

PR #10689, opened 2023-05-08
Compute entity-level commit order for entity insertions

mpdude:
Make the UoW find a commit order for entity insertions not on the class, but at the entity level. This is the third step to break https://github.com/doctrine/orm/pull/10547 into smaller PRs suitable for reviewing. It uses the new topological sort implementation from #10592 and the refactoring from #10651.

#### Current situation

`UnitOfWork::getCommitOrder()` would compute the entity sequence on the class level with the following code:

https://github.com/doctrine/orm/blob/70477d81e96c0044ad6fd8c13c37b2270d082792/lib/Doctrine/ORM/UnitOfWork.php#L1310-L1325

#### Suggested change

* Instead of considering the classes of all entities that need to be inserted, updated or deleted, consider the new (inserted) entities only. We only need to find a sequence in situations where there are foreign key relationships between two _new_ entities.
* In the dependency graph, add edges for all to-one association target entities.
* Make edges "optional" when the association is nullable.

#### Extra bonus

This is what the DALL·E AI thinks it looks like when the `UnitOfWork` is scheduling the sequence of entity insertions.

DALL·E 2023-05-08 18 32 54

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -57,7 +57,6 @@ use function array_filter;
 use function array_key_exists;
 use function array_map;
 use function array_merge;
-use function array_pop;
 use function array_sum;
 use function array_values;
 use function assert;
@@ -1365,81 +1364,6 @@ class UnitOfWork implements PropertyChangedListener
         }
 
         return $sort->sort();
-    }
-
-    /**
-     * Gets the commit order.
-     *
-     * @return list<ClassMetadata>
-     */
-    private function getCommitOrder(): array
-    {
-        $calc = $this->getCommitOrderCalculator();
-
-        // See if there are any new classes in the changeset, that are not in the
-        // commit order graph yet (don't have a node).
-        // We have to inspect changeSet to be able to correctly build dependencies.
-        // It is not possible to use IdentityMap here because post inserted ids
-        // are not yet available.
-        $newNodes = [];
-
-        foreach (array_merge($this->entityInsertions, $this->entityUpdates, $this->entityDeletions) as $entity) {
-            $class = $this->em->getClassMetadata(get_class($entity));
-
-            if ($calc->hasNode($class->name)) {
-                continue;
-            }
-
-            $calc->addNode($class->name, $class);
-
-            $newNodes[] = $class;
-        }
-
-        // Calculate dependencies for new nodes
-        while ($class = array_pop($newNodes)) {
-            foreach ($class->associationMappings as $assoc) {
-                if (! ($assoc['isOwningSide'] && $assoc['type'] & ClassMetadata::TO_ONE)) {
-                    continue;
-                }
-
-                $targetClass = $this->em->getClassMetadata($assoc['targetEntity']);
-
-                if (! $calc->hasNode($targetClass->name)) {
-                    $calc->addNode($targetClass->name, $targetClass);
-
-                    $newNodes[] = $targetClass;
-                }
-
-                $joinColumns = reset($assoc['joinColumns']);
-
-                $calc->addDependency($targetClass->name, $class->name, (int) empty($joinColumns['nullable']));
-
-                // If the target class has mapped subclasses, these share the same dependency.
-                if (! $targetClass->subClasses) {
-                    continue;
-                }
-
-                foreach ($targetClass->subClasses as $subClassName) {
-                    $targetSubClass = $this->em->getClassMetadata($subClassName);
-
-                    if (! $calc->hasNode($subClassName)) {
-                        $calc->addNode($targetSubClass->name, $targetSubClass);
-
-                        $newNodes[] = $targetSubClass;
-                    }
-
-                    $calc->addDependency($targetSubClass->name, $class->name, 1);
-                }
-            }
-        }
-
-        // Remove duplicate class entries
-        $result = [];
-        foreach ($calc->sort() as $classMetadata) {
-            $result[$classMetadata->name] = $classMetadata;
-        }
-
-        return array_values($result);
     }
 
     /**
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1266,20 +1266,51 @@ class UnitOfWork implements PropertyChangedListener
     /** @return list<object> */
     private function computeInsertExecutionOrder(): array
     {
-        $commitOrder = $this->getCommitOrder();
-        $result      = [];
-        foreach ($commitOrder as $class) {
-            $className = $class->name;
-            foreach ($this->entityInsertions as $entity) {
-                if ($this->em->getClassMetadata(get_class($entity))->name !== $className) {
+        $sort = new TopologicalSort();
+
+        // First make sure we have all the nodes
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1265,36 +1263,64 @@ private function executeDeletions(array $entities): void
...
+
+                // If there is no entity that we need to refer to, or it is already in the
+                // database (i. e. does not have to be inserted), no need to consider it.
+                if ($targetEntity === null || ! $sort->hasNode($targetEntity)) {
+                    continue;
+                }
+
+                // According to https://www.doctrine-project.org/projects/doctrine-orm/en/2.14/reference/annotations-reference.html#annref_joincolumn,
+                // the default for "nullable" is true. Unfortunately, it seems this default is not applied at the metadata driver, factory or other
+                // level, but in fact we may have an undefined 'nullable' key here, so we must assume that default here as well.
```
derrabus, line 1298: IIRC, join columns are nullable by default for one-to-many relations and not nullable for many-to-many relations. If at this point you need to know these defaults, maybe `ClassMetadata` should amend the `nullable` key if it's missing.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1265,36 +1263,64 @@ private function executeDeletions(array $entities): void
...
+                //
+                // From the dependency computation point of view, `false` would be a safer default, but that won't work with all examples and
+                // reproduce cases given.
```
derrabus: This comment merely sounds like a personal opinion than an explanation of the code ahead. I this we should remove it.

tests/Doctrine/Tests/Models/PersistentObject/PersistentEntity.php
```diff
@@ -29,6 +30,8 @@ class PersistentEntity extends PersistentObject
     protected $name;
 
     /**
+     * @ORM\JoinColumn(nullable=true)
```
derrabus: Why is this annotation necessary now?

lib/Doctrine/ORM/UnitOfWork.php
mpdude: changed

lib/Doctrine/ORM/UnitOfWork.php
mpdude: We have similar code here...

https://github.com/doctrine/orm/blob/70477d81e96c0044ad6fd8c13c37b2270d082792/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L2079-L2083

... or here ...

https://github.com/doctrine/orm/blob/70477d81e96c0044ad6fd8c13c37b2270d082792/lib/Doctrine/ORM/Tools/EntityGenerator.php#L1260-L1264

... and for fields in general (not only join columns) for example here:

https://github.com/doctrine/orm/blob/70477d81e96c0044ad6fd8c13c37b2270d082792/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L1394

I agree that it would be better if the `nullable` (and possibly other keys in those join column definition or field mapping configuration arrays) were initialized with defaults in the CMF; but that's out of scope here, so can we leave it for another PR?

tests/Doctrine/Tests/Models/PersistentObject/PersistentEntity.php
mpdude: Not sure, I copied that from the #10547 PR. Let's see if tests pass without it.

mpdude: In fact, it’s not necessary and I removed it.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1265,36 +1263,61 @@ private function executeDeletions(array $entities): void
...
+                // We only need to consider the owning sides of to-one associations,
+                // since many-to-many associations are persisted at a later step and
+                // have no insertion order problems (all entities already in the database
+                // at that time).
+                if (! ($assoc['isOwningSide'] && $assoc['type'] & ClassMetadata::TO_ONE)) {
                     continue;
                 }
 
-                $result[] = $entity;
+                $targetEntity = $class->getFieldValue($entity, $assoc['fieldName']);
```
mpdude, line 1289: We only need to schedule the order among new insertions. In case this value turns out to be a proxy, it means the other entity _must_ already be in the DB, so we can safely ignore it.

mpdude:
@greg0ire @SenseException and @derrabus  what do you think about this? 

We need to agree on this one to make progress for the #10547 epic.

tests/Doctrine/Tests/ORM/Functional/OneToOneSelfReferentialAssociationTest.php
```diff
@@ -74,9 +74,11 @@ public function testFind(): void
 
     public function testEagerLoadsAssociation(): void
     {
-        $this->createFixture();
+        $customerId = $this->createFixture();
+
+        $query = $this->_em->createQuery('select c, m from Doctrine\Tests\Models\ECommerce\ECommerceCustomer c left join c.mentor m where c.id = :id');
+        $query->setParameter('id', $customerId);
```
SenseException, line 80: What's the story about this changed test?

mpdude: The test assumed that the entities were inserted (and have their IDs assigned) in the order they were added to or discovered by the entity manager.

But, since Luke has Obi-Wan as his mentor, Obi-Wan will now be inserted first. This allows us to make the `mentor` connection without an extra `UPDATE` query.

I wrote the `TopologicalSort` class so that it keeps objects in order as much as possible (otherwise, it would have broken lots of tests), but associations and avoiding extra queries are the main reason for re-ordering inserts.

SenseException (approved):

mpdude:
Rebased and resolved conflicts after #10566 has been merged.

I wanted to spare clean-ups (remove the now unused `UoW::getCommitOrder()` method) for another PR, but static analysis checks forced me to remove it right away.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1336,81 +1366,6 @@ private function computeDeleteExecutionOrder(): array
         return $sort->sort();
     }
 
-    /**
-     * Gets the commit order.
-     *
-     * @return list<ClassMetadata>
-     */
-    private function getCommitOrder(): array
```
mpdude, line 1344: As of this PR, this method is no longer used.

mpdude:
@greg0ire and @SenseException 

I think this is the last PR regarding changes in implementation. Once we've got this merged, I can start making PRs to pick/collect/provide (regression) tests for all the fixed issues.

That means the hard work would be done, hopefully reviewing tests added is easy :-)

mpdude:
@greg0ire could you 👀 please? 🙏🏻

greg0ire (approved):

mpdude:
Please merge this, so I can continue with more tests on the `entity-level-commit-order` branch.

PR #10695, opened 2023-05-11
Changing return type of `getSingleScalarResult()` to `array|bool|float|int|string|null`

ThomasLandauer:
Reason: I'm using `intval()` to sanitize the result of `getSingleScalarResult()`. But now, PHPStan 1.10.15 complains that `intval()` only accepts `bool|float|int|resource|string|null` as input: https://github.com/phpstan/phpstan/issues/9295
So I threw all of these (except `resource`) in here :-)
I didn't doublecheck anything - waiting for some feedback first.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -1010,7 +1010,7 @@ abstract class AbstractQuery
      *
      * Alias for getSingleResult(HYDRATE_SINGLE_SCALAR).
      *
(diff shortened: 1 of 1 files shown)


phansys:
I didn't find this PR before, so I've created #10721. Sorry for that.

>I didn't doublecheck anything - waiting for some feedback first.

Based on n my tests, I can confirm this method can not return `array` or `null`.

ThomasLandauer:
Never mind - so I'm closing here :-)

PR #10701, opened 2023-05-13
Introduce convenience methods to narrow types

greg0ire:
These methods assert the type of the mapping provided by the collection according to the name of the class they are in: the one to many persister only ever deals with one to many associations, and the many to many persister only ever deals with many to many associations.

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
@@ -12,6 +12,7 @@ use Doctrine\DBAL\LockMode;
 use Doctrine\ORM\Mapping\AssociationMapping;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\InverseSideMapping;
+use Doctrine\ORM\Mapping\ManyToManyAssociationMapping;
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
```diff
@@ -757,4 +758,13 @@ private function getLimitSql(Criteria $criteria): string
 
         return $this->platform->modifyLimitQuery('', $limit, $offset ?? 0);
     }
+
+    private function getMapping(PersistentCollection $collection): AssociationMapping&ManyToManyAssociationMapping
+    {
+        $mapping = $collection->getMapping();
+
+        assert($mapping instanceof ManyToManyAssociationMapping);
```
SenseException, line 766: Is this assert even needed here when the type hint only allows `AssociationMapping` instances that implement `ManyToManyAssociationMapping`?

SenseException (approved):

lib/Doctrine/ORM/Persisters/Collection/ManyToManyPersister.php
greg0ire: Yes, otherwise SA will complain that nothing proves we return that type although we claim we do.

PR #10704, opened 2023-05-15
Remove unreachable piece of code

greg0ire:
`mappedBy` is never defined on an inverse relationship.
Bi-directional self-referencing should IMO still result in 2 separate
associations, on 2 separate fields of the same class, like mentor or
mentee.


mpdude (approved):
Plausible

SenseException (approved):
No test seems to fail.

PR #10707, opened 2023-05-16
Merge 2.16.x up into 3.0.x

greg0ire:


tests/Doctrine/Tests/Models/GH10334/GH10334ProductType.php
```diff
@@ -0,0 +1,44 @@
...
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\OneToMany;
+
+#[Entity]
+class GH10334ProductType
+{
+    #[Id]
+    #[Column(type: 'string', enumType: 'Doctrine\Tests\Models\GH10334\GH10334ProductTypeId', length: 255)]
```
greg0ire, line 18: I had to add a `length` here, otherwise there was an error message about MariaDB requiring a length… but I think that was MariaDB, the DBAL platform, not MariaDB the actual RDBMS, and that this would be due to using `doctrine/dbal` 4. It seems to do the trick.

greg0ire:
ping @derrabus @SenseException

tests/Doctrine/Tests/Models/GH10334/GH10334ProductType.php
SenseException: Should this be reversed after the MariaDB platform was fixed?

SenseException (approved):

tests/Doctrine/Tests/Models/GH10334/GH10334ProductType.php
greg0ire: Not sure what you mean :thinking: … is MariaDB broken in that it's not supposed to require a length? Also, why the past tense? Has it been fixed recently?

SenseException: I understand that the length was added because of a buggy-behaviour that the MariaDB DBAL platform introduced. Since this should actually work without the length too, I'd expect that the added length could be removed again once it was fixed in dbal.

PR #10708, opened 2023-05-16
Correct docs link

mbabker:
https://www.doctrine-project.org/projects/doctrine-orm/en/2.15/reference/inheritance-mapping.html has a broken link, this should fix it (it looks like the intent is to have `ResolveTargetEntityListener` shown as a code element but the parser might not be working?).


greg0ire (approved):
It does look like the correct syntax according to other occurrences in the project.

greg0ire:
Thanks @mbabker !

PR #10712, opened 2023-05-18
Fix break line in caching.rst

jaroslavtyc:
Remove a trailing space in the middle of a sentence causing new line by `<br>` in result.

```html
<p>In this section we'll describe how you can fully utilize the API of
the cache drivers to save data to a cache, check if some cached data<br>exists, fetch the cached data and delete the cached data. We'll use the
<code>ArrayCache</code> implementation as our example here.</p>
```

### DIFF
--- a/docs/en/reference/caching.rst
+++ b/docs/en/reference/caching.rst
@@ -142,7 +142,7 @@ Using Cache Drivers
 -------------------
 
 In this section we'll describe how you can fully utilize the API of
-the cache drivers to save data to a cache, check if some cached data 
+the cache drivers to save data to a cache, check if some cached data
 exists, fetch the cached data and delete the cached data. We'll use the
 ``ArrayCache`` implementation as our example here.
 


greg0ire:
Why are you targeting 2.8.x?

jaroslavtyc:
@greg0ire Got there by Google, did not notice.

Latest https://www.doctrine-project.org/projects/doctrine-orm/en/2.15/reference/caching.html#using-cache-drivers does not have that typo (nor the sentence itself) so we can close this as solved already.

jaroslavtyc:
Note: For example Symfony docs have warning "You are browsing old documentation. Consider switch to latest" https://symfony.com/doc/5.3/service_container/alias_private.html

Would be nice for Doctrine docs too.

greg0ire:
@jaroslavtyc well it's already the case, isn't it? There is a header with a warning on old Doctrine docs.

greg0ire:
The real question is: why does Google send you to an old version? Is there some search engine optimization we could do?

jaroslavtyc:
@greg0ire 

> here is a header with a warning on old Doctrine docs.

And I checked it before I wrote it. Dunno why I have missed it (twice), sorry. Maybe my brain just filtered it out as there is no "Call to action - click here to get to latest version you fool".

> The real question is: why does Google send you to an old version? Is there some search engine optimization we could do?

I have used "doctrine redis cache vs native sql" as a search term. Tried it now in private browser window and still the first result leads to old docs. There will be probably some trick how to tell Google to lower old docs rank in search result.

greg0ire:
I'm away from my computer now but can you please research "link rel canonical" and see if we implement it? IIRC it should fix that kind of issue.

jaroslavtyc:
Yeah, it is there (old docs) https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/caching.html#using-cache-drivers, canonical link to current version `<link rel="canonical" href="[https://www.doctrine-project.org/projects/doctrine-orm/en/2.15/reference/caching.html](view-source:https://www.doctrine-project.org/projects/doctrine-orm/en/2.15/reference/caching.html)" />` and seems valid.

Looks like you did maximum to avoid such situation.

greg0ire:
That href value looks weird, actually, doesn't it?

greg0ire:
Ah no it's just what happens when copy pasting from firefox's view-source I suppose. I double checked with wget and it is fine. What keywords did you use when searching on Google?

greg0ire:
Something must be off because both duckduckgo and google seem to show old versions of the docs when typing for instance: "doctrine using cache drivers"

greg0ire:
Ah that text does not appear on the page of version 2.15. I guess you must be facing a similar situation.

jaroslavtyc:
@greg0ire 

> Ah that text does not appear on the page of version 2.15. I guess you must be facing a similar situation.

Yup

PR #10716, opened 2023-05-19
Fixes #10715 - Incorrect subclasses dependencies weight while calculating commit orders

### LINKED ISSUE #10715: Incorrect subclasses dependencies weight while calculating commit orders
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.15.1

#### Summary

The weight of subclasses' dependencies is always 1 while the parent class' dependencies weight comes from the nullable attribute of the join columns.

#### Current behavior

Parent class and subclasses' dependencies don't have the same weight.

#### How to reproduce

Imagine this setup with 3 entities
- City
- Street (can be Street, Boulevard, or Avenue as inheritance mapping)
- House

- House has a Street (not null)
- Street has a City (not null)
- City has a main Street (nullable as we have to first insert the City before the Street due to foreign key constraints)
- Street has a main House (nullable as we have to first insert the Street before the House due to foreign key constraints)

Expected dependencies are:
- from House to Street, Boulevard, and Avenue with weight=0
- from Street, Boulevard, and Avenue to City with weight=0
- from City to Street, Boulevard, and Avenue with weight=1
- from Street, Boulevard, and Avenue to House with weight=1

Current dependencies are:
- from House to Street, Boulevard, and Avenue with weight=0
- from Street to City with weight=0
- from Boulevard and Avenue to City with weight=1 KO
- from City to Street, Boulevard, and Avenue with weight=1
- from Street, Boulevard, and Avenue to House with weight=1

This comes from `UnitOfWork::getCommitOrder()` which always uses 1 as weight when calling `$calc->addDependency()` for subclasses (see https://github.com/doctrine/orm/pull/10716 about a suggested fix)

#### Expected behavior

Parent class and subclasses' dependencies have the same weight.

sylfabre:
This PR fixes https://github.com/doctrine/orm/issues/10715 

It makes `UnitOfWork::getCommitOrder()` use consistent weight for both the parent class and subclasses' dependencies

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1322,7 +1322,8 @@ class UnitOfWork implements PropertyChangedListener
 
                 $joinColumns = reset($assoc['joinColumns']);
 
-                $calc->addDependency($targetClass->name, $class->name, (int) empty($joinColumns['nullable']));
+                $weight = (int) empty($joinColumns['nullable']);
+                $calc->addDependency($targetClass->name, $class->name, $weight);
 
                 // If the target class has mapped subclasses, these share the same dependency.
                 if (! $targetClass->subClasses) {
@@ -1338,7 +1339,7 @@ class UnitOfWork implements PropertyChangedListener
                         $newNodes[] = $targetSubClass;
                     }
 
-                    $calc->addDependency($targetSubClass->name, $class->name, 1);
+                    $calc->addDependency($targetSubClass->name, $class->name, $weight);
                 }
             }
         }


sylfabre:
I would like to add a functional test to prove the bug and that my fix is working but I don't know how to do it:
- May someone point me in the right direction with an existing functional test I can use as a base
- Where should I put such a functional test in the repository? Is `tests/Doctrine/Tests/ORM/Functional` the right path? What should be the filename?

Thanks 🙏

mpdude:
Hey there 👋🏼,

I am actively working on major changes to the commit order computation. Could you please have a look at #10547 and/or try if the branch `entity-level-commit-order` (you can Composer-install it as `dev-entity-level-commit-order`!) solves the issues for you? If so, please leave a note in #10547.

sylfabre:
This PR is replaced by https://github.com/doctrine/orm/pull/10547/files

Thanks @mpdude

PR #10719, opened 2023-05-20
Deleted entity should stay deleted "forever"

### LINKED ISSUE #10718: Deleted entity is causing "A new entity was found through the relationship" exception in the case of multiple flush() calls
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.15.1

#### Summary

Deleted entity is perceived as new after first `EM::flush()` call.
Therefore potentially causing this exception in the future EM::flush() calls:
```
Doctrine\ORM\ORMInvalidArgumentException
A new entity was found through the relationship
```

Example use case could be deleting many interconnected entities in batches.
This is currently no possible to do without  throwing `ORMInvalidArgumentException`. No other EM::flush() call will succeed.

#### Expected behavior

Deleted should stay deleted "forever".

forgie1:
Closes https://github.com/doctrine/orm/issues/10718

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1260,13 +1260,10 @@ class UnitOfWork implements PropertyChangedListener
             unset(
                 $this->entityDeletions[$oid],
                 $this->entityIdentifiers[$oid],
-                $this->originalEntityData[$oid],
-                $this->entityStates[$oid]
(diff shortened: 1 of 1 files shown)


derrabus:
Your change is not covered by a new test, and it breaks existing tests.

forgie1:
Our `onFlush` event was the problem, I'm closing the issue and MR.

PR #10724, opened 2023-05-24
Temporarily allow `symfony/console` 7

derrabus:
This change helps Symfony to test Symfony 7 against the ORM. We will likely release ORM 2.16 before Symfony 7, which is why we will need to revert this change before releasing 2.16.0. I'll prepare a PR to do that once this one is merged.

cc @nicolas-grekas

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "doctrine/lexer": "^2",
         "doctrine/persistence": "^2.4 || ^3",
         "psr/cache": "^1 || ^2 || ^3",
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Maybe you could open branch `2.17.x` right now if that makes things a bit easier? Not sure, but sharing in case :)

derrabus:
We don't have a fixed schedule for feature releases. It might very well be that we release 2.42.0 before Symfony 7. 😅

greg0ire (approved):

PR #10727, opened 2023-05-25
Revert "Allow symfony/console 7"

derrabus:
Follows #10724.

I'm adding this PR to the 2.16.0 milestone so we won't forget to revert Symfony 7 support before the final release.


SenseException (approved):
Insta-Approve to be ready to merge once the time for this is right.

PR #10732, opened 2023-05-31
Add a test case to show #10348 has been fixed

mpdude:
This is part of the series of issues fixed by #10547. In particular, the changes from #10566 were relevant.

See #10348 for the bug description.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
@@ -1,0 +1,108 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+final class GH10348Test extends OrmFunctionalTestCase
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH10348Person::class,
+            GH10348Company::class,
+        ]);
+    }
+
+    public function testTheORMRemovesReferencedEmployeeBeforeReferencingEmployee(): void
+    {
+        $person1         = new GH10348Person();
+        $person2         = new GH10348Person();
+        $person2->mentor = $person1;
+
+        $company = new GH10348Company();
+        $company->addEmployee($person1)->addEmployee($person2);
+
+        $this->_em->persist($company);
+        $this->_em->flush();
+
+        $company = $this->_em->find(GH10348Company::class, $company->id);
+
+        $this->_em->remove($company);
+        $this->_em->flush();
+
+        self::assertEmpty($this->_em->createQuery('SELECT c FROM ' . GH10348Company::class . ' c')->getResult());
+        self::assertEmpty($this->_em->createQuery('SELECT p FROM ' . GH10348Person::class . ' p')->getResult());
+    }
+}
+
+/**
+ * @ORM\Entity
+ */
+class GH10348Person
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue
+     *
+     * @var ?int
+     */
+    public $id = null;
+
+    /**
+     * @ORM\ManyToOne(targetEntity="GH10348Company", inversedBy="employees")
+     *
+     * @var ?GH10348Company
+     */
+    public $employer = null;
+
+    /**
+     * @ORM\ManyToOne(targetEntity="GH10348Person", cascade={"remove"})
+     *
+     * @var ?GH10348Person
+     */
+    public $mentor = null;
+}
+
+/**
+ * @ORM\Entity
+ */
+class GH10348Company
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue
+     *
+     * @var ?int
+     */
+    public $id = null;
+
+    /**
+     * @ORM\OneToMany(targetEntity="GH10348Person", mappedBy="emplo", cascade={"persist", "remove"})
+     *
+     * @var Collection
+     */
+    private $employees;
+
+    public function __construct()
+    {
+        $this->employees = new ArrayCollection();
+    }
+
+    public function addEmployee(GH10348Person $person): self
+    {
+        $person->employer = $this;
+        $this->employees->add($person);
+
+        return $this;
+    }
+}


tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
```diff
@@ -0,0 +1,109 @@
...
+        $this->expectNotToPerformAssertions();
+        $this->_em->flush();
+    }
+}
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="gh10348_child_entities")
+ */
+class GH10348ChildEntity
```
greg0ire: I think the vocabulary parent-child is confusing as there is no PHP inheritance here.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH10348ChildEntity::class,
+            GH10348ParentEntity::class,
+        ]);
+    }
+
+    public function testCanRemoveParentWithChildRelatesToOwnEntity(): void
```
greg0ire: I think it would be best to use the same vocabulary as the ORM. This is a self-referencing entity.
```suggestion
    public function testCanRemoveParentWithSelfReferencingChild(): void
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
mpdude: I've been thinking about this but was unable to come up with names that would better reveal the intentions. Do you have any suggestions?

https://github.com/doctrine/orm/issues/10348#issuecomment-1455678557 shows the object graph and the problem. Yes, we have a self-referencing association between the "child" classes, but both of them also depend on the "parent" class.

greg0ire: Sure!

```suggestion
class GH10348Person
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
```diff
@@ -0,0 +1,109 @@
...
+     * @var ?int
+     */
+    public $id = null;
+
+    /**
+     * @ORM\ManyToOne(targetEntity="GH10348ParentEntity", inversedBy="children")
+     *
+     * @var ?GH10348ParentEntity
+     */
+    public $parent = null;
```
greg0ire:
```suggestion
    public $employer = null;
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * @ORM\ManyToOne(targetEntity="GH10348ChildEntity", cascade={"remove"})
+     *
+     * @var ?GH10348ChildEntity
+     */
+    public $origin = null;
```
greg0ire:
```suggestion
    public $mentor = null;
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="gh10348_parent_entities")
+ */
+class GH10348ParentEntity
```
greg0ire:
```suggestion
class GH10348Company
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
```diff
@@ -0,0 +1,109 @@
...
+     * @var ?int
+     */
+    public $id = null;
+
+    /**
+     * @ORM\OneToMany(targetEntity="GH10348ChildEntity", mappedBy="parent", cascade={"persist", "remove"})
+     *
+     * @var Collection
+     */
+    private $children;
```
greg0ire:
```suggestion
    private $employees;
```

mpdude:
@greg0ire Thank you for the suggestion! That makes it a lot better.

I've re-checked that this test indeed fails on 2.16, so it shows the issue has been fixed.

However, I wonder how we could make the test a bit more "forcing" – that is, make it obvious that naturally the ORM would delete `$person1` before `$person2`, and that it is a merit that it recognized the situation and turns things around.

Any ideas?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH10348Person::class,
+            GH10348Company::class,
+        ]);
+    }
+
+    public function testCanRemoveParentWithSelfReferencingChild(): void
```
greg0ire: Maybe by changing the title of the test again?
```suggestion
    public function testTheORMRemovesReferencedEmployeeBeforeReferencingEmployee(): void
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10348Test.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->_em->persist($company);
+        $this->_em->flush();
+
+        $company = $this->_em->find(GH10348Company::class, $company->id);
+
+        $this->_em->remove($company);
+        $this->_em->flush();
+
+        self::assertEmpty($this->_em->createQuery('SELECT c FROM ' . GH10348Company::class . ' c')->getResult());
+        self::assertEmpty($this->_em->createQuery('SELECT p FROM ' . GH10348Person::class . ' p')->getResult());
```
greg0ire, line 42: Are these assertions maybe superfluous and distracting? We're not testing the cascade here, are we? I think it would be best to remove them and use `@doesNotPerformAssertions`

mpdude: In another test that we merged recently Claudio suggested to have a check like this instead of `@doesNotPerformAssertions`. After all, the query returning an emtpy result is the best indication of things actually happening (entities being deleted).

I'm indifferent – you decide.

greg0ire: I see, we're not only checking there is no crash, we are checking that the reason there is no crash isn't nothing actually happening. Kind of far-fetched because I would expect that to be fairly covered already, but who knows. Works for me I suppose.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @mpdude !

PR #10734, opened 2023-05-31
Document pdo_sqlite requirement for tests

greg0ire:
The error message you get if you don't enable pdo_sqlite mentions a driver, but does not tell the users that it's provided by this PHP extension.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,11 @@ cd orm
 composer install
 ```
 
+You will also need to enable the PHP extension that provides the SQLite driver
+for PDO: `pdo_sqlite`. How to do so depends on your system, but checking that it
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -40,6 +40,11 @@ cd orm
 composer install
 ```
 
+You will also need to enable the PHP extension that provides the SQLite driver
+for PDO: `pdo_sqlite`. How to do so depends on your system, but checking can it
+is enabled can universally be done with `php -m`: that command should list the
```
SenseException: "checking can it is enabled can"?

greg0ire: Woops! Fixed!

derrabus (approved):

PR #10735, opened 2023-06-01
Make it possible to have non-NULLable self-referencing associations when using application-provided IDs

### LINKED ISSUE #7877: Extra update for self-referencing Many-To-One with NONE generator strategy during persist

mpdude:
This change improves scheduling of extra updates in the `BasicEntityPersister`.

Extra updates can be avoided when
* the referred-to entity has already been inserted during the current insert batch/transaction and the post-insert (database) ID is already available in the entity
* we have a self-referencing entity with application-provided ID values (the `NONE` generator strategy).

As a corollary, with this change applications that provide their own IDs can define self-referencing associations as not NULLable.

One caveat, though: In the absence of entity-level commit ordering (#10547), it is not guaranteed that entities with self-references (at the class level) will be inserted in a suitable order. The order depends on the sequence in which the entities were added with `persist()`. 

Fixes #7877, closes #7882.

Co-authored-by: Sylvain Fabre <sylvain.fabre@assoconnect.com>

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -266,7 +266,7 @@ class BasicEntityPersister implements EntityPersister
         $stmt      = $this->conn->prepare($this->getInsertSQL());
         $tableName = $this->class->getTableName();
 
-        foreach ($this->queuedInserts as $entity) {
+        foreach ($this->queuedInserts as $key => $entity) {
             $insertData = $this->prepareInsertData($entity);
 
             if (isset($insertData[$tableName])) {
@@ -291,9 +291,16 @@ class BasicEntityPersister implements EntityPersister
             if ($this->class->requiresFetchAfterChange) {
                 $this->assignDefaultVersionAndUpsertableValues($entity, $id);
             }
-        }
 
-        $this->queuedInserts = [];
+            // Unset this queued insert, so that the prepareUpdateData() method knows right away
+            // (for the next entity already) that the current entity has been written to the database
+            // and no extra updates need to be scheduled to refer to it.
+            //
+            // In \Doctrine\ORM\UnitOfWork::executeInserts(), the UoW already removed entities
+            // from its own list (\Doctrine\ORM\UnitOfWork::$entityInsertions) right after they
+            // were given to our addInsert() method.
+            unset($this->queuedInserts[$key]);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mpdude:
@sylfabre it ain't that simple – references between different entities also schedule extra updates, so we're not done yet.

mpdude:
We need to process the post-insert IDs earlier: Not after executeInserts() has been finished, but rather during the execution. → #10743

lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
```diff
@@ -683,10 +690,37 @@ protected function prepareUpdateData($entity, bool $isInsert = false)
...
+                // that the UnitOfWork has already removed entities from its own list at the time
+                // they were passed to our addInsert() method.
+                //
+                // Then, there is one extra exception we can make: An entity that references back to itself
+                // _and_ uses an application-provided ID (the "NONE" generator strategy) also does not
+                // need the extra update, although it is still in the list of insertions itself.
+                // This looks like a minor optimization at first, but is the capstone for being able to
+                // use non-NULLable, self-referencing associations in applications that provide IDs (like UUIDs).
+                if (
+                    (isset($this->queuedInserts[$oid]) || $uow->isScheduledForInsert($newVal) || !$uow->isInIdentityMap($newVal))
```
mpdude, line 721: This has to check the identity map once for every association value during inserts. 

Performance issue?

mpdude:
😭 Converting to draft – using `isInIdentityMap()` is not safe. This does not guarantee that the entity has been written to the database. There are other code paths possible that make an entity end up in the map before being persisted.

derrabus (approved):

sylfabre:
Thanks a lot @mpdude 💪

mpdude:
@sylfabre could you please close #7882 and #7877 when this change here solved the issue?

Thanks!

PR #10736, opened 2023-06-01
Fix Lexer deprecation

jmsche:
Hi,

This (very) small PR to fix a lexer deprecation:

> Accessing Doctrine\Common\Lexer\Token properties via ArrayAccess is deprecated, use the value, type or position property instead (Token.php:104 called by Parser.php:585, https://github.com/doctrine/lexer/pull/79, package doctrine/lexer)

I think the fix is okay for 2.15.x branch as it requires doctrine/lexer:^2.


greg0ire:
Thanks for this, but I'll close in favor of #10737 , which covers more cases.

PR #10737, opened 2023-06-01
Fix deprecations from doctrine/lexer

nicolas-grekas:
Not triggering self-deprecations is easy when you have the right process.

Since we enabled reporting of Doctrine deprecations on the CI of Symfony, we can tell that the processes here are severely broken, as in: E_TOO_MANY_SELF_DEPRECATIONS.

Here is one of them, the community will send more fixes as we make progress on the topic.

Doctrine v3 is going to be a risky release if the processes are not fixed before.

This feels so Symfony-2.7 :(

Sorry for the bad-mood message but it's not like I didn't extensively explain that this was about to happen when doctrine/deprecations was introduced...

With the right processes, deprecations would have been MUCH cheaper to deal with for everybody: maintainers and users.

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -582,7 +582,7 @@ class Parser
         assert($this->lexer->lookahead !== null);
 
         return in_array(
-            $this->lexer->lookahead['type'],
+            $this->lexer->lookahead->type,
             [Lexer::T_ALL, Lexer::T_ANY, Lexer::T_SOME],
             true
         );
@@ -978,7 +978,7 @@ class Parser
             $this->match(Lexer::T_IDENTIFIER);
             assert($this->lexer->token !== null);
 
-            return $this->lexer->token['value'];
+            return $this->lexer->token->value;
         }
 
         $this->match(Lexer::T_ALIASED_NAME);
@@ -989,10 +989,10 @@ class Parser
             'doctrine/orm',
             'https://github.com/doctrine/orm/issues/8818',
             'Short namespace aliases such as "%s" are deprecated and will be removed in Doctrine ORM 3.0.',
-            $this->lexer->token['value']
+            $this->lexer->token->value
         );
 
-        [$namespaceAlias, $simpleClassName] = explode(':', $this->lexer->token['value']);
+        [$namespaceAlias, $simpleClassName] = explode(':', $this->lexer->token->value);
 
         return $this->em->getConfiguration()->getEntityNamespace($namespaceAlias) . '\\' . $simpleClassName;
     }


greg0ire (approved):

greg0ire:
It's not the same package, so technically, those are not self deprecations :P 
Thanks for fixing this though, I'll try to think about improving the processes, there is probably room for improvement indeed.

greg0ire:
Thanks @nicolas-grekas ! I'll do a release right now since we got 2 PRs for this already.

nicolas-grekas:
Thanks for merging.
You can wait a bit if you want, at least until we fixed other deprecations in Symfony.
(we don't need a release for the CI of Symfony)

greg0ire:
Well… that ship has sailed already. No harm done IMO, but I will wait for our signal next time.

MalikOvaiz:
Hi,

I am using version **2.16** of **doctrine/orm** and I still get indirect deprecation notices.

> Accessing Doctrine\Common\Lexer\Token properties via ArrayAccess is deprecated, use the value, type or position property instead (Token.php:104 called by DocParser.php:691, https://github.com/doctrine/lexer/pull/79, package doctrine/lexer)

Is this not supposed to be fixed after this merge request?

greg0ire:
@MalikOvaiz you're reporting this to the wrong package: `DocParser.php` is part of `doctrine/annotations`.

pfpro:
@greg0ire 
and what is your opinion about this deprecation: ?

User Deprecated: Accessing Doctrine\Common\Lexer\Token properties via ArrayAccess is deprecated, use the value, type or position property instead (Token.php:104 called by TimestampDiff.php:25, https://github.com/doctrine/lexer/pull/79, package doctrine/lexer)

(doctrine 2.17.1)

greg0ire:
@pfpro no opinion, I do not know that `TimestampDiff.php` file you are mentioning.

pfpro:
Unfortunately, neither do I :)
Thanks for your opinion anyway

greg0ire:
Yeah but you, on the other hand, have that file on your disk, presumably so maybe you could search for it, you know?

pfpro:
good point, i'll keep on researching

its the `TimestampDiff.php` from beberlei/doctrineextensions

greg0ire:
Then you know where to report this now.

pfpro:
Not really ...

> Issues are disabled on this repository, if a custom DQL function that you want isn't provided, or does not support the arguments you want to pass, pull requests are open and we would love to have your contributions.

This is the full log:

`{"exception":"[object] (ErrorException(code: 0): User Deprecated: Accessing Doctrine\\Common\\Lexer\\Token properties via ArrayAccess is deprecated, use the value, type or position property instead (Token.php:104 called by TimestampDiff.php:25, https://github.com/doctrine/lexer/pull/79, package doctrine/lexer) at /usr/www/.../vendor/doctrine/deprecations/lib/Doctrine/Deprecations/Deprecation.php:210)"} []`

greg0ire:
Okay, so send a pull request to apply the advice in the error message to `TimestampDiff` line 25… it must be using array access instead of object access.

pfpro:
How about you take care of it in this case? You seem to know exactly how to solve this mission. Shout outs to Paris!

greg0ire:
I'm away from keyboard, can't do it, sorry!

PR #10741, opened 2023-06-01
Cleanup bc layers

greg0ire:
Note that we also remove `UpdateCommandTest::testItPrintsTheSql()` because
it relied on not passing --complete to work.


lib/Doctrine/ORM/Tools/Console/Command/SchemaTool/UpdateCommand.php
```diff
@@ -29,7 +29,7 @@ protected function configure(): void
         $this->setName($this->name)
              ->setDescription('Executes (or dumps) the SQL needed to update the database schema to match the current mapping metadata')
              ->addOption('em', null, InputOption::VALUE_REQUIRED, 'Name of the entity manager to operate on')
-             ->addOption('complete', null, InputOption::VALUE_NONE, 'If defined, all assets of the database which are not relevant to the current metadata will be dropped.')
+             ->addOption('complete', null, InputOption::VALUE_NONE, 'This option is a no-op and should be deprecated in 3.1.0')
```
SenseException: This is a weird description for an option, because the option description. It's like being already deprecated, but without the logging.

If this does nothing different anymore, then why do we have it in an upcoming major release?

greg0ire: I'm hesitant about this, but right now we are saying that specifying the option is mandatory: people should prepare for the command to behave as if it was always provided. If we remove the command in the next major then they will get an error after upgrading, without the option being deprecated beforehand.

SenseException: I wonder if this could be handled as you suggested in https://github.com/doctrine/doctrine-website/pull/485#issuecomment-1572533717 and another minor release, but I don't know if this will be harder to maintain for 3.0.

This is no real blocker from my point of view if you have no other way to introduce this deprecation before 3.0. It just looks like unfinished business in a major release.

greg0ire: Yeah, I know it's unsatisfying, but I don't see a way around it.

derrabus: This is fine. We need the option in 2.x to opt into the new behavior and in order to keep the upgrade path smooth, we need to make sure that the opt-in does not error in 3.x.

derrabus:
```suggestion
             ->addOption('complete', null, InputOption::VALUE_NONE, 'This option is a no-op and will be removed in 4.0')
```

derrabus (approved):

PR #10743, opened 2023-06-02
Make EntityPersisters tell the UoW about post insert IDs early

mpdude:
This refactoring does two things:

* We can avoid collecting the post insert IDs in a cumbersome array structure that will be returned by the EntityPersisters and processed by the UoW right after. Instead, use a more expressive API: Make the EntityPersisters tell the UoW about the IDs immediately.
* IDs will be available in inserted entities a tad sooner. That may help to resolve #10735, where we can use the IDs to skip extra updates.

Reviewers: Is the way I implemented the deprecation that EntityPersisters shall no longer return these arrays and change their return type to `void` ok?


tests/Doctrine/Tests/ORM/Persisters/JoinedSubclassPersisterTest.php
```diff
@@ -1,38 +0,0 @@
...
-    protected function setUp(): void
-    {
-        parent::setUp();
-
-        $this->em        = $this->getTestEntityManager();
-        $this->persister = new JoinedSubclassPersister($this->em, $this->em->getClassMetadata(RootClass::class));
-    }
-
-    /** @group DDC-3470 */
-    public function testExecuteInsertsWillReturnEmptySetWithNoQueuedInserts(): void
```
mpdude, line 34: This was the only test in here, and it is now meaningless.

greg0ire:
> Reviewers: Is the way I implemented the deprecation that EntityPersisters shall no longer return these arrays and change their return type to void ok?

I don't see anything wrong with that. Please document the deprecation though.

mpdude:
Added the deprecation in the UPGRADE file

lib/Doctrine/ORM/Persisters/Entity/EntityPersister.php
```diff
@@ -114,12 +114,12 @@ public function addInsert($entity);
      *
      * If no inserts are queued, invoking this method is a NOOP.
      *
-     * @psalm-return list<array{
+     * @psalm-return void|list<array{
      *                   generatedId: int,
      *                   entity: object
-     *               }> An array of any generated post-insert IDs. This will be
+     *               }> Returning an array of any generated post-insert IDs. This will be
```
greg0ire: Why did you feel the need to add "Returning an" here? It feels like you wanted to write "Returning an array of… is deprecated", but stopped halfway.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3790,4 +3784,30 @@ private function normalizeIdentifier(ClassMetadata $targetClass, array $flatIden
...
+     * Assign a post-insert generated ID to an entity
+     *
+     * This is used by EntityPersisters after they inserted entities into the database.
+     * It will place the assigned ID values in the entity's fields and start tracking
+     * the entity in the identity map.
+     *
+     * @param object $entity
+     * @param mixed  $generatedId
+     */
+    public function assignPostInsertId($entity, $generatedId): void
```
greg0ire: Should this method be `final`?

mpdude: IMHO the entire UoW should be final :-)

mpdude: Even if technically it is not, we should treat it as if it were 😈

lib/Doctrine/ORM/Persisters/Entity/EntityPersister.php
mpdude: fixed

lib/Doctrine/ORM/UnitOfWork.php
mpdude: fixed

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @mpdude !

Issue #9745, opened 2022-05-09
`orphanRemoval` not working for `OneToMany` collections when using a custom type for the identifier and oracle database


lordrhodos:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.3

#### Summary

When using a custom type for the identifier of a `OneToMany` relation the `orphanRemoval` setting will trigger an error. 

#### Current behavior

In my case I am using a `OneToMany` relation to `SessionTag`:

**Session**

```php
    /**
     * @ORM\OneToMany(
     *     targetEntity="SessionTag",
     *     mappedBy="session",
     *     fetch="EXTRA_LAZY",
     *     cascade={"remove"},
     * )
     */
    private Collection $sessionTags;
```

`SessionTag` uses a custom uuid type as id:

**SessionTag**
```php
    /**
     * @ORM\Id
     * @ORM\Column(name="id", type="uuid")
     */
    private UuidInterface $id;
```

When I add the `orphanRemoval=true` attribute to the annotation and remove entities from the collection, a `flush` will trigger the following error:

    [2022-05-09T10:35:18.563416+02:00] request.CRITICAL: Uncaught PHP Exception PHPUnit\Framework\Error\Warning: "oci_bind_by_name(): Invalid variable used for bind" at /var/www/app/vendor/doctrine/dbal/src/Driver/OCI8/Statement.php line 86 {"exception":"[object] (PHPUnit\\Framework\\Error\\Warning(code: 2): oci_bind_by_name(): Invalid variable used for bind at /var/www/app/vendor/doctrine/dbal/src/Driver/OCI8/Statement.php:86)"} []

#### Root cause
The cause of the issue is the missing support for custom types in the method `\Doctrine\ORM\Persisters\Collection\OneToManyPersister::deleteEntityCollection`. The `executeStatement` is called without passing the types, triggering the error:

https://github.com/doctrine/orm/blob/4af1aa317713d5f3dbc57b22237d1acb8477079c/lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php#L169-L187

#### Expected behavior
The `orphanRemoval` attribute should be supported for custom types.

derrabus:
Thank you for your report. Are you able to work on a fix?

lordrhodos:
> Thank you for your report. Are you able to work on a fix?

@derrabus the fix is ready, but I had no luck providing a failing test with sqlite so far. Will work on that 😉

PR #10747, opened 2023-06-02
fix: OneToManyPersister does not take custom identifier types into account for orphan removal

wtfzdotnet:
I've taken a look at the test suite, but I am kind of mind-boggled as of where to begin, could you please point me in the right direction to make sure this change gets merged? (I'll tidy up this PR later when it's done)

Thanks! :-)

### DIFF
--- a/lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
@@ -16,6 +16,7 @@ use function array_reverse;
 use function array_values;
 use function assert;
 use function implode;
+use function is_int;
 use function is_string;
 
 /**
@@ -183,7 +184,11 @@ class OneToManyPersister extends AbstractCollectionPersister
         $statement = 'DELETE FROM ' . $this->quoteStrategy->getTableName($targetClass, $this->platform)
             . ' WHERE ' . implode(' = ? AND ', $columns) . ' = ?';
 
-        return $this->conn->executeStatement($statement, $parameters);
+        $numAffected = $this->conn->executeStatement($statement, $parameters);
+
+        assert(is_int($numAffected));
+
+        return $numAffected;
     }
 
     /**
@@ -246,6 +251,8 @@ class OneToManyPersister extends AbstractCollectionPersister
         $statement = $this->platform->getDropTemporaryTableSQL($tempTable);
 
         $this->conn->executeStatement($statement);
+
+        assert(is_int($numDeleted));
 
         return $numDeleted;
     }
--- a/lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
@@ -175,16 +175,18 @@ class OneToManyPersister extends AbstractCollectionPersister
         $targetClass = $this->em->getClassMetadata($mapping['targetEntity']);
         $columns     = [];
         $parameters  = [];
+        $types       = [];
 
         foreach ($targetClass->associationMappings[$mapping['mappedBy']]['joinColumns'] as $joinColumn) {
             $columns[]    = $this->quoteStrategy->getJoinColumnName($joinColumn, $targetClass, $this->platform);
             $parameters[] = $identifier[$sourceClass->getFieldForColumn($joinColumn['referencedColumnName'])];
+            $types[]      = PersisterHelper::getTypeOfColumn($joinColumn['referencedColumnName'], $sourceClass, $this->em);
         }
 
         $statement = 'DELETE FROM ' . $this->quoteStrategy->getTableName($targetClass, $this->platform)
             . ' WHERE ' . implode(' = ? AND ', $columns) . ' = ?';
 
-        $numAffected = $this->conn->executeStatement($statement, $parameters);
+        $numAffected = $this->conn->executeStatement($statement, $parameters, $types);
 
         assert(is_int($numAffected));
 
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1236,14 +1236,6 @@
                 $mapping['indexBy']  => $index,
             ]]]></code>
     </InvalidArrayOffset>
-    <InvalidReturnStatement>
-      <code>$numDeleted</code>
-      <code><![CDATA[$this->conn->executeStatement($statement, $parameters)]]></code>
-    </InvalidReturnStatement>
-    <InvalidReturnType>
-      <code>int</code>
-      <code>int</code>
-    </InvalidReturnType>
     <PossiblyNullArgument>
       <code><![CDATA[$collection->getOwner()]]></code>
       <code><![CDATA[$collection->getOwner()]]></code>
--- a/tests/Doctrine/Tests/ORM/Functional/GH10747Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/GH10747Test.php
@@ -1,0 +1,195 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\Type as DBALType;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\Table;
+use Doctrine\Tests\DbalTypes\CustomIdObject;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+use function method_exists;
+use function str_replace;
+
+/**
+ * Functional tests for asserting that orphaned children in a OneToMany relationship get removed with a custom identifier
+ *
+ * @group GH10747
+ */
+final class GH10747Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        if (! DBALType::hasType(GH10747CustomIdObjectHashType::class)) {
+            DBALType::addType(GH10747CustomIdObjectHashType::class, GH10747CustomIdObjectHashType::class);
+        }
+
+        $this->setUpEntitySchema([GH10747Article::class, GH10747Credit::class]);
+    }
+
+    public function testOrphanedOneToManyDeletesCollection(): void
+    {
+        $object = new GH10747Article(
+            new CustomIdObject('article')
+        );
+
+        $creditOne = new GH10747Credit(
+            $object,
+            'credit1'
+        );
+
+        $creditTwo = new GH10747Credit(
+            $object,
+            'credit2'
+        );
+
+        $object->setCredits(new ArrayCollection([$creditOne, $creditTwo]));
+
+        $this->_em->persist($object);
+        $this->_em->persist($creditOne);
+        $this->_em->persist($creditTwo);
+        $this->_em->flush();
+
+        $id = $object->id;
+
+        $object2 = $this->_em->find(GH10747Article::class, $id);
+
+        $creditThree = new GH10747Credit(
+            $object2,
+            'credit3'
+        );
+
+        $object2->setCredits(new ArrayCollection([$creditThree]));
+
+        $this->_em->persist($object2);
+        $this->_em->persist($creditThree);
+        $this->_em->flush();
+
+        $currentDatabaseCredits = $this->_em->createQueryBuilder()
+            ->select('c.id')
+            ->from(GH10747Credit::class, 'c')
+            ->getQuery()
+            ->execute();
+
+        self::assertCount(1, $currentDatabaseCredits);
+    }
+}
+
+/**
+ * @Entity
+ * @Table
+ */
+class GH10747Article
+{
+    /**
+     * @Id
+     * @Column(type="Doctrine\Tests\ORM\Functional\GH10747CustomIdObjectHashType")
+     * @var CustomIdObject
+     */
+    public $id;
+
+    /**
+     * @ORM\OneToMany(targetEntity="GH10747Credit", mappedBy="article", orphanRemoval=true)
+     *
+     * @var Collection<int, GH10747Credit>
+     */
+    public $credits;
+
+    public function __construct(CustomIdObject $id)
+    {
+        $this->id      = $id;
+        $this->credits = new ArrayCollection();
+    }
+
+    public function setCredits(Collection $credits): void
+    {
+        $this->credits = $credits;
+    }
+
+    /** @return Collection<int, GH10747Credit> */
+    public function getCredits(): Collection
+    {
+        return $this->credits;
+    }
+}
+
+/**
+ * @Entity
+ * @Table
+ */
+class GH10747Credit
+{
+    /**
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue()
+     *
+     * @Id()
+     * @var int|null
+     */
+    public $id = null;
+
+    /** @var string */
+    public $name;
+
+    /**
+     * @ORM\ManyToOne(targetEntity="GH10747Article", inversedBy="credits")
+     *
+     * @var GH10747Article
+     */
+    public $article;
+
+    public function __construct(GH10747Article $article, string $name)
+    {
+        $this->article = $article;
+        $this->name    = $name;
+    }
+}
+
+class GH10747CustomIdObjectHashType extends DBALType
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        return $value->id . '_test';
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        return new CustomIdObject(str_replace('_test', '', $value));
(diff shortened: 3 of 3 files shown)


mpdude:
Maybe the functional tests named `GHxxxTest` according to issue numbers give you enough inspiration to get started?

wtfzdotnet:
> Maybe the functional tests named `GHxxxTest` according to issue numbers give you enough inspiration to get started?

Yeah I would've just placed this next to the existing ManyToManyPersister test, so good I asked, thanks :-)

wtfzdotnet:
Okay I've made a unit test that validates this behavior ( tested on mysql, and it fails as expected ), adding Ramsey as a vendor obviously doesn't sense, but I could use a little help with ideas as of how to go about testing this via another route :-).

composer.json
```diff
@@ -45,6 +45,8 @@
         "phpstan/phpstan": "~1.4.10 || 1.10.14",
         "phpunit/phpunit": "^7.5 || ^8.5 || ^9.6",
         "psr/log": "^1 || ^2 || ^3",
+        "ramsey/uuid": "^4.7",
+        "ramsey/uuid-doctrine": "^2.0",
```
greg0ire: Other route: create a custom type in the test: https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/types.html#custom-mapping-types

wtfzdotnet: I did that in the first place, but SQLite couldn't be bothered in memory with my assertions for some reason. After I've set up a docker MySQL instance I got further, I'll try this again later with a simplistic example. Heading out to enjoy the sun for a bit now :-)

greg0ire: > SQLite couldn't be bothered in memory with my assertions for some reason

I didn't quite get that sentence, but enjoy the :sun_with_face: !

wtfzdotnet: Oh, when I ran the test suite early just with the default SQLite in memory configuration i couldn't get a failing assertion to my surprise. But I was also asserting against what's still left over in the unit of work instead of really querying manually again, so i think that's what caused confusion.

wtfzdotnet:
@greg0ire think we are good now for a review :-)

composer.json
wtfzdotnet: SQLite also failed with manual querying so I think that's where I made a wrong assumption :-)

wtfzdotnet:
Also fixes #9745

tests/Doctrine/Tests/ORM/Functional/GH10747Test.php
```diff
@@ -0,0 +1,194 @@
...
+     * @Id
+     * @Column(type="Doctrine\Tests\ORM\Functional\GH10747CustomIdObjectHashType")
+     * @var CustomIdObject
+     */
+    public $id;
+
+    /**
+     * @ORM\OneToMany(targetEntity="GH10747Credit", mappedBy="article", orphanRemoval=true)
+     *
+     * @var Collection
```
greg0ire: Let's be consistent, please.

```suggestion
     * @var Collection<int, GH10747Credit>
```

greg0ire:
> think we are good now for a review 

There are PHPStan issues. See https://www.doctrine-project.org/contribute/#running-checks-before-pushing for how to run checks before pushing.

wtfzdotnet:
> > think we are good now for a review
> 
> There are PHPStan issues. See https://www.doctrine-project.org/contribute/#running-checks-before-pushing for how to run checks before pushing.

I ran this earlier locally, and again after your comment ( I did mess up phpcs, and psalm apparently ).

```
➜  orm git:(feature/fix-one-to-many-custom-id-orphan-removal) vendor/bin/phpstan
Note: Using configuration file /home/m/git/doctrine/orm/phpstan.neon.
 454/454 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


                                                                                                                        
 [OK] No errors                                                                                                         
```

PHPCS issue is fixed, inconsistency is fixed, but I cannot grasp how the PHPStan errors from the report are related to my PR, could you please double check if I am a dummy?

As for the psalm error, it seems related to the underlying existing documentation which already supports `$types`, most of the code is just borrowed from the `ManyToManyPersister` to begin with so I am not sure what I'd have to fix here, I am not so familiar with psalm (unfortunately).

greg0ire:
> PHPCS issue is fixed, incosistency is fixed, but I cannot grasp how the PHPStan errors from the report are related to my PR, could you please double check if I am a dummy?

Ah I didn't check myself, maybe it's related to a recent release, let me check.

greg0ire:
Some of it is definitely related to a recent release, I'll fix it.

wtfzdotnet:
Thanks let me know when I should rebase, and if I'd have to do anything else in regards to the psalm error :-), and thanks for your prompt feedback!

greg0ire:
Ok so you can ignore the second Psalm error, and focus on the 2 other ones, they must be caused by this. See #10754 for the other errors.

wtfzdotnet:
> 

You'd want me to just rebase on top of that branch or wait until it merges?

wtfzdotnet:
@greg0ire as I said I am really not familiar with all the ins and outs of psalm, I ran into these baseline errors, after some searching around I ran the `vendor/bin/psalm --update-baseline`. ( Assuming the baseline is ignoring stuff now that is irrelevant, which cause the errors? )



> Ok so you can ignore the second Psalm error, and focus on the 2 other ones, they must be caused by this. See #10754 for the other errors.

I updated the psalm baseline and committed this, assuming that's what is supposed to happen at this stage. As far as I understand it's a sort of "ignore stuff from here", and the errors triggered seem related. I might be misunderstanding this so if I need to do something else please give me a little more guidance, psalm noob :-).

I could imagine this might have more changes than it should, but I am not familiar as of how to treat this file.

greg0ire:
You will need to wait for #10754 to be merged. The baseline is a file that allows you to ignore errors. Setting Psalm to a very high level generates many errors, too many for us, so we baselined many of them. That way, new code is held to a very high standard, and we can reduce the baseline progressively and make old code also comply with high standards. In your case, it's OK to remove stuff from the baseline if your changes somehow fix them, or make them no longer relevant.

psalm-baseline.xml
```diff
@@ -1851,12 +1837,6 @@
     </UninitializedProperty>
   </file>
   <file src="lib/Doctrine/ORM/Query/Exec/MultiTableUpdateExecutor.php">
-    <InvalidReturnStatement>
```
greg0ire: You shouldn't touch sections for files you didn't modify, that's for sure. I suppose that was done automatically, so maybe you don't have the correct version of Psalm installed? Make sure to run `composer update`.

wtfzdotnet: Yeah so basically I'd have to go over the change of the baseline file and make sure it only fixes stuff in-scope of my pr, gotcha. I'll sort it out.

wtfzdotnet:
> You will need to wait for #10754 to be merged. The baseline is a file that allows you to ignore errors. Setting Psalm to a very high level generates many errors, too many for us, so we baselined many of them. That way, new code is held to a very high standard, and we can reduce the baseline progressively and make old code also comply with high standards. In your case, it's OK to remove stuff from the baseline if your changes somehow fix them, or make them no longer relevant.

Assumed that was the case, thanks for taking the time to explain :-)

lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
```diff
@@ -166,24 +166,26 @@ public function loadCriteria(PersistentCollection $collection, Criteria $criteri
     }
 
     /** @throws DBALException */
-    private function deleteEntityCollection(PersistentCollection $collection): int
```
wtfzdotnet: psalm kept whining about the return type either being `int|string`, given php version i changed these into `mixed`

greg0ire: That won't work on old versions of PHP

greg0ire: I think the right way to deal with this is to store the result in a variable, then assert it's an integer, and not touch the return type.

wtfzdotnet: Will fix it after lunch break, thanks.

wtfzdotnet: @greg0ire got distracted for a bit there :-), making assertions on the integers doesn't satisfy psalm, so I think this should simply be ignored in either the baseline or another way?

greg0ire: Can't see why it wouldn't satisfy Psalm. Can you show me what you tried, and what's the result, by pushing your attempt to this branch?

wtfzdotnet: @greg0ire assumed this is what you meant, which results in

```
ERROR: MissingReturnType - lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php:169:22 - Method Doctrine\ORM\Persisters\Collection\OneToManyPersister::deleteEntityCollection does not have a return type, expecting int|string (see https://psalm.dev/050)
    private function deleteEntityCollection(PersistentCollection $collection)


ERROR: RedundantConditionGivenDocblockType - lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php:190:40 - Docblock-defined type string for $numAffected is always string (see https://psalm.dev/156)
        assert(is_int($numAffected) || is_string($numAffected));


ERROR: MissingReturnType - lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php:203:22 - Method Doctrine\ORM\Persisters\Collection\OneToManyPersister::deleteJoinedEntityCollection does not have a return type, expecting int|string (see https://psalm.dev/050)
    private function deleteJoinedEntityCollection(PersistentCollection $collection)


ERROR: RedundantConditionGivenDocblockType - lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php:256:39 - Docblock-defined type string for $numDeleted is always string (see https://psalm.dev/156)
        assert(is_int($numDeleted) || is_string($numDeleted));
```

lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
```diff
@@ -166,24 +166,30 @@ public function loadCriteria(PersistentCollection $collection, Criteria $criteri
...
+            $types[]      = PersisterHelper::getTypeOfColumn($joinColumn['referencedColumnName'], $sourceClass, $this->em);
         }
 
         $statement = 'DELETE FROM ' . $this->quoteStrategy->getTableName($targetClass, $this->platform)
             . ' WHERE ' . implode(' = ? AND ', $columns) . ' = ?';
 
-        return $this->conn->executeStatement($statement, $parameters);
+        $numAffected = $this->conn->executeStatement($statement, $parameters, $types);
+
+        assert(is_int($numAffected) || is_string($numAffected));
```
greg0ire: I meant this:

```suggestion
        assert(is_int($numAffected));
```

In that case, you will have to ignore an error about the possibility of `$numAffected` being a string, in case it's gigantic I guess.

psalm-baseline.xml
```diff
@@ -1236,14 +1236,6 @@
...
             ]]]></code>
     </InvalidArrayOffset>
-    <InvalidReturnStatement>
-      <code>$numDeleted</code>
-      <code><![CDATA[$this->conn->executeStatement($statement, $parameters)]]></code>
-    </InvalidReturnStatement>
-    <InvalidReturnType>
-      <code>int</code>
-      <code>int</code>
-    </InvalidReturnType>
```
greg0ire, line 1246: This should probably be kept once you've addressed my comment above. Part of this, at least.

wtfzdotnet: @greg0ire at that stage ( just resetting the InvalidReturnType node ), I am left with;

```
ERROR: MissingReturnType - lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php:169:22 - Method Doctrine\ORM\Persisters\Collection\OneToManyPersister::deleteEntityCollection does not have a return type, expecting int (see https://psalm.dev/050)
    private function deleteEntityCollection(PersistentCollection $collection)


ERROR: MissingReturnType - lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php:203:22 - Method Doctrine\ORM\Persisters\Collection\OneToManyPersister::deleteJoinedEntityCollection does not have a return type, expecting int (see https://psalm.dev/050)
    private function deleteJoinedEntityCollection(PersistentCollection $collection)


ERROR: UnusedBaselineEntry - lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php:0:0 - Baseline for issue "InvalidReturnType" has 2 extra entries. (see https://psalm.dev/316)
```

greg0ire: Well now that you asserted that it's an integer, you should add a return type declaration at least to `deleteEntityCollection`, right? Same for the other method.

wtfzdotnet: Ok that made me feel like a clown, I was presuming it was nagging about the return type hints again

wtfzdotnet:
phpstan happy, psalm happy, phpcs happy, I hope we are finally good now :pray:

lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
```diff
@@ -165,25 +166,35 @@ public function loadCriteria(PersistentCollection $collection, Criteria $criteri
...
     }
 
-    /** @throws DBALException */
-    private function deleteEntityCollection(PersistentCollection $collection): int
+    /**
+     * @return int
+     *
+     * @throws DBALException
+     */
+    private function deleteEntityCollection(PersistentCollection $collection)
```
greg0ire: :thinking: I don't understand why this was not reverted.

wtfzdotnet: The @throws  ?

greg0ire: The removal of the type declaration.

wtfzdotnet: Apologies, really haven't been having my day, bad night of sleep :sleeping:

wtfzdotnet:
Let me know if I still need to squash these commits, sorry about the going back and forth :heart:

lib/Doctrine/ORM/Persisters/Collection/OneToManyPersister.php
```diff
@@ -247,6 +254,8 @@ private function deleteJoinedEntityCollection(PersistentCollection $collection):
 
         $this->conn->executeStatement($statement);
 
+        assert(is_int($numDeleted));
```
greg0ire, line 257: Did you test this locally?

greg0ire:
> Let me know if I still need to squash these commits, sorry about the going back and forth heart

No worries. I think there should be 2 commits:
- one commit that adds the assertions and changes the baseline
- the commit with the changes you wanted to add

That's because right now, I don't think there is a relation between the 2 changes. Let me know if you need help going from 3 to 2 commits.

wtfzdotnet:
> > Let me know if I still need to squash these commits, sorry about the going back and forth heart
> 
> No worries. I think there should be 2 commits:
> 
> * one commit that adds the assertions and changes the baseline
> * the commit with the changes you wanted to add
> 
> That's because right now, I don't think there is a relation between the 2 changes. Let me know if you need help going from 3 to 2 commits.

@greg0ire okay that's done as well :sweat:

greg0ire:
The code looks good, I think. The commit message should be improved, so you should probably read https://www.doctrine-project.org/contribute/#crafting-meaningful-commit-messages

Specifically, the commit message only mentions the problem you're addressing, but does not detail what happens (do you have an error message? Or nothing happens. Also, it does not explain why why the issue exists, what you are doing in this PR, and why that fixes the issue.

Maybe what's happening is that at some point, the ORM is comparing objects with their representation in the database, when it should really convert the objects first, is that it?

wtfzdotnet:
> The code looks good, I think. The commit message should be improved, so you should probably read https://www.doctrine-project.org/contribute/#crafting-meaningful-commit-messages
> 
> Specifically, the commit message only mentions the problem you're addressing, but does not detail what happens (do you have an error message? Or nothing happens. Also, it does not explain why why the issue exists, what you are doing in this PR, and why that fixes the issue.
> 
> Maybe what's happening is that at some point, the ORM is comparing objects with their representation in the database, when it should really convert the objects first, is that it?

As was in the example I've given before with the gnarly ramsey/uuid implementation ( for the test case ), basically what happens is a custom doctrine type of Uuid object is converted to string by simply casting, resulting in a hex `DELETE FROM x WHERE id = ?` query, whilst it should've been converted along the way to it's  binary representation ( in our database choice case ). This leads to no deletions being made at all as you would expect making use of doctrine custom type's as an identifier.

wtfzdotnet:
Rewrote last commit's message to address this.

greg0ire (approved):

SenseException (approved):

norkunas:
And #10075 just was ignored, thanks guys

greg0ire:
Thanks @wtfzdotnet !

@norkunas , that's unfortunate your PR got ignored, but I don't see why you feel entitled to complain about it. I mean, we're not paid to review all incoming PRs, are we?

norkunas:
I don't complain. I stated a fact, and then thanked. At least after so long somebody gave a duck and completed this

greg0ire:
If you don't complain then we're cool :slightly_smiling_face: We in the team get drowned daily in a shitload of notifications, and it's indeed cool that people don't get discouraged by this and stuff ends up getting fixed. So thank you both for caring!

PR #10748, opened 2023-06-02
Update docs that `name` in @Index/@UniqueConstraint is optional

### LINKED ISSUE #4283: DDC-3473: @Index and @UniqueConstraint do not require the name parameter
Jira issue originally created by user benjamin:

The docs for [@Index](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/annotations-reference.html#annref-index) and [@UniqueConstraint](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/annotations-reference.html#uniqueconstraint) both say:

{quote}
Required attributes:
name: Name of the Index
{quote}

However, the `name` parameter does not seem to be required in practice. Omitting it makes Doctrine auto-generate a hash-based name, which is consistent with the indexes generated for foreign keys and is therefore good IMO.

If this is not just a side effect and omitting the `name` parameter is actually supported, it would be nice to update the docs to reflect this.

mpdude:
Fixes #4283.

### DIFF
--- a/docs/en/reference/annotations-reference.rst
+++ b/docs/en/reference/annotations-reference.rst
@@ -545,12 +545,12 @@ has meaning in the SchemaTool schema generation context.
 Required attributes:
 
 
--  **name**: Name of the Index
 -  **fields**: Array of fields. Exactly one of **fields**, **columns** is required.
 -  **columns**: Array of columns. Exactly one of **fields**, **columns** is required.
 
 Optional attributes:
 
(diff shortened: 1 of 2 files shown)


greg0ire (approved):

mpdude:
In fact, I checked that omitting the `name` in the `@Annotation` style configuration works, and that a provided name is picked up – at least when dumping the schema configuration.

SenseException (approved):

PR #10754, opened 2023-06-03
Fix build

greg0ire:
This is caused by recent static analysis improvements made to `doctrine/deprecations`.


SenseException (approved):
Build fixed.

PR #10755, opened 2023-06-04
Add test to show #9192 has been fixed

mpdude:
This test implements the situation described in #9192. The commit order computation will be fixed by #10547.


greg0ire:
Thanks @mpdude !

PR #10759, opened 2023-06-05
Don't call deprecated getSQLResultCasing and usesSequenceEmulatedIdentityColumns when we know the platform

nicolas-grekas:
Saves triggering the deprecation on PHP 7 where only DBAL v2 can be used:

> "AbstractPlatform::getSQLResultCasing is deprecated without replacement and removed in DBAL 3.Use Portability\\Connection with PORTABILITY_FIX_CASE to get portable result cases. (AbstractPlatform.php:3515 called by SQLResultCasing.php:30, https://github.com/doctrine/dbal/pull/4229, package doctrine/dbal)": 246,

### DIFF
--- a/lib/Doctrine/ORM/Internal/SQLResultCasing.php
+++ b/lib/Doctrine/ORM/Internal/SQLResultCasing.php
@@ -9,7 +9,9 @@ use Doctrine\DBAL\Platforms\DB2Platform;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
 
+use function get_class;
 use function method_exists;
+use function strpos;
 use function strtolower;
 use function strtoupper;
 
@@ -26,7 +28,7 @@ trait SQLResultCasing
             return strtolower($column);
         }
 
-        if (method_exists(AbstractPlatform::class, 'getSQLResultCasing')) {
+        if (strpos(get_class($platform), 'Doctrine\\DBAL\\Platforms\\') !== 0 && method_exists(AbstractPlatform::class, 'getSQLResultCasing')) {
             return $platform->getSQLResultCasing($column);
         }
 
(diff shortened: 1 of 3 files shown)


greg0ire:
Hi! Before you spend too much effort on this: I think [this comment](https://github.com/doctrine/orm/pull/10760#issuecomment-1576641483) applies here too: no one should be using PHP 7 anymore, and the same holds true for DBAL 2. That being said, there are already a bunch of `if ($platform instanceof…)` in here, so it's maybe not that straightforward. @derrabus , what your thoughts on this?

nicolas-grekas:
That's not the same. This covers the situation where DBAL v2 is used with ORM v2.15. Right now, ORM calls a deprecated API of DBALv2 while it can and thus should avoid doing so when possible, which is the case covered here.

greg0ire:
Ok well, you managed to get a green build anyway 👍
@derrabus authored that piece of code, so I'll defer to him.

derrabus:
The rationale behind keeping the call was that the platform classes are in fact extension points in DBAL. You can override `SQLServerPlatform` if you want to change how the DBAL interacts with SQL Server for instance. A consumer of DBAL could override `getSQLResultCasing()` in its own platform.

Your `instanceof` checks would not detect that case and break such an app, I'm afraid.

nicolas-grekas:
That's why I checked explicitly the platforms we know about: changing a child class won't change the behavior of the DB engine, so that there is nothing to allow customizing in this regard for the listed platforms.

nicolas-grekas:
PR updated to also skip calling deprecated usesSequenceEmulatedIdentityColumns in ClassMetadataFactory

derrabus:
> there is nothing to allow customizing in this regard for the listed platforms.

I agree and that's probably one of the reasons this method has been removed. I wonder if we can instead find a better way to silence that deprecation instead.

nicolas-grekas:
Why would we need another way? We know the server's behavior => done.

nicolas-grekas:
(just rebased)

derrabus (approved):

PR #10760, opened 2023-06-05
Trigger deprecation in AnnotationDriver only if PHP >= 8.0

fancyweb:
Hello, I think it would make sense to not trigger this deprecation when it's impossible to migrate to the new AttributeDriver because of the PHP version. The deprecation is not actionable: I don't really consider switching to the XML driver a viable migration.
People running PHP < 8.0 won't be able to use Doctrine 3.0 anyway since it requires PHP 8.1.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php
@@ -28,6 +28,8 @@ use function get_class;
 use function is_array;
 use function is_numeric;
 
+use const PHP_VERSION_ID;
+
 /**
  * The AnnotationDriver reads the mapping metadata from docblock annotations.
  */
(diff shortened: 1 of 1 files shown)


greg0ire (dismissed):
Sounds sensible.

derrabus:
tbh, I'd leave everything as it is. The deprecation can be silenced individually, if it annoys you.

> The deprecation is not actionable: I don't really consider switching to the XML driver a viable migration.

Well, first of all: yes, XML *is* a viable option. Apart from that, the preferred migration path would be to upgrade to PHP 8 and switch to attributes. And I consider this path to be actionable and if it isn't, the impediment is not on the ORM's side.

> People running PHP < 8.0 won't be able to use Doctrine 3.0 anyway since it requires PHP 8.1.

They don't _have to_ fix any deprecations then for as long as it's "impossible" for them to migrate to ORM 3. I realize that not every deprecation is trivial to fix and deprecations don't need to be fixed right away. That is why the deprecation framework that we use allows apps to silence deprecations by identifier.

Really nobody should be on PHP 7 anymore. And that deprecation could even add more arguments for moving forward to PHP 8. If we'd hide it on PHP 7, people might be surprised to see new deprecations after having upgraded to PHP 8.

fancyweb:
Thanks for your fast reviews, I'll close.

PR #10762, opened 2023-06-05
Don't call canEmulateSchemas in SchemaTool when possible

nicolas-grekas:


greg0ire (approved):
It makes sense not calling `canEmulateSchemas` if all assets are in the default schema, because in that case there are no assets to remove.

derrabus (approved):

PR #10763, opened 2023-06-05
Defer removing removed entities from to-many collections until after transaction commit

### LINKED ISSUE #10752: No ForeignKeyConstraintViolationException exception when trying to remove an entity
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | kinda
| Version     | 2.15.2

#### Summary

Since the following change
CleanShot 2023-06-03 at 08 04 00@2x
Doctrine stopped throwing `ForeignKeyConstraintViolationException` and (from what I noticed) it removes the entity from the database. What's more interesting, if we add the deleted lines, and leave the same code at the end of the method – it doesn't work. So putting this `if` statement somehow breaks the expected behavior.

#### Current behavior

No `ForeignKeyConstraintViolationException` 🤷🏼‍♂️.
#### How to reproduce

#10753

#### Expected behavior

`ForeignKeyConstraintViolationException` should be thrown.

mpdude:
This PR fixes regressions introduced in 2.15.1 related to how the `UnitOfWork` and `EntityPersister`s implement removal of entities that are part of many-to-many collections.

#### Background

To better understand the problem and regressions caused, we need to take a look at how the ORM deals with many-to-many associations when an entity shall be removed that is part of such an association.

`BasicEntityPersister::deleteJoinTableRecords()` is the method responsible for efficient removal of join-table rows in the database when a single entity is removed. It will `DELETE` all join-table rows referring to the entity in question before the entity is removed itself. This only works when the many-to-many association is declared on the entity's side (ie. its a bi-directional many-to-many association, or the entity is the owning side). Otherwise, since the ORM cannot detect associations that are declared in other classes only, join tables are left alone and database-level constraints apply.

`BasicEntityPersister::deleteJoinTableRecords()` implements a rather efficient join-table row removal, since a single `DELETE` query per removed entity and association can be used to possibly delete multiple rows (for associations with different other entities).

Besides deleting join-table rows in the database, we also want to update collections to no longer contain the removed entity. Removing the entity from these collections is the in-memory equivalent to removing the join table rows in the database.

Early during the commit phase, the `UnitOfWork::computeAssociationChanges()` method is responsible for looking at all associations in all in-memory entities. Among a few other tasks, it will identify all "dirty" collections – those that have been changed and need to be flushed to the database. This method will also check all collections for entities pending removal, and it will remove (unset) these entities from the collections.

#### Current situation

#10485 complained that the `computeAssociationChanges` would first take note of the "dirty" collections and then unset entities pending removal. The consequences of this were that 
* in collections that were already dirty (changed) _before_ `flush()` was called, the join-table row would be removed with a dedicated `DELETE` statement, while bringing the collection and database in sync
* more importantly, previously clean collections would be dirty after `flush()` returns (the opposite of what `flush()` is supposed to achieve), and subsequent `flush()` operations would fail since a `DELETE` statement would be issued for an entity that was no longer in the identity map.

The fix in #10486 (contained in 2.15.1) tried to address this by _first_ unsetting the to-be-removed entities _before_ taking note of (then) dirty collections. This caused unintended behavior changes that were not spotted by tests:
* Since collection updates are done first, we now use single `DELETE` statements for every join-table row deletion. `BasicEntityPersister::deleteJoinTableRecords()` was effectively bypassed for those collections that were loaded into memory at that time. More queries are used than necessary.
* #10752 reports a use case where the to-be-removed entity _does not_ declare the many-to-many association. Pre #10486, this would always mean database-level constraints apply, since the ORM does not touch join-table rows in that case. But with the #10486 changes _and_ given a collection containing the to-be-removed entity is in memory, we ended up doing join-table row removals first – circumventing database-level constraints.

#### Suggested fix

* Improve `ManyToManyBasicAssociationTest` to make sure collections are clean after `flush()` returns, and include assertions regarding the expected query count. This would have spotted per-row `DELETE`s instead of using the optimized persister method
* Do not modify collections during `computeAssociationChanges`, but instead track which entity needs to be removed from which collection in a dedicated array
* This way, join-tables will be dealt with by the `deleteJoinTableRecords` method only
* After all changes have been flushed successfully, in the finishing phase of the UoW commit, go back to the list of tracked entity removals. Remove the entities from the respective collections immediately before taking new snapshots (= defining the new "clean" state). This makes sure we only do in-memory cleanups and cause no additional database operations due to removals.
* Improve documentation to raise awareness of how this proess works

Fixes #10752, closes #10753.

### DIFF
--- a/docs/en/reference/working-with-objects.rst
+++ b/docs/en/reference/working-with-objects.rst
@@ -286,16 +286,52 @@ as follows:
 After an entity has been removed, its in-memory state is the same as
 before the removal, except for generated identifiers.
 
-Removing an entity will also automatically delete any existing
-records in many-to-many join tables that link this entity. The
-action taken depends on the value of the ``@joinColumn`` mapping
-attribute "onDelete". Either Doctrine issues a dedicated ``DELETE``
-statement for records of each join table or it depends on the
-foreign key semantics of onDelete="CASCADE".
+During the ``EntityManager#flush()`` operation, the removed entity
+will also be removed from all collections in entities currently
+loaded into memory.
+
+.. _remove_object_many_to_many_join_tables:
+
+Join-table management when removing from many-to-many collections
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Regarding existing rows in many-to-many join tables that refer to
+an entity being removed, the following applies.
+
+When the entity being removed does not declare the many-to-many association
+itself (that is, the many-to-many association is unidirectional and
+the entity is on the inverse side), the ORM has no reasonable way to
+detect associations targeting the entity's class. Thus, no ORM-level handling
+of join-table rows is attempted and database-level constraints apply.
+In case of database-level ``ON DELETE RESTRICT`` constraints, the
+``EntityManager#flush()`` operation may abort and a ``ConstraintViolationException``
+may be thrown. No in-memory collections will be modified in this case.
+With ``ON DELETE CASCADE``, the RDBMS will take care of removing rows
+from join tables.
+
+When the entity being removed is part of bi-directional many-to-many
+association, either as the owning or inverse side, the ORM will
+delete rows from join tables before removing the entity itself. That means
+database-level ``ON DELETE RESTRICT`` constraints on join tables are not
+effective, since the join table rows are removed first. Removal of join table
+rows happens through specialized methods in entity and collection persister
+classes and take one query per entity and join table. In case the association
+uses a ``@JoinColumn`` configuration with ``onDelete="CASCADE"``, instead
+of using a dedicated ``DELETE`` query the database-level operation will be
+relied upon.
+
+.. note::
+
+    In case you rely on database-level ``ON DELETE RESTRICT`` constraints,
+    be aware that by making many-to-many associations bidirectional the
+    assumed protection may be lost.
+
+
+Performance of different deletion strategies
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Deleting an object with all its associated objects can be achieved
 in multiple ways with very different performance impacts.
-
 
 1. If an association is marked as ``CASCADE=REMOVE`` Doctrine ORM
    will fetch this association. If its a Single association it will
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -24,6 +24,7 @@ use Doctrine\ORM\Event\PrePersistEventArgs;
 use Doctrine\ORM\Event\PreRemoveEventArgs;
 use Doctrine\ORM\Event\PreUpdateEventArgs;
 use Doctrine\ORM\Exception\ORMException;
+use Doctrine\ORM\Exception\UnexpectedAssociationValue;
 use Doctrine\ORM\Id\AssignedGenerator;
 use Doctrine\ORM\Internal\CommitOrderCalculator;
 use Doctrine\ORM\Internal\HydrationCompleteHandler;
@@ -955,6 +956,15 @@ class UnitOfWork implements PropertyChangedListener
 
             $state = $this->getEntityState($entry, self::STATE_NEW);
 
+            if (! ($entry instanceof $assoc['targetEntity'])) {
+                throw UnexpectedAssociationValue::create(
+                    $assoc['sourceEntity'],
+                    $assoc['fieldName'],
+                    get_debug_type($entry),
+                    $assoc['targetEntity']
+                );
+            }
+
             switch ($state) {
                 case self::STATE_NEW:
                     if (! $assoc['isCascadePersist']) {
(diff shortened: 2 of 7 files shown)


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -916,6 +934,14 @@ private function computeAssociationChanges(array $assoc, $value): void
...
         }
 
+        // If this collection is dirty, schedule it for updates
+        if ($value instanceof PersistentCollection && $value->isDirty()) {
+            $coid = spl_object_id($value);
+
+            $this->collectionUpdates[$coid]  = $value;
+            $this->visitedCollections[$coid] = $value;
+        }
+
```
mpdude, line 945: This reverts the main change from #10486.

docs/en/reference/working-with-objects.rst
```diff
@@ -286,17 +286,53 @@ as follows:
...
+may be thrown. No in-memory collections will be modified in this case.
+With ``ON DELETE CASCADE``, the RDBMS will take care of removing rows
+from join tables.
+
+When the entity being removed is part of bi-directional many-to-many
+associations, either as the owning or inverse side, the ORM will
+delete rows from join tables before removing the entity itself. That means
+database-level ``ON DELETE RESTRICT`` constraints on join tables are not
+effective, since the join table rows are removed first. Removal of join table
+rows happens through specialized methods in entity and collection persisterr
```
greg0ire:
```suggestion
rows happens through specialized methods in entity and collection persister
```

docs/en/reference/working-with-objects.rst
```diff
@@ -286,17 +286,53 @@ as follows:
...
+detect associations targeting the entity's class. Thus, no ORM-level handling
+of join-table rows is attempted and database-level constraints apply.
+In case of database-level ``ON DELETE RESTRICT`` constraints, the
+``EntityManager#flush()`` operation may abort and a ``ConstraintViolationException``
+may be thrown. No in-memory collections will be modified in this case.
+With ``ON DELETE CASCADE``, the RDBMS will take care of removing rows
+from join tables.
+
+When the entity being removed is part of bi-directional many-to-many
+associations, either as the owning or inverse side, the ORM will
```
greg0ire:
```suggestion
association, either as the owning or inverse side, the ORM will
```

docs/en/reference/working-with-objects.rst
```diff
@@ -286,17 +286,53 @@ as follows:
...
+classes and take one query per entity and join table. In case the association
+uses a ``@JoinColumn`` configuration with ``onDelete="CASCADE"``, instead
+of using a dedicated ``DELETE`` query the database-level operation will be
+relied upon.
+
+.. note::
+
+    In case you rely on database-level ``ON DELETE RESTRICT`` constraints,
+    be aware that by making many-to-many associations bidirectional the
+    assumed protection may be lost.
```
greg0ire, line 327: :bulb: it might be interesting to warn about this when validating a schema.

tests/Doctrine/Tests/ORM/Functional/GH10752Test.php
```diff
@@ -0,0 +1,157 @@
...
+
+    private function createOrder(): GH10752Order
+    {
+        return new GH10752Order();
+    }
+
+    private function createPromotion(): GH10752Promotion
+    {
+        return new GH10752Promotion();
+    }
```
greg0ire: What is the point of these methods?

tests/Doctrine/Tests/ORM/Functional/GH10752Test.php
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/**
+ * @ORM\Entity
+ */
+class GH10752Order
```
greg0ire, line 77: Some methods below are never used. Let's keep entities as concise as possible please.

tests/Doctrine/Tests/ORM/Functional/GH10752Test.php
mpdude: *shrugs* I just picked that from #10752 without thinking about it

mpdude: I'll change it

docs/en/reference/working-with-objects.rst
mpdude: Could you create an issue please if you think we should pursue that? It'll get lost here.

greg0ire: See https://github.com/doctrine/orm/issues/10783

greg0ire (approved):

mpdude:
Do you think there should be a mapping configuration setting/switch to tell the ORM not to handle the join table itself and always leave it to the database layer?

Probably that has to be a completely new setting. Might be beneficial for performance, to get consistent behavior regardless of using uni/bidirectional associations and to be able to `RESTRICT` removing entities that have m2m links.

SenseException (approved):

greg0ire:
@mpdude I don't know what level of flexibility people expect, or what's the sensible thing to expect here. I think we should keep the ORM as simple as possible. If people let us know they need this, we can always implement it.

greg0ire:
Thanks @mpdude !

PR #10764, opened 2023-06-05
Cleanup bc layers

greg0ire:


lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -157,14 +158,12 @@ public function setInflector(Inflector $inflector): void
     public function loadMetadataForClass(string $className, PersistenceClassMetadata $metadata): void
     {
         if (! $metadata instanceof ClassMetadata) {
-            Deprecation::trigger(
-                'doctrine/orm',
-                'https://github.com/doctrine/orm/pull/249',
-                'Passing an instance of %s to %s is deprecated, please pass a %s instance instead.',
-                $metadata::class,
+            throw new TypeError(sprintf(
+                'Argument #2 passed to %s() must be an instance of %s, %s given.',
```
greg0ire, line 163: I didn't provide an upgrade note for that one because it's not precisely a realistic scenario to provide anything else than an ORM metadata instance to this method.

lib/Doctrine/ORM/Mapping/Driver/DatabaseDriver.php
```diff
@@ -157,14 +158,12 @@ public function setInflector(Inflector $inflector): void
...
                 __METHOD__,
                 ClassMetadata::class,
-            );
+                $metadata::class,
```
derrabus:
```suggestion
                get_debug_type($metadata),
```

When embedding a class into a message, we should prefer `get_debug_type()` over `::class` because the former plays more nicely with anonymous classes.

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
```diff
@@ -34,22 +33,14 @@ class AttributeDriver implements MappingDriver
 
     private readonly AttributeReader $reader;
 
-    /** @param array<string> $paths */
+    /**
+     * @param array<string> $paths
+     * @param bool          $reportFieldsWhereDeclared no-op, to be removed in 4.0
+     */
     public function __construct(array $paths, bool $reportFieldsWhereDeclared = false)
```
derrabus:
```suggestion
     * @param true          $reportFieldsWhereDeclared no-op, to be removed in 4.0
     */
    public function __construct(array $paths, bool $reportFieldsWhereDeclared = true)
```

Also, let's throw if `false` has been passed explicitly.

derrabus (approved):

derrabus (changes requested):
Wrong button. I should go to bed. 🙈

lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php
greg0ire: Or even better, let's use `func_num_args()`

greg0ire: Let's switch to checking `func_num_args()` when deprecating this on 3.1.x

greg0ire:
@mpdude please review

derrabus (approved):

mpdude (commented):
A bit too late, but 👍🏻 from me

greg0ire:
Thanks for reviewing anyway :)

PR #10767, opened 2023-06-06
Limit Count in Entity persister

### LINKED ISSUE #10766: Avoid Using unqualifed COUNT in EntityPersister
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

Might be a dupe of #5960

Using Symfony. I switched from using `findBy` to `matching` in my services. I then checked the profiler for an unrelated feature I was working on and noticed Doctrine was issuing a `COUNT (*)` query. That's a bit concerning for larger tables, so I dug into the code and traced it to `getCountSql`

https://github.com/doctrine/orm/blob/3827dd769edf25d59fa29a7d63efd6ba15b3a789/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1136-L1156

If the $criteria is a Criteria object and doesn't have any expressions, this will issue a count against the entire table. I think to help avoid this, it should check if maxResults is set like so (rough example):
```php
$sql =  'SELECT COUNT(*) '
        . 'FROM ' . $tableName . ' ' . $tableAlias
        . (empty($conditionSql) ? '' : ' WHERE ' . $conditionSql);

if ($criteria instanceof Criteria and null !== $criteria->getMaxResults()) {
    $sql .= ' LIMIT ' . $criteria->getMaxResults();
}

return $sql;
```
I don't know if that check could go in one of the helper methods used in here, but it would save some io when the actual result set is bigger than the given limit.

Aweptimum:
Closes #10766

I wasn't sure how to test it beyond following what was already in there, so if more rigorous testing is desired please suggest what you'd like to see. That said, I think it's pretty straightforward. If the maintainers would prefer, I can put the added code in a helper method (`getCriteriaLimit`) to make it less ugly.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterTypeValueSqlTest.php
+++ b/tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterTypeValueSqlTest.php
@@ -149,6 +149,11 @@ class BasicEntityPersisterTypeValueSqlTest extends OrmTestCase
         $criteria  = new Criteria(Criteria::expr()->eq('value', 'bar'));
         $statement = $persister->getCountSQL($criteria);
         self::assertEquals('SELECT COUNT(*) FROM "not-a-simple-entity" t0 WHERE t0."simple-entity-value" = ?', $statement);
+
+        // Using a criteria object w/ Limit
+        $criteria  = new Criteria(Criteria::expr()->eq('value', 'bar'), null, null, 5000);
+        $statement = $persister->getCountSQL($criteria);
+        self::assertEquals('SELECT COUNT(*) FROM "not-a-simple-entity" t0 WHERE t0."simple-entity-value" = ? LIMIT 5000', $statement);
     }
 
     public function testCountEntities(): void
--- a/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
+++ b/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php
@@ -1150,9 +1150,15 @@ class BasicEntityPersister implements EntityPersister
                 : $filterSql;
         }
 
-        return 'SELECT COUNT(*) '
-            . 'FROM ' . $tableName . ' ' . $tableAlias
-            . (empty($conditionSql) ? '' : ' WHERE ' . $conditionSql);
+        $sql =  'SELECT COUNT(*) '
+        . 'FROM ' . $tableName . ' ' . $tableAlias
+        . (empty($conditionSql) ? '' : ' WHERE ' . $conditionSql);
+
+        if ($criteria instanceof Criteria && $criteria->getMaxResults() !== null) {
+            $sql .= ' LIMIT ' . $criteria->getMaxResults();
+        }
+
+        return $sql;
     }
 
     /**


greg0ire:
Although the change looks sensible, I can imagine how it would break a pagination system based on an `EntityRepository`, where you would use the same criteria to get the item of the current page, but also to get a count.

Aweptimum:
That's fair. In my use case I have timeseries data with billions of records, so I'm using keyset pagination. However, I'm also using timescaledb (built on postgres) to store it, so I could create an aggregate that keeps an updated count of records for my history data and avoid having to wait 5 minutes for the query to finish.

tests/Doctrine/Tests/ORM/Persisters/BasicEntityPersisterTypeValueSqlTest.php
```diff
@@ -149,6 +149,11 @@ public function testCountCondition(): void
         $criteria  = new Criteria(Criteria::expr()->eq('value', 'bar'));
         $statement = $persister->getCountSQL($criteria);
         self::assertEquals('SELECT COUNT(*) FROM "not-a-simple-entity" t0 WHERE t0."simple-entity-value" = ?', $statement);
+
+        // Using a criteria object w/ Limit
+        $criteria  = new Criteria(Criteria::expr()->eq('value', 'bar'), null, null, 5000);
+        $statement = $persister->getCountSQL($criteria);
+        self::assertEquals('SELECT COUNT(*) FROM "not-a-simple-entity" t0 WHERE t0."simple-entity-value" = ? LIMIT 5000', $statement);
```
greg0ire, line 156: Wait a minute… what will this return if there are 5001 elements? 5001, right? If yes, what's the point?

Aweptimum: Yes it'll return 5001. I'll have to test it better tomorrow, but from my understanding - if the user passes a criteria with a limit to `matching`, the returned collection won't contain more items than specified by the limit. But the count doesn't use the limit, so I think the below scenario is entirely possible:
```php
$c = new Criteria(null, null, null, 5000);
$results = $repository->matching($c);
echo $results->count(); // 2000000
```

greg0ire: Ok, then what's the point?

Aweptimum: Of adding the Limit? Performance. I have a default limit of 10k records for a query, which should be fast, but checking the size of the collection (using `isEmpty`) triggers ` SELECT COUNT(*)` on the entire table. But I would also expect the collection, which is a result of the criteria, to have no more records in it than the maxResults set in the criteria (if it is set). It if isn't set, then yeah the user is asking for a count of the whole table.

greg0ire: You say "`SELECT COUNT(*)` on the entire table" But `SELECT COUNT(*) LIMIT 1` is also on the entire table, and the `LIMIT 1` part seems useless here. I can't think how the `LIMIT 1` would improve performance here, because it just counts the entire table, then returns a row containing the count, which will always be just the one row.

Aweptimum: Ohhh, right, yes. Sorry for being dense, I haven't gotten enough sleep the past few days.
Yeah, to be performant it would have to use a subquery
```sql
SELECT COUNT(*) FROM (SELECT * FROM table WHERE ... LIMIT 1) as f
```
I could refactor it to do that - do you think that's a good idea?

greg0ire: I don't think it's a good idea either, sorry. I mean, it might be performant, but it is also useless as soon as the limit kicks in.

PR #10776, opened 2023-06-15
Fix missing setFilterSchemaAssetsExpression in phpdoc

dmitryuk:
Starting from `doctrine/dbal`  `2.9.0` [MR](https://github.com/doctrine/dbal/pull/3316) method `setFilterSchemaAssetsExpression` is deprecated and finally removed in `3.0.0` [MR](https://github.com/doctrine/dbal/pull/3518)

The PR updates docs to the current schema

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/ConvertMappingCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ConvertMappingCommand.php
@@ -80,7 +80,7 @@ on a global level:
             $assetName = $assetName->getName();
         }
 
-        return str_starts_with($assetName, 'audit_');
+        return !str_starts_with($assetName, 'audit_');
     });
 EOT
              );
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Tools/Console/Command/ConvertMappingCommand.php
```diff
@@ -75,7 +75,13 @@ protected function configure()
 by the ORM, you can use a DBAL functionality to filter the tables and sequences down
 on a global level:
 
-    $config->setFilterSchemaAssetsExpression($regexp);
+    $config->setSchemaAssetsFilter(function (string|AbstractAsset $assetName) use ($regexp): bool {
+        if ($assetName instanceof AbstractAsset) {
+            $assetName = $assetName->getName();
+        }
+
+        return (bool) preg_match($regexp, $assetName);
```
greg0ire: I don't think we should stick with a regexp example:

- it involves an undefined variable, which we pass via `use`;
- it involves a boolean cast.

Can you please use something more simple, like maybe `in_array`, or `str_starts_with`?

dmitryuk: @greg0ire done, take a look again, please

greg0ire (approved):

greg0ire:
Thanks @dmitryuk !

PR #10777, opened 2023-06-17
Turn deprecation into an exception

greg0ire:
I think the upgrade entry introduced in https://github.com/doctrine/orm/pull/10520 already covers this breaking change.


greg0ire:
@mpdude please review

mpdude (approved):

PR #10780, opened 2023-06-18
Avoid partial

greg0ire:
Partial objects are deprecated. They were handy to make the generated
SQL more legible, but we should refrain from relying on them.


SenseException (approved):
Is this a partial change to avoid partial objects and will further changes follow? This PR just changes a test class on line that contained `ORMQuery::HINT_FORCE_PARTIAL_LOAD`

greg0ire:
> Is this a partial change 

lol I see what you did here :wink: 

There might be more PRs against 2.15.x, if I notice more tests need to be migrated while attempting to remove PARTIAL from 3.0.x IIRC this was the biggest source of changes.

PR #10785, opened 2023-06-22
Add a safeguard against multiple objects competing for the same identity map entry

mpdude:
While trying to understand #3037, I found that it may happen that we have more entries in `\Doctrine\ORM\UnitOfWork::$entityIdentifiers` than in `\Doctrine\ORM\UnitOfWork::$identityMap`.

The former is a mapping from `spl_object_id` values to ID hashes, the latter an array first of entity class names and then from ID hash to entity object instances.

(Basically, "ID hash" is a concatenation of all field values making up the `@Id` for a given entity.)

This means that at some point, we must have _different_ objects representing the same entity, or at least over time different objects are used for the same entity without the UoW properly updating its `::$entityIdentifiers` structure.

I don't think it makes sense to overwrite an entity in the identity map, since that means a currently `MANAGED` entity is replaced with something else.

If it makes sense at all to _replace_ an entity, that should happen through dedicated management methods to first detach the old entity before persisting, merging or otherwise adding the new one. This way we could make sure the internal structures remain consistent.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1591,6 +1591,30 @@ class UnitOfWork implements PropertyChangedListener
         $className     = $classMetadata->rootEntityName;
 
         if (isset($this->identityMap[$className][$idHash])) {
+            if ($this->identityMap[$className][$idHash] !== $entity) {
+                throw new RuntimeException(sprintf(
+                    <<<'EXCEPTION'
+While adding an entity of class %s with an ID hash of "%s" to the identity map,
+another object of class %s was already present for the same ID. This exception
+is a safeguard against an internal inconsistency - IDs should uniquely map to
+entity object instances. This problem may occur if:
+
+- you use application-provided IDs and reuse ID values;
+- database-provided IDs are reassigned after truncating the database without 
+  clearing the EntityManager;
+- you might have been using EntityManager#getReference() to create a reference 
+  for a nonexistent ID that was subsequently (by the RDBMS) assigned to another 
+  entity. 
+
+Otherwise, it might be an ORM-internal inconsistency, please report it. 
+EXCEPTION
+                    ,
+                    get_class($entity),
+                    $idHash,
+                    get_class($this->identityMap[$className][$idHash])
+                ));
+            }
+
             return false;
         }
 
@@ -2811,23 +2835,18 @@ class UnitOfWork implements PropertyChangedListener
             }
 
             $this->originalEntityData[$oid] = $data;
+
+            if ($entity instanceof NotifyPropertyChanged) {
+                $entity->addPropertyChangedListener($this);
+            }
         } else {
             $entity = $this->newInstance($class);
             $oid    = spl_object_id($entity);
-
-            $this->entityIdentifiers[$oid]  = $id;
-            $this->entityStates[$oid]       = self::STATE_MANAGED;
-            $this->originalEntityData[$oid] = $data;
-
-            $this->identityMap[$class->rootEntityName][$idHash] = $entity;
+            $this->registerManaged($entity, $id, $data);
 
             if (isset($hints[Query::HINT_READ_ONLY])) {
                 $this->readOnlyObjects[$oid] = true;
             }
-        }
-
-        if ($entity instanceof NotifyPropertyChanged) {
-            $entity->addPropertyChangedListener($this);
         }
 
         foreach ($data as $field => $value) {
@@ -2987,20 +3006,7 @@ class UnitOfWork implements PropertyChangedListener
                                 break;
                             }
 
-                            // PERF: Inlined & optimized code from UnitOfWork#registerManaged()
-                            $newValueOid                                                     = spl_object_id($newValue);
-                            $this->entityIdentifiers[$newValueOid]                           = $associatedId;
-                            $this->identityMap[$targetClass->rootEntityName][$relatedIdHash] = $newValue;
-
-                            if (
-                                $newValue instanceof NotifyPropertyChanged &&
-                                ( ! $newValue instanceof Proxy || $newValue->__isInitialized())
-                            ) {
-                                $newValue->addPropertyChangedListener($this);
-                            }
-
-                            $this->entityStates[$newValueOid] = self::STATE_MANAGED;
-                            // make sure that when an proxy is then finally loaded, $this->originalEntityData is set also!
+                            $this->registerManaged($newValue, $associatedId, []);
                             break;
                     }
 
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC1238Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC1238Test.php
@@ -56,7 +56,7 @@ class DDC1238Test extends OrmFunctionalTestCase
 
         $user2 = $this->_em->getReference(DDC1238User::class, $userId);
 
-        $user->__load();
+        //$user->__load();
 
         self::assertIsInt($user->getId(), 'Even if a proxy is detached, it should still have an identifier');
 
--- a/tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
@@ -19,6 +19,7 @@ use Doctrine\Tests\Models\CMS\CmsPhonenumber;
 use Doctrine\Tests\Models\CMS\CmsUser;
 use Doctrine\Tests\OrmFunctionalTestCase;
 use InvalidArgumentException;
+use RuntimeException;
 
 use function get_class;
 
@@ -1289,6 +1290,35 @@ class BasicFunctionalTest extends OrmFunctionalTestCase
 
         $this->_em->persist($user);
 
+        $this->_em->flush();
+    }
+
+    public function testItThrowsWhenReferenceUsesIdAssignedByDatabase(): void
+    {
+        $user           = new CmsUser();
+        $user->name     = 'test';
+        $user->username = 'test';
+        $this->_em->persist($user);
+        $this->_em->flush();
+
+        // Obtain a reference object for the next ID. This is a user error - references
+        // should be fetched only for existing IDs
+        $ref = $this->_em->getReference(CmsUser::class, $user->id + 1);
+
+        $user2           = new CmsUser();
+        $user2->name     = 'test2';
+        $user2->username = 'test2';
+
+        // Now the database will assign an ID to the $user2 entity, but that place
+        // in the identity map is already taken by user error.
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessageMatches('/another object .* was already present for the same ID/');
+
+        // depending on ID generation strategy, the ID may be asssigned already here
+        // and the entity be put in the identity map
+        $this->_em->persist($user2);
+
+        // post insert IDs will be assigned during flush
         $this->_em->flush();
     }
 }
--- a/tests/Doctrine/Tests/ORM/Functional/ProxiesLikeEntitiesTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/ProxiesLikeEntitiesTest.php
@@ -95,7 +95,7 @@ class ProxiesLikeEntitiesTest extends OrmFunctionalTestCase
      */
     public function testProxyAsDqlParameterPersist(): void
     {
-        $proxy     = $this->_em->getProxyFactory()->getProxy(CmsUser::class, ['id' => $this->user->getId()]);
+        $proxy     = $this->_em->getReference(CmsUser::class, ['id' => $this->user->getId()]);
         $proxy->id = $this->user->getId();
         $result    = $this
             ->_em
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7869Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7869Test.php
@@ -55,7 +55,7 @@ class GH7869Test extends OrmTestCase
         $uow->clear();
         $uow->triggerEagerLoads();
 
-        self::assertSame(2, $em->getClassMetadataCalls);
+        self::assertSame(4, $em->getClassMetadataCalls);
     }
 }
 
--- a/tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
+++ b/tests/Doctrine/Tests/ORM/UnitOfWorkTest.php
@@ -41,6 +41,7 @@ use Doctrine\Tests\Models\GeoNames\Country;
 use Doctrine\Tests\OrmTestCase;
 use Doctrine\Tests\PHPUnitCompatibility\MockBuilderCompatibilityTools;
 use PHPUnit\Framework\MockObject\MockObject;
+use RuntimeException;
 use stdClass;
 
 use function assert;
@@ -108,6 +109,8 @@ class UnitOfWorkTest extends OrmTestCase
         $driverConnection = $this->createMock(Driver\Connection::class);
         $driverConnection->method('prepare')
             ->willReturn($driverStatement);
+        $driverConnection->method('lastInsertId')
+            ->willReturnOnConsecutiveCalls(1, 2, 3, 4, 5, 6);
 
         $driver = $this->createMock(Driver::class);
         $driver->method('getDatabasePlatform')
@@ -922,6 +925,25 @@ class UnitOfWorkTest extends OrmTestCase
         // Collection is clean, snapshot has been updated
         self::assertFalse($user->phonenumbers->isDirty());
         self::assertEmpty($user->phonenumbers->getSnapshot());
+    }
+
+    public function testItThrowsWhenApplicationProvidedIdsCollide(): void
+    {
+        // We're using application-provided IDs and assign the same ID twice
+        // Note this is about colliding IDs in the identity map in memory.
+        // Duplicate database-level IDs would be spotted when the EM is flushed.
+
+        $phone1              = new CmsPhonenumber();
+        $phone1->phonenumber = '1234';
+        $this->_unitOfWork->persist($phone1);
+
+        $phone2              = new CmsPhonenumber();
+        $phone2->phonenumber = '1234';
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessageMatches('/another object .* was already present for the same ID/');
+
+        $this->_unitOfWork->persist($phone2);
     }
 }
 


tests/Doctrine/Tests/ORM/Functional/Ticket/DDC1238Test.php
```diff
@@ -56,7 +56,7 @@ public function testIssueProxyClear(): void
 
         $user2 = $this->_em->getReference(DDC1238User::class, $userId);
 
-        $user->__load();
```
mpdude, line 59: @beberlei Do you still remember what you intended to do here? Why should it be possible or allowed to load an entity through a reference when the entity manager has been `clear()`ed in the meantime?

mpdude:
Draft: There are other locations updating the identity map as well, also need to cover those

flack:
just a quick note: I tried running this PR against some of my code, and I can hit the exception by roughly doing the following:

1. create/persist entity `$a` and detach it
2. create/persist entity `$b` which has an association to entity `$a` and detach it
3. call `$em->find()` with entity `$a`'s ID
4. this returns a `Proxy`, I call `__load` on the proxy (and then detach it)
5. I call `$em->merge` with entity `$b` ==> Exception is thrown

If I skip step 4 (specifically, the `__load` call), then everything works.

I know this is not really a great analysis, but maybe it'll give someone an idea what goes wrong here. I find this proxy stuff really hard to debug, but it looks like it's causing this issue

mpdude:
@flack thank you for the observation!

Regarding 2) – what kind of association does B have to A, and does it cascade-persist A as well?

flack:
@mpdude it looks like this in `ClassMetadata`'s `associationMappings` for B: 

```
"linkToA" => array:19 [
      "fieldName" => "linkToA"
      "targetEntity" => "a_entity"
      "joinColumns" => array:1 [
        0 => array:2 [
          "name" => "linkToA"
          "referencedColumnName" => "id"
        ]
      ]
      "type" => 2
      "mappedBy" => null
      "inversedBy" => null
      "isOwningSide" => true
      "sourceEntity" => "b_entity"
      "fetch" => 2
      "cascade" => []
      "isCascadeRemove" => false
      "isCascadePersist" => false
      "isCascadeRefresh" => false
      "isCascadeMerge" => false
      "isCascadeDetach" => false
      "sourceToTargetKeyColumns" => array:1 [
        "linkToA" => "id"
      ]
      "joinColumnFieldNames" => array:1 [
        "linkToA" => "linkToA"
      ]
      "targetToSourceKeyColumns" => array:1 [
        "id" => "linkToA"
      ]
      "orphanRemoval" => false
    ]
```

mpdude:
Maybe todo:

Add some tests illustrating invalid usage patterns that should be spotted by this new check.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1569,6 +1569,14 @@ public function addToIdentityMap($entity)
         $className     = $classMetadata->rootEntityName;
 
         if (isset($this->identityMap[$className][$idHash])) {
+            if ($this->identityMap[$className][$idHash] !== $entity) {
+                throw new RuntimeException(sprintf(
```
flack, line 1595: reposting part of my comment from #10797: Maybe we could even try to be smarter here and allow a non-proxy to replace a proxy? Because it is really easy to put an invalid into the `$identityMap` with `getReference`, and we could try to heal this inconsistency here

mpdude: I don’t think that’s possible: 

Even if an object started out as an uninitialized proxy, it may turn into a fully initialized instance and is the only object instance representing that particular entity identity. It may have gone into different associations, collections and such – we cannot swap all those out for another object.

flack: whether or not the Proxy is initialized can be checked with `__isInitialized()`, so that could be solved, but yeah, it would probably get quite involved and in the end it would only be a workaround for the fact that `getReference` does not give any guarantees that the referenced entity actually exists.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1597,6 +1597,15 @@ public function addToIdentityMap($entity)
...
+                    'While adding an entity of class %s with an ID hash of "%s" to the identity map, another object of class %s was already present for the same ID. This exception is a safeguard against an internal inconsistency - IDs should uniquely map to entity object instances. This problem may occur if you use application-provided IDs and reuse ID values. Or, if database-provided IDs are reassigned after truncating the database without clearing the EntityManager. Or, you might have been using EntityManager#getReference() to create a reference for a nonexistent ID that was subsequently (by the RDBMS) assigned to another entity. Otherwise, it might be an ORM-internal inconsistency, please report it.',
```
greg0ire: Did you consider using linebreaks here? If yes, what made you avoid them?

mpdude: Yes – that I did not find a good example right away that would show me what the desired code style is in that case.

greg0ire: It might be personal but I'd go with wrapping at the 80th column, and use nowdoc so as not to have to worry about quotes.

greg0ire: Example: https://github.com/doctrine/orm/blob/55d477dc507a52675c5a61267a00d389c71182f4/lib/Doctrine/ORM/Exception/NotSupported.php#L19-L23

mpdude: I don't like nowdocs because they break out of indentation, suddenly all text is at the left 😔

mpdude: What about so?

```diff
diff --git a/lib/Doctrine/ORM/UnitOfWork.php b/lib/Doctrine/ORM/UnitOfWork.php
index 4b15fa47a0..f31eeaffec 100644
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1599,7 +1599,18 @@ class UnitOfWork implements PropertyChangedListener
         if (isset($this->identityMap[$className][$idHash])) {
             if ($this->identityMap[$className][$idHash] !== $entity) {
                 throw new RuntimeException(sprintf(
-                    'While adding an entity of class %s with an ID hash of "%s" to the identity map, another object of class %s was already present for the same ID. This exception is a safeguard against an internal inconsistency - IDs should uniquely map to entity object instances. This problem may occur if you use application-provided IDs and reuse ID values. Or, if database-provided IDs are reassigned after truncating the database without clearing the EntityManager. Or, you might have been using EntityManager#getReference() to create a reference for a nonexistent ID that was subsequently (by the RDBMS) assigned to another entity. Otherwise, it might be an ORM-internal inconsistency, please report it.',
+                    'While adding an entity of class %s with an ID hash of "%s" ' .
+                    'to the identity map, another object of class %s was already ' .
+                    'present for the same ID. This exception is a safeguard against ' .
+                    'an internal inconsistency - IDs should uniquely map to entity ' .
+                    'object instances. This problem may occur if you use ' .
+                    'application-provided IDs and reuse ID values. Or, if database-provided ' .
+                    'IDs are reassigned after truncating the database without ' .
+                    'clearing the EntityManager. Or, you might have been using ' .
+                    'EntityManager#getReference() to create a reference for a ' .
+                    'nonexistent ID that was subsequently (by the RDBMS) assigned ' .
+                    'to another entity. Otherwise, it might be an ORM-internal ' .
+                    'inconsistency, please report it.',
                     get_class($entity),
                     $idHash,
                     get_class($this->identityMap[$className][$idHash])
```

greg0ire: > I don't like nowdocs because they break out of indentation, suddenly all text is at the left

Yeah, but that's not the case with recent versions of PHP, right? So I think it would be better to use nowdoc early, because on 3.0.x you will just have to indent everything instead of resorting to concatenation.

> What about so?

I see 2 issues with this:

- it's harder to maintain than nowdoc
- the output text has no linebreaks.

People reading the error message would want linebreaks as well in this case, I think. In recent versions of PHP, with nowdoc, you can have such strings with linebreaks _without introducing extra indentation in output text_.

mpdude: TIL!

I've changed it to a nowdoc.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1597,6 +1597,26 @@ public function addToIdentityMap($entity)
...
+another object of class %s was already present for the same ID. This exception
+is a safeguard against an internal inconsistency - IDs should uniquely map to
+entity object instances. This problem may occur if you use application-provided
+IDs and reuse ID values. Or, if database-provided IDs are reassigned after
+truncating the database without clearing the EntityManager. Or, you might have
+been using EntityManager#getReference() to create a reference for a nonexistent
+ID that was subsequently (by the RDBMS) assigned to another entity. Otherwise,
+it might be an ORM-internal inconsistency, please report it. 
+EXCEPTION
+                    ,
```
flack, line 1611: if linebreaks are now kept, could also do

```
                    <<<'EXCEPTION'
While adding an entity of class %s with an ID hash of "%s" to the identity map,
another object of class %s was already present for the same ID. This exception
is a safeguard against an internal inconsistency - IDs should uniquely map to
entity object instances. This problem may occur if 

- you use application-provided IDs and reuse ID values.
- database-provided IDs are reassigned after truncating the database without 
  clearing the EntityManager.
- you might have been using EntityManager#getReference() to create a reference 
  for a nonexistent ID that was subsequently (by the RDBMS) assigned to another 
  entity. 

Otherwise, it might be an ORM-internal inconsistency, please report it. 
EXCEPTION
```

mpdude: Nice! I'll use that.

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1597,6 +1597,30 @@ public function addToIdentityMap($entity)
         $className     = $classMetadata->rootEntityName;
 
         if (isset($this->identityMap[$className][$idHash])) {
+            if ($this->identityMap[$className][$idHash] !== $entity) {
+                throw new RuntimeException(sprintf(
+                    <<<'EXCEPTION'
+While adding an entity of class %s with an ID hash of "%s" to the identity map,
+another object of class %s was already present for the same ID. This exception
+is a safeguard against an internal inconsistency - IDs should uniquely map to
+entity object instances. This problem may occur if
```
greg0ire:
```suggestion
entity object instances. This problem may occur if:
```

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1597,6 +1597,30 @@ public function addToIdentityMap($entity)
...
         if (isset($this->identityMap[$className][$idHash])) {
+            if ($this->identityMap[$className][$idHash] !== $entity) {
+                throw new RuntimeException(sprintf(
+                    <<<'EXCEPTION'
+While adding an entity of class %s with an ID hash of "%s" to the identity map,
+another object of class %s was already present for the same ID. This exception
+is a safeguard against an internal inconsistency - IDs should uniquely map to
+entity object instances. This problem may occur if 
+
+- you use application-provided IDs and reuse ID values.
```
greg0ire:
```suggestion
- you use application-provided IDs and reuse ID values;
```

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -1597,6 +1597,30 @@ public function addToIdentityMap($entity)
...
+- database-provided IDs are reassigned after truncating the database without 
+  clearing the EntityManager.
```
greg0ire:
```suggestion
  clearing the EntityManager;
```

lib/Doctrine/ORM/UnitOfWork.php
greg0ire: It's beautiful!

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2817,25 +2841,20 @@ public function createEntity($className, array $data, &$hints = [])
             }
 
             $this->originalEntityData[$oid] = $data;
+
+            if ($entity instanceof NotifyPropertyChanged) {
+                $entity->addPropertyChangedListener($this);
+            }
```
greg0ire, line 2841: How does this change and the ones below relate to the safeguard part?

mpdude: Well spotted!

My goal was to use `registerManaged` for the code that follows in the `else` branch a few lines below, to have all updates to the Identity Map go through that method.

`registerManaged` already includes adding those `PropertyChangedListeners`. So, I do no longer add those listeners below, after the `if`/`else` branching. Instead, that code is moved up and now happens in the `if` part only.

greg0ire: So… does it relate to the safeguard part or not? I'm confused now.

mpdude: Yes, it belongs here. It makes sure we run the safeguard check also for entities created in this method.

greg0ire: … because `registerManaged()` calls `addToIdentityMap()`, I get it now, thanks!

greg0ire (approved):

greg0ire:
To be squash-merged (or squashed)

mpdude:
Squashed + rebased.

tests/Doctrine/Tests/ORM/Functional/Ticket/DDC1238Test.php
beberlei: I believe this tests that the proxy generation has a shortcut for `getId` methods that returns the value passed to getReference(.., $id);

I am not sure this needs the calls to __load anymore, but really don't know anymore.

mpdude: Do you think it should be possible to initialize a proxy _after_ the EM it was obtained from has been cleared?

I don’t think so, the EM has forgotten about that proxy instance.

flack: maybe I'm missing some part of the conversation, but isn't this the same as initializing the proxy before anything was put into the EM? I mean looking at the three lines of code shown in Gh, it wasn't cleared between `getReference` and `__load`, so it should be fine, no?

SenseException (approved):

greg0ire:
Thanks @mpdude !

akadlec:
Hi @mpdude what about to change `Runtime` exception to something other, more specific? I'm using application defined IDs and I have a mechanism that will catch exception if user will try to create new entity with sane ID (UniqueConstraintViolationException). But now it is not possible, it is just Runtime exception with a lot of text

mpdude:
So you got a `UniqueConstraintViolationException` before this change?

akadlec:
@mpdude yep, before 2.16

mpdude:
Can you provide a test that gives the `UniqueConstraintViationException` before 2.16 and the `RuntimeException` now?

gbirke:
I also have this problem with the `UniqeConstraintViolation` becoming a `RuntimeException` after the update. It's [a test](https://github.com/wmde/fundraising-payments/blob/main/tests/Integration/DataAccess/DoctrinePaymentRepositoryTest.php#L68-L78) for [a method](https://github.com/wmde/fundraising-payments/blob/main/src/DataAccess/DoctrinePaymentRepository.php#L16-L23) that converts the exception into a domain exception

mpdude:
Haven’t run the code, but my guess is:

The EM now notices the ID collision already when you pass the second entity to `persist()`, since the ID is already in the identity map. Previously, it would have replaced the identity map entry (thereby corrupting other internal data structures) and only figure out during `flush()` that the ID has already been taken.

@derrabus Maybe it would make sense to use something more specific than `RuntimeException` for this safeguard. Do you have any thoughts of what the exception hierarchy should look like? I don’t think it makes sense to have a common base class with `UniqueConstraintViolationException` though..

derrabus:
This is indeed a behavior change that might be unexpected for existing apps that trusted that the database would reject a colliding ID. We might need to add a feature flag for that. 😕

mpdude:
But that means with the flag disabled, we risk the corruption again, right? Because we accept to have two different object instances representing the same thing.

flack:
The database still will reject a colliding ID, it's just that now in most cases, you don't even get that far, because the ORM layer rejects it first. So now your code would need to catch both exceptions. 

Ideally, this duplication of functionality should be avoided, and as the DB backend is kind of the final source of truth the check should ideally be there, but then, the identyMap logic would need to be different to still be able to avoid corruption

akadlec:
That was my point, to change `RuntimeException` to something more specific to be able to catch it and handle it.

derrabus:
> But that means with the flag disabled, we risk the corruption again, right? Because we accept to have two different object instances representing the same thing.

Yes, but apps could live with this kind of corruption previously. If the flush resulted in a `UniqueConstraintViolationException`, the developer knows that the entity graph needs to be trashed, which is acceptable in a short-lived web process.

I believe, we should do both:

* feature-flag this detection and
* introduce a more specific exception.

mpdude:
Feature flag – that needs to go into a 2.17 release?

derrabus:
No, let's treat this as a bugfix.

flack:
> Yes, but apps could live with this kind of corruption previously. If the flush resulted in a UniqueConstraintViolationException, the developer knows that the entity graph needs to be trashed, which is acceptable in a short-lived web process.

Out of curiosity: Is there a design document somewhere that outlines the rationale for the entity graph/identity map? Because if the face of concurrent requests it seems to me that the graph should always be treated as a slightly out-of-date snapshot of the actual DB.

ps-mattstuart:
I don't mean to drag up an old issue but we battled this for a while.

While processing thousands of the same entity types we'd have issues with regenerating the same ID. I'm not remembering perfectly and I'll go back to try and resurrect the exact issue flow that would cause this. What was happening was the UOW was either not finding the entity because it had a persistent record of the OID having been seen and the newly asked for entity generated the same OID. I believe we're releasing the entity objects which based on the PHP definition of the function can lead to a re-use of the ID and we were running into this.

From php.net:

> This function returns a unique identifier for the object. The object id is unique for the lifetime of the object. Once the object is destroyed, its id may be reused for other objects. This behavior is similar to [spl_object_hash()](https://www.php.net/manual/en/function.spl-object-hash.php).

Does this address that case? We're currently testing the latest version to see if it resolves all of the bugs that we have internally resolved but looking at the commit for this ticket it doesn't appear to cover the issue we were seeing. Given that the OID could conflict naturally (there's no guarantee of unique once objects are getting destroyed) throwing an exception I'm unsure was the right idea here.

PR #10790, opened 2023-06-23
Work around slevomat/coding-standard issues

greg0ire:
I tweaked the code so that it would not fall victim to
slevomat/coding-standard#1585 or
slevomat/coding-standard#1586, thus fixing the
phpcs job without losing information or breaking other jobs.

### DIFF
--- a/lib/Doctrine/ORM/Configuration.php
+++ b/lib/Doctrine/ORM/Configuration.php
@@ -208,9 +208,9 @@ class Configuration extends \Doctrine\DBAL\Configuration
     }
 
     /**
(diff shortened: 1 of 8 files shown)


greg0ire:
@mpdude please review, I'd like to merge this and unblock you.

mpdude:
Hm, don't really know how to judge this.

I'd say when Psalm stays happy, the tests still pass and no other tool complains about CS, it's :+1:

greg0ire:
If you can't spot any obvious mistake, then go 👍

PR #8994, opened 2021-09-09
DDC-2332: Ensure managed entities are always tracked by UOW

bigfoot90:
This is a rebase of MR https://github.com/doctrine/orm/pull/7407 on `2.9.x` branch.
Fixes https://github.com/doctrine/orm/issues/3037

PR #10791, opened 2023-06-23
Avoid creating unmanaged proxy instances for referred-to entities during `merge()`

mpdude:
This PR tries to improve the situation/problem explained in #3037:

Under certain conditions – there may be multiple and not all are known/well-understood – we may get inconsistencies between the `\Doctrine\ORM\UnitOfWork::$entityIdentifiers` and `\Doctrine\ORM\UnitOfWork::$identityMap` arrays.

Since the `::$identityMap` is a plain array holding object references, objects contained in it cannot be garbage-collected.
`::$entityIdentifiers`, however, is indexed by `spl_object_id` values. When those objects are destructed and/or garbage-collected, the OID may be reused and reassigned to other objects later on.

When the OID re-assignment happens to be for another entity, the UoW may assume incorrect entity states and, for example, miss INSERT or UPDATE operations.

One cause for such inconsistencies is _replacing_ identity map entries with other object instances: This makes it possible that the old object becomes GC'd, while its OID is not cleaned up. Since that is not a use case we need to support (IMHO), #10785 is about adding a safeguard against it.

In this test shown here, the `merge()` operation is currently too eager in creating a proxy object for another referred-to entity. This proxy represents an entity already present in the identity map at that time, potentially leading to this problem later on.

Cross-reference https://github.com/doctrine/orm/issues/3037#issuecomment-369211489.

Closes #7407, closes #8994.

Not enough information in #3037 to tell whether it would be fixed.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -3669,14 +3669,18 @@ class UnitOfWork implements PropertyChangedListener
                                 $targetClass = $this->em->getClassMetadata($assoc2['targetEntity']);
                                 $relatedId   = $targetClass->getIdentifierValues($other);
 
-                                if ($targetClass->subClasses) {
-                                    $other = $this->em->find($targetClass->name, $relatedId);
-                                } else {
-                                    $other = $this->em->getProxyFactory()->getProxy(
-                                        $assoc2['targetEntity'],
-                                        $relatedId
-                                    );
-                                    $this->registerManaged($other, $relatedId, []);
+                                $other = $this->tryGetById($relatedId, $targetClass->name);
+
+                                if (! $other) {
+                                    if ($targetClass->subClasses) {
+                                        $other = $this->em->find($targetClass->name, $relatedId);
+                                    } else {
+                                        $other = $this->em->getProxyFactory()->getProxy(
+                                            $assoc2['targetEntity'],
+                                            $relatedId
+                                        );
+                                        $this->registerManaged($other, $relatedId, []);
+                                    }
                                 }
                             }
 
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7407Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7407Test.php
@@ -76,6 +76,7 @@ class GH7407Test extends OrmFunctionalTestCase
         ], $identityMap, 'The identity map contains exactly two objects, the article and the user.');
     }
 
+    /** @return mixed */
     private function grabProperty(string $name, UnitOfWork $uow)
     {
         $reflection = new ReflectionClass($uow);
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7407Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH7407Test.php
@@ -1,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\UnitOfWork;
+use Doctrine\Tests\Models\CMS\CmsArticle;
+use Doctrine\Tests\Models\CMS\CmsUser;
+use Doctrine\Tests\OrmFunctionalTestCase;
+use ReflectionClass;
+
+use function spl_object_id;
+
+class GH7407Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        $this->useModelSet('cms');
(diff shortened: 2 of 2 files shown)


mpdude:
@Ocramius Your thoughts would be very welcome

greg0ire:
Blocked by #10790

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7407Test.php
```diff
@@ -0,0 +1,95 @@
...
+        $mergedUser    = $this->_em->merge($user);
+        $mergedUserOid = spl_object_id($mergedUser);
+
+        // 3. Merge the article object back in,
+        // the returned entity object is the article instance as it is managed by the EM/UoW
+        $mergedArticle    = $this->_em->merge($article);
+        $mergedArticleOid = spl_object_id($mergedArticle);
+
+        // The $mergedArticle's #user property should hold the $mergedUser we obtained previously,
+        // since that's the only legetimate object instance representing the user from the UoW's
```
greg0ire:
```suggestion
        // since that's the only legitimate object instance representing the user from the UoW's
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7407Test.php
```diff
@@ -0,0 +1,95 @@
...
+        // point of view.
+        self::assertSame($mergedUser, $mergedArticle->user);
```
greg0ire: I love this comment, and I think you should move it to the 3rd argument of that assertion, so that people can see it when it fails:

```suggestion
        self::assertSame($mergedUser, $mergedArticle->user, <<<'MESSAGE'
The $mergedArticle's #user property should hold the $mergedUser we obtained previously,
since that's the only legetimate object instance representing the user from the UoW's
point of view.
MESSAGE
);
```

There are probably other comments that would make good `$message` arguments below.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7407Test.php
mpdude: fixed

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7407Test.php
mpdude: I hate those heredocs for the visual breakage they cause, but I've changed comments to messages.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH7407Test.php
```diff
@@ -0,0 +1,87 @@
...
+        self::assertArrayHasKey($mergedArticleOid, $entityIdentifiers);
+        self::assertArrayHasKey($mergedUserOid, $entityIdentifiers);
+
+        self::assertSame([
+            CmsUser::class => [$user->id => $mergedUser],
+            CmsArticle::class => [$article->id => $mergedArticle],
+        ], $identityMap, 'The identity map contains exactly two objects, the article and the user.');
+    }
+
+    private function grabProperty(string $name, UnitOfWork $uow)
```
greg0ire, line 80:
```suggestion
    /** @return mixed */
    private function grabProperty(string $name, UnitOfWork $uow)
```

greg0ire (approved):

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3669,14 +3669,18 @@ private function mergeEntityStateIntoManagedCopy($entity, $managedCopy): void
...
-                                } else {
-                                    $other = $this->em->getProxyFactory()->getProxy(
-                                        $assoc2['targetEntity'],
-                                        $relatedId
-                                    );
-                                    $this->registerManaged($other, $relatedId, []);
+                                $other = $this->tryGetById($relatedId, $targetClass->name);
+
+                                if (! $other) {
+                                    if ($targetClass->subClasses) {
```
greg0ire, line 3675: Codecov seems to say the inside of this if block is never covered now :thinking:

mpdude: Given that this code has seemingly been used for a decade without being covered by tests, and that it belongs to the `merge` feature removed in 3.0, is it worth adding tests?

mpdude: Oh. Just realized that this change is probably going to conflict when merging up into 3.0... the test should not be added, and the fixed method has been removed.

SenseException (approved):

greg0ire:
Thanks @mpdude !

PR #10795, opened 2023-06-23
Merge 2.16.x up into 3.0.x

greg0ire:


greg0ire:
@derrabus what should be done w.r.t https://github.com/doctrine/dbal/pull/5838 on the ORM?

greg0ire:
Ah I think I know: we should replace the `int` in AbstractSQLExecutor with `ArrayParameterType::*` (it's an integer with DBAL 3, and it's an enum with DBAL 4. I hope SA undestands `Enum::*` the same way it understands `Enum` :crossed_fingers:

lib/Doctrine/ORM/Query/Exec/AbstractSqlExecutor.php
```diff
@@ -49,10 +53,9 @@ public function removeQueryCacheProfile(): void
     /**
      * Executes all sql statements.
      *
-     * @param Connection $conn The database connection that is used to execute the queries.
-     * @psalm-param list<mixed>|array<string, mixed> $params The parameters.
-     * @psalm-param array<int, int|string|Type|null>|
-     *              array<string, int|string|Type|null> $types The parameter types.
+     * @param Connection                       $conn   The database connection that is used to execute the queries.
+     * @param list<mixed>|array<string, mixed> $params The parameters.
+     * @psalm-param WrapperParameterTypeArray  $types  The parameter types.
```
greg0ire, line 58: Ported as https://github.com/doctrine/orm/pull/10796

greg0ire:
@mpdude please review

mpdude:
I don’t know what to specifically watch out for which these merge-ups 🙁

mpdude:
Did you have to resolve any conflicts, or was it a smooth merge?

greg0ire:
It wasn't smooth at all, far from it :sweat: 

I wish Github would indicate where conflict happened, but I think right now the only way is to compare the diff of this PR with https://github.com/doctrine/orm/compare/3.0.x...2.16.x

derrabus (approved):

mpdude:
🫣

what is the source of the conflicts? Refactorings on 3.0?

greg0ire:
Yes, mostly. Sometimes it is a dumb as cs rules that are not applicable on 3.x, for instance trailing commas after the last argument passed to a method. Some other times, it's more complicated :P

PR #10796, opened 2023-06-24
Narrow integer to param class constants

greg0ire:
You cannot just pass any integer as a type, it has to map to something meaningful.

The psalm types introduced here are copied from `doctrine/dbal` 4, with a small difference: since we are not dealing with actual enums yet, but with class constants, we add `::*`, which is compatible with both cases (class or enum).


greg0ire:
Cannot be done if we still support DBAL 2 :thinking:

PR #10798, opened 2023-06-25
Resort on Query::HINT_FORCE_PARTIAL_LOAD less

greg0ire:
A lot of our tests mention it, but I do not think it is important to the test. Maybe it was a way to have more efficient tests? Most times, removing the hint does not affect the test outcome.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Hydration/ObjectHydratorTest.php
+++ b/tests/Doctrine/Tests/ORM/Hydration/ObjectHydratorTest.php
@@ -8,7 +8,6 @@ use Doctrine\ORM\Internal\Hydration\ObjectHydrator;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\PersistentCollection;
 use Doctrine\ORM\Proxy\ProxyFactory;
-use Doctrine\ORM\Query;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\Tests\Mocks\ArrayResultFactory;
 use Doctrine\Tests\Models\CMS\CmsAddress;
@@ -65,7 +64,7 @@ class ObjectHydratorTest extends HydrationTestCase
     }
 
     /**
-     * SELECT PARTIAL u.{id,name}
+     * SELECT u
      *   FROM Doctrine\Tests\Models\CMS\CmsUser u
      */
     public function testSimpleEntityQuery(): void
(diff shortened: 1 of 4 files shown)


SenseException:
The tests didn't seem to check the other sides of the partial objects. Were those all created for the partial objects?

greg0ire:
Sorry, what do you mean by "other sides"?

SenseException:
> Most times, removing the hint does not affect the test outcome.

They check the properties that are part of the partial, but not the ones left out of the query, that aren't hydrated. Or at least that's what I assume considering the DQL in the comments. So there couldn't be much of a different outcome after the partial-parts were removed.

greg0ire:
> The tests didn't seem to check the other sides of the partial objects. Were those all created for the partial objects?

I do not think the tests were created for testing partial objects, but for testing hydration, because "partial" is not mentioned in the title of these tests. Partial objects might have been used as a way to speed up the tests.

beberlei:
The tests seem to be for partial loading wirh the hydrator, dont they throw an exception in 3.0 for missing properties?

greg0ire:
I will try removing support for partial objects on 3.0 and see what happens.

greg0ire:
:thinking: whey I try _accessing_ `CmsUser::status` in `testSimpleEntity`, I get an error about the fact that it is not initialized. And I get it only if I try performing that access. Am I supposed to get an exception before that?

beberlei:
@greg0ire i was under the impression that the hydrator or UoW::createEntity would just flat out refuse to hydrate the object and bail.

greg0ire:
It does not seem to care :slightly_smiling_face:

tests/Doctrine/Tests/ORM/Hydration/ObjectHydratorTest.php
```diff
@@ -200,7 +199,7 @@ public function testSimpleMultipleRootEntityQuery(): void
     }
 
     /**
-     * SELECT PARTIAL u.{id, name} AS user, PARTIAL a.{id, topic}
+     * SELECT u AS user, a
      *   FROM Doctrine\Tests\Models\CMS\CmsUser u, Doctrine\Tests\Models\CMS\CmsArticle a
      */
     public function testSimpleMultipleRootEntityQueryWithAliasedUserEntity(): void
```
greg0ire, line 205: I don't think these tests are _about_ partial objects. For instance, according to the title, this one seems to be about testing that the objecthydrator works when the rsm as several root entities, and one of them is aliased. I think they still make sense, even after removing partial objects.

SenseException: Seems like the partial objects really were just unrelated to the tests.

SenseException:
> I get an error about the fact that it is not initialized. And I get it only if I try performing that access.

Is this happening when you access a property that isn't part of the partial object hydration?

greg0ire:
Yes, and that's because I deliberately added that access to see what would happen.

SenseException:
Should this target `2.16.x` now before merging? Seems like this can be approved from what we are knowing.

greg0ire:
I don't think so: it only touches the test suite. Having branches as similar as possible is best when there it does not come with a stability impact, IMO.

SenseException (approved):
Okay.

PR #10799, opened 2023-06-25
Improve and add test to set to-one and to-many associations with reference objects

### LINKED ISSUE #9534: Using Proxies obtained through EntityManager::getReference() in associations?
I have two entity classes, `Cart` and `Item`, where `Cart` has a many-to-many association with `Item`.

Inside a functional test, I'd like to shortcut things a bit, and thus I obtain a reference (proxy) for `Item` through `EntityManager::getReference(Item::class, 42)`.

I then create a new `Cart` instance, and the constructor will initialize the `Cart::$containedItems` collection as a new `ArrayCollection`. To this collection, I add the `Cart` reference.

When I try to persist the `Cart`, I get an exception stating that `Item` was found as a new Entity through the relationship `Cart::$containedItems` that was not configured to cascade persist operations.

I know I am on thin ice here, but the documentation on [Reference Proxies](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/reference/advanced-configuration.html#reference-proxies) states:

> This is useful, for example, as a performance enhancement, when you want to establish an association to an entity for which you have the identifier. You could simply do this:
> `$item = $em->getReference('MyProject\Model\Item', $itemId);`
> `$cart->addItem($item);`

#### Details of what's happening

When the data is flushed, `UnitOfWork::computeAssociationChanges()` will look at the `containedItems` association. It iterates over the contained entities and tries to determine their state:

https://github.com/doctrine/orm/blob/193c3abf0e5f7a7e2699e7c7ac37c1e713d79ef8/lib/Doctrine/ORM/UnitOfWork.php#L914-L919

`UnitOfWork` does not have a state for the reference/proxy object in its `::$entityStates` array, and thus returns `STATE_NEW`. This leads to taking note of the allegedly new entity a few lines later:

https://github.com/doctrine/orm/blob/193c3abf0e5f7a7e2699e7c7ac37c1e713d79ef8/lib/Doctrine/ORM/UnitOfWork.php#L930-L939

Nothing around that area tries to recognize `$entry` as a proxy. When I try to skip `instanceof Proxy` values in that place, we will later get into `UnitOfWork::getEntityIdentifier()` with the proxy as `$entity`.

This method, again, is not prepared to deal with a proxy:

https://github.com/doctrine/orm/blob/193c3abf0e5f7a7e2699e7c7ac37c1e713d79ef8/lib/Doctrine/ORM/UnitOfWork.php#L3079-L3083

The proxy is not in the `::$entityIdentifiers` array. It carries the values we'd expect to find in that array in its fields, but my guess would be we'd need a suitable `ProxyDefinition` to obtain them.

#### Applicable test cases

I have found `\Doctrine\Tests\ORM\Functional\BasicFunctionalTest::testSetSetAssociationWithGetReference` which suggests to test the use case I described initially. This test, however, notes:

https://github.com/doctrine/orm/blob/193c3abf0e5f7a7e2699e7c7ac37c1e713d79ef8/tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php#L547-L551

That makes me believe the test never really worked with an uninitialized proxy. When the proxy is initialized by accessing properties, we probably have a fully loaded, state-tracked entity with known identifier values in the `UnitOfWork`. 

<hr/>

So, first question probably: Is this accepted as a bug and something that would get support being fixed?

mpdude:
This PR tweaks a test case that checks initializing a to-one association with a proxy object, which is obtained through `EntityManager::getReference()`. With the change suggested here, we can avoid initializing the proxy, thus showing that things work even in that case.

Also, a second test case is added to test the same for a to-many (collection-valued) association.

Closes #9534.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
@@ -15,6 +15,7 @@ use Doctrine\Tests\IterableTester;
 use Doctrine\Tests\Models\CMS\CmsAddress;
 use Doctrine\Tests\Models\CMS\CmsArticle;
 use Doctrine\Tests\Models\CMS\CmsComment;
+use Doctrine\Tests\Models\CMS\CmsGroup;
 use Doctrine\Tests\Models\CMS\CmsPhonenumber;
 use Doctrine\Tests\Models\CMS\CmsUser;
 use Doctrine\Tests\OrmFunctionalTestCase;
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/BasicFunctionalTest.php
```diff
@@ -515,42 +516,78 @@ public function testInitializeCollectionWithNewObjectsRetainsNewObjects(): void
...
+        $this->_em->flush();
+        $this->_em->clear();
+
+        // Assume we only got the identifier of the user and now want to attach
+        // the article to the user without actually loading it, using getReference().
+        $groupRef = $this->_em->getReference(CmsGroup::class, $group->id);
+        self::assertInstanceOf(Proxy::class, $groupRef);
+        self::assertFalse($groupRef->__isInitialized());
 
-        $user->setAddress($addressRef); // Ugh! Initializes address 'cause of $address->setUser($user)!
```
mpdude, line 542: This is what I want to avoid.

greg0ire (approved):

SenseException (approved):

PR #843, opened 2013-11-08
[DDC-2785] Add failing test

flack:
Add a failing test for

http://www.doctrine-project.org/jira/browse/DDC-2785

PR #10800, opened 2023-06-25
Explain `EntityManager::getReference()` peculiarities

### LINKED ISSUE #10797: Better explain the `EntityManager::getReference()` method
As one takeaway from https://github.com/doctrine/orm/issues/3037#issuecomment-1605657003 and #843, we should look into better explaining the `EntityManager::getReference()` method, it’s semantics, caveats and potential responsibilities placed on the user.

mpdude:
As one takeaway from https://github.com/doctrine/orm/issues/3037#issuecomment-1605657003 and #843, we should look into better explaining the `EntityManager::getReference()` method, it’s semantics, caveats and potential responsibilities placed on the user.

This PR tries to do that, so it fixes #10797.

### DIFF
--- a/docs/en/reference/advanced-configuration.rst
+++ b/docs/en/reference/advanced-configuration.rst
@@ -311,10 +311,12 @@ Reference Proxies
 
 The method ``EntityManager#getReference($entityName, $identifier)``
 lets you obtain a reference to an entity for which the identifier
-is known, without loading that entity from the database. This is
-useful, for example, as a performance enhancement, when you want to
-establish an association to an entity for which you have the
-identifier. You could simply do this:
+is known, without necessarily loading that entity from the database.
+This is useful, for example, as a performance enhancement, when you
+want to establish an association to an entity for which you have the
+identifier.
+
+Consider the following example:
 
 .. code-block:: php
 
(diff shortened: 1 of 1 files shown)


mpdude:
As a side note, I noticed the ``getReference()`` method and proxies are mentioned in the "advanced configuration" section only.

That seems strange, I don't see what that topic has to do with it.

WDYT about having a dedicated page for proxies or moving that topic elsewhere?

greg0ire:
> WDYT about having a dedicated page for proxies or moving that topic elsewhere?

Makes sense to me 👍

docs/en/reference/advanced-configuration.rst
```diff
@@ -324,15 +326,33 @@ identifier. You could simply do this:
...
-database. If you access any state that isn't yet available in the
-Item instance, the proxying mechanism would fully initialize the
-object's state transparently from the database. Here
-$item is actually an instance of the proxy class that was generated
-for the Item class but your code does not need to care. In fact it
-**should not care**. Proxy objects should be transparent to your
+Whether the object being returned from ``EntityManager#getReference()``
+is a proxy or a direct instance of the entity class may depend on different
+factors, including whether the entity has already been loaded into memory
+or entity inheritance being used. But, your code does not need to care
```
greg0ire:
```suggestion
or entity inheritance being used. But your code does not need to care
```

greg0ire (approved):

SenseException (approved):

PR #10804, opened 2023-06-27
Follow recommendation about multiline type

greg0ire:
Apparently, there is consensus about multiline types between:

- PHPStan
- Psalm
- Slevomat Coding Standard

See https://github.com/slevomat/coding-standard/issues/1586#issuecomment-1610195706

Using parenthesis is supposed to be less ambiguous, although I cannot explain how it is ambiguous before… maybe if we had the pipes at the beginning of lines instead of the end it would be ambiguous because you could not clearly tell if the return is multiline or not?

Anyway, the change has a positive impact on the Psalm baseline, showing that psalm-return annotation was not really understood previously. Thanks @ondrejmirtes for the heads up!

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -2566,8 +2566,7 @@ class Parser
      *      EmptyCollectionComparisonExpression | CollectionMemberExpression |
      *      InstanceOfExpression
      *
-     * @return AST\Node
-     * @psalm-return AST\BetweenExpression|
+     * @return (AST\BetweenExpression|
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
> although I cannot explain how it is ambiguous before

It's a parser. The text is always open to multiple interpretations.

When a line ends with `|` the parser doesn't know if it's part of the type, or of it's just a description, and whether we're supposed to parse the next line.

If you start a type with `(`, you can parse until you find a `)`.

greg0ire:
Makes sense! I'll amend my commit message 👍

derrabus (approved):

Issue #6776, opened 2017-10-16
oneToMany orphanRemoval with unique - constraint violation


akomm:
I currently do not have time for the test, also because this blocks me: #6738. I try to add one asap.

Issue is the following:

When you remove and insert the Many-Items from the One-side and the Many-Items have an unique constraint in a single transaction/flush, you get a constraint violation. 

Reason: 
INSERTS for the Many side are executed prior to DELETE. This is not a problem with auto generated identifiers, but a problem when the Many side has an unique constraint and newly added items have same value (or value-pairs in composite) as the old one you delete.

From a quick check, It looks like it is the same as #2310, just with OneToMany.

lcobucci:
@akomm thanks for reporting the issue, we do need a test though (we're also short on time). Send us something when you're available and we discuss on the PR.

akomm:
I resolved the #6738 issue but could not make a reproduction "quick" and worked around it removing the unique constraint temporary relying on code as long as the app is not in prod. Want to come back to it when rest is done so I can enable the constraint again.

julienb-allopneus:
Hi,
We're facing the same problem but we won't just remove the constraint ;-)
Anyone working on this topic?

cc @nio-hevea @Myloth

lcobucci:
@julienb-allopneus could you please send us a failing test case that reproduces that behaviour? It would help us a lot to identify and fix the issue you're describing.

You can find examples on https://github.com/doctrine/doctrine2/tree/388afb46d0cb3ed0c51332e8df0de9e942c2690b/tests/Doctrine/Tests/ORM/Functional/Ticket

julienb-allopneus:
Hi @lcobucci ,
I'll work on this today, and keep you informed as soon as I can.

julienb-allopneus:
Hi guys @lcobucci & @akomm ,

I have written a test in the following PR https://github.com/doctrine/doctrine2/pull/6838 , please have a look and tell me what's next :-)
FYI, I have no experience in doctrine contribution and doctrine internals .

Thanks,

Julien.

cc @nio-hevea @Myloth

julienb-allopneus:
Hi all,

Any news/advice on the test I wrote?
@lcobucci , is it ok for you?

Thanks,

Julien.

julienb-allopneus:
Hi @lcobucci and @akomm ,

Any news on this issue?
I guess that we may remove the "Missing tests" label and requalify the issue.

Thanks,

Julien.

[...]

Issue #5742, opened 2016-03-26
Flush could delete entities before inserting new ones


MacDada:
I have a unique (not PK) column. I need to remove old entities first and then add new ones (replacements).

Yeah, I've seen this: https://github.com/doctrine/doctrine2/issues/5368

But I've also seen a "could be done" [solution](https://github.com/doctrine/doctrine2/issues/5109#issuecomment-162370533):

> A function "setShouldPerformDeletesFirst" as described in http://docs.oracle.com/cd/B31017_01/web.1013/b28218/uowadv.htm#i1137484 would be very useful in those cases. 

What I propose, is that this gets implemented. From the dev point of view, it would be cool, I one could do that like this:

`* @ORM\OneToMany(targetEntity="Whatever", mappedBy="Me", cascade={"persist"}, orphanRemoval=true, removeBeforeAdd=true)`

ureimers:
I'm not quite sure if `removeBeforeAdd` should only be available on relations. You could easily think of a case where you replace entity A with B with both having the same uniquely constrained fields.

IMM0rtalis:
+1 here

johandouma:
+1. 
This issue seems related to #6776 - in case other people are searching.

SmasherHell:
+1 That is quite a challenge to respect OCP when doctrine removes "merge". There is no way to update an object from a clone within a transaction. How to have a separation of concerns between logic and persistance when we must flush a state in the middle of logic

hhamon:
I faced this issue as well on my client project. We had to remove the SQL unique constraint on the table to ensure the Doctrine SQL transaction can be complete. But ideally we would need both the SQL unique constraint (to always ensure consistency in the database) and in the SQL transaction when Doctrine flush statement is performed.

mpdude:
@SmasherHell I am not sure if it would support your use case – probably it may be difficult to deal with two entities at the same time that share the same primary key value, and to have one replace the other in a single transaction.

The OP request is more about being able to meet other (non-primary key) constraints.

mpdude:
#10809 adds an example that shows why we cannot – at least not, in general – change the current order of operations in the UoW.

maddanio:
may I ask why the order is not simply preserved? that would likely be the expectation of the library user. doctrine reordering actions is susprising.

mpdude:
[...]

PR #10809, opened 2023-06-28
Add test to show why delete-before-insert may be challenging

mpdude:
There are a few requests (#5742, #5368, #5109, #6776, #7721, #7722, #2310) that ask to change the order of operations in the UnitOfWork to perform "deletes before inserts", or where such a switch appears to solve a reported problem.

I don't want to say that this is not doable. But this PR at least adds two tricky examples where INSERTs need to be done before an UPDATE can refer to new database rows; and where the UPDATE needs to happen to release foreign key references to other entities before those can be DELETEd.

So, at least as long as all operations of a certain type are to be executed in blocks, this example allows no other order of operations than the current one.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5742Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH5742Test.php
@@ -1,0 +1,158 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH5742Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->createSchemaForModels(
+            GH5742Person::class,
+            GH5742Toothbrush::class,
+            GH5742ToothpasteBrand::class
+        );
+    }
+
+    public function testUpdateOneToOneToNewEntityBeforePreviousEntityCanBeRemoved(): void
+    {
+        $person             = new GH5742Person();
+        $oldToothbrush      = new GH5742Toothbrush();
+        $person->toothbrush = $oldToothbrush;
+
+        $this->_em->persist($person);
+        $this->_em->persist($oldToothbrush);
+        $this->_em->flush();
+
+        $oldToothbrushId = $oldToothbrush->id;
+
+        $newToothbrush      = new GH5742Toothbrush();
+        $person->toothbrush = $newToothbrush;
+
+        $this->_em->remove($oldToothbrush);
+        $this->_em->persist($newToothbrush);
+
+        // The flush operation will have to make sure the new toothbrush
+        // has been written to the database
+        // _before_ the person can be updated to refer to it.
+        // Likewise, the update must have happened _before_ the old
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH5742Test.php
```diff
@@ -0,0 +1,150 @@
...
+class GH5742Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->createSchemaForModels(
+            GH5742Person::class,
+            GH5742Toothbrush::class,
+            GH5742ToothpasteBrand::class
```
greg0ire, line 21: Aaaah a funny test at last… this is going to be good! :popcorn:

tests/Doctrine/Tests/ORM/Functional/Ticket/GH5742Test.php
```diff
@@ -0,0 +1,150 @@
...
+        $person->toothbrush = new GH5742Toothbrush(); // to satisfy not-null constraint
+        $this->_em->persist($person);
+
+        $oldMice = new GH5742ToothpasteBrand();
+        $this->_em->persist($oldMice);
+
+        $person->preferredBrands->set(1, $oldMice);
+        $this->_em->flush();
+
+        $newSpice = new GH5742ToothpasteBrand();
```
greg0ire, line 70: "old mice"? "new spice" What's the joke? I don't get it… :thinking:

greg0ire (dismissed):

tests/Doctrine/Tests/ORM/Functional/Ticket/GH5742Test.php
mpdude: I designed it that way particularly with you in mind 😉

tests/Doctrine/Tests/ORM/Functional/Ticket/GH5742Test.php
mpdude: Old mice 🐁 is not so yum.

New spice might me something fresh like peppermint?

greg0ire: Makes sense :)

SenseException:
What do we expect when those tests fail? An exception that breaks the tests?

mpdude:
Yes, something related to foreign keys, constraint violation and such

tests/Doctrine/Tests/ORM/Functional/Ticket/GH5742Test.php
```diff
@@ -0,0 +1,150 @@
...
+        $this->createSchemaForModels(
+            GH5742Person::class,
+            GH5742Toothbrush::class,
+            GH5742ToothpasteBrand::class
+        );
+    }
+
+    public function testUpdateOneToOneToNewEntityBeforePreviousEntityCanBeRemoved(): void
+    {
+        $this->expectNotToPerformAssertions();
```
derrabus: Maybe it should perform assertions? Right now, a no-op entity manager would pass the test. We could clear the EM after flush, select our person and assert that the right toothbrush is attached. Also, selecting the old toothbrush should yield an empty result.

mpdude: I'll add assertions, that makes it better

mpdude: done

SenseException (approved):

greg0ire:
Thanks @mpdude !

PR #10811, opened 2023-06-29
Add a test case to demonstrate GH-3251

mpdude:
This shows #3251 is still an open issue.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH3251Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH3251Test.php
@@ -1,0 +1,95 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Tests\Models\CMS\CmsGroup;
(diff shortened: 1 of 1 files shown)


mpdude:
I need to learn more about different code paths how collections may get initialized.

When an uninitialized `PersistentCollection` needs to load, `\Doctrine\ORM\PersistentCollection::doInitialize()` will be reached:

*  It tucks newly added collection elements away
* then uses `\Doctrine\ORM\UnitOfWork::loadCollection()`, which in turn calls `loadOneToManyCollection` or `loadManyToManyCollection` in `EntityPersister` classes
* restore new collection entries

→ Works, given that the collection was used lazily (only appending, or `[]`'ing, to the collection.

On the other hand, fetch-joins of collections will reach `\Doctrine\ORM\Internal\Hydration\ObjectHydrator::initRelatedCollection()`. This method somewhat manages `PersistenCollection` instances "from the outside" (set their initialized and dirty state, clearing them...) and does not consider new entries the `PersistentCollection` might hold.

Question is, can we solve this by somehow refactoring `PersistentCollection` in a direction where it better encapsulates state and is not modified that much from the outside?

PR #10813, opened 2023-07-04
Let XML driver treat <id> field attributes same as regular <field> tag

Greg0:
According to other drivers, `id` field, should be treated as regular field and marked as an `id`.
In XML driver for id-field, there was code duplication in parsing field attributes. However, it was not updated at all for a while, so regular field growth with new attributes (e.g. `enum-type`) but id-field not.

Similar to other drivers, id-field attributes are parsed by same method in regular-field and enriched by `id` flag.

For example, without that fix, it is impossible to use XML driver and enum field as part of composite primary key.

Let me know if you need more details for that change.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -376,7 +376,7 @@ class XmlDriver extends FileDriver
                 continue;
             }
 
-            $mapping = $this->columnToArray($idElement);
+            $mapping       = $this->columnToArray($idElement);
             $mapping['id'] = true;
 
             $metadata->mapField($mapping);
--- a/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
@@ -376,30 +376,8 @@ class XmlDriver extends FileDriver
                 continue;
             }
 
-            $mapping = [
-                'id' => true,
-                'fieldName' => (string) $idElement['name'],
-            ];
-
-            if (isset($idElement['type'])) {
-                $mapping['type'] = (string) $idElement['type'];
-            }
-
-            if (isset($idElement['length'])) {
-                $mapping['length'] = (int) $idElement['length'];
-            }
-
-            if (isset($idElement['column'])) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
```diff
@@ -376,30 +376,8 @@ public function loadMetadataForClass($className, PersistenceClassMetadata $metad
...
-            }
-
-            if (isset($idElement['column-definition'])) {
-                $mapping['columnDefinition'] = (string) $idElement['column-definition'];
-            }
-
-            if (isset($idElement->options)) {
-                $mapping['options'] = $this->parseOptions($idElement->options->children());
-            }
+            $mapping = $this->columnToArray($idElement);
```
greg0ire: I am not sure this is correct. Before `columnToArray()` was introduced, the id mapping evaluation looked like this:

https://github.com/doctrine/orm/blob/4df3c753216d144859dd71b8f8021a01881728d1/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php#L292-L302

On the other hand, the field mapping evaluation looked like this: https://github.com/doctrine/orm/blob/4df3c753216d144859dd71b8f8021a01881728d1/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php#L231-L269

That being said, you mention the other drivers already do what you describe, so I guess we could do this for the sake of consistency.

greg0ire (commented):
Please run `vendor/bin/phpcbf` and amend your commit.

Greg0:
> Please run `vendor/bin/phpcbf` and amend your commit.

@greg0ire code style issues fixed

lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php
Greg0: As you said XmlDriver lack of consistency in comparison to other drivers. With annotations/attributes, we declare a standard field and only mark it as ID.
```php
@Id()
@Column(...)
```
In XML, we must declare the whole field as a different tag and it should accept all standard field parameters.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @Greg0 !

PR #10814, opened 2023-07-04
Deprecate classes related to old commit order computation

mpdude:
Here is a last addition for #10547: With the new `TopologicalSort` implementation, we do no longer need the old `CommitOrderCalculator` and related classes.


mpdude:
Closing - putting this directly into #10547 instead.

PR #10818, opened 2023-07-06
Throw when loading a proxy without identifiers

MatTheCat:
Related to https://github.com/symfony/symfony/discussions/50874

Loading a proxy without identifiers currently leads to its hydration using its table’s first row; it is better to throw in that case.
Not sure if it should be done in `EntityPersister:loadById` though :thinking:

### DIFF
--- a/tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
@@ -240,6 +240,18 @@ class ProxyFactoryTest extends OrmTestCase
         self::assertSame(1000, $cloned->getSalary(), 'Expect properties on the CompanyEmployee class to be cloned');
         self::assertSame('Bob', $cloned->getName(), 'Expect properties on the CompanyPerson class to be cloned');
     }
+
+    public function testProxyWithoutIdentifierThrowWhenLoaded(): void
+    {
+        $proxy = $this->proxyFactory->getProxy(ECommerceFeature::class, ['id' => 'toBeRemoved']);
+        $id    = new ReflectionProperty(ECommerceFeature::class, 'id');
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
This is not supposed to happen so this should not be needed.
The root issue is fixed in #10819

MatTheCat:
Well it happened :sweat_smile: 
Luckily the bug was found, but an exception would have been more appropriate.

tests/Doctrine/Tests/ORM/Proxy/ProxyFactoryTest.php
```diff
@@ -240,6 +240,18 @@ public function testProxyClonesParentFields(): void
         self::assertSame(1000, $cloned->getSalary(), 'Expect properties on the CompanyEmployee class to be cloned');
         self::assertSame('Bob', $cloned->getName(), 'Expect properties on the CompanyPerson class to be cloned');
     }
+
+    public function testProxyWithoutIdentifierThrowWhenLoaded(): void
```
greg0ire:
```suggestion
    public function testProxyWithoutIdentifierThrowsWhenLoaded(): void
```

greg0ire:
Might be interesting to see if there are other places in the code where an empty array will cause hard to troubleshoot issues and if yes, introducing an alternate method that throws instead of returning an empty array. Also, finding out what method should not be called with an empty array and annotate them with [`non-empty-array`](https://phpstan.org/writing-php-code/phpdoc-types#general-arrays) instead of just `array`. For instance, does it make sense to call `EntityPersister::loadById()` with an empty array?

MatTheCat:
Indeed. I guess such a work would supersede this PR?

greg0ire:
The first part would, I suppose.

MatTheCat:
Closing as itʼs not really an issue.

PR #10819, opened 2023-07-06
Fix cloning entities when using lazy-ghost proxies

### LINKED ISSUE #10817: Wrong LazyGhost initialization on cloned entity
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.15.3

#### Summary

When you clone an entity with a oneToOne relationship, the cloned entity is filled with data that already exists and has nothing to do with it.

#### Current behavior

The clone is reloaded from the database even though it is a new entity

#### How to reproduce

You can find a reproducer here: https://github.com/Erwui/lazy_ghost_bug
Go on "/"

#### Expected behavior

When setting "enable_lazy_ghost_objects" to false the clone is correct

nicolas-grekas:
Fixes #10817
Replaces #10818
Replaces https://github.com/symfony/symfony/pull/50899

### DIFF
--- a/lib/Doctrine/ORM/Proxy/ProxyFactory.php
+++ b/lib/Doctrine/ORM/Proxy/ProxyFactory.php
@@ -48,30 +48,18 @@ class <proxyShortClassName> extends \<className> implements \<baseProxyInterface
 {
     <useLazyGhostTrait>
 
-    /**
-     * @internal
-     */
-    public bool $__isCloning = false;
-
-    public function __construct(?\Closure $initializer = null)
+    public function __construct(?\Closure $initializer = null, ?\Closure $cloner = null)
     {
+        if ($cloner !== null) {
+            return;
+        }
+
         self::createLazyGhost($initializer, <skippedProperties>, $this);
     }
 
     public function __isInitialized(): bool
     {
         return isset($this->lazyObjectState) && $this->isLazyObjectInitialized();
-    }
-
-    public function __clone()
-    {
-        $this->__isCloning = true;
-
-        try {
-            $this->__doClone();
-        } finally {
-            $this->__isCloning = false;
-        }
     }
 
     public function __serialize(): array
@@ -97,6 +85,9 @@ EOPHP;
      * @var IdentifierFlattener
      */
     private $identifierFlattener;
+
+    /** @var ProxyDefinition[] */
+    private $definitions = [];
 
     /**
      * Initializes a new instance of the <tt>ProxyFactory</tt> class that is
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -98,6 +86,9 @@ public function __serialize(): array
      */
     private $identifierFlattener;
 
+    /** @var ProxyDefinition[] */
```
stof, line 89: this looks like `array<class-string, ProxyDefinition>`

nicolas-grekas: I borrowed from the parent class ;)

nicolas-grekas:
All green :green_apple:

stof (approved):

greg0ire:
Thanks @nicolas-grekas !

mpdude:
Can you guys provide an example when/how it is useful to clone entities and what the semantics of this operation are?

Does the new (cloned) entity let go of its ID? Should the EntityManager know about it? Is it in `MANAGED` or `DETACHED` state etc.?

stof:
@mpdude the cloned object will not be managed by the entity manager

mpdude:
When the entity that is cloned is an uninitialized proxy at that point in time, what happens when I call a method on the clone that triggers loading?

mpdude:
Coming from #3037 and #10785 where one source of "weird" problems is corruption of internal UoW state-tracking fields. This may be caused when using proxies directly created by the ProxyFactory. 

My concern is that cloning a proxy object might be no different:

Suddenly there are two objects claiming to proxy the same entity. But, "there can be only one", only the proxy object the UoW knows about should be initialized and is the sole object representing that entity.

TheBoringGuy1908:
> When the entity that is cloned is an uninitialized proxy at that point in time, what happens when I call a method on the clone that triggers loading?

I stumbled upon this while investigating an issue that seems to be related to this fix, and your suspicion seems to be confirmed.

Since doctrine/orm 2.15.4 I encounter the following problem:
- I have a lazy-loaded entity `original`
- I clone _original_ into `clone`
- I read some data from `original`, causing it to load the remaining data
- I modify some data on `original`

Up until now, everything is as expected. However, once I read some data from `clone` , it reads the missing data back from the database and populates it into `clone` **and** `original`, causing the modifications on `original` to be lost.

nicolas-grekas:
Please open a dedicated issue with a reproducer. Commenting on closed issues/PRs isn't tracked.

PR #10822, opened 2023-07-06
Use --fail-on-log instead of arcane bash tricks

greg0ire:
`--fail-on-log` is a brand new option. As its name indicates, it will fail not just on warnings, and that should be fine.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -44,4 +44,4 @@ jobs:
           dependency-versions: "highest"
 
       - name: "Run guides-cli"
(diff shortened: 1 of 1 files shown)


SenseException (approved):
> it will fail not just on warnings, and that should be fine.

I don't know what is supposed to happen here. Should it fail and break the workflow on a log entry? If so, I'll say this should be tested in the wild :+1:

greg0ire:
Yes, it should break the workflow on a log entry. Any log produced by the tool is something we want to fix. What do you mean by "tested in the wild"?

greg0ire:
We should probably pin to a previous commits because with the latest version, there are lots of errors about unknown directive that we would need to define in a new extension for the guides. They are errors so since we are only grep'ing warnings, it's not an issue for now.

PR #10824, opened 2023-07-07
Add dummy title to the sidebar before checking docs

greg0ire:
The way we have our docs, the sidebar is a separate document and as such, needs a title. Let us prepend a dummy title until the guides-cli provides a way to ignore that particular error.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -43,5 +43,9 @@ jobs:
         with:
           dependency-versions: "highest"
 
+      - name: "Add dummy title to the sidebar"
+        run: |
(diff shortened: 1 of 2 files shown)


SenseException:
What's the result of the change? That "Dummy Page" will be shown?

greg0ire:
The result of this change is the disappearance of a warning about the title missing, and affects only the CI. It doesn't impact what is show on the website, since that dummy title is never under version control. I've discussed this people @linawolf, and she told me the right way to fix this will be to handle [the `:orphan:` metadata](https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html#metadata) in the guides, and tolerate missing title for documents with that metadata. Until that's done, we want to be able to get a green build.

SenseException (approved):

PR #10826, opened 2023-07-09
Undeprecate flushing and clearing a single entity

greg0ire:
Both features have been deprecated without a replacement, making it hard for people to upgrade to ORM 3.0. Let us undeprecate them.

### DIFF
--- a/lib/Doctrine/ORM/EntityRepository.php
+++ b/lib/Doctrine/ORM/EntityRepository.php
@@ -162,13 +162,6 @@ class EntityRepository implements ObjectRepository, Selectable
      */
     public function clear()
     {
-        Deprecation::trigger(
-            'doctrine/orm',
-            'https://github.com/doctrine/orm/issues/8460',
-            'Calling %s() is deprecated and will not be supported in Doctrine ORM 3.0.',
-            __METHOD__
-        );
-
         if (! class_exists(PersistentObject::class)) {
             throw NotSupported::createForPersistence3(sprintf(
                 'Partial clearing of entities for class %s',
--- a/tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
@@ -1093,13 +1093,11 @@ class EntityRepositoryTest extends OrmFunctionalTestCase
         }
     }
 
-    public function testDeprecatedClear(): void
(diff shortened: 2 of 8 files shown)


greg0ire:
I'm targeting the patch branch because the sooner our users stop following the deprecation, the better.

lib/Doctrine/ORM/EntityRepository.php
```diff
@@ -162,13 +162,6 @@ public function createNativeNamedQuery($queryName)
...
-        Deprecation::trigger(
-            'doctrine/orm',
-            'https://github.com/doctrine/orm/issues/8460',
-            'Calling %s() is deprecated and will not be supported in Doctrine ORM 3.0.',
-            __METHOD__
-        );
-
         if (! class_exists(PersistentObject::class)) {
             throw NotSupported::createForPersistence3(sprintf(
                 'Partial clearing of entities for class %s',
```
greg0ire, line 167: I'm not sure what should be done here… should we create persistence 3.3 with this PR reverted: https://github.com/doctrine/persistence/pull/174 ?

greg0ire:
Please disregard the docs build failure, I will handle it separately.

derrabus:
Please target 2.16. This is not a bugfix.

tests/Doctrine/Tests/ORM/Functional/EntityRepositoryTest.php
```diff
@@ -1092,18 +1091,4 @@ public function testFindMultipleByNullValueInMultipleInCriteriaValues(): void
...
-
-        if (class_exists(PersistentObject::class)) {
-            $this->expectDeprecationWithIdentifier('https://github.com/doctrine/orm/issues/8460');
-        } else {
-            $this->expectException(NotSupported::class);
-            $this->expectExceptionMessage(CmsAddress::class);
-        }
-
-        $repository->clear();
-    }
```
derrabus: This test covers the `PersistentObject` removal. We should keep it in some form.

lib/Doctrine/ORM/EntityRepository.php
derrabus: > I'm not sure what should be done here… should we create persistence 3.3 with this PR reverted: https://github.com/doctrine/persistence/pull/174 ?

That would be a BC break.

I think, it's fine to not reintroduce that into Persistence. At this point, partial clearing is a feature the ORM adds on top of the Persistence interfaces.

PR #10828, opened 2023-07-09
Match namespace in XML file with namespace in XSD file

greg0ire:
In #8791, a global search and replace was used for http and https.
This broke the documentation examples in that as soon as you turn on XSD validation, it will fail because the namespace in the XML file does not match the ones defined in the XSD file, which do not exhibit the https.

Note that this is not a security concern, because these URIs are not meant to be actually resolved, but to serve as a unique identifier for the namespace in which we define our elements.

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -281,10 +281,10 @@ specific to a particular entity class's lifecycle.
 
         <?xml version="1.0" encoding="UTF-8"?>
 
-        <doctrine-mapping xmlns="https://doctrine-project.org/schemas/orm/doctrine-mapping"
-              xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
-              xsi:schemaLocation="https://doctrine-project.org/schemas/orm/doctrine-mapping
-                                  https://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
+        <doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
+                          xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
+                          xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
+                                              https://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
             <entity name="User">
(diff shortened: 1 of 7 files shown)


greg0ire:
The docs job error should be handled by #10824 and https://github.com/phpDocumentor/guides/issues/486

derrabus:
Is there a way to actually migrate towards HTTPS? I understand that the documentation is not correct, but it feels like it should be. 🙈

derrabus (approved):

greg0ire:
> Is there a way to actually migrate towards HTTPS? I understand that the documentation is not correct, but it feels like it should be.

Why would we want to? If you browse the URLs modified here, you will get 404s. They're not really URLs, they're more like unique identifiers that mention `http://doctrine-project.org` because we own that domain, so it's a good way to make the string unique. If you want the  migration towards HTTPS in order to avoid well-meaning people to make the same mistake again and change the documentation to something wrong, then I'm afraid I don't see a good way to do that. We'd have to have an alternate XSD file, and to somehow use one or the other depending on the incoming XML file. Sounds like a lot of effort for not much.

Greg0:
What about differences in XSD files 

- https://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd
- https://raw.githubusercontent.com/doctrine/orm/2.15.x/doctrine-mapping.xsd

There are pointing to different protocols (HTTP/HTTPS)

Also opening unsecured `http://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd` redirects now to HTTPS

greg0ire:
Maybe the best would be to use `vendor/doctrine/orm/doctrine-mapping.xsd` in the XML files instead. Doctrine uses that itself internally: https://github.com/doctrine/orm/blob/ec7a8a7a0f854e9874249dc77e2ca3bcba4d5937/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php#L1005-L1007

So I think the only thing calling https://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd might be IDEs, but I use Neovim without a plugin doing such a thing, so I wouldn't know. I wonder what is happening for PHPStorm users. Are they getting errors about invalid XML files?

PR #10830, opened 2023-07-10
 Allow `DateTimeImmutable` be part of composite key

Greg0:
I've added implementation for the attached previously use-case.
The whole description can be found in the linked issue [#10831](https://github.com/doctrine/orm/issues/10831)


greg0ire:
> There are some github action checks that should be run again to handle changes.


What do you mean?

greg0ire:
You may disregard the documentation build, it's not stable yet.

Greg0:
When can I expect a review of PR?

docs/en/tutorials/composite-primary-keys.rst
```diff
@@ -14,12 +14,12 @@ General Considerations
 Every entity with a composite key cannot use an id generator other than "NONE". That means
 the ID fields have to have their values set before you call ``EntityManager#persist($entity)``.
 
-Primitive Types only
-~~~~~~~~~~~~~~~~~~~~
+Allowed Types
+~~~~~~~~~~~~~
 
 You can have composite keys as long as they only consist of the primitive types
```
greg0ire, line 20: You changed the title probably because you're considering `DateTimeImmutable` isn't a primitive type. Should this sentence be changed as well then?

tests/Doctrine/Tests/ORM/Utility/IdentifierFlattenerDateTimeImmutableIdTest.php
```diff
@@ -0,0 +1,94 @@
...
+        $this->articleRepository      = $this->_em->getRepository(Article::class);
+        $this->articleAuditRepository = $this->_em->getRepository(ArticleAudit::class);
+    }
+
+    /** @group utilities */
+    public function testFlattenIdentifierWithDateTimeId(): void
+    {
+        $article = new Article('Some title');
+        $article->changeTitle('New title');
+        sleep(1);
```
greg0ire: Please find a way to avoid this. Maybe define a class that extends `DateTimeImmutable` and makes sure to add 1 second when the constructor is called, I don't know.

BTW, doesn't this point at a weakness of having time part of the composite key? What happens in high throughput scenarios? People might end up with entities with the same ID, might they not? I'm starting to have some doubts about this…

greg0ire (commented):
Codecov reported 2 blocks of code as uncovered, did you notice that?

Greg0:
> Codecov reported 2 blocks of code as uncovered, did you notice that?

I've only noticed issues with documentation checks
obraz

docs/en/tutorials/composite-primary-keys.rst
Greg0: Whole sentence is 
> You can have composite keys as long as they only consist of the primitive types ``integer`` and ``string`` **or** object of ``DateTimeImmutable``

tests/Doctrine/Tests/ORM/Utility/IdentifierFlattenerDateTimeImmutableIdTest.php
Greg0: I consider such validation as a part of entity design. Ofc I can add that validation in tested entities to get rid of that `sleep` function.

It is not the responsibility of Doctrine to think for programmers on the consistency of their entities.

greg0ire:
> I've only noticed issues with documentation checks

It appears in the form of comments on the files tab, when you have a successful build (which explains why it isn't available right now).

Greg0:
> > I've only noticed issues with documentation checks
> 
> It appears in the form of comments on the files tab, when you have a successful build (which explains why it isn't available right now).

Everything passed now

greg0ire:
This is what I'm talking about:

Greg0:
@greg0ire Can you point me to a place where tests that cover `UnitOfWork::createEntity` are located? Or where should I place it? I looked for them and didn't find anything promising.

greg0ire:
I'm on my phone so no. Have you tried throwing an exception in nearby code and running the test suite.

tests/Doctrine/Tests/ORM/Utility/IdentifierFlattenerDateTimeImmutableIdTest.php
greg0ire: It's not that I want to tell people what's wrong or right, but I would like to avoid people opening bug reports, have us help them debug only to find they are hitting this limitation. That's why I'm weary of code that works only if people know what they are doing.

Greg0: Agree but requirement of having unique entity identifier is not magic knowledge when you are dealing with ORM and even database I guess. What do you think?

greg0ire: It sounds sensible. Let's keep that thread open, and see what other maintainers think.

Greg0:
> I'm on my phone so no. Have you tried throwing an exception in nearby code and running the test suite.

I figured out how to check that part of the code. I need more time for that and I'll let you know when it will be finished

arkhamvm:
Any news? Faced same problem with `DateTime` in `3.x` branch.

arkhamvm:
> Any news? Faced same problem with `DateTime` in `3.x` branch.

Still actual problem.

Greg0:
I'm faced too many issues in entity management when working on covering code with tests so I gave up. Maybe someone else based on my changes will have better idea how to solve it proper.

AFAIR main issue was: to treat datetime field as part of key you must cast it to string, but you can't cast it to string because it has to be object field in other processes like persistence. With `enum` type it was easier because `enum` in fact is regular field with type=string which is only marked in mapping as enum driven https://www.doctrine-project.org/projects/doctrine-laminas-hydrator/en/3.4/enum-strategy.html#the-enum-and-entity

PR #10832, opened 2023-07-10
Decouple public API from Doctrine\Persistence\Proxy

nicolas-grekas:
Related to https://github.com/doctrine/persistence/pull/334
On the path to deprecating `Doctrine\Persistence\Proxy` in the future.

Adding `EntityManager::isUninitializedEntity()` will allow userland to decouple from the `Proxy` interface:

before:
```php
if ($entity instanceof Proxy && !$entity->__isInitialized()) {
    $entity->__load();
}
```

after:
```php
if ($em->isUninitializedEntity($entity)) {
    $em->initializeObject($entity);
}
```

### DIFF
--- a/lib/Doctrine/ORM/EntityManager.php
+++ b/lib/Doctrine/ORM/EntityManager.php
@@ -954,6 +954,14 @@ class EntityManager implements EntityManagerInterface
     }
 
     /**
+     * {@inheritDoc}
+     */
+    public function isUninitializedObject($obj): bool
+    {
+        return $this->unitOfWork->isUninitializedObject($obj);
+    }
+
+    /**
      * Factory method to create EntityManager instances.
      *
      * @deprecated Use {@see DriverManager::getConnection()} to bootstrap the connection and call the constructor.
--- a/lib/Doctrine/ORM/Proxy/InternalProxy.php
+++ b/lib/Doctrine/ORM/Proxy/InternalProxy.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Proxy;
+
+use Doctrine\Persistence\Proxy;
+
+/**
+ * @internal
+ *
+ * @template T of object
+ * @template-extends Proxy<T>
+ *
+ * @method void __setInitialized(bool $initialized)
+ */
+interface InternalProxy extends Proxy
+{
+}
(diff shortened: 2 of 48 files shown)


lib/Doctrine/ORM/EntityManager.php
```diff
@@ -953,6 +953,14 @@ public function initializeObject($obj)
         $this->unitOfWork->initializeObject($obj);
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function isUninitializedEntity($obj)
```
stof: you should add the parameter and return types

lib/Doctrine/ORM/Proxy/InternalProxy.php
```diff
@@ -0,0 +1,16 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Proxy;
+
+use Doctrine\Persistence\Proxy;
+
+/**
+ * @internal
+ *
+ * @method void __setInitialized(bool $initialized)
```
stof, line 15: if it is an internal API, can't we define the method in the interface directly ?

lib/Doctrine/ORM/EntityManager.php
nicolas-grekas: return type done, `mixed` for param is not compatible with PHP 7.1. `@inheritDoc` to the rescue, from https://github.com/doctrine/persistence/pull/334

lib/Doctrine/ORM/Proxy/InternalProxy.php
nicolas-grekas: Not possible because this overlaps with exiting declarations on legacy Proxy interfaces / implementations.

nicolas-grekas:
PR green :green_apple:  (ignoring unrelated doc failures.)

beberlei:
This changes code from not requiring the EntityManager, to requiring the entity manager to make this check.

Are we ok with this, since explicit proxy loading puts the code into "persistence context" that probably has the entity manager already?

nicolas-grekas:
> Are we ok with this, since explicit proxy loading puts the code into "persistence context" that probably has the entity manager already?

That's my though also yes: the code shouldn't care about proxy or not (as stated in the doc). But when caring, it means you have the EM not too far away (if not already there). The concept of a Doctrine proxy is in the same domain as Doctrine ORM/ODM in this context.

stof (dismissed):

lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -2832,7 +2819,8 @@ public function createEntity($className, array $data, &$hints = [])
                 }
             }
 
-            if ($entity instanceof Proxy && ! $entity->__isInitialized()) {
+            if ($this->isUninitializedObject($entity)) {
+                assert($entity instanceof InternalProxy);
```
derrabus: A proper `@psalm-assert-if-true` annotation on the new `isUninitializedObject()` method and we should not need this `assert()`.

https://psalm.dev/docs/annotating_code/adding_assertions/

nicolas-grekas: Updated, thanks.

derrabus (approved):

PR #10837, opened 2023-07-12
Deprecate not-enabling lazy-ghosts and decouple from doctrine/common's proxies

nicolas-grekas:
This PR deprecates not-enabling lazy-ghosts so that ORM v3 can drop support for doctrine/common's proxies.

It also rewrites ProxyFactory so that when lazy-ghosts are enabled, none of the code in the `Doctrine\Common\Proxy` namespace is run. This enables deprecating everything related to proxies in doctrine/common.


lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -102,23 +164,38 @@ public function __serialize(): array
...
-        $proxyGenerator = new ProxyGenerator($proxyDir, $proxyNs);
-
-        if ($em->getConfiguration()->isLazyGhostObjectEnabled()) {
-            $proxyGenerator->setPlaceholder('baseProxyInterface', Proxy::class);
-            $proxyGenerator->setPlaceholder('useLazyGhostTrait', Closure::fromCallable([$this, 'generateUseLazyGhostTrait']));
-            $proxyGenerator->setPlaceholder('skippedProperties', Closure::fromCallable([$this, 'generateSkippedProperties']));
-            $proxyGenerator->setPlaceholder('serializeImpl', Closure::fromCallable([$this, 'generateSerializeImpl']));
-            $proxyGenerator->setProxyClassTemplate(self::PROXY_CLASS_TEMPLATE);
-        } else {
+        if (! $em->getConfiguration()->isLazyGhostObjectEnabled()) {
```
stof, line 164: should we trigger this deprecation only on PHP 8.1+ ? When using older PHP versions, they cannot enable it as they cannot install the needed dependency.

lib/Doctrine/ORM/ORMInvalidArgumentException.php
```diff
@@ -261,6 +261,32 @@ public static function invalidEntityName($entityName)
         return new self(sprintf('Entity name must be a string, %s given', get_debug_type($entityName)));
     }
 
+    /** @param mixed $value */
+    public static function invalidAutoGenerateMode($value): self
+    {
+        return new self(sprintf('Invalid auto generate mode "%s" given.', $value));
```
derrabus: We cannot be certain that `mixed $value` can be converted to a string.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -127,19 +204,57 @@ public function __construct(EntityManagerInterface $em, $proxyDir, $proxyNs, $au
...
+        $proxyFactory = $this->proxyFactories[$className] ?? $this->getProxyFactory($className);
+
+        return $proxyFactory($identifier);
+    }
+
+    /**
+     * Generates proxy classes for all given classes.
+     *
+     * @param ClassMetadata[] $classes  The classes (ClassMetadata instances) for which to generate proxies.
+     * @param string          $proxyDir The target directory of the proxy classes. If not specified, the
```
derrabus:
```suggestion
     * @param string|null     $proxyDir The target directory of the proxy classes. If not specified, the
```

lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -181,6 +292,8 @@ protected function createProxyDefinition($className)
     /**
      * Creates a closure capable of initializing a proxy
      *
+     * @deprecated ProxyFactory::createInitializer() is deprecated and will be removed in version 3.0 of doctrine/orm.
```
derrabus, line 294: Do we need to deprecate a private method?

lib/Doctrine/ORM/Proxy/ProxyFactory.php
nicolas-grekas: We can but this won't save ppl from this one: https://github.com/doctrine/common/pull/1002/files#diff-58c63ccd37f4120e07f5995493641146490fc19779d4865aa4669fc1ce6282ef

I have a slight preference to trigger the deprecation regardless of the PHP version. Please advise.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
nicolas-grekas: I added those to help when cleaning for 3.0. Let me know.

nicolas-grekas:
PR updated, thanks for the review @derrabus

lib/Doctrine/ORM/Proxy/ProxyFactory.php
stof: Well, this one should probably also be conditional, or the ORM should bump its PHP min version first.

nicolas-grekas: I added the version check.

nicolas-grekas:
PR rebased, all green!

lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -10,29 +10,92 @@
...
+
+    /**
+     * Autogenerate the proxy class when the proxy file does not exist or
+     * when the proxied file changed.
+     *
+     * This strategy causes a file_exists() call whenever any proxy is used the
+     * first time in a request. When the proxied file is changed, the proxy will
+     * be updated.
+     */
+    public const AUTOGENERATE_FILE_NOT_EXISTS_OR_CHANGED = 4;
```
derrabus, line 98: Can be done later: We could think about moving those constants into a separate class, which would allow us to convert that class into an enum in 3.0.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -97,23 +160,40 @@ public function __serialize(): array
...
-            $proxyGenerator->setPlaceholder('skippedProperties', Closure::fromCallable([$this, 'generateSkippedProperties']));
-            $proxyGenerator->setPlaceholder('serializeImpl', Closure::fromCallable([$this, 'generateSerializeImpl']));
-            $proxyGenerator->setProxyClassTemplate(self::PROXY_CLASS_TEMPLATE);
-        } else {
+        if (! $em->getConfiguration()->isLazyGhostObjectEnabled()) {
+            if (PHP_VERSION_ID >= 80100) {
+                Deprecation::trigger(
+                    'doctrine/orm',
+                    'https://github.com/doctrine/orm/pull/10837/',
+                    'Not enabling lazy ghost objects is deprecated and will not be supported in Doctrine ORM 3.0. Ensure Doctrine\ORM\Configuration::setLazyGhostObjectEnabled(true) is called to enable them.'
```
derrabus, line 169: We should deprecate calling `Configuration::setLazyGhostObjectEnabled(false)` as well because `true` is going to be the only valid setting in 3.0.

derrabus (dismissed):

lib/Doctrine/ORM/Proxy/ProxyFactory.php
nicolas-grekas: I'm not sure it's worth the trouble for the community, especially because those might not be needed at all with the anticipated native lazy objects.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
nicolas-grekas: It's kinda already the case, because if you set this to false, you'll get the deprecation added to ProxyFactory.

derrabus: I see.

derrabus:
Rebase needed.

nicolas-grekas:
Rebased and :green_apple:

derrabus (approved):

PR #10848, opened 2023-07-20
Use properties instead of getters to read property/class names via reflection

nicolas-grekas:
Saves a few methods calls + object instantiations for `getDeclaringClass()`


derrabus (approved):

derrabus:
Thank you!

PR #10849, opened 2023-07-23
Fix templated phpdoc return type

mvorisek:
When split between `@param` and `@psalm-param`, only the `@psalm-param` is used by psalm/phpstan, ie `class-string<T>` is used instead of (too wide) `class|class-string<T>`.

In the 2nd commit, I fixed all other simillar problems in the whole repo. I used `@.*(string.*\|class-s|class-string.*\|.*string)` regex to find the lines to check/fix.

### DIFF
--- a/lib/Doctrine/ORM/EntityManagerInterface.php
+++ b/lib/Doctrine/ORM/EntityManagerInterface.php
@@ -335,7 +335,8 @@ interface EntityManagerInterface extends ObjectManager
     /**
      * {@inheritDoc}
      *
-     * @psalm-param string|class-string<T> $className
+     * @param string $className
+     * @psalm-param class-string<T> $className
      *
      * @return Mapping\ClassMetadata
      * @psalm-return Mapping\ClassMetadata<T>
--- a/lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
+++ b/lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
@@ -22,7 +22,7 @@ use const PHP_VERSION_ID;
 /** @psalm-type ScalarName = 'array'|'bool'|'float'|'int'|'string' */
 final class DefaultTypedFieldMapper implements TypedFieldMapper
 {
-    /** @var array<class-string|ScalarName, class-string<Type>|string> $typedFieldMappings */
+    /** @var array<class-string|ScalarName, class-string<Type>> $typedFieldMappings */
     private $typedFieldMappings;
 
     private const DEFAULT_TYPED_FIELD_MAPPINGS = [
@@ -36,7 +36,7 @@ final class DefaultTypedFieldMapper implements TypedFieldMapper
         'string' => Types::STRING,
     ];
 
-    /** @param array<class-string|ScalarName, class-string<Type>|string> $typedFieldMappings */
+    /** @param array<class-string|ScalarName, class-string<Type>> $typedFieldMappings */
     public function __construct(array $typedFieldMappings = [])
     {
         $this->typedFieldMappings = array_merge(self::DEFAULT_TYPED_FIELD_MAPPINGS, $typedFieldMappings);
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -136,11 +136,6 @@ parameters:
 			path: lib/Doctrine/ORM/EntityManager.php
 
 		-
-			message: "#^Template type T of method Doctrine\\\\ORM\\\\EntityManagerInterface\\:\\:getClassMetadata\\(\\) is not referenced in a parameter\\.$#"
-			count: 1
-			path: lib/Doctrine/ORM/EntityManagerInterface.php
-
-		-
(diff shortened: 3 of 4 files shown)


mvorisek:
I regenerated the psalm baseline locally, but the CI is still failing. Please regenerate it before merge.

derrabus:
I don't believe that this change is correct. Can you elaborate what issue you're trying to fix here?

mvorisek:
phpstan cannot otherwise infer the resulting type for `EntityManagerInterface::getClassMetadata()` method, see the no longer needed baseline ignore [1]

@derrabus why do you think this change is not correct?

see other methods phpdocs in `EntityManagerInterface` class, they are all, except this one, in the form I fixed this one to

[1] and in my usecase, I needed to add `Unable to resolve the template type T in call to method static method Doctrine\\Common\\Util\\ClassUtils::getClass\(\)` phpstan ignore /wo this PR

derrabus:
> see the no longer needed baseline ignore
> […]
> why do you think this change is not correct?

See all the new Psalm errors that pop up after your change. 🙂

mvorisek:
> > see the no longer needed baseline ignore
> > […]
> > why do you think this change is not correct?
> 
> See all the new Psalm errors that pop up after your change. 🙂

@derrabus looking into the psalm errors -> code: https://github.com/doctrine/orm/blob/2.15.4/lib/Doctrine/ORM/Cache.php#L42 it seems the are many places that accept a class name but are annotated as `string` instead of `class-string`.

I am not paid to maintain this project so I would be happy if this PR can be accepted and the psalm baseline regenerated (I tried myself on Windows, but the updated baseline did not fix all errors).

derrabus:
> I am not paid to maintain this project

Neither am I. 🤷🏻‍♂️ 

> I would be happy if this PR can be accepted

I cannot merge a PR with a red CI and I won't fix other people's PRs in my freetime.

SenseException:
@mvorisek What kind of problems do you have that you want to fix with this PR? Your last comment doesn't sound like that this is relevant to your daily work with ORM but more of an aid. We appreciate the effort people put into PRs as users (hopefully) appreciate the time we spent into the Doctrine project repositories on a regular basis.

We would like to know how this should be handled, because, as @derrabus already mentioned, this would rather be a hinderance to our also unpaid freetime if the failed build doesn't get fixed. I'd like to ask you to go the second half of what you've started.

lib/Doctrine/ORM/EntityManagerInterface.php
```diff
@@ -335,7 +335,8 @@ public function hasFilters();
     /**
      * {@inheritDoc}
      *
-     * @psalm-param string|class-string<T> $className
+     * @param string $className
+     * @psalm-param class-string<T> $className
```
greg0ire, line 339: `EntityManager::getClassMetadata()`  also supports aliased names and needs to do so until Persistence 2 support is removed:

https://github.com/doctrine/orm/blob/f7e4b61459692f9b747f40696e6bf72080390f2d/lib/Doctrine/ORM/EntityManager.php#L305-L316

Should the same be expected of any `EntityManagerInterface` implementation? If yes I'd say you're not so much fixing stuff as you are breaking it.

> Fix templated phpdoc return type 

Did you mean to write "parameter type" here?

greg0ire (changes requested):
All 4 of your commit messages are unhelpful, or sometimes even plain wrong, it seems. As I probably have already told you in the past, you should read the contribution guidelines to figure out what to write, they're fairly extensive on the topic.

As for the the CI build, if you want to contribute to Doctrine, you should setup a proper development environment.

mvorisek:
> What kind of problems do you have that you want to fix with this PR?

here is a phpstan repro: https://phpstan.org/r/b2cbe847-a748-4094-bc49-44bd3e200e12 explaining the need

notice the dumped type change when l6 is changed to `@psalm-param class-string<T> $className`

lib/Doctrine/ORM/EntityManagerInterface.php
mvorisek: @greg0ire see https://github.com/doctrine/orm/pull/10849#issuecomment-1648223127 for repro, I meant "return type", as the "return type" is currently not correctly narrowed as the  "parameter type" does allow any `string`

derrabus:
This function accepts strings that are not FQCN, so changing the parameter type to `class-string` would be wrong. But if we a `class-string<SomeClass>` is passed, the return type to expect is `ClassMetadata<SomeClass>`.

Can we solve this with a conditional return type?

lib/Doctrine/ORM/EntityManagerInterface.php
greg0ire: Related: https://github.com/doctrine/orm/pull/10853

VincentLanglet:
```
* @param string $className
     * @psalm-param class-string<T> $className
```

Might be a good way to deprecate string param for static analysis user
and solve https://github.com/phpstan/phpstan/issues/5175

VincentLanglet:
> This function accepts strings that are not FQCN, so changing the parameter type to `class-string` would be wrong. But if we a `class-string<SomeClass>` is passed, the return type to expect is `ClassMetadata<SomeClass>`.
> 
> Can we solve this with a conditional return type?

There was an issue about this in phpstan
https://github.com/phpstan/phpstan/issues/5175

mvorisek:
Closing as it seems this change is not wanted, at least for ORM 2.x.

VincentLanglet:
> Closing as it seems this change is not wanted, at least for ORM 2.x.

I think the string support was deprecated so maybe the issue does not exist in 3.x anymore (or the phpdoc could be updated).

Need to check

PR #10855, opened 2023-07-25
fix: use platform options instead of deprecated custom options

dunglas:
DBAL's custom schema options have been deprecated in favor of platform options.

This patch uses the recommended method to prevent deprecations.

Closes #10854.


greg0ire (approved):
See https://github.com/doctrine/dbal/pull/5476 for details.

According to the build, no need to bump the DBAL version, which makes sense when reading the pull request above: the platform options seem to have existed for a long time.

derrabus (approved):

PR #10856, opened 2023-07-26
Fix/Self deprecation with getQueryCacheImpl

VincentLanglet:
Hi @greg0ire, I got another "non solvable deprecation" from doctrine.

`getQueryCacheImpl` is used in `QueryCommand::execute`

### DIFF
--- a/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/QueryCommand.php
+++ b/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/QueryCommand.php
@@ -17,6 +17,7 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
+use function assert;
 use function get_debug_type;
 use function sprintf;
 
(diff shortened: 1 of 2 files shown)


greg0ire:
Looking at that command, it looks like it could be reworked to avoid calls to that method when `$cache` is not `null`. I'm assuming you cannot configure both a `psr/cache` and a `doctrine/cache`, but that might be a wrong assumption, please check.

lib/Doctrine/ORM/Tools/Console/Command/ClearCache/QueryCommand.php
```diff
@@ -63,32 +64,46 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            }
         }
 
         $ui->comment('Clearing <info>all</info> Query cache entries');
 
-        $result  = $cache ? $cache->clear() : $cacheDriver->deleteAll();
+        if ($cache) {
+            $result = $cache->clear();
+        } else {
+            assert($cacheDriver !== null);
```
greg0ire, line 103: :thinking: is caching mandatory? Why couldn't both `$cache` and `$cacheDriver` be null?

VincentLanglet: Previously there was 
```
$result = $cache ? $cache->clear() : $cacheDriver->deleteAll();
```

the fact is that with the refacto $cache and $cacheDriver cannot be null at the same time but unlike phpstan, psalm doesn't understand this so an assert is required.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @VincentLanglet !

PR #10865, opened 2023-08-02
Add possibility to set reportFieldsWhereDeclared in ORMSetup

W0rma:
The parameter `$reportFieldsWhereDeclared` was introduced to the `AttributeDriver` in https://github.com/doctrine/orm/pull/10455 and setting it to `false` is deprecated.

However, `ORMSetup::createAttributeMetadataConfiguration()` doesn't provide a way to fix this warning:

```
DEPRECATION: In ORM 3.0, the AttributeDriver will report fields for the classes where they are declared. This may uncover invalid mapping configurations. To opt into the new mode today, set the "reportFieldsWhereDeclared" constructor parameter to true. (AttributeDriver.php:82 called by ORMSetup.php:107, https://github.com/doctrine/orm/pull/10455, package doctrine/orm) /var/www/html/vendor/doctrine/deprecations/lib/Doctrine/Deprecations/Deprecation.php:209
```

### DIFF
--- a/docs/en/reference/advanced-configuration.rst
+++ b/docs/en/reference/advanced-configuration.rst
@@ -29,7 +29,7 @@ steps of configuration.
 
     $config = new Configuration;
     $config->setMetadataCache($metadataCache);
-    $driverImpl = new AttributeDriver(['/path/to/lib/MyProject/Entities']);
+    $driverImpl = new AttributeDriver(['/path/to/lib/MyProject/Entities'], true);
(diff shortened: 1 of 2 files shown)


derrabus:
Thank you. I agree that we should treat this as a bugfix. @mpdude WDYT?

derrabus (approved):

mpdude:
Agree.

Seems I added the flag in `\Doctrine\ORM\ORMSetup::createDefaultAnnotationDriver()`, but missed that method below.

PR #10870, opened 2023-08-03
Fix return type of `getSingleScalarResult`

whatUwant:
`Doctrine\ORM\AbstractQuery::getSingleScalarResult()` does not have `null` as a possible type of returned data while `NoResultException` is listed as one of the possible exceptions thrown.

However `NoResultException` is thrown only when hydration mode is not `HYDRATE_SINGLE_SCALAR`. 

```php
    public function getSingleResult($hydrationMode = null)
    {
        $result = $this->execute(null, $hydrationMode);

        if ($this->_hydrationMode !== self::HYDRATE_SINGLE_SCALAR && ! $result) {
            throw new NoResultException();
        }

```

In `getSingleScalarResult` hydration mode was set to `HYDRATE_SINGLE_SCALAR` via `Doctrine\ORM\AbstractQuery::execute()` which makes it impossible to throw that particular type of exception.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -1010,9 +1010,8 @@ abstract class AbstractQuery
      *
      * Alias for getSingleResult(HYDRATE_SINGLE_SCALAR).
      *
-     * @return bool|float|int|string The scalar result.
+     * @return bool|float|int|string|null The scalar result.
      *
-     * @throws NoResultException        If the query returned no result.
      * @throws NonUniqueResultException If the query result is not unique.
      */
     public function getSingleScalarResult()


derrabus (approved):

darenas31415:
> However `NoResultException` is thrown only when hydration mode is not `HYDRATE_SINGLE_SCALAR`.

That's not quite correct as `$result = $this->execute(null, $hydrationMode);` can still throw `NoResultException` exception.

Please see screenshot:

image

How to reproduce:

```
$qb = $this->em->createQueryBuilder();

$qb->select('p.code')
   ->from('App:Product', 'p')
   ->where('p.name = :name')
   ->setMaxResults(1)
   ->setParameter('name', 'missing');

$code = $qb->getQuery()->getSingleScalarResult();
```

derrabus:
You're right. Please send a PR to add back the `@throws` annotation.

darenas31415:
Shall I add back only `@throws` annotation or revert this PR completely?

I can't make `getSingleScalarResult` return `null` under any condition, I think the original code was correct.

I'd suggest to revert this PR altogether unless @whatUwant can provide a test case for returning `null`.

whatUwant:
It's correct that `NoResultException` can still be thrown. 
`null` can be returned when aggregate functions such as `MIN` `MAX` are used.
```
$qb = $this->em->createQueryBuilder();

$qb->select('MAX(p.id)')
   ->from('App:Product', 'p')
   ->where('p.name = :name')
   ->setParameter('name', 'missing');

$code = $qb->getQuery()->getSingleScalarResult();
```

derrabus:
`getSingleScalarResult()` with a `SELECT` on a nullable field should yield `null`, shouldn't it?

darenas31415:
> `getSingleScalarResult()` with a `SELECT` on a nullable field should yield `null`, shouldn't it?

You are totally right :+1:

Issue #10868, opened 2023-08-02
IdentityMap


tasselchof:
### Bug Report

|    Q        |   A
|------------ | ------
| Version     | 2.16.0

#### Summary

I am getting an exception about entity in identity map

#### Current behavior

While adding an entity of class Proxy was already present for the same ID. This exception\nis a safeguard against an internal inconsistency - IDs should uniquely map to\nentity object instances. This problem may occur if:\n\n- you use application-provided IDs and reuse ID values;\n- database-provided IDs are reassigned after truncating the database without \n  clearing the EntityManager;\n- you might have been using EntityManager#getReference() to create a reference \n  for a nonexistent ID that was subsequently (by the RDBMS) assigned to another \n  entity. \n\nOtherwise, it might be an ORM-internal inconsistency, please report it.

#### How to reproduce

I am not sure how to reproduce it, but I am getting this in that case:

I am getting some values from database with native query (I am using composite keys) and generating reference from it (I can fetch from database this value - no matter, everything is fine):

```php
$productOffer = $this->getObjectManager()->getReference(
    Offer::class,
    [
        'shop' => $locationLog['sclr_0'],
        'id'   => $locationLog['sclr_1'],
    ]
);
```

I am using this reference in further request and if I generate spl_object_id it's always same object. But later I am fetching one another entity from database:

```php
$acceptanceItem = $this->getObjectManager()->getRepository(
    Acceptance\Item::class
)->find($acceptanceItem);
```

This entity contains property productOffer with is linked with the same ids (so technically it contains same object).

And if I am checking `spl_object_id($acceptanceItem->getProductOffer())` it's different object.

I am using the value of this object in preFlush to populate search index:

```php
public function preFlush(PreFlushEventArgs $args)
    {
        $extendedValues = [];
        if (! empty($this->getProductOffer())) {
            $productOffer = $this->getProductOffer();

            $extendedValues[] = $productOffer->getName();
            $extendedValues[] = $productOffer->getSku();
            $extendedValues[] = $productOffer->getArticle();
```

And exactly here I am getting this exception:

```
#0  Doctrine\ORM\UnitOfWork->addToIdentityMap() called at [/home/dev/domains/alpha.eu/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:3399]
[...]

PR #10873, opened 2023-08-03
(Try to) add a reproducer for #10868

mpdude:
See #10868

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10868Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10868Test.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\ORM\Functional\Ticket;
 
+use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\ORM\Mapping as ORM;
 use Doctrine\Tests\OrmFunctionalTestCase;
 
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH10868Test.php
```diff
@@ -0,0 +1,192 @@
...
+        $this->_em->persist($shop);
+        $this->_em->persist($offer);
+        $this->_em->persist($order);
+        $this->_em->persist($orderProduct);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $order->orderProducts->count();
+
+        $reference = $this->_em->getReference(GH10868Offer::class, [
```
mpdude, line 46: This is a reference (proxy) for a `GH10868Offer` object.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10868Test.php
```diff
@@ -0,0 +1,192 @@
...
+        $order->orderProducts->add($orderProduct);
+
+        $this->_em->persist($shop);
+        $this->_em->persist($offer);
+        $this->_em->persist($order);
+        $this->_em->persist($orderProduct);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $order->orderProducts->count();
```
mpdude, line 44: We should not do that, since `$order` belongs to an EM instance that has been cleared.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10868Test.php
```diff
@@ -0,0 +1,192 @@
...
+
+        $reference = $this->_em->getReference(GH10868Offer::class, [
+            'shop' => $shop->id,
+            'id' => $offer->id,
+        ]);
+
+        $orderProductFromOrder = $order->orderProducts->first();
```
mpdude, line 51: We should not do anything off `$order`, since it is an entity from an EM that has been cleared.

tasselchof:
Ok, I didn't paid enough attention to this thing as I was not trying to isolate such cases before. After getting order once more it worked. The question is how to reproduce case that I have in live code...

Issue #10869, opened 2023-08-03
2.16.0 - Issue with persisting entities


dmaicher:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.16.0

#### Summary

I have a case in our test suite where entities are not persisted properly anymore. This worked fine on 2.15.

The insert statement does not bind any parameters and fails:

```
PDOException: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' at line 1

/var/www/app/symfony/vendor/doctrine/dbal/src/Driver/PDO/Statement.php:121
/var/www/app/symfony/vendor/doctrine/dbal/src/Driver/Middleware/AbstractStatementMiddleware.php:69
/var/www/app/symfony/vendor/doctrine/dbal/src/Logging/Statement.php:98
/var/www/app/symfony/vendor/doctrine/dbal/src/Driver/Middleware/AbstractStatementMiddleware.php:69
/var/www/app/symfony/vendor/symfony/doctrine-bridge/Middleware/Debug/Statement.php:68
/var/www/app/symfony/vendor/doctrine/dbal/src/Statement.php:190
/var/www/app/symfony/vendor/doctrine/dbal/src/Statement.php:249
/var/www/app/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php:280
/var/www/app/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:1177
/var/www/app/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:442
/var/www/app/symfony/vendor/doctrine/orm/lib/Doctrine/ORM/EntityManager.php:403
``` 

The logs (see `parameters: []`):

```
[...]

Issue #8695, opened 2021-05-17
PostFlush event is dispatched before unitofwork is cleanup


alexander-schranz:
Currently the postflush event is dispatched before the unitofwork is cleanup. We tried to add some listener to it but got into some problems because the `UnitOfWork` was not yet cleanup. The code can be found here:

https://github.com/doctrine/orm/blob/3a194ad69946ec3e4289e50a877f96b8aad8b223/lib/Doctrine/ORM/UnitOfWork.php#L378-L380

https://github.com/doctrine/orm/blob/3a194ad69946ec3e4289e50a877f96b8aad8b223/lib/Doctrine/ORM/UnitOfWork.php#L472-L474

For our case it would be great if the `UnitOfWork` cleanup is done before dispatching the post events e.g.:

```diff
-            $this->dispatchPostFlushEvent();
+            $this->postCommitCleanup($entity);

-            $this->postCommitCleanup($entity);
+            $this->dispatchPostFlushEvent();
 ```
 
 I understand that this could be a bc break so an alternative would be add a new `postFlushCleanup` event which we could listen to it. What do you think?

alexander-schranz:
closing this as it has todo with https://github.com/doctrine/orm/issues/11827 which has a reproducer.

PR #10874, opened 2023-08-03
Emit `postFlush` events as the very last thing in `UoW::flush()`

mpdude:
This PR moves the `postFlush` event dispatching _after_ final clean-up steps in `UnitOfWork::commit()`.

The idea is that at this time, `commit()` is "as done as it gets", and it might be possible to call `EntityManager::flush()` again from event listeners – we'd return from that method on the very next line, and everything would be in the same state if users dediced to call `flush()` again themselves.

Obviously, infinite recursion is one thing either the ORM or users would need to think about.

The intent is to make it somehow possible for users to have event listeners that get notified e. g. through `postPersist` to prepare "cascading" changes. These changes could be collected and applied in a `postFlush` listener, which might then call `flush()` again.

This sparked off from #10869, where the reporting user is calling `flush()` even from `postPersist` – a questionable practice, probably never endorsed but it "somehow worked" until 2.16.0.

⚠️ BC Break: If `postFlush` listeners inspect UoW state through methods like `isSchedulesForInsert` or `getEntityChangeSet`, this information will no longer available with this change. → Maybe we need an additional event?

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -402,9 +402,8 @@ class UnitOfWork implements PropertyChangedListener
                 $this->orphanRemovals)
         ) {
             $this->dispatchOnFlushEvent();
-            $this->dispatchPostFlushEvent();
-
             $this->postCommitCleanup($entity);
+            $this->dispatchPostFlushEvent();
 
             return; // Nothing to do.
         }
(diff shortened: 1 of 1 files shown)


alexander-schranz:
This would also solve: https://github.com/doctrine/orm/issues/8695

mpdude:
Probably this needs to be a new event for BC reasons.

We also need to think about transaction boundaries: Should the second round of `flush()` (that’s what we want this event for, right?) happen within the current transaction?

dmaicher:
> We also need to think about transaction boundaries: Should the second round of flush() (that’s what we want this event for, right?) happen within the current transaction?

For my use-case ideally everything should be transactional as a whole. But I guess I could also work around this myself by wrapping everything in a transaction inside the listener or so.

mpdude:
Closing for the time being - the current idea is to make it possible to call `flush()` also from event listeners. See #10982.

PR #10878, opened 2023-08-03
Turn identity map collisions from exception to deprecation notice

### LINKED ISSUE #10871: Add a feature flag for the exception added in #10785
... to follow up on https://github.com/doctrine/orm/pull/10785#issuecomment-1663606294.

mpdude:
In #10785, a check was added that prevents entity instances from getting into the identity map when another object for the same ID is already being tracked.

This caused regressions for users that work with application-provided IDs and expect this condition to fail with `UniqueConstraintViolationExceptions` when flushing to the database.

Thus, this PR turns the exception into a deprecation notice. Users can opt-in to the new behavior. In 3.0, the exception will be used.

Implements #10871.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,19 @@
 # Upgrade to 2.16
 
+## Deprecated accepting duplicate IDs in the identity map
+
+For any given entity class and ID value, there should be only one object instance
+representing the entity. 
+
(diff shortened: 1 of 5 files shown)


derrabus (approved):

mpdude:
Will we need a switch for this in `DoctrineBundle`?

derrabus:
Yes.

PR #10879, opened 2023-08-04
Add null to AbstractQuery::getSingleScalarResult()'s return type

jlherren:
Single scalar results can very well also be the NULL value.


derrabus:
Duplicate of #10870

Issue #10872, opened 2023-08-03
Use a dedicated exception for the check added in #10785


mpdude:
... to follow up on https://github.com/doctrine/orm/pull/10785#issuecomment-1663606294.

gbirke:
I'm very much looking forward to this :smile: 
I suggest creating the exception in the `Doctrine\ORM` namespace and not reusing the one from DBAL, because in other places of the code (i.e. `flush`), @Ocramius mentioned he doesn't like the underlying layer bleeding into the ORM layer, see https://github.com/doctrine/orm/issues/7780#issuecomment-517436821

mpdude:
Implemented in #10881

PR #10881, opened 2023-08-04
Use a dedicated exception for the check added in #10785

mpdude:
This adds a dedicated exception for the case that objects with colliding identities are to be put into the identity map.

Implements #10872.

### DIFF
--- a/lib/Doctrine/ORM/Exception/EntityIdentityCollisionException.php
+++ b/lib/Doctrine/ORM/Exception/EntityIdentityCollisionException.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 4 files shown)


mpdude:
I thought it might be worthwhile to extend from `ORMException`, but strictly speaking, that makes it a BC break because the generic `RuntimeException` was used previously.

I am open for all suggestions regarding naming or other exception base classes.

mpdude:
@gbirke FYI

gbirke (approved):
I like the change, esp. with the `throws` annotation. BC Break (No RuntimeException thrown) won’t be a problem for me

derrabus (approved):

PR #10884, opened 2023-08-05
Fix broken changeset computation for entities loaded through fetch=EAGER + using inheritance

### LINKED ISSUE #10880: `EntityManager::find()` calls `UnitOfWork::registerManaged()` twice and that causes Change Detection to fail
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.16.0

#### Summary

I have a Symfony 5.4 based Application that uses Framework Extra Bundle and its Param Converter feature. After upgrading from 2.15.3 to 2.16.0 Doctrine has stopped detecting Entity Changes in Entities loaded by the ParamConverter, due to the new logic introduced by this commit: https://github.com/doctrine/orm/commit/01a14327d2404885897961851659b467ec966d74

After some snooping I've found out, that the `UnitOfWork::registerManaged()` is actually called twice:
- from the line 2805: https://github.com/doctrine/orm/blame/2.16.x/lib/Doctrine/ORM/UnitOfWork.php#L2875

- from the line 3038: https://github.com/doctrine/orm/blame/2.16.x/lib/Doctrine/ORM/UnitOfWork.php#L3038

The second call rewrites previously loaded/hydrated Entity and sets the `UnitOfWork::$originalEntityData` property to `[]`. This causes the change detection to fail, because the UoW is not able to correctly compare the Actual Data to the Original Data.

The Framework Extra Bundle uses the `EntityManager::find()` method under the hood, so I wouldn't blame the lib for the failure. The fact is that Doctrine has altered it's behavior significantly because of that change.

#### Previous behavior

when an Entity wired by the ParamConverter was changed, Doctrine has correctly detected changes made in the Entity.

#### Current behavior

Doctrine doesn't detect changes in the Entities wired by the ParamConverter.

#### How to reproduce

1. Create a Symfony 5.4 Application with Framework Extra Bundle
2. Autowire an Entity to a Controller
3. change the Entitie's state and save
4. change detection fails

I am not sure if this is the case for every such configuration, but I am positive that the `UnitOfWork::registerManaged` method can be called twice for a single `createEntity` method call.

#### Suggested Fix

Any tampering with not calling the `UnitOfWork::registerManaged()` a second time has caused the ORM to blow up in some other place. However, modifying the `UnitOfWork::registerManaged()` method itself to not allow to overwrite the Entity seems to have helped:

```
   public function registerManaged($entity, array $id, array $data)
    {
        $oid = spl_object_id($entity);
        //don't allow to overwrite Entity once it has been registered
        if(array_key_exists($oid, $this->entityIdentifiers)) {
            return;
        }
        ....
```

P.S.

This is not the only regression caused by this change. I can also see multiple tickets that point to this change as a culprit. Don't mean to tell you how to do things, but this seems like a very impactful and drastic change for a minor release.

mpdude:
#10880 reports a case where the changes from #10785 cause entity updates to be missed.

Upon closer inspection, this change seems to be causing it:

https://github.com/doctrine/orm/pull/10785/files#diff-55a900494fc8033ab498c53929716caf0aa39d6bdd7058e7d256787a24412ee4L2990-L3003

The code was changed to use `registerManaged()` instead, which basically does the same things, but (since #10785) also includes an additional check against duplicate entity instances.

But, one detail slipped through tests and reviews: `registerManaged()` also updates `\Doctrine\ORM\UnitOfWork::$originalEntityData`, which is used to compute entity changesets. An empty array `[]` was passed for $data here.

This will make the changeset computation assume that a partial object was loaded and effectively ignore all field updates here:

https://github.com/doctrine/orm/blob/a616914887ea160db4158d2c67752e99624f7c8a/lib/Doctrine/ORM/UnitOfWork.php#L762-L764

I think that, effectively, it is sufficient to call `registerManaged()` only in the two cases where a proxy was created.

Calling `registerManaged()` with `[]` as data for a proxy object is consistent with e. g. `\Doctrine\ORM\EntityManager::getReference()`.

In the case that a full entity has to be loaded, we need not call `registerManaged()` at all, since that will already happen inside `EntityManager::find()` (or, more specifically, `UnitOfWork::createEntity()` called inside it).

Note that the test case has to make some provisions so that we actually reach this case:
* Load an entity that uses `fetch="EAGER"` on a to-one association
* That association being against a class that uses inheritance (why's that?)

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -3048,6 +3048,7 @@ EXCEPTION
                                 // We are negating the condition here. Other cases will assume it is valid!
                                 case $hints['fetchMode'][$class->name][$field] !== ClassMetadata::FETCH_EAGER:
                                     $newValue = $this->em->getProxyFactory()->getProxy($assoc['targetEntity'], $normalizedAssociatedId);
+                                    $this->registerManaged($newValue, $associatedId, []);
                                     break;
 
                                 // Deferred eager load only works for single identifier classes
@@ -3056,6 +3057,7 @@ EXCEPTION
                                     $this->eagerLoadingEntities[$targetClass->rootEntityName][$relatedIdHash] = current($normalizedAssociatedId);
 
                                     $newValue = $this->em->getProxyFactory()->getProxy($assoc['targetEntity'], $normalizedAssociatedId);
+                                    $this->registerManaged($newValue, $associatedId, []);
                                     break;
 
                                 default:
@@ -3063,13 +3065,6 @@ EXCEPTION
                                     $newValue = $this->em->find($assoc['targetEntity'], $normalizedAssociatedId);
                                     break;
                             }
-
-                            if ($newValue === null) {
-                                break;
-                            }
-
-                            $this->registerManaged($newValue, $associatedId, []);
-                            break;
                     }
 
                     $this->originalEntityData[$oid][$field] = $newValue;
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10880Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10880Test.php
@@ -1,0 +1,119 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+use function reset;
+
+class GH10880Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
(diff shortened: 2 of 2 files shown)


mpdude:
@eXsio does this change fix it for you?

eXsio:
@mpdude yup, that does it, the tests are green :)

eXsio:
One thing I am wondering ( as a total ignorant, so I may be completely off here :) ) but is there a valid scenario when the `registerManaged()` is called twice for a single Entity instance? Shouldn't there be a validation or a warning (inside of that method) if such thing happen? Breaking the Change Detection seems to me like a worst error that can possibly be in any ORM, because it breaks the contract and has potential of effectively poisoning the client's data. There is no error, there is **"only"** data inconsistency. If not for our single behat test that failed, we would've been happy to release it into prod and acted only when our clients discovered data errors on their side.

Just a thought.

derrabus:
I've fixed the PHPCS failure for you. ✌🏻

derrabus (approved):

PR #10893, opened 2023-08-09
ProxyFactory: do not generate long file names

JanTvrdik:
This is a follow up to https://github.com/doctrine/common/pull/1005, because the logic was moved from doctrine/common to doctrine/orm in 2.17.

---

The maximum file name length when using [eCryptfs](https://www.ecryptfs.org/) is 143 bytes. This causes issues with Doctrine proxies where with current implementation this limit can be exceeded.

This PR changes how proxy file name is generated to avoid hitting this limit.

> eCryptfs can only store filenames of up to 143 characters when filename encryption is enabled. The remaining 112 characters are used for storing metadata such as the encrypted filename prefix, the signature of the filename encryption key, and an identifier for the cipher used, as well as random bytes to make /foo/file and /bar/file encrypt to different ciphertext and padding bytes to achieve proper cipher block size alignment.
> https://bugs.launchpad.net/ecryptfs/+bug/344878

### DIFF
--- a/lib/Doctrine/ORM/Proxy/ProxyFactory.php
+++ b/lib/Doctrine/ORM/Proxy/ProxyFactory.php
@@ -432,8 +432,8 @@ EOPHP;
     {
         $baseDirectory = $baseDirectory ?: $this->proxyDir;
 
-        return rtrim($baseDirectory, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . InternalProxy::MARKER
-            . str_replace('\\', '', $className) . '.php';
+        return rtrim($baseDirectory, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . InternalProxy::MARKER . DIRECTORY_SEPARATOR
+            . str_replace('\\', DIRECTORY_SEPARATOR, $className) . '.php';
     }
 
     private function getProxyFactory(string $className): Closure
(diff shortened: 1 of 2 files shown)


derrabus:
Thank you. Can you please have a look at the failing test? It looks related.

Furthermore, I think we should document this change in the UPGRADE.md file.

derrabus:
Also, please have a look at the `Doctrine\ORM\Proxy\Autoloader` class. It's deprecated and we discourage its usage, but as long as the class exists, we must not break it. I think that after your change, the generated proxies are written to a different location than this autoloader would expect them to be.

Thanks for the hint, @nicolas-grekas!

JanTvrdik:
I'm not sure how to solve the `Doctrine\ORM\Proxy\Autoloader` behavior without at least partially reverting #10837. The `Autoloader` is a static class with behavior that has to be kept in sync both with doctrine/common and doctrine/orm. As long as the `Autoloader` exists in common, the file name logic cannot be safely moved to ORM. 

I updated [the original PR in doctrine/common](https://github.com/doctrine/common/pull/1005) to change the `Autoload` behavior as well.

derrabus:
That's difficult indeed. So maybe, we won't get around changing doctrine/common after all.

We could release a new version of doctrine/common with the new file name mapping and then have ORM 2.17 depend on that new release.

PR #10895, opened 2023-08-09
PHPStan 1.10.28, Psalm 5.14.1

derrabus:


derrabus:
~~Apparently, Psalm does not infer the return type of `str_replace(string, string, string)` as string anymore and wants us to handle possible array return values. I've baselined all errors caused by that issue.~~

edit: I could fix all of those errors.

PR #10906, opened 2023-08-10
Mitigate problems with `EntityManager::flush()` reentrance since 2.16.0

mpdude:
This PR addresses the issue brought up in #10869. It happens when users use event listeners – `postPersist` in particular – to make changes to entities and/or persist new entities and finally call `EntityManager::flush()`, while the `UnitOfWork` is currently within the commit phase.

There is a discussion in #10900 whether this kind of reentrance should be deprecated in 2.x and prevented in 3.x. But, in order to prevent complete breakage with the 2.16.0 update, this PR tries to apply a band-aid 🩹.

A few things changed inside the UoW commit implementation in 2.16, and for sure this PR does not restore all details of the previous behavior. Take it with a grain of salt.

Here's the details.

The issue appears when `UoW::commit()` is performing entity insertions, and `postPersist` listener causes `commit()` reentrance when there are pending insertions left.

In that situation, the "inner" (reentrant) call will start working through all changesets. Eventually, it finishes with all insertions being performed and `UoW::$entityInsertions` being empty.

The entity insertion order, an array computed at the beginning of `UnitOfWork::executeInserts()`, still contains entities that now have been processed already. This leads to the reported failure down the road. The mitigation is to check for this condition and skip such entities.

Before #10547 (pre-2.16), things worked a bit differently:

The UoW did not build a list of entities (objects) as the commit order, but a sequence of _classes_ to process.

For every entity _class_, it would find all entity instances in `UoW::$entityInsertions` and pass them to the persister in a batch. The persister, in turn, would execute the `INSERT` statements for _all_ those entities _before_ it dispatched the `postInsert` events.

That means when a `postInsert` listener was notified pre-2.16, the UoW was in a state where 1) all insertions for a particular class had been written to the database already and 2) the UoW had not yet gathered the next round of entities to process.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1165,31 +1165,12 @@ class UnitOfWork implements PropertyChangedListener
     private function executeInserts(): void
     {
         $entities = $this->computeInsertExecutionOrder();
+        $eventsToDispatch = [];
 
         foreach ($entities as $entity) {
-            $oid = spl_object_id($entity);
-
-            // Mitigation for GH-10869:
-            // Users may use postPersist and similar listeners to make entity updates and call
-            // EM::flush() -> UoW::commit() again, while a transaction is currently running.
-            // This "somehow" worked pre 2.16, although it was never officially endorsed and/or
-            // is disputed (and there is no guarantee that the UoW will be able to deal with this,
-            // does not lose updates etc.).
-            // https://github.com/doctrine/orm/pull/10900 is a discussion about deprecating this
-            // kind of reentrance and disallowing it in 3.0.
-            //
-            // However, to ease the pain somewhat for users in 2.16, this condition covers that
-            // a reentrant call into UoW::commit() may have processed pending insertions that we
-            // had in our computed insertion order. So, after the second (inner) commit() returned
-            // and the outer one continues, deal with the situation that entities are no longer in
-            // the set of pending insertions.
-            if (! isset($this->entityInsertions[$oid])) {
-                continue;
-            }
-
+            $oid       = spl_object_id($entity);
             $class     = $this->em->getClassMetadata(get_class($entity));
             $persister = $this->getEntityPersister($class->name);
-            $invoke    = $this->listenersInvoker->getSubscribedSystems($class, Events::postPersist);
 
             $persister->addInsert($entity);
 
@@ -1216,9 +1197,18 @@ class UnitOfWork implements PropertyChangedListener
                 $this->addToEntityIdentifiersAndEntityMap($class, $oid, $entity);
             }
 
+            $invoke = $this->listenersInvoker->getSubscribedSystems($class, Events::postPersist);
+
             if ($invoke !== ListenersInvoker::INVOKE_NONE) {
-                $this->listenersInvoker->invoke($class, Events::postPersist, $entity, new PostPersistEventArgs($entity, $this->em), $invoke);
+                $eventsToDispatch[] = ['class' => $class, 'entity' => $entity, 'invoke' => $invoke];
             }
+        }
+
+        // Mitigation for GH-10869:
+        // Defer dispatching `postPersist` events to until all entities have been inserted and there
+        // are no pending insertions left.
+        foreach ($eventsToDispatch as $event) {
+            $this->listenersInvoker->invoke($event['class'], Events::postPersist, $event['entity'], new PostPersistEventArgs($event['entity'], $this->em), $event['invoke']);
         }
     }
 
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -1167,7 +1167,26 @@ class UnitOfWork implements PropertyChangedListener
         $entities = $this->computeInsertExecutionOrder();
 
         foreach ($entities as $entity) {
-            $oid       = spl_object_id($entity);
+            $oid = spl_object_id($entity);
+
+            // Mitigation for GH-10869:
+            // Users may use postPersist and similar listeners to make entity updates and call
+            // EM::flush() -> UoW::commit() again, while a transaction is currently running.
+            // This "somehow" worked pre 2.16, although it was never officially endorsed and/or
+            // is disputed (and there is no guarantee that the UoW will be able to deal with this,
+            // does not lose updates etc.).
+            // https://github.com/doctrine/orm/pull/10900 is a discussion about deprecating this
+            // kind of reentrance and disallowing it in 3.0.
+            //
+            // However, to ease the pain somewhat for users in 2.16, this condition covers that
+            // a reentrant call into UoW::commit() may have processed pending insertions that we
+            // had in our computed insertion order. So, after the second (inner) commit() returned
+            // and the outer one continues, deal with the situation that entities are no longer in
+            // the set of pending insertions.
+            if (! isset($this->entityInsertions[$oid])) {
+                continue;
+            }
+
             $class     = $this->em->getClassMetadata(get_class($entity));
             $persister = $this->getEntityPersister($class->name);
             $invoke    = $this->listenersInvoker->getSubscribedSystems($class, Events::postPersist);
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10869Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10869Test.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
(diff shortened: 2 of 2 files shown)


dmaicher:
@mpdude this test fails for me on 2.16.x but works on 2.15.x

I added a few more entities that are persisted so I guess the flush during postPersist messes with the persisting of the other entities following afterwards?

```php
<?php

declare(strict_types=1);

use Doctrine\DBAL\Logging\LoggerChain;
use Doctrine\ORM\Event\PostPersistEventArgs;
use Doctrine\ORM\Events;
use Doctrine\Tests\DbalExtensions\SqlLogger;
use Doctrine\Tests\Models\CMS\CmsUser;
use function sprintf;

class GH10869Test extends OrmFunctionalTestCase
{
    {
        $this->useModelSet('cms');

        parent::setUp();
    }

    public function testPostPersistListenerUpdatingObject(): void
    {
        foreach (range(1, 3) as $i) {
            $joe           = new CmsUser();
            $joe->username = "joe-$i";
            $joe->name     = "Joe $i";
            $this->_em->persist($joe);
        }

        $this->_em->getEventManager()->addEventListener(Events::postPersist, new class {
            public function postPersist(PostPersistEventArgs $args): void
            {
                $object = $args->getObject();

                if (! $object instanceof CmsUser) {
                    return;
                }

                $objectManager = $args->getObjectManager();

                if ($object->status === null) {
                    $object->status = sprintf('joe-id=%s', $object->id);
                }

                $objectManager->flush();
            }
        });

        $this->_em->flush();
        $this->_em->clear();

        $joeReloaded = $this->_em->find(CmsUser::class, $joe->id);

        self::assertNotNull($joeReloaded->status);
        self::assertSame($joe->status, $joeReloaded->status);
    }
}
```

mpdude:
@dmaicher Please check and report if this fixes the issue for you

dmaicher:
@mpdude thanks for all your efforts on this :+1: 

The SQL error with missing parameters is indeed gone.

However I don't think the behavior is quite the same as on 2.15 :confused: Now some other test expectations are failing that seem related to the new entities that are persisted within `postPersist`. 

I think its too dangerous to flush again while another flush is ongoing. So instead of somehow fixing this I would definitely prefer a new event and do another clean flush.

here a test that fails with your changes but works on 2.15:

```php
<?php

declare(strict_types=1);

use Doctrine\ORM\Event\PostPersistEventArgs;
use Doctrine\ORM\Events;

use function sprintf;

class GH10869Test extends OrmFunctionalTestCase
{
    {
        parent::setUp();

        $this->setUpEntitySchema([
            GH10869Entity::class,
        ]);
    }

    public function testPostPersistListenerUpdatingObjectFieldWhileOtherInsertPending(): void
    {
        $entity1 = new GH10869Entity();
        $this->_em->persist($entity1);

        $entity2 = new GH10869Entity();
        $this->_em->persist($entity2);

        $this->_em->getEventManager()->addEventListener(Events::postPersist, new class {
            public function postPersist(PostPersistEventArgs $args): void
            {
                $object = $args->getObject();

                if ($object->field === 'foo') {
                    return;
                }

                $objectManager = $args->getObjectManager();
                $object->field = sprintf('test %s', $object->id);

                foreach (range(1, 2) as $ignored) {
                    $newObject = new GH10869Entity();
                    $newObject->field = 'foo';
                    $objectManager->persist($newObject);
                }

                $objectManager->flush();
            }
        });

        $this->_em->flush();
        $this->_em->clear();

        $entity1Reloaded = $this->_em->find(GH10869Entity::class, $entity1->id);
        self::assertSame($entity1->field, $entity1Reloaded->field);

        $entity2Reloaded = $this->_em->find(GH10869Entity::class, $entity2->id);
        self::assertSame($entity2->field, $entity2Reloaded->field);

        $objectsPersistedDuringPostPersist = $this->_em->getRepository(GH10869Entity::class)->count(['field' => 'foo']);
        self::assertSame(4, $objectsPersistedDuringPostPersist);
    }
}

/**
 * @ORM\Entity
 */
class GH10869Entity
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     *
     * @var ?int
     */
    public $id;

    /**
     * @ORM\Column(type="text", nullable=true)
     *
     * @var ?string
     */
    public $field;
}
```

mpdude:
What exactly is this difference?

dmaicher:
> What exactly is this difference?

you mean how my test case is different from yours?

```diff
--- a/tests/Doctrine/Tests/ORM/Functional/GH10869Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/GH10869Test.php
@@ -35,8 +35,19 @@ class GH10869Test extends OrmFunctionalTestCase
             {
                 $object = $args->getObject();
 
+                if ($object->field === 'foo') {
+                    return;
+                }
+
                 $objectManager = $args->getObjectManager();
                 $object->field = sprintf('test %s', $object->id);
+
+                foreach (range(1, 2) as $ignored) {
+                    $newObject = new GH10869Entity();
+                    $newObject->field = 'foo';
+                    $objectManager->persist($newObject);
+                }
+
                 $objectManager->flush();
             }
         });
@@ -49,6 +60,9 @@ class GH10869Test extends OrmFunctionalTestCase
 
         $entity2Reloaded = $this->_em->find(GH10869Entity::class, $entity2->id);
         self::assertSame($entity2->field, $entity2Reloaded->field);
+
+        $objectsPersistedDuringPostPersist = $this->_em->getRepository(GH10869Entity::class)->count(['field' => 'foo']);
+        self::assertSame(4, $objectsPersistedDuringPostPersist);
     }
``` 

the last assert basically fails. On your branch its 6 and on 2.15 its 4 which would be correct.

mpdude:
No, I mean what exactly the difference is between 2.15 and 2.16... does it miss updates? Is only the ordering different?

dmaicher:
> No, I mean what exactly the difference is between 2.15 and 2.16... does it miss updates? Is only the ordering different?

Hard to tell so far. It seems some new objects (added during postPersist) are not even persisted at all or so. I will check in more detail.

dmaicher:
@mpdude ok after further checking it has to do with the changed order of entity insertion.

In the postPersist listener I expect some related entity to be already persisted (written to DB and assigned ID) and only then I can properly create the new entity.

So I think I need a new event :yum: Too brittle to make (false as we see now) assumptions about the order in which entities are persisted.

mpdude:
Yes, you were assuming that all other entities of the same class have already been written before the event is dispatched.

Maybe we can move the `postInsert` event a bit back – I'll have to think this through.

dmaicher:
> Yes, you were assuming that all other entities of the same class have already been written before the event is dispatched.

actually in my case its not even about entities of the same class :confused: Its about another related class. The structure is a bit more complex.

Its like


```
    A
  ´   `
B       C
``` 

I listen to `postPersist` for `C` and expect that `B` is also already persisted. Which luckily was the case on 2.15 but now not anymore.

mpdude:
https://github.com/doctrine/orm/blob/d2de4ec03c63ddd7bdfda0421e237e11343c75ee/lib/Doctrine/ORM/UnitOfWork.php#L1215-L1217

This is the 2.15 version of `UnitOfWork::executeInserts()`, and it moved all the events to the very end of the method. If that would be possible in 2.16 as well, it might cover this case. Plus, we probably don't need the patch suggested in this PR.

mpdude:
@dmaicher Would this approach b66c3b4a1d82143ce5491eb224d85acad3c049d0 work for your use case?

mpdude:
Regarding https://github.com/doctrine/orm/pull/10906#issuecomment-1682366559 – I cannot reproduce that. For me, it seems to pass also based on 1895cf8167bd4b4f01ee50577f4350d7a692c460

dmaicher:
> @dmaicher Would this approach [b66c3b4](https://github.com/doctrine/orm/commit/b66c3b4a1d82143ce5491eb224d85acad3c049d0) work for your use case?

@mpdude this works indeed :tada: Let me try to run the full testsuite (after fixing some tests that reply on Ids and the order of persisted entites). 

EDIT: all tests green now on my end! :blush:

derrabus:
PHPCS fails. 🤓

mpdude:
Converting to draft for the time being. Will open a new PR later since the solution we landed at is somewhat different from what I initially proposed. To keep the PR history understandable, I think it would be better to move the "new" solution to a dedicated discussion.

Extra things I'd like to include:
* Check/fix that also `postRemove` (and `postUpdate`) should have event dispatching deferred in a similar way
* Include test to check that `postPersist` is called at a time where _all_ entities have been persisted and their IDs assigned, but not necessarily all extra updates and collections have been processed. Check with two entities of different classes cross-referencing each other.
* Update docs and explain this for the `postPersist` event
* Optionally: State that calling flush() from event listeners is discouraged (maybe dedicated PR?)

mpdude:
Closing in favor of #10915.

PR #10909, opened 2023-08-15
Continue merge up

greg0ire:
I want to check if the imerge steps also pass the build when running on Github with more jobs.

### DIFF
--- a/lib/Doctrine/ORM/Proxy/InternalProxy.php
+++ b/lib/Doctrine/ORM/Proxy/InternalProxy.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Proxy;
+
+use Doctrine\Persistence\Proxy;
+
+/**
+ * @internal
+ *
+ * @template T of object
+ * @template-extends Proxy<T>
+ *
+ * @method void __setInitialized(bool $initialized)
+ */
+interface InternalProxy extends Proxy
+{
+}
--- a/docs/en/reference/unitofwork.rst
+++ b/docs/en/reference/unitofwork.rst
@@ -37,8 +37,8 @@ will still end up with the same reference:
     public function testIdentityMapReference(): void
     {
         $objectA = $this->entityManager->getReference('EntityName', 1);
-        // check for proxyinterface
-        $this->assertInstanceOf('Doctrine\Persistence\Proxy', $objectA);
+        // check entity is not initialized
+        $this->assertTrue($this->entityManager->isUninitializedObject($objectA));
 
         $objectB = $this->entityManager->find('EntityName', 1);
 
--- a/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
+++ b/lib/Doctrine/ORM/Cache/DefaultQueryCache.php
@@ -16,7 +16,6 @@ use Doctrine\ORM\PersistentCollection;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\Query\ResultSetMapping;
 use Doctrine\ORM\UnitOfWork;
-use Doctrine\Persistence\Proxy;
 
 use function array_map;
 use function array_shift;
@@ -345,7 +344,7 @@ class DefaultQueryCache implements QueryCache
             $assocIdentifier = $this->uow->getEntityIdentifier($assocValue);
             $entityKey       = new EntityCacheKey($assocMetadata->rootEntityName, $assocIdentifier);
 
-            if (! $assocValue instanceof Proxy && ($key->cacheMode & Cache::MODE_REFRESH) || ! $assocRegion->contains($entityKey)) {
+            if (! $this->uow->isUninitializedObject($assocValue) && ($key->cacheMode & Cache::MODE_REFRESH) || ! $assocRegion->contains($entityKey)) {
                 // Entity put fail
                 if (! $assocPersister->storeEntityCache($assocValue, $entityKey)) {
                     return null;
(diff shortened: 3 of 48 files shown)


greg0ire:
:thinking: I don't reproduce a single error in this build locally…

greg0ire:
Ah, I do reproduce them when using `ORM_PROXY_IMPLEMENTATION=common`… which does seem super related :stuck_out_tongue: 

UPD: I notice the first failing test seems to be dependent on the previous test. ORM 2 uses PHPUnit 9, while ORM 3 uses PHPUnit 10. ~I feel this is going to be the reason for this difference in behavior.~ Downgrading to PHPUnit 9 does not fix the issue. Merging both tests together makes the issue disappear. Using `#[Depends]` does not.

https://github.com/doctrine/orm/blob/5577d51c447172ff8fdb81433277e625f4945cfe/tests/Doctrine/Tests/ORM/Cache/DefaultEntityHydratorTest.php#L41-L69

lib/Doctrine/ORM/Proxy/InternalProxy.php
```diff
@@ -0,0 +1,19 @@
...
+ * @internal
+ *
+ * @template T of object
+ * @template-extends Proxy<T>
+ *
+ * @method void __setInitialized(bool $initialized)
+ */
+interface InternalProxy extends Proxy
+{
+}
```
greg0ire, line 19: In #10832, this interface was introduced by the `Proxy` interface in the same namespace. Now, that interface no longer exists (since #10551). I'm not sure how to resolve that conflict.

greg0ire:
@nicolas-grekas @derrabus can you please look into this one? :pray:

nicolas-grekas:
Sure, ASAP!

greg0ire:
In the first failing test, it looks like Doctrine interacts with the DBAL at some point, performing an SQL query, when it really should not: the test does not create an entity in the database, it merely creates entities in the cache, and I think what's expected is that the entity is fetched from the cache. I think we expect the closure inside `ProxyFactory::createInitializer` to exit after checking that the proxy is initialized (it does not): https://github.com/doctrine/orm/blob/62ed63bbbe3f59dcacb7dadb574895f5cb27a257/lib/Doctrine/ORM/Proxy/ProxyFactory.php#L194-L196

The test line that causes the initialization on 3.0.x is this one: https://github.com/doctrine/orm/blob/62ed63bbbe3f59dcacb7dadb574895f5cb27a257/tests/Doctrine/Tests/ORM/Cache/DefaultEntityHydratorTest.php#L61

It calls that one: https://github.com/doctrine/orm/blob/62ed63bbbe3f59dcacb7dadb574895f5cb27a257/lib/Doctrine/ORM/Cache/DefaultEntityHydrator.php#L187

This is caused by a change in a UnitOfWork code block from 

from

```php
            if ($entity instanceof Proxy && ! $entity->__isInitialized()) {
                $entity->__setInitialized(true);
            } else {
                if (
                    ! isset($hints[Query::HINT_REFRESH])
                    || (isset($hints[Query::HINT_REFRESH_ENTITY]) && $hints[Query::HINT_REFRESH_ENTITY] !== $entity)
                ) {
                    return $entity;
                }
            }

```

to 

```php
            if ($this->isUninitializedObject($entity)) {
                $entity->__setInitialized(true);
            } else {
                if (
                    ! isset($hints[Query::HINT_REFRESH])
                    || (isset($hints[Query::HINT_REFRESH_ENTITY]) && $hints[Query::HINT_REFRESH_ENTITY] !== $entity)
                ) {
                    return $entity;
                }
            }
```

On 3.0.x, the `if` branch is used, while on `imerge/continue-merge-up`, the `else` branch is used.

greg0ire:
Unsurprisingly, the issue is caused by my unresolved comment: the proxy does not implement `InternalProxy`, which causes `isUninitializedObject` to return `false`.

greg0ire:
@derrabus @nicolas-grekas I pushed a commit with a very ugly fix, please have a look at it and let me know whether you think I should introduce an interface (`InternalCommonProxy` ?) that would extend `InternalProxy` and the `Proxy` from `doctrine/common`.

stof:
@greg0ire shouldn't proxy from Common not be used anymore in the 3.0.x branch ?

greg0ire:
@stof I think so, but I try to keep merges up diffs as small as possible. The cleanup should IMO happen in a separate PR.

stof:
@greg0ire the conflict comes from the fact that the old proxy logic is **already** removed from the 3.0.x (based on the version of the codebase before the refactoring you are trying to merge up). So you cannot delay the cleanup here. You need to solve the conflicts between the _past_ cleanup and the refactoring.

greg0ire:
> the conflict comes from the fact that the old proxy logic is already removed from the 3.0.x

@stof really? I don't think I truly understand what you mean by "old proxy logic" because the proxy logic relying on `doctrine/common` seems very much still in use on 3.0.x: 

https://github.com/doctrine/orm/blob/9a1085d72fbf6bbcc637dbc184ebfb4f3b4f8eff/.github/workflows/continuous-integration.yml#L42-L43

https://github.com/doctrine/orm/blob/9a1085d72fbf6bbcc637dbc184ebfb4f3b4f8eff/tests/Doctrine/Tests/TestUtil.php#L94-L112

https://github.com/doctrine/orm/blob/9a1085d72fbf6bbcc637dbc184ebfb4f3b4f8eff/composer.json#L27

Can you please clarify what you are referring to?

greg0ire:
> let me know whether you think I should introduce an interface (InternalCommonProxy ?) that would extend InternalProxy and the Proxy from doctrine/common.

As pointed out by @stof , this is all going to disappear soon, so let's not bother with a short-lived interface. This addresses my only concern about this, so let's mark this as ready for review.

Regarding the docs job, it should get fixed in a subsequent merge up.

lib/Doctrine/ORM/Proxy/InternalProxy.php
greg0ire: I think the solution is to have the proxy classes also implement this interface. That's what I did in my second commit anyway.

greg0ire:
@derrabus @SenseException can I get an approval on this? I'd like to continue and possibly finish the merge up before the hackathon.

derrabus (approved):

derrabus:
By all means, please continue.

PR #10910, opened 2023-08-15
Remove PHPUnit polyfills

greg0ire:
We are using PHPUnit 10, so none of this is needed now.


derrabus (changes requested):
Do we sill need the empty class?

greg0ire:
It was empty when introduced in 22e94ac594ae5b9b851140eae4b3ee553efb814d, and stayed that way for a bit more than a decade, until 1f4e6ebeeb4efeefcb28053145af395da779b7b7 . But I suppose we can remove it.

greg0ire:
Plus it looks like it wasn't systematically extended.

derrabus (approved):

Issue #10864, opened 2023-08-01
Creation order parent-children inversed


GregoireHebert:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.16.0

#### Summary

Record order creation has changed.
Probably following this : https://github.com/doctrine/orm/pull/10547

#### Previous behavior

Creation of an entity an its child. 

```json
{
      "_links": {
        "self": {
          "href": "/max_depth_eager_dummies/1"
        },
        "child": {
          "href": "/max_depth_eager_dummies/2"
        }
      },
      "_embedded": {
        "child": {
          "_links": {
            "self": {
              "href": "/max_depth_eager_dummies/2"
            }
          },
          "id": 2,
          "name": "level 2"
        }
      },
      "id": 1,
      "name": "level 1"
    }
```

#### Current behavior

Now the child is created before the parent.

```json
        {
          "_links": {
              "self": {
                  "href": "/max_depth_eager_dummies/2"
              },
              "child": {
                  "href": "/max_depth_eager_dummies/1"
              }
          },
          "_embedded": {
              "child": {
                  "_links": {
                      "self": {
                          "href": "/max_depth_eager_dummies/1"
                      }
                  },
                  "id": 1,
                  "name": "level 2"
              }
          },
          "id": 2,
          "name": "level 1"
      }
```

#### How to reproduce

Get the last version of API Platform and run this scenario: 
https://github.com/api-platform/core/actions/runs/5727605051/job/15520391811?pr=5675#step:9:94

Downgrading to 2.15.5 fixes the problem.

mpdude:
Can you show me the code where the entities are defined, created and persisted?

greg0ire:
Can you please state the obvious and say why this should be considered a breaking change in your opinion? Right now the only difference I can spot on my phone is a difference in ids. A diff might help.

dunglas:
@mpdude the entity definition: https://github.com/api-platform/core/blob/3.1/tests/Fixtures/TestBundle/Entity/MaxDepthEagerDummy.php
In this test, persistence is handled "automatically" by the Doctrine bridge of API Platform: https://github.com/api-platform/core/blob/3.1/src/Doctrine/Common/State/PersistProcessor.php#L102-L107

[...]

PR #10912, opened 2023-08-16
(Try to) add a reproducer for a delete ordering issue

mpdude:
In https://github.com/doctrine/orm/issues/10864#issuecomment-1676405188 and subsequent comments, @Dallas62 reported that since the upgrade to 2.16.1, they have an issue with the order of `DELETE` statements. The order of `DELETES` leads to a foreign key constraint violation.

Since the report does not exactly match the original issue topic in #10864, I am creating this one here so we can track it individually.

In order to reproduce, a demo Symfony application was provided in https://github.com/Dallas62/doctrine-orm-reproducer-deletion-issue, but obviously we cannot include that in the ORM test suite.

I have picked the two entities I deemed relevant, but as of now, this is not sufficient to reproduce the bug.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10912Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10912Test.php
@@ -21,18 +21,25 @@ class GH10912Test extends OrmFunctionalTestCase
 
         $this->setUpEntitySchema([
             GH10912User::class,
+            GH10912Profile::class,
             GH10912Room::class,
         ]);
     }
 
     public function testIssue(): void
     {
-        $user = new GH10912User();
-        $room = new GH10912Room($user);
+        $user    = new GH10912User();
+        $profile = new GH10912Profile();
+        $room    = new GH10912Room();
+
         $user->rooms->add($room);
+        $user->profile = $profile;
+        $profile->user = $user;
+        $room->user    = $user;
 
         $this->_em->persist($room);
         $this->_em->persist($user);
+        $this->_em->persist($profile);
         $this->_em->flush();
         $this->_em->clear();
 
@@ -48,9 +55,16 @@ class GH10912Test extends OrmFunctionalTestCase
             return strpos($entry['sql'], 'DELETE') === 0;
         }));
 
-        self::assertCount(2, $queries);
+        self::assertCount(3, $queries);
         self::assertSame('DELETE FROM GH10912Room WHERE id = ?', $queries[0]['sql']);
-        self::assertSame('DELETE FROM GH10912User WHERE id = ?', $queries[1]['sql']);
+        self::assertSame('DELETE FROM GH10912Profile WHERE id = ?', $queries[1]['sql']);
+        self::assertSame('DELETE FROM GH10912User WHERE id = ?', $queries[2]['sql']);
+
+        // The EntityManager is aware that all three entities have been deleted
+        $im = $this->_em->getUnitOfWork()->getIdentityMap();
+        self::assertEmpty($im[GH10912Profile::class]);
+        self::assertEmpty($im[GH10912User::class]);
+        self::assertEmpty($im[GH10912Room::class]);
     }
 }
 
@@ -73,10 +87,39 @@ class GH10912User
      */
     public $rooms;
 
+    /**
+     * @ORM\OneToOne(targetEntity=GH10912Profile::class, cascade={"remove"})
+     * @ORM\JoinColumn(onDelete="cascade")
+     *
+     * @var GH10912Profile
+     */
+    public $profile;
+
     public function __construct()
     {
         $this->rooms = new ArrayCollection();
     }
+}
+
+/** @ORM\Entity */
+class GH10912Profile
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue
+     *
+     * @var int
+     */
+    public $id;
+
+    /**
+     * @ORM\OneToOne(targetEntity=GH10912User::class)
+     * @ORM\JoinColumn(onDelete="cascade")
+     *
+     * @var GH10912User
+     */
+    public $user;
 }
 
 /** @ORM\Entity */
@@ -92,15 +135,10 @@ class GH10912Room
     public $id;
 
     /**
-     * @ORM\JoinColumn(nullable=false)
      * @ORM\ManyToOne(targetEntity=GH10912User::class, inversedBy="rooms")
+     * @ORM\JoinColumn(nullable=false)
      *
      * @var GH10912User
      */
     public $user;
-
-    public function __construct(GH10912User $user)
-    {
-        $this->user = $user;
-    }
 }
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10912Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH10912Test.php
@@ -1,0 +1,106 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+use function array_filter;
+use function array_values;
+use function strpos;
+
+class GH10912Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH10912User::class,
+            GH10912Room::class,
+        ]);
+    }
+
+    public function testIssue(): void
+    {
+        $user = new GH10912User();
+        $room = new GH10912Room($user);
+        $user->rooms->add($room);
+
+        $this->_em->persist($room);
+        $this->_em->persist($user);
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $userReloaded = $this->_em->find(GH10912User::class, $user->id);
+
+        $queryLog = $this->getQueryLog();
+        $queryLog->reset()->enable();
+
+        $this->_em->remove($userReloaded);
+        $this->_em->flush();
+
+        $queries = array_values(array_filter($queryLog->queries, static function ($entry) {
+            return strpos($entry['sql'], 'DELETE') === 0;
+        }));
+
+        self::assertCount(2, $queries);
+        self::assertSame('DELETE FROM GH10912Room WHERE id = ?', $queries[0]['sql']);
+        self::assertSame('DELETE FROM GH10912User WHERE id = ?', $queries[1]['sql']);
+    }
+}
+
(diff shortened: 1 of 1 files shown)


mpdude:
@Dallas62 I think what you're seeing is different from #10864. So let's continue the discussion here.

I have tried to pick the relevant classes from your repo and strip it down to the bare minimum needed to run as a test. But, as you can see, this test currently passes.

So, there must be more to it...

Can you experiment with the test as it can be found in https://github.com/mpdude/doctrine2/tree/delete-foreign-key-issue, and feel free to open PRs against it? We need a failing test to proceed, and it has to be a test that we can include and run in the ORM test suite.

Thanks!

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10912Test.php
```diff
@@ -0,0 +1,106 @@
...
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue
+     *
+     * @var int
+     */
+    public $id;
+
+    /**
+     * @ORM\OneToMany(targetEntity=GH10912Room::class, mappedBy="user", cascade={"remove"})
```
mpdude, line 84: If I remove `cascade: remove`, I'll get the error

> Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException : An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`doctrine_tests`.`GH10912Room`, CONSTRAINT `FK_7FCB83EEA76ED395` FOREIGN KEY (`user_id`) REFERENCES `GH10912User` (`id`))

This is because we have a `GH10912Room` pointing to the user that is to be removed, the association is not nullable and we have no database-level `ON DELETE` setting.

TIL that `cascade` seemingly also works on the inverse side of an association (@derrabus )

mpdude: Adding `orphanRemoval`, as can be seen in

https://github.com/Dallas62/doctrine-orm-reproducer-deletion-issue/blob/6e14b758be3484273ddddd2366f465a3d2eead30/src/Entity/User.php#L33

does not make difference.

Dallas62:
Thanks @mpdude for your time, I will take a look later this day. I found during making the reproducer, it only happen when the `OneToOne` relationship between `User` and `UserProfile` is added. In this particular structure I could have merge both tables but there might be some other impact in the orm.

mpdude:
I have added the `Profile` entity and the two `OneToOne` associations between `User` and `Profile` (side note: that's two individual, unidirectional associations). Seems it still works.

@Dallas62 looking forward to your suggestions!

Dallas62:
Ok, so I have continue digging on this subject.
- I tried to set all classes on your test case => No change.
- I tried to reproduce it by switching the pdo_mysql to pdo_sqlite in_memory on the reproducer repository => The issue is not present.
So the driver seems to be important, not sure how to mock it without changing the driver.

Next:
- I will try to reduce the number of entities in the reproducer to have the minimal exemple.
- Then try to reproduce it on your test case (at least with pdo_mysql).

mpdude:
Not sure whether SQLite and/or in_memory implements referential integrity. Without that, the issue won't surface.

Dallas62:
I simplified the reproducer:
- Remove `#[HasLifecycleCallbacks]`
- Remove `Plant` and `PlantPicture`
- Use `integer` instead of `ulid` and default `GeneratedValue` configuration
- remove unnecessary arguments
Still able to reproduce, but not in this PR (with pdo_mysql) 🤔

Dallas62:
@mpdude  After more digging into this issue, I finally got the cause of the reproducibility ! 🎉

The `$this->_em->remove($userReloaded);` seems to push data inside `$unitOfWork->entityDeletions` based on the list of properties.

So:
- if `profile` is before `rooms`, it fails
- if `profile` is after `rooms`, everything is working fine.

Good news, working with pdo_sqlite in_memory.

Dallas62:
```patch
@@ -80,13 +80,6 @@ class GH10912User
      */
     public $id;
 
-    /**
-     * @ORM\OneToMany(targetEntity=GH10912Room::class, mappedBy="user", cascade={"remove"})
-     *
-     * @var Collection<int, GH10912Room>
-     */
-    public $rooms;
-
     /**
      * @ORM\OneToOne(targetEntity=GH10912Profile::class, cascade={"remove"})
      * @ORM\JoinColumn(onDelete="cascade")

@@ -95,6 +88,13 @@ class GH10912User
      */
     public $profile;
 
+    /**
+     * @ORM\OneToMany(targetEntity=GH10912Room::class, mappedBy="user", cascade={"remove"})
+     *
+     * @var Collection<int, GH10912Room>
+     */
+    public $rooms;
+
     public function __construct()
     {
         $this->rooms = new ArrayCollection();

```

mpdude:
X-ref #10566; maybe there’s more conditions we need to consider?

mpdude:
@Dallas62 it's still preliminary, but you might want to give #10913 a try.

Dallas62:
Thanks, I will take a look asap.

I will also extends your tests based on some particular cases I have in mind.

Dallas62:
I tried with the given repository (at commit https://github.com/Dallas62/doctrine-orm-reproducer-deletion-issue/commit/6e14b758be3484273ddddd2366f465a3d2eead30)
And with an extra `ManyToOne` on `Profile` entity, everything is working fine. The order of deletion is still based on order of property but in tested cases, this as no impact on the result.

For me it seems ok, let some more experimented contributors review changes and give feedbacks.

Just in case:

```php
<?php

declare(strict_types=1);

use function array_filter;
use function array_values;
use function strpos;

class GH10912Case2Test extends OrmFunctionalTestCase
{
    {
        parent::setUp();

        $this->setUpEntitySchema([
            GH10912User::class,
            GH10912Profile::class,
            GH10912Room::class,
            GH10912AnotherDependency::class
        ]);
    }

    public function testIssue(): void
    {
        $user       = new GH10912User();
        $profile    = new GH10912Profile();
        $room       = new GH10912Room();
        $dependency = new GH10912AnotherDependency();

        $user->rooms->add($room);
        $user->profile = $profile;
        $profile->user = $user;
        $room->user = $user;
        $dependency->profile = $profile;
        $profile->dependencies->add($dependency);

        $this->_em->persist($room);
        $this->_em->persist($user);
        $this->_em->persist($profile);
        $this->_em->persist($dependency);
        $this->_em->flush();
        $this->_em->clear();

        $userReloaded = $this->_em->find(GH10912User::class, $user->id);

        $queryLog = $this->getQueryLog();
        $queryLog->reset()->enable();

        $this->_em->remove($userReloaded);
        $this->_em->flush();

        $queries = array_values(array_filter($queryLog->queries, static function ($entry) {
          return strpos($entry['sql'], 'DELETE') === 0;
        }));

        self::assertCount(4, $queries);
        self::assertSame('DELETE FROM GH10912AnotherDependency WHERE id = ?', $queries[0]['sql']);
        self::assertSame('DELETE FROM GH10912Room WHERE id = ?', $queries[1]['sql']);
        self::assertSame('DELETE FROM GH10912User WHERE id = ?', $queries[2]['sql']);
        self::assertSame('DELETE FROM GH10912Profile WHERE id = ?', $queries[3]['sql']);

        // The EntityManager is aware that all three entities have been deleted
        $im = $this->_em->getUnitOfWork()->getIdentityMap();
        self::assertEmpty($im[GH10912Profile::class]);
        self::assertEmpty($im[GH10912User::class]);
        self::assertEmpty($im[GH10912Room::class]);
        self::assertEmpty($im[GH10912AnotherDependency::class]);
    }
}

/** @ORM\Entity */
class GH10912User
{
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue
     *
     * @var int
     */
    public $id;

    /**
     * @ORM\OneToOne(targetEntity=GH10912Profile::class, cascade={"remove"})
     * @ORM\JoinColumn(onDelete="cascade")
     *
     * @var GH10912Profile
     */
    public $profile;

    /**
     * @ORM\OneToMany(targetEntity=GH10912Room::class, mappedBy="user", cascade={"remove"})
     *
     * @var Collection<int, GH10912Room>
     */
    public $rooms;

    public function __construct()
    {
        $this->rooms = new ArrayCollection();
    }
}

/** @ORM\Entity */
class GH10912Profile
{
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue
     *
     * @var int
     */
    public $id;

    /**
     * @ORM\OneToOne(targetEntity=GH10912User::class)
     * @ORM\JoinColumn(onDelete="cascade")
     *
     * @var GH10912User
     */
    public $user;

    /**
     * @ORM\OneToMany(targetEntity=GH10912AnotherDependency::class, mappedBy="profile", cascade={"remove"})
     *
     * @var Collection<int, GH10912AnotherDependency>
     */
    public $dependencies;

    public function __construct()
    {
        $this->dependencies = new ArrayCollection();
    }
}

/** @ORM\Entity */
class GH10912Room
{
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue
     *
     * @var int
     */
    public $id;

    /**
     * @ORM\ManyToOne(targetEntity=GH10912User::class, inversedBy="rooms")
     *
     * @var GH10912User
     */
    public $user;
}


/** @ORM\Entity */
class GH10912AnotherDependency
{
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue
     *
     * @var int
     */
    public $id;

    /**
     * @ORM\ManyToOne(targetEntity=GH10912Profile::class, inversedBy="dependencies")
     *
     * @var GH10912Profile
     */
    public $profile;
}

```

asmafarhat2006:
@mpdude 
Im facing a similar issue where deleting an entity which has a child association triggers this error 

> "Cannot delete or update a parent row: a foreign key constraint fails"

 , in 2.16, and it used to work fine in 2.15
Adding an inverse relation on the entity seems to resolve the problem , but this solution wont work for a large codebase.
Am I looking at the right thread or is my issue slightly different. ?

eigan:
@asmafarhat2006 Might not be helpful, but we got the same problem in our codebase and I dug arround and found out why. Of course, there is multiple ways to trigger this same error.

```php
class  MyEntity {
   # ManyToOne
   public OtherFooEntity $foo;
}


# This will give you problem in 2.16 and not in earlier version (we came from 2.14):
$myEntity->foo = null;
$em->remove($myEntity->foo);
$em->remove($myEntity);

$em->flush(); // Crash

// Doctrine thinks its OK to remove OtherFooEntity before MyEntity
// because it looks at the current state and not the original and current state in db.

// Was easy for us to refactor this, as it didnt make much sense to set foo = null 
```

asmafarhat2006:
Thanks for looking into this
But my issue is slightly different .the child entity does exist but doesnt
have the  annotations representing the relation in the entity class.
It used to work fine without adding the relation on the inverse side in
2.15 but it crashes in 2.16

mpdude:
The fix from #10913 will be included in the next 2.17 bugfix release.

PR #10913, opened 2023-08-17
Include `ON DELETE CASCADE` associations in the delete order computation

### LINKED ISSUE #10348: Order of referenced child deletions not correct when referencing to itself

mpdude:
In order to resolve #10348, some changes were included in #10547 to improve the computed _delete_ order for entities. 

One assumption was that foreign key references with `ON DELETE SET NULL` or `... CASCADE` need not need to be taken into consideration when planning the deletion order, since the RDBMS would unset or cascade-delete such associations by itself when necessary. Only associations that do _not_ use RDBMS-level cascade handling would be sequenced, to make sure the referring entity is deleted before the referred-to one.

This assumption is wrong for `ON DELETE CASCADE`. The following examples give reasons why we need to also consider such associations, and in addition, we need to be able to deal with cycles formed by them.

In the following diagrams, `odc` means `ON DELETE CASCADE`, and `ref` is a regular foreign key with no extra `ON DELETE` semantics.

```mermaid
graph LR;
C-->|ref| B;
B-->|odc| A;
```

In this example, C must be removed before B and A. If we ignore the B->A dependency in the delete order computation, the result may not to be correct. ACB is not a working solution.

```mermaid
graph LR;
A-->|odc| B;
B-->|odc| A;
C-->|ref| B;
```

This is the situation in #10912. We have to deal with a cycle in the graph. C must be removed before A as well as B. If we ignore the B->A dependency (e.g. because we set it to "optional" to get away with the cycle), we might end up with an incorrect order ACB.

```mermaid
graph LR;
A-->|odc| B;
B-->|odc| A;
A-->|ref| C;
C-->|ref| B;
```

This example has no possible remove order. But, if we treat `odc` edges as optional, A -> C -> B would wrongly be deemed suitable.

```mermaid
graph LR;
A-->|ref| B;
B-->|odc| C;
C-->|odc| B;
D-->|ref| C;
```

Here, we must first remove A and D in any order; then, B and C in any order. If we treat one of the `odc` edges as optional, we might find the invalid solutions ABDC or DCAB.

#### Solution implemented in this PR

First, build a graph with a node for every to-be-removed entity, and edges for `ON DELETE CASCADE` associations between those entities. Then, use [Tarjan's algorithm](https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm) to find strongly connected components (SCCs) in this graph. The significance of SCCs is that whenever we remove one of the entities in a SCC from the database (no matter which one), the DBMS will immediately remove _all_ the other entities of that group as well.

For every SCC, pick one (arbitrary) entity from the group to represent all entities of that group. 

Then, build a second graph. Again we have nodes for all entities that are to be removed. This time, we insert edges for all regular (foreign key) associations and those with `ON DELETE CASCADE`. `ON DELETE SET NULL` can be left out. The edges are not added between the entities themselves, but between the entities representing the respective SCCs.

Also, for all non-trivial SCCs (those containing more than a single entity), add dependency edges to indicate that all entities of the SCC shall be processed _after_ the entity representing the group. This is to make sure we do not remove a SCC inadvertedly by removing one of its entities too early.

Run a topological sort on the second graph to get the actual delete order. Cycles in this second graph are a problem, there is no delete order.

Fixes #10912.

### DIFF
--- a/lib/Doctrine/ORM/Internal/StronglyConnectedComponents.php
+++ b/lib/Doctrine/ORM/Internal/StronglyConnectedComponents.php
@@ -4,7 +4,12 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Internal;
 
+use InvalidArgumentException;
+
 use function array_keys;
+use function array_pop;
+use function array_push;
+use function min;
 use function spl_object_id;
 
 /**
@@ -22,7 +27,7 @@ final class StronglyConnectedComponents
 {
     private const NOT_VISITED = 1;
     private const IN_PROGRESS = 2;
-    private const VISITED = 3;
+    private const VISITED     = 3;
 
     /**
      * Array of all nodes, indexed by object ids.
@@ -61,9 +66,7 @@ final class StronglyConnectedComponents
      */
     private $lowlink = [];
 
-    /**
-     * @var int
-     */
+    /** @var int */
     private $maxdfs = 0;
 
     /**
@@ -73,18 +76,16 @@ final class StronglyConnectedComponents
      */
     private $representingNodes = [];
 
-    /**
-     * @var array<int>
-     */
+    /** @var array<int> */
     private $stack = [];
 
     /** @param object $node */
     public function addNode($node): void
     {
-        $id = spl_object_id($node);
-        $this->nodes[$id] = $node;
+        $id                = spl_object_id($node);
+        $this->nodes[$id]  = $node;
         $this->states[$id] = self::NOT_VISITED;
-        $this->edges[$id] = [];
+        $this->edges[$id]  = [];
     }
 
     /** @param object $node */
@@ -102,7 +103,7 @@ final class StronglyConnectedComponents
     public function addEdge($from, $to): void
     {
         $fromId = spl_object_id($from);
-        $toId = spl_object_id($to);
+        $toId   = spl_object_id($to);
 
         $this->edges[$fromId][$toId] = true;
     }
@@ -118,15 +119,15 @@ final class StronglyConnectedComponents
 
     private function tarjan(int $oid): void
     {
-        $this->dfs[$oid] = $this->lowlink[$oid] = $this->maxdfs++;
+        $this->dfs[$oid]    = $this->lowlink[$oid] = $this->maxdfs++;
         $this->states[$oid] = self::IN_PROGRESS;
         array_push($this->stack, $oid);
 
-        foreach ($this->edges[$oid] as $adjacentId => $_) {
+        foreach ($this->edges[$oid] as $adjacentId => $ignored) {
             if ($this->states[$adjacentId] === self::NOT_VISITED) {
                 $this->tarjan($adjacentId);
                 $this->lowlink[$oid] = min($this->lowlink[$oid], $this->lowlink[$adjacentId]);
-            } else if ($this->states[$adjacentId] === self::IN_PROGRESS) {
+            } elseif ($this->states[$adjacentId] === self::IN_PROGRESS) {
                 $this->lowlink[$oid] = min($this->lowlink[$oid], $this->dfs[$adjacentId]);
             }
         }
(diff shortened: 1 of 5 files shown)


mpdude:
X-ref #9376

Dallas62:
Hi @mpdude,
Do you have an ETA on the landing of this fix ?
Thanks !

mpdude:
@Dallas62 Could you please try the change suggested in this PR?

lib/Doctrine/ORM/Internal/StronglyConnectedComponents.php
```diff
@@ -0,0 +1,166 @@
...
+    private $maxdfs = 0;
+
+    /**
+     * Nodes representing the SCC another node is in, indexed by lookup-node object ID
+     *
+     * @var array<int, object>
+     */
+    private $representingNodes = [];
+
+    /** @var array<int> */
```
greg0ire:
```suggestion
    /** @var list<int> */
```

Should there also be a note about what `int` represents here?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10912Test.php
```diff
@@ -0,0 +1,166 @@
...
+        $this->_em->remove($profile);
+        $this->_em->remove($user);
+        $this->_em->remove($room);
+
+        $queryLog = $this->getQueryLog();
+        $queryLog->reset()->enable();
+
+        $this->_em->flush();
+
+        $queries = array_values(array_filter($queryLog->queries, static function ($entry) {
```
greg0ire:
```suggestion
        $queries = array_values(array_filter($queryLog->queries, static function (array $entry): bool {
```

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10912Test.php
```diff
@@ -0,0 +1,166 @@
...
+            return strpos($entry['sql'], 'DELETE') === 0;
+        }));
+
+        self::assertCount(3, $queries);
+
+        // $room deletion is the first query
```
greg0ire: This comment could be converted to a `$message` argument to `assertSame`

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10913Test.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private function flushAndAssertNumberOfDeleteQueries(int $expectedCount): void
+    {
+        $queryLog = $this->getQueryLog();
+        $queryLog->reset()->enable();
+
+        $this->_em->flush();
+
+        $queries = array_values(array_filter($queryLog->queries, static function ($entry) {
```
greg0ire:
```suggestion
        $queries = array_values(array_filter($queryLog->queries, static function (array $entry): bool {
```

mpdude:
Feedback addressed

greg0ire (approved):
Looks good to me 👍
More than that, it was a very interesting read! Thanks for all the effort you put in this!

PR #10918, opened 2023-08-23
Add space before backquote

greg0ire:
According to the RST docs,

> [inline markup] it must be separated from surrounding text by non-word
> characters. Use a backslash escaped space to work around that: thisis\ *one*\ word.

Because we were missing a space before backquotes here, the links were not rendered. Escaping the space allow not to actually produce a space in the output.

See https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup


greg0ire:
Note that although `doctrine/rst-parser` cares about this, the Parser used by Github does not seem to care, hence, no diff.

greg0ire:
Thanks @linawolf for the solution!

linawolf (approved):

SenseException (approved):

PR #10919, opened 2023-08-23
Support Symfony 7 by adding return types conditionally

derrabus:
Symfony 7 will add more return types to the console component. This PR uses conditional traits to add the necessary return types if they're present upstream. This way, we won't break downstream projects that override our commands unless those projects upgrade to Symfony 7 themselves.


X-Coder264:
Is there any chance of getting this into the 2.15.x branch? Our app's test suite has over 7000 tests and hundreds if not thousands of them started failing after the recent UoW commit order changes that were made in the 2.16 release. Fixing all of that is gonna be a huge task and if this only gets merged into 2.17 it's effectively gonna prevent us from updating to Symfony 7 for a long while.

derrabus:
No, we don't maintain 2.15.x anymore. If you want to stay on 2.15, you'll have to fork that branch eventually anyway.

greg0ire (approved):

Issue #10927, opened 2023-09-01
IDENTITY identifier strategy for PostgreSQL breaks when inherited from MappedSuperclass


DemoniacDeath:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.16.0

#### Summary

Persisting an `@Entity` which has it's `@Id` defined in it's parent (`@MappedSuperclass`) with `IDENTITY` generator strategy on PostgreSQL platform.
It seems that the breaking change occured in this PR https://github.com/doctrine/orm/pull/10455, specifically changes in `ClassMetadataFactory` that removed the condtion of `$rootEntityFound` for calling `inheritIdGeneratorMapping()`

#### Previous behavior

In version `2.15.5` the `IdentityGenerator` used to retrieve inserted id came from the Entity itself with `$sequenceName` matching the schema. So the value was retrieved from the correct sequence.

#### Current behavior

Now the `IdentityGenerator` used to retrieve inserted id comes from the parent `@MappedSuperclass` and `$sequenceName` is wrong (name of parent class + sequence suffix) and non-existent in schema, so a PostgreSQL error about relation not existing is thrown.

#### How to reproduce

E.g. (irrelevant code and domain-specific terminology is ommited):
```php
/**
 * @ORM\MappedSuperclass
 */
abstract class AbstractTaskOperatorHistoryRecord
{
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="IDENTITY")
     */
    private ?int $id = null;
}
```

```php
/**
 * @ORM\Entity()
 * @ORM\Table(name="foo_bar_task_operator_history")
 */
class FooBarTaskOperatorHistoryRecord extends AbstractTaskOperatorHistoryRecord
{
}
```

```php
$this->_em->persist(new FooBarTaskOperatorHistoryRecord());
$this->_em->flush();
```

The PostgreSQL sequence used by `id` column in `foo_bar_task_operator_history` table is `foo_bar_task_operator_history_id_seq`.
The sequence used to retrieve inserted id was `foo_bar_task_operator_history_id_seq` before the breaking change. Now it is trying to get the value from `abstracttaskoperatorhistoryrecord_id_seq` which obviously does not and should not exist. The error thrown is
```
PDOException

SQLSTATE[42P01]: Undefined table: 7 ERROR: relation "abstracttaskoperatorhistoryrecord_id_seq" does not exist
```

[...]

PR #10928, opened 2023-09-01
added functional test case for issue#10927

DemoniacDeath:
added test functional test case for issue https://github.com/doctrine/orm/issues/10927
at the moment in version 2.16.0 the test fails with exception
```
PDOException: SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "gh10927parent_id_seq" does not exist
```
but passes on version 2.15.5

test is meant to run against PostgreSQL RDBMS


DemoniacDeath:
Closes #10927
Disallows generation of sequence name for `IdentityGenerator` if the class is a `MappedSuperclass`

DemoniacDeath:
Nevermind, the fix is bad

PR #10930, opened 2023-09-02
Improve doc job

greg0ire:


docs/en/sidebar.rst
```diff
@@ -1,3 +1,5 @@
+:orphan:
```
SenseException: In a local test this `:orphan:` was written out as text on the website.

Auswahl_001

greg0ire: Wow, thanks for checking this! I guess `doctrine/rst-parser` does not handle it at all :thinking: 
I just dropped that last commit.

SenseException: At least the recent 0.5.x that is used by the website. I don't know about 0.6.

SenseException (approved):

PR #10933, opened 2023-09-05
docs: use modern named arguments syntax

kaznovac:
use official named arguments syntax in example instead of pre php 8 codestyle for 'named' arguments

### DIFF
--- a/docs/en/reference/dql-doctrine-query-language.rst
+++ b/docs/en/reference/dql-doctrine-query-language.rst
@@ -1381,7 +1381,7 @@ Result Cache API:
     $query->setResultCacheDriver(new ApcCache());
 
     $query->useResultCache(true)
-          ->setResultCacheLifeTime($seconds = 3600);
+          ->setResultCacheLifeTime(3600);
 
     $result = $query->getResult(); // cache miss
 
@@ -1392,7 +1392,7 @@ Result Cache API:
     $result = $query->getResult(); // saved in given result cache id.
 
     // or call useResultCache() with all parameters:
-    $query->useResultCache(true, $seconds = 3600, 'my_query_result');
+    $query->useResultCache(true, 3600, 'my_query_result');
     $result = $query->getResult(); // cache hit!
 
     // Introspection
@@ -1457,7 +1457,7 @@ several methods to interact with it:
 
 -  ``Query::setQueryCacheDriver($driver)`` - Allows to set a Cache
    instance
--  ``Query::setQueryCacheLifeTime($seconds = 3600)`` - Set lifetime
+-  ``Query::setQueryCacheLifeTime($seconds)`` - Set lifetime
    of the query caching.
 -  ``Query::expireQueryCache($bool)`` - Enforce the expiring of the
    query cache if set to true.
--- a/docs/en/cookbook/aggregate-fields.rst
+++ b/docs/en/cookbook/aggregate-fields.rst
@@ -211,7 +211,7 @@ Now look at the following test-code for our entities:
     {
         public function testAddEntry()
         {
-            $account = new Account("123456", $maxCredit = 200);
+            $account = new Account("123456", maxCredit: 200);
             $this->assertEquals(0, $account->getBalance());
     
             $account->addEntry(500);
@@ -223,7 +223,7 @@ Now look at the following test-code for our entities:
     
         public function testExceedMaxLimit()
         {
-            $account = new Account("123456", $maxCredit = 200);
+            $account = new Account("123456", maxCredit: 200);
     
             $this->expectException(Exception::class);
             $account->addEntry(-1000);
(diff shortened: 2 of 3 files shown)


greg0ire:
Good idea. I can find more occurrences of this:

```
$ rg 'new \w+\(.*\$.*=' docs
docs/en/tutorials/pagination.rst
18:    $paginator = new Paginator($query, $fetchJoinCollection = true);

docs/en/cookbook/aggregate-fields.rst
214:            $account = new Account("123456", $maxCredit = 200);
226:            $account = new Account("123456", $maxCredit = 200);
```

```
$ rg '>\w+\([^\)]*\$[^\)]*=' docs
docs/en/reference/dql-doctrine-query-language.rst
1384:          ->setResultCacheLifeTime($seconds = 3600);
1395:    $query->useResultCache(true, $seconds = 3600, 'my_query_result');
```

Can you please fix those as well?

kaznovac:
Of course, I'll look into it later

kaznovac:
@greg0ire done

searched through the rest of documentation, no further instances

---

however:

there is this one, looks a bit like definition of default value:
https://github.com/doctrine/orm/blob/eafc0c66afe0c5340701f878882f8003b6455821/docs/en/reference/dql-doctrine-query-language.rst?plain=1#L1460-L1461

but method definition does not specify the default value:
https://github.com/doctrine/orm/blob/d81afdb6e3454007477ac00b8c98481c6f99af0f/lib/Doctrine/ORM/Query.php#L544

how would you like to proceed?
i'd guess this one should also be rewritten as named argument

greg0ire:
>  i'd guess this one should also be rewritten as named argument

Yeah but would you then change the `Query` part to `$query`? Maybe what's best would be to drop the `= 3600` part instead :thinking:

greg0ire:
Please squash your commits together, you shouldn't do one commit per file.

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -1381,7 +1381,7 @@ Result Cache API:
     $query->setResultCacheDriver(new ApcCache());
 
     $query->useResultCache(true)
-          ->setResultCacheLifeTime($seconds = 3600);
+          ->setResultCacheLifeTime(seconds: 3600);
```
SenseException: Isn't this one `lifetime`?

```suggestion
          ->setResultCacheLifeTime(lifetime: 3600);
```

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -1392,7 +1392,7 @@ Result Cache API:
     $result = $query->getResult(); // saved in given result cache id.
 
     // or call useResultCache() with all parameters:
-    $query->useResultCache(true, $seconds = 3600, 'my_query_result');
+    $query->useResultCache(true, seconds: 3600, 'my_query_result');
```
SenseException: This would be lifetime too.

```suggestion
    $query->useResultCache(true, lifetime: 3600, 'my_query_result');
```

SenseException (changes requested):
I'm not sure about this change in the docs. Introducing this would need us to check the docs everytime when we change an internal variable name of an argument when it's a method with just one argument.

The variable `$seconds` was just there for context. Should the seconds value be used directly?

greg0ire:
> everytime when we change an internal variable name of an argument when it's a method with just one argument.

How often are we going to do that? Because of named arguments, this is arguably a breaking change now.

docs/en/reference/dql-doctrine-query-language.rst
kaznovac: This method is deprecated since 2.7

should the documentation still promote its use?

kaznovac: I've asked as I'd like to remove this reference, or should we fix it now and remove it in the future?

SenseException:
> How often are we going to do that? Because of named arguments, this is arguably a breaking change now.

I suppose not that often, but it's still a bad practice to use this when we only have one argument or when all arguments are already covered. Question is if we keep named arguments for context or just use a variable again.

greg0ire:
Why is it a bad practice to use named arguments when there is only one argument? I mean in some situations, it might make things clearer. It's not always obvious what an argument is. For instance, if you pass 60, it can be handy to know we are talking seconds and not minutes.

SenseException:
That's why I ask if this should be kept for context. My change request still included the named arguments.

SenseException (approved):

greg0ire:
Thanks @kaznovac !

docs/en/reference/dql-doctrine-query-language.rst
derrabus: Yes, we should document the non-deprecated method here.

kaznovac: I've asked because method `userResultCache` is deprecated, and this section is written in a language that kind of 'promotes' it's use. I'd suggest to replace these lines with  suggested replacement `enableResultCache`, but that's for another PR

https://github.com/doctrine/orm/blob/9f555ea8fb4f7ae9a362285dc71e85776a1c5cc3/lib/Doctrine/ORM/AbstractQuery.php#L709-L726

derrabus: Yes, please send a PR. 🙂

PR #10946, opened 2023-09-15
Adds metadata field type validation against Entity property type

### LINKED ISSUE #10661: Metadata field type validation against Entity property type

greg0ire:
Copied from #10662 

This avoid situations where you might map a decimal to a float, when it
should really be mapped to a string. This is a big pitfall because in
that situation, the ORM will consider the field changed every time.

Closes #10662 , closes #10661

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -4,32 +4,72 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Tools;
 
+use Doctrine\DBAL\Types\AsciiStringType;
+use Doctrine\DBAL\Types\BigIntType;
+use Doctrine\DBAL\Types\BooleanType;
+use Doctrine\DBAL\Types\DecimalType;
+use Doctrine\DBAL\Types\FloatType;
+use Doctrine\DBAL\Types\GuidType;
+use Doctrine\DBAL\Types\IntegerType;
+use Doctrine\DBAL\Types\JsonType;
+use Doctrine\DBAL\Types\SimpleArrayType;
+use Doctrine\DBAL\Types\SmallIntType;
+use Doctrine\DBAL\Types\StringType;
+use Doctrine\DBAL\Types\TextType;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use ReflectionNamedType;
 
 use function array_diff;
+use function array_filter;
 use function array_key_exists;
+use function array_map;
+use function array_push;
 use function array_search;
 use function array_values;
+use function assert;
 use function class_exists;
 use function class_parents;
 use function count;
 use function get_class;
 use function implode;
 use function in_array;
+use function sprintf;
+
+use const PHP_VERSION_ID;
 
 /**
  * Performs strict validation of the mapping schema
  *
  * @link        www.doctrine-project.com
+ *
+ * @psalm-import-type FieldMapping from ClassMetadata
  */
 class SchemaValidator
 {
     /** @var EntityManagerInterface */
     private $em;
+
+    /**
+     * It maps built-in Doctrine types to PHP types
+     */
+    private const BUILTIN_TYPES_MAP = [
+        AsciiStringType::class => 'string',
+        BigIntType::class => 'string',
+        BooleanType::class => 'bool',
+        DecimalType::class => 'string',
+        FloatType::class => 'float',
+        GuidType::class => 'string',
+        IntegerType::class => 'int',
+        JsonType::class => 'array',
+        SimpleArrayType::class => 'array',
+        SmallIntType::class => 'int',
+        StringType::class => 'string',
+        TextType::class => 'string',
+    ];
 
     public function __construct(EntityManagerInterface $em)
     {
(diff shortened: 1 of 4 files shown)


greg0ire:
I'm reworking some stuff, bear with me…

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10661/GH10661Test.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->validator = new SchemaValidator($this->em);
+    }
+
+    public function testMetadataFieldTypeNotCoherentWithEntityPropertyType(): void
+    {
+        $class = $this->em->getClassMetadata(InvalidEntity::class);
+        $ce    = $this->validator->validateClass($class);
+
+        self::assertEquals(
+            ["The field 'Doctrine\Tests\ORM\Functional\Ticket\GH10661\InvalidEntity#property1' has the property type 'float' that differs from the metadata field type 'string' returned by the 'decimal' DBAL type."],
```
greg0ire, line 34: I added the "returned by the 'decimal' DBAL type" part, so that people understand where we get the second type from.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH10661/InvalidEntity.php
```diff
@@ -0,0 +1,25 @@
...
+     * @var int
+     * @Id
+     * @Column
+     */
+    protected $key;
+
+    /**
+     * @Column(type="decimal")
+     */
+    protected float $property1;
```
greg0ire, line 24: I switched to decimal vs float here because I think it is a more common mistake, and it was the one I had in mind so I wanted to see how it looked for that case.

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -304,4 +349,66 @@ public function getUpdateSchemaList(): array
 
         return $schemaTool->getUpdateSchemaSql($allMetadata, true);
     }
+
+    /** @return list<string> containing the found issues */
+    private function validatePropertiesTypes(ClassMetadataInfo $class): array
```
greg0ire, line 354: We have no choice but to use `ClassMetadataInfo` here for now because of the calling code, and that should be fine, although I considered using `if (PHP_VERSION_ID >= 70400 && $class instanceof ClassMetadata) {`… that would be a bit weird.

greg0ire:
Cc @DavideBicego

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -4,33 +4,73 @@
...
+        DecimalType::class => 'string',
+        FloatType::class => 'float',
+        GuidType::class => 'string',
+        IntegerType::class => 'int',
+        JsonType::class => 'array',
+        SimpleArrayType::class => 'array',
+        SmallIntType::class => 'int',
+        StringType::class => 'string',
+        TextType::class => 'string',
+    ];
```
greg0ire, line 72: IMO we should get rid of this harcoded map when DBAL v4 is out and switch to reflection on the DBAL type instead. The signatures will be typed.

Ocramius: This list is intentional though: `instanceof` is **NOT** sufficient, because subclassing allows for types to be customized in userland.

If `doctrine/dbal:^4` has `final` on all types, then we can use `instanceof`

Ocramius: Ah, I understand what you mean now: `convertToPhpType(): T`, so we look at `T`?

greg0ire: Yes. It could go in a property that is wider than T, or even narrower but that's kind of dangerous I'd say. At the very least, the intersection of both types must not be empty.

lib/Doctrine/ORM/Tools/SchemaValidator.php
SenseException: I'm not sure I understand this comment. Is this because of a possible different Persistence version and `validateClass(ClassMetadataInfo $class)`?

greg0ire: I'm saying that in the unique call site of this method, we have no guarantee that we have a CM and not just a CMI (so yes, it does have to do with `validateClass(ClassMetadataInfo $class)`. This explains why we use this type declaration here.

greg0ire:
ping @derrabus @SenseException :slightly_smiling_face:

SenseException (approved):

greg0ire:
Thanks @DavideBicego !

cuberinooo:
any recommendation for using type mixed ? For resource we used the type mixed, but unfortunatly php does not have type hint for resource therefore we used mixed. But now the validation gives the following error: 
 "... has the property type 'mixed' that differs from the metadata field type 'string' returned by the 'string' DBAL type. " We save the resource directly to the db as bytea type and it is typed as string on dbal type.

norkunas:
@cuberinooo I was asked in #11076 to check for the mixed type and ignore it always, so it will be resolved when it get's merged (if get's).

Issue #10943, opened 2023-09-14
Call to a member function removeQueryCacheProfile() on null


cmodijk:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | Not sure
| Version     | 2.16.2

#### Summary

I have create a [this](https://github.com/JCID/removeQueryCacheProfile) example repository to reproduce the error. When i grap this project using PHP `7.4` and al the dependencies in it I get this error.

The first call to the method tries to to connect to a database. When i ignore this and then refresh the page i get this error. It seems to happen after the cache is stored on disk and then it is tried to be reloaded. 

On the same error i already found https://github.com/doctrine/DoctrineBundle/issues/1654 and https://github.com/doctrine/orm/pull/10684 but this pull request does not fixes it. After debugging when i remove the `__unserialize` method from `lib/Doctrine/ORM/Query/ParserResult.php` it seems to work not sure what is going on here. 

#### Current behavior

Results in a error

>Call to a member function removeQueryCacheProfile() on null

#### How to reproduce

```bash
git clone git@github.com:JCID/removeQueryCacheProfile.git
cd removeQueryCacheProfile
composer install
symfony serve 
```

#### Expected behavior

To not have a error.

greg0ire:
Reproduced with PHP 8 (you can drop the php 7 constraint for your reproducer).

Here is the full stack trace after I add a return type to `getSQLExecutor()`:

```
TypeError:
Doctrine\ORM\Query\ParserResult::getSqlExecutor(): Return value must be of type Doctrine\ORM\Query\Exec\AbstractSqlExecutor, null returned

  at vendor/doctrine/orm/lib/Doctrine/ORM/Query/ParserResult.php:94
  at Doctrine\ORM\Query\ParserResult->getSqlExecutor()
     (vendor/doctrine/orm/lib/Doctrine/ORM/Query.php:286)
  at Doctrine\ORM\Query->_doExecute()
     (vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php:1212)
  at Doctrine\ORM\AbstractQuery->executeIgnoreQueryCache()
     (vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php:1166)
  at Doctrine\ORM\AbstractQuery->execute()
     (vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php:901)
  at Doctrine\ORM\AbstractQuery->getResult()
[...]

PR #10948, opened 2023-09-15
Add compatibility with the Symfony 4.4 VarExporter

cmodijk:
This issue tries to solve the weird Symfony `__unserialize` data Symfony `4.4` is giving to the method. This seems to be a bug that Symfony solved in `5.4` in https://github.com/symfony/symfony/commit/6ccb85e185bb1774778f687f866d80ab0db96e76. Original report / investigation https://github.com/doctrine/orm/issues/10943.

### DIFF
--- a/tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
@@ -116,6 +116,25 @@ class ParserResultSerializationTest extends OrmFunctionalTestCase
         yield '2.17.0' => [rtrim(file_get_contents(__DIR__ . '/ParserResults/single_select_2_17_0.txt'), "\n")];
     }
 
+    public function testSymfony44ProvidedData(): void
+    {
+        $sqlExecutor      = $this->createMock(SingleSelectExecutor::class);
+        $resultSetMapping = $this->createMock(ResultSetMapping::class);
+
+        $parserResult = new ParserResult();
+        $parserResult->setSqlExecutor($sqlExecutor);
+        $parserResult->setResultSetMapping($resultSetMapping);
+        $parserResult->addParameterMapping('name', 0);
+
+        $exported     = VarExporter::export($parserResult);
+        $unserialized = eval('return ' . $exported . ';');
+
+        $this->assertInstanceOf(ParserResult::class, $unserialized);
+        $this->assertInstanceOf(ResultSetMapping::class, $unserialized->getResultSetMapping());
+        $this->assertEquals(['name' => [0]], $unserialized->getParameterMappings());
+        $this->assertInstanceOf(SingleSelectExecutor::class, $unserialized->getSqlExecutor());
+    }
+
     private static function parseQuery(Query $query): ParserResult
     {
         $r = new ReflectionMethod($query, 'parse');
--- a/lib/Doctrine/ORM/Query/ParserResult.php
+++ b/lib/Doctrine/ORM/Query/ParserResult.php
@@ -141,7 +141,9 @@ class ParserResult
     {
         foreach (self::LEGACY_PROPERTY_MAPPING as $property => $legacyProperty) {
             $this->$property = $data[sprintf("\0%s\0%s", self::class, $legacyProperty)]
+                ?? $data[self::class][$legacyProperty]
                 ?? $data[sprintf("\0%s\0%s", self::class, $property)]
+                ?? $data[self::class][$property]
                 ?? $this->$property
                 ?? null;
         }


greg0ire (dismissed):

derrabus:
Is there a way we can test this? 🤔 

Also, can we revert this change on 3.0.x?

greg0ire:
> Is there a way we can test this? 🤔

In https://github.com/doctrine/orm/pull/10684, you introduced txt files containing serialized data. Maybe there should be a new text file with the 3rd format in it?

cmodijk:
>Is there a way we can test this? 🤔

The txt files containing serialized data was not really working for me as Symfony directly call's the `__unserialize` with the wrong formated data. So i added a new test instead to check if that mapped data is handled. 

>Also, can we revert this change on 3.0.x?

I'm not sure. If `3.0.x` is removing support for Symfony `4.4` then yes. In the end it depends on if Doctrine ORM whants to support Symfony `4.4`. When Symfony `4.4` support is dropped then this can be removed.

But maybe the whole `__unserialize` should be dropped at some point as it still tries to reflect the old data from version `2.14.3` and when the `__unserialize` method is not there Symfony `4.4` also work's as expected.

derrabus:
> I'm not sure. If `3.0.x` is removing support for Symfony `4.4` then yes.

Symfony 4.4 is EOL, so yes, we won't support it on the 3.0.x branch.

derrabus:
Can we actually use `VarExporter` in that test? Otherwise nobody will understand where that data structure is coming from.

cmodijk:
>Can we actually use VarExporter in that test? Otherwise nobody will understand where that data structure is coming from.

If there is a way to ensure we get the Symfony `4.4` version of the `VarExporter` in that test then yes. But I don't know how i could achief that. For context to @derrabus this seems to be fixed in Symfony `5.4` and higher.

derrabus:
We do run our tests against VarExporter 4.4, yes.

cmodijk:
@derrabus I have checked out the PHPUnit [check](https://github.com/doctrine/orm/actions/runs/6247060166/job/16959519746?pr=10948) but can't firure out where Symfony `4.4` is tested?

greg0ire:
It's tested here: https://github.com/doctrine/orm/actions/runs/6247060166/job/16959520070?pr=10948#step:4:171

cmodijk:
I have rewriten the test to only validate it when symfony/var-exporter versie `4.4` is used and then let it export the data and then unserialize it back to the prefered data. Would love to hear feedback if this is a correct way of testing this.

derrabus:
> I have rewriten the test to only validate it when symfony/var-exporter versie `4.4` is used

Why? I understand that the problem is only reproducible with VarExporter 4, but the test should pass with any version, shouldn't it?

cmodijk:
>Why? I understand that the problem is only reproducible with VarExporter 4, but the test should pass with any version, shouldn't it?

Yes, you are right did not think about that. I removed the skip.

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
```diff
@@ -61,6 +64,31 @@ public static function provideSerializedSingleSelectResults(): Generator
...
+
+        $parserResult = new ParserResult();
+        $parserResult->setSqlExecutor($sqlExecutor);
+        $parserResult->setResultSetMapping($resultSetMapping);
+        $parserResult->addParameterMapping('name', 0);
+
+        $exported = VarExporter::export($parserResult);
+        $data     = file_put_contents($tempFile, '<?php return ' . $exported . ';');
+
+        $unserialized = require $tempFile;
```
derrabus: This is a complicated way to perform an `eval()`, isn't it?

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
```diff
@@ -61,6 +64,31 @@ public static function provideSerializedSingleSelectResults(): Generator
...
+
+        $this->assertInstanceOf(ParserResult::class, $unserialized);
+        $this->assertInstanceOf(ResultSetMapping::class, $unserialized->getResultSetMapping());
+        $this->assertEquals(['name' => [0]], $unserialized->getParameterMappings());
+        $this->assertInstanceOf(SingleSelectExecutor::class, $unserialized->getSqlExecutor());
+
+        unlink($tempFile);
```
derrabus: What happens to this file if the test fails?

derrabus (changes requested):

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
cmodijk: I'm not sure how i can solve this i followd the Symfony [docs](https://symfony.com/doc/current/components/var_exporter.html) which has noted it to use it this way. 

```php
use Symfony\Component\VarExporter\VarExporter;

$exported = VarExporter::export($someVariable);
// store the $exported data in some file or cache system for later reuse
$data = file_put_contents('exported.php', '<?php return '.$exported.';');

// later, regenerate the original variable when you need it
$regeneratedVariable = require 'exported.php';
```

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
cmodijk: I'm not that famillair with PHPUnit, is there a way to executed something even if the test fails? Maybe i could use `try/finally`?

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
cmodijk: What we maybe could do is dump the file once, commit it and then use the output to validate. But then we still need a way to import the file which is the part your question is about. So maybe this is still not the way. 🤔

derrabus: You don't need a file. PHP has `eval()`. Use that.

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
derrabus: > is there a way to executed something even if the test fails?

Yes, the `tearDown()` method.

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
cmodijk: I have pushed a update to use `eval` instead.

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
cmodijk: The `unlink` method is not used after https://github.com/doctrine/orm/pull/10948#discussion_r1387758034

cmodijk:
@derrabus I have updated the code with the feedback. When you have time can you check if this is what you expected?

derrabus (dismissed):

derrabus (approved):

derrabus:
Thank you!

cafferata:
Thank you too!

cmodijk:
Yes, Thank you too! ❤️

PR #10952, opened 2023-09-24
Doctrine does not provide reconciliation and orphanRemoval tries to remove entities that are not orphans

bizmate:
The objective of this pull request is to fully understand how complex objects with several relations can be properly updated seemlessly using the Cascade and orphan removal approaches as described in the doctrine docs

- CASCADE: all is described to perform create, update and delete if and when an entity and its related objects are updated, created, modified and deleted.
- the orphan-removal also needs to be enabled to remove ophans indeed.

I have tried to create code on the basis of the above as shown in my long thread on Slack -> Symfony 
https://symfony-devs.slack.com/archives/C3FQPE6LE/p1687177566412739

So far I know the Doctrine is not achieving the removal of all the related entities if the top object that owns them is modified and this the relation with its sub-related entities. 

As such I would like to add a test with proof of concept to see if indeed the update performed in my code can be done without totally destroying the object but indeed just by modifing the code.

As such could anyone readying provide guidance as to how to add the 
- relevant configuration so that i can reproduce the following mappings in these entities....

```
App\Entity\Business:
  type: entity
  table: business
  repositoryClass: App\Infrastructure\Persistence\Doctrine\BusinessRepository
  id:
    id:
      type: string
      generator:
        strategy: NONE
  fields:
    name:
      type: string
      length: 255
      nullable: false

  manyToMany:
    reviews:
      targetEntity: Review
      joinTable:
        name: businesses_reviews
        joinColumns:
          business_id:
            referencedColumnName: id
            onDelete: CASCADE
        inverseJoinColumns:
          review_id:
            referencedColumnName: id
            unique: true
            onDelete: CASCADE
      #cascade: [ persist, remove ]
      cascade: [ all ]
      orphanRemoval: true
      fetch: EAGER

App\Entity\Review:
  type: entity
  table: review
  id:
    id:
      type: string
  fields:
    rating:
      type: integer
      nullable: false
    text:
      type: text
      nullable: true


  manyToOne:
    user:
      targetEntity: User
      joinColumn:
        name: user_id
        referencedColumnName: id
      cascade: [ all ]
      orphanRemoval: true
      fetch: EAGER

App\Entity\User:
  type: entity
  table: user
  id:
    id:
      type: string
  fields:
    name:
      type: string
      length: 255
      nullable:
        false
```

I can see the fields are in the entities themselves but I am not sure where the config that I would usually write in Yaml is placed in the repo,

### DIFF
--- a/tests/Doctrine/Tests/Models/Rating/Business.php
+++ b/tests/Doctrine/Tests/Models/Rating/Business.php
@@ -1,0 +1,92 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\Rating;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\JoinColumn;
+use Doctrine\ORM\Mapping\JoinTable;
+use Doctrine\ORM\Mapping\ManyToMany;
+use Doctrine\ORM\Mapping\Table;
+
+/**
+ * @Entity
+ * @Table(name="ratings_business")
+ */
+class Business
+{
+    /**
+     * @Id
+     * @Column(type="string")
+     * @GeneratedValue(strategy="NONE")
+     */
+    private string $id;
+
+    /**
+     * @psalm-var string|null
+     * @Column(type="string", length=255)
+     */
+    private string $name;
+
+    /**
+     * @psalm-doc https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/association-mapping.html#one-to-many-unidirectional-with-join-table
+     * @psalm-var Collection<Review>
+     * @ManyToMany(targetEntity="Review", cascade={"all"}, orphanRemoval=true, fetch="EAGER")
+     * @JoinTable(name="ratings_businesses_reviews",
+     *      joinColumns={@JoinColumn(name="business_id", referencedColumnName="id", onDelete="CASCADE")},
+     *      inverseJoinColumns={@JoinColumn(name="review_id", referencedColumnName="id", unique=true, onDelete="CASCADE")}
+     *      )
+     */
+    public $reviews;
+
+    public function __construct(string $id, string $name, ?Reviews $reviews = null)
+    {
+        $this->id   = $id;
+        $this->name = $name;
+        if ($reviews === null) {
+            $this->reviews = new ArrayCollection();
+        } else {
+            $this->reviews = $reviews;
+        }
+    }
+
+    public function getId(): string
+    {
+        return $this->id;
+    }
+
+    public function getName(): ?string
+    {
+        return $this->name;
+    }
+
+    /** @psalm-return Collection<int, Review> */
+    public function getReviews(): Collection
+    {
+        return $this->reviews;
+    }
+
+    /**
+     * @return $this
+     */
+    public function setReviews(Collection $reviews)
+    {
+        $this->reviews = $reviews;
+
+        return $this;
+    }
+
+    public static function loadMetadata(ClassMetadata $metadata): void
+    {
+        $metadata->setPrimaryTable(
+            ['name' => 'rating_business']
+        );
+    }
+}
--- a/tests/Doctrine/Tests/Models/Rating/Reviews.php
+++ b/tests/Doctrine/Tests/Models/Rating/Reviews.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\Rating;
+
+use Doctrine\Common\Collections\ArrayCollection;
+
+class Reviews extends ArrayCollection
+{
+}
--- a/tests/Doctrine/Tests/ORM/Functional/OneToManyToOneOrphanRemovalTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/OneToManyToOneOrphanRemovalTest.php
@@ -1,0 +1,348 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional;
+
+use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\Common\Collections\Expr\Comparison;
+use Doctrine\Common\Collections\Selectable;
+use Doctrine\Tests\Models\Rating\Business;
+use Doctrine\Tests\Models\Rating\Review;
+use Doctrine\Tests\Models\Rating\Reviews;
+use Doctrine\Tests\Models\Rating\User;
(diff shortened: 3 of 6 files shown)


tests/Doctrine/Tests/ORM/Functional/OneToManyToOneOrphanRemovalTest.php
```diff
@@ -0,0 +1,131 @@
...
+
+        $phonenumber = $user->getPhonenumbers()->remove(0);
+
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $query  = $this->_em->createQuery('SELECT p FROM Doctrine\Tests\Models\CMS\CmsPhonenumber p');
+        $result = $query->getResult();
+
+        self::assertCount(1, $result, 'CmsPhonenumber should be removed by orphanRemoval');
```
greg0ire: Why? You've only updated the inverse side of the relationship here, right? According to  https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/unitofwork-associations.html#important-concepts, Doctrine should not do anything here.

bizmate: hi @greg0ire I have not done a thorough update as currently I am still trying to figure out how the whole test suite works. I have noticed....
- data sets are inside models, and then model sets are in the OrmFunctionalTestCase
- I need to work out exactly where the metadata/configuration for the app goes as per YAML in the main description. 
- Then I will try to modify the test to represent what I am trying to achieve within doctrine 
other things might be missing but I am still working my way around things

tests/Doctrine/Tests/Models/Rating/Business.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return $this->reviews;
+    }
+
+    /**
+     * @return $this
+     */
+    public function setReviews(Reviews $reviews)
+    {
+        $this->reviews = $reviews;
```
stof, line 81: Please don't create a test fixture showing a case of replacing the whole collection. This is a very bad practice because it breaks the change tracking of Doctrine (as the collection is not the PersistentCollection anymore).

When modifying an entity loaded from the database, you must mutate the PersistentCollection that is in the property, not replacing it.

tests/Doctrine/Tests/Models/Rating/Reviews.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\Rating;
+
+use Doctrine\Common\Collections\ArrayCollection;
+
+class Reviews extends ArrayCollection
```
stof, line 9: Given that the ORM does not support using custom collection types when loading the entity, I suggest avoiding to create this class as it would show a bad practice.

tests/Doctrine/Tests/Models/Rating/Business.php
bizmate: In many cases I need to refresh an entity with its relations from outside the entity manages ie rebuilding it and then I need to save it. In the saving I need to know what to do as based on documentation my understanding was that a cascade operation with orphan removal would tell doctrine to do the reconciliation.

If this is not possible could you confirm in your opinion what the best approach would be? If i need to do some manual reconciliation do you have any suggestions on how to do it as it is not obvious to me that such changes would not be handled by the ORM

tests/Doctrine/Tests/ORM/Functional/OneToManyToOneOrphanRemovalTest.php
```diff
@@ -0,0 +1,178 @@
...
+
+    public function testOrphanRemovalRemoveWithExternallyBuiltArrayCollection(): void
+    {
+        $business = $this->_em->find(Business::class, $this->businessId);
+
+        $newReviews    = new Reviews();
+        $newReviewUser = new User('newUser', 'New User');
+        $newReviews->add(new Review('newReview1', 'new review 1', 4, $newReviewUser));
+
+        $business->setReviews($newReviews);
```
bizmate: @stof  this is an example of how my app might like to update an entity by just replacing the whole collection. In my experience with other solutions if the objects passed have the same IDs they are reconciled and or removed according to the new object. However in this use case, for instance here, the result is that the previous persisted collection had 4 items. I set a new collection with one item. I would expect the old items to be replaced/deleted and the new one to be saved (including the many to one relation) 
However the result is that they new review is added and the total becomes 5 .

stof: Well, as I said on Slack, if you replace the PersistentCollection doing the tracking, you cannot expect the change tracking to work.

tests/Doctrine/Tests/ORM/Functional/OneToManyToOneOrphanRemovalTest.php
```diff
@@ -0,0 +1,269 @@
...
+        $query  = $this->_em->createQuery('SELECT u FROM Doctrine\Tests\Models\Rating\User u');
+        $result = $query->getResult();
+
+        self::assertCount(2, $result, 'User should be removed by orphanRemoval');
+    }
+
+    public function testOrphanRemovalRemoveUpdateWithNewCollection(): void
+    {
+        $business = $this->_em->find(Business::class, $this->businessId);
+        $business->setReviews($this->replaceReviews);
```
bizmate: Cascade all does not detect existing user and does not do reconciliation
There was 1 error:

1) Doctrine\Tests\ORM\Functional\OneToManyToOneOrphanRemovalTest::testOrphanRemovalRemoveUpdateWithNewCollection
Doctrine\DBAL\Exception\UniqueConstraintViolationException: An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'user1' for key 'ratings_user.PRIMARY'

tests/Doctrine/Tests/ORM/Functional/OneToManyToOneOrphanRemovalTest.php
bizmate: I am not replacing PersistentCollections but indeed just using a Collection provided by an external request, such as an update to a DTO, and feeding this DTO, that contains Collections, to the repository.

It is clear to me now how this whole thing work, ie you need to perform reconciliation and cannot use PPO, ie DTOs, directly with Doctrine save unless you are indeed creating a new record for sure.

Ideally this should be more obvious from documentation, usually showing rather simple use cases but also it would really be great if a reconciliation mechanism would be set in place. The latter is probably complex such as a serializer/mapping logic but it would really streamline interaction with repositories and make them very slim.

About documentation the only thing i see close to this matter is in this section
https://www.doctrine-project.org/projects/doctrine-orm/en/current/tutorials/getting-started.html#adding-behavior-to-entities

Here there is a User -> updateFromProfile function supposed to be responsible to do such reconciliation but again it is not clear. The only way I could see this working is by adding logic or a processor within the repository because with complex relations you need to continuously dig down the structure and take care of checking if an item is in a collection, update it if it is, remove it if removed in the DTO and take care of all its sub relations so the entity manager is needed continuously.


I will close the PR for now as it does not look like it is of interest anyway and it is rather complex for me to even design a possible solutinon

stof: your setter replaces the persistent collection with the external collection, instead of modifying the persistent collection that tracks changes done in the relation.

PR #10955, opened 2023-09-28
docs: in text, refer to PHP attributes when talking about metadata

dannyvankooten:
I was going through the getting started tutorial and was momentarily confused reading about `many-to-many` tags and `many-to-one` tags because I was looking at the default code examples, which uses PHP attributes. 

This PR changes the text to talk about metadata in terms of attributes, as that is what is visible by default (and probably what most people will use nowadays, but I could be wrong there).

### DIFF
--- a/docs/en/tutorials/getting-started.rst
+++ b/docs/en/tutorials/getting-started.rst
@@ -1049,21 +1049,21 @@ which translates the YYYY-mm-dd HH:mm:ss database format
 into a PHP DateTime instance and back.
 
 After the field definitions, the two qualified references to the
-user entity are defined. They are created by the ``many-to-one``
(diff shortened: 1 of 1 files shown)


greg0ire:
You're absolutely right, but why target 3.0.x?

dannyvankooten:
> You're absolutely right, but why target 3.0.x?

No particular reason, it's just that that was the documentation I was reading when I noticed this. Wasn't sure whether doc updates to any of the 2.x branches would be merged in 3.x too.

Want me to update the latest 2.x branch as well?

greg0ire:
Please retarget this PR to 2.x, and your changes should eventually get merged up in 3.x

PR #1215, opened 2014-12-08
Don't make this class Final

DerMonozukurist:
Why: It makes testing painful

https://github.com/phpspec/prophecy/issues/102

PR #10957, opened 2023-09-28
Expose commonly used Query methods in existing interface

ostrolucky:
There were few previous attempts to solve problems with Query class being final:
https://github.com/doctrine/orm/pull/1215
https://github.com/doctrine/orm/pull/6180
https://github.com/doctrine/orm/pull/6181

As well as complaints
https://github.com/doctrine/orm/discussions/10947

Removing final keyword has been rejected, as well as adding new interface.

This provides compromise: We don't remove final keyword nor add new interface. Instead, we move few methods to `AbstractQuery`. They are not required for classes extending `AbstractQuery` (they aren't abstract methods), hence no BC break there. Most people feeling the need to mock `Query` class do that because of these few methods and this would work for them.

### DIFF
--- a/lib/Doctrine/ORM/AbstractQuery.php
+++ b/lib/Doctrine/ORM/AbstractQuery.php
@@ -33,8 +33,10 @@ use function array_shift;
 use function assert;
 use function count;
 use function func_num_args;
+use function get_debug_type;
 use function in_array;
 use function is_array;
+use function is_int;
 use function is_numeric;
 use function is_object;
 use function is_scalar;
@@ -166,6 +168,20 @@ abstract class AbstractQuery
 
     /** @var int */
     protected $lifetime = 0;
+
+    /**
+     * The first result to return (the "offset").
+     *
+     * @var int
+     */
+    private $firstResult = 0;
+
+    /**
+     * The maximum number of results to return (the "limit").
+     *
+     * @var int|null
+     */
+    private $maxResults = null;
 
     /**
      * Initializes a new instance of a class derived from <tt>AbstractQuery</tt>.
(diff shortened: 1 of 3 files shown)


derrabus:
Your change would add these methods to `NativeQuery` as well, although they don't have any effect there. Wouldn't that be confusing?

ostrolucky:
Indeed. Then what do you suggest? I would say mark it `/* @final */` instead, or add a new interface. I would go with former personally.

beberlei:
Why do we want to support people mocking Query? Its painful to mock fluent apis and mocking third party code is generally a bad practice. We could debate if its our place to educate users

ostrolucky:
It's needed in order to write unit test for pagination logic in applications, like so
```php
$queryBuilder = $this->createMock(QueryBuilder::class);
$query = $this->createMock(Query::class);
$queryBuilder->expects($this->exactly(3))->method('getQuery')->willReturn($query);
$query->expects($this->exactly(3))->method('setFirstResult')->withConsecutive([0], [100], [200])->willReturn($query);
$query->expects($this->exactly(3))->method('setMaxResults')->with(150)->willReturn($query);
$this->queryBuilder->expects($this->exactly(3))->method('getQuery')->willReturn($queryBuilder) ;

$this->handler->execute($foo, $bar);
```

I don't think there is any education possible here. If you replace these interfaces with your own so that you don't test them directly, you still can't write tests for your adapters that make a bridge between your interfaces and ORM interfaces, because of same problem. This is the problem with "don't mock what you don't own" mantra. People keep repeating it, but I don't see an alternative.

derrabus:
> It's needed in order to write unit test for pagination logic in applications, like so

I'm sorry, but that code is painful to read and even more painful to maintain. Mocking the QueryBuilder will almost certainly result in a brittle test.

But if you want to go down that road: You don't really need a mock of the query here. That test might already work if you create a `Query` instance via reflection. Or in other words: If we were to build some tooling that gives you a dummy `Query` instance that and be passed around and that passes type guards, your problem would be solved?

> This is the problem with "don't mock what you don't own" mantra. People keep repeating it, but I don't see an alternative.

The alternative is to test against the real thing. And yes, that means that your pagination logic is tested against a real database. Could be an SQLite3 in-memory DB, fwiw.

ostrolucky:
> I'm sorry, but that code is painful to read and even more painful to maintain. Mocking the QueryBuilder will almost certainly result in a brittle test.

That's quite subjective. We have it in project since 2020 without issues and this is how unit tests traditionally look like, nothing special there.

> The alternative is to test against the real thing. And yes, that means that your pagination logic is tested against a real database. Could be an SQLite3 in-memory DB, fwiw.

Integration tests are too slow and go against test pyramid. Majority tests are supposed to be unit tests. Why do consumers have to write integration test here just because this one class?

> But if you want to go down that road: You don't really need a mock of the query here. That test might already work if you create a Query instance via reflection. Or in other words: If we were to build some tooling that gives you a dummy Query instance that and be passed around and that passes type guards, your problem would be solved?

Real query object won't work, because you can't assert methods in it were called. But yes, if you ship dummy query instance without missing methods that people can mock and ideally passes all the typehints (Eg. `Parser::__construct`, `QueryBuilder::getQuery`, `Paginator`) that would work. I don't see how that's preferrable over removing `final` from `Query`, or what would we as maintainers gain from that over removing `final`, but sure.

derrabus:
> That's quite subjective.

It's the kind of tests that makes people feel miserable about writing tests. If people feel like they have to write tests like this, I think we should at least evaluate if we could provide a better experience. That's what I was trying to do here.

I won't reply to the other arguments you've made because I believe that would only waste your time and mine. You seem very determined about this style of testing.

> Real query object won't work, because you can't assert methods in it were called.

I think, this is the important part here. The `Query` is not a pure value object. It encapsulates behavior with side effects (the `execute()` method). I understand why one would want to intercept calls to `execute()` and friends.

Let's go with a soft `final` by annotation, see #10990.

PR #10964, opened 2023-10-06
Updated order of mapping attribute parameters

soltmar:
Currently, when defining entities, PhpStorm complains about order of Doctrine Mapping PHP Attribute parameter order being wrong:

image

This PR changes parameter order for \Doctrine\ORM\Mapping\Index.php and \Doctrine\ORM\Mapping\OneToMany.php to match what is shown in documentation:
image
image


greg0ire (changes requested):
This is a breaking change for people not using named arguments. I don't think there is a way for us to deprecate not using named arguments, which means I don't see a way to do this on 3.0.x AND provide an upgrade path. Hopefully somebody will.

soltmar:
You're right. I haven't thought about this as been using named arguments since the beginning. Just wondering how big is usage without named arguments. Maybe we can start some wider discussion about it ? I know this is a minor thing but I'm surprised it wasn't picked before.

Also technically an upgrade would be to swap order of arguments for Index  and OneToMany on entities which shouldn't be too much work (although boring one). I could see this going to 3.0 since 3.0 already have breaking changes comparing to 2.x

greg0ire:
By upgrade path, I mean some change you could do while on 2.x that would allow you to be forward-compatible with 3.0.x (meaning you wouldn't need to perform any change after updating to 3.0.0)

greg0ire:
But since I don't think a lot of people omit argument names, I agree that this can go to 3.0.x. Please retarget, and add something about this in UPGRADE.md

soltmar:
Done. Let me know if you need Different / better explanation in UPGRADE.md

Thanks

greg0ire (dismissed):

UPGRADE.md
```diff
@@ -621,6 +621,11 @@ following classes and methods:
 
 Use `toIterable()` instead.
 
+## Minor BC BREAK: Changed order of arguments passed to `OneToOne`, `ManyToOne` and `Index` mapping PHP attributes
+
+To keep PHP mapping attributes consistent, order of arguments passed to above attributes has been changed
+so `$targetEntity` is a first argument now. This change affects only non-named arguments usage.
+
```
derrabus: New entries should be added at the top of the file.

soltmar: Pushed to the top. Thanks

SenseException:
Am I understanding it correctly that this change is introduced to satisfy one IDE or IDE plugin?

soltmar:
Hi @SenseException , not just to satisfy IDE but also to keep arguments of mapping classes in order and according to what you have in doctrine docs which I personally think will be more intuitive.

greg0ire (dismissed):

derrabus:
> Am I understanding it correctly that this change is introduced to satisfy one IDE or IDE plugin?

The PhpStorm inspection is bogus. I agree that the inspection complaining is not enough to motivate this change. However, the parameters are currently ordered inconsistently which is something we can fix in 3.0.

SenseException (approved):

greg0ire:
Thanks @soltmar !

PR #10967, opened 2023-10-07
Remove CommitOrderCalculator and related classes

greg0ire:
They are unused, and since they are internal, it should be fine to remove them without a deprecation.

This is an alternative to #10967

Closes #10966


mpdude:
We discussed this IRL and agreed it's fine to remove these classes in 2.x

mpdude (approved):

PR #10968, opened 2023-10-07
Reject ID collisions in identity map unconditionally

greg0ire:


greg0ire:
Cc @mpdude :slightly_smiling_face:

lib/Doctrine/ORM/Configuration.php
```diff
@@ -605,13 +605,18 @@ public function setLazyGhostObjectEnabled(bool $flag): void
         $this->attributes['isLazyGhostObjectEnabled'] = $flag;
     }
 
+    /** To be deprecated in 3.1.0 */
     public function setRejectIdCollisionInIdentityMap(bool $flag): void
```
derrabus, line 609: We should throw if `$flag === false`.

derrabus (changes requested):

lib/Doctrine/ORM/Configuration.php
greg0ire: done

derrabus (approved):

mpdude:
Thank you :+1:

PR #10969, opened 2023-10-07
Use lazy ghosts unconditionally

greg0ire:
A nice consequence of this is that we can almost get rid of `doctrine/common` now. I think we might only depend on [`ClassUtils`](https://github.com/doctrine/common/blob/3.4.x/src/Util/ClassUtils.php) now.


lib/Doctrine/ORM/Proxy/ProxyFactory.php
```diff
@@ -191,14 +164,13 @@ public function __construct(
     }
 
     /**
-     * {@inheritDoc}
+     * @param class-string $className
+     * @param array<mixed> $identifier
+     *
+     * @return InternalProxy
```
greg0ire, line 170: Using a native return type here will involve modify some tests that mock calls to `getProxy` and make them return actual entities.

greg0ire:
Cc @nicolas-grekas

lib/Doctrine/ORM/Configuration.php
```diff
@@ -581,28 +577,19 @@ public function setSchemaIgnoreClasses(array $schemaIgnoreClasses): void
...
+     * @return true
+     */
     public function isLazyGhostObjectEnabled(): bool
     {
-        return $this->attributes['isLazyGhostObjectEnabled'] ?? false;
+        return true;
     }
 
+    /** To be deprecated in 3.1.0 */
     public function setLazyGhostObjectEnabled(bool $flag): void
```
derrabus, line 592: We should throw if `$flag === false`.

lib/Doctrine/ORM/Proxy/ProxyFactory.php
derrabus: I think we should fix those tests, but that could be done in a follow-up.

derrabus (commented):
Please modify composer.json and make `symfony/var-exporter` a mandatory dependency.

greg0ire:
> I think we should fix those tests, but that could be done in a follow-up.

Yeah me too… but I haven't found a clear solution so far.

composer.json
```diff
@@ -33,7 +33,8 @@
         "doctrine/lexer": "^2.1 || ^3",
         "doctrine/persistence": "^3.1.1",
         "psr/cache": "^1 || ^2 || ^3",
-        "symfony/console": "^5.4 || ^6.0 || ^7.0"
+        "symfony/console": "^5.4 || ^6.0 || ^7.0",
+        "symfony/var-exporter": "^5.4 || ^6.2"
```
derrabus:
```suggestion
        "symfony/var-exporter": "^6.2"
```

Lazy ghost objects don't exist in v5.4. We need version 6.

greg0ire: :thinking: this makes me notice that for some reason, we don't have a job with `--prefer-lowest` on this branch.

greg0ire: @derrabus apparently you removed it in https://github.com/doctrine/orm/pull/9069 … do you recall why?

derrabus: No, let's add it back.

greg0ire: Done in #10971

derrabus (approved):

PR #10970, opened 2023-10-07
Allow to disable "DISTINCT" in LimitSubqueryWalker

### LINKED ISSUE #10553: LimitSubqueryWalker performance issue with DISTINCT
### Feature Request
|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary
 `\Doctrine\ORM\Tools\Pagination\LimitSubqueryWalker` Adds a `DISTINCT` keyword to paginate with no side effects queries that potentially contain multiple occurrences because of a join table that adds additional rows...

See https://github.com/doctrine/orm/blob/f82485e651763fbd1b34879726f4d3b91c358bd9/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php#L69

The `DISTINCT` might be expensive in some cases. 

In some cases it is known (not by doctrine but by the developer writing the query) that despite having joins, they are not adding extra rows, thus `DISTINCT` will not remove any row.

I would like to add a way into `LimitSubqueryWalker` to not use the distinct keyword.

Would it be acceptable something as this:

```php
$AST->selectClause->isDistinct  =  !$query->hasHint('opt_out_distinct') || !$query->getHint('opt_out_distinct');
```

in this way when writing the query, it would be possible to set the `opt_out_distinct` to skip the `DISTINCT`.

```php
$query->setHint('opt_out_distinct', true);
```
What do you think? is it acceptable?

A similar issue was reported in API platform in https://github.com/api-platform/api-platform/issues/792

If yes i can prepare a PR.

goetas:
Closes https://github.com/doctrine/orm/issues/10553

~Im not sure what is the correct branch to target.. 2.16 or 2.17...~

### DIFF
--- a/docs/en/tutorials/pagination.rst
+++ b/docs/en/tutorials/pagination.rst
@@ -43,3 +43,18 @@ the future.
 .. note::
 
     ``fetchJoinCollection`` argument set to ``true`` might affect results if you use aggregations in your query.
+
+By using the ``Paginator::HINT_ENABLE_DISTINCT`` you can instruct doctrine that the query to be executed
+will not produce "duplicate" rows (only to-one relations are joined), thus the SQL limit will work as expected.
+In this way the `DISTINCT` keyword will be omitted and can bring important performance improvements.
+
+.. code-block:: php
+
+    <?php
+    use Doctrine\ORM\Tools\Pagination\Paginator;
+
+    $dql = "SELECT u, p FROM User u JOIN u.mainPicture p";
+    $query = $entityManager->createQuery($dql)
+                           ->setHint(Paginator::HINT_ENABLE_DISTINCT, false)
+                           ->setFirstResult(0)
+                           ->setMaxResults(100);
--- a/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
@@ -50,12 +50,14 @@ class LimitSubqueryWalker extends TreeWalkerAdapter
             throw new RuntimeException('Paginating an entity with foreign key as identifier only works when using the Output Walkers. Call Paginator#setUseOutputWalkers(true) before iterating the paginator.');
         }
 
-        $this->_getQuery()->setHint(
+        $query = $this->_getQuery();
+
+        $query->setHint(
             self::IDENTIFIER_TYPE,
             Type::getType($rootClass->fieldMappings[$identifier]['type'])
         );
 
-        $this->_getQuery()->setHint(self::FORCE_DBAL_TYPE_CONVERSION, true);
+        $query->setHint(self::FORCE_DBAL_TYPE_CONVERSION, true);
 
         $pathExpression = new PathExpression(
             PathExpression::TYPE_STATE_FIELD | PathExpression::TYPE_SINGLE_VALUED_ASSOCIATION,
@@ -66,7 +68,7 @@ class LimitSubqueryWalker extends TreeWalkerAdapter
         $pathExpression->type = PathExpression::TYPE_STATE_FIELD;
 
         $AST->selectClause->selectExpressions = [new SelectExpression($pathExpression, '_dctrn_id')];
-        $AST->selectClause->isDistinct        = true;
+        $AST->selectClause->isDistinct        = ($query->getHints()[Paginator::HINT_ENABLE_DISTINCT] ?? true) === true;
 
         if (! isset($AST->orderByClause)) {
             return;
--- a/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
+++ b/lib/Doctrine/ORM/Tools/Pagination/Paginator.php
@@ -34,6 +34,8 @@ class Paginator implements Countable, IteratorAggregate
 {
     use SQLResultCasing;
 
+    public const HINT_ENABLE_DISTINCT = 'paginator.distinct.enable';
+
     /** @var Query */
     private $query;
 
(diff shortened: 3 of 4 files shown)


greg0ire:
Hi @goetas ! You should target 2.17, this is a new feature, right?

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
```diff
@@ -66,7 +68,7 @@ public function walkSelectStatement(SelectStatement $AST)
         $pathExpression->type = PathExpression::TYPE_STATE_FIELD;
 
         $AST->selectClause->selectExpressions = [new SelectExpression($pathExpression, '_dctrn_id')];
-        $AST->selectClause->isDistinct        = true;
+        $AST->selectClause->isDistinct        = ! $query->hasHint(Paginator::HINT_DISABLE_DISTINCT) || ! $query->getHint(Paginator::HINT_DISABLE_DISTINCT);
```
greg0ire: This would be less convoluted with something positively named, like `HINT_ENABLE_DISTINCT`. You wouldn't need to call `hasHint`, or to negate anything.

lib/Doctrine/ORM/Tools/Pagination/Paginator.php
```diff
@@ -34,6 +34,8 @@ class Paginator implements Countable, IteratorAggregate
 {
     use SQLResultCasing;
 
+    public const HINT_DISABLE_DISTINCT = 'paginator.distinct.disable';
```
greg0ire: Should this be documented somewhere?

goetas:
Thanks for the review, all suggestions make sense, will address them tomorrow

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
goetas: I have tried this but `getHint` does not allow you to detect if the int is not there or is set to `false`. Since the `distinct` has to be used by default, i could not find a way to make it backward compat

goetas: Only is to use `getHints` and then check if the hint is present in the array

beberlei:
The distinct behavior was the default before, so using an opt in enable flag will be a bc break. Its better to distinct by default, assume the worst, take a perf hit, bur at least result is guaranteed to be correct

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
```diff
@@ -66,7 +68,7 @@ public function walkSelectStatement(SelectStatement $AST)
         $pathExpression->type = PathExpression::TYPE_STATE_FIELD;
 
         $AST->selectClause->selectExpressions = [new SelectExpression($pathExpression, '_dctrn_id')];
-        $AST->selectClause->isDistinct        = true;
+        $AST->selectClause->isDistinct        = ($query->getHints()[Paginator::HINT_ENABLE_DISTINCT] ?? null) !== false;
```
beberlei:
```suggestion
        $AST->selectClause->isDistinct        = ($query->getHints()[Paginator::HINT_ENABLE_DISTINCT] ?? true) !== false;
```

Makes more sense to make the default more obvious.

beberlei (dismissed):

docs/en/tutorials/pagination.rst
```diff
@@ -43,3 +43,18 @@ the future.
 .. note::
 
     ``fetchJoinCollection`` argument set to ``true`` might affect results if you use aggregations in your query.
+
+By using the ``Paginator::HINT_ENABLE_DISTINCT`` you can instruct doctrine that the query to be executed
+will not produce "duplicate" rows (only to-one relations are joined), thus the SQL limit will work as expected.
+In this way the `DISTINCT` keyword will be omitted and can bing important performance improvements.
```
greg0ire:
```suggestion
In this way the `DISTINCT` keyword will be omitted and can bring important performance improvements.
```

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
greg0ire: Oh right. So that would be the only hint defaulting to `true`? Or are there other cases?

goetas: at the moment i see only true/false options, im not sure if it makes sense to have it as string and to call it "strategy" and have for now possible values such as: distinct (default) and none (new option).

goetas:
suggestions applied

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
greg0ire: I'm OK with it being a boolean, I was more wondering about what other hints look like right now (and why there is this `false` default). If this hint you introduce is the only one that is defaulting to `true` then it might be best to revert to the previous situation with the negative naming like you did at the beginning.

derrabus (approved):

lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php
greg0ire: I've taken a look at other hints in the codebase, and it looks like this is the first time we have one that we want to default to `true`. So what you did is fine 👍

greg0ire:
Thanks @goetas !

goetas:
:tada:

PR #10971, opened 2023-10-07
Restore job with --prefer-lowest

greg0ire:
It is useful to catch misconfigured dependency constraints. It was removed in #9069.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,11 +39,11 @@
         "doctrine/coding-standard": "^12.0",
         "phpbench/phpbench": "^1.0",
         "phpstan/phpstan": "1.10.35",
-        "phpunit/phpunit": "^10.0.14",
+        "phpunit/phpunit": "^10.4.0",
         "psr/log": "^1 || ^2 || ^3",
         "squizlabs/php_codesniffer": "3.7.2",
         "symfony/cache": "^5.4 || ^6.2",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -32,18 +32,18 @@
         "doctrine/instantiator": "^1.3 || ^2",
         "doctrine/lexer": "^2.1 || ^3",
         "doctrine/persistence": "^3.1.1",
-        "psr/cache": "^1 || ^2 || ^3",
+        "psr/cache": "^3",
```
greg0ire: We require the fluent interface provided by 3.0.0: https://github.com/doctrine/orm/blob/62ed63bbbe3f59dcacb7dadb574895f5cb27a257/lib/Doctrine/ORM/Query.php#L276

derrabus: The fluent interface is part of the standard since v1.

greg0ire: :thinking: I guess the issue must be with a wrong implementation then.

greg0ire: > MockObject_CacheItemInterface_9583ebe3

Ah yes, that makes sense. The mocking framework won't enforce a phpdoc-provided return type.

greg0ire: Fixed properly.

SenseException (approved):

PR #10979, opened 2023-10-09
Deprecate WITH in joined associations

### LINKED ISSUE #10978: Deprecate JOIN ... WITH keyword for join associations
There are two ways to join entities and both can be restricted further using the `WITH` keyword:

1. using join assocations. 
```
SELECT u, a FROM App\Entity\User u JOIN u.addresses a WITH a.country = 'DE'
```

2. using range variable declaration:
```
SELECT u, a FROM App\Entity\User u JOIN App\Entity\Address a WITH u.id = a.user AND a.country = 'DE'
```

The first one will fetch all addresses into the collection User::$addresses, the second one will cause a mixed result to emit both entities at the same time.

Case 1 will causes problems, because looking at a collection, you never know if its fully or only partially loaded. And reloading the collection is also not possible in a meaningful way. See this comment https://github.com/doctrine/orm/issues/2878#issuecomment-162361011

As such, we are looking to deprecate and remove case 1, and also deprecate restricting by an association that is fetched joined, say:

```
SELECT u, a FROM App\Entity\User u JOIN u.addresses a WHERE a.country = 'DE'
```

What the users mean to do is probably this:

```
SELECT u, a FROM App\Entity\User u JOIN u.addresses a WHERE u.id IN (SELECT a2.user FROM App\Entity\Address a2 WHERE a2.country = 'DE')
```

beberlei:
Deprecates `JOIN path.association WITH` type of expressions as they create persistent collections with a subset of related entities and look like they are fully loaded, causing all kinds of trouble.

A user could potentially work around this and move the condition to the `WHERE` clause. I checked the code and its not really possible in this case to detect it without adding tons of more code. So the WITH deprecation is what we can do about this. 

Fixes #10978

### DIFF
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -1748,14 +1748,6 @@ class Parser
 
         // Check for ad-hoc Join conditions
         if ($adhocConditions) {
-            if ($joinDeclaration instanceof AST\JoinAssociationDeclaration) {
-                Deprecation::trigger(
-                    'doctrine/orm',
-                    'https://github.com/doctrine/orm/issues/10978',
-                    'WITH join conditions are deprecated for joins on associations. Use either a sub-select with another occurance of this table or use Collection::matching(Criteria) in follow-up code.'
-                );
-            }
-
             $this->match(Lexer::T_WITH);
 
             $join->conditionalExpression = $this->ConditionalExpression();
--- a/lib/Doctrine/ORM/Query/Parser.php
+++ b/lib/Doctrine/ORM/Query/Parser.php
@@ -1752,7 +1752,7 @@ class Parser
                 Deprecation::trigger(
                     'doctrine/orm',
                     'https://github.com/doctrine/orm/issues/10978',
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -1748,6 +1748,14 @@ public function Join()
 
         // Check for ad-hoc Join conditions
         if ($adhocConditions) {
+            if ($joinDeclaration instanceof AST\JoinAssociationDeclaration) {
+                Deprecation::trigger(
+                    'doctrine/orm',
+                    'https://github.com/doctrine/orm/issues/10978',
+                    'WITH join conditions are deprecated for joins on associations. Use either a sub-select with another occurance of this table or use Collection::matching(Criteria) in follow-up code.'
```
greg0ire:
```suggestion
                    'WITH join conditions are deprecated for joins on associations. Use either a sub-select with another occurrence of this table or use Collection::matching(Criteria) in follow-up code.'
```

mpdude:
Looks good regarding the `WITH` keyword in JoinAssociation parts of the DQL. 

In order to fix #10978, we additionally need to take care of WHERE expressions referring to fetch-join aliases.

beberlei:
@mpdude ah yes, I checked that - and its not possible, because you don't really know how its used.

PR #10981, opened 2023-10-10
Take `ReflectionType::allowsNull` into account

beberlei:
This reintroduces the behavior removed in #8732 due to #8723 for 3.0

When you have a column with a typed property, then the null status of the property determines if the column's `nullable` mapping.

This is only done for fields and not for associations, because often association join columns need to be nullable to be updated later after the insert due to extra update and ordering reasons.

For 2.x, we will make this an opt-in behavior in the `DefaultTypedFieldMapper` and emit a deprecation if user is not opted-in.

Follow Up: 
* This allows clean up nullable check in `computeInsertExecutionOrder` and other places mentioned there in a comment.

### DIFF
--- a/docs/en/reference/attributes-reference.rst
+++ b/docs/en/reference/attributes-reference.rst
@@ -175,8 +175,9 @@ Optional parameters:
 -  **unique**: Boolean value to determine if the value of the column
    should be unique across all rows of the underlying entities table.
 
--  **nullable**: Determines if NULL values allowed for this column.
-    If not specified, default value is ``false``.
+-  **nullable**: Determines if NULL values allowed for this column.  If not
+   specified, default value is ``false``.  When using typed properties on entity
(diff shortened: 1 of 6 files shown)


docs/en/reference/attributes-reference.rst
```diff
@@ -175,8 +175,9 @@ Optional parameters:
 -  **unique**: Boolean value to determine if the value of the column
    should be unique across all rows of the underlying entities table.
 
--  **nullable**: Determines if NULL values allowed for this column.
-    If not specified, default value is ``false``.
+-  **nullable**: Determines if NULL values allowed for this column.  If not
+   specified, default value is ``false``.  When using typed properties on entity
+   class defaults to true when property is nullable.
```
greg0ire, line 180:
```suggestion
   class defaults to ``true`` when property is nullable.
```

PR #10985, opened 2023-10-10
Remove partial syntax in DQL

greg0ire:
It is no longer possible to use the `PARTIAL` keyword in a DQL query, or to artificially build an AST with a partial object expression. It is still possible to use the result set mapping API to build partial objects.


greg0ire:
Forbidding building partial objects will be attempted in a separate PR, it will require a substantial rework of the tests, they seem to heavily rely on that ability. A possible solution might be to build a factory, allowing us to specify just the properties that we care about. Note that building such objects through that API does not trigger any deprecation.

greg0ire:
Converting to draft as I've found more stuff to remove

UPGRADE.md
```diff
@@ -1,5 +1,13 @@
 # Upgrade to 3.0
 
+## BC BREAK: Partial objects are removed
+
+- The `PARTIAL` keyword no longer exists.
```
beberlei:
```suggestion
- The `PARTIAL` keyword in DQL no longer exists.
```

beberlei (dismissed):

beberlei (approved):

walva:
I wonder what was the reasoning behind this move. I use this feature time to time and I know ApiPlatform rely on it.

greg0ire:
The place to discuss that is https://github.com/doctrine/orm/issues/8471

PR #10986, opened 2023-10-10
Drop support for unserializing old serialization formats

greg0ire:


derrabus:
Should we keep the test for the 2.15 format? It's still the current one, isn't it?

mpdude (dismissed):

morozov (dismissed):

SenseException (dismissed):
> Should we keep the test for the 2.15 format? It's still the current one, isn't it?

Should it be renamed?

greg0ire:
> Should it be renamed?

I don't think so, if it does not change.

greg0ire:
Note that this removes txt files that I should have removed in my last merge up.

SenseException (approved):

PR #10992, opened 2023-10-10
Throw exceptions for cases where attributes are illegal on Embeddables.

### LINKED ISSUE #8763: Throw exception on unsupported mappings in Embeddable
For 3.0 mapping drivers should be more strict on which annotations/attributes are *not* allowed on properties / classes. For example currently you can define assocations on `Embeddable` but they are just ignored by drivers. For 2.x we could start deprecating this behavior and then throw a mapping exception in 3.0.

beberlei:
Only a very limited set of attributes is allowed on `Embedded` objects. All the other ones should throw exceptions.

Fixes #8763


mpdude (dismissed):
Looks good!

Do you plan to add tests and/or deprecations for this in 2.x?

beberlei:
@mpdude yes, i will work on the PR with deprecations for 2.x this morning.

PR #10995, opened 2023-10-11
Fix array parameters with DBAL 4

derrabus:
Calling `AbstractQuery::setParameter()` with an array parameter on DBAL 4 currently results in a `TypeError`. This PR fixes that. I'm going to backport the new tests to the 2.16.x branch.


morozov (approved):

greg0ire (approved):
Live long and prosper 🖖

PR #10998, opened 2023-10-11
Sever ties to common

greg0ire:
Companion PR: https://github.com/doctrine/orm/pull/10999

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,6 @@
         "composer-runtime-api": "^2",
         "ext-ctype": "*",
         "doctrine/collections": "^2.1",
-        "doctrine/common": "^3.3",
         "doctrine/dbal": "^3.6 || ^4",
         "doctrine/deprecations": "^0.5.3 || ^1",
         "doctrine/event-manager": "^1.2 || ^2",
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK: `Doctrine\ORM\Proxy\Autoloader` no longer extends `Doctrine\Common\Proxy\Autoloader`
+
+Make sure to use the former when writing a type declaration or an `instanceof` check.
+
 ## Minor BC BREAK: Changed order of arguments passed to `OneToOne`, `ManyToOne` and `Index` mapping PHP attributes
 
 To keep PHP mapping attributes consistent, order of arguments passed to above attributes has been changed
@@ -327,10 +331,6 @@ Use `Doctrine\Persistence\Mapping\Driver\StaticPHPDriver` and
 
 The second argument to `UnderscoreNamingStrategy::__construct()` was dropped,
 the strategy can no longer be unaware of numbers.
-
-## BC BREAK: Remove `Doctrine\ORM\Proxy\Autoloader`
-
-Use `Doctrine\Common\Proxy\Autoloader` instead.
 
 ## BC BREAK: Remove `Doctrine\ORM\Tools\DisconnectedClassMetadataFactory`
 
--- a/lib/Doctrine/ORM/Proxy/Autoloader.php
+++ b/lib/Doctrine/ORM/Proxy/Autoloader.php
@@ -1,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Proxy;
+
+use Closure;
+
+use function file_exists;
+use function ltrim;
+use function spl_autoload_register;
+use function str_replace;
+use function strlen;
+use function strpos;
+use function substr;
+
+use const DIRECTORY_SEPARATOR;
+
+/**
+ * Special Autoloader for Proxy classes, which are not PSR-0 compliant.
+ */
+final class Autoloader
+{
+    /**
+     * Resolves proxy class name to a filename based on the following pattern.
+     *
+     * 1. Remove Proxy namespace from class name.
+     * 2. Remove namespace separators from remaining class name.
+     * 3. Return PHP filename from proxy-dir with the result from 2.
+     *
+     * @psalm-param class-string $className
+     *
+     * @throws NotAProxyClass
+     */
+    public static function resolveFile(string $proxyDir, string $proxyNamespace, string $className): string
+    {
+        if (strpos($className, $proxyNamespace) !== 0) {
+            throw new NotAProxyClass($className, $proxyNamespace);
+        }
+
+        // remove proxy namespace from class name
+        $classNameRelativeToProxyNamespace = substr($className, strlen($proxyNamespace));
+
+        // remove namespace separators from remaining class name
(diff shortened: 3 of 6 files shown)


SenseException (dismissed):

lib/Doctrine/ORM/Proxy/ClassUtils.php
```diff
@@ -0,0 +1,54 @@
...
+
+use function get_class;
+use function strrpos;
+use function substr;
+
+/**
+ * Class-related functionality for objects that might or not be proxy objects
+ * at the moment.
+ *
+ * @internal
```
stof: Should this be an internal class ? The need to present the actual class name in error message or similar things also exists outside the ORM.
The class in doctrine/common was not internal.

Maybe this class should be moved to doctrine/persistence near the proxy interface.

greg0ire: > Maybe this class should be moved to doctrine/persistence near the proxy interface.

It might be best, here are the usages: https://github.com/search?q=org%3Adoctrine%20ClassUtils&type=code

- `doctrine/orm`, obviously
- `DoctrineBundle` also uses it
- `orientdb-odm` as well, whatever that is
- `phpcr-odm`

The mongoDB ODM seems to have moved away from it, they say so in their UPGRADE file. Seeing that reminded me of something @alcaeus did in `doctrine/persistence`: he contributed code very similar to what I'm contributing here, maybe we can reuse that: https://github.com/doctrine/persistence/pull/145

It would imply extracting the anonymous class holding the code publicly first, though.

stof: The drawback of the class name resolver is that it requires access to the Configuration class. But the advantage is that it allows to resolve things even for persistence projects that follow a different naming convention for their proxy classes (which is the case in the ODM due to their usage of ocramius/proxy-manager as the implementation)

greg0ire: Looking at doctrine/persistence#145 a bit more, it seems to me that the way to go would to have the `ProxyClassNameResolver` implementation live in the ORM. That way, `doctrine/persistence` would not have a public default implementation. I think it has a private one purely for BC reasons. Once we have that implementation in the ORM, we could inject it into the inside `ClassMetadataFactory::__construct()`, ensuring we use the same implementation in the code of `doctrine/persistence` as well, without relying on copy/paste.



> The drawback of the class name resolver is that it requires access to the Configuration class.

Looking at the usages inside `doctrine/orm`, the Configuration instance is available in most places, so I don't think that would be a big deal.

greg0ire: @stof I pushed a new implementation that relies on what I think you had in mind. Let me know if that would suit you and I will port something similar to 2.17.x (since there is no breaking change).

lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
```diff
@@ -35,6 +35,7 @@ abstract class AbstractEntityPersister implements CachedEntityPersister
 {
     protected UnitOfWork $uow;
     protected ClassMetadataFactory $metadataFactory;
+    protected DefaultProxyClassNameResolver $proxyClassNameResolver;
```
stof: shouldn't this use the interface as type ?

lib/Doctrine/ORM/Configuration.php
```diff
@@ -108,6 +109,20 @@ public function setProxyNamespace(string $ns): void
         $this->attributes['proxyNamespace'] = $ns;
     }
 
+    /**
+     * Gets the proxy class name resolver.
+     *
+     * @internal
+     */
+    public function getProxyClassNameResolver(): DefaultProxyClassNameResolver
```
stof: shouldn't this use the interface as type ?

lib/Doctrine/ORM/Proxy/DefaultProxyClassNameResolver.php
```diff
@@ -0,0 +1,46 @@
...
+class DefaultProxyClassNameResolver implements ProxyClassNameResolver
+{
+    /**
+     * Gets the real class name of an object (even if its a proxy).
+     *
+     * @param Proxy<T>|T $object
+     *
+     * @return class-string<T>
+     *
+     * @template T of object
```
stof: Do we need this ? Isn't it inherited from the interface ?

lib/Doctrine/ORM/Cache/DefaultCache.php
```diff
@@ -22,6 +22,7 @@ class DefaultCache implements Cache
 {
     private readonly UnitOfWork $uow;
     private readonly CacheFactory $cacheFactory;
+    private readonly DefaultProxyClassNameResolver $proxyClassNameResolver;
```
stof: I would use the interface as type

composer.json
```diff
@@ -24,7 +24,6 @@
         "composer-runtime-api": "^2",
         "ext-ctype": "*",
         "doctrine/collections": "^2.1",
-        "doctrine/common": "^3.3",
```
stof, line 27: The RunDqlCommand still uses it for now.

lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
greg0ire: The interface is not enough to fulfill the needs of the ORM: it provides one method which is use solely in `doctrine/persistence` and does not provide the other method (`resolveClass`).

stof: `->resolveClass($arg)` is just `->resolveClassName(get_class($arg))`. Do we need it ?

greg0ire: I can sure copy/paste that `get_class` everywhere. An alternative could be to introduce an ORM-specific interface called `ProxyClassResolver` and to use `ProxyClassNameResolver&ProxyClassResolver` as a type hint for the property stored in `Configuration`, and just `ProxyClassResolver` everywhere else.

greg0ire: @stof feedback addressed in https://github.com/doctrine/orm/pull/11008. I will reuse this PR for the `Autoloader` part, and to do the removal of `doctrine/common`.

beberlei:
I don't know if this must be configurable to end users, I prefer to hardcode the logic like it was before with `ClassUtils::getClass`.

greg0ire:
@beberlei I didn't provide a setter in the `Configuration` object, so it is not really configurable. I'm using `Configuration` as a convenient way to get that service everywhere I need it, without resorting to the singleton pattern, but maybe there is a better way?

derrabus (approved):

PR #10999, opened 2023-10-11
Undeprecate autoloader class

greg0ire:
We plan to sunset doctrine/common, and should move the `Autoloader` class
to `doctrine/orm`

Companion PR: https://github.com/doctrine/orm/pull/10998

PR #11000, opened 2023-10-12
Copy Debug class from doctrine/common

greg0ire:
This is an alternative to https://github.com/doctrine/orm/pull/10999/commits/4c8934f2539f5f700eb7a724f94af8d109057a66 (the other commit in the associated PR still stands).

This reduces our dependency to this shared library that now holds very little code we use.
The class has not been copied verbatim:
- Unused parameters and methods have been removed.
- The class is final and internal.
- Coding standards have been enforced, including enabling `strict_types`, which lead to casting a variable to string before feeding it to `explode()`.
- A bug found by static analysis has been addressed, where an INI setting obtained with `ini_get()` was compared with true, which is never returned by that function.

A follow-up of this PR could be to leverage Symfony's VarDumper component when available.

### DIFF
--- a/lib/Doctrine/ORM/Tools/Debug.php
+++ b/lib/Doctrine/ORM/Tools/Debug.php
@@ -1,0 +1,168 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Tools;
+
+use ArrayIterator;
+use ArrayObject;
+use DateTimeInterface;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Proxy\DefaultProxyClassNameResolver;
+use Doctrine\Persistence\Proxy;
+use stdClass;
+
+use function array_keys;
+use function count;
+use function end;
+use function explode;
+use function extension_loaded;
+use function get_class;
+use function html_entity_decode;
+use function ini_get;
+use function ini_set;
+use function is_array;
+use function is_object;
+use function ob_end_clean;
+use function ob_get_contents;
+use function ob_start;
+use function strip_tags;
+use function var_dump;
+
+/**
+ * Static class containing most used debug methods.
+ *
+ * @internal
+ *
+ * @link   www.doctrine-project.org
+ */
+final class Debug
+{
+    /**
+     * Private constructor (prevents instantiation).
+     */
+    private function __construct()
+    {
+    }
+
+    /**
+     * Prints a dump of the public, protected and private properties of $var.
+     *
+     * @link https://xdebug.org/
+     *
+     * @param mixed $var      The variable to dump.
+     * @param int   $maxDepth The maximum nesting level for object properties.
+     */
+    public static function dump($var, int $maxDepth = 2): string
+    {
+        $html = ini_get('html_errors');
+
+        if ($html !== '1') {
+            ini_set('html_errors', 'on');
+        }
+
+        if (extension_loaded('xdebug')) {
+            $previousDepth = ini_get('xdebug.var_display_max_depth');
+            ini_set('xdebug.var_display_max_depth', (string) $maxDepth);
+        }
+
+        try {
+            $var = self::export($var, $maxDepth);
+
+            ob_start();
+            var_dump($var);
+
+            $dump = ob_get_contents();
+
+            ob_end_clean();
+
+            $dumpText = strip_tags(html_entity_decode($dump));
+        } finally {
+            ini_set('html_errors', $html);
+
+            if (isset($previousDepth)) {
+                ini_set('xdebug.var_display_max_depth', $previousDepth);
+            }
+        }
+
+        return $dumpText;
+    }
+
+    /**
+     * @param mixed $var
+     *
+     * @return mixed
+     */
+    public static function export($var, int $maxDepth)
+    {
+        if ($var instanceof Collection) {
+            $var = $var->toArray();
+        }
+
+        if (! $maxDepth) {
+            return is_object($var) ? get_class($var)
+                : (is_array($var) ? 'Array(' . count($var) . ')' : $var);
+        }
+
+        if (is_array($var)) {
+            $return = [];
+
+            foreach ($var as $k => $v) {
+                $return[$k] = self::export($v, $maxDepth - 1);
+            }
+
+            return $return;
+        }
+
+        if (! is_object($var)) {
+            return $var;
+        }
+
+        $return = new stdClass();
+        if ($var instanceof DateTimeInterface) {
+            $return->__CLASS__ = get_class($var);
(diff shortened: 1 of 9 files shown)


tests/Doctrine/Tests/ORM/Tools/DebugTest.php
```diff
@@ -0,0 +1,146 @@
...
+        $var = Debug::export($it, 5);
+        self::assertContains('foobar', $var->__STORAGE__);
+    }
+
+    /**
+     * @param array<string, int> $expected
+     *
+     * @dataProvider provideAttributesCases
+     *
+     * @requires PHP < 8.1.0
```
derrabus: Why?

tests/Doctrine/Tests/ORM/Tools/DebugTest.php
```diff
@@ -0,0 +1,146 @@
...
+     */
+    public function testExportParentAttributes(TestAsset\ParentClass $class, array $expected): void
+    {
+        $actualRepresentation   = print_r($class, true);
+        $expectedRepresentation = print_r($expected, true);
+
+        $actualRepresentation   = substr($actualRepresentation, strpos($actualRepresentation, '('));
+        $expectedRepresentation = substr($expectedRepresentation, strpos($expectedRepresentation, '('));
+
+        self::assertSame($actualRepresentation, $expectedRepresentation);
```
greg0ire: note to self: it should be expected, actual

tests/Doctrine/Tests/ORM/Tools/DebugTest.php
greg0ire: Apparently, the representation differs. I can make the data provider output different orders based on the PHP version so that we continue testing this on 3.0

derrabus: Do the representations only differ in the order of the array keys? I think PHP 8 changed the order in which parent and child properties are dumped.

If that's the case, let's put the PHP 8.1 representation into the data provider and if we're on PHP 7, we `ksort()` both arrays before comparing them.

greg0ire: Yes, that's the only difference. I'll try what you are recommending.

greg0ire: `ksort` does not do the trick the order seems to be alphabetical, then public, protected, private :P: https://github.com/doctrine/orm/actions/runs/6500148405/job/17654937509?pr=11000#step:5:89

greg0ire: Alternatively, if we don't care, I could use https://docs.phpunit.de/en/10.4/assertions.html#assertequalscanonicalizing

derrabus: Yes, let's try that.

greg0ire: Actually that won't work: we're comparing strings (generated by `print_r`) here, not arrays. Besides, on 3.0.x, this canonicalization will be useless. As it stands, the code makes it more clear that something can be removed.

derrabus: All right.

derrabus (dismissed):

lib/Doctrine/ORM/Tools/Debug.php
```diff
@@ -0,0 +1,164 @@
...
+    public static function dump($var, int $maxDepth = 2): string
+    {
+        $html = ini_get('html_errors');
+
+        if ($html !== '1') {
+            ini_set('html_errors', 'on');
+        }
+
+        if (extension_loaded('xdebug')) {
+            ini_set('xdebug.var_display_max_depth', (string) $maxDepth);
```
stof, line 66: shouldn't we restore the ini settings in a finally block at the end of the method ?

lib/Doctrine/ORM/Tools/Debug.php
```diff
@@ -0,0 +1,164 @@
...
+        ob_start();
+        var_dump($var);
+
+        $dump = ob_get_contents();
+
+        ob_end_clean();
+
+        $dumpText = strip_tags(html_entity_decode($dump));
+
+        ini_set('html_errors', $html);
```
stof: this should probably be in a finally block so that the state reset is done also on failures

lib/Doctrine/ORM/Tools/Debug.php
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @return mixed
+     */
+    public static function export($var, int $maxDepth)
+    {
+        $return = null;
+        $isObj  = is_object($var);
+
+        if ($var instanceof Collection) {
+            $var = $var->toArray();
```
stof, line 99: It would be great to add a test exporting a Collection to cover this

lib/Doctrine/ORM/Tools/Debug.php
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * @param mixed $var
+     *
+     * @return mixed
+     */
+    public static function export($var, int $maxDepth)
+    {
+        $return = null;
+        $isObj  = is_object($var);
```
stof: this is used only once, so I would inline it in the usage

lib/Doctrine/ORM/Tools/Debug.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * @param mixed $var
+     *
+     * @return mixed
+     */
+    public static function export($var, int $maxDepth)
+    {
+        $return = null;
```
stof: this assignment is dead code as it is already re-assigned before any usage.

lib/Doctrine/ORM/Tools/Debug.php
```diff
@@ -0,0 +1,164 @@
...
+        $return = new stdClass();
+        if ($var instanceof DateTimeInterface) {
+            $return->__CLASS__ = get_class($var);
+            $return->date      = $var->format('c');
+            $return->timezone  = $var->getTimezone()->getName();
+
+            return $return;
+        }
+
+        $return->__CLASS__ = ClassUtils::getClass($var);
```
stof: How would you replace this in 3.0 if you don't port ClassUtils there ?

greg0ire: Converting to draft because of this comment. The Command that uses debug has  access to the entity manager, so if we validate the design in #10998 , then I will inject it here somehow.

derrabus (approved):

PR #11004, opened 2023-10-13
Remove useless check

dbu:
i was trying to understand the AttributeReader and stumbled over this bit.

intellij complains that we are not supposed to ever return RepeatableAttributeCollection according to the PHPDoc. reading the code, i think the additional isRepeatable check makes no sense, as we already checked above and errored if the property is repeatable.

this code is also in the 2.x branches, not sure which branch would be the right place to fix it.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php
+++ b/lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php
@@ -69,8 +69,7 @@ final class AttributeReader
             ));
         }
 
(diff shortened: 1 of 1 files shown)


greg0ire:
I would target 2.17.x, since this does not fix a bug, and is not a breaking change.

dbu:
doing that in #11006 - i guess you regularly merge 2.17.x to 3.x and we can close this MR here?

greg0ire:
Yes, that's what we (try to) do.

PR #11007, opened 2023-10-13
Address split of doctrine/common

greg0ire:
doctrine/common has been split in several packages. A lot of what was true about doctrine/common is true about doctrine/persistence today, so let us simply reuse the existing paragraphs and mention persistence instead of common.

### DIFF
--- a/docs/en/reference/architecture.rst
+++ b/docs/en/reference/architecture.rst
@@ -24,28 +24,34 @@ performance it is also recommended that you use APC with PHP.
 Doctrine ORM Packages
 -------------------
 
-Doctrine ORM is divided into three main packages.
+Doctrine ORM is divided into four main packages.
 
--  Common
--  DBAL (includes Common)
--  ORM (includes DBAL+Common)
+-  `Collections <https://www.doctrine-project.org/projects/doctrine-collections/en/stable/index.html>`_
+-  `Event Manager <https://www.doctrine-project.org/projects/doctrine-event-manager/en/stable/index.html>`_
+-  `Persistence <https://www.doctrine-project.org/projects/doctrine-persistence/en/stable/index.html>`_
+-  `DBAL <https://www.doctrine-project.org/projects/doctrine-dbal/en/stable/index.html>`_
+-  ORM (depends on DBAL+Persistence+Collections)
 
 This manual mainly covers the ORM package, sometimes touching parts
-of the underlying DBAL and Common packages. The Doctrine code base
+of the underlying DBAL and Persistence packages. The Doctrine code base
 is split in to these packages for a few reasons and they are to...
 
 
 -  ...make things more maintainable and decoupled
--  ...allow you to use the code in Doctrine Common without the ORM
-   or DBAL
+-  ...allow you to use the code in Doctrine Persistence and Collections
+  without the ORM or DBAL
 -  ...allow you to use the DBAL without the ORM
 
-The Common Package
-~~~~~~~~~~~~~~~~~~
+Collection, Event Manager and Persistence
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The Common package contains highly reusable components that have no
-dependencies beyond the package itself (and PHP, of course). The
-root namespace of the Common package is ``Doctrine\Common``.
+The Collection, Event Manager and Persistence packages contain highly
+reusable components that have no dependencies beyond the packages
+themselves (and PHP, of course). The root namespace of the Persistence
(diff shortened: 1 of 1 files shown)


docs/en/reference/architecture.rst
```diff
@@ -26,26 +26,26 @@ Doctrine ORM Packages
 
 Doctrine ORM is divided into three main packages.
 
--  Common
--  DBAL (includes Common)
--  ORM (includes DBAL+Common)
+-  Persistence
+-  DBAL
+-  ORM (includes DBAL+Persistence)
```
stof: I would say "depends on" rather than "includes"

docs/en/reference/architecture.rst
```diff
@@ -26,26 +26,26 @@ Doctrine ORM Packages
...
+-  ...allow you to use the code in Doctrine Persistence without the ORM
    or DBAL
 -  ...allow you to use the DBAL without the ORM
 
-The Common Package
+The Persistence Package
 ~~~~~~~~~~~~~~~~~~
 
-The Common package contains highly reusable components that have no
+The Persistence package contains highly reusable components that have no
```
stof: This sentence probably refers more to doctrine/collections than doctrine/persistence. Maybe it should be mentioned as well.

docs/en/reference/architecture.rst
```diff
@@ -26,26 +26,26 @@ Doctrine ORM Packages
 
 Doctrine ORM is divided into three main packages.
 
--  Common
--  DBAL (includes Common)
--  ORM (includes DBAL+Common)
+-  Persistence
+-  DBAL
```
stof: As those are about separate packages that have their own docs, should we have links to their docs ?

docs/en/reference/architecture.rst
```diff
@@ -24,28 +24,31 @@ performance it is also recommended that you use APC with PHP.
...
-Doctrine ORM is divided into three main packages.
+Doctrine ORM is divided into four main packages.
 
--  Common
--  DBAL (includes Common)
--  ORM (includes DBAL+Common)
+-  `Collections <https://www.doctrine-project.org/projects/doctrine-collections/en/stable/index.html>`_
+-  `Persistence <https://www.doctrine-project.org/projects/doctrine-persistence/en/stable/index.html>`_
+-  `DBAL <https://www.doctrine-project.org/projects/doctrine-dbal/en/stable/index.html>`_
+-  ORM (depends on DBAL+Persistence)
```
stof: and also on collections

docs/en/reference/architecture.rst
```diff
@@ -24,28 +24,31 @@ performance it is also recommended that you use APC with PHP.
...
 
 This manual mainly covers the ORM package, sometimes touching parts
-of the underlying DBAL and Common packages. The Doctrine code base
+of the underlying DBAL and Persistence packages. The Doctrine code base
 is split in to these packages for a few reasons and they are to...
 
 
 -  ...make things more maintainable and decoupled
--  ...allow you to use the code in Doctrine Common without the ORM
+-  ...allow you to use the code in Doctrine Persistence without the ORM
```
derrabus: … and Collections?

docs/en/reference/architecture.rst
```diff
@@ -24,28 +24,31 @@ performance it is also recommended that you use APC with PHP.
...
-~~~~~~~~~~~~~~~~~~
+The Collection and Persistence Packages
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The Common package contains highly reusable components that have no
-dependencies beyond the package itself (and PHP, of course). The
-root namespace of the Common package is ``Doctrine\Common``.
+The Collection and Persistence package contains highly reusable
+components that have no dependencies beyond the package itself (and PHP,
+of course). The root namespace of the Persistence package is
```
SenseException: Persistence has a dependency to the Event-Manager. I'm not sure what this paragraph was supposed to mean back then and now. That these packages are lightweight without a dependency chain?

stof: I guess this paragraph was true for common: at that time, event-manager was part of common as well.

greg0ire: I should use the plural, which will make the sentence "beyond the package themselves". If I add Persistence to the paragraph, this "themselves" will be vague enough that it can mean "each other" I think.

SenseException (approved):

docs/en/reference/architecture.rst
SenseException: Works for me :smile:

PR #11008, opened 2023-10-13
Implement proxy name resolver

greg0ire:
It is important to have the same implementation as used in `doctrine/persistence` without relying on copy/paste.
Besides, this decouples the ORM from `doctrine/common` a bit more.


greg0ire:
Closing in favor of #11009

PR #11009, opened 2023-10-14
Implement proxy name resolver

greg0ire:
It is important to have the same implementation as used in doctrine/persistence without relying on copy/paste.

Simplified alternative to #11008


beberlei (approved):
Yes! This keeps the surface of this code low and hidden from sight.

derrabus (approved):

PR #11012, opened 2023-10-15
Use native return type declarations in Debug class

greg0ire:


lib/Doctrine/ORM/Tools/Debug.php
```diff
@@ -144,12 +139,8 @@ public static function export($var, int $maxDepth)
     /**
      * Fill the $return variable with class attributes
      * Based on obj2array function from {@see https://secure.php.net/manual/en/function.get-object-vars.php#47075}
-     *
-     * @param object $var
-     *
-     * @return mixed
      */
-    private static function fillReturnWithClassAttributes($var, stdClass $return, int $maxDepth)
+    private static function fillReturnWithClassAttributes(object $var, stdClass $return, int $maxDepth): mixed
```
derrabus: Reading the code of the method, my expectation would be that we always return a `stdClass` instance.

greg0ire: I think you're not wrong :sweat_smile:

derrabus (approved):

PR #11013, opened 2023-10-15
Enforce native type declarations

greg0ire:
I tried to avoid changes in `tests`, so as to not unnecessarily widen the
gap between 2.x and 3.x.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -19,9 +19,6 @@
     <exclude-pattern>*/tests/Doctrine/Tests/ORM/Tools/Export/export/*</exclude-pattern>
 
     <rule ref="Doctrine">
-        <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingAnyTypeHint" />
-        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingReturnTypeHint"/>
-        <exclude name="SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingNativeTypeHint" />
         <exclude name="SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly.ReferencedGeneralException"/>
         <exclude name="SlevomatCodingStandard.ControlStructures.EarlyExit"/>
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming"/>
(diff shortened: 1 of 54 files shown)


phpcs.xml.dist
```diff
@@ -44,8 +40,17 @@
         <exclude-pattern>tests/Doctrine/Tests/ORM/Functional/ValueObjectsTest.php</exclude-pattern>
     </rule>
 
+    <rule ref="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingAnyTypeHint">
+        <exclude-pattern>tests/*</exclude-pattern>
+    </rule>
+
+    <rule ref="SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingNativeTypeHint">
+        <exclude-pattern>tests/*</exclude-pattern>
+    </rule>
```
derrabus: We should enforce those rules on our tests as well. Excluding (selected) fixtures, maybe.

greg0ire: It will make the merges up harder, and can be done after 3.0.0 is released. I don't remember what exactly what the plan is regarding the merges up and the release of 3.0.0, but I guess the merges up will stop or at least slow down at some point. I was planning on removing these rules then (they result in a huge diff).

derrabus: Sure, that was more like a future TODO. 👍

greg0ire: Oh right, OK 👍

tests/Doctrine/Tests/ORM/Hydration/ObjectHydratorTest.php
```diff
@@ -1021,18 +1024,21 @@ public function testCreatesProxyForLazyLoadingWithForeignKeys(): void
...
+            ProxyFactory::AUTOGENERATE_ALWAYS,
+        ) extends ProxyFactory {
+            public function getProxy(string $className, array $identifier): InternalProxy
+            {
+                TestCase::assertSame(ECommerceShipping::class, $className);
+                TestCase::assertSame(['id' => 42], $identifier);
+
+                return parent::getProxy($className, $identifier);
+            }
+        };
```
greg0ire, line 1041: @derrabus I pushed a new commit, it reworks this test and the one below, allowing to use a native return type declaration for `ProxyFactory::getProxy()`

SenseException (approved):
I flew over the changes and they seem to fit. :+1:

PR #11014, opened 2023-10-16
Fix pruning all collections in table for inherited elements.

DnwK98:
When we have two collections with inherited types in one entity, in this example it's `PetStore` containing `Cat` and `Dog`
```
Cat extends Pet
Dog extends Pet
// Cat and dog are stored in same table using entity_discriminator

PetStore
   Cat[] cats
   Dog[] dogs
```
And we clear one of them using `clear()`, second gets cleared too, due to DELETE statement without proper filter.

I think it's bug.

There is fix and test for it.

_Bypass before this is merged:
Instead of using_
```
collection.clear()
// or
collection = new ArrayCollection(newElements)
```
_load collection, and clear elements manually in your code_
```
foreach(collection as element){
    collection.removeElement(element)
}
foreach(newElements as newElement){
    collection.add(newElement)
}

### CODE REVIEW

**Comment by DnwK98** [comment]
Can I do something to help in review?

File: src/Persisters/Collection/OneToManyPersister.php [hunk 1]
```diff
@@ -183,6 +183,12 @@ private function deleteEntityCollection(PersistentCollection $collection): int
             $types[]      = PersisterHelper::getTypeOfColumn($joinColumn['referencedColumnName'], $sourceClass, $this->em);
         }
 
+        if ($targetClass->discriminatorValue && $targetClass->discriminatorColumn) {
```
beberlei, line 186: I am not sure if this will work for joined table inheritance as well, can you use joined inheritance instead of single table in the test please? Single table is a special case of joined table

DnwK98: This change doesn't affect joined inheritance, because its collections are removed using `deleteJoinedEntityCollection()` instead of `deleteEntityCollection()`

For joined inheritance this bug doesn't appear, because separate collections are filtered using JOIN to exact tables.

```
...
        $sql = $query->getSQL();
        //   SELECT c0_.id AS id_0 FROM Dog d1_
        //        INNER JOIN collection_with_inheritance_pet c0_ ON d1_.id = c0_.id WHERE d1_.petStore_id = ?

        assert(is_string($sql));
        $statement = 'INSERT INTO ' . $tempTable . ' (' . $idColumnList . ') ' . $sql;
        //   INSERT INTO collection_with_inheritance_pet_id_tmp (id)
        //        SELECT c0_.id AS id_0 FROM Dog d1_
        //            INNER JOIN collection_with_inheritance_pet c0_ ON d1_.id = c0_.id
        //        WHERE d1_.petStore_id = ?

        $parameters = array_values($sourceClass->getIdentifierValues($collection->getOwner()));
        $numDeleted = $this->conn->executeStatement($statement, $parameters);

        // 3) Delete records on each table in the hierarchy
        $classNames = array_merge($targetClass->parentClasses, [$targetClass->name], $targetClass->subClasses);

        foreach (array_reverse($classNames) as $className) {
            $tableName = $this->quoteStrategy->getTableName($this->em->getClassMetadata($className), $this->platform);
            $statement = 'DELETE FROM ' . $tableName . ' WHERE (' . $idColumnList . ')'
                . ' IN (SELECT ' . $idColumnList . ' FROM ' . $tempTable . ')';
            //   DELETE FROM collection_with_inheritance_pet
            //   WHERE (id) IN (SELECT id FROM collection_with_inheritance_pet_id_tmp)

            $this->conn->executeStatement($statement);
        }
...
```

DnwK98:
As it fixes bug maybe it should target 2.19.x

DnwK98:
Fixed in the meantime in https://github.com/doctrine/orm/pull/11501

PR #11015, opened 2023-10-16
Make phpdoc accurate

greg0ire:
When transforming these phpdoc types into native types, things break down. They are correct according to the EBNF, but in practice, there are so-called phase 2 optimizations that allow using `ConditionalPrimary`, `ConditionalFactor` and `ConditionalTerm` instances in places where `ConditionalExpression` is used.

Here is the relevant piece of code: https://github.com/doctrine/orm/blob/42af7cabb7496d77104132f9d164f9c74da6a629/lib/Doctrine/ORM/Query/SqlWalker.php#L2057-L2104

Once this PR is merged up, we should be able to use phpcbf to convert phpdoc to native types on 3.0.x without that causing any problem.

### DIFF
--- a/lib/Doctrine/ORM/Query/SqlWalker.php
+++ b/lib/Doctrine/ORM/Query/SqlWalker.php
@@ -1010,9 +1010,9 @@ class SqlWalker implements TreeWalker
     /**
      * Walks down a JoinAssociationDeclaration AST node, thereby generating the appropriate SQL.
      *
-     * @param AST\JoinAssociationDeclaration $joinAssociationDeclaration
-     * @param int                            $joinType
-     * @param AST\ConditionalExpression      $condExpr
+     * @param AST\JoinAssociationDeclaration                             $joinAssociationDeclaration
+     * @param int                                                        $joinType
+     * @param AST\ConditionalExpression|AST\Phase2OptimizableConditional $condExpr
      * @psalm-param AST\Join::JOIN_TYPE_* $joinType
      *
      * @return string
@@ -2048,7 +2048,7 @@ class SqlWalker implements TreeWalker
     /**
      * Walk down a ConditionalExpression AST node, thereby generating the appropriate SQL.
      *
-     * @param AST\ConditionalExpression $condExpr
+     * @param AST\ConditionalExpression|AST\Phase2OptimizableConditional $condExpr
      *
      * @return string
      *
@@ -2068,7 +2068,7 @@ class SqlWalker implements TreeWalker
     /**
      * Walks down a ConditionalTerm AST node, thereby generating the appropriate SQL.
      *
-     * @param AST\ConditionalTerm $condTerm
+     * @param AST\ConditionalTerm|AST\ConditionalFactor|AST\ConditionalPrimary $condTerm
      *
      * @return string
      *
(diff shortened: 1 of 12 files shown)


phpstan-baseline.neon
```diff
@@ -440,6 +440,11 @@ parameters:
 			count: 1
 			path: lib/Doctrine/ORM/Query/SqlWalker.php
 
+		-
+			message: "#^Parameter \\#1 \\$condTerm of method Doctrine\\\\ORM\\\\Query\\\\SqlWalker\\:\\:walkConditionalTerm\\(\\) expects Doctrine\\\\ORM\\\\Query\\\\AST\\\\ConditionalFactor\\|Doctrine\\\\ORM\\\\Query\\\\AST\\\\ConditionalPrimary\\|Doctrine\\\\ORM\\\\Query\\\\AST\\\\ConditionalTerm, Doctrine\\\\ORM\\\\Query\\\\AST\\\\Phase2OptimizedConditional given\\.$#"
```
greg0ire: PHPStan does not seem to have support for `psalm-inheritors` yet.

lib/Doctrine/ORM/Query/SqlWalker.php
```diff
@@ -2068,7 +2068,7 @@ public function walkConditionalExpression($condExpr)
     /**
      * Walks down a ConditionalTerm AST node, thereby generating the appropriate SQL.
      *
-     * @param AST\ConditionalTerm $condTerm
+     * @param AST\ConditionalTerm|AST\ConditionalFactor|AST\ConditionalPrimary $condTerm
```
greg0ire, line 2071: I've tried using the interface here instead, but that fails, presumably because Psalm does not manage to compute that that interface minus `AST\ConditionalTerm` is `AST\ConditionalFactor|AST\ConditionalPrimary`

lib/Doctrine/ORM/Tools/Pagination/WhereInWalker.php
```diff
@@ -96,10 +95,7 @@ public function walkSelectStatement(SelectStatement $AST)
                         ),
                     ]
                 );
-            } elseif (
-                $AST->whereClause->conditionalExpression instanceof ConditionalExpression
-                || $AST->whereClause->conditionalExpression instanceof ConditionalFactor
-            ) {
```
SenseException, line 102: Was this `elseif` superfluous that it could be changed into an `else`.

lib/Doctrine/ORM/Query/AST/Phase2OptimizedConditional.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Doctrine\ORM\Query\AST;
+
+/**
+ * Marks types that can be used in place of a ConditionalExpression as a phase
+ * 2 optimization.
+ *
+ * @psalm-inheritors ConditionalPrimary|ConditionalFactor|ConditionalTerm
+ */
+interface Phase2OptimizedConditional
```
SenseException: I understand that it should be declared as internal because it exists just as an interim, right?.

SenseException (changes requested):

lib/Doctrine/ORM/Tools/Pagination/WhereInWalker.php
greg0ire: I believe so, yes.

lib/Doctrine/ORM/Query/AST/Phase2OptimizedConditional.php
greg0ire: Not sure what you mean by interim (it should live forever), but you're right, it should be marked as internal.

SenseException: Then I misunderstood the interface's lifetime.

SenseException (dismissed):

PR #11018, opened 2023-10-17
Fix the support for enum types in the ResultSetMappingBuilder

### LINKED ISSUE #11017: ResultSetMappingBuilder does not properly support enumType
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.16.2

#### Summary

`Doctrine\ORM\Query\ResultSetMappingBuilder` is not aware of the `enumType` feature

#### Current behavior

When using `\Doctrine\ORM\Query\ResultSetMappingBuilder::addRootEntityFromClassMetadata` or `\Doctrine\ORM\Query\ResultSetMappingBuilder::addJoinedEntityFromClassMetadata` with an entity that uses `enumType` for some fields, the info about a field needing the enum conversion is not added automatically in the ResultSetMapping based on the metadata.

#### Expected behavior

The enum conversion should be added automatically in the ResultSetMapping based on the metadata, like other features defined in the metadata.

stof:
Fixes https://github.com/doctrine/orm/issues/11017

### DIFF
--- a/lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
+++ b/lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
@@ -154,6 +154,11 @@ class ResultSetMappingBuilder extends ResultSetMapping
             }
 
             $this->addFieldResult($alias, $columnAlias, $propertyName);
+
+            $enumType = $classMetadata->getFieldMapping($propertyName)['enumType'] ?? null;
+            if (! empty($enumType)) {
+                $this->addEnumResult($columnAlias, $enumType);
+            }
         }
 
         foreach ($classMetadata->associationMappings as $associationMapping) {
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH11017/GH11017Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH11017/GH11017Test.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket\GH11017;
+
+use Doctrine\ORM\AbstractQuery;
+use Doctrine\ORM\Query\ResultSetMappingBuilder;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+use function sprintf;
+
+/**
+ * @requires PHP 8.1
+ */
+class GH11017Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH11017Entity::class,
+        ]);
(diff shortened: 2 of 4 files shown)


greg0ire:
The failures look unrelated, but I think they actually are somehow: PHPUnit is happy with `origin/2.16.x` right now.

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
```diff
@@ -154,6 +154,11 @@ protected function addAllClassFields($class, $alias, $columnAliasMap = [])
             }
 
             $this->addFieldResult($alias, $columnAlias, $propertyName);
+
+            $enumType = $classMetadata->getFieldMapping($columnName)['enumType'] ?? null;
```
greg0ire: This is the line that is causing the issue. Getting the field mapping seems to have unwanted consequences.

greg0ire: Hang on… you're calling a field mapping with a column name here. This is why these tests are failing.

greg0ire:
```suggestion
            $enumType = $classMetadata->getFieldMapping($propertyName)['enumType'] ?? null;
```

This leaves one failure (so no errors), with the test you're introducing.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH11017/GH11017Test.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $tableName = $this->_em->getClassMetadata(GH11017Entity::class)->getTableName();
+        $sql       = sprintf('SELECT %s FROM %s e WHERE id = :id', $rsm->generateSelectClause(), $tableName);
+
+        $query = $this->_em->createNativeQuery($sql, $rsm)
+            ->setParameter('id', $entity1->id);
+
+        $entity1Reloaded = $query->getSingleResult(AbstractQuery::HYDRATE_ARRAY);
+        self::assertNotNull($entity1Reloaded);
+        self::assertSame($entity1->field, $entity1Reloaded['field']);
```
greg0ire, line 47:
```suggestion
        self::assertSame($entity1->field->value, $entity1Reloaded['field']);
```

lib/Doctrine/ORM/Query/ResultSetMappingBuilder.php
stof: indeed, I messed up (and my test was running fine because it was a case where both were equal)

stof:
@greg0ire this is fixed now

greg0ire (approved):

greg0ire:
Thanks @stof !

PR #11026, opened 2023-10-25
Hydrate AbstractSqlExecutor::sqlStatements

greg0ire:
This test caused deprecations because
`AbstractSqlExecutor::_sqlStatements`, a property that no longer exists since it was renamed was hydrated instead.


greg0ire:
@derrabus when working on this yesterday, I was happy to see I managed to fix the breaking test, and then I saw other test broke, involving classes inheriting from the one I modified. I see 3 possibilites now:

1. Removing support for unserializing 2.x format
2. Fixing the newly broken tests, and accepting the fact that executors outside this repository might break as well if they have properties.
3. Doing something on 2.x, maybe with `__sleep` / `__serialize` so that we store forward-compatible representations, in addition to 1.

What do you think?

greg0ire:
Alternate PR: https://github.com/doctrine/orm/pull/11027

PR #11027, opened 2023-10-26
Make serialized SQL executors forward compatible

greg0ire:
The idea here is that instead of having a backward compatibility layer
in the next major branch, we can have a forward compatibility layer in
this branch.

Closes #11026


lib/Doctrine/ORM/Query/Exec/AbstractSqlExecutor.php
```diff
@@ -60,4 +72,15 @@ public function removeQueryCacheProfile()
      * @return Result|int
      */
     abstract public function execute(Connection $conn, array $params, array $types);
+
+    /** @return list<string> */
+    public function __sleep(): array
+    {
+        return array_diff_key(array_keys((array) $this), ['*_sqlStatements']);
```
greg0ire: This does not seem to work for old versions of PHP… or does it have to do with `VarExporter`? I still didn't quite get how it comes into play for https://github.com/doctrine/orm/blob/609647a51a1b28a674de270204a5cea3af57d554/lib/Doctrine/ORM/Query/ParserResult.php#L143-L147

greg0ire: :thinking: why did I use `array_diff_key` here… makes no sense…

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
```diff
@@ -33,12 +34,19 @@ public function testSerializeParserResult(): void
 
         $parserResult = self::parseQuery($query);
         $serialized   = serialize($parserResult);
+        $this->assertSame(rtrim(file_get_contents(__DIR__ . '/ParserResults/single_select_3_0_0.txt'), "\n"), $serialized);
```
greg0ire: This fails because of how `CompanyEmployee.startDate` results in different case for the SQL alias depending on the platform. With PostgreSQL, it seems to result in lowercase, not sure why. I will instead check that there is no occurrence of `_sqlStatements`

lib/Doctrine/ORM/Query/Exec/AbstractSqlExecutor.php
```diff
@@ -60,4 +73,22 @@ public function removeQueryCacheProfile()
...
+    public function __sleep(): array
+    {
+        /* Two reasons for this:
+           - we do not need to serialize the deprecated property, we can
+             rebuild the reference to the new property in __wakeup()
+           - not having the legacy property in the serialized data means the
+             serialized representation becomes compatible with 3.0.x, meaning
+             there will not be a deprecation warning about a missing property
+             when unserializing data */
+        return array_values(array_diff(array_keys((array) $this), ["\0*\0_sqlStatements"]));
```
greg0ire, line 87: This weirdness can be understood by reading https://www.phpinternalsbook.com/php5/classes_objects/serialization.html

tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
```diff
@@ -33,12 +34,23 @@ public function testSerializeParserResult(): void
...
         $this->assertInstanceOf(ParserResult::class, $unserialized);
         $this->assertInstanceOf(ResultSetMapping::class, $unserialized->getResultSetMapping());
         $this->assertEquals(['name' => [0]], $unserialized->getParameterMappings());
         $this->assertInstanceOf(SingleSelectExecutor::class, $unserialized->getSqlExecutor());
+        $this->assertIsString($unserialized->getSqlExecutor()->getSqlStatements());
+
+        $r = new ReflectionProperty($unserialized->getSqlExecutor(), '_sqlStatements');
+        $r->setAccessible(true);
+
+        $this->assertIsString($r->getValue($unserialized->getSqlExecutor()));
```
SenseException: Even though `_sqlStatements` is referencing `sqlStatements` this test would be more understandable if you compare the contents of both properties.

```suggestion
        $this->assertSame($r->getValue($unserialized->getSqlExecutor()), $unserialized->getSqlExecutor()->getSqlStatements());
```

Maybe you want to create a separated test method for your changes instead of re-using `testSerializeParserResult()`. One tests the serialized ParserResult, the other your forward compatiility.

SenseException (changes requested):
Next to the review I did also a check of the occurrences of used `_sqlStatements` in case if some were forgotten, but you get anyone of those rascals. :grin:

lib/Doctrine/ORM/Query/Exec/AbstractSqlExecutor.php
SenseException: I did once some PHP shenanigans by creating private/protected stdClass properties just for fun. :joy:

derrabus:
Can you add a the new format to `ParserResultSerializationTest::provideSerializedSingleSelectResults()`?

SenseException (approved):

PhilETaylor:
upgraded from `doctrine/orm: '2.17.x-dev#16028e4fd3202e21cecbd87071e46b40f7eda8bf'`
to `doctrine/orm: '2.17.x-dev#f08159eb87ec2feb1aaf3a25bf138704a052edd7'`

and now code like this: 

```
return $this->createQueryBuilder('a')
            ->where('a.lastVisitDate > :d')
            ->andWhere('a.username != :un')
            ->setParameter('d', (new DateTime('now'))->modify($period))
            ->setParameter('un', "rabbit")
            ->orderBy('a.lastVisitDate', 'desc')
            ->setMaxResults($limit)
            ->getQuery()
            ->getResult();
```

results in Exception

```
Doctrine\DBAL\Connection::executeQuery(): Argument #1 ($sql) must be of type string, null given, called in /var/www/current/vendor/doctrine/orm/lib/Doctrine/ORM/Query/Exec/SingleSelectExecutor.php on line 33
```

greg0ire:
Thanks for reporting this, I will take a look.

greg0ire:
@PhilETaylor fixed

PhilETaylor:
Awesome - thanks.

alcohol:
@greg0ire you sure it was fixed? Our production environment just broke while rolling out with `2.17.0`. See https://github.com/doctrine/orm/issues/11063 -- from the comments it looks like I'm not the only one who ran into this also.

greg0ire:
@alcohol yes, I'm sure, see https://github.com/doctrine/orm/pull/11048, I added a test that covers what's reported here. What you see in #11063 must be slightly different I suppose.

PR #11039, opened 2023-11-03
Adds metadata field type and enumType validation against Entity property type

### LINKED ISSUE #11037: Enum field mapping validation
The recently properties types validation implemented in https://github.com/doctrine/orm/pull/10946 is not working properly for enum mappings:

>$ bin/console doctrine:schema:validate
>
>Mapping
>
> [FAIL] The entity-class App\Model\Product mapping is invalid:
> * The field 'App\Model\Product#status' has the property type 'App\Model\ProductStatus' that differs from the metadata field type 'string' returned by the 'string' DBAL type.

https://github.com/doctrine/orm/blob/16028e4fd3202e21cecbd87071e46b40f7eda8bf/lib/Doctrine/ORM/Tools/SchemaValidator.php#L384-L387

where `$propertyType` is a BackedEnum class and `$metadataFieldType` is `string` or `int`.

**Possible solution**

I'd add another check (just below that block) to validate the enum mapping, something like this worked for me:
```php
if (
    is_a($propertyType, BackedEnum::class, true)
    && isset($fieldMapping['enumType'])
    && $propertyType === $fieldMapping['enumType']
    && $metadataFieldType === get_debug_type($propertyType::cases()[0]->value)
) {
    return null;
}
```
It checks the property type against the `enumType` instead, but also checks if the BackedEnum variant (string or int) matches the metadata field type.

Does it make sense to you?

Affected versions: `2.17.x`+

yceruto:
Closes #11037

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\ORM\Tools;
 
+use BackedEnum;
 use Doctrine\DBAL\Types\AsciiStringType;
 use Doctrine\DBAL\Types\BigIntType;
 use Doctrine\DBAL\Types\BooleanType;
@@ -21,6 +22,7 @@ use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use ReflectionEnum;
 use ReflectionNamedType;
 
 use function array_diff;
@@ -37,6 +39,7 @@ use function count;
 use function get_class;
 use function implode;
 use function in_array;
+use function is_a;
 use function sprintf;
 
 use const PHP_VERSION_ID;
@@ -384,6 +387,23 @@ class SchemaValidator
                         // If the property type is the same as the metadata field type, we are ok
                         if ($propertyType === $metadataFieldType) {
                             return null;
+                        }
+
+                        if (
+                            is_a($propertyType, BackedEnum::class, true)
+                            && $metadataFieldType === (string) (new ReflectionEnum($propertyType))->getBackingType()
+                        ) {
+                            if (! isset($fieldMapping['enumType']) || $propertyType === $fieldMapping['enumType']) {
+                                return null;
+                            }
+
+                            return sprintf(
+                                "The field '%s#%s' has the property type '%s' that differs from the metadata enumType '%s'.",
+                                $class->name,
+                                $fieldName,
+                                $propertyType,
+                                $fieldMapping['enumType']
+                            );
                         }
 
                         return sprintf(
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH11037/GH11037Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH11037/GH11037Test.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket\GH11037;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Tools\SchemaValidator;
+use Doctrine\Tests\OrmTestCase;
+
+/**
+ * @requires PHP >= 8.1
+ */
+final class GH11037Test extends OrmTestCase
+{
+    /** @var EntityManagerInterface */
+    private $em;
+
+    /** @var SchemaValidator */
+    private $validator;
+
+    protected function setUp(): void
+    {
+        $this->em        = $this->getTestEntityManager();
+        $this->validator = new SchemaValidator($this->em);
+    }
+
+    public function testMetadataFieldTypeCoherentWithEntityPropertyType(): void
+    {
+        $class = $this->em->getClassMetadata(ValidEntityWithTypedEnum::class);
+        $ce    = $this->validator->validateClass($class);
+
+        self::assertEquals([], $ce);
+    }
+
+    public function testMetadataFieldTypeNotCoherentWithEntityPropertyType(): void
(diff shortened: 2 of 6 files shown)


tests/Doctrine/Tests/ORM/Functional/Ticket/GH11037/GH11037Test.php
```diff
@@ -0,0 +1,49 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket\GH11037;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Tools\SchemaValidator;
+use Doctrine\Tests\OrmTestCase;
+
+/**
+ * @requires PHP >= 8.1
```
yceruto, line 12: I've updated the PHP version from which Enums are supported

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -386,6 +389,24 @@ function (array $fieldMapping) use ($class): ?string {
                             return null;
                         }
 
+                        if (
+                            is_a($propertyType, BackedEnum::class, true)
+                            && $metadataFieldType === get_debug_type($propertyType::cases()[0]->value)
```
greg0ire: `get_debug_type` has a name that indicates it is probably not intended for this. Is there a way to do the same with the reflection API?

greg0ire: Also, why does the backing type matter? As soon as the property type enum is not the exact same as the mapping enum, it's an issue, isn't it?

yceruto: > get_debug_type has a name that indicates it is probably not intended for this. Is there a way to do the same with the reflection API?

Sure, using `ReflectionEnum` instead

yceruto: > Also, why does the backing type matter? As soon as the property type enum is not the exact same as the mapping enum, it's an issue, isn't it?

Still, we need to check that the backing type matches the mapping type. The checking order is as it's due to different validation message. I've added a test case to check that.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH11037/InvalidEntityWithTypedEnum.php
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * @Column(type="integer", enumType=StringEntityStatus::class)
+     */
+    protected StringEntityStatus $status1;
+
+    /**
+     * @Column(type="integer", enumType=StringEntityStatus::class)
+     */
+    protected IntEntityStatus $status2;
```
greg0ire, line 30: Here the enum is not the same for the property and the mapping. Are there cases where that is not a problem? If not, then there is no point in looking at the backing type, is there?

yceruto: True, there is no point in checking the backing type if the `enumType` doesn't match the enum class at that moment. Conversely, there is no point in checking the `enumType` if the mapping `type` doesn't match the backing type (both cases are being tested here).

Since this is an AND condition (forced by the early return approach), it's necessary to validate both the mapping `type` and the `enumType` against the property type, so I needed to choose one of them to check first, and I opted to check the mapping `type` against the backing type first because, if it fails, it returns the default error message.

Please let me know if anything remains unclear or if you have a preferred code in mind.

greg0ire: Thanks, I finally understand what I overlooked.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH11037/GH11037Test.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function testMetadataFieldTypeNotCoherentWithEntityPropertyType(): void
+    {
+        $class = $this->em->getClassMetadata(InvalidEntityWithTypedEnum::class);
+        $ce    = $this->validator->validateClass($class);
+
+        self::assertEquals(
+            [
+                "The field 'Doctrine\Tests\ORM\Functional\Ticket\GH11037\InvalidEntityWithTypedEnum#status1' has the property type 'Doctrine\Tests\ORM\Functional\Ticket\GH11037\StringEntityStatus' that differs from the metadata field type 'int' returned by the 'integer' DBAL type.",
+                "The field 'Doctrine\Tests\ORM\Functional\Ticket\GH11037\InvalidEntityWithTypedEnum#status2' has the property type 'Doctrine\Tests\ORM\Functional\Ticket\GH11037\IntEntityStatus' that differs from the metadata enumType 'Doctrine\Tests\ORM\Functional\Ticket\GH11037\StringEntityStatus' returned by the 'integer' DBAL type.",
```
greg0ire: I don't think that this is accurate. The DBAL has no knowledge of enums. The transformation from integer to enums does not happen in the DBAL type, but here: https://github.com/doctrine/orm/blob/609647a51a1b28a674de270204a5cea3af57d554/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php#L701-L710

Anyway, I think the message should not mention such implementation details: the fact that the contents of `enumType` differs from the property type is enough to tell that there is something wrong.

yceruto: I wasn't sure about it, thanks! 👍

greg0ire (approved):

greg0ire:

How to do that?

1. `git rebase -i origin/2.17.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

SenseException (approved):

greg0ire:
Thanks @yceruto !

PR #11044, opened 2023-11-05
Recommend SEQUENCE until doctrine/dbal 4 is released

greg0ire:
Using `IDENTITY` with `doctrine/dbal` 3 results in `SERIAL`, which is not recommended.

On 3.0.x, we should make a different recommendation based on the version of `doctrine/dbal`

Companion PR: https://github.com/doctrine/orm/pull/11045

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -34,7 +34,6 @@ use ReflectionException;
 
 use function assert;
 use function class_exists;
-use function constant;
 use function count;
 use function end;
 use function explode;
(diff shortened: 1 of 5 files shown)


derrabus (dismissed):

greg0ire:
Sorry, I forgot that the code was greatly simplified on 3.0.x, so now that the recommended and legacy no longer differ, I will apply the same simplifications that were applied by @beberlei in 2db1f76dee92961b76bf3ea747b0b958bd2c2329

lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -762,27 +745,25 @@ private function determineIdGeneratorStrategy(AbstractPlatform $platform): int
             }
         }
 
-        foreach (self::LEGACY_DEFAULTS_FOR_ID_GENERATION as $platformFamily => $strategy) {
+        foreach (self::NON_IDENTITY_DEFAULT_STRATEGY as $platformFamily => $strategy) {
             if (is_a($platform, $platformFamily)) {
-                $recommendedStrategyName = self::RECOMMENDED_STRATEGY[$platformFamily];
-                if ($strategy !== constant('Doctrine\ORM\Mapping\ClassMetadata::GENERATOR_TYPE_' . $recommendedStrategyName)) {
+                if ($platform instanceof Platforms\PostgreSQLPlatform || is_a($platform, 'Doctrine\DBAL\Platforms\PostgreSqlPlatform')) {
```
SenseException, line 750: What is `is_a($platform, 'Doctrine\DBAL\Platforms\PostgreSqlPlatform')` checking that's not already been checked?

greg0ire: There is a difference in case (`SQL` vs `Sql`). This allows supporting DBAL 2.

SenseException (approved):

PR #11045, opened 2023-11-05
Dynamically resolve AUTO to SEQUENCE or IDENTITY

greg0ire:
With DBAL 3.x, `IDENTITY` results in `SERIAL`.
With DBAL 4.x, it results in the standard `GENERATED BY DEFAULT AS IDENTITY`.

### DIFF
--- a/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
@@ -34,6 +34,7 @@ use function explode;
 use function in_array;
 use function is_a;
 use function is_subclass_of;
+use function method_exists;
 use function str_contains;
 use function strlen;
 use function strtolower;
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php
```diff
@@ -616,6 +617,13 @@ private function determineIdGeneratorStrategy(AbstractPlatform $platform): int
             }
         }
 
+        $nonIdentityDefaultStrategy = self::NON_IDENTITY_DEFAULT_STRATEGY;
```
SenseException, line 620: What's happening with `$nonIdentityDefaultStrategy` and where is it used? What am I missing?

greg0ire: Sorry, I forgot to use it below :facepalm:

greg0ire: I added tests.

SenseException (approved):

PR #11053, opened 2023-11-07
Handle ENUM Primary Key in ObjectHydrator.php

wucdbm:
Where `App\Core\Nomenclature\Country` is an ENUM used as a Primary Key of a relation, so it tries to cast it to string -- should be simple enough to reproduce, I an not allowed proper time to create a proper PR with tests and all

I've fixed it on our end by copying the source file and applying the same patch, I hope this helps

Error:
Object of class App\Core\Nomenclature\Country could not be converted to string

  at vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:1903
  at Doctrine\ORM\UnitOfWork->tryGetByIdHash(object(Country), 'App\\Core\\Entity\\Location\\CountryGroupCountry')
     (vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php:304)
  at Doctrine\ORM\Internal\Hydration\ObjectHydrator->getEntityFromIdentityMap('App\\Core\\Entity\\Location\\CountryGroupCountry', array('id' => object(Country)))
     (vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php:403)


wucdbm:
@derrabus I've no clue where/how to test this one, please advice

derrabus:
Create a functional test that reproduces your problem. We have plenty of those. Have a look at the existing tests and you should figure it out.

PR #11061, opened 2023-11-15
Fix calls to removed lock methods

derrabus:
`getReadLockSQL()` and `getWriteLockSQL()` have been removed from the DBAL. This PR restores the functionality in order to get the pipeline green again. I'll try to leverage the DBAL query builder instead when I find some time. 😓


derrabus:
I'm merging this so we have a green pipeline again.

jwage:
@derrabus should this be backported to 2.x so that we're not relying on deprecated DBAL methods in the ORM?

derrabus:
We can do that, yes.

jwage:
I will make a PR for 2.17.x

PR #11067, opened 2023-11-17
Drop support for Symfony 4

greg0ire:
It is unmaintained and makes contributions unnecessarily hard to do.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "doctrine/lexer": "^2",
         "doctrine/persistence": "^2.4 || ^3",
         "psr/cache": "^1 || ^2 || ^3",
-        "symfony/console": "^4.2 || ^5.0 || ^6.0 || ^7.0",
+        "symfony/console": "^5.0 || ^6.0 || ^7.0",
         "symfony/polyfill-php72": "^1.23",
         "symfony/polyfill-php80": "^1.16"
     },
@@ -46,9 +46,9 @@
         "phpunit/phpunit": "^7.5 || ^8.5 || ^9.6",
         "psr/log": "^1 || ^2 || ^3",
         "squizlabs/php_codesniffer": "3.7.2",
-        "symfony/cache": "^4.4 || ^5.4 || ^6.0",
-        "symfony/var-exporter": "^4.4 || ^5.4 || ^6.2",
-        "symfony/yaml": "^3.4 || ^4.0 || ^5.0 || ^6.0",
+        "symfony/cache": "^5.4 || ^6.0",
+        "symfony/var-exporter": "^5.4 || ^6.2",
+        "symfony/yaml": "^5.0 || ^6.0",
         "vimeo/psalm": "4.30.0 || 5.15.0"
     },
     "conflict": {
--- a/tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
+++ b/tests/Doctrine/Tests/ORM/Functional/ParserResultSerializationTest.php
@@ -13,7 +13,6 @@ use Doctrine\Tests\OrmFunctionalTestCase;
 use Generator;
 use ReflectionMethod;
 use ReflectionProperty;
-use Symfony\Component\VarExporter\Instantiator;
 use Symfony\Component\VarExporter\VarExporter;
 
 use function file_get_contents;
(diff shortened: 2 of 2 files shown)


greg0ire:
Ah… PHP 7.1, of course.

derrabus:
You can't do that without dropping PHP 7.1. 🤷🏻

greg0ire:
Maybe we should ?

derrabus:
These are the packagist statistics for the 2.16 branch (2.17 is too new, 2.15 looks similar).

Bildschirmfoto 2023-11-17 um 09 01 50

So, people who do upgrade their ORM, don't run PHP 7.1 anymore. And we don't need to care about people who don't upgrade their ORM anyway, I guess.

Let's bump to 7.4 and kick out DBAL 2 while we're at it?

Then again, all of this is what we've planned for ORM 3 anyway. Maybe we just flag issues related to Symfony 4 as won't fix and try to push ORM 3 out of the door instead?

greg0ire:
> Maybe we just flag issues related to Symfony 4 as won't fix

As you can see, it's more involving than setting a label on some issues. Sometimes, we have to go out of our way to make sure a test does not run when using Symfony 4. And I think the same happens for PHP 7.1.

derrabus:
I know. I can live with both solutions. Pick one. 🙂

greg0ire:
> Let's bump to 7.4 and kick out DBAL 2 while we're at it?

I'd say let's try to notice when supporting an unmaintained version becomes a hindrance, and drop exactly what causes the hindrance, otherwise @beberlei is going to send counter-PRs because stuff actually still works with PHP 7.2, for instance. So to me the balance we need to find is between "it works" and "it makes contributions and maintenance" painful.

derrabus:
> I'd say let's try to notice when supporting an unmaintained version becomes a hindrance

But that's not really the case here. The only real Symfony dependencies we have are Console and maybe YAML and Cache. Neither of them causes trouble.

If you're referring to the serialization issues with VarExporter 4.4: We don't use anything from VarExporter 4.4. And we don't tell anyone to even use VarExporter to serialize our stuff. If we merge any compatibility fixes for VarExporter 4.4, it's because we're being nice and not because we're identifying the incompatibility to a EOL library that we don't even depend on as a bug. Dropping support for Console 4.4 (which is _effectively_ what your PR does) would not change that policy for me.

greg0ire:
I was indeed referring to the serialization issues with VarExporter

> We don't use anything from VarExporter 4.4

:thinking: we use the lazy ghosts, do we not? We do depend on that lib, it's just an optional dependency, and it becomes a regular dependency on 3.0.x

I'm dropping support for Symfony 4 because in general, I think our policy is to drop a version of Symfony accross all Symfony Components.

derrabus:
> > We don't use anything from VarExporter 4.4
> 
> 🤔 we use the lazy ghosts, do we not?

Yes, from VarExporter >= 6.2. VarExporter 4.4 does not have lazy ghosts. It's a neutral package for us. We don't actively support it because the is nothing to support.

This PR drops support for Console 4.4 (which does not cause us trouble) and does not even prevent people from using VarExporter 4.4 on ORM classes. It fixes no problem that we actually have.

I see two options:

* Get ORM 3 out of the door soon and leave ORM 2 as it is. Feature development will continue on 3.1.x with shiny new dependencies.
* Do dependency housekeeping on ORM 2: PHP >= 7.4, DBAL >= 3.7, Symfony >= 5.4. Given that we will maintain ORM 2 for a while after 3.0.0, this might be a reasonable thing to do anyway.

greg0ire:
Ok I see why the 4.4 was important in "VarExporter 4.4" now.

> It fixes no problem that we actually have.

I'd argue we have needlessly complex tests and contributions, and that it fixes that. So, no end user problem fixed, but I think we should take contributors and maintainers into account.

> Do dependency housekeeping on ORM 2: PHP >= 7.4, DBAL >= 3.7, Symfony >= 5.4. Given that we will maintain ORM 2 for a while after 3.0.0, this might be a reasonable thing to do anyway.

If @beberlei signs off on this I'm more than happy to go with that solution.

composer.json
```diff
@@ -34,7 +34,7 @@
         "doctrine/lexer": "^2",
         "doctrine/persistence": "^2.4 || ^3",
         "psr/cache": "^1 || ^2 || ^3",
-        "symfony/console": "^4.2 || ^5.0 || ^6.0 || ^7.0",
+        "symfony/console": "^5.0 || ^6.0 || ^7.0",
```
Chris53897, line 37: I guess this line (and for symfony/yaml) can be changed to `^5.4 || ^6.0 || ^7.0` to only support LTS Versions?

greg0ire: yeah I guess

greg0ire:
Clear no from Benjamin.

PR #11076, opened 2023-11-20
Fix JSON mapping linting against subset of builtin types

### LINKED ISSUE #11072: JSON field mapping validation
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.17

#### Summary

After #10946 when linting entities started getting:

> The field 'App\Entity\Entity#value' has the property type 'mixed' that differs from the metadata field type 'array' returned by the 'json' DBAL type.
> The field 'App\Entity\AnotherEntity#value' has the property type 'int' that differs from the metadata field type 'array' returned by the 'json' DBAL type.

JSON is not only array, but also can be a string or a number.

#### Current behavior

Wrong validation

#### How to reproduce

Declare json columns with other than array type.

#### Expected behavior

No errors

norkunas:
Closes #11072

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -371,7 +371,7 @@ class SchemaValidator
                         }
 
                         // If the property type is not a named type, we cannot check it
-                        if (! ($propertyType instanceof ReflectionNamedType)) {
+                        if (! ($propertyType instanceof ReflectionNamedType) || $propertyType->getName() === 'mixed') {
                             return null;
                         }
 
@@ -404,6 +404,13 @@ class SchemaValidator
                                 $propertyType,
                                 $fieldMapping['enumType']
                             );
+                        }
+
+                        if (
+                            $fieldMapping['type'] === 'json'
+                            && in_array($propertyType, ['string', 'int', 'float', 'bool', 'true', 'false', 'null'], true)
+                        ) {
+                            return null;
                         }
 
                         return sprintf(
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH11072/GH11072EntityBasic.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH11072/GH11072EntityBasic.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket\GH11072;
+
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+
+/**
+ * @Entity
+ */
+class GH11072EntityBasic
+{
+    /**
+     * @Id
+     * @Column(type="integer")
+     * @GeneratedValue
+     * @var int
+     */
+    public $id;
+
+    /** @Column(type="json") */
+    public string $jsonString = 'test';
+
+    /** @Column(type="json") */
+    public int $age = 99;
+
+    /** @Column(type="json") */
+    public float $score = 0.0;
+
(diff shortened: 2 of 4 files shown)


greg0ire (dismissed):

SenseException (dismissed):

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -406,6 +406,13 @@ function (array $fieldMapping) use ($class): ?string {
                             );
                         }
 
+                        if (
+                            $fieldMapping['type'] === 'json'
+                            && in_array($propertyType, ['string', 'int', 'float', 'bool', 'true', 'false', 'null'], true)
```
janatjak, line 411: Can you please add `mixed` ?

example
```php
    /** @Column(type="json") */
    public mixed $data = null;
```

janatjak:
```suggestion
                            && in_array($propertyType, ['string', 'int', 'float', 'bool', 'true', 'false', 'mixed', 'null'], true)
```

norkunas: > Can you please add `mixed` ?
> 
> example
> 
> ```
>     /** @Column(type="json") */
>     public mixed $data = null;
> ```

sure, added

derrabus:
How does this fix play with union types? For example:

```php
#[Column(type: 'json')]
private array|bool|null $data;
```

norkunas:
> How does this fix play with union types? For example:
> 
> ```
> #[Column(type: 'json')]
> private array|bool|null $data;
> ```

It is not linted, because it's already ignored here:
https://github.com/doctrine/orm/blob/93d205797bc0bba76861ae4f50566c3194d4e75e/lib/Doctrine/ORM/Tools/SchemaValidator.php#L373-L376

derrabus:
Ah, perfect. 😎

tests/Doctrine/Tests/ORM/Functional/Ticket/GH11072/GH11072EntityBasic.php
```diff
@@ -0,0 +1,42 @@
...
+    public float $score = 0.0;
+
+    /** @Column(type="json", nullable=true) */
+    public ?bool $trinary = null;
+
+    /** @Column(type="json") */
+    public array $metadata = [];
+
+    /** @Column(type="json") */
+    public mixed $anything;
```
derrabus: This won't work on PHP 7. Let's move it to `GH11072EntityAdvanced`?

norkunas: If it's ok for you to run it only on 8.2 then I could, but locally now added a intermediary for the `mixed` type which would run on >= 8.0

derrabus: > If it's ok for you to run it only on 8.2

I think this is fine.

derrabus:
> It is not linted, because it's already ignored here:
> 
> https://github.com/doctrine/orm/blob/93d205797bc0bba76861ae4f50566c3194d4e75e/lib/Doctrine/ORM/Tools/SchemaValidator.php#L373-L376

Actually, I think we should extend this condition to also include `mixed`. Performing this validation on `mixed` properties is pretty pointless, given that `mixed` is a union of every value type known to PHP. We're about to build a special case for JSON on `mixed`, but I think this applies to more or less any DBAL type, built-in or custom.

norkunas:
Updated to always ignore mixed type

norkunas:
Can we get this merged? :slightly_smiling_face:

greg0ire (approved):

derrabus (approved):

norkunas:
Thank you

gnutix:
@derrabus @greg0ire Would you mind releasing a 2.17.2 version with this fix ? :pray:

PR #11082, opened 2023-11-23
Do not defer eager collection loading when in iteration context

### LINKED ISSUE #11081: Collections data loaded twice when using EAGER loading in combination with iterables
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.17.0/2.17.1

#### Summary

With #8391 optimisations where added with relation to eager loading collections. However, the eager load of the collection is not triggered when using an iterable result, causing the collection to be hydrated twice when actually accessed.

#### Current behavior

When an uninitialised collection within an entity coming from a iterable result is hit, it will trigger its load. However, this load also triggers the unit of work to process the stored eagerly loaded collection to load, resulting in the collection being loaded twice: it will contain duplicate items.

#### How to reproduce

See the added test, and proposed fix in #11082.

#### Expected behavior

The collection should not contains each item twice. I believe the eager load for the collection should be skipped by the UoW as it is already being loaded at that time.

bobvandevijver:
The test shows that eager collection loading is broken when used in conjunction with iterating over a query result. I also added a proposed fix to not defer eager collection loading when the iteration hint has been set.

Fixes #11081

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -3104,9 +3104,10 @@ EXCEPTION
                     $reflField->setValue($entity, $pColl);
 
                     if ($hints['fetchMode'][$class->name][$field] === ClassMetadata::FETCH_EAGER) {
-                        if ($assoc['type'] === ClassMetadata::ONE_TO_MANY) {
+                        $isIteration = isset($hints[Query::HINT_INTERNAL_ITERATION]) && $hints[Query::HINT_INTERNAL_ITERATION];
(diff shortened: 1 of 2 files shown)


bobvandevijver:
@beberlei Any chance you can take a look at this? This is blocking our upgrade to 2.17

bobvandevijver:
@greg0ire / @beberlei This is still blocking and quite hefty bug for iterators.. any chance this can be looked at?

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @bobvandevijver !

PR #11085, opened 2023-11-24
XmlDriver: remove dead code

staabm:
the variable `$mappings` is always containing a empty-array and therefore the code in question is dead


derrabus:
Thank you. Please make PHPStan happy.

derrabus (approved):

PR #11090, opened 2023-11-28
Failed getting entity with fetch eager property

dbannik:
Here, I provided tests for reproduction but without the solution to the problem itself.

Diff **WithoutFetchEager\AbstractRemoveControl** and **WithFetchEager\AbstractRemoveControl**
image


lib/Doctrine/ORM/UnitOfWork.php
```diff
@@ -3187,6 +3187,10 @@ private function eagerLoadCollections(array $collections, array $mapping): void
             foreach ($found as $targetValue) {
                 $sourceEntity = $targetProperty->getValue($targetValue);
 
+                if ($sourceEntity === null) {
+                    return;
```
dbannik: Exclude collection initialization if one of the collection elements has not completed full hydration

dbannik:
test passed. please check my fix

beberlei (dismissed):
I didn't think about this case when working on https://github.com/doctrine/orm/pull/8391

The test and fix are valid.

I cleaned it up a little.

greg0ire (approved):

greg0ire:
Thanks @dbannik and @beberlei !

PR #11104, opened 2023-12-04
Fix method name in code example

flack:
As discussed in https://github.com/doctrine/orm/pull/9961#discussion_r1414403326


derrabus (approved):

derrabus:
Thank you!

PR #11105, opened 2023-12-05
Deprecate `Join::ON` ?

HypeMC:
I've been wondering about this for years, what is the purpose of the `Join::ON` constant? This constant seems completely useless, as the DQL parser doesn't even support the `ON` keyword in a join:

```php
$fooRepository->createQueryBuilder('f')
    ->innerJoin('f.bar', 'b', Join::ON, 'b.condition = 1')
    ->getQuery()
    ->execute()
;
```
```
[Syntax Error] line 0, col 51: Error: Expected end of string, got 'ON'
```

I haven't found any issues related to this, nor any explanations regarding the original idea behind this constant.

Also, if it's decided to deprecate and remove this constant, the `$conditionType` parameters of the `QueryBuilder::*Join()` methods and `Expr\Join` class will become redundant since currently, the only two options are `WITH` and `ON`. Therefore, it might be a good idea to consider deprecating them as well.

https://github.com/doctrine/orm/blob/23d36c0d5293ac796f35850b4f082aef60c6fc52/lib/Doctrine/ORM/QueryBuilder.php#L1028-L1052

### DIFF
--- a/src/Query/Expr/Join.php
+++ b/src/Query/Expr/Join.php
@@ -16,6 +16,7 @@ class Join
     public const INNER_JOIN = 'INNER';
     public const LEFT_JOIN  = 'LEFT';
 
+    /** @deprecated This constant is deprecated and will be removed in Doctrine ORM 3.0. */
     public const ON   = 'ON';
     public const WITH = 'WITH';
 


derrabus:
Sorry that we haven't picked this up yet.

Does that mean that `Join::WITH` is the only valid value we can pass as the third argument of `innerJoin()`? If so, does it make sense to maintain a parameter with only a single valid value? Can we also deprecate the `Join::WITH` constant?

HypeMC:
> Sorry that we haven't picked this up yet.
> 
> Does that mean that `Join::WITH` is the only valid value we can pass as the third argument of `innerJoin()`? If so, does it make sense to maintain a parameter with only a single valid value? Can we also deprecate the `Join::WITH` constant?

@derrabus Yes, that's right, `Join::WITH` is the only valid value. IMO, it would make sense to deprecate the parameter altogether, as it's currently useless.

HypeMC:
I'm still willing to work on this, so if there's interest, it'd be great if it could be reopened.

PR #11109, opened 2023-12-06
Original entity data resolves inverse 1-1 joins

### LINKED ISSUE #11108: Loading inverse side of a one-to-one fails if target entity has an association id
### Bug Report

An inverse one-to-one is resolved with a separate query in EntityPersister->loadOneToOneEntity, which is called by UnitOfWork immediately after entity construction. If this is not the owning side of the association, a freshly constructed entity is provided as the source with the assumption that sufficient data is available in the entity to resolve the association.

The problem is that this is simply not true for an entity that uses an association as its identifier. In this case, the identifying association has not been initialized, so the entity has exactly zero useful information in it. This throws in loadOneToOneEntity indicating that the join column is not mapped.

The result is that the owning side of the relationship needs to be pulled into the UoW before the inverse side. Obviously, I should be able to query either object without failure

|    Q        |   A
|------------ | ------
| BC Break    | no (assuming BC means _Backwards Compatibility_, which is NO (_Not Obvious_))
| Version     | 2.17.1 (ongoing)

#### Summary

Entity identified with an association cannot be loaded if it is the inverse side of a non-identifying one-to-one relationship.

#### Current behavior

Exception is thrown indicating that the join column cannot be identified.

#### How to reproduce

Snippets pulled from associated pull request (which does not fail because it includes a fix). Any attempt to directly retrieve an InverseSide (findOneBy, getResult, etc.) that is not already cached will throw.
```php
/**
 * @Entity()
 * @Table(name="one_to_one_inverse_side_assoc_id_load_inverse")
 */
class InverseSide
{
    /**
     * Associative id (owning identifier)
     *
     * @var InverseSideIdTarget
     * @Id()
     * @OneToOne(targetEntity=InverseSideIdTarget::class, inversedBy="inverseSide")
     * @JoinColumn(nullable=false, name="associativeId")
     */
    public $associativeId;

    /**
     * @var OwningSide
     * @OneToOne(targetEntity=OwningSide::class, mappedBy="inverse")
     */
    public $owning;
}

/**
 * @Entity()
 * @Table(name="one_to_one_inverse_side_assoc_id_load_owning")
 */
class OwningSide
{
    /**
     * @var string
     * @Id()
     * @Column(type="string", length=255)
     * @GeneratedValue(strategy="NONE")
     */
    public $id;

    /**
     * Owning side
     *
     * @var InverseSide
     * @OneToOne(targetEntity=InverseSide::class, inversedBy="owning")
     * @JoinColumn(name="inverse", referencedColumnName="associativeId")
     */
    public $inverse;
}

/**
 * @Entity()
 * @Table(name="one_to_one_inverse_side_assoc_id_load_inverse_id_target")
 */
class InverseSideIdTarget
{
    /**
     * @var string
     * @Id()
     * @Column(type="string", length=255)
     * @GeneratedValue(strategy="NONE")
     */
    public $id;

    /**
     * @var InverseSide
     * @OneToOne(targetEntity=InverseSide::class, mappedBy="associativeId")
     */
    public $inverseSide;
}
```
#### Expected behavior

The expected behavior is that I can directly retrieve OwningSide and InverseSide in a symmetric fashion.

#### Approach to fix
The exception indicates that there is no field/column mapping for the associativeId field because it is used in association join column, so the first attempt to this added code to resolve the column. This was ultimately fruitless because even though the field was resolved, its contents are null because the owning relationship for the identifier has not been processed at this point.

The sourceEntity instance passed to loadOneToOneEntity is created immediately before this call, which moves data from the provided column-keyed array to the entity columns, with associations populated afterward. This means that we still have the original entity data array. I simply added a sourceEntityData argument to loadOneToOneEntity and provided this initial data. If the non-owning side cannot find data in the entity then it uses the original data array (which contains a superset of the entity data at this point) to form the query. This is a straightforward fix that solves the problem.

mcurland:
If the source entity for an inverse (non-owning) 1-1 relationship is identified by an association then the identifying association may not be set when an inverse one-to-one association is resolved. This means that no data is available in the entity to resolve the needed column value for the join query.

Provide the original entity data for the source entity as a fallback to resolve the query conditions.

The new test will fail if the fix is removed (comment out lines 840-847 in BasicEntityPersister.php).

Fixes #11108

### DIFF
--- a/src/Persisters/Entity/BasicEntityPersister.php
+++ b/src/Persisters/Entity/BasicEntityPersister.php
@@ -832,17 +832,42 @@ class BasicEntityPersister implements EntityPersister
 
         $computedIdentifier = [];
 
+        /** @var array<string,mixed>|null $sourceEntityData */
+        $sourceEntityData = null;
+
         // TRICKY: since the association is specular source and target are flipped
         foreach ($owningAssoc['targetToSourceKeyColumns'] as $sourceKeyColumn => $targetKeyColumn) {
             if (! isset($sourceClass->fieldNames[$sourceKeyColumn])) {
-                throw MappingException::joinColumnMustPointToMappedField(
-                    $sourceClass->name,
-                    $sourceKeyColumn
-                );
-            }
+                // The likely case here is that the column is a join column
+                // in an association mapping. However, there is no guarantee
+                // at this point that a corresponding (generally identifying)
+                // association has been mapped in the source entity. To handle
+                // this case we directly reference the column-keyed data used
+                // to initialize the source entity before throwing an exception.
+                $resolvedSourceData = false;
+                if (! isset($sourceEntityData)) {
+                    $sourceEntityData = $this->em->getUnitOfWork()->getOriginalEntityData($sourceEntity);
+                }
 
-            $computedIdentifier[$targetClass->getFieldForColumn($targetKeyColumn)] =
-                $sourceClass->reflFields[$sourceClass->fieldNames[$sourceKeyColumn]]->getValue($sourceEntity);
+                if (isset($sourceEntityData[$sourceKeyColumn])) {
+                    $dataValue = $sourceEntityData[$sourceKeyColumn];
+                    if ($dataValue !== null) {
+                        $resolvedSourceData                                                    = true;
+                        $computedIdentifier[$targetClass->getFieldForColumn($targetKeyColumn)] =
+                            $dataValue;
+                    }
+                }
+
+                if (! $resolvedSourceData) {
+                    throw MappingException::joinColumnMustPointToMappedField(
+                        $sourceClass->name,
+                        $sourceKeyColumn
+                    );
+                }
+            } else {
+                $computedIdentifier[$targetClass->getFieldForColumn($targetKeyColumn)] =
+                    $sourceClass->reflFields[$sourceClass->fieldNames[$sourceKeyColumn]]->getValue($sourceEntity);
+            }
         }
 
         $targetEntity = $this->load($computedIdentifier, null, $assoc);
--- a/tests/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/InverseSide.php
+++ b/tests/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/InverseSide.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\OneToOneInverseSideWithAssociativeIdLoad;
+
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\JoinColumn;
+use Doctrine\ORM\Mapping\OneToOne;
+use Doctrine\ORM\Mapping\Table;
+
+/**
+ * @Entity()
+ * @Table(name="one_to_one_inverse_side_assoc_id_load_inverse")
+ */
+class InverseSide
+{
+    /**
+     * Associative id (owning identifier)
+     *
+     * @var InverseSideIdTarget
+     * @Id()
+     * @OneToOne(targetEntity=InverseSideIdTarget::class, inversedBy="inverseSide")
+     * @JoinColumn(nullable=false, name="associativeId")
+     */
+    public $associativeId;
+
+    /**
+     * @var OwningSide
+     * @OneToOne(targetEntity=OwningSide::class, mappedBy="inverse")
+     */
+    public $owning;
+}
--- a/tests/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/InverseSideIdTarget.php
+++ b/tests/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/InverseSideIdTarget.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\OneToOneInverseSideWithAssociativeIdLoad;
+
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\OneToOne;
+use Doctrine\ORM\Mapping\Table;
+
+/**
+ * @Entity()
+ * @Table(name="one_to_one_inverse_side_assoc_id_load_inverse_id_target")
+ */
+class InverseSideIdTarget
+{
+    /**
+     * @var string
+     * @Id()
+     * @Column(type="string", length=255)
+     * @GeneratedValue(strategy="NONE")
+     */
+    public $id;
+
+    /**
+     * @var InverseSide
+     * @OneToOne(targetEntity=InverseSide::class, mappedBy="associativeId")
+     */
+    public $inverseSide;
+}
--- a/tests/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/OwningSide.php
+++ b/tests/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/OwningSide.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\OneToOneInverseSideWithAssociativeIdLoad;
+
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\JoinColumn;
+use Doctrine\ORM\Mapping\OneToOne;
+use Doctrine\ORM\Mapping\Table;
+
+/**
+ * @Entity()
+ * @Table(name="one_to_one_inverse_side_assoc_id_load_owning")
+ */
+class OwningSide
+{
+    /**
+     * @var string
+     * @Id()
+     * @Column(type="string", length=255)
+     * @GeneratedValue(strategy="NONE")
+     */
+    public $id;
+
+    /**
+     * Owning side
+     *
+     * @var InverseSide
+     * @OneToOne(targetEntity=InverseSide::class, inversedBy="owning")
+     * @JoinColumn(name="inverse", referencedColumnName="associativeId")
+     */
+    public $inverse;
+}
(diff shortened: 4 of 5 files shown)


mcurland:
It does not appear that any static or dynamic failure here has anything to do with the changes in this commit. Is the current 2.17.x branch head passing its own test suite?

greg0ire:
Very weird… I tried locally, no issue on 2.17.x, and there were issues with your branch but… with Psalm :thinking: 
Let's close and reopen to trigger a whole run of the CI.

greg0ire:
Ah my bad, I forgot to fetch before trying the build on 2.17.x Now I can see I was testing your branch with Psalm 2.15.

lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
```diff
@@ -582,9 +582,9 @@ public function loadOneToManyCollection(array $assoc, $sourceEntity, PersistentC
     /**
      * {@inheritDoc}
      */
-    public function loadOneToOneEntity(array $assoc, $sourceEntity, array $identifier = [])
+    public function loadOneToOneEntity(array $assoc, $sourceEntity, array $identifier = [], array $sourceEntityData = [])
```
greg0ire: The class is not marked as internal. This kind of breaking change could be avoided with `func_get_args()`

lib/Doctrine/ORM/Persisters/Entity/EntityPersister.php
```diff
@@ -217,16 +217,21 @@ public function loadById(array $identifier, $entity = null);
      * association from another entity.
      *
      * @param object $sourceEntity The entity that owns the association (not necessarily the "owning side").
-     * @psalm-param array<string, mixed> $identifier The identifier of the entity to load. Must be provided if
-     *                                               the association to load represents the owning side, otherwise
-     *                                               the identifier is derived from the $sourceEntity.
-     * @psalm-param AssociationMapping $assoc        The association to load.
+     * @psalm-param array<string, mixed> $sourceEntityData The data used to create the sourceEntity. If the sourceEntity
+     *                                                     is identified by an association, then that association may
+     *                                                     not be initialize before this call. This original data is used
```
greg0ire:
```suggestion
     *                                                     not be initialized before this call. This original data is used
```

tests/Doctrine/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/InverseSideIdTarget.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * @Entity()
+ * @Table(name="one_to_one_inverse_side_assoc_id_load_inverse_id_target")
+ */
+class InverseSideIdTarget
+{
+    /**
+     * @var string
+     * @Id()
+     * @Column(type="string", length=255)
```
greg0ire: The length is not relevant to the issue here, is it?

```suggestion
     * @Column(type="string")
```

tests/Doctrine/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/InverseSide.php
```diff
@@ -0,0 +1,34 @@
...
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\JoinColumn;
+use Doctrine\ORM\Mapping\OneToOne;
+use Doctrine\ORM\Mapping\Table;
+
+/**
+ * @Entity()
+ * @Table(name="one_to_one_inverse_side_assoc_id_load_inverse")
+ */
+class InverseSide
```
greg0ire: Ok so… this entity is the inverse side of the second relationship… and the owning side of the first relationship. I think your goal was to clarify things here, but I'm afraid it achieves the opposite :sweat_smile: 

Maybe you should come up with something a more realistic example, where it is obvious which entity would own which.

mcurland: This is the inverse side of the relationship the defect focuses on, which is any inverse 1-1 in an entity identified by an association. I just needed another entity for the identifier to target, so InverseSideIdTarget seemed like a reasonable name.

This was just an edit of the OneToOneInverseSideLoad model and test case, which was close to the pattern I needed. Clearly the id relationship has to be owning. The only different here is that the id column on InverseSide maps to an association and join column instead of the direct column declaration in the prior test case. If you compare the two I think it is pretty obvious what the difference is here.

The original names in the model that broke this were Action (the base type/id target here) and StateAction (the InverseSide) the other class was just State (OwningSide), which owns 1-1 Entry and Exit actions that reference StateAction. Out of context I'm not sure those names make anything any easier.

This is just a separated subtype pattern that isn't declared as such in the Doctrine ORM. It's nice to have that feature-I don't recall it being available when I first did this generated mapping in 2014 timeframe-but for the systems I'm working on this is just one of many layers generated from the same object-role model (the real breaking model produces 9k lines of Doctrine declarations sandwiched between 1400 of generated DDL and 20k lines that writes a JSON StateChange spec and reads a JSON-based query specification of the model). In other words, I very rarely interact with the Doctrine entities directly, so the final entity shape doesn't really affect me much. It doesn't matter if it is declared as a subtype or a 1-1, or if the database maps it with a separate table or combined in the same table, or even which side is owning/inverse, I just need to be able to start with any type and have the query succeed. I'm well aware this is a different perspective than using the entities as primary code artifacts, where entity aesthetics may be much more important.

tests/Doctrine/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/InverseSideIdTarget.php
mcurland: No, it isn't relevant here or in the OneToOneInverseSideLoad model this was directly copied from and is also the pattern in other tests I looked at. I generally have length-bounds on any column used in a key or index because dbs have limits, so I didn't think to remove it.

The more interesting question would be a different identifying type altogether (int or UUID, etc.).

The one concern I had with this fix, though, is that I wasn't sure if I should be using the raw input data. Normally this is processed through a reflField set/get value calls. I did not look into whether those ever normalize data or not. My rationale for skipping this question is that the raw data has just arrived from a SQL query and is being fed straight back into a sql query here, so should already be in a form suitable to use as a SQL parameter. Do you agree with this analysis?

tests/Doctrine/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/InverseSide.php
greg0ire: Fair enough, let's leave it as is then.

tests/Doctrine/Tests/Models/OneToOneInverseSideWithAssociativeIdLoad/InverseSideIdTarget.php
greg0ire: > No, it isn't relevant here or in the OneToOneInverseSideLoad model this was directly copied from and is also the pattern in other tests I looked at.

I think I would remove any distracting things that aren't strictly necessary to reproduce the issue.

> The one concern I had with this fix, though, is that I wasn't sure if I should be using the raw input data. Normally this is processed through a reflField set/get value calls. I did not look into whether those ever normalize data or not.

I don't know either, hopefully other maintainers that are more familiar with the internals of the ORM will.

> My rationale for skipping this question is that the raw data has just arrived from a SQL query and is being fed straight back into a sql query here, so should already be in a form suitable to use as a SQL parameter. Do you agree with this analysis?

I think I do, but again, I'm not sure my opinion is worth much here.

lib/Doctrine/ORM/Cache/Persister/Entity/AbstractEntityPersister.php
mcurland: I can't call that one in someone else's code base. I see some internal methods, but no internal classes. Is there anywhere you know of where this is used by something other than a UnitOfWork?

I would assume I'd get a lot of pushback from automated code checkers if I tried to slip something like this in the back door and pass more arguments than declared.

An id hash isn't available (all of the id data is still null in the entity), but the same original entity data should be available from the persister with $this->em->getUnitOfWork()->getOriginalEntityData($sourceEntity). This seems like a lot of work to do instead of just forwarding $data, but I would much rather take this approach than use func_get_args. This would also eliminate signature changes in most of the files.

Would you like me to take this approach to preserve the existing public signature?

greg0ire: Yes, if by "a lot of work" you mean typing a few extra chars as opposed to a lot of work for php. I'm on my phone so can't check myself. As for internal classes, maybe they 're in another project, or just in an `Internal` namespace

mcurland: The runtime hit is small, so I removed the signature change. The fix is now limited to BasicEntityPersister (plus the tests).

I'm going to tangent for an opinion (for a possible separate defect in the same code path):
The identifier is not provided by UoW for the inverse one-to-one path. This looks like a defect, and may be the source of an issue I've known about for a while. Basically, in 1-1 scenarios, if you modify an entity that was pulled as part of a 1-1 then later request a related 1-1 entity from the repository before flushing, then the entity will be replaced with fresh data from the db and your edits will be lost.

I accepted this as 'part of the library' and spit code to cache all related 1-1s when an entity is retrieved, which redirects any request for a previously fetched related 1-1 before the UoW sees it. However, entity caching is the job of the UoW, so I'd rather not keep the secondary hack cache.

I think omitting the identifier argument on the loadOneToOneEntity call in UnitOfWork.php line 2966 is the likely cause of this problem. Do you think this should be investigated? (No rush, you probably don't want to do this one from your phone).

greg0ire: After reading https://www.doctrine-project.org/projects/doctrine-orm/en/2.17/reference/working-with-objects.html#entities-and-the-identity-map, my opinion is that this is indeed a bug since that behavior is not mentioned as a known issue or particular case. I think you can investigate it and that whatever you find, it could result in either a bugfix or a documentation improvement.

mcurland:
Moved from 2.17.x to 2.20.x and consolidated change sets (after we changed the approach to get the original data).

If you want this in the default branch (currently 3.2.x) instead please let me know.

Could someone please review this? IMO this was ready to go 9 months ago.

greg0ire:
According to https://github.com/doctrine/orm/issues/11208, since this is a bug, you should pick 2.19.x

greg0ire:
Rebased.

mcurland:
Thanks for the rebase Grégoire, even if my local is now confused : )

Now let's get this reviewed and in, please. This change is much more localized (no method signature changes) than the original proposal.

greg0ire:
Nothing a good old `git fetch --all && git reset --hard @{upstream}` won't fix. @derrabus @SenseException @beberlei could you please review this?

SenseException (approved):
Not sure if I understand if this is a bugfix or not, but it doesn't look wrong to put this into a current 2.x branch.

greg0ire:
Thanks @mcurland !

PR #11116, opened 2023-12-12
Implement named arguments for NEW DTO syntax

### LINKED ISSUE #9182: DQL NEW Operator to support PHP 8.x named arguments.
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Would it be possible for DQL NEW operator (https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/dql-doctrine-query-language.html#new-operator-syntax)  to support PHP 8.0 named arguments (https://stitcher.io/blog/php-8-named-arguments) additionally to regular positional ones? As far as I see, DQL NEW() syntax is great for transforming Doctrine results to DTO objects but only drawback is use of positional arguments, which can be easily mixed up in the long-term (human error). Named arguments would fix that. 

Consider class:

```
class BookDTO
{
    public function __construct(
        public int $id,
        public string $title,
        public int $publish_year,
    ) { }
}
```

Current DQL:

`$query = $em->createQuery('SELECT NEW BookDTO(b.id, b.name, b.publish_year) FROM Book b');`

Proposed DQL:

`$query = $em->createQuery('SELECT NEW BookDTO(id: b.id, name: b.name, publish_year: b.publish_year) FROM Book b');`

Zuruuh:
I love using DTOs but it can get pretty annoying to understand what's going on when using a ton of arguments, having to keep track of what is what, in what order, etc... Using named arguments should really help a lot with that I think :+1:  
Small note but I also made it that trailing commas don't crash the lexer which is a nice qol feature to have imo. I can remove it / move it to another PR if needed 👌

Now the following DQL works as expected
```sql
SELECT NEW CustomerDTO(
  email: e.email,
  name: u.name,
  city: a.address,
) FROM ...
```

Closes #9182

### DIFF
--- a/docs/en/reference/dql-doctrine-query-language.rst
+++ b/docs/en/reference/dql-doctrine-query-language.rst
@@ -562,6 +562,16 @@ And then use the ``NEW`` DQL keyword :
 
 Note that you can only pass scalar expressions to the constructor.
 
+The ``NEW`` operator also supports named arguments:
+
+.. code-block:: php
+
+    <?php
+    $query = $em->createQuery('SELECT NEW CustomerDTO(email: e.email, name: c.name, address: a.city) FROM Customer c JOIN c.email e JOIN c.address a');
+    $users = $query->getResult(); // array of CustomerDTO
+
+Note that you must not pass ordered arguments after named ones.
+
 Using INDEX BY
 ~~~~~~~~~~~~~~
 
@@ -1650,7 +1660,7 @@ Select Expressions
     SelectExpression        ::= (IdentificationVariable | ScalarExpression | AggregateExpression | FunctionDeclaration | "(" Subselect ")" | CaseExpression | NewObjectExpression) [["AS"] ["HIDDEN"] AliasResultVariable]
     SimpleSelectExpression  ::= (StateFieldPathExpression | IdentificationVariable | FunctionDeclaration | AggregateExpression | "(" Subselect ")" | ScalarExpression) [["AS"] AliasResultVariable]
     NewObjectExpression     ::= "NEW" AbstractSchemaName "(" NewObjectArg {"," NewObjectArg}* ")"
-    NewObjectArg            ::= ScalarExpression | "(" Subselect ")"
+    NewObjectArg            ::= ScalarExpression | NamedScalarExpression | "(" Subselect ")"
 
 Conditional Expressions
 ~~~~~~~~~~~~~~~~~~~~~~~
--- a/src/Internal/Hydration/ObjectHydrator.php
+++ b/src/Internal/Hydration/ObjectHydrator.php
@@ -554,7 +554,8 @@ class ObjectHydrator extends AbstractHydrator
             foreach ($rowData['newObjects'] as $objIndex => $newObject) {
                 $class = $newObject['class'];
                 $args  = $newObject['args'];
-                $obj   = $class->newInstanceArgs($args);
+
+                $obj = $class->newInstanceArgs($args);
 
                 if ($scalarCount === 0 && count($rowData['newObjects']) === 1) {
                     $result[$resultKey] = $obj;
--- a/src/Internal/Hydration/ObjectHydrator.php
+++ b/src/Internal/Hydration/ObjectHydrator.php
@@ -554,8 +554,7 @@ class ObjectHydrator extends AbstractHydrator
             foreach ($rowData['newObjects'] as $objIndex => $newObject) {
                 $class = $newObject['class'];
                 $args  = $newObject['args'];
-
-                $obj = $class->newInstanceArgs($args);
+                $obj   = $class->newInstanceArgs($args);
 
                 if ($scalarCount === 0 && count($rowData['newObjects']) === 1) {
                     $result[$resultKey] = $obj;
--- a/src/Query/AST/NamedScalarExpression.php
+++ b/src/Query/AST/NamedScalarExpression.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Query\AST;
+
+use Doctrine\ORM\Query\SqlWalker;
+
+class NamedScalarExpression extends Node
+{
+    public function __construct(
+        public readonly Node $innerExpression,
+        public readonly string|null $name = null,
+    ) {
+    }
+
+    public function dispatch(SqlWalker $walker): string
+    {
+        return $this->innerExpression->dispatch($walker);
+    }
+}
--- a/src/Query/Parser.php
+++ b/src/Query/Parser.php
@@ -1634,12 +1634,20 @@ final class Parser
 
         $this->match(TokenType::T_OPEN_PARENTHESIS);
 
-        $args[] = $this->NewObjectArg();
+        $arg                   = $this->NewObjectArg();
+        $namedArgAlreadyParsed = $arg instanceof AST\NamedScalarExpression;
+        $args                  = [$arg];
 
         while ($this->lexer->isNextToken(TokenType::T_COMMA)) {
             $this->match(TokenType::T_COMMA);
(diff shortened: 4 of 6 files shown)


derrabus:
Thank you. Your new tests are failing, however.

Zuruuh:
yeah you're right, I only tested locally with a php version above 8 (and it worked fine), but it seems like `ReflectionClass::newInstanceArgs` didn't work with an associative array before the introduction of named params in php, so it will require a bit more work to handle php 7.x versions

tests/Doctrine/Tests/ORM/Functional/NewOperatorTest.php
```diff
@@ -1076,69 +1076,85 @@ public function testClassCantBeInstantiatedException(): void
...
+                FROM
+                    Doctrine\Tests\Models\CMS\CmsUser u
+                JOIN
+                    u.email e
+                JOIN
+                    u.address a
+                ORDER BY
+                    u.name',
+            ],
+            'Both named and unnamed arguments without trailling comma' => [
```
greg0ire:
```suggestion
            'Both named and unnamed arguments without trailing comma' => [
```

tests/Doctrine/Tests/ORM/Functional/NewOperatorTest.php
```diff
@@ -1076,69 +1076,85 @@ public function testClassCantBeInstantiatedException(): void
         $this->_em->createQuery($dql)->getResult();
     }
 
-    public function testQueryWithOnlyNamedArgumentsInOrder(): void
+    /** @return array<string, string> */
```
greg0ire:
```suggestion
    /** @return array<string, array{string}> */
```

Zuruuh:
Tested locally with php 7.1, tests should pass fine now :+1:

greg0ire:
This probably needs to be updated as well:

https://github.com/doctrine/orm/blob/2b91edc525bde4f23f9ada22a82c021b5c18c92b/docs/en/reference/dql-doctrine-query-language.rst?plain=1#L1690-L1691

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -1938,17 +1946,28 @@ public function NewObjectExpression()
     }
 
     /**
-     * NewObjectArg ::= ScalarExpression | "(" Subselect ")"
+     * NewObjectArg ::= ScalarExpression | NamedScalarExpression | "(" Subselect ")"
      *
      * @return mixed
      */
-    public function NewObjectArg()
+    public function NewObjectArg(bool $namedArgAlreadyParsed = false)
```
greg0ire: The Parser is neither final nor internal. This means this is a breaking change for extending classes. You can obtain the new argument with `func_get_args()` without changing the signature.
```suggestion
    public function NewObjectArg(/* bool $namedArgAlreadyParsed = false */)
```

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -584,6 +584,16 @@ And then use the ``NEW`` DQL keyword :
 
 Note that you can only pass scalar expressions to the constructor.
 
+The ``NEW`` operator also supports named arguments, similarly to php 8.0 :
```
derrabus:
```suggestion
The ``NEW`` operator also supports named arguments:
```

docs/en/reference/dql-doctrine-query-language.rst
```diff
@@ -584,6 +584,16 @@ And then use the ``NEW`` DQL keyword :
...
+
+.. code-block:: php
+
+    <?php
+    $query = $em->createQuery('SELECT NEW CustomerDTO(email: e.email, name: c.name, address: a.city) FROM Customer c JOIN c.email e JOIN c.address a');
+    $users = $query->getResult(); // array of CustomerDTO
+
+Note that you cannot pass unnamed arguments after named ones, just like in php.
```
derrabus:
```suggestion
Note that you must not pass ordered arguments after named ones.
```

lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php
```diff
@@ -565,7 +568,28 @@ protected function hydrateRowData(array $row, array &$result)
             foreach ($rowData['newObjects'] as $objIndex => $newObject) {
                 $class = $newObject['class'];
                 $args  = $newObject['args'];
-                $obj   = $class->newInstanceArgs($args);
+
+                if (PHP_VERSION_ID >= 80000) {
+                    $obj = $class->newInstanceArgs($args);
+                } else {
+                    $constructor = $class->getConstructor();
+                    $unnamedArgs = [];
```
derrabus:
```suggestion
                    $orderedArgs = [];
```

lib/Doctrine/ORM/Query/AST/NamedScalarExpression.php
```diff
@@ -0,0 +1,28 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\ORM\Query\AST;
+
+class NamedScalarExpression extends Node
+{
+    /** @var Node */
+    public $innerExpression;
+
+    /** @var ?string */
```
derrabus:
```suggestion
    /** @var string|null */
```

derrabus (changes requested):

lib/Doctrine/ORM/Query/Parser.php
```diff
@@ -1938,17 +1947,30 @@ public function NewObjectExpression()
...
     /**
-     * NewObjectArg ::= ScalarExpression | "(" Subselect ")"
+     * NewObjectArg ::= ScalarExpression | NamedScalarExpression | "(" Subselect ")"
      *
      * @return mixed
      */
-    public function NewObjectArg()
+    public function NewObjectArg(/* bool $namedArgAlreadyParsed = false */)
     {
+        $namedArgAlreadyParsed = func_get_args()[0] ?? false;
```
derrabus: Do we actually need this parameter? Couldn't `NewObjectExpression()` just check if the kind of argument it receives is supported at its position?

Zuruuh: I think this should work (but idk how to feel abt it): when parsing a `NewObjectExpression`, we parse the current one, consume the tokens, then parse the next one (if there is any) without consuming it, and throw if it's not named while the current one is. It would work without the need of the parameter, but i feel like it's kinda doing too much ? also we're parsing twice all the arguments while it's not really needed (idk if it's really impactant in terms of perf). wdyt ?

Zuruuh: Hello @derrabus, I didn't want to ping earlier since it was holiday season; what do you think about the approach describe above ? did you have something else in mind ?

Zuruuh:
just saw #11208, I will update the mr to target 3.1 :+1:

Zuruuh:
Update done
 - dropped the custom implementation for php versions prior to `8.0` since the lib requires `>=8.1`
 - Implemented the changes that would have been breaking in `src/Query/Parser.php` since the class is now finalized

greg0ire (dismissed):

src/Internal/Hydration/ObjectHydrator.php
```diff
@@ -554,7 +554,8 @@ protected function hydrateRowData(array $row, array &$result): void
             foreach ($rowData['newObjects'] as $objIndex => $newObject) {
                 $class = $newObject['class'];
                 $args  = $newObject['args'];
-                $obj   = $class->newInstanceArgs($args);
+
+                $obj = $class->newInstanceArgs($args);
```
norkunas: what's the point of this diff?

Zuruuh: there used to be code before switching target branch, will remove :+1:

greg0ire (approved):

beberlei:
This was taken over by #11575

PR #11122, opened 2023-12-14
Fix enum mapping validation

### LINKED ISSUE #11119: BackedEnum type with interface validation mismatch
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.17.1

#### Summary

After upgrading to version 2.17.1 I've encountered this validation error:

```
$ bin/console doctrine:schema:validate

[FAIL] The entity-class App\Entity\CertificateEvent mapping is invalid:
 * The field 'App\Entity\CertificateEvent#type' has the property type 'App\Enum\EventType' that differs from the metadata field type 'string' returned by the 'string' DBAL type.
```

#### Previous behavior
Schema validation pass

#### Current behavior
Schema validation fails

#### How to reproduce

```php
interface EventType
{
}

enum CertificateEventType: string implements EventType
{
}
```

the mapping is:
```xml
<entity name="App\Entity\CertificateEvent">
    <field name="type" type="string" column="type" enum-type="App\Enum\CertificateEventType" nullable="false" />
```

CertificateEvent looks like:
```php
class CertificateEvent extends CaseEvent
{
    public function __construct(
        CertificateEventType $type
    ) {
      parent::__construct($case, $type);
    }
```

```php
abstract class CaseEvent
{
    public function __construct(
        protected EventType $type,
    ) {
```

yceruto:
Closes #11119

### DIFF
--- a/lib/Doctrine/ORM/Tools/SchemaValidator.php
+++ b/lib/Doctrine/ORM/Tools/SchemaValidator.php
@@ -39,6 +39,7 @@ use function count;
 use function get_class;
 use function implode;
 use function in_array;
+use function interface_exists;
 use function is_a;
 use function sprintf;
 
@@ -389,10 +390,20 @@ class SchemaValidator
                             return null;
                         }
 
-                        if (
-                            is_a($propertyType, BackedEnum::class, true)
-                            && $metadataFieldType === (string) (new ReflectionEnum($propertyType))->getBackingType()
-                        ) {
+                        if (is_a($propertyType, BackedEnum::class, true)) {
+                            $backingType = (string) (new ReflectionEnum($propertyType))->getBackingType();
+
+                            if ($metadataFieldType !== $backingType) {
+                                return sprintf(
+                                    "The field '%s#%s' has the property type '%s' with a backing type of '%s' that differs from the metadata field type '%s'.",
+                                    $class->name,
+                                    $fieldName,
+                                    $propertyType,
+                                    $backingType,
+                                    $metadataFieldType
+                                );
+                            }
+
                             if (! isset($fieldMapping['enumType']) || $propertyType === $fieldMapping['enumType']) {
                                 return null;
                             }
@@ -403,6 +414,28 @@ class SchemaValidator
                                 $fieldName,
                                 $propertyType,
                                 $fieldMapping['enumType']
+                            );
+                        }
+
+                        if (
+                            isset($fieldMapping['enumType'])
+                            && $propertyType !== $fieldMapping['enumType']
+                            && interface_exists($propertyType)
+                            && is_a($fieldMapping['enumType'], $propertyType, true)
+                        ) {
+                            $backingType = (string) (new ReflectionEnum($fieldMapping['enumType']))->getBackingType();
+
+                            if ($metadataFieldType === $backingType) {
+                                return null;
+                            }
+
+                            return sprintf(
+                                "The field '%s#%s' has the metadata enumType '%s' with a backing type of '%s' that differs from the metadata field type '%s'.",
+                                $class->name,
+                                $fieldName,
+                                $fieldMapping['enumType'],
+                                $backingType,
+                                $metadataFieldType
                             );
                         }
 
--- a/tests/Doctrine/Tests/ORM/Functional/Ticket/GH11037/GH11037Test.php
+++ b/tests/Doctrine/Tests/ORM/Functional/Ticket/GH11037/GH11037Test.php
@@ -40,8 +40,9 @@ final class GH11037Test extends OrmTestCase
 
         self::assertEquals(
             [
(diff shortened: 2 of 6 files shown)


lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -406,6 +407,25 @@ function (array $fieldMapping) use ($class): ?string {
                             );
                         }
 
+                        if (
+                            isset($fieldMapping['enumType'])
+                            && $propertyType !== $fieldMapping['enumType']
+                            && interface_exists($propertyType)
+                            && is_a($fieldMapping['enumType'], $propertyType, true)
+                        ) {
+                            if ($metadataFieldType === (string) (new ReflectionEnum($fieldMapping['enumType']))->getBackingType()) {
```
greg0ire: You're basing this on a comparison between `$metadataFieldType` and the backing type of the enum, yet only one of those appears in the error message.

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -406,6 +407,25 @@ function (array $fieldMapping) use ($class): ?string {
...
+                                return null;
+                            }
+
+                            return sprintf(
+                                "The field '%s#%s' has the metadata enumType '%s' that differs from the metadata field type '%s'.",
```
greg0ire: So let's mention it here please?

```suggestion
                                "The field '%s#%s' has the metadata enumType '%s' with a backing type of %s that differs from the metadata field type '%s'.",
```

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -406,6 +407,28 @@ function (array $fieldMapping) use ($class): ?string {
...
+                            }
+
+                            return sprintf(
+                                "The field '%s#%s' has the metadata enumType '%s' with a backing type of '%s' that differs from the metadata field type '%s'.",
+                                $class->name,
+                                $fieldName,
+                                $fieldMapping['enumType'],
+                                $backingType,
+                                $metadataFieldType
+                            );
```
greg0ire, line 439: Should this be extracted somehow, and reused on line 400? There could be the same improvement here (right?). Or maybe both conditions should somehow be combined? I having a hard time figuring out why you are not touching code already dealing with enums within this class :thinking:

lib/Doctrine/ORM/Tools/SchemaValidator.php
```diff
@@ -406,6 +407,28 @@ function (array $fieldMapping) use ($class): ?string {
                             );
                         }
 
+                        if (
+                            isset($fieldMapping['enumType'])
+                            && $propertyType !== $fieldMapping['enumType']
+                            && interface_exists($propertyType)
```
greg0ire, line 423: Why do we need to check that it's an interface?

lib/Doctrine/ORM/Tools/SchemaValidator.php
yceruto: I'm having a hard time figuring out the same concern. I'm still checking, but they seem like different scenarios.

lib/Doctrine/ORM/Tools/SchemaValidator.php
yceruto: For two reasons: 1) The static analyzer complained about passing a non-class-string as the second argument of the `is_a` function; 2) It's only possible for enums to implement interfaces.

lib/Doctrine/ORM/Tools/SchemaValidator.php
yceruto: Note that in this use case, the property type is an interface, so we don't really know what it's for. The only thing we can check is whether the mapped enum type is an implementation of that interface, and whether it matches the mapping type, nothing more.

Anyway, I will give it a try to see if by inverting the first block, I can merge something there.

greg0ire (dismissed):
After re-reading the first case, my brain hurts but I think it's indeed really different. I think you actually went with the right way, combining the conditions is probably going to produce something very hard to understand.

yceruto:
Still, I improved the first error message to include the backing type information.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH11037/GH11037Test.php
```diff
@@ -40,8 +40,9 @@ public function testMetadataFieldTypeNotCoherentWithEntityPropertyType(): void
 
         self::assertEquals(
             [
-                "The field 'Doctrine\Tests\ORM\Functional\Ticket\GH11037\InvalidEntityWithTypedEnum#status1' has the property type 'Doctrine\Tests\ORM\Functional\Ticket\GH11037\StringEntityStatus' that differs from the metadata field type 'int' returned by the 'integer' DBAL type.",
+                "The field 'Doctrine\Tests\ORM\Functional\Ticket\GH11037\InvalidEntityWithTypedEnum#status1' has the property type 'Doctrine\Tests\ORM\Functional\Ticket\GH11037\StringEntityStatus' with a backing type of 'string' that differs from the metadata field type 'int'.",
```
greg0ire, line 43: Ah yes, that's way better 👍

SenseException (approved):

greg0ire:
Thanks @yceruto !

Issue #11073, opened 2023-11-20
Invalid field error reported for bigint DBAL type


emodric:
### Bug Report

```
The field 'App\Entity\Message#id' has the property type 'int' that differs
from the metadata field type 'string' returned by the 'bigint' DBAL type.
```

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.17.1

#### Summary

I have an entity with a `bigint` primary key, wich has an `int` typehint in the entity class. I realize from the error message that `bigint` uses `string` internally, but so far, there was no issues with using `int` typehint, especially, since we do not need values larger than 64bit integers.

#### Current behavior

After upgrading to Doctrine ORM 2.17.1, the above error is shown in logs.

#### How to reproduce

```xml
<entity name="App\Entity\Message">
    <id name="id" type="bigint">
        <generator />
    </id>
</entity>
```

```php
class Message
{
    private int $id;
}
```

#### Expected behavior

No error message is reported.

greg0ire:
```
Expected behavior

No error message is reported.
```
Why? As you mentioned yourself, `bigint` uses `string` internally. Either you should pick another type, or you should write a custom type that wraps the bigint type and returns an `int`. Note that you might not be aware of performance issues this can cause, see https://github.com/doctrine/orm/issues/10944

emodric:
What I really wanted to say is that it would be nice to allow usage of `bigint` with `int` PHP typehint, allowing up to 64bit size integer values, without having the warning in the logs.

dmamchyts:
The same issue after upgrade from
```
        "doctrine/doctrine-bundle": "^2.11.0",
        "doctrine/orm": "^2.16.2",
```

to 
```
        "doctrine/doctrine-bundle": "^2.11.1",
        "doctrine/orm": "^2.17.1",
```

Entity: 
```
#[ORM\Entity(repositoryClass: UserRepository::class)]
#[ORM\HasLifecycleCallbacks]
#[ORM\Table(name: 'users')]
class User extends AbstractEntity implements UserInterface
{
...
    #[ORM\Column(type: Types::BIGINT, nullable: true, unique: true)]
    protected ?int $facebookId = null;
...
```


error: 
```
	
The field 'App\Common\Entity\User#facebookId' has the
property type 'int' that differs from the metadata field type 'string' returned by the 'bigint' DBAL type.
```

greg0ire:
[...]

PR #11130, opened 2023-12-20
Allow to skip property type validation

derrabus:
See https://github.com/doctrine/orm/issues/11073#issuecomment-1822705925 for context.

#10946 introduced validation of property types for the `ValidateSchemaCommand`. For existing projects, this new feature might uncover quite a few valid problems, that need a little time to fix. This is problematic because if schema validation is run in CI, a project basically can't upgrade until all of the newly discovered errors have been fixed. This leaves two options:

* Disable schema validation completely and upgrade to 2.17.
* Stay on 2.16 until all error have been resolved.

Both strategies are not desirable which is why I propose to add a flag to the command that allows us to opt out of the validation of property types. This allows this kind of projects to upgrade to 2.17 and fix the type mismatches iteratively.


greg0ire (approved):

kevinpapst:
Thank you @derrabus , I had to deactivate `doctrine:schema:validate` in CI due to that new validation, which reported "false negatives" on a `mixed` type, which was manually type-handled.

Issue #11098, opened 2023-11-30
Regression in `Events::postRemove` event 


onEXHovia:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.17.1

#### Summary

When updating orm with version `2.14.1` to `2.17.1`, some events stopped working. These are quite old tests and they still worked.
The idea was that after deleting an entity, check if there are more related entities, if no, then delete the parent(owner).

#### Current behavior

Parent(owner) entity stopped being deleted.

#### How to reproduce

```php
namespace App\EventListener;

use Doctrine\Common\EventSubscriber;
use Doctrine\ORM\EntityManagerInterface;
use Doctrine\ORM\Events;
use Doctrine\Persistence\Event\LifecycleEventArgs;
use App\Entity\PullRequest\Comment;

final class NoteAutoRemoveListener implements EventSubscriber
{
    public function getSubscribedEvents(): array
    {
        return [
            Events::postRemove,
        ];
    }

    public function postRemove(LifecycleEventArgs $args): void
    {
        $entity = $args->getObject();
        $entityManager = $args->getObjectManager();

        if ($entity instanceof Comment) {
            $this->removeDiscussionNote($entity, $entityManager);
        }
    }

    private function removeDiscussionNote(Comment $comment, EntityManagerInterface $entityManager): void
    {
        $discussion = $comment->getNote();

        if (null !== $discussion && 0 === $discussion->getComments()->count()) {
            $entityManager->remove($discussion);
        }
    }
}
```

and test

```php
namespace App\Tests\EventListener;

use Doctrine\ORM\EntityManagerInterface;
use App\Entity\PullRequest\Comment;
use App\Entity\PullRequest\DiscussionNote;
use App\Entity\PullRequest\Note;
use App\Entity\User;
use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;

final class NoteAutoRemoveListenerTest extends KernelTestCase
{
    private EntityManagerInterface $em;

    protected function setUp(): void
    {
        parent::bootKernel();

        $this->em = self::getContainer()->get(EntityManagerInterface::class);
    }

    public function testRemoveDiscussionNoteAfterRemoveComment(): void
    {
        $pullRequest = $this->createPullRequest(); // method create entity
        $this->em->persist($pullRequest);
        $this->em->flush();

        $comment = (new Comment())
            ->setBody('text')
[...]

PR #11132, opened 2023-12-22
Update in-memory collections before dispatching the `postRemove` event

mpdude:
Make sure in-memory collections have removed entities unset before the `postRemove` event is dispatched. This is related to #11098, although by itself probably not going to fix the regression.

#### Background 

When an entity is removed from the database, the UoW also takes care of removing that entity from all in-memory collections. #10763 moved this cleanup and taking snapshots of updated collections to a very late stage during the commit phase, in order to avoid other side effects.

Now, part of the issue in #11098 is that `postRemove` event listeners will be called at a point where the database-level `DELETE` has happened (although the transaction has not yet been committed), but the in-memory collections have not yet been updated.

#### Suggested change

This PR moves the code part that unsets removed entities from in-memory collections and takes collection snapshots (makes collections "clean") from after transaction commit to before the `postRemove` event. That brings the in-memory update closer to the database-level execution.

In the case of a transaction failure/abort, this leaves us with updated and snapshotted in-memory collections, but no database-level updates. But, at this point, probably things got inconsistent anyways, the EM will be closed and we need not be too worried about the state.

On the other hand, it might not be worth the effort to have clean/consistent collections at `postRemove` time, since that still leaves us with dirty/inconsistent states during `postPersist` or `postUpdate`.

### DIFF
--- a/lib/Doctrine/ORM/UnitOfWork.php
+++ b/lib/Doctrine/ORM/UnitOfWork.php
@@ -469,6 +469,8 @@ class UnitOfWork implements PropertyChangedListener
             // (first delete entities depending upon others, before deleting depended-upon entities).
             if ($this->entityDeletions) {
                 $this->executeDeletions();
+            } else {
+                $this->finalizeCollectionUpdates();
             }
 
             // Commit failed silently
@@ -490,7 +492,12 @@ class UnitOfWork implements PropertyChangedListener
         }
 
         $this->afterTransactionComplete();
+        $this->dispatchPostFlushEvent();
+        $this->postCommitCleanup($entity);
+    }
 
+    private function finalizeCollectionUpdates(): void
+    {
         // Unset removed entities from collections, and take new snapshots from
         // all visited collections.
         foreach ($this->visitedCollections as $coid => $coll) {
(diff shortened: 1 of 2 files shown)


mpdude:
What happens when we `DELETE` entities from the database, remove them from in-memory collections as well and then the transaction fails?

Is a transaction failure something that leaves us in an inconsistent state anyway, thus closing the EM?

mpdude:
Converting to draft: maybe it would be better if collections were in a clean state at the time the event is dispatched. That would be a more consistent state in case an event handler starts a new flush() phase.

beberlei:
> Is a transaction failure something that leaves us in an inconsistent state anyway, thus closing the EM?

Yes

mpdude:
@beberlei 

After your comment https://github.com/doctrine/orm/issues/11098#issuecomment-1869551396, would you say that we should close this PR here and not try to change or fix anything?

This means in-memory collections might still contain removed entities at the time `postRemove` is called (and additionally, be in a dirty state), whereas the database-level operations have been executed already, but the transation has not been committed yet.

mpdude:
Closing this.

As of #11146, the documentation is more explicit about the in-memory collection states. For other events, this kind of consistency isn't achieved either, so I am not sure it's worth bothering.

PR #11140, opened 2023-12-28
Flatten directory tree on 3.0

greg0ire:
It will make fuzzy matchers more efficient, and configuration files more readable.

- `lib/Doctrine/ORM` becomes just `src`
- `tests/Doctrine/` becomes just `tests`


derrabus:
The PR title is misleading. We've been using PSR-4 for quite a while already. 😉

greg0ire:
Note: I think this will make merge ups more difficult in the cases of files that have too different versions in 2.x vs 3.x

derrabus:
> Note: I think this will make merge ups more difficult in the cases of files that have too different versions in 2.x vs 3.x

Shall we apply this change to 2.18.x instead then?

greg0ire:
I think you're right. Let me first do a bunch of merges up to start on a clean state, and then I will:
- do this on 2.18.x
- rebase this PR
- do a merge up with the `-s ours` flag

SenseException (approved):

PR #11142, opened 2023-12-28
Remove inheritance

greg0ire:
Spotted while trying to merge https://github.com/doctrine/orm/pull/11076 (among other things) up into 3.0.x. On that branch, it is no longer possible for an entity to extend another entity without specifying an inheritance mapping type.

I think the goal of that inheritance was just to reuse the identifier anyway, so let's just duplicate the identifier declaration instead.


greg0ire:
Cc @norkunas

norkunas:
Sorry :)

greg0ire:
No worries, just wanted to be sure there was nothing necessary with that inheritance :)

SenseException (approved):

PR #11146, opened 2024-01-02
Mention that ``postRemove`` may still see removed entities in in-memory collections

mpdude:
... plus minor tweaks.

Depending on how we proceed with #11132, we might want to document that `post*` Events should not expect collections to be in a clean state, and that the transaction has not yet been committed at that point.

The warning about `flush()` reentrance belongs to a more general place, not only to `postPersist`/`postRemove`/`postUpdate` events.


greg0ire (approved):

SenseException (approved):

mpdude:
#11132 might change this a tad to also update the in-memory collections before the event is omitted. OTOH, this consistency isn't guaranteed for the other events either, so I am not sure it's worth to bother about it.

PR #11148, opened 2024-01-02
Flatten directory tree on 2.18

greg0ire:
It will make fuzzy matchers more efficient, and configuration files more readable.

- `lib/Doctrine/ORM` becomes just `src`
- `tests/Doctrine/` becomes just `tests`

Companion PR: https://github.com/doctrine/orm/pull/11140


SenseException (approved):
It doesn't look like that there are affected files that need to be included by users with a direct path that could be considered a BC break. But it will affect open PRs that were created before this change.

greg0ire:
@SenseException or @derrabus  can you please approve #11140 ? I'd like to merge both in quick succession, and then perform a merge up

Issue #3591, opened 2013-12-03
DDC-2838: Leaky abstraction when applying Criteria to hydrated/non-hydrated PersistentCollection


doctrinebot:
Jira issue originally created by user ptlis:

When applying a Criteria to a PersistentCollection that has been hydrated the field names must be camel case, if the collection has not yet been hydrated the field names must be underscore separated.

The github repo linked here contains a simplified testcase for the matrix of hydrated/non-hydrated entities and camel case/underscore separated fields.

https://github.com/ptlis/DoctrineTestcase

ostrolucky:
I tried this and it seems to be no longer reproducible with provided demo repository. @ptlis can you confirm?

ostrolucky:
This has been fixed via https://github.com/doctrine/collections/pull/57/commits/32d524547eb452718f85aa76b5cee4f7460c8fd0 you can close this guys @Ocramius

alcaeus:
Thanks @ostrolucky!

PR #11160, opened 2024-01-12
`Selectable` should compare field values, not call getters for initialized collections

mpdude:
This revisits #3591 which I think is not fixed correctly.

When using the [`Selectable` API](https://www.doctrine-project.org/projects/doctrine-collections/en/stable/index.html#selectable-methods) to filter elements from a collection, the result depends on the initialization state of the collection.

When the collection is uninitialized, the filtering happens in the database at the SQL level. The given field name is translated to a database column name and a database column-based lookup (`WHERE`) is performed.

When the collection is initialized, we go through hoops to derive potential getter/isser names and/or access `$object->$field` as a last resort, but completely fail on private properties without getter method access.

To the user, it should not make a difference whether a collection is initialized. We should use plain, direct field access (through reflection?) since that is what most closely resembles the database-based lookup. I think this is what Marco refers to in https://github.com/doctrine/collections/pull/149#issuecomment-475982285 as "rely only on state for lookups". 

I am not aware of any other places where the ORM would be concerned with deriving accessor names, which makes this an additional inconsistency.

Obviously, preferring code (getter based) access is not an option for uninitialized collections.

Note that even for plain "state" lookups, there is a remaining chance for issues when DBAL types come into play that do database-to-PHP value conversion. A SQL level column-value comparison might not come to the same result as a PHP-level comparison for certain DBAL type implementations. Anyways.

#### Suggestions needed

* How can we fix this, given that most of the `Selectable` API lives in doctrine/collections and uses static method calls? Is this an ORM issue or a collections issue?
* Are we risking BC breaks, so this fix needs to come as an opt-in switch with a deprecation?
* If so, is it too late in 2.x now?

Related:
* https://github.com/doctrine/collections/issues/170

### DIFF
--- a/tests/Tests/ORM/Functional/Ticket/GH3591Test.php
+++ b/tests/Tests/ORM/Functional/Ticket/GH3591Test.php
@@ -1,0 +1,229 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\Common\Collections\Expr\Comparison;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+use Generator;
+
+class GH3591Test extends OrmFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->setUpEntitySchema([
+            GH3591Entity::class,
+            GH3591CollectionEntry::class,
+        ]);
+    }
+
+    /**
+     * @dataProvider provideFieldNamesAndInitializationState
+     */
+    public function testMatchingApiOnUninitializedCollection(string $fieldName): void
+    {
+        $criteria = $this->createComparisonCriteria($fieldName, 'test value');
+        $entity   = $this->createAndLoadEntityWithCollectionEntry('test value');
+
+        $matches = $entity->entries->matching($criteria);
+
+        self::assertCount(1, $matches);
+    }
+
+    /**
+     * @dataProvider provideFieldNamesAndInitializationState
+     */
+    public function testMatchingApiOnInitializedCollection(string $fieldName): void
+    {
+        $criteria = $this->createComparisonCriteria($fieldName, 'test value');
+        $entity   = $this->createAndLoadEntityWithCollectionEntry('test value');
+        $entity->entries->initialize();
+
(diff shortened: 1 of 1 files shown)


mpdude:
Check: What happens when there are class hierarchies or mapped superclasses and we cannot tell which class contains a private field?

tests/Doctrine/Tests/ORM/Functional/Ticket/GH3591Test.php
```diff
@@ -0,0 +1,183 @@
...
+     * @var string
+     */
+    public $publicField;
+
+    /**
+     * @ORM\Column(type="string")
+     *
+     * @var string
+     */
+    public $fieldWithAkwardGetter;
```
greg0ire:
```suggestion
    public $fieldWithAwkwardGetter;
```

greg0ire:
> Are we risking BC breaks, so this fix needs to come as an opt-in switch with a deprecation?

That would be the case if any of the getter/isser has side effects. I don't know if we have documentation where we forbid that.

greg0ire:
> If so, is it too late in 2.x now?

You can do a breaking change only on 3.x, and you have until Feb 1st to do that. Even after Feb 1st, it will still be possible (and encouraged) to add deprecation layers to 2.x.

greg0ire:
> How can we fix this, given that most of the Selectable API lives in doctrine/collections and uses static method calls? Is this an ORM issue or a collections issue?

If feels like a `doctrine/collections` issue, and might affect the ODM as well, I don't know. Maybe try to reproduce the issue with the collections API only? 
If what you are worried is the upgrade path, I'd say a static property on `ClosureExpressionVisitor` telling whether to check getters and issers might help (but it would be global to the application, and libraries probably shouldn't interfere with it, just the end user). And then we remove it as well as the getter lookup in collections 3?

mpdude:
Let's ignore for the time being that I have no idea what the deprecation strategy/BC might look like and how to fix this as a cross-library issue...

In order to deal with private fields, you need to know the class in which the field is declared. For the ORM, field names are currently unique, so for a given field name, we can figure that out... as long as we have access to the ClassMetadata or the ClassMetadataFactory.

We probably can get hold of that in `PersistentCollection`s. 

But, what about entity classes that have just been created, where users assigned `$this->collection = new ArrayCollection()` and they then use the `Selectable` API? How would that `ArrayCollection` be able to obtain class metadata? Through static calls to locate a ManagerRegistry, to find the right EntityManager for the class containing the field? In addition, `ArrayCollection` comes from `doctrine/collections`, where we'd not want to have code that is concerned with such details of the ORM.

mpdude:
I think https://github.com/doctrine/collections/pull/472 can fix this.

PR #11165, opened 2024-01-17
Allow XML validation to be disabled but keep it enabled by default.

jwage:
After discussing this issue, we decided to allow XML validation to be disabled for now until we have a proper solution to allow validating against custom schemas. Disabling the XML validation is the only solution available at the moment so we want to continue allowing that until we have an alternative solution to offer.

### DIFF
--- a/src/Mapping/Driver/XmlDriver.php
+++ b/src/Mapping/Driver/XmlDriver.php
@@ -56,7 +56,7 @@ class XmlDriver extends FileDriver
             );
         }
 
-        if (! extension_loaded('dom')) {
+        if ($isXsdValidationEnabled && ! extension_loaded('dom')) {
             throw new LogicException(
                 'XSD validation cannot be enabled because the DOM extension is missing.',
             );
--- a/src/Mapping/Driver/XmlDriver.php
+++ b/src/Mapping/Driver/XmlDriver.php
@@ -43,26 +43,17 @@ class XmlDriver extends FileDriver
 
     /**
      * {@inheritDoc}
-     *
-     * @param true $isXsdValidationEnabled no-op, will be removed in 4.0
      */
     public function __construct(
         string|array|FileLocator $locator,
         string $fileExtension = self::DEFAULT_FILE_EXTENSION,
-        bool $isXsdValidationEnabled = true,
+        private readonly bool $isXsdValidationEnabled = true,
     ) {
         if (! extension_loaded('simplexml')) {
             throw new LogicException(
                 'The XML metadata driver cannot be enabled because the SimpleXML PHP extension is missing.'
                 . ' Please configure PHP with SimpleXML or choose a different metadata driver.',
             );
-        }
-
-        if (! $isXsdValidationEnabled) {
-            throw new InvalidArgumentException(sprintf(
-                'The $isXsdValidationEnabled argument is no longer supported, make sure to omit it when calling %s.',
-                __METHOD__,
-            ));
         }
 
         if (! extension_loaded('dom')) {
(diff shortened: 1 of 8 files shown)


greg0ire (dismissed):

greg0ire:
:thinking: I'm wondering how the deprecation messages on 2.17.x need to be adjusted now. Maybe the flag should be made mandatory in this PR, and what would become deprecated would be to rely on the default `false` value (since it's going to change)? But with the bundle calling `createXMLMetadataConfiguration`, the flag is probably always provided anyway, with the bundle's default value. So maybe what's best is to remove the deprecation entirely, and let people find about the deprecations on 3.0.x the hard way.

jwage:
@greg0ire Seems like we should just remove the deprecation from 2.17.x?

greg0ire:
Maybe yes, I wonder what other maintainers think.

src/Mapping/Driver/XmlDriver.php
```diff
@@ -914,6 +905,10 @@ protected function loadMappingFile($file)
 
     private function validateMapping(string $file): void
     {
+        if (! $this->isXsdValidationEnabled) {
+            return;
+        }
```
SenseException, line 910: Shouldn't this be tested? As far as I understand the proper solution might take some time and we'll have this change for a while.

jwage: I will check to see how or if it was tested before it was removed and reintroduce the test. If it wasn't tested before then I will write a new test to cover it.

jwage:
> Maybe yes, I wonder what other maintainers think.

@beberlei what do you think? Can it be as simple as just backing out the deprecation for this from 2.x?

greg0ire:
Gave it a little more thought, and I think a good solution might  be to:
1. remove existing deprecations
2. add a deprecation here, if `func_num_args()` is 1. The deprecation should warn that the argument is going to become mandatory in 3.0.0: https://github.com/doctrine/orm/blob/85034699cb29702a551ef33aecc76a9afd57389b/lib/Doctrine/ORM/ORMSetup.php#L125
3. In `DoctrineBundle`, remove the default value for the config node, and in the DI extension, pass the argument to the service only if specified in the configuration. If not specified, maybe trigger another deprecation saying that the configuration node is going to become mandatory in the next major version of the bundle?
4. Change this PR so that the `ORMSetup` argument mentioned in point 2. becomes mandatory.
5. In the Symfony recipe, set the configuration node to `true`.

That way, it's fine to disable deprecations, it just has to be done explicitly.

src/Mapping/Driver/XmlDriver.php
jwage: Looks like it was not tested before as far as I can tell. I added a test for enabling/disabling xml validation that should take care of it!

tests/Tests/ORM/Mapping/XmlMappingDriverTest.php
```diff
@@ -310,6 +311,39 @@ public function testDisablingXmlValidationIsPossible(): void
             false,
         );
     }
+
+    public function testXmlValidationEnabled(): void
+    {
+        self::expectException(MappingException::class);
+        self::expectExceptionMessage("libxml error: Element '{http://doctrine-project.org/schemas/orm/doctrine-mapping}field', attribute 'invalid': The attribute 'invalid' is not allowed.");
```
greg0ire: Please move both lines right before the line supposed to trigger them, it's a best practice.

jwage: Moved

SenseException (approved):

greg0ire:
Thanks @jwage !

PR #11166, opened 2024-01-17
Merge 2.18.x up into 3.0.x

greg0ire:

PR #11167, opened 2024-01-18
Use foreach on iterable to prevent table locks during tests

bobvandevijver:
Should fix table lock issues when merging into 3.x (see #11166). The iterator was not marked as finished, as this particular test case does not clear it records between the test, which meant that the test I added actually has 5 owners in the iterator.

By cleaning on setup and using a foreach which will yield only a single record, the iterator is closed and the table lock is released.

I am open for improvements to ensure the table contents is cleared between runs, but for now this works.


greg0ire (approved):

greg0ire:
Thanks @bobvandevijver !

Issue #11173, opened 2024-01-19
Calling method on null when non-backed Enums are used


thePanz:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.17.3

#### Summary

When non-backed enums are used, an error occurs: `Call to a member function getName() on null`.

I know that **only backed enums** are supported, but guiding the developer to fix the issue in an error message should be implemented.

#### Current behavior

Running `bin/console doctrine:mapping:info` throws

```
In DefaultTypedFieldMapper.php line 67:
                                               
  [Error]                                      
  Call to a member function getName() on null                                             

Exception trace:
  at /www/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php:67
 Doctrine\ORM\Mapping\DefaultTypedFieldMapper->validateAndComplete() at /www/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php:1594
 Doctrine\ORM\Mapping\ClassMetadataInfo->validateAndCompleteTypedFieldMapping() at /www/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php:1644
 Doctrine\ORM\Mapping\ClassMetadataInfo->validateAndCompleteFieldMapping() at /www/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php:2727
[...]
```

#### How to reproduce

Use the following classes. Notes:
- the Enum is not backed
- there is no "type" defined for the column

```
enum Suit {
    case Hearts;
    case Diamonds;
    case Clubs;
    case Spades;
}

#[Entity]
class Card
{
    /** ... */

    #[Column(enumType: Suit::class)]
    public Suit $suit;
}
```

#### Expected behavior

An error message about using a Backed enum should be reported for better DX

I would suggest some changes as to throw an exception in case the enum is not backed, stating which enum is used and in which class and property the error occurred.

```
    public function validateAndComplete(array $mapping, ReflectionProperty $field): array
    {
        $type = $field->getType();

        if (
            ! isset($mapping['type'])
            && ($type instanceof ReflectionNamedType)
        ) {
            if (PHP_VERSION_ID >= 80100 && ! $type->isBuiltin() && enum_exists($type->getName())) {
                $mapping['enumType'] = $type->getName();

                $reflection = new ReflectionEnum($type->getName());
[...]

PR #11176, opened 2024-01-19
Throw exception when trying to use non-backed enum types

thePanz:
This solves #11173

### DIFF
--- a/lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
+++ b/lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
@@ -57,7 +57,15 @@ final class DefaultTypedFieldMapper implements TypedFieldMapper
                 $mapping['enumType'] = $type->getName();
 
                 $reflection = new ReflectionEnum($type->getName());
-                $type       = $reflection->getBackingType();
+                if (! $reflection->isBacked()) {
+                    throw MappingException::backedEnumTypeRequired(
+                        $field->class,
+                        $mapping['fieldName'],
+                        $mapping['enumType']
+                    );
+                }
+
+                $type = $reflection->getBackingType();
 
                 assert($type instanceof ReflectionNamedType);
             }
--- a/tests/Doctrine/Tests/Models/Enums/FaultySwitch.php
+++ b/tests/Doctrine/Tests/Models/Enums/FaultySwitch.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Models\Enums;
+
+use Doctrine\ORM\Mapping\Column;
+
+class FaultySwitch
+{
+    #[Column(type: 'string')]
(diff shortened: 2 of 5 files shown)


greg0ire:
It seems there are several CI jobs failing. Please take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools) for more on how to handle those.

thePanz:
@greg0ire Pushed some more fixes.

I did not run phpunit still as I do not have all the required DB backends tho

greg0ire:
You only need SQLite to run most tests, and you need nothing at all to run the tests you wrote.

lib/Doctrine/ORM/Mapping/DefaultTypedFieldMapper.php
```diff
@@ -57,7 +57,17 @@ public function validateAndComplete(array $mapping, ReflectionProperty $field):
                 $mapping['enumType'] = $type->getName();
 
                 $reflection = new ReflectionEnum($type->getName());
-                $type       = $reflection->getBackingType();
+                if (
+                    $reflection->isBacked() === false
+                ) {
```
greg0ire:
```suggestion
                if(! $reflection->isBacked()) {
```

tests/Doctrine/Tests/ORM/Mapping/TypedEnumFieldMapperTest.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        return new DefaultTypedFieldMapper();
+    }
+
+    public function testNotBackedEnumThrows(): void
+    {
+        $reflectionClass = new ReflectionClass(FaultySwitch::class);
+
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessage(sprintf(
```
greg0ire: I don't think you should use `sprintf` here. Please show the final exception message, as the user will see it instead.

tests/Doctrine/Tests/Models/Enums/FaultySwitch.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    #[Column(type: 'string')]
+    public string $value;
+
+    /**
+     * The following line is ignored on psalm and phpstan so that we can test
+     * that the mapping is throwing an exception when a non-backed enum is used.
+     *
+     * @psalm-suppress InvalidArgument
+     */
```
greg0ire, line 19: Ah so this means we already have static analysis for this kind of issue. Good. I suppose it cannot hurt to also validate it at runtime.

thePanz: yes, the library is covered by phpstan and psalm
Better having a runtime check as 3rd party users might not have the static analysis in place, or the static-analysis did not run yet (like my case) :)

greg0ire (approved):

SenseException (approved):

PR #11179, opened 2024-01-20
Modernize code in documentation

greg0ire:
Somehow, there still were code samples relying on annotations.


SenseException:
Maybe because ORM 2.17 still works with PHP 7.1?

derrabus (approved):

PR #11180, opened 2024-01-21
Allow doctrine/lexer 3

greg0ire:


SenseException (approved):

derrabus:
Why haven't we done this earlier? 🤔

greg0ire:
I asked myself the same question, and I have no clear answer for now… but I think it will make it harder to upgrade to ORM 3.0 if we don't allow it, so it definitely seems logical. I think maybe at some point we had `1 || 2` and felt satisfied with it, when the goal should have been `2 || 3`.


EDIT: rather, we couldn't have `1 || 2 || 3`, so went with just `2` in https://github.com/doctrine/orm/pull/10329 as a backport of what was done in 3.0.x, and then forgot to add the `|| 3` in a follow-up PR / didn't realise it was possible.

Issue #11163, opened 2024-01-17
orderBy not being respected


brian-codes:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 2.17.2

#### Summary

I have a relationship between Threads and Messages. One thread can contain many Messages. Using XML mapping there is an orderBy entry stating the collection of messages should be ordered by their created at date. This works correctly on v2.16.3 but open upgrading to 2.17.2 this is no longer respected and the collection of messages is ordered by it's UUID.

#### Current behavior

orderBy entry not being respected and instead sorting the collection by UUID

#### How to reproduce

Use QueryBuilder to get a collection of Threads.

```
    <entity repository-class="App\ThreadRepository" name="App\Thread" table="threads">
        <one-to-many field="messages" target-entity="App\Message" mapped-by="thread" fetch="EAGER">
            <order-by>
                <order-by-field name="createdAt"/>
            </order-by>
        </one-to-many>
    </entity>
```

```
    <entity repository-class="App\MessageRepository" name="App\Message" table="messages">
        <field name="createdAt" type="datetime" column="created_at"/>
        <many-to-one field="thread" target-entity="App\Thread" inversed-by="messages"/>
    </entity>
```

```
    $qb = $this->getEntityManager()->createQueryBuilder();

    $qb
    ->select('thread')
    ->from(Thread::class, 'thread');

    return $qb->getQuery()->getResult();
```

#### Expected behavior

orderBy entry to be respected and order the collection by the stated field.

mpdude:
Could you please check whether it is specific for the XML mapping and/or using fetch=EAGER?

grossmannmartin:
the same problem occurs with annotation mapping (learned the hard way today 😆 )

greg0ire:
@grossmannmartin are you using `fetch=EAGER`?

grossmannmartin:
Yes, Yes I was. 
In the end, I rewrote it without EAGER and that resolved my issue.

greg0ire:
Then it is probably caused by https://github.com/doctrine/orm/pull/8391, Cc @beberlei

Alvaro948:
Tested con 2.18.0, does not work:

```php
#[ORM\OneToMany(mappedBy: 'parent', targetEntity: VariableEntity::class, fetch: 'EAGER')]
#[ORM\OrderBy(value: ['name' => 'ASC'])]
private Collection $variables;
```

[...]

Issue #11149, opened 2024-01-02
Doctrine ORM 2.17.x Update Causes TypeError with indexBy on Doctrine Type (custom UUID)


mkarolczyk:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.17.x

#### Summary
After updating doctrine/orm to version 2.17.x, the index-by for doctrine type stopped working even though the object (custom UUID) has the __toString() method. Up to version 2.16.x everything worked fine.

`TypeError: Doctrine\Common\Collections\ArrayCollection::set(): Argument #1 ($key) must be of type string|int, App\Common\Domain\Uuid given, called in /home/xxx/vendor/doctrine/orm/lib/Doctrine/ORM/PersistentCollection.php on line 191`

#### Previous behavior

Until version 2.16.x, indexBy for doctrine type objects (custom UUID) with the __toString() method worked fine.

```xml
<one-to-many field="subaccountRecipients" target-entity="App\Message\Preparation\Domain\SubaccountRecipient" mapped-by="message" fetch="EAGER" orphan-removal="true" index-by="subaccountRecipientId">
            <cascade>
                <cascade-persist />
                <cascade-remove />
            </cascade>
        </one-to-many>
```

#### Current behavior

Currently, Doctrine is unable to read such an object and shows an error. Before creating an ArrayCollection, there is no cast to a simple type, in this case string.

`TypeError: Doctrine\Common\Collections\ArrayCollection::set(): Argument #1 ($key) must be of type string|int, App\Common\Domain\Uuid given, called in /home/xxx/vendor/doctrine/orm/lib/Doctrine/ORM/PersistentCollection.php on line 191`

#### How to reproduce

- Have an existing Symfony application using Doctrine 2.17.x
- Add in the indexBy field, which is an object mapped by doctrine type
- Use findOneBy method

greg0ire:
@mkarolczyk hey :wave: 

I tried finding what may have caused this, and failed, so instead, I wrote [this blog post](https://dev.to/greg0ire/bisecting-vendors-12kd)  especially for you. Can you please try to apply what you read here to this issue, and let me know if anything is unclear? Thanks.

gordinskiy:
Got same error after trying to use [Optimized eager fetch](https://github.com/doctrine/orm/pull/8391) instead of [Multi-step hydration](https://ocramius.github.io/blog/doctrine-orm-optimization-hydration/)

[...]

Issue #11097, opened 2023-11-30
`OneToMany::indexBy` no longer works when refering a database column


aprat84:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.17.0

#### Summary

Indexing by database column worked, now it doesn't.
Better look at previous/current behavior.

#### Previous behavior

```php
class Language
{
    #[ORM\Id]
    #[ORM\Column(type: 'string', length: 20)]
    private ?string $id = null;
}

class Product {
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column(type: 'integer', options: ['unsigned' => true])]
    private ?int $id = null;

    #[ORM\OneToMany(
        mappedBy: 'product',
        targetEntity: ProductTranslation::class,
        cascade: ['persist', 'merge'],
        fetch: 'EAGER',
        indexBy: 'language_id'
    )]
    private Collection $translations;
}

class ProductTranslation {
    #[ORM\Id]
    #[ORM\ManyToOne(targetEntity: Product::class, inversedBy: 'translations')]
    #[ORM\JoinColumn(nullable: false, onDelete: 'CASCADE')]
    private Product $product;

    #[ORM\Id]
    #[ORM\ManyToOne(targetEntity: Language::class)]
    #[ORM\JoinColumn(nullable: false)]
    private Language $language;
}
```

This worked, indexing `Product::$translations` by `language_id` (`Language::$id`).

#### Current behavior

I get a warning and an error:

```txt
Warning: Undefined array key "language_id"
vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php:846
```

```txt
Error: Call to a member function getValue() on null
vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:3195
```

#### How to reproduce

I've created a simple symfony project: https://github.com/aprat84/doctrine-2.17-bug

- Clone it
- Execute migrations
- Execute tests in each commit (one for 2.16.3 and one for 2.17.1)

aprat84:
Added a repo with simple project to reproduce the bug:
https://github.com/aprat84/doctrine-2.17-bug

igoooor:
Have you found a solution or a workaround to this? I have the same issue and don't know how to proceed

aprat84:
This is fixed by #11380

PR #11185, opened 2024-01-25
Eagerly loaded indexed associations use hydrated values instead of database values

wmouwen:
### Description
**Update:** bug appears to exists in `3.0.x` as well

There is a breaking change between `2.16.x` and `2.17.x` where for eagerly loaded collections `indexBy` started looking at the hydrated set of values instead of the raw values returned from the database. This broke setups for all eager associations using either non-scalar fields (Uuid, other entities) or associations where the inverse field's database name didn't match the property name.

### Affected versions
- `2.17`
- `2.18`
- `2.19`
- `3.0`
- `3.1`

### Caused by
- #8391

### Issues
- #11149
- #11097

### Related
- #11163
- #11203

### DIFF
--- a/tests/Tests/ORM/Functional/Ticket/GH11149/EagerProduct.php
+++ b/tests/Tests/ORM/Functional/Ticket/GH11149/EagerProduct.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket\GH11149;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+
+#[ORM\Entity]
+#[ORM\Table('gh11149_eager_product')]
+class EagerProduct
+{
+    #[ORM\Id]
+    #[ORM\Column(type: 'integer')]
+    public int $id;
+
+    #[ORM\OneToMany(
+        targetEntity: EagerProductTranslation::class,
+        mappedBy: 'product',
+        fetch: 'EAGER',
+        indexBy: 'locale_code',
+    )]
+    public Collection $translations;
+
+    public function __construct(int $id)
+    {
+        $this->id           = $id;
+        $this->translations = new ArrayCollection();
+    }
+}
--- a/tests/Tests/ORM/Functional/Ticket/GH11149/EagerProductTranslation.php
+++ b/tests/Tests/ORM/Functional/Ticket/GH11149/EagerProductTranslation.php
@@ -7,20 +7,20 @@ namespace Doctrine\Tests\ORM\Functional\Ticket\GH11149;
 use Doctrine\ORM\Mapping as ORM;
 
 #[ORM\Entity]
-#[ORM\Table('gh11149_product_translation')]
-class ProductTranslation
+#[ORM\Table('gh11149_eager_product_translation')]
+class EagerProductTranslation
 {
     #[ORM\Id]
-    #[ORM\ManyToOne(targetEntity: Product::class, inversedBy: 'translations')]
+    #[ORM\ManyToOne(targetEntity: EagerProduct::class, inversedBy: 'translations')]
     #[ORM\JoinColumn(nullable: false)]
-    public Product $product;
+    public EagerProduct $product;
 
     #[ORM\Id]
     #[ORM\ManyToOne(targetEntity: Locale::class)]
     #[ORM\JoinColumn(name: 'locale_code', referencedColumnName: 'code', nullable: false)]
     public Locale $locale;
 
-    public function __construct(Product $product, Locale $locale)
+    public function __construct(EagerProduct $product, Locale $locale)
     {
         $this->product = $product;
         $this->locale  = $locale;
--- a/tests/Tests/ORM/Functional/Ticket/GH11149/GH11149Test.php
+++ b/tests/Tests/ORM/Functional/Ticket/GH11149/GH11149Test.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\ORM\Functional\Ticket\GH11149;
 
+use Doctrine\ORM\PersistentCollection;
 use Doctrine\Tests\OrmFunctionalTestCase;
 
 class GH11149Test extends OrmFunctionalTestCase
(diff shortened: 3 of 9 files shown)


wmouwen:
In order to load collections in a same way as other spots, I've chosen to reuse the existing method `\Doctrine\ORM\UnitOfWork::loadCollection`. This way the functionality of eager loaded collections is in line with both older versions (`<= 2.16.x`) and non-eager loads.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH11149Test.php
```diff
@@ -0,0 +1,112 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Tests\ORM\Functional\Ticket;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Tests\OrmFunctionalTestCase;
+
+class GH11149Test extends OrmFunctionalTestCase
```
greg0ire: To avoid issues with php < 8.1, you should use `@require PHP 8.1` on this class, and move at least the enum to a separate file. To keep things tidy, please put all this inside a `tests/Doctrine/Tests/ORM/Functional/Ticket/GH11149`.

wmouwen: Done.

tests/Doctrine/Tests/ORM/Functional/Ticket/GH11149/GH11149Test.php
```diff
@@ -0,0 +1,42 @@
...
+        $this->_em->persist(new ProductTranslation($product, $locale));
+
+        $this->_em->flush();
+        $this->_em->clear();
+
+        $product = $this->_em->find(Product::class, 11149);
+        static::assertInstanceOf(Product::class, $product);
+        static::assertCount(1, $product->translations);
+
+        $translation = $product->translations->get(LocaleCode::French->value);
```
greg0ire: Ah now we have another issue. The code inside the test file itself does not parse. Pretty sure we've fixed a similar issue before, I'll try to find the solution.

greg0ire: Ah well I see you've found a solution yourself, great! Otherwise, we have another way out: https://github.com/doctrine/persistence/commit/f5844651d4f1e23f7e2e4d035c7700564a1e4787

wmouwen: We can set the test file to be in a PHP 7 compatible syntax, we just need to avoid using the backed enum in this class. Luckily that was an easy fix.

greg0ire:
In case you want to run stuff locally, you can take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools).

wmouwen:
> In case you want to run stuff locally, you can take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools).

That's the fun part; it works on my machine, and the reason why the test EagerFetchCollectionTest has been altered. Apparently the values between my machine (PHP 8.3) and the failing GitHub actions (PHP 7.2) differ, which isn't weird considering that the eager fetch logic has been altered. Guess the next step is to have a good look at whether EagerFetchCollectionTest is still valid.

wmouwen:
Forcing a run on PHP 7.2.34 with the same settings as the failing GitHub action ([link](https://github.com/doctrine/orm/actions/runs/7660992767/job/20879563088)), still passes the tests.

```
> docker run -it -v "$PWD":/app -w /app --rm php:7.2-cli vendor/bin/phpunit -c ci/github/phpunit/pdo_sqlite.xml --exclude-group performance,non-cacheable,locking_functional tests/Doctrine/Tests/ORM/Functional/EagerFetchCollectionTest.php
PHPUnit 8.5.35 by Sebastian Bergmann and contributors.


Using DB driver Doctrine\DBAL\Driver\PDO\SQLite\Driver
....                                                                4 / 4 (100%)

Time: 54 ms, Memory: 12.00 MB

OK (4 tests, 12 assertions)
```

I'm a bit clueless as to what is causing the difference in behavior. I have chosen to disable parts of the broken tests, that way we can see if it is the only class to fail and narrow down the search.

greg0ire:
@wmouwen I just had a hunch that your issue was caused by the second level cache. If you look more closely at the failing action in your link above, you will notice that there's a successful "Run PHPUnit" step, followed by a failing "Run PHPUnit with Second Level Cache". You should be able to reproduce the issue with `ENABLE_SECOND_LEVEL_CACHE=1 vendor/bin/phpunit`.

wmouwen:
@greg0ire I'm failing to find a way to get the raw database values of the correct `indexBy` column in `\Doctrine\ORM\UnitOfWork::eagerLoadCollections`.

Secondly, #11163 has been created for a different problem caused by the same PR #8391. As #8391 was an optimization, I would suggest reverting #8391 for now until a new optimization can be written without breaking changes.

cc @beberlei

wmouwen:
Updated the test for `3.0.x` and rebased. Still couldn't find a solution except for reverting the original changes and incurring a performance loss, help is welcome.

greg0ire:
@mpdude , maybe you can help with this?

wmouwen:
The below code fetches entities for associations of multiple entities at once.

https://github.com/doctrine/orm/blob/b187bc85881cc85d36b28e7ed9dbe2071d472e30/src/UnitOfWork.php#L2659

We can't inject an 'indexBy' property here. Even if we could, the result would be ambiguous, as what is unique for one entity does not have to be unique across multiple entities.

The `->loadAll()` call hydrates the entities. This causes the variable `$indexByProperty` a few lines below to try and find a non-existent column.

https://github.com/doctrine/orm/blob/b187bc85881cc85d36b28e7ed9dbe2071d472e30/src/UnitOfWork.php#L2672-L2674

The column `locale_code` in the testcases doesn't exist on the hydrated object.

There is no clear way to combine `$targetClass` (`Doctrine\ORM\Mapping\ClassMetadata`) and `$mapping` (`Doctrine\ORM\Mapping\OneToManyAssociationMapping`) to get the value for indexing from the hydrated object (`$targetValue`) given the database column name.

@beberlei @mpdude Does either of you know a way?

beberlei:
Superseded by #11380

PR #11187, opened 2024-01-26
Remove XML validation disabling deprecation.

jwage:
This goes with this https://github.com/doctrine/orm/pull/11165

We are keeping this functionality for now so we are removing this deprecation.


greg0ire (approved):

greg0ire:
Thanks @jwage !

